PR #697, opened 2020-11-25
Ubuntu/bionic

TheRealFalcon:


blackboxsw (changes requested):
your Bionic  has changes beyond 47f4229ebcef9f83df8b549bb869a2dbf6dff17c
I think we need this to regenerate `new-upstream-snapshot --sru-bug 1905599 47f4229ebcef9f83df8b549bb869a2dbf6dff17c`

blackboxsw (approved):
Approved. only minor manual nit was to redact the LP: out of one of the commit message summaries that ended up in debian/changelog. Will apply the same manual to ubuntu/focal & xenial

PR #702, opened 2020-12-01
integration_tests: introduce skipping of tests by OS

OddBloke:
## Proposed Commit Message
```
integration_tests: introduce skipping of tests by OS

This introduces an optional, more complex OS_IMAGE format (`<image
id>::<os>::<release>`) which allows the specification of the OS/OS
release which the given image ID corresponds to.  This information is
used to skip tests which do not apply to the image.
 
This commit is comprised of the following discrete changes:
* introduce the IntegrationImage class, to handle parsing and storing
   the new OS_IMAGE format
* support inferring the OS and OS release of Ubuntu series, so that we
  can continue to set OS_IMAGE to just a series name and have test
  skipping work
* add documentation on Image Selection to integration_tests.rst
* introduce the actual skipping behaviour based on OS marks
* apply the `ubuntu` mark to all tests that should be skipped on
  non-Ubuntu operating systems
```

## Additional Context

See the added "Image Selection" doc section for an overview.

## Test Steps

Travis should continue to pass, and it should be observed that tests are skipped on non-Ubuntu releases. This can be seen by using an Ubuntu image but lying about its origin with a full `OS_IMAGE` specification:

```
$ CLOUD_INIT_OS_IMAGE="ubuntu:focal::notubuntu::noreally" pytest --log-cli-level=INFO tests/integration_tests/ -rsXx
...
INFO     integration_testing:clouds.py:74 Detected image: image_id=ubuntu:focal os=notubuntu release=noreally
...
=== short test summary info ===
SKIPPED [12] tests/integration_tests/conftest.py:69: Cannot run on OS notubuntu
```

## Checklist:

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -6,7 +6,7 @@ from pycloudlib import EC2, GCE, Azure, OCI, LXDContainer, LXDVirtualMachine
 from pycloudlib.lxd.instance import LXDInstance
 
 import cloudinit
-from cloudinit.subp import subp
+from cloudinit.subp import subp, ProcessExecutionError
 from tests.integration_tests import integration_settings
 from tests.integration_tests.instances import (
     IntegrationEc2Instance,
@@ -23,6 +23,19 @@ except ImportError:
 
 
 log = logging.getLogger('integration_testing')
+
+
+def _get_ubuntu_series() -> list:
+    """Use distro-info-data's ubuntu.csv to get a list of Ubuntu series"""
+    out = ""
+    try:
+        out, _err = subp(["ubuntu-distro-info", "-a"])
+    except ProcessExecutionError:
+        log.info(
+            "ubuntu-distro-info (from the distro-info package) must be"
+            " installed to guess Ubuntu os/release"
+        )
+    return out.splitlines()
 
 
 class ImageSpecification:
@@ -48,6 +61,12 @@ class ImageSpecification:
         os: "Optional[str]" = None,
         release: "Optional[str]" = None,
     ):
+        if image_id in _get_ubuntu_series():
+            if os is None:
+                os = "ubuntu"
+            if release is None:
+                release = image_id
+
         self.image_id = image_id
         self.os = os
         self.release = release
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -25,6 +25,46 @@ except ImportError:
 log = logging.getLogger('integration_testing')
 
 
+class ImageSpecification:
+    """A specification of an image to launch for testing.
+
+    If either of ``os`` and ``release`` are not specified, an attempt will be
+    made to infer the correct values for these on instantiation.
+
+    :param image_id:
+        The image identifier used by the rest of the codebase to launch this
+        image.
+    :param os:
+        An optional string describing the operating system this image is for
+        (e.g.  "ubuntu", "rhel", "freebsd").
+    :param release:
+        A optional string describing the operating system release (e.g.
+        "focal", "8"; the exact values here will depend on the OS).
+    """
+
+    def __init__(
+        self,
+        image_id: str,
+        os: "Optional[str]" = None,
+        release: "Optional[str]" = None,
+    ):
+        self.image_id = image_id
+        self.os = os
+        self.release = release
+        log.info(
+            "Detected image: image_id=%s os=%s release=%s",
+            self.image_id,
+            self.os,
+            self.release,
+        )
+
+    @classmethod
+    def from_os_image(cls):
+        """Return an ImageSpecification for integration_settings.OS_IMAGE."""
+        parts = integration_settings.OS_IMAGE.split("::", 2)
+        return cls(*parts)
+
+
 class IntegrationCloud(ABC):
     datasource = None  # type: Optional[str]
     integration_instance_cls = IntegrationInstance
@@ -57,13 +97,11 @@ class IntegrationCloud(ABC):
         raise NotImplementedError
 
     def _get_initial_image(self):
-        _released_image_id = self.settings.OS_IMAGE
+        image = ImageSpecification.from_os_image()
         try:
-            _released_image_id = self.cloud_instance.released_image(
-                self.settings.OS_IMAGE)
+            return self.cloud_instance.released_image(image.image_id)
         except (ValueError, IndexError):
-            pass
-        return _released_image_id
+            return image.image_id
 
     def _perform_launch(self, launch_kwargs):
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -10,12 +10,13 @@ from pathlib import Path
 
 from tests.integration_tests import integration_settings
 from tests.integration_tests.clouds import (
+    AzureCloud,
     Ec2Cloud,
     GceCloud,
-    AzureCloud,
-    OciCloud,
+    ImageSpecification,
     LxdContainerCloud,
     LxdVmCloud,
+    OciCloud,
 )
 from tests.integration_tests.instances import IntegrationInstance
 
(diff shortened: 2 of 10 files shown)


tests/integration_tests/conftest.py
```diff
@@ -172,6 +172,26 @@ def _client(request, fixture_utils, session_cloud):
     launch_kwargs = {}
     if name is not None:
         launch_kwargs = {"name": name}
+    if (
```
TheRealFalcon, line 175: Do you think it makes sense to do image filtering/skipping in the same way and at the same time as we do the [platform filtering/skipping](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/conftest.py#L39)?

TheRealFalcon (commented):
Overall the idea makes sense and seems like a good way forward.

tests/integration_tests/clouds.py
```diff
@@ -25,14 +26,100 @@
...
+        A string describing the operating system release (e.g. "focal", "8";
+        the exact values here will depend on the OS).
+    """
+    def __init__(
+        self,
+        image_specification: str,
+        image_id: str,
+        os: Optional[str] = None,
+        release: Optional[str] = None
+    ) -> None:
```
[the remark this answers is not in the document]
OddBloke: Yep: https://github.com/psf/black/blob/master/docs/the_black_code_style.md#how-black-wraps-lines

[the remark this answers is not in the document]
OddBloke: (Well, I'm missing a trailing comma.)

tests/integration_tests/conftest.py
OddBloke: Yes, I think it probably does. That'll require some refactoring to drop the dependency on `session_cloud` from the current implementation; we don't have access to that in `pytest_runtest_setup`. I'll see what that looks like and report back.

OddBloke: One consequence that occurs to me: this will remove the ability to use the cloud to lookup the image's OS and release. LXD images from streams, for example, do have this metadata:

```
$ lxc image show ubuntu:ffae848ee5a0
auto_update: false
properties:
  architecture: amd64
  description: ubuntu 20.04 LTS amd64 (release) (20200529.1)
  label: release
  os: ubuntu
  release: focal
  serial: "20200529.1"
  type: root.tar.xz
  version: "20.04"
public: true
expires_at: 2025-04-23T00:00:00Z
profiles: []
```

I don't know that this is a particularly huge loss; I suspect that such reverse lookups will not be generically implementable anyway (on EC2, for example, we'd be parsing something like "ubuntu/images-testing/hvm-ssd/ubuntu-trusty-daily-amd64-server-20191107"), so the odds of them being used would be pretty low.

tests/integration_tests/clouds.py
```diff
@@ -25,14 +26,100 @@
...
+            reader = csv.DictReader(f)
+            return [row["series"] for row in reader]
+    except FileNotFoundError:
+        log.info(
+            "distro-info-data must be installed to guess Ubuntu os/release"
+        )
+    return []
+
+
+class IntegrationImage:
```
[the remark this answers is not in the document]
OddBloke: I think we'll be able to write the mark matching logic more succinctly if this class implements a `matches(os, release)` (or something broadly along those lines). Roughly:

```diff
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -75,6 +75,11 @@ class ImageSpecification:
             "Image tuple: (%s, %s, %s)", self.image_id, self.os, self.release
         )
 
+    def matches(self, os, release):
+        matches_os = self.os is None or self.os == os
+        matches_release = self.release is None or self.release == release
+        return matches_os and matches_release
+
     @classmethod
     def from_os_image(cls):
         parts = integration_settings.OS_IMAGE.split("::", 2)
diff --git a/tests/integration_tests/conftest.py b/tests/integration_tests/conftest.py
index fc52a8de..b482cfa2 100644
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -64,14 +64,7 @@ def pytest_runtest_setup(item):
     image = ImageSpecification.from_os_image()
     if "only_bionic" in test_marks:
         # TODO: implement generic matching logic
-
-        # Run Ubuntu tests if we either know this image is Ubuntu, or don't
-        # know what it is at all
-        run_ubuntu_tests = image.os is None or image.os == "ubuntu"
-        # Run bionic tests if we either know this image is bionic, or don't
-        # know what it is at all
-        run_bionic_tests = image.release is None or image.release == "bionic"
-        if not (run_ubuntu_tests and run_bionic_tests):
+        if not image.matches("ubuntu", "bionic"):
             pytest.skip("Only runs on Ubuntu bionic")
```

(And even if I didn't think the above: arguably a namedtuple with a function that constructs it is just a class without a namespace to put `__init__` in. :wink:)

tests/integration_tests/clouds.py
```diff
@@ -25,14 +26,100 @@
...
+
 class IntegrationCloud(ABC):
     datasource = None  # type: Optional[str]
     integration_instance_cls = IntegrationInstance
 
     def __init__(self, settings=integration_settings):
         self.settings = settings
         self.cloud_instance = self._get_cloud_instance()
-        self.image_id = self._get_initial_image()
+        self.image = self._get_initial_image()
```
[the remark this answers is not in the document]
OddBloke: Nope, we don't! And in the new draft we retain the `self.image_id = self._get_initial_image()` logic.

tests/integration_tests/clouds.py
```diff
@@ -57,13 +113,11 @@ def _get_cloud_instance(self):
...
-        _released_image_id = self.settings.OS_IMAGE
+        image = ImageSpecification.from_os_image()
         try:
-            _released_image_id = self.cloud_instance.released_image(
-                self.settings.OS_IMAGE)
+            return self.cloud_instance.released_image(image.image_id)
         except (ValueError, IndexError):
-            pass
-        return _released_image_id
+            return image.image_spec
```
TheRealFalcon, line 120: Is this supposed to be `image_id`? I don't see where `image_spec` comes from.

OddBloke:
On Tue, Dec 01, 2020 at 02:51:37PM -0800, James Falcon wrote:
> @TheRealFalcon commented on this pull request.
> 
> 
> 
> >          except (ValueError, IndexError):
> -            pass
> -        return _released_image_id
> +            return image.image_spec
> 
> Is this supposed to be `image_id`? I don't see where `image_spec` comes from.

Yep, this is what I get for renaming things right before pushing. :p

TheRealFalcon (approved):

tests/integration_tests/clouds.py
```diff
@@ -25,6 +26,65 @@
 log = logging.getLogger('integration_testing')
 
 
+def _get_ubuntu_series() -> list:
+    """Use distro-info-data's ubuntu.csv to get a list of Ubuntu series"""
+    try:
+        with open("/usr/share/distro-info/ubuntu.csv") as f:
```
blackboxsw, line 32: Why not avoid the csv import and use `distro-info -a` directly to simlify parsing?

tests/integration_tests/clouds.py
```diff
@@ -25,6 +26,65 @@
...
+            reader = csv.DictReader(f)
+            return [row["series"] for row in reader]
+    except FileNotFoundError:
+        log.info(
+            "distro-info-data must be installed to guess Ubuntu os/release"
```
blackboxsw, line 37: If you end up using distro-info directly, we'd change this
```suggestion
            "distro-info must be installed to guess Ubuntu os/release"
```

tests/integration_tests/clouds.py
OddBloke: Done, but with `ubuntu-distro-info` (else we'd guess Debian releases as "ubuntu:<release>" when the tests run on a Debian machine).

blackboxsw (approved):
LGTM!

PR #586, opened 2020-09-24
Change default authorizedkeys file

otubo:
## Proposed Commit Message
The following commit merged all ssh keys into a default user file
`~/.ssh/authorized_keys` in sshd_config had multiple files configured for
AuthorizedKeysFile:

commit f1094b1a539044c0193165a41501480de0f8df14
Author: Eduardo Otubo <otubo@redhat.com>
Date:   Thu Dec 5 17:37:35 2019 +0100

    Multiple file fix for AuthorizedKeysFile config (#60)

This commit ignored the case when sshd_config would have a single file for
AuthorizedKeysFile, but a non default configuration, for example
`~/.ssh/authorized_keys_foobar`. In this case cloud-init would grab all keys
from this file and write a new one, the default `~/.ssh/authorized_keys`
causing the bug.

rhbz: #1862967

Signed-off-by: Eduardo Otubo <otubo@redhat.com>

## Additional Context
None

## Test Steps
1. Prepare a VM with cloud-init enabled
2. Modify `/etc/ssh/sshd_config` to remove default `.ssh/authorized_keys` and change to another file e.g.:
```
# cat /etc/ssh/sshd_config | grep AuthorizedKeysFile
AuthorizedKeysFile .ssh/authorized_keys2
```
3. Remove /var/lib/cloud/instance/sem/config_ssh
4. systemctl restart cloud-init ; systemctl restart sshd
5. Try to ssh login through the cloud-init created user

Actual results:
Cannot login: `Permission denied (publickey,gssapi-keyex,gssapi-with-mic)`
The public key is written into `.ssh/authorized_keys` but not `.ssh/authorized_keys2`

Expected results:
Can login successfully. The public key is written into `.ssh/authorized_keys2`

## Checklist:
 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #706, opened 2020-12-02
Integration test for pull #586

TheRealFalcon:
## Proposed Commit Message
Integration test for pull #586

If a non-default AuthorizedKeysFile is specified in
/etc/ssh/sshd_config,  ensure we can still ssh as expected

## Additional Context
See https://github.com/canonical/cloud-init/pull/586

## Test Steps
Until SRU ends...
Run `pytest -s tests/integration_tests/bugs/test_gh586.py` to reproduce issue.
Run `CLOUD_INIT_CLOUD_INIT_SOURCE=PROPOSED pytest -s tests/integration_tests/bugs/test_gh586.py` to verify.

## Checklist:


TheRealFalcon:
Let's see if github complains about us storing keys in a repo :wink:

tests/integration_tests/assets/__init__.py
```diff
@@ -0,0 +1,16 @@
+import os
+from collections import namedtuple
+from pathlib import Path
+
+ASSET_DIR = Path(os.path.dirname(os.path.realpath(__file__)))
+PRIVATE_KEY_PATH = ASSET_DIR / 'test_id_rsa'
+PUBLIC_KEY_PATH = ASSET_DIR / 'test_id_rsa.pub'
+
+
+def get_test_keypair():
```
OddBloke, line 10: Do we want to make this interface `get_asset(asset_id: str) -> Any` to discourage ourselves from littering this with inconsistent utility functions in future? (The implementation doesn't need to be generic; `if asset_id != "test_keypair": raise Exception("Unknown asset")` seems like a fine placeholder for us to fix when we need to.)

(Perhaps this would be better as a fixture, to enforce a consistent single "entrypoint"; that wouldn't work at module-level for USER_DATA mark content, however, and I think that's a puzzle to be solved another time.)

tests/integration_tests/bugs/test_gh586.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@pytest.mark.sru_2020_11
+@pytest.mark.user_data(USER_DATA)
+def test_non_default_authorized_keys(client):
+    sshd = client.read_from_file('/etc/ssh/sshd_config')
+    assert 'AuthorizedKeysFile .ssh/authorized_keys2' in sshd
+
+    ssh_dir = client.execute('ls /home/ubuntu/.ssh').stdout
+    assert 'authorized_keys2' in ssh_dir
```
OddBloke, line 29: Should we assert the absence of the regular config line/files here?

tests/integration_tests/bugs/test_gh586.py
```diff
@@ -0,0 +1,37 @@
...
+
+    ssh = paramiko.SSHClient()
+    ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    paramiko_key = paramiko.RSAKey.from_private_key(StringIO(private_key))
```
OddBloke, line 33: As we need it to be an RSA key for this test, let's include the fact that it's an RSA key in the name of the asset: we might save someone from changing it to a DSA key for their test and then having to debug this failure.

tests/integration_tests/bugs/test_gh586.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Will fail with paramiko.ssh_exception.AuthenticationException
+    # if this bug isn't fixed
+    ssh.connect(client.instance.ip, username='ubuntu', pkey=paramiko_key)
```
OddBloke, line 37: It would be good to not hardcode "ubuntu" here; this test could (and should, as it was reported on a non-Ubuntu system) run on non-Ubuntu systems; I think `client.instance.username` would work (for now, at least)?

tests/integration_tests/assets/__init__.py
TheRealFalcon: Hmmmm, I think I'd rather have 20 random utility functions than 20 random strings in which we check which asset is tied to which string. Even if we had something like an Enum, that's essentially just adding a level of indirection to get to the private function doing what could be done with a public function. Is there a reason you want to restrict the number of entry points?

tests/integration_tests/bugs/test_gh586.py
TheRealFalcon: I'm adding a check for sshd_config, but the original authorized_keys file actually still exists from the image snapshot. It doesn't affect this test, but I wonder if that means we should we doing more than just a `cloud-init clean --logs` when creating an image snapshot.

tests/integration_tests/assets/__init__.py
OddBloke: > Hmmmm, I think I'd rather have 20 random utility functions than 20 random strings in which we check which asset is tied to which string.

I don't think these are equivalent: utility functions can have completely different signatures to one another, whereas strings are flat identifiers. I think that's the core of my concern: if these can take parameters then they can behave in substantially different ways to one another.

> Even if we had something like an Enum, that's essentially just adding a level of indirection to get to the private function doing what could be done with a public function. Is there a reason you want to restrict the number of entry points?

Experience tells me that we'll end up with test logic being implemented in functions in this file, which means (a) our tests are harder to read/write because they're more spread out, and (b) the functions won't be (easily) usable outside of the specific test they were written for (because they'll be written with whatever that test author sees as the "assets interface" at that moment in time).

That said, the `-> Any` in my proposal isn't ideal; it makes it impossible to reason about what `get_asset` would return without going and reading the assets code.

You aren't introducing an issue here, so we don't need to take care of this to get this in; let's be mindful of such issues as we introduce more assets in future but for now: YAGNI (WAGNI?).

tests/integration_tests/bugs/test_gh586.py
OddBloke: Yeah, I think we probably should be. Not only do we have that file still around, we also have the default user already created. (When we decide what else we want to do, we should consider expanding `clean` to perform it for us so it isn't specific to the tests.)

This is what the cloud_tests do, so not much to be learned from that: https://github.com/canonical/cloud-init/blob/bd76d5cfbde9c0801cefa851db82887e0bab34c1/tests/cloud_tests/releases.yaml#L14-L17

tests/integration_tests/bugs/test_gh586.py
```diff
@@ -0,0 +1,42 @@
...
+If a non-default AuthorizedKeysFile is specified in /etc/ssh/sshd_config,
+ensure we can still ssh as expected.
+"""
+import paramiko
+import pytest
+from io import StringIO
+from tests.integration_tests.assets import get_test_rsa_keypair
+
+
+public_rsa_key, private_rsa_key = get_test_rsa_keypair()
```
OddBloke, line 12: Capitalise these?

TheRealFalcon: Capitalize RSA or treat the variables as constants? If just RSA, I don't think it's typical to capitalize abbreviations in normal variables names. If treat like constants, the contents are derived from reading a file, so I'd say their immutable, not constant, and thus should remain lowercase.

OddBloke: I'm happy to leave it as-is. I've never really considered all caps to mean constant, so much as it means "module-level" but I guess I'll take you agreeing with PEP-8 as a victory. :wink:

TheRealFalcon: *ahem*
image

OddBloke: > can reach agreement on this issue

This sort of passive aggressive behaviour is very self-defeating. :grin:

TheRealFalcon: Haha, all in good fun (at least for me :wink: )!

OddBloke (approved):
Thanks!

tests/integration_tests/bugs/test_gh586.py
OddBloke: Absolutely!

PR #709, opened 2020-12-03
Network config generation exception

trstringer:
## Proposed Commit Message

Prevent network interfaces without IP addresses from being added to the generated network configuration.

## Additional Context

There are situations where a network interface might not have any associated IP addresses with it (this is by-design). In this case, the network configuration that is generated would currently still include these network interfaces. This could cause issues and delays in provisioning, for example if `systemd-networkd-wait-online` is running it would be waiting until timeout because any interface without an IP address would not ever come online.

This PR adds a check to make sure that we are only generating network configuration for devices with IP addresses.

## Checklist:


anhvoms:
Can you add a unit test to cover the case where if ipaddress isn't there, the network config for the interface would not be generated?

OddBloke (approved):
This looks good to me, thanks!

trstringer:
Thanks for the review and approval, @OddBloke!

PR #710, opened 2020-12-03
Add wait-for-disk functionality to disk_setup (#1832645)

nibalizer:
## Proposed Commit Message

Add wait-for-disk functionality to disk_setup

In a cloud environment, sometimes disks will attach while cloud-init is
running and get missed. This adds a configurable timeout to wait for
those disks.

LP: #1832645

> summary: no more than 70 characters
>
> A description of what the change being made is and why it is being
> made, if the summary line is insufficient.  The blank line above is
> required. This should be wrapped at 72 characters, but otherwise has
> no particular length requirements.
>
> If you need to write multiple paragraphs, feel free.
>
> LP: #NNNNNNN (replace with the appropriate bug reference or remove
> this line entirely if there is no associated bug)

## Checklist:

### DIFF
--- a/cloudinit/config/cc_disk_setup.py
+++ b/cloudinit/config/cc_disk_setup.py
@@ -95,6 +95,7 @@ specified using ``filesystem``.
           partition: <"auto"/"any"/"none"/<partition number>>
           overwrite: <true/false>
           replace_fs: <filesystem type>
+          timeout: <seconds>
 """
 
 from cloudinit.settings import PER_INSTANCE
@@ -133,7 +134,13 @@ def handle(_name, cfg, cloud, log, _args):
             if not isinstance(definition, dict):
                 log.warning("Invalid disk definition for %s" % disk)
                 continue
-
+            if definition.get("timeout"):
+                tmout = int(definition.get("timeout"))
+                need = util.wait_for_files([disk],
+                                           maxwait=tmout,
+                                           naplen=0.1)
+                if len(need) > 0:
+                    log.warning("Timeout expired waiting for %s" % disk)
             try:
                 log.debug("Creating new partition table/disk")
                 util.log_time(logfunc=LOG.debug,
--- a/doc/examples/cloud-config-disk-setup.txt
+++ b/doc/examples/cloud-config-disk-setup.txt
@@ -98,6 +98,7 @@ disk_setup:
 #       table_type: 'mbr'
 #       layout: <LAYOUT|BOOL>
 #       overwrite: <BOOL>
+#       timeout: <INT>
 #
 # Where:
 #   <DEVICE>: The name of the device. 'ephemeralX' and 'swap' are special
@@ -159,6 +160,10 @@ disk_setup:
 #               'true' is cowboy mode. There are no checks and things are
 #                   done blindly. USE with caution, you can do things you
 #                   really, really don't want to do.
+#
+#   timeout=<INT>: Seconds to wait for a disk to appear. If omitted, no wait
+#               is performed. This is useful if a cloud provider has late or
+#               inconsistent disk attachments and you need a delay.
 #
 #
 # fs_setup: Setup the file system


nibalizer:
This is a work in progress

cloudinit/config/cc_disk_setup.py
```diff
@@ -827,6 +829,36 @@ def mkpart(device, definition):
     LOG.debug("Partition table created for %s", device)
 
 
+def block_until_file_exists(path, timeout=30):
```
smoser, line 832: is this necessarily different from utils.wait_for_files ?

cloudinit/config/cc_disk_setup.py
```diff
@@ -133,8 +134,9 @@ def handle(_name, cfg, cloud, log, _args):
             if not isinstance(definition, dict):
                 log.warning("Invalid disk definition for %s" % disk)
                 continue
-
             try:
+                if definition.get("timeout"):
+                    block_until_file_exists(disk, definition.timeout())
```
smoser, line 139: are you *calling* timeout ?
i think you would want `block_until_file_exists(disk, definition.get('timeout'))`

```python

if definition.get("timeout"):
    utils.wait_for_files([disk], definition.get('timeout'), naplen=1)
```

smoser (changes requested):

cloudinit/config/cc_disk_setup.py
nibalizer: Yes I think you're correct whoops

cloudinit/config/cc_disk_setup.py
nibalizer: I just wasn't aware of that function, I'll use it instead

cloudinit/config/cc_disk_setup.py
```diff
@@ -133,8 +133,11 @@ def handle(_name, cfg, cloud, log, _args):
             if not isinstance(definition, dict):
                 log.warning("Invalid disk definition for %s" % disk)
                 continue
-
             try:
+                if definition.get("timeout"):
```
OddBloke, line 137: I suspect `definition.get("timeout")` is a string here, and `wait_for_files` treats its `maxwait` parameter as a number. Do we need to convert this to an `int`?

OddBloke, line 137: The docstring at the top of this file is used to generate the documentation for this module; it should also be updated.

cloudinit/config/cc_disk_setup.py
```diff
@@ -133,8 +133,11 @@ def handle(_name, cfg, cloud, log, _args):
...
+                    util.wait_for_files([disk],
```
OddBloke, line 138: Should this move outside the `try` block here? This call failing would not really be a "Failed partitioning operation", and perhaps should not be ignored in the same way.

OddBloke (commented):
Thanks for this branch @nibalizer! Some comments inline.

cloudinit/config/cc_disk_setup.py
nibalizer: Good catch thanks

nibalizer: No problem, thanks

cloudinit/config/cc_disk_setup.py
nibalizer: I took a whack at this, let me know if you think this works better.

HeroCC:
Could this be reopened? This feature would be quite useful.

TheRealFalcon:
Github won't let us reopen this one because the base branch has changed, but we can certainly open a new PR referencing this one. @HeroCC or @nibalizer , are you interested in pushing this across the finish line? I think what we have currently looks good, but we'd need to also add/modify some unit testing before we'd be ready to merge it.

dirien:
Yeah stumled also on this on Equinix Metal... Would be nice to get this activated again...

PR #711, opened 2020-12-04
Add ability to keep snapshotted images in integration tests

TheRealFalcon:
## Proposed Commit Message
Add ability to keep snapshotted images in integration tests

For SRU test development, every single time we start a new test run,
we need to first install the PROPOSED version and create an image
snapshot. Instead of automatically deleting a snapshot, add an
integration setting to allow us to keep the snapshot. The end of the
test run will log the image name which can then be used as the
OS_IMAGE in subsequent test runs.

## Test Steps
n/a

## Checklist:


blackboxsw (approved):
LGTM!

PR #570, opened 2020-09-08
Nocloud/OVF seedfrom vendordata

jqueuniet:
This PR modifies the `util.seedfrom` helper function to query optional vendor-data. I started with the idea of implementing for NoCloud, but as OVF is using the same helper, I figured it would be easier to do it for both.

PR #712, opened 2020-12-04
Integration test for 570

TheRealFalcon:
## Proposed Commit Message
Integration test for #570

Test that we can add optional vendor-data to the seedfrom file in a
NoCloud environment.

Also added the option to pass `raise_on_cloudinit_failure` through
an instance restart so we get automatic failure checking when
we need to manually reboot.

## Additional Context
#570

## Test Steps
should fail:
pytest tests/integration_tests/bugs/test_gh570.py

should pass:
CLOUD_INIT_CLOUD_INIT_SOURCE=PROPOSED pytest tests/integration_tests/bugs/test_gh570.py

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_gh570.py
+++ b/tests/integration_tests/bugs/test_gh570.py
@@ -34,5 +34,5 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
         VENDOR_DATA,
     )
     client.execute('cloud-init clean --logs')
-    client.restart()
+    client.restart(raise_on_cloudinit_failure=True)
     assert 'seeded_vendordata_test_file' in client.execute('ls /var/tmp')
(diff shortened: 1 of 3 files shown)


tests/integration_tests/bugs/test_gh570.py
```diff
@@ -0,0 +1,38 @@
...
+        "/var/lib/cloud/seed/nocloud-net/meta-data".format(seed_dir=seed_dir)
+    )
+    assert result.return_code == 0
+
+    client.write_to_file(
+        '{}/vendor-data'.format(seed_dir),
+        VENDOR_DATA,
+    )
+    client.execute('cloud-init clean --logs')
+    client.restart()
```
blackboxsw, line 37: Now that we [ignore restart errors from cloud-init in integration test instances](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/instances.py#L47). don't we need to either add a raise_errors=True param to IntegrationInstance.restart method or just call client.instance.restart() directly here which would wait and raise errors if we saw them across reboot?

blackboxsw (approved):

PR #714, opened 2020-12-07
Add support to resize rootfs if using LVM

otubo:
## Proposed Commit Message
This patch adds support to resize a single partition of a VM if it's using an
LVM underneath. The patch detects if it's LVM if the given block device
is a device mapper by its name (e.g. `/dev/dm-1`) and if it has slave
devices under it on `sysfs`. After that `syspath` is updated to the real
block device and `growpart` will be called to resize it (and automatically
its Physical Volume).

The Volume Group will be updated automatically and a final call to
extend the `rootfs` to the remaining space available will be made.

Using the same growpart configuration, the user can specify only one
device to be resized when using LVM and growpart, otherwise cloud-init
won't know which one should be resized and will fail.

rhbz: #1810878

## Additional Context
The commit also includes changes to the documentation as well as to unit tests to support the changes. The unit test is a copy from the existing test but using `/dev/XXdm-0` variants for LVM tests.

## Test Steps
One can verify the fix by installing cloud-init on an instance using LVM and configure it to expand one single partition, e.g. the `rootfs`.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -216,6 +216,7 @@ def device_part_info(devpath):
     rpath = os.path.realpath(devpath)
 
     bname = os.path.basename(rpath)
+    is_lvm = True if re.search('^dm.*', bname) else False
     syspath = "/sys/class/block/%s" % bname
 
     # FreeBSD doesn't know of sysfs so just get everything we need from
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_growpart.py
```diff
@@ -216,18 +216,25 @@ def device_part_info(devpath):
     rpath = os.path.realpath(devpath)
 
     bname = os.path.basename(rpath)
+    is_lvm = True if re.search('^dm.*', bname) else False
```
igalic, line 219: maybe this should be a function

igalic (commented):
🙋🏻‍♀️

otubo:
I'm gonna go ahead and close this PR because I've been testing and changing this code a lot lately. Will open a new one. Thanks for the reviews!

akkim31:
421ca235caf70d2e9f10c7e458b5f3a1b3aabf52

akkim31 (commented):
#714

PR #567, opened 2020-09-06
create a shutdown_command method in distro classes

emmanuelthome:
See https://www.freebsd.org/cgi/man.cgi?shutdown(8)

PR #717, opened 2020-12-08
Add integration test for power_state_change module

TheRealFalcon:
## Proposed Commit Message
Add integration test for power_state_change module

## Additional Context
Based on https://github.com/canonical/cloud-init/pull/567 , but not naming the test after it because the changes that affected Ubuntu were only a refactor.

In the test, I've included adding a timeout to the user_data, but I don't know of a way to actually test it because the power state module is the last module to run, so even with a timeout of 1, cloud-init finishes before the timeout is triggered.

## Test Steps
pytest tests/integration_tests/modules/test_power_state_change.py

## Checklist:

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -123,14 +123,13 @@ class IntegrationCloud(ABC):
             return image.image_id
 
     def _perform_launch(self, launch_kwargs):
-        wait = launch_kwargs.pop('wait', True)
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
-        if wait:
-            pycloudlib_instance.wait(raise_on_cloudinit_failure=False)
         return pycloudlib_instance
 
-    def launch(self, user_data=None, launch_kwargs=None,
+    def launch(self, user_data=None, launch_kwargs=None, wait=True,
                settings=integration_settings):
+        if launch_kwargs is None:
+            launch_kwargs = {}
         if self.settings.EXISTING_INSTANCE_ID:
             log.info(
                 'Not launching instance due to EXISTING_INSTANCE_ID. '
@@ -139,12 +138,15 @@ class IntegrationCloud(ABC):
                 self.settings.EXISTING_INSTANCE_ID
             )
             return
+        if 'wait' in launch_kwargs:
+            raise Exception("Specify 'wait' directly to launch, "
+                            "not in 'launch_kwargs'")
         kwargs = {
             'image_id': self.image_id,
             'user_data': user_data,
+            'wait': False,
         }
-        if launch_kwargs:
-            kwargs.update(launch_kwargs)
+        kwargs.update(launch_kwargs)
         log.info(
             "Launching instance with launch_kwargs:\n{}".format(
                 "\n".join("{}={}".format(*item) for item in kwargs.items())
@@ -152,7 +154,8 @@ class IntegrationCloud(ABC):
         )
 
         pycloudlib_instance = self._perform_launch(kwargs)
-
+        if wait:
+            pycloudlib_instance.wait(raise_on_cloudinit_failure=False)
         log.info('Launched instance: %s', pycloudlib_instance)
         return self.get_instance(pycloudlib_instance, settings)
 
@@ -259,7 +262,7 @@ class _LxdIntegrationCloud(IntegrationCloud):
 
     def _perform_launch(self, launch_kwargs):
         launch_kwargs['inst_type'] = launch_kwargs.pop('instance_type', None)
-        wait = launch_kwargs.pop('wait', True)
+        launch_kwargs.pop('wait')
         release = launch_kwargs.pop('image_id')
 
         try:
@@ -276,8 +279,6 @@ class _LxdIntegrationCloud(IntegrationCloud):
         if self.settings.CLOUD_INIT_SOURCE == 'IN_PLACE':
             self._mount_source(pycloudlib_instance)
         pycloudlib_instance.start(wait=False)
-        if wait:
-            pycloudlib_instance.wait(raise_on_cloudinit_failure=False)
         return pycloudlib_instance
 
 
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -123,8 +123,10 @@ class IntegrationCloud(ABC):
             return image.image_id
 
     def _perform_launch(self, launch_kwargs):
+        wait = launch_kwargs.pop('wait', True)
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
-        pycloudlib_instance.wait(raise_on_cloudinit_failure=False)
+        if wait:
+            pycloudlib_instance.wait(raise_on_cloudinit_failure=False)
         return pycloudlib_instance
 
     def launch(self, user_data=None, launch_kwargs=None,
@@ -140,7 +142,6 @@ class IntegrationCloud(ABC):
         kwargs = {
             'image_id': self.image_id,
             'user_data': user_data,
-            'wait': False,
         }
         if launch_kwargs:
             kwargs.update(launch_kwargs)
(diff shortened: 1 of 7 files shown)


tests/integration_tests/log_utils.py
```diff
@@ -0,0 +1,13 @@
+def ordered_items_in_text(to_verify: list, text: str):
```
TheRealFalcon, line 1: `log_utils.py` is better than `utils.py`, right???

But in all seriousness, I'm open to suggestions of a better place to put something like this.

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,79 @@
...
+@pytest.mark.sru_2020_11
+class TestPowerChange:
+    @pytest.mark.parametrize('mode,delay,timeout,expected', [
+        ('poweroff', 'now', '10', 'will execute: shutdown -P now msg'),
+        ('reboot', 'now', '0', 'will execute: shutdown -r now msg'),
+        ('halt', '+1', '0', 'will execute: shutdown -H +1 msg'),
+    ])
+    def test_poweroff(self, session_cloud: IntegrationCloud,
+                      mode, delay, timeout, expected):
+        with session_cloud.launch(
```
OddBloke, line 54: This is a nice example of the flexibility of the framework we have; good job!

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,79 @@
...
+            if mode == 'reboot':
+                _detect_reboot(instance)
+            else:
+                instance.instance.wait_for_stop()
+                instance.instance.start(wait=True)
+            log = instance.read_from_file('/var/log/cloud-init.log')
+            assert instance.running
+        lines_to_check = [
+            'Running module power-state-change',
+            expected.format(delay=delay),
```
OddBloke, line 68: Looks like all of the values `expected` can take are straight-up strings, so I think we can drop the `.format` here?

tests/integration_tests/log_utils.py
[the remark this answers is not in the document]
OddBloke: (And I suppose `s/list/Iterable[str]/` but let's not import typing just for that. ;))

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,79 @@
...
+
+Test that the power state config options work as expected.
+"""
+
+from tests.integration_tests.log_utils import ordered_items_in_text
+import pytest
+import time
+
+from tests.integration_tests.clouds import IntegrationCloud
+from tests.integration_tests.instances import IntegrationInstance
```
OddBloke, line 11: https://www.python.org/dev/peps/pep-0008/#imports :runner: :dash: 

(A nit.)

tests/integration_tests/clouds.py
```diff
@@ -123,8 +123,10 @@ def _get_initial_image(self):
             return image.image_id
 
     def _perform_launch(self, launch_kwargs):
+        wait = launch_kwargs.pop('wait', True)
```
OddBloke, line 126: If `wait` isn't in `launch_kwargs` then it will default to `True`, meaning that cloud-init failures will raise an exception. (More on this issue further down.)

OddBloke (commented):
Thanks James! The test itself LGTM, with a few minor nits/comments; the framework changes, however, I think may not be correct.

tests/integration_tests/instances.py
```diff
@@ -53,6 +53,14 @@ def __init__(self, cloud: 'IntegrationCloud', instance: BaseInstance,
         self.instance = instance
         self.settings = settings
 
+    @property
+    def running(self):
+        # TODO: Implement something better in pycloudlib
+        try:
+            return True if self.execute('true').return_code == 0 else False
```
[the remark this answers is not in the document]
TheRealFalcon: Bah! `True if ... else False` is a code peeve of mine and I just committed it! (but yes, I'll update this :wink: )

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,79 @@
...
+
+def _detect_reboot(instance: IntegrationInstance):
+    # We'll wait for instance up here, but we don't know if we're
+    # detecting the first boot or second boot, so we also check
+    # the logs to ensure we've booted twice. If the logs show we've
+    # only booted once, wait until we've booted twice
+    instance.instance.wait()
+    for _ in range(600):
+        try:
+            log = instance.read_from_file('/var/log/cloud-init.log')
```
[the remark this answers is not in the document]
TheRealFalcon: I agree for a general implementation, but in this specific case, the test code isn't initiating the reboot. Since we don't clear the reboot count when we generate the snapshot image, we can't definitively know how many reboots we should be looking for.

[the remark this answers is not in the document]
OddBloke: Very good point! You _could_ do something with `bootcmd` to determine it, but it would still be test-specific so there's really no reason to think it would be better.

OddBloke:
This isn't running successfully for me locally; I've seen this a few times (but not _every_ time):

```
E           OSError: Failed reading remote file via cat: /var/log/cloud-init.log
E           Return code: 1
E           Stderr: sudo: unknown user: ubuntu
E           sudo: unable to initialize policy plugin
E           Stdout:
```

But also:

```
>       assert ordered_items_in_text(lines_to_check, log)
E       assert False
E        +  where False = ordered_items_in_text(['Running module power-state-change', 'will execute: shutdown -r now msg', "running 'init-local'", 'config-power-state-change already ran'], "2020-12-08 22:30:56,727 - util.py[DEBUG]: Cloud-init v. 20.3-2-g371b392c-0ubuntu1~20.04.1 running 'init-local' at Tue...nds (0.12)\n2020-12-08 22:31:23,855 - handlers.py[DEBUG]: finish: init-local: SUCCESS: searching for local datasources")
```

(That assert could probably use a message, if you can figure out a good one to emit.)

OddBloke:
> This isn't running successfully for me locally

(Other than this, the code LGTM now, thanks!)

OddBloke:
> This isn't running successfully for me locally

(Let me know if there's anything I can do to help debug!)

TheRealFalcon:
@OddBloke I don't think it's worth investing anymore time in this test at this point. I get different weird errors (or just bugs in pycloudlib and lxd) from different clouds, and it's not worth the time fixing them now. I can either remove the test and run things manually, or keep the test around knowing its flaky and remove it from the normal test runs. I think there is value in keeping the test (hence my last commit), but I also understand not wanting to dirty our test base with flaky tests, so I'd be ok removing it too. What are your thoughts?

tests/integration_tests/conftest.py
```diff
@@ -71,6 +71,8 @@ def pytest_runtest_setup(item):
     supported_os_set = set(os_list).intersection(test_marks)
     if current_os and supported_os_set and current_os not in supported_os_set:
         pytest.skip("Cannot run on OS {}".format(current_os))
+    if 'unstable' in test_marks:
```
OddBloke, line 74: Is it worth adding `RUN_UNSTABLE` as a setting, so the interface for forcing a run is `CLOUD_INIT_RUN_UNSTABLE=true` rather than a code modification?

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,90 @@
+"""Integration test of the cc_power_state_change module.
+
+Test that the power state config options work as expected.
+"""
+
+import pytest
+import time
```
OddBloke, line 7: A nit, which I'll apply before merging:

```suggestion
import time

import pytest
```

(pytest is a third-party module.)

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _can_connect(instance):
+    return instance.execute('true').ok
+
+
+# This test is marked unstable because even though it should be able to
+# run anywhere, I can only get it to run in an lxd container, and even then
+# occasionally some timing issues will crop up.
+@pytest.mark.unstable
```
OddBloke, line 52: It might be nice to have a timed element like we have in the curtin tests: `unstable_until("2021/02/01")` to force us to fix it _at some point_. Not needed for now, though.

OddBloke (approved):
LGTM, thanks! Ideally we'd have a stable test here, but pragmatically we're better landing this (which we know is inert by default, and useful enough for SRU verification) and moving forward.

PR #669, opened 2020-11-17
add --no-tty option to gpg

riedel:
## Proposed Commit Message
Make sure the gpg works even if VM provides no /dev/tty (e.g. debian cloud init)

LP:#1813396

## Additional Context
fixes https://bugs.launchpad.net/cloud-init/+bug/1813396

## Test Steps
on a VM without a terminal mounted as /dev/tty define an external keyserver such used for docker

```
apt:
  sources:
   docker:
     source: 'deb [arch=amd64] https://download.docker.com/linux/debian stretch stable'
     keyserver: keyserver.ubuntu.com
     keyid: 0EBFCD88
```

## Checklist:
 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #719, opened 2020-12-09
Integration test for lp-1813396

TheRealFalcon:
## Proposed Commit Message
Integration test for lp-1813396 and #669 

Ensure gpg is called with no tty flag.
Also, refactored the "ordered_items_in_text" to assert if the line
is missing and provide a more useful error message.

## Additional Context
#669

## Test Steps
fail:
CLOUD_INIT_CLOUD_INIT_SOURCE=NONE pytest tests/integration_tests/bugs/test_lp1813396.py
pass:
CLOUD_INIT_CLOUD_INIT_SOURCE=PROPOSED pytest tests/integration_tests/bugs/test_lp1813396.py

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_lp1813396.py
+++ b/tests/integration_tests/bugs/test_lp1813396.py
@@ -1,21 +1,22 @@
 """Integration test for lp-1813396
 
-Ensure gpg works even if VM provides no /dev/tty"""
+Ensure gpg is called with no tty flag.
+"""
 
 import pytest
 
 from tests.integration_tests.instances import IntegrationInstance
-from tests.integration_tests.log_utils import ordered_items_in_text
+from tests.integration_tests.log_utils import verify_ordered_items_in_text
 
 
 USER_DATA = """\
 #cloud-config
 apt:
   sources:
-    docker:
-      source: 'deb [arch=amd64] https://download.docker.com/linux/debian stretch stable'
+    cloudinit:
+      source: 'deb [arch=amd64] http://ppa.launchpad.net/cloud-init-dev/daily/ubuntu focal main'
       keyserver: keyserver.ubuntu.com
-      keyid: 0EBFCD88
+      keyid: E4D304DF
 """  # noqa: E501
 
 
@@ -25,9 +26,9 @@ def test_gpg_no_tty(client: IntegrationInstance):
     log = client.read_from_file('/var/log/cloud-init.log')
     to_verify = [
         "Running command ['gpg', '--no-tty', "
-        "'--keyserver=keyserver.ubuntu.com', '--recv-keys', '0EBFCD88'] "
+        "'--keyserver=keyserver.ubuntu.com', '--recv-keys', 'E4D304DF'] "
         "with allowed return codes [0] (shell=False, capture=True)",
-        "Imported key '0EBFCD88' from keyserver 'keyserver.ubuntu.com'",
+        "Imported key 'E4D304DF' from keyserver 'keyserver.ubuntu.com'",
         "finish: modules-config/config-apt-configure: SUCCESS",
     ]
-    assert ordered_items_in_text(to_verify, log)
+    verify_ordered_items_in_text(to_verify, log)
--- a/tests/integration_tests/bugs/test_lp1813396.py
+++ b/tests/integration_tests/bugs/test_lp1813396.py
@@ -1,0 +1,33 @@
+"""Integration test for lp-1813396
+
+Ensure gpg works even if VM provides no /dev/tty"""
+
+import pytest
+
+from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.log_utils import ordered_items_in_text
+
+
+USER_DATA = """\
+#cloud-config
+apt:
+  sources:
+    docker:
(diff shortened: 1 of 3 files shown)


tests/integration_tests/bugs/test_lp1813396.py
```diff
@@ -0,0 +1,33 @@
+"""Integration test for lp-1813396
+
+Ensure gpg works even if VM provides no /dev/tty"""
```
OddBloke, line 3: This isn't what we're doing: we're testing that cloud-init passes `--no-tty` to gpg. I think that's sufficient for validation on Ubuntu (particularly as lxd doesn't support _not_ providing `/dev/tty` to a system: https://lxd.readthedocs.io/en/latest/instances/#devices-configuration), but there _could_ be an integration test for this bug which masks/removes `/dev/tty` somehow and then tests that we successfully configure the repo.

tests/integration_tests/bugs/test_lp1813396.py
```diff
@@ -0,0 +1,33 @@
...
+from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.log_utils import ordered_items_in_text
+
+
+USER_DATA = """\
+#cloud-config
+apt:
+  sources:
+    docker:
+      source: 'deb [arch=amd64] https://download.docker.com/linux/debian stretch stable'
```
OddBloke, line 16: Are we better using a PPA for this? Easier to access when we're running tests inside our network, and we don't introduce a dependency on an "external" service.

tests/integration_tests/bugs/test_lp1813396.py
```diff
@@ -0,0 +1,33 @@
...
+def test_gpg_no_tty(client: IntegrationInstance):
+    log = client.read_from_file('/var/log/cloud-init.log')
+    to_verify = [
+        "Running command ['gpg', '--no-tty', "
+        "'--keyserver=keyserver.ubuntu.com', '--recv-keys', '0EBFCD88'] "
+        "with allowed return codes [0] (shell=False, capture=True)",
+        "Imported key '0EBFCD88' from keyserver 'keyserver.ubuntu.com'",
+        "finish: modules-config/config-apt-configure: SUCCESS",
+    ]
+    assert ordered_items_in_text(to_verify, log)
```
OddBloke, line 33: This could also use a better failure message: now that we're using it in two places, perhaps `ordered_items_in_text` should return something (the first unfound item?) that could be used for a richer message? Or it could become `assert_ordered_items_in_text` and be responsible for raising an `AssertionError` with a richer message?

(A nit for this PR, but I would like to see this improved.)

TheRealFalcon: I guess I'm wondering what a different error message will buy us here. If it doesn't pass, we'll need to manually inspect the log file to see why, and I don't think a different error will change that.

OddBloke:
On Wed, Dec 09, 2020 at 03:27:17PM -0800, James Falcon wrote:
> I guess I'm wondering what a different error message will buy us here.
> If it doesn't pass, we'll need to manually inspect the log file to see
> why, and I don't think a different error will change that.

That's a fair question.  If someone is specifically running a test which
uses this helper, then it doesn't buy us much, I agree.  (And, I also
agree that the usage _here_ specifically wouldn't benefit hugely from an
improved message: all of these messages are expected to be very close
together in the log; hence it being a nit. :)

However, most of the time we're going to be running tests using this
helper as part of a larger run.  In that case, a clearer assertion
message makes failures in "unknown" tests much easier to understand and
deal with.  It also makes failures easier to differentiate from one
another: this is really asserting a bunch of things (all these items are
in the log, item 1 < item 2, item 2 < item 3, etc.), and each of those
different failures could indicate a different underlying issue that the
test is revealing.  Being able to compare these failures across test
runs without debugging is valuable, and a message could draw a
developer's attention to inconsistencies across releases (e.g. if xenial
is failing at "Running command ..." but everything else is failing at
"finish: modules-config ..." then we likely have two distinct issues,
and you don't have to launch an instance for every single Ubuntu release
to start narrowing things down).

> we'll need to manually inspect the log file

This helper is not constrained to use on log files; it could well be
used to inspect a string which is only in memory; once the test is over,
this message is all the record we'll have of why it wasn't correct.
(The same also applies if we have a test which is only failing in CI: we
currently don't collect or store logs in Travis.)

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -80,8 +80,7 @@ def test_poweroff(self, session_cloud: IntegrationCloud,
             "running 'init-local'",
             'config-power-state-change already ran',
         ]
-        assert ordered_items_in_text(lines_to_check, log), (
-            'Expected data not in logs')
+        assert verify_ordered_items_in_text(lines_to_check, log)
```
OddBloke, line 83:
```suggestion
        verify_ordered_items_in_text(lines_to_check, log)
```

(I'll apply this on merge.)

OddBloke (approved):
Thanks! I've confirmed that this fails/passes as expected, and the change to the power state test (with my suggestion applied) also continues to behave as expected.

PR #721, opened 2020-12-10
Add support to resize rootfs if using LVM

otubo:
## Proposed Commit Message
This patch adds support to resize a single partition of a VM if it's using an
LVM underneath. The patch detects if it's LVM if the given block device
is a device mapper by its name (e.g. `/dev/dm-1`) and if it has slave
devices under it on sysfs. After that syspath is updated to the real
block device and growpart will be called to resize it (and automatically
its Physical Volume).

The Volume Group will be updated automatically and a final call to
extend the rootfs to the remaining space available will be made.

Using the same growpart configuration, the user can specify only one
device to be resized when using LVM and growpart, otherwise cloud-init
won't know which one should be resized and will fail.

rhbz: #1810878
LP: #1799953

## Additional Context
The commit also includes changes to the documentation as well as to unit tests to support the changes. The unit test is a copy from the existing test but using `/dev/XXdm-0` variants for LVM tests.

## Test Steps
One can verify the fix by installing cloud-init on an instance using LVM and configure it to expand one single partition, e.g. the `rootfs`.

## Checklist:


cloudinit/config/cc_growpart.py
```diff
@@ -316,6 +342,22 @@ def resize_devices(resizer, devices):
                          "failed to resize: disk=%s, ptnum=%s: %s" %
                          (disk, ptnum, e),))
 
+        if is_lvm and type(resizer).__name__ == "ResizeGrowPart":
```
otubo, line 345: Quick comment on this line: I'm pretty sure there's a better to do check if `resizer` is `ResizeGrowPart`.

cloudinit/config/cc_growpart.py
```diff
@@ -210,11 +210,35 @@ def get_size(filename):
...
 
     # input of /dev/vdb or /dev/disk/by-label/foo
     # rpath is hopefully a real-ish path in /dev (vda, sdb..)
     rpath = os.path.realpath(devpath)
 
+    is_lvm = True if re.search('.*dm.*', rpath) else False
+    if is_lvm:
+        try:
+            (out, _err) = subp.subp(["lvm", "lvs", devpath, "-o", "vgname",
+                                     "--noheadings"], env=myenv)
```
smoser, line 224: setting LANG is good if you're processing output.  thank you.
you can do it without the full copy though.

just use `update_env`

```python
myenv = {'LANG': 'C'}
subp.subp(["lvm", "lvs", ...], update_env=env)
```

and also `--output=vgname`

cloudinit/config/cc_growpart.py
```diff
@@ -210,11 +210,35 @@ def get_size(filename):
...
 def device_part_info(devpath):
     # convert an entry in /dev/ to parent disk and partition number
+    myenv = os.environ.copy()
+    myenv['LANG'] = 'C'
 
     # input of /dev/vdb or /dev/disk/by-label/foo
     # rpath is hopefully a real-ish path in /dev (vda, sdb..)
     rpath = os.path.realpath(devpath)
 
+    is_lvm = True if re.search('.*dm.*', rpath) else False
```
smoser, line 220: I think use of device mapper doesn't necessarily mean lvm.
and fwiw, the above is the same as:

```python
is_lvm = 'dm' in rpath
```

Perhaps a better func would be:

```python
def is_lvm_lv(devpath):
    # all lvm lvs will have a realpath as a 'dm-*' name.
    rpath = os.path.realpath(devpath)
    if not os.path.basename(rpath).startswith("dm-"):
        return False
    out, _ = subp.subp("udevadm", "info", devpath)
    # lvs should have DM_LV_NAME=<lvmuuid> and also DM_VG_NAME
    return 'DM_LV_NAME=' in out
```

igalic, line 220: i still think this should be a standalone function
that way we

- can lru cache it
- call it, instead of passing it around
- check for is_Linux(), which is also cached
- (check foris_AIX(), the only other platform with LVM support, as soon as someone adds AIX support to cloud-init)

igalic (changes requested):
🙋🏻‍♀️

cloudinit/config/cc_growpart.py
otubo: Oh yeah, forgot to address this comment from your last review. Sorry! I think we could do it, even though the optimization would not be so big. And even though this is a tiny one liner :-) But if you guys like the idea I can change it, no problem for me.

cloudinit/config/cc_growpart.py
```diff
@@ -210,11 +210,35 @@ def get_size(filename):
...
+                                     "--noheadings"], env=myenv)
+            vgname = out.strip()
+        except subp.ProcessExecutionError as e:
+            if e.exit_code != 0:
+                util.logexc(LOG, "Failed: can't get Volume Group information "
+                                 "from %s", devpath)
+                raise ResizeFailedException(e) from e
+
+        try:
+            (out, _err) = subp.subp(["lvm", "vgs", vgname.strip(), "-o",
```
smoser, line 233: use `--options=pvname` rather than '-o' 'pvname'

cloudinit/config/cc_growpart.py
```diff
@@ -210,11 +210,35 @@ def get_size(filename):
...
+                                    "pvname", "--noheadings"], env=myenv)
+            rpath = out.strip()
```
smoser, line 235: a VG can have more than one PV. if it does, this would give you a string like '/dev/vdb1\n/dev/vdb2'

cloudinit/config/cc_growpart.py
```diff
@@ -210,11 +210,35 @@ def get_size(filename):
...
     # convert an entry in /dev/ to parent disk and partition number
+    myenv = os.environ.copy()
+    myenv['LANG'] = 'C'
 
     # input of /dev/vdb or /dev/disk/by-label/foo
     # rpath is hopefully a real-ish path in /dev (vda, sdb..)
     rpath = os.path.realpath(devpath)
 
+    is_lvm = True if re.search('.*dm.*', rpath) else False
+    if is_lvm:
```
smoser, line 221: I think the whole block of code between 222 and 241 would work well as a function and allow for specific testing of it.

```python

func get_pvs_for_lv(devpath):
    """return a list of pvs that compose the logical volume 'devpath'
    ...
    return [pvs]

File: cloudinit/config/cc_growpart.py (hunk 1 above)
Comment (smoser) [reply cloudinit/config/cc_growpart.py:345]: I think just:

```python
if isinstance(resizer, ResizeGrowPart):
```

smoser (changes requested):
one general comment, and some changes requested inline.
If there is no 'lvm', then we should not try to resize with lvm (i think right now that would atually just fail/stacktrace if there is no 'lvm' command).  just check `if subp.which('lvm'):`

cloudinit/config/cc_growpart.py
igalic: this whole function should check for Linux first, and even then, you're not guaranteed to have udevadm, like on Alpine

dermotbradley: @igalic Hi. Alpine cloud-init package maintainer here - while Alpine normally uses mdev instead of udev, in the case of the cloud-init package I explicitly added a dependancy on eudev (Alpine's packaged version of udev) as a couple of the cloud-init modules (persistent net device naming & disk_setup) expect udev to be in use.

Packaged cloud-init dependancies visible here: https://pkgs.alpinelinux.org/package/edge/community/x86_64/cloud-init

smoser:
This is what it looks like without LVM:

 |name         | vda1 | vda2   | vda3  10G        | <--- extra 100G ---> |
 | ----------  | ---- | -----  | ---------------- | --------------------- |
 |abstraction  | none | none   | none             |                       |
 |             | EFI  | /boot  | /                |                       |

Based on 'growpart's simplified view of the world,
growpart currently accepts that the user wanted it to add:
 * add extra space to vda3
 * grow the filesystem on /

So after growpart, we end up with:

 |name         | vda1 | vda2   | vda3    110G                             |
 | ----------  | ---- | -----  | ---------------------------------------- |
 |abstraction  | none | none   | none                                     |
 |             | EFI  | /boot  | / 110G                                   |


In a simple LVM case (one PV on a VG) we might see a disk like this:

 | name        | vda1 | vda2  | vda3  10G                  | <--- extra 100G  ---> |
 |:----------- | ---- | ----- | -------------------------- | --------------------- |
 | abstraction | none | none  | LVM PV VDA3                |                       |
 |             |      |       | LVM VG VG0                 |                       |
 |             | EFI  | /boot | [ LV / 8G ] [ LV /home 2G] |                       |


Assuming we can determine a simple case like this.
Then the behavior consistent with the no-LVM behavior would be
to turn the above into:

| name        | vda1 | vda2  | vda3  110G                               |
| ----------- | ---- | ----- |:---------------------------------------- |
| abstraction | none | none  | LVM PV VDA3                              |
|             |      |       | LVM VG VG0  (automatically)              |
|             | EFI  | /boot | [ LV / 108G              ] [LV /home 2G] |


What we've done there is:
 1. grow the VDA3 partition
 2. grow the VDA3 PV
 3. grow the VG (I'm not sure if that happens automatically)
 4. grown the / LV to take 100% of the added
 5. resized the filesystem on /

There are all sorts of ways the above could be more complicated. The vg0 could be spread across multiple PVS via RAID or stripe or linear.

Lets focus on the simple case.

I think the operations above simplify down to 2 groups (1,2,3 and 4,5).
You basically have 2 choices here:
 a.) should you grow that partition and PV and VG (1,2,3)?  The user could have had plans to add another partition and use it another way, but we already take that choice away in the simple case above.
 b.) If you did so, then do you then grow the LV (4).  The user could have had plans to use the new space in the VG otherwise. If you do grow the LV, then step 5 is obvious.

I think I'm fine with saying 'yes' to both as the default behavior, and if the user does not like it, they should disable growpart.

Also, lets limit operations to the simple case described above. If the / LV is on a VG that spans multiple PVs, then we should just do nothing.

cloudinit/config/cc_growpart.py
otubo:
```
[root@huzhaolvmfordev01220919-vm1 ~]# lvm vgs rootvg --output vgname  --noheadings
vgs: unrecognized option '--output'
  Error during parsing of command line.
```
Currently I'm testing on a `RHEL-8.3` with LVM `2.03.09(2)-RHEL8`:
```
[root@huzhaolvmfordev01220919-vm1 ~]# lvm version
  LVM version:     2.03.09(2)-RHEL8 (2020-05-28)
  Library version: 1.02.171-RHEL8 (2020-05-28)
  Driver version:  4.42.0
  Configuration:   ./configure --build=x86_64-redhat-linux-gnu --host=x86_64-redhat-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-default-dm-run-dir=/run --with-default-run-dir=/run/lvm --with-default-pid-dir=/run --with-default-locking-dir=/run/lock/lvm --with-usrlibdir=/usr/lib64 --enable-fsadm --enable-write_install --with-user= --with-group= --with-device-uid=0 --with-device-gid=6 --with-device-mode=0660 --enable-pkgconfig --enable-cmdlib --enable-dmeventd --enable-blkid_wiping --with-cluster=internal --enable-cmirrord --with-udevdir=/usr/lib/udev/rules.d --enable-udev_sync --with-thin=internal --with-cache=internal --enable-lvmpolld --enable-lvmlockd-dlm --enable-lvmlockd-dlmcontrol --enable-lvmlockd-sanlock --enable-dbus-service --enable-notify-dbus --enable-dmfilemapd --with-writecache=internal --with-vdo=internal --with-vdo-format=/usr/bin/vdoformat --with-integrity=internal --disable-silent-rules
```

otubo:
Thanks for all the input @smoser . I believe I fixed most of the things. Feel free to leave comments on the rework :-) Also, had to force push because of outdated branch, I personally don't like merge commits. Also^2: Running into lots of tox/pylint errors, even on a clean checkout. I'll file a bug for that (but that's why I'm not running tox in its full extent on my local laptop)

cloudinit/config/cc_growpart.py
```diff
@@ -367,10 +370,10 @@ def resize_devices(resizer, devices):
                          "failed to resize: disk=%s, ptnum=%s: %s" %
                          (disk, ptnum, e),))
 
-        if is_lvm and type(resizer).__name__ == "ResizeGrowPart":
+        if is_lvm and isinstance(resizer, ResizeGrowPart)::
             try:
                 if len(devices) == 1:
-                    (_out, _err) = subp.subp(["lvm", "lvextend", "---extents",
+                    (_out, _err) = subp.subp(["lvm", "lvextend", "--extents",
```
smoser, line 376: "--extents=100%FREE"

cloudinit/config/cc_growpart.py
```diff
@@ -92,6 +93,52 @@ class RESIZE(object):
...
+            return False
+        out, _ = subp.subp("udevadm", "info", devpath)
+        # lvs should have DM_LV_NAME=<lvmuuid> and also DM_VG_NAME
+        return 'DM_LV_NAME=' in out
+    else:
+        False
+
+@lru_cache()
+def get_pvs_for_lv(devpath):
+    myenv = os.environ.copy()
```
smoser, line 111: you dont need the copy here, as update_env does it for you.

cloudinit/config/cc_growpart.py
```diff
@@ -92,6 +93,52 @@ class RESIZE(object):
...
+    myenv = {'LANG': 'C'}
+
+    if util.is_Linux():
+        try:
+            (out, _err) = subp.subp(["lvm", "lvs", devpath, "-options=vgname",
```
smoser, line 116: -options ? or --options.

cloudinit/config/cc_growpart.py
```diff
@@ -92,6 +93,52 @@ class RESIZE(object):
...
+            vgname = out.strip()
+        except subp.ProcessExecutionError as e:
+            if e.exit_code != 0:
+                util.logexc(LOG, "Failed: can't get Volume Group information "
+                            "from %s", devpath)
+                raise ResizeFailedException(e) from e
+
+        try:
+            (out, _err) = subp.subp(["lvm", "vgs", vgname, "-options=pvname",
+                                     "--noheadings"], update_env=myenv)
```
smoser, line 127: --options

cloudinit/config/cc_growpart.py
```diff
@@ -92,6 +93,52 @@ class RESIZE(object):
...
+            pv = out.strip().splitlines()
```
smoser, line 128: you probably have to splitlines and then strip each item in it. lvm is just bad... it is going to indent each entry.
doesn't specifically matter here as you're only expecting one and you're going to strip that line.

but really it should be more like:

    pvs = [p.strip() for p in out.splitlines()]

smoser: and you used 'pv' rather than 'pvs'. which will cause a problem below.

cloudinit/config/cc_growpart.py
```diff
@@ -92,6 +93,52 @@ class RESIZE(object):
...
+            if len(pvs) > 1:
+                util.logexc(LOG, "Failed: cloud-init currently supports only "
+                            "one Physical Volume per Volume Group")
```
smoser, line 131: as i said in IRC, i think that this should be an error if the user explicitly asked for resizing, but if this is just defaults, then debug or info "do not know how to resize multiple pvs".

cloudinit/config/cc_growpart.py
```diff
@@ -272,6 +323,8 @@ def devent2dev(devent):
 
 def resize_devices(resizer, devices):
     # returns a tuple of tuples containing (entry-in-devices, action, message)
+    myenv = os.environ.copy()
+    myenv['LANG'] = 'C'
```
smoser, line 327: don't need copy()

smoser (changes requested):

cloudinit/config/cc_growpart.py
```diff
@@ -316,6 +369,23 @@ def resize_devices(resizer, devices):
                          "failed to resize: disk=%s, ptnum=%s: %s" %
                          (disk, ptnum, e),))
 
+        if is_lvm and isinstance(resizer, ResizeGrowPart):
+            try:
+                if len(devices) == 1:
+                    (_out, _err) = subp.subp(["lvm", "lvextend", "--extents=",
+                                             "100%FREE", blockdev],
```
smoser, line 376: this is probably wrong, or at best looks wierd.

you'll execute:

    lvm lvextend --extents= 100%FREE <blockdev>

Even if that works, what we *want* to have is '--extents=100%FREE' as one argument not two.

    lvm lvextend --extents=100%FREE <blockdev>

I realize pycodestyle is causing you pain here. this should work:

```python

                   (_out, _err) = subp.subp(
                        ["lvm", "lvextend", "--extents=100%FREE", blockdev],
                        update_env={'LANG': 'C'})
```

cloudinit/config/cc_growpart.py
```diff
@@ -316,6 +369,23 @@ def resize_devices(resizer, devices):
...
+                                             update_env={'LANG': 'C'})
+                    info.append((devent, RESIZE.CHANGED,
+                                 "Logical Volume %s extended" % devices[0],))
+                else:
+                    raise ValueError("Exactly one device should be configured "
+                                     "to be resized when using LVM. More than "
+                                     "one configured: %s" % devices)
```
smoser, line 383: maybe just an info message.... I really think that we do not want to raise error (and have cloud-init fail) just because someone has added a pv to their root vg.

smoser (changes requested):
we're almost there, @otubo.  Thank you for your patience.

otubo:
> we're almost there, @otubo. Thank you for your patience.

Thank you, Scott for all the review and help merging this. Just finished the final fixes and pushed.

cloudinit/config/cc_growpart.py
otubo: I was having code style problems because I couldn't run tox on my laptop because of the issue of PR #791 . But now I can and everything is better :-)

cloudinit/config/cc_growpart.py
smoser: If you util.logexc, log gets a WARN ... which really means fail in cloud-init.  I might be spewing dated knowledge, but I think if you get here, then the log will have a WARN and even if not it will have a stack trace that just doesn't give warm fuzzies. Other than this, I am happy with the PR at this point.

@blackboxsw or @OddBloke can you give your input?  in summary... we'll get here if the system has it's root filesystem on an LV that is part of a VG that includes more than one PVs.  It isn't really an error... cloud-init just doesn't know how to handle this situation.

I'm concerned that with this change, someone who has been happily using a system/image like this will now have stack traces showing up in their logs.

smoser (commented):
@otubo thanks... I just have the one comment. And seeking further input from @OddBloke or @blackboxsw .  It really is not a huge deal, but I just don't like stack traces that aren't errors.

cloudinit/config/cc_growpart.py
blackboxsw: I'm +1 on reducting the amount of illegible Tracebacks in cloud-init where we can. We have a couple of feature requests that are looking for easier logs to read (or less logging) from cloud-init. In this case, it oes seems that a standalone logging.info would capture the extent of the shortfall for cloud-init. It doesn't have to block this PR, but it's nice to keep in mind as "Tracebacks" in cloud-init.log do get treated as failures in integration tests and quick system inspection during triage.

cloudinit/config/cc_growpart.py
```diff
@@ -93,6 +94,53 @@ class RESIZE(object):
...
+                util.logexc(LOG, "Failed: can't get Volume Group information "
+                            "from %s", devpath)
+                raise ResizeFailedException(e) from e
+
+        try:
+            (out, _err) = subp.subp(["lvm", "vgs", vgname, "--options=pvname",
+                                     "--noheadings"], update_env=myenv)
+            pvs = [p.strip() for p in out.splitlines()]
+            if len(pvs) > 1:
+                util.logexc(LOG, "Do not know how to resize multiple Physical"
```
blackboxsw, line 131: This should be probably be a logging.warning since we aren't really adding logexc value here. We also should raise a  ResizeFailedException here right? otherwise I think we fall through to the end of the function and return None which then is used by 
bname = os.path.basename(rpath) which will die with a TypeError here 
```TypeError: expected str, bytes or os.PathLike object, not NoneType```

File: cloudinit/config/cc_growpart.py [hunk 17]
```diff
@@ -93,6 +94,53 @@ class RESIZE(object):
...
+                            " Volumes")
+            else:
+                return pvs[0]
+        except subp.ProcessExecutionError as e:
+            if e.exit_code != 0:
+                util.logexc(LOG, "Failed: can't get Physical Volume "
+                            "information from Volume Group %s", vgname)
+                raise ResizeFailedException(e) from e
+    else:
+        return False
```
blackboxsw, line 141: Returning False in failure cases versus a "string" in success cases feels confusing for `get_pvs_for_lv`. It looks like behavior like this could cause a problem because we don't check the return value and blindly call `    bname = os.path.basename(rpath) which ultimately leads to a confusing type error in all failure cases here.
` 
I think we probably should emit a warning message like ("No support for get_pvs_for_lv on {system}".format(systemd=platform.system()). If we want to continue with just a warning message, then we probably should be should be checking the return value from get_pvs_for_lv before setting rpath at line 268 `rpath = get_pvs_for_lv(devpath)`

blackboxsw (changes requested):
Thanks @otubo; minor suggestion about a possible failure path on Linux and non-linux systems that probably requires a bit of work on `get_pvs_for_lv` and handling.

Minimally, I'd like to see get_pvs_for_lv values get checked before putting it into rpath and allowing os.path.basename to field a boolean return value.

Ideally get_pvs_for_lv would return only one data type, not str or False.

Please also change the two util.logexc to logging.warning in cases where we would like to proceed or actually raise an exception if we really want cloud-init to fall over. It's confusing to a user to have a logged exception in cloud-init that isn't actually raising an exception.

cloudinit/config/cc_growpart.py
otubo: I added a check and the warning message :) thanks for the review!

cloudinit/config/cc_growpart.py
smoser: @otubo , @blackboxsw 

> This should be probably be a logging.warning since we aren't really adding logexc value here

I think this was the one sticking point that both Chad and I had.

I think this is at most 'info'.  WARNING means something went wrong.  I'd like to not introduce WARN into someone's logs that was happily using cloud-init and LVM previously.  ie, if someone has been using cloud-init with multiple PVs and they upgrade their cloud-init, they will now start seeing this WARN in their logs, even though nothing change for them (before: it didn't resize this.  after: it didn't resize this).

Thats my $0.02.

cloudinit/config/cc_growpart.py
```diff
@@ -93,6 +95,55 @@ class RESIZE(object):
...
+                            "from %s", devpath)
+                raise ResizeFailedException(e) from e
+
+        try:
+            (out, _err) = subp.subp(["lvm", "vgs", vgname, "--options=pvname",
+                                     "--noheadings"], update_env=myenv)
+            pvs = [p.strip() for p in out.splitlines()]
+            if len(pvs) > 1:
+                LOG.info("Do not know how to resize multiple Physical"
+                         " Volumes")
```
smoser, line 133: so this is a weird path, and if we get here, we will end up not returning anything (the same as 'return None').

Then, down below on line 144 we return 'False'.

Returning one of three different types (False, string, or None) from any function is odd at best.

We can get this down to string or None easily enough with just removing the 'return False' on line 144 and adding a return None outside the 'else' block there.

cloudinit/config/cc_growpart.py
```diff
@@ -93,6 +95,55 @@ class RESIZE(object):
...
+            else:
+                return pvs[0]
+        except subp.ProcessExecutionError as e:
+            if e.exit_code != 0:
+                util.logexc(LOG, "Failed: can't get Physical Volume "
+                            "information from Volume Group %s", vgname)
+                raise ResizeFailedException(e) from e
+    else:
+        LOG.warning("No support for get_pvs_for_lv on %s",
+                    platform.system())
```
smoser, line 143: I really do not want a warning here.  Doing so means that we have a warning in cases that previously did not.  Someone had an image that was working perfectly fine for them (two pvs for /).  It just was not resizing anything.  But now they'll have a warning in their logs, and in cloud-init speak, warnings are pretty much errors.

cloudinit/config/cc_growpart.py
```diff
@@ -316,6 +373,23 @@ def resize_devices(resizer, devices):
...
+                if len(devices) == 1:
+                    (_out, _err) = subp.subp(
+                        ["lvm", "lvextend", "--extents=100%FREE", blockdev],
+                        update_env={'LANG': 'C'})
+                    info.append((devent, RESIZE.CHANGED,
+                                 "Logical Volume %s extended" % devices[0],))
+                else:
+                    util.logexc(LOG, "Exactly one device should be configured "
+                                     "to be resized when using LVM. More than "
+                                     "one configured: %s" % devices)
```
smoser, line 387: and I'm pretty sure that stack traces are WARNings, so I'd really like this to be at strongest an INFO message.

smoser (changes requested):

smoser (approved):
Thanks, @otubo

cloudinit/config/cc_growpart.py
```diff
@@ -208,13 +259,18 @@ def get_size(filename):
...
-def device_part_info(devpath):
+def device_part_info(devpath, is_lvm):
     # convert an entry in /dev/ to parent disk and partition number
 
     # input of /dev/vdb or /dev/disk/by-label/foo
     # rpath is hopefully a real-ish path in /dev (vda, sdb..)
     rpath = os.path.realpath(devpath)
 
+    # first check if this is an LVM and get its PVs
+    lvm_rpath = get_pvs_for_lv(devpath)
```
blackboxsw, line 270: Won't this call raise an exception on any system that doesn't have "lvm" command installed?

blackboxsw (changes requested):
Needs fixing just so that we have a pre-flight subp.which('lvm') check before erroring and calling lvm commands.

cloudinit/config/cc_growpart.py
smoser: yeah... we shouldnt stack trace on that. we should log info (or less) and not fail.

smoser (approved):
I'm ok with this as it is.  But if you wanted to move the sanity check to the top of get_pvs_for_lv, i would approve that change also.

Thank you so much for your patience, @otubo 

```python
@lru_cache
def get_pvs_for_lv(devpath):
    if not util.is_Linux():
        LOG.info("No support for LVM on %s", platform.system())
        return None
    if not subp.which('lvm'):
        LOG.info("No 'lvm' command present")
        return None

    # continue on with less indentation
    try:
        (out, _err) = subp.subp(...
```

otubo:
> I'm ok with this as it is. But if you wanted to move the sanity check to the top of get_pvs_for_lv, i would approve that change also.
> 
> Thank you so much for your patience, @otubo
> 
> ```python
> @lru_cache
> def get_pvs_for_lv(devpath):
>     if not util.is_Linux():
>         LOG.info("No support for LVM on %s", platform.system())
>         return None
>     if not subp.which('lvm'):
>         LOG.info("No 'lvm' command present")
>         return None
> 
>     # continue on with less indentation
>     try:
>         (out, _err) = subp.subp(...
> ```

I think that makes sense. Just updated. Thanks!

smoser (approved):

blackboxsw (approved):
+1 thanks a lot @otubo all good on my front with this review. I've just clicked update branch to get latest from cloud-init tip and someone can land it if they get to it before I see it tomorrow morning.

PR #726, opened 2020-12-11
Some test_upgrade fixes

TheRealFalcon:
## Proposed Commit Message
Some test_upgrade fixes

- workaround pad.lv/1890528 for restarting instances
- move wait param from launch_kwargs to launch call
- remove name param as it's not universally supported
- add platfrom to log names

## Additional Context
n/a

## Test Steps
pytest tests/integration_tests/test_upgrade.py

## Checklist:

### DIFF
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -1,5 +1,6 @@
 import logging
 import pytest
+import time
 from pathlib import Path
 
 from tests.integration_tests.clouds import ImageSpecification, IntegrationCloud
@@ -40,6 +41,19 @@ def _output_to_compare(instance, file_path, netcfg_path):
             f.write(instance.execute(command) + '\n')
 
 
+def _restart(instance):
+    # work around pad.lv/1890528
+    try:
+        instance.restart(raise_on_cloudinit_failure=True)
+    except OSError as e:
+        for _ in range(10):
+            time.sleep(5)
+            result = instance.execute('cloud-init status --wait --long')
+            if result.ok:
+                return
+        raise e
+
+
 @pytest.mark.sru_2020_11
 def test_upgrade(session_cloud: IntegrationCloud):
     source = get_validated_source()
@@ -50,9 +64,7 @@ def test_upgrade(session_cloud: IntegrationCloud):
         return  # type checking doesn't understand that skip raises
 
     launch_kwargs = {
-        'name': 'integration-upgrade-test',
         'image_id': session_cloud._get_initial_image(),
-        'wait': True,
     }
 
     image = ImageSpecification.from_os_image()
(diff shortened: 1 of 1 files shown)


tests/integration_tests/test_upgrade.py
```diff
@@ -50,17 +64,16 @@ def test_upgrade(session_cloud: IntegrationCloud):
...
         'image_id': session_cloud._get_initial_image(),
-        'wait': True,
     }
 
     image = ImageSpecification.from_os_image()
 
     # Get the paths to write test logs
     output_dir = Path(session_cloud.settings.LOCAL_LOG_PATH)
     output_dir.mkdir(parents=True, exist_ok=True)
-    base_filename = 'test_upgrade_{os}_{{stage}}_{time}.log'.format(
```
mitechie, line 63: Is there an issue/etc for this change? Was the filename always wrong?

mitechie (approved):

tests/integration_tests/test_upgrade.py
TheRealFalcon: It's the output log of the running test. It outputs some of the same commands before and after the upgrade and writes the output to before and after log files. When running tests in batch, getting a number of log files like:
```
test_upgrade_gce_focal_before_1.log
test_upgrade_ec2_focal_before_2.log
test_upgrade_lxd_focal_before_3.log
```
is much preferable to
```
test_upgrade_focal_before_1.log
test_upgrade_focal_before_2.log
test_upgrade_focal_before_3.log
```

tests/integration_tests/test_upgrade.py
```diff
@@ -40,6 +41,19 @@ def _output_to_compare(instance, file_path, netcfg_path):
             f.write(instance.execute(command) + '\n')
 
 
+def _restart(instance):
+    # work around pad.lv/1890528
```
OddBloke, line 45: Could you explain in a little more detail how this is doing that? My reading of that bug is that it only triggers if `cloud-init status --wait` exits non-zero: if that's exiting non-zero (and therefore raising the `OSError`, I assume?) then `result.ok` in the `for` loop below is never going to be `True`, so we'll never `return` early. What am I missing?

(Even if I'm not missing something and this is a bug, this doesn't need to block landing; we should fix it soon hereafter though.)

OddBloke (approved):
One nit inline, but otherwise this looks good, thanks!

tests/integration_tests/test_upgrade.py
TheRealFalcon: Hmmm, maybe this is a different bug then. On ec2, cloud-init returns early with non-zero status (I'm inferring this from pycloudlib behavior). Connecting again, it'll complete and return 0. Nothing in the logs or result.json or status.json indicate any sort of failure or problem anywhere.

TheRealFalcon: I'll file a separate bug with logs and update the reference here

OddBloke: Thanks!

PR #727, opened 2020-12-14
Ensure overriding test vars with env vars works for booleans

TheRealFalcon:
## Proposed Commit Message
Ensure overriding test vars with env vars works for booleans

## Additional Context
n/a

## Test Steps
CLOUD_INIT_KEEP_INSTANCE=False pytest <any_test>
Ensure the instance isn't kept around.

## Checklist:

### DIFF
--- a/tests/integration_tests/integration_settings.py
+++ b/tests/integration_tests/integration_settings.py
@@ -1,6 +1,8 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 import os
 
+from distutils.util import strtobool
+
 ##################################################################
 # LAUNCH SETTINGS
 ##################################################################
@@ -109,6 +111,12 @@ except ImportError:
 # Perhaps a bit too hacky, but it works :)
 current_settings = [var for var in locals() if var.isupper()]
 for setting in current_settings:
-    globals()[setting] = os.getenv(
+    env_setting = os.getenv(
         'CLOUD_INIT_{}'.format(setting), globals()[setting]
     )
+    if isinstance(env_setting, str):
+        try:
+            env_setting = bool(strtobool(env_setting.strip()))
+        except ValueError:
+            pass
+    globals()[setting] = env_setting


tests/integration_tests/integration_settings.py
```diff
@@ -109,6 +111,15 @@
 # Perhaps a bit too hacky, but it works :)
 current_settings = [var for var in locals() if var.isupper()]
 for setting in current_settings:
-    globals()[setting] = os.getenv(
+    env_setting = os.getenv(
         'CLOUD_INIT_{}'.format(setting), globals()[setting]
     )
+    if isinstance(env_setting, str):
+        if env_setting.lower() == 'none':
```
mitechie, line 118: should we also be doing things like whitespace trimming here?

mitechie (approved):
one question on just doing all the right things for processing.

mitechie:
It'd be good to note the test/change that drove this as well to help justify the change in practice.

TheRealFalcon:
> It'd be good to note the test/change that drove this as well to help justify the change in practice.

It's a fix for a test running feature that hasn't been widely used yet. The only specific thing that drove it was I was about to run a fairly large set of integration tests so wanted to make sure `KEEP_INSTANCE` was set to false in the environment. I started setting the env var and then remembered we didn't do any particular boolean parsing so went and fixed it real quick.

tests/integration_tests/integration_settings.py
```diff
@@ -1,6 +1,8 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 import os
 
+from distutils.util import strtobool
```
OddBloke, line 4: :notes: hello util my old friend :notes:

OddBloke (approved):
Thanks!

(I wonder if we have any cloud-init code that could be simplified with this? Probably not unless the interface is bug-compatible.)

PR #626, opened 2020-10-24
Make wakeonlan Network Config v2 setting actually work

dermotbradley:
## Proposed Commit Message

Add code so that specifying "wakeonlan: true" actually results in relevant
configuration entry appearing in /etc/network/interfaces, NetPlan, and
sysconfig for RHEL and OpenSuse.

Add testcases for the above.

## Test Steps

In a Network Config v2 YAML file specify the following:

```
version: 2
ethernets:
    eth0:
        dhcp4: true
        wakeonlan: true
```

Then check the resultant Netplan file to see if wakeonlan specified there.

Wakeonlan is not supported in a v1 config file.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #728, opened 2020-12-14
Integration test for gh-626

TheRealFalcon:
## Proposed Commit Message
Integration test for gh-626

Ensure if wakeonlan is specified in the network config that it is
rendered in the /etc/network/interfaces or netplan config.

## Additional Context
See #626

## Test Steps
pytest tests/integration_tests/bugs/test_gh626.py

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_gh626.py
+++ b/tests/integration_tests/bugs/test_gh626.py
@@ -1,0 +1,42 @@
+"""Integration test for gh-626.
+
+Ensure if wakeonlan is specified in the network config that it is rendered
+in the /etc/network/interfaces or netplan config.
+"""
+
+import pytest
+import yaml
+
+from tests.integration_tests.clouds import ImageSpecification
+from tests.integration_tests.instances import IntegrationInstance
+
+
+NETWORK_CONFIG = """\
+version: 2
+ethernets:
+  eth0:
+    dhcp4: true
+    wakeonlan: true
+"""
+
+EXPECTED_ENI_END = """\
+iface eth0 inet dhcp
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
It looks like, at least under Ubuntu, the netplan config renders as expected, even before the changes we're testing (but not ENI). Is this expected?

tests/integration_tests/bugs/test_gh626.py
```diff
@@ -0,0 +1,43 @@
...
+
+@pytest.mark.sru_2020_11
+@pytest.mark.ubuntu
+@pytest.mark.lxd_container
+@pytest.mark.lxd_vm
+@pytest.mark.lxd_config_dict({
+    'user.network-config': NETWORK_CONFIG
+})
+def test_wakeonlan(client: IntegrationInstance):
+    if ImageSpecification.from_os_image().release == 'xenial':
```
TheRealFalcon, line 35: Checking this way isn't ideal, but I see it as being short term until we come up with a better way of checking

OddBloke: I'm not even sure if this is desirable, but this would DTRT even if we haven't detected a specified image as `xenial`:

```suggestion
    if client.execute("lsb_release --short --codename") == "xenial":
```

(Presumably "a better way" would be separate tests with appropriate release tagging?)

OddBloke (approved):
Code LGTM, and confirmed the test passes/fails as expected. (As netplan is just passthrough, we expect it to pass already; this is regression-testing that, really.) Thanks!

PR #732, opened 2020-12-17
integration_tests: port ca_certs tests from cloud_tests

OddBloke:
## Test Steps

Run `pytest --log-cli-level=INFO tests/integration_tests/modules/test_ca_certs.py` and observe success.

## Checklist:


TheRealFalcon (approved):
Why is this ubuntu only?

OddBloke:
> Why is this ubuntu only?

It isn't exactly: the test relies on Ubuntu's file naming, but (sh|w)ould also run on Debian and Alpine; we don't have marks for those. The module under test also supports RHEL, but uses different filenames for that (so would need updating). I've added a couple of TODO items to the module-level docstring to capture this.

PR #745, opened 2021-01-01
first commit for first cloud-init PR

beantaxi:
## Proposed Commit Message

```
Script handlers for per-boot, per-instance, and per-once.

Several (6?) months ago I inquired on #cloud-init about how to include multiple init scripts in user data, one per instance and one per boot. I was told this wasn't supported. I needed the behavior, so I had a discussion about implementation, and when I suggested a simple custom part handler for each frequency, to be used during the normal make-mime process, I was told that sounded good and it might be supported.

I've been using my solution in my own work for months, and since a client is about to go live with it, it seems like a good time to work towards getting it included in cloud-init proper.
```

## Test Steps

I'm currently running a command like the following

```
cloud-init devel make-mime --force \
                                --attach "$myCloudInit/cloudinit/cmd/devel/shell_script_per_boot.py":part-handler \
                                --attach "$myCloudInit/cloudinit/cmd/devel/shell_script_per_instance.py":part-handler \
                                --attach "$myCloudInit/cloudinit/cmd/devel/shell_script_per_once.py":part-handler \
                                --attach "$myScriptFolder./init-boot.sh":x-shellscript-per-boot \
                                --attach" $myScriptFolder/init-instance.sh":x-shellscript-per-instance
```                                
Then, I deploy the file as used data, check logfiles, and check /var/lib/cloud/scripts/per-boot and -instance to make sure my files are there.

I'd love for the script files to be placed in cloud-init such that they do not need to be explicitly added to userdata. But I didn't know how to do that, so I put them in cloudinit/cmd/devel instead.

## Checklist:

(I am happy to do these once my code looks like it's on the right track)


cloudinit/cmd/devel/shell_script_per_boot.py
```diff
@@ -0,0 +1,25 @@
...
+def list_types():
+    # return a list of mime-types that are handled by this module
+    LOG.debug("in shell_script-per-boot.list_types() ...")
+    return(["text/x-shellscript-per-boot"])
+
+
+def handle_part(data, ctype, script_path, payload):
+    if script_path is not None:
+        LOG.info("in shell_script-per-boot.handle_part() ...")
+        LOG.debug("script_path=%s", script_path)
```
raharper, line 20: Let's drop the verbose logging in these handlers; util.write_file() already logs which file it writes.

raharper (changes requested):
Thanks for putting up this PR!   This is a useful addition to cloud-init script handling, thank you!

I've included some feedback in the code.  I would also like to see some unittests for these.  We have some existing tests you can add to, I think ```tests/unittests/test_builtin_handlers.py``` could add a TestScriptPerHandler class testing the common scripts-per-frequency common implementation.

smoser (changes requested):
Hi.

If I understand correctly, you're adding handlers for 3 new mime types. 

   * text/x-shellscript-per-once
   * text/x-shellscript-per-always
   * text/x-shellscript-per-instance

You should be adding these to a new file in cloudinit/handlers/ and then
hooking them into cloudinit stages via cloudinit/stages.py.

cloudinit/cmd/devel/shell_script_per_boot.py
```diff
@@ -0,0 +1,25 @@
...
+        (folder, filename) = os.path.split(script_path)
+        LOG.debug("folder=%s filename=%s", folder, filename)
+        path = f"/var/lib/cloud/scripts/per-boot/{filename}"
+        LOG.debug("path=%s", path)
+        util.write_file(path, payload, 0o700)
```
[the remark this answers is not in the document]
beantaxi: @raharper Happy to add tests, but I didn't see any shell script handler tests in the file you mentioned, or elsewhere, so I wasn't quite sure where to put tests. I'm also not quite sure how to unit test a module that simply saves a file to a folder,  but I am open to ideas.

chrislalos:
Revisiting this ... 

I've just been reading some cloud-init code. Should I be trying to reuse some code from cc_write_files.py, like cc_write_files#write_files() to do what I want? Even if that requires lifting the code out of cc_write_files.py and into somewhere common, would that be the best way to go?

chrislalos:
Also, in the comment for the bot-closure of this issue, it says "(If the pull request is closed, please do feel free to reopen it if you wish to continue working on it.)" I'm happy to reopen but I'm not seeing where. Thanks!

OddBloke:
> I'm happy to reopen but I'm not seeing where. Thanks!

Hmm, I guess that must only be available to folks with commit rights; we'll fix that message. (I'll reopen it for you!)

OddBloke:
> I've just been reading some cloud-init code. Should I be trying to reuse some code from cc_write_files.py, like cc_write_files#write_files() to do what I want? Even if that requires lifting the code out of cc_write_files.py and into somewhere common, would that be the best way to go?

I've only scanned through the PR, but I don't see where this would be needed: `write_files` handles the configuration format that is specified under `write_files` in cloud-config YAML. `util.write_file` should be sufficient for your needs, I think.

mitechie:
Looks like the bot got cranky and re-closed since the tag was there. I've removed the stale-pr tag and reopened.

cloudinit/cmd/devel/shell_script_per_boot.py
```diff
@@ -0,0 +1,25 @@
...
+    return(["text/x-shellscript-per-boot"])
+
+
+def handle_part(data, ctype, script_path, payload):
+    if script_path is not None:
+        LOG.info("in shell_script-per-boot.handle_part() ...")
+        LOG.debug("script_path=%s", script_path)
+        (folder, filename) = os.path.split(script_path)
+        LOG.debug("folder=%s filename=%s", folder, filename)
+        path = f"/var/lib/cloud/scripts/per-boot/{filename}"
```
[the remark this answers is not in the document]
beantaxi: Thanks. I've made this change. Interestingly, the shell_script.py handler actually doesn't do what you described. It does some hocus pocus in the constructor to extract that scripts path, and assign it to a variable, which it then uses later in handle_part. I am not sure which is better -- yours is certainly simpler -- and maybe it'd be better to unify shell_script.py and my new handlers.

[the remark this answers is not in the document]
beantaxi: Correction: I could not get this change to build, because of  a circular reference issue: cloudinit.cmd.devel:__init.py__ imports stages.py, which imports all the handlers including mine, so my handlers can't import cloud.cmd.devel to get that utility method. I ended up copy/pasting the shell_script.py approach.

The method you pointed out seems like a useful method though ... would it make sense to put that functionality up somewhere higher, like in cloudinit.util?

beantaxi:
I think I finally have committed a viable solution. I've hooked my handlers into cloudinit stages, and I'm not harcoding the scripts_dir any more; instead I'm doing what shell_script.py. Thank you for bearing with many commits; I've gradually learned how to run the various travis steps locally, so hopefully I am no longer relying on `git push` to see if my code builds.

The biggest deviation I made from the existing handlers, is I didn't use the `prefixes` idiom for `list_types()`. To be honest, I didn't care for it. It seemed unnecessarily complex, and unnecessarily invasive of very common code, to add entries to a common prefixes map, and then define a prefixes constant in my handles class(es), and then make list_types() do a lookup in this map. It seems clearer to simply have `list_types` return the correct types. But let me know if I'm missing something.

I also have 3 classes when I could probably get away with one. At 2am I didn't want to take a shot at that, but it looks like a reasonable change now.

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,84 @@
...
+    path = os.path.join(folder, filename)
+    payload = util.dos2unix(payload)
+    util.write_file(path, payload, 0o700)
+
+
+# per-boot
+class ShellScriptPerBootPartHandler(Handler):
+    def __init__(self, paths, **_kwargs):
+        Handler.__init__(self, PER_ALWAYS)
+        self.scripts_dir = paths.get_ipath_cur('scripts')
```
TheRealFalcon, line 37: You want `get_cpath` for this and the line two lines below this. 
`/var/lib/cloud/scripts/{per-boot,per-instance,per-once}`

There's no boot distinction (or sub-directories) in `/var/lib/cloud/instance/scripts/`. Anything placed there will run every boot.

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def handle_part(self, data, ctype, script_path, payload, frequency):
+        if script_path is not None:
+            LOG.debug("script_path=%s", script_path)
+            write_script_by_frequency(script_path, payload, PER_ALWAYS,
+                                      self.scripts_dir)
+
+
+# per-instance
+class ShellScriptPerInstancePartHandler(Handler):
```
TheRealFalcon, line 52: You mentioned this in your previous comment, but all of the functionality in these classes should live in a single base class, then define sub-classes with something like `script_type` and `script_frequency` as class variables and nothing else.

TheRealFalcon (changes requested):
This is definitely coming along!

There's no need for `shell_script_per_boot.py`, `shell_script_per_instance.py`, and `shell_script_per_once.py` anymore, correct? These should be removed.

> The biggest deviation I made from the existing handlers, is I didn't use the prefixes idiom for list_types(). To be honest, I didn't care for it. It seemed unnecessarily complex, and unnecessarily invasive of very common code, to add entries to a common prefixes map, and then define a prefixes constant in my handles class(es), and then make list_types() do a lookup in this map. It seems clearer to simply have list_types return the correct types. But let me know if I'm missing something.

You're not entirely wrong in this specific case. Normally these prefixes are very useful because they allow us to understand what type of file we're looking at when passed some arbitrary user data. However, in this case, since an executable file requires the `#!` at the top, there's no way we can pass this as userdata except as a MIME encoded file. The other thing that defining a prefix gives us is that we don't have to use the `--force` when calling `cloud-init devel make-mime`, and don't get any warnings. For this reason, I think it would still be useful to have values for these in the `INCLUSION_TYPES_MAP` in `handlers/__init__.py`.

cloudinit/handlers/shell_script_by_frequency.py
beantaxi: If I'm reading these correctly ...

* It sounds like the INCLUSION_TYPES_MAP is doing doubleduty: it's both a whitelist of mime-types to suppress warnings from `cloud-init devel make-mime`, and a mapping of prefixes to mime-types. So if I include an 'identity mapping' of 'x-shellscript-per-xxx': 'x-shellscript-per-xxx' for all 3 frequencies, I'll benefit from the first feature and vestigially participate in the second.

* `paths.get_ipath_cur()` returns /var/cloud/instance/scripts, and I need /var/cloud/scripts, which can get by just switching to `paths.get_cpath()`.

* You'd like to see a base class and 3 subclasses. It's too bad I can't go one better, and just write a single class, and let `handle_parts()`do the right thing based on frequency. But it looks like I can't, because of the Handler subclass:

```
    def __init__(self, paths, **_kwargs):
        Handler.__init__(self, PER_ONCE)
```

If we don't actually need to call Handler's c'tor, and/or that frequency doesn't actually matter, then we can get away with a single class. But that feels sloppy. If I have to use 3 subclasses that's ok.

Is that everything? What am I missing ...

beantaxi:
I think the last 3 commits resolve thee 3 issues that were described.

I needed some `#pylint disable`s to make pylint happy, as it appeared to be throwing false negatives on base constructor calls, for too many positional parameters. pylint had some older issues with false negatives for too many positional arguments, which were closed, but it was enough to dissuade me from lengthy investigation. I'm happy to leave the next step up to y'all.

beantaxi:
One more commit, to merge in changes from master (no conflicts)

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+# per-boot
+class ShellScriptPerBootPartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-boot"]
+
+    def __init__(self, paths, **_kwargs):
+        # pylint: disable=too-many-function-args
+        ShellScriptByFreqPartHandler.__init__(self, paths, PER_ALWAYS,
+                                              **_kwargs)
```
TheRealFalcon, line 68:
```suggestion
    frequency = PER_ALWAYS
```

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+# per-instance
+class ShellScriptPerInstancePartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-instance"]
+
+    def __init__(self, paths, **_kwargs):
+        # pylint: disable=too-many-function-args
+        ShellScriptByFreqPartHandler.__init__(self, paths, PER_INSTANCE,
+                                              **_kwargs)
```
TheRealFalcon, line 78:
```suggestion
    frequency = PER_INSTANCE
```

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+# per-once
+class ShellScriptPerOncePartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-once"]
+
+    def __init__(self, paths, **_kwargs):
+        # pylint: disable=too-many-function-args
+        ShellScriptByFreqPartHandler.__init__(self, paths, PER_ONCE,
+                                              **_kwargs)
```
TheRealFalcon, line 88:
```suggestion
    frequency = PER_ONCE
```

TheRealFalcon (changes requested):
Hey @beantaxi , thanks for the updates. I have a few more (much more minor) comments inline.

One major issue though, is that I would have expected we no longer need the original three `shell_script_per_boot.py`, `shell_script_per_instance.py`, and `shell_script_per_once.py` files, but are you still using them somehow? How do you generate your mime archive now?

I can nuke those three files entirely and create the mime with something like:
```bash
cloud-init devel make-mime -a /tmp/script1.sh:x-shellscript-per-boot -a /tmp/script2.sh:x-shellscript-per-instance -a /tmp/script3.sh:x-shellscript-per-once > /tmp/user-data
```
If there's no longer a use for these files, then lets get rid of them.

cloudinit/handlers/__init__.py
```diff
@@ -51,6 +51,10 @@
     '#cloud-config-archive': 'text/cloud-config-archive',
     '#cloud-config-jsonp': 'text/cloud-config-jsonp',
     '## template: jinja': 'text/jinja2',
+    'text/x-shellscript-per-boot': 'text/x-shellscript-per-boot',
```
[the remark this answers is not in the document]
beantaxi: Done

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
+# part-handler
```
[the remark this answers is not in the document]
beantaxi: Done (but see below)

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
...
+# vi: syntax=python ts=4
```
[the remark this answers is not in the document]
beantaxi: Done .. but note those were lines were there because of https://cloudinit.readthedocs.io/en/latest/topics/format.html, so someone might wish to change the docs as well.

cloudinit/handlers/shell_script_by_frequency.py
```diff
@@ -0,0 +1,88 @@
...
+    filename = util.clean_filename(filename)
+    folder = get_script_folder_by_frequency(frequency, scripts_dir)
+    path = os.path.join(folder, filename)
+    payload = util.dos2unix(payload)
+    util.write_file(path, payload, 0o700)
+
+
+# This is purely to allow packaging args up into a single object to please
+# pylint and avoid 'too many positional args' complaints. I'd be happy to
+# have an alernative.
```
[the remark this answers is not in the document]
beantaxi: My fault ... in an attempt to address my `pylint` issue, I added a class to package up all positional args -- which were only two -- into a single arg. When that didn't help `pylint`, and Google had other examples of people with the same problem (though some folks didn't like yourself) ... when I read that, I scrapped the class and went with `#pylint:disable`. But I didn't scrap the comment that went with the class, which led to your understandable confusion.

cloudinit/handlers/shell_script_by_frequency.py
beantaxi: See below ...

cloudinit/handlers/shell_script_by_frequency.py
beantaxi: See below ...

cloudinit/handlers/shell_script_by_frequency.py
beantaxi: I can make these changes, but they seem odd. `frequency` is assigned by the `Handler` c'tor, which all these classes inherit from. Is it necessary in Python for subclasses to assign instance variables, even if they're assigned in a parent class? I wouldn't think so but I have been surprised by Python before.

TheRealFalcon: Hmmm, there was one other change that I thought I requested, but I'm not seeing now. These other changes won't make sense without it. It was to change `ShellScriptByFreqPartHandler` to look like this:
```python
class ShellScriptByFreqPartHandler(Handler):
    def __init__(self, paths, **_kwargs):
        Handler.__init__(self, self.frequency)
```

Looking at the `Handler` base class now, I can see how this would be a little confusing given that there is also a `frequency` instance variable there.

Since I'm proposing we define class variables named `frequency` in the subclasses, the call to that `Handler.__init__(self, self.frequency)` is passing a reference to the class variable named frequency which then sets the instance variable named `frequency`.

If that's confusing, I'd be ok with changing the names of the class variables (along with the corresponding `Handler.__init__` call) to something like `freq` so it's not so easy to confuse class variable names with instance variable names.

Does that make sense?

cloudinit/handlers/shell_script_by_frequency.py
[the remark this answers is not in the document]
TheRealFalcon: Thanks for the feedback. It is silly for us to have that `#vi` comment in there (I'll put up a PR to get that removed). The `# part-handler` is valid for user-defined part handlers, but isn't necessary for this file because it is no longer a user-defined part handler, but becoming a built-in one.

cloudinit/handlers/shell_script_by_frequency.py
beantaxi: Ahhhh I suspected my mediocre Python skills played a role here. You were requesting class variables in the subclasses, not instance variables, although there happen to be instance variables of the same name.

The whole `frequency` handling is confusing to me, since `Handler` takes a single frequency in its c'tor and then maintains it as in ivar, but then `handle_part()` takes a frequency as an argument as well.

Meanwhile had this idea: I just checked in `stages` where these handlers actually get created, and it looks like this:

```
        shellscript_per_boot_handler = ShellScriptPerBootPartHandler(**opts)
        shellscript_per_instance_handler = \
            ShellScriptPerInstancePartHandler(**opts)
        shellscript_per_once_handler = ShellScriptPerOncePartHandler(**opts)
```

In theory, those could all be c'tor invocations of a single class, parameterized on `frequency`, and then the subclasses can go away entirely. There'd still be the apparent redundancy of frequency as an arg to `handle_parts()` but that's ok.

TheRealFalcon: Ah, right. Your proposed solution makes sense and sounds like a good idea. Go for it!

beantaxi: Thanks! Just committed. I had to add a prefixes collection to the base class, of all 3 prefixes, but that's the only part I was a bit unsure about.

TheRealFalcon:
@beantaxi , this PR looks good to me at this point. One additional thing I'd like to have before we merge it is an integration test. The docs for our integration tests are at https://cloudinit.readthedocs.io/en/latest/topics/integration_tests.html , and the tests themselves can be found at `tests/integration_tests`.

I've got the beginnings of one here:
https://paste.ubuntu.com/p/F8Gqhnc5bC/

The test itself should be verifying that the scripts exist on the instance (via assert statements...not just printing stuff), and that they've all executed at least once. I don't actually think we need to verify the 'per-boot, per-instance, per-once' functionality as that is testing the behavior of those modules, not this new code.

My one major hangup is that the make-mime code prints directly to the console rather than returning any sort of text anywhere, so my `USER_DATA = handle_args('make-mime', args)` doesn't actually give us anything useful. That's...annoying. I think the `make_mime.py` could be refactored a bit to accommodate this.

I unfortunately have some higher priority work happening right now, so I may not be able to do any additional work on it in the next few days. If you'd like to poke at it, feel free, but I also realize it's a lot to digest for somebody new to this code base and these tests

beantaxi:
@TheRealFalcon I'm in. A couple questions on your last comment and the paste ...

"The test itself should be verifying ... that they've [the freq-specific scripts] all executed at least once" 

I am inferring that you'd be happy with something as simple as sample userdata scripts that each write a file, eg `/tmp/boot-worked.txt`, `/tmp/instance-worked.txt`, `/tmp/once-worked.txt`, and then tests that assert those files got created. Anything that indicates scripts ran. If you meant something, else, feel free to let me know.

As far as the paste goes, it looks like there is a fair amount of module-level code that creates `USER_DATA` which then gets 'injected' into downstream test code via Magic (aka `@pytest.mark`). When I do this for real, would all that be in a fixture? Or still module level. Thanks!

TheRealFalcon:
> I am inferring that you'd be happy with something as simple as sample userdata scripts that each write a file, eg /tmp/boot-worked.txt, /tmp/instance-worked.txt, /tmp/once-worked.txt, and then tests that assert those files got created. Anything that indicates scripts ran. If you meant something, else, feel free to let me know.

Yes, that plus verifying that the "per-boot, per-instance, per-once" scripts themselves are put in their proper places in /var/lib/cloud/scripts. What you have posted would implicitly test that behavior, but better to be explicit.

> As far as the paste goes, it looks like there is a fair amount of module-level code that creates USER_DATA which then gets 'injected' into downstream test code via Magic (aka @pytest.mark). When I do this for real, would all that be in a fixture? Or still module level. Thanks!

Either one could work, but module-level would be easiest here. I know it is sometimes frowned upon to do so much at the module level, but given that this is test code and tests are more independent than a large running program, I don't think it is a problem. The reason it is easier to do it this way rather than using a fixture is that the user data has to be defined before the test definition in order to use the `@pytest.mark.user_data` mark. Without that, you can't really use the `client` fixture either and would have to go the route of launching and instance manually after you've created your user data. [We have tests that do that](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/bugs/test_gh671.py#L27), but it's not as straightforward as just using the client fixture.

beantaxi:
"My one major hangup is that the make-mime code prints directly to the console rather than returning any sort of text anywhere ..." (etc)

After reading the code, I took this to mean that your `make_mime.handle_args()` call as written should be treated as pseudocode, since it does not actually return `USER_DATA`, or anything other than a status code. But I also took your broader point to be that it would be nice if there were a way to actually capture the work produced by `make_mime.handle_args()`, instead of just letting `handle_args()` dump to the console and return 0 or 1.

To that end I extracted a `make_mime.create_user_data()` function out of `handle_args()`, such that `handle_args()` calls `create_user_data()`, but an integration test could also call `create_user_data()` as well to set `USER_DATA` etc. But all that left me with some questions:

- Are my inferences correct?
- If so, is there really no existing function that simple creates userdata? Or am I duplicating an existing function.
- If there is no current `create_user_data()` function, then how is userdata creation actually getting tested? Is it only tested indirectly via various integration tests?

Thanks!

beantaxi:
Pending the answers to my previous comment, I _think_ I'm ready to go. I spent many hours reading about `pytest` & pytest fixtures, which I won't pretend fully understand, and the use of `pytest` in `cloud-init`, which I also won't pretend to fully understand. And a bit of `pycloudlib` which I look forward to spending more time with. But I think all I have to do is to create a few files, package them into userdata, let Test Fixture Magic fire up a new cloudinit VM with the userdata, and then use `client.execute()` to interact with the new VM, eg to check if files exist.

The one thing I can think of, is that it would be big help to be able to do `client.execute()`s in isolation, or even interactively, to let me get a feel for it and to come up with the right tests. Eg I don't know how I'd test for a file's existence with `client.execute()`, and iteratively running `pytest tests/integration_tests mytest.py` would not be very efficient. Unfortunately, despite numerous dive attempts, I have yet to comprehend exactly how the test fixtures are creating the VM and returning a reference, so I am currently limited to doing any `client.execute()` exploration from within my integration test. If that's how it is, I can manage, but if not, I figured it was worth asking.

Thanks!

TheRealFalcon:
> After reading the code, I took this to mean that your make_mime.handle_args() call as written should be treated as pseudocode, since it does not actually return USER_DATA, or anything other than a status code. But I also took your broader point to be that it would be nice if there were a way to actually capture the work produced by make_mime.handle_args(), instead of just letting handle_args() dump to the console and return 0 or 1.

Yes, that's correct.

> To that end I extracted a make_mime.create_user_data()

It'd be slightly more accurate to call it something like `make_mime_message`. It will be used as the user data for that particular integration test, but within the context of the `make_mime.py` file, it's just a mime message.

> Are my inferences correct?

I think so, but it's hard to understand completely without seeing the changes yet. If you want to push what you have so far, it'd be easier to understand exactly what you're doing.

> If so, is there really no existing function that simple creates userdata? Or am I duplicating an existing function.

I don't think that you're duplicating an existing function. The userdata is technically always provided by the user, so cloud-init usually isn't involved in the creating of that data. What we have here is a simple tool to create a mime message that can be provided as user data, but the vast majority of our use cases don't use this sort of thing, but rather provide a single shell script or  cloud-config. 

> then how is userdata creation actually getting tested?

With this integration test :slightly_smiling_face: . There's certainly gaps in our testing in the code base, and especially for a convenience tool like this which won't break boot like a bug in cloud-init proper would. Going forward though, we're trying to ensure all changes get properly tested, hence my asking for an integration test here.

> The one thing I can think of, is that it would be big help to be able to do client.execute()s in isolation, or even interactively, to let me get a feel for it and to come up with the right tests. Eg I don't know how I'd test for a file's existence with client.execute(), and iteratively running pytest tests/integration_tests mytest.py would not be very efficient. Unfortunately, despite numerous dive attempts, I have yet to comprehend exactly how the test fixtures are creating the VM and returning a reference, so I am currently limited to doing any client.execute() exploration from within my integration test. If that's how it is, I can manage, but if not, I figured it was worth asking.

That's some good feedback. Being able to interact with an already running instance would be useful for debugging in a lot of cases. But just FYI, the `client.execute(...)` should more-or-less act the same as executing commands in a root shell in the running instance. So for something like `Eg I don't know how I'd test for a file's existence with client.execute()`, the answer is "however you would do that in bash". There's a separate apt-related integration test that verifies that a file **doesn't** exist like so:
```python
conf_exists = class_client.execute(
    'test -f /etc/apt/apt.conf.d/90cloud-init-pipelining'
).ok
assert conf_exists is False
```

beantaxi:
Thanks for the response, especially the color around how at the end of the day `make_mime` is a `cloud-init devel` feature that serves edge cases, as opposed to being the main use case of single-script or cloud-config initialization. So it's quite natural for may be a gap here and there to surface from time to time. Just as long as I'm not rewriting existing tests.

Thanks also for the detail on `client.execute()` invocation, which is very helpful. I do still have a 'wishlist' of a handful of calls I'd like to see the results of, but I can get those just by batching up a bunch of calls in my initial integration test(s).

I'll get to this later today or tomorrow, & then it sounds like it might be good to go from my end.

chrislalos:
Real quickly ... I got around to creating and running an integration test, and it appears that my tests are failing; my scripts didn't get copied to the proper scripts directory, and they didn't execute.

I can't think of a next step to testing, other than installing a my-branch version of cloud-init on a VM and seeing what happens. I can think of 3 ways of doing so ...

1. Studying `pycloudlib` and using `pycloudlib` to do the heavy lifting
2. Studying existing VMs (eg EC2 instances), and seeing if I can easily monkeypatch my own cloud-init code changes changes onto an existing VM and going from there
3. Something else that's even better than someone here can point out, eg an lxd trick to specify a cloud-init commit hash to be used on VM creation.

All 3 have value to me, so in the absence of feedback I'll probably just start with #1 and see what happens. Thanks!

TheRealFalcon:
Hi @chrislalos . Are you working with @beantaxi on this?

Broadly, when it comes to debugging integration tests, I'll usually ensure the following are set in my `tests/integration_tests/user_settings.py`:
```
PLATFORM = 'lxd_container'
CLOUD_INIT_SOURCE = 'IN_PLACE'
KEEP_INSTANCE = True
```

This allows me to connect after a test has completed to see what happened. If that doesn't help, it's mostly print/logging statements and examining the logs. Be sure to set `KEEP_INSTANCE` back to `False` when you're done to ensure you don't leave a bunch of containers/VMs lying around after any future test runs.

For this issue in particular, `create_mime_message` is always returning failure. I think the latest commit meant to include `rc` in the return tuple rather than 1. Also, since it's returning a tuple now, that won't work as USERDATA anymore in the integration test.

How about something like this for the refactored `create_mime_message`:
```python
def create_mime_message(files):
    sub_messages = []
    errors = []
    for i, (fh, filename, format_type) in enumerate(files):
        contents = fh.read()
        sub_message = MIMEText(contents, format_type, sys.getdefaultencoding())
        sub_message.add_header('Content-Disposition',
                               'attachment; filename="%s"' % (filename))
        content_type = sub_message.get_content_type().lower()
        if content_type not in get_content_types():
            msg = ("content type %r for attachment %s "
                   "may be incorrect!") % (content_type, i + 1)
            errors.append(msg)
        sub_messages.append(sub_message)
    combined_message = MIMEMultipart()
    for msg in sub_messages:
        combined_message.attach(msg)
    return (combined_message, errors)
```
I like this because it doesn't write to stdout/stderr, so it can be called from other code without having side effects that assume we're running from the CLI. It also takes the `files` argument directly so we don't need to worry about needing an argparse object to use it. 

Then it can be called further down with something like this:
```python
    combined_message, errors = create_mime_message(args.files)
    if errors:
        level = "WARNING" if args.force else "ERROR"
        for error in errors:
            sys.stderr.write("{}: {}\n".format(level, error))
        sys.stderr.write("Invalid content-types, override with --force\n")
        if not args.force:
            return 1
    print(combined_message)
    return 0
```

Then using it in an integration test becomes much simpler too:
```python
FILES = [
    (PER_BOOT_FILE, 'boot.sh', 'x-shellscript-per-boot'),
    (PER_INSTANCE_FILE, 'instance.sh', 'x-shellscript-per-instance'),
    (PER_ONCE_FILE, 'once.sh', 'x-shellscript-per-once'),
]

USER_DATA, _ = create_mime_message(FILES)
```

(Note that I haven't thoroughly tested these code snippets, so feel free to correct something I missed)

One thing I did notice when trying this is that all of the scripts are ending up in the `per-boot` directory instead of their respective `per-boot,per-instance,per-once` directories. I'm not sure why, but that's one last thing we'd need to fix.

chrislalos:
May I request a reopen please? It's been a while, but seeing as this was a New Years Resolution for 2021, and I hadn't finished it, I figured I'd take a look last night / this morning for old time's sake ... and to my surprise I was able to get my integration test passing this morning!

TheRealFalcon:
@chrislalos , this PR cannot be re-opened as we no longer have a `master` branch (it was renamed to `main`). Feel free to open a new PR referencing this one.

PR #753, opened 2021-01-06
Missing IPV6_AUTOCONF=no to render sysconfig dhcp6 stateful on RHEL

otubo:
## Proposed Commit Message
IPV6_AUTOCONF needs to be set to 'no' on RHEL so NetworkManager can
properly acquire ipv6 address.

rhbz: #1859695

## Additional Context
According to additional tests from the reporter Tom Barron on the Original Bugzilla ([1859695](https://bugzilla.redhat.com/show_bug.cgi?id=1859695)) the option IPV6_AUTOCONF should also be set to `no` on this specific case. Most of the BZ is marked as private, though.

## Test Steps
(as described on the above mentioned BZ)
Steps to Reproduce:

1. Create StorageNFS network with IPv6 CIDR and address range defined
1. Create a Nova instance with 2 NICs, 2nd NIC connected to StorageNFS network
1. Login to VM and examin Eth1 IP address

Actual results:
The Nova assigned IPv6 address was not assigned to device Eth1

Expected results:
Expected the Nova assigned IP address to be assigned to Eth1 interface

## Checklist:

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -402,6 +402,7 @@ class Renderer(renderer.Renderer):
                     iface_cfg['BOOTPROTO'] = 'dhcp'
                     iface_cfg['DHCPV6C'] = True
                     iface_cfg['IPV6INIT'] = True
+                    iface_cfg['IPV6_AUTOCONF'] = False
                 else:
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Thanks!

PR #760, opened 2021-01-11
Fix regression by forcing no IMDS ssh keys

trstringer:
## Proposed Commit Message

Fix regression with handling of IMDS ssh keys.

```
With the changes for ssh public keys to be retrieved from IMDS as a first option, when a key is passed through not in the raw ssh public key format it causes an issue and the key is not added to the user's authorized_keys file.

This PR will temporarily disable this behavior until a permanent fix is put in place.
```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -654,6 +654,7 @@ class DataSourceAzure(sources.DataSource):
             raise KeyError(
                 "Not using public SSH keys from IMDS"
             )
+            # pylint:disable=unreachable
             ssh_keys = [
                 public_key['keyData']
                 for public_key
@@ -1278,6 +1279,7 @@ class DataSourceAzure(sources.DataSource):
                 raise KeyError(
                     "Not using public SSH keys from IMDS"
                 )
+                # pylint:disable=unreachable
                 public_keys = self.metadata['imds']['compute']['publicKeys']
                 LOG.debug(
                     'Successfully retrieved %s key(s) from IMDS',
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -651,6 +651,9 @@ class DataSourceAzure(sources.DataSource):
         LOG.debug('Retrieving public SSH keys')
         ssh_keys = []
         try:
+            raise KeyError(
+                "Not using public SSH keys from IMDS"
+            )
             ssh_keys = [
                 public_key['keyData']
                 for public_key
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -651,6 +651,9 @@ def get_public_ssh_keys(self):
         LOG.debug('Retrieving public SSH keys')
         ssh_keys = []
         try:
+            raise KeyError(
+                "Not using public SSH keys from IMDS"
```
OddBloke, line 655: This message isn't used in the exception handler, so neither the log nor the diagnostic event will indicate the reason for the failure (i.e. it will just log 'Unable to get keys from IMDS, falling back to OVF'). Is the message just here as documentation when reading the code, or was the expectation that it would be exposed outside cloud-init somewhere?

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1272,6 +1275,9 @@ def _negotiate(self):
 
             pubkey_info = None
             try:
+                raise KeyError(
+                    "Not using public SSH keys from IMDS"
```
OddBloke, line 1279: Same Q here as above.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1272,6 +1275,9 @@ def _negotiate(self):
...
+                )
                 public_keys = self.metadata['imds']['compute']['publicKeys']
```
OddBloke, line 1281: This is something of an aside (which I noticed when checking if `public_keys` being unset would be an issue), but it doesn't look like `public_keys` was used anywhere other than logging; is that a bug in the original implementation, or expected?

OddBloke (commented):
Thanks for this fix, Thomas! We'll need to discuss how to remediate this problem in a longer-term fashion, but this fix addresses our immediate need.

I have one important question inline: I want to make sure that the logging/diagnostics are aligned with your expectations. (I also have a side note, which doesn't really need addressing here.)

cloudinit/sources/DataSourceAzure.py
trstringer: Exactly. I put the message in there as self-documenting code in lieu of an actual comment. No expectations of this string being exposed anywhere outside of source code.

cloudinit/sources/DataSourceAzure.py
trstringer: Answered above!

cloudinit/sources/DataSourceAzure.py
trstringer: That's expected that `public_keys` isn't used. That code path is just to see if the public keys are available from IMDS, otherwise if they are not than the `except` clause would need to populate a different var `pubkey_info` for further processing. Also `public_keys` is used in logging.

OddBloke:
Thanks for the clarifications! I'm happy with this code. I'm now performing some testing locally, I will approve once I've confirmed this works.

OddBloke (approved):
I've confirmed that this fixes the issue using the test that I have just proposed for review in #761 (your review over there would be appreciated!). Thanks!

trstringer:
Thanks, @OddBloke!

PR #771, opened 2021-01-12
cc_seed_random: update documentation and fix integration test

OddBloke:
## Proposed Commit Message
```
cc_seed_random: update documentation and fix integration test

The documentation did not mention that the given data may not be the
exact string written: the cloud's random data may be added to it.
Additionally, the documentation of the command key was incorrect.

test_seed_random_data was updated to check that the given data is a
prefix of the written data, to match cloud-init's expected (and, now,
documented) behaviour.

LP: #1911227
```

## Test Steps

Run `CLOUD_INIT_PLATFORM=azure pytest tests/integration_tests/modules/test_seed_random_data.py` and observe a pass (vs. a fail previously).

## Checklist:

### DIFF
--- a/cloudinit/config/cc_seed_random.py
+++ b/cloudinit/config/cc_seed_random.py
@@ -24,15 +24,19 @@ Configuration for this module is under the ``random_seed`` config key. The
 optionally be specified in encoded form, with the encoding specified in
 ``encoding``.
 
+If the cloud provides its own random seed data, it will be appended to ``data``
+before it is written to ``file``.
+
 .. note::
     when using a multiline value for ``data`` or specifying binary data, be
     sure to follow yaml syntax and use the ``|`` and ``!binary`` yaml format
     specifiers when appropriate
 
-Instead of specifying a data string, a command can be run to generate/collect
-the data to be written. The command should be specified as a list of args in
-the ``command`` key. If a command is specified that cannot be run, no error
-will be reported unless ``command_required`` is set to true.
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_seed_random.py
```diff
@@ -24,15 +24,19 @@
...
 .. note::
     when using a multiline value for ``data`` or specifying binary data, be
     sure to follow yaml syntax and use the ``|`` and ``!binary`` yaml format
     specifiers when appropriate
 
-Instead of specifying a data string, a command can be run to generate/collect
-the data to be written. The command should be specified as a list of args in
-the ``command`` key. If a command is specified that cannot be run, no error
-will be reported unless ``command_required`` is set to true.
+A command can be specified (which will run after any data is written). The
```
smoser, line 35: (the parens are not necessary here).  It also wasn't completely clear to me what or where "data was written".  That is possibly because I was just reading this as diff output, so maybe I lost context.

Maybe:

    If a ``command`` key is present, the command will be executed after ``file`` has been populated.
    The command's environment will contain the value of ``file`` specified above in the key 
    ``RANDOM_SEED_FILE``. If a command is specified that cannot be run, no error will 
    be reported unless ``command_required`` is set to true.

smoser (approved):
I'm good with this as it is or with any amount of my feedback taken.  (But I do think the paren are not necessary :wink:).

smoser (approved):

PR #779, opened 2021-01-20
New upstream snapshot to devel release

OddBloke:
Including the fix we SRU'd earlier in the week.


ubuntu-server-builder (approved):
Confirmed:
- negligible diff: debian/changelog uploader
- build-package and sbuild successes

PR #780, opened 2021-01-20
Parse and handle SSH public keys in different formats

trstringer:
## Proposed Commit Message

If SSH public keys from IMDS are in the OpenSSH format, then consume them directly. Otherwise, utilize the OVF path to retrieve the OpenSSH-formatted public keys.

## Additional Context

This is the long-term fix for the regression that was fixed with the temporary/short-term PR https://github.com/canonical/cloud-init/pull/760

## Checklist:


cloudinit/sources/DataSourceAzure.py
```diff
@@ -1275,24 +1293,36 @@ def _negotiate(self):
...
+                report_diagnostic_event(log_msg, logger_func=LOG.debug)
+
+            if not keys_from_imds:
                 pubkey_info = self.cfg.get('_pubkeys', None)
+                log_msg = 'Retrieved {} keys from OVF'.format(
+                    len(pubkey_info)
+                    if pubkey_info is not None
+                    else 0
+                )
+                report_diagnostic_event(log_msg, logger_func=LOG.debug)
```
OddBloke, line 1325: This logic all looks quite repetitious; is there something we could do to reduce the duplication with `get_public_ssh_keys`?

OddBloke (commented):
Hey Thomas, a couple of high-level observations here; I haven't had a chance to dig in in more detail yet.

cloudinit/sources/DataSourceAzure.py
trstringer: Done! And good point. I refactored some of this logic out for deduplication.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1355,6 +1385,15 @@ def region(self):
         return self.metadata.get('imds', {}).get('compute', {}).get('location')
 
 
+def _key_is_openssh_formatted(key):
+    """
+    Validate whether or not the key is OpenSSH-formatted. This
+    is done by looking for a prefix of "ssh-".
```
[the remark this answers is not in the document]
trstringer: This should suffice at the moment as the keys in OpenSSH format are almost all entirely "ssh-rsa" (and this is [what is documented](https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/createorupdate#sshpublickey)). We should be safe to move forward with this.

trstringer:
@OddBloke, let me know if there's anything else you want to see changed in this PR! // @mitechie

trstringer:
Can I get this PR re-opened? @OddBloke @mitechie

OddBloke:
Yep, and let me remove that tag too to stop it happening again.

cloudinit/sources/DataSourceAzure.py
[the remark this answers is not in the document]
smoser: > I don't have a physical Yubikey or similar to check with, but my reading of https://github.com/openssh/openssh-portable/blob/master/sshkey.c also suggests that "SK" keys generated using one will also not match an `ssh-` prefix.

See the reference to sshkey.c in cloudinit/ssh_util.py.  It seems that it might make more sense to pass it to AuthKeyLineParser:

```python
import ssh_util


def _key_is_openssh_formatted(key):
    parser = ssh_util.AuthKeyLineParser()
    try:
        akl = parser.parse(key)
    except TypeError:
        return False

    return akl.keytype is not None
```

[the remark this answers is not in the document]
trstringer: That's a good idea, thanks! I made these code changes!

cloudinit/sources/DataSourceAzure.py
```diff
@@ -719,30 +727,52 @@ def get_public_ssh_keys(self):
...
             report_diagnostic_event(log_msg, logger_func=LOG.debug)
+
+        if not keys_from_imds:
             try:
                 ssh_keys = self.metadata['public-keys']
-                LOG.debug('Retrieved keys from OVF')
+                log_msg = 'Retrieved {} keys from OVF'.format(
+                    len(ssh_keys)
+                    if ssh_keys is not None
+                    else 0
```
smoser, line 765: i'm almost certain that ssh_keys has to be an array at this point. so the `if` here is not necessary.

trstringer: Good point, thanks!

trstringer:
@OddBloke, I've implemented @smoser's suggestions and this PR is ready for review. Thanks!

cloudinit/sources/DataSourceAzure.py
```diff
@@ -719,30 +727,48 @@ def get_public_ssh_keys(self):
...
                 ssh_keys = self.metadata['public-keys']
-                LOG.debug('Retrieved keys from OVF')
+                log_msg = 'Retrieved {} keys from OVF'.format(len(ssh_keys))
             except KeyError:
                 log_msg = 'No keys available from OVF'
+            finally:
                 report_diagnostic_event(log_msg, logger_func=LOG.debug)
 
-        return ssh_keys
+        return dict(
```
OddBloke, line 768: A nit, but using a dict here feels a little off: this is really an immutable 2-tuple. I appreciate that using a dict allows us to use names (instead of indexing) for these: a more Pythonic way of achieving the same would be to use a `namedtuple`:

```py
In [8]: keys_from_imds = False

In [9]: ssh_keys = ["ssh-rsa definitelyavalidkey", "ssh-dsa thisonesbadthough"]

In [10]: return_value = namedtuple("AzureSSHKeys", ("keys_from_imds", "ssh_keys"))(keys_from_imds, ssh_keys)

In [11]: return_value.keys_from_imds
Out[11]: False

In [12]: return_value.ssh_keys
Out[12]: ['ssh-rsa definitelyavalidkey', 'ssh-dsa thisonesbadthough']
```

The namedtuple definition should probably go at module level, though.

(You could also consider returning a `(2, tuple)` and consistently deconstructing it with `keys_from_imds, ssh_keys = self._get_public_ssh_keys_and_source()`.)

OddBloke (approved):
Thanks, Thomas, this looks good to me! I have one proposed improvement inline, but that's a nit so I think this is fine to land without it.

What I would like to see is an integration test! I know that you have other work contingent on this landing, so we can do that as a follow-up PR to make life easier. This will likely require a little framework modification, but I believe we have the broad support for it in place. I've just filed https://bugs.launchpad.net/cloud-init/+bug/1922801, I'll expand over there.

cloudinit/sources/DataSourceAzure.py
trstringer: Good idea! Change made (and tested).

trstringer:
Closing this PR as its code was merged in https://github.com/canonical/cloud-init/pull/865.

PR #782, opened 2021-01-21
[VMware]:disable network config if no network provided in metadata

xiaofengw-vmware:
summary: disable network config if no network provided in metadata

vmware vSphere users want to re-customize the VM without changing network,
which means the network could be absent in metadata.  In this case,  we need to
set the network config to disabled in datasource.

## Checklist:

 - [X ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [X ] I have updated or added any unit tests accordingly

### DIFF
--- a/cloudinit/sources/DataSourceOVF.py
+++ b/cloudinit/sources/DataSourceOVF.py
@@ -172,9 +172,8 @@ class DataSourceOVF(sources.DataSource):
                 if network:
                     self._network_config = network
                 else:
-                    self._network_config = (
-                        self.distro.generate_fallback_config()
-                    )
+                    LOG.debug("Network not found, disable network config.")
+                    self._network_config = {'config': 'disabled'}
 
             except safeyaml.YAMLError as e:
                 _raise_error_status(
--- a/tests/unittests/test_datasource/test_ovf.py
+++ b/tests/unittests/test_datasource/test_ovf.py
@@ -631,6 +631,55 @@ class TestDatasourceOVF(CiTestCase):
 
         self.assertIn('is not found', str(context.exception))
 
+    def test_get_data_cloudinit_metadata_no_network(self):
+        """Test network config disabled if no network in meta data
+        """
+        paths = Paths({'cloud_dir': self.tdir})
+        ds = self.datasource(
+            sys_cfg={'disable_vmware_customization': False}, distro={},
+            paths=paths)
+
+        # Prepare the conf file
+        conf_file = self.tmp_path('test-cust', self.tdir)
+        conf_content = dedent("""\
+            [CLOUDINIT]
+            METADATA = test-meta
+            USERDATA = test-user
+            """)
+        util.write_file(conf_file, conf_content)
+
+        # Prepare the meta data file
+        metadata_file = self.tmp_path('test-meta', self.tdir)
+        # no network in meta data
+        metadata_content = dedent("""\
+            instance-id: cloud-vm
+            local-hostname: my-host.domain.com
+            """)
+        util.write_file(metadata_file, metadata_content)
+
+        # Prepare the user data file
+        userdata_file = self.tmp_path('test-user', self.tdir)
+        userdata_content = "This is the user data"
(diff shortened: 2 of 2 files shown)


xiaofengw-vmware:
Hi all,
Please help to review when you have time. Thanks a lot.

TheRealFalcon:
Hi @xiaofengw-vmware , Can I get a little more context for this change? What does it mean to re-customize the VM without changing the network? Why does this require that the network be disabled? Does this only apply to the first boot or to subsequent boots?

In general, I'm very wary of simply disabling the network completely, because it could potentially cut off all network access (and ability to reach the VM) for people that may be currently relying on the fallback functionality. Is there a reason `{'cloud': 'disabled'}` can't be passed as the network metadata for the use case you're considering?

xiaofengw-vmware:
> Hi @xiaofengw-vmware , Can I get a little more context for this change? What does it mean to re-customize the VM without changing the network? Why does this require that the network be disabled? Does this only apply to the first boot or to subsequent boots?
> 
> In general, I'm very wary of simply disabling the network completely, because it could potentially cut off all network access (and ability to reach the VM) for people that may be currently relying on the fallback functionality. Is there a reason `{'cloud': 'disabled'}` can't be passed as the network metadata for the use case you're considering?

Hi @TheRealFalcon ,
Regarding to re-customize VM without changing the network, I could give you an user case. VMware vSphere customer could configure the VM with network when he deploy the VM. Later he wants to apply a script (user data) to this VM without changing the network. Then he could re-customize it with an empty meta data and a valid user data. In this case, we want to set the network-config=disabled so that cloud-init won't configure network. That's our design purpose.
Passing `{'cloud': 'disabled'}` as the network metadata is also a choice, but it need customer to be aware of how to disable it in metadata. 
I agree that it will have risk to cut off all network access. But I think it's customer's choice to not changing the network.  They could re-customize the network if there is network problem. How do you think? Thanks

xiaofengw-vmware:
Ping, please help to review. Thanks a lot.

TheRealFalcon:
@OddBloke , Do you mind taking a look at this one? I'm not sure how valid my concerns are.

OddBloke:
The most obvious concern I have is first boot. Currently, if no network data is configured for a VMWare host taking the `if use_raw_data` codepath, fallback networking configuration is generated: this configuration configures the system to DHCP on a single interface. With the proposed change, I believe that if no network data is configured for such a host, we will instead disable cloud-init's network generation: this will mean that on first boot, instances will come up with no configured networking (and therefore, of course, will not be accessible over the network). I don't think this is behaviour we want.

> vmware vSphere users want to re-customize the VM without changing network

I'd like to understand this a little better: can you explain the technical details of how re-customisation happens? I don't quite understand how we end up re-running some of these code paths, and I think understanding that would help me here.

Thanks!

xiaofengw-vmware:
> The most obvious concern I have is first boot. Currently, if no network data is configured for a VMWare host taking the `if use_raw_data` codepath, fallback networking configuration is generated: this configuration configures the system to DHCP on a single interface. With the proposed change, I believe that if no network data is configured for such a host, we will instead disable cloud-init's network generation: this will mean that on first boot, instances will come up with no configured networking (and therefore, of course, will not be accessible over the network). I don't think this is behaviour we want.
> 
> > vmware vSphere users want to re-customize the VM without changing network
> 
> I'd like to understand this a little better: can you explain the technical details of how re-customisation happens? I don't quite understand how we end up re-running some of these code paths, and I think understanding that would help me here.
> 
> Thanks!

Hi Daniel,
Thanks for your comments. I have discussed this topic with our product manager, we have agreed that we will not deliver this fix. That means we will set the network as dhcp if no network provided in meta data. If user wants to re-customize this VM without changing the network, the user needs to disable the network in their meta data.

Re-customize is a vSphere feature. After the VM is deployed, the vSphere administrator still could customize the VM later. For example, the administrator could change the ip configuration, launch a script etc.

PR #783, opened 2021-01-22
includedir in suoders can be prefixed by "arroba"

jordimassaguerpla:
## Proposed Commit Message

```
Since version 1.9.1, @includedir can be used in the sudoers files
instead of #includedir:

https://github.com/sudo-project/sudo/releases/tag/SUDO_1_9_1

This happens to be the default in SUSE Linux enterprise sudoer package,
so cloudinit should take this into account.

Otherwise, cloudinit was adding an extra #includedir, which was
resulting on the files under /etc/sudoers.d being included twice, one by
@includedir from the SUSE package, one by the @includedir from
cloudinit. The consequence of this, was that if you were defining an
Cmnd_Alias inside any of those files, this was being defined twice and
creating an error when using sudo.

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>```

## Additional Context
<!-- If relevant -->

## Test Steps
<!-- Please include any steps necessary to verify (and reproduce if
this is a bug fix) this change on a live deployed system,
including any necessary configuration files, user-data,
setup, and teardown. Scripts used may be attached directly to this PR. -->

Add "@includedir /etc/sudoers" into /etc/sudoers

Apply cloud-init and see you have both, "@includedir /etc/sudoers" and "#includedir /etc/sudoers".

After the fix, you should only have "@includedir"

## Checklist:
that apply. -->

### CODE REVIEW

**Comment by jordimassaguerpla** [comment]
Hi. I already signed the individual contributors agreement. Could you retry the github action?

**Comment by jordimassaguerpla** [comment]
I was missing https://github.com/canonical/cloud-init/pull/787

**Review by TheRealFalcon — CHANGES_REQUESTED** [review]
@jordimassaguerpla Thanks for this. The change looks good, but can you add a unit test or two as well? It would look similar and go in the same place as the one I'm linking here, but ensure that when a line containing the # or @ include is encountered, that it correctly keeps the line rather than appending a new one.
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_distros/test_generic.py#L109

File: tests/unittests/test_distros/test_generic.py [hunk 1]
```diff
@@ -119,6 +119,18 @@ def test_sudoers_ensure_append(self):
...
+        cls = distros.fetch("ubuntu")
+        d = cls("ubuntu", {}, None)
+        self.patchOS(self.tmp)
+        self.patchUtils(self.tmp)
+        util.write_file("/etc/sudoers", "@includedir /b")
+        d.ensure_sudo_dir("/b")
+        contents = util.load_file("/etc/sudoers")
+        self.assertIn("includedir /b", contents)
+        self.assertTrue(os.path.isdir("/b"))
+        self.assertEqual(1, contents.count("includedir /b"))
```
TheRealFalcon, line 132:
```suggestion
        for char in ['#', '@']:
            util.write_file("/etc/sudoers", "{}includedir /b".format(char))
            d.ensure_sudo_dir("/b")
            contents = util.load_file("/etc/sudoers")
            self.assertIn("includedir /b", contents)
            self.assertTrue(os.path.isdir("/b"))
            self.assertEqual(1, contents.count("includedir /b"))
```

TheRealFalcon (commented):
Thanks for the test. I left one small suggestion just to make sure we're hitting both include characters. After that, we should be good to merge!

TheRealFalcon (approved):
Looks good. Thanks!

PR #785, opened 2021-01-23
VMware: Fix default route configuration when no primary NIC exists

ntoofu:
## Proposed Commit Message

```
VMware: Fix default route configuration when no primary NIC exists

Because a subnet which gateway belongs to is usually treated
as a "directly connected" on the host, adding a route for the subnet has
no effect.

Therefore, `0.0.0.0/0` should be used as destinations of routes
associated with gateways of non-primary NICs instead.
```

## Test Steps

I verified in this way:

1. Deploy Ubuntu 20.04 virtual machine
2. Add `disable_vmware_customization: false` to `cloud.cfg` in the VM
3. Clone the VM with VMware guest customization by using [Ansible module](https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_guest_module.html)
4. Check the route table in a cloned VM

### Playbook used in step 3

```
- hosts: all
  gather_facts: no
  tasks:
  - community.vmware.vmware_guest:
      hostname: "{{ vcenter_hostname }}"
      username: "{{ vcenter_username }}"
      password: "{{ vcenter_password }}"
      validate_certs: no
      datacenter: Datacenter
      folder: /test
      name: testvm
      state: poweredon
      template: ubuntu-20.04-template
      disk:
      - type: thin
        datastore: datastore
        size_gb: 30
      hardware:
        memory_mb: 1024
        num_cpus: 2
        num_cpu_cores_per_socket: 1
        scsi: paravirtual
      networks:
      - name: VM Network
        ip: 192.168.0.11
        netmask: 255.255.255.0
        gateway: 192.168.0.1
        wait_for_ip_address: true
    delegate_to: localhost
```

### Network settings generated without this patch

```
root@testvm:~# ip addr show
    inet6 ::1/128 scope host
2: ens160: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:50:56:ba:86:41 brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.11/24 brd 192.168.0.255 scope global ens160
    inet6 fe80::250:56ff:feba:8641/64 scope link

root@testvm:~# ip route show
192.168.0.0/24 dev ens160 proto kernel scope link src 192.168.0.11
192.168.0.0/24 via 192.168.0.1 dev ens160 proto static metric 10000
```

No default route is generated.

### Network settings generated with this patch

```
    inet6 ::1/128 scope host
2: ens160: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:50:56:ba:db:32 brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.11/24 brd 192.168.0.255 scope global ens160
    inet6 fe80::250:56ff:feba:db32/64 scope link

root@testvm:~# ip route show
default via 192.168.0.1 dev ens160 proto static metric 10000
192.168.0.0/24 dev ens160 proto kernel scope link src 192.168.0.11
```

A default route is generated as expected.

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vmware/imc/config_nic.py
+++ b/cloudinit/sources/helpers/vmware/imc/config_nic.py
@@ -9,27 +9,10 @@ import logging
 import os
 import re
 
-from cloudinit.net.network_state import mask_to_net_prefix
 from cloudinit import subp
 from cloudinit import util
 
 logger = logging.getLogger(__name__)
-
-
-def gen_subnet(ip, netmask):
-    """
-    Return the subnet for a given ip address and a netmask
-    @return (str): the subnet
-    @param ip: ip address
-    @param netmask: netmask
-    """
-    ip_array = ip.split(".")
-    mask_array = netmask.split(".")
-    result = []
-    for index in list(range(4)):
-        result.append(int(ip_array[index]) & int(mask_array[index]))
-
-    return ".".join([str(x) for x in result])
 
 
 class NicConfigurator(object):
@@ -167,24 +150,21 @@ class NicConfigurator(object):
         # Add routes if there is no primary nic
         if not self._primaryNic and v4.gateways:
             subnet.update(
-                {'routes': self.gen_ipv4_route(nic, v4.gateways, v4.netmask)})
+                {'routes': self.gen_ipv4_default_route(nic, v4.gateways)})
 
         return ([subnet], route_list)
 
-    def gen_ipv4_route(self, nic, gateways, netmask):
+    def gen_ipv4_default_route(self, nic, gateways):
         """
-        Return the routes list needed to configure additional Ipv4 route
+        Return the routes list to configure additional Ipv4 default route
         @return (list): the route list to configure the gateways
         @param nic (NicBase): the nic to configure
         @param gateways (str list): the list of gateways
         """
         route_list = []
 
-        cidr = mask_to_net_prefix(netmask)
-
         for gateway in gateways:
-            destination = "%s/%d" % (gen_subnet(gateway, netmask), cidr)
-            route_list.append({'destination': destination,
+            route_list.append({'destination': '0.0.0.0/0',
                                'type': 'route',
                                'gateway': gateway,
                                'metric': 10000})
--- a/tests/unittests/test_vmware_config_file.py
+++ b/tests/unittests/test_vmware_config_file.py
@@ -17,7 +17,6 @@ from cloudinit.sources.DataSourceOVF import read_vmware_imc
 from cloudinit.sources.helpers.vmware.imc.boot_proto import BootProtoEnum
 from cloudinit.sources.helpers.vmware.imc.config import Config
 from cloudinit.sources.helpers.vmware.imc.config_file import ConfigFile
-from cloudinit.sources.helpers.vmware.imc.config_nic import gen_subnet
 from cloudinit.sources.helpers.vmware.imc.config_nic import NicConfigurator
 from cloudinit.tests.helpers import CiTestCase
 
@@ -173,16 +172,6 @@ class TestVmwareConfigFile(CiTestCase):
         self.assertEqual(['eng.vmware.com', 'proxy.vmware.com'],
                          dns_suffixes,
                          "suffixes")
-
-    def test_gen_subnet(self):
-        """Tests if gen_subnet properly calculates network subnet from
-           IPv4 address and netmask"""
-        ip_subnet_list = [['10.20.87.253', '255.255.252.0', '10.20.84.0'],
-                          ['10.20.92.105', '255.255.252.0', '10.20.92.0'],
-                          ['192.168.0.10', '255.255.0.0', '192.168.0.0']]
-        for entry in ip_subnet_list:
-            self.assertEqual(entry[2], gen_subnet(entry[0], entry[1]),
-                             "Subnet for a specified ip and netmask")
 
     def test_get_config_dns_suffixes(self):
         """Tests if get_network_config_from_conf properly
@@ -447,7 +436,7 @@ class TestVmwareNetConfig(CiTestCase):
                   {'control': 'auto', 'type': 'static',
                    'address': '10.20.87.154', 'netmask': '255.255.252.0',
                    'routes':
-                       [{'type': 'route', 'destination': '10.20.84.0/22',
+                       [{'type': 'route', 'destination': '0.0.0.0/0',
                          'gateway': '10.20.87.253', 'metric': 10000}]}]}],
             nc.generate())
 
@@ -489,7 +478,7 @@ class TestVmwareNetConfig(CiTestCase):
                   {'control': 'auto', 'type': 'static',
                    'address': '100.115.223.75', 'netmask': '255.255.255.0',
                    'routes':
-                       [{'type': 'route', 'destination': '100.115.223.0/24',
+                       [{'type': 'route', 'destination': '0.0.0.0/0',
                          'gateway': '100.115.223.254', 'metric': 10000}]}]}],
             nc.generate())
 
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -24,6 +24,7 @@ manuelisimo
 marlluslustosa
 matthewruffell
 nishigori
+ntoofu
 olivierlemasle
 omBratteng
 onitake


PengpengSun:
@ntoofu If a VM has 2 or more nics, which nic's gateway ip should be set to default gateway?

ntoofu:
@PengpengSun Thanks for your comment.

> Why do you want this change?

I want this change to set the default route without depending on (probably) undocumented behavior of VMware vSphere. 
 
> Is that nic configured as Primary Nic during guest customization by default? 
 
I'm sorry I don't know either how to configure the NIC as primary or its default behavior. I read documents, but couldn't find out the way. I guess it is undocumented. 
 
Tools such as Ansible cannot configure the NIC as primary because the [API specification](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.vm.customization.Specification.html) of VMware vSphere shows no property specifying "primary NIC" for guest customization. 
 
Primary NIC might be specified in the guest OS, but the [`primary-nics` setting](https://docs.vmware.com/en/VMware-Tools/11.2.0/com.vmware.vsphere.vmwaretools.doc/GUID-ECCF9D01-3666-40CE-B9FD-7EE0738AB5D9.html#set-primary-and-low-priority-interfaces-1) did not take effect as far as I tried. 
 
> See https://docs.vmware.com/en/VMware-vSphere/6.7/rn/vsphere-vcenter-server-67u2-release-notes.html#guest-os-issues-resolved 
 
This description is probably written about a vCenter Server Appliance (vCSA) itself and general virtual machines are unmentioned. 
 

In any case, the original code generates a route whose destination is a network the interface is directly connected to. I cannot imagine situations in which such a route is required. So I think this change will not cause unwanted side effects.

PengpengSun:
@ntoofu Yes,  [API specification](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.vm.customization.Specification.html) does NOT support "Primary Nic" setting. But the [vCenter Server 6.7u2 release note](https://docs.vmware.com/en/VMware-vSphere/6.7/rn/vsphere-vcenter-server-67u2-release-notes.html#guest-os-issues-resolved) I mentioned is the official release note, which means if your vCenter Server version is 6.7u2 or latter,  customize a VM (any Linux GOS which supported in [GOSC Matrix](http://partnerweb.vmware.com/programs/guestOS/guest-os-customization-matrix.pdf)) will set the first Nic to be a Primary Nic if it's customized to static IP and a gateway. I assume Ansible calls vSphere customize API to customize a VM, so it should have the same behavior. 

>In any case, the original code generates a route whose destination is a network the interface is directly connected to. I cannot imagine situations in which such a route is required. So I think this change will not cause unwanted side effects.

RHEL can set static route and default gateway, check this [Redhat doc](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s1-networkscripts-static-routes).
And consider a Ubuntu VM has more than 1 NICs,  all NICs have gateway configured, which one should be set to default gateway?

ntoofu:
@PengpengSun 

I see. I will try again with vSphere 7.0U1 this weekend. 
 
> RHEL can set static route and default gateway, check this Redhat doc. 
 
Yes, of course Linux can set static routes along with a default route.
But the destination calculated by `gen_ipv4_route` is incomprehensible. 

As shown in my experiment, `gen_ipv4_route` gives the same distination as locally connected network, which has no effect. 
 
``` 
192.168.0.0/24 dev ens160 proto kernel scope link src 192.168.0.11    # <== locally connected network 
192.168.0.0/24 via 192.168.0.1 dev ens160 proto static metric 10000   # <== calculated static route 
``` 
 
Adding a static route whose destination is the directly connected network is meaningless. A gateway address is required when a host want to send packets to a network other than that the host is connected to. If the target host is on the same network, packets can be transferred to the target by using a MAC address. 
 
Because destination networks cannot be specified explicitly in guest customization, the only thing cloud-init can do is to assume the destination is `0.0.0.0/0`.

PengpengSun:
@ntoofu I think if a default gateway can be set, then everything could work in your case. Yes, please try with vSphere 7.0U1 and share the result.
> A gateway address is required when a host want to send packets to a network other than that the host is connected to.

This is what default gateway does.

ntoofu:
@PengpengSun 
 
> Yes, please try with vSphere 7.0U1 and share the result. 
 
Thank you for your patience. When I tried with vSphere 7.0U1, the NIC is recognized as primary and the default gateway is configured successfully. The resulting route table is shown below. 
 
``` 
root@testvm:~# ip addr show 
    inet6 ::1/128 scope host 
2: ens160: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000 
    link/ether 00:50:56:99:21:d4 brd ff:ff:ff:ff:ff:ff 
    inet 192.168.0.11/24 brd 192.168.0.255 scope global ens160 
    inet6 fe80::250:56ff:fe99:21d4/64 scope link 
 
root@testvm:~# ip route show 
default via 192.168.0.1 dev ens160 proto static 
192.168.0.0/24 dev ens160 proto kernel scope link src 192.168.0.11 
``` 
 
And the following message was written in `/var/log/cloud-init.log`, which had not been present when trying with vSphere 6.5. 
 
> config_file.py[DEBUG]: ADDED KEY-VAL :: 'NIC1|PRIMARY' = 'yes' 
 
As above, the problem has been solved. Thank you for the helpful advice. 
 
--- 
 
However, I think this fix still has a benefit. Routes generated in `gen_ipv4_route` do not work regardless of the existence of a default route, because they are never selected by Linux kernel network routing. On the other hand, even if default gateway is configured properly, an additional route whose destination is `0.0.0.0/0` with large metric value is still useful for redundancy. When the link connected to the default gateway is down, the additional route will be selected.

PengpengSun:
@ntoofu I'm glad your case is working with newer vsphere version. To solve this problem completely,  a new "PRIMARY" field should be added to [API specification](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.vm.customization.Specification.html) to mark which nic should be configured to be primary,  PMs will plan on this.

IMHO, static route is configured for local sub-network, currently the destination is calculated by ip & netmask, so it should not act as a redundancy of default gateway.

ntoofu:
@PengpengSun 
 
I also think it is the best solution that vSphere users can explicitly mark a NIC as primary when making API calls.
 
Apart from that, the current `gen_ipv4_route` has an issue. 
Let's assme the following scenario: 
 
* You are deploying a VM with a non-primary NIC `ens192` 
* IP address of `ens192` is `192.168.0.11` 
* `ens192` is connected to `192.168.0.0/24` network 
* On the network, there is a gateway whose IP address is `192.168.0.1` 
 
In this case, `gen_ipv4_route` gives the following route: 
 
* Destination: `192.168.0.0/24` 
* Gateway: `192.168.0.1` 
* Metric: `10000` 
 
This means that packets whose destination address is in `192.168.0.0/24` will be forwarded to gateway `192.168.0.1`, but there is no need to forward such packets to the gateway because the destination host can be reachable by using MAC address. Regardless of the given route by `gen_ipv4_route`, the following route exist when IP address of `ens192` is configured properly: 
 
* Destination: `192.168.0.0/24` 
* Gateway: N/A 
* Interface: `ens192` 
* Scope: `link` 
 
With this route, packets whose destination address is in `192.168.0.0/24` will be transmitted to the destination host directly. 
Additionally, the routes generated by `gen_ipv4_route` is not selected because they have high metric values. 
 
In short, `gen_ipv4_route` generates routes with wrong destination and the routes are never selected by Linux kernel.

PengpengSun:
@ntoofu Sorry for late response, back to work after national holidays. 

**gateway** is not a mandatory field in [API specification](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.vm.customization.IPSettings.html), so user can leave gateway unset. While If a user set a gateway for non-primary nic on purpose, then it becomes a router, it's reasonable to let all packages target to the same sub-network ip through the router.

ntoofu:
@PengpengSun Sorry, I'm not sure what you meant. When a gateway is set for non-primary NIC in API call, is it expected that packets targeted for a host on the same sub-network are transmitted to the gateway first and then forwarded to the host?

PengpengSun:
@ntoofu Yes, IMHO, gateway is optional for non-primary NIC, while a gateway can be set for security or other reason, so I said it's reasonable. I think only primary NIC should be able to connect outside network(Internet), non-primary NIC should be internal network only. So changing non-primary NIC gateway destination to 0.0.0.0 is not proper.

This back to the requirement that [API specification](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.vm.customization.Specification.html) should have a flag to mark anyone of NICs as primary nic, but for now only the first nic can be primary if using vSphere customization API.

ntoofu:
@PengpengSun The routes generated by `gen_ipv4_route` do not work as you expect. They do nothing. Effects of such routes can be easily confirmed by the following procedure:

1. Run a container on a new Docker network and take a note of the MAC address of `eth0` interface.

    ```
    $ docker network create test
    d0bd9c859500a0a9d74015a76f6a07fffb84f7d6b1fe8fb680549fbd86e817e4

    $ docker run -it --rm --network=test alpine
    / # ip a
    1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    2: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN qlen 1000
        link/sit 0.0.0.0 brd 0.0.0.0
    53: eth0@if54: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP 
        link/ether 02:42:c0:a8:10:02 brd ff:ff:ff:ff:ff:ff
        inet 192.168.16.2/20 brd 192.168.31.255 scope global eth0
    ```
2. Run another container on the same Docker network with NET_ADMIN privilege and install `tcpdump`.

    ```
    $ docker run -it --rm --network=test --cap-add=NET_ADMIN alpine
    / # apk add --update tcpdump
    ...
    ```

3. Add a route into the second container in exactly the same way as `gen_ipv4_route`.

    ```
    / # ip route
    default via 192.168.16.1 dev eth0
    192.168.16.0/20 dev eth0 scope link  src 192.168.16.3

    / # ip route add 192.168.16.0/20 via 192.168.16.1 metric 10000

    / # ip route
    default via 192.168.16.1 dev eth0
    192.168.16.0/20 dev eth0 scope link  src 192.168.16.3
    192.168.16.0/20 via 192.168.16.1 dev eth0  metric 10000
    ```

4. Then send ICMP echo request from the second container to the first container, and capture it with `tcpdump`. The MAC addresses of captured packets show they are transmitted to the first container directly and not to the gateway. There is no difference evne without the route added in the previous step.

    ```
    / # ping 192.168.16.2 >/dev/null 2>&1 &

    / # tcpdump -i eth0 -e -nn
    tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
    listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
    18:38:52.496482 02:42:c0:a8:10:03 > 02:42:c0:a8:10:02, ethertype IPv4 (0x0800), length 98: 192.168.16.3 > 192.168.16.2: ICMP echo request, id 3072, seq 20, length 64
    18:38:52.496527 02:42:c0:a8:10:02 > 02:42:c0:a8:10:03, ethertype IPv4 (0x0800), length 98: 192.168.16.2 > 192.168.16.3: ICMP echo reply, id 3072, seq 20, length 64
    ```

> I think only primary NIC should be able to connect outside network(Internet), non-primary NIC should be internal network only.

If so, I think these routes associated with non-primary NICs should be removed totally.

PengpengSun:
@ntoofu Yes, thanks for the explanation, you are right. When there is no primary nic set, the non-primary nic's router should take effect. When there is a primary nic set, the non-primary nics' router will be ignored.
I'm ok with this change, please get one of cloud-init reviewers have a look.

PR #786, opened 2021-01-23
Network config V2 -> V1 conversion: handle accept-ra

ajmyyra:
## Handle accept_ra in network config conversion
Network config V2 -> V1 conversion: handle accept-ra

```
Network config version 2 is converted back to version 1 when rendered with a legacy render (ENI). This conversion is done by the handle functions in net/network_state.py, for example `handle_ethernets` that takes care of the ethernets part in network config v2. Subnet creation happens in function `_v2_to_v1_ipcfg` that is missing the accept-ra functionality available in both network config versions. This PR adds that.

LP: #1912545
```

## Checklist:

### DIFF
--- a/cloudinit/net/network_state.py
+++ b/cloudinit/net/network_state.py
@@ -771,8 +771,8 @@ class NetworkStateInterpreter(metaclass=CommandHandlerMeta):
             subnets.append(subnet)
         if cfg.get('dhcp6'):
             subnet = {'type': 'dhcp6'}
-            if 'accept_ra' in cfg:
-                if cfg.get('accept_ra'):
+            if 'accept-ra' in cfg:
+                if cfg.get('accept-ra'):
                     subnet = {'type': 'ipv6_dhcpv6-stateless'}
                 else:
                     subnet.update({'accept-ra': False})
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Hey @ajmyyra , it looks like you were using `accept_ra` instead of `accept-ra` in your configuration, the latter being the supported option. If I use the following configuration,
```
network:
  version: 2
  ethernets:
    ens5:
      match:
        macaddress: '3a:d6:ba:4a:29:51'
      dhcp6: True
      accept-ra: True
```
I wind up with:
```
auto lo
iface lo inet loopback

auto ens5
iface ens5 inet6 dhcp
    accept_ra 1
```

Does this accomplish what you're looking for?

ajmyyra:
Hi!

Having accept_ra instead of accept-ra seems to be a mistake from my part. Thank for noticing, I'll change it to a correct one!

I'm trying to achieve similar config that "ipv6_dhcpv6-stateless" produces in v1. So an interface with both dhcp6 and accept-ra would result in the following ENI configuration:

```
iface ens5 inet6 auto
    dhcp 1
```

This is a special case with DHCPv6 where address is autoconfigured through router advertisements and other items (DNS servers, routes etc) are then fetched from DHCP. Netplan knows what to do with just by setting dhcp6 and accept-ra as true, but ENI isn't quite that sophisticated.

ajmyyra:
Admitting my mistake, @TheRealFalcon. :) Turns out newer Debian's with ENI (as well as CoreOS, if that's still counted) explicitly need that configured with DHCP and accept_ra specified below. Modified the commit for only adding the accept-ra as is, and not using the specific stateless config option. I'll check if I've just made the same accept_ra mistake originally, with the correct one working just fine.

ajmyyra:
Yup, originally made mistake that caused the issue, not a bug in cloud-init. Sorry for the trouble!

PR #788, opened 2021-01-25
Remove 'remove-raise-on-failure' calls from integration_tests

TheRealFalcon:
Pycloudlib no longer raises exceptions when cloud-init fails to start,
and the API has been updated accordingly. Changes have been made to
integration tests accordingly

## Proposed Commit Message

```
Remove 'remove-raise-on-failure' calls from integration_tests

Pycloudlib no longer raises exceptions when cloud-init fails to start,
and the API has been updated accordingly. Changes have been made to
integration tests accordingly
```

## Additional Context
https://github.com/canonical/pycloudlib/pull/93

Tested on both lxd containers and lxd vms. Failures were unrelated.

## Checklist:

### DIFF
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -59,9 +59,7 @@ class IntegrationInstance:
     def restart(self):
         """Restart this instance (via cloud mechanism) and wait for boot.
 
-        This wraps pycloudlib's `BaseInstance.restart` to pass
-        `raise_on_cloudinit_failure=False` to `BaseInstance.wait`, mirroring
-        our launch behaviour.
+        This wraps pycloudlib's `BaseInstance.restart`
         """
         log.info("Restarting instance and waiting for boot")
         self.instance.restart()
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -56,18 +56,15 @@ class IntegrationInstance:
     def destroy(self):
         self.instance.delete()
 
-    def restart(self, raise_on_cloudinit_failure=False):
+    def restart(self):
         """Restart this instance (via cloud mechanism) and wait for boot.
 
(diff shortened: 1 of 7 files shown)


tests/integration_tests/instances.py
```diff
@@ -56,18 +56,15 @@ def __init__(self, cloud: 'IntegrationCloud', instance: BaseInstance,
     def destroy(self):
         self.instance.delete()
 
-    def restart(self, raise_on_cloudinit_failure=False):
+    def restart(self):
         """Restart this instance (via cloud mechanism) and wait for boot.
 
         This wraps pycloudlib's `BaseInstance.restart` to pass
         `raise_on_cloudinit_failure=False` to `BaseInstance.wait`, mirroring
         our launch behaviour.
```
OddBloke, line 64: This needs updating.

tests/integration_tests/modules/test_power_state_change.py
```diff
@@ -27,13 +27,13 @@ def _detect_reboot(instance: IntegrationInstance):
     # detecting the first boot or second boot, so we also check
     # the logs to ensure we've booted twice. If the logs show we've
     # only booted once, wait until we've booted twice
```
OddBloke, line 29: Total aside, but I just noticed that this test could fail (by exiting early) if run on an instance which has booted more than once without cloud-init.log removal: the "first" boot would have multiple entries. Generally for this sort of check, I think we need a "before" count which is expected to increase before we consider the reboot complete (and this is why we should capture this in library code, so every .restart() call DTRT).

Anyway, moving on.

OddBloke (changes requested):
One docstring update needed (plus an unrelated observation). I'm going to test this locally before I Approve.

OddBloke:
Tested this locally so LGTM; +1 once the docstring update happens.

OddBloke (approved):
Thanks!

PR #792, opened 2021-01-26
Add Rick Harding (mitechie) to CLA signers.

mitechie:
## Proposed Commit Message
Add Rick Harding (mitechie) to the CLA signers.


OddBloke (approved):
Thanks!

PR #793, opened 2021-01-27
Add flexibility to IMDS api-version

trstringer:
## Proposed Commit Message

Add flexibility to IMDS api-version by having both a desired IMDS api-version and a minimum api-version. The desired api-version will be used first, and if that fails it will fall back to the minimum api-version.

## Additional Context

This PR will allow us to rollout features in cloud-init to take advantage of newer features in IMDS.

## Test Steps

The following shell script was used to validate this end-to-end both using an existing desired IMDS api-version and then using an impossible (future) api-version, forcing a fallback to the minimum version. This script relies on [helper functions from my az-cli-helpers repo](https://github.com/trstringer/az-cli-helpers).

```bash
#!/bin/bash -i

# First create a VM and install the updated cloud-init from my build. Ensure
# that we are able to utilize the "desired" IMDS api version 2020-09-01.
RESOURCE_NAME=$(resource_name)
echo "$(date) - Using resource name $RESOURCE_NAME"
az_vm_create_default
az_vm_deb_install "$RESOURCE_NAME" ~/dev/cloud-init/cloud-init_all.deb
OLD_RESOURCE_NAME=$RESOURCE_NAME
RESOURCE_NAME=$(resource_name)
az_vm_create_from_vm "$OLD_RESOURCE_NAME"

if az_vm_ssh "$RESOURCE_NAME" "grep 'Falling back to IMDS api-version' /var/log/cloud-init.log"; then
    echo "$(date) Failed: api-version not found"
    exit 1
else
    echo "$(date) api-version found successfully"
fi

if az_vm_ssh "$RESOURCE_NAME" "grep 'UrlError' /var/log/cloud-init.log"; then
    echo "$(date) Failed: found UrlError"
    exit 1
else
    echo "$(date) no UrlError in logs"
fi

# For the second test, modify the source code directly so that the desired
# IMDS api-version is impossible to be used (a future date). This test
# should verify that we fall back to the minimum api-version for IMDS.
echo "$(date) Modifying to impossible IMDS version for wanted"
az_vm_ssh "$RESOURCE_NAME" "sudo sed -i 's/2020-09-01/2022-09-01/g' /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py"
az_vm_ssh "$RESOURCE_NAME" "sudo cloud-init clean --logs"
OLD_RESOURCE_NAME=$RESOURCE_NAME
RESOURCE_NAME=$(resource_name)
az_vm_create_from_vm "$OLD_RESOURCE_NAME"

if ! az_vm_ssh "$RESOURCE_NAME" "grep 'Falling back to IMDS api-version' /var/log/cloud-init.log"; then
    echo "$(date) Failed: no fall back for IMDS version"
    exit 1
else
    echo "$(date) Success for fall back IMDS version"
fi

echo "$(date) - All tests succeded"
```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -2125,6 +2125,7 @@ def _get_metadata_from_imds(
         api_version=IMDS_VER_MIN):
     url = "{}?api-version={}".format(md_type.value, api_version)
     headers = {"Metadata": "true"}
+    LOG.info("IMDS url: %s", url)
     try:
         response = readurl(
             url, timeout=IMDS_TIMEOUT_IN_SECONDS, headers=headers,
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1018,7 +1018,7 @@ class DataSourceAzure(sources.DataSource):
     def _poll_imds(self):
         """Poll IMDS for the new provisioning data until we get a valid
         response. Then return the returned JSON object."""
-        url = metadata_type.reprovisiondata.value
+        url = "{}?api-version={}".format(metadata_type.reprovisiondata.value, IMDS_VER_MIN)
         headers = {"Metadata": "true"}
         nl_sock = None
         report_ready = bool(not os.path.isfile(REPORTED_READY_MARKER_FILE))
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -635,7 +635,11 @@ class DataSourceAzure(sources.DataSource):
         return True
 
     @azure_ds_telemetry_reporter
-    def get_imds_data(self, fallback_nic, retries, md_type=metadata_type.compute):
+    def get_imds_data(
+                self,
+                fallback_nic,
+                retries,
+                md_type=metadata_type.compute):
         LOG.info("Using IMDS api-version: {}".format(self.api_version))
         if self.failed_desired_api_version:
             return get_metadata_from_imds(
@@ -654,7 +658,10 @@ class DataSourceAzure(sources.DataSource):
                     self.api_version
                 )
             except UrlError as err:
-                LOG.info("UrlError with IMDS api-version: {}".format(self.api_version))
+                LOG.info(
+                    "UrlError with IMDS api-version: {}".format(
+                        self.api_version
+                ))
                 if err.code == 400:
                     self.api_version = IMDS_VER_MIN
                     self.failed_desired_api_version = True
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -634,6 +634,48 @@ def _get_data(self):
...
+                        self.api_version
+                    )
+                    report_diagnostic_event(
+                        log_msg,
+                        logger_func=LOG.info
+                    )
+                    self.failed_desired_api_version = True
+                raise
+        except Exception:
+            return self.get_imds_data(fallback_nic, retries - 1, md_type)
```
TheRealFalcon, line 677: I think a non-recursive solution might be more straightforward. Would something like this be equivalent to what's happening in this method?
```python
        if not self.failed_desired_api_version:
            for _ in range(retries):
                try:
                    return get_metadata_from_imds(
                        fallback_nic=fallback_nic,
                        retries=0,
                        md_type=md_type,
                        api_version=IMDS_VER_WANT,
                    )
                except UrlError as err:
                    # Log stuff...
                    if err.code == 400:
                        # Log stuff and report diagnostic event
                        self.failed_desired_api_version = True
                        break
        return get_metadata_from_imds(
            fallback_nic=fallback_nic,
            retries=retries,
            md_type=md_type,
            api_version=IMDS_VER_MIN,
        )
```
This would also remove the need for a separate `api_version` instance variable (assuming the test could live without it).

cloudinit/sources/DataSourceAzure.py
```diff
@@ -634,6 +634,48 @@ def _get_data(self):
...
+                retries=retries,
+                md_type=md_type,
+                api_version=self.api_version
+            )
+
+        try:
+            try:
+                return get_metadata_from_imds(
+                    fallback_nic=fallback_nic,
+                    retries=1,
```
TheRealFalcon, line 656: `retries=1` means try twice. `retries=0` means try once. Did you intend `retries=1`?

TheRealFalcon (changes requested):
Overall looks good. Comments are inline.

trstringer:
Thanks for the review, @TheRealFalcon! I've made those requested changes and retested with the original test script (had to modify log searches, but essentially it remained the same). Let me know if you see anything else you want changed!

cloudinit/sources/DataSourceAzure.py
```diff
@@ -634,6 +636,57 @@ def _get_data(self):
...
+        if not self.failed_desired_api_version:
+            for _ in range(retries):
+                try:
+                    LOG.info(
+                        "Attempting IMDS api-version: %s",
+                        IMDS_VER_WANT
+                    )
+                    return get_metadata_from_imds(
+                        fallback_nic=fallback_nic,
+                        retries=retries,
```
TheRealFalcon, line 662: Since we're looping for the number of retries, this should be 0, right?

TheRealFalcon (commented):
Just one more inline comment, but otherwise looks good to me!

cloudinit/sources/DataSourceAzure.py
trstringer: Good catch!! Yes, you are correct.

TheRealFalcon (approved):
This looks good. Thanks!

PR #796, opened 2021-01-29
archlinux: Fix broken locale logic

klausenbusk:
## Proposed Commit Message
```
archlinux: Fix broken locale logic

The locale wasn't persisted correct nor set.

LP: https://bugs.launchpad.net/cloud-init/+bug/1402406
```

## Additional Context
See LP bug: https://bugs.launchpad.net/cloud-init/+bug/1402406

## Test Steps
This has been tested manually and `/etc/locale.gen` is in the expected format now and the locale is written to `/etc/locale.conf`.

## Checklist:

### DIFF
--- a/cloudinit/distros/arch.py
+++ b/cloudinit/distros/arch.py
@@ -23,7 +23,7 @@ LOG = logging.getLogger(__name__)
 
 
 class Distro(distros.Distro):
-    locale_conf_fn = "/etc/locale.gen"
+    locale_gen_fn = "/etc/locale.gen"
     network_conf_dir = "/etc/netctl"
     resolve_conf_fn = "/etc/resolv.conf"
     init_cmd = ['systemctl']  # init scripts
@@ -43,16 +43,20 @@ class Distro(distros.Distro):
         cfg['ssh_svcname'] = 'sshd'
 
     def apply_locale(self, locale, out_fn=None):
-        if not out_fn:
-            out_fn = self.locale_conf_fn
-        subp.subp(['locale-gen', '-G', locale], capture=False)
-        # "" provides trailing newline during join
+        if out_fn is not None and out_fn != "/etc/locale.conf":
+            util.logexc(LOG, "Invalid locale_configfile %s, only supported "
+                        "value is /etc/locale.conf", out_fn)
         lines = [
             util.make_header(),
-            'LANG="%s"' % (locale),
+            # Hard-coding the charset isn't ideal, but there is no other way.
+            '%s UTF-8' % (locale),
             "",
         ]
-        util.write_file(out_fn, "\n".join(lines))
+        util.write_file(self.locale_gen_fn, "\n".join(lines))
+        subp.subp(['locale-gen'], capture=False)
+        # In the future systemd can handle locale-gen stuff:
+        # https://github.com/systemd/systemd/pull/9864
+        subp.subp(['localectl', 'set-locale', locale], capture=False)
 
     def install_packages(self, pkglist):
         self.update_package_sources()
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -20,6 +20,7 @@ johnsonshi
 jordimassaguerpla
 jqueuniet
 jsf9k
+klausenbusk
 landon912
 lucasmoura
 lungj


cloudinit/distros/arch.py
```diff
@@ -43,16 +43,17 @@ def __init__(self, name, cfg, paths):
         cfg['ssh_svcname'] = 'sshd'
 
     def apply_locale(self, locale, out_fn=None):
-        if not out_fn:
```
TheRealFalcon, line 46: This `out_fn` should stay. The location of the config file is user-overridable, and we shouldn't just ignore that.

cloudinit/distros/arch.py
```diff
@@ -43,16 +43,17 @@ def __init__(self, name, cfg, paths):
...
+            # Hard-coding the charset isn't ideal, but there is no other way.
+            '%s UTF-8' % (locale),
             "",
         ]
-        util.write_file(out_fn, "\n".join(lines))
+        util.write_file(self.locale_gen_fn, "\n".join(lines))
+        subp.subp(['locale-gen'], capture=False)
+        # In the future systemd can handle locale-gen stuff:
+        # https://github.com/systemd/systemd/pull/9864
+        subp.subp(['localectl', 'set-locale', locale], capture=False)
```
TheRealFalcon, line 56: What does running `localectl` give us that running the code above it didn't?

TheRealFalcon (changes requested):
I've verified that the CLA has been signed. Thanks for this PR.

I left a couple comments inline. Additionally, can you provide a unit test for this behavior? There's a very small unit test file at unittests/test_distros/test_arch.py . Ensuring we write out the correct locale config would be a very useful unit test.

cloudinit/distros/arch.py
klausenbusk: The old code was broken, the line written to `/etc/locale.gen` was wrong and it didn't write to `/etc/locale.conf`.

Sure, I could just fix the code so it write the correct lines to both `/etc/locale.gen` and `/etc/locale.conf`, but `localectl` also updates the systemd [environment block](https://www.freedesktop.org/software/systemd/man/systemctl.html#Environment%20Commands) and emits a `PropertiesChanged` signal for the `org.freedesktop.locale1` `Locale` property.

... also in the future I want to remove the `locale.gen` logic as systemd can handle it (https://github.com/systemd/systemd/pull/9864), just waiting for it to ship.

cloudinit/distros/arch.py
klausenbusk: `out_fn` isn't used and it isn't configurable in systemd AFAIK, so ignoring it make sense.

klausenbusk:
> There's a very small unit test file at unittests/test_distros/test_arch.py . Ensuring we write out the correct locale config would be a very useful unit test.

`localectl` requires `systemd-localed` to be running (communicating over D-Bus). Is that something which can be tested in the CI?

cloudinit/distros/arch.py
TheRealFalcon: I'm saying it's user configurable via https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=locale#locale

If the user provides us an option, even if it's completely invalid, we don't just ignore it. If no other value other than '/etc/locale.gen' is valid, then check that they've passed '/etc/locale.gen'. You could log a warning and write anyway, or you could raise an exception, but we shouldn't just ignore user input.

TheRealFalcon:
> localectl requires systemd-localed to be running (communicating over D-Bus). Is that something which can be tested in the CI?

We use mocks in our unit tests. For testing this particular functionality, it would make sense to have tests that show:
- `/etc/locale.gen` contains what we expect
- `localectl set-locale` is called with the right argument (showing that it worked correctly would be beyond the scope of the test)
- Invalid user input is handled appropriately (per our conversation above)

I earlier linked you to some Arch tests, but looking again, it'd probably make more sense for them to go into the [test_handler_locale tests](https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_handler/test_handler_locale.py) . There are also other examples in there showing how some of the mocking works.

I hope that all makes sense. Let me know if you have any questions or need some additional guidance. Further test guidelines can be found here:
https://cloudinit.readthedocs.io/en/latest/topics/testing.html

cloudinit/distros/arch.py
klausenbusk: Oh. I added a warning.

TheRealFalcon: You should use `LOG.warning` for the log instead of `util.logexc`. `util.logexc` is meant to be used in an except block and will gather exception information.

klausenbusk: Thanks. I got a bit confused so I just guessed :)

klausenbusk:
Real life is getting in the way and writing tests is a bit more complicated than writing plain Python code. I hope to find some time in the near future.

klausenbusk:
@mitechie can you reopen?

@TheRealFalcon I pushed the unit test code but the PR is apparently not updated when it is closed.

TheRealFalcon:
@klausenbusk , unfortunately it appears that github doesn't allow reopening of PRs that have had their branch force-pushed to.

Feel free to open a new PR with your changes.

klausenbusk:
> @klausenbusk , unfortunately it appears that github doesn't allow reopening of PRs that have had their branch force-pushed to.

TIL. I opened a new PR: https://github.com/canonical/cloud-init/pull/841

PR #798, opened 2021-02-01
azure: case-insensitive UUID to avoid new IID during kernel upgrade

blackboxsw:
## Proposed Commit Message
Kernel's newer than 4.15 present /sys/dmi/id/product_uuid as a
lowercase value. Previously UUID was uppercase.

Azure datasource reads the product_uuid directly as their platform's
instance-id. This presents a problem if a kernel is either
upgraded or downgraded across the 4.15 kernel version boundary because
the case of the UUID will change, resulting in cloud-init seeing a
"new" instance id and re-running all modules.

Re-running cc_ssh in cloud-init deletes and regenerates ssh_host keys
on a system which can cause concern on long-running instances that 
somethingnefarious has happened.

Also add:
 - An integration test for this for Azure Bionic Ubuntu FIPS upgrading from
   a FIPS kernel with uppercase UUID to a lowercase UUID in linux-azure
 - A new pytest.mark.sru_next to collect all integration tests related to our
   next SRU
                                                             
LP: #1835584

## Additional Context
Integration test will add a 4.14 generic kernel, add grub config to prefer the "generic" flavor of that kernel across reboot.
This triggers the uppercase product_uuid which on current cloud-init would generate new ssh host keys.
New cloud-init will not trigger this ssh host key creation.

## Test Steps
```bash
# create a passwordless ssh pub/private keypair
ssh-keygen 


 CLOUD_INIT_PUBLIC_SSH_KEY=/root/.ssh/id_rsa_az.pub CLOUD_INIT_PLATFORM=azure  .tox/integration-tests/bin/py.test -v tests/integration_tests/bugs/test_lp1835584.py
# Expect failure running against current cloud-init: 
# E       AssertionError: config_ssh ran too many times 2
# E       assert 1 == 2
# E         +1
# E         -2

make deb
# Expect success running using CLOUD_INIT_CLOUD_INIT_SOURCE=<deb_from_this_pr>
```

## Checklist:

 - [n/a] I have updated or added any documentation accordingly

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -683,10 +683,18 @@ class DataSourceAzure(sources.DataSource):
     def _iid(self, previous=None):
         prev_iid_path = os.path.join(
             self.paths.get_cpath('data'), 'instance-id')
-        iid = dmi.read_dmi_data('system-uuid')
+        # Older kernels than 4.15 will have UPPERCASE product_uuid.
+        # We don't want Azure to react to an UPPER/lower difference as a new
+        # instance id as it rewrites SSH host keys.
+        # LP: #1835584
+        iid = dmi.read_dmi_data('system-uuid').lower()
         if os.path.exists(prev_iid_path):
             previous = util.load_file(prev_iid_path).strip()
-            if is_byte_swapped(previous, iid):
+            if previous.lower() == iid:
+                # If uppercase/lowercase equivalent, return the previous value
+                # to avoid new instance id.
+                return previous
+            if is_byte_swapped(previous.lower(), iid):
                 return previous
         return iid
 
--- a/tests/integration_tests/bugs/test_lp1835584.py
+++ b/tests/integration_tests/bugs/test_lp1835584.py
@@ -1,0 +1,104 @@
+""" Integration test for LP #1835584
+
+Upstream linux kernels prior to 4.15 providate DMI product_uuid in uppercase.
+More recent kernels switched to lowercase for DMI product_uuid. Azure
+datasource uses this product_uuid as the instance-id for cloud-init.
+
+The linux-azure-fips kernel installed in PRO FIPs images, that product UUID is
+uppercase whereas the linux-azure cloud-optimized kernel reports the UUID as
+lowercase.
+
+In cases where product_uuid changes case, ensure cloud-init doesn't
+recreate ssh hostkeys across reboot (due to detecting an instance_id change).
+
+This currently only affects linux-azure-fips -> linux-azure on Bionic.
+This test won't run on Xenial because both linux-azure-fips and linux-azure
+report uppercase product_uuids.
+
+The test will launch a specific Bionic Ubuntu PRO FIPS image which has a
+linux-azure-fips kernel known to report product_uuid as uppercase. Then upgrade
+and reboot into linux-azure kernel which is known to report product_uuid as
+lowercase.
+
+Across the reboot, assert that we didn't re-run config_ssh by virtue of
+seeing only one semaphore creation log entry of type:
+
+ Writing to /var/lib/cloud/instances/<UUID>/sem/config_ssh -
+
+https://bugs.launchpad.net/cloud-init/+bug/1835584
+"""
+import re
+
+import pytest
+
+from tests.integration_tests.instances import IntegrationAzureInstance
+from tests.integration_tests.clouds import (
+    ImageSpecification, IntegrationCloud
+)
+from tests.integration_tests.conftest import get_validated_source
+
+
+IMG_AZURE_UBUNTU_PRO_FIPS_BIONIC = (
+    "Canonical:0001-com-ubuntu-pro-bionic-fips:pro-fips-18_04:18.04.202010201"
+)
+
+
+def _check_iid_insensitive_across_kernel_upgrade(
+    instance: IntegrationAzureInstance
+):
+    uuid = instance.read_from_file("/sys/class/dmi/id/product_uuid")
+    assert uuid.isupper(), (
+        "Expected uppercase UUID on Ubuntu FIPS image {}".format(
+            uuid
+        )
+    )
+    orig_kernel = instance.execute("uname -r").strip()
+    assert "azure-fips" in orig_kernel
+    result = instance.execute("apt-get update")
+    # Install a 5.4+ kernel which provides lowercase product_uuid
+    result = instance.execute("apt-get install linux-azure --assume-yes")
+    if not result.ok:
+        pytest.fail("Unable to install linux-azure kernel: {}".format(result))
+    instance.restart()
+    new_kernel = instance.execute("uname -r").strip()
+    assert orig_kernel != new_kernel
+    assert "azure-fips" not in new_kernel
+    assert "azure" in new_kernel
+    new_uuid = instance.read_from_file("/sys/class/dmi/id/product_uuid")
+    assert (
+        uuid.lower() == new_uuid
+    ), "Expected UUID on linux-azure to be lowercase of FIPS: {}".format(uuid)
+    log = instance.read_from_file("/var/log/cloud-init.log")
+    RE_CONFIG_SSH_SEMAPHORE = r"Writing.*sem/config_ssh "
+    ssh_runs = len(re.findall(RE_CONFIG_SSH_SEMAPHORE, log))
+    assert 1 == ssh_runs, "config_ssh ran too many times {}".format(ssh_runs)
+
+
+@pytest.mark.azure
+@pytest.mark.sru_next
+def test_azure_kernel_upgrade_case_insensitive_uuid(
+    session_cloud: IntegrationCloud
+):
+    cfg_image_spec = ImageSpecification.from_os_image()
+    if (cfg_image_spec.os, cfg_image_spec.release) != ("ubuntu", "bionic"):
+        pytest.skip(
+            "Test only supports ubuntu:bionic not {0.os}:{0.release}".format(
+                cfg_image_spec
+            )
+        )
+    source = get_validated_source(session_cloud)
+    if not source.installs_new_version():
+        pytest.skip(
+            "Provide CLOUD_INIT_SOURCE to install expected working cloud-init"
+        )
+    image_id = IMG_AZURE_UBUNTU_PRO_FIPS_BIONIC
+    with session_cloud.launch(
+        launch_kwargs={"image_id": image_id}
+    ) as instance:
+        # We can't use setup_image fixture here because we want to avoid
+        # taking a snapshot or cleaning the booted machine after cloud-init
+        # upgrade.
+        instance.install_new_cloud_init(
+            source, take_snapshot=False, clean=False
+        )
+        _check_iid_insensitive_across_kernel_upgrade(instance)
--- a/tests/unittests/test_datasource/test_azure.py
+++ b/tests/unittests/test_datasource/test_azure.py
@@ -201,6 +201,7 @@ IMDS_NETWORK_METADATA = {
 }
 
 MOCKPATH = 'cloudinit.sources.DataSourceAzure.'
+EXAMPLE_UUID = 'd0df4c54-4ecb-4a4b-9954-5bdf3ed5c3b8'
 
 
 class TestParseNetworkConfig(CiTestCase):
@@ -630,7 +631,7 @@ scbus-1 on xpt0 bus 0
         return dsaz
 
     def _get_ds(self, data, agent_command=None, distro='ubuntu',
-                apply_network=None):
+                apply_network=None, instance_id=None):
 
         def dsdevs():
             return data.get('dsdevs', [])
@@ -659,7 +660,10 @@ scbus-1 on xpt0 bus 0
         self.m_ephemeral_dhcpv4 = mock.MagicMock()
         self.m_ephemeral_dhcpv4_with_reporting = mock.MagicMock()
 
-        self.instance_id = 'D0DF4C54-4ECB-4A4B-9954-5BDF3ED5C3B8'
+        if instance_id:
+            self.instance_id = instance_id
+        else:
+            self.instance_id = EXAMPLE_UUID
 
         def _dmi_mocks(key):
             if key == 'system-uuid':
@@ -910,7 +914,7 @@ scbus-1 on xpt0 bus 0
             'azure_data': {
                 'configurationsettype': 'LinuxProvisioningConfiguration'},
             'imds': NETWORK_METADATA,
-            'instance-id': 'D0DF4C54-4ECB-4A4B-9954-5BDF3ED5C3B8',
+            'instance-id': EXAMPLE_UUID,
             'local-hostname': u'myhost',
             'random_seed': 'wild'}
 
@@ -1613,6 +1617,32 @@ scbus-1 on xpt0 bus 0
         self.assertTrue(ret)
         self.assertEqual('value', dsrc.metadata['test'])
 
+    def test_instance_id_case_insensitive(self):
+        """Return the previous iid when current is a case-insensitive match."""
+        lower_iid = EXAMPLE_UUID.lower()
+        upper_iid = EXAMPLE_UUID.upper()
+        # lowercase current UUID
+        ds = self._get_ds(
+            {'ovfcontent': construct_valid_ovf_env()}, instance_id=lower_iid
+        )
+        # UPPERCASE previous
+        write_file(
+            os.path.join(self.paths.cloud_dir, 'data', 'instance-id'),
+            upper_iid)
+        ds.get_data()
+        self.assertEqual(upper_iid, ds.metadata['instance-id'])
+
+        # UPPERCASE current UUID
+        ds = self._get_ds(
+            {'ovfcontent': construct_valid_ovf_env()}, instance_id=upper_iid
+        )
+        # lowercase previous
+        write_file(
+            os.path.join(self.paths.cloud_dir, 'data', 'instance-id'),
+            lower_iid)
+        ds.get_data()
+        self.assertEqual(lower_iid, ds.metadata['instance-id'])
+
     def test_instance_id_endianness(self):
         """Return the previous iid when dmi uuid is the byteswapped iid."""
         ds = self._get_ds({'ovfcontent': construct_valid_ovf_env()})
@@ -1628,8 +1658,7 @@ scbus-1 on xpt0 bus 0
             os.path.join(self.paths.cloud_dir, 'data', 'instance-id'),
             '644CDFD0-CB4E-4B4A-9954-5BDF3ED5C3B8')
         ds.get_data()
-        self.assertEqual(
-            'D0DF4C54-4ECB-4A4B-9954-5BDF3ED5C3B8', ds.metadata['instance-id'])
+        self.assertEqual(self.instance_id, ds.metadata['instance-id'])
 
     def test_instance_id_from_dmidecode_used(self):
         ds = self._get_ds({'ovfcontent': construct_valid_ovf_env()})
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -116,7 +116,8 @@ class IntegrationInstance:
     def install_new_cloud_init(
         self,
         source: CloudInitSource,
-        take_snapshot=True
+        take_snapshot=True,
+        clean=True,
     ):
         if source == CloudInitSource.DEB_PACKAGE:
             self.install_deb()
@@ -133,7 +134,8 @@ class IntegrationInstance:
             )
         version = self.execute('cloud-init -v').split()[-1]
         log.info('Installed cloud-init version: %s', version)
-        self.instance.clean()
+        if clean:
+            self.instance.clean()
         if take_snapshot:
             snapshot_id = self.snapshot()
             self.cloud.snapshot_id = snapshot_id
--- a/tox.ini
+++ b/tox.ini
@@ -184,5 +184,6 @@ markers =
     instance_name: the name to be used for the test instance
     sru_2020_11: test is part of the 2020/11 SRU verification
     sru_2021_01: test is part of the 2021/01 SRU verification
+    sru_next: test is part of the next SRU verification
     ubuntu: this test should run on Ubuntu
     unstable: skip this test because it is flakey


tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.sru_next
+def test_upgrade_azure_uuid_case_insensitive(
+    setup_image, session_cloud: IntegrationCloud
+):
+    try:
+        with session_cloud.launch(launch_kwargs={
+            'image_id': UBUNTU_PRO_FIPS_IMG
+        }) as instance1:
+            _check_iid_insensitive_across_kernel_upgrade(instance1)
+    finally:
```
TheRealFalcon, line 61: When the `session_cloud.launch(...)` context manager exits, the instance is automatically destroyed. A snapshot won't be taken if `setup_image` from above is removed.

You should be able to remove this `finally` block

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+    log = instance.read_from_file('/var/log/cloud-init.log')
+    RE_CONFIG_SSH_SEMAPHORE = r'Writing.*sem/config_ssh '
+    matches = re.findall(RE_CONFIG_SSH_SEMAPHORE, log)
+    assert 1 == len(matches), "config_ssh ran too many times {}".format(matches)
+
+
+@pytest.mark.azure
+@pytest.mark.sru_next
+def test_upgrade_azure_uuid_case_insensitive(
+    setup_image, session_cloud: IntegrationCloud
```
TheRealFalcon, line 54: `setup_image` sets up the image to use in our tests. Since we're passing in the image id, we shouldn't need `setup_image` here.

TheRealFalcon (commented):
Only looked at the integration test. Can do fuller review tomorrow.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
+""" Integration test for LP #1835584
+
+Kernels prior to 4.15 uppercased the dmi product_uuid, Azure
+relies on system UUID as their "instance-id". Upgrades to kernels >= 4.15
+will represent the same product_uuid as lowercase. System UUID should be a
+case-insensitive comparison.
+
+The test will launch an Ubuntu PRO FIPS image on Azure which has a 4.15 kernel.
```
OddBloke, line 8: Does this need to be a Pro image? Could we use an older general-purpose Ubuntu image to bring this test more closely inline with the rest of our testing?

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+
+UBUNTU_PRO_FIPS_IMG = 'Canonical:0001-com-ubuntu-pro-bionic-fips:pro-fips-18_04'
+
+def _check_iid_insensitive_across_kernel_upgrade(instance):
+    uuid = instance.read_from_file('/sys/class/dmi/id/product_uuid')
+    assert uuid.isupper(), "UUID does not appear to be upppercase {}".format(
+      uuid
+    )
+    kernel = instance.read_from_file('uname -r').strip()
+    assert kernel == '4.15.0-1002-azure-fips'
```
OddBloke, line 35: This will start failing once the Pro image and its kernel are revved, right?

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+    if not instance.execute('sudo apt-get install linux-azure --assume-yes').ok:
+        raise Exception("Failed installing linux-azure: {}".format(result))
+    instance.restart()
+    kernel = instance.execute('uname -r').strip()
+    assert kernel == '5.4.0-1039-azure'
```
OddBloke, line 40: This is similarly fragile.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+import pytest
+
+from tests.integration_tests.clouds import IntegrationCloud
+
+
+UBUNTU_PRO_FIPS_IMG = 'Canonical:0001-com-ubuntu-pro-bionic-fips:pro-fips-18_04'
+
+def _check_iid_insensitive_across_kernel_upgrade(instance):
+    uuid = instance.read_from_file('/sys/class/dmi/id/product_uuid')
+    assert uuid.isupper(), "UUID does not appear to be upppercase {}".format(
```
OddBloke, line 31:
```suggestion
    assert uuid.isupper(), "UUID does not appear to be uppercase {}".format(
```

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    log = instance.read_from_file('/var/log/cloud-init.log')
+    RE_CONFIG_SSH_SEMAPHORE = r'Writing.*sem/config_ssh '
+    matches = re.findall(RE_CONFIG_SSH_SEMAPHORE, log)
+    assert 1 == len(matches), "config_ssh ran too many times {}".format(matches)
+
+
+@pytest.mark.azure
+@pytest.mark.sru_next
+def test_upgrade_azure_uuid_case_insensitive(
```
OddBloke, line 53: As written, this should only run on bionic.

OddBloke (commented):
Some quick integration test thoughts.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,88 @@
...
+
+USER_DATA = """\
+#cloud-config
+runcmd:
+{wgets}
+  - 'dpkg -i *.deb'
+""".format(wgets=wgets)
+
+
+def _check_iid_insensitive_across_kernel_upgrade(client):
```
OddBloke, line 51: Should we just inline this in the test function?

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,88 @@
+""" Integration test for LP #1835584
+
+Kernels prior to 4.15 report dmi product_uuid as uppercase, Azure
+relies on system UUID as their "instance-id". Upgrades to kernels >= 4.15
+will represent the same product_uuid as lowercase. System UUID should be a
+case-insensitive comparison and avoid triggering "new instance" detection
+in cloud-init on Azure platform.
+
+The test will launch an Ubuntu image which has a 5.4 kernel. We allow
```
OddBloke, line 9: This isn't true: this test is not constrained by Ubuntu release, so will launch whatever version of the kernel is in the latest Azure image for the target Ubuntu release.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+import re
+
+import pytest
+
+from tests.integration_tests.instances import IntegrationInstance
+
+
+KERNEL_URL = "https://kernel.ubuntu.com/~kernel-ppa/mainline/v4.14.120/"
+PKG_VER = "4.14.120-0414120-generic_4.14.120-0414120.201905161610_amd64"
```
OddBloke, line 30: It's my understanding that the generic kernel won't boot on Azure in all cases, and this will presumably only become more true over time. Do we need to find a way of installing an old Azure kernel instead?

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,88 @@
+""" Integration test for LP #1835584
+
+Kernels prior to 4.15 report dmi product_uuid as uppercase, Azure
+relies on system UUID as their "instance-id". Upgrades to kernels >= 4.15
+will represent the same product_uuid as lowercase. System UUID should be a
+case-insensitive comparison and avoid triggering "new instance" detection
+in cloud-init on Azure platform.
+
+The test will launch an Ubuntu image which has a 5.4 kernel. We allow
+cloud-init to detect the instance-id on the 5.4 kernel, set a 4.15 kernel as
```
OddBloke, line 10:
```suggestion
cloud-init to detect the instance-id on the 5.4 kernel, set a 4.14 kernel as
```

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,88 @@
...
+default and reboot the system. Expectation is that cloud-init retains same
+instance-id status and avoids triggering ssh host key regeneration across that
+reboot.
+
+Across the reboot, assert that we didn't re-run config_ssh by virtue of
+seeing only one semaphore creation log entry of type:
+
+ Writing to /var/lib/cloud/instances/<UUID>/sem/config_ssh -
+
+https://bugs.launchpad.net/cloud-init/+bug/1897099
```
OddBloke, line 20: Don't think this link is right.

OddBloke (commented):
Just integration test review for now. I've left some specific comments, but I'd like us to step back and consider whether this is the right path for this test. Previously you were launching an image that would present with one UUID case, and then upgrading its kernel in a way that would present the other case. I suggested an iteration of that approach: using an old standard Ubuntu image to avoid the dependency on Pro images. I think it's likely that would give us a simpler test (with fewer external dependencies); did you try that approach before settling on this one?

cloudinit/sources/DataSourceAzure.py
```diff
@@ -683,10 +683,16 @@ def check_instance_id(self, sys_cfg):
...
+        # Older kernels than 4.15 will have UPPERCASE product_uuid.
+        # We don't want Azure to react to an UPPER/lower difference as a new
+        # instance id as it rewrites SSH host keys.
+        # LP: #1835584
+        iid = dmi.read_dmi_data('system-uuid').lower()
         if os.path.exists(prev_iid_path):
             previous = util.load_file(prev_iid_path).strip()
-            if is_byte_swapped(previous, iid):
+            if previous.lower() == iid:
+                return previous
```
OddBloke, line 694: `is_byte_swapped` specifically considers `previous_id == current_id` to mean that they _don't_ match. Is this addressing an identified bug in that behaviour?

tests/unittests/test_datasource/test_azure.py
```diff
@@ -1613,6 +1617,32 @@ def test_fabric_data_included_in_metadata(self):
         self.assertTrue(ret)
         self.assertEqual('value', dsrc.metadata['test'])
 
+    def test_instance_id_case_insensitive(self):
+        """Return the previous iid when current is a case-insensitive match."""
+        lower_iid = EXAMPLE_UUID
```
OddBloke, line 1622:
```suggestion
        lower_iid = EXAMPLE_UUID.lower()
```

in case someone comes through and capitalises it?

OddBloke (commented):
OK, and now some not-integration-test review.

cloudinit/sources/DataSourceAzure.py
blackboxsw: The False returned from `is_byte_swapped` reports that prev_id and cur_id are actually a match and not "byte swapped" values which means that returning previous or current won't show up as a new instance id.

A False from is_byte_swapped means don't return the previous value, just return current which makes no different if previous_id == current_id.   This special check is because we also need to return previous_id if previous_id and current_id are case-insensitive matches of each other. In that event, we always want to return previous (the original uuid detected) whether or not that was uppercase or lowercase to avoid triggering new-instance-id detection in cloud-init.

OddBloke: Aha, right, makes sense to me now; thanks!

tests/integration_tests/bugs/test_lp1835584.py
blackboxsw: I've pinned the serial of the Ubuntu PRO bionic FIPS image so this won't break in this iteration of the integration test as we are only upgrading from stale FIPS -> newer kernel on Bionic.

tests/integration_tests/bugs/test_lp1835584.py
blackboxsw: In our discussion, the only real path for this currently is Bionic Ubuntu PRO FIPS Azure -> non-FIPS linux-azure kernel. Even linux-azure on Xenial still exhibits a product_uuid that is uppercase and matches the FIPS kernel product_uuid. So, I think we have to limit this test to Bionic to expose the real upgrade path issue folks are seeing.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,82 @@
...
+
+In cases where product_uuid changes case, ensure cloud-init doesn't
+recreate ssh hostkeys across reboot (due to detecting an instance_id change).
+
+This currently only affects linux-azure-fips -> linux-azure on Bionic.
+This test won't run on Xenial because both linux-azure-fips and linux-azure
+report uppercase product_uuids.
+
+The test will launch a Bionic Ubuntu PRO FIPS image which has
+linux-azure-fips kernel and reports product_uuid as uppercase. Then upgrade
```
OddBloke, line 19:
```suggestion
The test will launch a specific Bionic Ubuntu PRO FIPS image which has a
linux-azure-fips kernel known to report product_uuid as uppercase. Then upgrade
```

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+    log = instance.read_from_file('/var/log/cloud-init.log')
+    RE_CONFIG_SSH_SEMAPHORE = r'Writing.*sem/config_ssh '
+    ssh_runs = len(re.findall(RE_CONFIG_SSH_SEMAPHORE, log))
+    assert 1 == ssh_runs, "config_ssh ran too many times {}".format(ssh_runs)
+
+
+@pytest.mark.azure
+@pytest.mark.sru_next
+@pytest.mark.not_xenial
```
OddBloke, line 75: This will still run on every release bionic+; we *could* introduce an `only_bionic` mark or similar. However, given that this test doesn't use the `client` fixtures, we don't actually need to perform the skip via marks to avoid the instance launch: we can do it in the body of the test before we call `launch` instead.

Ideally, we'd be able to do something like `if (session_cloud.image.os, session_cloud.image.release) != ("ubuntu", "bionic"): pytest.skip(...)` but we don't currently store the `ImageSpecification` that we determine on `session_cloud`. That's probably not too painful to hook up (though may require some thought around snapshotting), but I don't think we need to for this, re-parsing `OS_IMAGE` should be sufficient: something like `image_spec = ImageSpecification.from_os_image(); if (image_spec.os, image_spec.release) != ("ubuntu", "bionic"): pytest.skip(...)`.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,82 @@
...
+def test_azure_kernel_upgrade_case_insensitive_uuid(
+    session_cloud: IntegrationCloud
+):
+    with session_cloud.launch(launch_kwargs={
+        'image_id': IMG_AZURE_UBUNTU_PRO_FIPS_BIONIC
+    }) as instance1:
+        _check_iid_insensitive_across_kernel_upgrade(instance1)
```
OddBloke, line 82:
```suggestion
    }) as instance:
        _check_iid_insensitive_across_kernel_upgrade(instance)
```

OddBloke (commented):
Code, unit tests and general structure of the integration test looks good to me, thank you!

We still need to get this test to run only on bionic, see my inline thoughts.

tests/integration_tests/bugs/test_lp1835584.py
blackboxsw: @OddBloke funny, I put that type of pytest.skip in locally and decided to pullt it out based on looking at self.settings.os_image I think. I'll put some semblance of that back into this PR

tests/integration_tests/bugs/test_lp1835584.py
blackboxsw: We've now pinned the test to launch a specific Ubuntu PRO FIPS image that will be available forever and exhibits a UUID case change when upgrading to linux-azure from linux-azure-fips.

OddBloke (commented):
Thanks for the revisions, a few more thoughts inline.

One thing to note, if I try to run this test locally, I get:

```
E           msrestazure.azure_exceptions.CloudError: Azure Error: ResourcePurchaseValidationFailed
E           Message: User failed validation to purchase resources. Error message: 'You have not accepted the legal terms on this subscription: '9a37cc2c-dc4a-4097-84dd-45dd2b8cbc63' for this plan. Before the subscription can be used, you need to accept the legal terms of the image. To read and accept legal terms, use the Azure CLI commands described at https://go.microsoft.com/fwlink/?linkid=2110637 or the PowerShell commands available at https://go.microsoft.com/fwlink/?linkid=862451. Alternatively, deploying via the Azure portal provides a UI experience for reading and accepting the legal terms. Offer details: publisher='canonical' offer = '0001-com-ubuntu-pro-bionic-fips', sku = 'pro-fips-18_04', Correlation Id: 'b0f49067-fae7-4c9f-b26c-8b7fd2eec9a0'.'
```

I'm going to follow those steps now, but this confirms that we shouldn't be running this test by default.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,112 @@
...
+    source = get_validated_source(session_cloud)
+    install_cloudinit = source.installs_new_version()
+    image_id = IMG_AZURE_UBUNTU_PRO_FIPS_BIONIC
+    with session_cloud.launch(
+        launch_kwargs={"image_id": image_id}
+    ) as instance:
+        # Allow for upgrade of cloud-init in FIPS image for pre-release testing
+        if install_cloudinit:
+            instance.install_new_cloud_init(source, take_snapshot=False)
+            _restart_with_retries(instance)
```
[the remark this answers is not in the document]
blackboxsw: The reason we restart after upgrading cloud-init is because `install_new_cloud_init` calls instance.clean() which removes all cloud-init logs, semaphores and syslog though pycloudlib. What we are trying to establish is that the upgraded cloud-init detects uppercase UUID on first boot, then on 2nd boot into new kernel we don't accidentally detect a change in instance-id due to the now lowercase UUID.   I support instead of the reboot I could either avoid instance.clean() in `install_new_cloud_init` by providing a `clean=False` param. That'd save time on the reboot (because cloud-init will have already detected the uppercase instance-id even before we upgrade cloud-init on that instance.

tests/integration_tests/bugs/test_lp1835584.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+    source = get_validated_source(session_cloud)
+    install_cloudinit = source.installs_new_version()
+    image_id = IMG_AZURE_UBUNTU_PRO_FIPS_BIONIC
+    with session_cloud.launch(
+        launch_kwargs={"image_id": image_id}
+    ) as instance:
+        # Allow for upgrade of cloud-init in FIPS image for pre-release testing
+        if install_cloudinit:
+            instance.install_new_cloud_init(source, take_snapshot=False)
```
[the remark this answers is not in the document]
blackboxsw: yep. added clarifying comment. we don't want to clean and we don't want to snapshot so we can't quite use `setup_image`

tests/integration_tests/bugs/test_lp1835584.py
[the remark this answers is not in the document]
blackboxsw: I've pushed a change to avoid the extra-reboot and I'm testing this now.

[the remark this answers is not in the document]
OddBloke: Yeah, I'd support modifying `install_new_cloud_init` to allow for this case; a `clean` parameter sounds reasonable to me.

@TheRealFalcon How does that sound to you?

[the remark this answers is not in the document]
blackboxsw: added clean param and confirmed success run of the following (which a built deb from this PR)
```bash
 CLOUD_INIT_OS_IMAGE=bionic CLOUD_INIT_CLOUD_INIT_SOURCE=/cloud-init_20.4.1-86-g5c199e90-1~bddeb_all.deb CLOUD_INIT_PUBLIC_SSH_KEY=/mykey  CLOUD_INIT_PLATFORM=azure  .tox/integration-tests/bin/py.test -v tests/integration_tests/bugs/test_lp1835584.py
```

OddBloke:
> One thing to note, if I try to run this test locally, I get:
> 
> ```
> E           msrestazure.azure_exceptions.CloudError: Azure Error: ResourcePurchaseValidationFailed
> E           Message: User failed validation to purchase resources. Error message: 'You have not accepted the legal terms on this subscription: '9a37cc2c-dc4a-4097-84dd-45dd2b8cbc63' for this plan. Before the subscription can be used, you need to accept the legal terms of the image. To read and accept legal terms, use the Azure CLI commands described at https://go.microsoft.com/fwlink/?linkid=2110637 or the PowerShell commands available at https://go.microsoft.com/fwlink/?linkid=862451. Alternatively, deploying via the Azure portal provides a UI experience for reading and accepting the legal terms. Offer details: publisher='canonical' offer = '0001-com-ubuntu-pro-bionic-fips', sku = 'pro-fips-18_04', Correlation Id: 'b0f49067-fae7-4c9f-b26c-8b7fd2eec9a0'.'
> ```
> 
> I'm going to follow those steps now, but this confirms that we shouldn't be running this test by default.

To follow-up `az vm image terms accept --urn Canonical:0001-com-ubuntu-pro-bionic-fips:pro-fips-18_04:18.04.202010201` fixed this issue for me.

However, now that I've successfully run the test, it has failed. The cloud-init version in that image (intentionally) doesn't have the fix to make this test pass, which means this test can _never_ pass without an updated cloud-init being installed. (This is different to our usual tests, which we would expect to fail until the Ubuntu image is updated with the fixed cloud-init: this test will never use an image with a fixed cloud-init.) Should we skip the test if we don't have a CLOUD_INIT_SOURCE which `installs_new_version()`?

blackboxsw:
> > One thing to note, if I try to run this test locally, I get:
> > ```
> > E           msrestazure.azure_exceptions.CloudError: Azure Error: ResourcePurchaseValidationFailed
> > E           Message: User failed validation to purchase resources. Error message: 'You have not accepted the legal terms on this subscription: '9a37cc2c-dc4a-4097-84dd-45dd2b8cbc63' for this plan. Before the subscription can be used, you need to accept the legal terms of the image. To read and accept legal terms, use the Azure CLI commands described at https://go.microsoft.com/fwlink/?linkid=2110637 or the PowerShell commands available at https://go.microsoft.com/fwlink/?linkid=862451. Alternatively, deploying via the Azure portal provides a UI experience for reading and accepting the legal terms. Offer details: publisher='canonical' offer = '0001-com-ubuntu-pro-bionic-fips', sku = 'pro-fips-18_04', Correlation Id: 'b0f49067-fae7-4c9f-b26c-8b7fd2eec9a0'.'
> > ```
> > 
> > 
> > I'm going to follow those steps now, but this confirms that we shouldn't be running this test by default.
> 
> To follow-up `az vm image terms accept --urn Canonical:0001-com-ubuntu-pro-bionic-fips:pro-fips-18_04:18.04.202010201` fixed this issue for me.
> 
> However, now that I've successfully run the test, it has failed. The cloud-init version in that image (intentionally) doesn't have the fix to make this test pass, which means this test can _never_ pass without an updated cloud-init being installed. (This is different to our usual tests, which we would expect to fail until the Ubuntu image is updated with the fixed cloud-init: this test will never use an image with a fixed cloud-init.) Should we skip the test if we don't have a CLOUD_INIT_SOURCE which `installs_new_version()`?

Yes this a good point, I was thinking for current dev it gives us the change to run the test to see the failure currently without the image_source set. But, we probably don't want to instrument tests that we can easily cause to break. Reviewers could just remove such a pytest.skip  if they wanted to see failures.

TheRealFalcon (approved):
Integration test LGTM. I did a much shallower review of everything else given how much Dan has been involved.

PR #800, opened 2021-02-04
Decrease azure ephemeral resource disk wait time

johnsonshi:
## Proposed Commit Message

Azure: Support for VMs without ephemeral resource disks.

Changes:
* Only merge in default Azure cloud ephemeral disk configs
during `DataSourceAzure._get_data()` if the ephemeral disk
exists.
* `DataSourceAzure.address_ephemeral_resize()` (which is
invoked in `DataSourceAzure.activate()` should only set up
the ephemeral disk if the disk exists.

Azure VMs may or may not come with ephemeral resource disks
depending on the VM SKU. For VM SKUs that come with
ephemeral resource disks, the Azure platform guarantees that
the ephemeral resource disk is attached to the VM before
the VM is booted. For VM SKUs that do not come with
ephemeral resource disks, cloud-init currently attempts
to wait and set up a non-existent ephemeral resource
disk, which wastes boot time. It also causes disk setup
modules to fail (due to non-existent references to the
ephemeral resource disk).

`udevadm settle` is invoked by cloud-init very early in boot.
`udevadm settle` is invoked very early, before
`DataSourceAzure`'s `_get_data()` and `activate()` methods.

Within `DataSourceAzure`'s `_get_data()` and `activate()` methods,
the ephemeral resource disk path should exist if the
VM SKU comes with an ephemeral resource disk.
The ephemeral resource disk path should not exist if the
VM SKU does not come with an ephemeral resource disk.

LP: #1901011

#### Problem

For Azure VMs, cloud-init's `DataSourceAzure.py` formats and addresses ephemeral disk resizing. It does this for all Azure VM SKUs. See [code](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/cloudinit/sources/DataSourceAzure.py#L1314-L1316) and [code](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/cloudinit/sources/DataSourceAzure.py#L1453-L1455).

The [code right now waits up to 120 seconds for the ephemeral disk to appear](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/cloudinit/sources/DataSourceAzure.py#L1454-L1473) before either proceeding or giving up. It waits up to 120 secs for the symlink `/dev/disk/cloud/azure_resource` to appear.

For new Azure VM SKUs (such as `Dv4`, `Dsv4`, `Ev4`, `Esv4`) that do not come with ephemeral resource disks, cloud-init would wait up to 120 seconds before giving up. See [LP: #1901011](https://bugs.launchpad.net/cloud-init/+bug/1901011).

For these new Azure VM SKUs without ephemeral resource disks, the `disk_setup` module would also fail later in cloud-init because "builtin Azure ephemeral disk configs" are merged into DataSourceAzure metadata. These builtin configs reference the non-existent ephemeral disk, which causes the module to fail.

#### Why this approach was chosen

As of today, the Azure Instance Metadata Service (Azure IMDS) does not expose VM instance metadata indicating whether an ephemeral resource disk exists for the VM or not.

The Azure host also guarantees that the ephemeral resource disk is attached to the VM before it is booted during VM deployment.

Additionally, the ephemeral resource disk symlink (`/dev/disk/cloud/azure_resource`) that cloud-init waits for is actually created by [a udev rule that comes with cloud-init](https://github.com/canonical/cloud-init/blob/master/udev/66-azure-ephemeral.rules). Additional relevant [code](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/udev/66-azure-ephemeral.rules#L10), [code](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/udev/66-azure-ephemeral.rules#L26), and [code](https://github.com/canonical/cloud-init/blob/54e202a6480e48dbb8a72004f7a5003f7c4edfae/udev/66-azure-ephemeral.rules#L29-L31).

Because:
* the Azure host guarantees that resource disks are attached for VM SKUs that have resource disks, 
* because the symlinks are created by `udev` rules (created as soon as the kernel detects the disk and sends the event to `udev`),
* and because `udevadm settle` is invoked very early in boot by cloud-init (before DataSourceAzure runs),
it is **guaranteed** that the ephemeral resource disk symlink exists by the time DataSourceAzure runs.

#### No regression for VM SKUs _with_ ephemeral resource disk

* Created a custom image with this branch's cloud-init installed.
* Deployed a `Standard_DS1_V2` VM (has ephemeral resource disk) from this custom image.
* Cloud-init logs here: https://paste.ubuntu.com/p/qdxkfwYKRD/

```
2021-02-03 23:23:31,913 - __init__.py[DEBUG]: Found unstable nic names: ['eth0']; calling udevadm settle
2021-02-03 23:23:31,914 - subp.py[DEBUG]: Running command ['udevadm', 'settle'] with allowed return codes [0] (shell=False, capture=True)
2021-02-03 23:23:31,933 - util.py[DEBUG]: Waiting for udev events to settle took 0.019 seconds
...
2021-02-03 23:23:32,823 - handlers.py[DEBUG]: finish: azure-ds/crawl_metadata: SUCCESS: crawl_metadata
2021-02-03 23:23:32,823 - util.py[DEBUG]: Crawl of metadata service took 1.117 seconds
...
2021-02-03 23:23:32,824 - azure.py[DEBUG]: Ephemeral resource disk '/dev/disk/cloud/azure_resource' exists. Merging default Azure cloud ephemeral disk configs.
...
2021-02-03 23:23:38,563 - handlers.py[DEBUG]: start: azure-ds/activate: activate
2021-02-03 23:23:38,563 - handlers.py[DEBUG]: start: azure-ds/address_ephemeral_resize: address_ephemeral_resize
2021-02-03 23:23:38,564 - azure.py[DEBUG]: Ephemeral resource disk '/dev/disk/cloud/azure_resource' exists.
...
2021-02-03 23:23:38,894 - handlers.py[DEBUG]: finish: azure-ds/address_ephemeral_resize: SUCCESS: address_ephemeral_resize
```

#### Fix for VM SKUs _without_ ephemeral resource disk

* Created a custom image with this branch's cloud-init installed.
* Deployed a `Standard_D2s_v4` VM (_no_ ephemeral resource disk) from this custom image.
* Cloud-init logs here: https://paste.ubuntu.com/p/GmCqBNTQrG/

```
2021-02-03 23:20:53,217 - __init__.py[DEBUG]: Found unstable nic names: ['eth0']; calling udevadm settle
2021-02-03 23:20:53,217 - subp.py[DEBUG]: Running command ['udevadm', 'settle'] with allowed return codes [0] (shell=False, capture=True)
2021-02-03 23:20:53,235 - util.py[DEBUG]: Waiting for udev events to settle took 0.018 seconds
...
2021-02-03 23:20:54,388 - handlers.py[DEBUG]: finish: azure-ds/crawl_metadata: SUCCESS: crawl_metadata
2021-02-03 23:20:54,389 - util.py[DEBUG]: Crawl of metadata service took 1.363 seconds
...
2021-02-03 23:20:54,389 - azure.py[DEBUG]: Ephemeral resource disk '/dev/disk/cloud/azure_resource' does not exist. Not merging default Azure cloud ephemeral disk configs.
...
2021-02-03 23:20:59,846 - handlers.py[DEBUG]: start: azure-ds/activate: activate
2021-02-03 23:20:59,846 - handlers.py[DEBUG]: start: azure-ds/address_ephemeral_resize: address_ephemeral_resize
2021-02-03 23:20:59,847 - azure.py[DEBUG]: Ephemeral resource disk '/dev/disk/cloud/azure_resource' does not exist.
2021-02-03 23:20:59,847 - handlers.py[DEBUG]: finish: azure-ds/address_ephemeral_resize: SUCCESS: address_ephemeral_resize
```

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_lp1901011.py
+++ b/tests/integration_tests/bugs/test_lp1901011.py
@@ -4,8 +4,6 @@ Ensure an ephemeral disk exists after boot.
 
 See https://github.com/canonical/cloud-init/pull/800
 """
-import re
-
 import pytest
 
 from tests.integration_tests.clouds import IntegrationCloud
@@ -38,15 +36,23 @@ def test_ephemeral(instance_type, is_ephemeral,
         assert expected_log in log
 
         # Verify devices
-        dev_links = client.execute('ls -l /dev/disk/cloud')
-        assert 'azure_root -> ../../sda' in dev_links
-        assert 'azure_root-part1 -> ../../sda1' in dev_links
+        dev_links = client.execute('ls /dev/disk/cloud')
+        assert 'azure_root' in dev_links
+        assert 'azure_root-part1' in dev_links
         if is_ephemeral:
-            assert 'azure_resource -> ../../sdb' in dev_links
-            assert 'azure_resource-part1 -> ../../sdb1' in dev_links
+            assert 'azure_resource' in dev_links
+            assert 'azure_resource-part1' in dev_links
 
         # Verify mounts
-        blks = client.execute('lsblk')
-        assert re.search('sda1.*part /', blks) is not None
+        blks = client.execute('lsblk -pPo NAME,TYPE,MOUNTPOINT')
+        root_device = client.execute(
+            'realpath /dev/disk/cloud/azure_root-part1'
+        )
+        assert 'NAME="{}" TYPE="part" MOUNTPOINT="/"'.format(
+            root_device) in blks
         if is_ephemeral:
-            assert re.search('sdb1.*part /mnt', blks) is not None
+            ephemeral_device = client.execute(
+                'realpath /dev/disk/cloud/azure_resource-part1'
+            )
+            assert 'NAME="{}" TYPE="part" MOUNTPOINT="/mnt"'.format(
+                ephemeral_device) in blks
--- a/tests/integration_tests/bugs/test_lp1901011.py
+++ b/tests/integration_tests/bugs/test_lp1901011.py
@@ -20,17 +20,33 @@ def test_ephemeral(instance_type, is_ephemeral,
                    session_cloud: IntegrationCloud, setup_image):
     if is_ephemeral:
         expected_log = (
-            'Ephemeral resource disk .* exists. Merging default Azure cloud '
-            'ephemeral disk configs.'
+            "Ephemeral resource disk '/dev/disk/cloud/azure_resource' exists. "
+            "Merging default Azure cloud ephemeral disk configs."
         )
     else:
         expected_log = (
-            'Ephemeral resource disk .* does not exist. Not merging '
-            'default Azure cloud ephemeral disk configs.'
+            "Ephemeral resource disk '/dev/disk/cloud/azure_resource' does "
+            "not exist. Not merging default Azure cloud ephemeral disk "
+            "configs."
         )
 
     with session_cloud.launch(
         launch_kwargs={'instance_type': instance_type}
     ) as client:
+        # Verify log file
         log = client.read_from_file('/var/log/cloud-init.log')
-        assert re.search(expected_log, log) is not None
+        assert expected_log in log
+
+        # Verify devices
+        dev_links = client.execute('ls -l /dev/disk/cloud')
+        assert 'azure_root -> ../../sda' in dev_links
+        assert 'azure_root-part1 -> ../../sda1' in dev_links
+        if is_ephemeral:
+            assert 'azure_resource -> ../../sdb' in dev_links
+            assert 'azure_resource-part1 -> ../../sdb1' in dev_links
+
+        # Verify mounts
+        blks = client.execute('lsblk')
+        assert re.search('sda1.*part /', blks) is not None
+        if is_ephemeral:
+            assert re.search('sdb1.*part /mnt', blks) is not None
--- a/tests/integration_tests/bugs/test_lp1901011.py
+++ b/tests/integration_tests/bugs/test_lp1901011.py
@@ -1,0 +1,36 @@
+"""Integration test for LP: #1901011
+
+Ensure an ephemeral disk exists after boot.
+
+See https://github.com/canonical/cloud-init/pull/800
+"""
+import re
+
+import pytest
+
+from tests.integration_tests.clouds import IntegrationCloud
+
+
+@pytest.mark.azure
+@pytest.mark.parametrize('instance_type,is_ephemeral', [
+    ('Standard_DS1_v2', True),
+    ('Standard_D2s_v4', False),
+])
+def test_ephemeral(instance_type, is_ephemeral,
+                   session_cloud: IntegrationCloud, setup_image):
+    if is_ephemeral:
+        expected_log = (
+            'Ephemeral resource disk .* exists. Merging default Azure cloud '
+            'ephemeral disk configs.'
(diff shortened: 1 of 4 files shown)


johnsonshi:
@raharper @smoser @anhvoms The [previous PR](https://github.com/canonical/cloud-init/pull/716) was closed due to inactivity as I was away for a short while. I have reopened it with the suggested changes. Dropped the wait entirely as the platform guarantees the existence of the resource disk for VMs that come with them, and `udevadm settle` is invoked anyways by cloud-init early in boot.

Also updated the PR message above with new info.

smoser (approved):
Given the unfortunate situation of not being able to know if we should expect a given disk or not, I think this is this approach is the best we can do.

My only concern is that we'll only know of failures when a user reports a problem, and they're not even likely to immediately notice a problem.

I guess the way to address that would be in cloud-tests.  The test *would* be able to know (based on the requested instance-type or details) if it expected a disk or not. It could launch one of each and expect to find the disk or expect to have no WARN/errors in the log about one not being there.

@OddBloke or @TheRealFalcon is something like that possible?

TheRealFalcon:
Yes, I think something like that shouldn't be too hard.

TheRealFalcon:
@johnsonshi I created an example integration test that could be used/adapted here (note that is also requires the [latest pycloudlib](https://github.com/TheRealFalcon/cloud-init/blob/ephemeral-test/integration-requirements.txt#L4)):

https://github.com/TheRealFalcon/cloud-init/blob/ephemeral-test/tests/integration_tests/bugs/test_lp1901011.py

tests/integration_tests/bugs/test_lp1901011.py
```diff
@@ -0,0 +1,52 @@
...
+        log = client.read_from_file('/var/log/cloud-init.log')
+        assert expected_log in log
+
+        # Verify devices
+        dev_links = client.execute('ls -l /dev/disk/cloud')
+        assert 'azure_root -> ../../sda' in dev_links
+        assert 'azure_root-part1 -> ../../sda1' in dev_links
+        if is_ephemeral:
+            assert 'azure_resource -> ../../sdb' in dev_links
+            assert 'azure_resource-part1 -> ../../sdb1' in dev_links
```
OddBloke, line 46: I'm not certain, but I think the reason for these symlinks existing is that the underlying device names are not necessarily stable. I think this check is still good if you just `s/ ->.*//` each of the strings?

tests/integration_tests/bugs/test_lp1901011.py
```diff
@@ -0,0 +1,52 @@
...
+
+        # Verify mounts
+        blks = client.execute('lsblk')
+        assert re.search('sda1.*part /', blks) is not None
+        if is_ephemeral:
+            assert re.search('sdb1.*part /mnt', blks) is not None
```
OddBloke, line 52: Given the above, probably want to dereference the named symlinks for these disks and check for those?

OddBloke (approved):
Integration test LGTM, and passes locally. Thanks!

PR #801, opened 2021-02-05
test_gh668: fix failure on LXD VMs

OddBloke:
## Proposed Commit Message
```
test_gh668: fix failure on LXD VMs

In LXD containers, the default interface is named eth0.  In VMs, it
isn't; it's renamed by systemd (likely to enp5s0, but we can't rely on
that).  This means that, on VMs, the network configuration we specify
for "eth0" doesn't match an interface in the system and so is not
applied.

This modifies the test to set a MAC address in a match clause in the
network configuration and on the eth0 interface (which is the LXD name
in both containers and VMs pre-rename): this ensures that the specified
configuration applies in both cases.
```

## Test Steps

Run the test using a LXD VM.

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_gh668.py
+++ b/tests/integration_tests/bugs/test_gh668.py
@@ -12,6 +12,7 @@ from tests.integration_tests.instances import IntegrationInstance
 
 DESTINATION_IP = "172.16.0.10"
 GATEWAY_IP = "10.0.0.100"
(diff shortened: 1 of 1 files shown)


OddBloke:
This test still causes systemd-networkd to report the interface as "configuring" indefinitely, meaning that we have to wait out the wait-online timeout (~2mins) before cloud-init will complete.

TheRealFalcon (approved):

PR #802, opened 2021-02-05
integration_tests: introduce lxd_use_exec mark

OddBloke:
## Proposed Commit Message
```
integration_tests: introduce lxd_use_exec mark

pycloudlib has modified the way LXD executes tests
(https://github.com/canonical/pycloudlib/pull/114): it will always use
SSH to access them by default, instead of using `lxc exec`.  This
behaviour is transparent for the majority of cloud-init's integration
tests, but some currently depend on using `lxc exec` to access instances
with (intentionally) broken networking: obviously these are not
accessible via SSH.

pycloudlib retains support for switching an instance to use `lxc exec`.
This commit introduces the `lxd_use_exec` mark, which tests can use to
indicate to the integration testing framework that they should be so
switched, and applies it to all applicable tests.
```

## Additional Context

See https://github.com/canonical/pycloudlib/pull/114

## Test Steps

CI tests will indicate that LXD containers work, a manual run is required for LXD VMs.

## Checklist:

### DIFF
--- a/integration-requirements.txt
+++ b/integration-requirements.txt
@@ -1,5 +1,5 @@
 # PyPI requirements for cloud-init integration testing
 # https://cloudinit.readthedocs.io/en/latest/topics/integration_tests.html
 #
-pycloudlib @ git+https://github.com/canonical/pycloudlib.git@878981e3c7caaf583a8c7c5494dba9d9447acee8
+pycloudlib @ git+https://github.com/canonical/pycloudlib.git@3a6c668fed769f00d83d1e6bea7d68953787cc38
 pytest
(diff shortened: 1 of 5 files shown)


blackboxsw (commented):
As written this assumes the a test will always interact using `lxc exec` instead of ssh. Do you see any use-cases that we want to plan for where we really only want `lxc exec` for some commands and ssh for others?

OddBloke:
> Do you see any use-cases that we want to plan for where we really only want `lxc exec` for some commands and ssh for others?

None immediately present themselves to me. Such cases _can_ be handled today, something like:

```py
@pytest.mark.lxc_use_exec
def test_example(client):
    client.execute("unbreak-ssh-access-somehow")
    client.instance.execute_via_ssh = True
    client.execute("echo ssh works")
```

If we see more than one or two such cases, we should look at improving the API.

TheRealFalcon (approved):

blackboxsw (approved):
Thanks LGTM

PR #803, opened 2021-02-07
fqdn should be used as hostname

taoyama:
## Proposed Commit Message
_Set Hostname_ Document say:
[If both ``fqdn`` and ``hostname`` are set, ``fqdn`` will be used.](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#:~:text=If%20both%20fqdn%20and%20hostname%20are%20set%2C%20fqdn%20will%20be%20used.)

## Checklist:

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -257,9 +257,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                         "hostname to %s", hostname)
 
     def _select_hostname(self, hostname, fqdn):
-        # Prefer the short hostname over the long
-        # fully qualified domain name
-        if not hostname:
+        # If both fqdn and hostname are set, fqdn will be used.
+        if fqdn:
             return fqdn
         return hostname
 
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -37,6 +37,7 @@ slyon
 smoser
 sshedi
 TheRealFalcon
+taoyama
 tnt-dev
 tomponline
 tsanghan


TheRealFalcon:
Hi @taoyama , it's unfortunate that our documentation doesn't line up with the current implementation, but hostname is central to the functioning of other aspects of cloud-init and many instances after they've been booted. Given such a large regression potential, I don't think this is a change we can realistically make right now.

Is there a specific problem you're hitting that requires the fqdn from this call? Could we simply update the documentation instead?

taoyama:
Hi @TheRealFalcon , Thanks for your reply.

Yes. I understand a large regression potential risk.
I made this pull request because I realized that the behavior of `hostname` and `fqdn` are differ across distributions.
Distro like RHEL is already use fqdn over hostname.

[cloudinit/distros/rhel.py](https://github.com/canonical/cloud-init/blob/11fa714f3a7b031657b642d0c2f0155854d5ee0e/cloudinit/distros/rhel.py)
```python
    def _select_hostname(self, hostname, fqdn):
        # Should be fqdn if we can use it
        # See: https://www.centos.org/docs/5/html/Deployment_Guide-en-US/ch-sysconfig.html#s2-sysconfig-network # noqa
        if fqdn:
            return fqdn
        return hostname
```

I believe this is very confusing for users.
Updating documentation would be fine though consistency across distributions for hostname configuration should be preferable.
What are your thoughts on this?

TheRealFalcon:
> consistency across distributions for hostname configuration should be preferable

I agree that consistency is preferable, but not introducing regressions is an even higher priority.

 If we go back through git history, we can see that initially `_select_hostname` in `__init__.py` was abstract, and we left it to each distribution to select the hostname. We also initially started with the short hostname everywhere, but do to various bugs (e.g., https://bugs.launchpad.net/cloud-init/+bug/1076759 and https://bugs.launchpad.net/cloud-init/+bug/1246485), it evolved the way it has. Consistency would be great, but that's just really not an option anymore.

We could update the docs to say that it is distro dependent whether short name or fqdn is used.

taoyama:
> I agree that consistency is preferable, but not introducing regressions is an even higher priority.

That is absolutely correct.
I didn't even know the history. Thank you for your information.

> We could update the docs to say that it is distro dependent whether short name or fqdn is used.

Agreed. I would make a pull request to update the docs.

taoyama:
Closing this pull request as I opened #818.

PR #805, opened 2021-02-08
Remove wait argument from tests with session_cloud calls

TheRealFalcon:
During #788 , I forgot to update the calls to `session_cloud.launch(...)` to remove the `wait` argument.


OddBloke (approved):
Thanks!

PR #808, opened 2021-02-10
Port apt cloud_tests to integration tests

TheRealFalcon:
## Proposed Commit Message
Port apt cloud_tests to integration tests

## Additional Context
This should cover all of the apt_*.py test cases currently in the cloud tests. I ported things as close to 1:1 as possible, but added some different PPA sources to ensure a source from one test wasn't affected a check in another.

I also question the value of a couple of these tests. In particular, the ones under `TestDefaults`, and wondering if we need 3 different tests for what's in `test_key`, `test_keyserver`, and `test_ppa`. I think it's fine to keep them, but if reviewers have strong opinions on deleting anything, let me know.

## Test Steps
pytest tests/integration_tests/modules/test_apt.py

## Checklist:


tests/integration_tests/modules/test_apt.py
```diff
@@ -0,0 +1,291 @@
...
+    deb-src $PRIMARY $RELEASE universe restricted
+    deb $SECURITY $RELEASE-security multiverse
+    deb-src $SECURITY $RELEASE-security multiverse
+  sources:
+    test_keyserver:
+        keyid: 72600DB15B8E4C8B1964B868038ACC97C660A937
+        keyserver: keyserver.ubuntu.com
+        source: "deb http://ppa.launchpad.net/cloud-init-raharper/curtin-dev/ubuntu $RELEASE main"
+    test_ppa:
+      keyid: 3552C902B4DDF7BD38421821015D28D7441614D8
```
blackboxsw, line 43: One thing I noticed is that in the new tests we are using the full fingerprint not a short 8 character fingerprint. It's not critical, but since we have two keyid examples, we could have one use a short fingerprint instead so we are exercising more failure paths (even though it's not really a cloud-init failure path but a tooling path)

tests/integration_tests/modules/test_apt.py
```diff
@@ -0,0 +1,291 @@
...
+apt_pipelining: os
+"""  # noqa: E501
+
+EXPECTED_REGEXES = [
+    r"deb http://badarchive.ubuntu.com/ubuntu [a-z].* main restricted",
+    r"deb-src http://badarchive.ubuntu.com/ubuntu [a-z].* main restricted",
+    r"deb http://badarchive.ubuntu.com/ubuntu [a-z].* universe restricted",
+    r"deb-src http://badarchive.ubuntu.com/ubuntu [a-z].* universe restricted",
+    r"deb http://badsecurity.ubuntu.com/ubuntu [a-z].*security multiverse",
+    r"deb-src http://badsecurity.ubuntu.com/ubuntu [a-z].*security multiverse",
```
blackboxsw, line 86: To be a bit more strict
```suggestion
    r"deb http://badarchive.ubuntu.com/ubuntu [a-z]+ main restricted",
    r"deb-src http://badarchive.ubuntu.com/ubuntu [a-z]+ main restricted",
    r"deb http://badarchive.ubuntu.com/ubuntu [a-z]+ universe restricted",
    r"deb-src http://badarchive.ubuntu.com/ubuntu [a-z]+ universe restricted",
    r"deb http://badsecurity.ubuntu.com/ubuntu [a-z]+-security multiverse",
    r"deb-src http://badsecurity.ubuntu.com/ubuntu [a-z]+-security multiverse",
```

blackboxsw (approved):
Looks good, I appreciate reducing and consolidating the test cases anywhere we have the ability to test non-conflicting features. Integration tests are too time/CPU costly for setup/teardown to not leverage some of that setup for multiple tests where applicable. Even in the test/cloud_tests we talked about trying to combine some tests that would not be destructive for other feature tests. Expedite test where we can and where it doesn't cause us problems in terms of getting a valid signal on a test failure.

PR #809, opened 2021-02-11
integration_tests: fix test_gh626 on LXD VMs via SSH

OddBloke:
## Proposed Commit Message
```
integration_tests: fix test_gh626 on LXD VMs via SSH

Without a MAC address match clause, the test network configuration is
not applied to the primary interface in LXD VMs (which is named enp*s*
rather than eth0). This means that networking does not come up, so
SSH access is not possible.
```

## Test Steps

Run the test on a LXD VM (with either an SSH key configured, or with the new pycloudlib which uses SSH by default) and observe success.

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_gh626.py
+++ b/tests/integration_tests/bugs/test_gh626.py
@@ -11,13 +11,16 @@ from tests.integration_tests.clouds import ImageSpecification
 from tests.integration_tests.instances import IntegrationInstance
 
 
+MAC_ADDRESS = "de:ad:be:ef:12:34"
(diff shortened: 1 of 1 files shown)


OddBloke:
This test has the same bug that all other tests which are setting a MAC address do: it uses the same MAC address everywhere, so if instances from previous runs are around or tests are being run in parallel, failures will be observed due to traffic being routed to the "wrong" `DE:AD:BE:EF:12:34` interface.

I'll address that generally in a follow-up.

TheRealFalcon (approved):

PR #813, opened 2021-02-17
integration_tests: use unique MAC addresses for tests

OddBloke:
## Proposed Commit Message
```
integration_tests: use unique MAC addresses for tests

Using the same MAC address results in strange test behaviour if more
than one such instance is up: traffic gets routed to an arbitrary
interface with the given MAC address.  This can happen if running tests
in parallel, or on a system which retains test instances from previous
runs.

The introduction of tests/integration_tests/__init__.py means that
pylint now checks the integration tests: this commit also addresses
those failures.
```

## Additional Info

This introduces `tests/integration_tests/__init__.py` as a holding place for the utility function this introduces; I didn't want to introduce a `util`, but it also didn't feel like it fit well in anything we currently had.

## Test Steps

Run the affected tests in parallel a few times and observe a lack of collisions.

## Checklist:

### DIFF
--- a/tests/integration_tests/__init__.py
+++ b/tests/integration_tests/__init__.py
@@ -1,0 +1,12 @@
+import random
+
+
+def random_mac_address() -> str:
+    """Generate a random MAC address.
+
+    The MAC address will have a 1 in its least significant bit, indicating it
+    to be a locally administered address.
+    """
+    return "02:00:00:%02x:%02x:%02x" % (random.randint(0, 255),
+                                        random.randint(0, 255),
+                                        random.randint(0, 255))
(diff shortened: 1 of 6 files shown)


TheRealFalcon (approved):
Funny thing is I slightly prefer `util.py` to `__init__.py`. `__init__.py` has always seemed more "hidden" to me. But still +1 :smile:

TheRealFalcon:
Did we turn up a setting on pylint? The pylint issues don't seem related to what you've changed.

OddBloke:
> Did we turn up a setting on pylint? The pylint issues don't seem related to what you've changed.

I suspect that adding `__init__` means that pylint now recurses in when previously it didn't? I'll address them.

PR #817, opened 2021-02-20
tools/write-ssh-key-fingerprints: do not display empty header/footer

dermotbradley:
When console output of SSH info is disabled show no headers/footers.

Prevent risk of message text being interpreted as "logger" option by
appending "--" to logger options.

Correct syslog output that was tagged with "ec2" regardless of DataSource
in use. Now use "cloud-init" tag instead.

Various "shellcheck" corrections.

Add testcase for disabled output of SSH host keys.

LP: #1915460

Provides alternative solution to #811.

## Test Steps

In user-data config add either, or both, of the following lines:

```
ssh_fp_console_blacklist: [ 'ssh-dsa', 'ssh-dss', 'ssh-rsa', 'ssh-ed25519' ]
ssh_key_console_blacklist: [ 'ssh-dsa', 'ssh-dss', 'ssh-rsa', 'ssh-ed25519' ]
```

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_keys_to_console.py
+++ b/tests/integration_tests/modules/test_keys_to_console.py
@@ -10,6 +10,11 @@ ssh_fp_console_blacklist: [ssh-dss, ssh-dsa, ecdsa-sha2-nistp256]
 ssh_key_console_blacklist: [ssh-dss, ssh-dsa, ecdsa-sha2-nistp256]
 """
 
+BLACKLIST_ALL_KEYS_USER_DATA = """\
+#cloud-config
+ssh_fp_console_blacklist: [ssh-dsa, ssh-ecdsa, ssh-ed25519, ssh-rsa, ssh-dss, ecdsa-sha2-nistp256]
+"""  # noqa: E501
+
 DISABLED_USER_DATA = """\
 #cloud-config
 ssh:
@@ -29,6 +34,20 @@ class TestKeysToConsoleBlacklist:
     def test_included_keys(self, class_client, key_type):
         syslog = class_client.read_from_file("/var/log/syslog")
         assert "({})".format(key_type) in syslog
+
+
+@pytest.mark.user_data(BLACKLIST_ALL_KEYS_USER_DATA)
+class TestAllKeysToConsoleBlacklist:
+    """Test that when key blacklist contains all key types that
+    no header/footer are output.
+    """
+    def test_header_excluded(self, class_client):
+        syslog = class_client.read_from_file("/var/log/syslog")
+        assert "BEGIN SSH HOST KEY FINGERPRINTS" not in syslog
+
+    def test_footer_excluded(self, class_client):
+        syslog = class_client.read_from_file("/var/log/syslog")
+        assert "END SSH HOST KEY FINGERPRINTS" not in syslog
 
 
 @pytest.mark.user_data(DISABLED_USER_DATA)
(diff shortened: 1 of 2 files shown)


TheRealFalcon (changes requested):
Overall, this change looks good. Might be worth making a function out of the shellcheck comment  and the `logger` options to cleanup the calls to `logger`.

I appreciate the additional cloud tests, but unfortunately these tests are deprecated now. Docs for the new test framework are [here](https://cloudinit.readthedocs.io/en/latest/topics/integration_tests.html), but the TLDR for this PR would be to add a test to [test_keys_to_console.py](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/modules/test_keys_to_console.py) with a full blacklist in the cloud-config and assert there's no headers/footers (similar to the bottom two tests in the file)

dermotbradley:
Hi there

I'll find some time in the next few days to work on this.

Regarding the correction of the 'ec2' string - @OddBloke mentioned on IRC that there might be some stuff out there (Subquity?) relying on that value and he would look into it further. I can drop this portion of the PR if its likely to cause problems.

dermotbradley:
> Overall, this change looks good. Might be worth making a function out of the shellcheck comment and the `logger` options to cleanup the calls to `logger`.

Done.

> the TLDR for this PR would be to add a test to [test_keys_to_console.py](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/modules/test_keys_to_console.py) with a full blacklist in the cloud-config and assert there's no headers/footers (similar to the bottom two tests in the file)

Done.

dermotbradley:
@mitechie ping!

mitechie:
pong! I've cleared the tag and we'll get another round of reviews going. Thanks for the ping.

tests/integration_tests/modules/test_keys_to_console.py
```diff
@@ -10,6 +10,11 @@
 ssh_key_console_blacklist: [ssh-dss, ssh-dsa, ecdsa-sha2-nistp256]
 """
 
+BLACKLIST_ALL_KEYS_USER_DATA = """\
+#cloud-config
+ssh_key_console_blacklist: [ssh-dsa, ssh-ecdsa, ssh-ed25519, ssh-rsa]
+"""
```
TheRealFalcon, line 16:
```suggestion
BLACKLIST_ALL_KEYS_USER_DATA = """\
#cloud-config
ssh_fp_console_blacklist: [ssh-dsa, ssh-ecdsa, ssh-ed25519, ssh-rsa, ssh-dss, ecdsa-sha2-nistp256]
"""  # noqa: E501
```

tests/integration_tests/modules/test_keys_to_console.py
```diff
@@ -31,6 +36,21 @@ def test_included_keys(self, class_client, key_type):
         assert "({})".format(key_type) in syslog
 
 
+@pytest.mark.user_data(BLACKLIST_ALL_KEYS_USER_DATA)
+class TestAllKeysToConsoleBlacklist:
+    """Test that when key blacklist contains all key types that
+    no header/footer are output.
+    """
+    @pytest.mark.parametrize("key_type", ["DSA", "ECDSA", "ED25519", "RSA"])
```
TheRealFalcon, line 44:
```suggestion
```

TheRealFalcon (changes requested):
Looking good. Just a couple changes the integration test needed

tests/integration_tests/modules/test_keys_to_console.py
dermotbradley: Done

tests/integration_tests/modules/test_keys_to_console.py
dermotbradley: Done

TheRealFalcon (approved):
Thanks!

PR #821, opened 2021-02-23
Add helper script: tools/build-deb-prepare

paride:
## Proposed Commit Message

```
Add helper script: tools/build-deb-prepare
    
The script grabs the debian/ packaging directory from a packaging branch
(default: ubuntu/devel) and generates a suitable d/changelog entry and
an orig tarball. It is suitable for quickly building a deb package from
the current branch via debuild or sbuild.
```

## Additional Context

Will also be useful to run integration testing from Jenkins from a devel branch.

## Test Steps
Assuming that the cloud-init build-deps are installed this should produce a deb package from the current branch using the packaging from `ubuntu/devel`:
```
tools/build-deb-prepare
debuild -us -uc
```
## Checklist:

### DIFF
--- a/tools/build-deb-prepare
+++ b/tools/build-deb-prepare
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+: "${PACKAGING_BRANCH:=ubuntu/devel}"
+: "${RELEASE:=UNRELEASED}"
+: "${CLOGVER_DEBIAN:=0ubuntu1}"
+: "${DEBFULLNAME:=$(git config user.name)}"
+: "${DEBEMAIL:=$(git config user.email)}"
+
+export DEBFULLNAME DEBEMAIL
+
+fail() { echo "$@" 1>&2; exit 1; }
+
+if [[ "$1" = "-h" || "$1" = "--help" ]]; then
+    cat <<-EOF
+	Usage: ${0##*/}
+	Prepare the source for building a .deb package via e.g. debuild(1).
+	Build a debian/ directory the current git HEAD pulling the debian/
+	directory from branch '$PACKAGING_BRANCH' and updating the changelog,
+	then prepare an .orig tarball matching the new changelog entry.
+	EOF
+    exit
+fi
+
+git clean -f debian || fail "Failed to run git clean."
+
+# This script could work even if a debian/ directory is already existing,
+# but it's a confusing setup that should be avoided.
+[[ ! -e debian ]] || fail "A debian/ directory already exists."
+
+[[ $DEBFULLNAME ]] || fail "Must have DEBFULLNAME or git user.name set."
+[[ $DEBEMAIL ]] || fail "Must have DEBEMAIL or git user.email set."
+
+if [[ $(git status -s) ]]; then
+    cat 1>&2 <<-EOF
+	WARNING: There are uncommitted changes in your working directory.
+	         These changes will not be included in the orig archive.
+	EOF
+fi
+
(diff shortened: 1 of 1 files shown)


paride:
As @TheRealFalcon noded this in part overlaps with `packages/bddeb`. The main difference between the two tools is that `bddeb` uses the skel `debian/` packaging from https://github.com/canonical/cloud-init/tree/master/packages/debian, while `build-deb-prepare` pulls the `debian/` packaging directly from the `ubuntu/<release>` packaging branches, which is basically what the LP recipe building the daily deb does, see for example [1]. The idea is to get a deb which is as similar as possible to what we might release at a given point.

https://code.launchpad.net/~cloud-init-dev/+recipe/cloud-init-daily-focal

paride:
Given that my aim here is to replicate a daily build but from an arbitrary branch, there could be a better way. We could have a Jenkins job build a recipe from a template, in the same format of e.g. [1], and then use `git-build-recipe`  (more likely: `git-build-recipe --allow-fallback-to-native`) to build the source package in the same way Launchpad does. The job can then debuild/sbuild it and run the integration tests. I still have to try, but I don't see why it shouldn't work.

What do you think? Do you think this `build-deb-prepare` script could still be useful?

[1] https://code.launchpad.net/~cloud-init-dev/+recipe/cloud-init-daily-devel

tools/build-deb-prepare
```diff
@@ -0,0 +1,53 @@
...
+	EOF
+fi
+
+# git describe will output something either like '0.1.0' (a tag)
+# or TAG-N-gHASH where N is number of commits since TAG
+uver=$(git describe --abbrev=8) ||
+    fail "Failed to get upstream version with 'git describe'."
+clogver_new="${uver}-${CLOGVER_DEBIAN}"
+
+git fetch "$(git remote)" "$PACKAGING_BRANCH:$PACKAGING_BRANCH" ||
```
OddBloke, line 46: This fails for me locally, because:

```
$ git remote | wc -l
22
```

OddBloke:
This doesn't work well with our `build-package` script, which bundles up the git tree to ensure that uncommitted local modifications aren't included. If I don't use `build-package`, then `debuild` complains about unrepresentable changes to `.tox`.

I wonder if we should fold this functionality into `build-package`, so it applies the debian/ directory to the pristine checkout (perhaps optionally)?

tools/build-deb-prepare
paride: Heh, you spotted the latest line I added to the script, without checking it well enough apparently :)

The script works fine without that line, but a local copy of the packaging branch has to exist, otherwise we can't `git restore` from it. When cloning the `cloud-init` repo from scratch no local copy of the packaging branches exists. My attempt here was to create a local copy of the packaging branch without doing a `git checkout` roundtrip, e.g.:
```
git checkout ubuntu/focal
git checkout master
```
I couldn't find way to do so, and my attempt was buggy, as you found out. I'm not sure if git allows this.

paride:
Thanks for looking at this @OddBloke. I agree it makes sense to fold this into `build-package`, or more in general to add this script to the `uss-tableflip` toolbox instead of adding it to `cloud-init`. It's quite generic after all and may apply to other project using a similar branch structure.

I'll let this PR survive up to the next standup just in case there is anything to discuss, but in general I'm fine with the NACK for merging it in cloud-init

paride:
Closing this for now.

PR #829, opened 2021-03-03
net: exclude OVS internal interfaces in get_interfaces

OddBloke:
## Proposed Commit Message
```
net: exclude OVS internal interfaces in get_interfaces

`get_interfaces` is used to in two ways, broadly: firstly, to determine
the available interfaces when converting cloud network configuration
formats to cloud-init's network configuration formats; and, secondly, to
ensure that any interfaces which are specified in network configuration
are (a) available, and (b) named correctly.  The first of these is
unaffected by this commit, as no clouds support Open vSwitch
configuration in their network configuration formats.

For the second, we check that MAC addresses of physical devices are
unique.  In some OVS configurations, there are OVS-created devices which
have duplicate MAC addresses, either with each other or with physical
devices.  As these interfaces are created by OVS, we can be confident
that (a) they will be available when appropriate, and (b) that OVS will
name them correctly.  As such, this commit excludes any OVS-internal
interfaces from the set of interfaces returned by `get_interfaces`.

LP: #1912844
```

## Test Steps

Use included integration test.

## Checklist:

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -6,6 +6,7 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 import errno
+import functools
 import ipaddress
 import logging
 import os
@@ -19,6 +20,19 @@ from cloudinit.url_helper import UrlError, readurl
 LOG = logging.getLogger(__name__)
 SYS_CLASS_NET = "/sys/class/net/"
 DEFAULT_PRIMARY_INTERFACE = 'eth0'
+OVS_INTERNAL_INTERFACE_LOOKUP_CMD = [
+    "ovs-vsctl",
+    "--format",
+    "csv",
+    "--no-headings",
+    "--timeout",
+    "10",
+    "--columns",
+    "name",
+    "find",
+    "interface",
+    "type=internal",
+]
 
 
 def natural_sort_key(s, _nsre=re.compile('([0-9]+)')):
@@ -131,6 +145,52 @@ def master_is_openvswitch(devname):
         return False
     ovs_path = sys_dev_path(devname, path="upper_ovs-system")
     return os.path.exists(ovs_path)
+
+
+@functools.lru_cache(maxsize=None)
+def openvswitch_is_installed() -> bool:
+    """Return a bool indicating if Open vSwitch is installed in the system."""
+    ret = bool(subp.which("ovs-vsctl"))
+    if not ret:
+        LOG.debug(
+            "ovs-vsctl not in PATH; not detecting Open vSwitch interfaces"
+        )
+    return ret
+
+
+@functools.lru_cache(maxsize=None)
+def get_ovs_internal_interfaces() -> list:
+    """Return a list of the names of OVS internal interfaces on the system.
+
+    These will all be strings, and are used to exclude OVS-specific interface
+    from cloud-init's network configuration handling.
+    """
+    try:
+        out, _err = subp.subp(OVS_INTERNAL_INTERFACE_LOOKUP_CMD)
+    except subp.ProcessExecutionError as exc:
+        if "database connection failed" in exc.stderr:
+            LOG.info(
+                "Open vSwitch is not yet up; no interfaces will be detected as"
+                " OVS-internal"
+            )
+            return []
+        raise
+    else:
+        return out.splitlines()
+
+
+def is_openvswitch_internal_interface(devname: str) -> bool:
+    """Returns True if this is an OVS internal interface.
+
+    If OVS is not installed or not yet running, this will return False.
+    """
+    if not openvswitch_is_installed():
+        return False
+    ovs_bridges = get_ovs_internal_interfaces()
+    if devname in ovs_bridges:
+        LOG.debug("Detected %s as an OVS interface", devname)
+        return True
+    return False
 
 
 def is_netfailover(devname, driver=None):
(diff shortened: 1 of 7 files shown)


tests/integration_tests/bugs/test_lp1912844.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+
+@pytest.mark.lxd_vm
+def test_wip(session_cloud):
+    with session_cloud.launch(
+        user_data=SETUP_USER_DATA,
+    ) as instance:
+        instance.instance.clean()
+        session_cloud.snapshot_id = instance.snapshot()
```
OddBloke, line 54: Looking at this again, I wonder if this is going to mean that this OVS-enabled image is used for all future launches: I'll check now.

cloudinit/net/__init__.py
```diff
@@ -888,6 +948,8 @@ def get_interfaces(blacklist_drivers=None) -> list:
         driver = device_driver(name)
         if driver in blacklist_drivers:
             continue
+        if is_openvswitch_internal_interface(name):
+            continue
```
raharper, line 952: Micro optimization, if you move this up above the driver = device_driver line, we can avoid looking up the device_driver of something that would be skipped.

cloudinit/net/__init__.py
```diff
@@ -19,6 +20,19 @@
...
+    "--format",
+    "csv",
+    "--no-headings",
+    "--timeout",
+    "10",
+    "--columns",
+    "name",
+    "find",
+    "interface",
+    "type=internal",
```
raharper, line 34: How confident are we that this query will find the intersection of system devices and ovs devices?  I read some of the bug discussion; however I'm wondering if we want to enumerate all ovs devices and skip the intersection of system devices and ovs devices?  Would that help future proof from some other ovs interface not type=internal but an interface we need to skip?

raharper (commented):
This looks well done.  Some in-line questions.

cloudinit/net/__init__.py
OddBloke: > How confident are we that this query will find the intersection of system devices and ovs devices? I read some of the bug discussion; however I'm wondering if we want to enumerate all ovs devices and skip the intersection of system devices and ovs devices?

Part of the issue is that it isn't clear to me that we _only_ want to consider devices that OVS has as type=system. Consider `bond0` in an instance launched by the integration test from this PR:

```
# ovs-vsctl --columns=name,type find interface type!=internal
name                : bond0
type                : ""
```

While we will skip `bond0` for other reasons (it _is_ a bond per `/sys/class/net`), it's not clear to me that we want to reject it _because of its state in OVS_.

Expressed differently: I'm not sure what classes of interface will have a type that isn't "internal" or "system" but, in the configurations I've seen, "type=internal" skips every interface that isn't presented "normally" through `/sys/class/net`. I think those are the ones we need to exclude.

> Would that help future proof from some other ovs interface not type=internal but an interface we need to skip?

Potentially, but it could also mean that we skip a class of interfaces that we _shouldn't_. Given that we don't _know_, I'm erring on the side of caution for now: only exclude interfaces that we're sure should be excluded.

cloudinit/net/__init__.py
OddBloke: Good point; done.

cloudinit/net/__init__.py
```diff
@@ -133,6 +147,52 @@ def master_is_openvswitch(devname):
     return os.path.exists(ovs_path)
 
 
+@functools.lru_cache()
```
TheRealFalcon, line 150: Nit: `maxsize=None` (and on the next function)

tests/integration_tests/bugs/test_lp1912844.py
OddBloke: It is; working on this now.

TheRealFalcon (commented):
I think main implementation and unit tests look good. I'll revisit when the WIP tag is removed since you're still working on the integration test.

Not necessary for this PR but...
Given the fairly random places I'm seeing
```
@mock.patch(
    "cloudinit.net.is_openvswitch_internal_interface",
    mock.Mock(return_value=False),
)
```
I'm wondering if it might make sense to create a top-level autorun fixture that provides default mocks for things that we obviously don't want running in unit tests. In this case we would default `is_openvswitch_internal_interface` to `False`, and then actual openvswitch tests could change it to `True` when needed.

cloudinit/net/__init__.py
OddBloke: Initially I was going to say:

> Why? Per [the docs](https://docs.python.org/3/library/functools.html#functools.lru_cache):
>
> > If maxsize is set to None, the LRU feature is disabled and the cache can grow without bound.
>
> These functions don't take parameters, so I think the size of the cache will be at most 1.

but I've answered my own question: the `functools` code [special-cases `None`](https://github.com/python/cpython/blob/67148254146948041a77d8a2989f41b88cdb2f99/Lib/functools.py#L549-L562) to make it much simpler; TIL, thanks!

fnordahl (approved):
I'm speaking only for the Open vSwitch specific bits in this patch set, and as we have discussed on the LP this is a good approach and LGTM.

TheRealFalcon (approved):

cloudinit/net/__init__.py
raharper: I'm fine with the cautious approach.  If we find out that there are additional "internal" interfaces that aren't listed in the current query we can look at a different approach of enumeration/filtering.

raharper (approved):

PR #828, opened 2021-03-03
lp-to-git-users: adding B1Sandmann

B1Sandmann:
Mapped from b1sandmann

PR #831, opened 2021-03-04
bonding sysconfig: use bonding_module_opts instead of bonding_opts

B1Sandmann:
Current documentation[1][2] recommends to use BONDING_MODULE_OPTS
instead of BONDING_OPTS. SLES/wicked even ignores BONDING_OPTS
except for the bonding mode and requires BONDING_MODULE_OPTS.

[1] https://wiki.linuxfoundation.org/networking/bonding
[2] https://www.kernel.org/doc/Documentation/networking/bonding.txt

I also changed the name of the render function so that there is no difference between function_name and what it does.

## Additional Context
Currently bonding on SLES (I only tested 15sp2 but it's probably true for prior versions as well as even SLES9 docu only mentions BONDING_MODULE_OPTS) via Configdrive on OpenStack fails because cloud-init sets BONDING_OPTS but SLES ignores the BONDING_OPTS. SLES also requires that the miimon option is set. (SLES [config example](https://documentation.suse.com/sles/15-SP2/single-html/SLES-admin/#sec-network-iface-bonding-hotplug) ) Even if this is set in OpenStack and cloud-init writes it as part of the BONDING_OPTS, SLES ignores it. From what I can tell current best practice seems to be to use BONDING_MODULE_OPTS instead of BONDING_OPTS

## Test Steps
Deploy a SLES 15SP2 on OpenStack ironic that is using port-groups. The ifcfg-bond0 file gets generated  with BONDING_OPTS but SLES will ignore it with a `bonding validation, insufficient monitoring settings` error.

## Checklist:

I found no corresponding documentation to change and I it seems the `tools/net-convert.py` mentioned in `doc/rtd/topics/network-config.rst` is not present any longer.

Requires #828 before it can be merged

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -637,14 +637,15 @@ class Renderer(renderer.Renderer):
                     break
         if bond_opts:
             if flavor == 'suse':
-                # suse uses the sysconfig support which requires BONDING_MODULE_OPTS
-                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
-                # 3.1 Configuration with Sysconfig Support            
+                # suse uses the sysconfig support which requires
+                # BONDING_MODULE_OPTS see
+                # https://www.kernel.org/doc/Documentation/networking/bonding.txt
+                # 3.1 Configuration with Sysconfig Support
                 iface_cfg['BONDING_MODULE_OPTS'] = " ".join(bond_opts)
             else:
                 # rhel uses initscript support and thus requires BONDING_OPTS
-                # this is also the old default
-                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+                # this is also the old default see
+                # https://www.kernel.org/doc/Documentation/networking/bonding.txt
                 #  3.2 Configuration with Initscripts Support
                 iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
 
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -635,18 +635,17 @@ class Renderer(renderer.Renderer):
                         bond_value = " ".join(bond_value)
                     bond_opts.append(value_tpl % (bond_value))
                     break
-        # suse uses the sysconfig support which requires BONDING_MODULE_OPTS
-        # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
-        # 3.1 Configuration with Sysconfig Support
-        if flavor == 'suse':
-            if bond_opts:
+        if bond_opts:
+            if flavor == 'suse':
+                # suse uses the sysconfig support which requires BONDING_MODULE_OPTS
+                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+                # 3.1 Configuration with Sysconfig Support            
                 iface_cfg['BONDING_MODULE_OPTS'] = " ".join(bond_opts)
-        # rhel uses initscript support and thus requires BONDING_OPTS
-        # this is also the old default
-        # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
-        #  3.2 Configuration with Initscripts Support
-        else:
-            if bond_opts:
+            else:
+                # rhel uses initscript support and thus requires BONDING_OPTS
+                # this is also the old default
+                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+                #  3.2 Configuration with Initscripts Support
                 iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
 
     @classmethod
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -313,7 +313,8 @@ class Renderer(renderer.Renderer):
     }
 
     # If these keys exist, then their values will be used to form
-    # a BONDING_OPTS grouping; otherwise no grouping will be set.
+    # a BONDING_OPTS / BONDING_MODULE_OPTS grouping; otherwise no
+    # grouping will be set.
     bond_tpl_opts = tuple([
         ('bond_mode', "mode=%s"),
         ('bond_xmit_hash_policy', "xmit_hash_policy=%s"),
@@ -622,7 +623,7 @@ class Renderer(renderer.Renderer):
                             route_cfg[new_key] = route[old_key]
 
     @classmethod
-    def _render_bonding_opts(cls, iface_cfg, iface):
+    def _render_bonding_opts(cls, iface_cfg, iface, flavor):
         bond_opts = []
         for (bond_key, value_tpl) in cls.bond_tpl_opts:
             # Seems like either dash or underscore is possible?
@@ -634,8 +635,19 @@ class Renderer(renderer.Renderer):
                         bond_value = " ".join(bond_value)
                     bond_opts.append(value_tpl % (bond_value))
                     break
-        if bond_opts:
-            iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
+        # suse uses the sysconfig support which requires BONDING_MODULE_OPTS
+        # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+        # 3.1 Configuration with Sysconfig Support
+        if flavor == 'suse':
+            if bond_opts:
+                iface_cfg['BONDING_MODULE_OPTS'] = " ".join(bond_opts)
+        # rhel uses initscript support and thus requires BONDING_OPTS
+        # this is also the old default
+        # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+        #  3.2 Configuration with Initscripts Support
+        else:
+            if bond_opts:
+                iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
 
     @classmethod
     def _render_physical_interfaces(
@@ -663,7 +675,7 @@ class Renderer(renderer.Renderer):
         for iface in network_state.iter_interfaces(bond_filter):
             iface_name = iface['name']
             iface_cfg = iface_contents[iface_name]
-            cls._render_bonding_opts(iface_cfg, iface)
+            cls._render_bonding_opts(iface_cfg, iface, flavor)
 
             # Ensure that the master interface (and any of its children)
             # are actually marked as being bond types...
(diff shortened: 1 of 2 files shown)


B1Sandmann:
FYI my local tests fail with
```
========================================================================================== FAILURES ==========================================================================================
__________________________________________________________ TestRunCommands.test_run_command_logs_commands_and_exit_codes_to_stderr ___________________________________________________________

self = <test_snap.TestRunCommands testMethod=test_run_command_logs_commands_and_exit_codes_to_stderr>

    def test_run_command_logs_commands_and_exit_codes_to_stderr(self):
        """All exit codes are logged to stderr."""
        outfile = self.tmp_path('output.log', dir=self.tmp)
    
        cmd1 = 'echo "HI" >> %s' % outfile
        cmd2 = 'bogus command'
        cmd3 = 'echo "MOM" >> %s' % outfile
        commands = [cmd1, cmd2, cmd3]
    
        mock_path = 'cloudinit.config.cc_snap.sys.stderr'
        with mock.patch(mock_path, new_callable=StringIO) as m_stderr:
            with self.assertRaises(RuntimeError) as context_manager:
                run_commands(commands=commands)
    
>       self.assertIsNotNone(
            re.search(r'bogus: (command )?not found',
                      str(context_manager.exception)),
            msg='Expected bogus command not found')
E       AssertionError: unexpectedly None : Expected bogus command not found

cloudinit/config/tests/test_snap.py:200: AssertionError
====================================================================================== warnings summary ======================================================================================
../cloudinit-tests-venv/lib64/python3.9/site-packages/_pytest/config/__init__.py:1183
  /home/jens/git/cloudinit-tests-venv/lib64/python3.9/site-packages/_pytest/config/__init__.py:1183: PytestDeprecationWarning: The --strict option is deprecated, use --strict-markers instead.
    self.issue_config_time_warning(

conftest.py:68
  /home/jens/git/cloud-init/conftest.py:68: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(autouse=True)

conftest.py:169
  /home/jens/git/cloud-init/conftest.py:169: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def httpretty():

cloudinit/patcher.py:9
  /home/jens/git/cloud-init/cloudinit/patcher.py:9: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

cloudinit/distros/tests/test_networking.py:18
  /home/jens/git/cloud-init/cloudinit/distros/tests/test_networking.py:18: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def generic_networking_cls():

cloudinit/distros/tests/test_networking.py:44
  /home/jens/git/cloud-init/cloudinit/distros/tests/test_networking.py:44: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def sys_class_net(tmpdir):

cloudinit/sources/tests/test_oracle.py:97
  /home/jens/git/cloud-init/cloudinit/sources/tests/test_oracle.py:97: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def oracle_ds(request, fixture_utils, paths, metadata_version):

cloudinit/sources/tests/test_oracle.py:524
  /home/jens/git/cloud-init/cloudinit/sources/tests/test_oracle.py:524: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(params=[True, False])

cloudinit/tests/test_features.py:14
  /home/jens/git/cloud-init/cloudinit/tests/test_features.py:14: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture()

cloudinit/tests/test_stages.py:355
  /home/jens/git/cloud-init/cloudinit/tests/test_stages.py:355: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def init(self, paths):

cloudinit/tests/test_util.py:718
  /home/jens/git/cloud-init/cloudinit/tests/test_util.py:718: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    def already_mounted_device_and_mountdict(self):

-- Docs: https://docs.pytest.org/en/stable/warnings.html
================================================================================== short test summary info ===================================================================================
FAILED cloudinit/config/tests/test_snap.py::TestRunCommands::test_run_command_logs_commands_and_exit_codes_to_stderr - AssertionError: unexpectedly None : Expected bogus command not found
================================================================== 1 failed, 2672 passed, 10 skipped, 11 warnings in 55.21s ==================================================================
make: *** [Makefile:30: unittest] Fehler 1
```

But these errors are unrelated to my change and also happen if I execute them on the master without my commit. So I assume either my test setup is wrong or there is a issue in current master i.e. 106c57d511b862177c8356685cf6d49ddd6cb55f

B1Sandmann:
I#m waiting for a review here as well. I'm not sure if it's wise of me to rename the function. I can't apply the patch as it is on SLES as SLES is using an older cloud-init version. 

@mitechie could you find someone to take a look with me? Thanks :)

TheRealFalcon:
@B1Sandmann , thanks for the contribution.

My understanding of sysconfig is admittedly very little, but based on the documentation you've linked, it appears that both options are valid depending on the system you're on.

Using [this link](https://wiki.linuxfoundation.org/networking/bonding) you've supplied, under the heading "Configuration with sysconfig support", it specifies to use `BONDING_MODULE_OPTS`, but further down, under the heading "Configuration with initscripts support", it specifies to use `BONDING_OPTS`.

Based on a cursory glance of our source, [rhel appears to support initscripts](https://github.com/canonical/cloud-init/blob/master/cloudinit/distros/rhel.py#L45), whereas [suse does not](https://github.com/canonical/cloud-init/blob/master/cloudinit/distros/opensuse.py#L39).

If this analysis is correct, then I don't think it's correct to simply replace `BONDING_OPTS` with `BONDING_MODULE_OPTS`. We would need to write `BONDING_MODULE_OPTS` if we're on a suse system and `BONDING_OPTS` if we're on a rhel system.

Does that sound correct?

B1Sandmann:
@TheRealFalcon yeah the linuxfoundation link looks a bit outdated. It mentions [SLES 9](https://wiki.linuxfoundation.org/networking/bonding#configuration_with_sysconfig_support) under the heading you mentioned and [Rhel 3 / 4](https://wiki.linuxfoundation.org/networking/bonding#configuration_with_initscripts_support). It also seems to be the website version of https://www.kernel.org/doc/Documentation/networking/bonding.txt
I was able to deploy a rhel 7 on the same node as the SLES15SP2 before and the rhel instance works with both BONDING_OPTIONS and BONDING_MODULE_OPTIONS for me. But my initial impression that `BONDING_MODULE_OPTIONS` is a newer version of `BONDING_OPTIONS`  seems to be wrong.Instead it seems the difference between sysconfig & initscript is indeed how the OS handles bonding. Not sure how much of a difference there still is today, however the [rhel 7 documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/networking_guide/sec-using_channel_bonding) still mentions BONDING_OPTS instead of BONDING_MODULE_OPTS so it is probably wiser to switch depending on the distribution.

Not sure I'm good enough with python to implement this but maybe it's possible at the distro specific places you mentioned?

TheRealFalcon:
> Not sure I'm good enough with python to implement this but maybe it's possible at the distro specific places you mentioned?

I think the main thing is making sure the [line currently setting BONDING_OPTS](https://github.com/canonical/cloud-init/pull/831/files#diff-23ea39b54c92c8024359c38493fa3cf5b18f94fb8d8c1054d575c766a6a84228L638) is correct.

If you look elsewhere in `sysconfig.py`, you can see code that checks 
```python
if flavor == 'suse':
    # something suse specific
else:
    # something else
```

You could pass flavor into the `_render_bonding_opts` method, do the check there, and set the config appropriately depending on if you're on suse or not.

rjschwei (changes requested):
This will most likely break RHEL. The changes will have to be triggered based off the flavor setting for the renderer

B1Sandmann:
So I tried to follow the advise from you @TheRealFalcon and adjusted the unittest for suse as well. @rjschwei please also take another look. :bow:

cloudinit/net/sysconfig.py
```diff
@@ -634,8 +635,19 @@ def _render_bonding_opts(cls, iface_cfg, iface):
...
+        if flavor == 'suse':
+            if bond_opts:
+                iface_cfg['BONDING_MODULE_OPTS'] = " ".join(bond_opts)
+        # rhel uses initscript support and thus requires BONDING_OPTS
+        # this is also the old default
+        # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
+        #  3.2 Configuration with Initscripts Support
+        else:
+            if bond_opts:
+                iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
```
TheRealFalcon, line 650:
```suggestion
        if bond_opts:
            if flavor == 'suse':
                # suse uses the sysconfig support which requires BONDING_MODULE_OPTS
                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
                # 3.1 Configuration with Sysconfig Support            
                iface_cfg['BONDING_MODULE_OPTS'] = " ".join(bond_opts)
            else:
                # rhel uses initscript support and thus requires BONDING_OPTS
                # this is also the old default
                # see https://www.kernel.org/doc/Documentation/networking/bonding.txt
                #  3.2 Configuration with Initscripts Support
                iface_cfg['BONDING_OPTS'] = " ".join(bond_opts)
```

TheRealFalcon (approved):
Just a small suggestion to simplify the `if` statement slightly, but otherwise looks good @B1Sandmann !

I'll wait to see if @rjschwei has anything to add.

cloudinit/net/sysconfig.py
B1Sandmann: yes that makes more sense this way. I was unsure about the order of the ifs. I'd say we wait for more feedback and do it this way if no one objects.

rjschwei: Yes the single `if bond_opts:` check please.

rjschwei (approved):

TheRealFalcon:
@B1Sandmann , the update looks good. The only thing left is a few flake errors.
https://travis-ci.com/github/canonical/cloud-init/jobs/496953649

PR #832, opened 2021-03-04
Integration test for #783

TheRealFalcon:
## Proposed Commit Message

```
Integration test for #783

Newer verisons of /etc/sudoers prefer @includedir over
#includedir. Ensure we handle that properly and don't include an
additional #includedir when one isn't warranted.
```

## Additional Context
#783

## Test Steps
Run tests/integration_tests/modules/test_users_groups.py on Groovy or later. Pre-SRU, you'll also need an image including the relevant code.

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_users_groups.py
+++ b/tests/integration_tests/modules/test_users_groups.py
@@ -11,6 +11,9 @@ import re
 
 import pytest
 
+from tests.integration_tests.clouds import ImageSpecification
+from tests.integration_tests.instances import IntegrationInstance
+
 
 USER_DATA = """\
 #cloud-config
@@ -86,3 +89,30 @@ class TestUsersGroups:
         _, groups_str = output.split(":", maxsplit=1)
         groups = groups_str.split()
         assert "secret" in groups
+
+
+@pytest.mark.user_data(USER_DATA)
+def test_sudoers_includedir(client: IntegrationInstance):
+    """Ensure we don't add additional #includedir to sudoers.
+
+    Newer verisons of /etc/sudoers will use @includedir rather than
+    #includedir. Ensure we handle that properly and don't include an
+    additional #includedir when one isn't warranted.
+
+    https://github.com/canonical/cloud-init/pull/783
+    """
+    if ImageSpecification.from_os_image().release not in ['hirsute', 'groovy']:
+        raise pytest.skip(
+            'Test requires version of sudo installed on groovy and later'
+        )
+    client.execute("sed -i 's/#include/@include/g' /etc/sudoers")
+
+    sudoers = client.read_from_file('/etc/sudoers')
+    if '@includedir /etc/sudoers.d' not in sudoers:
+        client.execute("echo '@includedir /etc/sudoers.d' > /etc/sudoers")
+    client.execute('cloud-init clean --logs')
+    client.restart()
+    sudoers = client.read_from_file('/etc/sudoers')
+
+    assert '#includedir' not in sudoers
+    assert sudoers.count('includedir /etc/sudoers.d') == 1
--- a/tests/integration_tests/modules/test_users_groups.py
+++ b/tests/integration_tests/modules/test_users_groups.py
@@ -1,7 +1,7 @@
 """Integration tests for the user_groups module.
 
 TODO:
-* This modules assumes that the "ubuntu" user will be created when "default" is
+* This module assumes that the "ubuntu" user will be created when "default" is
   specified; this will need modification to run on other OSes.
 """
 import re
--- a/tests/integration_tests/modules/test_users_groups.py
+++ b/tests/integration_tests/modules/test_users_groups.py
@@ -1,10 +1,7 @@
-"""Integration test for the user_groups module.
-
-This test specifies a number of users and groups via user-data, and confirms
-that they have been configured correctly in the system under test.
+"""Integration tests for the user_groups module.
 
 TODO:
-* This test assumes that the "ubuntu" user will be created when "default" is
+* This modules assumes that the "ubuntu" user will be created when "default" is
   specified; this will need modification to run on other OSes.
 """
 import re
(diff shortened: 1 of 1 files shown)


tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
         _, groups_str = output.split(":", maxsplit=1)
         groups = groups_str.split()
         assert "secret" in groups
+
+
+@pytest.mark.user_data(USER_DATA)
+def test_sudoers_includedir(client: IntegrationInstance):
+    """Ensure we don't add additional #includedir to sudoers.
+
+    Newer verisons of /etc/sudoers will use @includedir rather than
```
OddBloke, line 98:
```suggestion
    Newer versions of /etc/sudoers will use @includedir rather than
```

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
...
+    #includedir. Ensure we handle that properly and don't include an
+    additional #includedir when one isn't warranted.
+
+    https://github.com/canonical/cloud-init/pull/783
+    """
+    if ImageSpecification.from_os_image().release not in ['hirsute', 'groovy']:
```
OddBloke, line 104: This looks like it'll require updating in ~2mos; should we do `in ['xenial', 'bionic', 'focal']` instead? (Or perhaps `@pytest.mark.not_xenial`, `.not_bionic`, and `.not_focal`?)

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
...
+        raise pytest.skip(
+            'Test requires version of sudo installed on groovy and later'
+        )
+    client.execute("sed -i 's/#include/@include/g' /etc/sudoers")
+
+    sudoers = client.read_from_file('/etc/sudoers')
+    if '@includedir /etc/sudoers.d' not in sudoers:
+        client.execute("echo '@includedir /etc/sudoers.d' > /etc/sudoers")
```
OddBloke, line 112: Probably a nit, but:

```suggestion
        client.execute("echo '@includedir /etc/sudoers.d' >> /etc/sudoers")
```

will avoid replacing all existing content of `/etc/sudoers`: given we use `sudo` to execute commands, that seems like it could backfire.

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
...
+    client.execute('cloud-init clean --logs')
```
OddBloke, line 113: This is inline with other tests, so fine to keep as-is here, but should we generally be doing:

```suggestion
    client.instance.clean()
```

in such cases?

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
...
+    client.restart()
```
OddBloke, line 114: Do we need to wait for reboot here?

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +89,30 @@ def test_user_root_in_secret(self, class_client):
         _, groups_str = output.split(":", maxsplit=1)
         groups = groups_str.split()
         assert "secret" in groups
+
+
+@pytest.mark.user_data(USER_DATA)
+def test_sudoers_includedir(client: IntegrationInstance):
+    """Ensure we don't add additional #includedir to sudoers.
```
OddBloke, line 96: At the moment, the module docstring describes the test above this one; should that be moved to a class-level docstring?

OddBloke (commented):
Thanks James! The core of this test looks sound, I have some potential cleanups and nits inline.

tests/integration_tests/modules/test_users_groups.py
TheRealFalcon: restart defaults to `wait=True` (though we should probably support passing `wait` through...)

tests/integration_tests/modules/test_users_groups.py
TheRealFalcon: Yep, good idea. I prefer the former these days because otherwise we'll have 8-10 marks specifying every current release and not-release (plus future releases), and that's just for ubuntu. In the back of my mind, I have an idea for having a single mark to specify OS/feature dependencies...but...that doesn't help us now :smile:

OddBloke: Yeah, I was just thinking that `@pytest.mark.ubuntu(not=["xenial", "bionic", "focal"])` or similar would work nicely for this.

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -86,3 +92,32 @@ def test_user_root_in_secret(self, class_client):
...
+    additional #includedir when one isn't warranted.
+
+    https://github.com/canonical/cloud-init/pull/783
+    """
+    if ImageSpecification.from_os_image().release in [
+        'xenial', 'bionic', 'focal'
+    ]:
+        raise pytest.skip(
+            'Test requires version of sudo installed on groovy and later'
+        )
```
OddBloke, line 112: I'm running for the weekend, but I just realised doing this instead of marks means that we'll launch the instance before deciding that we don't want to run this test (because we're using `client` and not `session_cloud` as we do elsewhere with this pattern).

tests/integration_tests/modules/test_users_groups.py
OddBloke: I was actually thinking of the case where we get into the instance before the reboot has happened; I'm blanking on whether or not we've made that work across all clouds.

tests/integration_tests/modules/test_users_groups.py
```diff
@@ -1,16 +1,16 @@
-"""Integration test for the user_groups module.
-
-This test specifies a number of users and groups via user-data, and confirms
-that they have been configured correctly in the system under test.
+"""Integration tests for the user_groups module.
 
 TODO:
-* This test assumes that the "ubuntu" user will be created when "default" is
+* This modules assumes that the "ubuntu" user will be created when "default" is
```
OddBloke, line 4:
```suggestion
* This module assumes that the "ubuntu" user will be created when "default" is
```

tests/integration_tests/modules/test_users_groups.py
OddBloke: We'll address this in #839

OddBloke (approved):
Thanks, one minor typo fix that I'll apply now through the GH UI.

PR #833, opened 2021-03-04
Tools for testing locally in Docker for pushing PRs

beantaxi:
## Proposed Commit Message
Tools for testing locally in Docker for pushing PRs

```
Tools for testing locally in Docker for pushing PRs

When I was submitting my first PR (PR #745), I spent a lot of time waiting for
GitHub+Travis to report flake8 and pylint errors, plus pytest failures, before
fixing, pushing, repeating. It was very slow, and I was concerned that 
all the churn would deter approvers from paying attention to my PR,
what with all the constant tweaking.

I tried to run the tests locally, but it looked like I was getting failures
related to running locally on OSX, not Linux. So I figured if I need to
run on Linux, why not use Docker.
```
## Test Steps
I don't think any test steps are applicable here, but here are some usage steps

```
# Build the cloudinit-tester docker image
$ ./cloudinit-tester-build.sh

# Run a single command using a cloudinit-tester container
$ ./cloudinit-tester-run.sh pylint ./cloudinit/my-new-file.py

# Run flake8, pylint, and pytest in a single command
$ ./cloudinit-tester-cmds.sh

# Launch a cloudinit-tester container and shell into it
$ ./cloudinit-tester-shell.sh
```

## Checklist:


OddBloke:
Thanks for this thought! Enabling people to use Docker for local development would be great: this would allow developers without ready access to a Linux system to more easily perform the checks that their PR will need to pass locally.

Have you come across https://github.com/tox-dev/tox-docker before? I've never used it (so it may not be fit for purpose, for all I know), but that looks like it might give us an easier way of integrating the use of Docker into our existing definitions, without carrying custom code to do it. (This _would_ require folks to install `tox` on their local machine: that may mean that we lose the ease-of-access that supporting Docker would give us, which may mean that it isn't the right path to pursue.)

OddBloke:
Another thought, setting tox-docker aside: to avoid having to keep these script in-step with any tox changes, we should probably drive tox within the Docker container (which also has the advantage of simplifying the container creation: we'd only need to install tox).

beantaxi:
@OddBloke I might know less about `tox` than anyone alive, so I can't immediately answer your question intelligently, but I'll definitely have a look. I will say that requiring the installation of `tox` -- or anything other than Docker -- was a non-starter for me, which was sort of why I figured out how to test in Docker in the first place. But `tox-docker` (I just looked quickly) does sound potentially useful for folks who already have `tox` and Docker, which might well describe every cloud-init hacker other than me, so if I see anything helpful for folks who already have both, I'll try and leverage that.

beantaxi:
The one piece of feedback I'm expecting, so I'll go ahead and pre-emptively provide it, is that it might be good for the #833 functionality to be invoked via `cloudinit devel xxx` commands, instead of some loose bash scripts. That seems like it might be a nice clean endstate, but I didn't want to do the work to figure out how to do the work, without first checking if the `cloudinit devel` CLI was open for extension. Thanks!

OddBloke:
> The one piece of feedback I'm expecting, so I'll go ahead and pre-emptively provide it, is that it might be good for the #833 functionality to be invoked via `cloudinit devel xxx` commands, instead of some loose bash scripts. That seems like it might be a nice clean endstate, but I didn't want to do the work to figure out how to do the work, without first checking if the `cloudinit devel` CLI was open for extension. Thanks!

I don't think `cloud-init devel` subcommands would meet your requirements: using them would then require cloud-init (and its requirements) to be installed locally and:

> I will say that requiring the installation of tox -- or anything other than Docker -- was a non-starter for me

(Plus: if you can install cloud-init and its dependencies, you presumably could just install tox as well.)

Would `make` be more likely to be available to folks? We could consider adding targets for these bits there?

beantaxi:
Really good point about requiring a `cloud-init` installation; I hadn't thought about that.

The idea of `cloudinit devel` as my homebase for cloud-init development is appealing, but I guess it's trickier than I'd considered.

That being the case I'm happy to just stick with the existing shell scripts, unless now that I've said something they're irreversibly repugnant.

PR #834, opened 2021-03-05
Allow user control over update events

TheRealFalcon:
## Proposed Commit Message

```
Allow user control over update events

Control is currently limited to boot events, though this should
allow us to more easily incorporate HOTPLUG support. Disabling
'instance-first-boot' is not supported as we apply networking config
too early in boot to have processed userdata (along with the fact
that this would be a pretty big foot-gun).

The concept of update events on datasource has been split into
supported update events and default update events. Defaults will be
used if there is no user-defined update events, but user-defined
events won't be supplied if they aren't supported.
When applying the networking config, we now check to see if the event
is supported by the datasource as well as if it is enabled.

Configuration looks like:
updates:
  network:
    when: ['boot']
```

## Additional Context
Most of the functionality here has been taken from the [initial hotplug PR](https://github.com/canonical/cloud-init/pull/47), though this PR is intentionally more limited in scope, and changes have been made as I saw fit.

## Test Steps
WIP


cloudinit/stages.py
```diff
@@ -692,34 +702,81 @@ def _find_networking_config(self):
...
+        current_boot_type = EventType.BOOT
+        if self.is_new_instance():
+            current_boot_type = EventType.BOOT_NEW_INSTANCE
+
+        if not any([
+            self.datasource is NULL_DATA_SOURCE,
+            self.is_new_instance(),  # or should it be self._is_first_boot(),
+            (self.update_event_enabled(current_boot_type, scope='network')
+             and self.datasource.update_metadata([current_boot_type]))
+        ]):
```
TheRealFalcon, line 769: This section has been where I've been banging my head most against the wall. (With a few exceptions), `apply_networking_config` gets called twice during boot. It get called first in LOCAL mode, and then again in NETWORK mode. 

Previous to this PR, on first boot,  if the datasource got initialized in LOCAL mode, we'd return at this point and not apply networking config in NETWORK mode (`is_new_instance()` returns `False` in network mode on first boot if there's a FILESYSTEM dep). On every subsequent boot, If the BOOT update event is enabled we apply networking config twice (one for LOCAL, one for NETWORK).

That same general behavior should still be true, but applying networking config twice on every boot seems wrong to me. Since `self.is_new_instance()` will always return False on every subsequent boot, that can't be our determining factor. I had considered checking datasource dependencies, and apply during LOCAL if there's a filesystem dependency, and at NETWORK if there's a network dependency, but that could cause problems on first boot because even if we find no local datasource, we still generate the fallback config and use that

TheRealFalcon:
@raharper , do you have any insight to my comment here?

cloudinit/stages.py
raharper: @TheRealFalcon  just caught up on mail so sorry for the delay.

I think we've had a bug for the longest time on apply_networking_config twice
as you've noticed.  Generally this isn't an issue (other than wasteful) in
that for Datasources that are local and have network-config (say OpenStack)
we already get the network config we're going to use.  In the case that the DS
is network-only (GCE) local never wrote a network config.

That said, I have wanted to apply it only once and the rub has been we've not
decided on how we can keep state between local and init that we've already
rendered network config.  If we want to do so we could use either explicit
state on the cloud object such that when we re-use the object at net time (we
load the object with existing=trust (stages.py:_restore_checked_cache() we'll
have something that can be used to determine if we've already applied.

If we want to avoid attributes on the instance object, we can use ephemeral
files in /run/cloud-init;  this won't have any state implications from boot to
boot but we should carefully think about the case of something re-running
cloud-init from the command line.

cloudinit/stages.py
```diff
@@ -692,34 +702,81 @@ def _find_networking_config(self):
...
+            LOG.debug(
+                "No network config applied. Neither a new instance"
+                " nor datasource network update on '%s' event",
+                EventType.BOOT)
```
OddBloke, line 773: Is this log message still correct?

cloudinit/stages.py
```diff
@@ -692,34 +702,81 @@ def _find_networking_config(self):
...
-                    self._apply_netcfg_names(netcfg)
-                    return
-                else:
-                    # refresh netcfg after update
-                    netcfg, src = self._find_networking_config()
+        current_boot_type = EventType.BOOT
+        if self.is_new_instance():
+            current_boot_type = EventType.BOOT_NEW_INSTANCE
+
+        if not any([
```
OddBloke, line 764: Using `any([...])` here means that we'll evaluate all of these conditions, ironically, unconditionally:

```py
In [2]: def foo(): 
   ...:     print("foo") 
   ...:     return True 
   ...:                                                                                                                                                                                                                                                                                   

In [3]: any([foo(), foo()])                                                                                                                                                                                                                                                               
foo
foo
Out[3]: True
```

rather than lazily:

```py
In [6]: foo() or foo()                                                                                                                                                                                                                                                                    
foo
Out[6]: True
```

This is particularly problematic here because it means there's a decent chance we will call `self.datasource.update_metadata` in cases where we previously wouldn't: that could cause us to re-fetch metadata in cases where it isn't necessary.

cloudinit/stages.py
```diff
@@ -692,34 +702,81 @@ def _find_networking_config(self):
...
-                        EventType.BOOT)
-                    # nothing new, but ensure proper names
-                    self._apply_netcfg_names(netcfg)
-                    return
-                else:
-                    # refresh netcfg after update
-                    netcfg, src = self._find_networking_config()
+        current_boot_type = EventType.BOOT
+        if self.is_new_instance():
+            current_boot_type = EventType.BOOT_NEW_INSTANCE
```
OddBloke, line 762: This feels like it should be a first-class attribute (presumably on `self`).

cloudinit/stages.py
```diff
@@ -692,34 +702,81 @@ def _find_networking_config(self):
...
+            LOG.debug(
+                "No network config applied. Neither a new instance"
+                " nor datasource network update on '%s' event",
+                EventType.BOOT)
+            # nothing new, but ensure proper names
+            self._apply_netcfg_names(netcfg)
+            return
+        else:
+            # refresh netcfg after update
+            netcfg, src = self._find_networking_config()
```
OddBloke, line 779: A nit, for sure, but as we unconditionally return in the `if` branch, we don't need the `else`:

```suggestion
        # refresh netcfg after update
        netcfg, src = self._find_networking_config()
```

doc/rtd/topics/events.rst
```diff
@@ -0,0 +1,81 @@
+.. _events:
+
+******************
+Events and Updates
```
OddBloke, line 4: This doc could use some work to make it a little more accessible for cloud-init users.

OddBloke (commented):
First pass, familiarising myself with the overall approach here. Thanks for the work, both Ryan and James!

cloudinit/stages.py
TheRealFalcon: Yes...would you word it differently?

cloudinit/stages.py
TheRealFalcon: I don't see any other method where this would currently be used. Can we take a YAGNI approach here and make it first class later if it's needed?

OddBloke: My fear is that people will just reimplement it (but probably slightly differently) instead of performing a refactor. I think it would also aid readability here, as it would be clear(er) that the `current_boot_type` is an invariant property of the Init stage we're running (in the same way that `is_new_instance` is an invariant property).

(But I certainly won't block landing on it.)

cloudinit/stages.py
OddBloke: Wow, sure wish I'd written more words here, sorry about that. Let's see if I can remember what I meant.

Should this be:

```suggestion
            LOG.debug(
                "No network config applied. Neither a new instance"
                " nor datasource network update on '%s' event",
                current_boot_type)
```

cloudinit/event.py
```diff
@@ -1,17 +1,72 @@
...
+    ])
+    LOG.debug('updates merged: %s', updates)
+
+    events = {}
+    for etype in [scope for scope, val in default_events.items()
+                  if type(val) == dict and 'when' in val]:
+        events[etype] = (
+            set([EventTypeMap.get(evt)
+                 for evt in updates.get(etype, {}).get('when', [])
+                 if evt in EventTypeMap]))
```
OddBloke, line 59: Our codebase has a bunch of places where we pass around dicts with a known set of attributes which have a bunch of semantics attached to them, and it's always difficult to follow what's going on because the semantics are defined wherever that dict happens to be used in the codebase. The complexity of these list comprehensions makes me wonder if we should be replacing/wrapping these dicts with a class which could encapsulate some of this complexity, to avoid going down the same path that we have before?

Something like (this very quick sketch):

```py
updates = UpdateEventsConfig.from_merge(config_events, default_events)
...
for scope in default_events.scopes:
    events[scope] = updates.events_for(scope)
```

is much easier to follow, for example.

(Potentially a good first step would be to split out each of these list comprehensions to functions: giving them a name and explanatory docstring/comment might be sufficiently clarifying to obviate the need for a class.)

cloudinit/sources/__init__.py
```diff
@@ -174,12 +174,20 @@ class DataSource(metaclass=abc.ABCMeta):
...
-    # network configuration on metadata changes.
-    # A datasource which supports writing network config on each system boot
-    # would call update_events['network'].add(EventType.BOOT).
+    # network configuration on metadata changes. These are defined in
+    # `supported_network_events`.
+    # The datasource also defines a set of default EventTypes that the
+    # datasource can react to. These are the event types that will be used
+    # if not overridden by the user.
+    # A datasource requiring to write network config on each system boot
+    # would call default_update_events['network'].add(EventType.BOOT).
```
OddBloke, line 183: I know you didn't introduce this behaviour, but it's always made me a little concerned: we're modifying the set of events that _all_ datasource classes will use. This _shouldn't_ be a problem, because we only ever use one DS, but as your head is in this space ATM, let me know if you can think of any problems with it.

cloudinit/sources/__init__.py
```diff
@@ -174,12 +174,20 @@ class DataSource(metaclass=abc.ABCMeta):
...
 
     # Default: generate network config on new instance id (first boot).
-    update_events = {'network': set([EventType.BOOT_NEW_INSTANCE])}
+    supported_update_events = {'network': set([
```
OddBloke, line 186: `frozenset`, maybe? (A nit, regardless.)

cloudinit/sources/__init__.py
```diff
@@ -174,12 +174,20 @@ class DataSource(metaclass=abc.ABCMeta):
...
+        EventType.BOOT_NEW_INSTANCE,
+        EventType.BOOT,
+    ])}
+    default_update_events = {'network': set([EventType.BOOT_NEW_INSTANCE])}
```
OddBloke, line 190: Should we use a constant for `'network'` so that we don't have a "magic" string affecting behaviour? (We'll notice `EventScope.NETWROK` a lot sooner than `'netwrok'` via linting, etc.)

cloudinit/stages.py
```diff
@@ -692,34 +705,77 @@ def _find_networking_config(self):
...
+
+        for evt_scope in scopes:
+            if event_source_type in allowed.get(evt_scope, []):
+                LOG.debug('Event Allowed: scope=%s EventType=%s',
+                          evt_scope, event_source_type)
+                return True
+
+        LOG.debug('Event Denied: scopes=%s EventType=%s',
+                  scopes, event_source_type)
+        return False
```
OddBloke, line 732: None of this references `self`: perhaps a candidate to move into `cloudinit.event` as a function?

cloudinit/tests/test_stages.py
```diff
@@ -338,12 +341,84 @@ def fake_network_config():
...
+    ):
+        """Apply network if datasource permits BOOT event."""
+        net_cfg = self._apply_network_setup(m_macs)
+
+        self.init.datasource.default_update_events = {
+            'network': [EventType.BOOT_NEW_INSTANCE, EventType.BOOT]
+        }
         self.init.apply_network_config(True)
-        self.init.distro.apply_network_config_names.assert_called_with(net_cfg)
+        self.init.distro.apply_network_config_names.assert_called_with(
```
OddBloke, line 359: Is this mock autospec'd?

cloudinit/tests/test_stages.py
```diff
@@ -338,12 +341,84 @@ def fake_network_config():
...
+        self, m_ubuntu, m_macs
+    ):
+        """Don't apply network config if unsupported.
+
+        Shouldn't work even when specified as userdata
+        """
+        self._apply_network_setup(m_macs)
+
+        self.init.datasource.supported_update_events = {
+            'network': [EventType.BOOT_NEW_INSTANCE]
```
OddBloke, line 412: Should this be a set instead of a list?

OddBloke (commented):
Thanks James, this is looking good! I have some suggestions to improve clarity in `cloudinit.event` as well as some thoughts on testing, but the core of this functionality LGTM: now I'm mostly concerned with making sure we're setting ourselves up well for the future, in terms of API choices.

cloudinit/sources/__init__.py
TheRealFalcon: Yeah, I'm not very comfortable with this either. I haven't thought of any easy solutions though.

TheRealFalcon:
@OddBloke Pushed a big-ish update. I (hopefully) simplified event.py a lot, and a lot of the other changes flow out from there.

cloudinit/stages.py
```diff
@@ -710,26 +751,35 @@ def _apply_netcfg_names(self, netcfg):
...
-                    netcfg, src = self._find_networking_config()
+        if not (
+            self.datasource is NULL_DATA_SOURCE or
+            self.boot_type == EventType.BOOT_NEW_INSTANCE or
+            (
+                self.update_event_enabled(
+                    self.boot_type, scope=EventScope.NETWORK
+                ) and self.datasource.update_metadata([self.boot_type])
+            )
+        ):
```
OddBloke, line 772: I'm having a hard time following the exact logic here; could we decompose it into one or more variables, that'll make it a little easier to understand what we're checking?

tests/integration_tests/modules/test_user_events.py
```diff
@@ -0,0 +1,104 @@
...
+    assert 'dummy0' not in client.execute('ls /sys/class/net')
+
+    _add_dummy_bridge_to_netplan(client)
+    client.execute('rm /var/log/cloud-init.log')
+    client.restart()
+    log = client.read_from_file('/var/log/cloud-init.log')
+
+    assert 'Event Allowed: scope=network EventType=boot' in log
+    assert 'Applying network configuration' in log
+    assert 'dummy0' not in client.execute('ls /sys/class/net')
```
OddBloke, line 104: Worth refactoring this into a `_test_network_config_applied_on_reboot` which both this and the Azure test can call?

tests/integration_tests/modules/test_user_events.py
```diff
@@ -0,0 +1,104 @@
+"""Test user-overridable events.
+
+This is currently limited to applying network config on BOOT events.
+Since we reapply the same config on reboot, verification is limited to
+checking log messages.
+
+Because the supported/default events are datasource specific, the datasource
+is being limited to LXD containers/vms (NoCloud).
+For NoCloud, BOOT_NEW_INSTANCE and BOOT are supported, but only
+BOOT_NEW_INSTANCE is default.
```
OddBloke, line 10: These tests look much better now, thank you! (But this docstring is now out-of-date. :runner::dash:)

OddBloke (commented):
Thanks James, this refactor has made it much easier to follow along! I think structurally we're pretty much there: I have some more specific comments/questions inline.

cloudinit/sources/DataSourceRbxCloud.py
```diff
@@ -206,10 +206,10 @@ def read_user_data_callback(mount_dir):
 
 class DataSourceRbxCloud(sources.DataSource):
     dsname = "RbxCloud"
-    update_events = {'network': [
+    default_update_events = {EventScope.NETWORK: set([
```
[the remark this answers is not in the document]
TheRealFalcon: It's a personal preference because I often think I'm seeing a dictionary when I see a set literal. I don't have a problem changing it though.

cloudinit/stages.py
```diff
@@ -702,6 +715,34 @@ def _find_networking_config(self):
...
+        user_events = userdata_to_events(self.cfg.get('updates', {}))
+        # A value in the first will override a value in the second
+        allowed = util.mergemanydict([
+            copy.deepcopy(user_events),
+            copy.deepcopy(default_events),
+        ])
+        LOG.debug('Allowed events: %s', allowed)
+
+        if not scope:
+            scopes = list(allowed.keys())
```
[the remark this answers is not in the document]
TheRealFalcon: Because of the "almost list" structures that dicts give us back.
```
>>> x = {1:2}
>>> [x.keys()]
[dict_keys([1])]
>>> list(x.keys())
[1]
```

Since I'm using a comprehension it later, I need it to be a "real" list.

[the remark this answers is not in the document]
OddBloke: `dict_keys` is iterable:

```py
In [22]: from collections.abc import Iterable
In [21]: isinstance({}.keys(), Iterable)
Out[21]: True
```

So is usable in comprehensions:

```py
In [3]: from enum import Enum

In [4]: class Test(Enum):
   ...:     thing = "string"
In [8]: d = {Test("string"): "foo"}

In [9]: [x for x in d.keys()]
Out[9]: [<Test.thing: 'string'>]

In [10]: [x.value for x in d.keys()]
Out[10]: ['string']
```

It isn't a sequence:

```py
In [22]: from collections.abc import Sequence
In [23]: isinstance({}.keys(), Sequence)
Out[23]: False
```

because it doesn't implement `__getitem__` (so you can't subscript it, which is perhaps what you were thinking of).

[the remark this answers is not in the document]
TheRealFalcon: Actually...no :smile: 

The code I was copying from had a `scopes = [allowed.keys()]` and I misinterpreted why that wasn't working. I thought it was leaving it as a `dict_keys`, which is why I converted it to a proper list, when in actuality it was wrapping the dict_keys into another list.

All this to say...I'll drop the `list(...)`

cloudinit/stages.py
```diff
@@ -702,6 +716,47 @@ def _find_networking_config(self):
...
+        scope_values = [s.value for s in scopes]
+        LOG.debug('Possible scopes for this event: %s', scope_values)
+
+        for evt_scope in scopes:
+            if event_source_type in allowed.get(evt_scope, []):
+                LOG.debug('Event Allowed: scope=%s EventType=%s',
+                          evt_scope.value, event_source_type)
+                return True
+
+        LOG.debug('Event Denied: scopes=%s EventType=%s',
```
TheRealFalcon, line 756: I'm still not happy with this. We attempt to apply networking config twice on every boot. Pre-this-PR behavior, if only BOOT_NEW_INSTANCE was enabled, we would apply exactly once on the first part of first boot, then deny on every subsequent attempt. If BOOT was enabled, we'd apply every single time (twice per boot, including first boot).

With this PR, on first boot we don't have access to userdata, so we fall back to the default update events of the datasource. If only BOOT_NEW_INSTANCE is enabled, we'll get an event denied message in the logs, even if BOOT is specified in the userdata. I'm concerned having an event denied message in the logs will be confusing given that it is enabled in the userdata.

I could add a check for first boot here and slightly alter the log message. It could be something like "Event Denied from data source...", but that also seems like overkill checking every boot for a not-much-more-helpful log message.

Any thoughts for improvement?

TheRealFalcon:
@OddBloke should be ready for another review

cloudinit/stages.py
OddBloke: Are there cases when we would only successfully apply networking configuration on the second attempt during first boot? If so, then does it follow that `BOOT_NEW_INSTANCE` should be the toggle for both of those runs? And, therefore, should we actually be using the check for first boot for more than just a log message?

cloudinit/stages.py
```diff
@@ -710,26 +765,36 @@ def _apply_netcfg_names(self, netcfg):
...
-                        "No network config applied. Neither a new instance"
-                        " nor datasource network update on '%s' event",
-                        EventType.BOOT)
-                    # nothing new, but ensure proper names
-                    self._apply_netcfg_names(netcfg)
-                    return
-                else:
-                    # refresh netcfg after update
-                    netcfg, src = self._find_networking_config()
+        def _boot_event_enabled_and_supported():
```
OddBloke, line 778: Thanks for the update here James, it's much easier to follow what the intent is here!

cloudinit/stages.py
```diff
@@ -710,26 +765,36 @@ def _apply_netcfg_names(self, netcfg):
...
+            return self.update_event_enabled(
+                EventType.BOOT, scope=EventScope.NETWORK
+            ) and self.datasource.update_metadata([EventType.BOOT])
```
OddBloke, line 781: This call still throws me off: this isn't just checking for support, but will actually cause the datasource to discard cached data and fetch it anew. That feels to me like it should be two separate calls: "does this datasource support $event_type" and "update the datasource's metadata for $event_type".

It also feels a little odd that `update_metadata` doesn't take a scope: we only want to update network metadata here, if (e.g.) storage metadata has changed then we don't particularly want to update it.

Relatedly, it may look like `update_metadata` should just update network metadata (or, generally, metadata for the explicitly configured scopes), but our `_get_data` implementations generally fetch and store All The Metadata, so I think this will completely refresh all metadata regardless: perhaps not a problem, but worth noting.

cloudinit/stages.py
OddBloke: It also occurs to me that we already have language for these occasions, from our script execution modules: once, instance, and boot. I'm not suggesting we change the language we're using here ("instance" is obviously ~contained within "BOOT_NEW_INSTANCE"), but it does make me wonder if we should be following those modules and tracking whether we have applied network configuration _for this instance_, rather than trying to infer what we should be doing based on the type of a boot we think we have.

If we haven't yet configured networking, `BOOT_NEW_INSTANCE` or `BOOT` will cause it to be applied; if we have, then only `BOOT` will.

What do you think?

TheRealFalcon: > Are there cases when we would only successfully apply networking configuration on the second attempt during first boot?

Yes, I've been assuming that in my explanations. We will be in BOOT_NEW_INSTANCE state until the datasource has been initialized. It's entirely possible that the first attempt can result in "wait till networking is up", so the second attempt is still BOOT_NEW_INSTANCE. We always attempt to apply networking when BOOT_NEW_INSTANCE

> If so, then does it follow that BOOT_NEW_INSTANCE should be the toggle for both of those runs? And, therefore, should we actually be using the check for first boot for more than just a log message?

I'm not sure I follow what `should be the toggle` means. But to the second question, I think the answer is no. First boot is really only relevant because we don't have userdata available yet.

> but it does make me wonder if we should be following those modules and tracking whether we have applied network configuration for this instance, rather than trying to infer what we should be doing based on the type of a boot we think we have.
If we haven't yet configured networking, BOOT_NEW_INSTANCE or BOOT will cause it to be applied; if we have, then only BOOT will.

This is essentially what I'm already doing. Or am I misunderstanding what you're saying?

Sorry if I didn't make it clear, the but the case I'm concerned about is *when* networking is applied on first attempt and userdata specifies BOOT (which we can't access on first boot), what's happening on the second attempt.

CURRENT BEHAVIOR:
```
master when datasource specifies BOOT_NEW_INSTANCE:
first boot, one: apply
first boot, two: skip
subsequent boot, one: skip
subsequent boot, two: skip

master when datasource specifies BOOT:
first boot, one: apply
first boot, two: apply
subsequent boot, one: apply
subsequent boot, two: apply
```

The above should be exactly the same on this branch when userdata specifies BOOT_NEW_INSTANCE. However, when userdata specifies BOOT:

BRANCH BEHAVIOR:
```
branch with userdata BOOT:
first boot, one: apply
first boot, two: apply or skip based on datasource default because we don't have userdata
subsequent boot, one: apply
subsequent boot, two: apply
```
(These scenarios are assuming networking gets applied on the first attempt)

If on `first boot, two` we skip because of the datasource default, we get a big ugly "Event Denied" log message that will likely be confusing for somebody that doesn't know userdata won't take effect till next boot.

Changing this to instead apply on both attempts isn't difficult, but it would also apply in the BOOT_NEW_INSTANCE case, which is a change in current behavior. This could be important for backwards compatibility, in case we specifically DON'T want network config applied more than once when `BOOT_NEW_INSTANCE` is specified.

Does my reasoning here make sense? If so, do you think it makes sense to modify the log message on first boot for the case I mentioned?

TheRealFalcon:
@OddBloke most recent commit uses the semaphore as we discussed. Still needs a ton of test changes though.

doc/rtd/topics/events.rst
```diff
@@ -0,0 +1,79 @@
...
+
+`Cloud-init`_ will fetch and apply cloud and user data configuration
+upon several event types. The two most common events for cloud-init
+are when an instance first boots and any subsequent boot thereafter (reboot).
+In addition to boot events, cloud-init users and vendors are interested
+in when devices are added. cloud-init currently supports the following
+event types:
+
+- **BOOT_NEW_INSTANCE**: ``New instance first boot``
+- **BOOT**: ``Any system boot other than 'BOOT_NEW_INSTANCE'``
```
OddBloke, line 18: We need to update this with the legacy event.

OddBloke (approved):
This LGTM now, thanks for all the work, James! I've confirmed that a full LXD container run and a full Azure run both succeed locally.

PR #827, opened 2021-03-03
Add Vultr support

ddymko:
## Proposed Commit Message

```
    Add support for Vultr cloud.

    This PR adds in support so that cloud-init can run on instances 
    deployed on Vultr cloud. This was originally brought up in #628.

    Co-authored-by: Eric Benner <ebenner@vultr.com>
```

## Additional Context
Greets from Vultr 👋 

This PR adds in support so that cloud-init can run on instances deploy on Vultr instances. This was originally brought up in https://github.com/canonical/cloud-init/pull/628

Please let me know if there are any updates or additions that may need to be made.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #777, opened 2021-01-19
openstack: read the dynamic metadata group vendor_data2.json

andrewbogott:
Add support for openstack's dynamic vendor data, which appears under openstack/latest/vendor_data2.json

This adds vendor_data2 to all pathways; it should be a no-op for non-OpenStack providers.

Bug: https://bugs.launchpad.net/cloud-init/+bug/1841104

PR #837, opened 2021-03-10
Fix vendordata arrays and dicts

eb3095:
## Fix vendordata arrays and dicts
Fix a bug with vendordata_raw when arrays or arrays of dicts are used

```
summary: An error is thrown when arrays or arrays of dicts are used

When you use an array or an array of dicts with vendordata_raw it is passed to write_files as a string from _store_rawdata causing an error to be thrown. Bellow is the error thrown and a sample of the code used from another PR to trigger the bug.
```

```
2021-03-10 05:00:19,275 - util.py[WARNING]: failed stage init
2021-03-10 05:00:19,275 - util.py[DEBUG]: failed stage init
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/cmd/main.py", line 652, in status_wrapper
    ret = functor(name, args)
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/cmd/main.py", line 376, in main_init
    init.update()
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/stages.py", line 368, in update
    'vendordata')
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/stages.py", line 398, in _store_rawdata
    util.write_file(self._get_ipath('%s_raw' % datasource), data, 0o600)
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/util.py", line 1837, in write_file
    content = encode_text(content)
  File "/usr/local/lib/python3.6/site-packages/cloud_init-21.1-py3.6.egg/cloudinit/util.py", line 117, in encode_text
    return text.encode(encoding)
AttributeError: 'list' object has no attribute 'encode'
```
## Additional Context
@smoser had given me two examples of usage for vendordata_raw for use in the bellow listed PR. However when attempting to use it I encountered an unexpected error regarding writing it to a file. The methods he provided worked otherwise leading me to believe this may be a bug.

https://github.com/canonical/cloud-init/pull/827#discussion_r590696349

## Test Steps
This is the code I used to trigger this failure. It originates from the following PR.

https://github.com/canonical/cloud-init/pull/827

```
    self.vendordata_raw = []
    for uscript in user_scripts:
        self.vendordata_raw.append(uscript)
    self.vendordata_raw.append("#cloud-config\n%s" % json.dumps(config))
```

## Checklist:

### DIFF
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -400,7 +400,7 @@ class Init(object):
     def _store_raw_vendordata(self, data, datasource):
         # Only these data types
         if data is not None and type(data) not in [bytes, str, list]:
-            raise TypeError("vendordata_raw is unsupported type '%s'",
+            raise TypeError("vendordata_raw is unsupported type '%s'" %
                             str(type(data)))
         # This data may be a list, convert it to a string if so
         if isinstance(data, list):
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -401,7 +401,7 @@ class Init(object):
         # Only these data types
         if type(data) not in [bytes, str, list, None]:
             raise TypeError("vendordata_raw is unsupported type '%s'",
-                            type(data))
+                            str(type(data)))
         # This data may be a list, convert it to a string if so
         if isinstance(data, list):
             data = util.json_dumps(data)
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -398,8 +398,11 @@ class Init(object):
         util.write_file(self._get_ipath('%s_raw' % datasource), data, 0o600)
 
     def _store_raw_vendordata(self, data, datasource):
-        # This data may be a list or dict, convert it to a string if so
-        if type(data) in [list, dict]:
+        # Only these data types
+        if type(data) not in [bytes, string, list, None]:
+            raise TypeError("vendordata_raw is unsupported type '%s'", type(data))
+        # This data may be a list, convert it to a string if so
+        if isinstance(data, list):
             data = util.json_dumps(data)
         self._store_rawdata(data, datasource)
 
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -364,12 +364,12 @@ class Init(object):
                             'userdata')
         self._store_processeddata(self.datasource.get_userdata(),
                                   'userdata')
-        self._store_rawdata(self.datasource.get_vendordata_raw(),
-                            'vendordata')
+        self._store_raw_vendordata(self.datasource.get_vendordata_raw(),
+                                   'vendordata')
         self._store_processeddata(self.datasource.get_vendordata(),
                                   'vendordata')
-        self._store_rawdata(self.datasource.get_vendordata2_raw(),
-                            'vendordata2')
+        self._store_raw_vendordata(self.datasource.get_vendordata2_raw(),
+                                   'vendordata2')
         self._store_processeddata(self.datasource.get_vendordata2(),
                                   'vendordata2')
 
@@ -396,6 +396,12 @@ class Init(object):
         if data is None:
             data = b''
         util.write_file(self._get_ipath('%s_raw' % datasource), data, 0o600)
+
+    def _store_raw_vendordata(self, data, datasource):
+        # This data may be a list or dict, convert it to a string if so
+        if type(data) is list:
+            data = util.json_dumps(data)
+        self._store_rawdata(data, datasource)
 
     def _store_processeddata(self, processed_data, datasource):
         # processed is a Mime message, so write as string.
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -399,7 +399,7 @@ class Init(object):
 
     def _store_raw_vendordata(self, data, datasource):
         # Only these data types
-        if type(data) not in [bytes, str, list, None]:
+        if data is None or type(data) not in [bytes, str, list]:
             raise TypeError("vendordata_raw is unsupported type '%s'",
                             str(type(data)))
         # This data may be a list, convert it to a string if so
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -400,7 +400,8 @@ class Init(object):
     def _store_raw_vendordata(self, data, datasource):
         # Only these data types
         if type(data) not in [bytes, str, list, None]:
-            raise TypeError("vendordata_raw is unsupported type '%s'", type(data))
+            raise TypeError("vendordata_raw is unsupported type '%s'",
+                            type(data))
         # This data may be a list, convert it to a string if so
         if isinstance(data, list):
             data = util.json_dumps(data)
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -399,7 +399,7 @@ class Init(object):
 
     def _store_raw_vendordata(self, data, datasource):
         # This data may be a list or dict, convert it to a string if so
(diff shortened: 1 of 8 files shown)


smoser:
@eb3095 It definitely feels like you found a bug here, that was introduced by #777 (3cebe0df1e002bd85c8aa78e89f0ca507c17195a).  But I'm not sure that this is the right fix.

@andrewbogott, @TheRealFalcon  perhaps you have more context?

Openstack datasource gets vendordata and vendordata2 from [read_v2](https://github.com/canonical/cloud-init/blob/9bd19645a61586b82e86db6f518dd05c3363b17f/cloudinit/sources/helpers/openstack.py#L218).  If I'm reading that correctly, that will return the loaded json object, so it could be any json object (dict, list, string ... ).  Then it calls [convert_vendordata](https://github.com/canonical/cloud-init/blob/9bd19645a61586b82e86db6f518dd05c3363b17f/cloudinit/sources/DataSourceOpenStack.py#L164) on that.  convert_vendordata says clearly that it will return a string or list.

End result is that @eb3095 is correct, and that store_rawdata has to take any type, not just bytes.

It feels like I must be reading this wrong though... because otherwise I don't see how it would have passed any actual testing.

Help?

cloudinit/stages.py
```diff
@@ -395,6 +395,15 @@ def _store_rawdata(self, data, datasource):
         # Raw data is bytes, not a string
         if data is None:
             data = b''
```
smoser, line 397: The intent of the 'rawdata' storage was to store bit-for-bit the data that was presented in the datasource. At least that was the intent for user-data, so that if cloud-init got a blob that it didn't understand, something else could read that blob from here. (/var/lib/cloud/instance/userdata) cloud-init would have then provided that data in a consistent place across different cloud providers... and the reader could then just look there.

So from that perspective, we don't want to serialize data in any way here.  We want to write the raw bytes that we found.

The problem with that is that the raw bytes have not been stored in openstack (we decoded json into ds.vendordata_raw, and then in some cases even took a subset of it)... very much not _raw.

It feels like the typeerror here is actually correct. :cry: but we have to adjust what datasources are storing in vendordata or userdata_raw.

TheRealFalcon:
@smoser I don't think this is a new issue. I can go back a few years and set `vendordata_raw` to a list and still get this traceback:
```
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 650, in status_wrapper
    ret = functor(name, args)
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 372, in main_init
    init.update()
  File "/usr/lib/python3/dist-packages/cloudinit/stages.py", line 363, in update
    self._store_vendordata()
  File "/usr/lib/python3/dist-packages/cloudinit/stages.py", line 392, in _store_vendordata
    util.write_file(self._get_ipath('vendordata_raw'), raw_vd, 0o600)
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 1766, in write_file
    content = encode_text(content)
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 157, in encode_text
    return text.encode(encoding)
AttributeError: 'list' object has no attribute 'encode'
```

cloudinit/stages.py
```diff
@@ -397,6 +397,12 @@ def _store_rawdata(self, data, datasource):
...
         util.write_file(self._get_ipath('%s_raw' % datasource), data, 0o600)
 
+    def _store_raw_vendordata(self, data, datasource):
+        # This data may be a list or dict, convert it to a string if so
+        if type(data) in [list, dict]:
+            data = util.json_dumps(data)
```
smoser, line 403: a few things here:
a.) I think i'd rather fail on a dict (with a good error message). As far as I can see, no datasources should be storing a dict in vendordata_raw.  Since it was really *supposed* to be bytes, lets not slip any further into letting more types in.
b.) You probably have to allow for None.
c.) you should error on any other type. If you happened to have an int there, then it will go on through to store_rawdata, which will probably error with a less useful message.

So, how about:

```python
def _store_raw_vendordata(self, data, datasource):
    # vendordata_raw should only be one of None, list, bytes, or string.
    if data is None:
        data = b''
    elif isinstance(data, list):
        data = util.json_dumps(data)
    elif isinstance(data, bytes):
        pass
    elif isinstance(data, string):
        data = data.encode('utf-8')
    else:
        raise TypeError("vendordata_raw is unsupported type '%s'", type(data))
    self._store_rawdata(data, datasource)
```

smoser (changes requested):
thank you.

One comment.

cloudinit/stages.py
eb3095:
```
def _store_raw_vendordata(self, data, datasource):
        # Only these data types
        if type(data) not in [bytes, string, list, None]:
            raise TypeError("vendordata_raw is unsupported type '%s'", type(data))
        # This data may be a list, convert it to a string if so
        if isinstance(data, list):
            data = util.json_dumps(data)
        self._store_rawdata(data, datasource)
```

This should do the same thing. Don't really need,

```
    if data is None:
        data = b''
```

This is done in _store_rawdata already.

```
elif isinstance(data, string):
        data = data.encode('utf-8')
```

Not sure about this line either, considering that,

util.py
```
content = encode_text(content)
```

was throwing the error in the first place in write_files, and thats just calling .encode with the benefit of being able to specify and, encoding='utf-8', so its the default as well.

So let me know if you still want those lines in there as it looks like it my just be duplicating code.

smoser: Yours looks good now.   thank you.

smoser (approved):

smoser (approved):

smoser:
@eb3095 could you write a single coherent commit message?

You can either put it here (in the issue description) or squash your commits and write a single good message and then force-push.

Here is a suggested commit message:

     Fix stack trace if vendordata_raw contained an array.

     The implementation in existing datasources means that vendordata_raw is
     not "raw" as it ideally would be. Instead, actual values may include
     bytes, string or list.  If the value was a list, then the attempt to
     persist that data to a file in '_store_rawdata' would raise a
     TypeError.

     The change is to encode with util.json_dumps (which is safe for
     binary data) before writing.

     Co-authored-by: Eric Benner <ebenner@vultr.com>


We want the commit message to have less than 74 chars per line.

Thanks for your patience.

eb3095:
> @eb3095 could you write a single coherent commit message?
> 
> You can either put it here (in the issue description) or squash your commits and write a single good message and then force-push.
> 
> Here is a suggested commit message:
> 
> ```
>  Fix stack trace if vendordata_raw contained an array.
> 
>  The implementation in existing datasources means that vendordata_raw is
>  not "raw" as it ideally would be. Instead, actual values may include
>  bytes, string or list.  If the value was a list, then the attempt to
>  persist that data to a file in '_store_rawdata' would raise a
>  TypeError.
> 
>  The change is to encode with util.json_dumps (which is safe for
>  binary data) before writing.
> 
>  Co-authored-by: Eric Benner <ebenner@vultr.com>
> ```
> 
> We want the commit message to have less than 74 chars per line.
> 
> Thanks for your patience.

Looks like this was merged, did you still need that?

TheRealFalcon:
> Looks like this was merged, did you still need that?

Sorry, no. I merged with the suggested commit message.

PR #844, opened 2021-03-15
Fix mis-detecting network configuration in initramfs cmdline

TheRealFalcon:
**Integration test still needs to be written**

## Proposed Commit Message

```
Fix mis-detecting network configuration in initramfs cmdline

klibc initramfs in debian allows the 'iscsi_target_ip=' cmdline
parameter to specify an iscsi device attachment. This can
cause cloud-init to mis-detect the cmdline paramter as a
networking config.

LP: #1919188
```

## Additional Context
https://bugs.launchpad.net/cloud-init/+bug/1919188

## Test Steps
**Integration test still needs to be written**

## Checklist:

### DIFF
--- a/cloudinit/net/cmdline.py
+++ b/cloudinit/net/cmdline.py
@@ -73,7 +73,7 @@ class KlibcNetworkConfigSource(InitramfsNetworkConfigSource):
                 (ii) an open-iscsi interface file is present in the system
         """
         if self._files:
-            if re.search(r'(^|\s)ip[6]*=', self._cmdline):
+            if re.search(r'(^|\s)ip6?=', self._cmdline):
                 return True
             if os.path.exists(_OPEN_ISCSI_INTERFACE_FILE):
                 # iBft can configure networking without ip=
--- a/cloudinit/net/cmdline.py
+++ b/cloudinit/net/cmdline.py
@@ -12,7 +12,7 @@ import gzip
 import io
 import logging
 import os
-import re
+import shlex
 
 from cloudinit import util
 
@@ -73,8 +73,9 @@ class KlibcNetworkConfigSource(InitramfsNetworkConfigSource):
                 (ii) an open-iscsi interface file is present in the system
         """
         if self._files:
-            if re.search(r'(^|\s)ip6?=', self._cmdline):
-                return True
+            for item in shlex.split(self._cmdline):
+                if item.startswith('ip=') or item.startswith('ip6='):
+                    return True
             if os.path.exists(_OPEN_ISCSI_INTERFACE_FILE):
                 # iBft can configure networking without ip=
                 return True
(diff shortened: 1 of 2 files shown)


cloudinit/net/cmdline.py
```diff
@@ -72,7 +73,7 @@ def is_applicable(self) -> bool:
                 (ii) an open-iscsi interface file is present in the system
         """
         if self._files:
-            if 'ip=' in self._cmdline or 'ip6=' in self._cmdline:
+            if re.search(r'(^|\s)ip[6]*=', self._cmdline):
```
OddBloke, line 76: I think `[6]*` is slightly incorrect, it would match, e.g., `ip666` (the addressing protocol of the beast); `6?` will match 0 or 1 "6"s. (Either way, the square brackets are superfluous: a character and a bracket expression containing only that character are equivalent.) This is borderline nit territory, but as regexes are hard enough to read at the best of times, I think it's worth being exact here.

I think we're also missing tests for the `^` choice of the subexpression here: nothing fails if I edit the regex to only have `\s` there instead of the subexpression. Can you add a test for that?

mwhudson: I'm just flying past here, but you should really "parse" the kernel cmdline with shlex.split, because that's essentially what the initramfs is doing. This regex would match something like the admittedly extremely contrived `opt="bunch of stuff with spaces and ip=foo"`. Also the fewer regexes in the codebase the better!

TheRealFalcon:
@OddBloke Updated based on review comments.

I attempted writing an integration test for this behavior, but I was unable to create one. It requires
1. Passing kernel command line arguments
2. The datasource indicating there is no network config

#1 is doable but hacky. I can't figure out a way to do #2. Can you think of a realistic way to do #2? If not, are you ok having just unit test coverage for this one?

TheRealFalcon:
Updated (as a separate commit) to do this using shlex instead.

mwhudson:
Thanks, that makes me much less queasy :-)

I think it would make sense to parse the kernel command line into a dict more generally but that's a much bigger change of course.

OddBloke (approved):
This LGTM, thanks! I wonder if we should add a test for quoted content (i.e. cmdline of `foo="ip=shouldntbedetected"`); I'll let you decide and land this with or without it.

TheRealFalcon:
Separate from this one? https://github.com/canonical/cloud-init/pull/844/files#diff-402c7850e179c77acffc2f2a35c464a2ba0bdb6517060dbf8a9f6a6a571432b3R4435

OddBloke:
> Separate from this one? https://github.com/canonical/cloud-init/pull/844/files#diff-402c7850e179c77acffc2f2a35c464a2ba0bdb6517060dbf8a9f6a6a571432b3R4435

That looks good, no idea how I missed it first time through!

PR #845, opened 2021-03-16
Fix apt default integration test

TheRealFalcon:
The apt default test wasn't ported over from cloud-tests correctly.
Fix it so it works as it did in the cloud tests.

Also add openstack specific test for apt configuration with no uri.

## Proposed Commit Message

```
Fix apt default integration test

The apt default test wasn't ported over from cloud-tests correctly.
uri should be specified in the test, but it was not, so the test
failed on openstack (and likely other platforms) because without
a specified uri, the default uri will vary by platform. I separated
this uri test out into a separate test function.

Also add openstack specific test for apt configuration with no uri.
Other platform-specific tests should be added here over time.
```

## Additional Context
n/a

## Test Steps
Run integration tests on openstack

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_apt.py
+++ b/tests/integration_tests/modules/test_apt.py
@@ -223,18 +223,10 @@ class TestDefaults:
 
         When no uri is provided.
         """
-        version = class_client.execute(
-            'curl http://169.254.169.254'
-        ).split('\n')[-1]
-        zone = class_client.execute(
-            'curl http://169.254.169.254/{}/meta-data/placement/'
-            'availability-zone'.format(version),
-        )
-
+        zone = class_client.execute('cloud-init query v1.availability_zone')
         sources_list = class_client.read_from_file('/etc/apt/sources.list')
         assert '{}.clouds.archive.ubuntu.com'.format(zone) in sources_list
 
-    @pytest.mark.user_data(DEFAULT_DATA)
     def test_security(self, class_client: IntegrationInstance):
         """Test apt default security sources.
 
--- a/tests/integration_tests/modules/test_apt.py
+++ b/tests/integration_tests/modules/test_apt.py
@@ -200,30 +200,41 @@ class TestApt:
         assert conf_exists is False
 
 
-DEFAULT_DATA = """\
+_DEFAULT_DATA = """\
 #cloud-config
 apt:
   primary:
     - arches:
       - default
+      {uri}
   security:
     - arches:
       - default
 """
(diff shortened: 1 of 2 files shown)


OddBloke:
Some information in the commit message about how the porting was incorrect would be good: I'm struggling to determine what here is remedial, and what is new testing.

tests/integration_tests/modules/test_apt.py
```diff
@@ -200,30 +200,41 @@ def test_os_pipelining(self, class_client: IntegrationInstance):
...
+        zone = class_client.execute(
+            'curl http://169.254.169.254/{}/meta-data/placement/'
+            'availability-zone'.format(version),
+        )
+
         sources_list = class_client.read_from_file('/etc/apt/sources.list')
-        assert 'deb http://archive.ubuntu.com/ubuntu' in sources_list
+        assert '{}.clouds.archive.ubuntu.com'.format(zone) in sources_list
 
+    @pytest.mark.user_data(DEFAULT_DATA)
```
OddBloke, line 237: Don't think we need this?

tests/integration_tests/modules/test_apt.py
```diff
@@ -200,30 +200,41 @@ def test_os_pipelining(self, class_client: IntegrationInstance):
...
-        tests/cloud_tests/testcases/modules/apt_configure_primary.py
+        When no uri is provided.
         """
+        version = class_client.execute(
+            'curl http://169.254.169.254'
+        ).split('\n')[-1]
+        zone = class_client.execute(
+            'curl http://169.254.169.254/{}/meta-data/placement/'
+            'availability-zone'.format(version),
+        )
```
OddBloke, line 232: Could we use `cloud-init query v1.availability_zone` here instead?

OddBloke: Generalising it is out-of-scope for this PR, but if we do, can this test run on more than just OpenStack? We perform region/AZ substitution on every cloud; an EC2 instance I have kicking around, for example:

```
deb http://eu-west-1.ec2.archive.ubuntu.com/ubuntu/ focal main restricted
```

TheRealFalcon: Yep, I noticed there there were different templates for different clouds, so I figured we would want multiple versions, but since I was just testing on openstack, I started with openstack.

TheRealFalcon:
I updated the commit message on the PR (not in the actual commit). Is that enough context?

OddBloke (approved):
Thanks, exactly what I needed. Confirmed that these tests pass locally.

PR #857, opened 2021-03-24
- Detect a Python version change and clear the cache

rjschwei:
+ When a distribution gets updated it is possible that the Python version
    changes. Python makes no guarantee that pickle is consistent across
    versions as such we need to purge the cache and start over.

## Proposed Commit Message

```
summary: Clear cache when a Python version change is detected

When a distribution gets updated it is possible that the Python version
changes. Python makes no guarantee that pickle is consistent across
versions as such we need to purge the cache and start over.
```

## Additional Context
# python
Python 2.7.17 (default, Nov 04 2019, 16:08:54) [GCC] on linux2
Type "help", "copyright", "credits" or "license" for more information.
import pickle

class A():
    def __init__(self):
        self.me = 'a'
    def get_me(self):
        return self.me

a = A()
fout = open('data.pickle', 'wb')
pickle.dump(a, fout)

# python3
Python 3.4.10 (default, Dec 13 2019, 16:20:47) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
import pickle
fin = open('data.pickle', 'rb')
a = pickle.load(fin)
AttributeError: Can't get attribute 'A' on <module '__main__' (built-in)>

## Test Steps
- Start any instance
- Add the file /var/lib/cloud/pyver with content 1.9 (or anything other than the major.minor number of the Python interpreter on the instance
- Reboot
- Purge message is in log file

## Checklist:

 - [x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)


TheRealFalcon:
Hi @rjschwei , thanks for the contribution. [According to the documentation](https://docs.python.org/3/library/pickle.html#pickle.load), when a pickle is loaded, the pickle version is auto-detected. Older versions of python can't read the newer pickle formats, but newer versions of python can read the older pickle formats. That means that upgrade shouldn't be a problem.

Using the example you provided, I can't even reload the pickle written using the exact same version of python because the class isn't provided when you're trying to read it. Here's an example where I can write out a pickle using python 2.7 and read it back using python 3.5:

```python
root@my-xenial:~# cat pickle_stuff.py
import pickle
import sys

class A():
    def __init__(self, me):
        self.me = me
    def get_me(self):
        return self.me

if sys.argv[1] == 'write':
    a = A('me')
    with open('/tmp/data.pickle', 'wb') as f:
        pickle.dump(a, f)
elif sys.argv[1] == 'read':
    with open('/tmp/data.pickle', 'rb') as f:
        a = pickle.load(f)
    print(a.me)

root@my-xenial:~# python --version
Python 2.7.12
root@my-xenial:~# python3 --version
Python 3.5.2
root@my-xenial:~# python pickle_stuff.py write
root@my-xenial:~# python3 pickle_stuff.py read
me
```

Given this, I don't think this PR is necessary. Please let me know if there's something I'm missing.

rjschwei:
Yes, I am aware of the documentation claiming that pickle is forward compatible. Yet we have reports from @Moustafa-Moustafa that there have been issues in Azure across distribution related to obj.pkl. I have not dug as deep as I probably should have. Maybe @Moustafa-Moustafa can fill in some details for me. A work around was put into place that indiscriminately removes obj.pkl from the unit file. I was not willing to follow this path and as such thought we should be more picky. Of course this implementation still has a brought stroke as it removes more than absolutely necessary on Py version changes.

Moustafa-Moustafa:
#

> Yes, I am aware of the documentation claiming that pickle is forward compatible. Yet we have reports from @Moustafa-Moustafa that there have been issues in Azure across distribution related to obj.pkl. I have not dug as deep as I probably should have. Maybe @Moustafa-Moustafa can fill in some details for me. A work around was put into place that indiscriminately removes obj.pkl from the unit file. I was not willing to follow this path and as such thought we should be more picky. Of course this implementation still has a brought stroke as it removes more than absolutely necessary on Py version changes.

On RHEL 7.9, cloud-init 19.4 is installed and it runs on Python 2.7. When the upgrade completes, we end up on RHEL 8.2 with cloud-init 19.4 but now it runs on Python 3.6. And the same goes when upgrading SLES12SP5 to SLES15.
Before the upgrade, cloud-init caches DataSourceAzure which includes the user-data. The user-data is MIME multipart object. On Python 3.6, MIME multipart has a new required attribute “policy” that didn’t exist in Python 2.7. After the upgrade, cloud-init initializes the DataSourceAzure based on the cache which would be missing the required attribute “policy” for the user-data object. That causes cloud-init “init” stage to fail and eventually not mounting the resource disk.

OddBloke:
Hey folks,

On Thu, Mar 25, 2021 at 09:23:18AM -0700, Moustafa Moustafa wrote:
> #
> 
> > Yes, I am aware of the documentation claiming that pickle is forward
> > compatible. Yet we have reports from @Moustafa-Moustafa that there
> > have been issues in Azure across distribution related to obj.pkl. I
> > have not dug as deep as I probably should have. Maybe
> > @Moustafa-Moustafa can fill in some details for me. A work around
> > was put into place that indiscriminately removes obj.pkl from the
> > unit file. I was not willing to follow this path and as such thought
> > we should be more picky. Of course this implementation still has a
> > brought stroke as it removes more than absolutely necessary on Py
> > version changes.
> 
> On RHEL 7.9, cloud-init 19.4 is installed and it runs on Python 2.7.
> When the upgrade completes, we end up on RHEL 8.2 with cloud-init 19.4
> but now it runs on Python 3.6. And the same goes when upgrading
> SLES12SP5 to SLES15.  Before the upgrade, cloud-init caches
> DataSourceAzure which includes the user-data. The user-data is MIME
> multipart object. On Python 3.6, MIME multipart has a new required
> attribute “policy” that didn’t exist in Python 2.7. After the upgrade,
> cloud-init initializes the DataSourceAzure based on the cache which
> would be missing the required attribute “policy” for the user-data
> object. That causes cloud-init “init” stage to fail and eventually not
> mounting the resource disk.

Thanks for the background here Moustafa!  This doesn't sound like a
fundamental issue with pickle formats, this sounds like a change in what
cloud-init's code (if we include the stdlib libraries it depends on)
expects the objects _within_ the pickle to contain.

We recently introduced a lightweight framework for handling upgrade
issues like this, and it sounds like this could be a usecase for it.
The code in question is
https://github.com/canonical/cloud-init/blob/master/cloudinit/persistence.py
and you can see it "in action" in the commit that introduced that file:
https://github.com/canonical/cloud-init/commit/4f2da1cc1d24cbc47025cc8613c0d3ec287a20f9#diff-2902e7f3123dd1e701482d6c60d6ca2e51d2dcbe3f643cb30be70727fde3fbfb
In that case, we're detecting that the pickle we're loading predates the
introduction of the `networking` API, and fix it up to have the
attributes we expect.

The issue described herein is not exactly the same as that one: it's a
CPython code change that's the problem, rather than a cloud-init code
change.  However, I suspect that, for our purposes, the issues are the
same: the data we're loading needs some processing performed before it
will work with the current version of (cloud-init + CPython).

Could we try addressing this via the upgrade framework, rather than via
this much heavier hammer?


Cheers,

Dan

rjschwei:
The upgrade framework is certainly more targeted and an elegant approach for stuff in cloud-init. My question would be, do we really want to chase every change across every CPython version for the cloud-init dependencies. From my point of view that answer to that question is "No".

Secondly I don't see blowing away the pickle file and subsequently re-crating it as a big issue. Yes there is the "a few milliseconds" performance issue that someone will undoubtedly raise. But the system has just gone through a major update and will require a reboot, i.e. we have a service outage. I just don't see where the remove-recreate cycle has any practical implications.

cloudinit/cmd/main.py
```diff
@@ -276,6 +276,24 @@ def main_init(name, args):
...
+        # pickled previously is useless.
+        if cached_pyver != pyver:
+            LOG.debug('Python version change detected purging cache')
+            init.purge_cache(True)
+    else:
+        LOG.debug(
+            'Could not determin ePython version used to write cache, purging'
+        )
+        init.purge_cache(True)
+    with open(pyrefver, 'w') as pyref:
```
TheRealFalcon, line 294: There is a `util.write_file` helper function. Let's use that here.

TheRealFalcon (commented):
Hi @rjschwei , my apologies for waiting so long on this one. There has been some debate within the team of whether we want to support this, which is why it has taken us so long. The good news is that this is something we want to do now.

This is a good start, but in order to merge this we would still need a few more changes. Could you move this functionality into its own function instead of being inline? Also, could you add some unit tests for this behavior? We'll want an integration test too, but that is something I can help with. I also left a few comments inline.

rjschwei:
@TheRealFalcon Added a couple of unit tests. These are not quite complete as I couldn't figure out how I would verify that `init.purge_cache(True)` was called. Help with testing is definitely appreciated, my mock skills are terrible.

TheRealFalcon:
@rjschwei , since you wanted some help, I took the liberty of pushing to your branch. We can always remove/change my commit if you don't like it. I also added integration tests for the behavior.

I decided a unit test for this might not be all that valuable due to the large number of mocks it would require. Literally every line of the function would need to be mocked except for the if statements, and given that we have the integration tests testing the actual behavior, I'm not sure how such unit test could add any value.

Let me know what you think overall.

@blackboxsw , would you be able to review my changes as well? Do you agree about the unit tests?

rjschwei (commented):
@TheRealFalcon thanks LGTM

cloudinit/cmd/main.py
```diff
@@ -210,6 +210,37 @@ def attempt_cmdline_url(path, network=True, cmdline=None):
...
+            init.purge_cache(True)
+            util.write_file(python_version_path, current_python_version)
+    else:
+        if os.path.exists(init.paths.get_ipath_cur('obj_pkl')):
+            # Only worrying about purging if there's cached instance data
+            LOG.debug(
+                'Could not determine Python version used to write cache. '
+                'Purging cache'
+            )
+            init.purge_cache(True)
```
blackboxsw, line 240: Thanks for this PR and discussion, I'm good with the principle of purging cache across python minor version upgrades as you are more likely to stumble over inconsistencies in stdlib dependencies that cloud-init doesn't control. I'm just not thrilled with this big hammer here in the "else" clause.

What it looks like we are doing here is blowing away the cache on all systems which upgrade cloud-init, because they don't yet track the python_version against which the pickle was created. purging all caches due to a cloud-init package upgrade then exposes all systems, regardless of how old they are and whether or not they have changed their python version, to forcibly re-run datasource detection and attempt th crawl their cloud metadata services. The problem in this case is that those systems could no longer have access to their metadata or configdrives etc and cannot possibly redetect or re-run the same setup.

 Wouldn't it instead make more sense for us to just skip this whole else clause, and just call the write_file so that subsequent python major.minor upgrades get triggered?

Alternatively, distro packaging could also remedy this is within the packaging scripts of cloud-init, in the postinst or %post scripts for the package. If postinst script has visibility to the package we are upgrading from (as debian postinst does) and we know the packaging versions present on RHEL7 vs RHEL8 triggers this change, couldn't the postinst say if prev_ver <= "<VER_FROM_RHEL7>" and current_ver >= "<VER_FROM_RHEL8>" then rm obj.pkl.


I'm trying to come up with alternatives because I really feel like this bug we are trying to fix here seems datasource-specific across an even more specific python major upgrade. I don't want all cloud-init consumers on every distro and to retrigger a metadata crawl across package upgrade when this is really a corner case user-story.  If we are able to drop the 'else' I have no qualms with instrumenting this solution "going-forward". It just means that RHEL7 would need to see an upgrade to cloud-init 21.X (which persists the python version) in order to support upgrades. (Or help documentation/script that tells a customer to write out python major.minor to /var/lib/cloud/instance/data/python-version before major distro upgrade)

TheRealFalcon: @blackboxsw , yes I agree. I didn't realize that it would trigger a metadata recrawl on every single upgraded instance. I pushed a new commit with that part removed.

In the shorter term, I agree that i makes more sense for things to be handled in a distro-specific way. If a distro can handle removing the pickle in their packaging/upgrade code, that seems ideal.

rjschwei: Yes no longer having access to the config-drive or metadata service is something I did not consider, thanks for catching this. As such purging blindly is not a good idea. That said the idea was to help the user which we miss again with the latest push from @TheRealFalcon and the change of the `else` clause because we just write the Python version, then at some point later the cache read may fail and we are in the same situation as we are today, meaning it is not trivial to figure out what's going on when the cache doesn't load. So at the very least, if we do not want to blindly clean the cache, which I agree with, we should write a message to the log that leaves a bread crum for debugging purposes.

TheRealFalcon: > we should write a message to the log that leaves a bread crum for debugging purposes.

@rjschwei , Can you explain what you mean by this? What would we write? When would we write it?

Is there code that could be added as part of the distro upgrade process? Once the initial upgrade has been performed to a version of cloud-init that has this change, this shouldn't be a problem anymore.

rjschwei: @TheRealFalcon I am not fond of the "what the distro can do"idea in this case as we will simply get different behavior across distros and that makes for an inconsistent user experience. Which is the reason I worked on this in the first place rather than just clearing the cache in %post and calling it a day. So if we go down the "distro package should do something path" I think we would need to clearly document what we want that something to be in the hopes that all packagers will adhere to our recommendation.

Also that eliminates the "someone has a cache" and builds their own Python environment + cloud-init use case. We can argue how important that is but the use case probably exists.

As far as the breadcrumbs are concerned the issue is that at the place in the process when we create the py_ver file we do not yet know if the cache can actually be loaded. So we might write the file when the cache should be purged. In this case we could emit a message "Writing py_ver file, but cache compatibility status is unknown" or something more verbose. Then we would at least leave a hint in the log that if there is a cache loading issue that the cache should have been purged to begin with.

TheRealFalcon: >  I am not fond of the "what the distro can do"idea in this case as we will simply get different behavior across distros and that makes for an inconsistent user experience. Which is the reason I worked on this in the first place rather than just clearing the cache in %post and calling it a day.

I understand the frustration, but do we have another option? We're already agreed that clearing the cache for ALL users is too disruptive. The only other way this could be fixed upstream is to do the thing that OddBloke initially suggested and use the upgrade framework to fix the root issue here:
`On Python 3.6, MIME multipart has a new required attribute “policy” that didn’t exist in Python 2.7.`.

If we don't do that, then the solution has to be distro-specific, right? If we do fix that, then (along with this cache clear on version change) this issue should disappear entirely. I'm now leaning towards these two things together being the optimal solution here.

> In this case we could emit a message "Writing py_ver file, but cache compatibility status is unknown" or something more verbose. Then we would at least leave a hint in the log that if there is a cache loading issue that the cache should have been purged to begin with.

That makes sense. If the cache exists and the python version file doesn't exist, I think logging a one-time warning is reasonable...but we shouldn't need to do this (or a packaging-specific solution) if we fix it via the upgrade framework.

rjschwei: Sorry if there is a misunderstanding. I am not advocating to revisit previous conversations. IMHO using the upgrade framework for this puts us into a more or less continuous chasing loop and thus I think that's not a good idea. I suggest that we add a logging message to the current proposal in the `else:` code branch and call it good.

TheRealFalcon: > IMHO using the upgrade framework for this puts us into a more or less continuous chasing loop and thus I think that's not a good idea.

I'm suggesting we do both. Purging the cache on python version upgrade fixes the issue so we don't have a continuous chasing loop, but it still doesn't fix this MIME issue that happens when upgrading to the fixed version. Fixing that with the upgrade framework can be a "one off" fix for a problem that we'll no longer see with the python version checking in place.

If we do both of these things together, then there's no need for distro-specific upgrade code. I think it's the easiest path forward.

We can also add a logging message.

TheRealFalcon: @rjschwei and @blackboxsw , does this plan work for you? If so, I think we can merge this current PR as-is, then I can create a new PR to fix the specific MIME issue that was mentioned here.

tests/integration_tests/modules/test_version_change.py
```diff
@@ -0,0 +1,55 @@
...
+def _assert_no_pickle_problems(log):
+    assert 'Failed loading pickled blob' not in log
+    assert 'Traceback' not in log
+    assert 'WARN' not in log
+
+
+def test_reboot_without_version_change(client: IntegrationInstance):
+    log = client.read_from_file('/var/log/cloud-init.log')
+    assert 'Python version change detected' not in log
+    assert 'Could not determine Python version used to write cache' not in log
```
blackboxsw, line 20: I think this needs to now be the following right throughout?
```suggestion
    assert 'Cache compatibility status is currently unknown.' not in log    
```

cloudinit/cmd/main.py
```diff
@@ -210,6 +210,35 @@ def attempt_cmdline_url(path, network=True, cmdline=None):
...
+        init.paths.get_cpath('data'), 'python-version'
+    )
+    if os.path.exists(python_version_path):
+        cached_python_version = open(python_version_path).read()
+        # The Python version has changed out from under us, anything that was
+        # pickled previously is likely useless due to API changes.
+        if cached_python_version != current_python_version:
+            LOG.debug('Python version change detected. Purging cache')
+            init.purge_cache(True)
+            util.write_file(python_version_path, current_python_version)
```
blackboxsw, line 232: This write_file and in else, could be pushed up scope below the if/else blocks

blackboxsw (commented):
LGTM, with the caveat that a followup integration test for MIME type handling covers our bases. Just some fixes on the integration test asserts given that the logging changed.

cloudinit/cmd/main.py
rjschwei: @TheRealFalcon LGTM Thanks

cloudinit/cmd/main.py
TheRealFalcon: No need to write it every time, right? This way will write it if it doesn't exist or if the python version has changed. Otherwise, it won't write it.

TheRealFalcon:
Thanks for the test update @blackboxsw . I also updated another integration test accordingly

blackboxsw (approved):
LGTM! thanks @TheRealFalcon and @rjschwei.

PR #861, opened 2021-03-31
Azure: eject the provisioning iso before reporting ready

anhvoms:
## Proposed Commit Message
Azure: eject the provisioning iso before reporting ready

Due to hyper-v implementations, iso ejection is more efficient if performed 
from within the guest. The code will attempt to perform a best-effort ejection.
Failure during ejection will not prevent reporting ready from happening. If iso
ejection is successful, later iso ejection from the platform will be a no-op.
In the event the iso ejection from the guest fails, iso ejection will still happen at
the platform level.

## Checklist:


TheRealFalcon (approved):
Looks good to me! I'm also pushing a rebase that resolves the conflict.

PR #862, opened 2021-04-01
ConfigDrive: Add support for bridge interfaces

EBCEEB:
Add support to configure network bridge interfaces via network_data.json (derived from bond config)


EBCEEB:
Unfortunately I'm unable to prepare tests for this change :-(

PR #863, opened 2021-04-07
Fix unpickle for source paths missing run_dir

lucasmoura:
## Proposed Commit Message
Fix unpickle for source paths missing run_dir

On the source class, we require the use of paths.run_dir to perform some operations. On older cloud-init version, the Paths class does not have the run_dir attribute. To fix that, we are now manually adding that attribute in the source paths
object if doesn't exist in the unpickle operation.

LP: #1899299

## Test Steps
To reproduce this issue, run the following script:
```sh
#!/bin/sh
set -x

series=trusty
name=$series-upgrade

lxc delete --force $name
lxc launch ubuntu-daily:$series $name
sleep 10

lxc exec $name -- add-apt-repository ppa:ua-client/stable -y
lxc exec $name -- sudo apt-get update
lxc exec $name -- sudo apt-get install ubuntu-advantage-tools -y
lxc exec $name -- ua version
lxc exec $name -- sudo ua attach <uaclient-token>
lxc exec $name -- sudo ua status
lxc exec $name -- sudo add-apt-repository ppa:cloud-init-dev/daily -y
lxc exec $name -- sudo apt-get update
lxc exec $name -- sudo mkdir -p /etc/update-manager/release-upgrades.d  
lxc exec $name -- sh -c "echo '[Sources]\nAllowThirdParty=yes' > allow.cfg"
lxc exec $name -- sudo mv allow.cfg /etc/update-manager/release-upgrades.d   
lxc exec $name -- sudo apt-get upgrade -y
lxc exec $name -- sudo apt-get dist-upgrade -y
lxc exec $name -- sudo do-release-upgrade --frontend DistUpgradeViewNonInteractive
lxc exec $name -- sudo ua status --wait

lxc exec $name -- sudo reboot
sleep 60
lxc exec $name -- sudo ua status --wait

set +x
```

By looking at the `/var/log/ubuntu-advantage.log` we will see that the error is because `cloud-id` fails with the error:
```
AttributeError: 'Paths' object has no attribute 'run_dir'
```

We can confirm that by running `cloud-id` on the container directly.

To verify that the fix works, run the following commands:
```
lxc file push cloudinit/helpers.py trusty-upgrade/usr/lib/python3/dist-packages/cloudinit/helpers.py
lxc exec trusty-upgrade -- reboot
sleep 30
lxc exec trusty-upgrade -- cloud-id
lxc exec trusty-upgrade -- ua status
```
## Checklist:


cloudinit/sources/__init__.py
```diff
@@ -221,6 +223,16 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
     def __str__(self):
         return type_utils.obj_name(self)
 
+    def _unpickle(self, ci_pkl_version: int) -> None:
+        """Perform deserialization fixes for Distro."""
+        if "run_dir" not in self.paths.__dict__:
```
OddBloke, line 228:
```suggestion
        if not hasattr(self.paths, "run_dir"):
```

because it handles properties appropriately:

```py
In [1]: class Foo:
   ...:     @property
   ...:     def run_dir(self):
   ...:         pass
   ...: 

In [2]: "run_dir" in Foo().__dict__
Out[2]: False

In [3]: hasattr(Foo(), "run_dir")
Out[3]: True
```

cloudinit/sources/__init__.py
```diff
@@ -221,6 +223,16 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
...
+            # On older versions of cloud-init the Paths class do not
+            # have the run_dir attribute. This is problematic because
+            # when loading the pickle object on newer versions of cloud-init
+            # we will rely on this attribute. To fix that, we are now
+            # manually adding that attribute here.
+            self.paths.run_dir = "/run/cloud-init"
```
OddBloke, line 234: This is likely safe, but `run_dir` _is_ configurable. It's possible that someone could have configured a non-default `run_dir` even without a cloud-init that supports it (e.g. by applying the same configuration file to all their images). To ensure that we pick such configuration up, I think we should probably reinstantiate `Paths` with the same configuration as the pickled one, and then pull `run_dir` out of it. Roughly:

```py
self.paths.run_dir = Paths(self.paths.cfgs, ds=self.paths.datasource).run_dir
```

Does that make sense? (Do you agree?)

cloudinit/sources/__init__.py
```diff
@@ -221,6 +223,16 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
     def __str__(self):
         return type_utils.obj_name(self)
 
+    def _unpickle(self, ci_pkl_version: int) -> None:
```
OddBloke, line 226: Does this need to be on `DataSource`? Could it instead be on `Paths` itself?

As background, the `networking` fix had to be in `Distro._unpickle` because affected `Networking` instances don't have any instance state: this means that they get unpickled differently, so the hooks we use don't get called. In this case, however, all `Paths` instances do have instance state, so the hooks we use _should_ get called.

cloudinit/sources/__init__.py
```diff
@@ -221,6 +223,16 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
...
+        """Perform deserialization fixes for Distro."""
```
OddBloke, line 227: This needs updating.

OddBloke (changes requested):
Thanks Lucas! It's good to see this framework seems to be up to the second challenge it's presented with! I have some inline thoughts, but this is a good first pass.

cloudinit/sources/__init__.py
lucasmoura: It does make sense. I missed that this could configured by the user. So doing that is the safest approach.

cloudinit/sources/__init__.py
lucasmoura: You are totally right here. I missed that the `Paths`  object can be instantiated directly. I will move the code there

lucasmoura:
@OddBloke Thank you for the review and the context here. I have updated the code

TheRealFalcon (approved):

cloudinit/helpers.py
```diff
@@ -317,7 +318,9 @@ def items(self):
         return list(self.registered.items())
 
 
-class Paths(object):
+class Paths(persistence.CloudInitPickleMixin, object):
```
OddBloke, line 321: No need to inherit from `object` at all in Python 3:

```suggestion
class Paths(persistence.CloudInitPickleMixin):
```

OddBloke (approved):
Thanks Lucas! One minor nit which I'll apply now.

PR #866, opened 2021-04-13
doc: Replace remaining references to user-scripts as a config module

raharper:
## Proposed Commit Message

```
doc: Replace remaining references to user-scripts as a config module

git-grep showed a few more locations where we refer to a "user-scripts"
config module which is really cc_scripts_user module.  Replace these
references with slightly different language so as not to confuse
future me when looking for "user-scripts" vs. "scripts-user"
```

## Test Steps

```
tox -e doc
```

## Checklist:
that apply. -->


OddBloke (approved):
Thanks!

PR #869, opened 2021-04-16
Fix error on upgrade caused by new vendordata2 attributes

TheRealFalcon:
## Proposed Commit Message

```
Fix error on upgrade caused by new vendordata2 attributes

In #777, we added 'vendordata2' and 'vendordata2_raw' attributes to
the DataSource class, but didn't use the upgrade framework to deal
with an unpickle after upgrade. This commit adds the necessary
upgrade code.

Additionally, added a smaller-scope upgrade test to our integration
tests that will be run on every CI run so we catch these issues
immediately in the future.

LP: #1922739
```

## Additional Context
We haven't caught this is our current upgrade test because we haven't been checking for error status of `cloud-init status`. I added a check for that in the larger upgrade test as well.

## Test Steps
Since the latest cloud-init version is already in LXD images, it's not as straightforward. You can launch a xenial instance and then upgrade to the latest PPA/deb. For other releases, you can launch an instance, manually downgrade to 20.4.1, run a `cloud-init clean`, then upgrade to latest. I've manually reproduced the bug both ways and manually verified the bug both ways.

I also locally reproduced/verified the new integration test by launching the test against a xenial instance (CI uses focal).

### DIFF
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -24,6 +24,7 @@ from cloudinit import util
 from cloudinit.atomic_helper import write_json
 from cloudinit.event import EventType
 from cloudinit.filters import launch_index
+from cloudinit.persistence import CloudInitPickleMixin
 from cloudinit.reporting import events
 
 DSMODE_DISABLED = "disabled"
@@ -134,7 +135,7 @@ URLParams = namedtuple(
     'URLParms', ['max_wait_seconds', 'timeout_seconds', 'num_retries'])
 
 
-class DataSource(metaclass=abc.ABCMeta):
+class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
     dsmode = DSMODE_NETWORK
     default_locale = 'en_US.UTF-8'
@@ -196,6 +197,8 @@ class DataSource(metaclass=abc.ABCMeta):
     # non-root users
     sensitive_metadata_keys = ('merged_cfg', 'security-credentials',)
 
+    _ci_pkl_version = 1
+
     def __init__(self, sys_cfg, distro, paths, ud_proc=None):
         self.sys_cfg = sys_cfg
         self.distro = distro
@@ -217,6 +220,13 @@ class DataSource(metaclass=abc.ABCMeta):
             self.ud_proc = ud.UserDataProcessor(self.paths)
         else:
             self.ud_proc = ud_proc
+
+    def _unpickle(self, ci_pkl_version: int) -> None:
+        """Perform deserialization fixes for Paths."""
+        if not hasattr(self, 'vendordata2'):
+            self.vendordata2 = None
+        if not hasattr(self, 'vendordata2_raw'):
+            self.vendordata2_raw = None
 
     def __str__(self):
         return type_utils.obj_name(self)
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import pytest
 import time
 from pathlib import Path
@@ -8,6 +9,8 @@ from tests.integration_tests.conftest import (
     get_validated_source,
     session_start_time,
 )
+from tests.integration_tests.instances import CloudInitSource
+
 
 log = logging.getLogger('integration_testing')
 
@@ -93,6 +96,26 @@ def test_upgrade(session_cloud: IntegrationCloud):
         instance.install_new_cloud_init(source, take_snapshot=False)
         instance.execute('hostname something-else')
         _restart(instance)
+        assert instance.execute('cloud-init status --wait --long').ok
         _output_to_compare(instance, after_path, netcfg_path)
 
     log.info('Wrote upgrade test logs to %s and %s', before_path, after_path)
+
+
+@pytest.mark.ci
+@pytest.mark.ubuntu
+def test_upgrade_package(session_cloud: IntegrationCloud):
+    if get_validated_source(session_cloud) != CloudInitSource.DEB_PACKAGE:
+        not_run_message = 'Test only supports upgrading to build deb'
+        if os.environ.get('TRAVIS'):
+            # If this isn't running on CI, we should know
+            pytest.fail(not_run_message)
+        else:
+            pytest.skip(not_run_message)
+
+    launch_kwargs = {'image_id': session_cloud._get_initial_image()}
+
+    with session_cloud.launch(launch_kwargs=launch_kwargs) as instance:
+        instance.install_deb()
+        instance.restart()
+        assert instance.execute('cloud-init status --wait --long').ok
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -66,7 +66,7 @@ def test_upgrade(session_cloud: IntegrationCloud):
         return  # type checking doesn't understand that skip raises
 
     launch_kwargs = {
-        'image_id': session_cloud._get_initial_image(),
+        'image_id': session_cloud.released_image_id,
     }
 
     image = ImageSpecification.from_os_image()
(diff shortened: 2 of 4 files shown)


realizelol:
Thank you,

this fix is definitvely working.

tests/integration_tests/test_upgrade.py
```diff
@@ -93,6 +96,26 @@ def test_upgrade(session_cloud: IntegrationCloud):
...
+        assert instance.execute('cloud-init status --wait --long').ok
         _output_to_compare(instance, after_path, netcfg_path)
 
     log.info('Wrote upgrade test logs to %s and %s', before_path, after_path)
+
+
+@pytest.mark.ci
+@pytest.mark.ubuntu
+def test_upgrade_package(session_cloud: IntegrationCloud):
+    if get_validated_source(session_cloud) != CloudInitSource.DEB_PACKAGE:
```
OddBloke, line 108: Could this also work for -proposed, PPAs and `UPGRADE`? (Doesn't need to happen right now, ofc., but if we can expand the times we perform this test then why not?)

tests/integration_tests/test_upgrade.py
```diff
@@ -93,6 +96,26 @@ def test_upgrade(session_cloud: IntegrationCloud):
...
+        not_run_message = 'Test only supports upgrading to build deb'
+        if os.environ.get('TRAVIS'):
+            # If this isn't running on CI, we should know
+            pytest.fail(not_run_message)
+        else:
+            pytest.skip(not_run_message)
+
+    launch_kwargs = {'image_id': session_cloud._get_initial_image()}
```
OddBloke, line 116: Should we promote `_get_initial_image` to the public API?

(Side note, which I noticed when reading `_get_initial_image`s code: should we be caching that value so that we avoid getting different "initial" images if we call that before and after an image publication?)

tests/integration_tests/test_upgrade.py
```diff
@@ -93,6 +96,26 @@ def test_upgrade(session_cloud: IntegrationCloud):
...
+
+    with session_cloud.launch(launch_kwargs=launch_kwargs) as instance:
+        instance.install_deb()
+        instance.restart()
+        assert instance.execute('cloud-init status --wait --long').ok
```
OddBloke, line 121: I've been wondering: should we default to asserting that `cloud-init status` was successful for every test, with a mark (or marks) for tests which are intentionally launching instances in which cloud-init will exhibit a failure state?

(We could/should probably also check that there are no `WARN` or `Traceback` lines in `cloud-init.log`, and any other such things.)

cloudinit/sources/__init__.py
```diff
@@ -218,6 +221,13 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
         else:
             self.ud_proc = ud_proc
 
+    def _unpickle(self, ci_pkl_version: int) -> None:
+        """Perform deserialization fixes for Paths."""
+        if not hasattr(self, 'vendordata2'):
+            self.vendordata2 = None
+        if not hasattr(self, 'vendordata2_raw'):
+            self.vendordata2_raw = None
```
OddBloke, line 229: Obviously we need to address this issue so this isn't even a nit, but is it a bit strange that `vendordata2` uses attributes on `DataSource` when it's an OpenStack-specific concept?

cloudinit/sources/__init__.py
```diff
@@ -218,6 +221,13 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
         else:
             self.ud_proc = ud_proc
 
+    def _unpickle(self, ci_pkl_version: int) -> None:
+        """Perform deserialization fixes for Paths."""
+        if not hasattr(self, 'vendordata2'):
```
OddBloke, line 226: These should have a corresponding test pickle added to `tests/data/old_pickles` and assertions adding to `cloudinit/tests/test_upgrade.py`.

OddBloke (commented):
We're missing some non-integration testing here, and I have concerns about how we got here, but this otherwise looks pretty good to me; I'm going to run the test locally now.

OddBloke:
> I'm going to run the test locally now.

I've confirmed that the test in this branch fails when run against an old focal image with a deb built from main:

```
$ CLOUD_INIT_OS_IMAGE=ubuntu:e0c3495ffd48::ubuntu::focal CLOUD_INIT_CLOUD_INIT_SOURCE=./cloud-init_21.1-34-g45db197c-1\~bddeb\~20.04.1_all.deb pytest tests/integration_tests/test_upgrade.py
...
FAILED tests/integration_tests/test_upgrade.py::test_upgrade_package - assert False
```

and passes with a deb built from this branch:

```
$ CLOUD_INIT_OS_IMAGE=ubuntu:e0c3495ffd48::ubuntu::focal CLOUD_INIT_CLOUD_INIT_SOURCE=./cloud-init_21.1-36-g5933b497-1\~bddeb\~20.04.1_all.deb pytest tests/integration_tests/test_upgrade.py
...
PASSED
```

tests/integration_tests/test_upgrade.py
TheRealFalcon: What do you mean by before and after an image publication?

Although, now that I look at it again, should I just make [released_image_id](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/clouds.py#L114) public and use that instead?

tests/integration_tests/test_upgrade.py
TheRealFalcon: Yeah, I agree. Added a trello card

cloudinit/sources/__init__.py
TheRealFalcon: Yeah, I mistakenly assumed that it had to sit at that same level of abstraction of the parsing of other vendor data. There's been some chatter happening about it here: https://github.com/canonical/cloud-init/pull/777#discussion_r591661062

tests/integration_tests/test_upgrade.py
OddBloke: > What do you mean by before and after an image publication?

We'll call `_get_initial_image` once when `session_cloud` is first instantiated (in the entire test run), and once again here: if a release image is released between those two calls, we'll run this test and the other tests against different images.

> Although, now that I look at it again, should I just make [released_image_id](https://github.com/canonical/cloud-init/blob/master/tests/integration_tests/clouds.py#L114) public and use that instead?

Aha, yeah, I guess we are caching the value already! :+1:

tests/integration_tests/test_upgrade.py
OddBloke: IMO `test_upgrade` (a) does too many things, and (b) isn't really a test, but a way of using our testing framework to emit specialised output. As I mentioned above, introducing failure there is fine (and leaving this as-is is too, to be clear), but I do see this as a potential opportunity for us to decompose `test_upgrade` a little, into more specific tests.

tests/integration_tests/test_upgrade.py
```diff
@@ -93,6 +96,26 @@ def test_upgrade(session_cloud: IntegrationCloud):
         instance.install_new_cloud_init(source, take_snapshot=False)
         instance.execute('hostname something-else')
         _restart(instance)
+        assert instance.execute('cloud-init status --wait --long').ok
```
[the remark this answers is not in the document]
TheRealFalcon: I agree. I'm unhappy with this test in general and would like to re-work it in the near future. This was just an obvious "we're currently missing something really important by not checking this", so I threw it in.

cloudinit/sources/__init__.py
TheRealFalcon: I added a new test. It failed without this branch against the already existing pickles. Is there a reason to add a new pickle? I'm not sure I understand the reason for adding new pickles provided we start at a sufficiently old starting point.

TheRealFalcon:
@OddBloke I believe I addressed your comments. I had a question/comment about the unit test.
> I added a new test. It failed without this branch against the already existing pickles. Is there a reason to add a new pickle? I'm not sure I understand the reason for adding new pickles provided we start at a sufficiently old starting point.

cloudinit/sources/__init__.py
OddBloke: That should be fine, I think.

OddBloke (approved):
Thanks!

PR #870, opened 2021-04-16
Azure: mount the default provisioning iso location before attempting to find it through device listing

anhvoms:
## Proposed Commit Message
Mount the default provisioning iso location before attempting to find it through device listing

```
With a few exceptions, Azure VM deployments receive provisioning metadata 
through the provisioning iso presented as a cdrom device (/dev/sr0).
The existing code attempts to find this device by calling blkid to find all devices
that have either type iso9660 or udf. This can be very expensive if the VM has
a lot of disks. This commit will attempt to mount the default iso location 
first and only tries to use blkid to locate the iso location if the default mounting
location fails

```

## Checklist:


cloudinit/sources/DataSourceAzure.py
```diff
@@ -445,19 +444,30 @@ def crawl_metadata(self):
                 else:
                     ret = load_azure_ds_dir(cdev)
 
-            except NonAzureDataSource:
-                report_diagnostic_event(
-                    "Did not find Azure data source in %s" % cdev,
-                    logger_func=LOG.debug)
+            except (NonAzureDataSource, util.MountFailedError) as e:
```
OddBloke, line 447: This for loop feels increasingly overloaded to me: it feels like we have some unnecessary indirection by putting "magic" values into the list, when it would probably be simpler to circumvent the list entirely. Would something like this pseudocode make sense (as a replacement from ~line 411):

```py
ret = None
if os.path.isfile(REPROVISION_MARKER_FILE):
    ret, reprovision = None, True
elif os.path.isfile(NIC_ATTACH_MARKER_PRESENT):
    ret, reprovision_after_nic_attach = None, True
else:
     try:
         ret = load_azure_ds_dir(self.seed_dir)
    except (...):
        for cdev in list_possible_azure_ds_devs():
            try:
                ret = util.mount_cb(...)
                break
            except (...):
                ...
                continue
    if not ret:
        ret = load_azure_ds_dir(ddir)
```

You may point out that this retains the previous behaviour, rather than the behaviour you're changing to. I'd suggest that we can address that by converting `list_possible_azure_ds_devs()` to a generator; very roughly:

```py
def list_possible_azure_ds_devs(...):
    yield DEFAULT_PROVISIONING_ISO_DEV
    for fstype in ("iso9660", "udf"):
        yield from util.find_devs_with("TYPE=%s" % fstype)
```

This will yield `DEFAULT_PROVISIONING_ISO_DEV` first, and only if a second value is requested will it execute `util.find_devs_with(...)`: if `DEFAULT_PROVISIONING_ISO_DEV` works, then we'll `break` out of the for loop before . (This would actually be an improvement even without the short-circuiting for this PR: it will only call `util_find_devs_with` for UDF if none of the ISO9660 options work.)

Additionally: I think that everything within the existing for loop between the exception handlers and `break` could move out of the for loop: it will be executed at most once anyway, due to the `break`: it doesn't need to go inside my proposed for loop here.

What do you think?

anhvoms: Thanks for the suggestion on switching list_possible_azure_ds_devs to a generator. I've made the changes

TheRealFalcon (approved):
Thanks @anhvoms , your changes here look good. I left a few nits inline, but nothing major.

There's a conflict now with master. Can you rebase the PR and then we should be able to get this merged soon.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -407,96 +407,92 @@ def crawl_metadata(self):
...
+        else:
+            for src in list_possible_azure_ds(self.seed_dir, ddir):
+                try:
+                    if src.startswith("/dev/"):
+                        if util.is_FreeBSD():
+                            ret = util.mount_cb(src, load_azure_ds_dir,
+                                                mtype="udf")
+                        else:
+                            ret = util.mount_cb(src, load_azure_ds_dir)
+                        self.iso_dev = metadata_source
```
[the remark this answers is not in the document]
anhvoms: yes - this is used in _report_ready to pass it to get_metadata_from_fabric (which will eventually call eject_iso before reporting ready)

PR #873, opened 2021-04-19
Ubuntu/devel

TheRealFalcon:
Should this be a hirsute branch now rather than ubuntu/devel?


OddBloke (approved):
Thanks, this both matches what I expect to see, and following the process locally renders the same output. I'll upload this shortly.

(As a reminder, I'll push to this branch once I've uploaded, rather than using GitHub's merge functionality.)

PR #874, opened 2021-04-19
Ubuntu/groovy

TheRealFalcon:


debian/changelog
```diff
@@ -1,3 +1,12 @@
+cloud-init (21.1-19-gbad84ad4-0ubuntu1~20.10.1ubuntu1) groovy; urgency=medium
```
OddBloke, line 1: This isn't correct; our tooling gets this wrong on cherry-pick, unfortunately. Our two choices are `21.1-19-gbad84ad4-0ubuntu1~20.10.2` or `21.1-19-gbad84ad4-0ubuntu2~20.10.1`. I would lean towards the latter, as "21.1-19-gbad84ad4-0ubuntu2" is the version in the devel release.

OddBloke (approved):
This LGTM, and matches what I reproduce locally; thanks!

PR #878, opened 2021-04-19
Azure: Retry getting network metadata during nic attach for all non-timeout errors

aswinrajamannar:
## Proposed Commit Message
Azure: Retry getting network metadata during nic attach for all non-timeout errors.

## Summary
When network interfaces are hot-attached to the VM, attempting to get network metadata might return 410 (or 500, 503 etc) because the info is not yet available. In those cases, we retry getting the metadata before giving up. The only case where we can move on to wait for more nic attach events is if the call times out despite retries, which means the interface is not likely a primary interface, and we should try for more nic attach events.

## Test Steps
* Deploy Pre-provisioned VM in Azure of type "savable" (internal to Azure)
* Reuse the VM with one or more network interfaces
* When repeating this multiple times, some of them might fail if the platform is not ready to serve the network metadata

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -16,6 +16,7 @@ from time import sleep
 from xml.dom import minidom
 import xml.etree.ElementTree as ET
 from enum import Enum
+import requests
 
 from cloudinit import dmi
 from cloudinit import log as logging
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Thanks!

PR #879, opened 2021-04-20
Add Vultr to default datasource_list.

smoser:
Vultr support was added under #827.  This enables it by default
in the debian package.


smoser:
@ddymko see above.

ddymko:
@smoser Thanks for letting me know

OddBloke (approved):
Thanks!

debian/po/templates.pot
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: cloud-init\n"
 "Report-Msgid-Bugs-To: cloud-init@packages.debian.org\n"
-"POT-Creation-Date: 2021-03-22 14:08-0500\n"
+"POT-Creation-Date: 2021-04-20 10:14-0400\n"
```
TheRealFalcon, line 11: Why do we change this?

smoser: I'm not really sure. I was following this commit.
e3e36b2870b70dc588c87ec492cf18a54f46395f

PR #880, opened 2021-04-20
cc_mounts: exclude netdev mounts from mount -a 

anhvoms:
## Proposed Commit Message
Exclude netdev mounts from mount -a 

```
mounts typically runs in 'init' phase, which is before 
network-online. In many distros, NFS mounts can't be 
mounted too early due to dependencies on rpc-statd, 
which typically runs after network-online. Excluding mounts
annotated with netdev in /etc/fstab from the mount -a
command will avoid boot issue caused by this cyclic dependency.
```

## Additional Context
https://bugzilla.redhat.com/show_bug.cgi?id=1858930

## Checklist:

### DIFF
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -534,7 +534,7 @@ def handle(_name, cfg, cloud, log, _args):
         need_mount_all = True
 
     if need_mount_all:
-        activate_cmds.append(["mount", "-a"])
+        activate_cmds.append(["mount", "-a", "-O", "no_netdev"])
         if uses_systemd:
             activate_cmds.append(["systemctl", "daemon-reload"])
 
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_mounts.py
```diff
@@ -534,7 +534,7 @@ def handle(_name, cfg, cloud, log, _args):
         need_mount_all = True
 
     if need_mount_all:
-        activate_cmds.append(["mount", "-a"])
+        activate_cmds.append(["mount", "-a", "-O", "no_netdev"])
```
raharper, line 537: Interesting option.  The open question is what ensures these nfs mounts get mounted before user-config runs which expects mounts to be up?

From the hip idea; add a drop-in config to cloud-config.service.conf.d which greps /etc/fstab for netdev and if found run mount -a -O _netdev to mount any of those that aren't mounted. 

This might not help any module which run after cc_mount but before cloud-config.service runs... namely:

- set_hostname
- update_hostname
- update_etc_hots
- resolv_conf
- ca-certs
- rsyslog
- users-groups
- ssh

I users/groups and ssh are the most likely hits (if someone is nfs mounting something over /home.

PR #881, opened 2021-04-21
Add Vultaire as contributor

Vultaire:
Add "Vultaire" as contributor

PR #882, opened 2021-04-22
Added support for importing keys via primary/security mirror clauses

Vultaire:
## Proposed Commit Message
Added support for importing keys via primary/security mirror clauses

```
Added support for importing keys via primary/security mirror clauses

Presently, mirror keys cannot be associated with primary/security
mirrors.  Unfortunately, this prevents use of Landscape-managed
package mirrors as the mirror key for the Landscape-hosted repository
cannot be provided.

This patch allows the same key-related fields usable on "sources"
entries to be used on the "primary" and "security" entries as well.

LP: #1925395
```

## Test Steps
Note: These steps require a running Landscape server.
* Following the [Landscape repository documentation](https://docs.ubuntu.com/landscape/en/repositories), set up a mirror key, create an ubuntu distribution with default pockets, and sync the pockets.  Run "landscape-api get-distributions" and grep for "apt" to find the mirror line to use.
* Via cloud-init userdata, add the mirror to the apt.packages and/or apt.security sections, including the public key via the key property, and use this to spawn a VM or container.
* Without the fix, the mirrors will be added but "apt update" will fail due to the repository's key not being recognized.  With the fix, it should work.

## Checklist:


Vultaire:
Please note: my initial commit to tools/.github-cla-signers (#881) is still pending.

TheRealFalcon (approved):
Thanks!

PR #888, opened 2021-05-04
Add platform restrictions to serial console password test

TheRealFalcon:
## Proposed Commit Message

```
Add platform restrictions to serial console password test
```

## Additional Context
Even on platforms where getting the console text is supported, not all platforms return the full console text. This test will either fail or be skipped on anything other than LXD containers.

### DIFF
--- a/tests/integration_tests/modules/test_set_password.py
+++ b/tests/integration_tests/modules/test_set_password.py
@@ -129,6 +129,7 @@ class Mixin:
         assert "dick:" not in cloud_init_output
         assert "harry:" not in cloud_init_output
 
+    @pytest.mark.lxd_container
     def test_random_passwords_emitted_to_serial_console(self, class_client):
         """We should emit passwords to the serial console. (LP: #1918303)"""
         try:


OddBloke:
I'm -0 on this: I suspect that this indicates a family of bugs in pycloudlib, rather than a fundamental lack of support in the underlying platforms. For example, the test in question starts passing on EC2 if I apply this to `pycloudlib`:

```diff
diff --git a/pycloudlib/ec2/instance.py b/pycloudlib/ec2/instance.py
index 4eec974..b7c5fae 100644
--- a/pycloudlib/ec2/instance.py
+++ b/pycloudlib/ec2/instance.py
@@ -109,11 +109,15 @@ class EC2Instance(BaseInstance):
             The console log or error message
 
         """
-        try:
-            # OutputBytes comes from platform._decode_console_output_as_bytes
-            response = self._instance.console_output()
-            return response['OutputBytes']
-        except KeyError:
+        start = time.time()
+        while time.time() < start + 180:
+            try:
+                # OutputBytes comes from platform._decode_console_output_as_bytes
+                response = self._instance.console_output()
+                return response['OutputBytes']
+            except KeyError:
+                time.sleep(10)
+        else:
             return 'No Console Output [%s]' % self._instance
 
     def delete(self, wait=True):
```

It looks like it can take 2-3 minutes for the console output to be populated in EC2, which is why we're seeing failures: pycloudlib "naively" assumes that we can grab it immediately.

That said, I appreciate from a pragmatic point of view that we don't necessarily have the time to fix those right now (hence -0): I wonder if we should introduce a timed skip like curtin has (skip under $conditions until $date, then fail instead) so we don't permanently lose track of this gap?

TheRealFalcon:
I think adding something like a 'wait' parameter to pycloudlib for this call is good, but even if we had this, I wouldn't want to run it on ec2 because I don't think it is worth the waiting. It'd be one thing if that was the only coverage we had, but since we have lxd coverage, I think an additional 2-3 minute wait is unreasonable.

I'm also against a timed skip. Failing tests because we don't have a process for addressing tech debt or as means to take very-low priority items and arbitrarily raise them up to high priority doesn't sound reasonable to me.

OddBloke:
On Tue, May 04, 2021 at 02:46:08PM -0700, James Falcon wrote:
> I think adding something like a 'wait' parameter to pycloudlib for
> this call is good, but even if we had this, I wouldn't want to run it
> on ec2 because I don't think it is worth the waiting. It'd be one
> thing if that was the only coverage we had, but since we have lxd
> coverage, I think an additional 2-3 minute wait is unreasonable.

I'm not sure I follow this: we don't run EC2 tests in CI, so the wait
will only affect either automation (where runtime is, largely,
immaterial) or folks who are explicitly running all tests against EC2.

There are a couple of reasons I can think of for the latter: they're
validating that a cloud-init release/SRU works in EC2 (that's us, we do
have broad coverage), or EC2 is their cloud platform of choice (existing
credentials/knowledge, a work-funded account, they want to develop
against their deployment target, etc.): for those people, they don't
have coverage on LXD.  Even if they're model contributors and run all
the integration tests they can before submission, they may still face a
failure in CI: this is avoidable.

(Note that I say _all_ tests earlier: a full test run on any
non-container cloud takes the best part of an hour already.  Making this
test take 2-3 minutes longer is roughly equivalent to adding 2-3 more
tests which require a new EC2 instance: we do this as a matter of
course, of course.  Anyone who _is_ time-sensitive when it comes to
integration tests is likely already running a subset anyway.)

All that said, I agree that slow tests are generally a Bad Thing for
contributors.  I wonder if we could address your concern head on and
introduce a "slow" tag (or "slow_on_ec2", or
`pytest.mark.slow(["ec2"])`, etc.): this would give an easy, generic
interface for folks who are iterating to exclude things which really
slow that down?

(As an aside, I wonder if serial console access is quite so slow on
non-t2.micro instances?)

TheRealFalcon:
I added https://github.com/canonical/pycloudlib/issues/139 for the pycloudlib piece.

I've thought about the slow mark before as well. It's probably something I'll add when the speed annoys me enough :stuck_out_tongue:  (it doesn't make sense to add here until we add the pycloudlib bit).

TheRealFalcon:
Closing this as we fixed https://github.com/canonical/pycloudlib/issues/139

PR #892, opened 2021-05-06
Ubuntu/devel

TheRealFalcon:
Have to manually readjust a ^M to be \r\n in debian/changelog (and now I'm wishing I had used a better commit message).

Also, is there a reason we use an annotated tag for upstream releases, but an unannotated one for Ubuntu releases, or is it unintentional?


OddBloke:
> Have to manually readjust a ^M to be \r\n in debian/changelog (and now I'm wishing I had used a better commit message).

`git` presents it correctly as a 4 character sequence ("\", "r", "\", "n") so it's something in our tooling which is mangling it, I think: we certainly _shouldn't_ have to avoid backslashes in commit messages.

> Also, is there a reason we use an annotated tag for upstream releases, but an unannotated one for Ubuntu releases, or is it unintentional?

The canonical source for cloud-init upstream releases is those tags, so we want a GPG signature so people can verify that the tag was intentionally created by a person they trust. The canonical source for cloud-init in Ubuntu is the archive: that has its own trust path already, of course, so the tagging in our repo is more informational than anything else. So: I don't know if it's intentional, but I don't think it's a particular issue. (FAOD, using an annotated tag wouldn't be a problem, it's just not _necessary_.)

debian/changelog
```diff
@@ -1,3 +1,49 @@
+cloud-init (21.2-0ubuntu1) hirsute; urgency=medium
```
OddBloke, line 1: Our tooling uses the last changelog release, so gets it wrong for the first run in a new release:

```suggestion
cloud-init (21.2-0ubuntu1) impish; urgency=medium
```

OddBloke:
I'm seeing a few test failures in my local package build:

```
=========================== short test summary info ============================
FAILED cloudinit/distros/tests/test_init.py::TestGetPackageMirrorInfo::test_valid_substitution[None-fk\tfake\t1-patterns21-expected21-True-ec2]
FAILED cloudinit/distros/tests/test_init.py::TestGetPackageMirrorInfo::test_valid_substitution[None-fk\nfake\n1-patterns22-expected22-True-ec2]
FAILED cloudinit/distros/tests/test_init.py::TestGetPackageMirrorInfo::test_valid_substitution[None-fk\rfake\r1-patterns25-expected25-True-ec2]
============ 3 failed, 2716 passed, 10 skipped, 1 warning in 33.01s ============
```

I can't reproduce in an impish daily container until I install packages from -proposed: at that point, it reproduces. My best from-the-hip guess is that this is a Python 3.9.4->3.9.5 difference, as that's available in -proposed ATM, but I haven't validated this.

OddBloke:
> I can't reproduce in an impish daily container until I install packages from -proposed: at that point, it reproduces. My best from-the-hip guess is that this is a Python 3.9.4->3.9.5 difference, as that's available in -proposed ATM, but I haven't validated this.

https://docs.python.org/release/3.9.5/whatsnew/changelog.html indicates that https://bugs.python.org/issue43882 was part of 3.9.5: "[security] urllib.parse should sanitize urls containing ASCII newline and tabs". So that looks like a pretty clear match with these failures!

TheRealFalcon:
Thanks for finding this! I put up #893 to address this.

TheRealFalcon:
Updated to `21.2-3-g899bfaa9`

OddBloke (approved):
Thanks!

PR #894, opened 2021-05-10
BSD: static network, set the mtu

goneri:
In the case of a static network, we now set the MTU according to the
meta-data.


TheRealFalcon (approved):
Thanks!

PR #897, opened 2021-05-11
Ubuntu/bionic

TheRealFalcon:


debian/changelog
```diff
@@ -1,3 +1,48 @@
+cloud-init (21.2-3-g899bfaa9-0ubuntu1~18.04.1) UNRELEASED; urgency=medium
```
OddBloke, line 1: s/UNRELEASED/bionic/

We also appear to have modified a bunch of older changelog entries (to strip trailing whitespace, presumably thanks to your editor): let's drop those as needless churn.

TheRealFalcon: Updated...and woe to those who save trailing whitespace! :wink:

PR #898, opened 2021-05-11
Ubuntu/hirsute

TheRealFalcon:
I wasn't prompted for the BUG ID and it expected to release as `21.2-3-g899bfaa9-0ubuntu1` rather than `21.2-3-g899bfaa9-0ubuntu1~21.04.1` (I manually updated both of these). Is this expected that we do this manually when doing the first SRU from what was the devel release, or have I missed a step somewhere in the process?


OddBloke (changes requested):
I ran `new-upstream-snapshot --first-sru upstream/master` (an option I think I re-discover ~every 6 months) and it produced a fairly different changelog to yours: different version, and stripped out bug references (because we only want to reference our process bug).

TheRealFalcon:
@OddBloke Ran script and got the updated changelog.

TheRealFalcon:
Also ran `debconf-updatepo`

OddBloke:
This matches what following our process created for me locally; performing some testing now.

PR #868, opened 2021-04-16
Fix chef module run failure when chef_license is set

bmhughes:
## Commit Message

Move `chef_license` from TPL_PATH_KEYS to TPL_KEYS as the chef license setting is not a path but must be added to the client config template. Fixes file or folder not found exception raised from `ensure_dirs`.

## Test Steps

Perform a cloud-init run with chef bootstrap configuration and chef_license set.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #901, opened 2021-05-12
Add integration test for #868

TheRealFalcon:
## Proposed Commit Message

```
Add integration test for #868

Ensure no Traceback when 'chef_license' is set
```

## Test Steps
pytest tests/integration_tests/bugs/test_gh868.py

## Checklist:

### DIFF
--- a/tests/integration_tests/bugs/test_gh868.py
+++ b/tests/integration_tests/bugs/test_gh868.py
@@ -1,0 +1,19 @@
+"""Ensure no Traceback when 'chef_license' is set"""
+import pytest
+from tests.integration_tests.instances import IntegrationInstance
+
+
+USERDATA = """\
+#cloud-config
+chef:
+  install_type: omnibus
+  chef_license: accept
+  server_url: https://chef.yourorg.invalid
+  validation_name: some-validator
+"""
+
+
+@pytest.mark.user_data(USERDATA)
+def test_chef_license(client: IntegrationInstance):
+    log = client.read_from_file('/var/log/cloud-init.log')
+    assert 'Traceback' not in log


tests/integration_tests/bugs/test_gh868.py
```diff
@@ -0,0 +1,19 @@
+"""Ensure no Traceback when 'chef_license' is set"""
+import pytest
+from tests.integration_tests.instances import IntegrationInstance
+
+
+USERDATA = """\
+#cloud-config
+chef:
+  install_type: omnibus
```
TheRealFalcon, line 9: Are there any concerns about this being in an integration test? This causes chef to install from `https://www.chef.io/chef/install.sh`

blackboxsw: I like this test because it tests a realworld application/use-case all the way to the source.

All good to have the remote script source. I don't mind this for two reasons:
 - security: not concerned as it's a short-lived instance that is torn down immediately after the test run
 - scripts on the internet changing which break our tests: let's deal/cope with that only if we find that this test is really flaky due to changes in the script.
 
Are there other concerns you had in mind?

TheRealFalcon: Just from a "good steward" point of view, but I can't image our jenkins runs would be more than a blip compared to the number of hits it receives.

Only reason I thought of it was because we have an older cloud test for chef that is disabled because it relies on a 3rd party repo. https://github.com/canonical/cloud-init/blob/master/tests/cloud_tests/testcases/examples/install_run_chef_recipes.yaml#L4

blackboxsw (approved):
+1 on this integration test.

tests/integration_tests/bugs/test_gh868.py
blackboxsw: I'm okay with us keeping this for now. That commit where we disabled was pointing to https://bugs.launchpad.net/cloud-init/+bug/1678145 which was basically a "don't let things on the internet break us when they change". If this ends up breaking repeatedly and causing maintenance problems, We can spend time to remedy/cache it or something else.

PR #902, opened 2021-05-13
New

stappersg:
New expire date in example doc
New contributor, hopes that the two separate commits stay two separate commits


TheRealFalcon (approved):
Thanks!

> New contributor, hopes that the two separate commits stay two separate commits 

Just FYI, due to changelog handling we squash all of our PRs. We know it's not ideal, but that's currently how things work.

PR #903, opened 2021-05-13
Fix unit tests breaking from new httpretty version

TheRealFalcon:
## Proposed Commit Message

```
Fix unit tests breaking from new httpretty version

Httpretty now logs all requests by default which gets mixed up with our
logging tests. Also we were incorrectly setting a logging level to
'None', which now also causes issues with the new httpretty version.
See https://github.com/gabrielfalcao/HTTPretty/pull/419
```

## Test Steps
Run pytest

## Checklist:

### DIFF
--- a/cloudinit/tests/test_url_helper.py
+++ b/cloudinit/tests/test_url_helper.py
@@ -8,6 +8,7 @@ from cloudinit import util
 from cloudinit import version
 
 import httpretty
+import logging
 import requests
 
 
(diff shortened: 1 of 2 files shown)


cloudinit/tests/test_url_helper.py
```diff
@@ -81,6 +82,9 @@ def test_read_file_or_url_str_from_url_redacting_headers_from_logs(self):
         url = 'http://hostname/path'
         headers = {'sensitive': 'sekret', 'server': 'blah'}
         httpretty.register_uri(httpretty.GET, url)
+        # By default, httpretty will log our request along with the header,
+        # so if we don't change this the secret will show up in the logs
+        logging.getLogger('httpretty.core').setLevel(logging.CRITICAL)
```
OddBloke, line 87: Do we want to set this even higher than CRITICAL, so we get no messages?

OddBloke (approved):
Thanks!

PR #904, opened 2021-05-17
add DragonFlyBSD support

goneri:
- Mostly based on FreeBSD, the main exception is that
  `find_devs_with_on_freebsd` does not work.
- Since we cannot get the CDROM or the partition labels,
  `find_devs_with_on_dragonflybsd()` has a more naive approach and
  returns all the block devices.
- Support for growpart (GPT only)
- Support for resizefs


goneri:
Note: you can try the DragonFly BSD 6.0.0 image from https://bsd-cloud-image.org/

cloudinit/distros/freebsd.py
```diff
@@ -28,6 +34,13 @@ class Distro(cloudinit.distros.bsd.BSD):
     pkg_cmd_update_prefix = ["pkg", "update"]
     pkg_cmd_upgrade_prefix = ["pkg", "upgrade"]
     prefer_fqdn = True  # See rc.conf(5) in FreeBSD
+    home_dir = '/usr/home'
+
+    def _select_hostname(self, hostname, fqdn):
```
TheRealFalcon, line 39: With the `prefer_fqdn` class attribute, we no longer need to override this method.

cloudinit/distros/freebsd.py
```diff
@@ -17,7 +17,13 @@
 LOG = logging.getLogger(__name__)
 
 
-class Distro(cloudinit.distros.bsd.BSD):
+class FreeBSD(cloudinit.distros.bsd.BSD):
```
TheRealFalcon, line 20: I realize `bsd.py` doesn't follow the trend of the class being named `Distro`, but can we leave that as the one exception and let this be named `Distro` here as well?

cloudinit/distros/freebsd.py
```diff
@@ -155,4 +171,9 @@ def update_package_sources(self):
             "update-sources", self.package_command,
             ["update"], freq=PER_INSTANCE)
 
+
+class Distro(FreeBSD):
```
TheRealFalcon, line 175: If we leave the other class named `Distro`, this I believe we can remove this class.

config/cloud.cfg.tmpl
```diff
@@ -60,11 +60,11 @@ cloud_init_modules:
 {% endif %}
  - bootcmd
  - write-files
-{% if variant not in ["netbsd"] %}
+{% if variant not in ["netbsd", "openbsd", "dragonfly"] %}
```
TheRealFalcon, line 63: Nit: Can we put "dragonfly" at the front of this list to keep it alphabetical?

config/cloud.cfg.tmpl
```diff
@@ -60,11 +60,11 @@ cloud_init_modules:
...
  - growpart
  - resizefs
 {% endif %}
-{% if variant not in ["freebsd", "netbsd"] %}
+{% if variant not in ["freebsd", "netbsd", "openbsd", "dragonfly"] %}
```
TheRealFalcon, line 67: Same comment about alphabetical here

cloudinit/distros/dragonflybsd.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2020-2021 Gonéri Le Bouder
+#
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import cloudinit.distros.freebsd
+
+
+class Distro(cloudinit.distros.freebsd.FreeBSD):
```
TheRealFalcon, line 8: If we follow my other comment, the base should be `cloudinit.distros.freebsd.Distro` then

TheRealFalcon (changes requested):
Hey @goneri , thanks for this addition. Overall, this looks very good. I left some comments inline along with some comments for additions and testing here.

Should we add dragonfly here?
https://github.com/canonical/cloud-init/blob/master/cloudinit/distros/__init__.py#L47
and here?
https://github.com/canonical/cloud-init/blob/master/tools/render-cloudcfg#L7

Additionally, we would like some unit testing to go along with this addition.

Can we get unit testing to make sure it is detected correctly? Similar to:
https://github.com/canonical/cloud-init/blob/master/cloudinit/tests/test_util.py#L464

Along with testing for the new `find_devs` implementation? Simlar to:
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_util.py#L890
or perhaps:
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_datasource/test_nocloud.py#L292

TheRealFalcon (commented):
Thanks for addressing my inline comments. Did you see my overall comments on the review though? It looks like they still need to be addressed. We still need a couple unit tests, and there might be a place or two you would want to add the distro.

Previous comment:
```
Should we add dragonfly here?
https://github.com/canonical/cloud-init/blob/master/cloudinit/distros/__init__.py#L47
and here?
https://github.com/canonical/cloud-init/blob/master/tools/render-cloudcfg#L7

Additionally, we would like some unit testing to go along with this addition.

Can we get unit testing to make sure it is detected correctly? Similar to:
https://github.com/canonical/cloud-init/blob/master/cloudinit/tests/test_util.py#L464

Along with testing for the new find_devs implementation? Simlar to:
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_util.py#L890
or perhaps:
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_datasource/test_nocloud.py#L292
```

goneri:
Oh indeed, I totally missed the comment. I will work on that. Thank you for the review!

goneri:
> Should we add dragonfly here?
> https://github.com/canonical/cloud-init/blob/master/cloudinit/distros/__init__.py#L47

I see I also need to add OpenBSD and NetBSD. Can I do that in another iteration?

> and here?
> https://github.com/canonical/cloud-init/blob/master/tools/render-cloudcfg#L7

I use `freebsd` variant for now and `tools/build-on-freebsd` works fine.

I'm working on the unit-tests.

TheRealFalcon:
Hey @goneri , I noticed some new commits. Are you looking for a re-review yet?

goneri:
I've yet to add a test in test_nocloud.py. But yes, an early feedback is welcome.

goneri:
Regarding nocloud, I don't think we need any special unittest since 72f6eb0339f7fcf7b8b02be2e86e5f7477cf731c. They current code base relies on `cloudinit.util` which is tested.

TheRealFalcon (approved):

goneri:
Thank you @TheRealFalcon!

PR #906, opened 2021-05-19
Add Rocky Linux support to cloud-init

nazunalika:
## Proposed Commit Message

```
summary: Add support for Rocky Linux

Rocky Linux is a RHEL-compatible distribution so all changes that have
been made should be trivial.
```

## Additional Context
https://rockylinux.org

We currently have RC1 out (for version 8.3) and currently working on 8.4. We
had to redo our patches for 8.4 as it was rebased. We realized an upstream PR
was not made for our distribution, hence this submission.

Please let me know if I made a mistake anywhere throughout this PR and I
will try to correct it.

Thank you!

## Checklist:


TheRealFalcon (changes requested):
Thanks for this submission. Overall this looks good. A few notes...

Most of the lists you added `rocky` to were in alphabetical order. Instead of adding rocky to the end of the list, can you add it to keep the ordering alphabetical?

One of the unit tests you wrote is failing on CI. It's because of the parsing of `/etc/redhat-release` is expecting a codename after the version number. Would it be possible to update [the implementation](https://github.com/canonical/cloud-init/blob/master/cloudinit/util.py#L432) to make the codename optional?

I suggested a change inline for the other failing unit test.

nazunalika:
I believe I addressed the alphabetical order lists and also the code name. We added a code name recently but I didn't provide an updated set of info. Hopefully that addressed that issue there. I couldn't find your suggestion in the modified file list. Or perhaps I've looked in the wrong place.

tests/unittests/test_cli.py
```diff
@@ -224,7 +224,8 @@ def test_wb_devel_schema_subcommand_doc_content(self):
         self._call_main(['cloud-init', 'devel', 'schema', '--docs', 'all'])
         expected_doc_sections = [
             '**Supported distros:** all',
-            '**Supported distros:** almalinux, alpine, centos, debian, fedora',
+            ('**Supported distros:** almalinux, alpine, centos, debian, '
+             'fedora, rocky'),
```
TheRealFalcon, line 228: This is checking that the list in cc_ntp.py matches some documentation output. Unfrotunately it hasn't been updated in a while, but this should fix it.
```suggestion
            ('**Supported distros:** almalinux, alpine, centos, debian, '
             'fedora, opensuse, rhel, rocky, sles, ubuntu'),
```

TheRealFalcon (commented):
Hmmm, weird...must have put that suggestion in a different tab then I closed or something. Should be here now.

TheRealFalcon (approved):
Looks good now, thanks!

PR #909, opened 2021-05-25
Add support for VMware PhotonOS

sshedi:
Signed-off-by: Shreenidhi Shedi <yesshedi@gmail.com>

## Proposed Commit Message

```
Add support for VMware PhotonOS
```

## Checklist:


sshedi:
@smoser @TheRealFalcon @OddBloke - I have gotten this far with this PR. Need your help to solve the test failures.

cloudinit/config/cc_resolv_conf.py
```diff
@@ -48,14 +48,15 @@
 
 from cloudinit import log as logging
 from cloudinit.settings import PER_INSTANCE
+from cloudinit.subp import subp
```
TheRealFalcon, line 51:
```suggestion
from cloudinit import subp
```
Unfortunately some tests rely on being able to patch the subp module, so we have to import the entire module here, not just the function.

If you replace all further instances of `subp` in this module with `subp.subp`, then it should work as expected.

TheRealFalcon (commented):
I haven't done a full review because this PR is WIP, but as I was investigating the test issue, I noticed a couple things around networking. 

You shouldn't need to provide a separate networkd renderer, as that is already provided by the netplan renderer. Is there something about the netplan renderer that doesn't work for your case? Additionally, you don't need to support both the v1 and the v2 config. If the v2 config works for you, then go with that.

sshedi:
> I haven't done a full review because this PR is WIP, but as I was investigating the test issue, I noticed a couple things around networking.
> 
> You shouldn't need to provide a separate networkd renderer, as that is already provided by the netplan renderer. Is there something about the netplan renderer that doesn't work for your case? Additionally, you don't need to support both the v1 and the v2 config. If the v2 config works for you, then go with that.

@TheRealFalcon  thanks a lot for the input. My tests are passing now.

About adding a separate network renderer:

1. netplan doesn't work for us because we don't have netplan package in our system, ours is a minimalist system and we    add packages only when necessary.

 2. We consume read yaml data of network config, parse and create network file, if some extra config is needed we will add it based on requirement which may not be supported in upstream systemd/netplan. This approach will scale better for us.

3. I understand your idea of having V2 net config support but we support both V1 & V2 as there are consumers for both and it's kinda tricky situation. So good for us to support both versions.

Hope this clarifies at least some of your questions.

sshedi:
Hi @TheRealFalcon - Can you please review my changes?

TheRealFalcon:
@sshedi , yep, will do! Though probably won't get to it till sometime next week.

sshedi:
> @sshedi , yep, will do! Though probably won't get to it till sometime next week.

Hi @TheRealFalcon - if you find time, can you please review my changes?

config/cloud.cfg.tmpl
```diff
@@ -90,11 +120,15 @@ cloud_config_modules:
...
  - snap
 {% endif %}
+{% if variant not in ["photon"] %}
  - ssh-import-id
  - locale
+{% endif %}
  - set-passwords
-{% if variant in ["rhel", "fedora"] %}
+{% if variant in ["rhel", "fedora", "photon"] %}
+{% if variant not in ["photon"] %}
```
TheRealFalcon, line 129: Do you want spacewalk or no? Only one of these lines should be used.

config/cloud.cfg.tmpl
```diff
@@ -121,6 +155,10 @@ cloud_config_modules:
  - byobu
 {% endif %}
 
+{% if variant in ["photon"] %}
+unverified_modules: ['yum-add-repo', 'ntp']
```
TheRealFalcon, line 159: Since you added `photon` to the `distros` list in both of these modules, there should be no reason to add them as `unverified_modules`. This block should be deleted.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,387 @@
...
+            (_out, err) = subp.subp(cmd)
+            if len(err):
+                LOG.warning('Running %s resulted in stderr output: %s',
+                            cmd, err)
+            return True
+        except subp.subp.ProcessExecutionError:
+            util.logexc(LOG, 'Running network bringup command %s failed', cmd)
+            return False
+
+    def _select_hostname(self, hostname, fqdn):
```
TheRealFalcon, line 239: I know your logic here is slightly different, but if you set the class variable `prefer_fqdn = True`, this override shouldn't be needed.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,387 @@
...
+        # Prefer the the fully-qualified domain name over the short host name.
+        # Otherwise it is not possible to use cloud-init to set an FQDN for a
+        # host because of the way the host name is configured with cloud-init
+        # and the datasource used by Photon.
+        if not hostname or (isinstance(fqdn, str) and fqdn.find('.') > 0):
+            return fqdn
+        return hostname
+
+    def _write_hostname(self, your_hostname, out_fn):
```
TheRealFalcon, line 248: Since you seem to deferring to systemd as much as possible, it might be worth looking at the hostname method implementations in `rhel.py` as an alternative to what's here. It's just a suggestion though; that may not be relevant here.

TheRealFalcon (changes requested):
@sshedi We have a good start here. I'll wait to review the test files until we change some of the datasource networking things.

I left some comments inline. In addition to those:

cc_resolv_conf.py:33 : Let's add photon to 'Supported distros' line

I noticed you're setting locale both in the distro and also using cc_resolv_conf.py. Is this intentional? cc_resolv_conf.py isn't run by default. Are the changes to this file needed? If not, ignore my inline comments in that file.

For each new file that has a VMWare copyright line, can you also add a line underneath that says
```
# This file is part of cloud-init.  See LICENSE file for license information.
```

cloudinit/config/cc_resolv_conf.py
```diff
@@ -104,12 +109,24 @@ def handle(name, cfg, cloud, log, _args):
...
+        resolv_conf = "/etc/systemd/resolved.conf"
+
+    template_fn = cloud.get_template_filename(resolv_conf)
     if not template_fn:
-        log.warning("No template found, not rendering /etc/resolv.conf")
+        log.warning("No template found, not rendering " + resolv_conf)
         return
 
-    generate_resolv_conf(template_fn=template_fn, params=cfg["resolv_conf"])
+    if distro_name == "photon":
```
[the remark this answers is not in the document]
sshedi: I partially understood your point here, what you are suggesting is correct but I'm not sure about achieving it via `distros/__init__.py`.

Do I need to have a function in `photon.py` which will return the `resolv_conf_fn` ? But then if user enables `- resolv_conf` in `cloud.cfg` how will `config/cc_resolv_conf.py` will consume it?

cloudinit/config/cc_resolv_conf.py
```diff
@@ -50,12 +50,13 @@
...
 from cloudinit import templater
 from cloudinit import util
+from cloudinit import subp
 
 LOG = logging.getLogger(__name__)
 
 frequency = PER_INSTANCE
 
-distros = ['alpine', 'fedora', 'opensuse', 'rhel', 'sles']
+distros = ['alpine', 'fedora', 'opensuse', 'photon', 'rhel', 'sles']
```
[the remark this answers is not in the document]
sshedi: I don't want it to run on every instance, only upon user request.

cloudinit/distros/photon.py
sshedi: Good suggestion, I'll try this.

config/cloud.cfg.tmpl
sshedi: I don't want spacewalk but I want yum-add-repo.

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,166 @@
...
+from cloudinit import util
+from cloudinit import subp
+from distutils.util import strtobool
+from cloudinit import log as logging
+
+
+LOG = logging.getLogger(__name__)
+
+
+class Renderer(renderer.Renderer):
```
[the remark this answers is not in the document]
sshedi: I have a problem with netplan renderer.

```
#!/usr/bin/env python3

from cloudinit.net import netplan, network_state
from cloudinit import distros, safeyaml

net_data = """\
network:
  version: 1
  config:
    - type: physical
      name: eth0
      mac_address: 00:50:56:ae:e3:4b
      subnets:
        - control: auto
          type: dhcp
    - type: nameserver
      address: []
      search:
        - test.com
"""

pre_ns = safeyaml.load(net_data)

if 'network' in pre_ns:
    pre_ns = pre_ns.get('network')
ns = network_state.parse_net_config_data(pre_ns)

print(netplan.network_state_to_netplan(ns))
```
Try the above code, the search domain `test.com` won't show up in netplan output and I need it in my case.

https://github.com/canonical/cloud-init/commit/d29eeccd2c422b8eb3b053fc13ca966ed6d74c78

The above commit changed the way of handling dns.

My thought here is, once netplan rectifies it, I will remove `v1tov2.py` or if you have any work around for this issue please let me know. I need a way to add dns domains to each input interface given by user.

sshedi:
> @sshedi We have a good start here. I'll wait to review the test files until we change some of the datasource networking things.
> 
> I left some comments inline. In addition to those:
> 
> cc_resolv_conf.py:33 : Let's add photon to 'Supported distros' line
> 
> I noticed you're setting locale both in the distro and also using cc_resolv_conf.py. Is this intentional? cc_resolv_conf.py isn't run by default. Are the changes to this file needed? If not, ignore my inline comments in that file.
> 
> For each new file that has a VMWare copyright line, can you also add a line underneath that says
> 
> ```
> # This file is part of cloud-init.  See LICENSE file for license information.
> ```

Thanks for the review comments. I have added the license info to all new files.
I didn't understand your point `setting locale both in the distro and also using cc_resolv_conf.py.`

TheRealFalcon:
> I didn't understand your point setting locale both in the distro and also using cc_resolv_conf.py.

Sorry, I mean resolve_conf, not locale. In your distro, [you have code to set the resolve_conf](https://github.com/canonical/cloud-init/pull/909/files#diff-73becea852e038c81ccc7126698b334c86caa696d07d654f9dfbc0fc9b3e8ca9R165-R183). If the distro is handling that, why do we also need changes to `cc_resolv_conf.py`? Unless there's a specific reason to use it, it can just go unsupported in your distro.

cloudinit/config/cc_resolv_conf.py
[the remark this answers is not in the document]
TheRealFalcon: It could be consumed with something along the lines of `cloud.distro.generate_resolv_conf()`. That would need to be defined in `distros/__init__.py` and then overridden in your distro.

sshedi:
> > I didn't understand your point setting locale both in the distro and also using cc_resolv_conf.py.
> 
> Sorry, I mean resolve_conf, not locale. In your distro, [you have code to set the resolve_conf](https://github.com/canonical/cloud-init/pull/909/files#diff-73becea852e038c81ccc7126698b334c86caa696d07d654f9dfbc0fc9b3e8ca9R165-R183). If the distro is handling that, why do we also need changes to `cc_resolv_conf.py`? Unless there's a specific reason to use it, it can just go unsupported in your distro.

Yes but that's inside `_write_network` function which will never be called, but I will think about it. I will try adding it in networkd.py

cloudinit/distros/__init__.py
```diff
@@ -572,6 +573,18 @@ def create_user(self, name, **kwargs):
...
+        # cloud-init supports 'passwd' option only for non-existing user
+        # To maintain compatibility, this piece of code is added
+        if util.get_linux_distro()[0] == "photon":
+            user_exists = True if util.is_user(name) else False
+            # Set hashed_passwd to passwd if hashed_passwd is not present
+            if (user_exists and 'passwd' in kwargs and kwargs['passwd'] and
+                    'hashed_passwd' not in kwargs):
+                kwargs['hashed_passwd'] = kwargs['passwd']
+                kwargs.pop('passwd')
+
```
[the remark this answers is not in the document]
sshedi: I believe this is same as overriding resolv_conf like you mentioned in above comment?

cloudinit/net/networkd.py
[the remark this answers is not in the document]
TheRealFalcon: Removing v1tov2.py doesn't require using netplan. It requires parsing the network data differently. The problem is that your networkd renderer is [parsing v2 config](https://github.com/canonical/cloud-init/pull/909/files?file-filters%5B%5D=.md&file-filters%5B%5D=.py#diff-27a9d06fdd06260be708d1436848e6d1597ce6d286f0cb411c2555062f997942R151), when instead it should parsing the internal representation. The use of `network_state.config` is wrong unless you're using v2 config and want to write netplan, which is not what you're trying to do.

If I modify your example:
```python
#!/usr/bin/env python3

from cloudinit.net import netplan, network_state
from cloudinit import distros, safeyaml

net_data = """\
network:
  version: 1
  config:
    - type: physical
      name: eth0
      mac_address: 00:50:56:ae:e3:4b
      subnets:
        - control: auto
          type: dhcp
    - type: nameserver
      address: []
      search:
        - test.com
"""

pre_ns = safeyaml.load(net_data)

if 'network' in pre_ns:
    pre_ns = pre_ns.get('network')
ns = network_state.parse_net_config_data(pre_ns)
print(ns.version)
print(ns.dns_nameservers)
print(ns.dns_searchdomains)
print(ns.has_default_route)
for iface in ns.iter_interfaces():
    print(iface)
for route in ns.iter_routes():
    print(route)
```

and then run it, here is the output:
```
1
[]
['test.com']
False
{'name': 'eth0', 'type': 'physical', 'mac_address': '00:50:56:ae:e3:4b', 'inet': 'inet', 'mode': 'manual', 'mtu': None, 'address': None, 'gateway': None, 'subnets': [{'control': 'auto', 'type': 'dhcp', 'routes': []}], 'accept-ra': None, 'wakeonlan': None}
```

Using the internal representation, you don't have to worry about whether the user a v1 config or a v2 config, both get converted to the same thing and that's what should be used by the renderer.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,380 @@
...
+                            cmd, err)
+        except subp.ProcessExecutionError:
+            util.logexc(LOG, 'Restart of localed using command %s failed', cmd)
+
+    def install_packages(self, pkglist):
+        # self.update_package_sources()
+        self.package_command('install', pkgs=pkglist)
+
+    def _write_network_config(self, netconfig):
+        if isinstance(netconfig, str):
```
sshedi, line 92: We have some internal dependency on this function, down the line I will make it right. As of now, let it be.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,380 @@
...
+                        resolv_conf)
+            return
+
+        generator_fptr(template_fn=template_fn,
+                      params=cfg['resolv_conf'],
+                      target_fname=resolv_conf)
+
+        subp.subp(['systemctl', 'restart', 'systemd-resolved'])
+
+    def _write_network_(self, settings):
```
sshedi, line 178: Calling this from _write_network_config now

TheRealFalcon:
```diff
diff --git a/cloudinit/cmd/devel/net_convert.py b/cloudinit/cmd/devel/net_convert.py
index 0668ffa3..5c649fd0 100755
--- a/cloudinit/cmd/devel/net_convert.py
+++ b/cloudinit/cmd/devel/net_convert.py
@@ -11,7 +11,7 @@ from cloudinit.sources import DataSourceAzure as azure
 from cloudinit.sources import DataSourceOVF as ovf
 
 from cloudinit import distros, safeyaml
-from cloudinit.net import eni, netplan, network_state, sysconfig
+from cloudinit.net import eni, netplan, networkd, network_state, sysconfig
 from cloudinit import log
 
 NAME = 'net-convert'
@@ -51,7 +51,7 @@ def get_parser(parser=None):
     parser.add_argument("--debug", action='store_true',
                         help='enable debug logging to stderr.')
     parser.add_argument("-O", "--output-kind",
-                        choices=['eni', 'netplan', 'sysconfig'],
+                        choices=['eni', 'netplan', 'networkd', 'sysconfig'],
                         required=True,
                         help="The network config format to emit")
     return parser
@@ -118,9 +118,14 @@ def handle_args(name, args):
         config['netplan_path'] = config['netplan_path'][1:]
         # enable some netplan features
         config['features'] = ['dhcp-use-domains', 'ipv6-mtu']
-    else:
+    elif args.output_kind == "networkd":
+        r_cls = networkd.Renderer
+        config = distro.renderer_configs.get('networkd')
+    elif args.output_kind == "sysconfig":
         r_cls = sysconfig.Renderer
         config = distro.renderer_configs.get('sysconfig')
+    else:
+        raise RuntimeError("Invalid output_kind")
 
     r = r_cls(config=config)
     sys.stderr.write(''.join([

```

It'd be helpful to apply a patch like this so I can more easily work/play with the networkd renderer. Even with this change, there will still be a few issues to fix before it works (I'll address those inline), but here is an example of how you could run it:
`cloud-init devel net-convert -p /path/to/config.yaml -k yaml -d /tmp -D photon -O networkd --debug`

config/cloud.cfg.tmpl
```diff
@@ -38,6 +48,24 @@ preserve_hostname: false
 # This should not be required, but leave it in place until the real cause of
 # not finding -any- datasources is resolved.
 datasource_list: ['NoCloud', 'ConfigDrive', 'Azure', 'OpenStack', 'Ec2']
+{% elif variant in ["photon"] %}
```
TheRealFalcon, line 51: the datasource_list above was added to work around a specific bug. Unless you have datasources not being detected, I think you can get rid of this section.

cloudinit/distros/photon.py
TheRealFalcon: Ok, I'll hold off from reviewing the networking stuff in this file for now. Let me know when it's the way you want it.

cloudinit/config/cc_resolv_conf.py
```diff
@@ -104,12 +104,16 @@ def handle(name, cfg, cloud, log, _args):
...
+    try:
+        cloud.distro.generate_resolv_conf(cfg, cloud, generate_resolv_conf)
+    except Exception:
+        template_fn = cloud.get_template_filename('resolv.conf')
+        if not template_fn:
+            log.warning("No template found, not rendering /etc/resolv.conf")
+            return
 
-    generate_resolv_conf(template_fn=template_fn, params=cfg["resolv_conf"])
+        generate_resolv_conf(template_fn=template_fn,
```
TheRealFalcon, line 115: Instead of this try/except, it would be nice if we could instead move the functionality of `generate_resolv_conf` into `distros/__init__.py` so we get behavior that looks more traditionally polymorphic.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,380 @@
...
+        if self.uses_systemd():
+            if not out_fn:
+                out_fn = self.systemd_locale_conf_fn
+        else:
+            if not out_fn:
+                out_fn = self.locale_conf_fn
+        locale_cfg = {
+            'LANG': locale,
+        }
+        rhutil.update_sysconfig_file(out_fn, locale_cfg)
```
TheRealFalcon, line 72: Is this correct? Are you using sysconfig?

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def create_network_file(self, link, conf):
+        LOG.debug('Setting Networking Config for %s', link)
+
+        net_fn = self.network_fname_fptr(self.network_conf_dir, link)
+        util.write_file(net_fn, conf)
+        util.chownbyname(net_fn, 'systemd-network', 'systemd-network')
+
+    def render_network_state(self, network_state, templates=None, target=None):
+        fpnplan = os.path.join(subp.target_path(target), self.network_conf_dir)
```
TheRealFalcon, line 179: This needs to be used when creating the network file. This is especially important when using the `net-convert` command.

E.g., if I run:
`cloud-init devel net-convert -p /tmp/test.yaml -k yaml -d /tmp -D ubuntu -O networkd --debug`

I'm saying with the `-d tmp` that  I want the renderer to write the files to `/tmp/etc/systemd/network`. As it is now, it'll write to `/etc/systemd/network`, which isn't what we want.

TheRealFalcon (commented):
Thanks, you've made some good changes. I haven't been able to fully review everything yet, but this should be plenty for now, and the changes will make it easier for me to try some things.

cloudinit/distros/photon.py
sshedi: Sure, once this PR gets merged I will send incremental patches. I have touched too many files already, so I don't want to disrupt this at the moment.

cloudinit/distros/photon.py
sshedi: Yes, it's correct. I haven't enabled locate in cloud.cfg, we need glibc-i18n for this to work and we don't ship it by default. It just happens to be same as rhel way and I just need that utility function to update locale info.

However, I will remove the condition check for systemd, ours is systemd by default.

.travis.yml
```diff
@@ -125,7 +125,7 @@ matrix:
         - python: 3.5
           env:
               TOXENV=xenial
-              PYTEST_ADDOPTS=-v  # List all tests run by pytest
+              PYTEST_ADDOPTS=-sv  # List all tests run by pytest
```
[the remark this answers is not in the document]
sshedi: I just added this to see why xenial tests are failing and not in my setup. I got my answer. I wasn't planning to push it as it is. :)

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,380 @@
...
+from cloudinit.settings import PER_INSTANCE
+from cloudinit.distros import rhel_util as rhutil
+from cloudinit.net.network_state import mask_to_net_prefix
+from cloudinit.distros.parsers.hostname import HostnameConf
+
+LOG = logging.getLogger(__name__)
+
+def network_file_name(dirname, dev_name):
+    network_file_pattern = '*' + dev_name + '.network*'
+    for node in os.listdir(dirname):
```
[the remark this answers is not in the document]
sshedi: good point :)

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,380 @@
...
+        if fnmatch.fnmatch(node, network_file_pattern):
+            node_fullpath = os.path.join(dirname, node)
+            util.del_file(node_fullpath)
+            return dirname + node
+
+    return dirname + '10-' + dev_name + '.network'
```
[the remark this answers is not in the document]
sshedi: We use `10-<iface-name>.network` format and we have another way of customizing vm instances using shell scripts which creates network files with same naming convention. 10-* takes precedence over 50-* so, I don't want to change it.

config/cloud.cfg.tmpl
sshedi: We need this, `VMwareGuestInfo` this is not available by default and we want to go in the order given in the list.

If you insist, I will remove this.

TheRealFalcon: It is ok to keep (though I'd remove the comments). It'd be nice to get to the point where we don't need it though.

sshedi: I understand. The reason why I have kept the commented items is, we run some tests on azure, gce, aws and our test scripts enables desired entry. I need to fix things in multiple places, it's okay. I will remove the commented ones. Thanks.

sshedi:
I have no idea why this is failing https://travis-ci.com/github/canonical/cloud-init/jobs/513191790
```
E             Full diff:

E             - [call(<ANY>, "# Your system has been configured with 'manage-resolv-conf' set to true.\n# As a result, cloud-init has written this file with configuration data\n# that it has been provided. Cloud-init, by default, will write this file\n# a single time (PER_ONCE).\n#\n\noptions foo\n", mode=<ANY>)]

E             ?       ^^^ ^^^^^^^^^^                                                                                                                                                                                                                                                                               ^^^^^

E             + [call('/etc/resolv.conf', "# Your system has been configured with 'manage-resolv-conf' set to true.\n# As a result, cloud-init has written this file with configuration data\n# that it has been provided. Cloud-init, by default, will write this file\n# a single time (PER_ONCE).\n#\n\noptions foo \n", mode=420)]

E             ?       ^^^^^^^^^^^^^^^^^^^^^^^ ^^^        
```

The latter has a an extra space at the end of `foo`, once again this doesn't happen in my setup. Only on xenial test.

TheRealFalcon:
Hey @sshedi , I took the liberty of pushing a commit to this branch for the `cc_resolv_conf.py`. Does it work for you? If not, we can always undo it.

I also removed the
```python
        if self.uses_systemd():
            subp.subp(['systemctl', 'restart', 'systemd-resolved'])
```

The standard systemd-resolved.service runs after systemd-networkd.service, which runs after cloud-init-local.service, which is when this `cc_resolv_conf.py` is running. Since the config is being generated before systemd-resolved is started, we shouldn't attempt to restart it here.

sshedi:
No problem, thank you. My idea was giving a bit more freedom in distro specific function by adding new function in distros/__init__.py. Ultimately resolve conf will be processed in same way. I am okay with this approach and it works for me. Also, good call on restarting resolved service along with networkd.

I have fixed the `cloudinit/config/tests/test_resolv_conf.py` script and resolved the merge conflicts.

cloudinit/net/network_state.py
```diff
@@ -303,6 +304,10 @@ def parse_config_v1(self, skip_broken=True):
                 ) from e
             try:
                 handler(self, command)
+                if command_type == 'physical':
```
TheRealFalcon, line 307: Unfortunately, this isn't going to work. The v1 config is unorderered so a nameserver can't correspond directly to a specific interface. This also fails to parse if we have a config like this:
```
network:
  version: 1
  config:
    - type: nameserver
      address:
        - 192.168.23.2
        - 8.8.8.8
      search:
        - exemplary
    - type: physical
      name: eth0
      mac_address: '00:11:22:33:44:55'
      subnets:
         - type: static
           address: 192.168.23.14/27
           gateway: 192.168.23.1
```

I wouldn't be opposed to expanding the v1 config to add an 'interface' key to the 'nameserver' type, but I think we'd be better served doing that in a separate PR.

sshedi: I understand, I was being optimistic about the order of configurations. What can we do about in this PR?

TheRealFalcon: I see 3 options:
1. Create and wait for the other PR to land and rebase it into this one
2. Apply the globally defined dns entries to every interface (this is what netplan does)
3. Disallow the networkd renderer from using v1 config until we can make the change

sshedi: I have decided to use option 2 for the time being. I will start working on adding new option net config v1.

cloudinit/net/network_state.py
```diff
@@ -706,16 +717,18 @@ def handle_wifis(self, command):
...
-            dns = cfg.get('nameservers').get('addresses', [])
-            name_cmd = {'type': 'nameserver'}
-            if len(search) > 0:
-                name_cmd.update({'search': search})
-            if len(dns) > 0:
-                name_cmd.update({'addresses': dns})
-            LOG.debug('v2(nameserver) -> v1(nameserver):\n%s', name_cmd)
-            self.handle_nameserver(name_cmd)
+        for iface, dev_cfg in cfg.items():
+            if 'nameservers' in cfg:
```
TheRealFalcon, line 721:
```suggestion
            if 'nameservers' in dev_cfg:
```

cloudinit/net/network_state.py
```diff
@@ -706,16 +717,18 @@ def handle_wifis(self, command):
...
+                search = dev_cfg.get('nameservers').get('search', [])
+                dns = dev_cfg.get('nameservers').get('addresses', [])
+                name_cmd = {'type': 'nameserver'}
+                if len(search) > 0:
+                    name_cmd.update({'search': search})
+                if len(dns) > 0:
+                    name_cmd.update({'address': dns})
+                LOG.debug('v2(nameserver) -> v1(nameserver):\n%s', name_cmd)
```
TheRealFalcon, line 729: I think the `v2->v1` bit here is a little confusing. We're really just going v2 parsed into the proper `NetworkState`. I would either update it to just `nameserver: %s` or remove the log entirely.

cloudinit/distros/photon.py
```diff
@@ -0,0 +1,354 @@
...
+        LOG.debug('Attempting to run: %s', cmd)
+        try:
+            (out, err) = subp.subp(cmd, capture=capture)
+            if err:
+                LOG.warning('Running %s resulted in stderr output: %s',
+                            cmd, err)
+            return True, out, err
+        except subp.ProcessExecutionError:
+            util.logexc(LOG, 'Command %s failed', cmd)
+            return False, out, err
```
TheRealFalcon, line 57: `out` and `err` won't exist here if the subp command raises the exception

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,239 @@
...
+            LOG.warning('Target file not given')
+            return
+
+        contents = self.get_final_conf()
+        LOG.debug('Final content: %s', contents)
+        util.write_file(target_fn, contents)
+
+
+class Renderer(renderer.Renderer):
+    ''' Renders network information in /etc/systemd/network '''
```
TheRealFalcon, line 60: Can we add a comment or docstring here specifying that this Renderer is currently experimental and doesn't support all the use cases supported by the other renderers yet.

TheRealFalcon (commented):
@sshedi , thanks for all of the updates. I think we're getting really close here. I left a few more comments inline, but these should be quicker to fix.

Additionally, I'm hoping we can get a few more tests in `test_net.py` for the networkd renderer. In particular, if you look at the [NETWORK_CONFIGS](https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_net.py#L822) definition, you'll see a number of different keys, such as [small](https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_net.py#L823) or [v4_and_v6](https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_net.py#L940). For each key, there's a yaml definition, along with `expected_*` definitions for the expected output from each respective renderer. I realize that at this point not every definition in that list will be supported with the current networkd renderer, but it would be nice to have the tests for the ones that do work. Then further below are defined tests that ensure the rendering works as expected.

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,229 @@
...
+                        cfg.update_section('Route', 'Gateway', v)
+                    elif k == 'dns_nameservers':
+                        cfg.update_section('Network', 'DNS', ' '.join(v))
+                    elif k == 'dns_search':
+                        cfg.update_section('Network', 'Domains', ' '.join(v))
+
+        cfg.update_section('Network', 'DHCP', dhcp)
+
+    def dhcp_domain(self, d, cfg):
+        for item in ['dhcp4domain', 'dhcp6domain']:
```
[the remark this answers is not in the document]
sshedi: Yes, Photon has a newer version of systemd which supports [DHCPv4] & [DHCPv6]
For that I'm using this. If it's okay, let it be. We have some users who use this feature although not officially supported by network config netplan config was giving this info and we used it. It will be a lot of trouble explaining people why we can't support it anymore.

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,229 @@
...
+        if 'search' in dns and dns['search']:
+            cfg.update_section(sec, 'Domains', ' '.join(dns['search']))
+
+        if 'nameservers' in dns and dns['nameservers']:
+            cfg.update_section(sec, 'DNS', ' '.join(dns['nameservers']))
+
+    def create_network_file(self, link, conf, nwk_dir):
+        LOG.debug('Setting Networking Config for %s', link)
+
+        net_fn = nwk_dir + '10-' + link + '.network'
```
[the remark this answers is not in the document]
sshedi: Okay, this works for me. I can use `10-cloud-init-<link>.network`, will fix it.

cloudinit/net/networkd.py
```diff
@@ -0,0 +1,229 @@
...
+def available(target=None):
+    expected = ['systemctl']
+    search = ['/usr/bin', '/bin']
+    for p in expected:
+        if not subp.which(p, search=search, target=target):
+            return False
+    return True
+
+
+def network_state_to_networkd(ns):
```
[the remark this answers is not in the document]
sshedi: This is kinda useful when I want to trial and error with some sample net config yaml files.
I can just import networkd and call this function with network_state data. But if you insist, no problem. I will remove it.

cloudinit/net/network_state.py
sshedi: Oops, sorry.

cloudinit/net/networkd.py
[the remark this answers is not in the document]
TheRealFalcon: So you're saying you're using our v2/netplan config, but you've added extra `dhcp4domain` and `dhcp6domain` keys to the config? Where did these keys come from? How is it different from the `dhcp4` and `dhcp6` keys?

It's not ideal. Can you add a comment to the method saying something like "This is to accommodate extra keys present in VMWare config"?

cloudinit/net/networkd.py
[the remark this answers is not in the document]
TheRealFalcon: If you're using it it's fine to stay.

TheRealFalcon:
I put up a PR for the DNS changes here: https://github.com/canonical/cloud-init/pull/923 . I borrowed a little from this PR if that's ok. If that PR lands, you shouldn't need any changes to `network_state.py` in this PR.

cloudinit/net/networkd.py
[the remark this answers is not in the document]
sshedi: Done

sshedi:
Hi James,

No problem. I have pushed my latest changes and I have added some more tests in test_net.py as you suggested.
I have kept my network_state.py changes, if this lands first please re-base #923<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fcanonical%2Fcloud-init%2Fpull%2F923&data=04%7C01%7Csshedi%40vmware.com%7C2a21fab482be40dc89e008d931123767%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637594776585724939%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=smcQlNPhupY2xYG%2FlwZtxWIc2R0n9ULLugdI4dYKYfg%3D&reserved=0>. If #923 lands first, I will rebase my changes.

I think you need to add some documentation about per interface dns in doc/rtd/topics/network-config-format-v1.rst

--
Shedi
________________________________

TheRealFalcon:
Hi @sshedi , thanks for those updates. 

#923 just landed. It did include a doc change to the v1 format. After this is rebased, I should be able to do one final review tomorrow and hopefully get this one done!

TheRealFalcon (approved):
@sshedi , I think this looks good now.

Thanks for working with me to get everything done needed to get this across the finish line!

sshedi:
> @sshedi , I think this looks good now.
> 
> Thanks for working with me to get everything done needed to get this across the finish line!

Pleasure was all mine. Thanks James for all your insightful comments and being patient with me all this while and making me understand things better. Enjoyed working with you.

Will be sending incremental fixes and improvements from now on. Have a good time ahead, thanks again.

PR #912, opened 2021-05-27
Docs: Replace all freenode references with libera (SC-68)

TheRealFalcon:
Libera has no webchat, so removed one link, and replaced another with a link to the main website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ If you need support, start with the [user documentation](https://cloudinit.readt
 
 If you need additional help consider reaching out with one of the following options:
 
-- Ask a question in the [``#cloud-init`` IRC channel on Freenode](https://webchat.freenode.net/?channel=#cloud-init)
+- Ask a question in the [``#cloud-init`` IRC channel on Libera](https://libera.chat/)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ If you need support, start with the [user documentation](https://cloudinit.readt
 
 If you need additional help consider reaching out with one of the following options:
 
-- Ask a question in the [``#cloud-init`` IRC channel on Freenode](https://webchat.freenode.net/?channel=#cloud-init)
+- Ask a question in the [``#cloud-init`` IRC channel on Libera](https://libera.chat/)
```
mitechie, line 29: is there an equiv webui direct link? Or direct link that might trigger an xdg-open of the irc client for a user? If not all good, but if so seems like it might be a more equiv. update.

mitechie (approved):

README.md
TheRealFalcon: Checked last night and couldn't find anything, but just checked again and updated it.

mitechie: :+1:

PR #914, opened 2021-06-04
init: azure observe excluded_drivers when copying a distro in init

blackboxsw:
In Stages, the init.distro property will reconstitute the discovered
distro when considering the fully merged system config.

Since distro class now carries a networking instance, cloudinit needs
to ensure any datasource.excluded_drivers are preserved in the newly
fetched distro.networking.blacklist_drivers.

This allows any datasource (currently just AzureDataSource) to
prescribe the types of network devices which will be ignored when
attempting to render full network on all viable network devices
via a call to distros.networking.Network.get_interfaces_by_mac()

LP: #1927124

## Proposed Commit Message

# WIP so I can write a better integration test for this.
```

```

## Test Steps
* launch azure xenial vm with advanced networking
* obtain UA token from https://ubuntu.com/advantage
* ua attach <token>
* ua enable fips
* sudo cloud-init clean --reboot --logs
* Expect system to never come back 

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -344,6 +344,7 @@ class DataSourceAzure(sources.DataSource):
         EventType.BOOT,
         EventType.BOOT_LEGACY
     }}
+    excluded_drivers = BLACKLIST_DRIVERS  # Drivers excluded from network cfg
 
     _negotiated = False
     _metadata_imds = sources.UNSET
@@ -626,7 +627,7 @@ class DataSourceAzure(sources.DataSource):
         except Exception as e:
             LOG.warning("Failed to get system information: %s", e)
 
-        self.distro.networking.blacklist_drivers = BLACKLIST_DRIVERS
+        self.distro.networking.blacklist_drivers = self.excluded_drivers
 
         try:
             crawled_data = util.log_time(
--- a/cloudinit/tests/test_stages.py
+++ b/cloudinit/tests/test_stages.py
@@ -21,8 +21,10 @@ TEST_INSTANCE_ID = 'i-testing'
 class FakeDataSource(sources.DataSource):
 
     def __init__(self, paths=None, userdata=None, vendordata=None,
-                 network_config=''):
+                 network_config='', excluded_drivers=None):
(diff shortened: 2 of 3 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -344,6 +344,7 @@ class DataSourceAzure(sources.DataSource):
         EventType.BOOT,
         EventType.BOOT_LEGACY
     }}
+    excluded_drivers = BLACKLIST_DRIVERS  # Drivers excluded from network cfg
```
blackboxsw, line 347: The PR after this will address non-functional renames of "BLACKLIST_DRIVERS" to "excluded_drivers" throughout cloudinit. I didn't want to address this renaming in this branch as it would be much broader than this azure-specific change (and will touch distros/networking, cloudinit.net.__init__.py docs and potentially some config modules.

cloudinit/tests/test_stages.py
```diff
@@ -54,6 +56,24 @@ def setUp(self):
...
+        self.init._distro = "cacheddistro"
+        self.assertEqual("cacheddistro", self.init.distro)
+
+    def test_distro_honors_datasource_exclude_drivers(self):
+        """When datasource.exclude_drivers exists copy to distro.networking."""
+        self.assertFalse(hasattr(self.init.datasource, "exclude_drivers"))
+        distro = self.init.distro
+        self.assertIsNone(getattr(distro.networking, "blacklist_drivers"))
+        self.init._distro = None  # clears _distro cache
+        self.init.datasource.exclude_drivers = ["baddriver1", "baddriver2"]
```
TheRealFalcon, line 73: Don't we want a test with `excluded_drivers` to show we set it correctly?

TheRealFalcon (approved):
I think this is good.

I'm wondering if long-term it would make sense to make `excluded_drivers` an attribute on all classes, not just Azure, then perhaps some logic "somewhere" in the base class that does the current `self.distro.networking.blacklist_drivers = self.excluded_drivers` that Azure currently does.

If we suspect this will never happen on other datasources, then that might be overkill.

Do you need help with the integration test?

PR #915, opened 2021-06-05
tune azure ds telemetry annotation

haitch:
## Proposed Commit Message
```
summary: tune azure ds telemetry annotation

remove unnecessary azure telemetry, to avoid **cloud-init analyze** generate multiple boot record for a single boot. 

the reason is generate_records in /cloudinit/analyze/show.py have a bug, if a DS event is start and this DS has been seen, it will generate a new boot record.

        if event_type(event) == 'start':
            if event.get('name') in stages_seen:

LP: #1930956 (replace with the appropriate bug reference or remove
```

## Additional Context
bug filed at: https://bugs.launchpad.net/cloud-init/+bug/1930956
this PR is not real fix for ```cloud-init analyze```,  it's a easy workaround we can take.

## Test Steps
manually tested

## Checklist:

### DIFF
--- a/cloudinit/analyze/show.py
+++ b/cloudinit/analyze/show.py
@@ -134,6 +134,7 @@ class SystemctlReader(object):
     '''
     Class for dealing with all systemctl subp calls in a consistent manner.
     '''
+
     def __init__(self, property, parameter=None):
         self.epoch = None
         self.args = ['/bin/systemctl', 'show']
@@ -310,9 +311,9 @@ def generate_records(events, blame_sort=False,
 
     unprocessed = []
     for e in range(0, len(sorted_events)):
-        event = events[e]
+        event = sorted_events[e]
         try:
-            next_evt = events[e + 1]
+            next_evt = sorted_events[e + 1]
         except IndexError:
             next_evt = None
 
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -2197,7 +2197,6 @@ def _generate_network_config_from_imds_metadata(imds_metadata) -> dict:
     return netconfig
 
 
-@azure_ds_telemetry_reporter
 def _generate_network_config_from_fallback_config() -> dict:
     """Generate fallback network config excluding blacklisted devices.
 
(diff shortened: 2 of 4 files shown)


blackboxsw (changes requested):
Thanks for the PR @haitch. I see you are not listed in tools/.github-cla-signers. Please sign the CLA if you haven't already as documented in https://cloudinit.readthedocs.io/en/latest/topics/hacking.html#submitting-your-first-pull-request   and add your username to tools/.github-cla-signers.

blackboxsw (changes requested):
@haitch for the PR and trying to address the issues with cloud-init analyze.

I think we have a couple of things we need to sort though:
  1. We can just drop azure_ds_telemetry_reporter because Azure backplane (their cloud) does track and report these metrics. Your motivation in this branch appears to be to avoid triggering the bug in `cloud-init analyze show`. Let's instead fix the root of the problem on cloud-init/analyze/show.py
  2.  The actual issue here seems to be that cloud-init/analyze/show.py doesn't handle repeated parent events well by assuming that a [repeated parent event means a new boot right here](https://github.com/canonical/cloud-init/blob/master/cloudinit/analyze/show.py#L320-L322)
  3. Instead of dropping azure telemetry, let's add a `CLOUDINIT_BOOT_STAGES = ["init-local", "init", "init-network", "modules-config", "modules-final"] ` to cloudinit/analyze/show.py and make sure we only append to stages_seen local var `if event.get('name') in CLOUDINIT_BOOT_STAGES`
  4. We need a unit test to cloudinit/analyze/tests/test_show.py which exercises logic faking some of these repeated events that aren't in CLOUDINIT_BOOT_STAGES to assert that analyze doesn't inject  "boot" record for repeated events.

 


I did confirm (as you discovered), if we simply dropped any of the  log lines containing  "(start|finish).*_generate_network_config_from_fallback_config"  and "(start|finish).*get_last_log_byte_pushed_to_kvp_index" then analyze show works and shows a single boot record properly.

This is only due to the fact that we no longer have a nested pair of events
start:<something>
start:<something_nested>
finish:<something_nested>
finish:<something>

PR #919, opened 2021-06-14
Add new network activators to bring up interfaces (SC-85)

TheRealFalcon:
I kind of hate the name "configurer", but couldn't think of anything better. I'd prefer the term "network manager" but can't use it for obvious reasons. I'm open to suggestions.

## Proposed Commit Message

```
Add new network activators to bring up interfaces

Currently _bring_up_interfaces() is a no-op for any distro using
renderers. We need to be able to support bringing up a single
interfaces, a list of interfaces, and all interfaces. This should be
independent of the renderers, as the network config is often
generated independent of the mechanism used to apply it.

Additionally, I included a refactor to remove
"_supported_write_network_config". We had a confusing call chain of
apply_network_config->_write_network_config->_supported_write_network_config.
The last two have been combined.
```

## Additional Context
This was motivated by needing to support hotplug.

`_bring_up_interfaces()` is only ever called by arch and gentoo these days, so I removed the other implementations.

## Test Steps
I added some unit tests. Integration testing is difficult. Given that `_bring_up_interfaces` was previously a no-op, it's hard to imagine real world situations where we can test this. The only time we exercise this code path is when we're running in network mode, and we were unable to obtain DataSource information during the local phase.

One way to simulate this would be to create a testing DataSource that returns a different instance id every time it is queried. It would be an ugly hack, but it would work. Given that we'll need to test this when hotplug lands, I'm not sure it's worth the integration testing effort at the moment.

## Checklist:


cloudinit/net/network_state.py
```diff
@@ -58,38 +58,6 @@
                'bridge_waitport': None}}
 
 
-def parse_net_config_data(net_config, skip_broken=True):
```
TheRealFalcon, line 61: I moved `parse_net_config_data` to the bottom of the file so I could return some type information (`NetworkConfig` isn't defined here yet).

I removed `parse_net_config` as it is dead code.

cloudinit/net/renderers.py
```diff
@@ -38,13 +43,13 @@ def search(priority=None, target=None, first=False):
         if render_mod.available(target):
             cur = (name, render_mod.Renderer)
             if first:
-                return cur
+                return [cur]
```
TheRealFalcon, line 46: This is much easier to reason about if we're always returning a list. The only place this function is being called is from the `select` function below (which I modified accordingly) and a couple unit tests.

cloudinit/net/network_state.py
```diff
@@ -1052,4 +1020,31 @@ def mask_and_ipv4_to_bcast_addr(mask, ip):
...
+        # v2 does not have explicit 'config' key so we
+        # pass the whole net-config as-is
+        config = net_config
+
+    if version and config is not None:
+        nsi = NetworkStateInterpreter(version=version, config=config)
+        nsi.parse_config(skip_broken=skip_broken)
+        state = nsi.get_network_state()
+
+    if not state:
```
TheRealFalcon, line 1041: Previously, we were either returning `NetworkState` or `None`. In `net_convert.py` we handled the `None` accordingly, but the renderers don't check for `None` and we'll likely get an uglier error if they're passed it. So failing early here makes more sense and makes the function easier to reason about.

cloudinit/net/network_state.py
```diff
@@ -1052,4 +1020,31 @@ def mask_and_ipv4_to_bcast_addr(mask, ip):
...
+        raise RuntimeError(
```
TheRealFalcon, line 1042: Previously, we were either returning `NetworkState` or `None`. In `net_convert.py` we handled the `None` accordingly, but the renderers don't check for `None` and we'll likely get an ugly `AttributeError` error if they're passed it. So failing early here makes more sense and makes the function easier to reason about.

blackboxsw:
Yeah I'm struggling a bit with calling it a Configurer, maybe NetworkBackpane? What stinks is that netplan already calls -NetworkManager and systemd-networkd renderers... yet cloud-init also calls netplan a renderer. "NetworkBackplaneManager/ BackplaneBringup?    I'll hold on the rest of the bike-shedding and just review the code and we can discuss more broadly what this class should be called when we figure our what exactly it needs to do.

TheRealFalcon:
What about InterfaceBringerUpper? NetworkInterfaceInterfacer?
(these aren't serious suggestions just in case you were wondering :wink: )

cloudinit/distros/arch.py
```diff
@@ -62,9 +62,9 @@ def install_packages(self, pkglist):
         self.update_package_sources()
         self.package_command('', pkgs=pkglist)
 
-    def _write_network_config(self, netconfig):
+    def _write_network_state(self, network_state):
```
blackboxsw, line 65: I think we should also  drop "_bring_up_interfaces" from arch as it as a near equivalent (yet less functional) of the parent class `Distro._bring_up_interfaces`

blackboxsw (changes requested):
@TheRealFalcon, the code and refactor looks good I think the only critical thing at this point is solving the gentoo.py issues which I **think** lead to a break in network rendering.

The refactors and rename made it a bit time-consuming to review and understand for me, but I think I've walked through the setup path in general and the end result is a much cleaner set of method/function names.

See if any of my alternative bike-shed suggestions to "Configurer" ring as more appropriate/clear.... (I'm not sure they do).

blackboxsw:
I was able to go through inplace upgrades on existing systems greenfield and brownfield runs without error so, as expected, we haven't broken any the networking paths in this refactor.

cloudinit/distros/__init__.py
```diff
@@ -391,20 +395,11 @@ def preferred_ntp_clients(self):
...
-            (_out, err) = subp.subp(cmd)
-            if len(err):
-                LOG.warning("Running %s resulted in stderr output: %s",
-                            cmd, err)
-            return True
-        except subp.ProcessExecutionError:
-            util.logexc(LOG, "Running interface command %s failed", cmd)
-            return False
+        """Deprecated. Remove if/when arch and gentoo support renderers."""
+        raise NotImplementedError
```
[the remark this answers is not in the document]
TheRealFalcon: It doesn't actually. The call hierarchy would work like this:
base:apply_network -> gentoo:_bring_up_interfaces->base:_bring_up_interfaces->gentoo:_bring_up_interface

The `distros.Distro._bring_up_interfaces` is confusing (it should just call super), but since it's passing self to the base class, the "normal" method calling rules still apply.

cloudinit/net/configurers.py
```diff
@@ -0,0 +1,48 @@
...
+            "Unknown configurers provided in priority list: %s" % unknown)
+
+    found = []
+    for configurer in priority:
+        if configurer.available(target):
+            found.append(configurer)
+    return found
+
+
+def select_configurer(priority=None, target=None) -> Type[NetworkConfigurer]:
```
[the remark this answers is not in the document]
TheRealFalcon: Yeah...I named it this way because I was frustrated when grepping for usages of the `renderers.select|search`. Searching for 'select' or 'search' will return a million results. I figured the function call would be slightly more explicit too.

That said, I'm not super strongly opinionated about it. If you are, I don't mind changing it.

TheRealFalcon:
@blackboxsw , I think I've addressed (or responded to) each of your comments. The 'configurers-activators' was a fairly large change, so it might be easier to review the other commits separately.

blackboxsw (approved):
LGTM! Thanks for working through this @TheRealFalcon

cloudinit/distros/__init__.py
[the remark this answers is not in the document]
blackboxsw: D'oh right.

blackboxsw:
One CI error is likely due to mock not operating as expected in travis build env:

E       AssertionError: assert [<class 'clou...anActivator'>] == [<class 'clou...anActivator'>]
3097E         At index 0 diff: <class 'cloudinit.net.activators.NetplanActivator'> != <class 'cloudinit.net.activators.NetworkManagerActivator'>
3098E         Right contains one more item: <class 'cloudinit.net.activators.NetplanActivator'>
3099E         Full diff:
3100E           [
3101E         -  <class 'cloudinit.net.activators.NetworkManagerActivator'>,
3102E            <class 'cloudinit.net.activators.NetplanActivator'>,
3103E           ]
3104
3105tests/unittests/test_net_activators.py:110: AssertionError


loca

blackboxsw (changes requested):
fix mocked test to add available_mocks. then +1

blackboxsw (approved):
LGTM!

PR #922, opened 2021-06-15
Replace deprecated collections.Iterable with abc replacement (SC-127)

TheRealFalcon:
## Proposed Commit Message

```
Replace deprecated collections.Iterable with abc replacement

LP: #1932048
```

https://docs.python.org/3.9/library/collections.html

I did a quick search for any other references to something in `collections` that should be coming from `collections.abc` and didn't find anything.

## Test Steps
n/a

## Checklist:

### DIFF
--- a/cloudinit/log.py
+++ b/cloudinit/log.py
@@ -8,7 +8,7 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import collections
+import collections.abc
 import io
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
Approve only python3.5 support needed for `collections.abc`. No need to support earlier per https://lists.launchpad.net/cloud-init/msg00336.html

PR #923, opened 2021-06-16
Fix DNS in NetworkState (SC-133)

TheRealFalcon:
## Proposed Commit Message

```
Fix DNS in NetworkState

v1 network config currently has no concept of interface-specific DNS,
which is required for certain renderers. To fix this, added an
optional 'interface' key on the v1 nameserver definition. If
specified, it makes the DNS settings specific to the interface.
Otherwise, it will be defined as global DNS as it always has.

Additionally, DNS for v2 wasn't being recognized correctly. For DNS
defined on a particular interface, these settings now also go into the
global DNS settings as they were intended.

```

## Additional Context
These changes would help enable a systemd-networkd renderer.

## Test Steps
Added new unit tests. This can't be integration tested yet as a renderer would need to pick up these changes.

## Checklist:

### DIFF
--- a/cloudinit/net/network_state.py
+++ b/cloudinit/net/network_state.py
@@ -237,6 +237,7 @@ class NetworkStateInterpreter(metaclass=CommandHandlerMeta):
         self._network_state = copy.deepcopy(self.initial_network_state)
         self._network_state['config'] = config
         self._parsed = False
+        self._interface_dns_map = {}
 
     @property
     def network_state(self):
@@ -310,6 +311,21 @@ class NetworkStateInterpreter(metaclass=CommandHandlerMeta):
                     LOG.warning("Skipping invalid command: %s", command,
                                 exc_info=True)
                     LOG.debug(self.dump_network_state())
+        for interface, dns in self._interface_dns_map.items():
+            iface = None
+            try:
+                iface = self._network_state['interfaces'][interface]
+            except KeyError as e:
+                raise ValueError(
+                    'Nameserver specified for interface {0}, '
+                    'but interface {0} does not exist!'.format(interface)
+                ) from e
+            if iface:
+                nameservers, search = dns
+                iface['dns'] = {
+                    'addresses': nameservers,
+                    'search': search,
+                }
 
     def parse_config_v2(self, skip_broken=True):
         for command_type, command in self._config.items():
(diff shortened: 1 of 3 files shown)


cloudinit/net/network_state.py
```diff
@@ -526,21 +542,40 @@ def handle_bridge(self, command):
...
+        if 'interface' in command:
+            self._interface_dns_map[command['interface']] = (
+                nameservers, search
+            )
+        else:
+            dns['nameservers'].extend(nameservers)
+            dns['search'].extend(search)
+
+    @ensure_command_keys(['address'])
+    def handle_individual_nameserver(self, command, iface):
```
blackboxsw, line 575: We might want to rename this because of ugly magic of CommandHandlerMeta which will attempt to map any 'handle_' method of NetworkStateInterpreter as a viable "command"
Note we've done the same with _handle_bond_bridge to avoid "registering" and handler for the top-level network config key "bond_bridge"

```suggestion
    def _handle_individual_nameserver(self, command, iface):
```

cloudinit/net/tests/test_network_state.py
```diff
@@ -2,12 +2,62 @@
...
 
 netstate_path = 'cloudinit.net.network_state'
 
 
+_V1_CONFIG_NAMESERVERS = """\
+network:
+  version: 1
+  config:
+    - type: nameserver
+      interface: {}
```
blackboxsw, line 19: Can we specifically name this format variable or give it an index so it's more clear that we will probably be using this as a format string. Initial glance at the string made me think we were expecting a dict type in tests.
```suggestion
      interface: {0}     # or {int_name}
```

blackboxsw (changes requested):
Thanks @TheRealFalcon minor rename needed for `_handle_individual_nameserver` I think but beyond that +1. 

Maybe worth expanding the doc example to use the interface key?
diff --git a/doc/rtd/topics/network-config-format-v1.rst b/doc/rtd/topics/network-config-format-v1.rst
index d379b437..71d5f000 100644
--- a/doc/rtd/topics/network-config-format-v1.rst
+++ b/doc/rtd/topics/network-config-format-v1.rst
@@ -356,6 +356,7 @@ the following keys:
         address:
           - 192.168.23.2
           - 8.8.8.8
+        interface: interface0  # Ties nameserver to interface0 only
         search:
           - exemplary

blackboxsw (approved):
LGTM!

PR #928, opened 2021-06-18
d/cloud-init.templates: remove comment (unsupported by debconf)

paride:
- Revert "d/cloud-init.templates: comment about why Oracle DS is absent"
- Update d/changelog (UNRELEASED)

## Proposed Commit Message

If it's OK not to squash PRs to packaging branches then maybe this can be merged non-squashed. Otherwise the commit message can be
```
d/cloud-init.templates: remove comment (unsupported by debconf)
```

## Additional Context

Spotted by our CI.

## Test Steps

When installing a deb package with the comment this happens:
```
debconf: unable to initialize frontend: Dialog
debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
debconf: falling back to frontend: Readline
Template parse error near `# Oracle DS absent to retain orig behavior: Oracle(Bionic) == OpenStack detected', in stanza #1 of /var/lib/dpkg/info/cloud-init.templates
dpkg: error processing package cloud-init (--configure):
installed cloud-init package post-installation script subprocess returned error exit status 255
```
The error goes away removing the comment.

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -3,7 +3,6 @@ cloud-init (21.2-3-g899bfaa9-0ubuntu2~18.04.2) UNRELEASED; urgency=medium
   * d/patches/ubuntu-advantage-revert-tip.patch: drop revert patch
     + ubuntu-advantage-tools completed SRU to bionic. Bionic now
       compatible with upstream ua python-client CLI behavior.
-  * d/cloud-init.templates: Add comment for Oracle DS ommission on Bionic
 
  -- Chad Smith <chad.smith@canonical.com>  Tue, 15 Jun 2021 10:37:40 -0600
 
--- a/debian/cloud-init.templates
+++ b/debian/cloud-init.templates
@@ -1,4 +1,3 @@
-# Oracle DS absent to retain orig behavior: Oracle(Bionic) == OpenStack detected
 Template: cloud-init/datasources
 Type: multiselect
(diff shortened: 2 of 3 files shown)


TheRealFalcon:
Looks good to me, but I want to make sure @blackboxsw sees this first. Any idea of a way we can communicate this info so we don't come back to this in 6 months and wonder why Oracle isn't in this list?

paride:
I had a look, `debconf-devel(7)` doesn't mention comments. The actual parsing of the template is done by `/usr/share/perl5/Debconf/Template.pm`, I skipped through it doesn't seem to handle comments. I thought about adding a `Comment: ` field, but the parser would issue a warning:
```
warn sprintf(gettext("Unknown template field '%s', in stanza #%s of %s\n"), $field, $., $file);
```
I think the *right* place for this comment would be the `debian/README.Debian` file, but we're going to forget to look there I fear. What about adding a `debian/cloud-init.templates.README` file, with a note for us? That's very visible and will be ignored  by all the tools.

paride:
README added. I didn't add back the note to `d/changelog` as this change won't affect the package in any way.

blackboxsw (approved):

blackboxsw:
Confirmed no breakage on pkg install. TIL thanks @paride

PR #929, opened 2021-06-22
Removed distro specific network code from Photon

sshedi:
Minor fixes in networkd renderer & fixed corresponding tests

## Proposed Commit Message
Removed distro specific network code from Photon

```
Minor fixes in networkd renderer & fixed corresponding tests
```

## Checklist:


sshedi:
Hi @TheRealFalcon - like we discussed during 1st PR, I have removed the `_write_network_` and related functions.
Also fixed few issues in networkd. Please review whenever you get time.

cloudinit/net/networkd.py
```diff
@@ -35,6 +36,7 @@ def update_section(self, sec, key, val):
         for k in self.conf_dict.keys():
             if k == sec:
                 self.conf_dict[k].append(key+'='+str(val))
+                self.conf_dict[k] = list(dict.fromkeys(self.conf_dict[k]))
```
TheRealFalcon, line 39: This line appears to be converting a list to a dict with no values, then converting the keys of the dict back into a list. What does this accomplish?

cloudinit/net/networkd.py
```diff
@@ -103,19 +105,24 @@ def generate_link_section(self, iface, cfg):
...
-                    tmp += '/' + str(conf['prefix'])
-                cfg.update_section(sec, 'Destination', tmp)
-            elif k == 'metric':
-                cfg.update_section(sec, 'Metric', v)
+            if k == 'netmask' or k == 'prefix':
+                continue
+            if k == 'network' and 'prefix' in conf:
+                v += '/' + str(conf['prefix'])
+            elif k == 'network' and 'netmask' in conf:
+                v += '/' + network_state.mask_to_net_prefix(conf['netmask'])
```
TheRealFalcon, line 120: I don't think this will work. `mask_to_net_prefix` should return an int, so adding a string to an int like this should result int a TypeError. Did you miss an `str` call?

Can you also add/update a unit test that will take this path?

TheRealFalcon (changes requested):
Thanks for the updates! I've left a few comments inline.

cloudinit/net/networkd.py
sshedi: This removes duplicates from a list

cloudinit/net/networkd.py
sshedi: Yes, this is not looking good. I will rework on this, thanks.

cloudinit/net/networkd.py
TheRealFalcon: Ah, right. I'm used to using sets for that. Can you add a comment specifying it removes duplicates? Also, a duplicate in one of the units tests would be helpful. Currently removing this line still allows all tests to pass.

cloudinit/net/networkd.py
sshedi: Cleaned this up and added a test for handling duplicates

TheRealFalcon (approved):
Thanks!

PR #930, opened 2021-06-27
bsd/net_v1 format: read MTU from interface config

goneri:
## Proposed Commit Message

```
bsd/net_v1 format: read MTU from interface config

We read the MTU from the subnet entries. With the v1 format, the MTU can
be set at the root level of the interface entry in the `config` section.

Limitation, we won't set the MTU if the interface use DHCP. This
would require a bit of refactoring.

Also simplify/clarify how we pass the target variable in `cloudinit.net.bsd`.

See: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=256309
Reported-by: Andrey Fesenko
```

## Checklist:

 - [ x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [ x] I have updated or added any unit tests accordingly
 - [ x] I have updated or added any documentation accordingly

### DIFF
--- a/tests/unittests/test_net_freebsd.py
+++ b/tests/unittests/test_net_freebsd.py
@@ -1,8 +1,24 @@
-from cloudinit import net
+import os
+import yaml
 
-from cloudinit.tests.helpers import (CiTestCase, mock, readResource)
+import cloudinit.net
+import cloudinit.net.network_state
+from cloudinit.tests.helpers import (CiTestCase, mock, readResource, dir2dict)
+
 
 SAMPLE_FREEBSD_IFCONFIG_OUT = readResource("netinfo/freebsd-ifconfig-output")
+V1 = """
+config:
+-   id: eno1
+    mac_address: 08:94:ef:51:ae:e0
+    mtu: 1470
+    name: eno1
+    subnets:
+    -   address: 172.20.80.129/25
+        type: static
+    type: physical
+version: 1
+"""
 
 
 class TestInterfacesByMac(CiTestCase):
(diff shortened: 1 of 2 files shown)


goneri:
@f0andrey, can you give this a try?

f-andrey:
I'm check this patch FreeBSD guest with NoCloud and SmartOS datasources. Work fine.

Thanks.

tests/unittests/test_net_freebsd.py
```diff
@@ -1,8 +1,27 @@
-from cloudinit import net
+import cloudinit.net
+import cloudinit.net.network_state
+
+
+from cloudinit.tests.helpers import (CiTestCase, mock, readResource, dir2dict)
+
+
+import os
+import yaml
```
TheRealFalcon, line 9: Nit: from the second bullet of the [style guide](https://www.python.org/dev/peps/pep-0008/#imports)

```suggestion
import os
import yaml

import cloudinit.net
import cloudinit.net.network_state
from cloudinit.tests.helpers import (CiTestCase, mock, readResource, dir2dict)
```

tests/unittests/test_net_freebsd.py
```diff
@@ -12,8 +31,51 @@ class TestInterfacesByMac(CiTestCase):
...
+        ns = cloudinit.net.network_state.parse_net_config_data(network_config)
+        files = self._render_and_read(state=ns)
+        assert files == {
+            '/etc/resolv.conf': '# dummy resolv.conf\n',
+            '/etc/rc.conf': (
+                "# dummy rc.conf\n"
+                "ifconfig_eno1="
+                "'172.20.80.129 netmask 255.255.255.128 mtu 1470'\n")}
+        expected = []
+        mock_subp.assert_has_calls(expected)
```
TheRealFalcon, line 81: If `expected` is an empty list, this isn't really checking anything. Can we remove these last two lines, or did you intend to check that the proper service calls were made?

TheRealFalcon (changes requested):
Overall looks good. I just have some minor testing comments inline.

tests/unittests/test_net_freebsd.py
goneri: My bad, this is something that I should have removed.

TheRealFalcon (approved):
Thanks!

PR #933, opened 2021-06-30
Remove legacy eth0 from default interfaces file

dvandok:
## Proposed Commit Message
Remove legacy eth0 from default interfaces file

```
summary: default dhcp config blocks static IPv6 configuration

The Debian cloud images have dhcp by default for eth0 in
/etc/network/interfaces. The cloud-init script will write additional
config to /etc/network/interfaces.d/50-cloud-init, but the static
config for eth0 for IPv6 will be ignored as long as the dhcp config
exists.

The function to remove the legacy config points to another file
by default, but it should also be called for /etc/network/interfaces.
```

## Test Steps
An updated Debian cloud image should be created with the fix,
and then the behaviour can be tested. I have observed this
on an Openstack cluster.

## Checklist:

### DIFF
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -113,6 +113,7 @@ class Distro(distros.Distro):
 
     def _write_network_config(self, netconfig):
         _maybe_remove_legacy_eth0()
+        _maybe_remove_legacy_eth0("/etc/network/interfaces")
         return self._supported_write_network_config(netconfig)
 
     def _bring_up_interfaces(self, device_names):


TheRealFalcon:
This code will delete entirely `/etc/network/interfaces` if it contains exactly the following content (minus comments):
```
auto eth0
iface eth0 inet dhcp
```

If the file contains anything else besides comments, it won't touch the file.

That's not going to work.

dvandok:
I misunderstood how the function `_maybe_remove_legacy_eth0` works. What I wanted to accomplish is **deleting** these exact lines from the file.
There is also some config for eth1 and eth2 in the file, just in case people want to add more interfaces. This way to allow hot-plugging and handing out addresses via DHCP is convenient, but it would also block any static configuration of addresses for these interfaces. This needs a bit of a rethink about how this is best done.

dvandok:
I see this is going to be a little more complicated to do. One option is to overwrite /etc/network/interfaces completely with just the line to source everything from /etc/network/interfaces.d/.

dvandok:
I've filed a bug report for this: https://bugs.launchpad.net/cloud-init/+bug/1934303

PR #939, opened 2021-07-07
VMware: add a new switch allow_raw_data to control raw data feature

xiaofengw-vmware:
…update the documentation. Fix bugs about max_wait.

summary: Introduce a new switch allow_raw_data to control raw data feature, its default value is true. Also fix bugs about max_wait. Update the documentation accordingly.

## Test Steps
Set allow_raw_data to True and False in datasource config, trigger the guest customization on VMware vsphere and verify cloud-init could get the execpted data source.

## Checklist:

 - [ X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [ X] I have updated or added any unit tests accordingly
 - [ X] I have updated or added any documentation accordingly

### DIFF
--- a/cloudinit/sources/DataSourceOVF.py
+++ b/cloudinit/sources/DataSourceOVF.py
@@ -98,9 +98,20 @@ class DataSourceOVF(sources.DataSource):
             found.append(seed)
         elif system_type and 'vmware' in system_type.lower():
             LOG.debug("VMware Virtualization Platform found")
+            allow_vmware_cust = False
+            allow_raw_data = False
             if not self.vmware_customization_supported:
                 LOG.debug("Skipping the check for "
                           "VMware Customization support")
+            else:
+                allow_vmware_cust = not util.get_cfg_option_bool(
+                    self.sys_cfg, "disable_vmware_customization", True)
+                allow_raw_data = util.get_cfg_option_bool(
+                    self.ds_cfg, "allow_raw_data", True)
+
+            if not (allow_vmware_cust or allow_raw_data):
+                LOG.debug(
+                    "Customization for VMware platform is disabled.")
             else:
                 search_paths = (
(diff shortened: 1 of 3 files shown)


cloudinit/sources/DataSourceOVF.py
```diff
@@ -440,6 +453,8 @@ def get_max_wait_from_cfg(cfg):
 def wait_for_imc_cfg_file(filename, maxwait=180, naplen=5,
                           dirpath="/var/run/vmware-imc"):
     waited = 0
+    if maxwait <= naplen:
+        naplen = 1
```
TheRealFalcon, line 457: Nit: would `naplen = maxwait` be better here?

TheRealFalcon (approved):
One nit inline.

Also, it might be time to start breaking up `_get_data` into smaller methods. I don't think it's necessary for this PR, but going through the layers of nesting is getting hard to read.

xiaofengw-vmware:
> One nit inline.
> 
> Also, it might be time to start breaking up `_get_data` into smaller methods. I don't think it's necessary for this PR, but going through the layers of nesting is getting hard to read.

Thanks, I agree with you. It needs a full test and I prefer to do it in another pull request. I have created an internal PR to address this.

cloudinit/sources/DataSourceOVF.py
xiaofengw-vmware: If user set maxwait <= naplen,  it seems he is sensitive to this time delay, so check the cfg file at each seconds might be better. How do you think?

xiaofengw-vmware: @TheRealFalcon

TheRealFalcon: Sure, that makes sense.

PR #942, opened 2021-07-13
run-container: fixup the centos repos baseurls when using http_proxy

paride:
## Proposed Commit Message

```
run-container: fixup the centos repos baseurls when using http_proxy

The EPEL repo file used to have download.fedoraproject.org as its
baseurl. That has now been replaced by `download.example`, which we need
to replace with dl.fedoraproject.org, the actual mirror we want to
download from.

We can't use download.fedoraproject.org or the mirrorlist (which is the
default way for finding mirrors) because of our internal proxy rules.

This change only applies if http_proxy is set, otherwise the mirrors are
reached in the default way.

```

## Additional Context

Should fix our `cloud-init-copr-build` job.

## Test Steps

Follow what Jenkins does step by step:

https://github.com/canonical/server-jenkins-jobs/blob/main/jenkins-jobs/cloud-init/centos.yaml#L60

## Checklist:


blackboxsw (approved):
Works great locally now. Thanks for the fix @paride

paride:
Closing and reopening as Travis didn't pick this PR up for some reason.

PR #946, opened 2021-07-14
Update pylint and fix the new issues its spots

paride:
## Proposed Commit Message

```
Update pylint to v2.9.3 and fix the new issues its spots

In CI run against pylint 2.9.3 and fix occurrences of:
 - W0237 (arguments-renamed)
 - W0402 (deprecated-module)

The W0402 deprecated-module was about module `imp`:

    cloudinit/patcher.py:9: [W0402(deprecated-module), ]
        Uses of a deprecated module 'imp'

The imp module is deprecated and replaced by importlib, which according
to the documentation has no replacement for acquire_lock() and
release_lock(), which are the only reason why `imp` is imported.
    
I see nothing about the code using this lock that actually requires it.
Let's remove the locking code and the import altogether.

Dropping the locking makes patcher.patch() an empty wrapper around
_patch_logging(). Rename _patch_logging() to patch_logging() and
call it directly instead. Drop patch().
```

## Additional Context
Spotted by our `cloud-init-style-tip` Jenkins job.

This is the job always running against the latest versions of the linters. The job allows to keep the code up-to-date with the new versions without breaking CI, which uses pinned versions instead.

## Checklist:

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -147,7 +147,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         return uses_systemd()
 
     @abc.abstractmethod
-    def package_command(self, cmd, args=None, pkgs=None):
+    def package_command(self, command, args=None, pkgs=None):
         raise NotImplementedError()
 
     @abc.abstractmethod
@@ -253,7 +253,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def _write_hostname(self, hostname, filename):
+    def _write_hostname(self, hostname, out_fn):
         raise NotImplementedError()
 
     @abc.abstractmethod
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -253,7 +253,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def _write_hostname(self, hostname, out_fn):
+    def _write_hostname(self, hostname, filename):
         raise NotImplementedError()
 
     @abc.abstractmethod
--- a/cloudinit/patcher.py
+++ b/cloudinit/patcher.py
@@ -19,7 +19,7 @@ class QuietStreamHandler(logging.StreamHandler):
         pass
 
 
-def _patch_logging():
+def patch_logging():
     # Replace 'handleError' with one that will be more
     # tolerant of errors in that it can avoid
     # re-notifying on exceptions and when errors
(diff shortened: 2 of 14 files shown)


paride:
This is marked as a WIP because it doesn't fix this yet:
```
cloudinit/patcher.py:9: [W0402(deprecated-module), ] Uses of a deprecated module 'imp'
```
as I'm not really sure on what to do. The `imp` module is deprecated and replaced by `importlib`, which according to the documentation has no replacement for `acquire_lock()` and `release_lock()`. I'm not sure if those calls are still needed, or if we should silence the warning for now.

Other projects are facing this, e.g. https://github.com/pypa/setuptools/issues/479

paride:
CI failure is of course expected for now:
```
cloudinit/patcher.py:9: [W0402(deprecated-module), ] Uses of a deprecated module 'imp'
```

TheRealFalcon:
> The imp module is deprecated and replaced by importlib, which according to the documentation has no replacement for acquire_lock() and release_lock(). I'm not sure if those calls are still needed, or if we should silence the warning for now.

I see nothing about the code using this lock that actually requires it. There's not multiple threads in play and we're not creating import hooks. I'm comfortable removing the locking code altogether.

paride:
Thanks! I agree it doesn't really seem needed. I'll update the branch and drop those.

paride:
Branch updated and proposed commit message updated. I removed the WIP as this is now ready for review.

cloudinit/distros/__init__.py
```diff
@@ -253,7 +253,7 @@ def _read_hostname(self, filename, default=None):
         raise NotImplementedError()
 
     @abc.abstractmethod
-    def _write_hostname(self, hostname, filename):
+    def _write_hostname(self, hostname, out_fn):
```
TheRealFalcon, line 256: I'd prefer we go the other direction with this one. In all of my previous programming contexts, "fn" has meant "function" and usually refers to some kind of function pointer. I think "fn" meaning "filename" in our codebase is confusing, and in general I think full names are better and more readable than abbreviations.

cloudinit/patcher.py
```diff
@@ -39,10 +38,6 @@ def handleError(self, record):
 
 
 def patch():
-    imp.acquire_lock()
-    try:
-        _patch_logging()
-    finally:
-        imp.release_lock()
+    _patch_logging()
```
TheRealFalcon, line 41: The only place `patch` is called is in `main.py`. Can we remove the `patch` function altogether and rename `_patch_logging` to `patch_logging` and call that instead?

TheRealFalcon (commented):
Some minor requests inline

cloudinit/distros/__init__.py
paride: Thanks for the review, I chose `out_fn` as it was used more commonly in the `_write_hostname()` definitions, but I see your point and I'm fine with `filename`. I'm updating the branch,

cloudinit/patcher.py
paride: Well I guess `patch()` was meant to potentially patch more things via separate `_patch_something()` calls, but that code was last touched in 2016 and I don't see what else we may need to patch that way, so +1 for this.

paride:
PR updated and proposed commit message also updated.

TheRealFalcon (approved):
Thanks!

PR #947, opened 2021-07-16
Summary: Add 'network-config' support in ovf-env.xml through VMware g…

PengpengSun:
…uestinfo.ovfEnv

Details:
1. Support guest set network config through guestinfo.ovfEnv using OVF
2. 'network-config' Property is optional
3. 'network-config' Property's value has to be base64 encoded

Added unittests and updated ovf-env.xml example

## Test Steps
1. Create a ovf-env.xml which includes 'network-config' Property
2. Set 'network-config' Property's value to base64 encoded network config
3. Load ovf-env.xml to VMware guestinfo.ovfEnv
4. Verify cloud-init can get network config from guestinfo.ovfEnv in OVF

## Checklist:


TheRealFalcon (approved):
LGTM. Thanks!

cloudinit/sources/DataSourceOVF.py
```diff
@@ -358,8 +358,11 @@ def _get_data(self):
                 if contents:
                     break
             if contents:
-                (md, ud, cfg) = read_ovf_environment(contents)
+                read_network = ('com.vmware.guestinfo' == name)
```
t-8ch, line 361: Should this not have been `com.vmware.guestInfo` (different capitalization) the same as in line 353 above?

t-8ch: Also why is this not supposed to work with the ISO transport?

PengpengSun: @t-8ch Thanks for catching this. I should add more tests on this change. Will file a PR to fix it and add tests.

PengpengSun: @t-8ch For ISO transport, It can be supported, but user need create ISO for each instance with specified network configuration, I don't see it's a common user case when I opened this PR.
And  `com.vmware.guestInfo` is VMware platform specified, I get this request from inside of VMware, the guestinfo.ovfEnv can be updated easily from outside of instance(vm).

t-8ch: Would not VMware itself be able to generate the ISO on demand, at least when specifying the ISO transport in VMware?

PengpengSun: Yes, AFAIK, ISO can be generated on demand on VMware platform, till now, network config is never included in such ISO.  The ISO transport is not VMware specified, if there is indeed such requirement, I can make the change and let cloud-init team review it.

PengpengSun: @t-8ch Please fell free to send the patch, I was hesitating to add ISO transport because it's not specified to VMware platform. Let's see if cloud-init team is ok with it too.

PR #60, opened 2019-11-22
Multiple file fix for AuthorizedKeysFile config

otubo:
Currently cloud-init does not know how to handle multiple file
configuration on section AuthorizedKeysFile of ssh configuration.
cloud-init will mess up the home user directory by creating bogus
folders inside it.

This patch provides a fix for this erroneous behavior. It gathers all
keys from all the files listed on the section AuthorizedKeysFile of ssh
configuration and merge all of them inside home user
~/.ssh/authorized_keys of the vm deployed.

Resolves: rhbz#1642008

Signed-off-by: Eduardo Otubo <otubo@redhat.com>

PR #948, opened 2021-07-19
Stop storing system keys in user's authorized keys file

TheRealFalcon:
In #60, one of the changes added system-wide authorized keys into user-specific authorized keys files.

After #937, this could lead to some unintended consequences.

Given this cloud-config:
```
#cloud-config
bootcmd:
 - sed -i 's;#AuthorizedKeysFile.*;AuthorizedKeysFile /etc/ssh/authorized_keys /home/test_user/.ssh/authorized_keys;' /etc/ssh/sshd_config
users:
- default
- name: test_user
  ssh_authorized_keys:
    - <key2>
- name: test_user2
  ssh_authorized_keys:
    - <key3>
```
we would see `test_user`'s key info getting written into `test_user2`'s authorized keys file.

I understand that this sort of configuration seems very unlikely, but given this, plus the fact that we already have the global authorized keys file, I don't see a benefit to copying system-wide keys into a user-specific key file.

@otubo , you initially made this change, and @esposem , you changed this code most recently. Does it make sense to stop copying all system-wide keys into each user's authorizedkeys? Is there a use case this addresses that I'm missing?

Note that this PR is WIP because we'd still need to update tests. I'm just looking for validation of the idea at the moment.

### DIFF
--- a/cloudinit/ssh_util.py
+++ b/cloudinit/ssh_util.py
@@ -283,8 +283,10 @@ def extract_authorized_keys(username, sshd_cfg_file=DEF_SSHD_CFG):
             "AuthorizedKeysFile has an user-specific authorized_keys, "
             "using %s", user_authorizedkeys_file)
 
-    # always store all the keys in the user's private file
-    return (user_authorizedkeys_file, parse_authorized_keys(auth_key_fns))
+    return (
+        user_authorizedkeys_file,
+        parse_authorized_keys([user_authorizedkeys_file])
+    )
 
 
 def setup_user_keys(keys, username, options=None):


esposem:
Hi @TheRealFalcon, thanks for pointing this out. I created an unit test following your example, and I can reproduce the issue you mention.
One way I thought about this is that if `test_user` key is in the AuthorizedKeysFile, it means that the user wants it so be public so accessible also by `test_user2`. Therefore this can be the expected behavior.

Another solution for this use case could be to only recognize global-specific keys, and pass them as to `parse_authorized_keys`. A simple fix could be to discard `/home/` paths in the AuthorizedKeysFile line. What do you think?
This will force the user to avoid putting any global authorized_keys in an user-specific folder, which I think is reasonable.

This is the fix:
```Python
 with util.SeLinuxGuard(ssh_dir, recursive=True):
        try:
            ssh_cfg = parse_ssh_config_map(sshd_cfg_file)
            key_paths_full = ssh_cfg.get("authorizedkeysfile",
                                         "%h/.ssh/authorized_keys")
            key_paths = ""
            home_folder, _ = os.path.split(pw_ent.pw_dir)
            for key_path in key_paths_full.split(" "):
                if key_path.startswith('{}/'.format(home_folder)) and \
                not key_path.startswith('{}/'.format(pw_ent.pw_dir)):
                    LOG.debug("#### Discarding key %s", key_path)
                else:
                    key_paths += key_path + " "

            auth_key_fns = render_authorizedkeysfile_paths(
                key_paths, pw_ent.pw_dir, username)
```
I also have some additional tests ready.

Regarding the use case, @otubo can probably give a better explanation. I think that by not copying any key at all is not ideal, as we completely lose this feature.

esposem:
There is also another problem related to #937. Consider this case:
`AuthorizedKeysFile .ssh/authorized_keys /etc/ssh/userkeys/%u`
With the current cloudinit version, `/etc/ssh/userkeys/%u` will be picked as `user_authorizedkeys_file`. This is a problem, because that file might not be accessible by the user. Instead, it should default to `.ssh/authorized_keys`. In fact, I think this is the reason on why we need to copy the global keys. Because we can't really control which permission the user provides to these files, a simple solution is to assume that every file inside `/home/<current_user>` can be accessed by the user, everything outside `/home/` can be considered as global, and everything inside `/home/<other_user>` should not be considered for this user.

In general, it seems that we should:
1. Pick keys from the files defined in AuthorizedKeysFile that are not in `/home/<other_user>` and
2. Copy these keys in either a file defined in AuthorizedKeysFile that is in `/home/<current_user>`, or `.ssh/authorized_keys`.

otubo:
> @otubo , you initially made this change, and @esposem , you changed this code most recently. Does it make sense to stop copying all system-wide keys into each user's authorizedkeys? Is there a use case this addresses that I'm missing?

@TheRealFalcon I don't believe we have a specific use case for that.

> Pick keys from the files defined in AuthorizedKeysFile that are not in /home/<other_user> and
> Copy these keys in either a file defined in AuthorizedKeysFile that is in /home/<current_user>, or .ssh/authorized_keys.

@esposem Does it make sense to check if user specific files listed on AuthorizedKeysFile (i.e. `/home/test_user/.ssh/authorized_keys`) contains a token and accept it as valid only if it does? I believe it is acceptable to have `/home/%u/.ssh/authorized_keys` but not a hard coded username.

Also, we should check for permissions on global files like `/etc/ssh/userkeys/%u `

TheRealFalcon:
Thanks for the discussion everybody!

@esposem , I think we've identified two different problems, and I think we can address them both:
1. We're copying keys from globally defined authorized_keys into user-specific authorized_keys
2. We're not writing keys into the most ideal file(s)

Let's assume for a second that we can perfectly solve 2. If, for example, we have a line like the following:
`AuthorizedKeysFile /path/to/my/global/keys .ssh/authorized_keys /etc/ssh/userkeys/%u`
if we can ensure that we're correctly writing the user keys into `.ssh/authorized_keys` and/or `/etc/ssh/userkeys/%u`, is there ever a good reason to copy keys out of `/path/to/my/global/keys` and into the user locations? That is what we're doing currently, and my answer would be no. As long as the permissions to `/path/to/my/global/keys` is correct, every user already has access to those keys simply because they're in a global authorized_keys file. If the permissions on the global files are incorrect, that's not cloud-init's fault or responsibility to fix.

> One way I thought about this is that if test_user key is in the AuthorizedKeysFile, it means that the user wants it so be public so accessible also by test_user2. Therefore this can be the expected behavior.

I think that if the user wants it to be public, then they shouldn't be specifying a path in a user's home directory. That doesn't currently work with ssh, so I don't think we'd be breaking any sort of expectation that such a thing should work. I don't think this functionality adds any value. Additionally, as currently implemented, this would still be unreliable.

Say we defined users `user1`, `user2`, and `user3`. If cloud-init processes them in that order, then a line like
```
AuthorizedKeysFile /home/user1/.ssh/authorized_keys
```
would give user1's keys to user2 and user3. But if we had a line like:
```
AuthorizedKeysFile /home/user3/.ssh/authorized_keys
```
user1 and user2 would _not_ get user3's keys, because user3 is the last user to be handled.

> Another solution for this use case could be to only recognize global-specific keys, and pass them as to parse_authorized_keys. A simple fix could be to discard /home/ paths in the AuthorizedKeysFile line. What do you think?

I think this fix is a workaround for behavior we just shouldn't have.

> I think that by not copying any key at all is not ideal, as we completely lose this feature.

Is there anything else about this feature that I haven't addressed?

For problem 2, I think you've successfully identified the permissions problems. I didn't realize that for something like `/etc/ssh/userkeys/%u`, the permissions of the parent directory are also relevant. Given this, I think that when deciding which file to write to, instead of just choosing the last user-specific file, we should choose a location with valid permissions, then preferring a location in the user's home directory. So if we get a line like `AuthorizedKeysFile .ssh/authorized_keys /etc/ssh/userkeys/%u`, we'd choose `.ssh/authorized_keys` instead of `/etc/ssh/userkeys/%u`. If we do something like that, is there any reason to be copying keys around?

esposem:
@TheRealFalcon ok, so just to be sure I understand correctly, to sum up we need to add:
1. Ignore global keys, for the reason you explained above. How to identify global keys? My guess is that they must not be in `/home/<user>` and/or must not have `%h` and `%u`.
2. Between all possible key candidates to use, pick an user-specific one that has the right access privileges (0700 on the folder, 0600 on the file itself).

If you can confirm this, I can try to write a patch. I think it is also very very important to add test cases, to avoid having such problems in the future. I already implemented two unit tests for the case described above.

> Say we defined users user1, user2, and user3. If cloud-init processes them in that order, then a line like
> `AuthorizedKeysFile /home/user1/.ssh/authorized_keys`
> would give user1's keys to user2 and user3. But if we had a line like:
> `AuthorizedKeysFile /home/user3/.ssh/authorized_keys`
> `user1` and `user2` would not get `user3`'s keys, because `user3` is the last user to be handled.

I am confused on this example. Is this what is happening now? I don't understand why in the second case `user1` and `user2` won't get `user3`'s keys (assuming there are already some in `/home/user3/.ssh/authorized_keys`).

TheRealFalcon:
> 1. Ignore global keys, for the reason you explained above. How to identify global keys? My guess is that they must not be in /home/<user> and/or must not have %h and %u.

Not ignore global keys, just stop copying keys to other files period. If we go with the change I have in this PR it should accomplish that. On master, [this](https://github.com/canonical/cloud-init/blob/main/cloudinit/ssh_util.py#L304) returns all of the files specified on the `AuthorizedKeysFile`, reads all of the keys from all of them, then writes them into the user file selected. With my change, it returns only the user file selected, so it will only collect the keys from that file and then write back to it.

I agree on point 2.

> Say we defined users user1, user2, and user3. If cloud-init processes them in that order, then a line like
AuthorizedKeysFile /home/user1/.ssh/authorized_keys
would give user1's keys to user2 and user3. But if we had a line like:
AuthorizedKeysFile /home/user3/.ssh/authorized_keys
user1 and user2 would not get user3's keys, because user3 is the last user to be handled.
...
I am confused on this example. Is this what is happening now? I don't understand why in the second case user1 and user2 won't get user3's keys (assuming there are already some in /home/user3/.ssh/authorized_keys).

We currently write ALL keys found in any file specified in AuthorizedKeysFile, so if we start by processing user1, we haven't yet processed user3. If we haven't processed user3, we haven't written their key out yet. If we haven't written it out yet, there's no way for us to copy it into user1's key.

esposem:
Ok, I agree with you.
One more thing: I am trying to implement a fix for the permissions, but more than that it seems that the problem is in the owner of the folder containing the authorized_keys.
For example, assuming we have `/etc/ssh/userkeys/%u` and `cloud-user` and `root`, cloudinit will correctly set the permissions:
```
# ll
total 36
-rw-------. 1 cloud-user cloud-user 15687 Jul 30 06:08 cloud-user
-rw-------. 1 root       root       17199 Jul 30 06:08 root
```
However, [this](https://github.com/canonical/cloud-init/blob/main/cloudinit/ssh_util.py#L307) changes the owner of `/etc/ssh/userkeys` to the last user processed (`root`), leaving the folder inaccessible to the other users. Should we get rid of this line, and leave the problem of permissions to the user? I am not sure if that breaks something else though.

Also, should cloud-init set permission by itself at all here? changing the ownership and permissions of `/etc/ssh/userkeys/%u`?

esposem:
@TheRealFalcon  I created a new PR with my changes, feel free to give it a look and see if it works for you!
The permission check I added creates some problems with the unit tests though, since the file permission won't be correct and the test users are fake.

esposem:
I further investigated on this, and together with @otubo we figured that we need also to consider the `StrictMode yes/no` option in `AuthorizedKeysFile`.
From `man sshd_config`:
```
StrictModes
             Specifies whether sshd(8) should check file modes and ownership
             of the user's files and home directory before accepting login.
```
and from `man sshd`:
```
~/.ssh/authorized_keys
             Lists the public keys (DSA, ECDSA, Ed25519, RSA) that can be used
             for logging in as this user.  The format of this file is de‐
             scribed above.  The content of the file is not highly sensitive,
             but the recommended permissions are read/write for the user, and
             not accessible by others.

             If this file, the ~/.ssh directory, or the user's home directory
             are writable by other users, then the file could be modified or
             replaced by unauthorized users.  In this case, sshd will not al‐
             low it to be used unless the StrictModes option has been set to
             “no”.
```

The permissions are usually:
```
0600 for .ssh, and /home/user (owner user)
0700 for authorized_keys (owner user)
```
This means that if StrictModes is "yes", giving options like `/etc/ssh/userkeys/%u` will not work at all for any user except `root`, because the permissions in `/etc/ssh/userkeys` can only be correct for one user (usually `root`), and not the other ones.
What I am not sure and that the documentation does not clarify is what to do if the user specifies something like
`/keys/%u/authorized_keys`, with
```
keys: owned by root, 0600
%u: owned by user u, 0600
authorized_keys: owned by user u, 0700
```
How up in the directory tree should cloud-init check the permissions? Because in this case, `authorized_keys`'s parent (`%u`) has the right permissions, but `keys` has not. So should we use it as good candidate for `user_authorized_keys` or not? In PR #956 I do accept it, but I am not sure about this.

If StrictModes is "no", then we don't check the permissions at all.

Regarding tests and to see if we are on the same page, I wrote down a lot of use cases with what expect as behavior from cloud-init. Please let me know if you agree with the results. It's a lot of cases, so feel free to skip the initial ones as they are very simple. 
[cloud-init-example.txt](https://github.com/canonical/cloud-init/files/6919606/cloud-init-example.txt)

TheRealFalcon:
Closing this in favor of #956

PR #952, opened 2021-07-21
Only invoke hotplug socket when functionality is enabled (SC-251)

TheRealFalcon:
## Proposed Commit Message

```
Only invoke hotplug socket when functionality is enabled

Alter's hotplug hook to have a query mechanism checking if the
functionality is enabled. This allows us to avoid using the hotplug
socket and service when hotplug is disabled.

`cloud-init devel hotplug-hook -s net query` will return 'enabled' or
'disabled' depending on hotplug availability.
```

## Additional Context
n/a

## Test Steps
see integration test

## Checklist:

### DIFF
--- a/cloudinit/cmd/devel/hotplug_hook.py
+++ b/cloudinit/cmd/devel/hotplug_hook.py
@@ -3,6 +3,7 @@
 import abc
 import argparse
 import os
+import sys
 import time
 
 from cloudinit import log
@@ -12,7 +13,7 @@ from cloudinit.net import activators, read_sys_net_safe
 from cloudinit.net.network_state import parse_net_config_data
 from cloudinit.reporting import events
 from cloudinit.stages import Init
-from cloudinit.sources import DataSource
+from cloudinit.sources import DataSource, DataSourceNotFoundException
 
 
 LOG = log.getLogger(__name__)
@@ -236,7 +237,13 @@ def handle_args(name, args):
     with hotplug_reporter:
         try:
             if args.hotplug_action == 'query':
-                hotplug_init.fetch(existing="trust")
+                try:
+                    hotplug_init.fetch(existing="trust")
+                except DataSourceNotFoundException:
+                    print(
+                        "Unable to determine hotplug state. No datasource "
+                        "detected")
+                    sys.exit(1)
                 enabled = is_enabled(hotplug_init, args.subsystem)
                 print('enabled' if enabled else 'disabled')
             else:
--- a/cloudinit/cmd/devel/hotplug_hook.py
+++ b/cloudinit/cmd/devel/hotplug_hook.py
@@ -31,15 +31,31 @@ def get_parser(parser=None):
         parser = argparse.ArgumentParser(prog=NAME, description=__doc__)
 
     parser.description = __doc__
-    parser.add_argument("-d", "--devpath", required=True,
-                        metavar="PATH",
-                        help="sysfs path to hotplugged device")
-    parser.add_argument("-s", "--subsystem", required=True,
-                        help="subsystem to act on",
-                        choices=['net'])
-    parser.add_argument("-u", "--udevaction", required=True,
-                        help="action to take",
-                        choices=['add', 'remove'])
+    parser.add_argument(
+        "-s", "--subsystem", required=True,
+        help="subsystem to act on",
+        choices=['net']
+    )
+
+    subparsers = parser.add_subparsers(dest='hotplug_action')
+
+    subparsers.add_parser(
+        'query',
+        help='query if hotplug is enabled for given subsystem'
+    )
+
+    parser_handle = subparsers.add_parser(
+        'handle', help='handle the hotplug event')
+    parser_handle.add_argument(
+        "-d", "--devpath", required=True,
+        metavar="PATH",
+        help="sysfs path to hotplugged device"
+    )
+    parser_handle.add_argument(
+        "-u", "--udevaction", required=True,
+        help="action to take",
+        choices=['add', 'remove']
+    )
 
     return parser
 
@@ -133,6 +149,14 @@ SUBSYSTEM_PROPERTES_MAP = {
 }
 
 
+def is_enabled(hotplug_init, subsystem):
+    scope = SUBSYSTEM_PROPERTES_MAP[subsystem][1]
+    return hotplug_init.update_event_enabled(
+        event_source_type=EventType.HOTPLUG,
+        scope=scope
+    )
+
+
 def handle_hotplug(
     hotplug_init: Init, devpath, subsystem, udevaction
 ):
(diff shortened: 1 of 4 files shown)


cloudinit/cmd/devel/hotplug_hook.py
```diff
@@ -200,29 +221,30 @@ def handle_args(name, args):
...
     with hotplug_reporter:
         try:
-            handle_hotplug(
-                hotplug_init=hotplug_init,
-                devpath=args.devpath,
-                subsystem=args.subsystem,
-                udevaction=args.udevaction,
-            )
+            if args.hotplug_action == 'query':
+                hotplug_init.fetch(existing="trust")
```
blackboxsw, line 238: If we run this before the datasource is detected, we trace here:
```bash
cloudinit.sources.DataSourceNotFoundException: Did not find any data source, searched classes: (DataSourceNone)
```
We might want to catch that case and at least report "unknown" or "no-datasource" etc.
```suggestion
                try:
                    hotplug_init.fetch(existing="trust")
                except cloudinit.sources.DataSourceNotFoundException:
                     print("Unable to determine hotplug state. No datasource detected")
                     sys.exit(1)              
```

Also is it worth just baking this hotplug_init.fetch(existing="trust") into is_enabled() function because I see this same potential problem in handle_hotplug too?

I realize generally this cmdline shouldn't normally get run until after the datasource detection has occurred, just a race there until that discovery happens.

cloudinit/cmd/devel/hotplug_hook.py
```diff
@@ -31,15 +31,31 @@ def get_parser(parser=None):
...
-                        help="action to take",
-                        choices=['add', 'remove'])
+    parser.add_argument(
+        "-s", "--subsystem", required=True,
+        help="subsystem to act on",
+        choices=['net']
+    )
+
+    subparsers = parser.add_subparsers(dest='hotplug_action')
+
```
blackboxsw, line 41: I think you need a `    subparsers.required = True` to make sure we choose either query or handle. otherwise we get this trace:
```bash
root@dev-b:~# cloud-init devel hotplug-hook -s net
AttributeError: 'Namespace' object has no attribute 'devpath'
```

```suggestion
    subparsers.required = True

```

cloudinit/cmd/devel/hotplug_hook.py
```diff
@@ -133,6 +149,14 @@ def device_detected(self) -> bool:
 }
 
 
+def is_enabled(hotplug_init, subsystem):
+    scope = SUBSYSTEM_PROPERTES_MAP[subsystem][1]
```
blackboxsw, line 153: run try/except hotplug_init.fetch(existing="trust") here?

blackboxsw (changes requested):
looks good. minor changes for CLI usability.

cloudinit/cmd/devel/hotplug_hook.py
TheRealFalcon: Our udev hook should disable any hotplug commands until cloud-init has finished its initial run. How did you get this exception to happen? I don't necessarily disagree with your suggestion, just trying to understand.

cloudinit/cmd/devel/hotplug_hook.py
TheRealFalcon: I don't think we want it here. I applied the try/catch to the other place you suggested, but I don't think we always want to print and exit 1 when we get that exception. E.g., on line 174, if we have a problem there, I want it written the logs. I'm ok with the exception just falling through at that point.

blackboxsw (changes requested):
Besides the minor nit about the subparser title I think there are two other things:

UeventHandler.update_metadata raises a RuntimeError when a datasource is not configured to with a matching self.supported_update_event type for the subsystem and or hotplug. This configuration is static, so retrying on this RuntimeError doesn't make sense.

I'd prefer we just try handler.update_metadata and if fail we break the retry look in handle_hotplug.

Example, trying to provide the following in a lxc container:/etc/cloud/cloud.cfg
```yaml
updates:
    network:
       when: ['boot', 'hotplug']
```
The result is blocking repeated RuntimeErrors logged with 
```
RuntimeError: Datasource DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud-net][dsmode=net] not updated for event hotplug
```


I think instead we probably want to break the retry loop with a more specific failure mode if someone tries to enable hotplug on an unsupported datasource.

Here's a diff to ponder that fixes up a couple things I've suggested
```diff
diff --git a/cloudinit/cmd/devel/hotplug_hook.py b/cloudinit/cmd/devel/hotplug_hook.py
index 5c85c623a..3e1ffeecc 100644
--- a/cloudinit/cmd/devel/hotplug_hook.py
+++ b/cloudinit/cmd/devel/hotplug_hook.py
@@ -38,7 +38,9 @@ def get_parser(parser=None):
         choices=['net']
     )
 
-    subparsers = parser.add_subparsers(dest='hotplug_action')
+    subparsers = parser.add_subparsers(
+        title="Hotplug Action", dest='hotplug_action'
+    )
     subparsers.required = True
 
     subparsers.add_parser(
@@ -100,14 +102,8 @@ class UeventHandler(abc.ABC):
         return self.success_fn()
 
     def update_metadata(self):
-        result = self.datasource.update_metadata_if_supported([
         return self.success_fn()
 
     def update_metadata(self):
-        result = self.datasource.update_metadata_if_supported([
+        return self.datasource.update_metadata_if_supported([
             EventType.HOTPLUG])
-        if not result:
-            raise RuntimeError(
-                'Datasource %s not updated for '
-                'event %s' % (self.datasource, EventType.HOTPLUG)
-            )
-        return result
 
 
 class NetHandler(UeventHandler):
@@ -194,7 +190,12 @@ def handle_hotplug(
         )
         try:
             LOG.debug('Refreshing metadata')
-            event_handler.update_metadata()
+            if not event_handler.update_metadata():
+                LOG.debug(
+                    "Datasource%s does not support hotplug events for"
+                    " subsystem %s", event_handler.datasource.dsname, subsystem
+                )
+                break  # Don't retry as supported_update_events are hard-coded
             LOG.debug('Detecting device in updated metadata')
             event_handler.detect_hotplugged_device()
             LOG.debug('Applying config change')
```

TheRealFalcon:
@blackboxsw , thanks, I definitely missed a condition there.

> I think instead we probably want to break the retry loop with a more specific failure mode if someone tries to enable hotplug on an unsupported datasource.

I updated the code so that before we attempt to do any metadata update, we first fetch the datasource and _ensure it is enabled_. I didn't have the enabled requirement when doing a query previously.

I still haven't tested yet (so sorry if there's major dumb errors), but let me know if this idea works for you.

blackboxsw (approved):
This is excellent @TheRealFalcon . tested on lxc and openstack, cli docs are good and behavior is expected 
```
2021-08-13 20:11:27,422 - handlers.py[DEBUG]: finish: hotplug-hook: SUCCESS: Handle reconfiguration on hotplug events
```
:)

PR #953, opened 2021-07-23
Datasource for VMware

akutz:
## Proposed Commit Message

```
Datasource for VMware

This patch finally introduces the Cloud-Init Datasource for VMware
GuestInfo as a part of cloud-init proper. This datasource has existed
since 2018, and rapidly became the de facto datasource for developers
working with Packer, Terraform, for projects like kube-image-builder,
and the de jure datasource for Photon OS.

The major change to the datasource from its previous incarnation is
the name. Now named DatasourceVMware, this new version of the
datasource will allow multiple transport types in addition to
GuestInfo keys.

This datasource includes several unique features developed to address
real-world situations:

  * Support for reading any key (metadata, userdata, vendordata) both
    from the guestinfo table when running on a VM in vSphere as well as
    from an environment variable when running inside of a container,
    useful for rapid dev/test.

  * Allows booting with DHCP while still providing full participation
    in Cloud-Init instance data and Jinja queries. The netifaces library
    provides the ability to inspect the network after it is online,
    and the runtime network configuration is then merged into the
    existing metadata and persisted to disk.

  * Advertises the local_ipv4 and local_ipv6 addresses via guestinfo
    as well. This is useful as Guest Tools is not always able to
    identify what would be considered the local address.

The datasource currently lives in its own GitHub repository at
https://github.com/vmware/cloud-init-vmware-guestinfo. Once the datasource
is merged into Cloud-Init, the old repository will be deprecated.
```

## Additional Context
* The primary author and current steward of this datasource spoke at Cloud-Init Con 2020 where there was interest in contributing this datasource to the Cloud-Init codebase.
* This datasource has an active community that provides immense support to the project. Cc @codenrhoden, @toabctl, @jaymzmac, @prologic, @MnrGreg, @janitha, @powersj

## Test Steps
* There is [a walkthrough](https://github.com/vmware/cloud-init-vmware-guestinfo#walkthrough) of using the datasource in its current repository. 
* The datasource has been used in production for several years as part of:
  * [Photon OS](https://github.com/vmware/photon)
  * [kube-image-builder](https://github.com/kubernetes-sigs/image-builder)
  * [vmware-haproxy](https://github.com/haproxytech/vmware-haproxy)

## Checklist:


akutz:
It says the CLA validate failed. I did sign it (this morning). I suppose it just is not approved yet?

powersj:
Yeah the CLA check takes 1-2 days on our end, so probably early next week

akutz:
Thanks @powersj ,

I will fix the lint issues. I also thought I removed the dependency on `deepmerge`. It would be ironic if that's a PR I did not yet...merge :)

akutz:
Hi @powersj,

I was looking to see if it possible to have a datasource respond to two different names, and while that does not appear to be the case, I did find that this datasource no longer fully matches the [DS docs](https://cloudinit.readthedocs.io/en/latest/topics/datasources.html). I am going to update this DS with an RST doc instead of the MD one I included, as well as determine which of the current methods the DS is not implementing. While I know the DS works, I'd rather it be merged at a baseline that is congruent with the current criteria. Thanks!

TheRealFalcon:
With regards to the CLA, I can see that you've signed the CLA, but you haven't yet added yourself to the CLA signers file. As part of this PR, please also add your name (alphabetically) to [the CLA signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers). The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html#submitting-your-first-pull-request).

Based on your last comment, it sounds like there's still more work you want to do as part of this PR. Are you looking for review feedback now, or would you like to make the updates first?

akutz:
> With regards to the CLA, I can see that you've signed the CLA, but you haven't yet added yourself to the CLA signers file. As part of this PR, please also add your name (alphabetically) to [the CLA signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers). The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html#submitting-your-first-pull-request).
> 
> 
> 
> Based on your last comment, it sounds like there's still more work you want to do as part of this PR. Are you looking for review feedback now, or would you like to make the updates first?

Thanks for the question! Well, you can provide feedback if you like :)

I think all I'm going to do for now is:

- Rename this to "DataSourceVMware" to follow the existing pattern of other DS names
- Change the MD to an RST per the docs

Anything else can come later, or as part of the review. I realize I'm missing some functions in the DS prototype per the current documentation. I'd rather y'all let me know what is missing. Some things are a result of maintaining Py2 compat, but that's no longer a concern once merged into CI proper.

akutz:
Okay, I've updated the name and docs. Have at ye! :)

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+            if isinstance(wait_on_ipv6_val, bool):
+                wait_on_ipv6 = wait_on_ipv6_val
+            else:
+                wait_on_ipv6 = bool(strtobool(wait_on_ipv6_val))
+
+    # Get information about the host.
+    host_info = None
+    while host_info is None:
+        host_info = get_host_info()
+        if wait_on_ipv4:
```
TheRealFalcon, line 725: The nesting in this function goes fairly deep in places. Would it be possible to use a variable to represent the state instead? I'm thinking something along the lines of...
```python
        network = host_info.get('network') or {}
        interfaces = network.get('interfaces') or {}
        by_ipv4 = interfaces.get('by-ipv4') or {}
        if wait_on_ipv4:
            ipv4_ready = len(by_ipv4) > 0 if by_ipv4 else False
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+    wait_on_ipv4 = False
+    wait_on_ipv6 = False
+    if WAIT_ON_NETWORK in metadata:
+        wait_on_network = metadata[WAIT_ON_NETWORK]
+        if WAIT_ON_NETWORK_IPV4 in wait_on_network:
+            wait_on_ipv4_val = wait_on_network[WAIT_ON_NETWORK_IPV4]
+            if isinstance(wait_on_ipv4_val, bool):
+                wait_on_ipv4 = wait_on_ipv4_val
+            else:
+                wait_on_ipv4 = bool(strtobool(wait_on_ipv4_val))
```
TheRealFalcon, line 713: In `cloudinit.util` there's a function called `translate_bool`. It can be used here and in the similar block just below here.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+    if VMWARE_RPCTOOL:
+        return VMWARE_RPCTOOL
+    return None
+
+
+_MERGE_STRATEGY_ENV_VAR = "CLOUD_INIT_VMWARE_GUEST_INFO_MERGE_STRATEGY"
+_MERGE_STRATEGY_DEEPMERGE = "deepmerge"
+
+
+def merge_dicts(a, b):
```
TheRealFalcon, line 770: I haven't grokked all the merging code here, but would `util.mergemanydict` work as a replacement?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+            elif "addr" in af_inet4[0]:
+                ipv4 = af_inet4[0]["addr"]
+
+    return ipv4, ipv6
+
+
+# patched socket.getfqdn() - see https://bugs.python.org/issue5004
+
+
+def getfqdn(name=""):
```
TheRealFalcon, line 567: does `util.get_fqdn_from_hosts()` work here?

TheRealFalcon (changes requested):
This looks good overall. I'm used to seeing cloud-init retrieving the network information from the datasource to apply to the machine and not the other way around, so this is a bit unique in that regard.

I left a number of inline comments, but most are pretty minor. The one major thing is the use of netifaces. We're a bit wary of adding new dependencies to cloud-init. We have some similar networking reporting tools in [cloudinit/netinfo.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/netinfo.py). Would you be able to use the functions there instead of pulling in `netifaces`?

Additionally, there are a few places where you reference support for older versions of cloud-init. Since this datasource will ship as part of whatever version of cloud-init we're shipping, compatibility with older versions won't be a concern anymore. I highlighted these inline as well.

You may want to add VMware to `Supported Private Clouds` section in `README.md`.

Each datasource has its own set of unit tests under [tests/unittests/test_datasource](https://github.com/canonical/cloud-init/tree/main/tests/unittests/test_datasource). Would you be able to add your own?

akutz:
> This looks good overall. I'm used to seeing cloud-init retrieving the network information from the datasource to apply to the machine and not the other way around, so this is a bit unique in that regard.
> 
> I left a number of inline comments, but most are pretty minor. The one major thing is the use of netifaces. We're a bit wary of adding new dependencies to cloud-init. We have some similar networking reporting tools in [cloudinit/netinfo.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/netinfo.py). Would you be able to use the functions there instead of pulling in `netifaces`?
> 
> Additionally, there are a few places where you reference support for older versions of cloud-init. Since this datasource will ship as part of whatever version of cloud-init we're shipping, compatibility with older versions won't be a concern anymore. I highlighted these inline as well.
> 
> You may want to add VMware to `Supported Private Clouds` section in `README.md`.
> 
> Each datasource has its own set of unit tests under [tests/unittests/test_datasource](https://github.com/canonical/cloud-init/tree/main/tests/unittests/test_datasource). Would you be able to add your own?

Hi @TheRealFalcon,

Thank you for the review! Before I respond/address/fix anything else, I wanted to comment on the `netifaces` remark. I raised this in cloud-init con 2020, and the leadership there did not think it would be an issue. 

Looking at `cloudinit/netinfo.py`, I noticed it shell execs out to commands like `ip`, `ifconfig`, and `netstat` to gather most of its information. However, this is not as portable as [C-level interop implemented by netifaces](https://github.com/al45tair/netifaces/blob/master/netifaces.c). Additionally, it's unclear to me if it supports determining the default gateway for a particular device or address family the way netifaces does and on which this datasource relies.

Netifaces supports Windows as well. While that's not relevant _here_ per se, Cluster API (CAPI), the Kubernetes project for bootstrapping Kube clusters with Kubernetes, supports Linux and Windows nodes. The bootstrap controller produces cloud-init payloads for Linux guests, and in order to support Windows, the CAPI team worked with Microsoft to [port this datasource to cloudbase-init](https://github.com/cloudbase/cloudbase-init/commit/a77477e16e89979684ec31ffd2b0d479cf18b245) in order to make it as seamless as possible for CAPI and its infrastructure providers (Azure, AWS, VMware, etc.) to support bootstrapping Windows nodes.

That's a long way to say, can I pretty please continue to use `netifaces`, if only for the initial merge and then perhaps figure out a future path where it's not needed? It's just a lot of up front work to change now. I'm not opposed to it, but as you are aware, any changes, no matter how small, require revalidation. I know what we have _works_, and while new unit tests might show some other method works, nothing is as definitive as two years of production workloads.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+        if (
+            data_access_method == VMWARE_RPCTOOL
+            and CLEANUP_GUESTINFO in self.metadata
+        ):
+            clear_guestinfo_keys(self.metadata[CLEANUP_GUESTINFO])
+
+        if self.metadata or self.userdata_raw or self.vendordata_raw:
+            return True
+        else:
+            return False
```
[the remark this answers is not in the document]
akutz: Perhaps. I like syntactic sugar as much as the next person, and if that's the typical cloud-init pattern, I'm happy to adapt it. But as the years go by, I guess I tend towards more explicit logic statements. If I do change the syntax, I'll add a comment that indicates something along the lines of:

```
# get_data returns true if there is any available metadata, userdata, or vendordata.
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+    """
+
+    dsname = "VMwareGuestInfo"
+
+    def __init__(self, sys_cfg, distro, paths, ud_proc=None):
+        sources.DataSource.__init__(self, sys_cfg, distro, paths, ud_proc)
+        if not get_data_access_method():
+            LOG.error("Failed to find vmware-rpctool")
+
+    def get_data(self):
```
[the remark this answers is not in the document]
akutz: Probably. This is part of the external DS still supporting Py2 in addition to Py3. There's definitely some optimizations I could make to remove the Py2 compatibility. I should touch upon this in a general comment. I'm hesitant to make any changes as part of the initial merge that would necessitate complete end-to-end testing. Then again, that's not necessarily a bad thing :)

doc/rtd/topics/datasources/vmware.rst
```diff
@@ -0,0 +1,359 @@
...
+
+#. Create the metadata file for the VM. Save the following YAML to a file named
+   ``metadata.yaml``\ :
+
+   .. code-block:: yaml
+
+       instance-id: cloud-vm
+       local-hostname: cloud-vm
+       network:
+       version: 2
```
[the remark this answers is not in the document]
akutz: Perhaps. It works as-is though (I tested it by running `make doc` and opening `index.html` locally). The problem is the `m2r` command does not convert the original markdown very well at all, and it does not offer a way to specify tab width. I am new to `rst`, and the other datasource docs use varying widths, two, three, four. It was unclear if there is a standard width (`m2r` chose three), and since the original markdown used four, I just verified things looked correct in the browser and was going to ask about indentation policy later (again, there doesn't seem to be a standard based on the docs for the other datasources).

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+        "network": {
+            "interfaces": {
+                "by-mac": collections.OrderedDict(),
+                "by-ipv4": collections.OrderedDict(),
+                "by-ipv6": collections.OrderedDict(),
+            },
+        },
+    }
+
+    hostname = getfqdn(socket.gethostname())
```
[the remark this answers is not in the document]
akutz: Hmm, I don't think it would? The logic paths are different. Elsewhere you remarked:

>  I'm used to seeing cloud-init retrieving the network information from the datasource to apply to the machine and not the other way around, so this is a bit unique in that regard.

Some background is helpful here:

* This datasource does not depend on L3 networking, even an ephemeral DHCP network, to fetch its network configuration. Instead this currently happens via a series of RPC calls for `guestinfo` key/value pairs.
* This datasource was created in an environment where DHCP was used, which was not initially a big deal, but...
* Eventually the need to support Cluster API (CAPI) meant supporting cloud-init instance data JINJA queries
* Except cloud-init instance data only includes the information passed _into_ the metadata
* This is why this datasource uses the `setup` function to inspect the network after it is online and repersists the metadata back out to disk as the instance data file
* Now the user and vendor data can execute JINJA queries in the network boot stage and beyond that discover hostname, local_ipv4, and other information that were not available until the VM's DHCP network was brought online in the local boot stage.

All this is to say the reason this datasource's `getfqdn` function does not look at the metadata is because:

* The result of `getfqdn` might be used to _update_ the metadata in order to persist the DHCP-provided hostname back to disk in its instance data
* The `get_hostname` function has comments that maybe it should return an error if the datasource doesn't provide a hostname when requested.

So to me it seems like this is behavior that should continue to exist in the datasource and not rely on an implementation of the base class that could change.

doc/rtd/topics/datasources/vmware.rst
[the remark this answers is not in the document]
TheRealFalcon: Sorry, for this I don't mean the rst indentation, I mean that the section that comes under `network` should be further indented as part of the network definition in the yaml.

cloudinit/sources/DataSourceVMware.py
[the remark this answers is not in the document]
TheRealFalcon: That's fine. It can stay as-is right now. It helps me to remember that isn't an entirely new datasource...just new to the cloud-init codebase.

cloudinit/sources/DataSourceVMware.py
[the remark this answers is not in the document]
TheRealFalcon: It's fine as is.

TheRealFalcon:
> That's a long way to say, can I pretty please continue to use netifaces, if only for the initial merge and then perhaps figure out a future plan for removing the dependency?

Only because you said pretty please :wink: .

I remember the conversations about netifaces and I also remember concluding that it shouldn't be too hard to replace it, so that's why I suggested it here. Your reasoning makes sense though, and I think it's ok to keep it currently, but I would also like to get a +1 from @blackboxsw on this.

doc/rtd/topics/datasources/vmware.rst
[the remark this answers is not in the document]
akutz: Ugh, yes. VSCode and it's block tab indentation. Usually I have to just do a find/replace with a regex `s/^/<WHITE_SPACE>/g` to achieve consistent, leading white space. I guess I did not do that here.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+            public_keys_data = self.metadata["public-keys-data"].splitlines()
+
+        public_keys = []
+        if not public_keys_data:
+            return public_keys
+
+        for public_key in public_keys_data:
+            public_keys.append(public_key)
+
+        return public_keys
```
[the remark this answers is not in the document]
akutz: Damn this is slick, thank you!

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,816 @@
...
+        raw_data = zlib.decompress(base64.b64decode(data), zlib.MAX_WBITS | 16)
+    elif enc_type == "base64" or enc_type == "b64":
+        LOG.debug("Decoding %s format %s", enc_type, key)
+        raw_data = base64.b64decode(data)
+    else:
+        LOG.debug("Plain-text data %s", key)
+        raw_data = data
+
+    if isinstance(raw_data, bytes):
+        return raw_data.decode("utf-8")
```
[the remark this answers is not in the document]
akutz: Not sure. I think you'd be a better arbiter of that call than me. Happy to remove it if you think it is safe to do so.

cloudinit/sources/DataSourceVMware.py
akutz: Only if the `/etc/hosts` file has been configured correctly, which does not always occur when DHCP is in use due to drift between the host config and the files. We found this out the hard way in https://github.com/kubernetes-sigs/cluster-api/issues/1580.

cloudinit/sources/DataSourceVMware.py
akutz: Great catch!

cloudinit/sources/DataSourceVMware.py
akutz: Great catch, yes it can!

akutz:
Hi @TheRealFalcon,

> Each datasource has its own set of unit tests under [tests/unittests/test_datasource](https://github.com/canonical/cloud-init/tree/main/tests/unittests/test_datasource). Would you be able to add your own?

Absolutely, of course. The only reason I did not was because I'm not sure how to mock the RPC calls. I can create tests for everything post the actual retrieval of the data. The other challenge is the fact the datasource relies on inspecting the host's network configuration. I am not sure how I should mock that so the answer is deterministic enough to be testable.

Do you have suggestions on how to do either? FWIW, this is why testing for me has always been firing up a VM on my desktop in Fusion and verifying the outcome. The datasource's RPC calls via `vmware-rpctool` will read data placed in the VM's `vmx` config file on disk. Here's [an example](https://github.com/kubernetes-sigs/image-builder/blob/master/images/capi/hack/image-post-create-config.sh) of how I tested it in kube-image-builder. 

I'm more than happy to add unit tests if we can figure out how to address the two items above. Heck, I'm happy to add some even without addressing those items, but given those items _should_ be tested, I would sure appreciate helping figuring out how to do so.

TheRealFalcon:
> Absolutely, of course. The only reason I did not was because I'm not sure how to mock the RPC calls. I can create tests for everything post the actual retrieval of the data. The other challenge is the fact the datasource relies on inspecting the host's network configuration. I am not sure how I should mock that so the answer is deterministic enough to be testable.

In general there's some helper base classes that do a lot of the mocking for you. Just as an example, if you take a look at [test_digitalocean.py](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_datasource/test_digitalocean.py#L138), you can see it inherits from `CiTestCase` and with very little additional setup you should be able to instantiate a datasource and start interacting with it. Beyond that, I think it's just mocking the individual parts of the datasource that make external calls or rely on specific instance state. In your case, you could probably mock `get_guestinfo_value`, or perhaps the subp/env vars within it. You'd also need to mock some of the netifaces/ip_address calls.

Does that help at all?

> FWIW, this is why testing for me has always been firing up a VM on my desktop in Fusion and verifying the outcome.

Yeah, unit testing is frustrating when we have to mock so much. It definitely helps in regression testing though. We do also have an integration testing framework that lets us do some more real-world test cases. These tests rely heavily on another library called [pycloudlib](https://github.com/canonical/pycloudlib). The scope would be beyond this PR, but if we could make VMware a supported "cloud" in pycloudlib, it would it very simple to write an integration test for this datasource that is end-to-end and doesn't require any mocking. I'll admit that I'm fairly oblivious to the VMware ecosystem, so I would need to come up to speed on how the VMs are launched and managed, but if you're interested in more representative tests, this would be the way to go. It'd be a welcome addition into pycloudlib.

akutz:
HI @TheRealFalcon,

Holy. crap.

`@mock.patch` is magic. It's pure magic. Seriously. This is amazing.

akutz:
Hi @TheRealFalcon,

Okay, the tests have been added. They're not 100% code coverage, but I think they cover the necessary bits. Please let me know what you think. Thanks again!

akutz:
Hi @TheRealFalcon,

I've added tests for `ds-identify` as well.

akutz:
Please note that once this is merged, the plan is to relocate the VMware-specific bits out of the generic OVF datasource into the VMware datasource. I just do not want to do that as part of _this_ PR since there is no reason to force that much change in a single PR IMO.

akutz:
Please do let me know if there is anything else I can do to help facilitate the decision on whether or not to merge this. I'm happy to address any concerns here or in a high-bandwidth setting. Thanks!

cloudinit/sources/DataSourceVMware.py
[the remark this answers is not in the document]
TheRealFalcon: Eh...leaving it doesn't hurt anything

TheRealFalcon (approved):
@akutz , thanks for the tests. I'm glad you enjoyed python's mocking system :slightly_smiling_face: .

I think this PR is generally in a good state. I'm still waiting for an additional +1, but I think we'll be able to get this merged soon. Thanks for the contribution!

akutz:
Thank you for taking the time to review it as well as help me figure out the testing strategy.

I look forward to getting this merged so the Guest Tools team here at VMware can start a PR to move their bits from the generic OVF DS to the new VMware DS.

That and I can finally stop hearing "When are you going to merge this into cloud-init?!" every few weeks :)

akutz:
HI @TheRealFalcon,

Could you please re-approve? I rebased onto `main` since this branch fell behind while waiting for a second approval. In doing the rebase I also lost your approval :(

akutz:
> HI @TheRealFalcon,
> 
> Could you please re-approve? I rebased onto `main` since this branch fell behind while waiting for a second approval. In doing the rebase I also lost your approval :(

Oh maybe I didn't lose the approval? I thought I did because it says "1 workflow awaiting approval", but I think that's the CLA validation, not your approval. Sorry about that.

TheRealFalcon:
Yeah, Github is trying to protect us by preventing first-time contributors to this repo from automatically executing certain github actions. In this case, the only thing it is preventing is the CLA check which is somewhat ironic.

I'll have to re-allow it on every update to the branch until merged.

akutz:
Wait, you have to explicitly allow the CLA check? That is...well, it would be hilarious if not so annoying for you.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+                vd = guestinfo_envvar("vendordata")
+
+                if md or ud or vd:
+                    self.data_access_method = DATA_ACCESS_METHOD_ENVVAR
+
+        if not self.data_access_method:
+            LOG.error("failed to find a valid data access method")
+            return False
+
+        LOG.info("using data access method %s", self.data_access_method)
```
blackboxsw, line 130: This is good information that should probably also be returned by a  DatasourceVMware._get_subplatform method to better describe to instance-data.json what "sub flavor" of metadata sources we have detected on the system. See [DataSourceOVF](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceOVF.py#L400-L404) and [DataSourceNoCloud](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceNoCloud.py#L204-L210) for examples of the types of descriptive strings that might allow someone to better triage where to datasource discovered metadata came from.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+            "Unexpected error while trying to set "
+            + "guestinfo key=%s to value=%s",
+            key,
+            value,
+        )
+
+    return None
+
+
+def guestinfo_clear_keys(keys, vmware_rpctool=VMWARE_RPCTOOL):
```
blackboxsw, line 410: bikeshed function rename just because we use "redact" in a number of places in cloudinit and it would be nice to have greps find them all in the future in case we standardize redact operations for all datasources, logging, instance data etc.
 
```suggestion
def guestinfo_redact_keys(keys, vmware_rpctool=VMWARE_RPCTOOL):
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+    load first attempts to unmarshal the provided data as JSON, and if
+    that fails then attempts to unmarshal the data as YAML. If data is
+    None then a new dictionary is returned.
+    """
+    if not data:
+        return {}
+    try:
+        return json.loads(data)
+    except (json.JSONDecodeError, TypeError):
+        return safeyaml.load(data)
```
blackboxsw, line 442: These could use both util.load_json and util.load_yaml, which would also allow handling binary data.... but maybe you don't want all the extra baked in decoding operations?
```suggestion
    try:
        return util.load_json(data)
    except (json.JSONDecodeError, TypeError):
        return util.load_yaml(data)
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+        network = data["network"]
+        del data["network"]
+
+    network_enc = None
+    if "network.encoding" in data:
+        network_enc = data["network.encoding"]
+        del data["network.encoding"]
+
+    if network:
+        LOG.debug("network data found")
```
blackboxsw, line 461: Probably can drop this log as you log the data below with the dynamic network value on line 469 of the visual diff.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+        pass
+    else:
+        for addr in addrs:
+            if addr[3]:
+                name = addr[3]
+                break
+    return name
+
+
+def is_valid_ip_addr(val):
```
blackboxsw, line 598: We are doing a lot of re-inventing, or improving on existing cloud-init utility functions in this datasource. While I appreciate the motivations here, I'm trying to review this section in more detail (as well as general netifaces imports) to see what we can re-use to avoid instrumenting network detection logic that needs to be maintained in two places in cloud-init. 


I need another day to get through this review, so I'll follow up tomorrow.

blackboxsw (commented):
Thanks @akutz for coming around with this set of commits. I'm really glad to see this materialize. I'm trying to digest the approaches you have here and rationalize keeping some of the network logic separate from cloudinit's existing logic. I'll have more concrete feedback tomorrow, but looks pretty good.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,780 @@
...
+                self.metadata["public-keys-data"]
+            )
+        if "public_keys_data" in self.metadata:
+            return sources.normalize_pubkey_data(
+                self.metadata["public_keys_data"]
+            )
+        elif "public-keys" in self.metadata:
+            return sources.normalize_pubkey_data(self.metadata["public-keys"])
+        elif "public_keys" in self.metadata:
+            return sources.normalize_pubkey_data(self.metadata["public_keys"])
```
blackboxsw, line 211: Worth just doing this?
```suggestion-
        for key_name in "public-keys-data", "public_keys_data", "public_keys", "public_keys":
            if key_name in self.metadata:
                return sources.normalize_pubkey_data(self.metadata[key_name])
            return sources.normalize_pubkey_data(
                self.metadata["public_keys_data"]
            )
        elif "public-keys" in self.metadata:
            return sources.normalize_pubkey_data(self.metadata["public-keys"])
        elif "public_keys" in self.metadata:
            return sources.normalize_pubkey_data(self.metadata["public_keys"])
```

cloudinit/sources/DataSourceVMware.py
akutz: I'm sure those are fine. I created a lot of this back in early 2018, so I'm not sure if those functions were available then. And if they were, it's likely I just did what I needed inside the DS to make it work without worrying too much about finding the helpers. Thanks for pointing this out.

cloudinit/sources/DataSourceVMware.py
akutz: FWIW, I've explained to others that I view CI and its datasources like an NES or SNES. CI is the console, and the datasources are the cartridges. Some datasources, like some cartridges, have extra functionality, i.e. specialized chips, that enhance the overall user experience. With respect to cloud-init, I always assumed this was on purpose.

But, you _are_ right that I've done some things that _could_ be in core CI. However, some of what I've done is somewhat unique to the VMware situation where networking derived from DHCP still should be present in the instance data. That and I felt that netifaces offered a more consistent and deterministic way to inspect the local host's networking stack that shell exec'ing to a series of ip commands.

cloudinit/sources/DataSourceVMware.py
akutz: Now you are just showing off :)

cloudinit/sources/DataSourceVMware.py
akutz: Sure thing!

cloudinit/sources/DataSourceVMware.py
akutz: Do you think we should implement `_get_subplatform` or just change `self.data_access_method` to `self.metadata_address`? It looks like the base DS will use that attribute for the sub platform if it exists.

akutz: Eh, I just did what you suggested as it was cleaner within the existing implementation.

akutz:
> Thanks @akutz for coming around with this set of commits. I'm really glad to see this materialize. I'm trying to digest the approaches you have here and rationalize keeping some of the network logic separate from cloudinit's existing logic. I'll have more concrete feedback tomorrow, but looks pretty good.

Hi @blackboxsw,

Thank you again for the feedback! I've addressed all of your comments, updated the docs, added two new tests for the `_get_subplatform` call, and verified tests pass. I'm still awaiting your review of the networking-specific bits, but hopefully the rebase should bring us closer to merging this :)

cloudinit/sources/DataSourceVMware.py
blackboxsw: Thanks for this repeated context here @akutz. I see I missed some of your earlier resolved responses on this PR.

@TheRealFalcon and I talked a bit about my minor misgivings about instrumenting network-specific helper functions within VMware datasource.

TLDR: would you be opposed to documenting a couple of TODO(...)  comments in the VMwareDatasource about long-term direction to try aligning with common cloud-init network utilities for the functions:
  - is_valid_ip_addr -> # TODO( extend cloudinit.net.is_ip_addr to ignore link_local/loopback or unspecified to drop this func)
  - 

I realize that my friction here is the following:
 * If we instrument duplicated (but better) detection logic down in VMware datasource, shared CI networking detection functions miss out on any improvements available via `netifaces`. Alternately, VMWare may miss out on corner cases that common CI network functions improve handling of and so we find ourselves trying to maintain two "separate but equal" solutions over time, which rarely happens unless we hit critical bugs that point us to the secondary implementations.
    * This example of dual-maintenance showed up a couple of times in long-term maintenance of the cloudinit/sources/helpers/vmware/imc modules which had a number of VMware-specific helper functions which ultimately do the same thing as cloud-init's common net functions

My misgivings on the network-related functions are more of a goal than a firm blocker to your work and can be deemed as a "nice to have" or TODO for subsequent iterations. I realize that you and other VMware devs will be best placed to iterate and maintain this code long-term given your access to VMware platforms and products far exceeds what upstream cloud-init devs have access to, so upstream committers are ill-placed to validate changes to this datasource functions. As a result I'd leave most of that validation work to whatever integration  testing runs you are able to perform on your side in our subsequent reviews.

That said, I think there are a couple of places that we can use common cloud-init utils if possible, the rest of the networking-related functions let's table for a subsequent iteration and just add some TODO breadcrumbs to get us thinking about that alignment in the future. I'll add those set of suggestions inline over the next hour and probably conslidate a patch with the diff of suggestions for your review.

akutz: > Thanks for this repeated context here @akutz. I see I missed some of your earlier resolved responses on this P
Not at all. I appreciate your time on this!

> TLDR: would you be opposed to documenting a couple of TODO(...) comments in the VMwareDatasource about long-term direction to try aligning with common cloud-init network utilities
Of course not, happy to do it.

Your concerns are reasonable, and honestly, if you decided to block this DS from merging until we sorted out everything to exactly the way you would like it, I would understand.

My primary goals are:

* Stop maintaining this DS in a separate repo
* Provide backwards compatibility for the user experience, even as we update the internals
* Make it easier for others to have eyes on it

I sense you are also aligned with these goals, so I'm fine with whatever you want to do here. Obviously the sooner it's merged the sooner more people can work to address the issues, but again, that's totally up to y'all. I just appreciate all of your help on this (and @TheRealFalcon 's help as well)!

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+            keys_to_redact = self.metadata[REDACT]
+        elif CLEANUP_GUESTINFO in self.metadata:
+            # This is for backwards compatibility.
+            keys_to_redact = self.metadata[CLEANUP_GUESTINFO]
+
+        if self.data_access_method == DATA_ACCESS_METHOD_GUESTINFO:
+            guestinfo_redact_keys(keys_to_redact, self.vmware_rpctool)
+
+
+def decode(key, enc_type, data):
```
blackboxsw, line 237: This really has a couple entry points that could be satisfied by the existing util.b64d, decode_binary and util.decomp_gzip. Please use these common util functions where possible to meet the needs of your decode function.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+    of its trailing whitespace, is empty or equal to GUESTINFO_EMPTY_YAML_VAL.
+
+    The return value is always a string, regardless of whether the input is
+    a bytes class or a string.
+    """
+
+    # If the provided value is a bytes class, convert it to a string to
+    # simplify the rest of this function's logic.
+    if isinstance(val, bytes):
+        val = val.decode()
```
blackboxsw, line 286: probaly worth the common util func here
```suggestion
    val = util.decode_binary(val)
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+        if not guestinfo_set_value(
+            key, GUESTINFO_EMPTY_YAML_VAL, vmware_rpctool
+        ):
+            LOG.error("failed to clear %s", key_name)
+        LOG.info("clearing %s.encoding", key_name)
+        if not guestinfo_set_value(key + ".encoding", "", vmware_rpctool):
+            LOG.error("failed to clear %s.encoding", key_name)
+
+
+def load(data):
```
blackboxsw, line 466: for clarify/bike-shed how about
```suggestion
def load_json_or_yaml(data):
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+        data["network"] = network
+
+    return data
+
+
+def get_datasource_list(depends):
+    """
+    Return a list of data sources that match this set of dependencies
+    """
+    return [DataSourceVMware]
```
blackboxsw, line 516: To avoid accidentally injecting the VMware datasource into a not-yet-defined cloud-init boot stage, can we adhere to the common practice of defining the mapping of authorized boot stages where DataSourceVmware can be discovered?
This way we don't find out later if we add a stage before init-local or init that we are spending time trying to detect VMware where it can't possibly run.
```suggestion
# Used to match classes to dependencies                                         
datasources = [                                                                 
    (DataSourceVMware, (sources.DEP_FILESYSTEM,)),  # Run at init-local       
    (DataSourceVMware, (sources.DEP_FILESYSTEM, sources.DEP_NETWORK)),             
]                                                                               
                                                                                
                                                                                
def get_datasource_list(depends):                                               
    """
    Return a list of data sources that match this set of dependencies
    """
    return sources.list_from_depends(depends, datasources)   
```

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+
+        network = host_info.get("network") or {}
+        interfaces = network.get("interfaces") or {}
+        by_ipv4 = interfaces.get("by-ipv4") or {}
+        by_ipv6 = interfaces.get("by-ipv6") or {}
+
+        if wait_on_ipv4:
+            ipv4_ready = len(by_ipv4) > 0 if by_ipv4 else False
+            if not ipv4_ready:
+                LOG.info("ipv4 not ready")
```
blackboxsw, line 778: Generally we want to avoid logging level info unless there is something truly important for a human to read. Generally throughout this datasource it seems that most of this information is likely just a debug-level message.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+                host_info = None
+
+        if wait_on_ipv6:
+            ipv6_ready = len(by_ipv6) > 0 if by_ipv6 else False
+            if not ipv6_ready:
+                LOG.info("ipv6 not ready")
+                host_info = None
+
+        if host_info is None:
+            LOG.info("waiting on network")
```
blackboxsw, line 788: Do we really want to log two messages each second while waiting on network? Maybe any entry log and exit success log would suffice?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+
+from cloudinit import log as logging
+from cloudinit import sources
+from cloudinit import util
+from cloudinit.subp import subp, ProcessExecutionError
+
+import netifaces
+
+
+PRODUC_UUID_FILE_PATH = "/sys/class/dmi/id/product_uuid"
```
blackboxsw, line 35:
```suggestion
PRODUCT_UUID_FILE_PATH = "/sys/class/dmi/id/product_uuid"
```

cloudinit/sources/DataSourceVMware.py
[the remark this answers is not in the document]
blackboxsw: There is a also a question of whether it might be better to use util.get_hostname() instead as well (it uses socket.gethostname() instead of talking the IMDS)

cloudinit/sources/DataSourceVMware.py
```diff
@@ -0,0 +1,814 @@
...
+        addr = ipaddress.ip_address(unicode(val))
+    except Exception:
+        return None
+
+    if addr.is_link_local or addr.is_loopback or addr.is_unspecified:
+        return False
+    return True
+
+
+def get_host_info():
```
blackboxsw, line 656: The whole structure here is useful networking information albeit in a slightly different format than netinfo.debug_info(). Long-term I'd like to see how we can align cloudinit common functions to support this type of network digest, and maybe that means ultimately moving cloud-init proper functions to use netifaces directly now that we are introducing an explicit dependency on it.

blackboxsw (changes requested):
Thanks for your patience @akutz 
Ultimately, please take what you think is reasonable, long-term I'd like to work out those TODOs to better cloud-init proper and share more functions across all datasourcees.

I've tried to consolidate a number of comments I had for your review into this diff
```diff
diff --git a/cloudinit/sources/DataSourceVMware.py b/cloudinit/sources/DataSourceVMware.py
index f845fe8f..3c33cea9 100644
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -13,7 +13,6 @@ A cloud init datasource for VMware.
 
 import base64
 import collections
-import collections.abc
 import copy
 from distutils.spawn import find_executable
 import ipaddress
@@ -32,7 +31,7 @@ from cloudinit.subp import subp, ProcessExecutionError
 import netifaces
 
 
-PRODUC_UUID_FILE_PATH = "/sys/class/dmi/id/product_uuid"
+PRODUCT_UUID_FILE_PATH = "/sys/class/dmi/id/product_uuid"
 
 LOG = logging.getLogger(__name__)
 NOVAL = "No value found"
@@ -130,7 +129,7 @@ class DataSourceVMware(sources.DataSource):
         LOG.info("using data access method %s", self._get_subplatform())
 
         # Get the metadata.
-        self.metadata = process_metadata(load(md))
+        self.metadata = process_metadata(load_json_or_yaml(md))
 
         # Get the user data.
         self.userdata_raw = ud
@@ -206,7 +205,7 @@ class DataSourceVMware(sources.DataSource):
         # read the file /sys/class/dmi/id/product_uuid for the instance ID.
         if self.metadata and "instance-id" in self.metadata:
             return self.metadata["instance-id"]
-        with open(PRODUC_UUID_FILE_PATH, "r") as id_file:
+        with open(PRODUCT_UUID_FILE_PATH, "r") as id_file:
             self.metadata["instance-id"] = str(id_file.read()).rstrip().lower()
             return self.metadata["instance-id"]
 
@@ -258,17 +257,16 @@ def decode(key, enc_type, data):
     raw_data = None
     if enc_type in ["gzip+base64", "gz+b64"]:
         LOG.debug("Decoding %s format %s", enc_type, key)
-        raw_data = zlib.decompress(base64.b64decode(data), zlib.MAX_WBITS | 16)
+        # would util.decomp_gzip(util.b64d(data)) work here instead?
+        raw_data = zlib.decompress(util.b64d(data), zlib.MAX_WBITS | 16)
     elif enc_type in ["base64", "b64"]:
         LOG.debug("Decoding %s format %s", enc_type, key)
-        raw_data = base64.b64decode(data)
+        raw_data = util.b64d(data)
     else:
         LOG.debug("Plain-text data %s", key)
         raw_data = data
 
-    if isinstance(raw_data, bytes):
-        return raw_data.decode("utf-8")
-    return raw_data
+    return util.decode_binary(raw_data)
 
 
 def get_none_if_empty_val(val):
@@ -282,9 +280,7 @@ def get_none_if_empty_val(val):
 
     # If the provided value is a bytes class, convert it to a string to
     # simplify the rest of this function's logic.
-    if isinstance(val, bytes):
-        val = val.decode()
-
+    val = util.decode_binary(val)
     val = val.rstrip()
     if len(val) == 0 or val == GUESTINFO_EMPTY_YAML_VAL:
         return None
@@ -463,7 +459,7 @@ def guestinfo_redact_keys(keys, vmware_rpctool=VMWARE_RPCTOOL):
             LOG.error("failed to clear %s.encoding", key_name)
 
 
-def load(data):
+def load_json_or_yaml(data):
     """
     load first attempts to unmarshal the provided data as JSON, and if
     that fails then attempts to unmarshal the data as YAML. If data is
@@ -500,7 +496,7 @@ def process_metadata(data):
             LOG.debug("network data to be decoded %s", network)
             dec_net = decode("metadata.network", network_enc, network)
             network = {
-                "config": load(dec_net),
+                "config": load_json_or_yaml(dec_net),
             }
 
         LOG.debug("network data %s", network)
@@ -509,11 +505,18 @@ def process_metadata(data):
     return data
 
 
+# Used to match classes to dependencies
+datasources = [
+    (DataSourceVMware, (sources.DEP_FILESYSTEM,)),  # Run at init-local
+    (DataSourceVMware, (sources.DEP_FILESYSTEM, sources.DEP_NETWORK)),
+]
+
+
 def get_datasource_list(depends):
     """
     Return a list of data sources that match this set of dependencies
     """
-    return [DataSourceVMware]
+    return sources.list_from_depends(depends, datasources)
 
 
 def get_default_ip_addrs():
@@ -524,6 +527,8 @@ def get_default_ip_addrs():
     addresses associated with the device used by the default route for a given
     address.
     """
+    # akutz, please site the shortcoming on MacOS for both iproute2mac and netstat to inform future-us why we have these functions depending on netifaces
+    # TODO(promote and use netifaces in cloudinit.net* modules)
     gateways = netifaces.gateways()
     if "default" not in gateways:
         return None, None
@@ -612,9 +617,11 @@ def getfqdn(name=""):
     """Get fully qualified domain name from name.
     An empty argument is interpreted as meaning the local host.
     """
+    # TODO(may want to promote this function to util.getfqdn)
+    # TODO(may want to extend util.get_hostname to accept fqdn=True param)
     name = name.strip()
     if not name or name == "0.0.0.0":
-        name = socket.gethostname()
+        name = util.get_hostname()
     try:
         addrs = socket.getaddrinfo(
             name, None, 0, socket.SOCK_DGRAM, 0, socket.AI_CANONNAME
@@ -634,6 +641,8 @@ def is_valid_ip_addr(val):
     Returns false if the address is loopback, link local or unspecified;
     otherwise true is returned.
     """
+    # TODO(extend cloudinit.net.is_ip_addr exclude link_local/loopback etc)
+    # TODO(migrate to use cloudinit.net.is_ip_addr)#
 
     addr = None
     try:
@@ -657,7 +666,7 @@ def get_host_info():
     """
     Returns host information such as the host name and network interfaces.
     """
-
+    # TODO(look to promote netifices use up in cloud-init netinfo funcs)
     host_info = {
         "network": {
             "interfaces": {
@@ -667,8 +676,7 @@ def get_host_info():
             },
         },
     }
-
-    hostname = getfqdn(socket.gethostname())
+    hostname = getfqdn(util.get_hostname())
     if hostname:
         host_info["hostname"] = hostname
         host_info["local-hostname"] = hostname
@@ -765,6 +773,9 @@ def wait_on_network(metadata):
     # Get information about the host.
     host_info = None
     while host_info is None:
+        # This loop + sleep results in two logs every second while waiting
+        # for either ipv4 or ipv6 up. Do we really need to log each iteration
+        # or can we log once and log on successful exit?
         host_info = get_host_info()
 
         network = host_info.get("network") or {}
@@ -775,19 +786,19 @@ def wait_on_network(metadata):
         if wait_on_ipv4:
             ipv4_ready = len(by_ipv4) > 0 if by_ipv4 else False
             if not ipv4_ready:
-                LOG.info("ipv4 not ready")
+                LOG.debug("ipv4 not ready")
                 host_info = None
 
         if wait_on_ipv6:
             ipv6_ready = len(by_ipv6) > 0 if by_ipv6 else False
             if not ipv6_ready:
-                LOG.info("ipv6 not ready")
+                LOG.debug("ipv6 not ready")
                 host_info = None
 
         if host_info is None:
-            LOG.info("waiting on network")
+            LOG.debug("waiting on network")
             time.sleep(1)
-
+    LOG.debug("waiting on network complete")
     return host_info
 
 
diff --git a/tests/unittests/test_datasource/test_vmware.py b/tests/unittests/test_datasource/test_vmware.py
index 6da78b5d..8122d253 100644
--- a/tests/unittests/test_datasource/test_vmware.py
+++ b/tests/unittests/test_datasource/test_vmware.py
@@ -20,7 +20,7 @@ import os
 
 PRODUCT_UUID = "82343CED-E4C7-423B-8F6B-0D34D19067AB"
 REROOT_FILES = {
-    DataSourceVMware.PRODUC_UUID_FILE_PATH: PRODUCT_UUID,
+    DataSourceVMware.PRODUCT_UUID_FILE_PATH: PRODUCT_UUID,
 }
 
 VMW_MULTIPLE_KEYS = [

```

cloudinit/sources/DataSourceVMware.py
akutz: FWIW, the structure is to do two things:

1. Support the existing AWS JINJA queries
2. Make it super simple for a tokenized lookup of any known network info by address family, IP, or MAC. Yes, there are ways to do the same with less duplicate info and multiple queries, but I wanted to lessen the burden on the people actually writing the queries.

cloudinit/sources/DataSourceVMware.py
akutz: Probably not? I was a little more aggressive with logging originally to help debug issues while developing the DS. I'll turn down the verbosity and/or clean things up a bit.

Good catch!

cloudinit/sources/DataSourceVMware.py
akutz: So I wondered about that. I took the approach that since the DS didn't need the network to get its metadata, that there was no difference to the DS as to why it was loaded.

The examples I saw such as EC2 used the FS level in order to bring up the ephemeral DHCP net to retrieve the metadata on the 169 address.

cloudinit/sources/DataSourceVMware.py
akutz: Sure. I did this in part for the logging about the keys being used, but as I adjust logging maybe this is no longer an issue and I can switch.

akutz:
Hi @blackboxsw,

I've updated the PR with your recommended changes. PTAL at your convenience. Thank you again for your help!

akutz:
Rebased onto main

requirements.txt
```diff
@@ -32,3 +32,12 @@ jsonpatch
...
 jsonschema
+
+# Used by DataSourceVMware to inspect the host's network configuration during
+# the "setup()" function.
+#
+# This allows a host that uses DHCP to bring up the network during BootLocal
+# and still participate in instance-data by gathering the network in detail at
+# runtime and merge that information into the metadata and repersist that to
+# disk.
+netifaces>=0.10.9
```
blackboxsw, line 43: One last question here: do we need something specific in version 0.10.9 for support here?
I ask because the versions of netifaces in Ubuntu 18.04 and 20.04 only have netifaces 0.10.4. This would block our ability to release back to earlier version of Ubuntu without additional packaging and publishing work around python3-netifaces.

blackboxsw: Per @akutz in IRC conversation:
Yeah, it looks like we should stick with 0.10.9. At least 0.10.5! https://github.com/al45tair/netifaces/blob/master/CHANGELOG
In 0.10.5 they fixed "Respect interface priorities when determining default gateway"
That's a huge thing for the DS
I recommend focusing post-merge on a plan to move away from netifaces or how to backport specific things?

<•blackboxsw> Chad Smith  +1 right
 
Yes I think we can definitely iterate on this netifaces dependency, the fact that Ubuntu 18.04 and 20.04 both have 10.4 probably prevents us from depending on netifaces for all datasources. We can work to make this a conditional dependency that only exists in Hirsute ++. Or better yet solve the use-cases need by VMware in the cloud-init proper network utility modules/functions so that VMware doesn't need this external dependency.

akutz: This was discussed on IRC, and it was decided to leave as-is for now since 0.10.5 and other more recent releases fix issues with determining default gateways, something on which this DS depends. This may mean that this DS cannot be included in older Ubuntu releases, but there is a post-merge effort to try and adjust/reduce/remove the dependency on netifaces anyway.

blackboxsw (approved):
At long last. Thanks much @akutz for working through these review comments. This looks really good and I think we can iterate on any remaining TODOs and or netifaces concerns prior to releasing VMware datasource back to earlier distros with netifaces < 0.10.5.

prologic:
Congrats to all involved! This is really awesome news! 🎉

akutz:
> Congrats to all involved! This is really awesome news! 🎉

Thanks again for all your help @prologic! I wish I could have just preserved all the original commits, but in lieu of that I definitely want to make sure all the contributors are recognized in this PR!

akutz:
> At long last. Thanks much @akutz for working through these review comments. This looks really good and I think we can iterate on any remaining TODOs and or netifaces concerns prior to releasing VMware datasource back to earlier distros with netifaces < 0.10.5.

Re the version of netifaces. FWIW, we currently build from Ubuntu 20.04 in the Kube image builder using netifaces 0.10.5. We just install via pip3 instead of the system package. Is this something that could be considered?

mitechie:
The issue with a non-system package is that we'd like to have the
software involved tracked via the Ubuntu Security Team for any known
CVE issues. They track things we'd have as the system packages but
it's not as trivial for things that come in through third-party
installers.

akutz:
> The issue with a non-system package is that we'd like to have the
> software involved tracked via the Ubuntu Security Team for any known
> CVE issues. They track things we'd have as the system packages but
> it's not as trivial for things that come in through third-party
> installers.
> 
> On Tue, Aug 10, 2021 at 12:51 AM Andrew Kutz ***@***.***> wrote:
> >
> > At long last. Thanks much @akutz for working through these review comments. This looks really good and I think we can iterate on any remaining TODOs and or netifaces concerns prior to releasing VMware datasource back to earlier distros with netifaces < 0.10.5.
> >
> > Re the version of netifaces. FWIW, we currently build from Ubuntu 20.04 in the Kube image builder using netifaces 0.10.5. We just install via pip3 instead of the system package. Is this something that could be considered?
> >
> > —
> > You are receiving this because you are subscribed to this thread.
> > Reply to this email directly, view it on GitHub, or unsubscribe.
> > Triage notifications on the go with GitHub Mobile for iOS or Android.
> 

Ah, makes perfect sense. Thanks Rick!

akutz:
Hi @mitechie / @blackboxsw,

I'm looking at the python3-netifaces 0.10.4 and 0.10.9 packages and noticed the only difference is their Python 3 dependency. And I know that 0.10.9 works on at least Python >=3.6.

image

image

My point is that the newer 0.10.9 package should be binary compatible with Ubuntu 20.04. Any thoughts on how we can use this to our advantage?

akutz:
@blackboxsw I'm going to validate netifaces 0.10.4 on several standard network configs. If it works as expected then I will downgrade the dependency to 0.10.4 so we can be on Ubuntu 20.04. I forgot it was the most recent LTS, so it's not unimportant to be there.

PR #955, opened 2021-07-29
testing: port remaining cloud tests to integration testing framework (SC-191)

TheRealFalcon:
## Proposed Commit Message

```
testing: port remaining cloud tests to integration testing framework

This should enable us to remove the cloud-tests entirely.
```

## Additional Context
Actually removing cloud tests can come separately.

## Test Steps
pytest tests/integration_tests

## Checklist:


tests/integration_tests/modules/test_combined.py
```diff
@@ -0,0 +1,175 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+"""A set of somewhat unrelated tests that can be combined into a single
```
blackboxsw, line 2: +1 on consolidation here of completely unrelated tests. If we find that the interaction of such tests cost us much in terms of lost signal or unexpected cross-module interaction we can separate them in the future. Our integration tests are expensive to setup and to run as far as time. so it's good to be as broad as possible with our tests.

blackboxsw (approved):
+1 LGTM.

PR #958, opened 2021-08-03
distros/photon.py: refactor hostname handling

sshedi:

### DIFF
--- a/cloudinit/distros/photon.py
+++ b/cloudinit/distros/photon.py
@@ -91,15 +91,6 @@ class Distro(distros.Distro):
         # self.update_package_sources()
         self.package_command('install', pkgs=pkglist)
 
-    def _bring_up_interfaces(self, device_names):
-        cmd = ['systemctl', 'restart', 'systemd-networkd', 'systemd-resolved']
-        LOG.debug('Attempting to run bring up interfaces using command %s',
-                  cmd)
-        ret, _out, err = self.exec_cmd(cmd)
-        if ret:
-            LOG.error('Failed to bringup interfaces: %s', err)
-        return ret
-
     def _write_hostname(self, hostname, filename):
         if filename and filename.endswith('/previous-hostname'):
             util.write_file(filename, hostname)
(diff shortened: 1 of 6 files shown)


sshedi:
Hi @TheRealFalcon - can you please review this change? You suggested to use this in my very first PR, I took a part of hostname related chnages from RHEL distro code. Now I have tried to make it better.

cloudinit/distros/photon.py
```diff
@@ -99,26 +98,23 @@ def _bring_up_interfaces(self, device_names):
...
-            pass
-        if not conf:
-            conf = HostnameConf('')
-        conf.set_hostname(hostname)
-        util.write_file(filename, str(conf), mode=0o644)
+        if filename.endswith('/previous-hostname'):
+            util.write_file(filename, hostname)
+        else:
+            _ret, _out, _err = self.exec_cmd(['hostnamectl', 'set-hostname',
+                                             str(hostname)])
```
TheRealFalcon, line 105:
```suggestion
            self.exec_cmd(['hostnamectl', 'set-hostname', str(hostname)])
```
If we're not using the results, then we shouldn't store them.

cloudinit/distros/photon.py
```diff
@@ -99,26 +98,23 @@ def _bring_up_interfaces(self, device_names):
...
 
     def _read_hostname(self, filename, default=None):
-        _ret, out, _err = self.exec_cmd(['hostname'])
+        if filename.endswith('/previous-hostname'):
+            return util.load_file(filename).strip()
 
+        _ret, out, _err = self.exec_cmd(['hostname', '-f'], capture=True)
+        if out:
+            out = out.strip()
         return out if out else default
```
TheRealFalcon, line 118:
```suggestion
        return out.strip() if out else default
```
Just a minor tweak.

TheRealFalcon (approved):
Looks good to me!

I added two small minor suggestions inline, but they won't block merge.

cloudinit/distros/photon.py
sshedi: Agree.

cloudinit/distros/photon.py
sshedi: Thanks.

sshedi:
@TheRealFalcon - I have added networkd activator in this PR now. Please take a look.
If you want me to send a different PR for network activator changes, I will do so.

TheRealFalcon (approved):
Looks good, thanks!

sshedi:
Welcome and thank you too.

PR #961, opened 2021-08-05
Azure: Limit polling network metadata on connection errors

aswinrajamannar:
## Proposed Commit Message

```
Azure: Limit polling network metadata on connection errors
```
## Description
This is a follow up to #878 where we currently limit polling on network metadata to 10 retries if the calls fail due to timeout errors. After enabling the feature in Azure, we noticed that if secondary nic is attached first, the calls could also fail with network unreachable ConnectionError. In these cases as well, we shouldn't poll indefinitely, and instead move on to listen for more nic attach events.

## Test Steps
- Deploy a multi-nic VM in Azure that goes through the nic hot-attach code path. In a few cases where the secondary nic is attached first, the VM deployment might fail if this fix is missing.

## Checklist:


cloudinit/sources/DataSourceAzure.py
```diff
@@ -1011,9 +1014,12 @@ def network_metadata_exc_cb(msg, exc):
                                             (msg, exc.cause, exc.code),
                                             logger_func=LOG.error)
 
-            if exc.cause and isinstance(exc.cause, requests.Timeout):
-                metadata_timeout_count = metadata_timeout_count + 1
-                return (metadata_timeout_count <= 10)
+            # Retry up to a certain limit for both timeout and network
+            # unreachable errors.
+            if exc.cause and (isinstance(exc.cause, requests.Timeout) or
+                              isinstance(exc.cause, requests.ConnectionError)):
```
TheRealFalcon, line 1020:
```suggestion
            if exc.cause and isinstance(
                exc.cause, (requests.Timeout, requests.ConnectionError)
            ):
```

tests/unittests/test_datasource/test_azure.py
```diff
@@ -2854,10 +2855,17 @@ def network_metadata_ret(ifname, retries, type, exc_cb, infinite):
...
+                    # We are expected to retry for a certain period for both
+                    # timeout errors and network unreachable errors.
+                    if _ < 5:
+                        cause = requests.Timeout('Fake connection timeout')
+                        error = url_helper.UrlError(cause=cause)
+                        eth1Retries.append(exc_cb("Connection timeout", error))
+                    else:
+                        cause = requests.ConnectionError('Network Unreachable')
+                        error = url_helper.UrlError(cause=cause)
+                        eth1Retries.append(exc_cb("Connection timeout", error))
```
TheRealFalcon, line 2868:
```suggestion
                    if _ < 5:
                        cause = requests.Timeout('Fake connection timeout')
                    else:
                        cause = requests.ConnectionError('Network Unreachable')
                    error = url_helper.UrlError(cause=cause)
                    eth1Retries.append(exc_cb("Connection timeout", error))
```

TheRealFalcon (commented):
Looks good to me! I left two nit suggestions inline.

TheRealFalcon (approved):

aswinrajamannar:
> Looks good to me! I left two nit suggestions inline.

Thanks. Committed the suggestions.

PR #967, opened 2021-08-10
Add VMware to default datasource_list

TheRealFalcon:
Impish packaging change. If this is correct I'll do hirsute/focal/bionic next.


blackboxsw (approved):
+1 LGTM! Let's refelect to Bionic and Focal

PR #969, opened 2021-08-11
cc_resolv_conf: fix typos

sshedi:


sshedi:
@TheRealFalcon can we add a test to this? I have no idea how it sneaked in.

cloudinit/config/cc_resolv_conf.py
```diff
@@ -108,18 +108,19 @@ def handle(name, cfg, cloud, log, _args):
 
     if "resolv_conf" not in cfg:
         log.warning("manage_resolv_conf True but no parameters provided!")
+        return
```
sshedi, line 111: should return, `resolv_conf` key is used in generate_resolv_conf call

cloudinit/config/cc_resolv_conf.py
```diff
@@ -108,18 +108,19 @@ def handle(name, cfg, cloud, log, _args):
...
 
     try:
         template_fn = cloud.get_template_filename(
-            RESOLVE_CONFIG_TEMPLATE_MAP[cloud.distro.resolv_conf_fn])
+            RESOLVE_CONFIG_TEMPLATE_MAP[cloud.distro.resolve_conf_fn])
```
sshedi, line 115: Another typo. I don't know how, luckily I saw this file for no reason. And found these blunders

TheRealFalcon (approved):
Ooph, thanks for the changes and the test here. Some of our modules still lack test coverage, and this particular one makes that very apparent.

One nit on the added test: It would be easier to read and help fix any potential failures if we had 5 separate tests, rather than checking the 5 conditions in one larger test. I won't block the PR on it, but it'd be a helpful small change.

Before I merge this, does [photon](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/photon.py#L28) need an update too? On the line I linked to, I noticed the key is `resolv_conf_fn`, but [networkd.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/networkd.py#L75) is looking for `resolved_conf_fn`. Should that be updated too?

sshedi:
> Ooph, thanks for the changes and the test here. Some of our modules still lack test coverage, and this particular one makes that very apparent.
> 
> One nit on the added test: It would be easier to read and help fix any potential failures if we had 5 separate tests, rather than checking the 5 conditions in one larger test. I won't block the PR on it, but it'd be a helpful small change.
> 
> Before I merge this, does [photon](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/photon.py#L28) need an update too? On the line I linked to, I noticed the key is `resolv_conf_fn`, but [networkd.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/networkd.py#L75) is looking for `resolved_conf_fn`. Should that be updated too?

Oops. I will fix it. Also break the test into separate tests. Need sometime, I'm caught up with some other task. Is that okay?

TheRealFalcon:
Yep, no rush!

PR #970, opened 2021-08-11
 Update dscheck_VMware's rpctool check

sshedi:
Other minor tweaks.

## Proposed Commit Message
Update dscheck_VMware's rpctool check

```
Update dscheck_VMware's rpctool check

This patch updates the dscheck_VMware function's use of "vmware-rpctool".

When checking to see if a "guestinfo" property is set.
Because a successful exit code can occur even if there is an empty
string returned, it is possible that the VMware datasource will be
loaded as a false-positive. This patch ensures that in addition to
validating the exit code, the emitted output is also examined to ensure
a non-empty value is returned by rpctool before returning "${DS_FOUND}"
from "dscheck_VMware()".

Other minor tweaks in ds-identify script.
```

## Checklist:

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -141,6 +141,7 @@ error() {
     debug 0 "$@"
     stderr "$@"
 }
+
 warn() {
     set -- "WARN:" "$@"
     debug 0 "$@"
@@ -344,7 +345,6 @@ geom_label_status_as() {
     return $ret
 }
 
-
 read_fs_info_freebsd() {
     local oifs="$IFS" line="" delim=","
     local ret=0 labels="" dev="" label="" ftype="" isodevs=""
@@ -404,7 +404,6 @@ cached() {
     [ -n "$1" ] && _RET="$1" && return || return 1
 }
 
-
 detect_virt() {
     local virt="${UNAVAILABLE}" r="" out=""
     if [ -d /run/systemd ]; then
@@ -450,7 +449,7 @@ detect_virt() {
 read_virt() {
     cached "$DI_VIRT" && return 0
     detect_virt
-    DI_VIRT=${_RET}
+    DI_VIRT="$(echo "${_RET}" | tr '[:upper:]' '[:lower:]')"
 }
 
 is_container() {
@@ -1384,16 +1383,20 @@ vmware_has_rpctool() {
     command -v vmware-rpctool >/dev/null 2>&1
 }
 
+vmware_rpctool_guestinfo() {
+    vmware-rpctool "info-get guestinfo.${1}" 2>/dev/null | grep "[[:alnum:]]"
+}
+
 vmware_rpctool_guestinfo_metadata() {
-    vmware-rpctool "info-get guestinfo.metadata"
+    vmware_rpctool_guestinfo "metadata"
 }
 
 vmware_rpctool_guestinfo_userdata() {
-    vmware-rpctool "info-get guestinfo.userdata"
+    vmware_rpctool_guestinfo "userdata"
 }
 
 vmware_rpctool_guestinfo_vendordata() {
-    vmware-rpctool "info-get guestinfo.vendordata"
+    vmware_rpctool_guestinfo "vendordata"
 }
 
 dscheck_VMware() {


tools/ds-identify
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
sshedi, line 1: @TheRealFalcon @blackboxsw 
This script is using `local` keyword in almost all functions & `local` is not POSIX compliant (shellcheck warns the same).
Can we use bash by default? Ubuntu links sh->dash and I'm unsure about other systems.
I'm assuming that bash is available in all distros by default.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
 
     # Check to see if running in a container and the VMware
     # datasource is configured via environment variables.
-    if vmware_has_envvar_vmx_guestinfo; then
-        if vmware_has_envvar_vmx_guestinfo_metadata || \
-            vmware_has_envvar_vmx_guestinfo_userdata || \
-            vmware_has_envvar_vmx_guestinfo_vendordata; then
+    if [ -n "${VMX_GUESTINFO}" ]; then
```
sshedi, line 1379: @akutz - do we need to use a new function for each of these checks? I thought otherwise and kept it simple.
If you think it's needed, please let me know. It will be good to know the reason as well.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
...
     # Activate the VMware datasource only if any of the fields used
     # by the datasource are present in the guestinfo table.
-    if { vmware_rpctool_guestinfo_metadata || \
-         vmware_rpctool_guestinfo_userdata || \
-         vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
-        return "${DS_FOUND}"
-    fi
+    local tmp=''
+    for tmp in "metadata" "userdata" "vendordata"; do
+      tmp="$(vmware-rpctool "info-get guestinfo.$tmp")"
```
sshedi, line 1397: This is the tricky part. If we do `vmware-rpctool "info-set guestinfo.metadata ..."` once, even after doing `vmware-rpctool "info-set guestinfo.metadata  "` it won't go aways and `info-get guestinfo.userdata` returns 0 with empty string in return.

This remains same until poweroff. (reboot won't make it go).

If a user decides to use `datasource_list: ['VMware', 'NoCloud', 'OVF', 'ConfigDrive', 'OpenStack']`
VMware DS is always found but with no data and cloud-init stops searching for other DS because VMware is already found. So, we need to check for alpha numeric string in return value.

tools/ds-identify
```diff
@@ -450,7 +449,7 @@ detect_virt() {
 read_virt() {
     cached "$DI_VIRT" && return 0
     detect_virt
-    DI_VIRT=${_RET}
+    DI_VIRT="$(echo "${_RET}" | tr '[:upper:]' '[:lower:]')"
```
sshedi, line 452: Just playing safe here, ensuring that DI_VIRT is lower case.

akutz:
Thank you for doing this @sshedi. However, can you please provide a more verbose and meaningful commit message?

tools/ds-identify
akutz: I would focus this PR on improving `dscheck_VMware` and separate general script improvements into a separate change.

tools/ds-identify
akutz: If the interpreter is BASH then the following should be possible:

```bash
DI_VIRT="${_RET,,}"
```

tools/ds-identify
sshedi: I think I got my answer by CI test results. It will be nice to know your clarification.

tools/ds-identify
akutz: The new functions make it possible to test in [`test_ds-identify.py`](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_ds_identify.py). Please change everything back to the way it was. Thanks!

tools/ds-identify
akutz: Please change `$tmp` to `${tmp}` to be consistent with the full pattern I've used elsewhere.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
...
+    command -v vmware-rpctool >/dev/null 2>&1 || return ${DS_NOT_FOUND}
 
     # Activate the VMware datasource only if any of the fields used
     # by the datasource are present in the guestinfo table.
-    if { vmware_rpctool_guestinfo_metadata || \
-         vmware_rpctool_guestinfo_userdata || \
-         vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
-        return "${DS_FOUND}"
-    fi
+    local tmp=''
```
akutz, line 1395: Please use a more descriptive variable name and certainly not one that is overloaded like `tmp`

akutz (changes requested):

sshedi:
> Thank you for doing this @sshedi. However, can you please provide a more verbose and meaningful commit message?

Okay, I will give a brief info in my commit message. If you have anything in mind, feel free to share :) I struggle a bit while writing commit messages and assigning names to variables.

tools/ds-identify
akutz: Also, these were net-new and I'm not sure what the thinking was removing them. In the future reaching out on Slack or async is probably a better path forward before making so many reactive changes. At least have a conversation first. I'm nothing if not talkative :)

akutz: Also, FWIW, the way you changed it was the way I had it in the original repo. As I indicated, the new model was to support the testing framework for ds-identify.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
...
+    for tmp in "metadata" "userdata" "vendordata"; do
+      tmp="$(vmware-rpctool "info-get guestinfo.$tmp")"
+      if [ $? -eq 0 ]; then
+          echo "$tmp" | grep -q "[[:alnum:]]" && return ${DS_FOUND}
+      fi
+    done
 
-    return "${DS_NOT_FOUND}"
+    return ${DS_NOT_FOUND}
```
akutz, line 1403: Please change this to `"${DS_FOUND}"` to be consistent with how other values are used. Quoting variables in BASH is preferred unless otherwise required.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
...
-    if { vmware_rpctool_guestinfo_metadata || \
-         vmware_rpctool_guestinfo_userdata || \
-         vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
-        return "${DS_FOUND}"
-    fi
+    local tmp=''
+    for tmp in "metadata" "userdata" "vendordata"; do
+      tmp="$(vmware-rpctool "info-get guestinfo.$tmp")"
+      if [ $? -eq 0 ]; then
+          echo "$tmp" | grep -q "[[:alnum:]]" && return ${DS_FOUND}
```
akutz, line 1399: Please change this to `"${DS_FOUND}"` to be consistent with how other values are used. Quoting variables in BASH is preferred unless otherwise required.

tools/ds-identify
```diff
@@ -1409,33 +1376,31 @@ dscheck_VMware() {
...
+           [ -n "${VMX_GUESTINFO_VENDORDATA}" ]; then
             return "${DS_FOUND}"
         fi
     fi
 
     # Do not proceed unless the detected platform is VMware.
-    if [ ! "${DI_VIRT}" = "vmware" ]; then
-        return "${DS_NOT_FOUND}"
-    fi
+    [ "${DI_VIRT}" = "vmware" ] || return ${DS_NOT_FOUND}
```
akutz, line 1388: Please change this back as I wanted the logic to be more clear instead of the syntactic sugar approach. Thanks.

akutz (changes requested):

akutz:
> > Thank you for doing this @sshedi. However, can you please provide a more verbose and meaningful commit message?
> 
> Okay, I will give a brief info in my commit message. If you have anything in mind, feel free to share :) I struggle a bit while writing commit messages and assigning names to variables.

Sure. Well, on the whole of it:

* The cleanup to be BASH specific shouldn't be part of this PR IMO. That should be an entirely separate PR.
* The only change I see that _should_ likely occur is with respect to the way the guest info data is checked and return codes. So I'd redact all other changes except that and make the commit message:

```
Update dscheck_VMware's rpctool check

This patch updates the dscheck_VMware function's use of "vmware-rpctool"
when checking to see if a "guestinfo" property is set. Because a successful
exit code can occur even if there is an empty string returned, it is possible
that the VMware datasource will be loaded as a false-positive. This patch
ensures that in addition to validating the exit code, the emitted output is
also examined to ensure a non-empty value is returned by rpctool.
```

tools/ds-identify
sshedi: DS_FOUND and DS_NOT_FOUND are integers here, no problem.

akutz: It does not matter if they are integers. In fact, POSIX integer comparison requires quoting, ex. `[ "${x}" -gt "1" ]`. Unless needing explicit expansion, shell variables should almost always be quoted based on my experience.

tools/ds-identify
akutz, line 1397: You could also rewrite this as:

```bash
local _key=""
local _val=""
for _key in "metadata" "userdata" "vendordata"; do
  if _val="$(vmware-rpctool "info-get guestinfo.${_key}")" && [ -n "${_val}" ]; then
    return "${DS_FOUND}"
  fi
done
```

akutz: In fact, it could be simplified further like so:

```bash
local _key=""
local _val=""
for _key in "metadata" "userdata" "vendordata"; do
  if [[ -n "$(vmware-rpctool "info-get guestinfo.${_key}")" ]]; then
    return "${DS_FOUND}"
  fi
done
```

The above expression will fail if the command returns a non-zero exit code or empty string. Now, this requires double braces, which are not POSIX-shell compliant.

tools/ds-identify
sshedi: Done

tools/ds-identify
sshedi: rpctool returns a white space character, so `-n` will return 0
Also, for the tests to work I need to keep all the functions as it is. I will figure something out.

akutz: >  I will figure something out.

Let's just leave the code as-is please. I'm not sure why you think it needs to change. Optimization for optimization's sake isn't always the right answer. I want to not require setting an env var in those tests as it makes them bound to serial execution.

As for rpctool, here you go :)

```bash
local _key=""
local _val=""
for _key in "metadata" "userdata" "vendordata"; do
  if _val="$(vmware-rpctool "info-get guestinfo.${_key}")" && [ -n "${_val//[:space:]]//}" ]; then
    return "${DS_FOUND}"
  fi
done
```

The use of `${_val//[:space:]]//}` strips the whitespace from the value.

tools/ds-identify
sshedi: Nope, can't assume that. sh is bash only in some distros. So, we need to use POSIX complaint method.

akutz: Oh sure, I was not assuming it. I was saying if we could switch the interpreter explicitly to BASH as you requested elsewhere.

akutz: Do you happen to know if DASH supports the same pattern?

tools/ds-identify
```diff
@@ -1429,11 +1428,15 @@ dscheck_VMware() {
 
     # Activate the VMware datasource only if any of the fields used
     # by the datasource are present in the guestinfo table.
-    if { vmware_rpctool_guestinfo_metadata || \
-         vmware_rpctool_guestinfo_userdata || \
-         vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
-        return "${DS_FOUND}"
-    fi
+    local _val=''
```
akutz, line 1431: Don't you need to also use:

```bash
local _key=""
```

tools/ds-identify
```diff
@@ -1429,11 +1428,15 @@ dscheck_VMware() {
...
+    for _key in "metadata" "userdata" "vendordata"; do
+      _val="$(eval vmware_rpctool_guestinfo_${_key} 2>&1)"
```
akutz, line 1433: There is no need to use `eval` here.

tools/ds-identify
```diff
@@ -1429,11 +1428,15 @@ dscheck_VMware() {
...
+      if [ $? -eq 0 ]; then
+          if echo "${_val}" | grep -q "[[:alnum:]]"; then
```
akutz, line 1435: Please use the provided example as it is more efficient by not requiring the additional process executions.

akutz (changes requested):

tools/ds-identify
akutz: Please note I updated the above script example to redirect stderr to `/dev/null` to prevent an error message from triggering activation.

tools/ds-identify
sshedi: `DI_VIRT="${_RET,,}"` doesn't work on dash. Checked it. It will be nice if we can use bash but it seems like there are distros which doesn't use bash. `#!/bin/sh` is a symbolic link to a posix complaint shell, it can be bash/ksh/dash or anything. Using bash seems not feasible.

akutz: And yet `local` is used elsewhere, huh? Weird. Seems like if `/bin/sh` is used then the rest of the file should be strictly POSIX compliant...

tools/ds-identify
akutz: Actually, we're overthinking this. This isn't what we should be doing. Hang on.

akutz: We should stick with this:

```shell
    if { vmware_rpctool_guestinfo_metadata || \
          vmware_rpctool_guestinfo_userdata || \
          vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
         return "${DS_FOUND}"
     fi
```

But update:

```shell
vmware_rpctool_guestinfo() {
    local _val=""
    if ! _val="$(vmware-rpctool "info-get guestinfo.${1}")"; then
        return "${?}"
    fi
    return [ -n "${val/ /}" ]
}

vmware_rpctool_guestinfo_metadata() {
    vmware_rpctool_guestinfo "metadata"
}

vmware_rpctool_guestinfo_userdata() {
    vmware_rpctool_guestinfo "userdata"
}

vmware_rpctool_guestinfo_vendordata() {
     vmware_rpctool_guestinfo "vendordata"
}
```

akutz: Okay, as @sshedi just pointed out on Slack, string replacement is not POSX-shell compliant. Okay, here's how we optimize this then:

```shell
    # Activate the VMware datasource only if any of the fields used
    # by the datasource are present in the guestinfo table.
    if { vmware_rpctool_guestinfo_metadata && \
         vmware_rpctool_guestinfo_userdata && \
         vmware_rpctool_guestinfo_vendordata; } 2>/dev/null | \
         grep -q "[[:alnum:]]"; then
        return "${DS_FOUND}"
    fi
```

akutz: We also just realized that the above example's best case will still require three RPC calls. I think we'd rather have three calls to `grep` than three RPC calls, so:

```shell
vmware_rpctool_guestinfo() {
    vmware-rpctool "info-get guestinfo.${1}" 2>/dev/null | grep "[[:alnum:]]"
}

vmware_rpctool_guestinfo_metadata() {
    vmware_rpctool_guestinfo "metadata"
}

vmware_rpctool_guestinfo_userdata() {
    vmware_rpctool_guestinfo "userdata"
}

vmware_rpctool_guestinfo_vendordata() {
     vmware_rpctool_guestinfo "vendordata"
}
```

Then the following can still work:

```shell
    if { vmware_rpctool_guestinfo_metadata || \
         vmware_rpctool_guestinfo_userdata || \
         vmware_rpctool_guestinfo_vendordata; } >/dev/null 2>&1; then
         return "${DS_FOUND}"
     fi
```

akutz: Just verified the following patch still passes the unit tests:

```diff
diff --git a/tools/ds-identify b/tools/ds-identify
index 234ffa81..9efb647f 100755
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1384,16 +1384,20 @@ vmware_has_rpctool() {
     command -v vmware-rpctool >/dev/null 2>&1
 }
 
+vmware_rpctool_guestinfo() {
+    vmware-rpctool "info-get guestinfo.${1}" 2>/dev/null | grep "[[:alnum:]]"
+}
+
 vmware_rpctool_guestinfo_metadata() {
-    vmware-rpctool "info-get guestinfo.metadata"
+    vmware_rpctool_guestinfo "metadata"
 }
 
 vmware_rpctool_guestinfo_userdata() {
-    vmware-rpctool "info-get guestinfo.userdata"
+    vmware_rpctool_guestinfo "userdata"
 }
 
 vmware_rpctool_guestinfo_vendordata() {
-    vmware-rpctool "info-get guestinfo.vendordata"
+     vmware_rpctool_guestinfo "vendordata"
 }
 
 dscheck_VMware() {
```

tools/ds-identify
```diff
@@ -1384,16 +1383,20 @@ vmware_has_rpctool() {
...
 }
 
 vmware_rpctool_guestinfo_userdata() {
-    vmware-rpctool "info-get guestinfo.userdata"
+    vmware_rpctool_guestinfo "userdata"
 }
 
 vmware_rpctool_guestinfo_vendordata() {
-    vmware-rpctool "info-get guestinfo.vendordata"
+     vmware_rpctool_guestinfo "vendordata"
```
akutz, line 1399: You copied my original example verbatim and missed the thing I missed too. There's a leading WS on line 1399 :)

sshedi: Sigh. It's 3am :)

akutz: Right, so go to bed! :) I was wondering why you were poking at this in the first place so late. We're not spring chickens anymore Shreenidhi :)

akutz:
This looks good to me, thank you @sshedi! Consider this approved from my PoV.

TheRealFalcon (approved):

tools/ds-identify
smoser, line 452: this is really bad.  We shouldn't do things "just in case".   That doesn't actually make things better, but  makes them worse for a couple reasons:
 * systemd documents the outputs of 'systemd-detect-virt', and they are all  lower case. Accepting upper case is just a waste.
  * it is dramatically slower than it was before (2 additional forks - admittedly forks are fast on linux, but the are much slower than string comparisons).
   * The general policy of "just in case" is silly.  Should we also accept misspellings? in case something spells it 'vmwhere' ?

smoser (commented):
I do realize that I'm not really active in cloud-init before, but I saw this go by and just want to mention that I really think this kind of software development makes things worse, not better.

tools/ds-identify
akutz: Hi @smoser,

I concur. I thought this was removed prior to merge (see my other comments about reducing this PR to just the bits in `dscheck_VMware`. To be clear, this is not for misspelling, but your point is well received. Even if not systemd (FreeBSD), the mapping table ensures lower case as well.

I propose we revert this line to the former value.

akutz: Hi @smoser,

Please see https://github.com/canonical/cloud-init/pull/978.

smoser: thanks.

sshedi: Point taken. Thanks.

PR #972, opened 2021-08-12
Check if nic is up after sleep when trying to bring it up

aswinrajamannar:
## Proposed Commit Message
Azure: Check if interface is up after sleep when trying to bring it up

```
summary:

When bringing interface up by unbinding and then binding hv_netvsc driver, it might take a short delay after binding for the link to be up. So before trying unbind/bind again after sleep, check if the link is up. This is a corner case when a preprovisioned VM is reused and the NICs are hot-attached.
```

## Test Steps
- Deploy savable Pre-provisioned VMs whose NICs get hot-attached when they get reprovisioned
- Reuse them for a multi-NIC VM. In some cases, this may show up as a delay for the VM to report ready because we aggressively unbind, bind and check of status immediately after the bind.

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -923,6 +923,16 @@ class DataSourceAzure(sources.DataSource):
 
             sleep(sleep_duration)
 
+            # Since we just did a unbind and bind, check again after sleep
+            # but before doing unbind and bind again to avoid races where the
+            # link might take a slight delay after bind to be up.
+            if self.distro.networking.is_up(ifname):
+                msg = ("Link is up after checking after sleeping for %d secs"
+                       " after %d attempts" %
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -923,6 +923,15 @@ def wait_for_link_up(self, ifname):
 
             sleep(sleep_duration)
 
+            # Since we just did a unbind and bind, check again after sleep
+            # but before doing unbind and bind again to avoid races where the
+            # link might take a slight delay after bind to be up.
+            if self.distro.networking.is_up(ifname):
+                msg = ("Link is up after checking after sleeping for %d secs"
```
anhvoms, line 930: I think you're missing a report_diagnostic_event to output this message?

TheRealFalcon (approved):

PR #979, opened 2021-08-13
Update ds-identify to pass shellcheck

akutz:
## Proposed Commit Message

```
Update ds-identify to pass shellcheck

This patch updates the ds-identify file to pass the
popular shell linter, shellcheck. Updates include:

* Global ignore for error code 3043, the warning about
  the "local" keyword not being POSIX compliant. This
  keyword is used everywhere and cannot be easily removed.

* Point-of-use ignore for error code 3013, the warning
  about the "-nt" comparator not being POSIX compliant.
  There's no POSIX compliant way to compare timestamps
  on two files without forking a new process, so this
  comparator remains.

* Several point-of-use ignores for error code 2254, the
  warning about quoting parameters. These warnings are
  ignored for the *_matches functions where wildcard
  tokens are expected to be used. Quoting the parameters
  would prevent the desired globbing and cause calls
  to these functions with wildcards to fail.

* A point-of-use ignore for error code 2086 for the
  same reason as the above ignore -- allowing globbing
  to take place for matching.

* Quoted parameters in the remainder of the locations
  where it was safe to do so and did not break the
  tests.
```

## Additional Context
NA

## Test Steps
```shell
$ make clean_pyc && PYTHONPATH="$(pwd)" python3 -m pytest -v tests/unittests/test_ds_identify.py >/dev/null 2>&1 && echo "success"
success
```

```shell
$ shellcheck --version && shellcheck tools/ds-identify && printf "\nsuccess\n"
ShellCheck - shell script analysis tool
version: 0.7.2
license: GNU General Public License, version 3
website: https://www.shellcheck.net

success
```

## Checklist:


TheRealFalcon (approved):
Thanks!

PR #980, opened 2021-08-13
testing: skip upgrade tests on LXD VMs

TheRealFalcon:
## Proposed Commit Message

```
testing: skip upgrade tests on LXD VMs

The issues we see on Bionic VMs don't appear anywhere else, including
when invoking kvm directly. It likely has to do with the extra
LXD agent setup happening on bionic. Given that we still have Bionic
covered on all other platforms, the risk of skipping bionic for LXD VM
tests seems low.
```

## Checklist:

### DIFF
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -3,7 +3,7 @@ import logging
 import os
 import pytest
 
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
LGTM we'll circle back on this.

PR #984, opened 2021-08-17
Fix home permissions modified by ssh module (SC-338)

TheRealFalcon:
## Proposed Commit Message

```
Fix home permissions modified by ssh module

In #956, we updated the file and directory permissions for keys not in
the user's home directory. We also unintentionally modified the
permissions within the home directory as well. These should not change,
and this commit changes that back.

LP: #1940233
```

## Additional Context
See #956 and https://bugs.launchpad.net/cloud-init/+bug/1940233

## Checklist:


TheRealFalcon:
FYI @esposem

cloudinit/ssh_util.py
```diff
@@ -321,17 +321,27 @@ def check_create_path(username, filename, strictmodes):
...
+                    mode = 0o755
+                    uid = root_pwent.pw_uid
+                    gid = root_pwent.pw_gid
+                    if parent_folder.startswith(user_pwent.pw_dir):
+                        mode = 0o700
+                        uid = user_pwent.pw_uid
+                        gid = user_pwent.pw_gid
+                    try:
+                        os.mkdir(parent_folder, mode=mode)
+                    except FileExistsError:
```
esposem, line 342: So with this change 
- `/`, `/home/` and `/home/<user>` are skipped and not created in any case. 
- then, we can have either a path that is not in the user home folder, and we chmod it to root ownership, or inside the user folder, and we chmod it as user-owned.

Only one question: what if `/home/` or `/home/<user>` do not exist? `os.mkdir` should launch a `FileNotFoundError` I think, we should handle also this case right?

esposem (changes requested):

cloudinit/ssh_util.py
TheRealFalcon: I think I agree with @paride in https://bugs.launchpad.net/cloud-init/+bug/1940233/comments/4 when he says it's not the responsibility of this module to create a user's home directory.

paride: A couple of thoughts about this:
 * The `authorized_keys` file is always created user-owned (chmod 600), even if its path is outside the user's `$HOME` directory. I think this is good, so users are still be able to manage their own `authorized_keys`.
 * Given that we've got bitten by this, I think it we should test two more scenarios:
   * AuthorizedKeysFile outside of `$HOME`, e.g. in `/etc/ssh/authorized_keys/%u/`
   * AuthorizedKeysFile deeper in `$HOME`, e.g. `%h/secret/ssh/`

cloudinit/ssh_util.py
```diff
@@ -321,17 +321,27 @@ def check_create_path(username, filename, strictmodes):
         home_folder = os.path.dirname(user_pwent.pw_dir)
         for directory in directories:
             parent_folder += "/" + directory
-            if home_folder.startswith(parent_folder):
+            if (home_folder.startswith(parent_folder) or
+                    parent_folder == user_pwent.pw_dir):
                 continue
 
             if not os.path.isdir(parent_folder):
```
paride, line 328: There is one weird case that is not covered by this `if`: `parent_folder` existing but not-a-directory (file, symlink). In this case cloud-init is going to fail creating deeper subdirectories or the authorized_keys file, possibly doing unexpected things in the case of a symlink. Do you think we should cover this case, returning early when `os.path.isdir(parent_folder)` but `os.path.exists(parent_folder)`?

paride: More specifically I'm worried by a user doing e.g. `ln -s .ssh /root/.ssh`, and of what would cloud-init do.

TheRealFalcon: Pushed https://github.com/canonical/cloud-init/pull/984/commits/240a0795d9484a18b4a5798fb0f08bd5b7c139bf , which I think addresses this.

cloudinit/ssh_util.py
esposem: @TheRealFalcon ok, seems reasonable
@paride regarding the right permission w.r.t. authorized_keys files, look [here](https://github.com/canonical/cloud-init/pull/948#issuecomment-891332457). The mistake in this code is that I did not think about the permissions that other ssh tools need, in this case `ssh-keygen`. Because if `.ssh` has permission 755 and it's owned by root, it should work w.r.t. AuthorizedKeysFile. 

After a quick test I see that the mode doesn't matter, the ownership is the problem (which makes sense because with root owner and 755 `ssh-keygen` on behalf of the user cannot write in the folder).
 
Regarding the first scenario, it is covered in [this](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_sshutil.py#L1177) and [this](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_sshutil.py#L1108) test.

TheRealFalcon: Added the requested integration tests. There's some overlap with the unit tests (and the unit tests are much more comprehensive), but better to have a basic integration check of the other use cases too.

esposem: Sounds good to me, thank you!

TheRealFalcon:
@esposem , Yep, my bad. I misread Paride's comment and went a direction based off a wrong assumption. What you say makes sense.

cloudinit/ssh_util.py
paride: LGTM, thanks!

esposem:
Everything now looks good to me, thank you for working on this!

tests/integration_tests/modules/test_ssh_keysfile.py
```diff
@@ -83,3 +73,103 @@ def test_authorized_keys(client: IntegrationInstance):
...
+        # See LP: #1940233
+        home_dir = '/home/{}'.format(user)
+        home_perms = '755'
+        if user == 'root':
+            home_dir = '/root'
+            home_perms = '700'
+        assert '{} {}'.format(user, home_perms) == client.execute(
+            'stat -c "%U %a" {}'.format(home_dir)
+        )
+        if client.execute("test -f {}/.ssh".format(home_dir)).ok:
```
blackboxsw, line 87: Does this actually work? I get exit 1  from this call meaning we don't drop in for the perms check.

blackboxsw: Also can we attempt to run something like the example ssh-keygen call in the bug report to confirm that all the plumbing works as expected without error?

blackboxsw (approved):
LGTM! Needs rebase and CI pass but otherwise +1

PR #985, opened 2021-08-17
Ignore hotplug socket when collecting logs (SC-345)

TheRealFalcon:
## Proposed Commit Message

```
Ignore hotplug socket when collecting logs

Update "cloud-init collect-logs" to ignore
/run/cloud-init/hook-hotplug-cmd as this will raise the error
"/run/cloud-init/hook-hotplug-cmd` is a named pipe" if included.

Also updated logs.py to continue writing the tarball if it fails
collecting a file rather than let the exception bubble up.

LP: #1940235
```

## Additional Context
https://bugs.launchpad.net/cloud-init/+bug/1940235

## Test Steps
Start an impish image (or `packages/bddeb`, upload to your instance, install deb and reboot).
Run `cloud-init collect-logs`
On main, fails as mentioned in the bug. On this branch, shows `Wrote /root/cloud-init.tar.gz` having `hook-hotplug-cmd` excluded from the tarball. From now on, if any other error is encountered, we should see something along the lines of:
```
Failed collecting file(s) due to error:
[('/run/cloud-init/hook-hotplug-cmd', '/tmp/tmpbsl70bx1/cloud-init-logs-2021-08-17/run/cloud-init/hook-hotplug-cmd', '`/run/cloud-init/hook-hotplug-cmd` is a named pipe')]
Wrote /root/cloud-init.tar.gz
```
with the error file(s) being excluded from the tarball.

We should really add integration tests for `cloud-init collect-logs`. I'd like to add it to this PR if there's time.

## Checklist:

### DIFF
--- a/cloudinit/cmd/devel/logs.py
+++ b/cloudinit/cmd/devel/logs.py
@@ -48,11 +48,15 @@ def get_parser(parser=None):
     return parser
 
 
-def _copytree_ignore_sensitive_files(curdir, files):
-    """Return a list of files to ignore if we are non-root"""
-    if os.getuid() == 0:
-        return ()
-    return (INSTANCE_JSON_SENSITIVE_FILE,)  # Ignore root-permissioned files
+def _copytree_rundir_ignore_files(curdir, files):
+    """Return a list of files to ignore for /run/cloud-init directory"""
+    ignored_files = [
+        'hook-hotplug-cmd',  # named pipe for hotplug
+    ]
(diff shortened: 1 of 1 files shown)


paride (approved):
Not an authoritative review but this lgtm!

blackboxsw (approved):
This works for me. We'll get a noise error if other files are root read-only in /run/cloud-init (which only cloud-init should be writing anyway).

PR #989, opened 2021-08-20
Document how to seed cloud-init by creating files in /var/lib/cloud/seed/nocloud/

Jille:
## Proposed Commit Message

```
Document how to seed cloud-init by creating files in /var/lib/cloud/seed/nocloud/

This took me quite a while to find. I hope this helps others.
```

## Checklist:

### DIFF
--- a/doc/rtd/topics/datasources.rst
+++ b/doc/rtd/topics/datasources.rst
@@ -52,6 +52,13 @@ The following is a list of documents for each supported datasource:
    datasources/vultr.rst
    datasources/vmware.rst
 
+Seeding
+=======
+
+If you want to embed cloud-init in a disk, create
+`/var/lib/cloud/seed/nocloud/user-data` and
+`/var/lib/cloud/seed/nocloud/meta-data`.
+
 Creation
 ========
 


Jille:
Looks like I've forgotten to explain what I was doing: I wanted to use cloud-init for setting up some VMs on my own KVM server. I pre-create their disk with a shell script, and want to embed a cloud-init config into that disk so that it configures itself on boot.

> What do you mean by "explicit reference" here? What would you want to keep here?
At least we should make it obvious that NoCloud is not yet-another-cloud-provider, but a module for those running outside of a cloud provider.

> > But I still think people will be looking on this page for figuring out how to supply user-data through the filesystem.
> But this won't be correct or relevant for most users. The seed directory will be different (or non-existent) for every other cloud. Furthermore, most people are interested in what happens on first boot, which means they'd rather use other means of instrumenting user data over dropping files onto an already-booted system. We have the FAQ for that.
I'm not very familiar with cloud-init's internals. Wouldn't it always read /var/lib/cloud/seed/nocloud/? Or only if no cloud provider was detected?


> I do think this is useful information to add. I just don't see it being very relevant outside of the NoCloud datasource.
Agreed, if we'd have an explicit reference (;)) to NoCloud being there for non-cloud-providers, we could explain the seed folder on the NoCloud page.

TheRealFalcon:
> Wouldn't it always read /var/lib/cloud/seed/nocloud/? Or only if no cloud provider was detected?

Correct, this will only happen if we're not on a "real" cloud. Just for some context, most clouds have a metadata service that can be queried via http, and user data or vendor data usually gets fetched from there. NoCloud is fairly abnormal in how it obtains its userdata.

How about at the bottom of https://cloudinit.readthedocs.io/en/latest/topics/datasources.html#known-sources , we put something like:
```
If you're working outside of a cloud-hosted environment, see the NoCloud datasource
```

and then added your information in this PR to the NoCloud page. Would that work?

PR #992, opened 2021-08-23
testing: Fix ssh keys integration test

TheRealFalcon:
## Proposed Commit Message

```
testing: Fix ssh keys integration test

Home directory permissions changed in hirsute. The integration test
assumed permissions from earlier releases. Test was fixed to take both
permissions into account
```

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_ssh_keysfile.py
+++ b/tests/integration_tests/modules/test_ssh_keysfile.py
@@ -3,6 +3,7 @@ import pytest
 from io import StringIO
 from paramiko.ssh_exception import SSHException
 
+from tests.integration_tests.clouds import ImageSpecification
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.util import get_test_rsa_keypair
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/modules/test_ssh_keysfile.py
```diff
@@ -77,13 +77,15 @@ def common_verify(client, expected_keys):
         # Ensure we haven't messed with any /home permissions
         # See LP: #1940233
         home_dir = '/home/{}'.format(user)
-        home_perms = '755'
+        # Home permissions aren't consistent between releases. On ubuntu
+        # this can change to 750 once focal is unsupported.
+        home_perms = '75'
```
blackboxsw, line 82: Can we do something a bit more explicit here?
 ?
```suggestion
        if ImageSpecification.from_os_image().release in ("xenial", "bionic"):
            home_perms = '755'
        else:
            home_perms = '750'
```

blackboxsw (approved):
LGTM since this test is "@ubuntu" only we can make this kind of conditional check work for this test.  if CI passes let's land it.

PR #997, opened 2021-08-23
Ubuntu/bionic

TheRealFalcon:


blackboxsw (commented):
1. diff looks good
```bash
@@ -101,7 +101,7 @@ cloud-init (21.3-1-g6803368d-0ubuntu1~18.04.1) bionic; urgency=medium
     - Fix unit tests breaking from new httpretty version (#903)
     - Allow user control over update events (#834)
 
- -- James Falcon <james.falcon@canonical.com>  Mon, 23 Aug 2021 17:28:07 -0500
+ -- Chad Smith <chad.smith@canonical.com>  Mon, 23 Aug 2021 20:22:04 -0600
 ```
2.  **NEEDS FIXING** sbuild is shows a couple of extra lintian warnings that I think we should fix (and probably should fix in separate ubuntu/devel later) Not a biggie, but it's better if we mitigate lintian warnings so we don't lose more significant warnings/errors in a wall-o-text.
```bash
W: cloud-init source: patch-file-present-but-not-mentioned-in-series ubuntu-advantage-revert-tip.patch
...
W: cloud-init: spelling-error-in-changelog unnecesary unnecessary
W: cloud-init: spelling-error-in-changelog Moustafa Moustafa (duplicate word) Moustafa
```


I think we should be `git rm debian/patches/ubuntu-advantage-revert-tip.patch` file right?  We no longer apply it as we have released an updated UA client that can handle the "newer" CLI commands. And it's not listed in the debian/patches/series file so it doesn't get applied anyway even if present.

blackboxsw (changes requested):
We should fix those lintian warnings in each series where applicable.

blackboxsw (approved):
Thanks for the lintian fixes
+1 LGTM

PR #1000, opened 2021-08-25
Fix `make style-check` errors

sshedi:
```
  Fix make style-check errors

Using flake8 inplace of pyflakes
Renamed run-pyflakes -> run-flake8
Changed target name to flake8 in Makefile

With pyflakes we can't suppress warnings/errors in few required places.
flake8 is flexible in that regard. Hence using flake8 seems to be a
better choice here.

flake8 does the job of pep8 anyway.
So, removed pep8 target from Makefile along with tools/run-pep8 script.

Included setup.py in flake8 checks
```

## Proposed Commit Message

```
 Fix make style-check errors

Using flake8 inplace of pyflakes
Renamed run-pyflakes -> run-flake8
Changed target name to flake8 in Makefile

With pyflakes we can't suppress warnings/errors in few required places.
flake8 is flexible in that regard. Hence using flake8 seems to be a
better choice here.

flake8 does the job of pep8 anyway.
So, removed pep8 target from Makefile along with tools/run-pep8 script.

Included setup.py in flake8 checks
```

## Test Steps
```
make style-check; echo $?
```

## Checklist:


cloudinit/cmd/main.py
```diff
@@ -42,6 +38,10 @@
 from cloudinit.config import cc_set_hostname
 from cloudinit import dhclient_hook
 
+from cloudinit import patcher
+patcher.patch_logging()
+from cloudinit import log as logging  # noqa: E402
+
```
TheRealFalcon, line 44: Let's not move this. I want the patch to happen before other cloud-init imports.

cloudinit/cmd/devel/hotplug_hook.py
```diff
@@ -13,7 +13,11 @@
 from cloudinit.net.network_state import parse_net_config_data
 from cloudinit.reporting import events
 from cloudinit.stages import Init
-from cloudinit.sources import DataSource, DataSourceNotFoundException
+
```
TheRealFalcon, line 16: Let's remove this extra line

tests/integration_tests/instances.py
```diff
@@ -13,7 +13,9 @@
 try:
     from typing import TYPE_CHECKING
     if TYPE_CHECKING:
-        from tests.integration_tests.clouds import IntegrationCloud
+        from tests.integration_tests.clouds import (  # noqa: F401
```
TheRealFalcon, line 16: Why make this 3 lines?

TheRealFalcon (changes requested):
* The `run-pep8` stuff can be removed entirely. It's functionality is entirely covered by flake8.
* How are you running linting such that you're getting "F401" on the typing imports? They're used in the type annotation comments which doesn't trigger a warning for me.
* Can we leave W503 and W504 in and revert the changes for them back? These were added intentionally and I see no reason to pull them out...especially since the guidance has reversed in recent years.

sshedi:
I ran checks on Ubuntu-16.04 and it reported these errors. Are there any plans to stop running CI tests on Ubuntu 16.04?
The reason why I'm using 16.04 is, it behaves different compared to other versions. If tests work on 16.04, then most probably it will work on higher versions. I had a hard time while writing pytests for some modules because of dictionary ordering issues.

tests/integration_tests/instances.py
sshedi: Adding `# noqa: F401` at the end of line crosses 80 characters, so broke this into multiple lines.

cloudinit/cmd/main.py
sshedi: Okay, understood. My question is - can we do it in some other place and avoid this warning?

Like in `/usr/bin/cloud-init` before anything starts?

sshedi:
I will remove run-pep8 target and corresponding script.Thanks.

You want me to keep the ignore W503 & W504 in tox.ini as well? If guidance is reversed, it seems redundant.

TheRealFalcon:
> Are there any plans to stop running CI tests on Ubuntu 16.04?

16.04 no longer receives standard support, so from an upstream perspective, we're not concerned about breaking things in 16.04. Practically, we still run it in our CI because it has the lowest version of python that we support (3.5), so it's an easy way to test our 3.5 dependencies.

> You want me to keep the ignore W503 & W504 in tox.ini as well?

Correct.

PR #1001, opened 2021-08-25
testing: integration_test updates (SC-362)

TheRealFalcon:
## Proposed Commit Message

```
Integration test upgrades for the 21.3-1 SRU:

* Update test_combined.py to allow either valid LXD subplatform
* Split jinja templated tests into separate module as they can be more
  fragile
* Move checks for warnings and tracebacks into dedicated utility
  function. This allows us to work around persistent and expected
  tracebacks/warnings on particular clouds.
* Update test_upgrade.py to allow either valid Azure datasource.
  /var/lib/waagent or a mounted device are both valid.
* Add specificity to test_ntp_servers.py
  Clouds will often specify their own ntp servers in the ntp
  configuration files, so make the tests manually specify their own.
* Account for additional keys on system in test_ssh_keysfiles.py
* Update tests to account for invalid cache
  test_user_events.py and test_version_change.py both have tests that
  assume we will have valid ds cache when rebooting.
  In test_user_events.py, subsequent boots should block applying
  network on boot if boot event is denied. However, if the cache is
  invalid, it is valid to apply networking config that boot.
  In test_version_change.py no cache found won't trigger the expected
  debug log. Additionally, the pickle used for that test on an older
  release triggered an unexpected issue that took a different error
  path.
* Ignore bionic in hotplug tests (LP: #1942247)
  On Bionic, we traceback when attempting to detect the hotplugged
  device in the updated metadata. This is because Bionic is
  specifically configured not to provide network metadata.
  See LP: #1942247 for more details.
* Fix date used in test_final_message.
  In test_final_message, we ensured the variable substitution works as
  expected. For $timestamp, we compared against the current date. It's
  possible for the host date to be massively different from the client
  date, so obtain date on client rather than host.
* Remove module success from lp1813396 test. Module may fail
  unrelatedly (in this case apt-get update is failing), but the test
  should still pass.
* Skip testing events if network is disabled
* Ensure we install expected version of cloud-init
  As part of test setup, we can install cloud-init from various
  sources, including PROPOSED, PPAs, etc. We were never checking that
  this install completes successfully, and on OCI, it wasn't
  completing successfully because of apt locking issues. Code has
  been updated to retry, and then fail loudly if we can't complete the
  install.
* Remove ubuntu-azure-fips metapkg which mandates FIPS-flavour kernel
  In test_lp1835584.py
* Update test_user_events.py to account for Azure behavior
  since Azure has a separate service to clear the pickled metadata
  every boot
* Change failure to warning in test_upgrade.py if initial boot errors
  If there's already a pre-existing cause for warnings or tracebacks,
  that shouldn't cause the new version to fail.
* Add retry to test_random_passwords_emitted_to_serial_console
  It's possible we haven't retrieved the entire log when the call returns,
  so retry a few times if the output isn't empty.
```

## Test Steps
Run integration tests

## Checklist:


blackboxsw:
@TheRealFalcon you'll also need this to fix the Azure FIps test
```diff
diff --git a/tests/integration_tests/bugs/test_lp1835584.py b/tests/integration_tests/bugs/test_lp1835584.py
index 660d2a2a..451a693c 100644
--- a/tests/integration_tests/bugs/test_lp1835584.py
+++ b/tests/integration_tests/bugs/test_lp1835584.py
@@ -57,6 +57,8 @@ def _check_iid_insensitive_across_kernel_upgrade(
     result = instance.execute("apt-get update")
     # Install a 5.4+ kernel which provides lowercase product_uuid
     result = instance.execute("apt-get install linux-azure --assume-yes")
+    # Remove ubuntu-azure-fips metapkg which mandates FIPS-flavour kernel
+    result = instance.execute("ua disable fips --assume-yes")
     if not result.ok:
         pytest.fail("Unable to install linux-azure kernel: {}".format(result))
     instance.restart()
```

TheRealFalcon:
Wahoo...Thanks!

tests/integration_tests/modules/test_jinja_templating.py
```diff
@@ -0,0 +1,30 @@
...
+  - echo {{v1.local_hostname}} > /var/tmp/runcmd_output
+  - echo {{merged_cfg._doc}} >> /var/tmp/runcmd_output
+"""
+
+
+@pytest.mark.user_data(USER_DATA)
+def test_runcmd_with_variable_substitution(client: IntegrationInstance):
+    """Test jinja substitution.
+
+    Ensure we can also substitue variables from instance-data-sensitive
```
blackboxsw, line 21:
```suggestion
    Ensure we can also substitute variables from instance-data-sensitive
```

tests/integration_tests/modules/test_set_password.py
```diff
@@ -137,6 +140,15 @@ def test_random_passwords_emitted_to_serial_console(self, class_client):
             # Assume that an exception here means that we can't use the console
             # log
             pytest.skip("NotImplementedError when requesting console log")
+            return
+        if console_log.lower() == 'no console output':
```
blackboxsw, line 144: Thanks for this addition/retry fix here.

blackboxsw (approved):
Thank you for all of these changes.

PR #1002, opened 2021-08-26
puppet config: add the start_agent option

andrewbogott:
The current code starts the puppet agent and also sets autostart
in all cases. This conflicts with a common pattern where puppet
itself manages the agent and autostart state.

For example, in my deploy puppet disables the puppet agent
and replaces it with a cron. This causes various races
both within this cloud-init unit and within puppet itself
while cloud-init and puppet fight over whether or not
to enable the service.

## Proposed Commit Message
puppet config: add the 'start_agent' option

this can be used in combination with the 'exec' option to perform a single
synchronous puppet run and then leave whatever happens next to
the system config as dictated by puppet.

## Checklist:


TheRealFalcon (approved):
Thanks!

PR #1005, opened 2021-08-31
Add retries to DataSourceGCE.py when connecting to GCE

vteratipally:
## Proposed Commit Message
Add retries to DataSourceGCE.py when connecting to GCE
```
summary: 

Add retries to DatasourceGCE when connecting to GCE. Sometimes when the trying to fetch the metadata, the cloud-init fails and the fallback datasource NoCloud is used which is not expected. So adding retries to ensure loading of the data source. 

```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceGCE.py
+++ b/cloudinit/sources/DataSourceGCE.py
@@ -29,6 +29,8 @@ class GoogleMetadataFetcher(object):
 
     def __init__(self, metadata_address):
         self.metadata_address = metadata_address
+        self.retries = 5
+        self.sec_between_retries = 1
 
     def get_value(self, path, is_text, is_recursive=False):
         value = None
@@ -36,7 +38,9 @@ class GoogleMetadataFetcher(object):
             url = self.metadata_address + path
             if is_recursive:
                 url += '/?recursive=True'
-            resp = url_helper.readurl(url=url, headers=HEADERS)
+            resp = url_helper.readurl(url=url, headers=HEADERS,
+                                      retries=self.retries,
+                                      sec_between=self.sec_between_retries)
         except url_helper.UrlError as exc:
             msg = "url %s raised exception %s"
             LOG.debug(msg, path, exc)


cloudinit/sources/DataSourceGCE.py
```diff
@@ -29,14 +29,18 @@ class GoogleMetadataFetcher(object):
 
     def __init__(self, metadata_address):
         self.metadata_address = metadata_address
+        self.retries = 5
```
blackboxsw, line 32: It might be worth adding a url_sec_between_retries class attribute in the base Datasource class in cloudinit.sources.__init__. 

There are some common base class attributes like [url_retries, url_timeout etc](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L177-L180) that are sourced by calling [self.get_url_params()](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L391).   


This is helpful from a datasource config and testing perspective statically set them as class attributes url_retries url_timeout etc because they can also be overridden in with datasource configuration on disk via a file like /etc/cloud/cloud.cfg.d/customize-gcp.cfg
```yaml
datasource:
    GCE:
      retries: 10 # (defaults to 5 times)
      sec_between_retries: 10 # (defaults to 1 second)
```

blackboxsw (changes requested):
Thanks you for this submission @vteratipally. I've confirmed that you have signed the cloud-init Contributor Agreement. 

I've provided a couple of comments about how we can best add this retry logic to GCE in a way that aligns with how other datasources instrument this setting in a more flexible way.


Please also add the following:
1. your github username alphabetically in ./tools/.github-cla-signers
2. unit test coverage of these changes if you can in tests/unittests/test_datasource/test_gce.py
 3.  documentation update to doc/rtd/topics/datasources/gce.rst with a  new section called

```
Configuration                                                                   
-------------  
The following configuration can be set for the datasource in system             
configuration (in `/etc/cloud/cloud.cfg` or `/etc/cloud/cloud.cfg.d/`).         
                                                                                
The settings that may be configured are:                                        
...
```
See either doc/rtd/topics/datasources/openstack.rst or doc/rtd/topics/datasources/ec2.rst for examples of the markdown in this section.

blackboxsw (commented):
Changes, docs and tests look good to me. will test this tomorrow and approve if no other concerns. Much thanks @vteratipally

blackboxsw (approved):
After long delay, finally got round to validating behavior on GCP. Thank you, thank you for this change and your patience.

vteratipally:
thank you @blackboxsw

PR #1011, opened 2021-09-03
Ubuntu/devel

TheRealFalcon:
Double check the version number. It autocompleted to ubuntu3, but I changed it since we never released an ubuntu2. Is that right?

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+cloud-init (21.3-1-g6803368d-0ubuntu2) UNRELEASED; urgency=medium
+
+  * cherry-pick 28e56d99: Azure: Retry dhcp on timeouts when polling
+    reprovisiondata
+
+ -- James Falcon <james.falcon@canonical.com>  Fri, 03 Sep 2021 14:36:28 -0500
+
 cloud-init (21.3-1-g6803368d-0ubuntu1) impish; urgency=medium
 
   * New upstream snapshot.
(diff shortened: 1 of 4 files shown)


blackboxsw (approved):
- [x] diff good
```diff
diff --git a/debian/changelog b/debian/changelog
index b2cb8a6c..b56773cf 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -5,7 +5,7 @@ cloud-init (21.3-1-g6803368d-0ubuntu2) impish; urgency=medium
   * cherry-pick e69a8874: Set Azure to only update metadata on
     BOOT_NEW_INSTANCE
 
- -- James Falcon <james.falcon@canonical.com>  Fri, 03 Sep 2021 14:36:41 -0500
+ -- Chad Smith <chad.smith@canonical.com>  Fri, 03 Sep 2021 13:51:30 -0600
 
 cloud-init (21.3-1-g6803368d-0ubuntu1) impish; urgency=medium
```
- [x] sbuild good and patches applied, no additional lintian warnings
- [x]  DatasourceAzure represents applied patches
```bash
dpkg-source: info: using patch list from debian/patches/series
dpkg-source: info: applying cpick-28e56d99-Azure-Retry-dhcp-on-timeouts-when-polling
dpkg-source: info: applying cpick-e69a8874-Set-Azure-to-only-update-metadata-on-BOOT_NEW_INSTANCE

```

PR #1013, opened 2021-09-08
Improve ug_util.py

sshedi:

```
Improve ug_util.py
```


cloudinit/distros/ug_util.py
```diff
@@ -16,86 +16,69 @@
...
+# Comma seperated string or a list or a dictionary
 #
-# The output is a dictionary of group
-# names => members of that group which
-# is the standard form used in the rest
-# of cloud-init
+# Returns dictionary of group names => members of that group which is the
+# standard form used in the rest of cloud-init
 def _normalize_groups(grp_cfg):
+    groups = {}
```
TheRealFalcon, line 25: Nit: I think it makes sense to keep groups where it was. It doesn't get used until right after where it was defined, so it makes sense to keep the declaration close.

cloudinit/distros/ug_util.py
```diff
@@ -107,181 +90,149 @@ def _normalize_users(u_cfg, def_user_cfg=None):
                 if u and u not in users:
                     users[u] = {}
         elif isinstance(user_config, dict):
+            n = 'default'
```
TheRealFalcon, line 93: Nit: I think these 3 lines could just be
```python
n = user_config.pop('name', 'default')
```

TheRealFalcon (approved):
Thanks for the refactor. I left two nits inline, but I'm ok merging without them.

Some of these changes definitely make the code clearer (removing unnecessary `len()` calls, simplying if blocks and loops, etc), but some are also purely cosmetic (comment line length, single vs double quotes, removing unnecessary parens). In general, in the future, I think I'd prefer to only make cosmetic changes if we're modifying the code in question, especially since we're likely to adopt black at some point in the future.

cloudinit/distros/ug_util.py
sshedi: Thanks, good catch.

sshedi:
> Thanks for the refactor. I left two nits inline, but I'm ok merging without them.
> 
> Some of these changes definitely make the code clearer (removing unnecessary `len()` calls, simplying if blocks and loops, etc), but some are also purely cosmetic (comment line length, single vs double quotes, removing unnecessary parens). In general, in the future, I think I'd prefer to only make cosmetic changes if we're modifying the code in question, especially since we're likely to adopt black at some point in the future.

I understand. There were some typos and other mistakes in comments so I started fixing them and ended up making modifications in all places. But your suggestion is valid, I will keep it in mind.

TheRealFalcon (approved):

PR #1014, opened 2021-09-10
Leave the details of service management to the distro.

citrus-it:
## Proposed Commit Message

```
Various modules restart services and they all have logic to try and
detect if they are running on a system that needs 'systemctl' or
'service', and then have code to decide which order the arguments
need to be etc. On top of that, not all modules do this in the same way.

The duplication and different approaches are not ideal but this also
makes it hard to add support for a new distribution that does not use
either 'systemctl' or 'service'.

This change adds a new manage_service() method to the distro class
and updates several modules to use it.```
```

## Checklist:


TheRealFalcon:
Hey @citrus-it , I noticed a lot of recent pushes to this branch. Is it ready for review yet, or should I hold off for a while longer? If it isn't ready for review, it'd be best to set the `wip` label on the PR.

citrus-it:
> Hey @citrus-it , I noticed a lot of recent pushes to this branch. Is it ready for review yet, or should I hold off for a while longer? If it isn't ready for review, it'd be best to set the `wip` label on the PR.

Hi, sorry about that, I had a little trouble getting the integration tests to all pass. CI is clean now so it's ready for review.

cloudinit/distros/__init__.py
```diff
@@ -795,6 +795,32 @@ def shutdown_command(self, *, mode, delay, message):
...
+        else:
+            cmds = {'stop': [service, 'stop'],
+                    'start': [service, 'start'],
+                    'enable': [service, 'start'],
+                    'restart': [service, 'restart'],
+                    'reload': [service, 'restart'],
+                    'try-reload': [service, 'restart'],
+                    }
+        cmd = list(init_cmd) + list(cmds[action])
+        try:
```
TheRealFalcon, line 818: I think it'd be better here to `return subp.subp(...)` and leave the exception handling up to the caller.

E.g., cc_fan currently logs a warning, but cc_ntp raises. 

We can add a docstring and specify that it may raise ProcessExecutionError

tests/unittests/test_distros/test_manage_service.py
```diff
@@ -0,0 +1,87 @@
...
+    @mock.patch("cloudinit.distros.subp.subp")
+    def test_manage_service_service_initcmd(self, m_subp, m_sysd):
+        self.dist.init_cmd = ['service']
+        m_sysd.return_value = False
+        self.dist.manage_service('start', 'myssh')
+        m_subp.assert_called_with(['service', 'myssh', 'start'], capture=True)
+
+    @mock.patch("cloudinit.distros.uses_systemd")
+    @mock.patch("cloudinit.distros.subp.subp")
+    def test_manage_service_service(self, m_subp, m_sysd):
```
TheRealFalcon, line 72: Is there any difference between this test and the one above it?

TheRealFalcon (changes requested):
Thanks! This is a great refactor overall.

I left a few inline comments, but nothing major.

PR #1023, opened 2021-09-17
Set Azure to apply networking config every BOOT

TheRealFalcon:
## Proposed Commit Message

```
Set Azure to apply networking config every BOOT

In #1006, we set Azure to apply networking config every
BOOT_NEW_INSTANCE because the BOOT_LEGACY option was causing problems
applying networking the second time per boot. However,
BOOT_NEW_INSTANCE is also wrong as Azure needs to apply networking
once per boot, during init-local phase.
```

## Test Steps
Run integration test

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_user_events.py
+++ b/tests/integration_tests/modules/test_user_events.py
@@ -31,8 +31,6 @@ def _add_dummy_bridge_to_netplan(client: IntegrationInstance):
 @pytest.mark.gce
 @pytest.mark.oci
 @pytest.mark.openstack
-@pytest.mark.azure
-@pytest.mark.not_xenial
 def test_boot_event_disabled_by_default(client: IntegrationInstance):
     log = client.read_from_file('/var/log/cloud-init.log')
     if 'network config is disabled' in log:
@@ -77,7 +75,7 @@ def _test_network_config_applied_on_reboot(client: IntegrationInstance):
     assert 'dummy0' not in client.execute('ls /sys/class/net')
 
     _add_dummy_bridge_to_netplan(client)
-    client.execute('rm /var/log/cloud-init.log')
+    client.execute('echo "" > /var/log/cloud-init.log')
     client.restart()
 
     log = client.read_from_file('/var/log/cloud-init.log')
@@ -92,6 +90,11 @@ def _test_network_config_applied_on_reboot(client: IntegrationInstance):
     assert 'dummy0' not in client.execute('ls /sys/class/net')
 
 
+@pytest.mark.azure
+def test_boot_event_enabled_by_default(client: IntegrationInstance):
+    _test_network_config_applied_on_reboot(client)
+
+
 USER_DATA = """\
 #cloud-config
 updates:
@@ -100,7 +103,6 @@ updates:
 """
 
 
-@pytest.mark.not_xenial
 @pytest.mark.user_data(USER_DATA)
 def test_boot_event_enabled(client: IntegrationInstance):
     _test_network_config_applied_on_reboot(client)
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -22,7 +22,7 @@ import requests
 from cloudinit import dmi
 from cloudinit import log as logging
 from cloudinit import net
-from cloudinit.event import EventType
+from cloudinit.event import EventScope, EventType
 from cloudinit.net import device_driver
 from cloudinit.net.dhcp import EphemeralDHCPv4
 from cloudinit import sources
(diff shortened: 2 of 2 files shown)


TheRealFalcon:
FYI @anhvoms

anhvoms:
> FYI @anhvoms

LGTM!

tests/integration_tests/modules/test_user_events.py
```diff
@@ -77,7 +75,7 @@ def _test_network_config_applied_on_reboot(client: IntegrationInstance):
     assert 'dummy0' not in client.execute('ls /sys/class/net')
 
     _add_dummy_bridge_to_netplan(client)
-    client.execute('rm /var/log/cloud-init.log')
+    client.execute('echo "" > /var/log/cloud-init.log')
```
blackboxsw, line 78: Why do we need to touch /var/log/cloud-init.log in this test?

blackboxsw, line 78: Is this just in case we haven't yet run cloud-init at all so that _test_network_config_applied_on_reboot doesn't fail?

Don't we want both rm /var/log/cloud-init.log and then echo "" > /var/log/cloud-init.log so that we don't accidentally match "old boot" logs?

blackboxsw (commented):
still walking through test scenarios one minor question on the integration test

tests/integration_tests/modules/test_user_events.py
blackboxsw: Also still hitting:
2021-09-17 19:48:09 INFO      pycloudlib.instance:instance.py:167 executing: sudo -- sh -c 'cat /var/log/cloud-init.log'
FAILED                                                                   [ 66%]

TheRealFalcon: `>` essentially does that, right?

blackboxsw: HA!. yes sorry, right.

TheRealFalcon: I changed the line because I was occasionally getting failures from `/var/log/cloud-init.log` not existing, which is really strange. I changed it to empty out the file completely, but then I didn't get any errors anymore. `restart()` has a built in wait, so the cloud-init log should be 'full' by the time we try to read it. I'm curious if adding a `sync` to Azure at the end of `wait` would fix it.

blackboxsw (approved):
@TheRealFalcon all good thanks for walking through this with me.

xiachen-rh:
@TheRealFalcon 
Hi James,  some customers report a problem that this change disables the way they used to add additional search domains,  then what is the correct way to manually add additional persistent search domains to a Azure VM now?

TheRealFalcon:
Hi Amy,

Setting search domains can be environment dependent, but a common method
for RHEL-based distros is to use the Resolv Conf module:
https://cloudinit.readthedocs.io/en/latest/topics/modules.html#resolv-conf

If you think that the change from this PR has caused a regression, please
file a bug at https://bugs.launchpad.net/cloud-init and include the tarball
from `cloud-init collect-logs --include-userdata` with any sensitive
userdata redacted.

Thanks,
James

xiachen-rh:
Hi James,
Thanks for your reply, and I will test the resolv conf module. 
And could you tell me why setting Azure to apply networking config every BOOT please? I would like to know more background information about this PR.

Thanks,
Amy

anhvoms:
@xiachen-rh Upon every boot, cloud-init needs to fetch network config again to make sure it has the most up-to-date network config for the VM because the user might add/remove network interface when the VM is stopped

xiachen-rh:
@anhvoms I compared to other Cloud datasource file, and found that it seems only Azure and AWS Non-VPC (aka Classic) Ec2 instances need to rewrite network config file every boot.  The AWS document shows that  EC2-Classic network is going to retired (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-classic-platform.html).  So my question is, why Azure needs to apply networking config every BOOT but other Cloud platforms don't need?

TheRealFalcon:
@xiachen-rh , Anh answered your question. Some clouds issue a new instance id when network interfaces are changed on the system, causing cloud-init to refetch networking config. Azure does not, so it needs to fetch networking config every boot. This PR is not new behavior; it is a fix to me breaking this behavior in #1006. It's also important to know that until a few releases ago, Azure instances contained a script on boot that would remove cloud-init's cache such that every boot fetched networking data regardless of this value.

If something is broken, it'd be more helpful to file a bug that describes what behavior was working along with the first version you see broken. Include the results from `cloud-init collect-logs --include-userdata` whenever possible.

zhaohuijuan:
@TheRealFalcon @anhvoms , thank you for the updates and explanations about this change, but as @xiachen-rh mentioned, 
 there are indeed some requirements from customers to persist their customized network config after reboot. 
We can provide workaround for some customers' requirements, but as all the network configs will be over-written after reboot,  maybe one specific workaround can not cover all scenarios, so I think we'd better considering how to enhance this patch and support the requirements. I reported a bug https://bugs.launchpad.net/cloud-init/+bug/2003048 to track this, could you please take a look? 
Thanks a lot!

PR #1029, opened 2021-09-21
tox: bump the pinned flake8 and pylint version (SC-422)

paride:
## Proposed Commit Message

```
tox: bump the pinned flake8 and pylint version

* pylint: fix W1406 (redundant-u-string-prefix)

The u prefix for strings is no longer necessary in Python >=3.0.

* pylint: disable W1514 (unspecified-encoding)
    
From https://www.python.org/dev/peps/pep-0597/ (Python 3.10):
    
The new warning stems form https://www.python.org/dev/peps/pep-0597,
which says:
    
  Developers using macOS or Linux may forget that the default encoding
  is not always UTF-8. [...] Even Python experts may assume that the
  default encoding is UTF-8. This creates bugs that only happen on Windows.
    
The warning could be fixed by always specifying encoding='utf-8',
however we should be careful to not break environments which are not
utf-8 (or explicitly state that only utf-8 is supported). Let's silence
the warning for now.

* _quick_read_instance_id: cover the case where load_yaml() returns None

Spotted by pylint:
 - E1135 (unsupported-membership-test)
 - E1136 (unsubscriptable-object)

LP: #1944414
```

## Additional Context

Spotted by the `cloud-init-style-tip` Jenkins job.

## Test Steps

Reproduce with: `tox -e tip-pylint`. Test this branch by just running `tox` from it.

## Checklist:

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -24,8 +24,9 @@ jobs=4
 # W0631(undefined-loop-variable)
 # W0703(broad-except)
 # W1401(anomalous-backslash-in-string)
+# W1514(unspecified-encoding)
 
-disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401
+disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401, W1514
 
 
 [REPORTS]
--- a/cloudinit/reporting/handlers.py
+++ b/cloudinit/reporting/handlers.py
@@ -137,8 +137,8 @@ class HyperVKvpReportingHandler(ReportingHandler):
         self._event_types = event_types
         self.q = queue.Queue()
         self.incarnation_no = self._get_incarnation_no()
-        self.event_key_prefix = u"{0}|{1}".format(self.EVENT_PREFIX,
-                                                  self.incarnation_no)
+        self.event_key_prefix = "{0}|{1}".format(self.EVENT_PREFIX,
+                                                 self.incarnation_no)
         self.publish_thread = threading.Thread(
             target=self._publish_event_routine
         )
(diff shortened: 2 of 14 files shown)


paride:
This is a WIP because I'm not sure on how to fix this:
```
************* Module cloudinit.sources.DataSourceNoCloud
cloudinit/sources/DataSourceNoCloud.py:250: [E1135(unsupported-membership-test), _quick_read_instance_id] Value 'md' doesn't support membership test
cloudinit/sources/DataSourceNoCloud.py:251: [E1136(unsubscriptable-object), _quick_read_instance_id] Value 'md' is unsubscriptable
```
This happens here: https://github.com/canonical/cloud-init/blob/e27c30748e88409b1646a552f994edf9ed9d017e/cloudinit/sources/DataSourceNoCloud.py#L250
and the reason is that `load_yaml()` can return `None`, which of course has no membership test and isn't subscriptable: https://github.com/canonical/cloud-init/blob/e27c30748e88409b1646a552f994edf9ed9d017e/cloudinit/util.py#L840

One way to fix this by applying this change to https://github.com/canonical/cloud-init/blob/e27c30748e88409b1646a552f994edf9ed9d017e/cloudinit/sources/DataSourceNoCloud.py#L250
```
-            if iid_key in md:
+            if md and iid_key in md:
                 return md[iid_key]
```
And this works, however I wonder if we should make `load_yaml()` output an empty dict by default instead of `None`. While it seems a good idea, it may be not because `yaml.safe_load("")` returns `None` and we may want to mimic that...

TheRealFalcon:
I don't think this encoding change is safe to do as a bulk change. I'm guessing that most of these calls will work fine specifying utf-8, but are we sure that every single `open` is safe? 

If I've specified my locale to use ISO-8859, are we sure that none of the files that we're reading will use ISO-8859 rather than utf-8? If we need to be explicit, then I think `locale.getpreferredencoding()` is the only safe option unless we're sure about every single open call.

For the yaml question, given that our load yaml definition is
```python
def load_yaml(blob, default=None, allowed=(dict,)):
```
I think it makes sense to keep the default as None and do None checks for what has been returned.

paride:
I see your point and I'm find dropping the `encoding=` bulk commit and disable W1514 in pylintrc. We'll have to deal with this at some point, but with more care. +1 on keeping `load_yaml()` current behavior.

TheRealFalcon (approved):
Thanks! LGTM assuming CI passes

PR #1031, opened 2021-09-21
Remove invalid ssh_import_id from examples

TheRealFalcon:
## Proposed Commit Message

```
Remove invalid ssh_import_id from examples
```

## Additional Context
`None` could be taken to mean, "this is valid but do nothing", but actually specifying None will give an error in the logs.
@akutz requested I remove the VMWare references as well.

We use `<ssh pub key 1>` in the examples, so I figure using a non-real `<Launchpad/Github id>` is fine.

### DIFF
--- a/doc/examples/cloud-config-user-groups.txt
+++ b/doc/examples/cloud-config-user-groups.txt
@@ -20,14 +20,18 @@ users:
     groups: users
     selinux_user: staff_u
     expiredate: '2032-09-01'
-    ssh_import_id: <Launchpad/Github id>
+    ssh_import_id:
+      - lp:falcojr
+      - gh:TheRealFalcon
     lock_passwd: false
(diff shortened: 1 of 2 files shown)


doc/examples/cloud-config-user-groups.txt
```diff
@@ -20,14 +20,14 @@ users:
     groups: users
     selinux_user: staff_u
     expiredate: '2032-09-01'
-    ssh_import_id: foobar
+    ssh_import_id: <Launchpad/Github id>
```
blackboxsw, line 23: These examples are incorrect as ssh_import_id needs to be a list.

Since these are two examples, maybe we could make them explicit for either gh: or lp:?
```suggestion
    ssh_import_id: [lp:falcojr, gh:TheRealFalcon]
```

blackboxsw (changes requested):
Let's make our examples the correct type since they are examples :)

TheRealFalcon:
Strings work too, but I updated the examples to be more inline with work you're looking for.

blackboxsw (approved):
Oops right! JSON schema needed here FTW :/

PR #1032, opened 2021-09-22
WIP: Resize encrypted partitions

TheRealFalcon:
## Proposed Commit Message

```
TODO...
```

## Test Steps
Can't test on a fully encrypted system yet, but approximated what we're trying to achieve on openstack with ephemeral drives.

Using this cloud-config:
```
#cloud-config
growpart:
  mode: auto
  devices: ["/", "/dev/mapper/disk2", "/dev/vdc1"]  # or /mnt instead of /dev/mapper/disk2
```

Launch an instance with:
```
openstack server create --flavor m1.small --image  827195be-5e47-40f1-8f40-32c30b4965b7 --key-name newt --nic net-id=falcojr_admin_net test-growpart --ephemeral size=1 --ephemeral size=1 --wait --user-data /home/james/my-user-data
```
add floating ip here
upload cloud-init deb and install it

Run:
```
echo '{"key":"XFmCwX2FHIQp0LBWaLEMiHIyfxt1SGm16VvUAVledlY=","slot":5}' > /cc_growpart_keydata  # write to /cc_growpart_keydata
umount /mnt
parted --script /dev/vdb mklabel gpt mkpart primary ext4 100 1000
parted --script /dev/vdc mklabel gpt mkpart primary ext4 100 1000
```

lsblk should show a /dev/vdb1 and /dev/vdc1 with 859M size of partition.
possibly reboot here because jbd2 won't let go of vdb and rerun parted

After reboot, run:
```
echo 'XFmCwX2FHIQp0LBWaLEMiHIyfxt1SGm16VvUAVledlY=' | base64 --decode | cryptsetup -y -v luksFormat /dev/vdb1 --key-file - --key-slot 5
echo 'XFmCwX2FHIQp0LBWaLEMiHIyfxt1SGm16VvUAVledlY=' | base64 --decode | cryptsetup luksOpen --key-file - /dev/vdb1 disk2
mkfs.ext4 /dev/mapper/disk2
mount /dev/mapper/disk2 /mnt
cloud-init clean --logs; cloud-init init --local; cloud-init init; cloud-init modules --mode=config; cloud-init modules --mode=final; cloud-init status --wait --long
```

lsblk should show a resized sizes for both encrypted and unencrypted device with no errors in `/var/log/cloud-init.log`

## Checklist:

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -338,7 +338,7 @@ def get_underlying_partition(blockdev):
 
 
 def resize_encrypted(blockdev):
-    subp.subp(['cryptsetup', '--key-file', '<keyfile>', 'resize', blockdev])
+    subp.subp(['cryptsetup', 'resize', blockdev])
 
 
 def resize_devices(resizer, devices):
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -63,7 +63,9 @@ growpart is::
             - "/dev/vdb1"
         ignore_growroot_disabled: <true/false>
 """
+import base64
 import copy
+import json
 import os
 import os.path
 import re
@@ -83,6 +85,8 @@ DEFAULT_CONFIG = {
     'devices': ['/'],
     'ignore_growroot_disabled': False,
 }
+
+KEYDATA_PATH = "/cc_growpart_keydata"
 
 
 class RESIZE(object):
@@ -287,35 +291,31 @@ def devent2dev(devent):
     return dev
 
 
-def is_mapped_device(blockdev) -> bool:
-    """
-    Check if a device is a mapped device.
+def get_mapped_device(blockdev):
+    """Returns underlying block device for a mapped device.
 
-    blockdev should look something like '/dev/mapper/disk1' if True,
-    otherwise something like /dev/vdb1.
+    If blockdev is a symlink pointing to a /dev/dm-* device, return
+    the device pointed to. Otherwise, return None.
     """
-    is_mapped = blockdev.startswith('/dev/mapper/')
-    LOG.debug("%s is %s a mapped device", blockdev, "" if is_mapped else "not")
-    return is_mapped
+    realpath = os.path.realpath(blockdev)
+    if realpath.startswith("/dev/dm-"):
+        LOG.debug("%s is a mapped device pointing to %s", blockdev, realpath)
+        return realpath
+    return None
 
 
 def is_encrypted(blockdev) -> bool:
     """
-    Check if a device is an encrypted device.
-
-    Will work for both mapped and raw devices.
+    Check if a device is an encrypted device. blockdev should have
+    a /dev/dm-* path.
     """
-    is_encrypted = False
-    if not subp.which('cryptsetup'):
-        LOG.debug('cryptsetup not found. Assuming no encrypted partitions')
+    if not subp.which("cryptsetup"):
+        LOG.debug("cryptsetup not found. Assuming no encrypted partitions")
         return False
+    is_encrypted = False
     with suppress(subp.ProcessExecutionError):
-        subp.subp(['cryptsetup', 'status', blockdev])
+        subp.subp(["cryptsetup", "status", blockdev])
         is_encrypted = True
-    if not is_encrypted:
-        with suppress(subp.ProcessExecutionError):
-            subp.subp(['cryptsetup', 'isLuks', blockdev])
-            is_encrypted = True
     LOG.debug(
         "Determined that %s is %s encrypted",
         blockdev,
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_growpart.py
```diff
@@ -274,12 +275,69 @@ def devent2dev(devent):
...
+        # 1 dependencies : (vdb1)
+        return '/dev/{}'.format(dep.split(': (')[1].split(')')[0])
+    except IndexError:
+        raise Exception(
+            "Ran `{}`, but received unexpected stdout: `{}`".format(
+                command, dep))
+
+
+def resize_encrypted(blockdev):
+    subp.subp(['cryptsetup', '--key-file', '<keyfile>', 'resize', blockdev])
```
TheRealFalcon, line 329: This can't work as-is, but not sure where I'm supposed to get the key from.

xnox: You would not need key-file at all, as it would be in kernel keyring already, because the drive is unlocked.

xnox: However, there is backup recovery keyfile on disk too

TheRealFalcon: Ok, I pushed a change to remove that argument. It should be testable now.

cloudinit/config/cc_growpart.py
```diff
@@ -274,12 +275,69 @@ def devent2dev(devent):
...
+        "Determined that %s is %s encrypted",
+        blockdev,
+        "" if is_encrypted else "not"
+    )
+    return is_encrypted
+
+
+def get_underlying_partition(blockdev):
+    command = 'dmsetup deps -o devname {}'.format(blockdev)
+    dep = subp.subp(command.split())[0]
```
smoser, line 317: Generally we should prefer the following for which does not use split or format and uses long format arguments (readability).

```python
dep = subp.subp(["dmsetup", "deps", "--options=devname", blockdev])[0]
```

Also, you really should check the error code... if this returns non-zero, and you try to parse stdout, then you'll likely get an IndexError, and raise a non-helpful exception.

TheRealFalcon: Thanks. Was planning some additional error checking in other places too...though...subp will raise error in non-0 exits already. I just don't remember how ugly that is in the logs.

cloudinit/config/cc_growpart.py
```diff
@@ -286,12 +287,69 @@ def devent2dev(devent):
...
 
 
+def is_mapped_device(blockdev) -> bool:
+    """
+    Check if a device is a mapped device.
+
+    blockdev should look something like '/dev/mapper/disk1' if True,
+    otherwise something like /dev/vdb1.
+    """
+    is_mapped = blockdev.startswith('/dev/mapper/')
```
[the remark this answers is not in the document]
mwhudson: /dev/mapper/whatever is a symlink to the underlying block device. It depends a bit what you're really asking here but "os.path.realpath(blockdev).startswith('dm-')" is one possible implementation of this logic that's better than what you have here.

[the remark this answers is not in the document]
chrisccoulson: Yes, sorry it wasn't clear - the block device path you get could be any one of the user-space managed links in /dev/mapper or /dev/disk/ - the more correct thing to do is to resolve the symbolic link to obtain the actual block device path, and then check if the path you get begins with /dev/dm-

cloudinit/config/cc_growpart.py
```diff
@@ -286,12 +287,69 @@ def devent2dev(devent):
...
+    is_encrypted = False
+    if not subp.which('cryptsetup'):
+        LOG.debug('cryptsetup not found. Assuming no encrypted partitions')
+        return False
+    with suppress(subp.ProcessExecutionError):
+        subp.subp(['cryptsetup', 'status', blockdev])
+        is_encrypted = True
+    if not is_encrypted:
+        with suppress(subp.ProcessExecutionError):
+            subp.subp(['cryptsetup', 'isLuks', blockdev])
```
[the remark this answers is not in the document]
TheRealFalcon: @chrisccoulson The goal here was to support both types of paths. If `cryptsetup status` passed, then we return `True`. If not, check to see if we've been passed the underlying device and try `cryptsetup isLuks` instead. Given how I'm passing things currently, that `isLusk` check is kind of pointless so I can just remove it.

Would it be valid to simply call `cryptsetup status` on the device, and return True if that returns 0?

TheRealFalcon:
PR updated based on comments. cloud-init now reads the key off a json file on disk as suggested.

holmanb (commented):
So far this looks good. I made one comment. I assume that testing/docs are planned prior to merge.

I don't see any integration tests for this module so far. I don't know if that's due to the complexity of LXC with filsystems or if we just never got around to it before. Assuming we want to add them, one idea is we could possibly try to mount a loop device (file backed drive) into an LXC container as a block device for testing. If LXC can't handle something about that, qemu would work, (but we should try LXC first I think to avoid adding a dependency to the integration tests). If help is wanted with integration tests for this, feel free to ping.

PR #1034, opened 2021-09-24
Wait for apt lock

TheRealFalcon:
## Proposed Commit Message

```
Wait for apt lock

Currently any attempt to run an apt command while another process holds
an apt lock will fail. We should instead wait to acquire the apt lock.

LP: #1944611
```

## Additional Context
See https://bugs.launchpad.net/cloud-init/+bug/1944611 . I have also seen this intermittently while testing.

## Test Steps
Beyond the unit tests, start an instance with the following cloud-config
```
#cloud-config
bootcmd:
  - python3 -c "import fcntl; from time import sleep; x = open('/var/lib/dpkg/lock', 'w'); fcntl.lockf(x, fcntl.LOCK_EX | fcntl.LOCK_NB); sleep(50000)" &
drivers:
  nvidia:
    license-accepted: true
```
Old behavior is to fail immediately. With this PR it will wait 30 seconds and then fail. Decrease the sleep an appropriate amount of time if you'd like to see it wait and then proceed when the lock is cleared.

## Checklist:

### DIFF
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -208,7 +208,7 @@ class Distro(distros.Distro):
                     func=subp.subp,
                     kwargs=subp_kwargs,
                 )
-            except subp.ProcessExecutionError as e:
+            except subp.ProcessExecutionError:
                 # Even though we have already waited for the apt lock to be
                 # available, it is possible that the lock was acquired by
                 # another process since the check. Since apt doesn't provide
@@ -220,7 +220,7 @@ class Distro(distros.Distro):
                 # raced us when we tried to acquire it, so raise the apt
                 # error received. If the lock is unavailable, just keep waiting
                 if self._apt_lock_available():
-                    raise e
+                    raise
                 LOG.debug('Another process holds apt lock. Waiting...')
                 time.sleep(1)
         raise TimeoutError('Could not get apt lock')
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -177,7 +177,7 @@ class Distro(distros.Distro):
             with open(lock, 'w') as handle:
                 try:
                     fcntl.lockf(handle, fcntl.LOCK_EX | fcntl.LOCK_NB)
-                except IOError:
+                except OSError:
                     return False
         return True
 
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -7,8 +7,10 @@
 # Author: Joshua Harlow <harlowja@yahoo-inc.com>
 #
 # This file is part of cloud-init. See LICENSE file for license information.
-
+import fcntl
 import os
+import time
+from contextlib import ExitStack
 
 from cloudinit import distros
 from cloudinit import helpers
@@ -22,6 +24,7 @@ from cloudinit.settings import PER_INSTANCE
 
 LOG = logging.getLogger(__name__)
 
+APT_LOCK_WAIT_TIMEOUT = 30
 APT_GET_COMMAND = ('apt-get', '--option=Dpkg::Options::=--force-confold',
                    '--option=Dpkg::options::=--force-unsafe-io',
                    '--assume-yes', '--quiet')
@@ -40,6 +43,12 @@ NETWORK_FILE_HEADER = """\
 
 NETWORK_CONF_FN = "/etc/network/interfaces.d/50-cloud-init"
 LOCALE_CONF_FN = "/etc/default/locale"
+
+APT_LOCK_FILES = [
+    '/var/lib/dpkg/lock',
+    '/var/lib/apt/lists/lock',
+    '/var/cache/apt/archives/lock',
+]
 
 
 class Distro(distros.Distro):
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
Currently, the sleep time is hardcoded, which isn't ideal. I know that we're generally against adding top-level keys, but there's already some precedent for having not-really-documented top-level apt keys. I'd prefer we find a way to document them better, but I'm not sure it would be all that bad to add something like `apt_lock_timeout` alongside the already existing `apt_get_command` and `apt_get_wrapper`.

We could also put it under the `apt` top-level key, but that feels like adding configuration that belongs to another module, so I don't really like that option.

I'm open to any strong opinions.

cloudinit/distros/debian.py
```diff
@@ -155,7 +164,70 @@ def _get_localhost_ip(self):
...
+                    ' '.join(subp_kwargs['args'])
+                )
+                return util.log_time(
+                    logfunc=LOG.debug,
+                    msg=log_msg,
+                    func=subp.subp,
+                    kwargs=subp_kwargs,
+                )
+            except subp.ProcessExecutionError as e:
+                if e.exit_code == 100 and 'Could not get lock' in e.stderr:
```
TheRealFalcon, line 216: Does this pose any internationalization concerns?

cloudinit/distros/debian.py
```diff
@@ -155,7 +164,72 @@ def _get_localhost_ip(self):
...
+        short_cmd: Name of command like "upgrade" or "install"
+        subp_kwargs: kwargs to pass to subp
+        """
+        start_time = time.time()
+        LOG.debug('Waiting for apt lock')
+        while time.time() - start_time < timeout:
+            if not self._is_apt_lock_available():
+                time.sleep(1)
+                continue
+            LOG.debug('apt lock clear')
```
blackboxsw, line 204:
```suggestion
            LOG.debug('apt lock available')
```

cloudinit/distros/debian.py
blackboxsw: A couple of concerns: 
 - Agree that hard-coding English strings to scope our logging of the exception reason is ineffective without using gettext.
 - I'm not sure this logging warrants us digging into gettext on the apt and apt-utils domain to do translations on this expected string type for just this log.
 - This loop looks to sleep and retry for 30 seconds on all CalledProcessErrors. Shouldn't we be raising and not retrying the called processerrors if exit is != 100 or if _is_apt_lock_available is True yet we received a CalledProcessError?

I'm not quite understanding what you are expecting to handle in this except clause:
- Are you thinking that our _is_apt_lock_available is not enough of a pre-flight check on obtaining- the locks or 
- Are you trying to handle a race where some other process obtained the APT lock between our success on _is_apt_lock_available and our subsequent call of subp?

cloudinit/distros/debian.py
```diff
@@ -155,7 +164,72 @@ def _get_localhost_ip(self):
...
+                    # have to wait for it
+                    pass
+            for handle in file_handles:
+                try:
+                    fcntl.lockf(handle, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except IOError:
+                    return False
+        return True
+
+    def _wait_for_apt_install(
```
blackboxsw, line 190: rename? since this is broader than install?
```suggestion
    def _wait_for_apt_lock(
```

blackboxsw (changes requested):
thanks @TheRealFalcon only significant question I have is whether our `_wait_for_apt_install` loop needs to raise in unexpected (non-lock related) errors? Exit 100 is raised in instances like invalid apt sources.list which won't be fixed by retries: 
```E: The repository 'http://archive.ubuntu.com/ubuntu2 xenial Release' does not have a Release file.```

File: cloudinit/distros/debian.py [hunk 4]
```diff
@@ -155,7 +164,72 @@ def _get_localhost_ip(self):
...
+                    file_handles.append(stack.enter_context(open(lock, 'w')))
+                except FileNotFoundError:
+                    # If the lock file doesn't even exist yet, then we don't
+                    # have to wait for it
+                    pass
+            for handle in file_handles:
+                try:
+                    fcntl.lockf(handle, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except IOError:
+                    return False
```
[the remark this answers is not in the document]
TheRealFalcon: Yes, mine was definitely more convoluted. Thanks!

TheRealFalcon:
> This loop looks to sleep and retry for 30 seconds on all CalledProcessErrors. Shouldn't we be raising and not retrying the called processerrors if exit is != 100 or if _is_apt_lock_available is True yet we received a CalledProcessError?

Yep, I forgot to add the raise. It's added now!

> I'm not quite understanding what you are expecting to handle in this except clause:
Are you thinking that our _is_apt_lock_available is not enough of a pre-flight check on obtaining- the locks or
Are you trying to handle a race where some other process obtained the APT lock between our success on _is_apt_lock_available and our subsequent call of subp?

The second option. Since snap is doing things at the same time as us, I don't think it's out of the realm of possibility for us both to attempt to do an update at the same time. I could have removed the lock check entirely and retry on apt failure, but  that would produce a lot of log spam and needless apt runs, so I have both here.

Do you think checking for a race condition is unnecessary? It would remove the exception text checking, which I'm not exactly sure the best way to approach. Though...I did have fun playing with apt in French, and that particular error message came back in English :smiley:

blackboxsw (approved):
Thanks for the gap handling there. Validated lock handling races are coped with in this case and can see retries amid intermittent apt upgrade call collisions.

julian-klode:
This change is wrong and breaks parallel apt processes. You need to acquire the frontend-lock before acquiring any other locks.

julian-klode:
The basic way to lock things is:

1. `/var/lib/dpkg/lock-frontend`
2. `/var/lib/dpkg/lock`
3. `/var/cache/apt/archives`

Presumably it's enough to just lock the frontend one.

`/var/lib/apt/lists` is locked independently of that install chain, and only locked during update, so you can acquire it either order. Also update does not acquire the dpkg frontend lock.

Presumably prepending `/var/lib/dpkg/lock-frontend` to the list of locks will cause the correct behavior.

TheRealFalcon:
Thanks @julian-klode ! We're still supporting 18.04, but it's good to know that once that goes EOL the locking code can be removed.

Is there any documentation for this locking behavior? I'm still not entirely sure what each lock is responsible for.

julian-klode:
We only have waiting for parallel install though (it only waits for the frontend lock), not update.

`/var/lib/dpkg/lock-frontend` ensures nobody else tries to run dpkg while apt runs dpkg (whether it's the user, another apt, or another dpkg frontend). This is important because apt runs dpkg multiple times, so  `/var/lib/dpkg/lock` is released in between - it locks an individual dpkg run. Before the frontend lock, you could end up losing the lock in the middle of an install/upgrade (between two dpkg runs).

`/var/lib/apt/lists/lock` is locked when metadata is being downloaded. It is not locked during metadata access, which might cause some race conditions. Wondering if I should try acquire that lock when opening the cache, but it might be hard to do. Also would be nice to add waiting for that one into apt.

`/var/cache/apt/archives/lock` is locked once we start downloading packages and held while packages are being installed.

(Both are the same lock, just for different download areas - package lists vs packages).

You'll mostly get by locking just frontends and the lists lock, but oh well, some other frontends might behave badly and not support the frontend lock themselves, and I'm not sure where it's all backported too. I think most Ubuntu releases, but not 100% sure. In any case, no issue acquiring it always.

TheRealFalcon:
Thanks! That's very helpful.

PR #1043, opened 2021-09-30
Pin jsonschema to 3.2.0 (SC-466)

TheRealFalcon:
## Proposed Commit Message

```
Pin jsonschema to 3.2.0

On unit tests, tox is attempting to install 4.0, which fails two of
the unit tests, and fails python 3.5 as it is not compatible.
```

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,7 +31,7 @@ requests
 jsonpatch
 
 # For validating cloud-config sections per schema definitions
-jsonschema
+jsonschema==3.2.0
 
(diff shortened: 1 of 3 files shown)


blackboxsw (approved):
Landing this as it unblocks CI, but the 4.0.1 release (released yesterday)  of jsonschema no longer rasises expected schemavalidation errrors. due to 'format': 'hostname' declarations. I think we should sort that inconsistency with a followup PR.

PR #1067, opened 2021-10-13
Update Azure _unpickle (SC-500)

TheRealFalcon:
## Proposed Commit Message

```
Update Azure _unpickle

When self.failed_desired_api_version was added to DataSourceAzure, the
attribute was never added to the _unpickle method using the upgrade
framework. This commit adds the attribute.

LP: #1946644
```

## Test Steps
Unfortunately, there's no easy way to test this via unit or integration test. Using a pre-existing pickle doesn't work because specific things like instance id or network interfaces will be different instance to instance and raise other issues.

To test manually:
Launch an Azure Bionic VM
On the instance, 'rm /lib/systemd/system/cloud-init-local.service.d/50-azure-clear-persistent-obj-pkl.conf'
Checkout cloud-init source at the 19.1 tag.
Build a package with the source (DEB_BUILD_OPTIONS=nocheck packages/bddeb -d)
Push the deb to the azure instance and install it
cloud-init clean --logs --reboot
Similarly, create deb of this branch and push and install
DON'T cloud-init clean
reboot, verify no failure due to attributes not found in Azure datasource.

## Checklist:

### DIFF
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -84,17 +84,19 @@ def test_clean_boot_of_upgraded_package(session_cloud: IntegrationCloud):
                 'There were errors/warnings/tracebacks pre-upgrade. '
                 'Any failures may be due to pre-upgrade problem')
 
-        # Upgrade and reboot
+        # Upgrade
         instance.install_new_cloud_init(source, take_snapshot=False)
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
@blackboxsw ,  yeah, you're definitely right...especially given that comment in the code stating such :man_facepalming: . I updated the test accordingly.

PR #1065, opened 2021-10-12
Fix unhandled apt_configure case (SC-526)

holmanb:
## Proposed Commit Message
```
Fix schema and fix unhandled apt_configure case.

Don't throw an exception when mirror arch is unspecified.
```

## Test Steps
Use the following [config](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#additional-apt-configuration-and-repositories).

## Checklist:
 - [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [X] I have updated or added any unit tests accordingly
 - [X] I have updated or added any documentation accordingly

PR #1068, opened 2021-10-14
Remove (deprecated) apt-key (SC-89)

holmanb:
```
Remove apt-key dependency.

Add the "signed by" option to source definitions. This enables users
to limit the scope of trust for individual keys.

LP: #1836336
```

## Test Steps
Use [this example config](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#additional-apt-configuration-and-repositories)* from the docs. Start an instance and run `apt-key list`. Note that the keys from the config were installed in `/etc/apt/trusted.gpg.d/`.

*Note: this example's config is technically incorrect and may throw an exception, this should get fixed in [PR #1065](https://github.com/canonical/cloud-init/pull/1065), but if you happen to test before that PR gets merged either define an architecture in `security` or remove the security source to avoid the . 

## Checklist:

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -1158,8 +1158,8 @@ def apt_key(command, output_file=None, data=None, hardened=False,
         """
         file_name = '/dev/null'
         if not output_file:
-            util.logexc(LOG, 'Unknown filename, failed to add key: {}'.format(
-                data))
+            util.logexc(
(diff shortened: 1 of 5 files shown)


TheRealFalcon (approved):
Thanks @holmanb !

julian-klode:
This does not seem to be super useful. It still uses gpg, the reason apt-key goes away is because gpg is not needed anymore, there should really be no gpg on images anymore.

PR #1074, opened 2021-10-20
Leave the details of service management to the distro.

citrus-it:
I allowed the previous PR at https://github.com/canonical/cloud-init/pull/1014 to become stale. Re-opening as requested.

## Proposed Commit Message

```
Leave the details of service management to the distro.

Various modules restart services and they all have logic to try and
detect if they are running on a system that needs 'systemctl' or
'service', and then have code to decide which order the arguments
need to be etc. On top of that, not all modules do this in the same way.

The duplication and different approaches are not ideal but this also
makes it hard to add support for a new distribution that does not use
either 'systemctl' or 'service'.

This change adds a new manage_service() method to the distro class
and updates several modules to use it.
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_fan.py
+++ b/cloudinit/config/cc_fan.py
@@ -52,35 +52,26 @@ BUILTIN_CFG = {
 }
 
 
-def stop_update_start(service, config_file, content, systemd=False):
-    if systemd:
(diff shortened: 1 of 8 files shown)


citrus-it:
Thanks for your help.

PR #1087, opened 2021-10-28
Add necessary mocks to test_ovf unit tests

TheRealFalcon:
We likely have a test somewhere not cleaning up its mocks causing tests to run after it to not have to mock things that should be mocked. When run in isolation, these tests will fail. We need to find this test and fix it, but for now, fix the lack of mocks.


blackboxsw (commented):
+1 on this. Though I think we probably should run through all tests independently to sort out other "leaks":
 I think the following also fail:
FAILED tests/unittests/test_data.py
FAILED tests/unittests/test_distros/test_netconfig.py

blackboxsw (approved):
+1 on this fix. confirmed fix locally. there might be a few others to capture later.

PR #1088, opened 2021-10-29
Allow libexec for hotplug (SC-532)

TheRealFalcon:
## Proposed Commit Message

```
Allow libexec for hotplug

When we added the install hotplug module, we forgot to update the
redhet/cloud-init.spec.in file and allow for execution on /usr/libexec.
This PR adds that functionality
```

## Test Steps
Verified rpm package builds, `/etc/udev/rules.d/10-cloud-init-hook-hotplug.rules` gets written and references `/usr/libexec/cloud-init/hoook-hotplug` and `/usr/libexec/cloud-init/hook-hotplug` exists. Also tested that integration test still passes on ubuntu.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_install_hotplug.py
+++ b/cloudinit/config/cc_install_hotplug.py
@@ -85,11 +85,11 @@ __doc__ = get_schema_doc(schema)
 
 
 HOTPLUG_UDEV_PATH = "/etc/udev/rules.d/10-cloud-init-hook-hotplug.rules"
-HOTPLUG_UDEV_RULES = """\
+HOTPLUG_UDEV_RULES_TEMPLATE = """\
 # Installed by cloud-init due to network hotplug userdata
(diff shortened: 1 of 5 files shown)


cloudinit/config/cc_install_hotplug.py
```diff
@@ -129,8 +129,12 @@ def handle(_name, cfg, cloud, log, _args):
         log.debug("Skipping hotplug install, udevadm not found")
         return
 
+    # This may need to turn into a distro property at some point
```
blackboxsw, line 132: +1 I'm good with this approach. Checking /usr/libexec dir first would allow us to prefer that on distros that have installed cloud-init to that path.

blackboxsw (approved):
+1 this is a reasonable runtime approach to installing to the right path

PR #1091, opened 2021-11-01
Release 21.4

TheRealFalcon:


blackboxsw (approved):
Changeset looks fine. But we haven't yet run a successful "daily" build with the LXD datasource commit enabled. I just kicked all daily recipe package builds for bionic, focal, hirsute,impish and devel so that our next "*-daily" integration test run uses the latest bits.

PR #1092, opened 2021-11-02
testing: Remove calls to 'install_new_cloud_init'

TheRealFalcon:
## Proposed Commit Message

```
testing: Remove calls to 'install_new_cloud_init'

In our integration tests, a few tests were modifying the environment and
then calling 'install_new_cloud_init'. This is problematic because it
updates the environment for all future tests.

Existing instances of 'install_new_cloud_init' aren't problematic because
they aren't modifying the underlying environment.
```

## Test Steps
Since this involves one test affecting tests run after it, the whole suite of tests was run for lxd containers. I also verified the updated GCE test on GCE.

## Checklist:

### DIFF
--- a/tests/integration_tests/datasources/test_lxd_discovery.py
+++ b/tests/integration_tests/datasources/test_lxd_discovery.py
@@ -2,30 +2,17 @@ import json
 import pytest
 import yaml
 
-from tests.integration_tests.clouds import IntegrationCloud
-from tests.integration_tests.conftest import get_validated_source
+from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.util import verify_clean_log
 
 
-def _setup_custom_image(session_cloud: IntegrationCloud):
-    """Like `setup_image` in conftest.py, but with customized content."""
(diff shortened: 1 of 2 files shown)


tests/integration_tests/datasources/test_lxd_discovery.py
```diff
@@ -2,79 +2,61 @@
...
+    client.restart()
 
 
 # This test should be able to work on any cloud whose datasource specifies
 # a NETWORK dependency
 @pytest.mark.lxd_container
 @pytest.mark.lxd_vm
 @pytest.mark.ubuntu  # Because netplan
-def test_lxd_datasource_discovery(session_cloud: IntegrationCloud):
+def test_lxd_datasource_discovery(client: IntegrationInstance):
```
TheRealFalcon, line 23: Most of the changes in this test are just indentation

TheRealFalcon:
I had a successful local lxd_container run

blackboxsw (approved):
Thanks for the fix. Makes sense here that we no longer modify the base image under test :(.

PR #1093, opened 2021-11-02
Anhvo/pps imds

anhvoms:
## Proposed Commit Message
Azure: Support pre-provisioning for images without UDF support

```
Without UDF support, DS Azure cannot mount the provisioning ISO,
which contains platform metadata necessary to support
pre-provisioning. The required metadata is made available in IMDS 
starting with api version 2021-08-01. This change will leverage IMDS
to obtain the required metadata to support pre-preprovisioning if 
provisioning ISO was not available.
```

## Checklist:


TheRealFalcon (approved):
Thanks!

PR #1101, opened 2021-11-05
Add Strict Metaschema Validation

holmanb:
## Proposed Commit Message

```
Improve schema validation.

This adds strict validation of config module definitions at testing
time, with plumbing included for future runtime validation. This
eliminates a class of bugs resulting from schemas that have definitions
that are incorrect, but get interpreted by jsonschema as
"additionalProperties" that are therefore ignored.


- Add strict meta-schema for jsonschema unit test validation
- Separate schema from module metadata structure
- Improve type annotations for various functions and data types

Cleanup:
- Remove unused jsonschema "required" elements 
- Eliminate manual memoization in schema.py:get_schema(),
        reference module.__doc__ directly
```

## Test Steps
Many of the files were only touched for function name/arg/schema/meta updates, however schema.py, test_schema.py, test_cli.py had the most substantial changes.

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -34,7 +34,7 @@ SCHEMA_DOC_TMPL = """
 {property_doc}
 {examples}
 """
-SCHEMA_PROPERTY_TMPL = '{prefix}**{prop_name}:** ({prop_type}) {description}'
+SCHEMA_PROPERTY_TMPL = "{prefix}**{prop_name}:** ({prop_type}) {description}"
 SCHEMA_LIST_ITEM_TMPL = (
     '{prefix}Each item in **{prop_name}** list supports the following keys:')
 SCHEMA_EXAMPLES_HEADER = '\n**Examples**::\n\n'
@@ -105,9 +105,7 @@ def get_jsonschema_validator():
     return (cloudinitValidator, FormatChecker, None)
 
 
-def validate_cloudconfig_metaschema(
-        schema: dict,
-        throw=True):
+def validate_cloudconfig_metaschema(schema: dict, throw=True):
     """Validate provided schema meets the metaschema definition. Return strict
     Validator and FormatChecker for use in validation
     @param schema: schema to validate
@@ -121,11 +119,12 @@ def validate_cloudconfig_metaschema(
     """
 
     from jsonschema.exceptions import SchemaError
+
     (cloudinitValidator, FormatChecker, _) = get_jsonschema_validator()
     try:
 
         # disable bottom-level keys
-        cloudinitValidator.META_SCHEMA['additionalProperties'] = False
+        cloudinitValidator.META_SCHEMA["additionalProperties"] = False
         cloudinitValidator.check_schema(schema)
     except SchemaError as e:
         if throw:
@@ -135,10 +134,8 @@ def validate_cloudconfig_metaschema(
 
 
 def validate_cloudconfig_schema(
-        config: dict,
-        schema: dict,
-        strict=False,
-        strict_metaschema=False):
+    config: dict, schema: dict, strict=False, strict_metaschema=False
+):
     """Validate provided config meets the schema definition.
 
     @param config: Dict of cloud configuration settings validated against
@@ -160,13 +157,14 @@ def validate_cloudconfig_schema(
             else validate_cloudconfig_metaschema(schema, throw=False)
         )
     except ImportError:
-        logging.debug(
-            'Ignoring schema validation. jsonschema is not present')
+        logging.debug("Ignoring schema validation. jsonschema is not present")
         return
     if err:
         logging.warning(
-            'Meta-schema validation failed, attempting to validate config'
-            ' anyway: %s', err)
+            "Meta-schema validation failed, attempting to validate config"
+            " anyway: %s",
+            err,
+        )
     validator = cloudinitValidator(schema, format_checker=FormatChecker())
     errors = ()
     for error in sorted(validator.iter_errors(config), key=lambda e: e.path):
@@ -361,10 +359,11 @@ def _get_property_type(property_dict: dict) -> str:
     """Return a string representing a property type from a given
     jsonschema.
     """
-    property_type = property_dict.get('type')
-    if property_type is None and property_dict.get('enum'):
+    property_type = property_dict.get("type")
+    if property_type is None and property_dict.get("enum"):
         property_type = [
-            str(_YAML_MAP.get(k, k)) for k in property_dict['enum']]
+            str(_YAML_MAP.get(k, k)) for k in property_dict["enum"]
+        ]
     if isinstance(property_type, list):
         property_type = '/'.join(property_type)
     items = property_dict.get('items', {})
@@ -375,8 +374,8 @@ def _get_property_type(property_dict: dict) -> str:
             sub_property_type += '/'
         sub_property_type += '(' + _get_property_type(sub_item) + ')'
     if sub_property_type:
-        return '{0} of {1}'.format(property_type, sub_property_type)
-    return property_type or 'UNDEFINED'
+        return "{0} of {1}".format(property_type, sub_property_type)
+    return property_type or "UNDEFINED"
 
 
 def _parse_description(description, prefix) -> str:
@@ -401,7 +400,7 @@ def _parse_description(description, prefix) -> str:
     return description
 
 
-def _get_property_doc(schema: dict, prefix='    ') -> str:
+def _get_property_doc(schema: dict, prefix="    ") -> str:
     """Return restructured text describing the supported schema properties."""
     new_prefix = prefix + '    '
     properties = []
@@ -410,12 +409,15 @@ def _get_property_doc(schema: dict, prefix='    ') -> str:
         description = prop_config.get('description', '')
 
         # Define prop_name and description for SCHEMA_PROPERTY_TMPL
-        properties.append(SCHEMA_PROPERTY_TMPL.format(
-            prefix=prefix,
-            prop_name=prop_key,
-            description=_parse_description(description, prefix),
-            prop_type=_get_property_type(prop_config)))
-        items = prop_config.get('items')
+        properties.append(
+            SCHEMA_PROPERTY_TMPL.format(
+                prefix=prefix,
+                prop_name=prop_key,
+                description=_parse_description(description, prefix),
+                prop_type=_get_property_type(prop_config),
+            )
+        )
+        items = prop_config.get("items")
         if items:
             if isinstance(items, list):
                 for item in items:
@@ -434,7 +436,7 @@ def _get_property_doc(schema: dict, prefix='    ') -> str:
 
 def _get_examples(meta: MetaSchema) -> str:
     """Return restructured text describing the meta examples if present."""
-    examples = meta.get('examples')
+    examples = meta.get("examples")
     if not examples:
         return ''
     rst_content = SCHEMA_EXAMPLES_HEADER
@@ -458,27 +460,31 @@ def get_meta_doc(meta: MetaSchema, schema: dict) -> str:
 
     # Some modules set this to falsy value, exit gracefully
     if not meta or not schema:
-        return ''
+        return ""
     keys = set(meta.keys())
-    expected = set({
-        'id',
-        'title',
-        'examples',
-        'frequency',
-        'distros',
-        'description',
-        'name'})
+    expected = set(
+        {
+            "id",
+            "title",
+            "examples",
+            "frequency",
+            "distros",
+            "description",
+            "name",
+        }
+    )
     if keys != expected:
         raise KeyError(
-            'Missing module metadata key(s) {}'.format(expected - keys))
+            "Missing module metadata key(s) {}".format(expected - keys)
+        )
 
     # cast away type annotation
     meta_copy = dict(deepcopy(meta))
-    meta_copy['property_doc'] = _get_property_doc(schema)
-    meta_copy['examples'] = _get_examples(meta)
-    meta_copy['distros'] = ', '.join(meta['distros'])
+    meta_copy["property_doc"] = _get_property_doc(schema)
+    meta_copy["examples"] = _get_examples(meta)
+    meta_copy["distros"] = ", ".join(meta["distros"])
     # Need an underbar of the same length as the name
-    meta_copy['title_underbar'] = re.sub(r'.', '-', meta['name'])
+    meta_copy["title_underbar"] = re.sub(r".", "-", meta["name"])
     template = SCHEMA_DOC_TMPL.format(**meta_copy)
     return template
 
@@ -494,20 +500,24 @@ def error(message):
 
 
 def load_doc(requested_modules: list) -> str:
-    '''Load module docstrings
+    """Load module docstrings
 
     Docstrings are generated on module load. Reduce, reuse, recycle.
-    '''
-    docs = ''
-    all_modules = list(get_modules().values()) + ['all']
+    """
+    docs = ""
+    all_modules = list(get_modules().values()) + ["all"]
     invalid_docs = set(requested_modules).difference(set(all_modules))
     if invalid_docs:
-        error('Invalid --docs value {}. Must be one of: {}'.format(
-            list(invalid_docs), ', '.join(all_modules)))
+        error(
+            "Invalid --docs value {}. Must be one of: {}".format(
+                list(invalid_docs), ", ".join(all_modules)
+            )
+        )
     for mod_name in all_modules:
-        if 'all' in requested_modules or mod_name in requested_modules:
+        if "all" in requested_modules or mod_name in requested_modules:
             (mod_locs, _) = importer.find_module(
-                mod_name, ['cloudinit.config'], ['schema'])
+                mod_name, ["cloudinit.config"], ["schema"]
+            )
             if mod_locs:
                 mod = importer.import_module(mod_locs[0])
                 docs += mod.__doc__ or ""
@@ -517,15 +527,18 @@ def load_doc(requested_modules: list) -> str:
 def get_schema() -> dict:
     """Return jsonschema coalesced from all cc_* cloud-config module."""
     full_schema = {
-        '$schema': 'http://json-schema.org/draft-04/schema#',
-        'id': 'cloud-config-schema', 'allOf': []}
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "id": "cloud-config-schema",
+        "allOf": [],
+    }
 
     for (_, mod_name) in get_modules().items():
         (mod_locs, _) = importer.find_module(
-            mod_name, ['cloudinit.config'], ['schema'])
+            mod_name, ["cloudinit.config"], ["schema"]
+        )
         if mod_locs:
             mod = importer.import_module(mod_locs[0])
-            full_schema['allOf'].append(mod.schema)
+            full_schema["allOf"].append(mod.schema)
     return full_schema
 
 
@@ -534,10 +547,11 @@ def get_meta() -> dict:
     full_meta = dict()
     for (_, mod_name) in get_modules().items():
         mod_locs, _ = importer.find_module(
-            mod_name, ['cloudinit.config'], ['meta'])
+            mod_name, ["cloudinit.config"], ["meta"]
+        )
         if mod_locs:
             mod = importer.import_module(mod_locs[0])
-            full_meta[mod.meta['id']] = mod.meta
+            full_meta[mod.meta["id"]] = mod.meta
     return full_meta
 
 
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -4,16 +4,19 @@
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit import importer
 from cloudinit.importer import MetaSchema
-from cloudinit.util import find_modules, load_file
+from cloudinit.util import find_modules, load_file, error
 
 import argparse
 from collections import defaultdict
 from copy import deepcopy
+from functools import partial
 import logging
 import os
 import re
 import sys
 import yaml
+
+error = partial(error, sys_exit=True)
 
 _YAML_MAP = {True: 'true', False: 'false', None: 'null'}
 CLOUD_CONFIG_HEADER = b'#cloud-config'
@@ -493,11 +496,6 @@ def get_modules() -> dict:
     return find_modules(configs_dir)
 
 
-def error(message):
-    print(message, file=sys.stderr)
-    sys.exit(1)
-
-
 def load_doc(requested_modules: list) -> str:
     """Load module docstrings
 
@@ -509,7 +507,7 @@ def load_doc(requested_modules: list) -> str:
     if invalid_docs:
         error(
             "Invalid --docs value {}. Must be one of: {}".format(
-                list(invalid_docs), ", ".join(all_modules)
+                list(invalid_docs), ", ".join(all_modules),
             )
         )
     for mod_name in all_modules:
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -3,6 +3,7 @@
 
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit import importer
+from cloudinit.importer import MetaSchema
 from cloudinit.util import find_modules, load_file
 
 import argparse
@@ -15,7 +16,6 @@ import sys
 import yaml
 
 _YAML_MAP = {True: 'true', False: 'false', None: 'null'}
-SCHEMA_UNDEFINED = b'UNDEFINED'
 CLOUD_CONFIG_HEADER = b'#cloud-config'
 SCHEMA_DOC_TMPL = """
 {name}
@@ -34,7 +34,7 @@ SCHEMA_DOC_TMPL = """
 {property_doc}
 {examples}
 """
-SCHEMA_PROPERTY_TMPL = '{prefix}**{prop_name}:** ({type}) {description}'
+SCHEMA_PROPERTY_TMPL = '{prefix}**{prop_name}:** ({prop_type}) {description}'
 SCHEMA_LIST_ITEM_TMPL = (
     '{prefix}Each item in **{prop_name}** list supports the following keys:')
 SCHEMA_EXAMPLES_HEADER = '\n**Examples**::\n\n'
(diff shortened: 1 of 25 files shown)


tests/unittests/test_handler/test_schema.py
```diff
@@ -507,9 +551,76 @@ def _get_schema_doc_examples():
...
+            types['string'] = (str, bytes)  # pylint: disable=E1137
+            cloudinitValidator = create(
+                meta_schema=Draft4Validator.META_SCHEMA,
+                validators=Draft4Validator.VALIDATORS,
+                version="draft4",
+                default_types=types)
+
+        mymeta = cloudinitValidator.META_SCHEMA
+
+        # this disables bottom-level keys
```
holmanb, line 613: This line is basically the only difference between an otherwise identical function in `schema.py`. I'm leaning towards adding a flag to the function signature in schema.py to set additionalProperties=False, which would reduce this diff by about 70 LOC.

doc/rtd/conf.py
```diff
@@ -67,11 +67,12 @@
...
 
-def generate_docstring_from_schema(app, what, name, obj, options, lines):
+def generate_docstring_from_schema(*args):
     """Override module docs from schema when present."""
-    if what == 'module' and hasattr(obj, "schema"):
+    (_, what, _, obj, _, lines) = args
+    if what == 'module' and hasattr(obj, "__doc__"):
         del lines[:]
-        lines.extend(get_schema_doc(obj.schema).split('\n'))
+        lines.extend(obj.__doc__)
```
blackboxsw, line 75: Minimally I think we need a split("\n") here because you are appending a multi-line string to lines (list). 

The current branch doens't show any module names in the bulleted list: `tox -e doc; xdg-open doc/rtd_html/topics/modules.html`

blackboxsw:
Thanks for this @holmanb. I'm just starting the review amd will have some significant feedback tomorrow.  In your suggested commit message could you capture the intent behind the strict_metaschema being a testing time validation of config module definitions rather than runtime validation.

cloudinit/config/schema.py
```diff
@@ -343,19 +364,16 @@ def _parse_description(description, prefix):
...
     properties = []
     for prop_key, prop_config in schema.get('properties', {}).items():
-        # Define prop_name and dscription for SCHEMA_PROPERTY_TMPL
-        description = prop_config.get('description', '')
-
+        # Define prop_name and description for SCHEMA_PROPERTY_TMPL
         properties.append(SCHEMA_PROPERTY_TMPL.format(
             prefix=prefix,
             prop_name=prop_key,
-            type=_get_property_type(prop_config),
```
holmanb, line 357: avoid builtin function name, replace with something more specific

cloudinit/config/schema.py
```diff
@@ -301,8 +320,10 @@ def _schemapath_for_cloudconfig(config, original_content):
     return schema_line_numbers
 
 
-def _get_property_type(property_dict):
-    """Return a string representing a property type from a given jsonschema."""
+def _get_property_type(property_dict: dict) -> Union[str, bytes]:
```
holmanb, line 323: If SCHEMA_UNDEFINED can be of type `str` rather than type `bytes`, this annotation would be simpler and the changes below would not be necessary. On line 307 if `property_dict.get('enum')` returns a falsy value, then SCHEMA_UNDEFINED is returned from this function unmodified as type `bytes`. Alternatively casting `property_type` to `str` would also accomplish basically the same thing.

tests/unittests/test_cli.py
```diff
@@ -218,18 +219,27 @@ def test_wb_devel_schema_subcommand_parser(self):
...
-             'rocky, sles, ubuntu, virtuozzo'),
-            '**Config schema**:\n    **resize_rootfs:** (true/false/noblock)',
-            '**Examples**::\n\n    runcmd:\n        - [ ls, -l, / ]\n'
-        ]
+        with contextlib.redirect_stdout(stdout):
+            self._call_main(['cloud-init', 'devel', 'schema', '--docs', 'all'])
+            expected_doc_sections = [
+                '**Supported distros:** all',
+                ('**Supported distros:** almalinux, alpine, centos, '
+                    'cloudlinux, debian, eurolinux, fedora, openEuler, '
```
blackboxsw, line 237: This line may to cause problems openEuler vs. openeuler with #1117 fix I think.

tests/unittests/test_cli.py
```diff
@@ -218,18 +219,27 @@ def test_wb_devel_schema_subcommand_parser(self):
             self.stderr.getvalue())
 
     def test_wb_devel_schema_subcommand_doc_content(self):
-        """Validate that doc content is sane from known examples."""
+        """Validate that doc content is sane from known examples.
```
blackboxsw, line 222: Since we are touching this line, let's update inclusive wording where we can.
```suggestion
        """Validate that doc content from known examples is well-structured.
```

or coherent or correct etc.

cloudinit/config/schema.py
```diff
@@ -390,45 +408,88 @@ def _get_schema_examples(schema, prefix=''):
...
+            if mod_locs:
+                mod = importer.import_module(mod_locs[0])
+                docs += mod.__doc__ or ""
+    return docs
 
 
-def get_schema():
+def get_schema() -> dict:
     """Return jsonschema coalesced from all cc_* cloud-config module."""
-    global FULL_SCHEMA
```
holmanb, line 413: This global was previously a reimplementation of features provided by standard library functionality (`lru_cache`). The new implementation uses `load_doc()`, which references the docstrings that are already computed on module load. This means that, like the previous implementation, docstrings are only generated once.

tests/unittests/test_handler/test_schema.py
```diff
@@ -507,9 +533,23 @@ def _get_schema_doc_examples():
...
     schema = get_schema()
 
-    @pytest.mark.parametrize("example_path", _get_schema_doc_examples())
+    @pytest.mark.parametrize("example_path", _get_meta_doc_examples())
     @skipUnlessJsonSchema()
     def test_schema_doc_examples(self, example_path):
         validate_cloudconfig_file(str(example_path), self.schema)
 
+
+class TestStrictMetaschema:
```
holmanb, line 542: This test provides test-time validation of schemas.

tests/unittests/test_cli.py
```diff
@@ -218,18 +219,27 @@ def test_wb_devel_schema_subcommand_parser(self):
...
+
+        Ensure that schema doc is returned
+        """
+
+        # Note: patchStdoutAndStderr() is convenient for reducing boilerplate,
```
blackboxsw, line 227: Ultimately, we'd like to move away from subclassing test_helpers.FilesystemMockingTestCase which is a subclass of unittest.TestCase derivatives where possible and prefer to use pytest framework for discovering any tests prefaced with "Test". This gives the ability to use capsys fixture. I'm not certain this particular case is a candidate for shifting aware from unittest.TestCase for the other unit tests in this test_cli module. But I can see that the particular test case is an easy shift away from the FileSystemMocking test case. Your call on what you'd prefer to do on this PR. It might be more work than it's worth for this iteration to restructure the tests in this module.

```diff
diff --git a/tests/unittests/test_cli.py b/tests/unittests/test_cli.py
index d6909e7a..64e7b8cb 100644
--- a/tests/unittests/test_cli.py
+++ b/tests/unittests/test_cli.py
@@ -282,4 +282,31 @@ class TestCLI(test_helpers.FilesystemMockingTestCase):
         self.assertFalse(parseargs.debug)
         self.assertFalse(parseargs.force)
 
+
+class TestUsePyTest:
+    def _call_main(self, sysv_args=None):
+        if not sysv_args:
+            sysv_args = ['cloud-init']
+        try:
+            return cli.main(sysv_args=sysv_args)
+        except SystemExit as e:
+            return e.code
+
+    def test_wb_devel_schema_subcommand_doc_content(self, capsys):
+        """Validate that doc content is sane from known examples.
+        Ensure that schema doc is returned
+        """
+        self._call_main(['cloud-init', 'devel', 'schema', '--docs', 'all'])
+        expected_doc_sections = [
+                '**Supported distros:** all',
+                ('**Supported distros:** almalinux, alpine, centos, '
+                    'cloudlinux, debian, eurolinux, fedora, openEuler, '
+                    'opensuse, photon, rhel, rocky, sles, ubuntu, virtuozzo'),
+                '**Config schema**:\n    **resize_rootfs:** '
+                '(true/false/noblock)',
+                '**Examples**::\n\n    runcmd:\n        - [ ls, -l, / ]\n'
+        ]
+        stdout, stderr = capsys.readouterr()
+        for expected in expected_doc_sections:
+            assert expected in stdout
 # : ts=4 expandtab
```

tests/unittests/test_handler/test_schema.py
```diff
@@ -507,9 +533,23 @@ def _get_schema_doc_examples():
...
+    '''Validate that schemas follow a stricter metaschema definition than
+    the default. This disallows arbitrary key/value pairs.
+    '''
+    def test_modules(self):
```
blackboxsw, line 546: If we don't have one, let's also add a unit test which asserts no invalid module-level schema raises a schema validation error when strict_metaschema=True. maybe with a pytest.mark.parametrize True,False on strict_metaschema to see validation error vs hiding it.

blackboxsw (commented):
@holmanb thanks for putting this together. A couple of minor requests inline and a general question for you about moving toward pytest to avoid all the stdout/error context manager work.

cloudinit/config/schema.py
```diff
@@ -96,21 +101,35 @@ def validate_cloudconfig_schema(config, schema, strict=False):
...
+
+        # this disables bottom-level keys
+        mymeta['additionalProperties'] = False
+
+        # encoding the base level jsonschema definitions
+        # necessary since (since additionalProperties=False)
+        cloudinitValidator.check_schema(schema)
+
+        cloudinitValidator(schema, format_checker=FormatChecker())
+        return
```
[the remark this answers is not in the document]
holmanb: > @holmanb What do you feel is the use-case we are solving where we want the early exit and ignore of the schema validation below if strict_metaschema=True?

My bad, the intention is not clear here.

Originally this was a separate function in the unittest, but I shaved about 70LOC from the diff by making it a single `if` statement in `schema.py`. I'm realizing that perhaps refactoring out a separate function `get_schema_validator()` for use in `validate_cloudconfig_schema()` and the unittest would probably have been better than muddying the purpose of `validate_cloudconfig_schema()` by making it dual-purpose.

Basically the logic is currently:
```
if strict_metaschema:
    <validate the schema using the metaschema>
    return
<validate the config using the schema>
```

maybe clearer would have been:
```
if strict_metaschema:
    <validate schema using metaschema>
else:
    <validate config using schema>
```
The unittest passes an empty config when calling `validate_cloudconfig_schema()`, since the purpose of that tests is to validate the schemas, not to validate a config with the schema.

> And is it worth a unittest to show that error conditions in schema don't get raised with strict_metaschema=True. (such a unit test docstring would also be a place to document why we don't want to raise errors at runtime).

This sounds like a valuable addition. I will add that.

TheRealFalcon (commented):
A few inline questions/comments, but overall looking good.

tests/unittests/test_cli.py
holmanb: I haven't used capsys directly before, but it looks like `pytest --capture=tee-sys` with capsys would accomplish most of what I was trying to do. This test in particular produces a large amount of output, so the contextmanager allowed me to programatically filter values before printing captured output for debugging, which I don't see as an option with capsys (however maybe I just missed that in the docs). For most use cases duplicating output to stdout/stderr would be more than enough.

I am happy to separate this case out using the suggestion above, but given the size of this PR already, I think I'd prefer to make a `s/FileSystemMocking/capsys/g` conversion of this file into a separate PR.

doc/rtd/conf.py
```diff
@@ -67,11 +67,12 @@
 # of the sidebar.
 html_logo = 'static/logo.png'
 
-def generate_docstring_from_schema(app, what, name, obj, options, lines):
+def generate_docstring_from_schema(*args):
     """Override module docs from schema when present."""
-    if what == 'module' and hasattr(obj, "schema"):
+    (_, what, _, obj, _, lines) = args
```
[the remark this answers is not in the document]
holmanb: My lsp server (pyright I think) didn't recognize the `def generate_docstring_from_schema(_app, what, _name, obj, _options, lines):` syntax, but it didn't mind what I did above. A PR/bug with my lsp server would be more appropriate than working around it in our code, sorry about that.

cloudinit/config/schema.py
```diff
@@ -96,21 +101,35 @@ def validate_cloudconfig_schema(config, schema, strict=False):
...
             validators=Draft4Validator.VALIDATORS,
             version="draft4",
             default_types=types)
+
+    if strict_metaschema:
+        mymeta = cloudinitValidator.META_SCHEMA
+
+        # this disables bottom-level keys
+        mymeta['additionalProperties'] = False
+
```
blackboxsw, line 126: Is this code necessary or referenced anywhere anymore?

cloudinit/config/schema.py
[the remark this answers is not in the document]
blackboxsw: @holmanb I'm defintiely all for early exits specifically for avoiding nested scopes too.  It feel clearer in my opinion as we can then avoid additional line-wraps due to overflow due to all the extra indents later in functions/methods. All good there.  I don't really mind the dual-purpose of the validate_cloudconfig_schema(), that seems to make sense to me as we are generally validating different scopes of the "schema".

My comment really was that don't see why the two options need to be mutually-exclusive. If we add a runtime call with `strict_metaschema=True`, I see a use-case where cloud-init may apply strict schema verification during runtime for both meta schema and module config schema. Maybe cloud-init would warn on invalid meta schema defintions (because that only affects docs) and error on invalid "config" schema.  Also our existing unit tests in this branch pass either way with the empty config.

Given that is is a "future" discussion, and we don't have any runtime call-sites passing `strict_metadata=True`, it's definitely not something that needs resolution in this PR.

doc/rtd/conf.py
```diff
@@ -67,11 +67,12 @@
 # of the sidebar.
 html_logo = 'static/logo.png'
 
-def generate_docstring_from_schema(app, what, name, obj, options, lines):
+def generate_docstring_from_schema(*args):
```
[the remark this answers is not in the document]
holmanb: This currently implements similar logic as before but without the memoized function call (since the __doc__ is given the output of `get_meta_doc()` already). Since the docstring already has what it needs I don't think this preprocessing is required at all.

tests/unittests/test_cli.py
blackboxsw: +1 this can wait. not needed for this branch.

cloudinit/config/schema.py
blackboxsw: Docstrings are only generated once, but get_schema is still recalculating what full_schema is each time it is called. If this function was called multiple times throughout cloud-init bringup, we'd probably want to put an `@lru_cache` on  this I think, but since we only have one call site for get_schema() `validate_cloudconfig_file` and it's a manual command on the CLI, I suggest we don't worry about that pre-mature caching that was done.

cloudinit/config/schema.py
```diff
@@ -34,7 +34,7 @@
 {property_doc}
 {examples}
 """
-SCHEMA_PROPERTY_TMPL = '{prefix}**{prop_name}:** ({type}) {description}'
+SCHEMA_PROPERTY_TMPL = '{prefix}**{prop_name}:** ({prop_type})'
```
blackboxsw, line 37: We have dropped the description of each property here in your commit 859bcc2f8ab371f2bfe19bacc4cd31430e50ece7. This makes both our online docs and the command-line docs a lot less helpful when people try to figure out how to use a given key.

This results in more sparse/terse documentation that no longer explains the intent and expectations of each property value for both commandline docs and online docs as we can see in the [write_files docs](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#write-files)

```
content: (string) Optional content to write to the provided path. When content is present and encoding is not ‘text/plain’, decode the content prior to writing. Default: ‘’
```

but in your local branch we've truncated descriptions leaving us with the less helpful
```
content: (string)
```

we can also quick test these renderd docs locally on the commandline with
```
PYTHONPATH=. python3 -m cloudinit.cmd devel schema --docs all
```

doc/rtd/conf.py
holmanb: This is fixed in commit  [e1b52e3b6196055](https://github.com/canonical/cloud-init/pull/1101/commits/e1b52e3b619605565311728108334e6da720c42a). I don't think further manipulation of the `__doc__` is required, so these functions can just be removed.

doc/rtd/conf.py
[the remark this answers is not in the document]
holmanb: Addressed in [e1b52e3b61960556](https://github.com/canonical/cloud-init/pull/1101/commits/e1b52e3b619605565311728108334e6da720c42a).

cloudinit/config/schema.py
```diff
@@ -474,15 +533,7 @@ def handle_schema_args(name, args):
...
-        schema_ids = [subschema['id'] for subschema in full_schema['allOf']]
-        schema_ids += ['all']
-        invalid_docs = set(args.docs).difference(set(schema_ids))
-        if invalid_docs:
-            error('Invalid --docs value {0}. Must be one of: {1}'.format(
-                  list(invalid_docs), ', '.join(schema_ids)))
-        for subschema in full_schema['allOf']:
-            if 'all' in args.docs or subschema['id'] in args.docs:
-                print(get_schema_doc(subschema))
+        print(load_doc(args.docs))
```
blackboxsw, line 536: So, with this change you have dropped the ability to specify individual schema modules on the commandline via:
```PYTHONPATH=. python3 -m cloudinit.cmd devel schema --docs cc_write_files``` while this may be an under-utilized or under-documented feature of the command line, it does represent a change that we should probably discuss the merit of. 

Minimally, if we are dropping that ability to specify individual modules for documentation on the command line, we need to make sure our [argument parser is updated to avoid prompting for an `--all` value which is now disregarded](https://github.com/canonical/cloud-init/blob/e63069297ba5d75e47c1ed1149dd6cd3f1ac9a92/cloudinit/config/schema.py#L507).

**Review by blackboxsw — CHANGES_REQUESTED** [review]
I think do need to include property descriptions in docs. We might be able to drop the individual module documentation from the command-line, but we need to fixup the argparse setup for the devel schema subcommand.

File: cloudinit/config/schema.py (hunk 13 above)
Comment (holmanb) [reply cloudinit/config/schema.py:124]: Yes, I'll update that to be more obvious. What is getting updated here is actually cloudinitValidator.META_SCHEMA['additionalProperties'].

File: cloudinit/config/schema.py (hunk 7 above)
Comment (holmanb) [reply cloudinit/config/schema.py:413]: And if we ever do get concerned about the performance of python string concatenation, [this might come in handy](https://gist.github.com/holmanb/84be00eab35477565cb95a1d62a741a9).

File: cloudinit/config/schema.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (holmanb) [reply cloudinit/config/schema.py:145]: @blackboxsw I see what you are getting at. Mutual exclusion is not a requirement. Nothing is stopping us from doing strict metadata validation at runtime. Once this lands and we feel comfortable with how it's working that will be a trivial switch to flip.

For now I've refactored a little for readability and added a test that asserts that a bad schema throws the correct exception. I think the change I just made makes it more obvious _what_ failed when a test fails.

File: cloudinit/config/schema.py (hunk 16 above)
Comment (holmanb) [reply cloudinit/config/schema.py:555]: That command doesn't work for me, but in a container [I see this](https://gist.github.com/holmanb/b69e142a2b6be5729a1f7b7bf4e0434c), which is what I would expect for specifying individual modules to print out. Is there something about running it as a module that causes different behavior perhaps? The logic for selecting which modules to print was moved into `load_doc()` and I think it should still work. The functional change was to concatenate module docstrings rather than call `get_schema_doc()`.

Comment (holmanb) [reply cloudinit/config/schema.py:555]: I just noticed that load_doc() isn't doing the same error checking which causes a behavioral change,. I think that this expected behavior probably warrants a unittest. I'll fix it and add one.

File: tests/unittests/test_cli.py (hunk 5 above)
Comment (holmanb) [reply tests/unittests/test_cli.py:237]: I don't think that issue in the PR is related to this. This test just validates that the supported distros in the `cc_ntp` meta config end up in the final doc. currently that list comes from here:

cloud-init/cloudinit/config/cc_ntp.py
```
distros = ['almalinux', 'alpine', 'centos', 'cloudlinux', 'debian',             
            'eurolinux', 'fedora', 'openEuler', 'opensuse', 'photon',            
            'rhel', 'rocky', 'sles', 'ubuntu', 'virtuozzo']  
```

so `system_info()` is not involved

File: cloudinit/config/schema.py (hunk 16 above)
Comment (holmanb) [reply cloudinit/config/schema.py:560]: @blackboxsw - I just pushed some [tests to verify expected behavior](https://github.com/canonical/cloud-init/pull/1101/commits/662564e19994df3e185d3506a7017bbcd2e30fab).

File: tests/unittests/test_handler/test_schema.py [hunk 17]
```diff
@@ -93,7 +130,7 @@ def test_validateconfig_schema_emits_warning_on_missing_jsonschema(self):
         with mock.patch.dict('sys.modules', **{'jsonschema': ImportError()}):
             validate_cloudconfig_schema({'p1': -1}, schema, strict=True)
         self.assertIn(
-            'Ignoring schema validation. python-jsonschema is not present',
```
holmanb, line 96: While I'm making changes, change this to python package name, rather than distro-specific name.

doc/rtd/conf.py
```diff
@@ -1,6 +1,8 @@
 import os
 import sys
 
+from cloudinit import version
```
holmanb, line 4: looks like we don't currently lint this file - fix import location

cloudinit/config/schema.py
```diff
@@ -390,48 +451,108 @@ def _get_schema_examples(schema, prefix=''):
...
+            "frequency",
+            "distros",
+            "description",
+            "name",
+        }
+    )
+    if keys != expected:
+        raise KeyError(
+            "Missing module metadata key(s) {}".format(expected - keys)
+        )
```
blackboxsw, line 479: This will also error if additional keys in meta exist beyond what is expected with a message like "Missing module metadata key(s) set()". Is it worth differentiating the errors here a bit?
Note formatting/whitespace might be off in the "suggestion" here.
```suggestion
    error_msg = ""
    if keys.difference(expected):
        error_msg = "Additional unexpected keys found in module meta: {}".format(keys.difference(expected) 
    elif expected.difference(keys):
        error_msg = "Missing expected keys in module meta: {}".format(expected.difference(keys))
    if error_msg:
       raise KeyError(error_msg)
```

blackboxsw (approved):
Thanks @holmanb any of these remaining comments are nits and I can +1 once you make any changes you think are relevant. Thank you for the restructure, I agree it is easier to read/maintain here and functionally works will with no degradation of existing behavior that I can see.

Approve pending any comments you decide to address and a passing CI run

cloudinit/config/schema.py
```diff
@@ -390,48 +451,108 @@ def _get_schema_examples(schema, prefix=''):
...
+    @raise KeyError: If metadata lacks an expected key.
     """
-    schema_copy = deepcopy(schema)
-    schema_copy['property_doc'] = _get_property_doc(schema)
-    schema_copy['examples'] = _get_schema_examples(schema)
-    schema_copy['distros'] = ', '.join(schema['distros'])
+
+    # Some modules set this to falsy value, exit gracefully
+    if not meta or not schema:
+        return ""
```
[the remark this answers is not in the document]
holmanb: This was required on a previous iteration for reasons that are no longer valid. I agree, throwing an exception on empty meta/schema is probably the right approach.

cloudinit/config/schema.py
```diff
@@ -390,48 +451,108 @@ def _get_schema_examples(schema, prefix=''):
...
 
-FULL_SCHEMA = None
+def get_modules() -> dict:
+    configs_dir = os.path.dirname(os.path.abspath(__file__))
+    return find_modules(configs_dir)
+
 
+def error(message):
+    print(message, file=sys.stderr)
+    sys.exit(1)
```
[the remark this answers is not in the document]
holmanb: factored out to util.py

blackboxsw (changes requested):
Combined diff of my musings for your thoughts. I think we want to minimally make the changes to simplify the get_json_validator return value to a 2-tuple instead of 3-tuple. But other diffs are included in case you find them reasonable.

```diff
diff --git a/cloudinit/config/schema.py b/cloudinit/config/schema.py
index 61f8cd9d..0c46cbdc 100644
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -100,40 +100,48 @@ def get_jsonschema_validator():
         # assignment-operation pylint warning which appears because this
         # code path isn't written against the latest jsonschema).
         types['string'] = (str, bytes)  # pylint: disable=E1137
+        # disable bottom-level keys in meta-schema
+        strict_metaschema = deepcopymeta_schema=Draft4Validator.META_SCHEMA)
+        strict_metaschema["additionalProperties"] = False
         cloudinitValidator = create(
-            meta_schema=Draft4Validator.META_SCHEMA,
+            meta_schema=strict_metaschema,
             validators=Draft4Validator.VALIDATORS,
             version="draft4",
             default_types=types)
-    return (cloudinitValidator, FormatChecker, None)
+    return (cloudinitValidator, FormatChecker)
 
 
-def validate_cloudconfig_metaschema(schema: dict, throw=True):
+def validate_cloudconfig_metaschema(
+    validator, schema: dict, throw: bool = True
+): 
     """Validate provided schema meets the metaschema definition. Return strict
     Validator and FormatChecker for use in validation
+    @param validator: Draft4Validator instance used to validate the schema
     @param schema: schema to validate
     @param throw: Sometimes the validator and checker are required, even if
-        the schema is invalid. Toggle for whether to raise SchemaErrors
-
-    @returns: Tuple(jsonschema.Validator, jsonschema.FormatChecker, err)
+        the schema is invalid. Toggle for whether to raise
+        SchemaValidationError orlog warnings.
 
     @raises: ImportError when jsonschema is not present
-    @raises: jsonschema.exceptions.SchemaError if the schema is invalid
+    @raises: SchemaValidationError when the schema is invalid
     """
 
     from jsonschema.exceptions import SchemaError
 
-    (cloudinitValidator, FormatChecker, _) = get_jsonschema_validator()
     try:
-
-        # disable bottom-level keys
-        cloudinitValidator.META_SCHEMA["additionalProperties"] = False
-        cloudinitValidator.check_schema(schema)
-    except SchemaError as e:
+        validator.check_schema(schema)
+    except SchemaError as err:
+        # Raise SchemaValidationError so we don't have jsonschema imports
+        # at call sites.
         if throw:
-            raise e
-        return (cloudinitValidator, FormatChecker, e)
-    return (cloudinitValidator, FormatChecker, None)
+            raise SchemaValidationError(
+              errors=(('.'.join([str(p) for p in err.path]), err.message),)
+            )
+        logging.warning(
+            "Meta-schema validation failed, attempting to validate config"
+            " anyway: %s",
+            err
+        )
 
 
 def validate_cloudconfig_schema(
@@ -154,20 +162,14 @@ def validate_cloudconfig_schema(
         against the provided schema.
     """
     try:
-        (cloudinitValidator, FormatChecker, err) = (
-            get_jsonschema_validator()
-            if not strict_metaschema
-            else validate_cloudconfig_metaschema(schema, throw=False)
-        )
+        (cloudinitValidator, FormatChecker) = get_jsonschema_validator()
+        if strict_metaschema:
+            validate_cloudconfig_metaschema(
+                cloudinitValidator, schema, throw=False
+            )
     except ImportError:
         logging.debug("Ignoring schema validation. jsonschema is not present")
         return
-    if err:
-        logging.warning(
-            "Meta-schema validation failed, attempting to validate config"
-            " anyway: %s",
-            err,
-        )
     validator = cloudinitValidator(schema, format_checker=FormatChecker())
     errors = ()
     for error in sorted(validator.iter_errors(config), key=lambda e: e.path):
```

holmanb:
> Combined diff of my musings for your thoughts. I think we want to minimally make the changes to simplify the get_json_validator return value to a 2-tuple instead of 3-tuple. But other diffs are included in case you find them reasonable.

Thanks for the help on this. I wasn't sure about the best approach to the failure paths with this one. Besides one item, I think these are all improvements.

> ```diff
> diff --git a/cloudinit/config/schema.py b/cloudinit/config/schema.py
> index 61f8cd9d..0c46cbdc 100644
> --- a/cloudinit/config/schema.py
> +++ b/cloudinit/config/schema.py
> @@ -100,40 +100,48 @@ def get_jsonschema_validator():
>          # assignment-operation pylint warning which appears because this
>          # code path isn't written against the latest jsonschema).
>          types['string'] = (str, bytes)  # pylint: disable=E1137
> +        # disable bottom-level keys in meta-schema
> +        strict_metaschema = deepcopymeta_schema=Draft4Validator.META_SCHEMA)
> +        strict_metaschema["additionalProperties"] = False
>          cloudinitValidator = create(
> -            meta_schema=Draft4Validator.META_SCHEMA,
> +            meta_schema=strict_metaschema,
>              validators=Draft4Validator.VALIDATORS,
>              version="draft4",
>              default_types=types)
> -    return (cloudinitValidator, FormatChecker, None)
> +    return (cloudinitValidator, FormatChecker)

The problem with this is that get_jsonschema_validator() conditionally produces validator/formatchecker differently based on jsonschema version. The change proposed here only applies to the pre-3.0 jsonschema branch. I think we want both branches to be capable of strict validation. Assuming I don't see anything else in need of change I plan on applying the other proposals you made, in addition to a modified version of this one.

Thanks!

blackboxsw (approved):
Thank you for all the discussion and work here @holmanb +1. LGTM

PR #1108, opened 2021-11-15
lxd: add preference for LXD cloud-init.* config keys over user keys

blackboxsw:
## Proposed Commit Message

```
lxd: add preference for LXD cloud-init.* config keys over user keys

LXD now adds cloud-init scoped configuration keys network-config,
user-data and vendor-data. The existing user.user-data,
user.vendor-data, user.network-config and meta-data will be
deprecated in newer LXD.

cloud-init will prefer LXD config keys cloud-init.* keys above
user.* keys even if both are present.

Expectation is that the configuration user.network-config,
user.meta-data, user.user-data and user.vendor-data* keys should
not be present at the same time as the comparable cloud-init.* keys.

Also drop awareness of user.network_mode as this was deprecated
from LXD 2.1.

```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -353,6 +353,11 @@ def read_metadata(
                     # fallback config on old LXD, with new cloud-init images.
                     if CONFIG_KEY_ALIASES[cfg_key] not in md:
                         md[CONFIG_KEY_ALIASES[cfg_key]] = response.text
+                    else:
+                        LOG.warning(
+                            "Ignoring LXD config %s in favor of %s value.",
+                            cfg_key, cfg_key.replace("user", "cloud-init", 1)
+                        )
             else:
                 LOG.debug(
                     "Skipping %s on [HTTP:%d]:%s",
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -10,6 +10,7 @@ Notes:
  * TODO( Hotplug support using websockets API 1.0/events )
 """
 
+from json.decoder import JSONDecodeError
 import os
 
 import requests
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceLXD.py
```diff
@@ -294,35 +288,58 @@ def read_metadata(
...
             "meta-data": md["meta-data"]
         }
-        for config_route in config_routes:
+
+        config_url = version_url + "config"
+        # Represent all advertized/available config routes under
+        # the dict path {LXD_SOCKET_API_VERSION: {config: {...}}.
+        response = session.get(config_url)
+        LOG.debug("[GET] [HTTP:%d] %s", response.status_code, config_url)
+        if not response.ok:
```
TheRealFalcon, line 317: Nice additional error checking in this section

cloudinit/sources/DataSourceLXD.py
```diff
@@ -331,9 +348,16 @@ def read_metadata(
                 md[LXD_SOCKET_API_VERSION]["config"][cfg_key] = response.text
                 # Promote common CONFIG_KEY_ALIASES to top-level keys.
                 if cfg_key in CONFIG_KEY_ALIASES:
```
TheRealFalcon, line 350: Nit: Do we want an `else` on this `if` to raise if the parsed key doesn't exist in our ALIASES? Probably not necessary, but would help indicate a problem in our handling of the datasource.

cloudinit/sources/DataSourceLXD.py
```diff
@@ -331,9 +348,16 @@ def read_metadata(
...
-                    md[CONFIG_KEY_ALIASES[cfg_key]] = response.text
+                    # Due to sort of config_routes, promote cloud-init.*
+                    # aliases before user.*. This allows user.* keys to act as
+                    # fallback config on old LXD, with new cloud-init images.
+                    if CONFIG_KEY_ALIASES[cfg_key] not in md:
```
TheRealFalcon, line 354: If I pass `cloud-init.user-data` and `user.user-data` simultaneously, then we'll ignore `user.user-data` in favor of `cloud-init.user-data`? If so, I think we should at least raise a warning if both are passed simultaneously.

cloudinit/sources/tests/test_lxd.py
```diff
@@ -182,4 +169,209 @@ def test_expected_viable(
         else:
             assert 0 == m_lstat.call_count
 
+
+class TestReadMetadata:
```
TheRealFalcon, line 173: I really like how this test is structured.

TheRealFalcon (changes requested):
Looks good. Only major comment is the warning.

TheRealFalcon (approved):
LGTM, thanks!

PR #1111, opened 2021-11-16
integration_tests: Speed up CI run time (SC-612)

TheRealFalcon:
## Proposed Commit Message

```
integration_tests: Speed up CI run time

Move more tests into test_combined.py and remove the CI mark from module
tests that aren't updated often or don't represent core functionality.
```

## Additional Context
Removed files have been merged into test_combined.py. All tests will still get run daily on our internal jenkins instance. This is for travis and tests that block merging.


blackboxsw (approved):
All good with consolidating unrelated modules into test_combined in general. If we run into failure scenarios because of interaction between those modules(unlikely), we can separate them back out. Likewise, if we find that the removal of pytest.mark.ci leads to frequent failures on our daily build after branches have landed, we can opt to re-integrate those tests in per-PR CI runs.

PR #1115, opened 2021-11-19
Fix adding a static IPv6 default route failed on RA networks

li-kunkun:
## Proposed Commit Message

```
There may be adding a static IPv6 default route failed on networks where
router advertisements are also present.

The flow of up an interface is this:
1. ip link set dev $interface up
2. set sysctl config
3. config ip addr
4. config default ip route

After setting the link up, the kernel might learning an address and a
default route from RA before the default route config. The default route
will fail to be added. If the RA route is expired, and not be refreshed
for some reasons，the system looses IPv6 network connectivity.

Proposed fix is to use "route replace" instead of "route add". When the
RA learned route is still present it gets replaced, when it is not present
the route gets added.

```

## Checklist:

### DIFF
--- a/cloudinit/net/eni.py
+++ b/cloudinit/net/eni.py
@@ -363,7 +363,7 @@ class Renderer(renderer.Renderer):
                  how-to-set-static-routes-in-ubuntu-server
         """
         content = []
-        up = indent + "post-up route add"
+        up = indent + "post-up route replace"
         down = indent + "pre-down route del"
         or_true = " || true"
         mapping = {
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
@li-kunkun , thanks for the submission.

I'm not sure I understand the need for this change. If you're expecting to use cloud-init configuration, then why not disable RA on this instance? If you're expecting to use RA, then why not ensure periodic RAs are working correctly? This seems to be a problem with your network configuration rather than a problem with cloud-init.

I can also imagine the case where somebody doesn't want cloud-init to overwrite a pre-existing route, and since this is current behavior, changing this behavior would result in a regression.

PR #1121, opened 2021-11-26
cc_ssh.py: Change private keyfile mode to 600

Jianlin-lv:
sshd required that private key files are not accessible by others.
Permissions 0640 for keyfile are too open.

One use case is provision one ubuntu 20.04 VM with sshd version
‘OpenSSH_8.2p1 Ubuntu-4ubuntu0.3’. 
ssh.service process is aborted, because ExecStartPre=‘/usr/sbin/sshd -t’ 
run failed. Can’t connect vm using ssh connection anymore. 
So change private key files to 600 mode.

Signed-off-by: Jianlin Lv <lvlin@mail.ustc.edu.cn>

## Proposed Commit Message

```
I provisioned one Ubuntu 20.04 VM and can't login with ssh.
ssh.service startup failed, because ExecStartPre=‘/usr/sbin/sshd -t’ returns an error.
The root cause is the permissions of private key are too open.
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -252,7 +252,7 @@ def handle(_name, cfg, cloud, log, _args):
                     if gid != -1:
                         # perform same "sanitize permissions" as sshd-keygen
                         os.chown(keyfile, -1, gid)
-                        os.chmod(keyfile, 0o640)
+                        os.chmod(keyfile, 0o600)
                         os.chmod(keyfile + ".pub", 0o644)
                 except subp.ProcessExecutionError as e:
                     err = util.decode_binary(e.stderr).lower()


TheRealFalcon:
Hey @Jianlin-lv , these permissions are set intentionally on RHEL/CentOS/Fedora due to a requirement on those systems. See https://github.com/canonical/cloud-init/pull/1070 where the change was added. These permissions should only get set if there's an "ssh_keys" group, which shouldn't exist on Ubuntu.

How are you deploying your instance? Do you have a manually created "ssh_keys" group?

Jianlin-lv:
> Hey @Jianlin-lv , these permissions are set intentionally on RHEL/CentOS/Fedora due to a requirement on those systems. See #1070 where the change was added. These permissions should only get set if there's an "ssh_keys" group, which shouldn't exist on Ubuntu.
> 
> How are you deploying your instance? Do you have a manually created "ssh_keys" group?

We created the "ssh_keys" group, which is inherited from Centos. 
It seems that deleting the "ssh_keys" group can solve the problem I encountered.

TheRealFalcon:
> We created the "ssh_keys" group, which is inherited from Centos.
It seems that deleting the "ssh_keys" group can solve the problem I encountered.

Thanks for the update. I'm going to close this PR then. If you believe there is still something to be addressed, feel free to re-open or create a new PR.

PR #1123, opened 2021-11-29
jinja: provide and document jinja-safe key aliases in instance-data (SC-622)

blackboxsw:
## Proposed Commit Message
```
Allow #cloud-config and cloud-init query to use underscore-delimited
"jinja-safe" key aliases for any instance-data.json keys
containing jinja operator characters.

This provides a means to use Jinja's dot-notation instead of square brackets
and quoting to reference "unsafe" obtain attribute names.

Support for these aliased keys is available to both #cloud-config user-data and
`cloud-init query`.

For example #cloud-config alias access can look like:
  {{ ds.config.user_network_config }}

  - instead of -
  - 
  {{ ds.config["user.network-config"] }}
```

## Test Steps

```bash
 tox -e integration-tests tests/integration_tests/datasources/test_lxd_discovery.py
```

## Checklist:

### DIFF
--- a/cloudinit/cmd/query.py
+++ b/cloudinit/cmd/query.py
@@ -19,7 +19,10 @@ import os
 import sys
 
 from cloudinit.handlers.jinja_template import (
-    convert_jinja_instance_data, render_jinja_payload)
+    convert_jinja_instance_data,
+    get_jinja_variable_alias,
+    render_jinja_payload
+)
 from cloudinit.cmd.devel import addLogHandlerCLI, read_cfg_paths
 from cloudinit import log
 from cloudinit.sources import (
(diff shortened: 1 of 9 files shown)


cloudinit/handlers/jinja_template.py
```diff
@@ -97,7 +101,9 @@ def render_jinja_payload_from_file(
 def render_jinja_payload(payload, payload_fn, instance_data, debug=False):
     instance_jinja_vars = convert_jinja_instance_data(
         instance_data,
-        decode_paths=instance_data.get('base64-encoded-keys', []))
+        decode_paths=instance_data.get('base64-encoded-keys', []),
+        include_key_aliases=True
```
TheRealFalcon, line 105: Is `include_key_aliases` being used in this function?

cloudinit/cmd/query.py
```diff
@@ -162,17 +165,45 @@ def handle_args(name, args):
...
+            if key_path_part in response:
+                response = response[key_path_part]
+            else:  # We are an underscore_delimited key alias
+                for key in response:
+                    if get_jinja_variable_alias(key) == key_path_part:
+                        response = response[key]
+                        break
+            if walked_key_path:
+                walked_key_path += "."
+            walked_key_path += key_path_part
```
TheRealFalcon, line 200: I think this section is hard to follow in a function that's already fairly long. It might help to break it down some or maybe even just add some more comments, but I don't think it needs to block merging.

tests/unittests/test_builtin_handlers.py
```diff
@@ -281,17 +282,44 @@ def test_jinja_template_handle_renders_jinja_content_missing_keys(self):
...
+                True,
+                {'my/key': 'my/val'},
+                {'my/key': 'my/val', 'my_key': 'my/val'}
+            ),
+        )
+    )
+    def test_convert_instance_data_operators_to_underscores(
+        self, include_key_aliases, data, expected
+    ):
+        """Replace Jinja operatiors keys with underscores in instance-data."""
```
TheRealFalcon, line 319: Nit: `operatiors` -> `operators`

tests/integration_tests/modules/test_jinja_templating.py
```diff
@@ -19,12 +20,15 @@ def test_runcmd_with_variable_substitution(client: IntegrationInstance):
     """Test jinja substitution.
 
     Ensure we can also substitute variables from instance-data-sensitive
-    LP: #1931392
+    LP: #1931392 and that underscore-delimited aliases exist for hyphenated
```
TheRealFalcon, line 23: Nit: I think we should probably still keep the LP at the bottom of the docstring.

TheRealFalcon (changes requested):
Looks good overall! I left mostly minor comments inline with the only major thing being removing the unused parameter.

cloudinit/handlers/jinja_template.py
blackboxsw: ahh no longer in this iteration.

blackboxsw: The `instance_jinja_vars` output relies on the included aliases in the return dict to pass into render_string(). Without the added aliases in the instance_jinja_vars dict we wouldn't get proper alias path substitutions.

TheRealFalcon: Ah right, I misread that.

TheRealFalcon (approved):
I'll leave the existing comments up to you, but I think this is good as-is.

cloudinit/cmd/query.py
blackboxsw: 118 line long function... I've done worse, but could do a lot better. breaking this down into a couple of separate functions for readability.

PR #1125, opened 2021-11-30
Fix missing metadata routes for Vultr

eb3095:
## Proposed Commit Message
Fix missing metadata routes for Vultr

```
Vultr uses 169.254.169.254 for the metadata server. Some distros are having trouble with this on IPv6 only servers because the route is not being assigned to the link-local interface by default as it is in other distros. This change sets that route before attempting to fetch the metadata avoiding the current issue.
```

## Test Steps
Deploy Arch as IPv6 Only on Vultr.

## Checklist:


TheRealFalcon (approved):
Thanks!

PR #1129, opened 2021-12-02
Fix exception when no activator found (SC-540)

TheRealFalcon:
## Proposed Commit Message

```
Fix exception when no activator found

Given that there are additional network management tools that we haven't
yet supported with activators, we should log a warning and continue
without network activation here, especially since this was a no-op for
years.

LP: #1948681
```

## Checklist:


blackboxsw (approved):
It makes sense to log a warning here. Let other distros/network stacks add support as needed.

PR #1132, opened 2021-12-03
find_devs/openbsd: accept ISO on disk

goneri:
When the metadata is an ISO image and is exposed through a disk,
the device is called `/dev/sd?a` internally. For instance `/dev/sd1a`.

It can then be mounted with `mount_cd9660 /dev/sd1a /mnt`.

For comparison, metadata in the FAT32 format are exposed as `/dev/sd?i` because this time there is a partition table.

With this change, cloud-init ttries to mount `/dev/sd?a` in addition to `/dev/sd?i`.

Closes: https://github.com/ContainerCraft/kmi/issues/12

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -1217,8 +1217,9 @@ def find_devs_with_openbsd(criteria=None, oformat='device',
             continue
         if entry == 'fd0:':
             continue
-        part_id = 'a' if entry.startswith('cd') else 'i'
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
@goneri , the change looks good to me, but a few unit tests are failing. Can you fix the failing unit tests?

TheRealFalcon (approved):
Thanks!

PR #1136, opened 2021-12-07
factor out function for getting top level directory of cloudinit

holmanb:
```
Add a test helper to get top level directory

Many tests need to get the location of files & dirs within the cloud-init
project directory.

Tests implement this in various different ways, and often those ways
depend on the current working directory of the pytest invocation.
Create helper functions (and tests) that gets the path of the top
directory or any sub directory under the top directory. This function
does not depend on the environment.
```

## Additional Context
- This is not a very high value refactor, but not being able to run pytest wherever I please has been bugging me.
- It should also makes a common task less slightly less error prone and tedious.
- I'm sure that I missed tests.
- The tests I converted were unittests that fail when running pytest in a subdirectory.

### DIFF
--- a/tests/unittests/helpers.py
+++ b/tests/unittests/helpers.py
@@ -510,7 +510,7 @@ if not hasattr(mock.Mock, 'assert_not_called'):
 def get_top_level_dir() -> Path:
     """Return the path to the top cloudinit project directory
 
-    @return Path('top-cloudinit-dir>')
+    @return Path('<top-cloudinit-dir>')
     """
     return Path(cloudinit.__file__).parent.parent.resolve()
 
--- a/tests/unittests/helpers.py
+++ b/tests/unittests/helpers.py
@@ -507,12 +507,12 @@ if not hasattr(mock.Mock, 'assert_not_called'):
     mock.Mock.assert_not_called = __mock_assert_not_called
 
 
-def get_top_level_dir():
+def get_top_level_dir() -> Path:
     """Return the path to the top cloudinit project directory
 
     @return Path('top-cloudinit-dir>')
     """
-    return Path(cloudinit.__file__).parent.parent
+    return Path(cloudinit.__file__).parent.parent.resolve()
 
 
 def CloudinitDir(sub_path: str) -> str:
--- a/tests/unittests/helpers.py
+++ b/tests/unittests/helpers.py
@@ -464,7 +464,7 @@ def wrap_and_call(prefix, mocks, func, *args, **kwargs):
 
 
 def resourceLocation(subname=None):
-    path = CloudinitDir('tests/data')
+    path = cloud_init_project_dir('tests/data')
     if not subname:
         return path
     return os.path.join(path, subname)
@@ -515,12 +515,12 @@ def get_top_level_dir() -> Path:
     return Path(cloudinit.__file__).parent.parent.resolve()
 
 
-def CloudinitDir(sub_path: str) -> str:
+def cloud_init_project_dir(sub_path: str) -> str:
     """Get a path within the cloudinit project directory
 
     @return str of the combined path
 
-    Example: CloudinitDir("my/path") -> "/path/to/cloud-init/my/path"
+    Example: cloud_init_project_dir("my/path") -> "/path/to/cloud-init/my/path"
     """
     return str(get_top_level_dir() / sub_path)
 
--- a/tests/unittests/helpers.py
+++ b/tests/unittests/helpers.py
@@ -508,7 +508,7 @@ if not hasattr(mock.Mock, 'assert_not_called'):
 
 
 def get_top_level_dir() -> Path:
-    """Return the path to the top cloudinit project directory
+    """Return the absolute path to the top cloudinit project directory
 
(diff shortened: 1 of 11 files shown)


tests/unittests/test_subp.py
```diff
@@ -232,13 +232,17 @@ def test_c_lang_can_take_utf8_args(self):
...
-            'import json, sys',
-            'from cloudinit.subp import subp',
-            'data = sys.stdin.read()',
-            'cmd = json.loads(data)',
-            'subp(cmd, capture=False)',
-            ''])
+        python_prog = '\n'.join(
+            [
+                'import json, sys',
+                'sys.path.append("{}")'.format(get_top_level_dir()),
```
holmanb, line 238: Setting the pythonpath environment variable in subp didn't seem to work. That would have been cleaner, but this works too.

tests/unittests/helpers.py
```diff
@@ -504,4 +506,18 @@ def __mock_assert_not_called(mmock):
...
+
+def get_top_level_dir() -> Path:
+    """Return the path to the top cloudinit project directory
+
+    @return Path('<top-cloudinit-dir>')
+    """
+    return Path(cloudinit.__file__).parent.parent.resolve()
+
+
+def CloudinitDir(sub_path: str) -> str:
```
TheRealFalcon, line 518: Is there a reason we're going with this naming? Usually functions are lowercase and snake case. Additionally, I don't think the naming tells me enough to know if this refers to the top-level project directory or the `cloudinit` directory within it.

I also wonder how much we need this additional helper. To take a random example from this PR, is there really much difference between:
`CloudinitDir('tools/ds-identify')`
vs
`get_top_level_dir() / "tools/ds-identify"`

(with maybe a `str` cast when needed?)

TheRealFalcon (commented):
I like the overall approach...just some inline questions about the function implementation.

Also, I think test_schema.py, line 53 needs this too. It currently looks like
`files = list(Path("../../cloudinit/config/").glob("cc_*.py"))`
which doesn't work for me since I'm not running pytest from that directory. I have it replaced locally with
```python
    files = list(
        (
            Path(__file__).parent.parent.parent.parent / "cloudinit" / "config"
        ).glob("cc_*.py")
    )
```
which is much less nice than using this would be.

tests/unittests/helpers.py
holmanb: > Is there a reason we're going with this naming? Usually functions are lowercase and snake case. Additionally, I don't think the naming tells me enough to know if this refers to the top-level project directory or the cloudinit directory within it.

I was trying to think of something short that got the point across and that's what I landed on. Happy to change it to something better. I actually had it implemented as a class in a previous iteration, hence the style.

`get_cloud_init_project_dir()`?

`cloud_init_parent_dir()`?

I'm open to suggestions.

> I also wonder how much we need this additional helper. To take a random example from this PR, is there really much difference between:
CloudinitDir('tools/ds-identify')
> vs
> get_top_level_dir() / "tools/ds-identify"
>
> (with maybe a str cast when needed?)

Most of the locations that I've replaced so far actually expect to do `str` things (+ concatenation, encode('utf-8'), etc) to the returned object. If we did that we would have to cast many of the returned values, which is why I chose this approach.

[A previous iteration](https://github.com/holmanb/cloud-init/commit/b03b911e52deac1cf225f0a8b720a8afa81a130e#) actually subclassed `Path()` and implemented the magic methods required currently (`__add__`, `__addr__`, as well as `encode()`). But something in Python 3.5 was failing which drove me to just return a `str` from the function. (iirc it was join() in 3.5 standard library `os` functions requiring the objects to actually be `str` objects not just implement the magic methods, or something like that).

TheRealFalcon: Thanks, that explanation makes sense. 

As far as naming, I think "project" or "root" is clearer than "parent". `get_cloud_init_project_dir()` or `cloud_init_root_dir()`? Pick your poison.

tests/unittests/test_subp.py
smoser: you probably want to prepend, not append.  The right 'cloudinit.subp' to import is the one in the top level dir, if there were others in the path for some reason then you'd get the wrong one.

that said, '' is (I think) usually the first token of sys.path, which is why this worked before.  So I don't really think you have to change anything unless the intent is to execute these tests with some other working directory.

tests/unittests/test_helpers.py
```diff
@@ -34,4 +35,42 @@ def test_get_ipath_and_empty_instance_id_returns_none(self):
...
+        Note: Recursively searching for .git/ fails during build tests due to
+        .git not existing. This implementation assumes that ../../../ is the
+        relative path to the cloud-init project directory form this file.
+        """
+        out = Path(__file__).parent.parent.parent.resolve()
+        return out
+
+    def test_top_level_dir(self):
+        """Assert the location of the top project directory is correct"""
+        assert 1 == len(
```
TheRealFalcon, line 55: Nit: `assert self.top_dir == self._get_top_level_dir_alt_implementation()` should work here, right?

TheRealFalcon (approved):
Thanks!

PR #1139, opened 2021-12-07
travis - don't run integration tests if no deb (SC-699)

holmanb:
```
If building the *.deb fails, exit

Currently integration tests will run and fail with a non-obvious
message. This makes it so the last thing in the logs is whatever
caused the build to fail.
```

## Additional Context
Consider the new output for a failed build:
https://app.travis-ci.com/github/canonical/cloud-init/jobs/551267972


TheRealFalcon (approved):

TheRealFalcon (approved):
Thanks!

PR #1140, opened 2021-12-07
Improve error message in cloud-init.log

KsenijaS:
Improve error message in cloud-init.log for images without ntfs support.

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -1752,12 +1752,10 @@ def mount_cb(device, callback, data=None, mtype=None,
                     mountpoint = tmpd
                     break
                 except (IOError, OSError) as exc:
-                    stderr = "Stderr: mount: unknown filesystem type 'ntfs'"
-                    if stderr in str(exc):
-                        LOG.debug("No support for mounting ntfs.")
-                    else:
-                        LOG.debug("Failed mount of '%s' as '%s': %s",
-                                  device, mtype, exc)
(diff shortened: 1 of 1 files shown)


holmanb:
Hi @KsenijaS, thanks for the PR!

I agree that the current debug message may leave some things to be desired.

Currently if running on an image without ntfs support, I think an attempt to mount an ntfs file system here would produce a debug message that looks something like: `"Failed mount of '/dev/sdaN' as 'ntfs': Stderr: mount: unknown filesystem type 'ntfs'"`. Is that correct?

I have a comment on the change you requested, as well as some potential ideas.

Since Linux (and BSD) has many different supported filesystems that may have similar error conditions, I think a more generalized solution would be preferred to one isolated error message for a specific fs/error condition combo.

There are a handful of questions that we don't currently answer with this error message that might be helpful to 
- What command was used to attempt the mount?
- Does the block device exist?
- What is does <ntfs/reiserfs/btrfs/etc> mean? (the message does not state that this is the mount/filesystem type, and several lesser known filesystem types could be very confusing to a large majority of users without adding the context that this is a filesystem type)

I think that these are things that we could add to keep the logic generalized to all filesystems, while still improving the error message.  

Something like this, perhaps?
```         
LOG.debug(
    "Failed to mount device: '%s' with type: '%s' using mount command: '%s' which caused exception: '%s'",
    device, mtype, ' '.join(mountcmd), exc)                                                             
```

TheRealFalcon (approved):
Thanks!

PR #1141, opened 2021-12-09
Disable unit tests during integration test run (SC-672)

holmanb:
```
Remove unittests from integration test job in travis

Currently the integration test job executes unittests as part of package build.
The unittests have their own jobs, so this coverage is redundant and
unnecessary. Save time and remove this test.
```

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,7 +121,7 @@ matrix:
             # Use sudo to get a new shell where we're in the sbuild group
             # Don't run integration tests when build fails
             - |
-                sudo -E su $USER -c 'sbuild --nolog --no-run-lintian --verbose --dist=xenial cloud-init_*.dsc' &&
+                sudo -E su $USER -c 'sbuild --nolog --no-run-lintian --no-run-autopkgtest --verbose --dist=xenial cloud-init_*.dsc' &&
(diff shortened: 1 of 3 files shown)


holmanb:
This shaves about 1 minute from the longest running travis test:

Without unittests: [12m 7s](https://app.travis-ci.com/github/canonical/cloud-init/builds/244421715)
With unittests: [13m 16s](https://app.travis-ci.com/github/canonical/cloud-init/builds/244361267) (a build from a current PR that doesn't change any integration test files)

TheRealFalcon:
Changes look good to me, but I'm still entirely sure all of the use cases of packages/debian and if it's appropriate to disable tests there. @blackboxsw , thoughts?

holmanb:
@TheRealFalcon I undid changes to the debian `rules` file. I don't think that change actually did anything, and based on the build log output it didn't run the unit tests after undoing that change, so I don't think it was needed. Now it's gone to be safe.

For context: this skips the tests by adding `DEB_BUILD_OPTIONS=nocheck` env variable to the .travis.yml which causes prevents `make check` from running in `debian/rules`

TheRealFalcon (approved):
LGTM, thanks!

PR #1144, opened 2021-12-09
Schema processing changes (SC-676)

TheRealFalcon:
## Proposed Commit Message

```
Schema processing changes

* Use proper logging
* Add parsing for patternProperties
* Add label to annotate patternProperties
* Log warning if schema parsing fails during metaschema processing
* Some schema test fixes
```

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -17,6 +17,7 @@ import sys
 import yaml
 
 error = partial(error, sys_exit=True)
+LOG = logging.getLogger(__name__)
 
 _YAML_MAP = {True: 'true', False: 'false', None: 'null'}
 CLOUD_CONFIG_HEADER = b'#cloud-config'
@@ -91,7 +92,9 @@ def get_jsonschema_validator():
     # This allows #cloud-config to provide valid yaml "content: !!binary | ..."
 
     strict_metaschema = deepcopy(Draft4Validator.META_SCHEMA)
-    strict_metaschema['additionalProperties'] = False
+    strict_metaschema["additionalProperties"] = False
+    strict_metaschema["properties"]["label"] = {"type": "string"}
+
     if hasattr(Draft4Validator, 'TYPE_CHECKER'):  # jsonschema 3.0+
         type_checker = Draft4Validator.TYPE_CHECKER.redefine(
             'string', is_schema_byte_string)
@@ -140,7 +143,7 @@ def validate_cloudconfig_metaschema(validator, schema: dict, throw=True):
                     ('.'.join([str(p) for p in err.path]), err.message),
                 )
             ) from err
-        logging.warning(
+        LOG.warning(
             "Meta-schema validation failed, attempting to validate config "
             "anyway: %s", err)
 
@@ -168,7 +171,7 @@ def validate_cloudconfig_schema(
             validate_cloudconfig_metaschema(
                 cloudinitValidator, schema, throw=False)
     except ImportError:
-        logging.debug("Ignoring schema validation. jsonschema is not present")
+        LOG.debug("Ignoring schema validation. jsonschema is not present")
         return
 
     validator = cloudinitValidator(schema, format_checker=FormatChecker())
@@ -180,8 +183,8 @@ def validate_cloudconfig_schema(
         if strict:
             raise SchemaValidationError(errors)
         else:
-            messages = ['{0}: {1}'.format(k, msg) for k, msg in errors]
-            logging.warning('Invalid config:\n%s', '\n'.join(messages))
+            messages = ["{0}: {1}".format(k, msg) for k, msg in errors]
+            LOG.warning("Invalid config:\n%s", "\n".join(messages))
 
 
 def annotated_cloudconfig_file(cloudconfig, original_content, schema_errors):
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
My use of "darker" has resulted in a few formatting-only changes. If it's annoying enough I can change them back.

cloudinit/config/schema.py
```diff
@@ -494,7 +516,11 @@ def get_meta_doc(meta: MetaSchema, schema: dict) -> str:
 
     # cast away type annotation
     meta_copy = dict(deepcopy(meta))
-    meta_copy["property_doc"] = _get_property_doc(schema)
+    try:
+        meta_copy["property_doc"] = _get_property_doc(schema)
+    except AttributeError:
+        LOG.warning("Unable to render property_doc due to invalid schema")
+        meta_copy["property_doc"] = ""
```
TheRealFalcon, line 523: If our schema doesn't include properties or otherwise doesn't adhere to the metaschema, this call would fail with a very obscure traceback rather than getting the test failure from the metaschema test telling us what we did wrong.

tests/unittests/config/test_schema.py
```diff
@@ -626,9 +627,11 @@ def _get_meta_doc_examples():
     examples_dir = Path(cloud_init_project_dir('doc/examples'))
     assert examples_dir.is_dir()
 
-    all_text_files = (f for f in examples_dir.glob('cloud-config*.txt')
-                      if not f.name.startswith('cloud-config-archive'))
-    return all_text_files
+    return (
```
TheRealFalcon, line 630: Without this, pytest will name the parametrized tests "example_path1", "example_path2", etc. With this, we see the actual path of the schema being tested.

cloudinit/config/schema.py
```diff
@@ -410,34 +418,53 @@ def _get_property_doc(schema: dict, prefix="    ") -> str:
...
+        schema.get("patternProperties", {}),
+    ]
+
+    for props in property_keys:
+        for prop_key, prop_config in props.items():
+            # Define prop_name and description for SCHEMA_PROPERTY_TMPL
+            description = prop_config.get("description", "")
+
+            # Define prop_name and description for SCHEMA_PROPERTY_TMPL
+            label = prop_config.get("label", prop_key)
```
holmanb, line 432: Glad to see this change is non-breaking (falls back to prop_key when label isn't defined).

cloudinit/config/schema.py
```diff
@@ -91,7 +92,14 @@ def get_jsonschema_validator():
     # This allows #cloud-config to provide valid yaml "content: !!binary | ..."
 
     strict_metaschema = deepcopy(Draft4Validator.META_SCHEMA)
-    strict_metaschema['additionalProperties'] = False
+    strict_metaschema["additionalProperties"] = False
+
+    # This additional label allows us to specify a different name
+    # than the property key. This is especially useful when using a
+    # "patternProperties" regex.
```
holmanb, line 99: Nit: This change happens in the `get_jsonschema_validator()`, but the purpose for this change is for generating meaningful doc labels. It might be helpful to explain "why it's helpful when using a patternProperties regex"

perhaps:
```
    # "patternProperties" regex, otherwise the property label in the generated docs will be a regular expression.
```

cloudinit/config/schema.py
holmanb: I like the addition of a useful message. Do you think we might want to add a unittest ensuring this isn't in the logs?

holmanb (approved):
This looks good to me.

I have a couple of minor nits and comments, but nothing worth holding up this PR over. It might be worthwhile to add a test to test_schema.py demonstrating how a a label in the schema overrides the property key. I would be fine with it getting merged as-is, however.

cloudinit/config/schema.py
TheRealFalcon: That's actually a little easier said than done. Since `get_meta_doc` runs at import time through various `__doc__ = get_meta_doc(meta, schema)` calls, we'll see this warning before the test has fully initialized and before the caplog fixture has a chance to start capturing logs.

Since this is really only a problem if we've defined a schema that fails against the metaschema, and since we test all of the module schemas against the metaschema, I'm ok leaving this one as is.

cloudinit/config/schema.py
```diff
@@ -410,34 +420,53 @@ def _get_property_doc(schema: dict, prefix="    ") -> str:
...
-                properties.append(SCHEMA_LIST_ITEM_TMPL.format(
-                    prefix=new_prefix, prop_name=prop_key))
-                new_prefix += '    '
-                properties.append(_get_property_doc(items, prefix=new_prefix))
-        if 'properties' in prop_config:
-            properties.append(
-                _get_property_doc(prop_config, prefix=new_prefix))
-    return '\n\n'.join(properties)
+                        SCHEMA_LIST_ITEM_TMPL.format(
+                            prefix=new_prefix, prop_name=prop_key
```
blackboxsw, line 455: This needs label I think right?
```suggestion
                            prefix=new_prefix, prop_name=label
```

blackboxsw (commented):
Good idea with the label. 
Only thing I think is really worth fixing 

non-blocking comments:
1. early skip from for loop if properties or patternProperties are empty
2. If you can find someway in metaschema to define that our objects need either "properties" or "patternProperties" go4it. Otherwise no concerns there as we do raise errors during processing of docs

blackboxsw:
> My use of "darker" has resulted in a few formatting-only changes. If it's annoying enough I can change them back.

I think this is fine here as I don't expect any downstreams to have patched the schema tool. It's worth us announcing the use of darker on the mailing list to align future development and make coding style standard. I think it is totally worth us doing this work to darken cloud-init code, but I want to generally do that in a separate non-functional set of PRs. This way downstreams won't have to sift through white-space related content versus functional changes in multiple PRs.

cloudinit/config/schema.py
```diff
@@ -91,7 +92,16 @@ def get_jsonschema_validator():
...
-    strict_metaschema['additionalProperties'] = False
+    strict_metaschema["additionalProperties"] = False
+
+    # This additional label allows us to specify a different name
+    # than the property key when generating docs.
+    # This is especially useful when using a "patternProperties" regex,
+    # otherwise the property label in the generated docs will be a
+    # regular expression.
+    # http://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties
+    strict_metaschema["properties"]["label"] = {"type": "string"}
```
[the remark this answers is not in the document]
TheRealFalcon: I'm not sure I follow here. What are you wanting to have happen?

cloudinit/config/schema.py
```diff
@@ -410,34 +420,53 @@ def _get_property_doc(schema: dict, prefix="    ") -> str:
...
-                description=_parse_description(description, prefix),
-                prop_type=_get_property_type(prop_config),
+    property_keys = [
+        schema.get("properties", {}),
+        schema.get("patternProperties", {}),
+    ]
+
+    for props in property_keys:
+        for prop_key, prop_config in props.items():
+            # Define prop_name and description for SCHEMA_PROPERTY_TMPL
```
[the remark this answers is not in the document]
TheRealFalcon: I don't think so. Properties definitions look something like:
```
'properties': {
    'thing1': {
        'type': 'object',
        ...
    },
    'thing2': {
        ....
    }
}
```
With this example, first time through, `prop_key` would be `thing1` and `prop_config` would be the dictionary it defines. The only time we wouldn't get a truthy value for prop_conig is if we had a definition of one of the properties like `"thing1": {}` or `"thing1": None`, which is weird and doesn't adhere to the metaschema.

If we want to do validation within the function, we could bail if prop_config isn't an instance of dict/mapping, but what to do then? I don't think we should just return given that something is wrong here. We could throw a ValueError rather than the AttributeError we'll naturally get, but I think I prefer leaving the responsibility of anything invalid to the caller. This is why we have the warning on [line 529](https://github.com/canonical/cloud-init/pull/1144/files#diff-3dab74160d0d2e18837e4c271eb03209be4263cf51d44cb8ca78bb739fd3f68cR529)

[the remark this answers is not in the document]
blackboxsw: Oops misread the looping. I thought prop_key and prop_config were actually going to be values like:
prop_key = "properties"
prop_cfg = {} but you are already going through items() of the properties under the "properties" key here.
+1 disregard.

blackboxsw (approved):
Thanks for the unit test on the last commit. LGTM.

PR #1148, opened 2021-12-10
Fix bddeb script for impish (SC-688)

holmanb:
```
Fix bddeb script for impish
```

## Additional Context
dh_systemd is now included in the default helper, no need to specify it anymore for impish

### DIFF
--- a/packages/bddeb
+++ b/packages/bddeb
@@ -94,6 +94,8 @@ def write_debian_folder(root, templ_data, cloud_util_deps):
     requires.extend(['python3'] + reqs + test_reqs)
     if templ_data['debian_release'] == 'xenial':
         requires.append('python3-pytest-catchlog')
+    elif templ_data['debian_release'] == 'impish':
+        requires.remove('dh-systemd')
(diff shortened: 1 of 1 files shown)


packages/pkg-deps.json
```diff
@@ -2,8 +2,7 @@
    "debian" : {
       "build-requires" : [
          "debhelper",
-         "dh-python",
-         "dh-systemd"
+         "dh-python"
```
smoser, line 5: wont that result in a missing build-dependency elsewhere?
surely 'dh-systemd' is not useless, it must be relevant to other releases, no?

holmanb: Hmm, yes it would. I see a workaround for Xenial in `packages/bddep` (which requires using the `--release` flag). I updated this PR to model after that approach. Adding support for release-specific definitions to `tools/read-dependencies` might be cleaner, especially if requirements continue to diverge. I'd rather not do that right now if the updated proposal is acceptable.

TheRealFalcon (approved):
Thanks!

PR #1149, opened 2021-12-10
find_devs_with_openbsd: ensure we return the last entry

goneri:
`sysctl -n hw.disknames` returns a trailing `\n`. We need to clean
this up. In addition, the criteria matching system is a source of
problem because:

- we don't have a way to look up the label of the partition
- we've got situation where an ISO image can be exposed through a virtio
  block device.

So we just totally ignore the value of `criteria`. We end-up with a
slightly longer loop of mount-retry. But this way we're sure we don't
miss a configuration disk.

Tested on Kubvirt with the help of Brady Pratt @jbpratt.


goneri:
I will iterate on this later to refresh the unit-tests.

TheRealFalcon (approved):
Thanks!

PR #1153, opened 2021-12-13
Include dpkg frontend lock in APT_LOCK_FILES (SC-650)

TheRealFalcon:
## Additional Context
https://github.com/canonical/cloud-init/pull/1034#issuecomment-986971376
https://github.com/canonical/cloud-init/pull/1034#issuecomment-987012207

I also included a link to the first comment in a code comment.

### DIFF
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -43,10 +43,17 @@ NETWORK_FILE_HEADER = """\
 NETWORK_CONF_FN = "/etc/network/interfaces.d/50-cloud-init"
 LOCALE_CONF_FN = "/etc/default/locale"
 
+# The frontend lock needs to be acquired first followed by the order that
+# apt uses. /var/lib/apt/lists is locked independently of that install chain,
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
Thanks for the James. Reasonable to add yet another apt lock file check without additional unittests as we already have coverage for this functionality. 

I was able to continue to validate that desired behavior and retries is working by camping on an `apt upgrade` in a terminal and running `cloud-init modules --mode=final` with userdata which tries to install packages
```bash
2021-12-15 03:21:07,618 - debian.py[DEBUG]: Waiting for apt lock
2021-12-15 03:21:17,631 - debian.py[DEBUG]: apt lock available
2021-12-15 03:21:17,631 - subp.py[DEBUG]: Running command ['eatmydata', 'apt-get', '--option=Dpkg::Options::=--force-confold', '--option=Dpkg::options::=--force-unsafe-io', '--assume-yes', '--quiet', 'update'] with allowed return codes [0] (shell=False, capture=False)
```

PR #1156, opened 2021-12-14
tests/cmd/query: fix test run as root and add coverage for defaults

cjp256:
test_handle_args_error_on_invalid_vaname_paths() would fail when run
as root due to invocation of load_userdata() on the default user/vendor
data locations under the instance link.

- Mock load_userdata() for this test case to avoid loads.
- Update _setup_paths() to configure cloud_dir in temporary location.
- Add new test case to verify that the default locations are loaded
  when unspecified.

### DIFF
--- a/tests/unittests/cmd/test_query.py
+++ b/tests/unittests/cmd/test_query.py
@@ -2,8 +2,9 @@
 
 import errno
 import gzip
-from io import BytesIO
 import json
+import os
+from io import BytesIO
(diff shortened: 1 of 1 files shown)


anhvoms:
This would fix https://bugs.launchpad.net/cloud-init/+bug/1825027 ?

TheRealFalcon (approved):
Thanks!

PR #1159, opened 2021-12-15
sources/azure: do not persist failed_desired_api_version flag

cjp256:
If get_imds_data_with_api_fallback() falls back to the minimum required
API version, it is effectively pinned to the old API version forever.

Remove the failed_desired_api_version property to prevent persistence of
the flag between calls and/or reboots.

The continued presence of this flag in obj.pkl should be harmless.


anhvoms:
LGTM

anhvoms (approved):

TheRealFalcon (approved):
LGTM, though we recently applied black formatting to the codebase, so this PR will need a rebase before we can merge it.

TheRealFalcon (approved):
Thanks!

PR #1160, opened 2021-12-15
Add ec2 IPv6 IMDS Support (SC-336)

holmanb:
```
Add support for dual stack ipv6/ipv4 IMDS to Ec2

- add support for parallel http(s) requests to wait_for_url()
- implementation based loosely on RFC6555: "Happy Eyeballs" [1]
- update ec2 datasource to support dual-stack ipv6/ipv4
- only "nitro" instances have ipv6 IMDS, favor ipv4 initially
- replace httpretty with requests for ec2 tests

[1] https://datatracker.ietf.org/doc/html/rfc6555
```

## Additional Context

In contrast to "happy eyeballs", this implementation supports parallel attempts to communicate with a small delay favoring ipv4. This is datasource-specific, and other datasources that wish favor ipv6 can do so trivially.

### Technical Description of Changes:
1.  use a threadpool executor to wrap synchronous calls in async scheduler (see `url_helper.py:dual_stack()`) that does 0.150s delay, cancels pending responses, etc (basically a stateless generic happy eyeballs implementation for synchronous libraries but at the http layer, not the socket/connection layer)
2. refactor `wait_for_url()` into reusable pieces to support optionally parallel http requests to endpoints
3. add tests that verify behavior by simulating a slow/unresponsive endpoint
4. update Ec2 datasource to support parallel GETs (PUTs are left synchronous)
5. update Ec2 tests to test with ipv6
6. make Ec2 favor ipv4 (for now), when majority of cloud has ipv6 support this can change to favor ipv6

### DIFF
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -56,8 +56,8 @@ class DataSourceEc2(sources.DataSource):
     # They will be checked for 'resolveability' and some of the
     # following may be discarded if they do not resolve
     metadata_urls = [
-        "http://[fd00:ec2::254]",
         "http://169.254.169.254",
+        "http://[fd00:ec2::254]",
         "http://instance-data.:8773",
     ]
 
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -55,7 +55,11 @@ class DataSourceEc2(sources.DataSource):
     # Default metadata urls that will be used if none are provided
     # They will be checked for 'resolveability' and some of the
     # following may be discarded if they do not resolve
-    metadata_urls = ["http://169.254.169.254", "http://instance-data.:8773"]
+    metadata_urls = [
+        "http://[fd00:ec2::254]",
+        "http://169.254.169.254",
+        "http://instance-data.:8773"
+    ]
 
     # The minimum supported metadata_version from the ec2 metadata apis
     min_metadata_version = "2009-04-04"
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -257,6 +257,7 @@ class DataSourceEc2(sources.DataSource):
                 exception_cb=self._imds_exception_cb,
                 request_method=request_method,
                 headers_redact=AWS_TOKEN_REDACT,
+                connect_synchronously=False,
             )
         except uhelp.UrlError:
             # We use the raised exception to interupt the retry loop.
@@ -322,7 +323,6 @@ class DataSourceEc2(sources.DataSource):
                 headers_redact=AWS_TOKEN_REDACT,
                 headers_cb=self._get_headers,
                 request_method=request_method,
-                connect_synchronously=False,
             )
 
             if url:
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -58,7 +58,7 @@ class DataSourceEc2(sources.DataSource):
     metadata_urls = [
         "http://[fd00:ec2::254]",
         "http://169.254.169.254",
-        "http://instance-data.:8773"
+        "http://instance-data.:8773",
     ]
 
     # The minimum supported metadata_version from the ec2 metadata apis
@@ -257,7 +257,8 @@ class DataSourceEc2(sources.DataSource):
                 exception_cb=self._imds_exception_cb,
                 request_method=request_method,
                 headers_redact=AWS_TOKEN_REDACT,
-                connect_synchronously=False)
+                connect_synchronously=False,
+            )
         except uhelp.UrlError:
             # We use the raised exception to interupt the retry loop.
             # Nothing else to do here.
@@ -315,10 +316,15 @@ class DataSourceEc2(sources.DataSource):
 
             start_time = time.time()
             url, _ = uhelp.wait_for_url(
-                urls=urls, max_wait=url_params.max_wait_seconds,
-                timeout=url_params.timeout_seconds, status_cb=LOG.warning,
-                headers_redact=AWS_TOKEN_REDACT, headers_cb=self._get_headers,
-                request_method=request_method, connect_synchronously=False)
+                urls=urls,
+                max_wait=url_params.max_wait_seconds,
+                timeout=url_params.timeout_seconds,
+                status_cb=LOG.warning,
+                headers_redact=AWS_TOKEN_REDACT,
+                headers_cb=self._get_headers,
+                request_method=request_method,
+                connect_synchronously=False,
+            )
 
             if url:
                 metadata_address = url2base[url]
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -257,7 +257,7 @@ class DataSourceEc2(sources.DataSource):
                 exception_cb=self._imds_exception_cb,
                 request_method=request_method,
                 headers_redact=AWS_TOKEN_REDACT,
-            )
+                connect_synchronously=False)
         except uhelp.UrlError:
             # We use the raised exception to interupt the retry loop.
             # Nothing else to do here.
@@ -315,14 +315,10 @@ class DataSourceEc2(sources.DataSource):
 
             start_time = time.time()
             url, _ = uhelp.wait_for_url(
-                urls=urls,
-                max_wait=url_params.max_wait_seconds,
-                timeout=url_params.timeout_seconds,
-                status_cb=LOG.warning,
-                headers_redact=AWS_TOKEN_REDACT,
-                headers_cb=self._get_headers,
-                request_method=request_method,
-            )
+                urls=urls, max_wait=url_params.max_wait_seconds,
+                timeout=url_params.timeout_seconds, status_cb=LOG.warning,
+                headers_redact=AWS_TOKEN_REDACT, headers_cb=self._get_headers,
+                request_method=request_method, connect_synchronously=False)
 
             if url:
                 metadata_address = url2base[url]
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -257,7 +257,6 @@ class DataSourceEc2(sources.DataSource):
                 exception_cb=self._imds_exception_cb,
                 request_method=request_method,
                 headers_redact=AWS_TOKEN_REDACT,
-                connect_synchronously=False,
             )
         except uhelp.UrlError:
             # We use the raised exception to interupt the retry loop.
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -11,7 +11,6 @@
 import copy
 import json
 import os
-import sys
 import time
 from concurrent.futures import ThreadPoolExecutor, TimeoutError, as_completed
 from email.utils import parsedate
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -28,7 +28,7 @@ from requests import Session, exceptions
 from requests.adapters import HTTPAdapter
 from urllib3._collections import RecentlyUsedContainer
 from urllib3.exceptions import InsecureRequestWarning
-from urllib3.poolmanager import PoolKey, key_fn_by_scheme
+from urllib3.poolmanager import key_fn_by_scheme
 from urllib3.util import Url
 
 from cloudinit import log as logging
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -427,11 +427,11 @@ def dual_stack(
     # in the constructor, otherwise the context manager would be preferred
     finally:
         # python 3.9 allows canceling futures, which may save some cycles
-        if sys.version_info.major >= 3 and sys.version_info.minor > 9:
+        try:
             executor.shutdown(  # pylint: disable=E1123
                 wait=False, cancel_futures=True
             )
-        else:
+        except TypeError:  # no cancel_futures support
             executor.shutdown(wait=False)
     return (returned_address, return_result)
 
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -18,6 +18,7 @@ from errno import ENOENT
 from functools import partial
 from http.client import NOT_FOUND
 from itertools import count
+from collections import namedtuple
 
 import urllib3
 from urllib.parse import urlparse, urlunparse, quote
@@ -35,6 +36,13 @@ from cloudinit import version
 
 LOG = logging.getLogger(__name__)
 
+# Store a url and its (optional) session together
+UrlSession = namedtuple("UrlSession", ["url", "session"])
+
+# Check if requests has ssl support (added in requests >= 0.8.8)
+SSL_ENABLED = False
+CONFIG_ENABLED = False  # This was added in 0.7 (but taken out in >=1.0)
+_REQ_VER = None
 REDACTED = "REDACTED"
 
 
@@ -163,9 +171,12 @@ class UrlError(IOError):
 
 class HTTPConnectionPoolEarlyConnect(urllib3.HTTPConnectionPool):
     """Allow socket connection prior to http request for dual-stack address
-    selection followed by socket reuse.
+    selection.
 
-    A connection pool enables connection reuse to a single host.
+    In HTTPConnectionPool "connections" are reused to a single host, but socket
+    doesn't actually get connected until the first http(s) request when using
+    HTTPConnectionPool. Allow early socket opening for negotiating address
+    selection.
     """
     def _validate_conn(self, conn) -> None:
         """
@@ -212,10 +223,11 @@ class PoolManagerEarlyConnect(urllib3.PoolManager):
     """Enable early connection to multiple "hosts" for dual-stack addresses
     selection.
 
+    PoolManager handles HTTPConnectionPool allocation based on connection.
     Requests and urllib3 expose a high level API wherein the socket connection
     occurs with the first http request. Allow initializing the connection
     separately which enables dual stack selection (rfc 6555, aka "happy
-    eyeballs") prior to the http request
+    eyeballs") prior to the http request.
     """
     proxy: Optional[Url] = None
     proxy_config: Optional[Any] = None
@@ -243,7 +255,7 @@ class PoolManagerEarlyConnect(urllib3.PoolManager):
     def connection_from_url(
         self, url: str, pool_kwargs: Optional[Dict[str, Any]] = None
         ) -> HTTPConnectionPoolEarlyConnect:
-        """Init pool and connect
+        """Override parent class: Init pool and connect
         """
         pool = super().connection_from_url(url, pool_kwargs)
         pool.connect()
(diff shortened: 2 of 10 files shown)


holmanb:
I'm currently tracking down an issue where tests pass with httpretty when ran directly, but when running pytest locally against all of cloud-init, they fail [1].

I'm fairly confident that the behavior those tests are supposed to be testing are not the cause for this issue. Also the tests pass via travis, so I switched this from draft to ready to review.

[1] The failure message is: `No mocking was registered, and real connections are not allowed (httpretty.allow_net_connect = False).`, which is odd, given that these tests _are_ mocked using `httpretty.register_uri()` which works correctly unless ran with the rest of the suite.

cloudinit/url_helper.py
```diff
@@ -442,18 +516,82 @@ def timeup(max_wait, start_time):
...
+
+    def read_url_serial():
+        """iterate over list of urls, request each one and handle responses
+        and thrown exceptions individually per url
+        """
+
+        def url_reader_serial(url):
+            return (url, read_url_cb(url))
+
+        nonlocal timeout
```
TheRealFalcon, line 588: I think the nonlocal here is a smell. If I follow it through, there's nothing dangerous about it, but it'd be easier to reason about if timeout was passed to the function and then passed again to `read_url_cb`.

TheRealFalcon (changes requested):
Thanks @holmanb . This has come together really nicely. I have just a few small comments.

Do we need to change the call on line 250 of DataSourceEc2.py to be asynchronous as well?
I think the `start_time` and `status_cb` assignments in `wait_for_url` might be better placed under all of the nested functions.

I went back and forth about how abstract this is. On one hand, the code is elegant and well structured. On the other hand, it can be hard to follow the flow through the layers of callbacks, partials, and closures. But then again, unwinding it all or adding types might cause it to be even harder to follow. I think at the end of the day, this is just a more functional approach, and it's really just a few functions, so I think it's fine.

Have you tested this on a live ec2 instance yet? I think it'd be worth ensuring we can fetch metadata in both ipv4-only and ipv6-only environments. If not, we should do that before we merge this. Can you post your test steps in the PR description so they can be easily reproduced in the future? 

I think it'd be worth figuring out how to instrument pycloudlib to launch an ipv6 instance so we can write an integration test, but that doesn't need to come with this PR...though it'd be nice before our next SRU.

cloudinit/url_helper.py
```diff
@@ -382,6 +385,73 @@ def _cb(url):
...
+        return func(addr)
+
+    executor = ThreadPoolExecutor(max_workers=len(addresses))
+    try:
+        futures = {
+            executor.submit(
+                _run_func,
+                func=func,
+                addr=addr,
+                delay=(None if i == 0 else stagger_delay),
```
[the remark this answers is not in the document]
holmanb: Good catch! That logic comes from an early iteration when dual_stack() supported only 2 addresses (per RFC 6555). I missed that when I extended this function to support >2 addresses.

cloudinit/url_helper.py
holmanb: Sounds good to me

holmanb:
>This has come together really nicely.

Thank you :)

> Do we need to change the call on line 250 of DataSourceEc2.py to be asynchronous as well?

I debated this myself, and honestly I still don't know the right answer. I need a better understanding of the semantics of AWS tokens. I should have noted this in the comments when I switched to requesting review, apologies.

I do know that if we do make the PUT on line 250 async, the assertion at `test_aws_token_redacted():673` will fail with len == 5, because the PUT happens multiple times. What does this actually mean?

> I think the `start_time` and `status_cb` assignments in `wait_for_url` might be better placed under all of the nested functions.

Agreed
 
> I went back and forth about how abstract this is. On one hand, the code is elegant and well structured. On the other hand, it can be hard to follow the flow through the layers of callbacks, partials, and closures. But then again, unwinding it all or adding types might cause it to be even harder to follow. I think at the end of the day, this is just a more functional approach, and it's really just a few functions, so I think it's fine.

My sentiments exactly. I should note that since this code already had lots of callbacks it didn't seem too out of place when writing it. That said, I can always restructure to something less functional if needed. One concern I have with doing that is that there would likely be more code duplication and resulting potential for subtle differences in behavior between the sync/parallel implementations. 

> Have you tested this on a live ec2 instance yet? I think it'd be worth ensuring we can fetch metadata in both ipv4-only and ipv6-only environments. If not, we should do that before we merge this. Can you post your test steps in the PR description so they can be easily reproduced in the future?

I have not yet. Since I haven't heard too strong of objections on this yet I'll get started on that next.

> I think it'd be worth figuring out how to instrument pycloudlib to launch an ipv6 instance so we can write an integration test, but that doesn't need to come with this PR...though it'd be nice before our next SRU.

I like that idea.

cloudinit/sources/DataSourceEc2.py
```diff
@@ -317,6 +321,7 @@ def wait_for_metadata_service(self):
                 headers_redact=AWS_TOKEN_REDACT,
                 headers_cb=self._get_headers,
                 request_method=request_method,
+                connect_synchronously=False,
```
blackboxsw, line 324: On EC2 official platform (which is the only Ec2 lookalike that we are currently aware of with IPv6 IMDS support) we should be providing  this connect_synchronously=False only to the call `uhelp.wait_for_url` in `_maybe_fetch_api_token` on line 259.  We shouldn't attempt this here as our first interaction to official EC2 IMDS platform is the latest/api/token which is an "allowed" URL for us to hit/retry multiple times without triggering "non-token IMDS interaction warnings/counts" on EC2 instances.

blackboxsw: And given that `_maybe_fetch_api_token` is only intended to support EC2 proper platforms, this feels like the right scope for this changeset anyway. If other EC2-lookaline platforms grow IPv6 IMDS support we can adapt in the future.

blackboxsw (commented):
This is great work. Please add a fairly detailed suggested commit message in the description of the PR as it merits a lot more context about the design/architecture here.
Also a pointer to the  RFC 6555 link in the commit message will be helpful.

cloudinit/url_helper.py
```diff
@@ -382,6 +385,74 @@ def _cb(url):
...
+        returned_address = futures[future]
+        return_result = future.result()
+        return_exception = future.exception()
+        if return_exception:
+            LOG.warning(
+                "Exception %s during request to %s",
+                return_exception,
+                returned_address,
+            )
+            raise return_exception
```
blackboxsw, line 435: Would we want to specifically look at requests.ConnectionError failures which basically tell us no route to host for IPv6 and ignore them as no route means we really aren't ever going to succeed there right?

holmanb: Interesting question. I don't think that "no route to host" should ever happen (because we're using link-local addresses[1]), but maybe I'm wrong. Generally the idea of "ignore connection failures in case the others succeed" is something I can get behind.

[1] To my knowledge Linux doesn't require fib entries for sending directly to a link-local addresses.

A quick experiment:
```
[ec2-user@ip-172-31-6-134 ~]$ ip r
default via 172.31.0.1 dev eth0 
169.254.169.254 dev eth0 
172.31.0.0/20 dev eth0 proto kernel scope link src 172.31.6.134 
[ec2-user@ip-172-31-6-134 ~]$ sudo ip r del 169.254.169.254
[ec2-user@ip-172-31-6-134 ~]$ ip route show table all | grep -e 169 -e ec2
[ec2-user@ip-172-31-6-134 ~]$ wget http://169.254.169.254
--2022-01-25 23:39:54--  http://169.254.169.254/
Connecting to 169.254.169.254:80... connected.
HTTP request sent, awaiting response... 200 OK
```
Even though the link local route was removed from the routing table it was still able to communicate using that address.

blackboxsw (changes requested):
forgot to set request changes for moving the connect_synchronously up into _maybe_fetch_api_token.

cloudinit/url_helper.py
blackboxsw: Ahh maybe I was misinterpreting the log I saw on an IPv4 only instance:
`2022-01-25 22:34:46,195 - url_helper.py[WARNING]: Calling 'None' failed [0/120s]: request error [HTTPConnectionPool(host='fd00:ec2::254', port=80): Max retries exceeded with url: /latest/api/token (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7facb9c98400>: Failed to establish a new connection: [Errno 101] Network is unreachable'))]`

holmanb: Thanks for the context. "Network is unreachable" is often related to routing issues, I see what you were getting at. Regardless of what "Network is unreachable" is caused by,  I think this suggestion is a good one. Specifically this would make connectivity possible in the case that there is a slow first address and an unreachable second address.

cloudinit/url_helper.py
```diff
@@ -382,6 +385,74 @@ def _cb(url):
...
+        elif not return_result:
+            LOG.warning("Empty result for address %s", returned_address)
+
+    # when max_wait expires
+    except TimeoutError:
+        LOG.warning("Timed out waiting for addresses: %s", " ".join(addresses))
```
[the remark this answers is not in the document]
holmanb: This is thrown from `as_completed()` if none of the endpoints responds before the timeout. If one responds, this does not get thrown.

Snippet from the [docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.as_completed):
> The returned iterator raises a concurrent.futures.TimeoutError if `__next__()` is called and the result isn’t available after timeout seconds from the original call to as_completed().

 This is the same timeout passed to urlopen and for Ec2 this is currently set to 50.

cloudinit/url_helper.py
```diff
@@ -382,6 +385,74 @@ def _cb(url):
...
+    # Executor doesn't provide kwargs for setting shutdown behavior
+    # in the constructor, otherwise the context manager would be preferred
+    finally:
+        # python 3.9 allows canceling futures, which may save some cycles
+        if sys.version_info.major >= 3 and sys.version_info.minor > 9:
+            executor.shutdown(  # pylint: disable=E1123
+                wait=False, cancel_futures=True
+            )
+        else:
+            executor.shutdown(wait=False)
```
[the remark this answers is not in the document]
holmanb: Agreed, this is easier to visually parse.

blackboxsw:
> I think it'd be worth figuring out how to instrument pycloudlib to launch an ipv6 instance so we can write an integration test, but that doesn't need to come with this PR...though it'd be nice before our next SRU.


pycloudlib does have capacity to launch dual-stack ipv4/ipv6 vpc (though this may not yet against exclusive IPv6 IMDS) via something like:
```python
vpc = ec2.get_or_create_vpc(name="cloud-nit-dualstack-vpc")
ec2.launch(image_id, vpc=vpc)
```

So I think we can adapt subclass  tests.integrationtests.cloud.Ec2Cloud._perform_launch to call the get_or_create_vpc  first in cases we want dual-stack support for testing

cloudinit/url_helper.py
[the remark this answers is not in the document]
holmanb: test added in `tests/unittests/test_url_helper.py:test_dual_stack_staggered()`

cloudinit/url_helper.py
holmanb: > Interesting question. I don't think that "no route to host" should ever happen (because we're using link-local addresses[1]), but maybe I'm wrong. 

Update: I was wrong.

My former argument here was invalid. The link local address used on ec2 instances is actually the destination address, not source address. Http(s) traffic leaves via the default route on a private address (RFC1918).

Remove default routes and this error could happen:

```
[ec2-user@ip-172-31-6-134 ~]$ wget http://[fd00:ec2::254]/latest/meta-data/
--2022-02-10 22:36:17--  http://[fd00:ec2::254]/latest/meta-data/
Connecting to [fd00:ec2::254]:80... failed: Network is unreachable.
```

I still think that the route I deleted in the example above (`169.254.169.254 dev eth0`) is unneeded, but for different reasons.

This comment doesn't imply a different course of action, but I wanted to set the record straight here.

TheRealFalcon:
The failing test is an annoying bug that was fixed a few years ago:
https://github.com/getsentry/responses/issues/212

Here's a workaround:
```
diff --git a/tests/unittests/sources/test_ec2.py b/tests/unittests/sources/test_ec2.py
index 3848924d..9e373632 100644
--- a/tests/unittests/sources/test_ec2.py
+++ b/tests/unittests/sources/test_ec2.py
@@ -530,6 +530,14 @@ class TestEc2(test_helpers.HttprettyTestCase):
             md={"md": old_metadata},
         )
         self.assertTrue(ds.get_data())
+
+        # Workaround https://github.com/getsentry/responses/issues/212
+        if hasattr(responses.mock, "_urls"):
+            for index, url in enumerate(responses.mock._urls):
+                if url["url"].startswith(
+                    "http://169.254.169.254/2009-04-04/meta-data/"
+                ):
+                    del responses.mock._urls[index]
         # Provide new revision of metadata that contains network data
         register_mock_metaserver(
             "http://169.254.169.254/2009-04-04/meta-data/", DEFAULT_METADATA
```

Since it's only one test, I'm fine just inserting it there. If this becomes a bigger issue we'd probably want to find a more elegant solution.

paride:
@TheRealFalcon nice! Also matches my findings about responses 0.17.0 being the first version without this issue. Maybe worth adding a line to the comment explaining that the workaround can be removed once there will be no need for testing with requests < 0.17.0 (that is, post-Focal EOL)?

holmanb:
> The failing test is an annoying bug that was fixed a few years ago: [getsentry/responses#212](https://github.com/getsentry/responses/issues/212)

Thanks! This works locally so I have included it.

Also, I had to rebase on the tip of main since the latest commit introduced a merge conflict in one of the tests I changed.

holmanb:
I believe this is ready for re-review.

TheRealFalcon:
I made an integration test for this. If you use https://github.com/canonical/pycloudlib/pull/188 and apply this patch, it should work
```diff
diff --git a/tests/integration_tests/clouds.py b/tests/integration_tests/clouds.py
index e5fe56d1..5ea41cbb 100644
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -207,10 +207,14 @@ class Ec2Cloud(IntegrationCloud):
 
     def _perform_launch(self, launch_kwargs, **kwargs):
         """Use a dual-stack VPC for cloud-init integration testing."""
-        launch_kwargs["vpc"] = self.cloud_instance.get_or_create_vpc(
-            name="ec2-cloud-init-integration"
-        )
+        if "vpc" not in launch_kwargs:
+            launch_kwargs["vpc"] = self.cloud_instance.get_or_create_vpc(
+                name="ec2-cloud-init-integration"
+            )
+        if "Ipv6AddressCount" not in launch_kwargs:
+            launch_kwargs["Ipv6AddressCount"] = 1
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
+        pycloudlib_instance.enable_ipv6_instance_metadata()
         return pycloudlib_instance
 
 
diff --git a/tests/integration_tests/datasources/test_ec2.py b/tests/integration_tests/datasources/test_ec2.py
new file mode 100644
index 00000000..8cde4dc9
--- /dev/null
+++ b/tests/integration_tests/datasources/test_ec2.py
@@ -0,0 +1,43 @@
+import re
+
+import pytest
+
+from tests.integration_tests.instances import IntegrationInstance
+
+
+def _test_crawl(client, ip):
+    assert client.execute("cloud-init clean --logs").ok
+    assert client.execute("cloud-init init --local").ok
+    log = client.read_from_file("/var/log/cloud-init.log")
+    assert f"Using metadata source: '{ip}'" in log
+    result = re.findall(
+        r"Crawl of metadata service took (\d+.\d+) seconds", log
+    )
+    if len(result) != 1:
+        pytest.fail(f"Expected 1 metadata crawl time, got {result}")
+    # 20 would still be a crazy long time for metadata service to crawl,
+    # but it's short enough to know we're not waiting for a response
+    assert float(result[0]) < 20
+
+
+@pytest.mark.ec2
+def test_dual_stack(client: IntegrationInstance):
+    # Drop IPv4 responses
+    assert client.execute("iptables -I INPUT -s 169.254.169.254 -j DROP").ok
+    _test_crawl(client, "http://[fd00:ec2::254]")
+
+    # Block IPv4 requests
+    assert client.execute("iptables -I OUTPUT -d 169.254.169.254 -j REJECT").ok
+    _test_crawl(client, "http://[fd00:ec2::254]")
+
+    # Re-enable IPv4
+    assert client.execute("iptables -D OUTPUT -d 169.254.169.254 -j REJECT").ok
+    assert client.execute("iptables -D INPUT -s 169.254.169.254 -j DROP").ok
+
+    # Drop IPv6 responses
+    assert client.execute("ip6tables -I INPUT -s fd00:ec2::254 -j DROP").ok
+    _test_crawl(client, "http://169.254.169.254")
+
+    # Block IPv6 requests
+    assert client.execute("ip6tables -I OUTPUT -d fd00:ec2::254 -j REJECT").ok
+    _test_crawl(client, "http://169.254.169.254")
```

It'll default all of our ec2 tests to being dual-stack with ipv6 metadata enabled, but I don't see any reason that would be a problem.

tests/unittests/sources/test_ec2.py
```diff
@@ -300,9 +301,14 @@ def register_helper(register, base_url, body):
             register(base_url, "not found", status=404)
 
     def myreg(*argc, **kwargs):
-        url = argc[0]
-        method = httpretty.PUT if ec2.API_TOKEN_ROUTE in url else httpretty.GET
-        return httpretty.register_uri(method, *argc, **kwargs)
+        url, body = argc
+        method = responses.PUT if ec2.API_TOKEN_ROUTE in url else responses.GET
+        print(
```
TheRealFalcon, line 306: This can get rather verbose, so I don't think we wanna keep this around.

tests/unittests/sources/test_ec2.py
```diff
@@ -680,10 +732,11 @@ def test_aws_token_redacted(self):
         logs_with_redacted_ttl = [log for log in all_logs if REDACT_TTL in log]
         logs_with_redacted = [log for log in all_logs if REDACT_TOK in log]
         logs_with_token = [log for log in all_logs if "API-TOKEN" in log]
-        self.assertEqual(1, len(logs_with_redacted_ttl))
-        self.assertEqual(83, len(logs_with_redacted))
+        assert len(logs_with_redacted_ttl)
```
TheRealFalcon, line 735: Do you know why this test changed?

cloudinit/url_helper.py
```diff
@@ -394,31 +517,94 @@ def wait_for_url(
...
+        url = None
+        try:
+            url, response = url_reader_cb(urls)
+            url_exc, reason = read_url_handle_response(response, url)
+            if not url_exc:
+                return (url, response)
+        except UrlError as e:
+            reason = "request error [%s]" % e
+            url_exc = e
+        except Exception as e:  # yikes, do we really think this is necessary?
```
TheRealFalcon, line 567: Why the yikes? If everything doesn't get caught by UrlError, we'd want a broader exception to catch anything else.

cloudinit/url_helper.py
```diff
@@ -394,31 +517,94 @@ def wait_for_url(
...
         if max_wait is None:
             return False
         return (max_wait <= 0) or (time.time() - start_time > max_wait)
 
-    loop_n = 0
-    response = None
-    while True:
-        if sleep_time_cb is not None:
-            sleep_time = sleep_time_cb(response, loop_n)
+    def read_url_handle_response(response, url):
```
TheRealFalcon, line 530: Can we rename this to something like `handle_url_response`? On first glance the name looks like it's involved in calling readurl.

TheRealFalcon (changes requested):
A few minor comments inline, but otherwise looks good. Once we get the dual-stack timeout fixed and the integration test passes, we should be good to go.

Additionally, 2 tickets we should create for follow-up:
1. Add `pytest-responses` to `Build-Depends` of all the packaging branches. I also plan on adding `pytest-mock`, so we could do those at the same time.
2. Replace httpretty with responses everywhere. No need to have 2 libraries doing the same thing.

cloudinit/url_helper.py
holmanb: Oops. That's a note-to-self that slipped past my final review, agreed.

tests/unittests/sources/test_ec2.py
holmanb: Reverted

TheRealFalcon (approved):
This looks good to me now. Thanks for all the work you put into this!

PR #1161, opened 2021-12-16
Add .git-blame-ignore-revs

TheRealFalcon:
## Proposed Commit Message

```
Add .git-blame-ignore-revs

So git doesn't use formatting changes for git blame
```

So I don't get blamed for everything :wink:


holmanb:
LGTM

blackboxsw (approved):

PR #1162, opened 2021-12-18
cli: cloud-id report not-run or disabled state as cloud-id

blackboxsw:
*     cli: status avoid reporting disabled before generator runs
    
    Once cloud-init's systemd generator runs it emits either:
       /run/cloud-init/enabled  OR /run/cloud-init/disabled
    
    If we have neither of those files, cloud-init status should
    still report not-run.

*     cli: cloud-id report not-run or disabled state as cloud-id
    
    Given that cloud-id could be run before cloud-init discovery has
    run report when cloud-init is in a 'disabled' state or whether it
    has 'not-run'.
    
    Adopt new exit codes:
     0: success
     1: error
     2: cloud-init is in disabled state
     3: cloud-init generator has not run yet

## Proposed Commit Message

```
    cli: cloud-id report not-run or disabled state as cloud-id

    This fix has two elements:
     - cloud-init status will not correctly report 'not-run' prior to systemd
       generator running. Only report "disabled" when generator has run
       and /run/cloud-init/disabled exists.
     -  Expose not-run and disabled state in cloud-id responses
     -  Add unique error codes from cloud-id for error, disabled and not-run.
    
   The new cloud-id exit codes:
     0: success
     1: error
     2: cloud-init is in disabled state
     3: cloud-init generator has not run yet
```

## Checklist:

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -16,12 +16,13 @@ from cloudinit.util import get_cmdline, load_file, load_json
 
 CLOUDINIT_DISABLED_FILE = "/etc/cloud/cloud-init.disabled"
 
+
 # customer visible status messages
 @enum.unique
 class UXAppStatus(enum.Enum):
     """Enum representing user-visible cloud-init application status."""
 
-    ENABLED_NOT_RUN = "not-run"
+    NOT_RUN = "not-run"
     RUNNING = "running"
     DONE = "done"
     ERROR = "error"
@@ -68,19 +69,17 @@ def handle_status_args(name, args):
     init.read_cfg()
     status, status_detail, time = get_status_details(init.paths)
     if args.wait:
-        while status in (UXAppStatus.ENABLED_NOT_RUN, UXAppStatus.RUNNING):
+        while status in (UXAppStatus.NOT_RUN, UXAppStatus.RUNNING):
             sys.stdout.write(".")
             sys.stdout.flush()
             status, status_detail, time = get_status_details(init.paths)
             sleep(0.25)
         sys.stdout.write("\n")
+    print("status: {0}".format(status.value))
     if args.long:
-        print("status: {0}".format(status.value))
         if time:
             print("time: {0}".format(time))
         print("detail:\n{0}".format(status_detail))
-    else:
-        print("status: {0}".format(status.value))
     return 1 if status == UXAppStatus.ERROR else 0
 
 
@@ -126,7 +125,7 @@ def get_status_details(paths=None):
         init.read_cfg()
         paths = init.paths
 
-    status = UXAppStatus.ENABLED_NOT_RUN
+    status = UXAppStatus.NOT_RUN
     status_detail = ""
     status_v1 = {}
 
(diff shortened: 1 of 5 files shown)


TheRealFalcon:
Should we be concerned about backwards compatibility?

TheRealFalcon (commented):
One inline nit. I think we should add a couple of tests to `test_cloud_id.py` for the new return codes.

Additionally, should we be concerned about backwards compatibility?

cloudinit/cmd/status.py
```diff
@@ -16,11 +17,15 @@
...
-STATUS_ENABLED_NOT_RUN = "not run"
-STATUS_RUNNING = "running"
-STATUS_DONE = "done"
-STATUS_ERROR = "error"
-STATUS_DISABLED = "disabled"
+@enum.unique
+class UXAppStatus(enum.Enum):
+    """Enum representing user-visible cloud-init application status."""
+
+    ENABLED_NOT_RUN = "not-run"
```
holmanb, line 24: Idea: Besides `"not-run"` the enum names are just the lower-case equivalent of the enum name. If they all matched, I would suggest making the return codes the enum value to store the rc alongside the return message (and then access the enum via `enum.name` or `enum.name.lower()` when needed for printing messages).

cloudinit/cmd/status.py
```diff
@@ -61,23 +66,22 @@ def handle_status_args(name, args):
...
-        print("status: {0}".format(status))
+        print("status: {0}".format(status.value))
         if time:
             print("time: {0}".format(time))
         print("detail:\n{0}".format(status_detail))
     else:
-        print("status: {0}".format(status))
-    return 1 if status == STATUS_ERROR else 0
+        print("status: {0}".format(status.value))
+    return 1 if status == UXAppStatus.ERROR else 0
```
holmanb, line 84: Nit: It seems odd to have a conditional in which both if and else start by doing the same thing. Perhaps something like this?
```
    print("status: {0}".format(status.value))
    if args.long:
        if time:
            print("time: {0}".format(time))
        print("detail:\n{0}".format(status_detail))
    return 1 if status == UXAppStatus.ERROR else 0
```

holmanb (commented):
I added a couple of minor comments (mostly nits). This looks good to me so far.

cloudinit/cmd/status.py
blackboxsw: I'm probably not understanding the intent of your comment here. If get_status_details returned the lower enum value to the caller `handle_status_args` wouldn't I then have to [decode those via UXAppStatus.name.lower() here too](https://github.com/canonical/cloud-init/pull/1162/files#diff-da437f8452f6c63679a5c5e6abb6ac5e999c9bf7ca6717a3e026de77da6b6608R71)

holmanb: You would, yes. The point I'm making is that we define an enum data structure that contains strings and later use the value of the enum to derive error codes [here](https://github.com/canonical/cloud-init/pull/1162/files#diff-df0c5bb55245c37f1e8cd155d17f3e29eaaf2f82bb7357445803e44e4a27c305R71).

This suggestion was intended to be "rather than derive error codes programmatically why not to store them together with their intended string representation, since it appears that we have a 1:1 mapping". The specifics I mentioned with `.lower()` is one way to do that given the current `enum` usage. Also this suggestion might be unnecessary for a data structure that isn't widely used - I'm fine with it as is if you prefer it the way it was.

blackboxsw: This is a really good pt @holman. I thought I would adopt this approach as suggested, but forgot that RUNNING vs DONE status types would have the common non-unique exit code of 0. There may be other enum status that grow over time that also might fall into exit-code 1 for general failure. So, I think I won't be able to use enum to represent this structure.

TheRealFalcon (approved):
LGTM, thanks!

PR #1163, opened 2021-12-28
bsd/resolv.d/ avoid duplicated entries

goneri:
Ensure we don't add duplicated nameserver or searchdomains.
This can happen on OpenBSD because of dhcpleased.


TheRealFalcon (approved):
Thanks!

PR #1164, opened 2021-12-28
puriflying!

DirtyApexAlpha:
## Proposed Commit Message

```

```

## Checklist:

 - [✓] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [✓] I have updated or added any unit tests accordingly
 - [✓] I have updated or added any documentation accordingly

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,36 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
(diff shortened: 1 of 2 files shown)


DirtyApexAlpha (commented):
appreciate if you actually know fact from fiction, theoretical from physical sciences, Mathematics from algorithms or lie from truth be told. Lying about stories fake, fictional, based off of  More lies, architecture doesn't actually with stand when you build on sandboxes so what's the foundation of the source have to say about nevertheless crumbs of tardy responses that make fun of people claiming bash coding language are hate crimes gone too far.

TheRealFalcon:
These are not valid changes

PR #1166, opened 2022-01-04
Shell script handlers by freq

chrislalos:
New PR to replace PR #745
```
Merge of upstream/main + fixes to pass integration tests. 

This commit, my first since May '21, merges in all changes on upstream since then, including the master -> main branch name change. I also included the last suggestions made by Mr. Falcon, which I thought were great.

There are some unit tests failures with some unit tests I'd written that used to pass, but no longer do. The failure does not appear functional in nature and happens far deep down in the call hierarchy. I don't feel great checking code with a unittest failure but I believe I will need help in solving the issue. Thanks!
```

## Checklist:


chrislalos:
Incidentally I've done stuff in the past on this PR as 'beantaxi' ... we're the same user.

.gitignore
```diff
@@ -34,3 +34,4 @@ cloud-init_*.upload
 
 # user test settings
 tests/integration_tests/user_settings.py
+misc/
```
TheRealFalcon, line 37: Let's remove this line since it isn't really relevant for other people. If you want to keep things locally from version control, you can always add the file/directory to `.git/info/exclude`

tests/unittests/test_builtin_handlers.py
```diff
@@ -473,4 +474,28 @@ def test_render_jinja_payload_replaces_missing_variables_and_warns(self):
         self.assertIn(expected_log, self.logs.getvalue())
 
 
+class TestShellScriptByFrequencyHandlers(CiTestCase):
```
TheRealFalcon, line 477: The `CiTestCase` base class isn't needed here. You can just remove it along with the `with_logs = True` on the following line.

tests/unittests/test_builtin_handlers.py
```diff
@@ -473,4 +474,28 @@ def test_render_jinja_payload_replaces_missing_variables_and_warns(self):
...
+    def do_test_frequency(self, frequency):
+        from cloudinit.cmd.devel import read_cfg_paths
+        from cloudinit.handlers.shell_script_by_frequency \
+            import (get_script_folder_by_frequency,
+                    pathMap)
+        ci_paths = read_cfg_paths()
+        scripts_dir = ci_paths.get_cpath('scripts')
+        testFolder = os.path.join(scripts_dir, pathMap[frequency])
+        folder = get_script_folder_by_frequency(frequency, scripts_dir)
+        self.assertEqual(testFolder, folder)
```
TheRealFalcon, line 489:
```suggestion
        assert testFolder == folder
```

tests/unittests/test_builtin_handlers.py
```diff
@@ -473,4 +474,28 @@ def test_render_jinja_payload_replaces_missing_variables_and_warns(self):
         self.assertIn(expected_log, self.logs.getvalue())
 
 
+class TestShellScriptByFrequencyHandlers(CiTestCase):
+    with_logs = True
+
+    def do_test_frequency(self, frequency):
+        from cloudinit.cmd.devel import read_cfg_paths
```
TheRealFalcon, line 481: Is there any reason these imports shouldn't be at the top of the file?

tests/integration_tests/shellscript-by-freq.py
```diff
@@ -0,0 +1,66 @@
...
+PER_FREQ_TEMPLATE = """\
+#!/bin/bash
+touch /tmp/test_per_freq_{}
+"""
+
+PER_BOOT_FILE = StringIO(PER_FREQ_TEMPLATE.format('boot'))
+PER_INSTANCE_FILE = StringIO(PER_FREQ_TEMPLATE.format('instance'))
+PER_ONCE_FILE = StringIO(PER_FREQ_TEMPLATE.format('once'))
+
+# args = SimpleNamespace(
```
TheRealFalcon, line 21: Lets remove this comment block

tests/integration_tests/shellscript-by-freq.py
```diff
@@ -0,0 +1,66 @@
...
+# )
+
+FILES=[
+    (PER_BOOT_FILE, 'boot.sh', 'x-shellscript-per-boot'),
+    (PER_INSTANCE_FILE, 'instance.sh', 'x-shellscript-per-instance'),
+    (PER_ONCE_FILE, 'once.sh', 'x-shellscript-per-once'),
+]
+
+USER_DATA, errors = create_mime_message(FILES)
+print(f'errors={errors}')
```
TheRealFalcon, line 38: We can remove all of the prints from this test.

tests/integration_tests/shellscript-by-freq.py
```diff
@@ -0,0 +1,66 @@
...
+print(f'USER_DATA=${USER_DATA}')
+@pytest.mark.user_data(USER_DATA)
+def test_per_freq(client: IntegrationInstance):
+    print('checking /v/l/c/scripts exists ...')
+    rc_ok = client.execute('test -d /var/lib/cloud/scripts').ok
+    assert rc_ok is True
```
TheRealFalcon, line 44: You can combine these lines. I.e., `assert client.execute('test -d /var/lib/cloud/scripts').ok` . You can do that for the rest of the asserts here.

tests/integration_tests/shellscript-by-freq.py
```diff
@@ -0,0 +1,66 @@
...
+    rc_ok = client.execute('test -f /tmp/test_per_freq_instance').ok
+    assert rc_ok is True
+    # Test per-once
+    print('checking /v/l/c/s/per-boot/once.sh exists ...')
+    rc_ok = client.execute('test -f /var/lib/cloud/scripts/per-boot/once.sh').ok
+    assert rc_ok is True
+    print('checking /tmp/c/test_per_freq_once exists ...')
+    rc_ok = client.execute('test -f /tmp/test_per_freq_once').ok
+    assert rc_ok is True
+    pass
```
TheRealFalcon, line 66: You can remove this line

TheRealFalcon (changes requested):
@chrislalos Thanks for bringing this back. Overall the changes are good. I left some comments inline, but they should all be fairly minor.

Some additional things:
* This branch needs a rebase in order to pass CI.
* Please run `tox -e flake8` and `tox -e pylint` and fix any issues.
* Please run `tox -e format` to format according to our (new) standards.
* Can you re-sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your new Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . This is just so we have your new github username on file. The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html#submitting-your-first-pull-request)

tests/unittests/test_builtin_handlers.py
beantaxi: Done

tests/unittests/test_builtin_handlers.py
beantaxi: Done (I may have been following some existing code at the time; not sure)

tests/unittests/test_builtin_handlers.py
beantaxi: Done

tests/integration_tests/shellscript-by-freq.py
beantaxi: Done

tests/integration_tests/shellscript-by-freq.py
beantaxi: Done (& good to know)

tests/integration_tests/shellscript-by-freq.py
beantaxi: Done

tests/integration_tests/shellscript-by-freq.py
beantaxi: Done

.gitignore
beantaxi: Done. I misread your comment initially, and removed the wrong line. Should be fine now.

beantaxi (commented):
I think all the suggestions have now been incorporated.

chrislalos (commented):
I believe all James Falcon's requests have been fulfilled

(I already commented to this effect but with the wrong user account)

.gitignore
```diff
@@ -34,4 +34,4 @@ cloud-init_*.tar.xz
 cloud-init_*.upload
 
 # user test settings
-tests/integration_tests/user_settings.py
+tests/integration_tests/user_settings.py
```
TheRealFalcon, line 37: Let's restore the newline here.

TheRealFalcon (changes requested):
@chrislalos , thanks for the updates. Things are definitely a lot cleaner.

I took a closer look and found this code isn't really working as expected. Your integration test passes because you're looking in the same directory for all of the scripts, rather than the directory each is supposed to be in.

The logic can be a little tricky, because even though we want the scripts to run per-*whatever*, the part handler itself should run per-always (similar to the existing shell script handler). On every boot, we should write the file to the correct directory, and then based what directory it is in, it runs as it should. 

The other problem is that we have one class defining the 3 valid prefixes, so then when we go to match our part against a matching class, the first instance matches all of them.

Here's a patch that you can `git apply` that should fix these issues
```diff
diff --git a/cloudinit/handlers/shell_script_by_frequency.py b/cloudinit/handlers/shell_script_by_frequency.py
index 71d8031d..9d5f368d 100644
--- a/cloudinit/handlers/shell_script_by_frequency.py
+++ b/cloudinit/handlers/shell_script_by_frequency.py
@@ -42,25 +42,44 @@ def write_script_by_frequency(script_path, payload, frequency, scripts_dir):
 class ShellScriptByFreqPartHandler(Handler):
     """Common base class for the frequency-specific script handlers."""
 
-    prefixes = [
-        "text/x-shellscript-per-boot",
-        "text/x-shellscript-per-instance",
-        "text/x-shellscript-per-once",
-    ]
-
-    def __init__(self, freq, paths, **_kwargs):
-        print("ShellScriptByFreqPartHandler: c'tor()")
-        Handler.__init__(self, freq)
+    def __init__(self, script_frequency, paths, **_kwargs):
+        self.script_frequency = script_frequency
+        Handler.__init__(self, PER_ALWAYS)
         self.scripts_dir = paths.get_cpath("scripts")
         if "script_path" in _kwargs:
             self.scripts_dir = paths.get_cpath(_kwargs["script_path"])
 
     def handle_part(self, data, ctype, script_path, payload, frequency):
-        print("ShellScriptByFreqPartHandler.handle_part()")
-        print("script_path=%s", script_path)
         if script_path is not None:
             filename = os.path.basename(script_path)
             filename = util.clean_filename(filename)
             write_script_by_frequency(
-                script_path, payload, self.frequency, self.scripts_dir
+                script_path, payload, self.script_frequency, self.scripts_dir
             )
+
+
+class ShellScriptPerBootPartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-boot"]
+
+    def __init__(self, paths, **_kwargs):
+        ShellScriptByFreqPartHandler.__init__(
+            self, PER_ALWAYS, paths, **_kwargs
+        )
+
+
+class ShellScriptPerInstancePartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-instance"]
+
+    def __init__(self, paths, **_kwargs):
+        ShellScriptByFreqPartHandler.__init__(
+            self, PER_INSTANCE, paths, **_kwargs
+        )
+
+
+class ShellScriptPerOncePartHandler(ShellScriptByFreqPartHandler):
+    prefixes = ["text/x-shellscript-per-once"]
+
+    def __init__(self, paths, **_kwargs):
+        ShellScriptByFreqPartHandler.__init__(
+            self, PER_ONCE, paths, **_kwargs
+        )
diff --git a/cloudinit/stages.py b/cloudinit/stages.py
index 899b4e5e..6288b654 100644
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -22,7 +22,9 @@ from cloudinit.handlers.cloud_config import CloudConfigPartHandler
 from cloudinit.handlers.jinja_template import JinjaTemplatePartHandler
 from cloudinit.handlers.shell_script import ShellScriptPartHandler
 from cloudinit.handlers.shell_script_by_frequency import (
-    ShellScriptByFreqPartHandler,
+    ShellScriptPerBootPartHandler,
+    ShellScriptPerInstancePartHandler,
+    ShellScriptPerOncePartHandler,
 )
 from cloudinit.handlers.upstart_job import UpstartJobPartHandler
 from cloudinit.net import cmdline
@@ -520,21 +522,12 @@ class Init(object):
         # TODO(harlowja) Hmmm, should we dynamically import these??
         cloudconfig_handler = CloudConfigPartHandler(**opts)
         shellscript_handler = ShellScriptPartHandler(**opts)
-        shellscript_per_boot_handler = ShellScriptByFreqPartHandler(
-            PER_ALWAYS, **opts
-        )
-        shellscript_per_instance_handler = ShellScriptByFreqPartHandler(
-            PER_INSTANCE, **opts
-        )
-        shellscript_per_once_handler = ShellScriptByFreqPartHandler(
-            PER_ONCE, **opts
-        )
         def_handlers = [
             cloudconfig_handler,
             shellscript_handler,
-            shellscript_per_boot_handler,
-            shellscript_per_instance_handler,
-            shellscript_per_once_handler,
+            ShellScriptPerBootPartHandler(**opts),
+            ShellScriptPerInstancePartHandler(**opts),
+            ShellScriptPerOncePartHandler(**opts),
             BootHookPartHandler(**opts),
             UpstartJobPartHandler(**opts),
         ]
```
You'll also need to update tests accordingly. Also, can we change all of our 'per-boot' to be 'per-always' to match the existing semantics?

Beyond that, I left a few more comments inline, but nothing else major.

.gitignore
chrislalos: I'm unsure of what you mean by 'restore the newline' here. This portion of this file currently matches `cloudinit/.gitignore` on main, so I could not find a clue there.

chrislalos:
I think I understand your larger comment. If I add 3 handlers to `stages/def_handlers`, and each of those handlers is an instance of a single class, and that class matches 3 prefixes ... then only one of those 3 handlers actually gets invoked, and it'll be the first one that gets added to `def_handlers`.

And as luck would have it, my integration tests passed anyway because they were mistakenly all checking the same folder (though for different script names), and that coincidentally was the folder written to by that lucky first handler. Thanks for sorting that out ... it sounds like I made that very annoying for you.

So you are proposing creating 3 handlers, restricting each handler to a single prefix, and then adding an instance of each to `def_handlers`. And then fixing the careless mistakes in my test. It took me a few reads but that's actually pretty straightforward.

I do have a few questions ...

- When I dropped from 3 handler classes to 1 handler class, etc, it was based on a conversation we have back on Mar 5 21, on the old PR (#745). I had proposed the idea and you thought it sounded good. If it's just an idea that sounded good in the moment, but then turned out not to work, then no worries. I just wanted to make sure we weren't missing anything.

- Sort of similarly to the above, it seems like the current code is using a class variable (prefixes) to customize the properties on objects. Sort of like a separate `Person` class per every common name, with a class var for `name`, instead of one class and an ivar. If my understanding is correct, then hypothetically that seems like a decent thing to fix, and among other things get back down to one class. But I don't have much enthusiasm to look at that now tbh.

- Regarding this comment of yours: "even though we want the scripts to run per-whatever, the part handler itself should run per-always (similar to the existing shell script handler). On every boot, we should write the file to the correct directory, and then based what directory it is in, it runs as it should." ... am I correct that this is current behavior, and I don't actually have to do anything? Or ... actually, on second thought ... it means my handler does in fact need to write a script to the freq-specific folder and it needs to do so on every boot, irrespective of frequency. And then current behavior manages the script invocation once written, at the right time based on subfolder.

I think that's it.

.gitignore
holmanb: Actually it doesn't match:
```
arc~/workspace/cloud-init(:d072fcd1|✔) git diff HEAD:.gitignore main:.gitignore                 
diff --git a/.gitignore b/.gitignore
index 7e38138b..6eae45c9 100644
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,4 @@ cloud-init_*.tar.xz
 cloud-init_*.upload
 
 # user test settings
-tests/integration_tests/user_settings.py
\ No newline at end of file
+tests/integration_tests/user_settings.py
```

You'll need to go to the end of the last line in the file and type "Enter", which will restore the newline in that file.

TheRealFalcon:
> When I dropped from 3 handler classes to 1 handler class, etc, it was based on a conversation we have back on Mar 5 21, on the old PR (first commit for first cloud-init PR #745). I had proposed the idea and you thought it sounded good. If it's just an idea that sounded good in the moment, but then turned out not to work, then no worries. I just wanted to make sure we weren't missing anything.

No, not missing anything in particular.

> Sort of similarly to the above, it seems like the current code is using a class variable (prefixes) to customize the properties on objects. Sort of like a separate Person class per every common name, with a class var for name, instead of one class and an ivar. If my understanding is correct, then hypothetically that seems like a decent thing to fix, and among other things get back down to one class. But I don't have much enthusiasm to look at that now tbh.

Heh, funny you mention it. I was looking at doing that before I posted my comment, so went back to it and doing it that way should work too. This might be preferable to the diff I posted above:
```diff
diff --git a/cloudinit/handlers/shell_script_by_frequency.py b/cloudinit/handlers/shell_script_by_frequency.py
index 71d8031d..c820bf4c 100644
--- a/cloudinit/handlers/shell_script_by_frequency.py
+++ b/cloudinit/handlers/shell_script_by_frequency.py
@@ -42,25 +42,18 @@ def write_script_by_frequency(script_path, payload, frequency, scripts_dir):
 class ShellScriptByFreqPartHandler(Handler):
     """Common base class for the frequency-specific script handlers."""
 
-    prefixes = [
-        "text/x-shellscript-per-boot",
-        "text/x-shellscript-per-instance",
-        "text/x-shellscript-per-once",
-    ]
-
-    def __init__(self, freq, paths, **_kwargs):
-        print("ShellScriptByFreqPartHandler: c'tor()")
-        Handler.__init__(self, freq)
+    def __init__(self, script_frequency, paths, **_kwargs):
+        Handler.__init__(self, PER_ALWAYS)
+        self.prefixes = [f"text/x-shellscript-{pathMap[script_frequency]}"]
+        self.script_frequency = script_frequency
         self.scripts_dir = paths.get_cpath("scripts")
         if "script_path" in _kwargs:
             self.scripts_dir = paths.get_cpath(_kwargs["script_path"])
 
     def handle_part(self, data, ctype, script_path, payload, frequency):
-        print("ShellScriptByFreqPartHandler.handle_part()")
-        print("script_path=%s", script_path)
         if script_path is not None:
             filename = os.path.basename(script_path)
             filename = util.clean_filename(filename)
             write_script_by_frequency(
-                script_path, payload, self.frequency, self.scripts_dir
+                script_path, payload, self.script_frequency, self.scripts_dir
             )
diff --git a/cloudinit/stages.py b/cloudinit/stages.py
index 899b4e5e..3f17294b 100644
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -520,21 +520,12 @@ class Init(object):
         # TODO(harlowja) Hmmm, should we dynamically import these??
         cloudconfig_handler = CloudConfigPartHandler(**opts)
         shellscript_handler = ShellScriptPartHandler(**opts)
-        shellscript_per_boot_handler = ShellScriptByFreqPartHandler(
-            PER_ALWAYS, **opts
-        )
-        shellscript_per_instance_handler = ShellScriptByFreqPartHandler(
-            PER_INSTANCE, **opts
-        )
-        shellscript_per_once_handler = ShellScriptByFreqPartHandler(
-            PER_ONCE, **opts
-        )
         def_handlers = [
             cloudconfig_handler,
             shellscript_handler,
-            shellscript_per_boot_handler,
-            shellscript_per_instance_handler,
-            shellscript_per_once_handler,
+            ShellScriptByFreqPartHandler(PER_ALWAYS, **opts),
+            ShellScriptByFreqPartHandler(PER_INSTANCE, **opts),
+            ShellScriptByFreqPartHandler(PER_ONCE, **opts),
             BootHookPartHandler(**opts),
             UpstartJobPartHandler(**opts),
         ]
```

> Regarding this comment of yours: "even though we want the scripts to run per-whatever, the part handler itself should run per-always (similar to the existing shell script handler). On every boot, we should write the file to the correct directory, and then based what directory it is in, it runs as it should." ... am I correct that this is current behavior, and I don't actually have to do anything? Or ... actually, on second thought ... it means my handler does in fact need to write a script to the freq-specific folder and it needs to do so on every boot, irrespective of frequency. And then current behavior manages the script invocation once written, at the right time based on subfolder.

If you apply either of my patches, this should be correct behavior and you shouldn't have to do anything.

.gitignore
chrislalos: @holmanb I've added the newline to .gitignore.

I actually assumed an ending newline was the requested change. However, I did not see one via browsing in github.com, or via `git checkout main` ie my local tracking branch, or `git checkout upstream/main` in detached against the upstream itself. And when I ran the diff you posted above, I didn't get a diff. I'm curious as to why I was not seeing an ending newline no matter what I tried, but both you and @TheRealFalcon were expecting the newline, and you're seeing it as well.

chrislalos:
@TheRealFalcon Glad to see my idea was not so strange after all, and thanks for the implementation. I actually wasn't able to apply your initial patch per se, simply because I had already deleted print() statements before trying the patch, and I couldn't coerce the patch to match the file in its current form. I did faithfully recreate it by hand, and I could do the same with your above patch as well. Alternately, if you consider it much better git-fu and etiquette to revert back to the file(s) as they were when you made your patches, and then apply the patches against those versions, I can try that as well. I'd need to figure out just how to do that in git, but it seems straightforward.

One thing I'd feel better about, is if I followed your integration testing ideas and tested using a persistent lxd container on a Linux VM (I don't run Linux locally), and then shelled in after the testing and looked at everything for myself. I did have a VM set up for that on a discount cloud provider, but that VMs having unrelated issues and I can't use it at the moment. Hopefully I can get to it soon.

chrislalos:
Incidentally, I've been surprised a few times that my code's passed CI, because there was something in my integration tests that seemed like it wouldn't work. But, I figured if CI passed than I guess I'm wrong.

However, I just checked the travis log for the integration tests step, and I could not find any mention of my tests or many other tests. The tests that did appear in there all seem to have a specific attribute: `@pytest.mark.ci`. Which sounds like marking the test so that it runs as part of CI.

Am I expected to mark my test with `@pytest.mark.ci`? Or should I leave that alone, and run integration tests myself some other way?

chrislalos:
Done ... and I definitely saw the extra newline this time :) It's been removed.

Thanks for the clarity on pytest.mark.ci. In the future I will plan on running integration tests myself if I feel I need to.

TheRealFalcon (commented):
@chrislalos , I think we're there!

Two final things (I promise!):
Please provide a commit message. When I merge this, all of the commits will be squashed into a single commit. Guidelines for commit messages are near the bottom of [this section](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html#do-these-things-for-each-feature-or-bug) (this PR has no accompanying Launchpad bug). You can put the commit message in the PR description.

Also, can you apply this patch to update the documentation? I tried pushing it to your branch but am unable to. Sorry for not thinking of it sooner.
```diff
diff --git a/doc/rtd/topics/format.rst b/doc/rtd/topics/format.rst
index 05580804..31fdcbba 100644
--- a/doc/rtd/topics/format.rst
+++ b/doc/rtd/topics/format.rst
@@ -38,6 +38,9 @@ Supported content-types are listed from the cloud-init subcommand make-mime:
     x-include-once-url
     x-include-url
     x-shellscript
+    x-shellscript-per-boot
+    x-shellscript-per-instance
+    x-shellscript-per-once
 
 
 Helper subcommand to generate mime messages
@@ -47,13 +50,26 @@ The cloud-init subcommand can generate MIME multi-part files: `make-mime`_.
 
 ``make-mime`` subcommand takes pairs of (filename, "text/" mime subtype)
 separated by a colon (e.g. ``config.yaml:cloud-config``) and emits a MIME
-multipart message to stdout.  An example invocation, assuming you have your
-cloud config in ``config.yaml`` and a shell script in ``script.sh`` and want
-to store the multipart message in ``user-data``:
+multipart message to stdout.
+
+Examples
+--------
+Create userdata containing both a cloud-config (``config.yaml``)
+and a shell script (``script.sh``)
+
+.. code-block:: shell-session
+
+    $ cloud-init devel make-mime -a config.yaml:cloud-config -a script.sh:x-shellscript > userdata
+
+Create userdata containing 3 shell scripts:
+
+- ``always.sh`` - Run every boot
+- ``instance.sh`` - Run once per instance
+- ``once.sh`` - Run once
 
 .. code-block:: shell-session
 
-    $ cloud-init devel make-mime -a config.yaml:cloud-config -a script.sh:x-shellscript > user-data
+    $ cloud-init devel make-mime -a always.sh:x-shellscript-per-boot -a instance.sh:x-shellscript-per-instance -a once.sh:x-shellscript-per-once
 
 .. _make-mime: https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/devel/make_mime.py
 
```

Also, I might have accidentally pinged you on another PR (accidentally intending to respond to this one). You can ignore that. :slightly_smiling_face:

TheRealFalcon (approved):
@chrislalos , I'd rather not let this PR get closed again from going stale, so I'm going to merge it as-is and then propose the doc change in a separate PR.

Thanks for all your hard work on this and getting it to the finish line!

beantaxi:
I could have sworn I made the docs changes and added the commit comment! I
definitely had the commit comment ready to go in giithub ... but maybe I
restarted for an OSX upgrade before I submitted.

In any case I'm very pleased this is in. Thank you so much for your
patience and guidance! I learned quite a bit.

What's next ... wait for the next clout-init release? I seem to recall
seeing mid Feb on the release calendar, would this PR be part of that/

Thanks!

C

chrislalos:
I found the docs changes to `doc/rtd/topics/format.rst` ... looks like I applied your PATCH on main and not the PR branch. Shall I turn around and create a new PR for those containing just the one file?

TheRealFalcon:
@chrislalos no worries, I added the patch as a separate PR already.

> What's next ... wait for the next clout-init release? I seem to recall
seeing mid Feb on the release calendar, would this PR be part of that

This PR will go into the next upstream release, yes. We have an upstream release scheduled for tomorrow, Feb 15th. It will still take more time for the release to trickle down into distro packages though. For Ubuntu, it will likely be a couple more weeks for the release to go through the [SRU](https://wiki.ubuntu.com/StableReleaseUpdates) process.

chrislalos:
Been a while :)

I was just thinking about this PR ... I notice it's Merged, but not Closed, and tagged as stale. According to the release linked above (22.1), this PR was included in that release.

Should this simply be closed? Or is there another step or so required from me?

TheRealFalcon:
Nope, nothing left to do. "Closed" really means "Closed without merging". Once a PR is merged, it is done.

If you opened a separate issue/bug for this, you could make sure that it is closed, but otherwise we're good here.

PR #1167, opened 2022-01-06
Remove 3.5 and xenial support (SC-711)

TheRealFalcon:
## Proposed Commit Message

```
Remove 3.5 and xenial support

Includes:
 - Update tox.ini and .travis.yml accordingly
 - Cleanup tox.ini with new tox syntax and cloud-init dependencies
 - Update documentation accordingly
 - Replace/remove xenial references where additional testing isn't required
 - Remove xenial checks in integration tests
 - Replace yield_fixture with fixture in pytest tests

Sections of code commented with lines like "Remove when Xenial is no
longer supported" still exist as they'll require additional testing.
```

## Additional Context
I created a separate ticket for dealing with code that would need additional testing.

## Checklist:

### DIFF
--- a/doc/rtd/topics/testing.rst
+++ b/doc/rtd/topics/testing.rst
@@ -54,50 +54,28 @@ Test Layout
 * pytest tests should use bare ``assert`` statements, to take advantage
   of pytest's `assertion introspection`_
 
-  * For ``==`` and other commutative assertions, the expected value
-    should be placed before the value under test:
-    ``assert expected_value == function_under_test()``
-
-
 ``pytest`` Version Gotchas
 --------------------------
 
-As we still support Ubuntu 16.04 (Xenial Xerus), we can only use pytest
-features that are available in v2.8.7.  This is an inexhaustive list of
+As we still support Ubuntu 18.04 (Bionic Beaver), we can only use pytest
+features that are available in v3.3.2.  This is an inexhaustive list of
 ways in which this may catch you out:
-
-* Support for using ``yield`` in ``pytest.fixture`` functions was only
-  introduced in `pytest 3.0`_.  Such functions must instead use the
-  ``pytest.yield_fixture`` decorator.
 
 * Only the following built-in fixtures are available [#fixture-list]_:
 
   * ``cache``
   * ``capfd``
-  * ``caplog`` (provided by ``python3-pytest-catchlog`` on xenial)
+  * ``capfdbinary``
+  * ``caplog``
   * ``capsys``
+  * ``capsysbinary``
+  * ``doctest_namespace``
   * ``monkeypatch``
   * ``pytestconfig``
   * ``record_xml_property``
   * ``recwarn``
   * ``tmpdir_factory``
   * ``tmpdir``
-
-* On xenial, the objects returned by the ``tmpdir`` fixture cannot be
-  used where paths are required; they are rejected as invalid paths.
-  You must instead use their ``.strpath`` attribute.
-
-  * For example, instead of ``util.write_file(tmpdir.join("some_file"),
-    ...)``, you should write
-    ``util.write_file(tmpdir.join("some_file").strpath, ...)``.
-
-* The `pytest.param`_ function cannot be used. It was introduced in
-  pytest 3.1, which means it is not available on xenial.  The more
-  limited mechanism it replaced was removed in pytest 4.0, so is not
-  available in focal or later.  The only available alternatives are to
-  write mark-requiring test instances as completely separate tests,
-  without utilising parameterisation, or to apply the mark to the
-  entire parameterized test (and therefore every test instance).
 
 Mocking and Assertions
 ----------------------
@@ -168,9 +146,9 @@ Test Argument Ordering
 .. [#fixture-list] This list of fixtures (with markup) can be
    reproduced by running::
 
-     py.test-3 --fixtures -q | grep "^[^ -]" | grep -v '\(no\|capturelog\)' | sort | sed 's/.*/* ``\0``/'
+     python3 -m pytest  --fixtures -q | grep "^[^ -]" | grep -v '\(no\|capturelog\)' | sort | sed 's/.*/* ``\0``/'
 
-   in a xenial lxd container with python3-pytest-catchlog installed.
+   in an ubuntu lxd container with python3-pytest installed.
 
 .. _pytest: https://docs.pytest.org/
 .. _pytest fixtures: https://docs.pytest.org/en/latest/fixture.html
--- a/doc/rtd/topics/testing.rst
+++ b/doc/rtd/topics/testing.rst
@@ -146,7 +146,7 @@ Test Argument Ordering
 .. [#fixture-list] This list of fixtures (with markup) can be
    reproduced by running::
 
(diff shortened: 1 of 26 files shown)


doc/rtd/topics/testing.rst
```diff
@@ -54,51 +54,29 @@ Test Layout
 * pytest tests should use bare ``assert`` statements, to take advantage
   of pytest's `assertion introspection`_
 
-  * For ``==`` and other commutative assertions, the expected value
-    should be placed before the value under test:
-    ``assert expected_value == function_under_test()``
-
-
```
TheRealFalcon, line 61: I snuck this removal in here as I was modifying this file as this rule seems arbitrarily restrictive and one I purposely ignore all the time :slightly_smiling_face:

doc/rtd/topics/testing.rst
```diff
@@ -168,9 +146,9 @@ Test Argument Ordering
 .. [#fixture-list] This list of fixtures (with markup) can be
    reproduced by running::
 
-     py.test-3 --fixtures -q | grep "^[^ -]" | grep -v '\(no\|capturelog\)' | sort | sed 's/.*/* ``\0``/'
+     python3 -m pytest  --fixtures -q | grep "^[^ -]" | grep -v '\(no\|capturelog\)' | sort | sed 's/.*/* ``\0``/'
```
holmanb, line 149: A few comments on this one-liner:

Presumably the "no" is to remove `no tests ran in 0.00s`, but this filters out `no_cover` on more recent pytest versions - future-proof with longer match string.

"capturelog" doesn't filter anything on pytest 6.2.5, I assume this is something old that isn't needed?

It looks like "[session scope]" is being manually removed after running the command. This can be updated with a small updated to the `sed` command.

Assuming others agree with the above comments, this is what I propose:
```
python3 -m pytest  --fixtures -q | grep "^[^ -]" | grep -v 'no tests ran in' | sort | sed -e 's/ \[session scope\]//g;s/.*/* ``\0``/g'
```

TheRealFalcon: Works for me, thanks!

holmanb (approved):
LGTM

Related: We still have a couple of comments saying "todo: remove when 2.7 is no longer supported" in the codebase. When we get around to removing 3.5-specific code, we may want to keep those in mind as well :)

tox.ini
```diff
@@ -1,144 +1,91 @@
...
     -r{toxinidir}/test-requirements.txt
 commands = {envpython} -m pytest \
             --durations 10 \
             {posargs:--cov=cloudinit --cov-branch \
             tests/unittests}
 
-[testenv:py27]
-basepython = python2.7
-deps = -r{toxinidir}/test-requirements.txt
+[testenv:lowest-supported]
```
blackboxsw, line 57: I don't think we can do away with a shared lowest-supported-deps definition just yet. 
Since `lowest-supported` is included in our default envlist it'll be run automatically in developer environments via `tox -p auto`. Without continuing to "pin" `httpretty=0.9.5` in a lowest-supported-dev target we'll fail to pass locally on focal, hirsute and I think impish too.

It'd like to ensure developers can run this build env to avoid getting a suprirse if CI runs it and fails. What do you think?

tox.ini
```diff
@@ -1,144 +1,91 @@
...
+    pyyaml==3.12
+    requests==2.18.4
+    jsonpatch==1.16
+    jsonschema==2.6.0
+    netifaces==0.10.4
+    # test-requirements
+    httpretty==0.8.14
+    pytest==3.3.2
+    pytest-cov==2.5.1
+    attrs==17.4.0  # Needed by pytest and default causes failures
```
blackboxsw, line 79: Inline comments still break on bionic tox. This comment needs to be a separate line otherwise we'll hit something like:
```bash
ERROR: Invalid requirement: 'attrs==17.4.0 #  Needed by pytest and default causes failures'
```

blackboxsw (changes requested):
Hey @TheRealFalcon this looks really good, only comments I have are some of the tox refactors as I think we need to be able to support running these targets on Bionic, Focal and Impish.  I think we still need a dev target that pins a different httpretty that the straight `lowest-supported` testenv. Also we need to make sure `tox -e lowest-supported` works on Bionic directly right?


Here's a diff of what I think works for my inline comments:
```diff
diff --git a/tox.ini b/tox.ini
index d57111331..1a6a441d2 100644
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py3, lowest-supported, flake8, pylint, black, isort
+envlist = py3, lowest-supported-dev, flake8, pylint, black, isort
 recreate = True
 
 [testenv]
@@ -54,7 +54,7 @@ commands = {envpython} -m pytest \
             {posargs:--cov=cloudinit --cov-branch \
             tests/unittests}
 
-[testenv:lowest-supported]
+[lowest-supported-deps]
 # Tox is going to install requirements from pip. This is fine for
 # testing python version compatibility, but when we build cloud-init, we are
 # building against the dependencies in the OS repo, not pip. The OS
@@ -73,11 +73,26 @@ deps =
     jsonschema==2.6.0
     netifaces==0.10.4
     # test-requirements
-    httpretty==0.8.14
     pytest==3.3.2
     pytest-cov==2.5.1
-    attrs==17.4.0  # Needed by pytest and default causes failures
+    # attrs==17.4.0  # Needed by pytest and default causes failures
+    attrs=17.4.0
+
+[testenv:lowest-supported]
 commands = {[testenv:py3]commands}
+deps =
+    {[lowest-supported-deps]deps}
+    # Official lowest-supported deps released in bionic-updates
+    httpretty==0.8.14
+
+[testenv:lowest-supported-dev]
+commands = {[testenv:py3]commands}
+deps =
+    {[lowest-supported-deps]deps}
+    # httpretty in bionic is 0.8.14, not 0.9.5.  The oldest version to work with
+    # Python 3.7+ is 0.9.5, because it is the first to include this commit:
+    # https://github.com/gabrielfalcao/HTTPretty/commit/5776d97da3992b9071db5e21faf175f6e8729060
+    httpretty==0.9.5
 
 [testenv:doc]
 deps =
```

blackboxsw:
Note I discovered a potentially related set of failures with running `tox -e py3 tests/unittests/test_ds_identify.py` on Bionic. Seems we have a test leaking into the underlying filesystem and getting back unexpected data: 
http://paste.ubuntu.com/p/MM2XjX7by5/

TheRealFalcon:
Really good comments @blackboxsw , thanks. I never knew why we had so many xenial definitions in our tox definition, and now I do :smile: .

Updated based on comments.

blackboxsw:
> Note I discovered a potentially related set of failures with running `tox -e py3 tests/unittests/test_ds_identify.py` on Bionic. Seems we have a test leaking into the underlying filesystem and getting back unexpected data: http://paste.ubuntu.com/p/MM2XjX7by5/

Put up the related PR to avoid leaks detecting a valid LXD datasource in test_ds_identify. https://github.com/canonical/cloud-init/pull/1178

blackboxsw (approved):
+1! Bring it home! Bye bye xenial.

PR #1173, opened 2022-01-07
3.10 Test Coverage (SC-719)

holmanb:
```
Add 3.10 support to test matrix

Multiple supported distros (arch/fedora) have changed their default
python version to 3.10. We should include it in the test matrix.
```
Context: https://travis-ci.community/t/add-python-3-10/12220/12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,7 +141,7 @@ matrix:
           env: TOXENV=doc
         # Test all supported Python versions (but at the end, so we schedule
         # longer-running jobs first)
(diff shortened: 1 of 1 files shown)


holmanb:
At some point presumably travis will add a 3.10 target that tracks the latest minor version, but until then "3.10.1" works.

TheRealFalcon (approved):
Oooh, thanks for this. I've tried 3.10 in the past and Travis just noped out on it. Didn't realize 3.10.1 was an option.

PR #1174, opened 2022-01-07
Ensure system_cfg read before ds net config on Oracle (SC-720)

TheRealFalcon:
## Proposed Commit Message

```
Ensure system_cfg read before ds net config on Oracle

In 2c52e6e88b19f5db8d55eb7280ee27703e05d75f, the order of
reading network config was changed for Oracle due to initramfs
needing to take lower precedence than the datasource. However,
this also bumped system_cfg to a lower precedence than ds, which
means that any network configuration specified in /etc/cloud will not
be applied. system_cfg should instead be moved above ds so network
configuration in /etc/cloud takes precedence.

LP: #1956788
```

## Additional Context
https://github.com/canonical/cloud-init/commit/2c52e6e88b19f5db8d55eb7280ee27703e05d75f

## Test Steps
Boot an Oracle instance using the Oracle datasource.
Create /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with contents `network: {config: disabled}`
rm /etc/netplan/50-cloud-init.yaml (or rename it if you need it)
`cloud-init clean --logs && cloud-init init --local`
Without patch, /etc/netplan/50-cloud-init.yaml should get written. With patch, it won't be written and logs should show no network written.

I can also include an Oracle-specific integration test, but wanted to get the PR up sooner rather than later.

## Checklist:

### DIFF
--- a/tests/integration_tests/network/test_net_config_load.py
+++ b/tests/integration_tests/network/test_net_config_load.py
@@ -23,6 +23,5 @@ def test_network_disabled_via_etc_cloud(client: IntegrationInstance):
         )
     _customize_envionment(client)
 
-    # assert client.execute("test -f /etc/netplan/50-cloud-init.yaml").failed
     log = client.read_from_file("/var/log/cloud-init.log")
     assert "network config is disabled by system_cfg" in log
--- a/tests/integration_tests/network/test_net_config_load.py
+++ b/tests/integration_tests/network/test_net_config_load.py
@@ -1,0 +1,28 @@
+"""Test loading the network config"""
+import pytest
+
+from tests.integration_tests.instances import IntegrationInstance
+
+
+def _customize_envionment(client: IntegrationInstance):
+    # Insert our "disable_network_config" file here
+    client.write_to_file(
+        "/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg",
+        "network: {config: disabled}\n",
+    )
+    client.execute("cloud-init clean --logs")
+    client.restart()
+
+
+def test_network_disabled_via_etc_cloud(client: IntegrationInstance):
+    """Test that network can be disabled via config file in /etc/cloud"""
+    if client.settings.CLOUD_INIT_SOURCE == "IN_PLACE":
+        pytest.skip(
+            "IN_PLACE not supported as we mount /etc/cloud contents into the "
+            "container"
+        )
+    _customize_envionment(client)
+
+    # assert client.execute("test -f /etc/netplan/50-cloud-init.yaml").failed
+    log = client.read_from_file("/var/log/cloud-init.log")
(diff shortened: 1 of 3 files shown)


tests/integration_tests/network/test_net_config_load.py
```diff
@@ -0,0 +1,28 @@
...
+def test_network_disabled_via_etc_cloud(client: IntegrationInstance):
+    """Test that network can be disabled via config file in /etc/cloud"""
+    if client.settings.CLOUD_INIT_SOURCE == "IN_PLACE":
+        pytest.skip(
+            "IN_PLACE not supported as we mount /etc/cloud contents into the "
+            "container"
+        )
+    _customize_envionment(client)
+
+    # assert client.execute("test -f /etc/netplan/50-cloud-init.yaml").failed
```
blackboxsw, line 26: Did you still want this assert here if so I suspect you'd also need a remove of /etc/netplan/50-cloud-init.yaml up in _customize_environment below cloud-init clean --logs

blackboxsw (commented):
Trying to run through an Oracle test now to confirm expected behavior.

tests/integration_tests/network/test_net_config_load.py
TheRealFalcon: Whoops, no. I meant to remove this. I initially tried that and then discovered that on a "normal" instance, removing the netplan config is a bad idea :)

blackboxsw (approved):
LGTM.
confirmed upgrade behavior without restart via `sudo cloud-init init --local; sudo cloud-init init` allows OracleDatasource to see the disabled config. Same across system reboot network is no longer rendered and instance remains accessible via ssh.
```bash
ubuntu@instance-20220115-0946:~$ grep 'disabled by' /var/log/cloud-init.log 
2022-01-15 17:44:55,657 - stages.py[DEBUG]: network config disabled by system_cfg
2022-01-15 17:44:55,657 - stages.py[INFO]: network config is disabled by system_cfg
2022-01-15 17:44:56,102 - stages.py[DEBUG]: network config disabled by system_cfg
2022-01-15 17:44:56,102 - stages.py[INFO]: network config is disabled by system_cfg
...
ubuntu@instance-20220115-0946:~$ ls /etc/netplan/
ubuntu@instance-20220115-0946:~$
ubuntu@instance-20220115-0946:~$ cloud-init status --long
status: done
time: Sat, 15 Jan 2022 17:52:25 +0000
detail:
DataSourceOracle
ubuntu@instance-20220115-0946:~$ grep Trace /var/log/cloud-init.log 
ubuntu@instance-20220115-0946:~$
```

cloud-init can be re-enabled for networking on oracle by ` sudo rm /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg `

PR #1177, opened 2022-01-10
Remove distutils usage

sshedi:
distutils is getting deprecated soon. Let's replace it with alternatives as suggested in:
https://www.python.org/dev/peps/pep-0632/

## Proposed Commit Message

```
Remove distutils usage
distutils is getting deprecated soon. Let's replace it with alternatives as suggested in:
https://www.python.org/dev/peps/pep-0632/
```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -68,14 +68,13 @@ import json
 import os
 import socket
 import time
-from distutils.spawn import find_executable
 
 import netifaces
 
 from cloudinit import dmi
 from cloudinit import log as logging
 from cloudinit import sources, util
-from cloudinit.subp import ProcessExecutionError, subp
+from cloudinit.subp import ProcessExecutionError, subp, which
 
 PRODUCT_UUID_FILE_PATH = "/sys/class/dmi/id/product_uuid"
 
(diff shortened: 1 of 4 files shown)


holmanb:
Hi @sshedi, thanks for the PR.

In addition to fixing the failing tests, I think that it would be better if we could avoid adding a dependency in this PR.  Since `distutils.version` is only used twice in one file I don't think the 3rd party dep is justified (distutils is standard library). 

Furthermore that usage is for a workaround for a version of `requests` that I doubt is supported with any version of python we support. According to the requests [1.0.0 pypi page](https://pypi.org/project/requests/1.0.0/), Python 3.3 was the highest Python version supported at the time of the 1.0.0 release.


It looks like requests on Ubuntu Bionic, for example, is currently 
```
root@me:~# python3
Python 3.6.9 (default, Dec  8 2021, 21:08:43) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests
>>> requests.__version__
'2.18.4'
>>> quit()
```

Removing this workaround code and adding a minimum version to requests might be a workable solution.

sshedi:
Thanks @holmanb for the inputs. I will do the needed.

sshedi:
@holmanb & @TheRealFalcon can you guys please take a look at integration test failures?

TheRealFalcon (changes requested):
Thanks @sshedi . It'll be good to avoid deprecation warnings on 3.10+.

I left some inline comments. The cause of the integration test failures is one of the changed imports. More details on that inline as well.

TheRealFalcon (approved):
Thanks!

PR #1178, opened 2022-01-10
tests: mock LXD datasource detection in ds-identify on LXD containers

blackboxsw:
## Proposed Commit Message
```
On LXD containers /dev/lxd/sock will always exist.

Mock dscheck_LXD to return 1 (NOT_FOUND) to avoid leaking into
test environment and returning LXD as detected.

We have integration tests covering proper LXD datasource detection
so we don't need a Unit test validating the [ -S /dev/lxd/sock] that
is in ds-identify.
```

## Test Steps
```bash
lxc launch ubuntu-daily:bionic dev-b
lxc config device add dev-b rohome disk source="/home/csmith/src/cloud-init" path="/root/cloud-init/"
lxc exec dev-b bash
tox -e py3 tests/unittests/test_ds_identify.py
```

## Checklist:

### DIFF
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -92,6 +92,7 @@ MOCK_VIRT_IS_CONTAINER_OTHER = {
     "RET": "container-other",
     "ret": 0,
 }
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
Thanks!

PR #1179, opened 2022-01-11
Add support for gentoo templates and cloud.cfg

vteratipally:
## Add support for gentoo in cloud.cfg and templates

```

Add support for gentoo in cloud.cfg and templates.
This is an already existing patch in gentoo upstream https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/cloud-init/files/cloud-init-20.4-gentoo-support-upstream-templates.patch.
Cherry-picking the patch to the upstream cloud-init.
```

## Checklist:

### DIFF
--- a/tools/render-cloudcfg
+++ b/tools/render-cloudcfg
@@ -5,7 +5,7 @@ import os
 import sys
 
 VARIANTS = ["almalinux", "alpine", "amazon", "arch", "centos", "cloudlinux", "debian",
-            "eurolinux", "fedora", "freebsd", "miraclelinux", "netbsd", "openbsd", "openEuler", "photon",
+            "eurolinux", "fedora", "freebsd", "gentoo", "miraclelinux", "netbsd", "openbsd", "openEuler", "photon",
             "rhel", "suse","rocky", "ubuntu", "unknown", "virtuozzo"]
 
 
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -176,7 +176,7 @@ cloud_final_modules:
 system_info:
    # This will affect which distro class gets used
 {% if variant in ["almalinux", "alpine", "amazon", "arch", "centos", "cloudlinux", "debian",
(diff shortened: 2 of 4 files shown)


holmanb:
Hi @vteratipally, thanks for the PR. A while ago somebody tried to get something similar merged, but it never landed. I think this is a welcome contribution, but tests don't currently pass. Once tests are passing I can add a review.

I did notice that the cloud-init ebuild in the gentoo repo seems to have duplicate patches listed, which makes me wonder if the gentoo patches even work. Additional context would be helpful. Would you happen to know a) if this package builds cleanly and b) where/how this package gets used? This information will help with testing.

vteratipally:
Yes the gentoo ebuild with the patches works fine. For example, it is used extensively here https://cos.googlesource.com/cos/overlays/board-overlays/+/refs/heads/master/project-lakitu/app-emulation/cloud-init/ for building the cloud-init package into the Container Optimized OS.

tools/render-cloudcfg
```diff
@@ -4,6 +4,9 @@ import os
 import sys
 import argparse
 
+VARIANTS = ["almalinux", "alpine", "amazon", "arch", "centos", "cloudlinux", "debian",
```
holmanb, line 7: This list changed locations on the main branch since your PR was opened. You'll need to update the list below rather than the change here.

holmanb (changes requested):
Thanks for the feedback @vteratipally. This does require a small change, but otherwise I think it's probably good.

tools/render-cloudcfg
vteratipally: Thank you for the review @holmanb . I have updated the file. Thanks.

TheRealFalcon:
@holmanb , this look good to you now? If so, I can merge it.

holmanb:
@TheRealFalcon - Yep, looks good to me!

TheRealFalcon (approved):
Thanks!

PR #1180, opened 2022-01-11
Fix NetworkManagerActivator error

wzfei012:
Load connection before execute "nmcli connection up"。By the way, reload network connections maybe required .

### DIFF
--- a/cloudinit/net/activators.py
+++ b/cloudinit/net/activators.py
@@ -141,6 +141,34 @@ class NetworkManagerActivator(NetworkActivator):
         return _alter_interface(cmd, device_name)
 
     @staticmethod
+    def bring_up_interfaces(device_names: Iterable[str]) -> bool:
+        """Bring up specified list of interfaces.
+
+        Return True is successful, otherwise return False
+        """
+        for device_name in device_names:
+            """Load new connection"""
+            cmd = ["nmcli", "connection", "load", f'/etc/sysconfig/network-scripts/ifcfg-{device_name}']
+            ok = _alter_interface(cmd, device_name)
+            if not ok:
+                return False
+            """Up new connection"""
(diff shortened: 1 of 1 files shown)


wzfei012:
I tried to configure network-scripts on my machine,but cloud-init does not effect at all 。Then I found NM_CONTROLLED was set default to "no"。When I set NM_CONTROLLED  to "yes" and try again, it still does not work. However,the scripts at /etc/sysconfig/network-scripts are all right.
I found while Bring up interface using nmcli,new connection has to be load first.
Sometimes,existing connections will effect new connections which we hope to configure.So I reload network at last

TheRealFalcon:
Looks like this solution is sysconfig specific. I'm not very knowledgeable of sysconfig. Is there a way to load the connection without referencing the specific file (i.e., in a non-sysconfig specific way)? If not, can you first check to see if the file exists before attempting to load it?

Also, why do we need to completely take down networking in order to bring up a new interface? We don't want to take down existing connections if we don't have to.

otubo:
This PR seems to fix the problem we're having on this bug: https://bugzilla.redhat.com/show_bug.cgi?id=2036060

@wzfei012 what's the status on making it move forward? I see it's been 8 days since last comment. I can take over and rework a bit if you will.

Thanks!

anhvoms:
> This PR seems to fix the problem we're having on this bug: https://bugzilla.redhat.com/show_bug.cgi?id=2036060
> 
> @wzfei012 what's the status on making it move forward? I see it's been 8 days since last comment. I can take over and rework a bit if you will.
> 
> Thanks!

@otubo  I don't have access to that bug. Can you add me to the bug?

tyll:
all commands except the nmcli connection load and nmcli connection up commands should not be necessary. The commit message does not explain the reasons the other commands are needed. What should be accomplished with this patch?

PR #1182, opened 2022-01-11
Debian: tell apt-get not to check dates

maxnet:
## Proposed Commit Message

```
Debian: tell apt-get not to check dates

* Date and time may not always be set correctly, especially on systems without RTC.
* "apt-get update" fails if current time is not set correctly, and the 
  Releases file mentions a Date in the future.
* Ideally, we would order systemd units to wait for NTP synchronization to be complete 
  before using any "apt" function. 
  However there may still be a chicken and the egg problem here if we
  do that, if NTP client is not installed yet in image, 
  and the cc_ntp module wants to use apt-get to install it...

So just tell apt-get not to perform date/time checks.

LP: #1951639 
```

## Checklist:

(Note, I am currently not in .github-cla-signers yet, but there is another PR that takes care of that).

### DIFF
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -24,6 +24,7 @@ APT_GET_COMMAND = (
     "apt-get",
     "--option=Dpkg::Options::=--force-confold",
     "--option=Dpkg::options::=--force-unsafe-io",
+    "--option=Acquire::Check-Date=false",
     "--assume-yes",
     "--quiet",
 )


holmanb:
> Any way to force Travis to do a rebuild without adding commits?

I typically force-push a no-op commit.

maxnet:
>I typically force-push a no-op commit.

Closed and reopened the PR, which also seems to work.

This PR passes checks now.
But noticed it was failing for other PRs on ssh-import-id as well: https://github.com/canonical/cloud-init/pull/1183/checks?check_run_id=4778858425
You may want to consider changing the setup to skip that test, if it is depending on external servers and therefore temperamental...

TheRealFalcon:
@maxnet , since this will disable all time related checks, which can be a security concern, I don't think this is this something we want to add as a default hardcoded option.

Would adding it into a cloud-config work for you? Something like:
```yaml
#cloud-config
apt:
  conf: |
    Acquire {
      Check-Date "false";
    };
```

maxnet:
>since this will disable all time related checks, which can be a security concern

Recall the date related headers in Release files were originally introduced to prevent stale mirrors being used, rather than being a real security measure.

>Would adding it into a cloud-config work for you? Something like:
>
>#cloud-config
>apt:
>  conf: |
>    Acquire {
>      Check-Date "false";
>    };

I would prefer to be able to use the same cloud-init file for more than one Linux distribution, and not add something so Debian specific there.


Perhaps another solution would be to create a new cloud-init module that can perform time synchronization in Python code, so that will always be available regardless whether NTP is in a functional state yet.
(Simple time synchronization against an arbitrary HTTP server doesn't have to take more than a handful lines of code, does not have to be anything more sophisticated).

TheRealFalcon:
> I would prefer to be able to use the same cloud-init config for more than one Linux distribution, and not add something so Debian specific there.

Templating would work here. Something like:
```yaml
## template: jinja
#cloud-config
{% if v1.distro in ["debian", "ubuntu"] %}
apt:
  conf: |
    Acquire {
      Check-Date "false";
    };
{% endif %}

```

maxnet:
>Templating would work here.

Would still require adjusting the template everytime somebody introduces a new OS.

I have the somewhat unique use-case that I do know in advance which operating system will be installed exactly by the end-user.
So any kind of OS independent abstraction to signal that the date/time of the system cannot be trusted would be preferred over templates that may need adjustment sooner or later.

TheRealFalcon:
Backing up a little...you wouldn't even need that jinja template for the apt section to do what you want. If you have the apt section defined, cloud-init will only run it on ubuntu and debian. It'll get ignored on other distros.

maxnet:
>If you have the apt section defined, cloud-init will only run it on ubuntu and debian.

And when variant is "unknown"?!

>{% if variant in ["ubuntu", "unknown", "debian"] %}
> - grub-dpkg
> - apt-pipelining
> - apt-configure
>{% endif %}

TheRealFalcon:
Sorry, the template was a red herring and not needed here at all. Modules can declare what distros they run on. [Apt will only run on debian and ubuntu](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_apt_configure.py#L33)

That's without templating using a cloud-config like this:
```yaml
#cloud-config
apt:
  conf: |
    Acquire {
      Check-Date "false";
    };
```

maxnet:
>Modules can declare what distros they run on.

Thanks for the info.
(I had wrongfully assumed cloud.cfg.tmpl decided what got run, and what modules declare was more advisory/for documentation purposes)

Ok, will play a bit with apt configuration then, and see how far that will get me.

PR #1183, opened 2022-01-11
Update parsing of ip addr show (SC-723)

TheRealFalcon:
## Proposed Commit Message

```
Update parsing of ip addr show

"ip addr show" may include a metric on the inet line.
Parse this while also keeping backwards compatibility
for lines without.
```

## Additional Context
Example `ip addr show` on Jammy:
```
root@jammy:~# ip addr show
    inet6 ::1/128 scope host 
118: eth0@if119: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 00:16:3e:80:4f:ea brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.85.130.57/24 metric 100 brd 10.85.130.255 scope global dynamic eth0
       valid_lft 2912sec preferred_lft 2912sec
    inet6 fd42:baa2:3dd:17a:216:3eff:fe80:4fea/64 scope global dynamic mngtmpaddr noprefixroute 
       valid_lft 3545sec preferred_lft 3545sec
    inet6 fe80::216:3eff:fe80:4fea/64 scope link 
```
Notice the `metric 100` on the inet line.

## Test Steps
Can be seen in jammy integration tests on LXD as a new unexpected warning.

## Checklist:

### DIFF
--- a/cloudinit/netinfo.py
+++ b/cloudinit/netinfo.py
@@ -57,8 +57,10 @@ def _netdev_info_iproute(ipaddr_out):
             devs[dev_name]["ipv6"].append(m.groupdict())
         elif "inet" in line:
             m = re.match(
-                r"\s+inet\s(?P<cidr4>\S+)(\sbrd\s(?P<bcast>\S+))?\sscope\s"
-                r"(?P<scope>\S+).*",
+                r"\s+inet\s(?P<cidr4>\S+)"
+                r"(\smetric\s(?P<metric>\d+))?"
+                r"(\sbrd\s(?P<bcast>\S+))?"
+                r"\sscope\s(?P<scope>\S+).*",
                 line,
             )
             if not m:
--- a/tests/data/netinfo/sample-ipaddrshow-output-metric
+++ b/tests/data/netinfo/sample-ipaddrshow-output-metric
@@ -1,0 +1,12 @@
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host \       valid_lft forever preferred_lft forever
+2: enp0s25: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
+    link/ether 50:7b:9d:2c:af:91 brd ff:ff:ff:ff:ff:ff
+    inet 192.168.2.18/24 metric 100 brd 192.168.2.255 scope global dynamic enp0s25
+       valid_lft 84174sec preferred_lft 84174sec
+    inet6 fe80::7777:2222:1111:eeee/64 scope global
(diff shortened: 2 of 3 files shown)


TheRealFalcon:
pytest parametrization doesn't work with the base class, so I removed the base class from the tests.

The only functional test change is in `test_netdev_iproute_pformat`

tests/data/netinfo/sample-ipaddrshow-output-metric
```diff
@@ -0,0 +1,12 @@
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
```
holmanb, line 3: Are the backslashes on lines 3 and 4 a typo?

TheRealFalcon: I just copied the `sample-ipaddrshow-output` file and added the metric part

cloudinit/netinfo.py
```diff
@@ -57,8 +57,10 @@ def _netdev_info_iproute(ipaddr_out):
             devs[dev_name]["ipv6"].append(m.groupdict())
         elif "inet" in line:
             m = re.match(
-                r"\s+inet\s(?P<cidr4>\S+)(\sbrd\s(?P<bcast>\S+))?\sscope\s"
-                r"(?P<scope>\S+).*",
+                r"\s+inet\s(?P<cidr4>\S+)"
+                r"(\smetric\s(?P<metric>\d+))?"
```
holmanb, line 61: I did a little bit of manual testing with this regex and didn't spot any issues.

holmanb (approved):
lgtm

tests/data/netinfo/sample-ipaddrshow-output-metric
holmanb: I see, nevermind. It doesn't matter what it says after `scope` since the regex ends with `.*`.

mwhudson:
Is there a reason you are not parsing the output of 'ip -json addr show' instead?

TheRealFalcon:
@mwhudson It was proposed, but we decided against it due to lack of availability everywhere.

mwhudson:
> @mwhudson It was proposed, but we decided against it due to lack of availability everywhere.

Ah yes I see it's not supported in xenial.

cloudinit/netinfo.py
blackboxsw: The bigger this regex gets, the more it makes me question whether we should really be extending a fragile regex parsing of human-readable output of ip.  While I originally weighed in to the contrary because cloud-init is supported even more broadly than just Ubuntu, I think it would probably be preferable given that we attempt to try `ip -json` format first, and fallback to the legacy parsing if `-json` param is not present. This way most of the latest *nix distros with recent `ip`  tooling can benefit from simple parsing of machine-readable output and any older distributions will continue to parse correctly with the older `ip` tooling and output formats.

TheRealFalcon:
Closing in favor of https://github.com/canonical/cloud-init/pull/1210

PR #1187, opened 2022-01-13
sources/azure: remove lease file parsing for wireserver IP

cjp256:
Always review the ephemeral DHCP lease when performing to find
the wireserver endpoint specified with option 245.

DHCP option 245 is guaranteed to be provided, which we will now
immediately parse when bringing up ephemeral networking with
_process_dhcp_lease(),  saving the parsed address as
_wireserver_addres (which safely defaults to the expected IP
address if the option is not found for whatever reason).

In practice, the fallback parsing of lease files has never been
necessary because the wireserver IP has never changed to date.

This allows us to safely remove the dhclient_lease_file option
and reduce some complexity through consolidation of lease
handling.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -12,7 +12,6 @@ import re
 import xml.etree.ElementTree as ET
 from collections import namedtuple
 from enum import Enum
-from functools import partial
 from time import sleep, time
 from xml.dom import minidom
 
@@ -28,7 +27,9 @@ from cloudinit.reporting import events
 from cloudinit.sources.helpers import netlink
 from cloudinit.sources.helpers.azure import (
     DEFAULT_REPORT_FAILURE_USER_VISIBLE_MESSAGE,
+    DEFAULT_WIRESERVER_ENDPOINT,
     EphemeralDHCPv4WithReporting,
+    WALinuxAgentShim,
     azure_ds_reporter,
     azure_ds_telemetry_reporter,
     build_minimal_ovf,
@@ -51,7 +52,6 @@ DEFAULT_METADATA = {"instance-id": "iid-AZURE-NODE"}
 # azure systems will always have a resource disk, and 66-azure-ephemeral.rules
 # ensures that it gets linked to this path.
 RESOURCE_DISK_PATH = "/dev/disk/cloud/azure_resource"
-LEASE_FILE = "/var/lib/dhcp/dhclient.eth0.leases"
 DEFAULT_FS = "ext4"
 # DMI chassis-asset-tag is set static for all azure instances
 AZURE_CHASSIS_ASSET_TAG = "7783-7084-3265-9085-8269-3286-77"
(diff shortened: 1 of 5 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -895,6 +896,18 @@ def _create_report_ready_marker(self):
             logger_func=LOG.debug,
         )
 
+    def _process_dhcp_lease(self, lease: dict) -> None:
+        """Maintain relevant info for leases."""
+        dhcp_opt_245 = lease.get("unknown-245")
+        if dhcp_opt_245:
+            self._wireserver_address = (
+                WALinuxAgentShim.get_ip_from_lease_value(dhcp_opt_245)
```
anhvoms, line 904: move this function out of the shim? Previously its usage was restricted to the shim, if we're using it outside of the shim, we should just put it as a helper function in azure.py

cloudinit/sources/DataSourceAzure.py
```diff
@@ -895,6 +896,18 @@ def _create_report_ready_marker(self):
...
+            )
+            LOG.debug(
```
anhvoms, line 906: report a diag event containing wireserver IP obtained from lease

cloudinit/sources/DataSourceAzure.py
```diff
@@ -959,7 +973,8 @@ def _check_if_nic_is_primary(self, ifname):
                 dhcp_ctx = EphemeralDHCPv4(
                     iface=ifname, dhcp_log_func=dhcp_log_cb
                 )
-                dhcp_ctx.obtain_lease()
+                lease = dhcp_ctx.obtain_lease()
```
anhvoms, line 976: nit: if we're going to call process_dhcp_lease every time we obtain_lease, it might be better to add an optional callback_fn to obtain_lease and invoke that after getting the lease?

tests/unittests/sources/test_azure.py
```diff
@@ -1734,7 +1734,7 @@ def test_dsaz_report_failure_description_msg(self):
             test_msg = "Test report failure description message"
             self.assertTrue(dsrc._report_failure(description=test_msg))
             self.m_report_failure_to_fabric.assert_called_once_with(
-                dhcp_opts=mock.ANY, description=test_msg
+                endpoint="168.63.129.16", description=test_msg
```
anhvoms, line 1737: import this from helpers.azure

anhvoms (changes requested):

cloudinit/sources/DataSourceAzure.py
cjp256: will do

cloudinit/sources/DataSourceAzure.py
cjp256: will do

cloudinit/sources/DataSourceAzure.py
cjp256: DHCP will be consolidated in a separate PR to simplify this.

tests/unittests/sources/test_azure.py
cjp256: will do

cjp256:
Closing temporarily, will revisit this as a follow up to another PR to prevent conflicts.

PR #1190, opened 2022-01-14
sources/azure: consolidate DHCP variants to EphemeralDHCPv4WithReporting

cjp256:
- Update EphemeralDHCPv4WithReporting to subclass EphemeralDHCPv4 for
  consistency (non-functional change).

- Replace all usage of EphemeralDHCPv4 with EphemeralDHCPv4WithReporting.

- Converging to one DHCP class exposed an issue with ExitStack patches
  being mixed with decorators.  Specifically, it appeared that tests
  that did not enable azure.EphemeralDHCPv4WithReporting mocks had it
  applied anyways from previous tests.

  Presumably ExitStack was overwriting the actual value with the
  mock provided by the decorator?  For now, remove some mock patches
  that trigger failures, but future work should move towards a
  consistent approach to prevent undetected effects.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -23,7 +23,6 @@ from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
 from cloudinit.net import device_driver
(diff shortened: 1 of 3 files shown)


anhvoms (approved):
LGTM

TheRealFalcon (approved):
Thanks!

PR #1191, opened 2022-01-14
[RFC] net/dhcp: set default dhclient config

cjp256:
When bringing up early networking, dhcp_discovery() invokes
dhclient which will use the host's default config.  In practice,
this may result in sending the system's hostname which may not
have been configured yet, polluting dynamic DNS services.

Write a dhclient.conf to requests the various DHCP options
without sending host-name or other details that may not be
correct until configured by cloud-init.

This will allow for standardizing/customizing behavior across
distros for ephemeral networking (for example, 300s timeout).
Configuration to customize these knobs may be desirable (either
from a data source or user perspective).

Thoughts? :D


TheRealFalcon:
cc @blackboxsw and @holmanb . Looks reasonable to me.

holmanb:
In general I don't have any issues with this.

We might need to update the docs. Currently the [hostname module](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#set-hostname) states:
> This module will run in the init-local stage before networking is configured if the hostname is set by metadata or user data on the local system.
>
> This will occur on datasources like nocloud and ovf where metadata and user data are available locally. This ensures that the desired hostname is applied before any DHCP requests are preformed on these platforms where dynamic DNS is based on initial hostname.

It sounds like early networking is setting the hostname before the hostname module runs, so this is out of date. Correct?

cjp256:
> It sounds like early networking is setting the hostname before the hostname module runs, so this is out of date. Correct?

Correct, at least for images that have `send host-name = gethostname();` or similar in their config (like Ubuntu's /etc/dhcp/dhclient.conf).

anhvoms:
In general this is the right direction because we want dhcp discovery behavior to be somewhat consistent. However, this might bring surprised to some customers and cloud providers. I would suggest that dhcp_discovery take in some option that allow datasources to choose whether or not they want to use the image's dhclient config, or use cloud-init builtin config.

TheRealFalcon:
> add a string flag like dhclient_conf: Optional[str] = None which gives the data source full control, but assumes all of the maintenance.

In this case, `None` would still be current behavior, but also allows you to pass the config, correct? If so, this is my preferred option.

PR #1192, opened 2022-01-14
Seek interfaces, skip dummy, fix region codes

eb3095:
## Proposed Commit Message
Seek interfaces on Vultr opposed to using the first, fix region codes

```
We were seeing issues where if anything showed up before the expected first adapter, booting could fail. This switches to seeking for a working interface to handle edge cases.

Also fixes region code handling.
```

## Checklist:


TheRealFalcon (approved):
Looks good, but please run `tox -e format` so the code passes the `black` check.

PR #1193, opened 2022-01-14
sources/azure: set ovf_is_accessible when OVF is read successfully

cjp256:
The if-statement set ovf_is_accessible to True if the OVF is read
from /dev/sr0, but not from other data sources.  It defaults to
True, but may get flipped to False while processing an invalid
source, and never get set back to True when reading from the data
directory.

Instead, default ovf_is_accessible to False, and only set it to
True once we've read an OVF successfully (and end the search).

This fixes an error when OVF is read from data_dir and IMDS
data is unavailable (failing with "No OVF or IMDS available").


anhvoms (approved):

TheRealFalcon (approved):
Thanks!

PR #1194, opened 2022-01-18
sources/azure: unpack ret tuple in crawl_metadata()

cjp256:
load_azure_ds_dir() always returns a tuple.  Instead of saving this
tuple as ret, expand it immediately as md, userdata_raw, cfg, files.

This allows for more fine-grained passing of data before getting
expanded later.

- Update _should_reprovision methods to use cfg instead of tuple.

- Update _should_reprovision methods to remove the ovf_md guard.
 This should be a safe refactor as the OVF is not required, and the
 config is initialized to an empty dict.  In practice, a mount failure
 would have initalized ret anyways if the OVF was not found. If a
 mount failure wasn't seen and ret was None, this guard could be
 causing other failures by ignoring the PPS state that should be
 available from IMDS metadata.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -338,7 +338,10 @@ class DataSourceAzure(sources.DataSource):
         ovf_is_accessible = False
         reprovision_after_nic_attach = False
         metadata_source = None
-        ret = None
+        md = {}
+        userdata_raw = ""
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

cjp256:
*rebased

TheRealFalcon (approved):
LGTM, thanks!

PR #1195, opened 2022-01-18
Enable OVF environment transport via ISO in example

megian:
This enables the OVF enviroment transport via ISO variable as defined
in the Open Virtualization Format Specification Version: 2.1.1.

8.1 VirtualHardwareSection
11.1 Transport media

Reference:
https://www.dmtf.org/sites/default/files/standards/documents/DSP0243_2.1.1.pdf

## Test Steps
* Create a template using the OVF example
* Instantiate a VM from that template
* Check if the content is available via ISO transport

## Checklist:
 - [x] I have updated or added any unit tests accordingly (N/A because it is just an example file)
 - [x] I have updated or added any documentation accordingly (N/A because it is just an example file)


TheRealFalcon (approved):
Thanks!

PR #1197, opened 2022-01-19
cmd: status and cloud-id avoid change in behavior for 'not run'

blackboxsw:
We don't want to break existing snapd. So, let's leave "not run" value untouched. Also note, if we decide to change cloud-init status exit codes, that also has the potential to break snapd too, so we'd need to float that by that team first and get buyin.

## Proposed Commit Message
```
snapd currrently looks for 'not run' from cloud-init status[1].
Avoid changing this behavior and revert "not-run" value to
"not run".

This avoids having to get snapd to change implementation and
release updates as far back as Bionic to handle a hyphenated
not-run string.

[1]: https://github.com/snapcore/snapd/blob/master/sysconfig/\
     cloudinit.go#L802
```

## Checklist:

### DIFF
--- a/cloudinit/cmd/cloud_id.py
+++ b/cloudinit/cmd/cloud_id.py
@@ -63,7 +63,7 @@ def handle_args(name, args):
 
     Print the canonical cloud-id on which the instance is running.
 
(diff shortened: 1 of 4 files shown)


TheRealFalcon (approved):
+1 assuming CI passes

PR #1199, opened 2022-01-19
sources/azure: always initialize _ephemeral_dhcp_ctx on unpickle

cjp256:
Avoid requirement of getattr() and ensure _ephemeral_dhcp_ctx isn't
persisted in the cache.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -518,6 +518,7 @@ class DataSourceAzure(sources.DataSource):
             use_cached_ephemeral = (
                 self.distro.networking.is_up(self.fallback_interface)
                 and self._ephemeral_dhcp_ctx
+                and self._ephemeral_dhcp_ctx.lease
             )
             if use_cached_ephemeral:
                 self._report_ready(lease=self._ephemeral_dhcp_ctx.lease)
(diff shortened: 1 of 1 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -513,9 +515,10 @@ def crawl_metadata(self):
 
         if perform_reprovision or perform_reprovision_after_nic_attach:
             LOG.info("Reporting ready to Azure after getting ReprovisionData")
-            use_cached_ephemeral = self.distro.networking.is_up(
-                self.fallback_interface
-            ) and getattr(self, "_ephemeral_dhcp_ctx", None)
+            use_cached_ephemeral = (
+                self.distro.networking.is_up(self.fallback_interface)
+                and self._ephemeral_dhcp_ctx
```
anhvoms, line 520: Do we need to check for self._ephemeral_dhcp_ctx.lease here?

anhvoms (approved):

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
cjp256: In practice I expect that a NoDHCPLeaseError exception would have been raised and prevented it from getting this far.

I added an explicit check and I have other PRs in queue to further improve this handling.

TheRealFalcon (approved):
Thanks!

PR #1205, opened 2022-01-21
Fix Gentoo Locales (SC-748)

holmanb:
```
Fix Gentoo locales
```

## Additional Context
Tracking down a stack trace in gentoo cloud LXD images pointed out an issue with localization on gentoo. It's possible that this code once "succeeded" with a different image build. Fix the trace and actually set a locale. 

https://wiki.gentoo.org/wiki/Localization/Guide/en

Stack Trace:
```
2022-01-24 20:04:10,553 - util.py[DEBUG]: Running module locale (<module 'cloudinit.config.cc_locale' from '/usr/lib/python3.9/site-packages/cloudinit/config/cc_locale.py'>) failed
  File "/usr/lib/python3.9/site-packages/cloudinit/stages.py", line 884, in _run_modules
    ran, _r = cc.run(run_name, mod.handle, func_args,
  File "/usr/lib/python3.9/site-packages/cloudinit/cloud.py", line 54, in run
  File "/usr/lib/python3.9/site-packages/cloudinit/helpers.py", line 186, in run
    results = functor(*args)
  File "/usr/lib/python3.9/site-packages/cloudinit/config/cc_locale.py", line 78, in handle
    cloud.distro.apply_locale(locale, locale_cfgfile)
  File "/usr/lib/python3.9/site-packages/cloudinit/distros/gentoo.py", line 43, in apply_locale
    subp.subp(['locale-gen', '-G', locale], capture=False)
  File "/usr/lib/python3.9/site-packages/cloudinit/subp.py", line 293, in subp
    raise ProcessExecutionError(stdout=out, stderr=err,
Command: ['locale-gen', '-G', 'en_US.UTF-8']
Exit code: 4
Reason: -
Stdout: -
Stderr: -
```

## Test Steps
1. Start gentoo cloud container (images:gentoo/cloud), note the above stack trace
2. Mount this cloud-init branch locally into container and check for error:
```
# Define BASE_DIR, NAME, and MY_CLOUD_CONFIG to appropriate values first
IMAGE_ID=images:gentoo/cloud
lxc init $IMAGE_ID $NAME -c user.user-data="$(cat $MY_CLOUD_CONFIG)"

# Gentoo currently uses 3.9 as system default 
lxc config device add $NAME host-cloud-init disk source=$BASE_DIR/cloudinit path=/usr/lib/python3.9/site-packages/cloudinit/
lxc start $NAME
lxc shell me 
```

### DIFF
--- a/cloudinit/distros/gentoo.py
+++ b/cloudinit/distros/gentoo.py
@@ -18,7 +18,7 @@ LOG = logging.getLogger(__name__)
 
 class Distro(distros.Distro):
     locale_conf_fn = "/etc/env.d/02locale"
-    locale_gen_file = "/etc/locale.gen"
+    locale_gen_fn = "/etc/locale.gen"
     network_conf_fn = "/etc/conf.d/net"
     hostname_conf_fn = "/etc/conf.d/hostname"
     init_cmd = ["rc-service"]  # init scripts
@@ -46,9 +46,7 @@ class Distro(distros.Distro):
         Locales need to be added to /etc/locale.gen and generated prior
         to selection. Default to en_US.UTF-8 for simplicity.
         """
-        util.write_file(
-            self.locale_gen_file, "\n".join(self.locales), mode=644
-        )
+        util.write_file(self.locale_gen_fn, "\n".join(self.locales), mode=644)
 
         # generate locales
         subp.subp(["locale-gen"], capture=False)
--- a/cloudinit/distros/gentoo.py
+++ b/cloudinit/distros/gentoo.py
@@ -17,10 +17,16 @@ LOG = logging.getLogger(__name__)
 
 
 class Distro(distros.Distro):
-    locale_conf_fn = "/etc/locale.gen"
+    locale_conf_fn = "/etc/env.d/02locale"
+    locale_gen = "/etc/locale.gen"
     network_conf_fn = "/etc/conf.d/net"
     hostname_conf_fn = "/etc/conf.d/hostname"
     init_cmd = ["rc-service"]  # init scripts
+    default_locale = "en_US.UTF-8"
+
+    # C.UTF8 makes sense to generate, but is not selected
+    # Add /etc/locale.gen entries to this list to support more locales
+    locale_gen = ["C.UTF8 UTF-8", "en_US.UTF-8 UTF-8"]
 
     def __init__(self, name, cfg, paths):
         distros.Distro.__init__(self, name, cfg, paths)
@@ -32,14 +38,25 @@ class Distro(distros.Distro):
         # Fix sshd restarts
         cfg["ssh_svcname"] = "/etc/init.d/sshd"
 
-    def apply_locale(self, locale, out_fn=None):
-        if not out_fn:
-            out_fn = self.locale_conf_fn
-        subp.subp(["locale-gen", "-G", locale], capture=False)
+    def apply_locale(self, _, out_fn=None):
+        """rc-only - not compatible with systemd
+
+        Locales need to be added to /etc/locale.gen and generated prior
+        to selection. Default to en_US.UTF-8 for simplicity.
+        """
+        util.write_file(self.locale_gen, "\n".join(self.locale_gen), mode=644)
+
+        # generate locales
+        subp.subp(["locale-gen"], capture=False)
+
+        # select locale
+        subp.subp(
+            ["eselect", "locale", "set", self.default_locale], capture=False
+        )
         # "" provides trailing newline during join
         lines = [
             util.make_header(),
-            'LANG="%s"' % locale,
+            'LANG="%s"' % self.default_locale,
             "",
         ]
         util.write_file(out_fn, "\n".join(lines))
--- a/cloudinit/distros/gentoo.py
+++ b/cloudinit/distros/gentoo.py
@@ -44,7 +44,9 @@ class Distro(distros.Distro):
         Locales need to be added to /etc/locale.gen and generated prior
         to selection. Default to en_US.UTF-8 for simplicity.
         """
-        util.write_file(self.locale_gen_file, "\n".join(self.locales), mode=644)
+        util.write_file(
+            self.locale_gen_file, "\n".join(self.locales), mode=644
+        )
 
         # generate locales
         subp.subp(["locale-gen"], capture=False)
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Is there any possibility of regressing older launches where this configuration may have been valid? Based on what I know of Gentoo, the answer is probably no, but do we know for sure?

holmanb:
> Is there any possibility of regressing older launches where this configuration may have been valid? Based on what I know of Gentoo, the answer is probably no, but do we know for sure?

I just noticed that downstream actually has a patch on this section of code. I'll see if I can get an opinion from a distro maintainer.

holmanb:
@prometheanfire - I see you are currently the maintainer of `app-emulation/cloud-init`. Would you mind reviewing this PR?

I noticed that [this patch](https://github.com/gentoo-mirror/gentoo/blob/stable/app-emulation/cloud-init/files/18.4-fix-filename-for-storing-locale.patch) overlaps with this section of code. My PR would introduce a change in behavior, since currently locale isn't currently set by cloud-init on Gentoo (it is set on all other distros). Would it be better to bring Gentoo in line with other distros behavior, or maintain current behavior?

cloudinit/distros/gentoo.py
```diff
@@ -31,18 +37,33 @@ def __init__(self, name, cfg, paths):
         self.osfamily = "gentoo"
         # Fix sshd restarts
         cfg["ssh_svcname"] = "/etc/init.d/sshd"
+        if distros.uses_systemd():
+            LOG.error("Cloud-init does not support systemd with gentoo")
```
prometheanfire, line 41: gentoo can use system (or not (another init)), I'm not sure how this function checks for systemd support.

cloudinit/distros/gentoo.py
```diff
@@ -31,18 +37,33 @@ def __init__(self, name, cfg, paths):
...
+        )
         # "" provides trailing newline during join
         lines = [
             util.make_header(),
-            'LANG="%s"' % locale,
+            'LANG="%s"' % self.default_locale,
             "",
         ]
-        util.write_file(out_fn, "\n".join(lines))
+        util.write_file(self.locale_conf_fn, "\n".join(lines))
```
prometheanfire, line 66: writing to `/etc/env.d/02locale` is not necessary if running the `eselect locale` as above.  `eselect locale` writes to `/etc/locale.conf` and `/etc/env.d/02locale` is a symlink to `/etc/locale.conf`


```
ls -l /etc/env.d/02locale
lrwxrwxrwx 14 root 2015-02-02 22:36  /etc/env.d/02locale -> ../locale.conf

cat /etc/locale.conf
# Configuration file for eselect
# This file has been automatically generated.
LANG="en_US.utf8"
```

prometheanfire (changes requested):
I think some of this is good, not sure if the systemd check is needed and the file write to 02locale is can be removed

cloudinit/distros/gentoo.py
holmanb: @prometheanfire - The reason I was considering adding that is because cloud-init doesn't currently have support for systemd with Gentoo. Cloud-init assumes openrc for gentoo. I'm not sure if the ebuild packaging allows cloud-init with systemd, but if so things may fail and this log may help confused users. Alternatively we could add support for systemd to cloud-init for gentoo. I don't think it would be much work to make that happen, but I don't know if anybody would find that useful.

prometheanfire: `distutils-r1_python_install --init-system=sysvinit_openrc,systemd --distro gentoo` is how we are installing cloud-init, so we are at least installing service files and the like already.  Systemd on Gentoo is fully supported, even if gentoo is better known for openrc (I personally use systemd on my laptop with gentoo).

holmanb: @prometheanfire: I think we may be misunderstanding each other in some way.

Gentoo fully supports systemd, but cloud-init unconditionally assumes that Gentoo's init system is OpenRC.

For example, note that the [init command here](https://github.com/canonical/cloud-init/blob/6a4b6dc0de94a9e75bb44fa6c03288c7d4dded7e/cloudinit/distros/gentoo.py#L23) is rc-service, and [`apply_locale()`](https://github.com/canonical/cloud-init/blob/6a4b6dc0de94a9e75bb44fa6c03288c7d4dded7e/cloudinit/distros/gentoo.py#L35) doesn't support the [systemd locale setup with `localectl`](https://wiki.gentoo.org/wiki/Localization/Guide/en#systemd_2).

I haven't tried using cloud-init with a systemd gentoo install, but since cloud-init assumes OpenRC-specific things, I don't think it  could start services correctly.

cloudinit/distros/gentoo.py
holmanb: It seems I misunderstood what this part was doing, thanks. In that case I guess this might have once succeeded and set the locale. It seems that something must have changed (either in image build or `locale-gen` behavior) in some way if `locale-gen -G en_US.UTF-8` used to succeed.

Either way, I think that this adds some improvement by ensuring that `C.UTF8 UTF-8` is contained in `/etc/locale.gen` before trying to generate the local.

cloudinit/distros/gentoo.py
holmanb: @prometheanfire - Thanks for the feedback, it's much appreciated. Would you mind taking a another look at this considering my latest commit and comments above regarding systemd/cloud-init?  We have an upcoming release (22.1) and I'd like to get this in before the release if possible.

cloudinit/distros/gentoo.py
```diff
@@ -17,10 +17,16 @@
 
 
 class Distro(distros.Distro):
-    locale_conf_fn = "/etc/locale.gen"
+    locale_conf_fn = "/etc/env.d/02locale"
+    locale_gen_file = "/etc/locale.gen"
```
prometheanfire, line 21: being pedantic, for consistency's sake the new var should probably be suffixed with `_fn`, but I don't care too much.

prometheanfire (approved):

cloudinit/distros/gentoo.py
holmanb: Thanks!

prometheanfire (approved):

holmanb:
The CI test failure before was unrelated to the change in this PR (a flakey test that relied on network connection). I rebased on main to force tests to run, no other changes. I think this is ready to merge (pending re-review by @TheRealFalcon).

TheRealFalcon (approved):
Thanks!

PR #1206, opened 2022-01-21
sources/azure: refactor _should_reprovision[_after_nic_attach]() logic

cjp256:
Consolidate _should_reprovision_after_nic_attach() with
_should_reprovision() into the following:

- _write_reprovision_marker() to write provisioning marker for
  reboot-during-provisioning case.

- PPSType enum and _determine_pps_type() for determining which to
  provisioning mode, if any, we're running under.

PPSType.UNKNOWN is when the reprovisioning marker is found and we
do not have the context to know what the original mode was. In this
scenario, we must resort to polling for reprovision data.

Tests:

- Introduce a simple data source fixture to for fine-grain
  control of mocking with pytest without unittest.

- Migrate relevant _should_reprovision() tests into a combination of
  TestDeterminePPSTypeScenarios cases.


BigBlueHundids (approved):

anhvoms (approved):
LGTM

TheRealFalcon (approved):
Thanks!

PR #1208, opened 2022-01-22
Fix lo being used for DHCP, try next on cmd fail

eb3095:
## Proposed Commit Message
Fix lo being used for DHCP, try next on cmd fail

```
Arch didn't like the last fix and lo found its way into this list. Checking for this and skipping interfaces that generate subp failures.
```

## Test Steps
Deployed Arch on Vultr

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -24,7 +24,7 @@ def get_metadata(url, timeout, retries, sec_between, agent):
         # Skip dummy, lo interfaces
         if "dummy" in iface[0]:
             continue
-        if 'lo' == iface[0]:
+        if "lo" == iface[0]:
             continue
         try:
             with EphemeralDHCPv4(
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -22,7 +22,9 @@ def get_metadata(url, timeout, retries, sec_between, agent):
     # Seek iface with DHCP
     for iface in net.get_interfaces():
         # Skip dummy, lo interfaces
-        if iface[0] in ['lo', 'dummy']:
+        if "dummy" in iface[0]:
+            continue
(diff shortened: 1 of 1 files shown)


cloudinit/sources/helpers/vultr.py
```diff
@@ -21,8 +21,8 @@ def get_metadata(url, timeout, retries, sec_between, agent):
 
     # Seek iface with DHCP
     for iface in net.get_interfaces():
-        # Skip dummy interfaces
-        if "dummy" in iface[0]:
+        # Skip dummy, lo interfaces
+        if iface[0] in ['lo', 'dummy']:
```
cjp256, line 25: This check would now fail to filter out numbered dummy interfaces, e.g. "dummy0".

FWIW I've been considering taking the filtering done for net.find_fallback_nic() and providing a method so we can query the set of viable interfaces without the data sources having to add similar filtering logic for themselves if fallback_interface is not sufficient.

eb3095: Ah indeed it would. Not sure why I didn't see that... Anyways its fixed, and it looks like find_candidate_nics is exactly what I need. Once that's merged in I will immediately refactor this to use that instead. Probably when I go through an simplify the generation of net metadata. I am looking at making a new endpoint on our end that feeds cloud-init format configs directly instead of doing what we are doing here to eliminate some of these inconsistencies we are seeing due to our metadata bumping heads with the cloud-init format here and there.

cjp256 (approved):

eb3095:
Seems like travis is stuck here? Am I able to kick that from my end to get it going? I dont see an option.

TheRealFalcon:
Hey @eb3095 , it should work now if you push anything to the branch. If there's nothing new, you can force push a trivial change (i.e., an extra space) in the commit message.

TheRealFalcon (approved):
Thanks!

PR #1209, opened 2022-01-24
net: introduce find_candidate_nics()

cjp256:
find_fallback_nic_on_linux(), etc. provides valuable filtering of
network interfaces in an effort to determine the best candidate
for the fallback interface.

Expose this logic with a new set of methods for finding the candidate
network interfaces.  These methods can be used by data sources which
cannot rely on the fallback interface being the correct choice.

Note that the the MAC address filtering is now part of
find_candidate_nics_on_linux().  This should be consistent behavior
as find_fallback_nic_on_linux() never selected an interface without
a MAC.  find_fallback_nic_on_linux() continues to prefer eth0,
but we make no such distinction in the candidate search.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -11,7 +11,7 @@ import ipaddress
 import logging
 import os
 import re
-from typing import Any, Dict
+from typing import Any, Dict, List, Optional
 
 from cloudinit import subp, util
 from cloudinit.net.network_state import mask_to_net_prefix
@@ -389,8 +389,23 @@ def is_disabled_cfg(cfg):
     return cfg.get("config") == "disabled"
 
 
-def find_fallback_nic(blacklist_drivers=None):
-    """Return the name of the 'fallback' network device."""
+def find_candidate_nics(blacklist_drivers: Optional[List] = None) -> List[str]:
+    """Get the list of network interfaces viable for networking.
+
+    @return List of interfaces, sorted naturally.
+    """
+    if util.is_FreeBSD() or util.is_DragonFlyBSD():
+        return find_candidate_nics_on_freebsd(blacklist_drivers)
+    elif util.is_NetBSD() or util.is_OpenBSD():
+        return find_candidate_nics_on_netbsd_or_openbsd(blacklist_drivers)
+    else:
+        return find_candidate_nics_on_linux(blacklist_drivers)
+
+
+def find_fallback_nic(
+    blacklist_drivers: Optional[List] = None,
+) -> Optional[str]:
+    """Get the name of the 'fallback' network device."""
     if util.is_FreeBSD() or util.is_DragonFlyBSD():
         return find_fallback_nic_on_freebsd(blacklist_drivers)
     elif util.is_NetBSD() or util.is_OpenBSD():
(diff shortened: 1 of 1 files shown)


anhvoms (approved):

cjp256:
One thing I was considering was to adjust find_candidate_nics_on_linux() was to instead return both connected and possibly connected, as a tuple so callers can determine for themselves if they want to process possibly connected interfaces (presumably after checking connected interfaces).

e.g.
`return sorted(connected, key=natural_sort_key), sorted(possibly_connected, key=natural_sort_key)`


...rather than only providing possibly_connected if there are no connected.

TheRealFalcon:
@cjp256 , This is a nice refactor. Thanks!

As far as your question, do have a particular use case in mind for the tuple? I don't really see a use case for using a possibly connected interface when you have interfaces you're sure are connected. Do you foresee additional filtering you'll need to do?

If you have a use case for it, then go for it. I would just request you return a namedtuple to keep call sites more readable. If you don't have an immediate use case, I prefer it as-is. If we have a need for it in the future, it shouldn't be a big refactor.

Since you put RFC in the title, I'm not sure how much more you're looking to add here, or if there's more to discuss. Let me know when you're ready for a final review/merge.

cjp256:
@TheRealFalcon Yeah the tuple was not a great idea :D

I added a commit which should address the issue of ignoring potential candidates just because some interfaces are connected.
 
The candidates will be ordered as:  

 1. DEFAULT_PRIMARY_INTERFACE, if connected.
 2. Remaining connected interfaces, naturally sorted.
 3. DEFAULT_PRIMARY_INTERFACE, if possibly connected.
 4. Remaining possibly connected interfaces, naturally sorted.
 
 This way the fallback calculation is consistent with previous
 behavior, but we don't ignore possibly connected interfaces just
 because others may be connected.

cloudinit/net/__init__.py
```diff
@@ -491,24 +545,34 @@ def find_fallback_nic_on_linux(blacklist_drivers=None):
...
-            return name
+    # Order the NICs:
+    # 1. DEFAULT_PRIMARY_INTERFACE, if connected.
+    # 2. Remaining connected interfaces, naturally sorted.
+    # 3. DEFAULT_PRIMARY_INTERFACE, if possibly connected.
+    # 4. Remaining possibly connected interfaces, naturally sorted.
+    sorted_interfaces = []
+    for interfaces in [connected, possibly_connected]:
+        interfaces = sorted(interfaces, key=natural_sort_key)
+        if DEFAULT_PRIMARY_INTERFACE in connected:
```
TheRealFalcon, line 556: This should be `interfaces` rather than `connected`, correct?

TheRealFalcon (commented):
Thanks @cjp256 , I like this approach. Let's go with it. Just one comment inline.

cloudinit/net/__init__.py
cjp256: Ah yes, indeed, thanks!

TheRealFalcon:
@cjp256 , not trying to rush you, but were you still planning on adding additional tests here?

cjp256:
> @cjp256 , not trying to rush you, but were you still planning on adding additional tests here?

Yeppers, I'll do it Monday thanks!

PR #1155, opened 2021-12-14
Ensure schemas for all modules starting A-D (SC-646)

TheRealFalcon:
## Proposed Commit Message

```
Ensure schemas for all modules starting A-D

Provide/update schemas (including docs) for the following modules:
  cc_apk_configure
  cc_apt_configure
  cc_apt_pipelining
  cc_bootcmd
  cc_byobu
  cc_ca_certs
  cc_debug
  cc_disable_ec2_metadata
  cc_disk_setup
```

## Additional Context
Checked/Updated so that for every schema defintion we:
* define the schema and meta dictionaries per metaschema
* define `__doc__` using `get_meta_doc`
* call `validate_cloudconfig_schema` early in `handle`
* ensure `test_schema.py` passes
* generated docs look correct

For very large schema and doc definitions (e.g., apt configure and disk setup), I moved their definitions to their own files. At times I lean towards defining these in non-python files entirely, but most schemas are much smaller and simpler.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [x] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #1211, opened 2022-01-24
Schema a d

blackboxsw:
Based on [James' schema-a-d PR 1155l](https://github.com/canonical/cloud-init/pull/1155) migrate legacy schema definitions to static cloud-init-schema.json.

modules migrated, extended and schema unittests added:
 - cc_apt_configure
 - cc_bootcmd
 - cc_byobu
 - cc_chef  (migrated)
 - cc_debug
 - cc_disable_ec2_metadata

## Proposed Commit Message

```
schema: add static JSON schema for some a-c config modules

Migrate from legacy schema or define new schema in 
cloud-init-schema.json, adding extensive schema tests for:
- cc_apt_configure
- cc_bootcmd
- cc_byobu
- cc_ca_certs
- cc_chef
- cc_debug
- cc_disable_ec2_metadata
- cc_disk_setup

- Add schema alias `ca_certs.remove_defaults` for `ca-certs.remove-defauts`
- Add deprecation warnings for use or ca-certs and remove-defaults
- Extend apt_configure schema
   - Define more strict schema below object opaque keys using
     patternProperties
  -  create common $def apt_configure.mirror for reuse in 'primary' 
     and 'security' schema definitions within cc_apt_configure

 Co-Authored-by: James Falcon <james.falcon@canonical.com>
```
## Test Steps

```bash
# review generated docs
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_apt_configure
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_bootcmd
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_byobu
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_chef
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_ca_certs
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_debug
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_disable_ec2_metadata

# Exercise valid/invalid schema
tox -e py3 tests/unittests/config
```

## Checklist:


holmanb (changes requested):
What I've seen looks good so far; I like the general approach. I still have more to review, but there are a couple of minor nits I'll share right now.

cloudinit/config/cc_bootcmd.py
```diff
@@ -13,7 +13,7 @@
 from textwrap import dedent
 
 from cloudinit import subp, temp_utils, util
-from cloudinit.config.schema import get_meta_doc, validate_cloudconfig_schema
+from cloudinit.config.schema import get_meta_doc
```
TheRealFalcon, line 16: Let's remove the comment on lines 21-25 of this file.

TheRealFalcon (approved):
LGTM! One inline nit and a question (in addition to Brett's comments).

You mentioned you thought there might be an issue with disk setup, but I don't remember the details. Can you explain the issue?

cloudinit/config/cloud-init-schema.json
```diff
@@ -60,10 +187,364 @@
...
+                  "type": "string",
+                  "default": "mbr",
+                  "enum": ["mbr", "gpt"],
+                  "description": "Specifies the partition table type, either ``mbr`` or ``gpt``. Default: ``mbr``."
+                },
+                "layout": {
+                  "type": ["string", "boolean", "array"],
+                  "default": false,
+                  "oneOf": [
+                    {"type": "string", "enum": ["auto", "remove"]},
```
blackboxsw, line 457: @TheRealFalcon here is the "error" in both disk-setup: documentation and schema definition I think. I can't find in our python code where layout of "auto" is honored or used so I think it might not be correct (maybe a thinko for the original module writer due to "partition" key supporting "auto"?

TheRealFalcon: @blackboxsw , yes, good catch. `auto` isn't valid here. I suspect it may have been sometime in the past because there was also an example that had `auto` (that I removed), but apparently I missed taking it out of the schema. We should be able to safely remove it.

blackboxsw (commented):
@holmanb @TheRealFalcon I've just pushed additional changes to add a warning log about deprecation of "ca-certs" key in favor of "ca_certs" to make sure we don't suprise folks by rejecting that in the new strict schema.

Also, I've aliased a "ca-certs" key to "ca_certs" in the cloud-init schema so schema validation doesn't raise an error or warning on the deprecated ca-certs key.


Two more questions I have: 
1.  `disk-setup:` layout seems to include an "auto" value in both docs and schema, but I don't see the matching cc_disk_setup logic that honors this value. So, I'm not certain we need it.
2. `apt_configure`: has a [convert_v1_to_v2_apt_format](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_apt_configure.py#L889) which supports a different schema for `apt_configure` than we have captured in our current strict schema. Shall we emit deprecation warnings for old uses of `apt` in this release? 
3. Should we define our current strict schema to still support the "old schema"  for apt_configure?

cloudinit/config/cloud-init-schema.json
blackboxsw: +1 will remove it now.

cloudinit/config/cc_ca_certs.py
```diff
@@ -162,20 +169,33 @@ def handle(name, cfg, cloud, log, _args):
     @param log: Pre-initialized Python logger object to use for logging.
     @param args: Any module arguments from cloud.cfg
     """
-    # If there isn't a ca-certs section in the configuration don't do anything
-    if "ca-certs" not in cfg:
+    if "ca-certs" in cfg:
+        log.warning(
+            "DEPRECATION: key 'ca-certs' is now deprecated. Use 'ca_certs' by"
+            " version 23.1."
```
TheRealFalcon, line 175: I don't think this date is reasonable unless it's only for upstream. In order to maintain backwards compatibility, we either will have to patch downstream or extend this date to April 2025.

cloudinit/config/cc_ca_certs.py
```diff
@@ -162,20 +169,33 @@ def handle(name, cfg, cloud, log, _args):
...
+        )
+        ca_cert_cfg = cfg["ca-certs"]
+    elif "ca_certs" not in cfg:
```
TheRealFalcon, line 178: We should change the debug text in this condition to say `ca_certs`

cloudinit/config/cc_ca_certs.py
```diff
@@ -162,20 +169,33 @@ def handle(name, cfg, cloud, log, _args):
...
         log.debug(
             "Skipping module named %s, no 'ca-certs' key in configuration",
             name,
         )
         return
-
-    ca_cert_cfg = cfg["ca-certs"]
+    else:
```
TheRealFalcon, line 184: There's that small possibility they provide both `ca-certs` and `ca_certs`. In that case I think we should log a warning and use `ca_certs`.

cloudinit/config/cc_ca_certs.py
```diff
@@ -162,20 +169,33 @@ def handle(name, cfg, cloud, log, _args):
...
     distro_cfg = _distro_ca_certs_configs(cloud.distro.name)
 
-    # If there is a remove-defaults option set to true, remove the system
+    # If there is a remove_defaults option set to true, remove the system
     # default trusted CA certs first.
-    if ca_cert_cfg.get("remove-defaults", False):
+    if "remove-defaults" in ca_cert_cfg:
+        log.warning(
+            "DEPRECATION: key 'ca-certs.remove-defaults' is now deprecated."
+            " Use 'ca_certs.remove_defaults' by version 23.1."
```
TheRealFalcon, line 193: Same thing about the date here

TheRealFalcon (changes requested):
> apt_configure: has a convert_v1_to_v2_apt_format which supports a different schema for apt_configure than we have captured in our current strict schema. Shall we emit deprecation warnings for old uses of apt in this release?

Yes, I think that makes sense, especially since we don't even document v1 anymore.

> Should we define our current strict schema to still support the "old schema" for apt_configure?

I don't think so. It's still just a warning, right? If they got the warning before, they'll continue to get the warning now.

holmanb:
>     2. `apt_configure`: has a [convert_v1_to_v2_apt_format](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_apt_configure.py#L889) which supports a different schema for `apt_configure` than we have captured in our current strict schema. Shall we emit deprecation warnings for old uses of `apt` in this release?

I'm generally in favor of deprecating old formats so we can lay this kind of conversion code to rest. That said, user experience around these kinds of "forced changes" is probably important.

Do we have a well defined user story around handling config deprecation in cloud-init? If we don't, maybe we could enhance `cloud-init devel schema` to warn of deprecated config format usage (which would maybe check some metadata stored in the schema?). In my mind this lands outside the scope of this PR, but since this PR touches on deprecation in a couple different ways I wanted to bring it up to see what others' thoughts are.

holmanb (changes requested):
Spelling nits:
```
spellintian cloudinit/config/cloud-init-schema.json
cloudinit/config/cloud-init-schema.json: dependancies -> dependencies
cloudinit/config/cloud-init-schema.json: partion -> partition
```

cloudinit/config/cc_ca_certs.py
blackboxsw: Agreed, setting this drop version or date is arbitrary.  Ideally, I'd like to drop this everywhere, but I don't know what line in the sand we should set for this. I don't love the idea of patching out behavior on older releases..... but in order to retain original behavior, maybe that is a debt we need to pay to move upstream forward to a unified/strict schema. We can sort how to handle deprecation timing separately in a different PR for these types of conditions (ca-certs -> ca_certs deprecation.  and apt_configure v1 vs v2 deprecation).

cloudinit/config/cc_ca_certs.py
blackboxsw: +1. as it is currently. I've just added the DEPRECATION warning and dropped the version suggestion from this PR. We can sort how/when to deprecate a number of schema related values as a separate PR since there will be multiple cases by the time we get through all cc_* modules.

blackboxsw:
>  Do we have a well defined user story around handling config deprecation in cloud-init? If we don't, maybe we could enhance cloud-init devel schema to warn of deprecated config format usage (which would maybe check some metadata stored in the schema?). In my mind this lands outside the scope of this PR, but since this PR touches on deprecation in a couple different ways I wanted to bring it up to see what others' thoughts a

Unfortunately, we don't. I agree with a deprecation warning generally from schema validate to better raise awareness of that expected failure in the future. I've seen a couple of public JSON schemas that add a "deprecated" bool to a property in the schema definition but we also risk mixing metadata w/ true/strict schema definitions.  Maybe it's still worth injecting that data into our schema to better raise awareness of the eventual drop of said properties. We could also leverage such an attribute for our generated docs to render that deprecation warning as well in readthedocs content.  +1 on outside the scope of this PR, but let's talk on this and come up with a consensus on best approach to handle this case and put a PR during our schema work to address it.

cloudinit/config/cc_ca_certs.py
blackboxsw: +1 added warning and unit test

TheRealFalcon (approved):
Looks good to me! Thanks for getting this the rest of the way to the finish line.

PR #1213, opened 2022-01-26
Add _netdev option to mount Azure ephemeral disk

otubo:
## Proposed Commit Message

The ephemeral disk depends on a functional network to be mounted. Even
though it depends on cloud-init.service, sometimes an ordering cycle is
noticed on the instance. If the option "_netdev" is added the problem is
gone.

rhbz: #1998445

## Test Steps
Deploy a RHEL-9/cloud-init-21.1 instance on Azure and observe that sometimes there's an ordering cycle on the logs and there's no network connectivity.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -392,7 +392,8 @@ def handle(_name, cfg, cloud, log, _args):
     def_mnt_opts = "defaults,nobootwait"
     uses_systemd = cloud.distro.uses_systemd()
     if uses_systemd:
-        def_mnt_opts = "defaults,nofail,x-systemd.requires=cloud-init.service"
+        def_mnt_opts = ("defaults,nofail,x-systemd.requires=cloud-init.service,"
+                        ",_netdev")
 
     defvals = [None, None, "auto", def_mnt_opts, "0", "2"]
     defvals = cfg.get("mount_default_fields", defvals)
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -392,8 +392,8 @@ def handle(_name, cfg, cloud, log, _args):
     def_mnt_opts = "defaults,nobootwait"
     uses_systemd = cloud.distro.uses_systemd()
     if uses_systemd:
-        def_mnt_opts = ("defaults,nofail,x-systemd.requires=cloud-init.service,"
-                        ",_netdev")
+        def_mnt_opts = ("defaults,nofail,"
+                        "x-systemd.requires=cloud-init.service,_netdev")
 
     defvals = [None, None, "auto", def_mnt_opts, "0", "2"]
     defvals = cfg.get("mount_default_fields", defvals)
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -392,8 +392,9 @@ def handle(_name, cfg, cloud, log, _args):
     def_mnt_opts = "defaults,nobootwait"
     uses_systemd = cloud.distro.uses_systemd()
     if uses_systemd:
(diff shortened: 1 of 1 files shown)


otubo:
For some reason tox didn't complain about the line length on my local copy. Anyways, just updated. Thanks!

anhvoms:
@otubo Can you elaborate on why this is the case "The ephemeral disk depends on a functional network to be mounted"? Why is this an issue for RHEL 9 and not other distros (or earlier versions of RHEL)?

otubo:
@anhvoms you probably want to check the whole discussion, more specifically [David's comment on the original BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1998445#c17). TL;DR he says a recent update on systemd made it visible, but the problem might have been there all along.

cjp256:
> @anhvoms you probably want to check the whole discussion, more specifically [David's comment on the original BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1998445#c17). TL;DR he says a recent update on systemd made it visible, but the problem might have been there all along.

Anh is out out for a couple of days. I don't have access to that bug so I cannot read along, but the mount already has a dependency on cloud-init.service, which itself should have a dependency on network being online.  Has that changed for RHEL9?

RHEL 8.5:
```
# /usr/lib/systemd/system/cloud-init.service

[Unit]
Description=Initial cloud-init job (metadata service crawler)
Wants=cloud-init-local.service
Wants=sshd-keygen.service
Wants=sshd.service
After=cloud-init-local.service
After=NetworkManager.service network.service
After=NetworkManager-wait-online.service
...
```

Can you share the error you are seeing?  I'm also not quite sure why networking needs to be online to mount the disk...

Thanks!

cjp256:
To summarize from the BZ...

A change in systemd (https://github.com/systemd/systemd/commit/fa138f5e26742f83a762b2809a1891978c6aa04a) triggers an ordering cycle because local-fs.target will now set After=mnt.mount (and local-fs.target is required for cloud-init).  Previously, it would just be Wants=mnt.mount, avoiding the order cycling.

> systemd-248-7.el9: - No mnt.mount in "After"
> # systemctl show -p After local-fs.target
> After=systemd-remount-fs.service tmp.mount run-user-0.mount
> systemd-fsck-root.service boot-efi.mount local-fs-pre.target

> systemd-249-2.el9: - Have mnt.mount in "After"
> # systemctl show -p After local-fs.target
> After=systemd-fsck-root.service mnt.mount run-user-0.mount boot-efi.mount
> run-user-1000.mount local-fs-pre.target systemd-remount-fs.service tmp.mount
> -.mount

The argument made in the BZ is that due to a dependency on cloud-init.service, the ephemeral disk should not be considered local, but instead mark it as _netdev so that is not considered a local filesystem.

It seems to me that this grouping is slightly at odds with the x-systemd.requires configuration, but from what I gather reading the systemd.mount manpage, mounts fall into two groups: local-fs.target or remote-fs.target, depending whether the file system is local or remote.  If our mount cannot make the local-fs.target, it should declare _netdev to group in with remote-fs.target.

Another option would be to instead use nofail so that it is not required by either local or remote targets - but that would likely cause other issues :)

From my perspective given what I've seen, this seems a reasonable change (and not specific to Azure). 👍  Are there possible side-effects from grouping with remote-fs.target?

anhvoms:
@cjp256  thanks for the explanation. This looks reasonable to me

TheRealFalcon (approved):
Thanks for the explanation @cjp256 . Given that this module runs after networking has been applied, it looks like a reasonable change.

TheRealFalcon:
@otubo , looks like there's still a formatting issue. Can you run `tox -e format` to fix it?

otubo:
I hate style issues, sorry for the inconvenience guys. Should be good to go now. Thanks!

PR #1214, opened 2022-01-26
RFC - tox changes (SC-763)

holmanb:
```
fix parallel tox execution
```

Changes:
- tox -p fails due to a shared directory clobbering files - use unique env dirs to avoid this (this does cause a longer initial serial run due to deps getting installed multiple times)
- remove unneeded dependencies for black, flake8, and isort environments (saves some CI cycles by avoiding installing unused deps - since travis doesn't share a cache this should be a net gain)
- add a new environment (for convenience) just for "checking formatting" - this moves current "format" env to "do_format" and adds "check_format". I'm happy to adjust names if others would rather "format" doesn't change or think there is some better naming scheme.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ setenv =
 passenv=
     PYTEST_ADDOPTS
 
-[version]
+[format_deps]
 flake8==3.9.2
 pylint==2.11.1
 black==21.12b0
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -9,39 +9,51 @@ setenv =
...
+[testenv:check_format]
+envlist = flake8, pylint, black, isort
+deps =
+    flake8=={[version]flake8}
+    pylint=={[version]pylint}
+    black=={[version]black}
+    isort=={[version]isort}
+    -r{toxinidir}/test-requirements.txt
+    -r{toxinidir}/integration-requirements.txt
+commands =
```
TheRealFalcon, line 49: looks like copypasta of the `do_format` command. I think instead we want something like
```
commands =
    {[testenv:flake8]commands}
    {[testenv:pylint]commands}
    {[testenv:black]commands}
    {[testenv:isort]commands}
```

TheRealFalcon (changes requested):
Overall approach looks good. I left some comments inline.

TheRealFalcon (approved):
Parallelize all the things!

PR #1216, opened 2022-01-27
Update .github-cla-signers

violethaze74:
## Proposed Commit Message

```

```

## Checklist:

### DIFF
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -45,6 +45,7 @@ jshen28
 klausenbusk
 KsenijaS
 landon912
+lmd-jazzededtech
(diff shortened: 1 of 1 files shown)


blackboxsw (changes requested):
Hi @violethaze74 I can confirm that you have signed the CLA with this username and that you signed this CLA on behalf of the 2nd github account lmd-jazzededtech. For our own accountability can you drop lmd-jazzededtech from this PR and create a separate PR from the @lmd-jazzedtech  account which only includes lmd-jazzededtech. That way we can be sure that the author of each PR is the only person who adds the specific github usernames and commits.

PR #1220, opened 2022-01-28
Move LXD to end of datasource_list

TheRealFalcon:
The names in the changelog were added by `dch`. I haven't seen that before so maybe a new update to the tool?

https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1959118


blackboxsw (approved):
The alternative to this PR would be to leave ds-id detection ordering LXD in front of OpenStack untouched and grow awareness of OpenStack inside dscheck_LXD to return DS_NOT_FOUND if we also saw. "OpenStack Nova" or "OpenStack Compute" from dmi_product_name_matches.

We don't want to grow the above support for two reasons:
  - [nova-lxd project has been retired 3 years ago](https://git.openstack.org/openstack/nova-lxd), so adding ongoing support for ignoring LXD datasource within OpenStack is probably not something we want to add to ds-identify just to support older/End-of-life versions of OpenStack (stein).
  -  re-ordering priority of datasource detection is a low-risk mechanical change. Generally new datasources are added to the back of the list specifically to avoid conditions like this bug showing up, where the new datasource is being detected in front of old. Given that Bionic/Focal cloudimage for LXD also contain NoCloud  seed metadata templates so regression potential is limited as LXD was already ordered after NoCloud datasource.


Changeset is good here and will fix the specific issue where ds-id detected LXD before OpenStack on older OpenStack deployment.

We need to also do the following:
- change of LXD ordering to [ds-identifiy:DI_DSLIST_DEFAULT  in upstream/main](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L126) because ds-id will use that default order when no /etc/cloud/cloud.cfg/90_dpkg.cfg exists on a system (non-Ubuntu). We want the ordering to remain the same for both ds-id and packaging  postinst generated defaults.
- reflect this changeset to ubuntu/focal and ubuntu/impish and ubuntu/devel branches as well.

Note: unreltaed to this branch we should provide a commit to ubuntu/bionic to fix the `tox -e format` errors on the debian/apport* module

debian/changelog
```diff
@@ -1,5 +1,6 @@
 cloud-init (21.4-82-g0d364a29-0ubuntu1~18.04.1) UNRELEASED; urgency=medium
 
+  [ Brett Holman ]
```
blackboxsw, line 3: Since we don't really call out individual contributions in debian/changelog for Canonical contributors in cloud-init, let's redact the separate authorship sections here and just add your one liner to the top of the changelog. Also generally because d/changelog entries in cloud-init are so large at SRU time, it's been requested that our team consolidate any package related changes  at the top of the changelog. So make sure to pull your d/cloud-init.templates entry up to the top of the changelog section.

blackboxsw (changes requested):
Requested diff before we merge this.
```diff
diff --git a/debian/changelog b/debian/changelog
index 5e15b1b7f..43340cc83 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,6 +1,6 @@
 cloud-init (21.4-82-g0d364a29-0ubuntu1~18.04.1) UNRELEASED; urgency=medium
 
-  [ Brett Holman ]
+  * d/cloud-init.templates: Move LXD to back of datasource_list
   * refresh patches:
    + debian/patches/ec2-dont-apply-full-imds-network-config.patch
    + debian/patches/openstack-no-network-config.patch
@@ -115,9 +115,6 @@ cloud-init (21.4-82-g0d364a29-0ubuntu1~18.04.1) UNRELEASED; urgency=medium
     - testing: stop chef test from running on openstack (#1102)
     - alpine.py: add options to the apk upgrade command (#1089) [dermotbradley]
 
-  [ James Falcon ]
-  * d/cloud-init.templates: Move LXD to back of datasource_list
-
  -- Brett Holman <brett.holman@canonical.com>  Mon, 24 Jan 2022 09:05:03 -0700
 
 cloud-init (21.4-0ubuntu1~18.04.1) bionic; urgency=medium
```

blackboxsw (approved):
+1, LGTM!

PR #1222, opened 2022-01-28
sources/azure: refactor _report_ready_if_needed and _poll_imds

cjp256:
Refactor _report_ready_if_needed() to work for both Savable PPS
and Runnable PPS:

- rename _report_ready_if_needed() to _report_ready_for_pps()

- return interface name from lease to support _poll_imds() behavior
  without changing it.

- fixes an issue where reporting ready return value was silently
  ignored for Savable PPS.

- add explicit handling for failure to obtain DHCP lease to
  result in sources.InvalidMetaDataException.

Refactor _poll_imds():

- use _report_ready_for_pps() for reporting ready, removing this logic
  to simplify loop logic.

- move netlink and vnetswitch out of while loop to simplify loop logic,
  leaving only reprovision polling in loop.

- add explicit handling for failure to obtain DHCP lease and
  retry in the next iteration.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -895,8 +895,6 @@ class DataSourceAzure(sources.DataSource):
         """Report ready for PPS, creating the marker file upon completion.
 
         :raises sources.InvalidMetaDataException: On error reporting ready.
-
-        :return: Name of interface used to report ready.
         """
         report_ready_succeeded = self._report_ready(lease=lease)
         if not report_ready_succeeded:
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -890,36 +891,20 @@ def _create_report_ready_marker(self):
...
     @azure_ds_telemetry_reporter
-    def _report_ready_if_needed(self):
-        """Report ready to the platform if the marker file is not present,
-        and create the marker file.
+    def _report_ready_for_pps(self, lease: dict) -> None:
+        """Report ready for PPS, creating the marker file upon completion.
+
+        :raises sources.InvalidMetaDataException: On error reporting ready.
+
+        :return: Name of interface used to report ready.
```
anhvoms, line 899: this function doesn't return anything though?

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
cjp256: Ah yes, old approach. Removed

cjp256: thanks :D

TheRealFalcon (approved):
Thanks!

PR #1229, opened 2022-02-01
sources/azure: consolidate ephemeral DHCP configuration

cjp256:
Introduce:

-  _setup_ephemeral_networking() to bring up networking.
  If no iface is specified, it will use net.find_fallback_nic()
  which is consistent with the previous usage of fallback_interface.
  This method now tracks the encoded address of the wireserver
  with a new property `_wireserver_endpoint`.

- _teardown_ephemeral_networking() to bring down networking.

- _is_ephemeral_networking_up() to check status.

Ephemeral networking is now:

- Brought up prior to checking IMDS.

- Torn down following metadata crawl.

- For Savable PPS, torn down prior to waiting for NIC detach.
  The link must be torn down in advance or we will see errors
  from cleaning up network after the interface is unplugged.

- For Running PPS, torn down after waiting for media switch.
  The link must be up for media switch to be detected.

- For all PPS, after network switch is complete, networking is
  brought back up to poll for reprovision data and report ready.
  It will be torn down after metadata crawl is complete like
  non-PPS paths.

Additionally:

- Remove EphemeralDHCPv4WithReporting variant in favor of directly
using EphemeralDHCPv4.  The reporting was only for __enter__ usage
which is no longer a used path.  Continue to use dhcp_log_cb
callback.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -6,6 +6,7 @@
 
 import base64
 import crypt
+import datetime
 import os
 import os.path
 import re
@@ -24,15 +25,16 @@ from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
 from cloudinit.net import device_driver
-from cloudinit.net.dhcp import NoDHCPLeaseError
+from cloudinit.net.dhcp import EphemeralDHCPv4, NoDHCPLeaseError
 from cloudinit.reporting import events
 from cloudinit.sources.helpers import netlink
 from cloudinit.sources.helpers.azure import (
     DEFAULT_REPORT_FAILURE_USER_VISIBLE_MESSAGE,
-    EphemeralDHCPv4WithReporting,
+    DEFAULT_WIRESERVER_ENDPOINT,
     azure_ds_reporter,
     azure_ds_telemetry_reporter,
     build_minimal_ovf,
+    dhcp_log_cb,
     get_boot_telemetry,
     get_metadata_from_fabric,
     get_system_info,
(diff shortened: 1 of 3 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -303,14 +305,15 @@ def __init__(self, sys_cfg, distro, paths):
...
+        self._wireserver_endpoint = DEFAULT_WIRESERVER_ENDPOINT
         self.iso_dev = None
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         super()._unpickle(ci_pkl_version)
 
         self._ephemeral_dhcp_ctx = None
-        if not hasattr(self, "iso_dev"):
-            self.iso_dev = None
+        self.iso_dev = None
```
anhvoms: This means the iso won't be ejected in init-network for normal (non-preprovisioned) VM.

anhvoms: I don't think we can set it to None here, the provisioning iso won't be ejected in init phase for regular (non-preprovisioned) VMs

anhvoms (changes requested):

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
```diff
@@ -326,6 +329,66 @@ def _get_subplatform(self):
...
+        if self._ephemeral_dhcp_ctx is not None:
+            raise RuntimeError(
+                "Bringing up networking when already configured."
+            )
+
+        if iface is None:
+            iface = net.find_fallback_nic()
+
+        if iface is None:
+            report_diagnostic_event(
```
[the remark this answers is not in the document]
cjp256: Fixed

cloudinit/sources/DataSourceAzure.py
cjp256: Fixed

TheRealFalcon (approved):
LGTM. There are a couple pylint errors that need to be fixed though.

cjp256:
> LGTM. There are a couple pylint errors that need to be fixed though.

Fixed, I accidentally pushed the wrong ref earlier 😵‍💫

PR #1230, opened 2022-02-01
docs: include upstream python version info

TheRealFalcon:


CONTRIBUTING.rst
```diff
@@ -197,6 +197,15 @@ Design
 This section captures design decisions that are helpful to know when
 hacking on cloud-init.
 
+Python Support
+--------------
+Cloud-init upstream currently supports Python 3.6 and above.
```
blackboxsw, line 202: Let's include the upstream version of cloud-init v. 22.1 as the point at which the python 3.6 min version support is expected.

Is it worth us trying to preserve a stable-21.4 branch at the cutoff point where 3.5 was dropped for other downstream reference? We did this for our python-2.7 drop via upstream/stable-19.4. 
I think there is limited utility for these separate stable-\* branches, as I don't think we've seen any downstream PRs against stable-19.4 for instance so I don't know if there is much demand there.

If we did want to capture/doc the python version deprecatoin branches in docs or process, this is the commit where [upstream dropped 3.5 support I think is here](https://github.com/canonical/cloud-init/commit/dc1aabfca851e520693c05322f724bd102c76364).

GIven that other distributions might trail upstream significantly. Is it worth us having a table showing last upstream version which supports min python version supported?
| cloud-init version | min python version |
|------------------------|---------------------------|
| 22.1 | python 3.6+ |
| 20.3 | python 3.5+ |
| 19.4 | python 2.7+ |

blackboxsw (commented):
Nice. I would like to minimally capture what upstream version folks should expect to see python 3.6 min support in. Take that however you'd like to present the information.

CONTRIBUTING.rst
TheRealFalcon: > Is it worth us trying to preserve a stable-21.4 branch at the cutoff point where 3.5 was dropped for other downstream reference? We did this for our python-2.7 drop via upstream/stable-19.4.
I think there is limited utility for these separate stable-* branches, as I don't think we've seen any downstream PRs against stable-19.4 for instance so I don't know if there is much demand there.

I don't think we should do that. 2.7->3.x was a huge transition with a ton of breaking changes. Moving between 3.x versions should be much less impactful.

I can capture the table in the doc.

TheRealFalcon:
@blackboxsw , included the table in the latest commit

blackboxsw (approved):
LGTM

PR #1232, opened 2022-02-02
sources/azure: fix metadata check in _check_if_nic_is_primary()

cjp256:
Currently it checks for imds_md is None, but imds_md is returned as
an empty dictionary on error fetching metdata.

Fix this check and the tests that are incorrectly vetting IMDS
polling code.

Additionally, use response.contents() instead of str(response) when
loding the JSON data returned from readurl.  This helps simplify the
mocking and avoids an unncessary conversion.


anhvoms (approved):
LGTM

TheRealFalcon (approved):
Thanks!

PR #1233, opened 2022-02-02
Fix missing continue

eb3095:
## Proposed Commit Message
Fix missing continue

```
A continue was missing to allow interface seeking.
```

## Test Steps
Deploy an image with a private network on Vultr

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -38,6 +38,7 @@ def get_metadata(url, timeout, retries, sec_between, agent):
         except (NoDHCPLeaseError, subp.ProcessExecutionError) as exc:
             LOG.error("DHCP Exception: %s", exc)
             exception = exc
+            continue
 
         return json.loads(v1)
     raise exception
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -35,12 +35,11 @@ def get_metadata(url, timeout, retries, sec_between, agent):
 
                 # Fetch the metadata
                 v1 = read_metadata(url, timeout, retries, sec_between, agent)
+
+                return json.loads(v1)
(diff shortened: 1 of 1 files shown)


eb3095:
Sorry, this one is a bit embarrassing. Apparently a continue went missing somewhere along the lines of the last pull request.

TheRealFalcon:
@eb3095 , would it be better for the `return` to go inside the `try`? We can return the data once we have it, and if more exceptions pop up each one doesn't need to have it's own `continue`.

eb3095:
> @eb3095 , would it be better for the `return` to go inside the `try`? We can return the data once we have it, and if more exceptions pop up each one doesn't need to have it's own `continue`.

The idea is not to have it fail out on the first exception and only the last. We are seeing private networks and lo get pushed through that loop. We want every interface from 0 to last to be tried until none are left to try.

TheRealFalcon:
Sure, I'm just saying currently you have a structure like
```
for:
    try:
        ...
    except:
        ...
        continue
    return ...
```

Would it be better to replace with a structure like
```
for:
    try:
        ...
        return ...
    except:
        ...
```

Either way the return is still in the loop and `return`s when we find valid metadata and `continue`s when we don't.

eb3095:
> Sure, I'm just saying currently you have a structure like
> 
> ```
> for:
>     try:
>         ...
>     except:
>         ...
>         continue
>     return ...
> ```
> 
> Would it be better to replace with a structure like
> 
> ```
> for:
>     try:
>         ...
>         return ...
>     except:
>         ...
> ```
> 
> Either way the return is still in the loop and `return`s when we find valid metadata and `continue`s when we don't.

Ah, my mistake, yeah, that actually makes a lot of sense! Not sure how I missed that one either! Getting it fixed now.

TheRealFalcon (approved):
Thanks!

PR #1234, opened 2022-02-02
integration: do not LXD bind mount /etc/cloud/cloud.cfg.d

blackboxsw:
## Proposed Commit Message

```
integration: do not LXD bind mount /etc/cloud/cloud.cfg.d
    
Since lxc bind mounts will be read-only as nobody:nogroup
we don't want to bind mount /etc/cloud/cloud.cfg.d into the
instance because some tests add artifacts to /etc/cloud/cloud.cfg.d.

Also make LXD push_file pull_file methods assert that the
file transfer was a success, otherwise we miss the root-cause
for test failures.

This resulted in failed Jenkins runs in test_lxd_discovery with a
symptom of NoCloud being detected instead of LXD datasource.
The root-case was that instance.file_push failed due to permission
errors for root on the bind mounted /etc/cloud/cloud.cfg.d.

Also bump pycloudlib commitish to get Azure Jammy image support.
```

## Test Steps
```bash
cat >/tmp/pycloudlib-lxd.toml <<EOF
[lxd]
EOF
CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_OS_IMAGE=focal CLOUD_INIT_COLLECT_LOGS=ON_ERROR PYCLOUDLIB_CONFIG=/tmp/pycloudlib-lxd.toml CLOUD_INIT_LOCAL_LOG_PATH=/tmp/ci-logs CLOUD_INIT_CLOUD_INIT_SOURCE=IN_PLACE tox -e integration-tests tests/integration_tests/datasources/test_lxd_discovery.py
```

## Checklist:

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -295,7 +295,10 @@ class _LxdIntegrationCloud(IntegrationCloud):
             cloudinit_path, "..", "config", "cloud.cfg.d"
         )
         for src_file in os.listdir(config_dir):
-            command = f"lxc file push {config_dir}/{src_file} {instance.name}/etc/cloud/cloud.cfg.d/"
+            command = (
+                f"lxc file push {config_dir}/{src_file} "
+                f"{instance.name}/etc/cloud/cloud.cfg.d/"
+            )
             subp(command.split())
 
     def _perform_launch(self, launch_kwargs, **kwargs):
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -259,10 +259,6 @@ class _LxdIntegrationCloud(IntegrationCloud):
         mounts = [
             (cloudinit_path, "/usr/lib/python3/dist-packages/cloudinit"),
             (
-                os.path.join(cloudinit_path, "..", "config", "cloud.cfg.d"),
-                "/var/tmp/etc/cloud/cloud.cfg.d",
-            ),
-            (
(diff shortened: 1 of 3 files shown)


tests/integration_tests/clouds.py
```diff
@@ -274,7 +274,7 @@ def _mount_source(instance: LXDInstance):
             (cloudinit_path, "/usr/lib/python3/dist-packages/cloudinit"),
             (
                 os.path.join(cloudinit_path, "..", "config", "cloud.cfg.d"),
-                "/etc/cloud/cloud.cfg.d",
+                "/var/tmp/etc/cloud/cloud.cfg.d",
```
TheRealFalcon, line 277: If we're pushing these files directly, there's no reason to mount this directory anymore, correct?

TheRealFalcon (approved):
Good find! One question inline, but +1 since it shouldn't be a problem either way.

You also have a conflict you'll need to fix first because of my merge :grimacing:

tests/integration_tests/clouds.py
blackboxsw: d'oh, leftover attempt to avoid the nobody/nogroup bin mount. I was originally trying to bind mount then copy the full directory from the bind-mount over to /etc/cloud/cloud.cfg.d, but I realized that I should just copy the files in the first place without the bind-mount anyway. So I'm dropping that mount defitinitions as it's unused now.

PR #1237, opened 2022-02-03
testing: stop universally overwriting /etc/cloud/cloud.cfg.d

TheRealFalcon:
## Proposed Commit Message

```
testing: stop universally overwriting /etc/cloud/cloud.cfg.d

As part of IN_PLACE testing, /etc/cloud/cloud.cfg.d get overwritten by
what's in the source tree. This can cause problems when the directory is
mounted in, because tests need the ability to modify files in
/etc/cloud. Attempting to 'lxc file push'  instead will fail on LXD VMs
because the LXD agent isn't available yet.

If such functionality is desired, one can temporarily use the
'lxd_setup' mark while writing a test and push the files manually.
```

## Additional Context
#1234

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -282,21 +282,6 @@ class _LxdIntegrationCloud(IntegrationCloud):
             ).format(**format_variables)
             subp(command.split())
 
-        # /etc/cloud/cloud.cfg.d is a directory we write to in some
-        # integration tests. We can't use lxc mount in the container
-        # as /etc/cloud/cloud.cfg.d will then be owned nobody:nogroup and be
-        # read-only for root.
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
This is fine as we only risk exposure for not seeing unreleased changes to ./config/ in our jenkins git jobs which set CLOUD_INIT_SOURCE=IN_PLACE. This currently would affect the cc_keyboard tests failing since we changed that on Jan 22 and don't have an SRU yet that released those config changes.

holmanb:
LGTM. For what it's worth, a quick spot check shows that the overwritten files in a test instance currently match files deployed in a bionic image, so I wouldn't expect that to cause a behavior change in cloud-init.

PR #1238, opened 2022-02-03
Include "Container-Optimized OS" in distro list

holmanb:
```
Include "Container-Optimized OS" in distro list

Google provides an OS image called Container-Optimized OS that is
based on Chromium OS (which is itself based on Gentoo). Since this is an
OS on a major cloud provider that uses cloud-init [1] for configuration,
we should document it.

[1] https://cloud.google.com/container-optimized-os/docs/how-to/create\
-configure-instance
```

Notes:
I'm not sure how to best handle URL in commit message wrt the 50/72 rule (and how strictly we even follow it). Feel free to modify accordingly


TheRealFalcon:
Ehh...this is kind of stretching the definition of support to me. We have no code specific to the OS at all, and we don't do any kind of detecting of it. There are a number of OSes (and clouds) that happen to work fine with cloud-init, but (with a few exceptions), we don't usually call them out as being supported.

@blackboxsw , thoughts?

blackboxsw (commented):
While I appreciate the sentiment given those google docs, I'm seeing when I've launched a couple of "container-optimized OS"  images in GCE and that support really doesn't seem to be there.

The images I've launched seem to be gentoo-based, with broken symlink artifacts in the file-system that like to the portage packaging utility. In those cases cloud-init incorrectly detects the distro system incorrectly as ubuntu-based. 

```grep ubuntu /var/log/cloud-init.log 
2022-02-07 16:30:03,845 - stages.py[DEBUG]: Using distro class <class 'cloudinit.distros.ubuntu.Distro'>
```

So, if folks try to provide any significant user-data that tries to install package or manipulate anything ubuntu-specific, we are likely in a place where that container will hit significant failures.

To garner OS support in cloud-init, I think we probably need Container-Optimized OS references to properly represent what the Container-Optimized OS is in cloudinit/distros/cos.py and cloudinit/config/cc_*py modules should properly represent "support"  in the per-module `distros` list  for each of the cloud-init modules behavior. I'd expect in COS case that only a subset of cloud-init modules are declared as supported by COS.

I think we should reject this inclusion until we define a specific Distro subclass and vet which cc_*  config modules are actually supported by cloud-init without error.

blackboxsw (changes requested):
marking request-changes. But expect it likely that we/upstream are probably not going to prioritize this work currently.

holmanb:
@blackboxsw - Thanks for the thorough review. I don't plan on making these downstream fixes anytime soon, so will close this for now. We can always revisit in the future if this is prioritized or if downstream makes the required fixes and wants to upstream them.

PR #1243, opened 2022-02-03
Fix extra ipv6 issues, code reduction and simplification

eb3095:
## Proposed Commit Message
Fix extra ipv6 issues, code reduction and simplification

```
I eliminated the duplicate code and now run the entire configuration routine against both public and private interfaces. I also addressed an inconsistency from our metadata api for ipv6 address configuration.
```

## Test Steps
Start a Vultr vm with additional IPv6's added to the instance.

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -165,19 +165,18 @@ def generate_network_config(interfaces):
 
     # Prepare interface 0, public
     if len(interfaces) > 0:
-        public = generate_public_network_interface(interfaces[0])
+        public = generate_interface(interfaces[0], primary=True)
         network["config"].append(public)
 
     # Prepare additional interfaces, private
     for i in range(1, len(interfaces)):
-        private = generate_private_network_interface(interfaces[i])
+        private = generate_interface(interfaces[i])
         network["config"].append(private)
 
     return network
 
 
-# Input Metadata and generate public network config part
-def generate_public_network_interface(interface):
+def generate_interface(interface, primary=False):
     interface_name = get_interface_name(interface["mac"])
     if not interface_name:
         raise RuntimeError(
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
@eb3095 , changes look good.

Would you be able to add a couple unit tests for this behavior? I think the last PR with the missing continue/return shows we could use some extra unit test coverage for this datasource.

eb3095:
> @eb3095 , changes look good.
> 
> Would you be able to add a couple unit tests for this behavior? I think the last PR with the missing continue/return shows we could use some extra unit test coverage for this datasource.

Ha, fair, indeed it could use a few more tests I think. I'll add that.

eb3095:
Alright, I added tests for all the recent failures and this latest one.

cloudinit/sources/helpers/vultr.py
```diff
@@ -45,45 +45,8 @@ def get_metadata(url, timeout, retries, sec_between, agent):
...
-                "add",
-                "169.254.169.254/32",
-                "dev",
-                iface,
-            ]
-        )
-    elif subp.which("route"):
-        subp.subp(["route", "add", "-net", "169.254.169.254/32", "100.64.0.1"])
+    if subp.which("route"):
+        subp.subp(["route", "add", "169.254.169.254/32", iface])
```
TheRealFalcon, line 49: Was this removal part of the original PR? You don't need the code represented here anymore?

Is it worth checking that the route add succeeds.

eb3095: It was not, it came to light during testing this on some new hardware during this process. Turns out that gateway is not as consistent as I thought. I'm not even sure what the failure is here. There is only one OS that fails, Arch Linux, and this fix shouldn't even work (though it does). I test the DHCP against the metadata IP to avoid the potential of having a customer router on one of the other network interfaces. One would assume that the DHCP url check here would fail as that route is not added, but it doesn't. However when it proceeds to the next step to pull the metadata, it fails, as the route is not there. Hence why I'm adding this there. Its very strange behavior I can't make heads or tails of. But again, it only happens on Arch Linux.

I was originally using ip and using route as a fallback if ip wasnt available such as on BSD. Though we don't actually have Cloud-Init on BSD's yet so it's not actually been tested against that particular use case as of now.

If you have any insight on this particular issue, there is likely a better solution then what I am doing here, this feels super hacky honestly. I'm running a QA round on the new images with this version against various problematic configurations now to be sure this is going to hold. I'de be fine letting this sit until thats over. This was tested, just against the current ongoing problem servers.

eb3095: Ok, to touch base here. I have a commit stored for adding back the checks for linux and the route existing but im not convinced this needs to be solved in cloud-init for this problem as it only affects arch. The route falling off for one distro between the dhcp url check and the metadata pull feels super specific to something I'm missing in this particular distro. Im going through the cloud-init code trying to find the package im missing here. I mean this code "works" it shouldn't because you would think the subp adding an already existing route would fail with a non-zero exit code on the route command, but its working across my QA on multiple distro's/hw. This as is without the waiting commit will fail on *BSD due to the difference in the route command so it will need this waiting commit to resolve that. But again, I want to make sure that this is absolutely necessary before submitting another borked commit to you guys here. Have to apologize for that as we have a very large offering spanning a wide variety of baremetal and VM profile offerings that make this incredibly difficult to get right in a single go.

TheRealFalcon: Hmmm, sorry, no particular insight here. I agree that definitely seems strange though.

Thanks for the context. Multiple commits is no problem. Would you rather I merge this as-is, or wait until you've figured out what's happening?

eb3095: I want to take a harder look. I would really prefer to remove this then add to it. But it needs one more commit to readd the route check and linux check to prevent failures if there is no other solution.

eb3095: As I had thought, this didn't need fixed here. Turns out Arch Linux by default does not install the appropriate config file to /etc/dhclient.conf. After adding this file this issue went away. Code removed.

eb3095:
OK This should be good to go now!

TheRealFalcon (approved):
LGTM, thanks!

PR #1247, opened 2022-02-06
Add Oracle Linux 8 support

marthydavid:
Add Oracle Linux 8 support
This is similar alternative as AlmaLinux or Rocky for RHEL.

## Additional Context
[oracle.com/linux](https://www.oracle.com/linux/)

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # cloud-init
 
-[![Build Status](https://travis-ci.com/canonical/cloud-init.svg?branch=main)](https://travis-ci.com/canonical/cloud-init) [![Read the Docs](https://readthedocs.org/projects/cloudinit/badge/?version=latest&style=flat)](https://cloudinit.readthedocs.org)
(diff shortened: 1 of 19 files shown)


marthydavid:
@TheRealFalcon could you give me some advice why the tests are failing?

holmanb (changes requested):
I added a couple of comments, but this mostly looks good (besides the failing tests).

TheRealFalcon:
@marthydavid , your unit test is failing because we expect rhel derivatives to have a code name, and your unit test does not include the code name in the name of the OS release. The regex used to parse the release info can be seen [here](https://github.com/canonical/cloud-init/blob/main/cloudinit/util.py#L482). If Oracle Linux doesn't have code names, then you can add `ol` to the section below the regex that checks for a distro not having a code name.

Additionally, you also have a flake8 failure that will need fixing.

PR #1251, opened 2022-02-08
sources/hetzner: Use EphemeralDHCPv4 instead of static configuration

asciiprod:
## Proposed Commit Message

sources/hetzner: Use EphemeralDHCPv4 instead of static configuration

Signed-off-by: Markus Schade <markus.schade@hetzner.com>
```
When the datasource was originally submitted, EphemeralDHCPv4 was not
yet available. Also avoid race conditions by skipping network
configuration if metadata service can be reached.

Also update test examples to current IP addresses.
```

## Checklist:


cloudinit/sources/DataSourceHetzner.py
```diff
@@ -110,7 +116,14 @@ def network_config(self):
         migration.
         """
 
-        if self._network_config:
+        if self._network_config is None:
+            LOG.warning(
+                "Found None as cached _network_config. " "Resetting to %s",
```
cjp256, line 121: nit: string formatting

otherwise LGTM

PR #1252, opened 2022-02-08
tests: when generating crypted password, generate in target env

blackboxsw:
This resolves ` AssertionError: assert None == '$y$j9T$la3Xe...iM0xEOjQwtjxB'`  failures on our [bionic Jenkins test runner when exercising lxd_container jammy-daily jobs](https://jenkins.ubuntu.com/server/view/cloud-init/job/cloud-init-integration-lxd_container-jammy-daily-ppa/4/console)

## Proposed Commit Message

```
There are inconsistencies for cryptographic libraries across
major distribution releases.

From a bionic host, attempting run run crypt.crypt locally using
salt and format info from a Jammmy /etc/shadow file will result in
a failure to produce an encrpted password.

To avoid inconsistencies of python cryptographic libs across Linux
releases, perform the password encryption on the system under test.
```

## Test Steps
From a bionic host system
```PYCLOUDLIB_CONFIG=pycloudlib.toml CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_SOURCE=ppa:cloud-init-dev/daily CLOUD_INIT_OS_IMAGE=jammy  .tox/integration-tests-jenkins/bin/pytest tests/integration_tests/modules/test_set_password.py::TestPasswordList::test_explicit_password_set_correctly
```

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_set_password.py
+++ b/tests/integration_tests/modules/test_set_password.py
@@ -8,8 +8,6 @@ other tests chpasswd's list being a string.  Both expect the same results, so
 they use a mixin to share their test definitions, because we can (of course)
 only specify one user-data per instance.
 """
-import crypt
-
 import pytest
 import yaml
 
@@ -162,9 +160,13 @@ class Mixin:
         shadow_users, _ = self._fetch_and_parse_etc_shadow(class_client)
 
         fmt_and_salt = shadow_users["tom"].rsplit("$", 1)[0]
-        expected_value = crypt.crypt("mypassword123!", fmt_and_salt)
-
-        assert expected_value == shadow_users["tom"]
+        GEN_CRYPT_CONTENT = (
+            "import crypt\n"
+            f"print(crypt.crypt('mypassword123!', '{fmt_and_salt}'))\n"
+        )
+        class_client.write_to_file("/gen_crypt.py", GEN_CRYPT_CONTENT)
+        result = class_client.execute("python3 /gen_crypt.py")
+        assert result.stdout == shadow_users["tom"]
 
     def test_shadow_expected_users(self, class_client):
         """Test that the right set of users is in /etc/shadow."""


TheRealFalcon (approved):
LGTM, but flake8 failure due to crypt import no longer needed.

Suggested condensed first line of commit:
`tests: generate crypted password in target env`

Feel free to merge after fixing the flake8.

holmanb:
Something doesn't seem right about this resolution to the problem.


Even though the user is added via  `useradd tom --password "$1$S7$tT1BEDIYrczeryDQJfdPe0" -m` (from `cloudinit/distros/__init__.py`, the hash read from `/etc/shadow` is `$y$j9T$dwk9jSfj/HlK68D989MAa1$tkF6ZjRJWXG9OZ2dS2SzCQCL9nuAogxSd9nRjHC6De6`. This shouldn't happen since we're passing a hash.

[`crypt`'s default password may have changed to yescrypt](https://manpages.debian.org/unstable/libcrypt-dev/crypt.5.en.html#yescrypt) between distros (hence the new`$y` format), but that shouldn't affect this since we're passing in a hash. We shouldn't be depending on local crypt settings with this.

What am I missing here?

I don't understand why this fixes the odd behavior I just mentioned.

holmanb (changes requested):
See previous comments

blackboxsw:
Ok  @TheRealFalcon and I debugged this further. A couple things to note:
1. Per your comment
>>  Even though the user is added via useradd tom --password "$1$S7$tT1BEDIYrczeryDQJfdPe0" -m (from cloudinit/distros/__init__.py, the hash read from /etc/shadow is $y$j9T$dwk9jSfj/HlK68D989MAa1$tkF6ZjRJWXG9OZ2dS2SzCQCL9nuAogxSd9nRjHC6De6. This shouldn't happen since we're passing a hash.

The userdata provided to the instance actually provides additional passwd via chpasswd which effectively overwrites the values provided in "users:" cloud-config. 

2. invoking crypt.crypt from python on jammy reads '/lib/x86_64-linux-gnu/libcrypt.so.1` delivered by libcrypt1:amd64 which is build from source pkg: libxcrypt.  There are huge changes for yescrypt support and handling between bionic (where our integration tests run) and Jammy (or test target that is failing). We can walk through some of those changes using git-ubuntu
```bash
git-ubuntu clone libxcrypt
cd libxcrypt
# All ubuntu package source uploads are represented under git branches pkg/ubuntu/(bionic|focal|jammy)-(proposed|devel|updates)?
# the defaut clone drops you into pkg/ubuntu/devel (which is current the ubuntu jammy release of libxcrypt)
git diff pkg/ubuntu/bionic-devel
```
The resulting diff shows us that bionic had no yescrypt support, so running python3 crypt.crypt locally on our bionic system will return None for salt/hash/formats it doesn't understand. The solution to validate hashed passwd via crypt on the target environment will avoid this type of failure.

holmanb:
> Ok @TheRealFalcon and I debugged this further. A couple things to note:
> 
>     1. Per your comment
> 
> 
> > > Even though the user is added via useradd tom --password "$1$S7$tT1BEDIYrczeryDQJfdPe0" -m (from cloudinit/distros/**init**.py, the hash read from /etc/shadow is $y$j9T$dwk9jSfj/HlK68D989MAa1$tkF6ZjRJWXG9OZ2dS2SzCQCL9nuAogxSd9nRjHC6De6. This shouldn't happen since we're passing a hash.
> 
> The userdata provided to the instance actually provides additional passwd via chpasswd which effectively overwrites the values provided in "users:" cloud-config.


Aha! This is the piece I was missing. Thanks for the explanation. I agree that passing through crypt behavior rather than assuming a specific version of crypt is the way to go. +1 on the approach

holmanb (approved):
In addition to this change I think we need to add yescrypt support (parsing `$y` in addition to `$1`/`$5`/`$6`/etc) with changes in [the docs](https://github.com/canonical/cloud-init/blob/bae9b11da9ed7dd0b16fe5adeaf4774b7cc628cf/cloudinit/config/cc_set_passwords.py#L32) and in the [parsing code](https://github.com/canonical/cloud-init/blob/bae9b11da9ed7dd0b16fe5adeaf4774b7cc628cf/cloudinit/config/cc_set_passwords.py#L168). 


While we're at it we might want to add support for the other recently added [hashing methods](https://manpages.debian.org/bullseye/libcrypt-dev/crypt.5.en.html) supported by `crypt` too (`$2b`/`$gy`/`$7`).


That doesn't need to be in this PR, however.

PR #1253, opened 2022-02-09
use PEP 589 syntax for TypeDict

holmanb:
```
Use PEP 589 syntax for TypeDict annotation.

Also fixes previously broken typing MetaSchema typing implementation.
```


cloudinit/importer.py
```diff
@@ -12,21 +12,19 @@
...
-            "title": str,
-            "description": str,
-            "distros": typing.List[str],
-            "examples": typing.List[str],
-            "frequency": str,
-        },
-    )
+if sys.version_info >= (3, 8):
+
+    class MetaSchema:
```
TheRealFalcon, line 19: This needs to inherit from TypedDict.

TheRealFalcon (commented):
The old version had 
```python
if sys.version_info >= (3, 8) and hasattr(typing, "TypeDict"):
```
That's `TypeDict` rather than `TypedDict`, so this never actually worked.

In order for this proposed version to work, we also need to add this type to every `meta` definition because apparently you can't coerce a raw dict into a TypedDict. E.g., on `cc_apk_configure`, I currently see:
image

Can we update these meta definitions in this PR too?

holmanb:
@TheRealFalcon I rebased on the new mypy commit and it looks like the fix you suggested allows us to remove a couple of files from the "exclude" list. This is ready for re-review.

TheRealFalcon (approved):
Nice! LGTM.

PR #1265, opened 2022-02-14
sources/azure: report ready in local phase

cjp256:
Pre-provisioned instances report ready early in the local phase and
again in the non-local phase, during setup().  Non-PPS only reports
ready during non-local phase.

Update the process to report ready during the local phase for all
cases.  Only attempt to do so if networking is up to prevent stalling
boot. We've already waited at least 20 minutes for DHCP if we're
provisioning, or 5 minutes for DHCP on normal boot requesting updated
network configuration.

- Extend _report_ready() with pubkey_info and raise exception
  on error to consolidate reporting done in _negotiate() and
  _report_ready().

- Remove setup(), moving relevant logic into crawl_metadata().

- Move remaining _negotiate() logic into _cleanup_markers() and
  _determine_wireserver_pubkey_info().

These changes effectively fix two issues that were present:

(1) _negotiated is incorrectly set to True

When failing to report ready.  _negotiate() squashed the exception and
the return value was not checked.  This was probably masked due to the
forced removal of obj.pkl on Ubuntu instances, but would be preferable
once we start persisting it to prevent unnecessary re-negotiation.

(2) provisioning media is not ejected for non-PPS

_negotiate() did not pass iso_dev parameter when reporting ready.  The
host will ensure this operation takes place, but it is preferable to
eject /dev/sr0 from within the guest when we're done with it.

Lastly, this removes any need for lease file parsing as the wireserver
addressed is tracked for ephemeral DHCP.  A follow-up PR will remove
this now-unused logic.


cjp256:
@TheRealFalcon any chance can we get this one in prior to release?

TheRealFalcon:
I'll start reviewing it.

TheRealFalcon (approved):

cloudinit/sources/DataSourceAzure.py
```diff
@@ -317,17 +317,16 @@ def __init__(self, sys_cfg, distro, paths):
...
         self._wireserver_endpoint = DEFAULT_WIRESERVER_ENDPOINT
-        self.iso_dev = None
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         super()._unpickle(ci_pkl_version)
 
         self._ephemeral_dhcp_ctx = None
-        if not hasattr(self, "iso_dev"):
-            self.iso_dev = None
+        self._iso_dev = None
```
blackboxsw, line 329: @TheRealFalcon @cjp256 Hrm, if my cached obj.pkl pickle already has self._iso_dev set to something other than none, this unpickle will clear it's value. Does this break us?

cjp256: It would not break anything.  We don't want to persist it anymore :)

PR #1266, opened 2022-02-14
docs: Add more details to runcmd docs

TheRealFalcon:
Multiple times I have seen confusion over runcmd and how to know what it's running. It'd help to know that runcmd isn't actually executing the script and what module is.

### DIFF
--- a/cloudinit/config/cc_runcmd.py
+++ b/cloudinit/config/cc_runcmd.py
@@ -37,7 +37,8 @@ meta = {
         ``sh``.
 
         Note that the ``runcmd`` module only writes the script to be run
-        later. The module that actually runs the script is ``scripts-user``.
+        later. The module that actually runs the script is ``scripts-user``
+        in the ref:`Final` boot stage.
 
         .. note::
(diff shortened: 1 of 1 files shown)


cloudinit/config/cc_runcmd.py
```diff
@@ -32,10 +32,12 @@
...
-        list, it will be properly executed as if passed to ``execve()`` (with
-        the first arg as the command). If the item is a string, it will be
-        written to a file and interpreted
-        using ``sh``.
+        list, it will be properly quoted. Each item is written to
+        ``/var/lib/cloud/instance/runcmd`` to be later interpreted using
+        ``sh``.
+
+        Note that the ``runcmd`` module only writes the script to be run
+        later. The module that actually runs the script is ``scripts-user``.
```
blackboxsw, line 40: Worth saying that this is run the the final boot stage?
```suggestion
        later. The module that actually runs the script is ``scripts-user``
        in the ref:`Final` boot stage.
```

PR #1270, opened 2022-02-16
Release to ubuntu/devel

holmanb:
```
    releasing cloud-init version 21.4-130-g8537237d-0ubuntu1~22.04.1
```
Assuming this is correct, I'll need someone with commit permissions to push a tag with:
```
git tag ubuntu/21.4-130-g8537237d-0ubuntu1_22.04.1
git push <remote> ubuntu/21.4-130-g8537237d-0ubuntu1_22.04.1
```


holmanb:
That failure is... interesting

blackboxsw (changes requested):
debian/changelog here has an SRU bug number and shouldn't be redacting individual bugs fixed in the changelog. This points to a problem in new-upstream-snapshot in detecting a devel release appropriately. 

A note that devel releases, as Jammy currently is, don't require an SRU process bug because the release is not considered officially stable yet.  Since we aren't publishing to a stable release in this PR we expect two things:
 - no need for an StableReleaseUpdate process bug
 - each individual bug that is fixed needs to be documented in the debian/changelog entries so that launchpad automation closes these bugs as Fix Released.

I've put up a PR that should correct that script to give us the expected upload for a devel release since Jammy. https://github.com/canonical/uss-tableflip/pull/85

blackboxsw (commented):
Also something is funky in this upload as you have 217 commits listed vs my branch has only 18 commits since last release. It makes me think you are using a stale ubuntu/devel checkout. Please ensure to `git fetch upstream; git checkout -B upstream/ubuntu/devel ubuntu/devel` before running new-upstream-snapshot

holmanb:
> Also something is funky in this upload as you have 217 commits listed vs my branch has only 18 commits since last release. It makes me think you are using a stale ubuntu/devel checkout. Please ensure to `git fetch upstream; git checkout -B upstream/ubuntu/devel ubuntu/devel` before running new-upstream-snapshot

I just noticed this PR is against ubuntu/daily/devel, not ubuntu/devel. I think this is the reason.

blackboxsw:
> I just noticed this PR is against ubuntu/daily/devel, not ubuntu/devel. I think this is the reason.

Ahh that's it. I don't know if you can retarget the PR or if you have to re-open a PR against the right target.

holmanb:
> > I just noticed this PR is against ubuntu/daily/devel, not ubuntu/devel. I think this is the reason.
> 
> Ahh that's it. I don't know if you can retarget the PR or if you have to re-open a PR against the right target.

Turns out you can (clicking "edit" turns the base branch into a dropdown menu).

blackboxsw:
> Turns out you can (clicking "edit" turns the base branch into a dropdown menu).

Thanks for the TIL. I had been closing and re-opening when I defined the wrong merge target.

blackboxsw (approved):
- [x] git diff looks good against my local copy, only authorship differences
- [x] build-package and sbuild success no new lintian warnings
- [x] quilt push/pop unnecessary as there are no quilt patches to be wary of

The isort/black CI failures can be addressed with a separate commit to this branch after this release and included in the next release just so CI passes without false/concerning alerts.

PR #1272, opened 2022-02-16
sources/azure: fix and consolidate imds retry logic

cjp256:
Extend retry_on_url_exc with some configurable knobs for error
codes and exception types.  Replace all custom callbacks with
this, replacing the log backoff logic with retry back-off, where
errors are logged between attempts.

Retry for codes:
- code 404 = not found (yet)
- code 410 = gone / unavailable (yet)
- code 429 = rate-limited/throttled
- code 500 = server error

Never retry for:
- code 400 = bad request (unsupported API version or malformed request)

In addition to the consolidation, this should address three issues:

1. get_imds_data_with_api_fallback() will attempt one request with
IMDS_VER_WANT.  If the connection fails due to a timeout or other
connection issue, an empty dictionary will be returned without
attempting the requested number of retries.

2. IMDS_VER_WANT will never be attempted if retries=0, such as when
fetching network metadata with infinite=True.

3. Only 404 error codes were retried in some cases.  Retry for all
above codes.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -12,6 +12,7 @@ import os.path
 import re
 import xml.etree.ElementTree as ET
 from enum import Enum
+from functools import partial
 from time import sleep, time
 from typing import Any, Dict, List, Optional
 from xml.dom import minidom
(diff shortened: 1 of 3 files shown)


cjp256:
Just for demonstration purposes atm, will revisit in the future.

PR #1273, opened 2022-02-16
Add json schemas for modules starting with P

TheRealFalcon:
## Proposed Commit Message

```
Add json schemas for modules starting with P

Includes:
* cc_package_update_upgrade_install
* cc_phone_home
* cc_power_state_change
* cc_puppet
```

## Additional Context
This was supposed to be P-RMC, but the two RMC related modules have no public documentation or configuration.

I also removed files from doc/examples that were a single example. It serves as a secondary, less visible spot for docs and examples, when the examples could just as easily go into the documentation.


cloudinit/config/cloud-init-schema.json
```diff
@@ -543,6 +543,244 @@
...
+            "apt_upgrade": {
+                "type": "boolean",
+                "default": false,
+                "description": "DEPRECATED. Alias for ``package_upgrade``. Default: ``false``"
+            },
+            "apt_reboot_if_required": {
+                "type": "boolean",
+                "default": false,
+                "description": "DEPRECATED. Alias for ``package_reboot_if_required``. Default: ``false``"
+            }
```
TheRealFalcon, line 590: Curious if we should just remove these from the schema altogether as a means of deprecation. Since we don't strictly enforce schemas when cloud-init runs, they'll still work but we'd see a schema validation error in the logs.

holmanb: It might be more user friendly to have a separate message for users when a config is invalid vs when they are using a deprecated key. I'd lean towards leaving it how you have it for now.

Related: a quick test with your code shows that removing these deprecated keys from the jsonschema doesn't cause warnings with `cloud-init devel schema --config-file file.yml`. I think we want errors with the cli command as well as with the cloud-init runtime validation warning, so I assume that will take some more work.

My test:
```
root@my-host:~# cloud-init devel schema --config-file tmp.yml 
Valid cloud-config: tmp.yml
root@my-host:~# cat tmp.yml 
#cloud-config
packages:
  - pwgen
  - pastebinit
  - [libpython3.8, 3.8.10-0ubuntu1~20.04.2]
apt_update: true
apt_upgrade: true
apt_reboot_if_required: true

# demonstrating that the keys do not exist in jsonschema
root@my-host:~# grep -e apt_update -e apt_upgrade -e apt_reboot_if_required /usr/lib/python3/dist-packages/cloudinit/config/cloud-init-schema.json

# demonstrating that this instance is using the new schema definition
root@my-host:~# grep -e apt_update -e apt_upgrade -e apt_reboot_if_required -e cc_package_update_upgrade_install /usr/lib/python3/dist-packages/cloudinit/config/cloud-init-schema.json
    "cc_package_update_upgrade_install": {
    { "$ref": "#/$defs/cc_package_update_upgrade_install" },

```

cloudinit/config/cloud-init-schema.json
```diff
@@ -543,6 +543,244 @@
...
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["mode"],
+          "properties": {
+            "delay": {
+              "description": "Time in minutes to delay after cloud-init has finished. Can be ``now`` or an integer specifying the number of minutes to delay. Default: ``now``",
+              "default": "now",
+              "oneOf": [
+                {"type": "integer"},
+                {"type": "string", "pattern": "^\\+*[0-9]+$"},
```
holmanb, line 648: What is the purpose of this regex and string type? Do we have documented usage of this as a string anywhere?

cloudinit/config/cloud-init-schema.json
```diff
@@ -543,6 +543,244 @@
...
+    },
+    "cc_phone_home": {
+      "type": "object",
+      "properties": {
+        "phone_home": {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["url"],
+          "properties": {
+            "url": {
```
holmanb, line 601: Since jsonschema has had a [builtin uri format going back to Draft 4](https://json-schema.org/understanding-json-schema/reference/string.html#format), this might be a good location to see if it can be used for validation of the URL here? I'm also not sure if it's worth the effort.

cloudinit/config/cloud-init-schema.json
TheRealFalcon: I want to "quietly" deprecate this syntax by just removing reference to it in the docs. If you look at [current docs](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#power-state-change), it's a +number format because that's traditional `shutdown` command syntax. We can still support the syntax, but to me it makes more sense to just provide an integer. 

Good point that it should still be documented somewhere, but json files don't support comments, so not sure the best place to put that.

cloudinit/config/cloud-init-schema.json
TheRealFalcon: > Related: a quick test with your code shows that removing these deprecated keys from the jsonschema doesn't cause warnings with cloud-init devel schema --config-file file.yml. I think we want errors with the cli command as well as with the cloud-init runtime validation warning, so I assume that will take some more work.

This is a more general comment that would apply to all schemas and not something to fix here, correct? Just making sure I understand correctly.

cloudinit/config/cc_power_state_change.py
```diff
@@ -4,67 +4,80 @@
...
 EXIT_FAIL = 254
 
+MODULE_DESCRIPTION = """\
+This module handles shutdown/reboot after all config modules have been run. By
+default it will take no action, and the system will keep running unless a
+package installation/upgrade requires a system reboot (e.g. installing a new
+kernel) and ``package_reboot_if_required`` is true.
+
+Using this module ensures that cloud-init is entirely finished with
+modules that would be executed, and avoids any error/log messages
```
holmanb, line 32: >... and avoids any error/log messages
that may go to the console as a result of system services like
syslog being taken down while cloud-init is running.

It sounds like adding this module solved a problem previously existed in cloud-init. This information seems valuable as a breadcrumb for future developers, but I doubt it is valuable / actionable for end users. Think we still need it?

cloudinit/config/cloud-init-schema.json
holmanb: > > Related: a quick test with your code shows that removing these deprecated keys from the jsonschema doesn't cause warnings with cloud-init devel schema --config-file file.yml. I think we want errors with the cli command as well as with the cloud-init runtime validation warning, so I assume that will take some more work.
> 
> This is a more general comment that would apply to all schemas and not something to fix here, correct? Just making sure I understand correctly.

Correct, I don't think this is a local problem. I just happened to be using your PR to see what kind of error messages I would get and was surprised I didn't see any. We can do that in a separate PR.

cloudinit/config/cloud-init-schema.json
holmanb: > I want to "quietly" deprecate this syntax by just removing reference to it in the docs. If you look at [current docs](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#power-state-change), it's a +number format because that's traditional `shutdown` command syntax. We can still support the syntax, but to me it makes more sense to just provide an integer.

Makes sense, thanks. I think the regex might be incorrect: `\\` means an escaped `\`, but I think you want to escape the `+`, correct? Also we shouldn't need the `*` since multiple plus signs (`++2`) isn't a valid notation.  Based on that, I think that `^\+[0-9]+$` is probably what we want. I haven't tested this with jsonschema specifically, but I have locally.

tests/unittests/config/test_cc_power_state_change.py
```diff
@@ -156,4 +163,29 @@ def check_lps_ret(psc_return, mode=None):
...
+            # Invalid mode
+            (
+                {"power_state": {"mode": "test"}},
+                r"'test' is not one of \['poweroff', 'reboot', 'halt'\]",
+            ),
+            # Delay can be a number, a +number, or "now"
+            ({"power_state": {"mode": "halt", "delay": "5"}}, None),
+            ({"power_state": {"mode": "halt", "delay": "now"}}, None),
+            ({"power_state": {"mode": "halt", "delay": "+5"}}, None),
+            ({"power_state": {"mode": "halt", "delay": "-5"}}, ""),
```
holmanb, line 179: This test should demonstrate why the pattern should use `{0,1}` instead of `*`.
```
           ({"power_state": {"mode": "halt", "delay": "++5"}}, ""),
```

doc/examples/cloud-config-puppet.txt
```diff
@@ -1,93 +0,0 @@
-#cloud-config
```
holmanb, line 1: Why remove the examples?

TheRealFalcon: I mentioned it in the PR description:
```I also removed files from doc/examples that were a single example. It serves as a secondary, less visible spot for docs and examples, when the examples could just as easily go into the documentation.```

I've noticed a lot of modules have an example, then there's an example file with almost the exact same example with comments that more of less serve as documentation. Seems it'd be best to have that all in one place, and save example files for times when we have multiple different examples for multiple different use cases.

**Review by blackboxsw — CHANGES_REQUESTED** [review]
A couple of suggestions on the deprecation expression in our schema definition. Doc rendering looks good. I'm walking through the examples and constraints this afternoon and will have a couple of additional comments. Thanks!

File: cloudinit/config/cloud-init-schema.json (hunk 1 above)
Comment (blackboxsw) [reply cloudinit/config/cloud-init-schema.json:590]: I agree that feature for rendering deprecation warnings is desirable for this current epic of work, but not for this specific PR.  I added this [work item](https://warthogs.atlassian.net/browse/SC-833)

File: cloudinit/config/cloud-init-schema.json [hunk 7]
```diff
@@ -543,6 +543,244 @@
...
+        "power_state": {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["mode"],
+          "properties": {
+            "delay": {
+              "description": "Time in minutes to delay after cloud-init has finished. Can be ``now`` or an integer specifying the number of minutes to delay. Default: ``now``",
+              "default": "now",
+              "oneOf": [
+                {"type": "integer"},
```
blackboxsw, line 647: Let's put a "minimum": 0 here too
```suggestion
                {"type": "integer", "minimum": 0},
```

cloudinit/config/cloud-init-schema.json
blackboxsw: I'd agree that we probably want to deprecate that "+\d" value from schema at some point to create concise/simple scheme for expected values. If we want to deprecate it fully, we can start logging warnings for uses of the "+\d" value in another PR when we handle proper deprecation notices.  Maybe we can add a checklist item for deprecation warnings for "delay" to SC-774

tests/unittests/config/test_cc_power_state_change.py
```diff
@@ -156,4 +163,29 @@ def check_lps_ret(psc_return, mode=None):
...
+    @pytest.mark.parametrize(
+        "config, error_msg",
+        [
+            # Invalid mode
+            (
+                {"power_state": {"mode": "test"}},
+                r"'test' is not one of \['poweroff', 'reboot', 'halt'\]",
+            ),
+            # Delay can be a number, a +number, or "now"
+            ({"power_state": {"mode": "halt", "delay": "5"}}, None),
```
blackboxsw, line 176: It's probably worth a couple of tests for the various types "condition" can be.

cloudinit/config/cloud-init-schema.json
```diff
@@ -543,6 +543,244 @@
...
+              "description": "Time in seconds to wait for the cloud-init process to finish before executing shutdown. Default: ``30``",
+              "type": "integer",
+              "default": 30
+            },
+            "condition": {
+              "description": "Apply state change only if condition is met. May be boolean true (always met), false (never met), or a command string or list to be executed. For command formatting, see the documentation for ``cc_runcmd``. If exit code is 0, condition is met, otherwise not. Default: ``true``",
+              "default": true,
+              "oneOf": [
+                {"type": "string"},
+                {"type": "boolean"}
```
blackboxsw, line 671: I think we also need a "type": "array"

cloudinit/config/cloud-init-schema.json
```diff
@@ -543,6 +543,244 @@
...
+                  "type": "object"
+                },
+                "agent": {
+                  "type": "object"
+                },
+                "user": {
+                  "type": "object"
+                },
+                "ca_cert": {
+                  "type": "string"
```
blackboxsw, line 764: Where are we grabbing these top-level section definitions? I agree we don't want to boil the ocean and define schema for al params that puppet provides, but how do we know that we aren't missing another config section. Docs I've seen look sparse on this as I couldn't find a [ca_cert] section

cloudinit/config/cloud-init-schema.json
TheRealFalcon: > Since your tests pass, I guess that the \\ should be fine - this appears consistent with re.search()'s behavior and is a side effect of python str handling (though it wouldn't match as a PCRE, I don't think).

Looks like backslash is an escape character in json (e.g., you can "\n" in json), so the double backslashes should be correct. Agree with the rest of the regex, though I think `?` should work instead of `{0,1}`, correct?

holmanb: > Looks like backslash is an escape character in json (e.g., you can "\n" in json), so the double backslashes should be correct.

Today I learned, thanks for digging into this :)

> Agree with the rest of the regex, though I think `?` should work instead of `{0,1}`, correct?

Agreed

tests/unittests/config/test_cc_package_update_upgrade_install.py
```diff
@@ -0,0 +1,24 @@
...
+class TestPackageUpdateUpgradeSchema:
+    @pytest.mark.parametrize(
+        "config, error_msg",
+        [
+            # packages list with single entry (2 required)
+            ({"packages": ["p1", ["p2"]]}, ""),
+            # packages list with three entries (2 required)
+            ({"packages": ["p1", ["p2", "p3", "p4"]]}, ""),
+            # empty packages list
+            ({"packages": []}, "is too short"),
```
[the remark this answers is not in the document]
TheRealFalcon: Yes, I figured for a single simple data types adding explicit testing seemed a bit overkill, and we have examples for testing the happy path. I'm fine adding more if you disagree though.

cloudinit/config/cloud-init-schema.json
TheRealFalcon: I'm taking this from the existing documentation:
"section names should be one of: ``main``, ``server``, ``agent`` or
``user`` and keys should be valid puppet configuration options."
https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_puppet.py#L46-L56

https://puppet.com/docs/puppet/7/config_file_main.html#config_file_main-puppet-conf-config-sections also lists 4 config sections.

`ca_cert` I'm stealing directly from the code as current documentation looks to be wrong:
https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_puppet.py#L282

TheRealFalcon:
I believe I have addressed all comments

tests/unittests/config/test_cc_package_update_upgrade_install.py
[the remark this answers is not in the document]
blackboxsw: All clear from my side, it's painful to add all these specific tests for things that should be validated at review time (and by examples testing)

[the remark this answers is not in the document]
blackboxsw: I think it's only worth us adding type asserts in complex anyOf cases where each type isn't already covered by an example in the schema which is covered by TestCloudConfigExamples

tests/unittests/config/test_cc_package_update_upgrade_install.py
```diff
@@ -0,0 +1,24 @@
...
+        ],
+    )
+    def test_schema_validation(self, config, error_msg):
```
blackboxsw, line 22: I think on any JSON schema validation tests we may still need the @skipUnlessJsonSchema decorator to allow test environments that may not have json schema dependencies installed.
```suggestion
    @skipUnlessJsonSchema()
    def test_schema_validation(self, config, error_msg):
```

blackboxsw (commented):
Changes LGTM! only thing to note I think is json schema unit tests may still needs the skipUnlessJsonSchema decorator.

cloudinit/config/cloud-init-schema.json
blackboxsw: That's good enough for me!

blackboxsw (approved):
Let's land it

PR #1276, opened 2022-02-17
net/dhcp: surface type of DHCP lease failure to caller

cjp256:
When performing DHCP, it is useful for the caller to have context
on the type of failure.  This can be done with some new exceptions
types, subclassing NoDHCPLeaseError so the caller's current
contract remains.

- Add the following errors:

  - NoDHCPLeaseInterfaceError if there are problems finding
  the (possibly specified) interface.

  - NoDHCPLeaseMissingDhclientError for missing dhclient.

- Update InvalidDHCPLeaseFileError to subclass NoDHCPLeaseError.

- Pass through these errors rather than catching it in obtain_lease().

Tests:

- Add missing mock for test_provided_nic_does_not_exist().

- Add new test coverage for EphemeralDHCPv4 errors.

- Update existing tests for maybe_perform_dhcp_discovery() to match
  new behavior.

### DIFF
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -28,7 +28,11 @@ LOG = logging.getLogger(__name__)
 NETWORKD_LEASES_DIR = "/run/systemd/netif/leases"
 
 
-class InvalidDHCPLeaseFileError(Exception):
+class NoDHCPLeaseError(Exception):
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM!

TheRealFalcon (approved):
Thanks!

PR #1284, opened 2022-02-21
Ubuntu/impish SRU

holmanb:
## Proposed Commit Message

```

```

## Checklist:


blackboxsw (changes requested):
Thanks @holmanb . This PR has a few extra commits beyond what we already released to ubuntu/devel(Jammy). For SRUs, Ubuntu policy and our cloud-init SRU exception doesn't let us typically upload "newer" content into stable releases than what was released to the later series unless there are specific bug fixes for the target. 

So, we'll need to make sure we only new-upstream-snapshot the same upstream commitish (edit: commitish from upstream/main) that we released to ubuntu/devel

This should suffice
```
new-upstream-snapshot b3d9acddd352f2c787ca16e132aa741dac00baef
```

blackboxsw (approved):
- [x] only authorship diffs changed and whitespace/formatting fix
- [x] no new lintain warnings/errors 
- [x]  package build success
- [x]  proper versioning and commit representative of what was officially released to ubuntu/devel b3d9acddd352f2

PR #1285, opened 2022-02-21
sources/azure: prevent tight loops for DHCP retries

cjp256:
With debug logging, tight loops may result in huge log file sizes, e.g.:
 "Unable to find fallback nic"
 
 1. Raise NoDHCPLeaseMissingDhclientError to caller if no dhclient found
  instead of retrying DHCP, retrying will not fix a missing dhclient.
 
 2. For other DHCP failures, retry after sleeping one second.
 
Depends on #1276 

Would be a good candidate for SRU, if possible.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -6,7 +6,6 @@
 
 import base64
 import crypt
-import datetime
 import functools
 import os
 import os.path
@@ -24,7 +23,12 @@ from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
 from cloudinit.net import device_driver
-from cloudinit.net.dhcp import EphemeralDHCPv4, NoDHCPLeaseError
+from cloudinit.net.dhcp import (
+    EphemeralDHCPv4,
+    NoDHCPLeaseError,
+    NoDHCPLeaseInterfaceError,
+    NoDHCPLeaseMissingDhclientError,
+)
 from cloudinit.reporting import events
 from cloudinit.sources.helpers import netlink
 from cloudinit.sources.helpers.azure import (
@@ -358,7 +362,11 @@ class DataSourceAzure(sources.DataSource):
 
     @azure_ds_telemetry_reporter
     def _setup_ephemeral_networking(
-        self, *, iface: Optional[str] = None, timeout_minutes: int = 5
+        self,
+        *,
+        iface: Optional[str] = None,
+        retry_sleep: int = 1,
+        timeout_minutes: int = 5,
     ) -> None:
         """Setup ephemeral networking.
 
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -390,16 +400,39 @@ def _setup_ephemeral_networking(
...
+                        "dhclient executable not found", logger_func=LOG.error
+                    )
+                    self._ephemeral_dhcp_ctx = None
+                    raise
                 except NoDHCPLeaseError:
-                    continue
+                    # Typical DHCP failure, continue with backoff increase.
+                    report_diagnostic_event(
+                        "Failed to obtain DHCP lease (iface=%s)" % iface,
+                        logger_func=LOG.error,
```
anhvoms, line 430: I assume this exception is only raised if dhclient successfully ran, but dhcp failed. dhclient already has its own backoff interval and retry, do we still need to backoff in this case?

cjp256: My paranoia here was anything that could result in a quick/early exit of dhclient.

Perhaps it would be better if we simply set a minimum time per iteration?

anhvoms: Agreed. That would be better.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -390,16 +400,31 @@ def _setup_ephemeral_networking(
...
+                except NoDHCPLeaseMissingDhclientError:
+                    # No dhclient, no point in retrying.
+                    report_diagnostic_event(
+                        "dhclient executable not found", logger_func=LOG.error
+                    )
+                    self._ephemeral_dhcp_ctx = None
+                    raise
                 except NoDHCPLeaseError:
-                    continue
+                    # Typical DHCP failure, continue with backoff increase.
```
anhvoms, line 420: comment no longer relevant

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
cjp256: Fixed, thanks!

cloudinit/sources/DataSourceAzure.py
```diff
@@ -265,6 +270,11 @@ def get_resource_disk_on_freebsd(port_id) -> Optional[str]:
     return None
 
 
+def datetime_utcnow() -> datetime.datetime:
```
TheRealFalcon, line 273: Is this helper really needed here? Also wondering if time.time() would be simpler than datetime.datetime.utcnow() and deltas.

TheRealFalcon (commented):
Overall looks good, but one inline comment/question and please rebase the previous merged PR into this one.

cloudinit/sources/DataSourceAzure.py
cjp256: time() makes more sense and removed the helper!

Thanks!

cjp256:
Also updated the loop so we don't sleep if we're past the timeout.

anhvoms (approved):

TheRealFalcon (approved):
Thanks!

PR #1290, opened 2022-02-21
fix apt source match

holmanb:
```
Fix apt test on azure
```
Tested via pycloudlib on azure and locally.


blackboxsw (approved):
Verified on both bionic and jammy. Thanks!

PR #1297, opened 2022-02-23
Better interface handling for Vultr, expect unexpected DHCP servers

eb3095:
## Proposed Commit Message
Better interface handling for Vultr, expect unexpected DHCP servers

```
If a private interface is configured, has a DHCP server on it, and Cloud-Init runs, the Datasource fails to fail over to the next interface interrupting seeking. Configuring it as such can also lead to delays in deploy time if it did. This sorts the interface list to limit the potential of this occurring. Checks for the metadata route before trying, and now appropriately fails over if those contingencies fails by appropriately continuing to seek on metadata failure as was originally intended. Ideally the upcoming candidate_nic would of been used here to avoid the interface list function.
```

## Additional Context
The complexity here is due to the wild inconsistency on the platform. An image may end up on a vm, or a baremetal server, or not even be made by us. Options to handle this has been limited. New unittests were added to keep this in check as well.

## Test Steps
Deploy a server with a private interface with a dhcp server running on the private network.

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -28,7 +28,7 @@ def get_metadata(url, timeout, retries, sec_between, agent):
                 iface=iface[0], connectivity_url_data={"url": url}
             ):
                 # Check for the metadata route, skip if not there
-                if not check_route():
+                if not check_route(url):
                     continue
 
                 # Fetch the metadata
@@ -65,7 +65,7 @@ def get_interface_list():
 
 # Check for /32 route that our dhcp servers inject
 # in order to determine if this a customer-run dhcp server
-def check_route():
+def check_route(url):
     # Get routes, confirm entry exists
     routes = netinfo.route_info()
 
@@ -75,7 +75,7 @@ def check_route():
 
     # Parse each route into a more searchable format
     for route in routes["ipv4"]:
-        if "169.254.169.254" in route["destination"]:
+        if route.get("destination", None) in url:
             return True
 
     return False
@@ -168,7 +168,7 @@ def generate_network_config(interfaces):
         interface = interfaces[i]
 
         # Skip interfaces set not to be configured
-        if "unconfigured" in interface and interface["unconfigured"]:
+        if interface.get("unconfigured"):
             continue
 
         private = generate_interface(interface)
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -5,9 +5,11 @@
 import json
 from functools import lru_cache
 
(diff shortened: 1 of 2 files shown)


cloudinit/sources/helpers/vultr.py
```diff
@@ -20,26 +22,65 @@ def get_metadata(url, timeout, retries, sec_between, agent):
...
+    # Get routes, confirm entry exists
+    routes = netinfo.route_info()
+
+    # If no tools exist and empty dict is returned
+    if "ipv4" not in routes:
+        return False
+
+    # Parse each route into a more searchable format
+    for route in routes["ipv4"]:
+        if "169.254.169.254" in route["destination"]:
```
TheRealFalcon, line 78: Is a check needed here? Can route ever not have destination?

cloudinit/sources/helpers/vultr.py
```diff
@@ -124,7 +165,13 @@ def generate_network_config(interfaces):
 
     # Prepare additional interfaces, private
     for i in range(1, len(interfaces)):
-        private = generate_interface(interfaces[i])
+        interface = interfaces[i]
+
+        # Skip interfaces set not to be configured
+        if "unconfigured" in interface and interface["unconfigured"]:
```
TheRealFalcon, line 171: Nit: This could be simplified to `if interface.get("unconfigured"):`

TheRealFalcon (approved):
Looks good with an inline question and nit.

cloudinit/sources/helpers/vultr.py
relwell:
```
if route.get("destination", None) == "169.254.159.254":
    return True
```

Might be more nilsafe and remain a one-liner.

relwell: I'm also curious as to whether we should be explicitly referring to the IP string literal here. Referencing `BULTIN_DNS_CONFIG["url"]` might be easier to maintain in the long run and more meaningful.

cloudinit/sources/helpers/vultr.py
eb3095: TIL. Fixed!

cloudinit/sources/helpers/vultr.py
eb3095: > I'm also curious as to whether we should be explicitly referring to the IP string literal here. Referencing `BULTIN_DNS_CONFIG["url"]` might be easier to maintain in the long run and more meaningful.

Could be useful if someone does something weird here or if we change this (doubtful). However I adjusted to see if the route is in the url to avoid having to do any wacky regex or parsing.

PR #1302, opened 2022-02-24
sources/azure: remove lease file parsing

cjp256:
With reporting ready now happening in local phase, we have access
to ephemeral DHCP lease options and no longer need to parse DHCP
lease files.

- Switch from tracking wireserver endpoint in its encoded form to the
  IP string, parsing it only when read from lease options.

- Drop fallback_lease_file and dhcp_options parameters in favor of
  processed endpoint string.

- Add some minor type information for mypy.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -34,6 +34,7 @@ from cloudinit.sources.helpers import netlink
 from cloudinit.sources.helpers.azure import (
     DEFAULT_REPORT_FAILURE_USER_VISIBLE_MESSAGE,
     DEFAULT_WIRESERVER_ENDPOINT,
+    WALinuxAgentShim,
     azure_ds_reporter,
     azure_ds_telemetry_reporter,
     build_minimal_ovf,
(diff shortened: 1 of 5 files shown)


anhvoms (approved):
LGTM!

TheRealFalcon (approved):
LGTM! I'm going to close/re-open to re-trigger CI

TheRealFalcon (changes requested):
Looks like there's still some broken tests.

cjp256:
Fixed tests & rebased

holmanb:
> There seems to be an unrelated failure in the tests:

This is the second time I've seen that failure in the last this week. I think this is a flaky test that should be removed from CI. This depends on multiple keyservers so remote failure is likely cause, not your changes. A no-op push should get tests passing. I am proposing a separate PR to remove that test from CI runs.

cjp256:
It failed twice this morning so I had just pushed -vv to the tests for better visibility.

TheRealFalcon:
We're putting up a PR to (re)move the test. It's failing because ssh-import-id is failing, which would be due to the Launchpad or Github service being down and unrelated to cloud-init.

TheRealFalcon:
I see that it failed again which indicates a different problem, but it is still a problem with ssh-import-id and not cloud-init, so I still moved the test out of CI. If you rebase, we should be able to pass now.

cjp256:
Done

TheRealFalcon (approved):
Thanks!

PR #1315, opened 2022-03-03
Remove Unused Init Local Artifact

holmanb:
```
Remove unused init local artifact
```
This isn't used in cloud-init and systemd provides commands to show that `cloud-init-local.service` has been reached. It seems like a debugging artifact that shouldn't be needed.

### DIFF
--- a/systemd/cloud-init-local.service.tmpl
+++ b/systemd/cloud-init-local.service.tmpl
@@ -19,7 +19,6 @@ RequiresMountsFor=/var/lib/cloud
 [Service]
 Type=oneshot
 ExecStart=/usr/bin/cloud-init init --local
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
+1 from me, but @blackboxsw can you review also in case there's more context I'm not aware of?

blackboxsw (approved):
LGTM. Will not merge until further discussion @holmanb as you marked it WIP.

holmanb:
> LGTM. Will not merge until further discussion @holmanb as you marked it WIP.

Thanks! This was marked WIP since I was considering adding another commit. I will put that change (which is unrelated but adjacent) in a separate PR.

PR #1317, opened 2022-03-03
Docs

holmanb:
```
Doc cleanups
```

## Additional Context

Remove todos


TheRealFalcon (approved):
The API for datasources is a useful bit of documentation we should have, but as-is it is incorrect, which I believe is worse than having no documentation at all.

PR #1319, opened 2022-03-04
tests: add common tag prefix for cloud instances and groups

blackboxsw:
## Proposed Commit Message
```
To aid in debug of test instances launched by cloud-init, use
a common tag prefix format:
   ci-<platform>-<release>-MMDD-<timestamp>[0-9a-z]{8}

Example instance names:
   ci-lxd-container-jammy-0304-144058la5mnhu9
```

## Additional Context
SC-840

## Test Steps
```
CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_KEEP_INSTANCE=true CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:cloud-init-dev/daily CLOUD_INIT_OS_IMAGE=jammy::ubuntu::jammy .tox/integration-tests/bin/pytest tests/integration_tests/modules/test_apt.py::TestApt::test_ppa_source

$ lxc ls ci
 lxc ls ci
+--------------------------------------------+---------+--------------------+-----------------------------------------------+-----------+-----------+
|                    NAME                    |  STATE  |        IPV4        |                     IPV6                      |   TYPE    | SNAPSHOTS |
+--------------------------------------------+---------+--------------------+-----------------------------------------------+-----------+-----------+
| ci-lxd-container-jammy-0304-144058la5mnhu9 | RUNNING | 10.69.12.43 (eth0) | fd42:fc73:a2ea:bdfb:216:3eff:feb7:47ca (eth0) | CONTAINER | 0         |
+--------------------------------------------+---------+--------------------+-----------------------------------------------+-----------+-----------+
```

## Checklist:

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -1,10 +1,9 @@
 # This file is part of cloud-init. See LICENSE file for license information.
-import datetime
 import logging
 import os.path
 import random
 import string
-from abc import ABC, abstractmethod
+from abc import ABC
 from typing import Optional, Type
 from uuid import UUID
 
@@ -92,9 +91,12 @@ class ImageSpecification:
 class IntegrationCloud(ABC):
     datasource: str
     cloud_instance: BaseCloud
+    instance_tag: str
+    pycloudlib_instance_cls: Type[BaseCloud]
 
     def __init__(self, settings=integration_settings):
         self.settings = settings
+        self.image_specification = ImageSpecification.from_os_image()
         self.cloud_instance: BaseCloud = self._get_cloud_instance()
         self.initial_image_id = self._get_initial_image()
         self.snapshot_id = None
@@ -114,16 +116,18 @@ class IntegrationCloud(ABC):
             )
         )
 
-    @abstractmethod
     def _get_cloud_instance(self):
-        raise NotImplementedError
+        release = self.image_specification.release
+        platform = integration_settings.PLATFORM.replace("_", "")
+        return self.pycloudlib_instance_cls(tag=f"ci-{platform}-{release}")
 
     def _get_initial_image(self):
-        image = ImageSpecification.from_os_image()
         try:
-            return self.cloud_instance.daily_image(image.image_id)
+            return self.cloud_instance.daily_image(
+                self.image_specification.image_id
+            )
         except (ValueError, IndexError):
-            return image.image_id
+            return self.image_specification.image_id
 
     def _perform_launch(self, launch_kwargs, **kwargs):
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
@@ -200,9 +204,8 @@ class IntegrationCloud(ABC):
 
 class Ec2Cloud(IntegrationCloud):
     datasource = "ec2"
-
-    def _get_cloud_instance(self):
-        return EC2(tag="ec2-integration-test")
+    cloud_instance: EC2
+    pycloudlib_instance_cls = EC2
 
     def _perform_launch(self, launch_kwargs, **kwargs):
         """Use a dual-stack VPC for cloud-init integration testing."""
@@ -215,19 +218,14 @@ class Ec2Cloud(IntegrationCloud):
 
 class GceCloud(IntegrationCloud):
     datasource = "gce"
-
-    def _get_cloud_instance(self):
-        return GCE(
-            tag="gce-integration-test",
-        )
+    cloud_instance: GCE
+    pycloudlib_instance_cls = GCE
 
 
 class AzureCloud(IntegrationCloud):
     datasource = "azure"
     cloud_instance: Azure
-
-    def _get_cloud_instance(self):
-        return Azure(tag="azure-integration-test")
+    pycloudlib_instance_cls = Azure
 
     def destroy(self):
         if self.settings.KEEP_INSTANCE:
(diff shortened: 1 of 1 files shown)


holmanb (approved):
Looks good to me. Tested locally and verified lxd container name is as expected. Approved, pending the failing CI tests.

tests/integration_tests/clouds.py
```diff
@@ -92,9 +91,12 @@ def from_os_image(cls):
 class IntegrationCloud(ABC):
     datasource: str
     cloud_instance: BaseCloud
+    instance_tag: str
+    pycloudlib_instance_cls = Type(BaseCloud)
```
paride, line 95: This fails for me (Python 3.10):
```
tests/integration_tests/conftest.py:16: in <module>
    from tests.integration_tests.clouds import (
tests/integration_tests/clouds.py:91: in <module>
    class IntegrationCloud(ABC):
tests/integration_tests/clouds.py:95: in IntegrationCloud
    pycloudlib_instance_cls = Type(BaseCloud)
/usr/lib/python3.10/typing.py:941: in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
E   TypeError: Type Type cannot be instantiated; use type() instead
```
Replacing with `type()` works.

tests/integration_tests/clouds.py
```diff
@@ -92,9 +91,12 @@ def from_os_image(cls):
 class IntegrationCloud(ABC):
     datasource: str
     cloud_instance: BaseCloud
+    instance_tag: str
```
paride, line 94: This looks unused?

tests/integration_tests/clouds.py
```diff
@@ -300,8 +290,7 @@ def _perform_launch(self, launch_kwargs, **kwargs):
         except KeyError:
             profile_list = self._get_or_set_profile_list(release)
 
-        prefix = datetime.datetime.utcnow().strftime("cloudinit-%m%d-%H%M%S")
-        default_name = prefix + "".join(
+        default_name = self.cloud_instance.tag + "".join(
```
paride, line 293: Maybe in pycloudlib we should make LXD use the `tag` to set the instances default name, like in the other clouds. (Not to be fixed here.)

tests/integration_tests/clouds.py
```diff
@@ -300,8 +290,7 @@ def _perform_launch(self, launch_kwargs, **kwargs):
...
             random.choices(string.ascii_lowercase + string.digits, k=8)
```
paride, line 294: Now we have to "moving parts" in the instance naming: the timestamp set in pycloudlib and this random string. In Jenkins we may want to use fixed instance names to prevent the proliferation of stale instances should we have bugs in the cleanup code. This would require disabling both this random string and the pycloudlib timestamp.

I was thinking of dropping this random string bit, leaving it to pycloudlib to create unique instance names based on the tag. Then we can add a `pycloudlib.toml` setting to make it not add the timestamp in the cases where we want fixed instance names instead.

We already have seconds resolution in the pycloudlib timestamping, hopefully that's enough to prevent collisions. Thoughts?

paride (changes requested):
Big improvement in handling the instance naming, thanks! I left a few inline comments, where the more important one is a RFC about dropping the random part of the instance names.

tests/integration_tests/clouds.py
```diff
@@ -114,16 +116,18 @@ def emit_settings_to_log(self) -> None:
             )
         )
 
-    @abstractmethod
     def _get_cloud_instance(self):
-        raise NotImplementedError
+        release = self.image_specification.release
+        platform = integration_settings.PLATFORM.replace('_', '-')
```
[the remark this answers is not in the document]
paride: In any case black wants double quotes here.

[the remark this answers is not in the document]
blackboxsw: +1 on that collapse is good.

tests/integration_tests/clouds.py
blackboxsw: I had a typo/thinko
```suggestion
    pycloudlib_instance_cls: Type[BaseCloud]
```

tests/integration_tests/clouds.py
blackboxsw: I think that makes sense in pycloudlib so all clouds behave the same-ish,

tests/integration_tests/clouds.py
blackboxsw: Yes this makes sense. If we have parallel jobs running in jenkins, it's likely they'll be different Ubuntu series or platforms so collisions should not generally occur. If we end up runing jobs for a single platform-series in parallel instead of sequentially at some point we can grow random chars at that point if we actually hit a collision problem. I'll drop this from this branch and we can incorporate that into pycloudlib in the future if needed.

paride: +1 on incorporating this into pycloudlib if needed.

blackboxsw: Had to delete my prior comment on this. I agree with the sentiement about pycloudlib randomizing instance name based on the cloud.tag prefix. The issue we have at the moment in cloud-init CI is that pycloudlib doesn't yet have this feature, and the tag is created only at the start of integration tests. So, until pycloudlib grows this feature to create names based on tags, this branch should either go to WIP/draft and await our work on pycloudlib, or add this random string to allow sequentially running mulitple integration tests in sequence. I'm good with WIP/draft on this PR if we want to add the pycloudlib feature first.

paride (commented):
+1 on landing this modulo the review comments, when they make sense :) Go ahead without my explicit approval as I'll be on pto until Monday.

blackboxsw:
Marked draft as I didn't get around to putting up the pycloudlib prerequisite PR so that we can drop the rand suffix up in cloud-init integration tests. Will get back to this in a week.

PR #1320, opened 2022-03-07
travis supports 3.10 now (SC-843)

holmanb:
```
Switch to python 3.10 in Travis CI

Travis 3.10 support previously required pinning to 3.10.1. This restriction
no longer exists.
```


TheRealFalcon (approved):
Thanks!

PR #1322, opened 2022-03-08
remove flaky test from ci (SC-850)

holmanb:
```
Remove flaky test

This ssh-import-id test uses multiple remote key servers
that fail periodically.
```

## Additional Context
Here is an example flaky failure:
https://github.com/canonical/cloud-init/pull/1302#issuecomment-1061763157

This same failure has happened to me as well.

### DIFF
--- a/tests/integration_tests/modules/test_combined.py
+++ b/tests/integration_tests/modules/test_combined.py
@@ -145,23 +145,6 @@ class TestCombined:
         client = class_client
         assert "hello world" == client.read_from_file("/var/tmp/runcmd_output")
 
-    @retry(tries=30, delay=1)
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
LGTM assuming tests pass.

Long term, I think it's worth seeing if ssh-import-id can give us any meaningful response other than failure. If it fails because on of the services are down, we can just skip the test for that run. If not, this is probably better replaced with a unit test.

cjp256 (approved):

holmanb:
> Long term, I think it's worth seeing if ssh-import-id can give us any meaningful response other than failure. 

Based on a quick glance at the source, it looks like return codes are hardcoded at 0 and 1, but in the case of failure (rc=1) it prints out error messages we could use.

PR #1326, opened 2022-03-10
Revert "Ensure system_cfg read before ds net config on Oracle (#1174)"

TheRealFalcon:
## Proposed Commit Message

```
This reverts commit b306633fd17e5ba0173ad3c41add59cb11884757.

While this ultimately seems like a better solution, currently the
file /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg exists on
all Oracle launched instances which will prevent networking from
being properly initialized.
```

## Additional Context
#1174
https://bugs.launchpad.net/cloud-init/+bug/1956788

## Checklist:


TheRealFalcon:
I built a deb and tested that `tests/integration_tests/modules/test_combined.py` runs successfully. Without this revert, we would not be able to connect to the instance.

blackboxsw (approved):
LGTM revert is clean revert here to what was already released and functional on Oracle.

PR #1327, opened 2022-03-11
Fix sysconfig render when set-name is missing

akutz:
## Proposed Commit Message

```
Fix sysconfig render when set-name is missing

This patch addresses an issue where the absence of set-name
in a network configuration potentially results in an unintended
network configuration.

Please note a function from cloudinit.net.networkstate was relocated
under cloudinit.net due to a circular dependency introduced by the
former calling get_interfaces_by_mac in the latter. A stub function
has been created to preserve backward compatibility.

LP: #1855945
```

## Additional Context
NA

## Test Steps
Still `TODO`

## Checklist:


cloudinit/net/network_state.py
```diff
@@ -683,6 +692,24 @@ def handle_ethernets(self, command):
...
+            # following in the order they are listed:
+            #   * set-name
+            #   * interface name looked up by mac
+            #   * value of "eth" key from this loop
+            name = eth
+            set_name = cfg.get("set-name", None)
+            if set_name:
+                name = set_name
+            elif mac_address and ifaces_by_mac:
+                lcase_mac_address = mac_address.lower()
```
TheRealFalcon, line 706: Nit: this line could be removed and `.lower()` could be called in the if statement.

TheRealFalcon (changes requested):
Thanks for this fix. Overall LGTM.

The for loop is juuuust starting to get long enough that I'd consider putting the code you added into its own method/function, but I think it's ok as-is too. I'll leave it up to you if you want to change it.

Re: the circular import, I'm not a huge fan of the proposed solution as it adds more indirection and imports an underscore prefixed name.

Given that your current solution is already a workaround, I think a simpler (though maybe slightly hackier) workaround is to move the `get_interfaces_by_mac` import from the top of `network_state.py` to the top of the `handle_ethernets` method where you added it.

Long term, I think that most of the functions at the bottom of `network_state.py` (`is_ipv6_addr` through `mask_and_ipv5_to_bcast_addr`) probably make more sense in `cloudinit/net/__init__.py`. That's obviously a bigger refactor though, and not one we should do in this PR.

akutz:
> Re: the circular import, I'm not a huge fan of the proposed solution as it adds more indirection and imports an underscore prefixed name.

Yeah, neither was I, but I also am not so actively involved with the project that I felt comfortable to refactor things too heavily.

cloudinit/net/network_state.py
TheRealFalcon: Yes, good point. Ignore my comment.

TheRealFalcon:
> Oh, are you suggesting relocating the actual import statement, not the method itself?

Yes, but since you did the refactor instead (thanks for that!), not needed.

TheRealFalcon (approved):
Thanks @akutz . Assuming CI passes, I'll merge this soon.

PR #1328, opened 2022-03-11
url_helper: add tuple support for readurl timeout

cjp256:
It may be useful to configure connection timeout and read timeout
separately.  Update readurl() to accept a tuple that is supported
by python requests to configure both.


TheRealFalcon (approved):
LGTM, thanks!

PR #1329, opened 2022-03-11
Fix fs label documentation and ds-identify (SC-854)

holmanb:
```
Make fs labels match for ds-identify and docs

- ConfigDrive datasource works with CONFIG-2, document it
- RbxCloud datasource works with cloudmd, document it
- OpenNebula datasource supports CDROM, but ds-identify doesn't check it
```


holmanb:
Note: OpenNebula currently doesn't have any existing tests in `test_ds_identify.py` so I opted not to add any for the change to `ds-identify`. If others prefer, I can add one.

PR #1330, opened 2022-03-11
Doc fixes (SC-857)

holmanb:
```
Various doc fixes:

    - fix rst to actually render the security page
    - remove headers of missing examples
    - m2r seems abandoned, switch to m2r2
    - hide deprecated content in a drop-down panel
    - bump sphinx to 4.3
    - ping sphinx_rtd_theme (workaround for bullet list rendering issue)
    - use inline literal rather than italics for code
    - fix a broken link
```

Note: it looks like docutils 0.17 / sphinx_rtd_theme < 0.5.2 breaks things, context can be found [here](https://github.com/readthedocs/sphinx_rtd_theme/issues/1115) (link included as a comment in doc-requirements.txt)

Planned future work: resolve the 40+ warnings that sphinx spews (mostly harmless, I think?) and then make warnings fail CI (this would have prevented some of the errors I fixed in this PR).

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -1,10 +1,9 @@
 doc8
-m2r
-sphinx<2
+m2r2
+sphinx
 sphinx_rtd_theme
 pyyaml
 
 # Indirect dependencies
 jinja2<3.1.0  # https://github.com/readthedocs/readthedocs.org/issues/9037
 docutils<0.18
-mistune<2.0.0  # https://github.com/miyakogi/m2r/issues/66
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -3,6 +3,7 @@ m2r2
 sphinx
 sphinx_rtd_theme
 pyyaml
+sphinx-panels
 
 # Indirect dependencies
 jinja2<3.1.0  # https://github.com/readthedocs/readthedocs.org/issues/9037
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -1,10 +1,10 @@
 doc8
 m2r2
-sphinx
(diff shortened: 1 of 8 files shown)


TheRealFalcon (changes requested):
Nice changes and good idea hiding deprecated instructions behind a dropdown.

However, I built this branch on rtd, and things aren't rendering correctly.

image

Left is current main, right is this branch. We lose bullets and some kind of formatting in multiple places.

Our version of sphinx is quite old, and I'm assuming the original reasons for limiting it no longer apply, so I'm happy to upgrade our version of it, but that may need to happen in a different PR in order to figure out the various quirks involved.

holmanb:
> However, I built this branch on rtd, and things aren't rendering correctly.

Thanks for the review @TheRealFalcon. I think I have things worked out now.

> image

It looks like this is a problem with specific version combinations of docutils and sphinx_rtd_theme. I've pinned the dependency versions for the bullet list rendering to work.

I also noticed that the cli page had issues rendering cli flags with two hyphens inside italics, so I switched those to use inline codeblocks, which also look better anyways in my opinion.

I don't see any other issues so I'm switching this back from draft to "ready for review".

.readthedocs.yaml
```diff
@@ -3,6 +3,6 @@ version: 2
 formats: all
 
 python:
-    install:
-        - requirements: doc-requirements.txt
-        - path: .
+  install:
+    - path: .
```
blackboxsw, line 7: I think this small changeset already landed in main per a James branch Friday. might need a `git rebase` to make sure something else in James' doc branches doesn't adversely affect your changes in rendering.

PR #1331, opened 2022-03-12
add Requires=cloud-init-hotplugd.socket

yangzz-97:
I think if the service depends on socket, it should change After=cloud-init-hotplugd.socket to Requires=cloud-init-hotplugd.socket in the cloud-init-hotplugd.service file.

### DIFF
--- a/systemd/cloud-init-hotplugd.service
+++ b/systemd/cloud-init-hotplugd.service
@@ -11,6 +11,7 @@
 
 [Unit]
 Description=cloud-init hotplug hook daemon
+After=cloud-init-hotplugd.socket
 Requires=cloud-init-hotplugd.socket
 
 [Service]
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
I'd be ok with adding a `Requires`, but I think the `After` should exist as well. The service shouldn't run unless the socket is already available, and `Requires` won't necessarily guarantee that as it allows them to start in parallel.

Additionally, before we can accept your change, you need to sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html#submitting-your-first-pull-request)

yangzz-97:
You're right. There will be parallel startup in Requires. After and Requires are both required. I'll modify them later

PR #1333, opened 2022-03-14
Expose https_proxy env variable to ssh-import-id cmd

michaelrommel:
The import of ssh keys for users does not work in corporate networks
behind proxies.

The ssh-import-id command does not support cmd line arguments or other
configurations for specifying a proxy. The sudo command in the module
executes the ssh-import-id command directly, so any proxy settings of
shells will not work. The only option would be either to specify the
proxy setting on the command line, but this wouldn't be easily user
editable.

So the only solution is to selectively forward any already set
https_proxy variable to the executed command. Since only one env
variable is used and forwarded, the security impact should be
acceptable.

### DIFF
--- a/cloudinit/config/cc_ssh_import_id.py
+++ b/cloudinit/config/cc_ssh_import_id.py
@@ -100,7 +100,7 @@ def import_ssh_ids(ids, user, log):
     except KeyError as exc:
         raise exc
 
-    cmd = ["sudo", "-Hu", user, "ssh-import-id"] + ids
+    cmd = ["sudo", "--preserve-env=https_proxy -Hu", user, "ssh-import-id"] + ids
     log.debug("Importing SSH ids for user %s.", user)
 
     try:
--- a/cloudinit/config/cc_ssh_import_id.py
+++ b/cloudinit/config/cc_ssh_import_id.py
@@ -100,7 +100,38 @@ def import_ssh_ids(ids, user, log):
     except KeyError as exc:
         raise exc
 
-    cmd = ["sudo", "--preserve-env=https_proxy -Hu", user, "ssh-import-id"] + ids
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
@michaelrommel I don't completely understand how this would work. Can you provide some more context? Where would the `https_proxy` env var come from? Who would set it and where?

michaelrommel:
I have a `user-data` file that includes a `bootcmd` section as follows:

```
bootcmd:
 - |
   cloud-init-per once env sh -c "mkdir -p /etc/systemd/system/cloud-config.service.d &&
   mkdir -p /etc/systemd/system/cloud-final.service.d && { cat > /etc/cloud/env <<-EOF
   http_proxy=http://192.168.10.212:3128/
   https_proxy=http://192.168.10.212:3128/
   EOF
   } && { cat > /etc/systemd/system/cloud-config.service.d/override.conf <<-EOF
   [Service]
   EnvironmentFile=/etc/cloud/env
   PassEnvironment=https_proxy http_proxy
   EOF
   } && { cat > /etc/systemd/system/cloud-final.service.d/override.conf <<-EOF
   [Service]
   EnvironmentFile=/etc/cloud/env
   PassEnvironment=https_proxy http_proxy
   EOF
   } && systemctl daemon-reload"
```

This leads to exporting the required environment variables for that particular environment being exported as part of the two standard cloud-init systemd services. These could then be passed on to the import-ssh-id command as I described above.

The creation of those files could also happen at some other place in the user-data file.

Hope that helps.

TheRealFalcon (approved):
Thanks @michaelrommel for the explanation. This is a valid use case, though I think we should probably handle it in a more holistic manner. Given that that may be a fair bit of work, I'm using your solution, but also pushed a big TODO comment for us in the future.

Assuming CI passes, I'll go ahead and merge it.

PR #1337, opened 2022-03-15
Add --no-update to add-apt-repostory call (SC-880)

TheRealFalcon:
## Proposed Commit Message

```
Add --no-update to add-apt-repostory call

In cc_apt_configure, "add-apt-repository" is called a few lines before
we make a call to update packages, so the additional update that
happens in "add-apt-repository" is redundant and a waste of time.
Additionally, the update in add-apt-repository can fail on
obtaining apt lock and cause the module to raise an exception.
```

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -571,7 +571,10 @@ def add_apt_sources(
 
         if aa_repo_match(source):
             try:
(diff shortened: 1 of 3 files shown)


holmanb (approved):
LGTM!

blackboxsw (approved):
LGTM

PR #1340, opened 2022-03-17
BUG 1473527: module ssh-authkey-fingerprints fails Input/output error…

andrew-lee-metaswitch:
This fixes https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1473527?comments=all

## Proposed Commit Message

```
Don't error if we cannot log to /dev/console

We've seen instances on VMware of serial consoles not being set up
correctly by the kernel, making /dev/ttyS0 not set up correctly, and 
hence /dev/console not writeable to.

In such circumstances, cloud-init should not fail, instead it should 
gracefully fall back to logging to stdout.

The only time cloud-init tries to write to `/dev/console` is in the
`multi_log` command- which is called by the
ssh-authkey-fingerprints module

LP: #1473527
```

## Checklist:

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -29,7 +29,7 @@ import socket
 import stat
 import string
 import subprocess
-import sys
+import 
 import time
 from base64 import b64decode, b64encode
 from errno import ENOENT
@@ -393,14 +393,23 @@ def multi_log(
         sys.stderr.write(text)
     if console:
         conpath = "/dev/console"
+        writing_to_console_failed = False
         if os.path.exists(conpath):
-            with open(conpath, "w") as wfh:
-                wfh.write(text)
-                wfh.flush()
-        elif fallback_to_stdout:
-            # A container may lack /dev/console (arguably a container bug).  If
-            # it does not exist, then write output to stdout.  this will result
-            # in duplicate stderr and stdout messages if stderr was True.
+            try:
+                with open(conpath, "w") as wfh:
+                    wfh.write(text)
+                    wfh.flush()
+            except OSError:
+                writing_to_console_failed = True
+    
+        if fallback_to_stdout or writing_to_console_failed:
+            # A container may lack /dev/console (arguably a container bug).
+            # Additionally, /dev/console may not be writable to on a VM (again
+            # likely a VM bug).
+            #
+            # If either of these is the case, then write output to stdout.
+            # This will result in duplicate stderr and stdout messages if
+            # stderr was True.
             #
             # even though upstart or systemd might have set up output to go to
             # /dev/console, the user may have configured elsewhere via
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2078,7 +2078,7 @@ def write_file(
     omode="wb",
     preserve_mode=False,
     *,
-    ensure_dir_exists=True
+    ensure_dir_exists=True,
 ):
     """
     Writes a file with the given content and sets the file mode as specified.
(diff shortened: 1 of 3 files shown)


andrew-lee-metaswitch:
Sorry - I need to get agreement that I can sigm the CLA from my company - so this will have to wait a few days. 

In the meantime I'd be grateful for any ideas on how to add regression testing for this - it's not immediately obvious to me how to emulate the OSError we are seeing in testing

cloudinit/util.py
```diff
@@ -393,14 +393,24 @@ def multi_log(
...
-        elif fallback_to_stdout:
-            # A container may lack /dev/console (arguably a container bug).  If
-            # it does not exist, then write output to stdout.  this will result
-            # in duplicate stderr and stdout messages if stderr was True.
+            try:
+                with open(conpath, "w") as wfh:
+                    wfh.write(text)
+                    wfh.flush()
+                writing_to_console_worked = True
+            except OSError:
```
anhvoms, line 403: so we get into this situation if console is enabled and /dev/console exists, which indicates some sort of issue with the image. I would log a warning here indicate that serial console is likely busted

andrew-lee-metaswitch: /dev/console being unwritable to is not necessarily a sign that the image is busted (see e.g. https://www.linuxquestions.org/questions/linux-software-2/enable-dev-ttys0-on-centos-7-vm-on-esx-server-4175606689/).

Regardless, I've added a log.

TheRealFalcon:
@andrew-lee-metaswitch , Given how transient this issue is, I think testing that we haven't regressed any existing use cases is the best we can do. That plus unit testing would be sufficient.

andrew-lee-metaswitch:
Thanks @TheRealFalcon - I've written some UTs, it wasn't actually hard in the end (though I still maintain that pytest/unittest and mock is a dark magic with an impenetrable syntax). I've had to alter the non-test code to make it mockable/testable.

TheRealFalcon:
@andrew-lee-metaswitch , I've been holding off reviewing this one since it says WIP and you had/have issues with the CLA. Have those issues been resolved? Is this ready for review?

andrew-lee-metaswitch:
Apologies, @TheRealFalcon , I've now got agreement from the legal team and have signed the CLA.

cloudinit/util.py
```diff
@@ -393,14 +399,25 @@ def multi_log(
...
-        elif fallback_to_stdout:
-            # A container may lack /dev/console (arguably a container bug).  If
-            # it does not exist, then write output to stdout.  this will result
-            # in duplicate stderr and stdout messages if stderr was True.
+            try:
+                write_to_console(conpath, text)
+                writing_to_console_worked = True
+            except OSError:
+                console_error = "Failed to write to /dev/console"
+                sys.stdout.write(console_error)
```
TheRealFalcon, line 409: Can we add a newline here and also update the unit test accordingly?
```suggestion
                sys.stdout.write(f"{console_error}\n")
```

TheRealFalcon (changes requested):
Looks good! Just a small request inline.

TheRealFalcon:
Oh also, regarding,
> One minor thing - flake8 defaults to 79 character lines, whereas black to 80 characters, meaning I can run do_format over my code and be left with lines of length 80. check-in my code and still have flake8 fail - it would be nice if the flake8 and black config were consistent.

I tried to reproduce this but couldn't. We also have black set to 79 width:
https://github.com/canonical/cloud-init/blob/main/pyproject.toml#L2

Curious if you can come up with a simple reproducer on main

TheRealFalcon:
Unit tests failed after my last requested update, so I took the liberty of pushing the fix. Assuming CI passes, this looks good to me now.

Thanks for the contribution!

TheRealFalcon (approved):

PR #1341, opened 2022-03-18
schema: add JSON defs for modules resize-salt (SC-654)

TheRealFalcon:
## Proposed Commit Message

```
schema: add JSON defs for modules resize-salt

Includes:
 - cc_resizefs
 - cc_resolv_conf
 - cc_rh_subscription
 - cc_rightscale_userdata
 - cc_rsyslog
 - cc_runcmd
 - cc_salt_minion
```

## Additional Context
There's a legacy syntax for rsyslog that became legacy 6 years ago. It will still work but I removed all references to it, don't include it in the schema, and log a warning if it's used as it was already considered legacy 6 years ago.

Additionally, I'm not exactly sure what's needed for rightscale userdata. The context around it makes it sound like it's a value embedded within other yaml, rather being standard yaml itself, but I'm hoping somebody else has some additional context there. 

Also, Rightscale and Salt don't even exist as independent entities anymore, but I'm assuming their modules can still be used by those who bought the technology.


cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+    "cc_rh_subscription": {
+      "type": "object",
+      "properties": {
+        "rh_subscription": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "username": {
+              "type": "string",
+              "description": "The username to use. Must be used with password. Should not be used with activation-key or org"
```
blackboxsw, line 845: Markdown sugar for key names
```suggestion
              "description": "The username to use. Must be used with password. Should not be used with ``activation-key`` or ``org``"
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+            },
+            "password": {
+              "type": "string",
+              "description": "The password to use. Must be used with username. Should not be used with activation-key or org"
```
blackboxsw, line 849:
```suggestion
              "description": "The password to use. Must be used with username. Should not be used with ``activation-key`` or ``org``"
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+            },
+            "activation-key": {
+              "type": "string",
+              "description": "The activation key to use. Must be used with org. Should not be used with username or password"
```
blackboxsw, line 853:
```suggestion
              "description": "The activation key to use. Must be used with org. Should not be used with ``username`` or ``password``"
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+            },
+            "org": {
+              "type": "integer",
+              "description": "The organization number to use. Must be used with activation-key. Should not be used with username or password"
```
blackboxsw, line 857:
```suggestion
              "description": "The organization number to use. Must be used with ``activation-key``. Should not be used with ``username`` or ``password``"
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+            },
+            "auto-attach": {
+              "type": "boolean",
+              "description": "Whether to attach subscriptions automatically"
+            },
+            "service-level": {
+              "type": "string",
+              "description": "The service level to use when subscribing to RH repositories. auto-attach must be true for this to be used"
```
blackboxsw, line 865:
```suggestion
              "description": "The service level to use when subscribing to RH repositories. ``auto-attach`` must be true for this to be used"
```

blackboxsw (changes requested):
Not for this PR, but I really don't like that rsyslog_config [silently sets default values if invalid values are provided in config](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_rsyslog.py#L238-L239). It's probably worth us taking a separate bug about this for future cleanup.

All else looks pretty good for this PR so take or leave comments as you want with the exception of my question on  manage_resolv_conf default value. Maybe I'm reading the code wrong there but it doesn't look like we default to True. I'll review your responses tomorrow on the rest of the items and we can close out on this branch. Thanks again

cloudinit/config/cloud-init-schema.json
```diff
@@ -787,6 +787,226 @@
...
+        }
+      }
+    },
+    "cc_resolv_conf": {
+      "type": "object",
+      "properties": {
+        "manage_resolv_conf": {
+          "type": "boolean",
+          "default": true,
+          "description": "Whether to manage the resolv.conf file. Default: ``true``"
```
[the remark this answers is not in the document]
TheRealFalcon: Brain-fart on my part. Yes, it defaults to false. The whole 'manage_resolv_conf` key is silly. It should just be that if you provide the config then we use it, but I don't think we really gain anything by deprecating it. I'll add something to the description.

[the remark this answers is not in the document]
TheRealFalcon: Filed https://warthogs.atlassian.net/browse/SC-878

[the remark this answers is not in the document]
blackboxsw: +1 LGTM!

blackboxsw (approved):
Thanks for the quick fixes. +1

PR #1346, opened 2022-03-22
Add 2021-07-15 EC2 instance meta version (Adds support for instance tags)

SeanHood:
By updating to the latest EC2 instance metadata version, EC2 instance tags are then available from within cloud-init.

Ref: https://aws.amazon.com/about-aws/whats-new/2022/01/instance-tags-amazon-ec2-instance-metadata-service/

This feels like a rather big change without added tests, however I didn't see any other tests which checked for arbitrary keys within the metadata. In a separate commit I've added a sample of the 2021-07-15 metadata.

https://github.com/SeanHood/cloud-init/commit/51fc6ea86fa76a931536a092465575992a61fa02

## Proposed Commit Message

```
summary: Add 2021-07-15 EC2 instance meta version

By updating to the latest EC2 instance metadata version, EC2 instance tags are then available from within cloud-init.

Ref: https://aws.amazon.com/about-aws/whats-new/2022/01/instance-tags-amazon-ec2-instance-metadata-service/
```

## Test Steps
1. Create EC2 instance with some tags and enable "Allow tags in instance metadata"

2. With this patch there should now be tags populated in the cloud-init metadata.
    ```
    $ jq '.ds."meta-data".tags.instance' /run/cloud-init/instance-data.json
    {
      "Env": "Dev",
      "Name": "Test-instance"
    }
    ```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -62,7 +62,7 @@ class DataSourceEc2(sources.DataSource):
 
     # Priority ordered list of additional metadata versions which will be tried
     # for extended metadata content. IPv6 support comes in 2016-09-02
-    extended_metadata_versions = ["2018-09-24", "2016-09-02"]
+    extended_metadata_versions = ["2021-07-15", "2018-09-24", "2016-09-02"]
 
     # Setup read_url parameters per get_url_params.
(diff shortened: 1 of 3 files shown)


SeanHood:
Dupe of: #1309

PR #1348, opened 2022-03-23
changed passwd to hashed_passwd within examples

StefanKaerst:
## improved example page

```
summary: improved example page regarding hashed_passwd

according to cloudinit/distros/__init__.py
 607         * ``plain_text_passwd``
 608         * ``hashed_passwd``

there is no keyword "passwd", so I changed it in the examples.

### DIFF
--- a/doc/examples/cloud-config-user-groups.txt
+++ b/doc/examples/cloud-config-user-groups.txt
@@ -24,7 +24,7 @@ users:
       - lp:falcojr
       - gh:TheRealFalcon
     lock_passwd: false
-    passwd: $6$j212wezy$7H/1LT4f9/N3wpgNunhsIqtMj62OKiS3nyNwuizouQc3u7MbYCarYeAHWYPYb2FT.lbioDm2RrkJPb9BZMN1O/
+    hashed_passwd: $6$j212wezy$7H/1LT4f9/N3wpgNunhsIqtMj62OKiS3nyNwuizouQc3u7MbYCarYeAHWYPYb2FT.lbioDm2RrkJPb9BZMN1O/
   - name: barfoo
     gecos: Bar B. Foo
     sudo: ALL=(ALL) NOPASSWD:ALL
@@ -63,7 +63,7 @@ users:
 #           SELinux user.
 #   lock_passwd: Defaults to true. Lock the password to disable password login
 #   inactive: Number of days after password expires until account is disabled
-#   passwd: The hash -- not the password itself -- of the password you want
+#   hashed_passwd: The hash -- not the password itself -- of the password you want
 #           to use for this user. You can generate a safe hash via:
 #               mkpasswd --method=SHA-512 --rounds=4096
 #           (the above command would create from stdin an SHA-512 password hash

### CODE REVIEW

**Comment by holmanb** [comment]
Hi @StefanKaerst, thanks for the PR. Do you have an example of this behavior being broken?

Here is an example of a working cloud-config that uses the `passwd` key.
```
#cloud-config
users:
  - name: holmanb
    groups: sudo
    expiredate: '2032-09-01'
    lock_passwd: false
    passwd: $6$l5sgH4fXTpFhLpbC$7WoVfs65hBbfX30u91nrXm2A8yow2RucL6kKtlO2evL.1C92aghN5UE.qpsQpydILyf5nLDETHqWMTIQiQtZD1
    primary_group: holmanb
    shell: '/bin/bash'
```

After initialization, the user can use the password "newpassword" to escalate.
```
root@me:~# su holmanb                     
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

holmanb@me:/root$ whoami
holmanb
holmanb@me:/root$ echo "newpassword" | sudo -S whoami
[sudo] password for holmanb: root
```

StefanKaerst:
Brett Holman wrote on 23/03/2022 16:11:
 > Hi @StefanKaerst <https://github.com/StefanKaerst>, thanks for the PR. Do you have an example of this behavior being
broken?
hi Brett,


 > Here is an example of a working cloud-config that uses the |passwd| key.
ok, thanks!

we use OpenStack to create compute nodes (VMs) using terraform. I'm not sure if this is important, but we faced a
problem using an Ubuntu image with an unknown version of cloud-init.
we tried to set a password for root according to the example provided here:
https://cloudinit.readthedocs.io/en/latest/topics/examples.html

at first we tried:
users:
   - name: root
     lock_passwd: false
     passwd: $6$rounds=4096$8WAM1FAnTHISE$WlFIIi27Obj...
   - default
   - name: ubuntu
     groups: users, sudo
     sudo: ALL=(ALL) NOPASSWD:ALL
     ssh_authorized_keys:
       - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCxot...

as this did not work (root had no password) I changed passwd to hashed_passwd as mentioned in the comments regarding
exceptions for existing users. I'm not sure what existing users means here, but I suppose a cloud-init-ready image for
VMs already contain the user root. after changing cloud-init config to use hashed_passwd, all new Ubuntu VMs were
created with root having our password.

btw. for me the example page is not clear enough about passwords in general. either I provide the password in clear and
therefore would use "plain_text_passwd" - or - I provide the password hashed and would use "hashed_passwd". why is there
a difference between existing users and new ones? does "passwd" work with clear passwords _and_ hashed passwords? maybe
there should be 4 directives instead: existinguser_hashed_passwd, existinguser_plain_text_passwd, newuser_hashed_passwd
and newuser_plain_text_passwd .. and no "passwd" at all. ;)   or, someone changes cloud-init in a way that both versions
of a password do work, as a hashed password is clearly distinguishable from a cleartext password. in this case only
"passwd" is needed. ;)      I personally would use "password" not "passwd" anyway, because it transports the _password_
not the command a linux sysadmin like me would use to change it.   just my thoughts.

thank you for your fast reply.
hopefully I could shed some light on this issue/PR

GL&HF!

--
Viele Grüße
Stefan Kärst

--------------------------------
***@***.***:~> fortune -s
UNIX was half a billion (500000000) seconds old on
Tue Nov  5 00:53:20 1985 GMT (measuring since the time(2) epoch).
                 -- Andy Tanenbaum

***@***.***:~> e2d 500000000
Tue 5 Nov 01:53:20 CET 1985
***@***.***:~>
--------------------------------

holmanb:
> I'm not sure what existing users means here, but I suppose a cloud-init-ready image for
VMs already contain the user root.

I would assume all images booting cloud-init _successfully_ will have `root` as an existing user (cloud-init runs as root).

> after changing cloud-init config to use hashed_passwd, all new Ubuntu VMs were
created with root having our password.

I see. This PR makes more sense with that information.

> as this did not work (root had no password) I changed passwd to hashed_passwd as mentioned in the comments regarding exceptions for existing users.

Which comments and exceptions are you referring to?

> why is there a difference between existing users and new ones

Linux has different commands to update a current users password vs create a new user with a specific password.

I think that setting a root password on startup is probably a common enough task that we should add this to our examples. I can't say that I've ever done this, since I always change my root password on first boot, but I would assume some (like yourself) might want this documented.

I think the example that you change in this PR probably doesn't need to change; they should "work" for a new user. Maybe you could propose an example of setting root user's password?

I do agree that different keys for different password types would be better than what we have right now, but I haven't yet taken the time to propose a code change.

StefanKaerst:
Brett Holman wrote on 23/03/2022 23:46:

>     as this did not work (root had no password) I changed passwd to hashed_passwd as mentioned in the comments regarding
>     exceptions for existing users.
>
> Which comments and exceptions are you referring to?
cloud-config-user-groups.txt

9   # Add users to the system. Users are added after groups are added.
10  # Note: Most of these configuration options will not be honored if the user
11  #       already exists. Following options are the exceptions and they are
12  #       applicable on already-existing users:
13  #       - 'plain_text_passwd', 'hashed_passwd', 'lock_passwd', 'sudo',
14  #         'ssh_authorized_keys', 'ssh_redirect_user'.


>     why is there a difference between existing users and new ones
>
> Linux has different commands to update a current users password vs create a new user with a specific password.
maybe creating users without passwords and changing their passwords afterwards would be a good strategy to only use the
"passwd" command. this would remove all exceptions mentioned above too.


> I think the example that you change in this PR probably doesn't need to change; they should "work" for a new user. Maybe
> you could propose an example of setting root user's password?
maybe you explicitly add the word "existing", as long as there are different keywords to be used. I'm not sure about the
default user and passwd vs. plain_text_passed (this raises the same problem as with root because it depends on the OS
image you use).   again, I would change cloud-init so all keywords only exist in one version. it is easy to find out
whether a user already exists or not because login names are unique.
I'm not sure how ansible would do this, but I suppose it takes the user config from the config file as the desired
target state, which is .. creating all users not already there and change existing users according to config provided.
this way the config does not contain any knowledge about existing users vs. new users.

#existing user
users:
   - name: root
     lock_passwd: false
     hashed_passwd: $6$rounds=4096$8WAM1FAnTHISE$WlFIIi27Obj...

#new user
users:
   - name: developer
     groups: users
     passwd: $6$rounds=4096$8WAM1FAnTHISE$WlFIIi27Obj...
  #new default user for Ubuntu
users:
   - default
   - name: ubuntu
     groups: users, sudo
     sudo: ALL=(ALL) NOPASSWD:ALL
     plain_text_passwd: changeme

PR #1349, opened 2022-03-23
Fix cloud-init status --wait when no datasource found

TheRealFalcon:
## Proposed Commit Message

```
Fix cloud-init status --wait when no datasource found

In 0de7acb1, we modified status checks to wait until we get an "enabled"
or "disabled" file from our generator. It never outputs a
"disabled" file, so "status --wait" will wait indefinitely if no
datasource is found.

LP: #1966085
```

## Additional Context
See: https://github.com/canonical/cloud-init/pull/1162

## Test Steps
Run tests/integration_tests/cmd/test_status.py

I also manually checked the generator script to verify that enabled and disabled files get created as expected and are mutually exclusive.

### DIFF
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -5,6 +5,7 @@ import os.path
 import random
 import string
 from abc import ABC, abstractmethod
+from copy import deepcopy
 from typing import Optional, Type
 from uuid import UUID
 
@@ -290,7 +291,8 @@ class _LxdIntegrationCloud(IntegrationCloud):
             ).format(**format_variables)
             subp(command.split())
 
-    def _perform_launch(self, launch_kwargs, **kwargs):
+    def _perform_launch(self, original_kwargs, **kwargs):
+        launch_kwargs = deepcopy(original_kwargs)
         launch_kwargs["inst_type"] = launch_kwargs.pop("instance_type", None)
         wait = launch_kwargs.pop("wait", True)
         release = launch_kwargs.pop("image_id")
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -291,14 +291,16 @@ class _LxdIntegrationCloud(IntegrationCloud):
             ).format(**format_variables)
             subp(command.split())
 
-    def _perform_launch(self, original_kwargs, **kwargs):
-        launch_kwargs = deepcopy(original_kwargs)
-        launch_kwargs["inst_type"] = launch_kwargs.pop("instance_type", None)
-        wait = launch_kwargs.pop("wait", True)
-        release = launch_kwargs.pop("image_id")
+    def _perform_launch(self, launch_kwargs, **kwargs):
+        instance_kwargs = deepcopy(launch_kwargs)
+        instance_kwargs["inst_type"] = instance_kwargs.pop(
+            "instance_type", None
+        )
+        wait = instance_kwargs.pop("wait", True)
+        release = instance_kwargs.pop("image_id")
 
         try:
-            profile_list = launch_kwargs["profile_list"]
+            profile_list = instance_kwargs["profile_list"]
         except KeyError:
             profile_list = self._get_or_set_profile_list(release)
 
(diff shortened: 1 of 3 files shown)


tests/integration_tests/cmd/test_status.py
```diff
@@ -0,0 +1,58 @@
...
+@pytest.mark.ubuntu
+@pytest.mark.lxd_container
+def test_wait_when_no_datasource(session_cloud: IntegrationCloud, setup_image):
+    """Ensure that when no datasource is found, we get status: disabled
+
+    LP: #1966085
+    """
+    with session_cloud.launch(
+        launch_kwargs={
+            "config_dict": {"security.devlxd": False},
```
holmanb, line 41: I don't think I understand this bit. Are these args the reason that no datasource is found?

TheRealFalcon: I added some comments. Do they answer your question?

holmanb: Perfect, thanks!

tests/integration_tests/clouds.py
```diff
@@ -291,12 +292,15 @@ def _mount_source(instance: LXDInstance):
             subp(command.split())
 
     def _perform_launch(self, launch_kwargs, **kwargs):
-        launch_kwargs["inst_type"] = launch_kwargs.pop("instance_type", None)
-        wait = launch_kwargs.pop("wait", True)
-        release = launch_kwargs.pop("image_id")
+        instance_kwargs = deepcopy(launch_kwargs)
```
blackboxsw, line 295: How did this mutable being modified problem manifest as a failure in our integration tests?  Did the popped 
"wait" end up not being honored or is this now a vestigial change because you added the _wait_for_cloud function in the integration test?

TheRealFalcon: pycloudlib takes the `wait`, but cloud-init also uses the `wait` parameter, which won't get respected if we pop it off the dict. https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/clouds.py#L165

It might be vestigial at this point though given that nothing in that cloud-init code should cause any kind of failure. I added it because I was really confused why my wait kwarg appeared to be disappearing.

tests/integration_tests/cmd/test_status.py
```diff
@@ -0,0 +1,64 @@
...
+        # Jammy and above will use LXD datasource by default
+        if ImageSpecification.from_os_image().release in [
+            "bionic",
+            "focal",
+            "impish",
+        ]:
+            _remove_nocloud_dir_and_reboot(client)
+        status = _wait_for_cloud_init(client)
+        assert status.stdout.strip() == "status: disabled"
+        assert client.execute("cloud-init status --wait").ok
```
blackboxsw, line 64: Minor change suggestion just to make sure our status is reporting the correct reason for the disable
```diff
diff --git a/tests/integration_tests/cmd/test_status.py b/tests/integration_tests/cmd/test_status.py
index 085680154..53bab5fd6 100644
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -13,7 +13,7 @@ def _wait_for_cloud_init(client: IntegrationInstance):
     last_exception = None
     for _ in range(30):
         try:
-            result = client.execute("cloud-init status")
+            result = client.execute("cloud-init status --long")
             if result and result.ok:
                 return result
         except Exception as e:
@@ -59,6 +59,7 @@ def test_wait_when_no_datasource(session_cloud: IntegrationCloud, setup_image):
             "impish",
         ]:
             _remove_nocloud_dir_and_reboot(client)
-        status = _wait_for_cloud_init(client)
-        assert status.stdout.strip() == "status: disabled"
+        status_out = _wait_for_cloud_init(client).stdout.strip()
+        assert "status: disabled" in status_out
+        assert "Cloud-init disabled by cloud-init-generator" in status_out
         assert client.execute("cloud-init status --wait").ok
```

blackboxsw (commented):
LGTM. minor suggestion integration test explicit test. I'm running this patch through now on Jammy and impish to validate

tests/integration_tests/cmd/test_status.py
blackboxsw: Confirmed success with a deb from this PR on impish and jammy runs with the above integration test patch. See expected failure using `CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:cloud-init-dev/proposed` deb without your fix.

blackboxsw (approved):

tests/integration_tests/cmd/test_status.py
TheRealFalcon: Good idea, thanks. Applied your patch.

blackboxsw:
+1 will await CI run on this

blackboxsw:
@valentindavid just FYI this fix is SRU'd to Bionic, Focal and Impish -updates pocket as version 22.1-14-g2e17a0d6-0ubuntu1~20.04.3 so daily cloudimages should have this fix you need.

PR #1353, opened 2022-03-24
testing: Add missing mock to networking test

TheRealFalcon:
Is failing for me locally (but not in CI)

### DIFF
--- a/tests/unittests/test_net_freebsd.py
+++ b/tests/unittests/test_net_freebsd.py
@@ -60,10 +60,11 @@ class TestFreeBSDRoundTrip(CiTestCase):
         renderer.render_network_state(ns, target=target)
         return dir2dict(target)
 
-    @mock.patch("cloudinit.net.get_interfaces_by_mac")
-    @mock.patch("cloudinit.subp.subp")
-    def test_render_output_has_yaml(self, m_get_interfaces, m_subp):
-
+    @mock.patch(
+        "cloudinit.subp.subp", return_value=(SAMPLE_FREEBSD_IFCONFIG_OUT, 0)
+    )
+    @mock.patch("cloudinit.util.is_FreeBSD", return_value=True)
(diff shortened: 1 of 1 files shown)


tests/unittests/test_net_freebsd.py
```diff
@@ -60,8 +60,9 @@ def _render_and_read(
         renderer.render_network_state(ns, target=target)
         return dir2dict(target)
 
+    @mock.patch("cloudinit.net.get_interfaces_by_mac")
     @mock.patch("cloudinit.subp.subp")
-    def test_render_output_has_yaml(self, mock_subp):
+    def test_render_output_has_yaml(self, m_get_interfaces, m_subp):
```
blackboxsw, line 65: Should this instead be the following diff? If not, minimally the order of the mocked params should switch
```diff
diff --git a/tests/unittests/test_net_freebsd.py b/tests/unittests/test_net_freebsd.py
index 3facb2bb4..bf850ceaf 100644
--- a/tests/unittests/test_net_freebsd.py
+++ b/tests/unittests/test_net_freebsd.py
@@ -61,8 +61,10 @@ class TestFreeBSDRoundTrip(CiTestCase):
         return dir2dict(target)
 
     @mock.patch("cloudinit.subp.subp")
-    def test_render_output_has_yaml(self, mock_subp):
+    @mock.patch("cloudinit.util.is_FreeBSD", return_value=True)
+    def test_render_output_has_yaml(self, _is_freebsd, mock_subp):
 
+        mock_subp.return_value = (SAMPLE_FREEBSD_IFCONFIG_OUT, 0)
         entry = {
             "yaml": V1,
         }
```

blackboxsw (changes requested):

tests/unittests/test_net_freebsd.py
TheRealFalcon: Yes, this is better

blackboxsw (approved):
LGTM

PR #1354, opened 2022-03-25
docs: Re-order readthedocs install (SC-884)

TheRealFalcon:
## Proposed Commit Message

```
docs: Re-order readthedocs install

Order building cloud-init before building doc-requirements.txt.
Building '.' invokes a 'pip install --upgrade .' which happens to also
re-install the latest versions of some previously pinned 
docs dependencies

Also updated the yaml to 2-space indents.
```

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,6 +3,6 @@ version: 2
 formats: all
 
 python:
-    install:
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
Nice find analysis and fix.

PR #1358, opened 2022-03-29
schema: add JSON schema for mcollective, migrator and mounts modules            

blackboxsw:
## Proposed Commit Message
```
schema: add JSON schema for mcollective, migrator and mounts modules

Define static JSON schema for cloud-config modules in
cloudinit/config/cloud-init-schema.json. Migrate existing doc/examples
into inline meta schema examples where applicable.

Includes:
 - cc_mcollective
 - cc_migrator
 - cc_mounts

Doc drops:
 - doc/examples/cloud-config-mcollective.txt because example is included in
   inline meta schema docs
```

## Test Steps
```
tox -re doc; xdg-open doc/rtd_html/topics/modules.html 
# validate visual changes vs https://cloudinit.readthedocs.io/en/latest/topics/modules.html#mcollective

# validate specific module rendered docs on cmdline
PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema --docs cc_mcollective
```
Unit tests will run through any included cc_<module>.meta['examples'] to confirm that they are correct schema. 

## Checklist:


cloudinit/config/cloud-init-schema.json
```diff
@@ -783,6 +783,97 @@
...
+                "public-cert": {
+                  "type": "string",
+                  "description": "Optional value of server public certificate which will be written to ``/etc/mcollective/ssl/server-public.pem``"
+                },
+                "private-cert": {
+                  "type": "string",
+                  "description": "Optional value of server private certificate which will be written to ``/etc/mcollective/ssl/server-private.pem``"
+                }
+              },
+              "patternProperties": {
```
TheRealFalcon, line 804: Do `properties` and `patternProperties` co-exist at the same level peacefully? Can we add a test where we set public-cert or private-cert to a boolean or integer and verify it fails to show that the `patternProperties` definition isn't replacing the `properties` definition?

cloudinit/config/cloud-init-schema.json
```diff
@@ -783,6 +783,97 @@
...
+      "properties": {
+        "mounts": {
+          "type": "array",
+          "items": {
+              "type": "array",
+              "items": {"type": "string"},
+              "minItems": 1,
+              "maxItems": 6
+          },
+          "description": "List of lists. Each inner list entry is a list of ``/etc/fstab`` entries of the format: [ fs_spec, fs_file, fs_vfstype, fs_mntops, fs-freq, fs_passno ]",
```
TheRealFalcon, line 841: We should have some mention that all 6 entries are not required.

cloudinit/config/cc_mounts.py
```diff
@@ -42,32 +59,35 @@
...
-        maxsize: <size in bytes>
-"""
-
-import logging
-import os
-import re
-from string import whitespace
-
-from cloudinit import subp, type_utils, util
+        filename: /my/swapfile
```
TheRealFalcon, line 72: In lines 66-72, is it worth removing quotes around options that don't need it (or possibly putting quotes around everything?). I'm just noticing the inconsistency between `auto` and `"auto"`

cloudinit/config/cc_mounts.py
```diff
@@ -42,32 +59,35 @@
 with ``filename``, the size of the swap file with ``size`` maximum size of
 the swap file if using an ``size: auto`` with ``maxsize``. By default no
 swap file is created.
+"""
```
TheRealFalcon, line 62: We should update the blurb mentioning the default:
```
It defaults to::
    mount_default_fields: [none, none, "auto", "defaults,nobootwait", "0", "2"]
On a systemd booted system that default is the mostly equivalent::
    mount_default_fields: [none, none, "auto",
       "defaults,nofail,x-systemd.requires=cloud-init.service", "0", "2"]
Note that `nobootwait` is an upstart specific boot option that somewhat
equates to the more standard `nofail`.
```
Is upstart default anywhere anymore? I think we should just specify the systemd default as the only default and then mention that other init systems may vary in their defaults.

cloudinit/config/cloud-init-schema.json
```diff
@@ -783,6 +783,97 @@
...
+          "minItems": 1
+        },
+        "mount_default_fields": {
+          "type": "array",
+          "description": "Default mount configuration for any mount entry with less than 6 options provided. When specified, 6 items are required and represent ``/etc/fstab`` entries. Systemd-based mount options will be ``defaults,nofail,x-systemd.requires=cloud-init.service,_netdev``",
+          "default": [null, null, "auto", "defaults,nobootwait", "0", "2"],
```
TheRealFalcon, line 847: See my other comment about defaults. I think we should use the systemd default here and not specify other defaults.

cloudinit/config/cloud-init-schema.json
```diff
@@ -783,6 +783,97 @@
...
+        "swap": {
+          "type": "object",
+          "properties": {
+            "filename": {
+              "type": "string",
+              "description": "Path to the swap file to create"
+            },
+            "size": {
+              "description": "The size in bytes of the swap file, or 'auto'",
+              "oneOf": [{"enum": ["auto"]}, {"type": "integer"}]
```
TheRealFalcon, line 866: String is valid here: https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_mounts.py#L378-L381

cloudinit/config/cloud-init-schema.json
```diff
@@ -783,6 +783,97 @@
...
+            },
+            "maxsize": {
+              "type": "integer",
```
TheRealFalcon, line 869: String is valid here: https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_mounts.py#L378-L381

TheRealFalcon (changes requested):
Some comments inline, but overall looking good.

cloudinit/config/cloud-init-schema.json
blackboxsw: Yes they do. I had wondered the same things, which is why we [already have a test for explicit property types like public-cert in the unittests of this PR](https://github.com/canonical/cloud-init/pull/1358/files#diff-226b092ec1d5779ac38368bed2be8bb6a67701d48155e01604039169005906c2R176-R178)

cloudinit/config/cloud-init-schema.json
blackboxsw: added `"pattern": "^([0-9]+)?\\.?[0-9]+[BKMGT]$"}` to schema for swap.size swap.maxsize and unittests

cloudinit/config/cloud-init-schema.json
```diff
@@ -838,13 +838,13 @@
               "minItems": 1,
               "maxItems": 6
           },
-          "description": "List of lists. Each inner list entry is a list of ``/etc/fstab`` entries of the format: [ fs_spec, fs_file, fs_vfstype, fs_mntops, fs-freq, fs_passno ]",
+          "description": "List of lists. Each inner list entry is a list of ``/etc/fstab`` mount declaratiosn of the format: [ fs_spec, fs_file, fs_vfstype, fs_mntops, fs-freq, fs_passno ]. A mount declaration with less than 6 items, will get remaining values from ``mount_default_fields``. A mount declaration with only `fs_spec` and no `fs_file` mountpoint will be skipped.",
```
TheRealFalcon, line 841: `declaratiosn`->`declarations`. Also the comma in the second sentence should be removed.

cloudinit/config/cc_mounts.py
```diff
@@ -45,15 +46,10 @@
...
-
-On a systemd booted system that default is the mostly equivalent::
-
     mount_default_fields: [none, none, "auto",
-       "defaults,nofail,x-systemd.requires=cloud-init.service", "0", "2"]
+        "defaults,nofail,x-systemd.requires=cloud-init.service", "0", "2"]
 
-Note that `nobootwait` is an upstart specific boot option that somewhat
-equates to the more standard `nofail`.
+Other init systems will vary in ``mount_default_fields``.
```
TheRealFalcon, line 52: Nit: might be clearer to change `Other` to `Non-systemd`

TheRealFalcon (changes requested):
One typo we should fix, and everything else optional. For some reason Github wouldn't let me do suggestions on this review.

PR #1362, opened 2022-03-31
tox: add types-* packages to check_format env (SC-898)

TheRealFalcon:
## Proposed Commit Message

```
tox: add types-* packages to check_format env
```

## Test Steps
`tox -e check_format` 
Fails on mypy because the type stubs aren't installed.


blackboxsw (approved):
Confirmed failure case on clean jammy system and this PR resolves the missing deps.

PR #1365, opened 2022-03-31
schema: add json defs for modules scripts-timezone (SC-801)

TheRealFalcon:
## Proposed Commit Message

```
schema: add json defs for modules scripts-timezone

Additionally change module documentation to not show the property
header if no property definition exists.

Migrate legacy or define new schemas for the following:
- cc_scripts_per_boot
- cc_scripts_per_instance
- cc_scripts_per_once
- cc_scripts_user
- cc_scripts_vendor
- cc_seed_random
- cc_set_hostname
- cc_set_passwords
- cc_snap
- cc_spacewalk
- cc_ssh_authkey_fingerprints
- cc_ssh_import_id
- cc_ssh
- cc_timezone
```


cloudinit/config/cc_scripts_per_once.py
```diff
@@ -5,30 +5,34 @@
...
-from cloudinit import subp
-from cloudinit.settings import PER_ONCE
-
-frequency = PER_ONCE
+meta: MetaSchema = {
+    "id": "cc_scripts_per_once",
+    "name": "Scripts Per Once",
+    "title": "Run one time scripts",
+    "description": MODULE_DESCRIPTION,
+    "distros": [ALL_DISTROS],
```
blackboxsw, line 30: I don't think at the moment any of our modules [can avoid defining a module level attribute distros](https://warthogs.atlassian.net/browse/SC-888) because the [cloudinit.config._fixup_modules sets mod.distros = [] when absent](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/__init__.py#L40-L41) and if mod.frequency is not definied cloud-init it'll be set to PER_INSTANCE (so good that you left this PER_ONCE as module-level attr

blackboxsw: Actually strike that on distros comment for modules with a distros = ["all"] case. when a config module doesn't define mod.distros -> fixup_modules sets mod.distros = [] -> [cloudinit.stages only checks/filters distros is distros is non-empty :)](https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L1119-L1129). Ok so modules which claim support of "all" distros can avoid defining mod.distros attribute as it'll be an inactive filter in that case.

cloudinit/config/cc_scripts_vendor.py
```diff
@@ -3,12 +3,17 @@
...
---------------
-**Summary:** run vendor scripts
+import os
+
+from cloudinit import subp, util
+from cloudinit.config.schema import MetaSchema, get_meta_doc
+from cloudinit.distros import ALL_DISTROS
+from cloudinit.settings import PER_INSTANCE
 
+frequency = PER_INSTANCE
```
blackboxsw, line 15: Can drop this as it's unused and cloud-init defaults to PER_INSTANCE if not getattr(mod, "frequency").

cloudinit/config/cc_scripts_vendor.py
```diff
@@ -3,12 +3,17 @@
...
+MODULE_DESCRIPTION = """\
 Any scripts in the ``scripts/vendor`` directory in the datasource will be run
 when a new instance is first booted. Scripts will be run in alphabetical order.
```
blackboxsw, line 18: I think we should update docs to somehow mention the run order of script config modules is [vendor, per-once, per-boot, per-instance, user].

cloudinit/config/cc_scripts_vendor.py
```diff
@@ -26,12 +31,18 @@
         prefix: <vendor data prefix>
```
blackboxsw, line 31: We need cloud-init-schema.json defined for cc_scripts_vendor supporting "vendor_data: prefix". Also there is an errant example specifying the "enable: True" is allow but I see no handling of that here.

blackboxsw (changes requested):

cloudinit/config/cc_set_passwords.py
```diff
@@ -5,85 +5,65 @@
...
+    "examples": [
+        dedent(
+            """\
+            ssh_pwauth: true
+            password: password1
+            """
+        ),
+        dedent(
+            """\
+            ssh_pwauth: false
```
blackboxsw, line 54: Do we want a multi-line contextual header comment in these examples to tell people what the snippet's behavior is?

cloudinit/config/cc_set_passwords.py
```diff
@@ -101,6 +81,12 @@ def handle_ssh_pwauth(pw_auth, distro):
     @return: None"""
     cfg_name = "PasswordAuthentication"
 
+    if isinstance(pw_auth, str):
+        LOG.warning(
+            "DEPRECATION: The 'ssh_pwauth' config key should be set to "
```
blackboxsw, line 86: Thanks for deprecation. it's not effective to support "" and "unchanged" which are equivalent of false.

cloudinit/config/cc_set_passwords.py
```diff
@@ -5,85 +5,65 @@
...
+    "id": "cc_set_passwords",
+    "name": "Set Passwords",
+    "title": "Set user passwords and enable/disable SSH password auth",
+    "description": MODULE_DESCRIPTION,
+    "distros": [ALL_DISTROS],
+    "frequency": PER_INSTANCE,
+    "examples": [
+        dedent(
+            """\
+            ssh_pwauth: true
```
blackboxsw, line 48: In this example we probably also want to include the following and document what this means as anyone setting their passwords and trying to use them would also probably not want them to auto expire on them :)
`chpasswd: { expire: False }`


and document that we 
As

blackboxsw (changes requested):
almost done w/ cc_set_passwords but wanted to add comments before I EOW. cc_snap looks insteresting as I wonder if we want to use patternProperties schema to handle reprensenting the underlying lists below the opaque-keys. will get back to this monday.

cloudinit/config/cc_scripts_vendor.py
TheRealFalcon: If one cares about ordering, they would need to check /etc/cloud/cloud.cfg, and I think we need to find a way to document that separately. I don't disagree that it should be documented, but I'm not sure the module documentation is the best place for it.

cloudinit/config/cc_scripts_vendor.py
TheRealFalcon: Thanks...I think I skipped over vendor data to understand it better later, and then forgot to come back to it later :man_facepalming:

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,282 @@
...
+              "description": "File to write random data to. Default: ``/dev/urandom``"
+            },
+            "data": {
+              "type": "string",
+              "description": "This data will be written to ``file`` before data from the datasource. When using a multiline value or specifying binary data, be sure to follow yaml syntax and use the ``|`` and ``!binary`` yaml format specifiers when appropriate"
+            },
+            "encoding": {
+              "type": "string",
+              "enum": ["raw", "base64", "b64", "gzip", "gz"],
+              "description": "Used to decode ``data`` provided. Allowed values are ``raw``, ``base64``, ``b64``, ``gzip``, or ``gz``.  Default is ``raw``"
```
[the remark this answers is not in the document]
TheRealFalcon: Good idea.

cloudinit/config/cc_set_hostname.py
```diff
@@ -5,24 +5,36 @@
...
-force the hostname use.
+the ``FQDN`` config key. If both ``FQDN`` and ``hostname`` are set,
+the ``prefer_fqdn_over_hostname`` will force the use of FQDN in all distros
+if true, and if false it will force the hostname use. Otherwise, the hostname
+to use is distro-dependent.
+
+.. note::
+    cloud-init performs no hostname input validation before sending the
+    hostname to distro-specific tools, and most tools will not accept a
+    trailing dot on the FQDN.
```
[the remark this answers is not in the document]
TheRealFalcon: Got a bug report a few days ago about this

cloudinit/config/cc_set_hostname.py
```diff
@@ -31,25 +43,28 @@
...
+    "name": "Set Hostname",
+    "title": "Set hostname and FQDN",
+    "description": MODULE_DESCRIPTION,
+    "distros": [ALL_DISTROS],
+    "frequency": frequency,
+    "examples": [
+        "perserve_hostname: true",
+        dedent(
+            """\
+            hostname: myhost
```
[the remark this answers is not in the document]
TheRealFalcon: Isn't a lot of it technically distro-dependent? Also the interaction between this module and update_hostname is still confusing to me and I think it could probably be simplified down into one module...but not here.

cloudinit/config/cc_set_passwords.py
TheRealFalcon: Eh...I think it is fine. You can `ssh ubuntu@<ip>`, enter `password1`, then it will prompt you to change the password. Having a default password for new users that need to be changed right away seems like a common enough use case...assuming you're not using keys which is the unreasonable thing ;)

I added a comment explaining that though.

TheRealFalcon:
Pushed 2 commits (one just for vendor data) addressing comments

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+      }
+    },
+    "cc_ssh": {
+      "type": "object",
+      "properties": {
+        "ssh_keys": {
+          "type": "object",
+          "description": "A dictionary entries for the public and private host keys of each desired key type. Entries in the ``ssh_keys`` config dict should have keys in the format ``<key type>_private``, ``<key type>_public``, and, optionally, ``<key type>_certificate``, e.g. ``rsa_private: <key>``, ``rsa_public: <key>``, and ``rsa_certificate: <key>``. Not all key types have to be specified, ones left unspecified will not be used. If this config option is used, then separate keys will not be automatically generated. In order to specify multiline private host keys and certificates, use yaml multiline syntax.",
+          "patternProperties": {
+            "^(rsa|dsa|ecdsa|ed25519)_(public|private|certificate)$": {
```
blackboxsw, line 1566: This regex looks a lot more strict that what our cc_ssh module documents as supported. I'll have to reread this content again to better understand whether this incorrectly excludes config directives.

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+              "label": "<key_type>",
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        "ssh_authorized_keys": {
+          "type": "array",
+          "minItems": 1,
+          "description": "The SSH public keys to add ``.ssh/authorized_keys`` in user's home directory",
```
blackboxsw, line 1576:
```suggestion
          "description": "The SSH public keys to add ``.ssh/authorized_keys`` in the ``default`` user's home directory",
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+          "description": "Remove host SSH keys. This prevents re-use of a private host key from an image with default host SSH keys. Default: ``true``"
+        },
+        "ssh_genkeytypes": {
+          "type": "array",
+          "description": "The SSH key types to generate. Default: ``[rsa, dsa, ecdsa, ed25519]``",
+          "default": ["rsa", "dsa", "ecdsa", "ed25519"],
+          "minItems": 1,
+          "items": {
+            "type": "string",
+            "enum": ["rsa", "dsa", "ecdsa", "ed25519"]
```
blackboxsw, line 1593: Newer versions of `ssh-keygen` support the following key types `dsa | ecdsa | ecdsa-sk | ed25519 | ed25519-sk | rsa`  so I think our enum should also permit these values as cloud-init passes them directly to ssh-keygen.

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+          }
+        },
+        "disable_root": {
+          "type": "boolean",
+          "default": true,
+          "description": "Disable root login. Default: ``true``"
+        },
+        "disable_root_opts": {
+          "type": "string",
```
blackboxsw, line 1602: Let's add the default: value here too.

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
+              "type": "boolean",
+              "default": true,
+              "description": "If true, will read host keys from /etc/ssh/*.pub and publish them to the datasource (if supported). Default: ``true``"
+            },
+            "blacklist": {
+              "type": "array",
+              "description": "The SSH key types to blacklist when publishing. Default: ``[dsa]``",
```
blackboxsw, line 1626:
```suggestion
              "description": "The SSH key types to ignore when publishing. Default: ``[dsa]``",
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+          }
+        }
+      }
+    },
+    "cc_timezone": {
+      "type": "object",
+      "properties": {
+        "timezone": {
+          "type": "string",
+          "description": "The timezone to use"
```
blackboxsw, line 1640:
```suggestion
          "description": "The timezone to use as represented in /usr/share/zoneinfo."
```

cloudinit/stages.py
```diff
@@ -757,6 +757,12 @@ def _consume_vendordata(self, vendor_source, frequency=PER_INSTANCE):
             LOG.debug("%s consumption is disabled.", vendor_source)
             return
 
+        if isinstance(enabled, str):
+            LOG.debug(
+                "Use of string for 'vendor_data:enabled' field is "
```
blackboxsw, line 762: Let's add the value here for debug/triage purposes.
```suggestion
                f"Use of string '{enabled}' for 'vendor_data:enabled' field is "
```

doc/examples/cloud-config-seed-random.txt
```diff
@@ -1,32 +0,0 @@
...
-#             the command will have RANDOM_SEED_FILE in its environment
-#             set to the value of 'file' above.
-#   command_required: default False
-#             if true, and 'command' is not available to be run
-#             then exception is raised and cloud-init will record failure.
-#             Otherwise, only debug error is mentioned.
-#
-# Note: command could be ['pollinate',
-#                         '--server=http://local.pollinate.server']
-# which would have pollinate populate /dev/urandom from provided server
```
blackboxsw, line 26: I don't like losing documented use-cases from examples and not representing them in module.Meta.examples. Given that the examples page is the page with the most daily hits, I don't want to diminish the value of that page without providing supplemental/comparable examples somewhere else. Please include a supplemental example that explains this use-case in cc_seed_random.

blackboxsw (changes requested):
I need to come back to cc_ssh but wanted to give you a second round of feedback first.

doc/examples/cloud-config-seed-random.txt
blackboxsw: Heh I just re-looked at full rendered docs and see that this example is already in cc_seed_random's MODULE_DESCRIPTION. Maybe we can just move that example and introductory header comment down into Meta['examples']

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,282 @@
...
+              "default": "/dev/urandom",
+              "description": "File to write random data to. Default: ``/dev/urandom``"
+            },
+            "data": {
+              "type": "string",
+              "description": "This data will be written to ``file`` before data from the datasource. When using a multiline value or specifying binary data, be sure to follow yaml syntax and use the ``|`` and ``!binary`` yaml format specifiers when appropriate"
+            },
+            "encoding": {
+              "type": "string",
+              "enum": ["raw", "base64", "b64", "gzip", "gz"],
```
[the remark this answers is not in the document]
TheRealFalcon: I'm not sure what you're asking for. I already set default as `raw`, and that is the default of the module

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,303 @@
...
+              "additionalProperties": {
+                "oneOf": [
+                  {"type": "string"},
+                  {"type": "array", "items": {"type": "string"}}
+                ]
+              }
+            },
+            "squashfuse_in_container": {
+              "type": "boolean",
+              "description": "**Development only**: The ``squashfuse_in_container`` boolean can be set true to install squashfuse package when in a container to enable snap installs. Default: ``false``."
```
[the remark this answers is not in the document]
TheRealFalcon: Created https://warthogs.atlassian.net/browse/SC-905

cloudinit/config/cloud-init-schema.json
TheRealFalcon: Once we add the ability to hide schema properties from the docs, I'd like to deprecate this and replace it with 'denylist'

cloudinit/config/cloud-init-schema.json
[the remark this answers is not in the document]
blackboxsw: Duplicate of https://warthogs.atlassian.net/browse/SC-903

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,305 @@
...
+        "vendor_data": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "enabled": {
+              "type": ["boolean", "string"],
+              "description": "Whether vendor data is enabled or not. Use of string for this value is DEPRECATED. Default: ``true``"
+            },
+            "prefix": {
+              "type": ["boolean", "string"],
```
blackboxsw, line 1359: I think we added wrong "types" in the latest commit for prefix it should be a 
```suggestion
              "type": ["array", "string"],
              "items": {"type": ["string", "integer"]},
              "minItems": 1,
```

cloudinit/config/cloud-init-schema.json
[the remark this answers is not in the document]
TheRealFalcon: Ahhh, thanks. Deleted

blackboxsw (commented):
Minor schema fix with vendor_data.prefix otherwise LGTM!

```diff
diff --git a/cloudinit/config/cloud-init-schema.json b/cloudinit/config/cloud-init-schema.json
index 43e90e1a4..96dec0a7c 100644
--- a/cloudinit/config/cloud-init-schema.json
+++ b/cloudinit/config/cloud-init-schema.json
@@ -1356,7 +1356,9 @@
               "description": "Whether vendor data is enabled or not. Use of string for this value is DEPRECATED. Default: ``true``"
             },
             "prefix": {
-              "type": ["boolean", "string"],
+              "type": ["array", "string"],
+              "items": {"type": ["string", "integer"]},
+              "minItems": 1,
               "description": "The command to run before any vendor scripts. Its primary use case is for profiling a script, not to prevent its run"
             }
           }
diff --git a/tests/unittests/config/test_cc_scripts_vendor.py b/tests/unittests/config/test_cc_scripts_vendor.py
index a8cbfb4f4..69a5fe801 100644
--- a/tests/unittests/config/test_cc_scripts_vendor.py
+++ b/tests/unittests/config/test_cc_scripts_vendor.py
@@ -1,3 +1,5 @@
+import re
+
 import pytest
 
 from cloudinit.config.schema import (
@@ -14,6 +16,10 @@ class TestScriptsVendorSchema:
         (
             ({"vendor_data": {"enabled": True}}, None),
             ({"vendor_data": {"enabled": "yes"}}, None),
+            (
+                {"vendor_data": {"prefix": []}},
+                re.escape("vendor_data.prefix: [] is too short"),
+            ),
         ),
     )
     @skipUnlessJsonSchema()
```

cloudinit/config/cloud-init-schema.json
```diff
@@ -1343,6 +1343,305 @@
...
+      }
+    },
+    "cc_ssh": {
+      "type": "object",
+      "properties": {
+        "ssh_keys": {
+          "type": "object",
+          "description": "A dictionary entries for the public and private host keys of each desired key type. Entries in the ``ssh_keys`` config dict should have keys in the format ``<key type>_private``, ``<key type>_public``, and, optionally, ``<key type>_certificate``, e.g. ``rsa_private: <key>``, ``rsa_public: <key>``, and ``rsa_certificate: <key>``. Not all key types have to be specified, ones left unspecified will not be used. If this config option is used, then separate keys will not be automatically generated. In order to specify multiline private host keys and certificates, use yaml multiline syntax.",
+          "patternProperties": {
+            "^(dsa|ecdsa|ecdsa-sk|ed25519|ed25519-sk|rsa)_(public|private|certificate)$": {
```
blackboxsw, line 1567: This schema looks ok,  but the corresponding code in [[cc_ssh.py will silently skip any key type that is not in GENERATE_KEY_NAMES](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_ssh.py#L205-L206)

I think we need to properly handle these other keys and integration testing of them to make sure all is well. Since it's ssh-related I'd like that work to be a separate PR to make sure we get good eyes on it separate from schema coverage

TheRealFalcon: I'll make a ticket

cloudinit/config/cloud-init-schema.json
blackboxsw: @TheRealFalcon CI is failing on this invalid schema definition I believe. Otherwise I think we are good to go.

TheRealFalcon: Gah, sorry. I made the change but never pushed it.

TheRealFalcon:
Fixed the CI issues. Note that this includes a small removal to the vendordata schema. On default LXD launch I'm getting
```
2022-04-08 13:08:59,487 - schema.py[WARNING]: Invalid cloud-config provided:
vendor_data.prefix: [] is too short
```

so we don't want a minItems defined on that.

blackboxsw (approved):
Bring in home!

sshedi:
@TheRealFalcon I am seeing something weird after this change.

My `/etc/hosts` looks like this:
```
$ cat /etc/hosts 
127.0.1.1    ph3dev.local ph3dev
127.0.0.1    localhost
```
And my `/etc/hostname`
```
ph3dev
```

Now if I do `systemctl restart cloud-init` , `/etc/hostname` is getting changed.
```
$ cat /etc/hostname 
ph3dev.local
```

And `hostname` command return fqdn instead of short name. Isn't it incorrect to have fqdn in `/etc/hostname`?

holmanb:
> TheRealFalcon I am seeing something weird after this change.

The changes to the hostname module appear docs related. Did you bisect to this commit specifically?

> Isn't it incorrect to have fqdn in /etc/hostname?

This is treated as distro-dependant if `prefer_fqdn_over_hostname` is unset. We would need to know what is in your user-data to know what behavior to expect.

@sshedi Please file a bug report with logs if you still believe this to be an issue.

sshedi:
Thanks for the response @holmanb.

I changed  `frequency = PER_ONCE` in `cc_set_hostname.py` and it doesn't modify `/etc/hostname` after that and this is how it was in `cloud-init <= v22.1` and even in `v22.1` we have `prefer_fqdn_over_hostname` set in `photon.py` but it doesn't modify `/etc/hostname` file.

Sure, I will file a bug report with necessary data.

holmanb:
> Sure, I will file a bug report with necessary data.

@sshedi Excellent, thank you.

FWIW, I do see `prefer_fqdn = True` in photon.py, so I would expect cloud-init to set the hostname with a fqdn (see how this changes behavior in hostname setting inside of `cloudinit/distros/__init__.py:Distro`). Let me know if I missed anything.

sshedi:
I have created bug with necessary info https://bugs.launchpad.net/cloud-init/+bug/1983811

Please let me know if you need more info.

PR #1369, opened 2022-04-06
Ubuntu/devel 22.1 hotfix

TheRealFalcon:
Don't squash merge.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,12 +1,15 @@
-cloud-init (22.1-14-g2e17a0d6-0ubuntu1~22.04.4ubuntu2) UNRELEASED; urgency=medium
+cloud-init (22.1-14-g2e17a0d6-0ubuntu1~22.04.4ubuntu3) UNRELEASED; urgency=medium
 
   * d/p/cpick-be9389c6-Work-around-bug-in-LXD-VM-detection-1325:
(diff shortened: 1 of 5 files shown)


blackboxsw (changes requested):
I'd like to block this upload on another corner case support for LXD.
When launching Jammy kvm instances from a Jammy host we'll get into the `systemd-detect-virt`  "qemu" case instead of "kvm". To avoid ds-id disabling those VM hosts, it also needs to grow awareness of "qemu". Once systemd 251 is out on our support matrix, we can drop that distinction as `hv_passthrough` qemu options won't generate incorrect virt type.

https://github.com/canonical/cloud-init/pull/1370

blackboxsw (changes requested):
I'll probably just force push the minor changelog changes to your branch and dput from there later today.

No big changes other than readable text on last changelog item and version correction `22.04.5`

quilt push -a; tox -e py; quilt pop -a success
build-package success
sbuild success no new debian packaging warnings errors errors
upload to my own PPA shows sensible diff https://launchpadlibrarian.net/595282136/cloud-init_22.1-14-g2e17a0d6-0ubuntu1~22.04.4_22.1-14-g2e17a0d6-0ubuntu1~22.04.5.diff.gz

TheRealFalcon:
Thanks, your change looks good. Failure is a docs failure which should be fine for a hotfix branch.

PR #1373, opened 2022-04-09
CloudStack Tests Occasionally Take a Long Time

holmanb:
```
Mock sleep in CloudStack test 
```

## Additional Context
I've seen this happen plenty of times, but I don't know where it's comming from and it's not reproducible. Think it would be alright to just mock the likely culprit?

```
15.18s call     tests/unittests/sources/test_cloudstack.py::TestCloudStackPasswordFetching::test_password_not_saved_if_bad_request
10.35s call     tests/unittests/sources/test_cloudstack.py::TestCloudStackPasswordFetching::test_valid_response_means_password_marked_as_saved
10.32s call     tests/unittests/sources/test_cloudstack.py::TestCloudStackPasswordFetching::test_saved_password_doesnt_create_config
10.22s call     tests/unittests/sources/test_cloudstack.py::TestCloudStackPasswordFetching::test_password_sets_password
10.22s call     tests/unittests/sources/test_cloudstack.py::TestCloudStackPasswordFetching::test_password_not_saved_if_empty
```

Suspecting that `wait_for_metadata()` is occasionally is unmocked somehow (again, not sure why/how) and we hit default timeout values here: https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L189

Thoughts?

### DIFF
--- a/tests/unittests/sources/test_cloudstack.py
+++ b/tests/unittests/sources/test_cloudstack.py
@@ -67,6 +67,7 @@ class TestCloudStackPasswordFetching(CiTestCase):
         ds.get_data()
         self.assertEqual({}, ds.get_config_obj())
 
+    @mock.patch("cloudinit.url_helper.time.sleep")
     @mock.patch(DS_PATH + ".wait_for_metadata_service")
     def test_password_sets_password(self, m_wait):
         m_wait.return_value = True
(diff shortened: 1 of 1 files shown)


holmanb:
hmmm, tests fail and it would be better to understand the root of the problem anyways. Closing for now.

PR #1374, opened 2022-04-10
Mypy stubs and other tox maintenance (SC-920)

holmanb:
```
Mypy stubs and other tox maintenance

 - add type stubs for mypy: jinja, jsonschema, oauthlib
 - specify mypy directories to avoid env/ false positive
 - add conftest.py and setup.py
 - make setup.py pass pylint:
   - check ./tools/render-cloudcfg return code in setup.py
   - find_sources() has no return code, don't check for one
```

find_sources has never had a return code: https://github.com/pypa/setuptools/commit/873ebe91a8de43df836c873d9106e5af31c302e4

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,6 @@ isort==5.10.1
 mypy==0.931
 pylint==2.11.1
 pytest==7.0.0
-types-Jinja2==2.11.9
 types-jsonschema==4.4.2
 types-oauthlib==3.1.6
 types-PyYAML==6.0.4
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -16,21 +16,24 @@ isort==5.10.1
 mypy==0.931
 pylint==2.11.1
 pytest==7.0.0
+types-Jinja2==2.11.9
```
TheRealFalcon, line 19: I think this one is wrong. Our check and do format jobs run against latest pip rather than minimum supported, and latest pip for 3.6 installs 3.0.3.

According to https://pypi.org/project/types-Jinja2/ , 
> Note: The Jinja2 package includes type annotations or type stubs since version 3.0. Please uninstall the types-Jinja2 package if you use this or a newer version.

TheRealFalcon (changes requested):
I think we need to remove all the `types-Jinja2` references, but otherwise looks good!

holmanb:
> I think we need to remove all the `types-Jinja2` references, but otherwise looks good!

@TheRealFalcon Thanks for the review! Fixed.

PR #1376, opened 2022-04-12
cloudinit.sub.py: Return a namedtuple

holmanb:
```
Return a namedtuple from subp()

This provides a minor readability improvement.

 subp.subp(cmd)[0] -> subp.subp(cmd).stdout
 subp.subp(cmd)[1] -> subp.subp(cmd).stderr
```

## Additional Context
Named tuples are an "easy win" in terms of readability. I'm open to suggestions if someone would rather a different name than `PipeOutput`.

### DIFF
--- a/cloudinit/gpg.py
+++ b/cloudinit/gpg.py
@@ -41,7 +41,7 @@ def dearmor(key):
 
     note: man gpg(1) makes no mention of an --armour spelling, only --armor
     """
-    return subp.subp(["gpg", "--dearmor"], data=key, decode=False)[0]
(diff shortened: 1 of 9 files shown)


TheRealFalcon (approved):
> Named tuples are an "easy win" in terms of readability. I'm open to suggestions if someone would rather a different name than PipeOutput.

image


How about `SubpResult` as a name?

holmanb:
> How about `SubpResult` as a name?

@TheRealFalcon that works - updated accordingly

PR #1380, opened 2022-04-13
Fix KeyError when rendering sysconfig IPv6 routes (SC-776)

TheRealFalcon:
## Proposed Commit Message

```
Fix KeyError when rendering sysconfig IPv6 routes

Route rendering code was expecting a netmask rather than using the
prefix. A prefix is provided to the renderer, but was being hidden
from the route rendering code. This commit exposes the prefix and
prefers it for IPv6, given how uncommon netmasks are for IPv6.

LP: #1958506
```

## Additional Context
See https://bugs.launchpad.net/cloud-init/+bug/1958506

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -4,17 +4,18 @@ import copy
 import io
 import os
 import re
+from typing import Mapping
 
 from cloudinit import log as logging
 from cloudinit import subp, util
 from cloudinit.distros.parsers import networkmanager_conf, resolv_conf
 from cloudinit.net import (
     IPV6_DYNAMIC_TYPES,
-    ipv6_mask_to_net_prefix,
     is_ipv6_addr,
     net_prefix_to_ipv4_mask,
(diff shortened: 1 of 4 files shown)


cloudinit/net/sysconfig.py
```diff
@@ -635,12 +636,9 @@ def _render_subnet_routes(cls, iface_cfg, route_cfg, subnets, flavor):
...
                         )
-                    # NOTE(harlowja): ipv6 and ipv4 default gateways
-                    gw_key = "GATEWAY0"
-                    nm_key = "NETMASK0"
-                    addr_key = "ADDRESS0"
-                    # The owning interface provides the default route.
-                    #
+                    # NOTE that instead of defining the route0 settings,
+                    # the owning interface provides the default route.
+
```
TheRealFalcon, line 641: Note that the 3 keys removed here were never used at all. I think linting previously never complained about it because they were also defined and then used in the else clause.

holmanb (approved):
LGTM, thanks!

PR #1381, opened 2022-04-15
Fix ds-identify not detecting NoCloud seed in config (SC-542)

TheRealFalcon:
## Proposed Commit Message

```
Fix ds-identify not detecting NoCloud seed in config

NoCloud seed config can be defined in /etc/cloud/cloud.cfg[.d].
However, ds-identify had no means of detecting this config and reported
NOT FOUND. This commit allows ds-identify to detect and report it
properly.

LP: #1876375
```

## Additional Context
LP: #1876375 notes that if you specify the datasource list in a cfg file, it works, but if you don't it doesn't. It has nothing to do with the reading of the config files. If the datasource list only contains one entry (or one + None), ds-identify skips datasource detection entirely and just uses the specified datasource. https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L1793-L1798 . The problem is that ds-identify just had no means of detecting this use case.

## Checklist:

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -851,6 +851,11 @@ dscheck_NoCloud() {
         return ${DS_FOUND}
     fi
 
+    # This is a bit hacky, but a NoCloud false positive isn't the end of the world
+    if check_config "NoCloud" && check_config "user-data" && check_config "meta-data"; then
+        return ${DS_FOUND}
+    fi
+
     return ${DS_NOT_FOUND}
 }
 
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -1162,11 +1162,11 @@ def read_runtime_config():
 def fetch_base_config():
     return util.mergemanydict(
         [
-            # builtin config
+            # builtin config, hardcoded in settings.py.
             util.get_builtin_cfg(),
             # Anything in your conf.d or 'default' cloud.cfg location.
             util.read_conf_with_confd(CLOUD_CONFIG),
-            # runtime config
+            # runtime config. I.e., /run/cloud-init/cloud.cfg
             read_runtime_config(),
             # Kernel/cmdline parameters override system config
             util.read_conf_from_cmdline(),
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -3,6 +3,7 @@
 import copy
 import os
 from collections import namedtuple
+from textwrap import dedent
 from uuid import uuid4
 
 from cloudinit import safeyaml, subp, util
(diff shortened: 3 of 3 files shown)


tools/ds-identify
```diff
@@ -851,6 +851,11 @@ dscheck_NoCloud() {
         return ${DS_FOUND}
     fi
 
+    # This is a bit hacky, but a NoCloud false positive isn't the end of the world
+    if check_config "NoCloud" && check_config "user-data" && check_config "meta-data"; then
```
holmanb, line 855: `user-data` and `meta-data` are [required seed files](https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html) for NoCloud, but it looks like `check_config` searches `/etc/cloud/cloud.cfg` and `/etc/cloud/cloud.cfg.d/*` for keys with this name (am I reading that right?). Since I don't see those as keys in the doc, could you expand on why this works?

TheRealFalcon: Yeah, this functionality needs to be documented better. I just learned of this behavior due to the bug report :smile:  The only documentation is [an example](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#configure-data-sources). The magic happens [in the datasource](https://github.com/canonical/cloud-init/blob/43c1cf433281844c30f0e4a19a938f4459bfccb7/cloudinit/sources/DataSourceNoCloud.py#L113). 

For the ds-identify piece, `check_config` just looks for a key at any nesting of the config (thanks bash). I'm basically just assuming that if 'NoCloud', 'user-data', and 'meta-data' exist somewhere in the config, then the datasource definition is happening there. If we're wrong, we waste a few cycles in python code checking for NoCloud.

TheRealFalcon: E.g., write create 
/etc/cloud/cloud.cfg.d/99_ds.cfg:
```yaml
datasource_list: ["NoCloud"]
datasource:
  NoCloud:
    user-data:  |
      #cloud-config
      runcmd: 
      - echo 'hi' > /var/tmp/hi 
    meta-data:
      instance-id: i-87018aed
      local-hostname: myhost.internal
```

Then `cloud-init clean --logs --reboot` and you should see those details reflected in your instance. With this branch, you shouldn't need the `datasource_list` line.

holmanb: Aha! Thanks. Yes this seems like something that should be documented better. I'm indifferent whether we should update the docs in this PR or a followup.

holmanb (approved):

tools/ds-identify
TheRealFalcon: Already in our docs tracking ticket

PR #1385, opened 2022-04-15
Use cc_* module meta defintion over hardcoded vars (SC-888)

TheRealFalcon:
## Proposed Commit Message

```
Use cc_* module meta defintion over hardcoded vars

Previously, each cc_* module required a hardcoded 'distro' and
'frequency' variable to control the respective distro to run on and
run frequency. If undefined or invalid, we silently changed this to
a default value.

Instead, this commit will validate the MetaSchema definition and raise
an exception if invalid. This means every module MUST contain a
MetaSchema definition.

Additionally, the Modules class was moved out of stages.py into its
own module, along with some helper functions in config/__init__.py.
```

## Additional Context
Review the commits in order. One of the commits is a big cut and paste (file moves) with no changes, so easiest to review the changes separately after that.


TheRealFalcon:
I need to add a test to ensure things fail when a module doesn't have the required MetaSchema, but otherwise this PR is reviewable as-is.

cloudinit/config/cc_refresh_rmc_and_interface.py
```diff
@@ -25,23 +25,29 @@
...
 frequency = PER_ALWAYS
 
+# This module is undocumented
+meta: MetaSchema = {
+    "id": "cc_refresh_rmc_and_interface",
+    "name": "Refresh IPv6 Interface and RMC",
+    "title": "Ensure Network Manager is not managing IPv6 interface",
+    "description": __doc__,
+    "distros": [ALL_DISTROS],
+    "frequency": PER_ALWAYS,
```
blackboxsw, line 47: Might want to still have the local 'frequency' variable in case someone changes that value and doesn't notice this one in the future.
```suggestion
    "frequency": frequency,
```

cloudinit/config/modules.py
```diff
@@ -35,28 +38,31 @@ def form_module_name(name):
...
+    ):
+        raise ValueError(
+            f"Module '{mod}' with name '{name}' MUST have a 'meta' attribute "
+            "of type 'MetaSchema'."
+        )
+    if mod.meta["frequency"] not in FREQUENCIES:
+        raise ValueError(
+            f"Module '{mod}' with name '{name}' has an invalid frequency "
+            f"{mod.meta['frequency']}."
+        )
```
blackboxsw, line 55: Is it worth adding the following?
```suggestion
        )
    if hasattr(mod, "schema"):
        raise ValueError(f"Module '{mod}' with name '{name}' has a JSON "schema" attribute defined. Please define schema in cloud-init-schema,json.")
```

cloudinit/config/modules.py
```diff
@@ -35,28 +38,31 @@ def form_module_name(name):
...
+    if (
+        not hasattr(mod, "meta")
+        or "frequency" not in mod.meta
+        or "distros" not in mod.meta
+    ):
+        raise ValueError(
+            f"Module '{mod}' with name '{name}' MUST have a 'meta' attribute "
+            "of type 'MetaSchema'."
+        )
+    if mod.meta["frequency"] not in FREQUENCIES:
```
blackboxsw, line 51: +1 noisy errors in misconfigured modules instead of silent DEFAULT to PER_INSTANCE. Strict behavior for the win.

cloudinit/config/modules.py
```diff
@@ -129,6 +144,22 @@ def _read_modules(self, name):
         return module_list
 
     def _fixup_modules(self, raw_mods):
```
blackboxsw, line 145: Worth typing here to give concrete validation of return values? Maybe a pain for this at the moment.

cloudinit/config/cc_refresh_rmc_and_interface.py
```diff
@@ -25,23 +25,29 @@
...
 
 from cloudinit import log as logging
 from cloudinit import netinfo, subp, util
+from cloudinit.config.schema import MetaSchema
+from cloudinit.distros import ALL_DISTROS
 from cloudinit.settings import PER_ALWAYS
 
 frequency = PER_ALWAYS
 
+# This module is undocumented
```
blackboxsw, line 40: Not true :)

I think we probably want to filter out cmdline docs for modules which have "meta" but don't have specific schema defined, or we want a "hidden" flag. Otherwise we get [docs for them on the commandline because the module contains a 'meta' attr](via https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schema.py#L709).


```
+Write Files Deferred
+------------
+**Summary:** Defer writing certain files
+
+This module is based on `'Write Files' <write-files>`__, and
+will handle all files from the write_files list, that have been
+marked as deferred and thus are not being processed by the
+write-files module.
+
+*Please note that his module is not exposed to the user through
+its own dedicated top-level directive.*
+
+Reset RMC
+------------
+**Summary:** reset rsct node id
+
+Reset RMC module is IBM PowerVM Hypervisor specific
+
+Reliable Scalable Cluster Technology (RSCT) is a set of software components,
+that  together provide a comprehensive clustering environment (RAS features)
+for IBM PowerVM based virtual machines. RSCT includes the Resource monitoring
+and control (RMC) subsystem. RMC is a generalized framework used for managing,
+monitoring, and manipulating resources. RMC runs as a daemon process on
+individual machines and needs creation of unique node id and restarts
+during VM boot.
+More details refer
+https://www.ibm.com/support/knowledgecenter/en/SGVKBA_3.2/admin/bl503_ovrv.htm
+
+This module handles
+- creation of the unique RSCT node id to every instance/virtual machine
+  and ensure once set, it isn't changed subsequently by cloud-init.
+  In order to do so, it restarts RSCT service.
+
+Prerequisite of using this module is to install RSCT packages.
+
 Mounts
 ------
 **Summary:** Configure mount points and swap files
@@ -2844,6 +2879,27 @@
         maxsize: 2G
     
 
+Refresh IPv6 interface and RMC
+------------------------------
+**Summary:** Ensure Network Manager is not managing IPv6 interface
+
+This module is IBM PowerVM Hypervisor specific
+
+Reliable Scalable Cluster Technology (RSCT) is a set of software components
+that together provide a comprehensive clustering environment(RAS features)
+for IBM PowerVM based virtual machines. RSCT includes the Resource
+Monitoring and Control (RMC) subsystem. RMC is a generalized framework used
+for managing, monitoring, and manipulating resources. RMC runs as a daemon
+process on individual machines and needs creation of unique node id and
+restarts during VM boot.
+More details refer
+https://www.ibm.com/support/knowledgecenter/en/SGVKBA_3.2/admin/bl503_ovrv.htm
+
+This module handles
+- Refreshing RMC
+- Disabling NetworkManager from handling IPv6 interface, as IPv6 interface
+  is used for communication between RMC daemon and PowerVM hypervisor.
+
 Foo
 ---
 **Summary:** Example module
```

cloudinit/config/cc_write_files_deferred.py
```diff
@@ -44,6 +49,3 @@ def handle(name, cfg, _cloud, log, _args):
         )
         return
     write_files(name, filtered_files)
-
-
-# vi: ts=4 expandtab
```
blackboxsw, line 49: Yeah good idea throughout. +1 if you wanted a commit to clear that out in general.

cloudinit/config/cc_write_files_deferred.py
```diff
@@ -2,31 +2,36 @@
...
+from cloudinit.distros import ALL_DISTROS
+from cloudinit.settings import PER_INSTANCE
+
+# This module isn't documented
+meta: MetaSchema = {
+    "id": "cc_write_files_deferred",
+    "name": "Write Files Deferred",
+    "title": "Defer writing certain files",
+    "description": __doc__,
+    "distros": [ALL_DISTROS],
```
blackboxsw, line 31: If you feel excited about cleaning out cruft, we could also drop any  `distros= ["all"]` module attributes use ALL_DISTROS instead
```
$ fgrep -r "distro" cloudinit/config/cc_* | grep '"all"'
cloudinit/config/cc_bootcmd.py:distros = ["all"]
cloudinit/config/cc_chef.py:distros = ["all"]
cloudinit/config/cc_keys_to_console.py:distros = ["all"]
cloudinit/config/cc_locale.py:distros = ["all"]
cloudinit/config/cc_mcollective.py:distros = ["all"]
cloudinit/config/cc_migrator.py:distros = ["all"]
cloudinit/config/cc_mounts.py:distros = ["all"]
cloudinit/config/cc_update_etc_hosts.py:distros = ["all"]
cloudinit/config/cc_update_hostname.py:distros = ["all"]
cloudinit/config/cc_users_groups.py:    "distros": ["all"],
cloudinit/config/cc_write_files.py:    "distros": ["all"],
```

blackboxsw (commented):
Generally looks great. I like actually creating a separate cohesive cloudinit.config.modules.  Couple change suggeestions and supplements.

blackboxsw:
The only real fix I think we need here is making sure cloud-init devel schema --docs all doesn't render documentation for modules without schema.

TheRealFalcon:
Addressed the comments as a separate commit (but force pushed due to rebase).

I'm not a huge fan of setting `__doc__ = ""`, but our schema docs are fundamentally based on `__doc__` and I don't think we should change that in this branch. I'm wondering if it's worth documenting every single module, even if there's no configuration involved. We do this for a handful of other modules (e.g., [scripts-per-boot](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#scripts-per-boot) ), so is it worth providing these modules in the docs as well?

blackboxsw (approved):
Thanks for the cleanup. LGTM. Docs are clean and migration to a separate module looks good.

PR #1387, opened 2022-04-16
Shave ~8ms off generator runtime

holmanb:
```
Shave ~8ms off generator runtime

- remove redundant file reads and subcommands[1]
- add make target that renders templates[2]
- add make target that benchmarks generator[3]

[1] cuts time in lxd from 17-20ms down to 8-13ms in lxc on my local machine
[2] example: make render-template FILE=./systemd/cloud-init-generator.tmpl
[3] usage: NUM_RUNS=42 make benchmark-generator
```

## Additional Context
`read_proc_cmdline()` gets run twice currently. I don't believe this is necessary. This is where the time savings come from.


TheRealFalcon (commented):
The generator changes LGTM, and I think it makes sense to have a `render-template` target in the Makefile. I'm not sure I agree with putting performance benchmarking in the makefile. It's unrelated to actually building anything, and is more or less shoehorning a shell script into Makefile syntax. I think it'd make sense as a separate script in the tools directory. Additionally, I think there should be some indication that the benchmarking shouldn't be run on a system not being used to test cloud-init.

holmanb:
> It's unrelated to actually building anything, and is more or less shoehorning a shell script into Makefile syntax. I think it'd make sense as a separate script in the tools directory.

I'm fine with moving the "benchmark" implementation to a shell script in `tools/`, but I would prefer if we keep a target in the makefile for easy discovery and invocation.

At a glance I see several other targets that are small shell scripts in the Makefile (some of which I wrote, admittedly). For consistency, would you prefer moving other tool implementations out of the Makefile and into `tools/`?

> Additionally, I think there should be some indication that the benchmarking shouldn't be run on a system not being used to test cloud-init.

+1

TheRealFalcon (approved):
I dub thee the official Makefile maintainer.

PR #1388, opened 2022-04-18
cloudinit.net refactor: apply_network_config_names

aciba90:
## Proposed Commit Message

```
cloudinit.net refactor: apply_network_config_names

- Remove `apply_network_config_names` from `cloudinit.net` and `Distro` classes so it is only accessible via the networking attribute.

LP: #1884602
```

## Checklist:

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -13,7 +13,18 @@ from cloudinit import log as logging
 from cloudinit import subp, util
 from cloudinit.settings import PER_INSTANCE
 
+from .networking import BSDNetworking, NetworkConfig
+
 LOG = logging.getLogger(__name__)
+
+
+class FreeBSDNetworking(BSDNetworking):
+    def apply_network_config_names(self, netcfg: NetworkConfig) -> None:
+        # This is handled by the freebsd network renderer. It writes in
+        # /etc/rc.conf a line with the following format:
+        #    ifconfig_OLDNAME_name=NEWNAME
+        # FreeBSD network script will rename the interface automatically.
+        pass
 
 
 class Distro(cloudinit.distros.bsd.BSD):
@@ -23,6 +34,7 @@ class Distro(cloudinit.distros.bsd.BSD):
     (N.B. DragonFlyBSD inherits from this class.)
     """
 
+    networking_cls = FreeBSDNetworking
     usr_lib_exec = "/usr/local/lib"
     login_conf_fn = "/etc/login.conf"
     login_conf_fn_bak = "/etc/login.conf.orig"
@@ -152,13 +164,6 @@ class Distro(cloudinit.distros.bsd.BSD):
                 util.logexc(
                     LOG, "Failed to restore %s backup", self.login_conf_fn
                 )
-
-    def apply_network_config_names(self, netconfig):
-        # This is handled by the freebsd network renderer. It writes in
-        # /etc/rc.conf a line with the following format:
-        #    ifconfig_OLDNAME_name=NEWNAME
-        # FreeBSD network script will rename the interface automatically.
-        pass
 
     def _get_pkg_cmd_environ(self):
         """Return environment vars used in *BSD package_command operations"""
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -13,18 +13,9 @@ from cloudinit import log as logging
 from cloudinit import subp, util
 from cloudinit.settings import PER_INSTANCE
 
-from .networking import BSDNetworking, NetworkConfig
+from .networking import FreeBSDNetworking
 
 LOG = logging.getLogger(__name__)
-
-
-class FreeBSDNetworking(BSDNetworking):
-    def apply_network_config_names(self, netcfg: NetworkConfig) -> None:
-        # This is handled by the freebsd network renderer. It writes in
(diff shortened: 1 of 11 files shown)


TheRealFalcon (commented):
@aciba90 , thanks for this contribution.

I think the end goal of the net refactor (which has admittedly stalled), is to remove methods from the distro classes entirely and into the networking classes. So the call in [stages.py](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/stages.py#L855) would be `self.distro.networking.apply_network_config_names(netcfg)` instead.

In order to do that, we would also need implementations in the BSD networking class. The current implementations (e.g., [1](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/bsd.py#L138) [2](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/freebsd.py#L156) [3](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/netbsd.py#L137) ) are unsupported or no-ops and would also need to be refactored into this networking class setup.

aciba90:
Thank you for your comments.

> I think the end goal of the net refactor (which has admittedly stalled), is to remove methods from the distro classes entirely and into the networking classes. So the call in [stages.py](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/stages.py#L855) would be `self.distro.networking.apply_network_config_names(netcfg)` instead.

I have removed those methods and made the change in `stages.py`.

> In order to do that, we would also need implementations in the BSD networking class. The current implementations (e.g., [1](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/bsd.py#L138) [2](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/freebsd.py#L156) [3](https://github.com/canonical/cloud-init/blob/e783ef5c8e5f7c46816161f6a902f2912a4cc93f/cloudinit/distros/netbsd.py#L137) ) are unsupported or no-ops and would also need to be refactored into this networking class setup.

I have moved current implementations into `BSDNetworking` and `FreeBSDNetworking`.

@TheRealFalcon please let me know if there is something else to improve.

cloudinit/distros/freebsd.py
```diff
@@ -13,16 +13,28 @@
 from cloudinit import subp, util
 from cloudinit.settings import PER_INSTANCE
 
+from .networking import BSDNetworking, NetworkConfig
+
 LOG = logging.getLogger(__name__)
 
 
+class FreeBSDNetworking(BSDNetworking):
```
TheRealFalcon, line 21: Let's put this class in `networking.py`. For now, I think it makes more sense there.

tests/unittests/distros/test_networking.py
```diff
@@ -229,3 +235,113 @@ def test_retrying_and_strict_behaviour(
             5 * len(wait_for_physdevs_netcfg["ethernets"])
             == m_settle.call_count
         )
+
+
+class TestLinuxNetworkingApplyNetworkCfgNames:
```
TheRealFalcon, line 240: Nice refactor of these tests! This is much nicer.

TheRealFalcon, line 240: Nice refactor of these tests! This is much nicer.

tests/unittests/distros/test_networking.py
```diff
@@ -229,3 +235,113 @@ def test_retrying_and_strict_behaviour(
...
+        )
+
+    @pytest.mark.parametrize(
+        ["config_attr"],
+        [
+            pytest.param("V2_CONFIG_NO_SETNAME", id="without_setname"),
+            pytest.param("V2_CONFIG_NO_MAC", id="without_mac"),
+        ],
+    )
+    def test_apply_v2_renames_skips_without_setname(self, config_attr: str):
```
TheRealFalcon, line 335: Nit: `test_apply_v2_renames_skips_without_setname_or_mac`

TheRealFalcon, line 335: Nit: `test_apply_v2_renames_skips_without_setname_or_mac`

TheRealFalcon (changes requested):
Thanks for addressing the requested changes. One more minor change in where the new networking class lives, and I think we'll be good here.

cloudinit/distros/freebsd.py
aciba90: Done

tests/unittests/distros/test_networking.py
aciba90: Thanks!

tests/unittests/distros/test_networking.py
aciba90: Done

aciba90:
> Thanks for addressing the requested changes. One more minor change in where the new networking class lives, and I think we'll be good here.

Thanks for the comments. Resolved!

TheRealFalcon (approved):
LGTM, thanks! I'm going to briefly close and then reopen this PR to trigger a Travis rebuild.

PR #1389, opened 2022-04-18
Move notes about refactorization out of CONTRIBUTING.rst and into separate doc (SC-949)

holmanb:
```
Move notes about refactorization out of contributing doc
```


TheRealFalcon (approved):
Thanks...this doesn't really belong in our public documentation.

PR #1390, opened 2022-04-18
Spring cleaning

holmanb:
```
Drop references to old dependencies and old centos script
```

## Additional Context
- boto dependency dropped in 0.7.5
- The Centos script has been deprecated for some time, and only supports 6 & 7. Rather than adapt to 8-stream, just drop it. 7 gets updates for a couple more years, but one would need to look back into git history to find a version of cloud-init that runs on python2 anyways, so no need to keep it in the main branch.


blackboxsw (approved):
LGTM

PR #1391, opened 2022-04-19
schema: render schema paths from _CustomSafeLoaderWithMarks

blackboxsw:
## Proposed Commit Message
```
Use a custom yaml.SafeLoader to track SchemaMarks to define
start and end row/column values of each object within the
original cloud-config content.

Fixes KeyError from cloud-init devel schema --annotate when
encountering invalid nested list values under a dict key.

Drop _CustomSafeLoaderWithMarks replaces the need for
_schemapath_for_cloudconfig logic.

Fixes: SC-929
```

## Test Steps
```bash
$ cat <baduser.txt  >>EOF
#cloud-config
users:
- default                                                                       
- name: ok
  sudo: false
- name: user2
  sudo: 2
EOF
 PYTHONPATH=. python3 -m cloudinit.cmd.main devel schema -c baduser.txt --annotate

# expect valid `E1` failure annotations on the correct line of the original content
```

## Checklist:

### DIFF
--- a/cloudinit/safeyaml.py
+++ b/cloudinit/safeyaml.py
@@ -135,7 +135,7 @@ def dumps(obj, explicit_start=True, explicit_end=True, noalias=False):
         explicit_start=explicit_start,
         explicit_end=explicit_end,
         default_flow_style=False,
-        Dumper=(NoAliasSafeDumper if noalias else yaml.dumper.Dumper),
+        Dumper=(NoAliasSafeDumper if noalias else yaml.dumper.SafeDumper),
     )
 
 
--- a/cloudinit/safeyaml.py
+++ b/cloudinit/safeyaml.py
@@ -23,7 +23,7 @@ class _CustomSafeLoader(yaml.SafeLoader):
 
 
 class _CustomSafeLoaderWithMarks(yaml.SafeLoader):
-    """A loader which catalogs line and column coordination for objects."""
+    """A loader preserving line and column start and end  marks for objects."""
 
     def __init__(self, stream):
         super().__init__(stream)
--- a/cloudinit/safeyaml.py
+++ b/cloudinit/safeyaml.py
@@ -12,6 +12,9 @@ import yaml
 
 YAMLError = yaml.YAMLError
 
+# SchemaPathMarks track the path to an element within a loaded YAML file.
+# The start_mark and end_mark contain the row and column indicators
+# which represent the coordinates where the schema element begins and ends.
 SchemaPathMarks = namedtuple(
     "SchemaPathMarks", ("path", "start_mark", "end_mark")
 )
@@ -23,7 +26,34 @@ class _CustomSafeLoader(yaml.SafeLoader):
 
 
 class _CustomSafeLoaderWithMarks(yaml.SafeLoader):
-    """A loader preserving line and column start and end  marks for objects."""
+    """A loader which provides line and column start and end marks for YAML.
+
+    If the YAML loaded represents a dictionary, get_single_data will inject
+    a top-level "schemamarks" key in that dictionary which can be used at
+    call-sites to process YAML paths schemamark metadata when annotating
+    YAML files for errors.
+
+    The schemamarks key is dictionary where each key is a dot-delimited path
+    into the YAML object. Each dot represents an element that is nested under
+    a parent and list items are represented with the format
+    `<parent>.<list-index>`.
+
+    The values in schemamarks will be the line number in the original content
+    where YAML element begins to aid in annotation when encountering schema
+    errors.
+
+    The example YAML shows expected schemamarks for both dicts and lists:
+
+      one: val1
+      two:
+        subtwo: val2
+      three: [val3, val4]
+
+    schemamarks == {
+        "one": 1, "two": 2, "two.subtwo": 3, "three": 4, "three.0": 4,
+        "three.1": 4
+    }
+    """
 
     def __init__(self, stream):
         super().__init__(stream)
@@ -77,7 +107,7 @@ class _CustomSafeLoaderWithMarks(yaml.SafeLoader):
 
     def get_single_data(self):
         data = super().get_single_data()
-        if isinstance(data, dict):  # valid cloud-config scheam is a dict
+        if isinstance(data, dict):  # valid cloud-config schema is a dict
             data["schemamarks"] = dict(
                 [
                     (v.path, v.start_mark.line + 1)  # 1-based human-readable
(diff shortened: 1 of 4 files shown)


holmanb (commented):
@blackboxsw Thanks for the PR.

I'm currently catching up on the context/purpose of our custom yaml classes for this review. Sorry question you on code you didn't write, but while doing some archeology to grok this change I noticed something odd in the commit history that I don't understand and I'm curious could help me understand.

I see that the same commit (cf30836645473c62599e838ab48b2d31677fa584) that added the safe dumper class also changed `yaml.safe_dump` to `yaml.dump` and elected to default to `yaml.dumper.Dumper` rather than `yaml.dumper.SafeDumper` as the default. This change doesn't really make sense to me in the context of the commit message. My understanding of the yaml api is superficial at present, so perhaps I misunderstand the usage. Would you happen to know offhand why this was changed from the "safe" api?

Aside: `safeyaml.py` seemed a bit light on comments / references in the code prior to your PR. Could we grab add some context from cf30836645473c62599e838ab48b2d31677fa584 and a7a9de1a079a70f5c8290ab5158661d3a33e5552 describing how this alters default yaml parsing while we are touching this code?

No objections so far, but I have more to review.

blackboxsw:
> @blackboxsw Thanks for the PR.
> 
> I'm currently catching up on the context/purpose of our custom yaml classes for this review. Sorry question you on code you didn't write, but while doing some archeology to grok this change I noticed something odd in the commit history that I don't understand and I'm curious could help me understand.
> 
> I see that the same commit ([cf30836](https://github.com/canonical/cloud-init/commit/cf30836645473c62599e838ab48b2d31677fa584)) that added the safe dumper class also changed `yaml.safe_dump` to `yaml.dump` and elected to default to `yaml.dumper.Dumper` rather than `yaml.dumper.SafeDumper` as the default. This change doesn't really make sense to me in the context of the commit message. My understanding of the yaml api is superficial at present, so perhaps I misunderstand the usage. Would you happen to know offhand why this was changed from the "safe" api?
> 
> Aside: `safeyaml.py` seemed a bit light on comments / references in the code prior to your PR. Could we grab add some context from [cf30836](https://github.com/canonical/cloud-init/commit/cf30836645473c62599e838ab48b2d31677fa584) and [a7a9de1](https://github.com/canonical/cloud-init/commit/a7a9de1a079a70f5c8290ab5158661d3a33e5552) describing how this alters default yaml parsing while we are touching this code?
> 
> No objections so far, but I have more to review.

Thanks for the digging on this. I tried going through original PRs and even LP merge requests on this original content to refresh myself on the intent of those first branches. I agree with you there is something unique about the transition from yaml.safe_dump to a default of a yaml.Dumper which may have actually slipped through the cracks that conflicted with the original intent. I want to dig a bit more on this tomorrow as I'm inclined to move to the `yaml.dumper.SafeDumper` class for the default noalias=False behavior but I want to confirm use cases we know of  here w/ MAAS and any call-sites we have grown in cloudinit to make sure we aren't inadvertently breaking a behavior that has grown since the commit that moved us from safe_dump -> dump with `dumper=yaml.dumper.Dumper`.

blackboxsw:
> > @blackboxsw Thanks for the PR.
> > I'm currently catching up on the context/purpose of our custom yaml classes for this review. Sorry question you on code you didn't write, but while doing some archeology to grok this change I noticed something odd in the commit history that I don't understand and I'm curious could help me understand.
> > I see that the same commit ([cf30836](https://github.com/canonical/cloud-init/commit/cf30836645473c62599e838ab48b2d31677fa584)) that added the safe dumper class also changed `yaml.safe_dump` to `yaml.dump` and elected to default to `yaml.dumper.Dumper` rather than `yaml.dumper.SafeDumper` as the default. This change doesn't really make sense to me in the context of the commit message. My understanding of the yaml api is superficial at present, so perhaps I misunderstand the usage. Would you happen to know offhand why this was changed from the "safe" api?
> > Aside: `safeyaml.py` seemed a bit light on comments / references in the code prior to your PR. Could we grab add some context from [cf30836](https://github.com/canonical/cloud-init/commit/cf30836645473c62599e838ab48b2d31677fa584) and [a7a9de1](https://github.com/canonical/cloud-init/commit/a7a9de1a079a70f5c8290ab5158661d3a33e5552) describing how this alters default yaml parsing while we are touching this code?
> > No objections so far, but I have more to review.
> 
> Thanks for the digging on this. I tried going through original PRs and even LP merge requests on this original content to refresh myself on the intent of those first branches. I agree with you there is something unique about the transition from yaml.safe_dump to a default of a yaml.Dumper which may have actually slipped through the cracks that conflicted with the original intent. I want to dig a bit more on this tomorrow as I'm inclined to move to the `yaml.dumper.SafeDumper` class for the default noalias=False behavior but I want to confirm use cases we know of here w/ MAAS and any call-sites we have grown in cloudinit to make sure we aren't inadvertently breaking a behavior that has grown since the commit that moved us from safe_dump -> dump with `dumper=yaml.dumper.Dumper`.

@holmanb, confirmed, that original changeset in 0836645473c62599e838ab48b2d31677fa584 inadvertently adapted the fallback option to move from `yaml.safe_dump` to yaml.dumper.Dumper as you suggested. There is nothing in callsites that would be negatively impacted by using the SafeDumper, so I'll shift this back to align more with the original implementation as I believe that was an oversight.

cloudinit/safeyaml.py
```diff
@@ -4,14 +4,87 @@
...
+            )
+            if line_num not in self.schemamarks_by_line:
+                self.schemamarks_by_line[line_num] = [marks]
+            else:
+                self.schemamarks_by_line[line_num].append(marks)
+        return sequence
+
+    def get_single_data(self):
+        data = super().get_single_data()
+        if isinstance(data, dict):  # valid cloud-config scheam is a dict
```
holmanb, line 80: `"scheam" should be schema

holmanb (approved):
Reviewing this required some digging into how the yaml object internals work, but I don't see any issues or concerns with it.

The SafeLoader subclass just squirrels away mark start/end data in `schemamarks_by_line` for later use as it is parsing the yaml. I was previously going to ask for some extra comments around this code since the purpose isn't immediately obvious, but the yaml module code is pretty well documented, and future behavioral modifications to that code will require reading the yaml anyways so there really isn't much to add besides what you already have.

I ran a couple of speed and memory benchmarks on it, since parsing can potentially be computationally intensive, but the results showed insignificant difference from the `yaml.SafeLoader` class, so I'm not concerned at all about this.

Thanks for switching these to the "Safe" versions too.

Thanks for this!

PR #1392, opened 2022-04-19
sources/azure: refactor http_with_retries to remove **kwargs

cjp256:
Remove unused functionality and general kwargs support in favor
of defined parameters.

As only headers and data kwargs are used, add those explicitly
to http_with_retries().

There should be no behavioral changes.

### DIFF
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -330,34 +330,22 @@ def get_ip_from_lease_value(fallback_lease_value):
 
 
 @azure_ds_telemetry_reporter
-def http_with_retries(url, **kwargs) -> url_helper.UrlResponse:
+def http_with_retries(
+    url, *, headers: dict, data=None
+) -> url_helper.UrlResponse:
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

cloudinit/sources/helpers/azure.py
```diff
@@ -330,34 +330,22 @@ def get_ip_from_lease_value(fallback_lease_value):
 
 
 @azure_ds_telemetry_reporter
-def http_with_retries(url, **kwargs) -> url_helper.UrlResponse:
+def http_with_retries(
+    url, *, headers: dict, data=None
```
blackboxsw, line 334: Since we are type hinting, let's get url too
```suggestion
    url: str, *, headers: dict, data=None
```

blackboxsw (commented):
Just a minor nit and +1

cloudinit/sources/helpers/azure.py
cjp256: done, for data too, thanks!

PR #1395, opened 2022-04-20
Bionic: d/control: add python3-responses for unittests

holmanb:
## Proposed Commit Message

```
d/control: add python3-responses for unittests
```

## Additional Context
ref: https://github.com/canonical/cloud-init/compare/ubuntu/bionic...holmanb:ubuntu/bionic?expand=1
## Test Steps
```
git checkout upstream/ubuntu/bionic
git merge <this_branch>
build-package
sbuild --resolve-alternatives --dist=bionic --arch=amd64 ../out/cloud-init*dsc
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (22.1-14-g2e17a0d6-0ubuntu1~18.04.4) UNRELEASED; urgency=medium
+
+  * d/control:
+    - add python3-responses for unittests
+    - add Suggests: ssh-import-id
+
+ -- Brett Holman <brett.holman@canonical.com>  Fri, 22 Apr 2022 11:55:09 -0600
+
 cloud-init (22.1-14-g2e17a0d6-0ubuntu1~18.04.1) bionic; urgency=medium
 
   * d/patches/retain-apt-partner-pocket.patch:
--- a/debian/control
+++ b/debian/control
@@ -20,7 +20,8 @@ Build-Depends: debhelper (>= 9.20160709),
                python3-requests,
                python3-serial,
                python3-setuptools,
-               python3-yaml
+               python3-yaml,
+               python3-responses
 XS-Python-Version: all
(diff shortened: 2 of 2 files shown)


debian/changelog
```diff
@@ -1,5 +1,8 @@
 cloud-init (22.1-14-g2e17a0d6-0ubuntu1~18.04.1) bionic; urgency=medium
 
+  * d/control:
```
blackboxsw, line 3: I think we need a new UNRELEASED changelog entry here because 22.1-14-g2e17a0d6-0ubuntu1~18.04.1 is already officially published. 
So you'll want to `dch -v 22.1-14-g2e17a0d6-0ubuntu1~18.04.4` and put this d/control content into that changelog entry as a separate commit from the d/control changes. This will make it easier to collapse the next new-upstream-snapshot into ubuntu/bionic in the future with changelog commits vs functional commits.

blackboxsw: `rmadison cloud-init tells` us we already released 18.04.3 so let's bump to the next available version (even though we'll be setting this to something like 22.2  before we officially release next (and syncing in the hotfix d/changelog entries we published from `upstream/ubuntu/bionic-22.1-hotfix`

blackboxsw (changes requested):
Confirmed tox -e doc changes will be resolved in CI once we new-upstream-snapshot for the next official release so it's a don't care for this PR. But we generally should not add content to d/changelog for releases that are already published. For those unreleased content, we generally add a new d/changelog entry (normally via dch -i) to queue any officially unlreleased content. Our new-upstream-snapshot tooling will collapse any UNRELEASED d/changelog entry into a single entry prior to official release.

blackboxsw (approved):
+1
- [x] build-package; sbuild SUCCESS
- [x]  can merge into upstream/main and build without error
- [x] tested install on system with neither ssh-import-id nor python3-responses  no negative hard-requirements for packages that are not critical
- [x] tox -e doc failures resolved during next `new-upstream-snapshot`

PR #1396, opened 2022-04-20
Focal: d/control: add python3-responses for unittests

holmanb:
## Proposed Commit Message

```
d/control: add python3-responses for unittests
```

## Additional Context
ref: https://github.com/canonical/cloud-init/compare/ubuntu/bionic...holmanb:ubuntu/bionic?expand=1
## Test Steps
```
git checkout upstream/ubuntu/focal
git merge <this_branch>
build-package
sbuild --resolve-alternatives --dist=focal --arch=amd64 ../out/cloud-init*dsc
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,8 @@
 cloud-init (22.1-43-gbf94945f-0ubuntu1~20.04.1) UNRELEASED; urgency=medium
 
+  * d/control:
+    - add python3-responses for unittests
+    - add Suggests: ssh-import-id
   * refresh patches:
    + debian/patches/retain-apt-partner-pocket.patch
   * New upstream snapshot. (LP: #SRU_BUG_NUMBER_HERE)
--- a/debian/control
+++ b/debian/control
@@ -21,7 +21,8 @@ Build-Depends: debhelper (>= 9.20160709),
                python3-requests,
                python3-serial,
                python3-setuptools,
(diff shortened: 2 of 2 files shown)


debian/changelog
```diff
@@ -1,5 +1,8 @@
 cloud-init (22.1-43-gbf94945f-0ubuntu1~20.04.1) UNRELEASED; urgency=medium
```
holmanb, line 1: Not sure if I got the version right here - does this one need 20.04.1 bumped to 20.04.4? Or is fine as is due to the snapshot previously bumping 22.1-14 -> 22.1-43

blackboxsw: All good here thanks for the question. Since we had already performed `new-upstream-snapshot` into `ubuntu/focal` and not yet officially released this branch to 22.04 we can and should append to this "open" UNRELEASED changelog entry without needing to bump the version. +1

blackboxsw: For our "process" the d/changelog entry becomes closed when we label UNRELEASED -> <some_series_name> to indicate to us that we have published this version to official Ubuntu -updates repos.

blackboxsw (approved):
- [x] build-package; sbuild SUCCESS
- [x]  quilt push/pop -a with tox SUCCESS
- [x]  installed deb on system with neither ssh-import-id nor python3-responses SUCCESS

PR #1397, opened 2022-04-20
Schema example doc fixes

blackboxsw:
## Proposed Commit Message

```
    tests: make module example tests very strict to specific schema
    
    Followup to c2bfd5. Schema module examples should be strictly
    validated against the specific schema. Given that
    cloud-init-schema.json defines a permissive 'allOf' set of schemas
    top-level additionalParoperties that are undefined in schema are
    permitted.
    
    This allowed for multiple typos to get through schema
    unittest validation.

     Also fix invalid schema definition in cc_seed_random which should define
     `random_seed` property instead of `seed_random`.
```

## Test Steps
Unit tests should now strictly validate and fail and specific example typos:
```
tox -e py3  tests/unittests/config/test_schema.py -vv
```

## Checklist:


holmanb (approved):
Strict schema validation in tests is a systemic solution. Thanks!

PR #1402, opened 2022-04-25
Promote cloud-init schema (SC-953)

holmanb:
```
Promote schema command from devel to top level subcommand
```

### DIFF
--- a/cloudinit/cmd/devel/parser.py
+++ b/cloudinit/cmd/devel/parser.py
@@ -6,8 +6,6 @@
 
 import argparse
 
-from cloudinit.config import schema
-
 from . import hotplug_hook, make_mime, net_convert, render
 
 
--- a/cloudinit/cmd/devel/parser.py
+++ b/cloudinit/cmd/devel/parser.py
@@ -28,12 +28,6 @@ def get_parser(parser=None):
             hotplug_hook.handle_args,
         ),
         (
-            "schema",
-            "Validate cloud-config files for document schema",
-            schema.get_parser,
-            schema.handle_schema_args,
-        ),
-        (
             net_convert.NAME,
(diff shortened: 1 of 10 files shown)


cloudinit/cmd/devel/parser.py
```diff
@@ -29,7 +29,8 @@ def get_parser(parser=None):
         ),
         (
             "schema",
-            "Validate cloud-config files for document schema",
+            "schema has been promoted from devel up to a top level subcommand,"
+            " invoking via `cloud-init devel schema` is now deprecated",
```
TheRealFalcon, line 33: We have a warning with our devel commands that they can and will change (and likely be promoted top-level), so I'm ok removing this entirely rather than deprecating it.

TheRealFalcon (changes requested):
Some really nice changes here. Normally I like small updates along the way, but it looks like there's more non-schema changes than there are schema ones. Do you mind moving the misc updates to another PR?

Additionally, there's some other places that need to be updated.
* If you search for "devel schema" in faq.rst, tutorial.rst, and tests/integration_tests/modules/test_cli.py, they all need updates.
* Search '"devel", "schema"' in tests/unittests/test_cli.py
* cli.rst has schema documentation under the devel heading.
* `bash_completion/cloud-init` has schema is under devel

holmanb:
I had to rebase for dropping commits, hence the force push. Since I already had to force push, I squashed the recommended "drop devel schema" change into the promotion commit since they overlapped.

TheRealFalcon (approved):
LGTM, thanks!

I trust you can fix the flake8 and spell check issue without another review.

holmanb:
After rerunning CI passes, merging.

PR #1403, opened 2022-04-26
testing: Add responses workaround for focal/impish (SC-954)

TheRealFalcon:
## Proposed Commit Message

```
testing: Add responses workaround for focal/impish

The responses workaround relies on an internal variable, and the
internals changed between Bionic and Focal, so the workaround needs
to be updated as well.
```

## Additional Context
This fixes tests failing on Focal/Impish


holmanb (approved):
LGMT, thanks!

PR #1411, opened 2022-04-27
Don't fail if IB and Ethernet devices 'collide'

holmanb:
```
Don't fail if InfiniBand/Ethernet addresses collide

Current behavior for the collision of Ethernet mac 
address and the "openstack 6 byte IB format" is to fail[1].
This isn't a problem for the respective Ethernet and IB
networks, so don't force cloud-init to fail.

[1] https://bugs.launchpad.net/cloud-init/+bug/1871326
```

I was on the fence about whether to propose as-is, or to move forward with refactoring out the openstack-specific code. I proposed because this represents a failure path that I don't think we want and the "fix" is relatively small.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1031,9 +1031,9 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
             if ib_mac not in ret:
                 ret[ib_mac] = name
                 LOG.warning(
-                    "Ethernet and InfiniBand interfaces have the same address",
+                    "Ethernet and InfiniBand interfaces have the same address"
                     " both '%s' and '%s' have address '%s'."
-                    % (name, ret[ib_mac], ib_mac)
+                    % (name, ret[ib_mac], ib_mac),
                 )
     return ret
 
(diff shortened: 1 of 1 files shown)


cloudinit/net/__init__.py
```diff
@@ -1002,16 +1002,41 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
...
+            # no sense to compare them to Ethernet mac addresses. This appears
+            # to be a # workaround for openstack-specific behavior[1], and for
+            # now leave it to avoid breaking openstack
+            # but this should be removed from get_interfaces_by_mac_on_linux()
+            # because IB GUIDs are not mac addresses, and operate on a separate
+            # L2 protocol so address collision doesn't matter.
+            #
+            # [1] sources/helpers/openstack.py:convert_net_json() expects
+            # net.get_interfaces_by_mac() to return IB addresses in this format
+            if ib_mac not in ret:
```
TheRealFalcon, line 1031: Is this section right? Don't we want to warn if `ib_mac in ret` but set the value regardless?

holmanb: Oops, yes that was my intention.

TheRealFalcon (approved):
LGTM

PR #1412, opened 2022-04-27
mypy: disable missing imports warning for httpretty

cjp256:
httpretty does not appear to support mypy stubs. Add configuration
to for mypy to ignore this import.

With this we can add checking for test_azure.py.


TheRealFalcon (approved):
Looks good assuming CI passes. We're hoping to replace httpretty with responses in the near future due to all the httpretty issues we've had.

PR #1414, opened 2022-04-27
sources/azure: remove reprovisioning marker

cjp256:
If we haven't reported ready for source PPS then we can treat
the recovery boot like any other.  The metadata on the OVF
and IMDS will indicate the PPS type correctly as the state
hasn't changed.

If we have reported ready for source PPS, we continue to fall
into _poll_imds() by way of setting pps_type to UNKNOWN if the
REPORTED_READY_MARKER is present and will not attempt to report
ready again.

This fixes a potential issue when recovering on Savable PPS.
If a recovery boot occurs after the recovery marker is created,
and without reporting ready, the subsequent boot will assume
pps type UNKNOWN and attempt to report ready in _poll_imds()
using the Running PPS netlink operations.

Add unit test coverage for complete recovery scenario.


cjp256:
NOTE: Prior to merging, needs approval @anhvoms :)

anhvoms (approved):
LGTM

TheRealFalcon (approved):
Thanks!

PR #1415, opened 2022-04-27
docs: Add docs for module creation (SC-961)

TheRealFalcon:
## Proposed Commit Message

```
docs: Add docs for module creation

Additionally remove the cc_foo module as any relevant documentation
there has been moved to the official documentation.
```


TheRealFalcon:
Pushed a rebase

holmanb:
I think this does a good job of covering what goes in the config module directory. Think we should add context for boot stage determination / cloud.cfg.tmpl / etc?

doc/rtd/topics/module_creation.rst
```diff
@@ -0,0 +1,94 @@
...
+    instance types.
+  * ``log``: A logger object that can be used to log messages.
+  * ``args``: An argument list. This is usually empty and is only populated
+    if the module is called independently from the command line.
+
+* If your module introduces any new cloud-config keys, you must provide a
+  schema definition in `cloud-init-schema.json`_.
+* The ``meta`` variable must exist and be of type `MetaSchema`_.
+
+  * ``distros``: Defines the list of supported distros. It can contain
```
blackboxsw, line 71: Is it worth noting that ``id``: Unique string beginning with the `cc_` prefix for this module.
Or maybe we just define our example in terms of the following... but then we're `import os` in each module. Though it's a paid cost anyway as we've loaded it before.

`'id':  os.path.basename(__file__).rpartition(".")[0]`

doc/rtd/topics/module_creation.rst
```diff
@@ -0,0 +1,94 @@
...
+  * ``frequency``: Defines how often module runs. It must be one of:
+
+    * ``PER_ALWAYS``: Runs on every boot.
+    * ``ONCE``: Runs only on first boot.
+    * ``PER_INSTANCE``: Runs once per instance. When exactly this happens
+      is dependent on the datasource but may triggered anytime there
+      would be a significant change to the instance metadata. An example
+      could be an instance being moved to a different subnet.
+
+  * ``examples``: Lists examples to be shown in the documentation.
```
blackboxsw, line 83:
```suggestion
  * ``examples``: Lists examples of any new cloud-config keys this module reacts to.
    These examples will be rendered in the module reference documentation.
```

doc/rtd/topics/module_creation.rst
```diff
@@ -0,0 +1,94 @@
...
+
+    meta: MetaSchema = {
+        "id": "cc_example",
+        "name": "Example Module",
+        "title": "Shows how to create a module",
+        "description": MODULE_DESCRIPTION,
+        "distros": [ALL_DISTROS],
+        "frequency": PER_INSTANCE,
+        "examples": [
+            "cc_example: example1",
```
blackboxsw, line 38: might want to express this as something a bit different from cc_example so it doesn't imply some sort of requirement to match the current module name.
```suggestion
            "some_new_config_key: allowed_value",
```

TheRealFalcon:
Addressed comments in a separate commit. Force push is only for rebase.

doc/rtd/topics/module_creation.rst
```diff
@@ -0,0 +1,117 @@
...
+    during testing.
+
+* ``__doc__ = get_meta_doc(meta)`` is necessary to provide proper module
+  documentation.
+
+Module Execution
+================
+
+In order for a module to be run, it must be defined in ``/etc/cloud`` on the
+launched instance. The three module sections are
```
blackboxsw, line 97:
```suggestion
In order for a module to be run, it must be defined in a module run section in 
``/etc/cloud/cloud.cfg`` or ``/etc/cloud/cloud.cfg.d`` on the launched instance.
The three module sections are
```

blackboxsw (approved):
Minor thought on being clear about where the config files exist for changing such a value. If we don't point to /etc/cloud/cloud.cfg specifically, someone may inadvertently add a `cloud_init_modules: [cc_my_mod]` which would drop all of our predefined modules in the base config.   

I'll defer to whether or not you think the specific comment is useful or needed but +1 regardless of your choice there in wording.

holmanb (approved):
Thanks for the `cloud.cfg.tmpl` addition. LGTM

PR #1418, opened 2022-04-29
Misc Tidying (SC-964)

holmanb:
```
Misc module cleanup

- move datasource helpers to dedicated directory
- drop unnecessary executable bit on shebangless python files
```

I can't say I'm in love with this in the git history, but I am also not in love with these inconsistencies so I threw together a PR.


aciba90 (approved):
+1 to me. Thanks!

TheRealFalcon (commented):
I think for the first 4 (cloud_id, make_mime, net_convert, and render), it makes more sense to add the python shebang as they have their own main functions and are intended to be run from the cli.

Otherwise PR LGTM.

holmanb:
> I think for the first 4 (cloud_id, make_mime, net_convert, and render), it makes more sense to add the python shebang as they have their own main functions and are intended to be run from the cli.


I've always seen them used from the command line as modules (`python -m cloudinit.cmd.<module>`) or, in the case of cloud-init/cloud-id commands, using the python "entry_points" tooling. Since neither method requires executable bit to be set, I didn't add them. That said, I don't see any reason why we can't / shouldn't do this, so I can make that change.

TheRealFalcon (approved):
LGTM. Approve assuming CI passes.

PR #1419, opened 2022-04-29
Fix doc warnings and enable errors.

aciba90:
## Proposed Commit Message

```
Fix doc warnings

- Remove Sphinx build warnings
- Turn Sphinx warnings into errors


LP: #1876341
SC-894
```

## Checklist:


cloudinit/config/cc_phone_home.py
```diff
@@ -43,7 +43,9 @@
 
 Data is sent as ``x-www-form-urlencoded`` arguments.
 
-**Example HTTP POST**::
+**Example HTTP POST**:
```
aciba90, line 46: In order to modify the global literal highlight language (yaml in our case) we have to use an explicit code-block directive.

[literal-blocks](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks)

TheRealFalcon (approved):
Good work! Some of the warnings were legitimate issues, so I'm glad we got that fixed.

PR #1422, opened 2022-05-02
Handle error if SSH service no present.

aciba90:
## Proposed Commit Message

```
Add pre-fight check to verify if SSH is running

- In negative case then perform an early return with a warning and no 
  modifications
- Handle an exception when user-data provides 'ssh_pwauth: True'
  and the open-ssh service is not available
- Add openssh-server as suggested package for debian-based distros

SC-968
LP: #1969526
```

## Test Steps

```sh
cat > ssh_needed.yaml << EOF
#cloud-config
ssh_pwauth: true
EOF

lxc launch ubuntu-daily:focal f1 -c user.user-data="$(cat ssh_needed.yaml)"
lxc exec f1 -- cloud-init status --wait
lxc exec f1 apt remove opensssh-server
# reset PasswordAuthentication off again so that cloud-init wants to reset ssh server
lxc exec f1 -- sed -i 's/PasswordAuthentication yes/PasswordAuthentication no/' /etc/ssh/sshd_config
# force clean reboot so cloud-init reruns
lxc exec f1 -- cloud-init clean --logs --reboot
lxc exec f1 -- cloud-init status --wait --long
```

Expect status Done and a warning in the logs instead of the error.
Warning: `Failed to restart the SSH deamon. ...`

## Checklist:

### DIFF
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -92,12 +92,15 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
         distro.manage_service("status", service)
     except subp.ProcessExecutionError as e:
         LOG.warning(
-            "Ignoring config 'ssh_pwauth: %s'. Service '%s' is not running.",
+            (
+                "Ignoring config 'ssh_pwauth: %s'. SSH deamon service '%s' is "
+                "not running."
+            ),
             pw_auth,
             service,
         )
         LOG.debug(
-            "Service '%s' is not running: %s",
+            "SSH deamon service '%s' is not running: %s",
             service,
             e,
         )
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -14,7 +14,7 @@ from textwrap import dedent
 from cloudinit import log as logging
 from cloudinit import subp, util
 from cloudinit.config.schema import MetaSchema, get_meta_doc
-from cloudinit.distros import ALL_DISTROS, ug_util
+from cloudinit.distros import ALL_DISTROS, Distro, ug_util
 from cloudinit.settings import PER_INSTANCE
 from cloudinit.ssh_util import update_ssh_config
 
@@ -79,7 +79,7 @@ LOG = logging.getLogger(__name__)
 PW_SET = "".join([x for x in ascii_letters + digits if x not in "loLOI01"])
 
 
-def handle_ssh_pwauth(pw_auth, distro):
+def handle_ssh_pwauth(pw_auth, distro: Distro):
     """Apply sshd PasswordAuthentication changes.
 
     @param pw_auth: config setting from 'pw_auth'.
@@ -87,6 +87,22 @@ def handle_ssh_pwauth(pw_auth, distro):
     @param distro: an instance of the distro class for the target distribution
 
     @return: None"""
+    service = distro.get_option("ssh_svcname", "ssh")
+    try:
+        distro.manage_service("status", service)
+    except subp.ProcessExecutionError as e:
+        LOG.warning(
+            "Ignoring config 'ssh_pwauth: %s'. Service '%s' is not running.",
+            pw_auth,
+            service,
+        )
+        LOG.debug(
+            "Service '%s' is not running: %s",
+            service,
+            e,
+        )
+        return
+
     cfg_name = "PasswordAuthentication"
 
     if isinstance(pw_auth, str):
@@ -112,13 +128,8 @@ def handle_ssh_pwauth(pw_auth, distro):
         LOG.debug("No need to restart SSH service, %s not updated.", cfg_name)
         return
 
-    service = distro.get_option("ssh_svcname", "ssh")
-    try:
-        distro.manage_service("restart", service)
-    except subp.ProcessExecutionError as e:
-        LOG.warning("Failed to restart the SSH deamon. %s: %s", service, e)
-    else:
-        LOG.debug("Restarted the SSH daemon.")
+    distro.manage_service("restart", service)
+    LOG.debug("Restarted the SSH daemon.")
 
 
 def handle(_name, cfg, cloud, log, args):
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -112,8 +112,13 @@ def handle_ssh_pwauth(pw_auth, distro):
         LOG.debug("No need to restart SSH service, %s not updated.", cfg_name)
         return
 
-    distro.manage_service("restart", distro.get_option("ssh_svcname", "ssh"))
-    LOG.debug("Restarted the SSH daemon.")
+    service = distro.get_option("ssh_svcname", "ssh")
+    try:
+        distro.manage_service("restart", service)
+    except subp.ProcessExecutionError as e:
+        LOG.warning("Failed to restart the SSH deamon. %s: %s", service, e)
+    else:
+        LOG.debug("Restarted the SSH daemon.")
 
 
 def handle(_name, cfg, cloud, log, args):
@@ -226,7 +231,7 @@ def handle(_name, cfg, cloud, log, args):
     handle_ssh_pwauth(cfg.get("ssh_pwauth"), cloud.distro)
 
     if len(errors):
-        log.debug("%s errors occured, re-raising the last one", len(errors))
+        log.debug("%s errors occurred, re-raising the last one", len(errors))
         raise errors[-1]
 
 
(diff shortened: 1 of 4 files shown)


cloudinit/config/cc_set_passwords.py
```diff
@@ -112,8 +112,13 @@ def handle_ssh_pwauth(pw_auth, distro):
         LOG.debug("No need to restart SSH service, %s not updated.", cfg_name)
         return
 
-    distro.manage_service("restart", distro.get_option("ssh_svcname", "ssh"))
-    LOG.debug("Restarted the SSH daemon.")
+    service = distro.get_option("ssh_svcname", "ssh")
+    try:
+        distro.manage_service("restart", service)
+    except subp.ProcessExecutionError as e:
+        LOG.warning("Failed to restart the SSH deamon. %s: %s", service, e)
```
blackboxsw, line 119: Generally let's try to avoid the words "Failed" and "error" in warning level logs as they are confusing when you grep in /var/log/cloud-init.log. If we stick with this approach, please replace Failed 
```suggestion
        LOG.warning(f"Wrote '{cfg_name} {cfg_val}' to {DEF_SSHD_CFG}. Unable to restart the SSH deamon {service}: {e}")
```

blackboxsw (changes requested):
Thanks for this PR @aciba90.

I don't think we want to spend time calling into `update_ssh_config` on a system that doesn't have the necessary service to run ssh.
This suggestion below may be more work than it's worth, I look forward to your thoughts either way.

Ideally I think we want a pre-flight check at the beginning of `handle_ssh_pwauth` function if the ssh_svcname in question does not exist for restarting. If not present, exit early with a `LOG.warning(f"Ignoring config 'ssh_pwauth: {pw_auth}'. Service {service} is not installed")`

This PR is not straight-forward for this situation because we can't easily invoke `subp.which` for the service we are trying to restart because we are talking to the `distro.manage_service` layer which has significantly different implementations on each distro for systemd vs. non-systemd init systems. 

I wonder if we might be able to add a "status" command to `distro.manage_service`  to report whether or not the service actually exists. We could then try/except around a call to `distro.manage_service("status", service)`.

cloudinit/config/cc_set_passwords.py
aciba90: Thanks for the suggestion @blackboxsw.
 
> I wonder if we might be able to add a "status" command to `distro.manage_service` to report whether or not the service actually exists. We could then try/except around a call to `distro.manage_service("status", service)`.

I was thinking exactly about that. Let me rework this PR.

cloudinit/config/cc_set_passwords.py
```diff
@@ -79,14 +79,33 @@
...
     @param distro: an instance of the distro class for the target distribution
 
     @return: None"""
+    service = distro.get_option("ssh_svcname", "ssh")
+    try:
+        distro.manage_service("status", service)
+    except subp.ProcessExecutionError as e:
+        LOG.warning(
+            (
+                "Ignoring config 'ssh_pwauth: %s'. SSH deamon service '%s' is "
```
aciba90, line 96: > Ideally I think we want a pre-flight check at the beginning of handle_ssh_pwauth function if the ssh_svcname in question does not exist for restarting. If not present, exit early with a LOG.warning(f"Ignoring config 'ssh_pwauth: {pw_auth}'. Service {service} is not installed")

I though that it was more generic to say that the service is not running, as a non 0 exit from `systemctl status service` no always does mean that the service is not installed.

aciba90 (commented):
@blackboxsw this one is ready to be reviewed again. I have implemented the status command and the pre-flight check as suggested.

Thanks for you comments.

blackboxsw (changes requested):
I think adding the "status" operation to manage_services is helpful, we could more specifically call `systemctl is-enabled` but then we'd have to provide a different option any distro which doesn't support "is-enabled` command. I think it is probably cleaner at this iteration to properly look at exit codes: 
if exit_code == 0: write config and restart
   exit_code == 3: write_config but warn about not restarting ssh because service is in disabled state
   exit_code == 4: warn sshd is not installed and return
    other exit code: warn "sshd is unavailable due to {e}", to provide the exception message and return

here's a mockup diff of what I'm alluding to if this makes sense. I'm open to discussion and or disagreement here if you think this suggestion doesn't make sense.

```diff
diff --git a/cloudinit/config/cc_set_passwords.py b/cloudinit/config/cc_set_passwords.py
index d952ba4d7..927d0bf7b 100644
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -88,23 +88,25 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
 
     @return: None"""
     service = distro.get_option("ssh_svcname", "ssh")
+    restart_ssh = True
     try:
         distro.manage_service("status", service)
     except subp.ProcessExecutionError as e:
-        LOG.warning(
-            (
-                "Ignoring config 'ssh_pwauth: %s'. SSH deamon service '%s' is "
-                "not running."
-            ),
-            pw_auth,
-            service,
-        )
-        LOG.debug(
-            "SSH deamon service '%s' is not running: %s",
-            service,
-            e,
-        )
-        return
+        if e.exit_code == 3:
+            # Continue to write ssh but don't invoke restart
+            restart_ssh = False
+        elif e.exit_code == 4:
+            LOG.warning(
+                f"Ignoring config 'ssh_pwauth: {pw_auth}'."
+                f" SSH daemon '{service}' is not installed."
+            )
+            return
+        else:
+            LOG.warning(
+                f"Ignoring config 'ssh_pwauth: {pw_auth}'."
+                f" SSH daemon '{service}' is unavailable. Error: {e}."
+            )
+            return
 
     cfg_name = "PasswordAuthentication"
 
@@ -131,7 +133,10 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
         LOG.debug("No need to restart SSH service, %s not updated.", cfg_name)
         return
 
-    distro.manage_service("restart", service)
+    if restart_ssh:
+        distro.manage_service("restart", service)
+    else:
+        LOG.debug("Not restarting SSH service: service is stopped.")
     LOG.debug("Restarted the SSH daemon.")
```

aciba90:
Thanks for the suggestion @blackboxsw, I should have reacted to error_codes.

Those error codes are part of a [Reference specification](https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html), therefore I though it is more likely that non-compliant init system adopt it than that they implement an `is-enabled` command.

sysvinit init systems seem to be also complaint with the specification, for example: [Fedora](https://fedoraproject.org/wiki/EPEL:SysVInitScripts#Exit_Codes_for_the_Status_Action).

Ready to be reviewed.

blackboxsw (approved):
Thank you for the supplemental work here @aciba90 and unittest coverage. I ran through various exit_code scenarios on a live environment and this behavior performs as expected.

@aciba90 please also provide pull requests to add `Suggests: openssh-server` to debian/control against upstream ubuntu/bionic, ubuntu/focal, ubuntu/jammy and ubuntu/devel branches.

TheRealFalcon:
I know this has already merged, but I think we need a systemd check around the error checking and handle non-systemd in a more generic way. @aciba90 , you point to the LSB, but the LSB hasn't been updated in 7 years and has never really been respected by many distros. I don't think we can say "do it LSB way or your distro is wrong". There's a few other sections on init scripts, and systemd breaks most of them.

E.g., on alpine, calling `service sshd status` returns 1 if not installed, 3 if stopped, and 32 if crashed.

What do you think?

aciba90:
> I know this has already merged, but I think we need a systemd check around the error checking and handle non-systemd in a more generic way. @aciba90 , you point to the LSB, but the LSB hasn't been updated in 7 years and has never really been respected by many distros. I don't think we can say "do it LSB way or your distro is wrong". There's a few other sections on init scripts, and systemd breaks most of them.

If non-systemd systems / services do not follow a convention we could either:

1) Only distinguish between 0 and non 0 exit codes. Apply the config and restart only if 0.
2) Concrete implementations for non-complaint exit_codes for every non-complaint `(init system, ssh service implementation, distro)`. As for example in System V, each init script implements its status logic.

My gut decision would be to go with 1), as 2) sound complex to maintain. What do you think?

> E.g., on alpine, calling service sshd status returns 1 if not installed, 3 if stopped, and 32 if crashed.

The only problematic thing here is that if `exit_code == 1` we say `service is not available` which is technically correct. I do not see a problem here.

@TheRealFalcon, do you think it makes sense to do a deeper investigation about exit codes?

TheRealFalcon:
To me option 1 makes the most sense with additional support for systemd.

aciba90:
> To me option 1 makes the most sense with additional support for systemd.

Thanks pointing out this situation. PR created: https://github.com/canonical/cloud-init/pull/1450

PR #1426, opened 2022-05-03
Fix bug in url_helper/dual_stack() logging

holmanb:
```
Log all exceptions in dual_stack()
```

While debugging something that uses the dual_stack() function, I noticed that exceptions are not being logged as originally intended.

All exceptions thrown from futures in dual_stack() should be logged for debugging purposes.

### DIFF
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -413,6 +413,7 @@ def dual_stack(
             return_exception = future.exception()
             if return_exception:
                 last_exception = return_exception
+                exceptions.append(last_exception)
             else:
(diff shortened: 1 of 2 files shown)


TheRealFalcon (commented):
Looks good, but probably worth some unit testing checking exception behavior.

holmanb:
@TheRealFalcon Assuming the last commit for black formatting, I think this is ready for re-review.

TheRealFalcon (approved):
LGTM...and I think you had a little too much fun with the text in your tests :smile:

PR #1431, opened 2022-05-05
Systemd and config: align rhel custom files with upstream

esposem:
## Proposed Commit Message
```
So far RHEL had its own custom .service and cloud.cfg files,
that diverged from upstream. We always replaced the generated files
with the ones we had.

This caused only confusion and made it harder to rebase and backport
patches targeting these files.
Hopefully this brings some alignment with upstream.
At the same time, we are going to delete our custom downstream-only files
and use the ones generated by .tmpl.

The mapping is
config/cloud.cfg.tmpl -> rhel/cloud.cfg
systemd/* -> rhel/systemd/*

Such rhel-specific files are open and available in the Centos repo:
https://gitlab.com/redhat/centos-stream/src/cloud-init

```

RHBZ 2082071

## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -34,7 +34,11 @@ disable_root: true
 
 {% if variant in ["almalinux", "alpine", "amazon", "centos", "cloudlinux", "eurolinux",
                   "fedora", "miraclelinux", "openEuler", "rhel", "rocky", "virtuozzo"] %}
+{% if variant == "rhel" %}
+mount_default_fields: [~, ~, 'auto', 'defaults,nofail,x-systemd.requires=cloud-init.service,_netdev', '0', '2']
+{% else %}
 mount_default_fields: [~, ~, 'auto', 'defaults,nofail', '0', '2']
+{% endif %}
 {% if variant == "amazon" %}
 resize_rootfs: noblock
 {% endif %}
@@ -64,6 +68,14 @@ datasource_list: ['NoCloud', 'ConfigDrive', 'Azure', 'OpenStack', 'Ec2']
 # Amazon Linux relies on ec2-net-utils for network configuration
 network:
   config: disabled
+{% endif %}
+
+{% if variant == "rhel" %}
+# Default redhat settings:
+ssh_deletekeys:   1
+ssh_genkeytypes:  ['rsa', 'ecdsa', 'ed25519']
+syslog_fix_perms: ~
+disable_vmware_customization: false
 {% endif %}
 
 # The modules that run in the 'init' stage
@@ -107,10 +119,15 @@ cloud_config_modules:
 {% endif %}
 {% if variant not in ["photon"] %}
  - ssh-import-id
- - keyboard
  - locale
+{% if variant not in ["rhel"] %}
+ - keyboard
+{% endif %}
 {% endif %}
  - set-passwords
+{% if variant in ["rhel"] %}
+ - rh_subscription
+{% endif %}
 {% if variant in ["rhel", "fedora", "photon"] %}
 {% if variant not in ["photon"] %}
  - spacewalk
@@ -239,6 +256,10 @@ system_info:
      name: ec2-user
      lock_passwd: True
      gecos: EC2 Default User
+{% elif variant == "rhel" %}
+     name: cloud-user
+     lock_passwd: true
+     gecos: Cloud User
 {% else %}
      name: {{ variant }}
      lock_passwd: True
@@ -254,6 +275,8 @@ system_info:
      groups: [adm, sudo]
 {% elif variant == "arch" %}
      groups: [wheel, users]
+{% elif variant == "rhel" %}
+     groups: [adm, systemd-journal]
 {% else %}
      groups: [wheel, adm, systemd-journal]
 {% endif %}
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -72,7 +72,7 @@ network:
 
 {% if variant == "rhel" %}
 # Default redhat settings:
-ssh_deletekeys:   1
+ssh_deletekeys:   true
 ssh_genkeytypes:  ['rsa', 'ecdsa', 'ed25519']
 syslog_fix_perms: ~
 disable_vmware_customization: false
@@ -119,10 +119,10 @@ cloud_config_modules:
 {% endif %}
 {% if variant not in ["photon"] %}
  - ssh-import-id
- - locale
 {% if variant not in ["rhel"] %}
  - keyboard
 {% endif %}
+ - locale
 {% endif %}
  - set-passwords
 {% if variant in ["rhel"] %}
--- a/systemd/cloud-config.service.tmpl
+++ b/systemd/cloud-config.service.tmpl
@@ -2,9 +2,7 @@
 [Unit]
 Description=Apply the settings specified in cloud-config
 After=network-online.target cloud-config.target
-{% if variant not in ["rhel"] %}
 After=snapd.seeded.service
-{% endif %}
 Wants=network-online.target cloud-config.target
 {% if variant == "rhel" %}
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
--- a/systemd/cloud-config.service.tmpl
+++ b/systemd/cloud-config.service.tmpl
@@ -2,8 +2,14 @@
 [Unit]
 Description=Apply the settings specified in cloud-config
 After=network-online.target cloud-config.target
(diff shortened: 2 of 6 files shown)


esposem:
@TheRealFalcon @raharper Hi!
I am (finally) trying to align rhel config and systemd files with cloud-init.
This is still work in progress, as we need to make sure that everything works as it should on our side too, but wanted to have some first impression and feedbacks from you too.
Thank you in advance

esposem:
Rhel-specific files are here: https://gitlab.com/redhat/centos-stream/src/cloud-init/-/tree/c9s/rhel

TheRealFalcon:
@esposem , thanks for this! It will be good to bring things closer to upstream. 

Comparing yours to the rhel-specific files you linked, I noticed a few differences:
* `cloud-init-local.service`: Yours missing line `ExecStart=/bin/touch /run/cloud-init/network-config-ready`.
* `cloud-init.service`: Yours includes `DefaultDependencies=no`
* `cloud.cfg`: A number of modules are re-ordered (or added from upstream)

Are all of these intentional?

Additionally, have you been able to build and test your changes?

esposem:
@TheRealFalcon  you are right, I did not notice these differences in `cloud-init-local.service` and `cloud-init.service`.
Regarding `cloud.cfg`, does order matter? I don't think so, right?
In any case, our order is probably pretty much outdated, so keeping yours should be fine.

Regarding testing yes, we have tested the changes (without the above you suggested), and all seems to work fine :tada:

esposem:
Oh sorry I understand what you mean with `cloud.cfg`. If you mean some modules are in the wrong section `cloud_final_modules` vs `cloud_config_modules`, it's an error from our side that actually caused also bugs. That's also why I am trying to align everything with upstream, because the static rhel/ files were not updated and were always replacing your updated files.
Anyways I am pushing again, I think some modules are supported by rhel (at least, by looking at `distros = [... "rhel" ]` in their `cc_module.py` file). I don't think it will hurt us having additional modules (which were supposed to be there anyways following the various releases and rebases).

esposem:
>     * `cloud-init-local.service`: Yours missing line `ExecStart=/bin/touch /run/cloud-init/network-config-ready`.

Actually, this was dropped in https://github.com/canonical/cloud-init/pull/1315 
Therefore I think we should drop it too.

I will test the whole build again, and see if everything works as expected. But I don't see why it shouldn't.

TheRealFalcon:
Once the build passes, let me know and I'll approve the PR.

systemd/cloud-config.service.tmpl
```diff
@@ -2,8 +2,14 @@
 [Unit]
 Description=Apply the settings specified in cloud-config
 After=network-online.target cloud-config.target
+{% if variant not in ["rhel"] %}
 After=snapd.seeded.service
+{% endif %}
 Wants=network-online.target cloud-config.target
+{% if variant == "rhel" %}
+ConditionPathExists=!/etc/cloud/cloud-init.disabled
+ConditionKernelCommandLine=!cloud-init=disabled
```
blackboxsw, line 11: Nice, today I learned about ConditionKernelCommandLine. Thanks!

config/cloud.cfg.tmpl
```diff
@@ -66,6 +70,14 @@ network:
   config: disabled
 {% endif %}
 
+{% if variant == "rhel" %}
+# Default redhat settings:
+ssh_deletekeys:   1
```
blackboxsw, line 75: Please make this `ssh_deletekeys: true` instead of 1. As we are trying to make cloud-init-schema.json more strict instead of permissive on different value types. Tip of master will currently emit a warning in logs (and eventually will error completely) with this log `Cloud config schema errors: ssh_deletekeys: 1 is not of type 'boolean'
`

systemd/cloud-config.service.tmpl
```diff
@@ -2,8 +2,14 @@
 [Unit]
 Description=Apply the settings specified in cloud-config
 After=network-online.target cloud-config.target
+{% if variant not in ["rhel"] %}
 After=snapd.seeded.service
```
blackboxsw, line 6: Any After=X generally won't have applicability/impact on any distro unless they have a snapd.seeded.service installed. Also, since snapd looks like it is installable from the `epel-release` stream, it seems like rhel images could have this installed and I think we still want cloud-config.service to start after the seeding completes.

config/cloud.cfg.tmpl
```diff
@@ -107,10 +119,15 @@ cloud_config_modules:
 {% endif %}
 {% if variant not in ["photon"] %}
  - ssh-import-id
- - keyboard
  - locale
+{% if variant not in ["rhel"] %}
+ - keyboard
```
blackboxsw, line 124: This inadvertently changes the order that modules are run, putting keyboard after locale. Should we move this back above locale?

systemd/cloud-final.service.tmpl
```diff
@@ -15,7 +19,16 @@ ExecStart=/usr/bin/cloud-init modules --mode=final
 RemainAfterExit=yes
 TimeoutSec=0
 KillMode=process
+{% if variant == "rhel" %}
+# Restart NetworkManager if it is present and running.
+ExecStartPost=/bin/sh -c 'u=NetworkManager.service; \
+ out=$(systemctl show --property=SubState $u) || exit; \
+ [ "$out" = "SubState=running" ] || exit 0; \
+ systemctl reload-or-try-restart $u'
```
blackboxsw, line 27: This is an interesting behavior in cloud-final as we should have earlier either emitted the proper network prior to NetworkManager.service first bringup for datasources which support discovery in init-local stage. Or cloud-init should have restarted network in 'init(-network)' stage for datasources detected in the init-network stage just after we write network config. I'm wondering what failure mode we are trying to work around here.

systemd/cloud-init-local.service.tmpl
```diff
@@ -1,23 +1,45 @@
 ## template:jinja
 [Unit]
 Description=Initial cloud-init job (pre-networking)
-{% if variant in ["ubuntu", "unknown", "debian"] %}
+{% if variant in ["ubuntu", "unknown", "debian", "rhel" ] %}
 DefaultDependencies=no
 {% endif %}
 Wants=network-pre.target
+{% if variant not in ["rhel"] %}
 After=hv_kvp_daemon.service
```
blackboxsw, line 9: Are there specific needs where RHEL doesn't want to log failures/warnings Azure's KVP backplane? Or is the desire to avoid this being present on non-Azure platforms?

blackboxsw (commented):
Thank you @esposem for pushing this content upstream. changeset looks pretty good, I am just trying to wrap my head around the intent behind a couple of changes. Please see questions inline.

systemd/cloud-final.service.tmpl
esposem: The original bug is this one: https://bugzilla.redhat.com/show_bug.cgi?id=1897616
Apparently we are covering the case where NM is not installed

systemd/cloud-config.service.tmpl
esposem: Makes sense, I was not sure what it was and decided to go safe and avoid adding something that could potentially break our downstream.

systemd/cloud-init-local.service.tmpl
esposem: Nope, as above I was just not sure. If you tell me that After=X does not have an impact if the service is not installed, then I am happy to leave this for rhel too :)

config/cloud.cfg.tmpl
esposem: Moving it back, thank you

systemd/cloud-init-local.service.tmpl
blackboxsw: Yes confirmed. We ship stock After=hv_kvp_deaemon.sevice definitions on other distros only impact systemd ordering if hv_kvp_daemon.service is present on the system (which should be Azure cloud images)

systemd/cloud-final.service.tmpl
blackboxsw: Ok this works for me. Thanks for the additional context and looks like it's been in streams for a while too! +1

blackboxsw (approved):
LGTM. Thank you @esposem for pushing these changes upstream.

otubo:
Just reviewed the PR and LGTM too. Thanks for taking that, @esposem. Huge benefit for everyone :+1:

esposem:
@TheRealFalcon @blackboxsw downstream tests are passing, feel free to merge this PR :tada:

PR #1432, opened 2022-05-05
run-container: add support for LXD VMs (stable-19.4)

paride:
run-container: add support for LXD VMs

Backport of d1149988323bb88 to stable-19.4.

Images of older releases (e.g. CentOS 7) can't run as containers on
newer Ubuntu releases because they need a CGroupV1 host system. They
can however run as VMs.


paride:
Wrong target branch.

PR #1433, opened 2022-05-05
run-container: add support for LXD VMs (SC-985)

paride:
## Proposed Commit Message

```
run-container: add support for LXD VMs

Backport of d1149988323bb88 to stable-19.4.

Images of older releases (e.g. CentOS 7) can't run as containers on
newer Ubuntu releases because they need a CGroupV1 host system. They
can however run as VMs.
```

## Additional Context

Needed to keep the tests working on the stable-19.4 branch.

## Test Steps

```
./tools/run-container centos/7 --source-package --artifacts=. --vm
```

## Checklist:

### DIFF
--- a/tools/run-container
+++ b/tools/run-container
@@ -41,6 +41,7 @@ Usage: ${0##*/} [ options ] [images:]image-ref
       -p | --package         build a binary package (.deb or .rpm)
       -s | --source-package  build source package (debuild -S or srpm)
       -u | --unittest        run unit tests
+           --vm              use a VM instead of a container
 
     Example:
       * ${0##*/} --package --source-package --unittest centos/6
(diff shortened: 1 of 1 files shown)


paride:
Note that `.travis.yml` from `stable-19.4` is very outdated and expected to fail.

TheRealFalcon:
Where is the `stable-19.4` branch still used?

paride:
Not that we expect much to happen there, but it's tested periodically on CentOS 7, as newer cloud-init versions do not support it anymore.

TheRealFalcon:
I see. This commit is fine, but I'm not sure we need to worry much about Centos 7 anymore. It's in maintenance only mode and probably won't be pulling anything down from upstream anymore other than cherry-picking any critical security issues that could pop up.

PR #1435, opened 2022-05-05
net: Prefer Netplan and NetworkManager renderers by default

Conan-Kudo:
## Proposed Commit Message

```
net: Prefer Netplan and NetworkManager renderers by default

NetworkManager is used by default on a variety of Linux distributions,
and exists as a cross-distribution network management service.

Additionally, add information about the NetworkManager renderer to
the cloud-init documentation.

Because Netplan can be explicitly used to manage NetworkManager,
it needs to be preferred before NetworkManager.

As part of this	change, delete a duplicate test case for network
configuration renderers.

Signed-off-by: Neal Gompa <ngompa13@gmail.com>
```

## Additional Context
This change is a follow-up to #1224, which added the native NetworkManager renderer. This patch has been deployed on Fedora's cloud-init package throughout the development of Fedora Linux 36 to verify that it works.

This should also make it tremendously easier for Linux distributions to use cloud-init because now a standard configuration is supported by default.

## Test Steps
This change is deployed in Fedora Cloud images for Fedora Linux 36, so you can see it in action there.

Otherwise, building an image with a package that has _this_ change and also NetworkManager installed in the image and booting it should result in NetworkManager being used by default instead of any of the legacy ones.

## Checklist:

### DIFF
--- a/cloudinit/net/renderers.py
+++ b/cloudinit/net/renderers.py
@@ -27,10 +27,10 @@ NAME_TO_RENDERER = {
 }
 
 DEFAULT_PRIORITY = [
-    "eni",
-    "sysconfig",
     "netplan",
     "network-manager",
+    "eni",
+    "sysconfig",
     "freebsd",
     "netbsd",
     "openbsd",
--- a/doc/rtd/topics/network-config.rst
+++ b/doc/rtd/topics/network-config.rst
@@ -188,6 +188,15 @@ generated configuration into an internal network configuration state. From
 this state `Cloud-init`_ delegates rendering of the configuration to Distro
 supported formats.  The following ``renderers`` are supported in cloud-init:
 
+- **NetworkManager**
+
+`NetworkManager <https://networkmanager.dev>`_ is the standard Linux network
+configuration tool suite. It supports a wide range of networking setups.
+Configuration is typically stored in ``/etc/NetworkManager``.
+
+It is the default for a number of Linux distributions, notably Fedora;
+CentOS/RHEL; and derivatives.
+
 - **ENI**
 
 /etc/network/interfaces or ``ENI`` is supported by the ``ifupdown`` package
@@ -212,9 +221,10 @@ Network Output Policy
 The default policy for selecting a network ``renderer`` in order of preference
 is as follows:
 
+- Netplan
+- NetworkManager
 - ENI
 - Sysconfig
-- Netplan
 
 When applying the policy, `Cloud-init`_ checks if the current instance has the
 correct binaries and paths to support the renderer.  The first renderer that
@@ -223,7 +233,7 @@ supplying an updated configuration in cloud-config. ::
 
   system_info:
     network:
-      renderers: ['netplan', 'eni', 'sysconfig', 'freebsd', 'netbsd', 'openbsd']
+      renderers: ['netplan', 'network-manager', 'sysconfig', 'eni', 'freebsd', 'netbsd', 'openbsd']
 
 
 Network Configuration Tools
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -7366,16 +7366,14 @@ class TestRenderersSelect:
             ),
             # -netplan +ifupdown -sys -nm -networkd selects eni
             ("eni", False, True, False, False, False),
-            # +netplan +ifupdown -sys -nm -networkd selects eni
-            ("eni", True, True, False, False, False),
-            # +netplan -ifupdown -sys -nm -networkd selects netplan
-            ("netplan", True, False, False, False, False),
+            # +netplan +ifupdown -sys -nm -networkd selects netplan
+            ("netplan", True, True, False, False, False),
             # +netplan -ifupdown -sys -nm -networkd selects netplan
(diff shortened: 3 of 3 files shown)


TheRealFalcon:
@Conan-Kudo ,

As the code is currently, network-manager will still get detected and used if it is installed and no other networking config is detected. Other renderers will just take priority if they get detected first. In the case of netplan, we want this because it is often expected that both netplan and network manager be installed at the same time. People with netplan installed want to use netplan, so netplan must come before network-manager.

For eni and sysconfig, I'm more concerned about breaking changes. If somebody running cloud-init 22.1 with eni or sysconfig upgrades to 22.2 containing this change, their network config will change from using eni or sysconfig to network-manager. Theoretically this should be harmless, but I'm concerned about a few things:
1. Cloud-init having written multiple network config files for multiple active network utilities could cause additional bugs.
2. There's a fairly large regression risk given how new the network-manager renderer is
3. There are almost certainly breaking differences between all of the network renderers

Because of these risks, I just don't see any benefit in putting network-manager first given that it will still be selected if no other networking configuration exists. Unless there's additional context that I'm not seeing, I'm -1 on this change (but would keep a docs change).

If you still strongly feel that network-manager should be first in Fedora, you could drop a file in /etc/cloud/cloud.cfg.d/ on Fedora images to change the [priority](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/__init__.py#L120), or if you want to override that entire `_write_network_state` method for Fedora and specify a custom priority for Fedora only, that would work too.

TheRealFalcon:
@Conan-Kudo 

> Is that the case anywhere for cases where cloud-init is used right now? Also, Ubuntu already ships altered configuration to use netplan first regardless of upstream.

It's not just about stock Ubuntu instances. Anybody can create an image with netplan installed. If we change this default, this can break anybody with both netplan and NetworkManger installed, which is a primary use case for netplan. If we don't change this default, nothing bad happens to somebody using NetworkManager unless they already have NetworkManager installed, decides to install netplan, and then decides they don't want netplan. The former is way more likely than the latter. I see no advantage at all to having network-manager above netplan. I only see regression risk. To the question you asked, I have no idea, and that's the point.

> The problem is that because NetworkManager supports reading both eni and sysconfig files with varying degrees of compatibility, those backends getting used causes problems for directly configuring NetworkManager. It has to be before those two.

Does eni actually represent a problem here? It doesn't get detected unless `ifquery` is found, which isn't installed by default on any RH based systems. For ifcfg, I get that you've had to carry a patch downstream, and that's unfortunate, but I don't see how that justifies breaking _currently working_ systems.

For context, when ubuntu switched to netplan as the default, [we had to patch all of the existing releases](https://github.com/canonical/cloud-init/blob/ubuntu/xenial/debian/patches/renderer-do-not-prefer-netplan.patch) to ensure that instances that were already launched didn't get their networking randomly changed on upgrade.

If we change these defaults, existing systems can break on upgrade. It would also require existing downstreams to patch it out until they're ready to switch. In light of that, it seems to me that carrying the ifcfg patch until Fedora 36 and RHEL 9 launch isn't that big of a deal. If the patches worked before, they will continue to work. When the ifcfg-rh plugin is no longer shipped, those patches can be removed and the default order will work as expected.

blackboxsw:
> If you still strongly feel that network-manager should be first in Fedora, you could drop a file in /etc/cloud/cloud.cfg.d/ on Fedora images to change the [priority](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/__init__.py#L120), or if you want to override that entire `_write_network_state` method for Fedora and specify a custom priority for Fedora only, that would work too.

Or an alternative would be to specify the renderer priority order for fedora/centos/rhel/rocky in the [cloud.cfg.tmpl](https://github.com/canonical/cloud-init/blob/main/config/cloud.cfg.tmpl). Though I'm not certain if fedora sources that rendered content during packaging?

Conan-Kudo:
> It's not just about stock Ubuntu instances. Anybody can create an image with netplan installed. If we change this default, this can break anybody with both netplan and NetworkManger installed, which is a primary use case for netplan. If we don't change this default, nothing bad happens to somebody using NetworkManager unless they already have NetworkManager installed, decide to install netplan, and then decide they don't want netplan. The former is way more likely than the latter. I see no advantage at all to having network-manager above netplan. I only see regression risk. To the question you asked, I have no idea, and that's the point.

So then if we changed it so it was netplan -> nm -> eni -> sysconfig, then it'd be acceptable, is that it?

TheRealFalcon:
> So then if we changed it so it was netplan -> nm -> eni -> sysconfig, then it'd be acceptable, is that it?

Please see the 2nd half of the comment you're responding to. It's an argument for not changing the order at all. If we were designing this from scratch, I would agree with you, but it's more important to not introduce regressions.

esposem:
Alternative PR to apply such changes only to RHEL and Fedora: #1465 
I think this is what @blackboxsw  is suggesting too
@Conan-Kudo you need to align Fedora to drop the hardcoded `cloud.cfg` file just as we did in RHEL. You probably need to also add "fedora" in some conditional statements in `cloud.cfg.tmpl`.

PR #1438, opened 2022-05-05
Fix network-manager activator availability and order (SC-991)

TheRealFalcon:
## Proposed Commit Message

```
Fix network-manager activator availability and order

The current network-manager activator availability check is too broad.
It triggers as available if chrony is installed. Its priority is also
higher than netplan, which means it can activate even if netplan config
has been rendered. This commit reverts the network-manager check to its
previous state and orders netplan above network-manager in priority.
```

## Additional Context
Currently the check for network-manager only checks if '/usr/lib/NetworkManager' exists ([here](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/network_manager.py#L372) and [here](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/network_manager.py#L22))

On an Azure instance, after upgrade (tests/integration_tests/test_upgrade.py::test_clean_boot_of_upgraded_package) I can see the following in the logs:
```
2022-05-05 18:10:14,595 - __init__.py[DEBUG]: Selected renderer 'netplan' from priority list: ['netplan', 'eni', 'sysconfig']
2022-05-05 18:10:14,595 - netplan.py[DEBUG]: V2 to V2 passthrough
2022-05-05 18:10:14,596 - util.py[DEBUG]: Writing to /etc/netplan/50-cloud-init.yaml - wb: [644] 481 bytes
2022-05-05 18:10:14,596 - subp.py[DEBUG]: Running command ['netplan', 'generate'] with allowed return codes [0] (shell=False, capture=True)
2022-05-05 18:10:14,686 - subp.py[DEBUG]: Running command ['udevadm', 'test-builtin', 'net_setup_link', '/sys/class/net/lo'] with allowed return codes [0] (shell=False, capture=True)
2022-05-05 18:10:14,690 - subp.py[DEBUG]: Running command ['udevadm', 'test-builtin', 'net_setup_link', '/sys/class/net/eth0'] with allowed return codes [0] (shell=False, capture=True)
2022-05-05 18:10:14,695 - __init__.py[DEBUG]: Bringing up newly configured network interfaces
2022-05-05 18:10:14,695 - activators.py[DEBUG]: Using selected activator: <class 'cloudinit.net.activators.NetworkManagerActivator'>
2022-05-05 18:10:14,695 - activators.py[DEBUG]: Attempting command ['nmcli', 'connection', 'load', '/etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection'] for device eth0
2022-05-05 18:10:14,695 - subp.py[DEBUG]: Running command ['nmcli', 'connection', 'load', '/etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection'] with allowed return codes [0]
(shell=False, capture=True)
2022-05-05 18:10:14,697 - util.py[WARNING]: Running interface command ['nmcli', 'connection', 'load', '/etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection'] failed
2022-05-05 18:10:14,697 - util.py[DEBUG]: Running interface command ['nmcli', 'connection', 'load', '/etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection'] failed
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 291, in subp
    sp = subprocess.Popen(
  File "/usr/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.9/subprocess.py", line 1821, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: b'nmcli'
```

network-manager is not installed, but the chrony package is creating the directory we use to check if network-manager is installed:
```
ubuntu@SRU-worked:~$ dpkg -L chrony | grep "/usr/lib/NetworkManager"
/usr/lib/NetworkManager
/usr/lib/NetworkManager/dispatcher.d
/usr/lib/NetworkManager/dispatcher.d/20-chrony-dhcp
/usr/lib/NetworkManager/dispatcher.d/20-chrony-onoffline
```

Because of this, I reverted the check back to checking for `/etc/NetworkManager/NetworkManager.conf` and the existance of `nmcli`.

@lkundrak and @Conan-Kudo , does this seem reasonable to you?

## Test Steps
```
DEB_BUILD_OPTIONS=nocheck packages/bddeb -d
CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_all.deb CLOUD_INIT_PLATFORM=azure CLOUD_INIT_OS_IMAGE=impish pytest tests/integration_tests/test_upgrade.py::test_clean_boot_of_upgraded_package
```

### DIFF
--- a/cloudinit/net/activators.py
+++ b/cloudinit/net/activators.py
@@ -255,8 +255,8 @@ class NetworkdActivator(NetworkActivator):
 # version to encompass both seems overkill at this point
 DEFAULT_PRIORITY = [
     IfUpDownActivator,
-    NetworkManagerActivator,
     NetplanActivator,
+    NetworkManagerActivator,
     NetworkdActivator,
 ]
 
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -373,7 +373,14 @@ def conn_filename(con_id, target=None):
 def available(target=None):
     config_present = os.path.isfile(subp.target_path(target, path=NM_CFG_FILE))
     nmcli_present = subp.which("nmcli", target=target)
-    return config_present and bool(nmcli_present)
+    service_active = True
+    if subp.which("systemctl"):
+        try:
+            subp.subp(["systemctl", "is-active", "NetworkManager.service"])
+        except subp.ProcessExecutionError:
+            service_active = False
+
+    return config_present and bool(nmcli_present) and service_active
 
 
 # vi: ts=4 expandtab
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -20,6 +20,7 @@ from . import renderer
 
 NM_RUN_DIR = "/etc/NetworkManager"
 NM_LIB_DIR = "/usr/lib/NetworkManager"
+NM_CFG_FILE = "/etc/NetworkManager/NetworkManager.conf"
 LOG = logging.getLogger(__name__)
 
 
@@ -370,8 +371,9 @@ def conn_filename(con_id, target=None):
 
 
 def available(target=None):
-    target_nm_dir = subp.target_path(target, NM_LIB_DIR)
-    return os.path.exists(target_nm_dir)
+    config_present = os.path.isfile(subp.target_path(target, path=NM_CFG_FILE))
+    nmcli_present = subp.which("nmcli", target=target)
+    return config_present and bool(nmcli_present)
 
 
 # vi: ts=4 expandtab
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -380,7 +380,7 @@ def available(target=None):
     service_active = True
     if uses_systemd():
         try:
-            subp.subp(["systemctl", "is-active", "NetworkManager.service"])
+            subp.subp(["systemctl", "is-enabled", "NetworkManager.service"])
         except subp.ProcessExecutionError:
             service_active = False
 
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -371,10 +371,14 @@ def conn_filename(con_id, target=None):
 
 
 def available(target=None):
+    # TODO: Move `uses_systemd` to a more appropriate location
+    # It is imported here to avoid circular import
+    from cloudinit.distros import uses_systemd
+
     config_present = os.path.isfile(subp.target_path(target, path=NM_CFG_FILE))
     nmcli_present = subp.which("nmcli", target=target)
     service_active = True
-    if subp.which("systemctl"):
+    if uses_systemd():
         try:
             subp.subp(["systemctl", "is-active", "NetworkManager.service"])
         except subp.ProcessExecutionError:
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -33,7 +33,6 @@ KNOWN_DISTROS = [
     "suse",
     "virtuozzo",
 ]
-NM_CFG_FILE = "/etc/NetworkManager/NetworkManager.conf"
 
 
 def _make_header(sep="#"):
--- a/tests/unittests/test_net_activators.py
+++ b/tests/unittests/test_net_activators.py
@@ -44,7 +44,9 @@ NETPLAN_CALL_LIST = [
 def available_mocks():
     mocks = namedtuple("Mocks", "m_which, m_file, m_exists")
     with ExitStack() as mocks_context:
-        mocks_context.enter_context(patch("cloudinit.subp.subp"))
+        mocks_context.enter_context(
+            patch("cloudinit.distros.uses_systemd", return_value=False)
+        )
         m_which = mocks_context.enter_context(
             patch("cloudinit.subp.which", return_value=True)
         )
@@ -61,7 +63,9 @@ def available_mocks():
 def unavailable_mocks():
     mocks = namedtuple("Mocks", "m_which, m_file, m_exists")
     with ExitStack() as mocks_context:
-        mocks_context.enter_context(patch("cloudinit.subp.subp"))
+        mocks_context.enter_context(
+            patch("cloudinit.distros.uses_systemd", return_value=False)
+        )
         m_which = mocks_context.enter_context(
             patch("cloudinit.subp.which", return_value=False)
         )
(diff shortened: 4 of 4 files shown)


cloudinit/net/activators.py
```diff
@@ -255,8 +255,8 @@ def bring_down_interface(device_name: str) -> bool:
 # version to encompass both seems overkill at this point
 DEFAULT_PRIORITY = [
     IfUpDownActivator,
-    NetworkManagerActivator,
     NetplanActivator,
+    NetworkManagerActivator,
     NetworkdActivator,
```
Conan-Kudo, line 260: I don't think this is necessary when you've changed it to check if the `nmcli` tool exists.

cloudinit/net/network_manager.py
```diff
@@ -370,8 +371,9 @@ def conn_filename(con_id, target=None):
 
 
 def available(target=None):
-    target_nm_dir = subp.target_path(target, NM_LIB_DIR)
-    return os.path.exists(target_nm_dir)
+    config_present = os.path.isfile(subp.target_path(target, path=NM_CFG_FILE))
+    nmcli_present = subp.which("nmcli", target=target)
```
Conan-Kudo, line 375: Wouldn't it also make sense to check if the NetworkManager service is enabled?

Conan-Kudo (changes requested):

cloudinit/net/activators.py
blackboxsw: Thanks for the review comments and previous PR too!

The presence of nmcli (just as the presence of systemd-networkd) doesn't mean that cloud-init should primarily use those tools directly to bring up specific interfaces. Given that netplan renders configuration aritfacts to either network backend it should be preferred because Images which have netplan.io installed have made an opinionated decision to install netplan to manage network config.  It feels a bit like undercutting that decision by going directly to the network backend tools even when netplan is installed and will likely result in more cases of misconfiguration as some configuraiton defined up in /etc/netplan/* config may have defined a separate `renderer: sytemd-networkd` configuration which nmcli will know nothing about.

Conan-Kudo: That makes sense. I have also adjusted my PR (#1435) to now order Netplan at the highest for that reason.

cloudinit/net/network_manager.py
TheRealFalcon: Thanks for the comment @Conan-Kudo . I have updated the PR to take service status into account.

Conan-Kudo: We'd probably want this for Netplan too, right?

TheRealFalcon: Netplan doesn't have a service associated with it, so I don't think there's anything more to be done.

cloudinit/net/network_manager.py
```diff
@@ -370,8 +371,16 @@ def conn_filename(con_id, target=None):
...
+    service_active = True
+    if subp.which("systemctl"):
```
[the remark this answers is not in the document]
blackboxsw: Comparison which vs uses_systemd
```bash
csmith@uptown:~/src/cloud-init$ cat 1.py
#!/usr/bin/python3
from cloudinit.subp import which
from cloudinit.distros import uses_systemd

def f1():
    which("systemctl")

def f2():
     uses_systemd()

if __name__ == "__main__":
  import timeit
  print(timeit.timeit('f1()', setup="from __main__ import f1", number=10000))
  print(timeit.timeit('f2()', setup="from __main__ import f2", number=10000))
csmith@uptown:~/src/cloud-init$ ./1.py 
0.6738533910029219
0.01924525299909874
```

blackboxsw (changes requested):
Good to tighten down the conditions under which we claim an activator is `available` and used. 

We probably at some point want to grow our `distros.manage_services` shim to handle this a bit more simply across distributions. But, that's some we can work toward longer term.

cloudinit/net/network_manager.py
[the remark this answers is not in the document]
TheRealFalcon: Yeah, I was specifically limiting this to systemd but not gating on it (but to your example. Alpine also doesn't use NetworkManager by default).

I tried `uses_systemd` at first, but that introduces a circular import. I thought that `subp.which` would be better than using a local import, but I can switch to the local import for now.

cloudinit/net/network_manager.py
```diff
@@ -370,8 +371,20 @@ def conn_filename(con_id, target=None):
...
+    # TODO: Move `uses_systemd` to a more appropriate location
+    # It is imported here to avoid circular import
+    from cloudinit.distros import uses_systemd
+
+    config_present = os.path.isfile(subp.target_path(target, path=NM_CFG_FILE))
+    nmcli_present = subp.which("nmcli", target=target)
+    service_active = True
+    if uses_systemd():
+        try:
+            subp.subp(["systemctl", "is-active", "NetworkManager.service"])
```
blackboxsw, line 383: TLDR: I think we actually want `is-enabled` not `is-active` to assert NetworkManager.service is "turned on". `is-active` means the service is running, not whether it is scheduled to run.

The more I test this, the more I can confirm there is a race in init-local stage with `systemctl is-active NetworkManager.service` because when network is being rendered, the network services have not yet come up fully.  NetworkManager.service is in fact blocked until `After=network-pre.target` but cloud-init-local.service (at least on ubuntu) is `Before=network-pre.target`. So, is-active will return `inactive` in this boot stage and exit non-zero.

I can confirm this behavior when installing network-manager on Ubuntu server and setting renderers: [network-manager, netplan]. `network_manager.available` will always return False in init-local stage and the default renderer chosen remains netplan. Given that both activators and renderers are use the same `network_manager.available` function we'll find ourselves in a state where NM config is rendered, but if activators are triggered, netplan is used to activate the network without any network config written.

blackboxsw (changes requested):

blackboxsw (approved):
Confirmed behavior on an Ubuntu system with both netplan and network-manager and an /etc/cloud/cloud.cfg specifying `network: renderers: [network-manager, netplan, sysconfig]` that network-manager renderer can be chosen as renderer in init-local stage. Validated failure of integration test without this PR and resolution with PR applied.

Long-term I wonder if we need to provide a mechanism by which network: activators can be specified with a priority override in /etc/cloud/cloud.cfg. Someone that wants to override network: renderers will also likely want to override the activators.

PR #1440, opened 2022-05-09
Handle file permission errors in commands

aciba90:
## Proposed Commit Message

```
Redact files with permission errors in commands.

Add tests covering this behavior for query, status and render
cmds.

Migrate `test_render.py` and `test_status.py` to Pytest.

LP: #1953430
SC-658
```

## Test Steps

```sh
$ CONTAINER_NAME=some-ubuntu
$ lxc launch ubuntu-daily:jammy $CONTAINER_NAME
$ lxc exec $CONTAINER_NAME -- cloud-init status --wait
$ lxc exec $CONTAINER_NAME -- \
    chmod 400 '/etc/cloud/cloud.cfg.d/90_dpkg.cfg'

# Case 1
$ lxc exec $CONTAINER_NAME --user 1000 -- cloud-init query --all
<config with redacted parts>

# Case 2
$ lxc exec $CONTAINER_NAME --user 1000 -- cloud-init status --long
status: done
time: Tue, 10 May 2022 07:25:25 +0000
detail:
DataSourceLXD
```
Case 1 did exist with 1. Shows config with redacted parts.
Case 2 did raise an exception. Shows correct output. Log messages do not appear because of how the log is configured atm.

## Checklist:


blackboxsw (changes requested):
Thanks for this PR @aciba90 and for both pytest migrations and type hints.

While the solution you have cleans up the traceback, I think we might want to make the behavior of  `cloud-init status` and `cloud-init query` still work for non-root users. Raising the exception prevents any non-root user from querying `cloud-init status` or `cloud-init query` until a root admin removes or changes perms on the offending files.  If we can, let's instead emit REDACTED warnings for any configuration file that a non-root user doesn't have permission to that aligns with the messaging we currently have for `cloud-init query vendordata` for non-root. Given that `cloud-init status` and `cloud-init query` are really only trying to source Paths.run_dir the likelihood that a root-read-only file in /etc/cloud/cloud.cfg.d is overriding `system_info: paths:`  is very slim to have an impact on the functionality of those tools.

As such, I'm thinking we can proceed with something more like the following:
```diff
diff --git a/cloudinit/util.py b/cloudinit/util.py
index 9fb0c6616..90c6c7a47 100644
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -32,7 +32,7 @@ import subprocess
 import sys
 import time
 from base64 import b64decode, b64encode
-from errno import ENOENT
+from errno import EACCES, ENOENT
 from functools import lru_cache
 from typing import List
 from urllib import parse
@@ -1010,8 +1010,13 @@ def read_conf_d(confd):
     # Load them all so that they can be merged
     cfgs = []
     for fn in confs:
-        cfgs.append(read_conf(os.path.join(confd, fn)))
-
+        try:
+            cfgs.append(read_conf(os.path.join(confd, fn)))
+        except OSError as e:
+            if e.errno == EACCES:
+                LOG.warning(
+                    f"REDACTED config part {confd}/{fn} for non-root user"
+                )
     return mergemanydict(cfgs)

```


Additionally something I'd like to see in this PR is that if we get an error in `cloud-init query --all` due to permissions issues or some other unexpected error, we get no message printed to console, just and exit 1. I'd like us to [handle exceptions and emit an appropriate error message](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/query.py#L263-L269) in that event.

cloudinit/cmd/status.py
```diff
@@ -64,17 +65,16 @@ def get_parser(parser=None):
     return parser
 
 
-def handle_status_args(name, args):
+def handle_status_args(name, args) -> int:
```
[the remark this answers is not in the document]
aciba90: It is a pleasure. I truly believe from past experiences that the type system helps to catch early bugs.

cloudinit/cmd/devel/__init__.py
```diff
@@ -16,10 +19,22 @@ def addLogHandlerCLI(logger, log_level):
...
-    """Return a Paths object based on the system configuration on disk."""
+def read_cfg_paths() -> Paths:
+    """Return a Paths object based on the system configuration on disk.
+
+    It handles file permission errors.
+    """
     init = Init(ds_deps=[])
-    init.read_cfg()
+    try:
+        init.read_cfg()
```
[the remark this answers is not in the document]
aciba90: Problematic files redacted.

aciba90:
Changed to redact configurations coming from files with permission errors.
Extra tests added to check that the sys-output is empty and that the log messages are correctly emitted.

Thank you for the comments and review. It is ready to be reviewed @blackboxsw

cloudinit/stages.py
```diff
@@ -953,23 +956,38 @@ def should_run_on_boot_event():
...
-            # runtime config. I.e., /run/cloud-init/cloud.cfg
-            read_runtime_config(),
-            # Kernel/cmdline parameters override system config
-            util.read_conf_from_cmdline(),
-        ],
-        reverse=True,
-    )
+    config_loaders = [
+        # builtin config, hardcoded in settings.py.
+        util.get_builtin_cfg,
```
blackboxsw, line 973: While I appreciate the sentiment in the event we get permissions errors, the following callables will not generate EACCES errors so it might be over-engineering to cope with scenarios which won't raise EACCESS:
- `get_builtin_cfg` (we are sourcing a static variable in the python code)
- `read_runtime_config` pulls from /run/cloud-init which we set at world-readable
- `read_conf_from_cmdline` pulls from /proc/cmdline which is also world-readable


The only function/callable that could generate that EACCES is read_conf_with_confd, so I don't think we need to add the layer of test/except handling here in a for loop if we handle this warning down in `read_conf_with_confd`

cloudinit/util.py
```diff
@@ -1010,13 +1011,24 @@ def read_conf_d(confd):
...
 
 
 def read_conf_with_confd(cfgfile):
-    cfg = read_conf(cfgfile)
+    try:
+        cfg = read_conf(cfgfile)
+    except OSError as e:
+        if e.errno == EACCES:
+            LOG.warning("REDACTED config part %s for non-root user", cfgfile)
+        raise
```
blackboxsw, line 1031: I don't think we want a raise here, we want to proceed to other "conf_d" files if available.

cloudinit/cmd/devel/render.py
```diff
@@ -63,7 +63,10 @@ def handle_args(name, args):
     if args.instance_data:
         instance_data_fn = args.instance_data
     else:
-        paths = read_cfg_paths()
+        try:
+            paths = read_cfg_paths()
+        except (IOError, OSError):
+            return 1
```
blackboxsw, line 69: Since init.read_cfg() no longer raises an OSError/IOError (and instead logs warnings) I don't think we need this try/except here anymore right?

cloudinit/stages.py
```diff
@@ -953,23 +956,38 @@ def should_run_on_boot_event():
 
 
 def read_runtime_config():
-    return util.read_conf(RUN_CLOUD_CONFIG)
```
blackboxsw, line 956: Cloud-init owns and creates this file as world-readable, so I don't believe we need the try/except handling around this.

blackboxsw (changes requested):
Thanks @aciba90 a couple of notes on this behavior I think we may be able to drop a couple of the exception handling cases since you are coping with EACCES errors within `read_conf_d`.

Please let me know if I've misunderstood some elements.

cloudinit/stages.py
aciba90: Removed, thanks.

cloudinit/stages.py
```diff
@@ -953,23 +956,38 @@ def should_run_on_boot_event():
...
+    ]
+    configs = []
+    for config_loader in config_loaders:
+        try:
+            configs.append(config_loader())
+        except OSError as e:
+            if e.errno == errno.EACCES:
+                pass  # Already logged in `config_loaders`.
+            else:
+                raise
```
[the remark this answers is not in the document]
aciba90: Done

cloudinit/util.py
aciba90: Reworked to proceed to other "conf_d files" if available.

cloudinit/stages.py
aciba90: After previous modifications, this is definitely not needed. Removed.

cloudinit/cmd/devel/render.py
aciba90: Removed.

blackboxsw (approved):
Good work on this PR and addressing all facets of the failure. Validated on server live install environments. LGTM

PR #1441, opened 2022-05-10
testing: Make integration tests kinetic friendly

TheRealFalcon:
## Proposed Commit Message

```
testing: Make integration tests kinetic friendly
```

### DIFF
--- a/tests/integration_tests/datasources/test_lxd_discovery.py
+++ b/tests/integration_tests/datasources/test_lxd_discovery.py
@@ -9,7 +9,7 @@ from tests.integration_tests.util import verify_clean_log
 
 
 def _customize_envionment(client: IntegrationInstance):
-    # Assert our platform can detect LXD during sytemd generator timeframe.
+    # Assert our platform can detect LXD during systemd generator timeframe.
     ds_id_log = client.execute("cat /run/cloud-init/ds-identify.log").stdout
     assert "check for 'LXD' returned found" in ds_id_log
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests/datasources/test_lxd_discovery.py
```diff
@@ -35,20 +35,10 @@ def _customize_envionment(client: IntegrationInstance):
         "/etc/cloud/cloud.cfg.d/99-detect-lxd-first.cfg",
         "datasource_list: [LXD, NoCloud]\n",
     )
-    if ImageSpecification.from_os_image().release == "jammy":
```
TheRealFalcon, line 38: @blackboxsw , this wasn't specifically related, but I don't see why this section would be needed/wanted in tests testing the LXD datasource. Is it ok to remove this or is there something I'm missing?

blackboxsw: I had originally put this in here for two reasons:
1. allow ds-identify to detect two vaible datasources and ensure a tracer through the system was available to ensure a tracer through the system the even when multiple datasources are detected, priority order is honored when we get to python datasource get_data discovery. But, that is test case is handled anyway though by integration tests `test_combined.py` for each platform which don't modify default datasource_list config and manage to detect the right datasource).
2. I also had the thought I wanted to be sure that we [configure more than 1 datasource so ds-identify actually does run the `dscheck_LXD` and `dscheck_NoCloud` functions instead of skipping those discovery functions](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L1797-L1803). But, I bet we can drop the seed file creation for two reasons:
   * we are configuring `datasouce_list: [LXD, NoCloud]` in `/etc/cloud/cloud.cfg.d` in this test which is what feeds ds-id's `read_datasource_list` so we don't limit ds-identify behavior because > 1 elements are in `datasource_list`
   * Any Bionic, Focal and Impish test is going to detect both NoCloud (from seed files) and LXD (from socket presence) so we have the inverse ordering test where 2 viable datasources are detected, but default priority order is honored by selecting NoCloud over LXD.

TheRealFalcon: Ah, thanks. That makes sense. I can put it back.

blackboxsw (approved):
LGTM . Will land when CI passes.

PR #1444, opened 2022-05-11
Github Actions

holmanb:
```
Move "formatting tests" to Github Actions
```

Start with formatting tests, we can always go back to travis.

## Benefits

- Github integration means users see test result on local branch by default, no setup
- Lower friction UI and UX due to to integration
- [free](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) (I think?)

#### Note:

the bonus tests (failing) will obviously be dropped before merge

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -1,0 +1,33 @@
+on:
+  push:
+  pull_request:
+
+jobs:
+  check_format:
+    strategy:
+      fail-fast: false
+      matrix:
+        env: [flake8, mypy, pylint, black, isort]
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout #1
+        uses: actions/checkout@v3.0.0
+
+      - name: Checkout #2 (for tools/read-version)
+        run: |
+          git fetch --unshallow
+          git remote add upstream https://git.launchpad.net/cloud-init
+
+      - name: Dependencies
+        run: |
+          sudo apt update
+          sudo apt install tox
+
+      - name: Print version
+        run: python --version
+
+      - name: Test
+        env:
+          # matrix env: not to be confused w/environment variables or testenv
+          TOXENV: ${{ matrix.env}}
+        run: tox
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -20,8 +20,8 @@ jobs:
 
       - name: Dependencies
         run: |
-          sudo apt update
-          sudo apt install tox
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
 
       - name: Print version
         run: python --version
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -31,7 +31,7 @@ jobs:
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
 
       - name: Print version
-        run: python --version
+        run: python3 --version
 
       - name: Test
         env:
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -1,3 +1,4 @@
+name: Lint Tests
 on:
   push:
   pull_request:
@@ -36,5 +37,5 @@ jobs:
       - name: Test
         env:
           # matrix env: not to be confused w/environment variables or testenv
-          TOXENV: ${{ matrix.env}}
+          TOXENV: ${{ matrix.env }}
         run: tox
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -2,6 +2,13 @@ on:
   push:
   pull_request:
 
+concurrency:
+  group: 'ci-${{ github.workflow }}-${{ github.ref }}'
+  cancel-in-progress: true
+defaults:
+  run:
+    shell: sh -ex {0}
+
 jobs:
   check_format:
     strategy:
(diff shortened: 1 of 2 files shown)


paride (commented):
This looks great. I left some minor inline comments. Another couple of useful settings we use in the UA CI are:
```
# Cancel any in-progress job or run
concurrency:
  group: 'ci-${{ github.workflow }}-${{ github.ref }}'
  cancel-in-progress: true

defaults:
  run:
    shell: sh -ex {0}
```
The first is probably only really useful in longer runs (e.g. integration tests).

The second basically makes `sh -ex` the default interpreter for `run:`.

.github/workflows/check_format.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Checkout #2 (for tools/read-version)
+        run: |
+          git fetch --unshallow
+          git remote add upstream https://git.launchpad.net/cloud-init
+
+      - name: Dependencies
+        run: |
+          sudo apt update
+          sudo apt install tox
```
paride, line 24: Better in automation:
```suggestion
          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox
```

.github/workflows/check_format.yml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+  pull_request:
+
+jobs:
+  check_format:
+    strategy:
+      fail-fast: false
+      matrix:
+        env: [flake8, mypy, pylint, black, isort]
+    runs-on: ubuntu-18.04
```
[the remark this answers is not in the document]
holmanb: No :)

I was testing with 20.04 previously, but since our current format tests run on 3.6 in travis, I downgraded to 18.04.

holmanb:
> This looks great. I left some minor inline comments. Another couple of useful settings we use in the UA CI are:

Thanks for the review @paride! I just included your suggestions.

.github/workflows/check_format.yml
```diff
@@ -0,0 +1,40 @@
...
+defaults:
+  run:
+    shell: sh -ex {0}
+
+jobs:
+  check_format:
+    strategy:
+      fail-fast: false
+      matrix:
+        env: [flake8, mypy, pylint, black, isort]
```
paride, line 17: I wonder if it's really worth having one test matrix element per TESTENV, instead of doing something like `tox -e flake8 mypy pylint black isort` or `tox --parallel -e ...` or similar via `TOXENV`.

We may want to run some environments on different Ubuntu releases (e.g. flake8 or pylint), to test them with different Python versions, and the test matrix easily starts growing. AIUI the GitHub enterprise account should allow 180 concurrent jobs (https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration), which is not little but also not infinite, so maybe we should use less jobs when it makes sense?

(I see you just replicated what we have in Travis, but maybe it's the right time to discuss this.)

paride (commented):
This basically LGTM, I just added another couple of inline comments/questions.

On merging: I think it's fine having this coexist with Travis for a little while, so we can still merge PRs with confidence even if this Actions workflow has any issue.

.github/workflows/check_format.yml
holmanb: If I understand our recent discussion about this correctly, it sounded like (please correct me if I'm wrong), we might want to hold off on increasing this for a) more information about number of jobs and tests concurrently on the enterprise account and b) potential for self-hosted test runners providing this coverage.

paride: Yes that's it, and again I tried to ask about the status of the github.com/canonical account but apparently I don't know *who* to ask as I've got no answer.

What do you see as the main advantage of having each environment in a separate job? Is it speed (everything parallel) or "failure clarity" (it's immediate to see which job failed, and therefore which environment failed)? If it's the latter, then a middle ground solution could be calling environments in separate steps, e.g.
```
- name: Tox: flake8
  run: tox -e pyenv
- name: Tox: black
  run: tox -e black
- ...
```
This keeps things pretty clear in the workflow logs, but clearly it will be slightly slower.

holmanb: > Yes that's it, and again I tried to ask about the status of the github.com/canonical account but apparently I don't know who to ask as I've got no answer.

Thanks for trying :)

Comments in the `.travis.yml` imply that it was optimized for fast failure reporting (tests run in parallel and report early) as well as for total time (comment implies that ordering integration tests before others means it starts earlier). These are defaults that I personally like, but I don't feel strongly about this if we think a different optimum would be better.

"Waiting for tests to finish" is usually something which can be done asynchronously, however serializing the tests would make the wait several minutes longer so I think more feedback from regular contributors is probably warranted if we're considering that. @blackboxsw @TheRealFalcon @aciba90 or others, any objections, preferences. or concerns on this?

paride (approved):
Replied to the "matrix vs serial" inline thread, but this LGTM already.

.github/workflows/check_format.yml
TheRealFalcon: Most of the formatting checkers take <5 seconds, so I think it's fine to run them serially. I'd put pylint at the end as that usually takes the longest. If/When we start introducing the unit tests and integration tests, I would probably start running those in parallel as they can take longer to run.

I'm not strongly opinionated at this point though, so I'm +1 either way.

TheRealFalcon (approved):

.github/workflows/check_format.yml
TheRealFalcon: To @paride's concern, given that it's just a single yaml file to update, I'm not really concerned about runner availability until it actually becomes a problem. If it does become a problem, it should be a fairly quick fix to run more things serially as needed.

PR #1448, opened 2022-05-11
Drop cc_snap.maybe_install_squashfuse.

aciba90:
cc_snap.maybe_install_squashfuse no longer needed in Bionic++.

squashfuse install no longer required as snapd feature was backported
and released (and Xenial support is no longer available for
cloud-init).

## Proposed Commit Message

```
Drop cc_snap.maybe_install_squashfuse.

cc_snap.maybe_install_squashfuse no longer needed in Bionic++.

squashfuse install no longer required as snapd feature was backported
and released (and Xenial support is no longer available for
cloud-init).
```

## Additional Context

SC-903

## Checklist:

### DIFF
--- a/cloudinit/config/cloud-init-schema.json
+++ b/cloudinit/config/cloud-init-schema.json
@@ -1819,11 +1819,6 @@
                   {"type": "array", "items": {"type": "string"}}
                 ]
               }
-            },
-            "squashfuse_in_container": {
-              "type": "boolean",
-              "default": false,
-              "description": "DEPRECATED: This key is no longer needed and will be removed in a future version of cloud-init."
             }
           }
         }
--- a/cloudinit/config/cc_snap.py
+++ b/cloudinit/config/cc_snap.py
@@ -182,23 +182,6 @@ def run_commands(commands):
         raise RuntimeError(msg)
 
 
-# RELEASE_BLOCKER: Once LP: #1628289 is released on xenial, drop this function.
-def maybe_install_squashfuse(cloud):
-    """Install squashfuse if we are in a container."""
-    if not util.is_container():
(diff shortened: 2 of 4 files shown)


cloudinit/config/cloud-init-schema.json
```diff
@@ -1819,11 +1819,6 @@
                   {"type": "array", "items": {"type": "string"}}
                 ]
               }
-            },
-            "squashfuse_in_container": {
-              "type": "boolean",
-              "default": false,
-              "description": "DEPRECATED: This key is no longer needed and will be removed in a future version of cloud-init."
```
aciba90, line 1826: I guess the moment to drop it is now. Do we have to explicitly say something in the changelog or something else? Or should we keep this for backward compatibility?

holmanb (approved):
Thanks for this cleanup PR. This looks good to me. To your question, I think input from @blackboxsw would be helpful. We are still working out versioning / publishing of the schema.



- [x] no further references of `squashfuse_in_container`
- [x] no remaining unnecessary references to "squash" or "fuse"
- [x] docs updated in this pr

cloudinit/config/cloud-init-schema.json
blackboxsw: I think we are clear enough here to drop. It was an optional thing to use in containers until about 2018 with snapd version 2.32.3.1. It is unnecessary and anybody using this config option doesn't really need to anymore. The workaround if someone really wanted to install squashfuse in a container would be to provide `packages: [squashfuse]` in #cloud-config  if necessary for their appllication (which it shouldn't be).

blackboxsw (approved):

PR #1449, opened 2022-05-11
testing: mock uses_systemd in ssh password tests (SC-1040)

TheRealFalcon:
## Proposed Commit Message

```
testing: mock uses_systemd in ssh password tests

Tests currently assume systemd, so add a mock such that uses_systemd
always returns True.
```

## Additional Context
Dailies are failing. See
https://launchpadlibrarian.net/600835820/buildlog_ubuntu-jammy-amd64.cloud-init_22.1-3228-g3e5938c6-0ubuntu1+1959~trunk~ubuntu22.04.1_BUILDING.txt.gz

## Test Steps
I haven't actually reproduced the errors on a local machine, but based on the failures I would be surprised if the new mock doesn't work.

PR #1450, opened 2022-05-12
Log generic warning on non-systemd systems.

aciba90:
If the SSH status is checked in a non-systemd system, do not
distinguish between service statuses and log a generic warning
as we cannot expect status exit codes fulfilling any convention.

## Proposed Commit Message

```
Log generic warning on non-systemd systems.

If the SSH status is checked in a non-systemd system, do not
distinguish between service statuses and log a generic warning
as we cannot expect status exit codes fulfilling any convention.
```

## Additional Context

https://github.com/canonical/cloud-init/pull/1422#issuecomment-1124936382

## Checklist:

### DIFF
--- a/tests/unittests/config/test_cc_set_passwords.py
+++ b/tests/unittests/config/test_cc_set_passwords.py
@@ -188,6 +188,105 @@ class TestHandleSSHPwauth:
         assert m_update_ssh_config.call_count == update_ssh_call_count
         m_subp.assert_not_called()
 
+    @mock.patch("cloudinit.distros.Distro.uses_systemd", return_value=False)
+    @mock.patch(MODPATH + "update_ssh_config", return_value=True)
+    @mock.patch("cloudinit.distros.subp.subp")
+    def test_failed_ssh_non_systemd_service_is_not_available_exit_code_25(
+        self, m_subp, m_update_ssh_config, m_uses_systemd
+    ):
+        """If the ssh service is not available in a system without systemd,
+        then no updates config, no restart and a generic message is logged.
+        """
+        cloud = self.tmp_cloud(distro="ubuntu")
+        cloud.distro.init_cmd = ["service"]
+        process_error = "Service is not available."
+        cloud.distro.manage_service = mock.Mock(
+            side_effect=subp.ProcessExecutionError(
+                stderr=process_error, exit_code=25
+            )
+        )
+
+        setpass.handle_ssh_pwauth(True, cloud.distro)
+        self.assertIn(
+            r"WARNING: Ignoring config 'ssh_pwauth: True'."
+            r" SSH service 'ssh' is not available. Error: ",
+            self.logs.getvalue(),
+        )
+        self.assertIn(process_error, self.logs.getvalue())
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


tests/unittests/config/test_cc_set_passwords.py
```diff
@@ -182,6 +182,105 @@ def test_failed_ssh_service_is_not_available(
...
+    @mock.patch(MODPATH + "update_ssh_config", return_value=True)
+    @mock.patch("cloudinit.distros.subp.subp")
+    def test_failed_ssh_non_systemd_service_is_not_available_exit_code_25(
+        self, m_subp, m_update_ssh_config, m_uses_systemd
+    ):
+        """If the ssh service is not available in a system without systemd,
+        then no updates config, no restart and a generic message is logged.
+        """
+        cloud = self.tmp_cloud(distro="ubuntu")
+        cloud.distro.init_cmd = ["service"]
```
blackboxsw, line 195: The PR in 1449,  which only just landed while you were working on this PR, refactored some unittests to use pytest instead of subclassing from CiTestCase. #1449 provided a parameterized TestHandleSSHPwauth test case `test_no_restart_when_service_is_not_running`. Would these new tests make more sense if parameterized in either `TestHandleSSHPwauth.test_no_restart_when_service_is_not_running` or a new parameterized testcase instead of separate unittests?

Two other minor points:

- generally in unittests, I think we want to start avoiding calls to `self.tmp_cloud` and prefer `tests.unittests.util.get_cloud` instead as we try to retired tmp_cloud (and CiTestCase) where possible.
- I don't think we need to set `cloud.distro.init_cmd = ["service"]` on ubuntu distro because default is [`init_cmd 
 = ["service"]`](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/__init__.py#L78) and this only gets reset to "systemctl" if uses_systemd == True, which you've mocked in this test.

blackboxsw (changes requested):

tests/unittests/config/test_cc_set_passwords.py
aciba90: Sorry for breaking the integration tests and thanks for pointing out the updates.

I have parametrized the tests.

aciba90:
@TheRealFalcon this PR is ready now. Thanks.

TheRealFalcon (approved):
LGTM, thanks for your flexibility on this!

blackboxsw (approved):
LGTM

dermotbradley:
@blackboxsw @TheRealFalcon. This change now means I get the following on the console for Alpine Linux systems (openrc/busybox init-based) where previously I did not:

```
2022-05-15 19:06:05,300 - cc_set_passwords.py[WARNING}: Ignoring config 'ssh_pwauth: False'. SSH service 'sshd' is not available. Error: unexpected error while running command.
Command: ['rc-service', 'sshd', 'status']
Exit code: 3
Reason: -
Stdout:  * status: stopped
Stderr: .
```

This is because the sshd service is started later in the boot sequence.

Also note that I have "ssh_pwauth: False" defined in /etc/cloud/cloud.cfg

PR #1451, opened 2022-05-13
Drop support of *-sk keys in cc_ssh.

aciba90:
- Delete *-sk keys from cloud-init-schema.json under
cc_ssh.{ssh_keys,ssh_genkeytypes}
- Log a warning if some given key is unsupported or unknown.
- Port tests to Pytests, add some types and increase unittest
coverage.

## Proposed Commit Message

```
Drop support of *-sk keys in cc_ssh.

- Delete *-sk keys from cloud-init-schema.json under
cc_ssh.{ssh_keys,ssh_genkeytypes}
- Log a warning if some given key is unsupported or unknown.
- Port tests to Pytests, add some types and increase unittest
coverage.
```

## Additional Context

SC-908

After the discussions in #cloud-init we decided to drop the support
of *-sk key types because:
1) In the autogeneration case, user interaction with the device is needed,
which does not fit with a cloud-context.
2) This type of keys are user-based, not hostkeys.

About the integration tests, without the need to support these key types,
I think is sufficient with the existent tests:
- [test_ssh_auth_key_fingerprints.py](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_ssh_auth_key_fingerprints.py)
- [test_ssh_generate.py](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_ssh_generate.py)
- [test_ssh_keys_provided.py](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_ssh_keys_provided.py)

## Checklist:

### DIFF
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -26,7 +26,7 @@ KEY_NAMES_NO_DSA = [
 
 
 @pytest.fixture(scope="function")
-def _publish_hostkey_test_setup(tmpdir):
+def publish_hostkey_test_setup(tmpdir):
     test_hostkeys = {
         "dsa": ("ssh-dss", "AAAAB3NzaC1kc3MAAACB"),
         "ecdsa": ("ecdsa-sha2-nistp256", "AAAAE2VjZ"),
@@ -62,45 +62,33 @@ class TestHandleSsh:
     """Test cc_ssh handling of ssh config."""
 
     @pytest.mark.parametrize(
-        "args,kwargs",
+        "keys,user,disable_root_opts",
         [
             # For the given user and root.
-            pytest.param([["key1"], "clouduser", False], {}, id="with_user"),
+            pytest.param(["key1"], "clouduser", False, id="with_user"),
             # For root only.
-            pytest.param([["key1"], None, False], {}, id="with_no_user"),
+            pytest.param(["key1"], None, False, id="with_no_user"),
             # For the given user and disable root ssh.
             pytest.param(
-                [
-                    ["key1"],
-                    "clouduser",
-                    True,
-                ],
-                {},
+                ["key1"],
+                "clouduser",
+                True,
                 id="with_user_disable_root",
             ),
             # No user and disable root ssh.
             pytest.param(
-                [
-                    ["key1"],
-                    None,
-                    True,
-                ],
-                {},
+                ["key1"],
+                None,
+                True,
                 id="with_no_user_disable_root",
             ),
         ],
     )
     def test_apply_credentials(
-        self,
-        m_setup_keys,
-        args,
-        kwargs,
+        self, m_setup_keys, keys, user, disable_root_opts
     ):
-        keys, user, *args, disable_root_opts = args
         options = ssh_util.DISABLE_USER_OPTS
-        cc_ssh.apply_credentials(
-            keys, user, *args, disable_root_opts, options, **kwargs
-        )
+        cc_ssh.apply_credentials(keys, user, disable_root_opts, options)
         if not disable_root_opts:
             expected_options = ""
         else:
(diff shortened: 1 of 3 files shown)


tests/unittests/config/test_cc_ssh.py
```diff
@@ -22,80 +25,92 @@
...
+        keys, user, *args, disable_root_opts = args
         options = ssh_util.DISABLE_USER_OPTS
-        cc_ssh.apply_credentials(keys, user, True, options)
-        options = options.replace("$USER", "NONE")
-        options = options.replace("$DISABLE_USER", "root")
-        self.assertEqual(
-            [mock.call(set(keys), "root", options=options)],
-            m_setup_keys.call_args_list,
+        cc_ssh.apply_credentials(
+            keys, user, *args, disable_root_opts, options, **kwargs
```
TheRealFalcon, line 102: I'm a little confused about this call. `apply_credentials` takes 4 arguments: keys, user, disable_root, and disable_root_opts. Why are you passing an *args and **kwargs too?

tests/unittests/config/test_cc_ssh.py
```diff
@@ -22,80 +25,92 @@
...
+            ),
+        ],
+    )
+    def test_apply_credentials(
+        self,
+        m_setup_keys,
+        args,
+        kwargs,
+    ):
+        keys, user, *args, disable_root_opts = args
```
TheRealFalcon, line 99: Why the `*args` here? I think this call should just be `keys, user, disable_root` = args

tests/unittests/config/test_cc_ssh.py
```diff
@@ -22,80 +25,92 @@
...
+@mock.patch(MODPATH + "ssh_util.setup_user_keys")
+class TestHandleSsh:
+    """Test cc_ssh handling of ssh config."""
 
-    def test_apply_credentials_with_no_user_disable_root(self, m_setup_keys):
-        """Apply keys no user and disable root ssh."""
-        keys = ["key1"]
-        user = None
+    @pytest.mark.parametrize(
+        "args,kwargs",
```
TheRealFalcon, line 65: Why `kwargs`? It's an empty dict in all of these definitions, so we shouldn't need it at all.

tests/unittests/config/test_cc_ssh.py
```diff
@@ -144,271 +155,144 @@ def test_dont_allow_public_ssh_keys(
...
+                id="cfg_without_disable_root",
+            ),
+        ],
+    )
     @mock.patch(MODPATH + "glob.glob")
     @mock.patch(MODPATH + "ug_util.normalize_users_groups")
     @mock.patch(MODPATH + "os.path.exists")
-    def test_handle_cfg_without_disable_root(
-        self, m_path_exists, m_nug, m_glob, m_setup_keys
+    def test_handle_default_root(
```
TheRealFalcon, line 190: Nit: I'd rename this to something like `test_cfg_disable_root`

tests/unittests/config/test_cc_ssh.py
```diff
@@ -22,80 +25,92 @@
 ]
 
 
-@mock.patch(MODPATH + "ssh_util.setup_user_keys")
-class TestHandleSsh(CiTestCase):
-    """Test cc_ssh handling of ssh config."""
+@pytest.fixture(scope="function")
+def _publish_hostkey_test_setup(tmpdir):
```
TheRealFalcon, line 29: Nit: I don't think we need to preprend local fixtures with an underscore.

TheRealFalcon (changes requested):
Overall looks good. Really nice test changes, though I have some comments inline.

tests/unittests/config/test_cc_ssh.py
aciba90: Same as in [r872517535](https://github.com/canonical/cloud-init/pull/1451/files/4252a5f6f0ca225afb78da86da7cbaeb4f1c6018#r872517535). Simplified.

aciba90 (commented):
Thank you @TheRealFalcon for pointing out the unneeded flexibility (`args`, `kwargs`) and the rest. I have simplified it. Ready to review after the build passes.

TheRealFalcon (approved):
Thanks!

PR #1453, opened 2022-05-16
cc_set_hostname: do not write "localhost" when no hostname is given

esposem:
## Proposed Commit Message
```
Systemd used to sometimes ignore localhost in /etc/hostnames, and many programs
like cloud-init used this as a workaround to set a default hostname.

From https://github.com/systemd/systemd/commit/d39079fcaa05e23540d2b1f0270fa31c22a7e9f1:

	We would sometimes ignore localhost-style names in /etc/hostname. That is
	brittle. If the user configured some hostname, it's most likely because they
	want to use that as the hostname. If they don't want to use such a hostname,
	they should just not create the config. Everything becomes simples if we just
	use the configured hostname as-is.

	This behaviour seems to have been a workaround for Anaconda installer and other
	tools writing out /etc/hostname with the default of "localhost.localdomain".
	Anaconda PR to stop doing that: rhinstaller/anaconda#3040.
	That might have been useful as a work-around for other programs misbehaving if
	/etc/hostname was not present, but nowadays it's not useful because systemd
	mostly controls the hostname and it is perfectly happy without that file.

	Apart from making things simpler, this allows users to set a hostname like
	"localhost" and have it honoured, if such a whim strikes them.

As also suggested by the Anaconda PR, we need to stop writing default "localhost"
in /etc/hostnames, and let the right service (networking, user) do that if they
need to. Otherwise, "localhost" will permanently stay as hostname and will
prevent other tools like NetworkManager from setting the right one.

RHBZ: 1980403
```

## Checklist:


TheRealFalcon:
@esposem , the goal here is a good one, but I think it might make more sense to change this at another level. Setting this in the set_hostname module as done in this PR would prevent anybody from setting localhost via userdata, and also doesn't take into account changes in the update_hostname module. I this this might be better solved by modifying the [default implementation of the datasource get_hostname](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L701). There we can ensure we don't use localhost if haven't explicitly gotten 'localhost' from userdata or our datasource.

esposem:
@TheRealFalcon Do you have a specific suggestion in mind? The problem of datasource `get_hostname` is that we need to always return an hostname, because that is being used by other modules. We can't just "ignore" the return value if it's localhost.

For example, I see `cc_phone_home` uses `distro.get_hostname()`. 

And inside `get_hostname` implementation, it's `socket.gethostname()` that returns `localhost` by default, so maybe we need to change the order of execution between cloud-init and NetworkManager? Or maybe the module execution phase (not sure in which phase it executes now)?

esposem:
Ok, it should be ready now. I used namedtuple as you suggested, and added the case also for `cc_update_hostname`.
Let me know what do you think about naming :smile:

TheRealFalcon (changes requested):
Sorry, I commented inline earlier and then never finished the review and instead commented. Hopefully you can see them now and it isn't too big of a change.

cloudinit/sources/__init__.py
```diff
@@ -158,6 +158,11 @@ def redact_sensitive_keys(metadata, redact_value=REDACT_SENSITIVE_VALUE):
     ],
 )
 
+DataSourceHostname = namedtuple(
+    "DataSourceHostname",
+    ["hostname", "default_hostname"],
```
TheRealFalcon, line 163: `is_default` instead of `default_hostname`?

cloudinit/util.py
```diff
@@ -1103,16 +1103,30 @@ def dos2unix(contents):
     return contents.replace("\r\n", "\n")
 
 
+HostnameFqdnInfo = namedtuple(
+    "HostnameFqdnInfo",
+    ["hostname", "fqdn", "default_hostname"],
```
TheRealFalcon, line 1108: `is_default` instead of `default_hostname`?

TheRealFalcon (changes requested):
Ok, the last inlines comments don't make as much sense anymore, so I'll delete the nonsensical ones :smile: 

`default_hostname` makes me think it'll be the name of a hostname that is the default, rather than a boolean indicating if it is default or not.

esposem:
@TheRealFalcon  Makes sense, thank you for the review!

TheRealFalcon (approved):
Thanks for the work here, this looks good!

PR #1456, opened 2022-05-16
Improve cc_set_passwords.

aciba90:
- If the system doesn't use systemd and the status of the ssh
service is not 0, then apply the pw_auth config. As distros as
Alpine could start the ssh service later in the boot sequence.

- Change log level from warning to debug when ssh_pwauth is
written but the ssh service is not restarted.

## Proposed Commit Message

```
Improve cc_set_passwords.

- If the system doesn't use systemd and the status of the ssh
service is not 0, then apply the pw_auth config. As distros as
Alpine could start the ssh service later in the boot sequence.

- Change log level from warning to debug when ssh_pwauth is
written but the ssh service is not restarted.
```

## Additional Context

SC-968
https://github.com/canonical/cloud-init/pull/1450#issuecomment-1127138478

## Checklist:

### DIFF
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -93,25 +93,7 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
         distro.manage_service("status", service)
     except subp.ProcessExecutionError as e:
         uses_systemd = distro.uses_systemd()
-        if uses_systemd and e.exit_code == 3:
-            # Service is not running. Write ssh config.
-            LOG.debug(
-                "Writing config 'ssh_pwauth: %s'. SSH service '%s'"
-                " will not be restarted because it is stopped.",
-                pw_auth,
-                service,
-            )
-            restart_ssh = False
-        elif uses_systemd and e.exit_code == 4:
-            # Service status is unknown
-            LOG.warning(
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_set_passwords.py
```diff
@@ -111,6 +111,15 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
                 service,
             )
             return
+        elif not uses_systemd:
```
blackboxsw, line 114: It probably makes sense to shuffle this condition up first so we can drop the `if uses_systemd and ...` in the other conditionals
```python
if not distro.uses_systemd():
    <your new debug log>
elif e.exit_code == 3:
   <existing exit_code 3 behavior>
elif e.exid_code == 4:
   < existing exit_code 4 behavior>
else:
   LOG.warning(Ignoring...)
   return
```

blackboxsw (changes requested):
Let's move this not uses_systemd conditional first so we can avoid compound conditionals for the uses_systemd exit_code 3 & 4 cases.

cloudinit/config/cc_set_passwords.py
aciba90: Simplified. Thanks for the suggestion @blackboxsw.

blackboxsw (approved):
+1 Thanks for this work @aciba90

PR #1459, opened 2022-05-17
testing: make fake cloud-init wait actually wait

TheRealFalcon:
## Proposed Commit Message

```
testing: make fake cloud-init wait actually wait
```

Fixes: https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-impish-lxd_container/21/testReport/junit/tests.integration_tests.cmd/test_status/test_wait_when_no_datasource/

### DIFF
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -13,7 +13,7 @@ def _wait_for_cloud_init(client: IntegrationInstance):
     last_exception = None
     for _ in range(30):
         try:
-            result = client.execute("cloud-init status --long")
+            result = client.execute("cloud-init status")
             if (
                 result
                 and result.ok


tests/integration_tests/cmd/test_status.py
```diff
@@ -14,7 +14,11 @@ def _wait_for_cloud_init(client: IntegrationInstance):
     for _ in range(30):
         try:
             result = client.execute("cloud-init status --long")
-            if result and result.ok:
+            if (
+                result
+                and result.ok
+                and ("running" not in result or "not run" not in result)
```
holmanb, line 20: Is there a reason we need `--long` here?

I don't think we are checking the extra output so to avoid future false positives ("running" or "not run" in the extra details) think we should drop it?

holmanb (approved):
One optional comment/question, otherwise looks good!

PR #1461, opened 2022-05-18
Fix test to pass with older pytest.

aciba90:
The test did not work with:
- pytest==4.6.9
- pytest-mock==3.0.0

## Proposed Commit Message

```
Fix daily focal build.

Fix the test to pass with:
- python3-pytest_4.6.9
- python3-pytest-mock_3.0.0
```

## Additional Context

https://launchpad.net/~cloud-init-dev/+archive/ubuntu/daily/+build/23756305

## Test Steps

Reproduce the daily build in focal.

## Checklist:

### DIFF
--- a/tests/unittests/cmd/devel/test_render.py
+++ b/tests/unittests/cmd/devel/test_render.py
@@ -112,10 +112,7 @@ class TestRender:
         with mock.patch("sys.stderr", new_callable=StringIO):
             with mock.patch("sys.stdout", new_callable=StringIO) as m_stdout:
                 assert render.handle_args("anyname", args) == 0
-        assert "Converted jinja variables\n{" in caplog.text
-        # TODO enable after pytest>=3.4
-        # more info: https://docs.pytest.org/en/stable/how-to/logging.html
-        # assert "Converted jinja variables\n{" in m_stderr.getvalue()
+        assert "Converted jinja variables\n" in caplog.text
         assert "rendering: jinja worked" == m_stdout.getvalue()
 
     @skipUnlessJinja()


TheRealFalcon:
@aciba90 , Thanks for looking into this. The oldest supported Ubuntu distro is Bionic. I haven't checked every dependency yet, but so far the ones I looked at are specified correctly, and are older than the ones you're updating this PR to. If you look at the build log of the latest build here, https://code.launchpad.net/~cloud-init-dev/+recipe/cloud-init-daily-bionic , you can see what dependencies are getting installed.

I am also skeptical this is related to pytest or mock versions, but rather some change in jinja rendering between versions. If you look at the caplog text, the focal version shows:
```
2022-05-18 12:37:41 DEBUG     cloudinit.handlers.jinja_template:jinja_template.py:117 Converted jinja variables
                                                                                      {
                                                                                       "my-var": "jinja worked",
                                                                                       "my_var": "jinja worked"
                                                                                      }
```
whereas my local dev version shows:
```
2022-05-18 12:37:16 DEBUG     cloudinit.handlers.jinja_template:jinja_template.py:118 Converted jinja variables
{
 "my-var": "jinja worked",
 "my_var": "jinja worked"
}
```

I'm curious as to why there's a difference there.

TheRealFalcon:
Looks like I was wrong, because I think there's some difference in the caplog fixtures between pytest versions. I verified that what gets written to the logs is the same between versions. However, if I change the line
`assert "Converted jinja variables\n{" in caplog.text`
to something like
`assert "Converted jinja variables\n{" in caplog.records[-1].getMessage()`

I get consistent results.

aciba90:
> Looks like I was wrong, because I think there's some difference in the caplog fixtures between pytest versions. I verified that what gets written to the logs is the same between versions.

I was preparing a reproducible script to show you that. There have been significant changes in pytest, in the way the configure the log and possibly in the way it is captured: https://docs.pytest.org/en/6.2.x/logging.html

>  However, if I change the line `assert "Converted jinja variables\n{" in caplog.text` to something like `assert "Converted jinja variables\n{" in caplog.records[-1].getMessage()`

Do you think it would be sufficient if if change the test to that? Or should we investigate it more deeply? @TheRealFalcon

aciba90:
@TheRealFalcon fixed. Thank you!

TheRealFalcon (approved):
LGTM

PR #1465, opened 2022-05-19
Use Network-Manager and Netplan as default renderers for RHEL and Fedora

esposem:
```
This is adapted from Neal Gompa's PR:
https://github.com/canonical/cloud-init/pull/1435

The only difference is that we are not modifying renderers.py (thus
modifying the priority of all distros), but just tweaking cloud.cfg to
apply this change to Fedora and RHEL. Other distros can optionally
add themselves afterwards.

	net: Prefer Netplan and NetworkManager renderers by default

	NetworkManager is used by default on a variety of Linux distributions,
	and exists as a cross-distribution network management service.

	Additionally, add information about the NetworkManager renderer to
	the cloud-init documentation.

	Because Netplan can be explicitly used to manage NetworkManager,
	it needs to be preferred before NetworkManager.

	As part of this	change, delete a duplicate test case for network
	configuration renderers.

	This change is a follow-up to #1224, which added the native
	NetworkManager renderer.
	This patch has been deployed on Fedora's cloud-init package throughout
	the development of Fedora Linux 36 to verify that it works.

	This should also make it tremendously easier for Linux distributions
	to use cloud-init because now a standard configuration is supported
	by default.

	Signed-off-by: Neal Gompa <ngompa13@gmail.com>

```

## Test Steps
This change is deployed in Fedora Cloud images for Fedora Linux 36, so you can see it in action there.

Otherwise, building an image with a package that has this change and also NetworkManager installed in the image and booting it should result in NetworkManager being used by default instead of any of the legacy ones.
## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -349,4 +349,7 @@ system_info:
 {% elif variant in ["dragonfly"] %}
    network:
       renderers: ['freebsd']
+{% elif variant in ["rhel", "fedora"] %}
+   network:
+      renderers: ['netplan', 'network-manager', 'networkd', 'sysconfig', 'eni']
 {% endif %}
--- a/doc/rtd/topics/network-config.rst
+++ b/doc/rtd/topics/network-config.rst
@@ -188,6 +188,15 @@ generated configuration into an internal network configuration state. From
 this state `Cloud-init`_ delegates rendering of the configuration to Distro
 supported formats.  The following ``renderers`` are supported in cloud-init:
 
+- **NetworkManager**
+
+`NetworkManager <https://networkmanager.dev>`_ is the standard Linux network
+configuration tool suite. It supports a wide range of networking setups.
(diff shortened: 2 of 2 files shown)


config/cloud.cfg.tmpl
```diff
@@ -349,4 +349,8 @@ system_info:
 {% elif variant in ["dragonfly"] %}
    network:
       renderers: ['freebsd']
+{% elif variant in ["rhel", "fedora"] %}
+   network:
+      renderers: ['network-manager', 'netplan', 'eni', 'sysconfig', 'freebsd', 'netbsd',
```
Conan-Kudo, line 354: For RHEL/Fedora, please use `['netplan', 'network-manager', 'networkd', 'sysconfig', 'eni']`

Conan-Kudo (changes requested):

config/cloud.cfg.tmpl
esposem: ok

PR #1481, opened 2022-05-26
FAQ: How can I make a module run on every boot?

holmanb:
```
doc: Document how to change module frequency
```

Also unify the qemu/multipass/lxd sections under a section header.

Here is a low-res screenshot of the rendering:

mod-freq

### DIFF
--- a/doc/rtd/topics/faq.rst
+++ b/doc/rtd/topics/faq.rst
@@ -182,6 +182,49 @@ Internal reasons:
 - nonstandard configurations that disable timeouts or set extremely high
   values ("never" is used in a loose sense here)
 
+How can I make a module run on every boot?
+==========================================
+Modules have a default frequency that can be overridden. This is done
+by modifying the module list in ``/etc/cloud/cloud.cfg``.
+
+1. Change the module from a string (default) to a list.
+2. Set the first list item to the module name and the second item to the
+   frequency.
+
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
`runcmd` probably isn't the best example here. It'll write the script out every time, but since `scripts-user` only runs once per instance, the script won't actually run every time. Given this example, I would expect `date >> /var/log/bootcmd.log` to run every boot.

holmanb:
@TheRealFalcon Thanks for the review. I've updated the example to use cc_phone_home, which has a [frequency of once-per-instance](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=bootcmd#phone-home).

doc/rtd/topics/faq.rst
```diff
@@ -261,8 +306,8 @@ custom network config.
 .. _Instance Configuration: https://linuxcontainers.org/lxd/docs/master/instances
 .. _Custom Network Configuration: https://linuxcontainers.org/lxd/docs/master/cloud-init
 
-cloud-localds
--------------
+Qemu
```
TheRealFalcon, line 309: Nit: Technically QEMU

TheRealFalcon (approved):
One minor nit inline but +1 either way

PR #1483, opened 2022-05-27
Ephemeral ipv6 (SC-938)

holmanb:
```
net: Implement link-local ephemeral ipv6

Also refactor network context managers into net.ephemeral

Currently EC2 is the only IMDS to make use of this.

IPv6 requires a link local address on interfaces. A
link local address is sufficient for the EC2 IMDS,
so no dhcp6 assignment is required for early boot
IMDS queries.

The kernel assigns this address using RFC 4291 [1]
during link initialization, so all cloud-init needs
to do is ensure that link is up.

This means that even if dhcp4 fails, an ipv6-enabled
instance may still succeed at crawling metadata.

[1] https://datatracker.ietf.org/doc/html/rfc4291#section-2.5.6
```


tests/integration_tests/datasources/test_ec2_ipv6.py
```diff
@@ -41,3 +41,11 @@ def test_dual_stack(client: IntegrationInstance):
     # Block IPv6 requests
     assert client.execute("ip6tables -I OUTPUT -d fd00:ec2::254 -j REJECT").ok
     _test_crawl(client, "http://169.254.169.254")
+
+    # Force NoDHCPLeaseError (by removing dhclient) and assert ipv6 still works
+    # Destructive test goes last
+    client.execute("rm /usr/sbin/dhclient").ok
```
aciba90, line 47:
```suggestion
    assert client.execute("rm /usr/sbin/dhclient").ok
```

I guess we want to make sure that the test precondition was correctly executed here.

aciba90 (approved):
Thanks for the improvements. LGTM.

I left some minor nit comments.

tests/integration_tests/datasources/test_ec2_ipv6.py
```diff
@@ -41,3 +41,11 @@ def test_dual_stack(client: IntegrationInstance):
     # Block IPv6 requests
     assert client.execute("ip6tables -I OUTPUT -d fd00:ec2::254 -j REJECT").ok
     _test_crawl(client, "http://169.254.169.254")
+
+    # Force NoDHCPLeaseError (by removing dhclient) and assert ipv6 still works
+    # Destructive test goes last
+    assert client.execute("rm /usr/sbin/dhclient").ok
+    client.restart()
+    client.wait()
```
TheRealFalcon, line 49: Nit: `wait()` should already be built into `restart()`

cloudinit/sources/DataSourceEc2.py
```diff
@@ -125,12 +127,30 @@ def _get_data(self):
                 LOG.debug("FreeBSD doesn't support running dhclient with -sf")
                 return False
             try:
-                with EphemeralDHCPv4(self.fallback_interface):
+                # ipv6 dualstack might succeed when dhcp4 fails
```
TheRealFalcon, line 130: This whole block feels a bit awkward, and I have a feeling we're going to be doing the same thing for other datasources in the near future. What if we have something like a `EphemeralNetwork` (or named whatever) context manager that takes arguments whether to use v4 and v6. This logic could move in there, and then the call logic could be a lot cleaner.

cloudinit/sources/DataSourceEc2.py
```diff
@@ -125,12 +127,30 @@ def _get_data(self):
...
+                        exception = e
+                        v6_only = " using link-local ipv6"
+
                     self._crawled_metadata = util.log_time(
                         logfunc=LOG.debug,
-                        msg="Crawl of metadata service",
+                        msg="Crawl of metadata service" + v6_only,
                         func=self.crawl_metadata,
                     )
+                    if exception:
```
TheRealFalcon, line 151: I don't think we actually want to raise here. If v6 succeeded but v4 failed, this will still cause the datasource to nope out, and there's a higher level try/except that should catch if both failed.

holmanb: Agreed, thanks. Looks like the exception raise happens after the message is logged, which I think explains why the test passed.

cloudinit/sources/DataSourceEc2.py
```diff
@@ -125,12 +126,15 @@ def _get_data(self):
                 LOG.debug("FreeBSD doesn't support running dhclient with -sf")
                 return False
             try:
-                with EphemeralDHCPv4(self.fallback_interface):
+                with EphemeralIPNetwork(
+                    self.fallback_interface, ipv6=True
```
holmanb, line 130: I'm on the fence whether to make ipv6 default to False (as it does now), which makes `EphemeralIPNetwork` more of a drop-in replacement for `EphemeralDHCPv4()`, or make ipv6 default to True, on the assumption that the new context manager will only be used where dual-stack is expected and is therefore a better default. I'm leaning towards option 2 but no strong preference. Thoughts?

holmanb:
@TheRealFalcon: This turned into more refactor than I thought it would be.

> Overall logic looks good, but I think we should build a separate helper for it.

Thanks for the suggestion, agreed. I went with `EphemeralIPNetwork()` as a wrapper for both, and to avoid a dependency cycle I pulled all the ephemeral network context managers out into `net.ephemeral`. I can't say I love this module name, alternative suggestions are welcome.

cloudinit/sources/DataSourceEc2.py
TheRealFalcon: I lean towards option 1 just so a common thing can be imported everywhere, but same...no strong preference

TheRealFalcon (approved):
Is the naming here confusing, or is it just me?

There's EphemeralIPv4Network and EphemeralIPv6Network, but EphemeralDHCPv4 is really more analogous to EphemeralIPv6Network, whereas EphemeralIPv4Network exists purely as a component of EphemeralIDHCPv4. If I'm not looking at prior art, I probably wouldn't choose the right one to import for a new datasource.

I know this all predates this PR, so this isn't something to change here, but might be something we want to address in the future.

holmanb:
> There's EphemeralIPv4Network and EphemeralIPv6Network, but EphemeralDHCPv4 is really more analogous to EphemeralIPv6Network, whereas EphemeralIPv4Network exists purely as a component of EphemeralIDHCPv4. If I'm not looking at prior art, I probably wouldn't choose the right one to import for a new datasource.

Excellent point. I'll plan a followup to rethink the naming scheme into a more coherent ensemble.

PR #1484, opened 2022-05-28
net-convert: use yaml.dump for debugging python NetworkState obj

blackboxsw:
## Proposed Commit Message

```
net-convert: use yaml.dump for debugging python NetworkState obj
    
When debugging python's NetworkState intance we cannot use
safeyaml.dumps because that leverages the yaml.SafeDumper which
does not allow rendering python objects.

Use yamls.dump instead.

* Additional fix for networkd renderer to create /etc/systemd/network if
  it does not exist

LP: #1975907
```

## Checklist:


cloudinit/net/networkd.py
```diff
@@ -219,15 +218,15 @@ def create_network_file(self, link, conf, nwk_dir):
...
 
     def render_network_state(self, network_state, templates=None, target=None):
-        fp_nwkd = self.network_conf_dir
+        network_dir = self.network_conf_dir
         if target:
-            fp_nwkd = subp.target_path(target) + fp_nwkd
+            network_dir = subp.target_path(target) + network_dir
 
-        util.ensure_dir(os.path.dirname(fp_nwkd))
+        util.ensure_dir(network_dir)
```
aciba90, line 225: Nice discovery here!

aciba90 (approved):
LGTM. Thanks for fixing both bugs and add unit tests.

tests/unittests/cmd/devel/test_net_convert.py
```diff
@@ -0,0 +1,187 @@
...
+            ),
+        ),
+    )
+    def test_convert_output_kind_artifacts(
+        self, output_kind, outfile_content, debug, capsys, tmpdir
+    ):
+        """Assert proper output-kind artifacts are written."""
+        network_data = tmpdir.join("network_data")
+        network_data.write(SAMPLE_NET_V1)
+        distro = "fedora" if output_kind == "sysconfig" else "ubuntu"
```
holmanb, line 161: Nit: Since fedora just switched to network-manager, maybe we should use a slower-moving distro for the sysconfig distro?

holmanb (approved):
I have one optional inline nit, but I don't think it currently changes anything functionally, so feel free to disregard. Lgtm!

PR #1485, opened 2022-05-31
Reuse library code, remove dead code

holmanb:
```
datasources: replace networking functions with stdlib and cloudinit.net code

util: remove unused function
sources/vmw: reuse cloudinit.net code
sources/rbx: replace homegrown IP address code with stdlib
```

These changes should be functionally equivalent with better code re-use.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1203,7 +1203,7 @@ def is_ip_network(address: str) -> bool:
     :return:
         A bool indicating if the string is an IPv4 address or not.
     """
-    return bool(maybe_get_address(ipaddress.ip_network, address))
+    return bool(maybe_get_address(ipaddress.ip_network, address, strict=False))
 
 
 def is_ipv4_network(address: str) -> bool:
@@ -1215,7 +1215,7 @@ def is_ipv4_network(address: str) -> bool:
     :return:
         A bool indicating if the string is an IPv4 address or not.
     """
-    return bool(maybe_get_address(ipaddress.IPv4Network, address))
+    return bool(maybe_get_address(ipaddress.IPv4Network, address, strict=False))
 
 
 def is_ipv6_network(address: str) -> bool:
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1139,6 +1139,13 @@ def has_url_connectivity(url_data: Dict[str, Any]) -> bool:
     return True
 
 
+def maybe_get_address(check_cb: Callable, address: str, **kwargs):
+    try:
+        return check_cb(address, **kwargs)
+    except ValueError:
+        return False
+
+
 def network_validator(check_cb: Callable, address: str, **kwargs) -> bool:
     """Use a function to determine whether address meets criteria.
 
@@ -1151,10 +1158,7 @@ def network_validator(check_cb: Callable, address: str, **kwargs) -> bool:
         A bool indicating if the string passed the test.
 
     """
-    try:
-        return bool(check_cb(address, **kwargs))
-    except ValueError:
-        return False
+    return bool(maybe_get_address(check_cb, address, **kwargs))
 
 
 def is_ip_address(address: str) -> bool:
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1139,15 +1139,9 @@ def has_url_connectivity(url_data: Dict[str, Any]) -> bool:
     return True
 
 
-def maybe_get_address(check_cb: Callable, address: str, **kwargs):
-    try:
-        return check_cb(address, **kwargs)
-    except ValueError:
-        return False
-
-
-def network_validator(check_cb: Callable, address: str, **kwargs) -> bool:
-    """Use a function to determine whether address meets criteria.
+def maybe_get_address(convert_to_address: Callable, address: str, **kwargs):
+    """Use a function to return an address. If conversion throws a ValueError
+    exception return False.
 
     :param check_cb:
         Test function, must return a truthy value
@@ -1155,10 +1149,13 @@ def network_validator(check_cb: Callable, address: str, **kwargs) -> bool:
         The string to test.
 
     :return:
-        A bool indicating if the string passed the test.
+        Address or False
 
     """
-    return bool(maybe_get_address(check_cb, address, **kwargs))
+    try:
+        return convert_to_address(address, **kwargs)
+    except ValueError:
+        return False
 
 
 def is_ip_address(address: str) -> bool:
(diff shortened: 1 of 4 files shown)


cloudinit/sources/DataSourceVMware.py
```diff
@@ -685,20 +685,10 @@ def is_valid_ip_addr(val):
...
-        addr = ipaddress.ip_address(val)
-    except ipaddress.AddressValueError:
-        addr = ipaddress.ip_address(str(val))
-    except Exception:
-        return None
-
-    if addr.is_link_local or addr.is_loopback or addr.is_unspecified:
-        return False
-    return True
+    addr = net.maybe_get_address(ipaddress.ip_address, val)
```
blackboxsw, line 688: I don't think this is ultimately what we want here as this introduces a small cost to all `net.is_ip*` functions which now have to call the `maybe_get_address` function.

Ideally, I'd like to see `is_ip_addr` grow a `allow_unroutable: bool = True` which would only invoke the is_link_local, is_loopback, is_unspecified class methods if `True`. Then we can drop this function altogether and just call `net.is_ip_addr(ip_info["addr"], allow_unroutable=False)` down in VMWare's `get_host_info`.

cloudinit/net/__init__.py
```diff
@@ -1139,6 +1139,13 @@ def has_url_connectivity(url_data: Dict[str, Any]) -> bool:
     return True
 
 
+def maybe_get_address(check_cb: Callable, address: str, **kwargs):
```
blackboxsw, line 1142: I don't think this is quite what we want as we now add another level of indirection to all `is_ip*` functions to now invoke yet another function under the hood. It looks like the reason you are centralizing this function is to provide VMware datasource with one less call to `if net.is_ip_address(val): <then check link_local etc>`.  I think I'd prefer if we can extend `net.is_ip_address` to do this optional checking itself given a new parameter (no_link_local=True, or allow_unroutable=False). See more historic discussion here https://github.com/canonical/cloud-init/pull/953/files#r682164843

blackboxsw (changes requested):
Thanks for carving this item up @holmanb. I think we might need to drop a few things from the python 2.7 support front and potentially have further discussion on the best way to leverage existing net.*ip* functions if possible. Let me know what you think.
Thanks

blackboxsw:
> ```
> Reuse library code, remove dead code
> 
> util: remove python 2.7 functionality
> util: delete unused function
> net: new helper function
> sources/vmw: reuse cloudinit.net code
> sources/rbx: replace homegrown IP address code with stdlib
> ```
> 
> These changes should be functionally equivalent with better code re-use.


@holmanb can we also make the commit message summary a bit more specific (which will probably be possible if we are able to drop some of the content in this branch). The title/summary as it stands won't tell future us in a debian/changelog or SRU release notes anything about what really was changed here, we'd have to dive into the full commit message to grok the intent of the PR.

holmanb:
> @holmanb can we also make the commit message summary a bit more specific (which will probably be possible if we are able to drop some of the content in this branch). The title/summary as it stands won't tell future us in a debian/changelog or SRU release notes anything about what really was changed here, we'd have to dive into the full commit message to grok the intent of the PR.

This strikes me as an example where rebase might work better for preserving history of commits (and associated commit messages). I suppose I could have simulated that by breaking it into separate commits.

:+1: will do

cloudinit/sources/DataSourceVMware.py
holmanb: Do I understand the concern correctly that an additional function call overhead would be too great and that new kwargs+conditional in `is_ip_addr` won't add as much overhead to each call?

Function wrapper vs additional kwargs is not something I've measured, but intuition says this tradeoff smells of performance over-optimization. Maybe I'm wrong?

cloudinit/util.py
```diff
@@ -1701,15 +1701,37 @@ def json_serialize_default(_obj):
         return "Warning: redacted unserializable type {0}".format(type(_obj))
 
 
+def json_preserialize_binary(data):
+    """Preserialize any discovered binary values to avoid json.dumps issues.
+
+    Used only on python 2.7 where default type handling is not honored for
+    failure to encode binary data. LP: #1801364.
+    TODO(Drop this function when py2.7 support is dropped from cloud-init)
```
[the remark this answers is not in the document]
holmanb: Thanks @aciba9! This commit reinstates that code so this PR won't clash with PR 1472.

cloudinit/net/__init__.py
blackboxsw: > The logic was basically: "this looks a lot like a version of is_ip_addr() that returns the address if true, rather than just a boolean". The extra layer of indirection allows us to use the existing library logic in the vmware datasource.

Yeah I think this logic in my mind went the other way: "This logic is almost like is_ip_addr() and only needs modification for VMWare datasource, why impact all other network_validator wrapper functions for something that seems specific to just is_ip_addr plus additional validation constraints".


Honestly, I don't know if other datasources would use specific additional constraints in is_ip_address if we built it into that function.

>  [Option 3]: network_validator() [could] return the address (probably also rename the function to something better describing the new functionality)
I hadn't thought of that option either. 

My comment on this shouldn't be blocking as it feels a bit more like personal preference of mine more than significant performance optimization.

I think either stay the course with what you have currently or go with option 3, which could be just drop the `network_validator` function, migrate the applicable parts of docstr of `network_validator` into  `maybe_get_address`  and perform the bool in the `is_ip_*`.
That at least would cut down on the number of functions we have to look through to see what is happening.

cloudinit/sources/DataSourceVMware.py
blackboxsw: The concern was that we were affecting all `is_ip_*` functions in order to tease out a need for the VMWare datasource that seemed like if could have been confined to just the `is_ip_address` function without any impact to the other utility functions. That said it was a minor concern and shouldn't block progress on this branch

blackboxsw (approved):
Thanks for the discussion @holmanb. I'm good with your support of your original suggestion here as I don't think it's really a significant performance diff either way. I'm withdrawing my Needs fixing as that part of the review was mostly along the need to avoid python2.7 removals which were a subset of another branch.

holmanb:
@blackboxsw Thanks for the review.

> ... or go with option 3, which could be just drop the network_validator function, migrate the applicable parts of docstr of network_validator into maybe_get_address and perform the bool in the is_ip_*.
That at least would cut down on the number of functions we have to look through to see what is happening.

+1 I went with this option. I know you've already approved the change, but this feels like enough LOC changed that I should hold off merging.

blackboxsw (approved):
Updated changes LGTM!

PR #1490, opened 2022-06-03
Adminpass metadata

linitio:
## Proposed Commit Message

```
Add support to use admin_pass key from metadata server to set password instance. For example in rebuild or rescue process
```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -68,6 +68,12 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
         mstr = "%s [%s,ver=%s]" % (root, self.dsmode, self.version)
         return mstr
 
+    def get_admin_pass(self):
+        if self.metadata and "admin_pass" in self.metadata:
+            password = self.metadata["admin_pass"]
+            expire = False
+        return (password, expire)
+
     def wait_for_metadata_service(self):
         urls = self.ds_cfg.get("metadata_urls", [DEF_MD_URL])
         filtered = [x for x in urls if util.is_resolvable_url(x)]
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -172,6 +172,13 @@ def handle(_name, cfg, cloud, log, args):
         password = util.get_cfg_option_str(cfg, "password", None)
 
     expire = True
+
+    # get_admin_pass() returns either:
+    # (None, True)
+    # or
+    # ("password", False)
+    if not password:
+        password, expire = cloud.datasource.get_admin_pass()
     plist = None
 
     if "chpasswd" in cfg:
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -361,6 +361,9 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
         if not attr_defaults:
             self._dirty_cache = False
 
+    def get_admin_pass(self):
+        return (None, True)
+
     def get_data(self) -> bool:
         """Datasources implement _get_data to setup metadata and userdata_raw.
 
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -54,6 +54,7 @@ kallioli
 klausenbusk
 KsenijaS
 landon912
+linitio
 lkundrak
 lucasmoura
 lucendio


holmanb:
Hi @linitio, thanks for contributing!

I think if this gets accepted, minimally we will want tests to cover this use-case.

I have a couple of questions to make sure I understand the context:

Is this `admin_pass` key in use on any clouds currently? I don't think that arbitrary keys are used much for cloud-init configuration, however it looks like this is used in cc_seed_random where Azure and openstack have a `random_seed` key, so there is some precedent for this.

Out of curiosity, is there any reason that the standard module config keys won't work during the rebuild or rescue process?

As written, this would only take effect if the module config in userdata isn't set, so I'm not sure I understand what benefit this adds over the following in your cloud-config:
```yaml
password: "super-secret-you-will-not-guess-this"
```

holmanb:
@linitio Ping! Any updates?

linitio:
Hi @holmanb,

Thanks for your reply. The key admin_pass in metadata or config drive is standard key from a long time ago. This key is set by nova when you rebuild or rescue an instance.

For example if you do this from openstack:

`nova rescue --password <PASSWORD> --image <IMAGE_ID> <INSTANCE_ID>`
or
`nova rescue --image <IMAGE_ID> <INSTANCE_ID>`

This command return this result:

`nova rescue  --image <IMAGE_ID> <INSTANCE_ID>
+-----------+---------------+
| Property     | Value                 |
+-----------+---------------+
| adminPass | Ut8DCueTKF9x |
+-----------+---------------+`

Which is either a randomly generated password or the password given in command line.

I've search in existing code on repository and i haven't found any occurence of admin_pass being used.

thomasgoirand:
Hi there!

This is a missing feature in Cloud Init that has been missing for a long time, and I very much would love to see this patch merged.

Cheers,

Thomas Goirand

holmanb:
> This is a missing feature in Cloud Init that has been missing for a long time, and I very much would love to see this patch merged.

@thomasgoirand: Thanks for the feedback :)

holmanb:
@linitio: Thanks for the info. I think the intended purpose sounds makes sense. I have a couple of questions (my openstack background is limited, please bear with me):

**Platform dependent behavior**
> The key admin_pass in metadata or config drive is standard key from a long time ago. This key is set by nova when you rebuild or rescue an instance.

Good to know. Since this sounds like a platform-specific key I think it would be better to isolate this behavior to the intended datasource so we aren't inserting platform behavior into a config module (see "risk" below). I'm thinking this should be a no-op on other datasources. Would you agree?

**Is a non-default config required to use this?**
The set_passwords module [only runs once-per-instance](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=cc_set_password#set-passwords) by default. Since the rescue command takes an instance ID, I assume that this is an instance that cloud-init has already initialized. In order to re-run this module, one would have to execute `cloud-init clean` or similar on the instance. The following boot will re-run cloud-init.

Is this an acceptable workflow for rescued images? I ask because if an admin doesn't already have a root shell on the system to run `cloud-init clean`, this feature won't be useful. This could be worked around by configuring this [module to always run](https://cloudinit.readthedocs.io/en/latest/topics/faq.html#how-can-i-make-a-module-run-on-every-boot) in images, but it would be worth documenting that a non-default config required for this feature to be useful on nova.

When rescuing images, which cloud-config gets used? The same one that launched it or a minimal "rescue userdata"? I could see arguments for doing it either way, and since a defined password in the userdata will override `admin_pass` I think this needs to be documented, since it "won't always work" (anytime the cloud-config has a set password in it). This question makes me wonder if we might be coupling implementation details of nova and cloud-init together too much.

**Risk:**

Some other cloud (or existing cloud users) might be using the same key for different purposes. It's unlikely that someone adds a `admin_pass` key of their own accord and _doesn't_ want this user to have root access. However, if that happens this change introduces a password to a (possibly privileged) default user. At a minimum this will need to be documented in the release notes. Limiting the scope of this change to the intended datasource will help mitigate some risk.

Please let me know if I misjudged anything.

thomasgoirand:
Hi @holmanb ,

Let me attempt to answer for my colleague.

Platform dependent behavior:

Yes, I do believe this is OpenStack specific.

Is a non-default config required to use this?

What happens is that when a VM is started with the --password option, the password should be set when the VM start (using the admin_pass key in the yaml). This, in Nova, can be only set once, when the VM is created, and never later (it can never be changed). Otherwise, when in rescue mode, the VM starts on the rescue image, which is ephemeral (ie: never, data are written back on the image or anywhere else). This means that, from the cloud-init point of view, fetching the admin_pass value always happen on the first boot, never later. Therefore, I'm not sure what is "cloud-init clean", but from my point of view, it doesn't seem useful or related to this feature we're trying to add.

I'm also not sure about what metadata is used when in rescue, but this is probably only the ssh key and the admin_pass. It's not really relevant to this feature as well anyways.

Risk:

I do agree that it may be preferable to have this feature (ie: reading admin_pass to write the root password) only used when using OpenStack as datasource. However, could you explain how can this be achieve? I don't know how to do it, and I'm not sure Mickael knows. If you explain, we'll come with a new patch to implement this how you suggest.

Thanks a lot for your reactivity,
Cheers,

Thomas Goirand

holmanb (changes requested):
See my previous comment.

Also @blackboxsw, if you have time at some point could I request a second opinion?

linitio:
We have implemented the way you suggested. Your thoughts on this new patch revision ?

cloudinit/sources/DataSourceOpenStack.py
```diff
@@ -68,6 +68,12 @@ def __str__(self):
         mstr = "%s [%s,ver=%s]" % (root, self.dsmode, self.version)
         return mstr
 
+    def get_admin_pass(self):
+        if self.metadata and "admin_pass" in self.metadata:
+            password = self.metadata["admin_pass"]
+            expire = False
+        return (password, expire)
```
holmanb, line 75: `password` and `expire` are undefined variables if the conditional evaluates to false

holmanb (changes requested):
Thanks! I think we also need additional test coverage to ensure behavior works as expected.

thomasgoirand:
Hi. What you pointed at was addressed in this last patch.
We're ok to add tests for this feature, but we have no experience doing so in cloud-init. Would you be able to help?

Cheers,

Thomas Goirand (zigo)

holmanb:
> What you pointed at was addressed in this last patch.

Oops, I guess it's just `expire` that is potentially unbound, not both. My bad. This does still require a small change.

> We're ok to add tests for this feature, but we have no experience doing so in cloud-init. Would you be able to help?

I can share some suggestions, but I don't plan on writing the tests myself.

I would try to encode the expectations we've made here into tests so that future changes to the surrounding code don't accidentally cause behavior change.

1. If both user-data config and admin_pass are provided, user-data default user password is not overridden.
2. If only admin_pass is provided, it is used.

```
tests/unittests/config/test_cc_set_passwords.py
tests/unittests/sources/test_openstack.py
```

Since this PR puts datasource-specific functionality into a config module, I'm not sure which is best to put tests in. It might be easier to put the tests in test_openstack.py that way tests can leverage the httpretty uri helpers for this key. That way you can just add the admin_pass key to the appropriate URL, define a couple of different cloud-config userdata definitions and make assertions to verify that cc_set_passwords.handle() behaves as expected.

If the tests go in test_cc_set_passwords, I think you will have to manually mock up a cloud definition with an openstack datasource, which I think would probably be more work.

holmanb:
> What you pointed at was addressed in this last patch.

Oops, I guess it's just `expire` that is potentially unbound, not both. My bad. This does still require a small change.

> We're ok to add tests for this feature, but we have no experience doing so in cloud-init. Would you be able to help?

I can share some suggestions, but I don't plan on writing the tests myself.

I would try to encode the expectations we've made here into tests so that future changes to the surrounding code don't accidentally cause behavior change.

1. If both user-data config and admin_pass are provided, user-data default user password is not overridden.
2. If only admin_pass is provided, it is used.

```
tests/unittests/config/test_cc_set_passwords.py
tests/unittests/sources/test_openstack.py
```

Since this PR puts datasource-specific functionality into a config module, I'm not sure which is best to put tests in. It might be easier to put the tests in test_openstack.py that way tests can leverage the httpretty uri helpers for this key. That way you can just add the admin_pass key to the appropriate URL, define a couple of different cloud-config userdata definitions and make assertions to verify that cc_set_passwords.handle() behaves as expected.

If the tests go in test_cc_set_passwords, I think you will have to manually mock up a cloud definition with an openstack datasource, which I think would probably be more work.

holmanb:
@linitio and @thomasgoirand Just checking, are you planning on continuing with this PR? There are a couple of outstanding requested changes I left in my previous review that still need to be addressed.

PR #1491, opened 2022-06-03
Update WebHookHandler to run as background thread (SC-456)

TheRealFalcon:
## Proposed Commit Message

```
Update WebHookHandler to run as background thread

When reporting via the WebHookHandler, the events are handled on the
main thread. This slows the execution of cloud-init significantly.
If there are timeouts in POSTing, this can make cloud-init appear to be
stuck.

This commit updates the WebHookHandler to run in the background as to
not block the main thread.

Documentation was also added as this feature wasn't previously
documented outside of examples.

LP: #1910552
```

## Additional Context
https://bugs.launchpad.net/maas/+bug/1910552

I intentionally didn't document the Azure-specific reporter as I'm not sure the current state of it or if they would want it enabled arbitrarily. I plan on discussing it at our next Azure sync.

## Test Steps
I used `docker run -p 55555:55555 darklynx/request-baskets` on my host machine to ensure requests are coming through or canceled as expected. 

Note that I moved two unit test files into a subdirectory for organization. No changes to those two files other than the move, and I added a 3rd file for testing this behavior.

## Checklist:


cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,110 @@
...
+                      "description": "The number of times to retry sending the webhook."
+                    }
+                  }
+                },
+                {
+                  "additionalProperties": false,
+                  "properties": {
+                    "type": {
+                      "type": "string",
+                      "enum": ["hyperv"],
```
aciba90, line 2300: Would not be useful to document this for the user?

aciba90 (commented):
Thanks for this improvement and schema addition. I really like the documentation to help users configuring this. I have left some suggestions to harden the schema addition.

cloudinit/config/schemas/schema-cloud-config-v1.json
TheRealFalcon: I brought it up in our Azure meeting this week. This isn't really intended to be user configurable and gets enabled automatically in `/etc/cloud` on Azure instances. I asked if they wanted this to be more broadly documented at our last sync up and they said they'd get back to me. For now, let's leave it.

cloudinit/reporting/__init__.py
```diff
@@ -32,19 +34,19 @@ def update_configuration(config):
             )
             continue
         handler_config = handler_config.copy()
-        cls = available_handlers.registered_items[handler_config.pop("type")]
+        cls: Type[HandlerType] = available_handlers.registered_items[
+            handler_config.pop("type")
+        ]
         instantiated_handler_registry.unregister_item(handler_name)
-        instance = cls(**handler_config)
+        instance = cls(**handler_config)  # pyright: ignore
```
holmanb, line 41: requires pyright >= 1.1.229

TheRealFalcon: Since it's a comment, it shouldn't require pyright at all, right? We already have a handful of pyright comments other places.

holmanb: No need for change here, this was more a note to self than anything I guess. I just was wondering why pyright wasn't respecting the comment locally and realized an upgrade was required to pick up the change.

holmanb (changes requested):
Overall, I like the way this is looking. I have a couple of small nits, and found some issues in the schema definition. I think we'll probably want some tests for the schema once we get it fixed too.

I wasn't very familiar with the reporting code, but I like the use of polymorphism with the `publish()`/`flush()` interface used.

Note: I have more to review, but I have enough text stored in ram/cache that I'd rather save my work and give you something to get started chewing on for this round of review.

cloudinit/reporting/handlers.py
```diff
@@ -81,34 +82,79 @@ def __init__(
...
+        self.event_processor.start()
+
+    def process_requests(self):
+        consecutive_failed = 0
+        while True:
+            if self.flush_requested and consecutive_failed > 2:
+                # Flushing generally happens when cloud-init is ready to
+                # exit, so if we have a queue of events piled up and recent
+                # events have failed, lets not waste time trying to post
+                # the rest, especially since a long timeout could block
```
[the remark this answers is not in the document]
holmanb: +1 Thanks for the clarification.

TheRealFalcon:
@holmanb , I believe I have addressed all of your existing comments.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,112 @@
...
+                    "type": {
+                      "type": "string",
+                      "enum": ["webhook"],
+                      "required": true
+                    },
+                    "endpoint": {
+                      "type": "string",
+                      "format": "uri",
+                      "description": "The URL to send the event to.",
+                      "required": true
```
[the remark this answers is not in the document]
holmanb: Ah, I think I figured it out. The `required` key [must contain at least one string](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties) in Draft4.

I think the change requires something this at the object level:
```yaml
                   "required": ["type", "endpoint"],
```

holmanb (changes requested):
I previously made a comment about the schema failing to validate schemas with missing required keys. I made an inline comment about the cause, here's what I did to get schema validation to warn about required properties.

```diff
diff --git a/cloudinit/config/schemas/schema-cloud-config-v1.json b/cloudinit/config/schemas/schema-cloud-config-v1.json
index 7f2971d2..e8e4232d 100644
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -2220,6 +2220,7 @@
       "properties": {
         "reporting": {
           "type": "object",
+          "additionalProperties": false,
           "patternProperties": {
             "^.+$": {
               "label": "<arbitrary_name>",
@@ -2254,6 +2255,7 @@
                 },
                 {
                   "additionalProperties": false,
+                  "required": ["type", "endpoint"],
                   "properties": {
                     "type": {
                       "type": "string",
```

After applying the above changes, schema validation fails (as expected), albeit not the most intuitive failure message:
```
arc~/cloud-init(reporting|✚1…) PYTHONPATH=. python3 -m cloudinit.cmd.main schema -c 1.yml
Valid cloud-config: 1.yml
arc~/cloud-init(reporting|✚1…) PYTHONPATH=. python3 -m cloudinit.cmd.main schema -c 2.yml
Error:
Cloud config schema errors: reporting.reporter_1: {'consumer_key': '42'} is not valid under any of the given schemas
arc~/cloud-init(reporting|✚1…) cat 2.yml                                                 
#cloud-config
reporting:
  reporter_1:
    consumer_key: "42"
```

In addition to that diff I think we'll want to clean up the `required: true/false` statements which don't appear to do anything.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,112 @@
           "additionalProperties": false
         }
       }
+    },
+    "reporting_config": {
+      "type": "object",
+      "properties": {
+        "reporting": {
+          "type": "object",
```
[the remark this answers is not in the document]
TheRealFalcon: I swear I tried this (and a hundred other combinations) the last time and it failed miserably, but now it's magically working for me.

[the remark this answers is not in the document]
holmanb: I spent more time than I want to admit working on this. You're not alone :)

We have a couple of other `properties` and `patternProperties` in our schema that are missing an `additionalProperties` (where I believe it should be set to false). I have a small local branch with these fixes. I'll wait until this PR lands before I propose it so you don't have to deal with a merge conflict.

TheRealFalcon:
Thanks for all the help here @holmanb . I updated the schema based on your comments and also pushed new schema unit tests as a new commit. The later force push is just a rebase and didn't change anything else.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,112 @@
...
+          "patternProperties": {
+            "^.+$": {
+              "label": "<arbitrary_name>",
+              "type": "object",
+              "oneOf": [
```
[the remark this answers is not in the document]
TheRealFalcon: Do you have an example of what you're concerned about? If the same keys are present, userdata will override anything defined before it, so I don't think there's a problem here, but I may not be understanding the problem.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,110 @@
...
+                      "type": "string",
+                      "enum": ["DEBUG", "INFO", "WARN", "ERROR", "FATAL"],
+                      "default": "DEBUG"
+                    }
+                  }
+                },
+                {
+                  "additionalProperties": false,
+                  "properties": {
+                    "required": ["type"],
```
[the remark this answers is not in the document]
holmanb: I think the "type" property is implicitly checked here without extra effort, but gives a slightly less valuable error message. The reason is that there is only one property key and absence of any properties in a config will be evaluated as nonetype/null. I expect will that current validation causes a warning like `is not of type 'object'`, rather than the marginally better user message: `not valid`.

I'm only adding this detail because this may require updating some unit tests match statements as well as the change here.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,110 @@
...
+                      "type": "integer",
+                      "minimum": 0,
+                      "description": "The number of times to retry sending the webhook."
+                    }
+                  }
+                },
+                {
+                  "additionalProperties": false,
+                  "properties": {
+                    "required": ["type"],
```
holmanb, line 2298: no effect, failing test:
```diff
diff --git a/tests/unittests/reporting/test_reporting.py b/tests/unittests/reporting/test_reporting.py
index 68fc52ad..ec8d872d 100644
--- a/tests/unittests/reporting/test_reporting.py
+++ b/tests/unittests/reporting/test_reporting.py
@@ -554,6 +554,7 @@ class TestReportingSchema:
                 "not valid",
             ),
             ({"reporting": {"a": {"type": "hyperv", "a": "b"}}}, "not valid"),
+            ({"reporting": {"a": {"kvp_file_path": "/path/to/glory/and/honor"}}}, "not valid"),
             # BAD: missing required properties
             ({"reporting": {"a": {"type": "webhook"}}}, "not valid"),
         ],
```

holmanb (changes requested):
A couple more minor inline issues.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2214,6 +2214,110 @@
           "additionalProperties": false
         }
       }
+    },
+    "reporting_config": {
+      "type": "object",
+      "properties": {
+        "reporting": {
+          "type": "object",
+          "additionalProperties": false,
```
holmanb, line 2223: Stylistic comment (no changes requested): I like the way this is structured. Our schema has a lot of individual statements tacked on to the end of long {} contexts, which I personally I find slightly harder to visually parse:
```

        }
      },
      "additionalProperties": true
    }
  ]
}
```

vs:
```json
       "additionalProperties": true,
       "properties": {

```
Going forward I would prefer to the second style where possible.

holmanb:
I just close/reopened to force a re-run of tests, but I think this will need a rebase to pick up the setuptools pin added last week.

cloudinit/reporting/handlers.py
```diff
@@ -81,34 +83,79 @@ def __init__(
...
+                # cloud-init for quite a long time.
+                LOG.warning(
+                    "Multiple consecutive failures in WebHookHandler. "
+                    "Cancelling all queued events."
+                )
+                while not self.queue.empty():
+                    self.queue.get_nowait()
+                    self.queue.task_done()
+                consecutive_failed = 0
+            args = self.queue.get(block=True)
```
holmanb, line 123: Won't this blocking `queue.get()` cause a deadlock when `.join()` is hit? I tried to put together a simple example to demonstrate:

```python3
from threading import Thread, Event
from queue import Queue

e = Event()
q = Queue()

def getter():
    print("entering getter")
    while True:
        print("before blocking get")
        val = q.get(block=True)
        print(f"after blocking get, received {val}")


Thread(target=getter).start()

e.set()
q.put("one")
q.join()
e.clear()
```
example:
```bash
arc~ python3 queue-model.py 
entering getter
before blocking get
after blocking get, received one
before blocking get
```

One way around this (I'm sure there are many) would be to put a sentinal value in the queue that, when received, causes the thread to exit.

It's entirely possible I oversimplified my example too much and missed something. Let me know.

TheRealFalcon: Yeah, `Queue.join()` is a little weird. According to the docs:
> If a [join()](https://docs.python.org/3/library/queue.html#queue.Queue.join) is currently blocking, it will resume when all items have been processed (meaning that a [task_done()](https://docs.python.org/3/library/queue.html#queue.Queue.task_done) call was received for every item that had been [put()](https://docs.python.org/3/library/queue.html#queue.Queue.put) into the queue).

So if all the items in the queue have been processed, join() should unblock. I think I have that covered in the `self.queue.task_done()` calls on lines 141 and 121.

To update your example, I added a `.task_done()` after the print in the while loop and made the thread a daemon, which I think more accurately reflects the code. With these changes, I don't see any deadlock.

```python
from queue import Queue
from threading import Event, Thread

e = Event()
q = Queue()


def getter():
    print("entering getter")
    while True:
        print("before blocking get")
        val = q.get(block=True)
        print(f"after blocking get, received {val}")
        q.task_done()


t = Thread(target=getter)
t.daemon = True
t.start()

e.set()
q.put("one")
print("before join")
q.join()
print("after join")
e.clear()
```

I welcome the deeper look though, and let me know if there are any cases I missed.

holmanb: > To update your example, I added a `.task_done()` after the print in the while loop and made the thread a daemon, which I think more accurately reflects the code. With these changes, I don't see any deadlock.

Ah yes, you're right, a manual exit from the thread is unnecessary. The `daemon=True` setting is what I missed in my example. (I had `.task_done()` locally, but accidentally deleted that while cleaning up to share.)

holmanb (approved):
Looks good to me!

PR #1493, opened 2022-06-06
3.11

holmanb:
```
ci: add python 3.11 (currently unreleased)
```

I pushed this PR with the intention of grabbing failure logs for a ticket tracking future changes in upstream python, but since this passed maybe we just merge it? I am ambivalent about it - if any objections feel free to close.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,6 +150,7 @@ matrix:
             - make check_spelling && tox
         # Test all supported Python versions (but at the end, so we schedule
         # longer-running jobs first)
+        - python: 3.11-dev
         - python: "3.10"
         - python: 3.9
         - python: 3.8


aciba90 (approved):
Thanks! If the resources are not a problem, I think it is a good idea to test against the python-dev version and catch errors in advance.

This is related to [lp #1874866](https://bugs.launchpad.net/cloud-init/+bug/1874866), which looks like outdated, but we could close it with this change, as we are expending our text matrix to all supported python versions + the development one.

holmanb:
> This is related to [lp #1874866](https://bugs.launchpad.net/cloud-init/+bug/1874866), which looks like outdated, but we could close it with this change, as we are expending our text matrix to all supported python versions + the development one.

Didn't realize that one existed, thanks for pointing that out. I agree - close with this change or if this doesn't get merged then we just close it anyways because it's outdated and has already been done.

TheRealFalcon (approved):
LGTM...and shorten the time period I'll need to add more money to the Travis account by that much more :wink:

PR #1495, opened 2022-06-07
Add user/group '-' deprecation warning (SC-1095)

TheRealFalcon:
## Proposed Commit Message

```
Add user/group '-' deprecation warning

Any user/group userdata key that has an '_' can also be used
with '-' instead. Add a deprecation warning if '-' is specified.

LP: #1977818
```

## Additional Context
See https://superuser.com/questions/1725127/invalid-config-for-cloud-init-but-apparently-still-works-how-do-i-remove-the/

Using the userdata in that post, I now see this in the logs:
```
2022-06-07 12:54:44,954 - ug_util.py[WARNING]: User/Group key 'ssh-authorized-keys' is deprecated and will be removed in a future version of cloud-init. Use 'ssh_authorized_keys' instead.
```

### DIFF
--- a/cloudinit/distros/ug_util.py
+++ b/cloudinit/distros/ug_util.py
@@ -110,9 +110,17 @@ def _normalize_users(u_cfg, def_user_cfg=None):
         for uname, uconfig in users.items():
             c_uconfig = {}
             for k, v in uconfig.items():
-                k = k.replace("-", "_").strip()
-                if k:
-                    c_uconfig[k] = v
+                new_k = k.replace("-", "_").strip()
+                if k != new_k:
+                    LOG.warning(
+                        "User/Group key '%s' is deprecated and will be "
+                        "removed in a future version of cloud-init. Use '%s' "
+                        "instead.",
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
[normalize_users_groups](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/ug_util.py#L164) gets called by a few different modules which is why we wind up getting the same warning multiple times. The results should be idempotent, so I tried caching the return values. This still leaves us with 2 identical warnings in the logs because they are called at different boot stages.  Trying to persist the results over multiple boot stages feels like overkill, so it seems best to either live with multiple warnings or pre-check our keys before calling `normalize_users_groups`.

holmanb (approved):
LGTM! Thanks @TheRealFalcon!

TheRealFalcon:
Since we have plans to issue deprecation warnings from the schema rather than issuing individual warnings across the modules, I'm going to close this for now. I'll keep the branch in case we decide we want to revive it.

PR #1496, opened 2022-06-07
Only run github actions on pull request (SC-1094)

TheRealFalcon:
Is there a reason we want it enabled for push and not just for PRs?


holmanb (approved):
No good reason +1

In my mind this was to let users see test results prior to submitting a PR, but since this is causing friction I don't see anyways to keep this around.

Thanks for this!

PR #1498, opened 2022-06-07
improve runcmd docs

holmanb:
```bash
doc: Add runcmd str vs. list arg behavior to module docs
```

Per discussion [here](https://bugs.launchpad.net/cloud-init/+bug/1977780) I think we should document this better.

Note: The comment about rc.local isn't as relevant as it used to be, but I don't think it should be removed at this time.

The rendering:

runcmd

### DIFF
--- a/cloudinit/config/cc_runcmd.py
+++ b/cloudinit/config/cc_runcmd.py
@@ -33,10 +33,6 @@ how it is executed:
 * If the item is a list, the items will be executed as if passed to execve(3)
   (with the first arg as the command).
 
-
-Each item is written to ``/var/lib/cloud/instance/runcmd`` to be later
-interpreted using ``sh``.
-
 Note that the ``runcmd`` module only writes the script to be run
 later. The module that actually runs the script is ``scripts-user``
 in the :ref:`topics/boot:Final` boot stage.
--- a/cloudinit/config/cc_runcmd.py
+++ b/cloudinit/config/cc_runcmd.py
@@ -24,7 +24,7 @@ from cloudinit.settings import PER_INSTANCE
 
 
 MODULE_DESCRIPTION = """\
-Run arbitrary commands at a rc.local like level with output to the
+Run arbitrary commands at a rc.local like time-frame with output to the
 console. Each item can be either a list or a string. The item type affects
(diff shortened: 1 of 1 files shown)


cloudinit/config/cc_runcmd.py
```diff
@@ -25,10 +25,17 @@
...
+how it is executed:
+
+
+* If the item is a string, it will be interpreted by ``sh``.
+* If the item is a list, the items will be executed as if passed to execve(3)
+  (with the first arg as the command).
+
+
+Each item is written to ``/var/lib/cloud/instance/runcmd`` to be later
+interpreted using ``sh``.
```
smoser, line 38: personally, I kind of view that as implementation detail.  And actually a bad implementation detail.

One reason that it is a bad implementation, and one you don't want to advertise to a user, is that it means you can do something like this:

```yaml
runcmd:
 - "foo=bar"
 -  "echo \"foo is $foo\""
```

i don't remember if the above actually would echo 'foo is bar' or not.  But I do recall there was a juju related issue based on the fact that cloud-init was rendering this all to a shell script and executing it.

It would be better if cloud-init handled the execution of these items in the 'runcmd' without  the state that 'sh' usage is providing. 

This is one of those things that you look back at 10 years later and think "why did I do that".

cloudinit/config/cc_runcmd.py
```diff
@@ -25,10 +25,17 @@
 
 MODULE_DESCRIPTION = """\
 Run arbitrary commands at a rc.local like level with output to the
```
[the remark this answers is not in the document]
holmanb: Works for me, will do. :)

cloudinit/config/cc_runcmd.py
holmanb: Agreed, I'll drop that.

holmanb:
@smoser Thanks for the review. I've updated the PR to include your suggestions.

> This is one of those things that you look back at 10 years later and think "why did I do that".

It happens :) Thanks for sharing the additional context.

TheRealFalcon (approved):
LGTM!

PR #1500, opened 2022-06-08
Fix cc_phone_home requiring 'tries' (SC-1101)

TheRealFalcon:
## Proposed Commit Message

```
Fix cc_phone_home requiring 'tries'

As part of dee8231, the code was updated to use a ValueError rather than
catching the top level Exception. This should have also included
catching TypeError. Also, calculate 'retries' correctly.

Additionally, add unit tests for module as it effectively had none.

LP: #1977952
```

I officially apologize for all of the younger-me rants about checked exceptions in other languages.

### DIFF
--- a/cloudinit/config/cc_phone_home.py
+++ b/cloudinit/config/cc_phone_home.py
@@ -130,7 +130,7 @@ def handle(name, cfg, cloud, log, args):
     tries = ph_cfg.get("tries")
     try:
         tries = int(tries)  # pyright: ignore
-    except ValueError:
+    except (ValueError, TypeError):
         tries = 10
         util.logexc(
             log,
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_phone_home.py
```diff
@@ -190,7 +191,7 @@ def handle(name, cfg, cloud, log, args):
         url_helper.read_file_or_url(
             url,
             data=real_submit_keys,
-            retries=tries,
+            retries=tries - 1,
```
blackboxsw, line 194: +1 on the logic correction here. While this does represent change in behavior, I'd defend that original logic was a bug. Generally this retry logic will only trigger in significant repeated error conditions.

cloudinit/config/cc_phone_home.py
```diff
@@ -130,7 +130,7 @@ def handle(name, cfg, cloud, log, args):
     tries = ph_cfg.get("tries")
     try:
         tries = int(tries)  # pyright: ignore
-    except ValueError:
+    except (ValueError, TypeError):
```
blackboxsw, line 133: works for me. This is a better suggestion than the alternative of `tries = ph_cfg.get("tries", DEFAULT_RETRIES)`, as someone in the off chance provided `tries: null` in user-data, we'd still fail with the TypeError exception

PR #1501, opened 2022-06-09
Rename find_modules() to avoid confusion

holmanb:
```
Rename find_modules() to avoid confusion

Currently find_modules() and find_module() are imported
from cloudinit.util and cloudinit.importer, respectively.
Rename find_module() to something more self-descriptive.
```

Since both functions are intended to be used outside the modules they are defined in, and in some parts of the code they are called right next to each other, disambiguation should improve readability.


TheRealFalcon (approved):
Looks good, thanks!

PR #1503, opened 2022-06-09
Ubuntu/bionic 22.2 hotfix

TheRealFalcon:


debian/patches/cpick-a2e62738-Fix-cc_phone_home-requiring-tries-1500
```diff
@@ -0,0 +1,158 @@
...
+
+LP: #1977952
+---
+ cloudinit/config/cc_phone_home.py            | 13 +--
+ tests/unittests/config/test_cc_phone_home.py | 87 ++++++++++++++++++++
+ 2 files changed, 94 insertions(+), 6 deletions(-)
+
+Index: cloud-init/cloudinit/config/cc_phone_home.py
+===================================================================
+--- cloud-init.orig/cloudinit/config/cc_phone_home.py	2022-06-09 16:18:04.443867768 -0500
```
blackboxsw, line 20: Generally, we don't want your local timestamps in the quilt patches as that'll introduce diffs depending on who creates these quilt patches for the cherry-picks. Basically what's missing in your environment is a ~/.quiltrc that defines

`QUILT_REFRESH_ARGS="--no-timestamps --backup"`

You can obtain the basic default quiltrc via something like 
```bash
 cp /etc/quilt.quiltrc  ~/.quiltrc   # ensure QUILT_REFRESH_ARGS="--no-timestamps --backup"
 quilt push -a
 quilt refresh # which removes the trailing timestamps  in your cpick file
quilt pop -a
git commit -am 'quilt refresh patch without timestamps'
git rebase -i HEAD~3 # and squash the quilt patch changes together dropping the timestamps.
```

Otherwise this is +1.

TheRealFalcon: I noticed you pushed the change requested here. If I run through this procedure, I see no changes compared to what you pushed here. Is there anything else we need to do here?

PR #1515, opened 2022-06-13
Sketch solution to skip inapplicable modules

aciba90:
## Proposed Commit Message

```
Skip inapplicable modules

TL:DR: Save time loading and calling handle function in 
56 modules when no applicable user-data config is
provided.

Pre-filter any inapplicable config modules from boot if both
following items hold:
- `skippable` meta property is True
- Check if there is user-data config intersecting module's
  schema defs 
```

## Additional Context

[SC-1065](https://warthogs.atlassian.net/browse/SC-1065)

## Test Steps

```sh
$ lxc launch ubuntu-daily:bionic dev-b
$ lxc exec dev-b -- cloud-init status --wait
$ lxc exec dev-b -- cat /var/log/cloud-init.log | grep applicable
2022-06-13 13:24:43,487 - modules.py[INFO]: Skipping modules 'snap,ssh-import-id,keyboard,ubuntu-advantage' because no applicable user-data config is provided
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ssh_import_id.py
+++ b/cloudinit/config/cc_ssh_import_id.py
@@ -43,6 +43,7 @@ meta: MetaSchema = {
             """
         )
     ],
+    "skippable": True,
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/modules.py
+++ b/cloudinit/config/modules.py
@@ -8,11 +8,12 @@
 
 import copy
 from collections import namedtuple
-from typing import List
+from typing import List, Optional
 
 from cloudinit import config, importer
 from cloudinit import log as logging
 from cloudinit import type_utils, util
+from cloudinit.config.schema import get_config_keys, get_schema
 from cloudinit.distros import ALL_DISTROS
 from cloudinit.helpers import ConfigMerger
 from cloudinit.reporting.events import ReportEventStack
@@ -63,6 +64,18 @@ def validate_module(mod, name):
             f"Module '{mod}' with name '{name}' has a JSON 'schema' attribute "
             "defined. Please define schema in cloud-init-schema,json."
         )
+
+
+def _is_applicable(
+    module_details: ModuleDetails, cfg: dict, schema: Optional[dict] = None
+) -> bool:
+    if not module_details.module.meta.get("skippable", False):
+        return True
+    normalized_name = f"cc_{module_details.name}".replace("-", "_")
+    config_keys = get_config_keys(normalized_name, schema=schema)
+    if not config_keys.intersection(cfg.keys()):
+        return False
+    return True
 
 
 class Modules(object):
@@ -155,7 +168,7 @@ class Modules(object):
     def _fixup_modules(self, raw_mods) -> List[ModuleDetails]:
         """Convert list of returned from _read_modules() into new format.
 
-        Invalid modules and arguments are ingnored.
+        Invalid modules and arguments are ignored.
         Also ensures that the module has the required meta fields.
         """
         mostly_mods = []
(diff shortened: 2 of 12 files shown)


TheRealFalcon:
I think in general this is a good approach, but I think we're going to wind up with a few thorny exceptions. There's some configuration shared between multiple modules but only defined under the schema of one (e.g., hostname is used by both cc_set_hostname and cc_update_hostname) and I think multiple modules might operate on ssh data and user/group data. In the short term, I think it's fine to define these modules as skippable and worry about how to skip them later.

Additionally, I think we need a fair amount of testing of this feature to ensure we're not accidentally skipping modules that we shouldn't be. It's probably better to use this PR as the infrastructure (and maybe one or two modules as an example), and then add groups of modules in follow-up PRs as we go.

@blackboxsw , wanna take a look too? I think this was initially your idea, so do you have any comments on what's being proposed here?

holmanb:
Another exception to consider: a couple of config modules pull metadata directly from the IMDS, not just the user-data. Here's a [related comment I made about this](https://github.com/canonical/cloud-init/pull/1490#issuecomment-1146103780) in an open pull request that is proposing such an addition to cc_set_password.

cloudinit/config/modules.py
```diff
@@ -265,16 +278,21 @@ def run_section(self, section_name):
...
+        inapplicable_mods = []
         active_mods = []
-        for (mod, name, _freq, _args) in mostly_mods:
+        for module_details in mostly_mods:
+            (mod, name, _freq, _args) = module_details
             if mod is None:
                 continue
             worked_distros = mod.meta["distros"]
-
+            if not _is_applicable(module_details, self.cfg, schema=schema):
```
blackboxsw, line 293: I think we want to also consider that this new behavior may not be something we turn on by default on stable series Bionic/Focal/Impish/Jammy. I'd like us to think about a configuration option maybe under a `modules` top-level key that can be provided minimally in /etc/cloud/cloud.cfg(.d/*). Ideally, we might also want to surface that config-key in user-data too so folks could opt to `skip_modules_by_applicable_schema: true` or something during a specific instance launch. By default that config setting would be False on stable series because we don't want to change behavior.

cloudinit/config/cc_ssh_import_id.py
```diff
@@ -43,6 +43,7 @@
             """
         )
     ],
+    "skippable": True,
```
blackboxsw, line 46: I wonder if we want the keys that a module expresses to be static in either schema or meta-schema so we don't have to calculate much per module during runtime.  `skip_without_config_keys` could be a list in meta schema. This could allow us to opinionate specific subsets of keys that trigger some modules but not others even though multple modules may support. A non-empty `skip_without_config_keys` value would indicate that a module is skippable.

cloudinit/config/modules.py
blackboxsw: In a team standup conversation on this, the intent of this PR is that it is a performance improvement that consumers of stable releases will want turned on by default anyway.  We agree it should be writen in such a way as to not change behavior from the consumers point of view. As such, we'll ensure this is covered adequately by unit and integration tests to ensure no regressions are there. There is no need for us to provide a user-data or file-system top-level configuration key to turn this feature on or off because it should just work. If it does not, it it a bug that needs to be addressed in the impementation.

aciba90:
> There's some configuration shared between multiple modules but only defined under the schema of one (e.g., hostname is used by both cc_set_hostname and cc_update_hostname) and I think multiple modules might operate on ssh data and user/group data.

I think this suggestion solves the previous problem.

> Additionally, I think we need a fair amount of testing of this feature to ensure we're not accidentally skipping modules that we shouldn't be. It's probably better to use this PR as the infrastructure (and maybe one or two modules as an example), and then add groups of modules in follow-up PRs as we go.

Agreed, better to separate it in different PRs as opposed to a gigantic one.

Thanks, @TheRealFalcon, for the review and comments.

cloudinit/config/cc_ssh_import_id.py
aciba90: > I wonder if we want the keys that a module expresses to be static in either schema or meta-schema so we don't have to calculate much per module during runtime. `skip_without_config_keys` could be a list in meta schema. This could allow us to describe specific subsets of keys that trigger some modules but not others as multiple modules may choose to react to keys not defined within their module specific schema. A non-empty `skip_without_config_keys: [a, b, c]` list would indicate that a module is skippable unless key 'a' 'b' or 'c' exists in merged config.

I chose to express it as boolean, only taking into account the naive idea that every module (named cc_**module_name**) reacts only to **module_name** config key. 

I think your suggested way of defining the non-applicability is better, as it would solve [@TheRealFalcon's commented problem](https://github.com/canonical/cloud-init/pull/1515#issuecomment-1155509669) of having multiple modules reacting to the same config keys.

Thanks, @blackboxsw, for the review and comments.

aciba90:
I am closing this PR for the moment. I will reopen it when ready to be merged. Thanks.

PR #1517, opened 2022-06-13
Improve Schema For cc_set_passwords (SC-1105)

holmanb:
```
Update chpasswd schema to deprecate password parsing

Additionally:

- chpasswd is specific to distro, make it a distro method
- add test to verify parity between new and old schema options
- update docs to encourage use of new schema keys
```

I've seen this come up as a concern multiple times from users. I'd also personally like to see this improved.

One remaining item to check before merging: ~~what happens when identical users are in both `list:` and `users:`? Would it be better to catch this with a runtime exception or allow any error to bubble up from `chpasswd`?~~ <- No longer a todo item: tests have been added that verify current behavior is preserved with the current change.


holmanb:
> So we need to keep the random password feature.

Sounds good. No problem, I'll add that to the new key as well.

holmanb:
> FYI, https://bugs.launchpad.net/cloud-init/+bug/1979065 . Not sure if it makes sense to address this here (I actually think it's probably better not to), but since you're in the middle of modifying some of this, you may want to incorporate this too.

I can do that in a follow up PR while this module is fresh in my brain.

holmanb:
I just added some tests to verify that duplicate users in input don't break anything in a new way. This is arguably invalid input, but since we don't currently check for it I just wanted to ensure it didn't lead to brokenness while merging the two supported inputs together.

cloudinit/config/cc_set_passwords.py
```diff
@@ -215,9 +216,9 @@ def handle(_name, cfg: dict, cloud: Cloud, log: Logger, args: list):
                     "cloud-init. Use the list format instead."
                 )
                 log.debug("Handling input for chpasswd as multiline string.")
-                plist = util.get_cfg_option_str(chfg, "list", plist)
-                if plist:
-                    plist = plist.splitlines()
+                multiline = util.get_cfg_option_str(chfg, "list")
```
holmanb, line 219: The change here should be a noop change and was made for typing reasons.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1762,6 +1762,46 @@
...
+                        "enum": ["hash", "text"],
+                        "default": "hash",
+                        "type": "string"
+                      },
+                      "password": {
+                        "type": "string"
+                      }
+                    }
+                  }
+		]
```
aciba90, line 1802: Indentation fix:

```suggestion
                ]
```

aciba90 (approved):
+1. Really nice UX schema addition.

I left you a minor nit. The indentation in that line was made with tabs instead of spaces.

cloudinit/config/cc_set_passwords.py
```diff
@@ -172,11 +194,18 @@ def handle(_name, cfg, cloud, log, args):
...
+    users_list: List = []
 
     if "chpasswd" in cfg:
         chfg = cfg["chpasswd"]
+        users_list = util.get_cfg_option_list(chfg, "users")
         if "list" in chfg and chfg["list"]:
+            log.warning(
+                "DEPRECATION: The list key is "
+                "deprecated and will be removed from a future version of "
+                "cloud-init. Use the users key instead."
```
blackboxsw, line 207: We should standardize on the deprecation format we've used elsewhere in other cc_* modules
```suggestion
                "DEPRECATION: key 'lists' is now deprecated. Use 'users'."
```

tests/unittests/config/test_cc_set_passwords.py
```diff
@@ -376,6 +527,96 @@ class TestSetPasswordsSchema:
...
+                            {
+                                "name": "sonata",
+                                "password": "dit",
+                                "dat": "dot",
+                            }
+                        ]
+                    }
+                },
+                "is not valid under any of the given schemas",
+            ),
```
blackboxsw, line 615: One thing that is a nice-to-have from unittest run perspective is aptly named test ids for these cases which allows us to avoid the comment, and we get to see specfic labelled test ids if they fall.
you get to see womthing like:
```test_cc_set_passwords.py::TestSetPasswordsSchema::test_schema_validation[dat_is_an_additional_property]```
instead of 
```test_cc_set_passwords.py::TestSetPasswordsSchema::test_schema_validation[config8-is not valid under any of the given schemas```

You can to something like this:
```suggestion
            pytest.param(
                {
                    "chpasswd": {
                        "users": [
                            {
                                "name": "sonata",
                                "password": "dit",
                                "dat": "dot",
                            }
                        ]
                    }
                },
                "is not valid under any of the given schemas",
                id="dat_is_an_additional_property"
            )
```

blackboxsw (approved):
@holmanb much delayed review. but looks great. Just a missing deprecated: true I think and minor text changes.
Do what you will with the unittest suggestions.

Non-blocking comments, you can merge once deprecated: true is in schema.

holmanb:
@blackboxsw - Great feedback, thanks! I committed your suggestions verbatim.

> Non-blocking comments, you can merge once deprecated: true is in schema.

It looks like adding the deprecation key causes schemavalidation errors to be raised under strict validation in some tests.

You already gave approval, but given this change wasn't discussed previously I'll wait until you've had a chance to review that part (last commit).

blackboxsw (approved):
LGTM, and yes incorporates @aciba90's deprecation handling now from ccea0d44752b86bd9da951708df048be17561545. - +1

PR #1520, opened 2022-06-15
Add support for OpenMandriva

berolinux:
## Proposed Commit Message

```
Add support for OpenMandriva (https://openmandriva.org/).

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>
```

## Additional Context

Verified to work with an OpenMandriva Cooker image in Oracle Cloud.
Will try a few more clouds and working with extra modules; a few more patches may be necessary to make all modules work, but all the basics are looking good already with this initial patch.

## Test Steps
Install OpenMandriva in a supported cloud and make sure cloud-init created the user, brought up the network, set up ssh keys, resized the filesystem, etc.

## Checklist:


config/cloud.cfg.tmpl
```diff
@@ -125,8 +125,8 @@ cloud_config_modules:
 {% if variant in ["rhel"] %}
  - rh_subscription
 {% endif %}
-{% if variant in ["rhel", "fedora", "photon"] %}
-{% if variant not in ["photon"] %}
+{% if variant in ["rhel", "fedora", "openmandriva", "photon"] %}
+{% if variant not in ["openmandriva", "photon"] %}
```
Conan-Kudo, line 129: It is possible for an OpenMandriva system to be managed by a Spacewalk/Uyuni deployment, since there's a DNF plugin for it.

Conan-Kudo (changes requested):

TheRealFalcon (changes requested):
Thanks for the contribution, @berolinux ! I have verified that you signed the CLA.

Overall the addition here looks good. In addition to the other comment already listed, can you also add a test to `tests/unittests/test_util.py`? Near all of the other `OS_RELEASE_` definitions, you can add a `OS_RELEASE_OPENMANDRIVA` with a test to accompany it. You should be able to use other tests in the file in you need help.

config/cloud.cfg.tmpl
berolinux: True. I was careful about it because it's untested, but yes, it's likely to work

berolinux: Yes, I was overly cautious because it's untested - but should work. Fixed.

berolinux:
Should be good now (sorry, the "Compare" output is somewhat messed up because I rebased to current master while at it)

TheRealFalcon:
Changes look good but looks like there's some formatting and testing failures. Looks like the test is `tests/unittests/test_cli.py::TestCLI::test_wb_schema_subcommand_doc_all_spot_check` and there's also black failures.

berolinux:
All tests (including black) pass here now.

TheRealFalcon (approved):
LGTM, thanks!

PR #1528, opened 2022-06-17
tox: integration-tests-jenkins: softfail if only some test failed (SC-1117)

paride:
## Proposed Commit Message

```
tox: integration-tests-jenkins: softfail if only some test failed

Prior to this change a single test failure when running the
integration-tests-jenkins tox environment caused Jenkins jobs to fully
fail (status FAILURE).

With this change the pytest's "some tests failed" return code [1] is
ignored, and the status of the Jenkins jobs is determined using the
junit report that pytest generates. Most of the time this will result
in UNSTABLE (yellow) jobs when some tests failed.

If the junit report is missing (or only reports failures) the job
will still be marked as FAILURE (red).

[1] https://docs.pytest.org/en/7.1.x/reference/exit-codes.html
```

## Additional Context

See the internal Jenkins `cloud-init` view.

## Test Steps

1. Sabotage a test and run the `integration-tests-jenkins` on a subset
of tests containing it. Verify that tox logs the failure but returns 0.
2. Pass a test directory that doesn't exist to the environment via posargs.
Verify that tox returns nonzero (should return 4 iirc).

## Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -200,7 +200,10 @@ setenv =
     PYTEST_ADDOPTS="-m ci and not adhoc"
 
 [testenv:integration-tests-jenkins]
-commands = {[testenv:integration-tests]commands}
+# Pytest's RC=1 means "Tests were collected and run but some of the tests failed".
+# Do not fail in this case, but let Jenkins handle it using the junit report.
+allowlist_externals = sh
+commands = sh -c "{envpython} -m pytest --log-cli-level=INFO -vv {posargs:tests/integration_tests/none} || [ $? -eq 1 ]"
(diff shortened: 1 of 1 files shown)


paride:
This should help distinguish tests failure from infrastructure/Jenkint/network/... issues. If you're wondering where we're telling pytest to generate the junit report it's here: https://github.com/canonical/server-jenkins-jobs/blob/4871903be5d8f411f7fe1b2cb0e7b8ff754e69fa/jenkins-jobs/cloud-init/integration.yaml#L167=.

TheRealFalcon (approved):
LGTM!

PR #1527, opened 2022-06-17
Add dcaro to the cla signers

david-caro:
Just signed the cla, so adding myself.

## Proposed Commit Message

```
summary: Add dcaro to the cla signers

```

PR #501, opened 2020-07-23
Updated OpenStack default timeouts

eandersson:
In some rare cases, the metadata service may not be reachable at the very first attempt. Other cloud-init sources set max_wait and the timeout to higher values to help combat this; e.g. the EC2 and CloudStack sources. Lets do the same for OpenStack to make sure that we give it ample time to detect the metadata service.

PR #1529, opened 2022-06-17
DataSourceOpenStack: retry also when first waiting

david-caro:
## Proposed Commit Message
There's a first step to wait and filter out non responsive metadata
services, and the `retries` option is not taken into account there.
This patch adds the retrying logic to that first step too.

```
summary: DataSourceOpenStack: retry also when first waiting

There's a first step to wait and filter out non responsive metadata
services, and the `retries` option is not taken into account there.
This patch adds the retrying logic to that first step too.

LP: #1979049
```

## Checklist:

This depends on #1527 (CLA signing)

### DIFF
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -68,7 +68,7 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
         mstr = "%s [%s,ver=%s]" % (root, self.dsmode, self.version)
         return mstr
 
-    def wait_for_metadata_service(self):
+    def wait_for_metadata_service(self, retries=5):
         urls = self.ds_cfg.get("metadata_urls", [DEF_MD_URL])
         filtered = [x for x in urls if util.is_resolvable_url(x)]
         if set(filtered) != set(urls):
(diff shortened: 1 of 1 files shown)


andrewbogott (approved):
Yes please! This would help with overloaded metadata servers and could also help if we're rotating through a set of servers some of which are unresponsive.

TheRealFalcon:
Depends on #1527

TheRealFalcon:
@david-caro , We should already be retrying until we hit the `max_wait`. See my inline comment. Does that sound right to you?

On some level, I think that if the datasource isn't ready within 10 seconds, that's really an issue that should be fixed on the datasource side, but I understand why longer timeouts can help. Historically, we've been hesitant to increase the default timeouts in openstack because on certain architectures that could lead to longer timeouts for non-openstack datasources. See #501 for context.

That said, those concerns aren't as relevant anymore as most other datasources do get reported correctly during early boot.

Can the issues you're seeing be fixed by [updating the configuration](https://cloudinit.readthedocs.io/en/latest/topics/datasources/openstack.html#configuration) for your particular openstack installation? Configuration can be set in /etc/cloud/cloud.cfg, /etc/cloud/cloud.cfg.d, or as vendordata to all of your instances.

david-caro:
> @david-caro , We should already be retrying until we hit the `max_wait`. See my inline comment. Does that sound right to you?
> 
Hmm, I think I understand the logic now, tell me if I'm right:

So when doing the `wait_for_metadata_service`, the helper function it uses, `url_helper.wait_for_url`, will try the urls passed, potentially one by one, with the timeout given for each connection, and until a maximum elapsed time of `max_wait_seconds`, so in order to make it retry the urls that are passed, expecting each of them to timeout in `timeout_seconds`, should be `max_wait_seconds > timeout_seconds * num_urls`, is that so?

If so, I can try tweaking the configuration yes,

> On some level, I think that if the datasource isn't ready within 10 seconds, that's really an issue that should be fixed on the datasource side, but I understand why longer timeouts can help. Historically, we've been hesitant to increase the default timeouts in openstack because on certain architectures that could lead to longer timeouts for non-openstack datasources. See #501 for context.
> 
I agree with that, but we are trying to minimize the impact until we fix it properly.

> That said, those concerns aren't as relevant anymore as most other datasources do get reported correctly during early boot.
> 
> Can the issues you're seeing be fixed by [updating the configuration](https://cloudinit.readthedocs.io/en/latest/topics/datasources/openstack.html#configuration) for your particular openstack installation? Configuration can be set in /etc/cloud/cloud.cfg, /etc/cloud/cloud.cfg.d, or as vendordata to all of your instances.

TheRealFalcon:
> so in order to make it retry the urls that are passed, expecting each of them to timeout in timeout_seconds, should be max_wait_seconds > timeout_seconds * num_urls, is that so?

Yes, but note that timeout is only relevant if the server is blocking on returning a response. If the request returns with error immediately then the `timeout_seconds` isn't really relevant.

david-caro:
Oh yes, changing the config seemed to work, thanks!

I'll close the pr/issue

PR #1532, opened 2022-06-17
schema: Use additionalProperties by default.

holmanb:
This suggestion/discussion item is that our current idiom:
```json
  "properties": {
    "number": { "type": "number" },
    ...
  },
  "additionalProperties": false
```
can largely be replaced by:
```json
  "additionalProperties": {
    "number": { "type": "number" },
    ...
  },
```

Currently cloud-init uses `additionalProperties` as a boolean, where we usually manually set `additionalProperties: False` to force `properties` and `patternProperties` to match or cause a validation failure. This works, but can be error prone to write and audit, especially with a multi-thousand-line json file. All it takes is one missing `additionalProperties: false` in an object definition to disable validation of all subschemas, hence this suggestion.

Per [the docs](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties):

> The additionalProperties keyword is used to control the handling of extra stuff, that is, properties whose names are not listed in the properties keyword or match any of the regular expressions in the patternProperties keyword. By default any additional properties are allowed.
> The value of the additionalProperties keyword is a schema that will be used to validate any properties in the instance that are not matched by properties or patternProperties. Setting the additionalProperties schema to false means no additional properties will be allowed.

Of course we would want to keep `properties` (with `additionalProperties` set to its default value) in places where we truly wish to allow passthrough of arbitrary data without validation.

I don't believe that this extends to `patternProperties`, so that will still require manually disabling `additionalProperties`.

I'll see if I can grab some examples of where validation is incorrect due to this issue. At a quick glance in my editor I think I see one or more examples of missing `additionalProperties: false` definitions, but I need to test to confirm.

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -19,7 +19,7 @@
           {"required": ["name"]},
           {"required": ["snapuser"]}
       ],
-      "properties": {
+      "additionalProperties": {
         "name": {
           "description": "The user's login name. Required otherwise user creation will be skipped for this user.",
           "type": "string"
@@ -130,8 +130,7 @@
           "description": "The user's ID. Default is next available value.",
           "type": "integer"
         }
-      },
-      "additionalProperties": false
+      }
     },
     "apt_configure.mirror": {
       "type": "array",


holmanb:
I'm still a fan of this idea, but since this is actually behavior defined in the [2019 draft](https://json-schema.org/draft/2019-09/json-schema-core.html#additionalProperties), some packaging work will be required before we can take advantage of this feature. The current python3-jsonschema package on Ubuntu jammy contains Drafts 3-7 (despite the outdated description).

[Upstream](https://python-jsonschema.readthedocs.io/en/stable/validate/?highlight=draft#versioned-validators) is a major version ahead of jammy and additionally has the 2019 and 2020 validators. Maybe we could request a version bump to be able to make use of this feature?

holmanb:
It looks like there is some work on [supporting 4.6.0 in debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1005757#59). There probably won't be anything actionable to do here until the issues get resolved in multiple upstream projects and then accepted into multiple downstream distros. Closing for now.

aciba90:
I like this initiative. Using `additionalProperties` as a schema to validate against not defined props could help us to add more meaning to our schema. Let's track `python3-jsonchema`'s version. Thanks for investigating this!

PR #1537, opened 2022-06-22
Fix LXD datasource crawl when BOOT enabled

TheRealFalcon:
## Proposed Commit Message

```
Fix LXD datasource crawl when BOOT enabled

In 53a995e2, the network config code was updated to initialize
variables to None rather than using "_unset" in order to simplify the
typing. On subsequent boots with the BOOT event enabled, this resulted
in network config being set to "_unset", causing cloud-init to fail
with traceback.
```

## Additional Context
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-jammy-lxd_container/47/testReport/tests.integration_tests.modules/test_user_events/test_boot_event_enabled/

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -13,7 +13,7 @@ import os
 import socket
 import stat
 from json.decoder import JSONDecodeError
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Union, cast
 
 import requests
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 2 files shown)


aciba90 (approved):
LGTM, thanks.

It would be interesting to know what component is checking against `unset`, in order to be able to simplify this in the future. But I have notice that a lot of DataSources use `unset`, I guess it will not be that easy.

holmanb (approved):
Looks good.

PR #1540, opened 2022-06-23
Pin setuptools for Travis (SC-1136)

TheRealFalcon:
```
Pin setuptools for Travis

Pinned to the latest version that works across versions.
See https://github.com/pypa/setuptools/issues/3118
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ install:
     # Required so `git describe` will definitely find a tag; see
     # https://github.com/travis-ci/travis-ci/issues/7422
     - git fetch --unshallow
+    # See https://github.com/pypa/setuptools/issues/3118
+    - pip install setuptools==59.6.0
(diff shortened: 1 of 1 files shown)


paride (approved):
+1, my only comment is that I find https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v5970= a better reference to justify the specific pinned version, as the linked setuptools issue doesn't seem to reference it.

holmanb (approved):
I'm fine with pinning the same for all versions, but perhaps a comment about which tests fail (3.7/3.8) will be appreciated in the future for context? I guess we should be able to just mine this PR on Github for context, so whether or not you want to add a comment, LGTM.

TheRealFalcon:
Added an additional comments about python 3.7 and 3.8. I'll merge this then if tests pass.

PR #1542, opened 2022-06-24
testing: Don't run custom cloud dir test on Bionic (SC-1138)

TheRealFalcon:
## Proposed Commit Message

```
testing: Don't run custom cloud dir test on Bionic

LXD inserts some agent setup code into VMs on Bionic under
/var/lib/cloud. The inserted script will cause this test to fail because
the test ensures nothing is running under /var/lib/cloud. Since LXD is
doing this and not cloud-init, we should just not run the test on Bionic
to avoid the failure.
```

## Additional Context
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-bionic-lxd_vm/42/

### DIFF
--- a/tests/integration_tests/test_paths.py
+++ b/tests/integration_tests/test_paths.py
@@ -50,7 +50,13 @@ class TestHonorCloudDir:
             collect_logs_result.ok
         ), f"collect-logs error: {collect_logs_result.stderr}"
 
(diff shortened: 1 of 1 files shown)


aciba90 (approved):
LGTM!

PR #1547, opened 2022-06-27
schema: Force stricter validation (SC-1166)

holmanb:
```
schema: add missing additionalProperties values to schema

Unless we want to disable schema validation under a specific key,
additionalProperties should be set to false in each object. When
we want to disable checking for subschemas, we should probably 
set additionalProprties to true, just to make it clear that it was
the initial intent (and not a forgotten key).

This PR adds:
- 8 tests that fail without the accompanying schema fixes
- multiple additionalProperties definitions to object definitions
- misc format modifications for ease of auditing schema
- a spelling fix to one key
```

## Additional Context
This work is somewhat related to https://github.com/canonical/cloud-init/pull/1532.

## Validation

```
git checkout HEAD^
tox -e py3  # observe 8 failures
git checkout -
tox -e py3  # observe no failures
```

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -2,6 +2,7 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "$defs": {
     "users_groups.groups_by_groupname": {
+      "additionalProperties": false,
       "patternProperties": {
         "^.+$": {
           "label": "<group_name>",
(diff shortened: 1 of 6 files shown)


holmanb:
I guess since I changed some schema values and tests didn't fail I suppose I aught to provide a couple of tests. I'm putting this on ice until those are added.

holmanb:
Assuming tests pass, I believe that this is ready for review.

TheRealFalcon (approved):
LGTM!

PR #1549, opened 2022-06-29
Schema deprecation handling

aciba90:
## Proposed Commit Message

```
schema: Add deprecation handling

Add machinery to react to the schema annotation tag `deprecated`
consisting of:

- Log deprecation warnings if the entrypoint is `main`
- Print to stdout if the entrypoint is `cloud-init schema`
- Render deprecation info in docs

Activate deprecations:

- cc_ca_certs.ca_certs.remove-defaults
- cc_ca_certs.ca-certs
- cc_package_update_upgrade_install.{apt_update,apt_upgrade,
  apt_reboot_if_required}
- cc_power_state_change.power_state.delay with `^\\+?[0-9]+$`
- cc_user_groups.users_groups.user.{groups,lock-passwd}
- distro.add_user.kwargs as group dict

As `jsonschema` does not return the annotations nor provide
more flexible hooks, we had to hook into it via 
extending/implementing `validators` and overriding
`Validator.is_valid`.
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-774

I included the suggestions of: https://github.com/canonical/cloud-init/pull/1273#discussion_r817359121

<details>
  <summary>Example of generated documentation:</summary>
  
  image
</details>

## Test Steps

```sh
cat <<EOT > /tmp/user-data
#cloud-config
packages:
  - htop
apt_update: true
apt_upgrade: true
apt_reboot_if_required: true
EOT
lxc launch ubuntu-daily:jammy jj --config=user.user-data="$(cat /tmp/user-data.yaml)"
lxc exec jj -- cloud-init status --wait

$ lxc exec jj -- grep -ni depr /var/log/cloud-init.log
310:2022-06-29 13:39:08,608 - schema.py[WARNING]: Deprecated cloud-config provided:
311:apt_reboot_if_required: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_reboot_if_required``. Default: ``false``
312:apt_update: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_update``. Default: ``false``
313:apt_upgrade: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_upgrade``. Default: ``false``
1005:2022-06-29 13:39:52,616 - schema.py[WARNING]: Deprecated cloud-config provided:
1006:apt_reboot_if_required: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_reboot_if_required``. Default: ``false``
1007:apt_update: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_update``. Default: ``false``
1008:apt_upgrade: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_upgrade``. Default: ``false`

lxc file push /tmp/user-data dev/root/user-data
$ lxc exec jj -- cloud-init schema -c user-data       
Cloud config schema deprecations: apt_reboot_if_required: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_reboot_if_required``. Default: ``false``, apt_update: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_update``. Default: ``false``, apt_upgrade: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_upgrade``. Default: ``false``
Valid cloud-config: user-data

$ lxc exec dev -- cloud-init schema --annotate -c user-data
#cloud-config
packages:
  - htop
apt_update: true		# D1
apt_upgrade: true		# D2
apt_reboot_if_required: true		# D3

# Deprecations: -------------
# D1: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_update``. Default: ``false``
# D2: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_upgrade``. Default: ``false``
# D3: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_reboot_if_required``. Default: ``false``


Valid cloud-config: user-data

```

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -12,7 +12,8 @@ import typing
 from collections import defaultdict
 from copy import deepcopy
 from functools import partial
-from typing import List, Optional, Tuple, Union, cast
+from itertools import chain
+from typing import List, NamedTuple, Optional, Tuple, Type, Union, cast
 
 import yaml
 
@@ -76,10 +77,45 @@ else:
 # pylint: enable=E1101
 
 
+class SchemaProblem(NamedTuple):
+    path: str
+    message: str
+
+    def format(self) -> str:
+        return f"{self.path}: {self.message}"
+
+
+class SchemaProblems(NamedTuple):
+    problems: Tuple[SchemaProblem, ...] = ()
+
+    def format(
+        self, prefix: Optional[str] = None, separator: str = ", "
+    ) -> str:
+        formatted = separator.join(map(lambda p: p.format(), self.problems))
+        if prefix:
+            formatted = f"{prefix}{formatted}"
+        return formatted
+
+    def __len__(self) -> int:
+        return len(self.problems)
+
+    def __bool__(self) -> bool:
+        return bool(self.problems)
+
+    def __add__(self, other) -> "SchemaProblems":
+        return SchemaProblems(
+            tuple(filter(None, chain(self.problems, other.problems)))
+        )
+
+
 class SchemaValidationError(ValueError):
     """Raised when validating a cloud-config file against a schema."""
 
-    def __init__(self, schema_errors=(), schema_deprecations=()):
+    def __init__(
+        self,
+        schema_errors: Optional[SchemaProblems] = None,
+        schema_deprecations: Optional[SchemaProblems] = None,
+    ):
         """Init the exception an n-tuple of schema errors.
 
         @param schema_errors: An n-tuple of the format:
@@ -87,19 +123,21 @@ class SchemaValidationError(ValueError):
         @param schema_deprecations: An n-tuple of the format:
             ((flat.config.key, msg),)
         """
-        error_messages = [
-            "{0}: {1}".format(config_key, message)  # TODO factor this
-            for config_key, message in schema_errors
-        ]
-        message = "Cloud config schema errors: {0}".format(
-            ", ".join(error_messages)
-        )
+        message = ""
+        if schema_errors:
+            message += schema_errors.format("Cloud config schema errors: ")
+        if schema_deprecations:
+            if message:
+                message += "\n\n"
+            message += schema_deprecations.format(
+                "Cloud config schema deprecations: "
+            )
         super().__init__(message)
         self.schema_errors = schema_errors
         self.schema_deprecations = schema_deprecations
 
     def has_errors(self) -> bool:
-        return bool(self.schema_errors)
+        return self.schema_errors is not None and bool(self.schema_errors)
 
 
 def is_schema_byte_string(checker, instance):
@@ -116,32 +154,40 @@ def is_schema_byte_string(checker, instance):
     ) or isinstance(instance, (bytes,))
 
 
-# TODO error handling
-from jsonschema import ValidationError
-class SchemaDeprecationError(ValidationError):
-    pass
+def _validator_deprecated(
+    _validator,
+    deprecated: bool,
+    _instance,
+    schema: dict,
+    error_type: Type[Exception],
+):
+    """Jsonschema validator for `deprecated` items.
 
-
-def validator_deprecated(validator, deprecated: bool, instance, schema: dict):
-    """
-    TODO
-    a validator instance
-    the value of the property being validated within the instance
-    the instance
-    the schema
-    raise:
+    It raises a instance of `error_type` if deprecated that must be handled,
+    otherwise the instance is consider faulty.
     """
     if deprecated:
-        yield SchemaDeprecationError(schema.get("description", "DEPRECATED"))
+        yield error_type(schema.get("description", "DEPRECATED"))
 
 
-def anyOf(validator, anyOf, instance, schema):
+def _anyOf(validator, anyOf, instance, _schema, error_type: Type[Exception]):
+    """Jsonschema validator for `anyOf`.
+
+    It treats occurrences of `error_type` as non-errors, but yield them for
+    external processing. Useful to process schema annotations, as `deprecated`.
+    """
+    from jsonschema import ValidationError
+
     all_errors = []
     all_deprecations = []
     for index, subschema in enumerate(anyOf):
-        all_errs = list(validator.descend(instance, subschema, schema_path=index))
-        errs = list(filter(lambda e: not isinstance(e, SchemaDeprecationError), all_errs))
-        deprecations = list(filter(lambda e: isinstance(e, SchemaDeprecationError), all_errs))
+        all_errs = list(
+            validator.descend(instance, subschema, schema_path=index)
+        )
+        errs = list(filter(lambda e: not isinstance(e, error_type), all_errs))
+        deprecations = list(
+            filter(lambda e: isinstance(e, error_type), all_errs)
+        )
         if not errs:
             all_deprecations.extend(deprecations)
             break
@@ -154,14 +200,25 @@ def anyOf(validator, anyOf, instance, schema):
     yield from all_deprecations
 
 
-def oneOf(validator, oneOf, instance, schema):
+def _oneOf(validator, oneOf, instance, _schema, error_type: Type[Exception]):
+    """Jsonschema validator for `oneOf`.
+
+    It treats occurrences of `error_type` as non-errors, but yield them for
+    external processing. Useful to process schema annotations, as `deprecated`.
+    """
+    from jsonschema import ValidationError
+
     subschemas = enumerate(oneOf)
     all_errors = []
     all_deprecations = []
     for index, subschema in subschemas:
-        all_errs = list(validator.descend(instance, subschema, schema_path=index))
-        errs = list(filter(lambda e: not isinstance(e, SchemaDeprecationError), all_errs))
-        deprecations = list(filter(lambda e: isinstance(e, SchemaDeprecationError), all_errs))
+        all_errs = list(
+            validator.descend(instance, subschema, schema_path=index)
+        )
+        errs = list(filter(lambda e: not isinstance(e, error_type), all_errs))
+        deprecations = list(
+            filter(lambda e: isinstance(e, error_type), all_errs)
+        )
         if not errs:
             first_valid = subschema
             all_deprecations.extend(deprecations)
@@ -182,6 +239,32 @@ def oneOf(validator, oneOf, instance, schema):
         )
     else:
         yield from all_deprecations
+
+
+class _DeprecationErrorCreator:
+    """Creates `SchemaDeprecationError`.
+
+    We need to create the error class at runtime as `jsonschema` is an optional
+    dependency.
+    """
+
+    _SchemaDeprecationError = None
+
+    @property
+    def error_type(self) -> Type[Exception]:
+        if self._SchemaDeprecationError is not None:
+            return self._SchemaDeprecationError
+        return self._create_error_class()
+
+    @classmethod
+    def _create_error_class(cls):
+        from jsonschema import ValidationError
+
+        class SchemaDeprecationError(ValidationError):
+            pass
+
+        cls._SchemaDeprecationError = SchemaDeprecationError
+        return cls._SchemaDeprecationError
 
 
 def get_jsonschema_validator():
@@ -227,10 +310,14 @@ def get_jsonschema_validator():
         # pylint: disable=E1123
         validator_kwargs = {"default_types": types}
 
+    # Add deprecation handling
+    SchemaDeprecationError = _DeprecationErrorCreator().error_type
     validators = dict(Draft4Validator.VALIDATORS)
-    validators[DEPRECATED_SCHEMA_KEY] = validator_deprecated
-    validators["oneOf"] = oneOf
-    validators["anyOf"] = anyOf
+    validators[DEPRECATED_SCHEMA_KEY] = partial(
+        _validator_deprecated, error_type=SchemaDeprecationError
+    )
+    validators["oneOf"] = partial(_oneOf, error_type=SchemaDeprecationError)
+    validators["anyOf"] = partial(_anyOf, error_type=SchemaDeprecationError)
 
     cloudinitValidator = create(
         meta_schema=strict_metaschema,
@@ -239,12 +326,16 @@ def get_jsonschema_validator():
         **validator_kwargs,
     )
 
+    # Add deprecation handling
     def is_valid(self, instance, _schema=None, **__):
         errors = filter(
-            lambda e: not isinstance(e, SchemaDeprecationError),
+            lambda e: not isinstance(  # pylint: disable=W1116
+                e, SchemaDeprecationError
+            ),
             self.iter_errors(instance, _schema),
         )
         return next(errors, None) is None
+
     cloudinitValidator.is_valid = is_valid
 
     return (cloudinitValidator, FormatChecker)
@@ -272,8 +363,12 @@ def validate_cloudconfig_metaschema(validator, schema: dict, throw=True):
         # sites
         if throw:
             raise SchemaValidationError(
-                schema_errors=(
-                    (".".join([str(p) for p in err.path]), err.message),
+                schema_errors=SchemaProblems(
+                    (
+                        SchemaProblem(
+                            ".".join([str(p) for p in err.path]), err.message
+                        ),
+                    )
                 )
             ) from err
         LOG.warning(
@@ -305,7 +400,7 @@ def validate_cloudconfig_schema(
     @param log_details: Boolean, when True logs details of validation errors.
        If there are concerns about logging sensitive userdata, this should
        be set to False.
-    @param log_deprecations: TODO
+    @param log_deprecations: Controls whether to log deprecations or not.
 
     @raises: SchemaValidationError when provided config does not validate
         against the provided schema.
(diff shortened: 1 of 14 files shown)


aciba90:
Example of generated documentation:

image

aciba90:
I have fixed conflicts and added some integration tests. There are some deprecated keys in our schema that I did not activate, as:

- [user_groups.sudo](https://github.com/canonical/cloud-init/blob/a79bc737577cdfb5937850ee865a341137a19880/cloudinit/config/schemas/schema-cloud-config-v1.json#L161)
- [cc_grub_dpkg.grub-dpkg](https://github.com/canonical/cloud-init/blob/a79bc737577cdfb5937850ee865a341137a19880/cloudinit/config/schemas/schema-cloud-config-v1.json#L803)
- etc.

This PR is getting bigger than I would like to admit. If you agree, we could activate them in follow-up PRs, with good testing.

cloudinit/config/schema.py
```diff
@@ -109,6 +155,164 @@ def is_schema_byte_string(checker, instance):
...
+    otherwise the instance is consider faulty.
+    """
+    if deprecated:
+        deprecated_time = schema.get(DEPRECATED_TIME_KEY)
+        description = schema.get("description")
+        msg = _add_deprecation_msg(description, deprecated_time)
+        yield error_type(msg)
+
+
+def _anyOf(validator, anyOf, instance, _schema, error_type: Type[Exception]):
```
TheRealFalcon, line 193: Is this, along with _oneOf and _allOf, using the same implementation in the base but just adding the yield?

cloudinit/config/schema.py
```diff
@@ -135,28 +339,55 @@ def get_jsonschema_validator():
...
+
+    cloudinitValidator = create(
+        meta_schema=strict_metaschema,
+        validators=validators,
+        version="draft4",
+        **validator_kwargs,
+    )
+
+    # Add deprecation handling
+    def is_valid(self, instance, _schema=None, **__):
```
TheRealFalcon, line 376: Does this mean that our schema will no longer be valid against the built-in validators?

TheRealFalcon (changes requested):
I left some comments inline, but overall I like the the direction.

Additionally, when using the following yaml:
```yaml
#cloud-config
packages:
  - htop
apt_update: true
apt_upgrade: true
apt_reboot_if_required: true
users:
  - name: james
    hashed-passwd: somepassword
```

when I run `cloud-init schema --annotate --config-file <path>`, I get

```
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 732, in validate_cloudconfig_file
    validate_cloudconfig_schema(
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 490, in validate_cloudconfig_schema
    raise SchemaValidationError(errors, deprecations)
cloudinit.config.schema.SchemaValidationError: Cloud config schema errors: users.0: {'name': 'james', 'hashed-passwd': 'somepassword'} is not valid under any of the given schemas

Cloud config schema deprecations: apt_reboot_if_required: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_reboot_if_required``. Default: ``false``, apt_update: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_update``. Default: ``false``, apt_upgrade: DEPRECATED. Dropped in EOL Ubuntu bionic. Use ``package_upgrade``. Default: ``false``

  File "/usr/bin/cloud-init", line 33, in <module>
    sys.exit(load_entry_point('cloud-init==22.2', 'console_scripts', 'cloud-init')())
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 1066, in main
    retval = util.log_time(
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 2621, in log_time
    ret = func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 1149, in handle_schema_args
    validate_cloudconfig_file(
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 738, in validate_cloudconfig_file
    annotated_cloudconfig_file(
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 641, in annotated_cloudconfig_file
    return _Annotator(cloudconfig, original_content, schemamarks).annotate(
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 614, in annotate
    errors_by_line = self._build_errors_by_line(schema_errors)
  File "/usr/lib/python3/dist-packages/cloudinit/config/schema.py", line 530, in _build_errors_by_line
    errors_by_line[self._schemamarks[path]].append(msg)
KeyError: 'users.0'
```

cloudinit/config/schema.py
```diff
@@ -75,24 +78,67 @@ class MetaSchema(typing.TypedDict):
...
+class SchemaProblem(NamedTuple):
+    path: str
+    message: str
+
+    def format(self) -> str:
+        return f"{self.path}: {self.message}"
+
+
+class SchemaProblems(NamedTuple):
+    problems: Tuple[SchemaProblem, ...] = ()
```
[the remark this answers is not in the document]
aciba90: Refactored, thanks for the suggestion.

cloudinit/config/schema.py
```diff
@@ -232,76 +469,178 @@ def validate_cloudconfig_schema(
...
             details = (
+                "Invalid cloud-config provided: "
                 "Please run 'sudo cloud-init schema --system' to "
                 "see the schema errors."
             )
-        LOG.warning("Invalid cloud-config provided: %s", details)
+        LOG.warning(details)
+
+
+class _Annotator:
```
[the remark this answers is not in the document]
aciba90: Thanks!

cloudinit/config/schema.py
```diff
@@ -75,24 +78,67 @@ class MetaSchema(typing.TypedDict):
...
-            for config_key, message in schema_errors
-        ]
-        message = "Cloud config schema errors: {0}".format(
-            ", ".join(error_messages)
-        )
-        super(SchemaValidationError, self).__init__(message)
+        self.schema_deprecations = schema_deprecations
+
+    def has_errors(self) -> bool:
+        return self.schema_errors is not None and bool(self.schema_errors)
```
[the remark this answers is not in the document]
aciba90: ::facepalm::
Fixed.

cloudinit/config/schema.py
aciba90: 1. `_validator_deprecated` yields a `SchemaDeprecationError` if `deprecated: true` is found in a sub-schema.
2. `_anyOf` same as base plus:
   * If the instance matches a subschema then yields `SchemaDeprecationError`s if present.
3. `_oneOf`: similar as `_anyOf`
4. `_allOf`: base + yield all `SchemaDeprecationError`s present in sub-schemas.
5. `is_valid`: base + ignore `SchemaDeprecationError`s.

I did try to document these functions, please let me know if you think we need to improve it.

cloudinit/config/schema.py
aciba90: No, it does not. We needed because we add extra meaning to the `deprecated` key, which in `draft4` is not defined and in newer drafts is treated as an [annotation](https://json-schema.org/understanding-json-schema/reference/generic.html?highlight=deprecated#annotations).

Other validators will ignore `deprecated` and `deprecated.time`.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1,6 +1,10 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "$defs": {
+    "eol_bionic": {
+      "deprecated": true,
+      "deprecated.time": "EOL Ubuntu bionic"
+    },
```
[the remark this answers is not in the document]
aciba90: I have changed its value to `Dropped after April 2027.` (End of standard support of Jammy).

TL;DR I dropped `deprecated.time` and include the extra info in `description`.

Regarding `deprecated.time`, I have thought about it and I came to the conclusion that it is not really needed.

I was solving two problems:

1. Using it as a kind of variable by referencing it in other sub-schemas, in order to avoid repetition.
2. Supplementing the deprecation description.

(1.) implied to merge `deprecated.time` and `description` within code.

Thinking deeper about it:

* The lack of variables, string concatenation, etc in `jsonschema` should not enforce this added complexity
* The moment we add this new key to the schema, we must maintain it to keep backwards compatibility, as we don't know if clients would be coupled to it.
* If the schema repetition / complexity increases a lot, then this problem sounds as more soluble by using a templating engine (`as Jinja2`).

I decided to KISS and add the deprecation date info to `description`.

cloudinit/config/schema.py
```diff
@@ -109,6 +157,106 @@ def is_schema_byte_string(checker, instance):
...
+            "%r is not valid under any of the given schemas" % (instance,),
+            context=all_errors,
+        )
+
+    more_valid = [s for i, s in subschemas if validator.is_valid(instance, s)]
+    if more_valid:
+        more_valid.append(first_valid)
+        reprs = ", ".join(repr(schema) for schema in more_valid)
+        yield ValidationError(
+            "%r is valid under each of %s" % (instance, reprs)
```
holmanb, line 250: Is this correct? It seems odd that this yields a ValidationError that says "this is valid".

aciba90: Yes, it is. When one declares a `oneOf` sub-schema, an instance is valid if and only if it is valid under only one `oneOf`'s sub-schema.

More info: https://json-schema.org/understanding-json-schema/reference/combining.html?highlight=oneof#schema-composition

cloudinit/config/schema.py
```diff
@@ -109,6 +157,106 @@ def is_schema_byte_string(checker, instance):
...
+    It raises a instance of `error_type` if deprecated that must be handled,
+    otherwise the instance is consider faulty.
+    """
+    if deprecated:
+        description = schema.get("description")
+        msg = _add_deprecation_msg(description)
+        yield error_type(msg)
+
+
+def _anyOf(validator, anyOf, instance, _schema, error_type: Type[Exception]):
```
holmanb, line 185: Rather than writing a custom validator implementations, could we extend the existing functions instead?

Poking around at how other projects extend jsonschema validators, there appears to be a general pattern that allows using the existing jsonschema validators and adding extra functionality:

```
def _extend_oneOf(self, validator_class):
        validate_oneOf = validator_class.VALIDATORS["oneOf"]

        def custom_oneOf(validator, properties, instance, schema):
            for error in validate_oneOf(
                validator, properties, instance, schema,
            ):
                yield error

            # add custom functionality here
        return validators.extend(
            validator_class, {"oneOf": custom_oneOf},
        ) 
```

Since the `.extend()` interface is intended for this use case, I'd like to see us use it so we don't have to worry about drift from upstream's validator function implementations, additional test coverage for basic jsonschema functionality, etc.

tests/unittests/config/test_schema.py
```diff
@@ -852,14 +1026,88 @@ def test_label_overrides_property_name(self):
...
+                    "$schema": "http://json-schema.org/draft-04/schema#",
+                    "$defs": {"eol_bionic": {"deprecated": True}},
+                    "properties": {
+                        "prop1": {
+                            "allOf": [
+                                {
+                                    "type": ["string", "integer"],
+                                    "description": "<description>",
+                                },
+                                {"$ref": "#/$defs/eol_bionic"},
```
TheRealFalcon, line 1068: I think this needs to be updated?

tests/unittests/config/test_schema.py
```diff
@@ -852,14 +1026,88 @@ def test_label_overrides_property_name(self):
...
+                                {"type": ["string", "integer"]},
+                                {"$ref": "#/$defs/eol_bionic"},
+                            ]
+                        }
+                    },
+                },
+                ("**prop1:** (string/integer) DEPRECATED. <description>"),
+            ),
+        ],
+    )
```
TheRealFalcon, line 1096: Might be worth adding an `anyOf` or `allOf` here with two different descriptions where one is deprecated and showing that the non-deprecated description shows up in docs.

TheRealFalcon (changes requested):
Thanks for the changes @aciba90 ! I have some more comments here in addition to Brett's, but overall I think we're looking good.

For the integration tests, I think we can consolidate some of them. I think we can remove the additions to the test_apt.py and test_users_groups.py integration tests completely. That behavior should be covered by unit tests and then general functionality in test_schema.py. In test_schema.py, I think can we can have tests centered around a single launch using a class and the class_client fixture. Within the class, there can be a test for checking the behavior of `cloud-init schema` (basically what you already have there), and another test checking `verify_clean_log` and that the proper warnings are in /var/log/cloud-init.log.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -34,10 +34,30 @@
...
+              },
+              "minItems": 1
+            },
+            {
+              "type": "object",
+              "patternProperties": {
+                "^.+$": {
+                  "label": "<group_name>",
+                  "description": "When providing an object for users.groups the ``<group_name>`` keys are the groups to add this user to",
+                  "deprecated": true,
```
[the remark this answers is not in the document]
aciba90: We do emit that warning. I have added tests to make sure it's covered:

* https://github.com/canonical/cloud-init/pull/1549/files#diff-665a8909889380de684d22ea92285daed075400bc82b20258487a15a12e420cfR384
* https://github.com/canonical/cloud-init/pull/1549/files#diff-ccf5aac2a6067d8a0782d74b172805f80e71225a212822e4f4e158f9ca1e6e3aR547

cloudinit/config/schema.py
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/schema.py
```diff
@@ -109,6 +157,106 @@ def is_schema_byte_string(checker, instance):
...
+        all_errors.extend(errs)
+    else:
+        yield ValidationError(
+            "%r is not valid under any of the given schemas" % (instance,),
+            context=all_errors,
+        )
+    yield from all_deprecations
+
+
+def _oneOf(validator, oneOf, instance, _schema, error_type: Type[Exception]):
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/schema.py
```diff
@@ -135,28 +283,53 @@ def get_jsonschema_validator():
...
-            default_types=types,
+        validator_kwargs = {"default_types": types}
+
+    # Add deprecation handling
+    validators = dict(Draft4Validator.VALIDATORS)
+    validators[DEPRECATED_KEY] = partial(
+        _validator_deprecated, error_type=SchemaDeprecationError
+    )
+    validators["oneOf"] = partial(_oneOf, error_type=SchemaDeprecationError)
+    validators["anyOf"] = partial(_anyOf, error_type=SchemaDeprecationError)
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1309,7 +1341,11 @@
               "default": "now",
               "oneOf": [
                 {"type": "integer", "minimum": 0},
-                {"type": "string", "pattern": "^\\+?[0-9]+$"},
+                {
+                    "type": "string",
+                    "pattern": "^\\+?[0-9]+$",
+                    "deprecated": true
```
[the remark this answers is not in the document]
aciba90: Done.

tests/unittests/config/test_schema.py
aciba90: Done.

tests/unittests/config/test_schema.py
```diff
@@ -852,14 +1026,88 @@ def test_label_overrides_property_name(self):
...
+                        "eol_bionic": {
+                            "deprecated": True,
+                            "description": "<description>",
+                        }
+                    },
+                    "properties": {
+                        "prop1": {
+                            "allOf": [
+                                {"type": ["string", "integer"]},
+                                {"$ref": "#/$defs/eol_bionic"},
```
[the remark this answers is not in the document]
aciba90: Done.

tests/unittests/config/test_schema.py
aciba90: Done.

cloudinit/config/schema.py
aciba90: That would be nice, but we cannot extend them in that way as:

Let's suppose we have the following schema and instance of it

```py
schema = {
    "$schema": "http://json-schema.org/draft-04/schema#",
    "properties": {
        "delay": {
            "oneOf": [
                {"type": "integer", "description": "Minutes of delay."},
                {
                    "type": "string",
                    "deprecated": True,
                    "description": "DEPRECATED. Use an integer.",
                },
            ]
        },
    },
}

instance = {"delay": "+5"}
```

Then if extend `onOf` as suggested, `validate_oneOf` wouldn't yield any error as the instance is valid and `custom_oneOf` wouldn't be able to determine to which sub-schema `delay` belongs to. And we should not emit false-positive deprecation warning.

Solving the above problem implies to have a tight integration with `oneOf`, `anyOf`, etc. As they do not offer an API / hooks to do so, the only way to do so (I think) is to override them.

Note: A similar test case covering this requirement is added here: https://github.com/canonical/cloud-init/pull/1549/files#diff-ccf5aac2a6067d8a0782d74b172805f80e71225a212822e4f4e158f9ca1e6e3aR514

aciba90:
Thanks, @holmanb and @TheRealFalcon, for your review and useful suggestions!

I think I have covered the remaining comments, please could you review it again?

One point that is still unresolved is: https://github.com/canonical/cloud-init/pull/1549#issuecomment-1177336456

tests/unittests/config/test_schema.py
```diff
@@ -852,14 +1040,127 @@ def test_label_overrides_property_name(self):
...
+                                },
+                                {
+                                    "type": "number",
+                                    "description": "<description>",
+                                },
+                            ]
+                        },
+                    },
+                },
+                "**prop1:** (UNDEFINED)\n",
```
TheRealFalcon, line 1146: Should this be `(UNDEFINED) <description>\n`?

TheRealFalcon:
```
Regarding your https://github.com/canonical/cloud-init/pull/1549#pullrequestreview-1028680743, I have been able to reproduced it in main. There are two things happening:

users.hashed-passwd is not yet added as a deprecated key.
during the annotation phase, an unhandled exception is raised.
1: I thought it would be better to add all the deprecated keys in follow-up PRs, in order to test them and not add unwanted ones.
2: As the error is independent of this PR, I think it would be cleaner to solve it as a separated PR.
```

Works for me. Do we have a ticket detailing the remaining items? If not, we should add one before we merge this PR.

Did my response in https://github.com/canonical/cloud-init/pull/1549#discussion_r916863266 make sense?

I added one more very minor comment as a one-off. Additionally, your explanation for the oneOf/anyOf issue makes sense. I'll give @holmanb till Monday to see if he has anything else to say about it, otherwise I think we're good here.

aciba90 (commented):
> There are some deprecated keys in our schema that I did not activate ...

https://warthogs.atlassian.net/browse/SC-1173 created.

Regarding your [commented error](https://github.com/canonical/cloud-init/pull/1549#pullrequestreview-1028680743), I have found that [lp1885598](https://bugs.launchpad.net/cloud-init/+bug/1885598) covers it.

And I made some questions on both comments, could you please clarify them?

Thanks for the review.

cloudinit/config/schemas/schema-cloud-config-v1.json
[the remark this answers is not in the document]
TheRealFalcon: > Something like <regex1> is deprecated if the given instance matches it?

More or less, yes.

My assumption was that when we encounter `"deprecated": true` in the schema, we'd find a way to log that without having to add separate logs to the code. That's obviously much harder to do in practice, and in cases like this, I'm not sure it even makes sense. I was more wondering if you were planning to try to handle cases like this. If not, that's totally fine, especially with how much extra work it might be to support such an obscure case.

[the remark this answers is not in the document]
aciba90: I think it is not that difficult. We already emit a warning even if the description is not given: https://github.com/canonical/cloud-init/pull/1549/files#diff-ccf5aac2a6067d8a0782d74b172805f80e71225a212822e4f4e158f9ca1e6e3aR555

We could easily extend the (given or not) description with the regex within [_validator_deprecated](_validator_deprecated) as that info is given in the `schema` argument.

Let me create a ticket.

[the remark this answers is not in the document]
aciba90: SC-1174

TheRealFalcon (approved):
I think we're good here now. Thanks for all the work getting this across the finish line @aciba90 !

tests/unittests/config/test_schema.py
aciba90: SC-1175

cloudinit/config/schema.py
holmanb: Sorry for the delayed response on this. +1 Thanks for the explanation!

PR #1550, opened 2022-06-29
sources/azure: refactor ovf-env.xml parsing

cjp256:
- Use ElementTree instead of minidom
- Use namespaces and case sensitive names
- Decouple parsing from usage in config/metadata dictionaries
- More clearly distinguish between NonAzureDataSource() and
  BrokenAzureDataSource() exceptions.  Only raise NonAzureDataSource()
  exception if the `ProvisioningSection` in the windowsazure namespace
  is not found. Any other parsing failures will result in
  BrokenAzureDataSource() being raised.
- Streamline log messages
- Move logic into Azure helper module

There should be no effective change in behavior unless some bad XML
is in the wild and being ignored or failing silently.

### DIFF
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -1122,8 +1122,8 @@ class OvfEnvXml:
         self,
         node,
         name: str,
+        required: bool,
         namespace: str = "wa",
-        error_if_missing: bool = True,
     ):
         matches = node.findall(
             "./%s:%s" % (namespace, name), OvfEnvXml.NAMESPACES
@@ -1131,7 +1131,7 @@ class OvfEnvXml:
         if len(matches) == 0:
             msg = "No ovf-env.xml configuration for %r" % name
             LOG.debug(msg)
-            if error_if_missing:
+            if required:
                 raise BrokenAzureDataSource(msg)
             return None
         elif len(matches) > 1:
@@ -1146,8 +1146,8 @@ class OvfEnvXml:
         self,
         node,
         name: str,
+        required: bool,
         decode_base64: bool = False,
-        error_if_missing: bool = False,
         parse_bool: bool = False,
         default=None,
     ):
@@ -1155,7 +1155,7 @@ class OvfEnvXml:
         if len(matches) == 0:
             msg = "No ovf-env.xml configuration for %r" % name
             LOG.debug(msg)
-            if error_if_missing:
+            if required:
                 raise BrokenAzureDataSource(msg)
             return default
         elif len(matches) > 1:
@@ -1179,34 +1179,45 @@ class OvfEnvXml:
         return value
 
     def _parse_linux_configuration_set_section(self, root):
-        provisioning_section = self._find(root, "ProvisioningSection")
+        provisioning_section = self._find(
+            root, "ProvisioningSection", required=True
+        )
         config_set = self._find(
             provisioning_section,
             "LinuxProvisioningConfigurationSet",
+            required=True,
         )
 
         self.custom_data = self._parse_property(
-            config_set, "CustomData", decode_base64=True
+            config_set,
+            "CustomData",
+            decode_base64=True,
+            required=False,
         )
         self.username = self._parse_property(
-            config_set, "UserName", error_if_missing=True
+            config_set, "UserName", required=True
         )
-        self.password = self._parse_property(config_set, "UserPassword")
+        self.password = self._parse_property(
+            config_set, "UserPassword", required=False
+        )
         self.hostname = self._parse_property(
-            config_set, "HostName", error_if_missing=True
+            config_set, "HostName", required=True
         )
         self.disable_ssh_password_auth = self._parse_property(
             config_set,
             "DisableSshPasswordAuthentication",
             parse_bool=True,
+            required=False,
         )
 
         self._parse_ssh_section(config_set)
 
     def _parse_platform_settings_section(self, root):
-        platform_settings_section = self._find(root, "PlatformSettingsSection")
+        platform_settings_section = self._find(
+            root, "PlatformSettingsSection", required=True
+        )
         platform_settings = self._find(
-            platform_settings_section, "PlatformSettings"
+            platform_settings_section, "PlatformSettings", required=True
         )
 
         self.preprovisioned_vm = self._parse_property(
@@ -1214,20 +1225,23 @@ class OvfEnvXml:
             "PreprovisionedVm",
             parse_bool=True,
             default=False,
+            required=False,
         )
         self.preprovisioned_vm_type = self._parse_property(
-            platform_settings, "PreprovisionedVMType"
+            platform_settings,
+            "PreprovisionedVMType",
+            required=False,
         )
 
     def _parse_ssh_section(self, config_set):
         self.public_keys = []
 
-        ssh_section = self._find(config_set, "SSH", error_if_missing=False)
+        ssh_section = self._find(config_set, "SSH", required=False)
         if ssh_section is None:
             return
 
         public_keys_section = self._find(
-            ssh_section, "PublicKeys", error_if_missing=False
+            ssh_section, "PublicKeys", required=False
         )
         if public_keys_section is None:
             return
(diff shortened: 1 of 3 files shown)


cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,213 @@ def dhcp_log_cb(out, err):
...
+        self.preprovisioned_vm = preprovisioned_vm
+        self.preprovisioned_vm_type = preprovisioned_vm_type
+
+    def __eq__(self, other) -> bool:
+        return self.__dict__ == other.__dict__
+
+    def __repr__(self) -> str:
+        props = [
+            "username=%s" % self.username,
+            "password=%s" % self.password,
```
cjp256, line 1102: This is probably a bad idea in case the object gets logged at some point.  I'll probably drop the repr altogether.

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,213 @@ def dhcp_log_cb(out, err):
...
+            "custom_data=%r" % self.custom_data,
+            "disable_ssh_password_auth=%s" % self.disable_ssh_password_auth,
+            "public_keys=%s" % self.public_keys,
+            "preprovisioned_vm=%s" % self.preprovisioned_vm,
+            "preprovisioned_vm_type=%s" % self.preprovisioned_vm_type,
+        ]
+        return "OvfEnvXml(" + " ".join(props) + ")"
+
+    @classmethod
+    def parse_text(cls, ovf_env_xml: str) -> "OvfEnvXml":
```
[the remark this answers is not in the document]
cjp256: Is there an advantage?

TIL about `Self`

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,213 @@ def dhcp_log_cb(out, err):
...
+
+class NonAzureDataSource(Exception):
+    pass
+
+
+class OvfEnvXml:
+    OVF_NAMESPACE = "http://schemas.dmtf.org/ovf/environment/1"
+    WA_NAMESPACE = "http://schemas.microsoft.com/windowsazure"
+
+    ns = {
```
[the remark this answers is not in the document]
cjp256: Sure, originally just `ns` for brevity, but that would make sense.  Will change

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,213 @@ def dhcp_log_cb(out, err):
...
+
+    def __init__(
+        self,
+        *,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        hostname: Optional[str] = None,
+        custom_data: Optional[bytes] = None,
+        disable_ssh_password_auth: Optional[bool] = None,
+        public_keys: Optional[list] = None,
```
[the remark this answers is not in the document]
cjp256: I think `list` is the preferred way going forward for python, but I have no data for that :D  Python 3.9 added support for list[T].  I'll update to List[dict] anyways to indicate the type of list.

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,213 @@ def dhcp_log_cb(out, err):
...
+            return
+
+        public_keys_section = self._find(
+            ssh_section, "PublicKeys", error_if_missing=False
+        )
+        if public_keys_section is None:
+            return
+
+        for public_key in public_keys_section.findall(
+            "./wa:PublicKey", self.ns
```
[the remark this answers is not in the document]
cjp256: I don't think there is a difference, but will change :D

cloudinit/sources/helpers/azure.py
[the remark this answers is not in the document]
cjp256: done!

cloudinit/sources/helpers/azure.py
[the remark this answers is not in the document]
cjp256: done!

cloudinit/sources/helpers/azure.py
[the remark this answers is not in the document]
vmarcella: I believe the only advantage would be that it's more semantic for pythons language server recommendations/validations, but it could have no benefit if the language server is already capable of picking up the type on the way you've done it!

Regardless, I wouldn't really classify this as a *blocking* change if you wanted to keep it how it is.

cjp256:
@TheRealFalcon @blackboxsw given the size of the azure modules, would it perhaps make sense to just bite the bullet and make an cloudinit.sources.azure package to better organize Azure bits?

vmarcella (approved):

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,212 @@ def dhcp_log_cb(out, err):
...
+        self,
+        node,
+        name: str,
+        namespace: str = "wa",
+        error_if_missing: bool = True,
+    ):
+        matches = node.findall(
+            "./%s:%s" % (namespace, name), OvfEnvXml.NAMESPACES
+        )
+        if len(matches) == 0:
```
anhvoms, line 1144:
```suggestion
        if not matches or len(matches) == 0:
```

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,212 @@ def dhcp_log_cb(out, err):
...
+            msg = "No ovf-env.xml configuration for %r" % name
+            LOG.debug(msg)
+            if error_if_missing:
+                raise BrokenAzureDataSource(msg)
+
+            return None
+
+        if len(matches) > 1:
```
anhvoms, line 1152:
```suggestion
        elif len(matches) > 1:
```

tests/unittests/sources/test_azure_helper.py
```diff
@@ -1414,4 +1416,241 @@ def test_instantiates_shim_with_kwargs(self):
...
+
+        with pytest.raises(azure_helper.NonAzureDataSource) as exc_info:
+            azure_helper.OvfEnvXml.parse_text(ovf)
+
+        assert (
+            str(exc_info.value)
+            == "Ignoring non-Azure ovf-env.xml: ProvisioningSection not found"
+        )
+
+
```
anhvoms, line 1655: Can you add a test for bad XML?

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,212 @@ def dhcp_log_cb(out, err):
...
+        """Parser for ovf-env.xml data.
+
+        :raises NonAzureDataSource: if XML is not in Azure's format.
+        :raises BrokenAzureDataSource: if XML is unparseable or invalid.
+        """
+        try:
+            root = ElementTree.fromstring(ovf_env_xml)
+        except Exception as e:
+            error_str = "Invalid ovf-env.xml: %s" % e
+            raise BrokenAzureDataSource(error_str) from e
```
anhvoms, line 1120: We should report a diagnostic event for invalid ovf-env.xml

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,212 @@ def dhcp_log_cb(out, err):
...
+
+        # If there's no provisioning section, it's not Azure ovf-env.xml.
+        if not root.find("./wa:ProvisioningSection", cls.NAMESPACES):
```
anhvoms, line 1123: report a diagnostic event here as well in the case of error

anhvoms (changes requested):

tests/unittests/sources/test_azure_helper.py
cjp256: Excellent point, will do!

cloudinit/sources/helpers/azure.py
cjp256: It does get reported in crawl_metadata 👍

cloudinit/sources/helpers/azure.py
```diff
@@ -1055,4 +1055,199 @@ def dhcp_log_cb(out, err):
...
+    ):
+        matches = node.findall("./wa:" + name, OvfEnvXml.NAMESPACES)
+        if not matches or len(matches) == 0:
+            msg = "No ovf-env.xml configuration for %r" % name
+            LOG.debug(msg)
+            if error_if_missing:
+                raise BrokenAzureDataSource(msg)
+
+            return default
+
```
cjp256, line 1164:
```suggestion
```

cloudinit/sources/helpers/azure.py
cjp256: (skipped the `not matches`)

anhvoms (approved):

TheRealFalcon:
> given the size of the azure modules, would it perhaps make sense to just bite the bullet and make an cloudinit.sources.azure package to better organize Azure bits?

@cjp256 Yes, I think that makes sense. If you want to do it here that's fine, but I'm assuming you mean in a future PR?

cjp256:
> > given the size of the azure modules, would it perhaps make sense to just bite the bullet and make an cloudinit.sources.azure package to better organize Azure bits?
> 
> @cjp256 Yes, I think that makes sense. If you want to do it here that's fine, but I'm assuming you mean in a future PR?

Yeah, just thinking out loud, we can do it in the future if the churn is worth it.

TheRealFalcon (approved):
LGTM!

PR #1551, opened 2022-06-29
Remove schema errors from log

TheRealFalcon:
Remove schema errors from log

When schema errors are encountered, the section of userdata in question
gets printed to the cloud-init log. As this could contain sensitive
data, so log a generic warning instead and redirect user to run
`cloud-init schema --system` as root.

LP: #1978422


blackboxsw (approved):
LGTM!  validated locally.

PR #1556, opened 2022-06-30
Resource leak cleanup

holmanb:
```
add tox target for tracing for resource leaks, fix some leaks
```

## Additional Context
I cleaned up a few leaks, but I'm loosing steam on this and would like to merge it as is. My original intent was to clean up all leaks which would make this more useful as a ci / jenkins job, but I have spent too much time on this already and need to move on to other things.

[resource-leak.txt](https://github.com/canonical/cloud-init/files/9023208/resource-leak.txt)

## Test Steps
run the added tox env

### DIFF
--- a/cloudinit/analyze/__main__.py
+++ b/cloudinit/analyze/__main__.py
@@ -6,7 +6,7 @@ import argparse
 import re
 import sys
 from datetime import datetime
-from typing import List, IO
+from typing import IO
 
 from cloudinit.util import json_dumps
 
--- a/cloudinit/analyze/__main__.py
+++ b/cloudinit/analyze/__main__.py
@@ -6,6 +6,7 @@ import argparse
 import re
 import sys
 from datetime import datetime
+from typing import List, IO
 
 from cloudinit.util import json_dumps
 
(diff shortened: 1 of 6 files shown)


cloudinit/analyze/__main__.py
```diff
@@ -192,6 +192,10 @@ def analyze_boot(name, args):
     }
 
     outfh.write(status_map[status_code].format(**kwargs))
+    if outfh != sys.stdout:
```
TheRealFalcon, line 195: I think that before merging this as-is, we should at least approach the return from `configure_io` consistently. If you'd rather not deal with that right now, I think I'd prefer we remove this section before merging.

holmanb: The conditional exists because the sys file descriptors are provided by the sys module and therefore I don't think we should close them, but if outfh/infh are allocated by us, we should clean them up.

What exactly do you mean by "approach the return from `configure_io` consistently?

TheRealFalcon: Sorry, just that the function is called a few times in this file, but we're only addressing it here.

holmanb: @TheRealFalcon Agreed, thanks! I just pushed a change addressing this comment.

TheRealFalcon (approved):
LGTM!

PR #1563, opened 2022-07-05
ubuntu/devel: fix lintian warnings

aciba90:
do not squash

https://warthogs.atlassian.net/browse/SC-1154

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,17 @@
+cloud-init (22.2-64-g1fcd55d6-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
+
+  * d/control: lintian fixes:
+    + Drop deprecated keys: XS-Python-Version and XB-Python-Version.
+  * d/cloud-init.lintian-overrides: lintian fixes:
+    + Fix systemd-service-file-refers-to-unusual-wantedby-target format.
+  * d/cloud-init.postinst: lintian fixes:
+    + Fix command-with-path-in-maintainer-script for grub-install.
+  * d/source/lintian-overrides: lintian fixes:
+    + silence binary-nmu-debian-revision-in-source bug:
+      https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1014584
+
+ -- Alberto Contreras <alberto.contreras@canonical.com>  Fri, 08 Jul 2022 13:14:41 +0200
+
 cloud-init (22.2-64-g1fcd55d6-0ubuntu1~22.10.1) kinetic; urgency=medium
 
   * d/control: add python3-debconf as Depends and Build-Depends
--- a/debian/cloud-init.postinst
+++ b/debian/cloud-init.postinst
@@ -140,7 +140,7 @@ fix_1336855() {
   [ -r /proc/cmdline ] || return 0
 
   # Don't do anything unless we have grub
-  command -v grub-install > /dev/null 2>&1 || return 0
+  command -v grub-install > /dev/null || return 0
 
   # First, identify the kernel device for the parent.
   for parm in $(cat /proc/cmdline); do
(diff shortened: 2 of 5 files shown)


aciba90:
https://warthogs.atlassian.net/browse/SC-1154

debian/changelog
```diff
@@ -1,3 +1,9 @@
+cloud-init (22.2-64-g1fcd55d6-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
+
+  * Fix lintian warnings
```
blackboxsw, line 3: When we touch debian/* packaging files we generally want to refer to the specific files changed and the reason behind it (as SRU tend to look more closely at the packaging related changes) and it is more intuitive for review to  see those packaging changes called out specifically

So we need to see something like the specific warning handled and the reason for each file
* d/source/lintian-overrides:
  + silence binary-nmu-debian-revision-in-source as deb ver includes series
* d/control: lintian fixes: drop -XS-Python-Version  and XB-Python-Version because XXX
* d/cloud-init.lintian-overrides:
  + format of systemd-service-file-refers-to-unusual-wantedby-target changed?
* d/source/cloud-init.postinst: avoid hard-coded paths for grub-install

blackboxsw (changes requested):
Thanks a lot for jumping on this @aciba90. I think we ultimately want to have separate commits for each debian/* file change which explain the reason for each lintian fix. And we want to improve the d/changelog entries with that same context to align with how we documented previous debian/* file changes in previous cloud-init releases.

blackboxsw (changes requested):
Thanks a lot for jumping on this @aciba90. I think we ultimately want to have separate commits for each debian/* file change which explain the reason for each lintian fix. And we want to improve the d/changelog entries with that same context to align with how we documented previous debian/* file changes in previous cloud-init releases.

debian/cloud-init.postinst
```diff
@@ -309,7 +309,7 @@ fix_lp1889555() {
     [ -f /var/lib/cloud/instance/sem/config_grub_dpkg ] || return 0
 
     # Don't do anything unless we have grub
-    [ -x /usr/sbin/grub-install ] || return 0
+    [ -x "$(which grub-install)" ] || return 0
```
blackboxsw, line 312: Let's align with how we are doing this in general to check for presence of a command in your path in ds-identify and other postinst script functions
```suggestion
    command -v grub-install > /dev/null 2>&1 || return 0
```

aciba90:
Thanks, @blackboxsw, for the explanations.

I have separated changes under `debian/` into one commit per file and added a supplementary commit with changes to `debian/changelog`. I hope this is ok. If we need one commit updating the changelog per commit, I can fix it.

debian/cloud-init.postinst
paride: +1 on this: `which` is half deprecated (prints a deprecation warning in Debian, which was silenced in Ubuntu *for now*).

debian/changelog
aciba90: Done.

aciba90:
> Pending: [#1566 (comment)](https://github.com/canonical/cloud-init/pull/1566#issuecomment-1177629649)

Done, ready to be reviewed.

blackboxsw (approved):
I've commented on the related bug. I think this is a lintian bug in matching ubuntu package verison that we can sort long-term.

One alternative we could take (and discuss later) is whether cloud-init pkg versioning can just use augmented versioning .XX.YY.1 instead of diminished ~XX.YY.1 versioning now that we happen to be including the release-specific suffix in Ubuntu devel pkg releases anyway.

In either case, I think this is really a bug we can work in lintian upstream as this feels like a reasonable packaging versioning policy we currently have. So, let's silence this nmu warning in the meantime.

PR #1568, opened 2022-07-07
Enable pytest to run in parallel

holmanb:
```
Enable pytest to run in parallel
    
- add a tox target for running unit tests in parallel
- add missing mocks
- make parametrized gzip bytes deterministic
```
Turn your cpu into a space heater with `tox -e py3-fast`.

Comparison:

tox -e py3
```python
======================== 3771 passed, 9 skipped, 1 deselected, 1 warning in 75.59s (0:01:15) =========================
______________________________________________________ summary _______________________________________________________
  py3: commands succeeded
  congratulations :)
tox -e py3  80.41s user 3.00s system 97% cpu 1:25.96 total
```
vs:

tox -e py3-fast
```python
============================================================================= 3771 passed, 10 skipped, 13 warnings in 14.00s ==============================================================================
_________________________________________________________________________________________________ summary _________________________________________________________________________________________________
  py3-fast: commands succeeded
  congratulations :)
tox -e py3-fast  101.85s user 5.63s system 450% cpu 23.877 total
```

Regarding gzip: The gzip change is required because during test collection if different processes collect different tests xdist assumes something is wrong and bails. Manually setting time makes these bytes deterministic.

### DIFF
--- a/tests/unittests/net/test_network_state.py
+++ b/tests/unittests/net/test_network_state.py
@@ -99,14 +99,14 @@ class TestNetworkStateParseConfig(CiTestCase):
 
 
 class TestNetworkStateParseConfigV2(CiTestCase):
-    @mock.patch("cloudinit.net.network_state.get_interfaces_by_mac")
-    def test_version_2_ignores_renderer_key(self, m_subp):
+    def test_version_2_ignores_renderer_key(self):
         ncfg = {"version": 2, "renderer": "networkd", "ethernets": {}}
-        nsi = network_state.NetworkStateInterpreter(
-            version=ncfg["version"], config=ncfg
-        )
-        nsi.parse_config(skip_broken=False)
-        self.assertEqual(ncfg, nsi.as_dict()["config"])
+        with mock.patch("cloudinit.net.network_state.get_interfaces_by_mac"):
+            nsi = network_state.NetworkStateInterpreter(
+                version=ncfg["version"], config=ncfg
+            )
+            nsi.parse_config(skip_broken=False)
+            self.assertEqual(ncfg, nsi.as_dict()["config"])
 
 
 class TestNetworkStateParseNameservers:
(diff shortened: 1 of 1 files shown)


aciba90:
> Turn your cpu into a space heater with tox -e py3-fast.

XD

I like this PR, hopefully it gets merged. And it gets the default at some point. Thanks!

tox.ini
```diff
@@ -135,6 +135,14 @@ commands = {envpython} -m pytest \
             -m "not hypothesis_slow" \
             {posargs:--cov=cloudinit --cov-branch tests/unittests}
 
+# experimental
+[testenv:py3-fast]
+deps =
+    -r{toxinidir}/test-requirements.txt
+    pytest-xdist
+commands = {envpython} -m pytest -n auto -m "not hypothesis_slow" -m "not serial"\
```
TheRealFalcon, line 143: Note that `--dist=loadscope` would probably result in similar times without requiring skipping any tests.

https://stackoverflow.com/a/56475200

TheRealFalcon (approved):
Looks good. I left a few comments inline, but the only change I'm adamant about is the `m_subp`.

Feel free to merge once you've made the change.

tox.ini
holmanb: This PR originally used loadscope, but when I switched away from it (and simultaneously removed the cov/duration tracking bits) it ended up about 20s faster. I didn't bother finding the source of the speedup so I don't know whether that's caused by loadscope or cov or duration tracking or something else entirely. Alternatively we could use loadgroup and mark them as groups, but when I did that I saw errors I didn't feel like debugging so I reverted back to this. Laziness prevailed here I guess.

tests/unittests/net/test_network_state.py
```diff
@@ -136,7 +138,9 @@ def test_v1_nameservers_invalid(self):
                 V1_CONFIG_NAMESERVERS_INVALID
             )
 
-    def test_v2_nameservers(self):
+    @mock.patch("cloudinit.net.network_state.get_interfaces_by_mac")
+    @mock.patch("cloudinit.net.get_interfaces_by_mac")
+    def test_v2_nameservers(self, *_):
```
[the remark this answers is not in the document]
holmanb: +1 to mocker, I'll ditch the current approach for something more pythonic

PR #1569, opened 2022-07-07
mounts: fix suggested_swapsize for > 64GB hosts

sstallion:
## Proposed Commit Message

```
mounts: fix suggested_swapsize for > 64GB hosts
```

## Additional Context
If provisioning hosts with more than 64GB of memory, swap will not be created and the following error messages will be logged by cloud-init:
```
2022-07-07 23:45:13,257 - cc_mounts.py[DEBUG]: suggest None swap for 257675.54296875 MB memory with '958841.80859375 MB' disk given max=16384.0 MB [max=16384.0 MB]'
2022-07-07 23:45:13,257 - cc_mounts.py[WARNING]: failed to setup swap: unsupported operand type(s) for /: 'NoneType' and 'int'
```
Upon inspection, this appears to be a bug in `suggested_swapsize`; I could not see a case where the existing conditional would be correct behavior, but I would like to hear from upstream in case I missed something obvious.

## Test Steps
Attempt to provision a system with more than 64GB of RAM with the following config snippet; no swap file will be created, and the above error messages should be present in `/var/log/cloud-init.log`.
```
swap:
  filename: /swap.img
  size: auto
  maxsize: 16G
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -9,6 +9,7 @@
 """Mounts: Configure mount points and swap files"""
 
 import logging
+import math
 import os
 import re
 from string import whitespace
@@ -110,6 +111,8 @@ NETWORK_NAME_RE = re.compile(NETWORK_NAME_FILTER)
 WS = re.compile("[%s]+" % (whitespace))
 FSTAB_PATH = "/etc/fstab"
 MNT_COMMENT = "comment=cloudconfig"
+MB = 2**20
+GB = 2**30
 
 LOG = logging.getLogger(__name__)
 
@@ -210,13 +213,12 @@ def suggested_swapsize(memsize=None, maxsize=None, fsys=None):
     if memsize is None:
         memsize = util.read_meminfo()["total"]
 
-    GB = 2**30
-    sugg_max = 8 * GB
+    sugg_max = memsize * 2
 
     info = {"avail": "na", "max_in": maxsize, "mem": memsize}
 
     if fsys is None and maxsize is None:
-        # set max to 8GB default if no filesystem given
+        # set max to default if no filesystem given
         maxsize = sugg_max
     elif fsys:
         statvfs = os.statvfs(fsys)
@@ -234,35 +236,17 @@ def suggested_swapsize(memsize=None, maxsize=None, fsys=None):
 
     info["max"] = maxsize
 
-    formulas = [
-        # < 1G: swap = double memory
-        (1 * GB, lambda x: x * 2),
-        # < 2G: swap = 2G
-        (2 * GB, lambda x: 2 * GB),
-        # < 4G: swap = memory
-        (4 * GB, lambda x: x),
-        # < 16G: 4G
-        (16 * GB, lambda x: 4 * GB),
-        # < 64G: 1/2 M up to max
-        (64 * GB, lambda x: x / 2),
-    ]
+    if memsize < 4 * GB:
+        minsize = memsize
+    elif memsize < 16 * GB:
+        minsize = 4 * GB
+    else:
+        minsize = round(math.sqrt(memsize / GB)) * GB
 
-    size = None
-    for top, func in formulas:
-        if memsize <= top:
-            size = min(func(memsize), maxsize)
-            # if less than 1/2 memory and not much, return 0
-            if size < (memsize / 2) and size < 4 * GB:
-                size = 0
-                break
-            break
-
-    if size is not None:
-        size = maxsize
+    size = min(minsize, maxsize)
 
     info["size"] = size
 
-    MB = 2**20
     pinfo = {}
     for k, v in info.items():
         if isinstance(v, int):
--- a/tests/unittests/config/test_cc_mounts.py
+++ b/tests/unittests/config/test_cc_mounts.py
@@ -1,13 +1,21 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
+import math
 import os.path
 import re
+from collections import namedtuple
 from unittest import mock
 
 import pytest
+from pytest import approx
 
 from cloudinit.config import cc_mounts
-from cloudinit.config.cc_mounts import create_swapfile
+from cloudinit.config.cc_mounts import (
+    GB,
+    MB,
+    create_swapfile,
+    suggested_swapsize,
+)
 from cloudinit.config.schema import (
     SchemaValidationError,
     get_schema,
@@ -523,6 +531,35 @@ class TestCreateSwapfile:
 
         msg = "fallocate swap creation failed, will attempt with dd"
         assert msg in caplog.text
+
+    # See https://help.ubuntu.com/community/SwapFaq
+    @pytest.mark.parametrize(
+        "memsize,expected",
+        [
+            (256 * MB, 256 * MB),
+            (512 * MB, 512 * MB),
+            (1 * GB, 1 * GB),
+            (2 * GB, 2 * GB),
+            (4 * GB, 4 * GB),
+            (8 * GB, 4 * GB),
+            (16 * GB, 4 * GB),
+            (32 * GB, 6 * GB),
+            (64 * GB, 8 * GB),
+            (128 * GB, 11 * GB),
+            (256 * GB, 16 * GB),
+            (512 * GB, 23 * GB),
+        ],
+    )
+    def test_suggested_swapsize(self, memsize, expected, mocker):
+        mock_stat = namedtuple("mock_stat", "f_frsize f_bfree")
+        mocker.patch(
+            "os.statvfs",
+            # Don't care about available disk space for the purposes of this
+            # test
+            return_value=mock_stat(math.inf, math.inf),
+        )
+        size = suggested_swapsize(memsize, math.inf, "dontcare")
+        assert expected == approx(size)
 
 
 class TestMountsSchema:
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -89,6 +89,7 @@ slingamn
 slyon
 smoser
 sshedi
+sstallion
 stappersg
 steverweber
 t-8ch


sstallion:
> Thanks, @sstallion, for making `cloud-init` better.
> 
> The problem is that there is no [formula](https://jenkins.canonical.com/server-team/view/cloud-init/) for values greater or equal than 64 GB.

Hi @aciba90! That was my first instinct as well, though I was concerned that no matter which limit I chose, there could always be a system (in the future) that would exceed any limit I would set. This led me to look at the conditional below - I couldn't see a case where we would want to override `size` if it had been set already, which led me to believe it was a logic error for this case.

Given the above, would you still prefer that I add a 1TB formula? Also, `x / 2` looks like a bit too much - using the Ubuntu FAQ as a reference, a 1TB system without hibernation would need around 32GB of swap.

> Another good thing would be to cover this bug with unit test under `tests/unittest/config/test_cc_mounts.py`.

I looked through `test_cc_mounts.py` prior to submitting the PR and it looks like `suggested_swapsize` has never been tested. I'm happy to add a test case, but I'm not sure I could do it justice especially given that this was a very minor change. Is there a better middle ground between a full unit test for the method versus none at all?

aciba90:
Thanks, @sstallion, for the detailed explanation!

> Hi @aciba90! That was my first instinct as well, though I was concerned that no matter which limit I chose, there could always be a system (in the future) that would exceed any limit I would set. This led me to look at the conditional below - I couldn't see a case where we would want to override `size` if it had been set already, which led me to believe it was a logic error for this exact condition.
> 
> Given the above, would you still prefer that I add a 1TB formula? Also, `x / 2` looks like a bit too much - using the Ubuntu FAQ as a reference, a 1TB system without hibernation would need around 32GB of swap.

I overlook that. Taken a deeper look, yes, it seems we had a bug in that line. The original intention was probably to suggest `maxsize` if `memsize` is bigger than 64 GB.

> 
> > Another good thing would be to cover this bug with unit test under `tests/unittest/config/test_cc_mounts.py`.
> 
> I looked through `test_cc_mounts.py` prior to submitting the PR and it looks like `suggested_swapsize` has never been tested. I'm happy to add a test case, but I'm not sure I could do it justice especially given that this was a very minor change. Is there a better middle ground between a full unit test for the method versus none at all?

I think it suffices adding a test just covering the error. Something on this lines:

```diff
diff --git a/tests/unittests/config/test_cc_mounts.py b/tests/unittests/config/test_cc_mounts.py
index 8ae28099..7ebdf427 100644
--- a/tests/unittests/config/test_cc_mounts.py
+++ b/tests/unittests/config/test_cc_mounts.py
@@ -7,7 +7,7 @@ from unittest import mock
 import pytest
 
 from cloudinit.config import cc_mounts
-from cloudinit.config.cc_mounts import create_swapfile
+from cloudinit.config.cc_mounts import create_swapfile, suggested_swapsize
 from cloudinit.config.schema import (
     SchemaValidationError,
     get_schema,
@@ -525,6 +525,15 @@ class TestCreateSwapfile:
         assert msg in caplog.text
 
 
+class TestSuggestedSwapsize:
+    def test_maxsize_bigger_than_formulas(self):
+        maxsize = 17_179_869_184  # 16 GB
+        assert maxsize == suggested_swapsize(
+            memsize=68_719_476_736 + 1,  # a bit bigger than 64 GB
+            maxsize=maxsize,
+        )
+
+
 class TestMountsSchema:
     @pytest.mark.parametrize(
         "config, error_msg",
```

Prior to your fix, `suggested_swapsize` returned `None`, which caused subsequent operations to fail.

sstallion:
> Thanks, @sstallion, for the detailed explanation!

My pleasure!

> I think it suffices adding a test just covering the error. Something on this lines:
> ...
> Prior to your fix, `suggested_swapsize` returned `None`, which caused subsequent operations to fail.

Sounds good! Thanks for providing an example to work off of! I should have some time later today to add a test case and add it to the PR.

sstallion:
> I took this even a little further 😄
> 
> I think it'd be good to also update the algorithm to account for larger memory sizes, otherwise it could use 90% of filesystem space for swap...which...seems excessive. If you don't mind applying this patch, I added that as well as a few more unit tests around that particular algorithm. It doesn't fully test what that function is doing, but it would be an improvement over what we have.

Will do! Out of curiosity, is there any interest in syncing these formulas up to the [SwapFaq](https://help.ubuntu.com/community/SwapFaq)? The reasoning in that document seem sound enough to me and I'm somewhat doubtful that cloud-init would be used on machines intended to hibernate.

sstallion:
@TheRealFalcon I believe this simplifies the function a bit further, what do you think?

```
diff --git a/cloudinit/config/cc_mounts.py b/cloudinit/config/cc_mounts.py
index d8b7101a..e5dfa7e4 100644
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -9,6 +9,7 @@
 """Mounts: Configure mount points and swap files"""
 
 import logging
+import math
 import os
 import re
 from string import whitespace
@@ -110,6 +111,7 @@ NETWORK_NAME_RE = re.compile(NETWORK_NAME_FILTER)
 WS = re.compile("[%s]+" % (whitespace))
 FSTAB_PATH = "/etc/fstab"
 MNT_COMMENT = "comment=cloudconfig"
+GB = 2**30
 
@@ -210,8 +212,7 @@ def suggested_swapsize(memsize=None, maxsize=None, fsys=None):
     if memsize is None:
         memsize = util.read_meminfo()["total"]
 
-    GB = 2**30
-    sugg_max = 8 * GB
+    sugg_max = memsize * 2
 
     info = {"avail": "na", "max_in": maxsize, "mem": memsize}
 
@@ -233,34 +234,7 @@ def suggested_swapsize(memsize=None, maxsize=None, fsys=None):
         maxsize = sugg_max
 
     info["max"] = maxsize
-
-    formulas = [
-        # < 1G: swap = double memory
-        (1 * GB, lambda x: x * 2),
-        # < 2G: swap = 2G
-        (2 * GB, lambda x: 2 * GB),
-        # < 4G: swap = memory
-        (4 * GB, lambda x: x),
-        # < 16G: 4G
-        (16 * GB, lambda x: 4 * GB),
-        # < 64G: 1/2 M up to max
-        (64 * GB, lambda x: x / 2),
-    ]
-
-    size = None
-    for top, func in formulas:
-        if memsize <= top:
-            size = min(func(memsize), maxsize)
-            # if less than 1/2 memory and not much, return 0
-            if size < (memsize / 2) and size < 4 * GB:
-                size = 0
-                break
-            break
-
-    if size is None:
-        size = maxsize
-
-    info["size"] = size
+    info["size"] = round(math.sqrt(memsize))
 
     MB = 2**20
     pinfo = {}
```

TheRealFalcon:
@sstallion , yes I think it is reasonable to use the updated chart. The (until now broken) formula in cloud-init is 8 years old with no context or reference as to where they came from, so I think it makes sense to update them. However, cloud instances can be hibernated, so I think we should stick with the "with hibernation" defaults. How about we stick with values from the table until we hit 16GB, then go with the sqrt approach you suggest for anything greater than 16?

sstallion:
@TheRealFalcon I just pushed a set of changes that strictly conforms to the sqrt approach before I saw your message. I can definitely increase the amount of swap for 4GB and below as suggested, but how many hosts are being provisioned these day given how much RAM is generally available now? Is it worth making the change as-is for clarity?

sstallion:
@TheRealFalcon Another alternative could be to add a `hibernate` key to the mounts schema to indicate which set of values to use - that might be a little cleaner than trying to guess for some memory ranges.

sstallion:
> @TheRealFalcon Another alternative could be to add a `hibernate` key to the mounts schema to indicate which set of values to use - that might be a little cleaner than trying to guess for some memory ranges.

To wit, if we had a `hibernate` key, this is all that would be needed to bring the function into conformance with the FAQ with the current set of changes:

```diff
diff --git a/cloudinit/config/cc_mounts.py b/cloudinit/config/cc_mounts.py
index 2b1b194a..6a93590e 100644
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -242,6 +242,9 @@ def suggested_swapsize(memsize=None, maxsize=None, fsys=None):
     else:
         minsize = memsize

+    if hibernate:
+        minsize += memsize
+
     size = min(minsize, maxsize)

     info["size"] = size
```

TheRealFalcon:
Personally, I think that having a `hiberate` key is overkill. At this point the user has opted into having a swapfile, and the user declined to manually specify swap size. Another thing to keep in mind is that swap recommendations can vary widely. E.g., Red Hat has a separate list of recommendations here: https://www.redhat.com/sysadmin/cloud-swap

I think if the user cares deeply about their swap numbers, they'll not use auto. If they don't care, they're not going to care much about the `hibernate` key. They just want something that isn't wildly inappropriate for their system.

> but how many hosts are being provisioned these day given how much RAM is generally available now? Is it worth making the change as-is for clarity?

For production works, probably not many, but the cheapest instances still go down to half a GB (not to mention other use cases of cloud-init like VMWare or Openstack), so I think we should still support them as best we can.

sstallion:
> For production works, probably not many, but the cheapest instances still go down to half a GB (not to mention other use cases of cloud-init like VMWare or Openstack), so I think we should still support them as best we can.

Makes sense. With the current set of changes, swap will be set to the amount of memory in the system for 1GB and below. Would you like me to raise that to 4GB?

TheRealFalcon:
@sstallion 2 loose rules I have in my mind is "1:1 for lower memory values" and "Min 4GB for higher memory values". 

What if we did:
* 1:1 RAM up to 4GB
* 4GB for RAM 4GB-16GB
* sqrt(RAM) for > 16GB

TheRealFalcon:
@sstallion One other thing...I noticed you added yourself to the CLA signers file, but have you signed the [actual CLA](https://ubuntu.com/legal/contributors/agreement) too? I'm not seeing your information there. If you did, can you provide the date that you signed it?

sstallion:
I apologize, I mistook the Code of Contact for a CLA agreement. I've submitted the agreement now.

PR #1571, opened 2022-07-08
sources/azure: increase read-timeout to 60 seconds for wireserver

cjp256:
When reporting ready, 5s is not always sufficient for read-timeout.
The wireserver connection is generally killed at 30s.  This change
increases that timeout to 60s to ensure the client doesn't prematurely
kill the connection, while maintaining a reasonably high limit
in case things go really wrong.

Currently http_with_retries worst-case timing is variable:
~0s 404 * 240 attempts + 5s sleep * 240 attempts = ~20 minutes
~5s timeout * 240 attempts + 5s sleep * 240 attempts = ~40 minutes

Increasing the read timeout to 60 would result in a worst-case of:
~60s timeout * 240 attempts + 5s sleep * 240 attempts = ~260 minutes

To reduce this impact, this changes http_with_retries to a time-based
timeout rather than a retry count.

- Increase read-timeout from 5 -> 60 seconds.

- Introduce time-based timeout of 20 minutes.

- Log every failure rather than periodically. With the 5 second sleep,
  we will have a maximum of 240 events recorded which is reasonable
  for this edge case.

- Update relevant tests to pytest and directly import
  time.time() and time.sleep() for testing purposes.

- Catch only UrlError from readurl() and update the tests using
  SentinelException to UrlErrors.

## Proposed Commit Message

```

```

## Checklist:


cjp256:
> LGTM!
> 
> Just a heads up (no need to change anything here), since I've seen you writing more mocks lately that like this:
> 
> ```
> @pytest.fixture
> def mock_readurl():
>     with mock.patch(MOCKPATH + "url_helper.readurl", autospec=True) as m:
>         yield m
> ```
> 
> we now include [pytest-mock](https://pypi.org/project/pytest-mock/) as a test dependency.
> 
> `mocker.patch(...)` works similar to mock.patch, but automatically cleans up at the end of a test and doesn't require a decorator or context-manager.

perfect, thanks!

PR #1572, opened 2022-07-08
UC02.1: subiquity autoinstall directives opaque schema support (SC-1160)

blackboxsw:
This PR is part of a spec to improve subiquity and cloud-init live-installer interaction and align both the [autoinstall config syntax](https://ubuntu.com/server/docs/install/autoinstall-reference) and cloud-init `#cloud-config` schema.

This PR is part of a series of improvements in cloud-init and live-installer interaction:
- [ ] UC01: cloud-init clean to provide a /etc/cloud/clean.d run-parts directory which allows subiquity to provide scripts which can clean initial install config artifacts to aid in golden image creation from server live-installer.
- [x] UC02.1 cloud-init add opaque schema validation for a new top-level autoinstall key in cloud-config user-data
- [ ] UC03: bug fix for cloud-init collect-logs to improve configuration artifact collection when cloud-init is disabled and no /run/cloud-init/* files exist.
- [ ] UC04: live-desktop installer to disable cloud-init by default after installed boot
- [ ] UC05: cloud-init status to provide machine-readable JSON/YAML format support

We want to allow #cloud-config to honor existing top-level opaque `autoinstall` directives in cloud-init's strict schema validation.
We would like to keep that schema opaque for a couple reasons:
 - avoid cloud-init containing out of date static copies of upstream subiquity schema as the installer grows their schema
 - The live-installer (desktop and server) already provide more detailed schema validation for most autoinstall schema keys so cloud-init doesn't need double-duty on that schema

## Proposed Commit Message
```
cc_autoinstall: support ubuntu live-installer autoinstall directives

New top-level "autoinstall" key can be provided in #cloud-config
user-data to support both Ubuntu live-server(subiquity) and
live-desktop(ubuntu-desktop_installer).

When "autoinstall" config directive is present:
  - validate minimal schema is supported (mandatory version key)
  - warn and skip autoinstall configuration if no autoinstaller snap
    is present on the system

The live-desktop and live-server installer both provide deep
validation of autoinstall config schema so cloud-init does not have
to.

Minimally we want to error loudly and set cloud-init status to
'error' state by raising a hard SchemaValidationError when
autoinstall config is inoperable.

This additional config schema validation is necessary because
current schema validation errors are only emitted as warnings which
can be easily disregarded.
```

## Test Steps
```
# docs
PYTHONPATH=. python3 -m cloudinit.cmd.main schema --docs cc_autoinstall

# schema validation
$ cat > invalid_autoinstall.yaml <<EOF
#cloud-config 
autoinstall: {}
EOF
$ PYTHONPATH=. python3 -m cloudinit.cmd.main schema -c invalid_autoinstall.yaml --annotate
#cloud-config
autoinstall:		# E1
 identity: 
  username: yep

# Errors: -------------
# E1: 'version' is a required property


$ cat > valid_autoinstall.yaml <<EOF
#cloud-config
autoinstall:
  version: 1
  identity:
    username: ubuntiu
    password: 7PzNd9ZG.sneicbR7rySz.mvdtOGTFN14TAQi/nxNFi3EiAGup.3HaM/NYTuhA860hLeFVnfx3/8Ra6rLMCd/
EOF

$ PYTHONPATH=. python3 -m cloudinit.cmd.main schema -c valid_autoinstall.yaml --annotate
Valid cloud-config: valid-autoinstall.yaml




# integration test run
make deb
CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_OS_IMAGE=focal CLOUD_INIT_CLOUD_INIT_SOURCE=./cloud-init_all.deb tox -e integration-tests tests/integration_tests/modules/test_ubuntu_autoinstall.py
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_autoinstall.py
+++ b/cloudinit/config/cc_autoinstall.py
@@ -8,6 +8,7 @@ from textwrap import dedent
 from cloudinit import log as logging
 from cloudinit.config.schema import (
     MetaSchema,
+    SchemaProblem,
     SchemaValidationError,
     get_meta_doc,
 )
@@ -115,26 +116,26 @@ def validate_config_schema(cfg):
     autoinstall_cfg = cfg["autoinstall"]
     if not isinstance(autoinstall_cfg, dict):
         raise SchemaValidationError(
-            (
-                (
+            [
+                SchemaProblem(
                     "autoinstall",
                     "Expected dict type but found:"
                     f" {type(autoinstall_cfg).__name__}",
-                ),
-            )
+                )
+            ]
         )
 
     if "version" not in autoinstall_cfg:
         raise SchemaValidationError(
-            (("autoinstall", "Missing required 'version' key"),)
+            [SchemaProblem("autoinstall", "Missing required 'version' key")]
         )
     elif not isinstance(autoinstall_cfg.get("version"), int):
         raise SchemaValidationError(
-            (
-                (
+            [
+                SchemaProblem(
                     "autoinstall.version",
                     f"Expected int type but found:"
                     f" {type(autoinstall_cfg['version']).__name__}",
-                ),
-            )
+                )
+            ]
         )
--- a/cloudinit/config/cc_autoinstall.py
+++ b/cloudinit/config/cc_autoinstall.py
@@ -1,0 +1,140 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+
+"""Autoinstall: Support ubuntu live-server autoinstall syntax."""
+
+import re
+from textwrap import dedent
+
+from cloudinit import log as logging
+from cloudinit.config.schema import (
+    MetaSchema,
+    SchemaValidationError,
+    get_meta_doc,
+)
+from cloudinit.settings import PER_INSTANCE
+from cloudinit.subp import subp
+
+LOG = logging.getLogger(__name__)
+
+distros = ["ubuntu"]
+
+meta: MetaSchema = {
+    "id": "cc_autoinstall",
+    "name": "Autoinstall",
+    "title": "Support Ubuntu live-server install syntax",
+    "description": dedent(
+        """\
+        Ubuntu's autoinstall syntax supports single-system automated installs
+        in either the live-server or live-desktop installers.
+        When "autoinstall" directives are provided in either
+        #cloud-config user-data or ``/etc/cloud/cloud.cfg.d`` validate
+        minimal autoinstall schema adherance and emit a warning if the
+        live-installer is not present.
+
+        The live-installer will use autoinstall directives to seed answers to
+        configuration prompts during system install to allow for a
+        "touchless" Ubuntu system install.
+
+        For more details on Ubuntu's autoinstaller:
+            https://ubuntu.com/server/docs/install/autoinstall
+    """
+    ),
+    "distros": distros,
+    "examples": [
+        dedent(
+            """\
+            # Tell the live-server installer to provide dhcp6 network config
+            # and LVM on a disk matching the serial number prefix CT
+            autoinstall:
+              version: 1
+              network:
+                version: 2
+                ethernets:
+                  enp0s31f6:
+                    dhcp6: yes
+              storage:
+                layout:
+                  name: lvm
+                  match:
+                    serial: CT*
+        """
+        )
+    ],
+    "frequency": PER_INSTANCE,
+}
+
+__doc__ = get_meta_doc(meta)
+
+
+LIVE_INSTALLER_SNAPS = ("subiquity", "ubuntu-desktop-installer")
+
+
+def handle(name, cfg, cloud, log, _args):
+
+    if "autoinstall" not in cfg:
+        LOG.debug(
+            "Skipping module named %s, no 'autoinstall' key in configuration",
+            name,
+        )
+        return
+
+    snap_list, _ = subp(["snap", "list"])
+    installer_present = None
+    for snap_name in LIVE_INSTALLER_SNAPS:
+        if re.search(snap_name, snap_list):
+            installer_present = snap_name
+    if not installer_present:
+        LOG.warning(
+            "Skipping autoinstall module. Expected one of the Ubuntu"
+            " installer snap packages to be present: %s",
+            ", ".join(LIVE_INSTALLER_SNAPS),
+        )
+        return
+    validate_config_schema(cfg["autoinstall"])
+    LOG.debug(
+        "Valid autoinstall schema. Config will be processed by %s",
+        installer_present,
+    )
+
+
+def validate_config_schema(cfg):
+    """Supplemental runtime schema validation for autoinstall yaml.
+
+    Schema validation issues currently result in a warning log currently which
+    can be easily ignored because warnings do not bubble up to cloud-init
+    status output.
+
+    In the case of the live-installer, we want cloud-init to raise an error
+    to set overall cloud-init status to 'error' so it is more discoverable
+    in installer environments.
+
+    # TODO(Drop this validation When cloud-init schema is strict and errors)
+
+    :raise: SchemaValidationError if any known schema values are present.
+    """
+    autoinstall_cfg = cfg["autoinstall"]
+    if not isinstance(autoinstall_cfg, dict):
+        raise SchemaValidationError(
+            (
+                (
+                    "autoinstall",
+                    "Expected dict type but found:"
+                    f" {type(autoinstall_cfg).__name__}",
+                ),
+            )
+        )
+
+    if "version" not in autoinstall_cfg:
+        raise SchemaValidationError(
+            (("autoinstall", "Missing required 'version' key"),)
+        )
+    elif not isinstance(autoinstall_cfg.get("version"), int):
+        raise SchemaValidationError(
+            (
+                (
+                    "autoinstall.version",
+                    f"Expected int type but found:"
+                    f" {type(autoinstall_cfg['version']).__name__}",
+                ),
+            )
+        )
--- a/cloudinit/config/cc_autoinstall.py
+++ b/cloudinit/config/cc_autoinstall.py
@@ -90,7 +90,7 @@ def handle(name, cfg, cloud, log, _args):
             ", ".join(LIVE_INSTALLER_SNAPS),
         )
         return
-    validate_config_schema(cfg["autoinstall"])
+    validate_config_schema(cfg)
     LOG.debug(
         "Valid autoinstall schema. Config will be processed by %s",
         installer_present,
--- a/cloudinit/config/cc_ubuntu_autoinstall.py
+++ b/cloudinit/config/cc_ubuntu_autoinstall.py
@@ -12,7 +12,7 @@ from cloudinit.config.schema import (
     SchemaValidationError,
     get_meta_doc,
 )
-from cloudinit.settings import PER_INSTANCE
+from cloudinit.settings import PER_ONCE
 from cloudinit.subp import subp
 
 LOG = logging.getLogger(__name__)
@@ -61,7 +61,7 @@ meta: MetaSchema = {
         """
         )
     ],
-    "frequency": PER_INSTANCE,
+    "frequency": PER_ONCE,
 }
 
 __doc__ = get_meta_doc(meta)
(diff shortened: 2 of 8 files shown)


blackboxsw:
@dbungert @ogayot just FYI on this PR as work that relates to subiquity

cloudinit/config/cc_autoinstall.py
```diff
@@ -0,0 +1,140 @@
...
+        if re.search(snap_name, snap_list):
+            installer_present = snap_name
+    if not installer_present:
+        LOG.warning(
+            "Skipping autoinstall module. Expected one of the Ubuntu"
+            " installer snap packages to be present: %s",
+            ", ".join(LIVE_INSTALLER_SNAPS),
+        )
+        return
+    validate_config_schema(cfg["autoinstall"])
```
aciba90, line 93: `validate_config_schema` expects the raw config.

```suggestion
    validate_config_schema(cfg)
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -188,6 +188,21 @@
...
+    "cc_autoinstall": {
+      "type": "object",
+      "properties": {
+        "autoinstall": {
+          "description": "Opaque autoinstall schema definition for Ubuntu autoinstall. Full schema processed by live-installer. See: https://ubuntu.com/server/docs/install/autoinstall-reference",
+          "type": "object",
+          "properties": {
+            "version": {
+              "type": "integer"
+            }
```
aciba90, line 200: Nit: I know `"additionalProperties": true` is the default behavior, but do you think would it add value to explicitly signal that the intention of this schema is to contain more props than `version` ?

```suggestion
            },
            "additionalProperties": true
```

aciba90 (changes requested):
Thanks, @blackboxsw, for the PR.

I found a minor error and suggested adding more unit tests.

cloudinit/config/schemas/schema-cloud-config-v1.json
blackboxsw: This is a good suggestion +1.

mwhudson:
I do wonder if we are going this route if in the long run we should move to a model where cloud-init reads this config and puts it somewhere for the installer to find it, rather than subiquity reading the cloud-init config itself and then writing it to disk. That would require careful coordination with changes to the installer and probably couldn't be considered completely done until 22.04 goes EOL... might not be worth it.

I guess we should also think about much validation it makes sense to do here. Depends what we want the experience to be on failure I guess. Happy to let @dbungert and @ogayot figure that out with you I think :)

tests/unittests/config/test_cc_autoinstall.py
```diff
@@ -0,0 +1,151 @@
...
+                {"autoinstall": {"version": 1}},
+                SAMPLE_SNAP_LIST_DESKTOP_INSTALLER,
+                [mock.call(["snap", "list"])],
+                [
+                    "Valid autoinstall schema. Config will be processed by"
+                    " ubuntu-desktop-installer"
+                ],
+                id="valid_autoinstall_schema_sees_desktop_installer",
+            ),
+            pytest.param(
```
aciba90, line 112: Duplicates https://github.com/canonical/cloud-init/pull/1572/files#diff-f64af37497fa03ef1a6145f72804c9017ef0b2a6aa97e5da075c053134e1fe8cR92

aciba90 (changes requested):
Thanks, @blackboxsw, for the improvements.

I have noticed that there is a duplicated test and the tests fail after the inclusion of https://github.com/canonical/cloud-init/pull/1549

Could you please perform a rebase and apply something like:

```diff
diff --git a/cloudinit/config/cc_autoinstall.py b/cloudinit/config/cc_autoinstall.py
index c9e3e087..7c42ab89 100644
--- a/cloudinit/config/cc_autoinstall.py
+++ b/cloudinit/config/cc_autoinstall.py
@@ -8,6 +8,7 @@ from textwrap import dedent
 from cloudinit import log as logging
 from cloudinit.config.schema import (
     MetaSchema,
+    SchemaProblem,
     SchemaValidationError,
     get_meta_doc,
 )
@@ -115,26 +116,26 @@ def validate_config_schema(cfg):
     autoinstall_cfg = cfg["autoinstall"]
     if not isinstance(autoinstall_cfg, dict):
         raise SchemaValidationError(
-            (
-                (
+            [
+                SchemaProblem(
                     "autoinstall",
                     "Expected dict type but found:"
                     f" {type(autoinstall_cfg).__name__}",
-                ),
-            )
+                )
+            ]
         )
 
     if "version" not in autoinstall_cfg:
         raise SchemaValidationError(
-            (("autoinstall", "Missing required 'version' key"),)
+            [SchemaProblem("autoinstall", "Missing required 'version' key")]
         )
     elif not isinstance(autoinstall_cfg.get("version"), int):
         raise SchemaValidationError(
-            (
-                (
+            [
+                SchemaProblem(
                     "autoinstall.version",
                     f"Expected int type but found:"
                     f" {type(autoinstall_cfg['version']).__name__}",
-                ),
-            )
+                )
+            ]
         )
diff --git a/tests/unittests/config/test_cc_autoinstall.py b/tests/unittests/config/test_cc_autoinstall.py
index 7b03df10..19ced74e 100644
--- a/tests/unittests/config/test_cc_autoinstall.py
+++ b/tests/unittests/config/test_cc_autoinstall.py
@@ -109,16 +109,6 @@ class TestHandleAutoinstall:
                 ],
                 id="valid_autoinstall_schema_sees_desktop_installer",
             ),
-            pytest.param(
-                {"autoinstall": {"version": 1}},
-                SAMPLE_SNAP_LIST_SUBIQUITY,
-                [mock.call(["snap", "list"])],
-                [
-                    "Valid autoinstall schema. Config will be processed by"
-                    " subiquity"
-                ],
-                id="valid_autoinstall_schema_sees_subiquity",
-            ),
         ],
     )
     def test_handle_autoinstall_cfg(
```

blackboxsw:
> 

Thanks @mwhudson. I agree long-term we should plan on cloud-init emitting a flat file artifact so that subiquity doesn't have to have more intimate knowledge of cloud-init internals or structured instance data.  On first pass of the related US015 spec, and the initial invocation of this PR, I was thinking cloud-init could read that config and emit an `/autoinstall.yaml` on the filesystem, which is what subiquity ultimately writes anyway. In discussion on that approach, we opted in this iteration to keep it simple and avoid the transition risk you alluded to in your comment because cloud-init may unintentionally overwrite /autoinstall.yaml put into the installer environment by some other means.  Maybe an output of this cycle's work on US015 is to generate a work item/issue or mini-spec that would establish a specific file unique to cloud-init processing of autoinstall that subiquity could source if present `/var/lib/cloud/autoinstall.yaml` or something. This way cloud-init can be sure it doesn't introduce racy interactions with any subiquity install paths.

blackboxsw:
Thanks for the review @aciba90 I rebased and fixed the existing unittests for expected `SchemaValidationError(schema_errors=[SchemaProblem])` param

TheRealFalcon:
`cc_autoinstall` feels a bit generic to me. Would something like `cc_ubuntu_autoinstall` or `cc_subiquity_autoinstall` make more sense?

aciba90 (approved):
Thanks. Apart from @TheRealFalcon's name change comment, LGTM!

blackboxsw:
`cc_ubuntu_autoinstall` FTW since this installer is unlikely to be scoped outside of ubuntu.

aciba90 (changes requested):
Thanks, @blackboxsw, for the renaming.

After it, the module definition under `cloud.cfg.tmpl` must be adapted.
~~I left an additional suggestion about the cloud-config key name.~~

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -214,6 +214,22 @@
       },
       "minProperties": 1
     },
+    "cc_ubuntu_autoinstall": {
+      "type": "object",
+      "properties": {
+        "autoinstall": {
```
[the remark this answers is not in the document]
aciba90: We cannot do this, as `autoinstall` is what we currently accept.

config/cloud.cfg.tmpl
```diff
@@ -85,6 +85,9 @@ cloud_init_modules:
  - seed_random
 {% endif %}
  - bootcmd
+{% if variant in ["ubuntu"] %}
+ - autoinstall
```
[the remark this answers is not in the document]
dbungert: Is it getting confusing that in some places we're using the term `autoinstall` and others `ubuntu_autoinstall`?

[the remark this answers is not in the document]
blackboxsw: @dbungert it was an oversight in my last rename commit per review comments. sorting the naming now all python modules/config names  should be ubuntu_autoinstall.... thought the top-level #cloud-config key will remain "autoinstall:"

cloudinit/config/schemas/schema-cloud-config-v1.json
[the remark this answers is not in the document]
blackboxsw: Correct this schema property needs to remain "autosinstall" even if we name our python confing module `cc_ubuntu_autoinstall`

cloudinit/config/cc_ubuntu_autoinstall.py
```diff
@@ -0,0 +1,141 @@
...
+                    dhcp6: yes
+              storage:
+                layout:
+                  name: lvm
+                  match:
+                    serial: CT*
+        """
+        )
+    ],
+    "frequency": PER_INSTANCE,
```
TheRealFalcon, line 64: `ONCE`?

cloudinit/config/schemas/schema-cloud-config-v1.json
TheRealFalcon: Setting `true` here doesn't negatively affect anything, but I'm not sure that it's supported either. https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties says false or a list of additional keys.

config/cloud.cfg.tmpl
```diff
@@ -85,6 +85,9 @@ cloud_init_modules:
  - seed_random
 {% endif %}
  - bootcmd
+{% if variant in ["ubuntu"] %}
+ - ubuntu_autoinstall
```
TheRealFalcon, line 89: I think this needs to be further down in the boot process. When I tried this, the boot hung for a minute or two, and I eventually got this in the logs:
```
2022-07-15 17:47:47,471 - util.py[DEBUG]: Running module ubuntu_autoinstall (<module 'cloudinit.config.cc_ubuntu_autoinstall' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_ubuntu_autoinstall.py'>) failed
  File "/usr/lib/python3/dist-packages/cloudinit/config/modules.py", line 231, in _run_modules
    ran, _r = cc.run(
  File "/usr/lib/python3/dist-packages/cloudinit/cloud.py", line 67, in run
  File "/usr/lib/python3/dist-packages/cloudinit/helpers.py", line 185, in run
    results = functor(*args)
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ubuntu_autoinstall.py", line 82, in handle
    snap_list, _ = subp(["snap", "list"])
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 335, in subp
    raise ProcessExecutionError(
Command: ['snap', 'list']
Exit code: 1
Reason: -
Stdout: 
Stderr: error: cannot list snaps: cannot communicate with server: Get http://localhost/v2/snaps: dial unix /run/snapd.socket: connect: no such file or directory
```

Maybe go under the snap module instead?

TheRealFalcon (commented):
Left some comments inline but otherwise looks good.

config/cloud.cfg.tmpl
blackboxsw: Ohh right. I had forgotten about that dependency... you are right on a couple counts:
1. and yes there'll be snap seeding races we don't want to cope with too on some cloud images
2. userdata may want to install the subiquity snap to allow autoinstall in some corner cases

cloudinit/config/cc_ubuntu_autoinstall.py
TheRealFalcon: @blackboxsw Did you see my frequency comment here? Should we do ONCE instead of PER_INSTANCE?

TheRealFalcon (approved):
LGTM! I'll let you merge in case you're waiting for feedback from anybody else.

PR #1574, opened 2022-07-11
sources/azure: refactor chassis asset tag handling

cjp256:
In preparation of adding a new tag to support, move the
current DMI chassis asset tag into an enum.  No change
in behavior should be present other than reporting.
- Create ChassisAssetTag enum for containing all Azure
  DMI chassis asset tags and logic to query system for it.
- Add current DMI asset tag to enum as AZURE_CLOUD.
- Reporting: drop event frame and report valid asset tag.
- Update tests for platform viability to pytest.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -12,6 +12,7 @@ import os.path
 import re
 import xml.etree.ElementTree as ET
 from enum import Enum
+from pathlib import Path
 from time import sleep, time
 from typing import Any, Dict, List, Optional
 
@@ -34,6 +35,7 @@ from cloudinit.sources.helpers.azure import (
     DEFAULT_REPORT_FAILURE_USER_VISIBLE_MESSAGE,
     DEFAULT_WIRESERVER_ENDPOINT,
     BrokenAzureDataSource,
+    ChassisAssetTag,
     NonAzureDataSource,
     OvfEnvXml,
     azure_ds_reporter,
@@ -60,8 +62,6 @@ DEFAULT_METADATA = {"instance-id": "iid-AZURE-NODE"}
 # ensures that it gets linked to this path.
 RESOURCE_DISK_PATH = "/dev/disk/cloud/azure_resource"
 DEFAULT_FS = "ext4"
-# DMI chassis-asset-tag is set static for all azure instances
-AZURE_CHASSIS_ASSET_TAG = "7783-7084-3265-9085-8269-3286-77"
 AGENT_SEED_DIR = "/var/lib/waagent"
 DEFAULT_PROVISIONING_ISO_DEV = "/dev/sr0"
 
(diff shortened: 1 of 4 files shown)


anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
```diff
@@ -669,7 +669,12 @@ def crawl_metadata(self):
 
     def _is_platform_viable(self):
         """Check platform environment to report if this datasource may run."""
-        return _is_platform_viable(self.seed_dir)
+        chassis_tag = ChassisAssetTag.query_system()
+        if chassis_tag is not None:
+            return True
+
+        # If no valid chassis tag, check for seeded ovf-env.xml
+        return Path(self.seed_dir, "ovf-env.xml").exists()
```
blackboxsw, line 677: Can we move this logic back out into a separate function outside the class named `def is_platform_viable`?
```suggestion
return is_platform_viable(self.seed_dir)
```

The reasons being :
 1: Eventually I'd like to see a standardized functional functional `is_platform_viable` entry point in each DataSource module  which allows for a targeted environment/artifact check of a running environment without looking at cached content or having to initialize a DataSource instance in order to invoke DataSource._is_platform_viable.
 2. Any datasources which define the `is_platform_viable` function will then make unit testing `ds-identify` behavior simpler to ensure the python datasource discovery code and the shell datasource discovery code align for the same use cases.

This also will help us in the future grow additional cloud-init CLI subcommands like `cloud-init check cloud-id` to check the status of the current running environment to see if current IMDS/DMI/environmental artifacts represent if the viable datasource discovery has changed (on a system that becomes sandboxed).

cloudinit/sources/helpers/azure.py
```diff
@@ -1063,6 +1064,32 @@ class NonAzureDataSource(Exception):
     pass
 
 
+class ChassisAssetTag(enum.Enum):
+    AZURE_CLOUD = "7783-7084-3265-9085-8269-3286-77"
```
blackboxsw, line 1068: Not for this PR, but as you add more enum values, we'll need to also make sure to update [ds-identifiy's is_azure_chassis function](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L1049) when we add new supported chassis values

cloudinit/sources/DataSourceAzure.py
```diff
@@ -2140,24 +2145,6 @@ def maybe_remove_ubuntu_network_config_scripts(paths=None):
                 util.del_file(path)
 
 
-def _is_platform_viable(seed_dir):
-    """Check platform environment to report if this datasource may run."""
-    with events.ReportEventStack(
```
blackboxsw, line 2145: Just for our understanding Is there a motivation in general to reduce event frame logging in Azure during early boot to avoid excessive or noisy eventing? I just want to know if we have general guidance of what type of events Azure would like to see vs not.

blackboxsw (commented):
Looks pretty good. one request to relocate _is_platform_viable out to a separate function, and a question. Otherwise Looks good to me.

cloudinit/sources/helpers/azure.py
cjp256: I was not aware of that, will do, thanks!

cloudinit/sources/DataSourceAzure.py
cjp256: Sounds good, will do!

cjp256: Done! I added an optional seed_path so it should work in both cases.

cloudinit/sources/DataSourceAzure.py
cjp256: Added @azure_ds_telemetry_reporter to is_platform_viable.

cjp256:
All feedback should have been addressed.  If there's anything more, I'll get back to it in ~12 days :D

blackboxsw (approved):
LGTM thanks @cjp256 for the comments and context. I'll perform a couple of integration runs with this locally and expect to land it unchanged.

PR #1577, opened 2022-07-11
Fix expire passwords for hashed passwords

SadeghHayeri:
## Proposed Commit Message

```
summary: fix expiration for hashed passwords

there was a bug in `cc_set_password` that only expires passwords when using a raw password for chpasswd.list. this fixes the issue and runs expiration for both users and also hashed_users
[refrence](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=chpasswd#set-passwords)

```


SadeghHayeri:
@holmanb Done!

holmanb:
> @holmanb Done!

My apologies, I should have mentioned that the file is in alphabetical order, could you please make that change?

SadeghHayeri:
Done.
I also fix the orders.

PR #1578, opened 2022-07-11
tox: Remove entries from default envlist (SC-1176)

TheRealFalcon:
## Proposed Commit Message

```
tox: Remove entries from default envlist

lowest-supported-dev requires running with python 3.6, which few people
will be running locally.

hypothesis-slow is being removed because fuzz testing
shouldn't be a gate for new features.

LP: #1980854
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,6 @@
 [tox]
 envlist =
     py3,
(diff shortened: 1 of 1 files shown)


holmanb (approved):
LGTM

PR #1579, opened 2022-07-12
Add Ansible Config Module

holmanb:
```
New ansible config module

The intended purpose of this commit is to augment cloud-init's
configuration capabilities during boot. This allows users
to run ansible playbooks against the local node as part of
the boot order. Current Ansible workflows on nodes booted
by cloud-init requires waiting for the booted node before
running the playbook against the online node from the external
control node. In the current state this could potentially
be automated to using phone-home or runcmd modules, however
neither of these options provides an obvious solution for
integration and both require an external service to operate.

This module enables users to seamlessly integrate cloud-init
and Ansible auto-configuration during boot using ansible-pull,
a less common mode of operation that differs from the normal
mode of operation by installing Ansible locally and running a
playbook pulled from a VCS repository.

Expected future work in this module includes an option to
configure management nodes.
```
I'm also interested in extending the module, however to get feedback, limit scope, and make this work easier to review, I'm starting with just `ansible-pull`.

### To Test
to test this in a local lxd container image, don't forget to add the module to `/etc/cloud/cloud.cfg` before rerunning cloud-init


cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,116 @@
...
+                cmd,
+                *[
+                    f"--{key}={value}" if value else f"--{key}"
+                    for key, value in cfg.items()
+                    if key != "playbook-name"
+                ],
+                playbook_name,
+            ]
+        )
+        if stderr:
```
holmanb, line 111: These logs would be really helpful for debugging, but maybe `/var/log/cloud-init-output.log` would be a better place for them?

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,108 @@
...
+                cmd,
+                *[
+                    f"--{key}={value}" if value else f"--{key}"
+                    for key, value in cfg.items()
+                ],
+                playbook_name,
+            ]
+        )
+        if stderr:
+            log.warn(f"{stderr}")
```
holmanb, line 104: I think `/var/log/cloud-init-output.log` is the right place to stuff this output?

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+            #cloud-config
+            ansible:
+              install: true
+              pull:
+                url: "https://github.com/holmanb/vmboot.git"
+                playbook-name: ubuntu.yml
+            """
+        ),
+    ],
+    "frequency": PER_INSTANCE,
```
blackboxsw, line 44: Can now add the following to ensure we don't even call this module's handle when no applicable config is provided.
```suggestion
    "activate_by_schema_keys": ["ansible"],
    "frequency": PER_INSTANCE,
```

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+}
+
+__doc__ = get_meta_doc(meta)
+
+
+def handle(name: str, cfg: dict, cloud: Cloud, log: Logger, _):
```
blackboxsw, line 50: Yeah long-term we probably should drop the log param from all handle calls. Wasn't the right way to do this a number of years ago when this was first spec'd

aciba90 (approved):
LGTM! I left a minor doc typo comment.

holmanb:
@paride Pinging you in case you want to take a look. I recall you mentioned experience with Ansible in a previous conversation.

cloudinit/config/cc_ansible.py
blackboxsw: not sure why we'd see stderr here that didn't raise a ProcessExecutionError. If we do have known password/credentials leaks in messages emitted by ansible-pull I agree this should be steered to cloud-init-output.log to avoid the world-readable cloud-init.log if we can think of an easy way to do that.

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+    ansible_cfg: dict = cfg.get("ansible", {})
+    if ansible_cfg:
+        install, ansible_config = get_and_validate_config(ansible_cfg)
+        install_ansible(cloud.distro, install)
+        run_ansible_pull(ansible_config, log)
+
+
+def get_and_validate_config(cfg: dict) -> Tuple[bool, dict]:
+    pull: dict = cfg.get("pull", {})
+    install: bool = cfg.get("install", False)
```
blackboxsw, line 60: Since ansible can be installed from pip or devel channels instead of just the distro deb/rpm, do we want to make this an install_source enum that would allow for devel vs pip vs distro package?  The cc_chef module does something similar though I'm not sure the best way to do it.

holmanb: I've always preferred distro-tested config management solutions for my OS configs, but you're right that others may prefer the PyPI package instead. Probably best to not be opinionated about this, +1.

cloudinit/config/cc_ansible.py
holmanb: > not sure why we'd see stderr here that didn't raise a ProcessExecutionError

+1

> If we do have known password/credentials leaks in messages emitted by ansible-pull 

I don't think ansible typically logs secrets to stdout, my concern is that this will log arbitrary ansible task descriptions which _could_ contain sensitive data. I don't feel strongly about this point, it's more of a hand wavy "lets not surprise users that may expect ansible logs to be root-only readable".

Perhaps a better argument for logging to `/etc/cloud/cloud-init-out.log` is that the output[1] is multiline, which I think fits better in `/etc/cloud/cloud-init-out.log`

[1] output example:
```
Starting Ansible Pull at 2022-07-22 04:09:06
/usr/bin/ansible-pull --url=https://github.com/holmanb/vmboot.git ubuntu.yml
[WARNING]: Could not match supplied host pattern, ignoring: me.lxd
[WARNING]: Could not match supplied host pattern, ignoring: me
localhost | CHANGED => {
    "after": "c5d28021d861cecf76905d1c2bc6ceb366c4ce95",
    "before": null,
    "changed": true
}
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that
the implicit localhost does not match 'all'
[WARNING]: Could not match supplied host pattern, ignoring: me.lxd
[WARNING]: Could not match supplied host pattern, ignoring: me

PLAY [127.0.0.1] ***************************************************************

TASK [Gathering Facts] *********************************************************
ok: [localhost]

TASK [apt : install packages] **************************************************
ok: [localhost]

TASK [apt : install packages] **************************************************
changed: [localhost] => (item=git)

PLAY RECAP *********************************************************************
localhost                  : ok=3    changed=1    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

cloudinit/config/cc_ansible.py
blackboxsw: Yes I think we need a bit of discussion to determine how to best represent how install_type can best imply desired behavior to support 'packages' (distro-based/default) vs pip 
- rename `install` to `install_type`, which could be a reserved string "packages", or something to represent pip-based install per some of the [ansible upstream installation recommendations](https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#selecting-an-ansible-package-and-version-to-install)


Also one thing we might want to account for in this PR is how we intend to handle of BSD as [ansible may require additional setup on BSD nodes](https://docs.ansible.com/ansible/latest/user_guide/intro_bsd.html). This may only apply to control node type deployments and interacting with the managed nodes over SSH, so it might be something we put off until we have ansib;e control node support. 

We can warn if we don't have support for it at the moment, but it looks like there are some BSD-isms that cloud-init should ultimately handle before trying to support

holmanb: > so it might be something we put off until we have ansib;e control node support.

+1 sounds good.

This module should work "as-is" on the BSDs without extra support. I'll add that if required in future ansible-related PRs.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -230,6 +230,92 @@
...
+                },
+                "clean": {
+                  "type": "boolean",
+                  "default": false
+                },
+                "full": {
+                  "type": "boolean",
+                  "default": false
+                },
+                "diff": {
```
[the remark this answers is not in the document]
holmanb: Good catch, thanks!

I'm more in favor of exiting from the module if we can't do what the user asked rather than trying to guess their intentions. `--check --diff` in an automated call[1] might be used for checking for unexpected changes in important files (i.e. `/etc/{group,shadow,passwd}`). A false negative caused by dropping `--diff` in that scenario would be bad.

[1] particularly if they set frequency to PER_ALWAYS or if we add a long poll option

Thoughts?

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+
+def filter_args(cfg: dict) -> dict:
+    """remove boolean false values"""
+    return {key: value for (key, value) in cfg.items() if value is not False}
+
+
+def run_ansible_pull(cfg: dict, log: Logger):
+    cmd = "ansible-pull"
+    check_deps(cmd)
+    playbook_name: str = deepcopy(cfg).pop("playbook-name")
```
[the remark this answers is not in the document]
holmanb: I moved the deepcopy() call into the run_ansible_pull() callsite, which should resolve this issue.

holmanb:
Thanks for the review on this one @blackboxsw.

I think I've addressed everything mentioned this far, I'll leave details some details below:

> Marking needs fixing for:
>    the discussion around what we should do for install_type vs an install bool

I added supported for pip vs distro install options. This slightly changes implementation details for version parsing, installing, and checking if installed. I opted to make use of this shared / split behavior with some OOP (a base class and separate classes for pip vs distro packaging).

The UX is loosely based on chef, but I opted to use 'distro' and `pip' as the 'install-method's. PyPI ships python packages, which would make it confusing to call distro-managed packages 'packages' - that would seem ambiguous.

Let me know your thoughts.

>    avoiding adding the --playbook-name param on the CLI

Fixed. +1 thanks for catching that.

>    what to do with --diff support

Unfortunately I don't see a semantic versioning parser in the stdlib (one in setup-tools???, but I think that would add deps). I wrote something with regex to work around this behavior difference. It probably requires more test coverage if this route looks acceptable. More comments inline.

Overall the pip/distro split caused a lot of refactor since the last review, but the general flow is the same.

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,215 @@
...
+              pull:
+                url: "https://github.com/holmanb/vmboot.git"
+                playbook-name: ubuntu.yml
+            """
+        ),
+        dedent(
+            """\
+            #cloud-config
+            ansible:
+              install-method: pip
```
blackboxsw, line 49: There are two main packages available for different ansible use-cases, `ansible` and a minimalist `ansible-core`. We might want to surface a  `package-name:` config option that is honored for either install-method: distro or pip which defaults to `ansible`, but could be change to be either `ansible-core` or something like ` ansible-core==2.12.3` in the case of pip installs wanting to pin a version.

holmanb: I like it. I'll plumb that through +1

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,217 @@
...
+
+class AnsiblePull:
+    cmd_version: list = []
+    cmd_pull: list = []
+    env = os.environ.copy()
+
+    def get_version(self) -> Optional[Version]:
+        stdout, _ = subp(self.cmd_version, env=self.env)
+        matches = re.search(
+            r"^ansible.*(\d+)\.(\d+).(\d+).*", stdout.splitlines().pop(0)
```
paride, line 78: I see some older but not ancient Ansible versions having four numeric fields (see `pip index versions ansible`), some where packaged in Debian/Ubuntu (see d/changelog). Also [1] says that ansible-core "does not use semantic versioning" and gives 2.11 and 2.12 as example versions, but I don't see any past version without the PATCH field.

The current code should work even with x.y.z.w versions (ignoring the .w field), but maybe it should be made more general, supporting versions in x.y[.z[.w]] format.

https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html

cloudinit/config/schemas/schema-cloud-config-v1.json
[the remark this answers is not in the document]
paride: +1 on staying on the safe side

aciba90 (changes requested):
Thanks, @holmanb, for the improvements. I left a suggestion about improving the validation and the schema definition and minor nit comments.

cloudinit/config/cc_ansible.py
holmanb: Thanks for the feedback.

I've turned this into a class in util that handles between 2 and 4 digits with comparison operator implementation as a namedtuple subclass (and much better test coverage).

See 5a3c873033b60bd458290bdb746929c28620328c.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -230,6 +230,101 @@
...
+          "additionalProperties": false,
+          "properties": {
+            "install-method": {
+              "type": "string",
+              "default": "distro",
+              "enum": [
+                "distro",
+                "pip"
+              ],
+              "description": "The type of installation for ansible. It can be one of the following values:\n\n - ``distro``\n - ``pip``\n - ``none``"
```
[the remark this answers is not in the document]
holmanb: I forgot to remove the null/none option. Since the commands differ between distro and pip installs, I don't think we should allow not installing it (which set of commands would we attempt?). The code currently checks if it is already installed prior to attempting to install, so that if it is installed during a different config module it doesn't repeat the call. Assuming there are no objections to this approach, I'll just update the description.

tests/unittests/config/test_cc_ansible.py
```diff
@@ -0,0 +1,359 @@
...
+                        "pull": {
+                            "url": "gophers://encrypted-gophers/",
+                        },
+                    }
+                },
+                "'playbook-name' is a required property",
+                id="require-url",
+            ),
+        ),
+    )
```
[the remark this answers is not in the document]
holmanb: see comment above

holmanb:
Thanks for the feedback @paride and @aciba90. I think I've addressed the concerns this far.

cloudinit/config/schemas/schema-cloud-config-v1.json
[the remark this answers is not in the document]
aciba90: Sorry, I thought it was a valid input. Then remove it.

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def handle(name: str, cfg: dict, cloud: Cloud, log: Logger, _):
+    ansible_cfg: dict = cfg.get("ansible", {})
+    if ansible_cfg:
+        validate_config(ansible_cfg)
+        install = ansible_cfg["install-method"]
+        pull_cfg = ansible_cfg.get("pull")
+        if pull_cfg:
+            if install == "pip":
```
aciba90, line 135: Nit: I think it would be better to type `ansible` for both branches as following, in order to not tie it to a `AnsiblePull`'s implementation:

```diff
diff --git a/cloudinit/config/cc_ansible.py b/cloudinit/config/cc_ansible.py
index a21a91e8..934c6538 100644
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -132,8 +132,9 @@ def handle(name: str, cfg: dict, cloud: Cloud, log: Logger, _):
         install = ansible_cfg["install-method"]
         pull_cfg = ansible_cfg.get("pull")
         if pull_cfg:
+            ansible: AnsiblePull
             if install == "pip":
-                ansible: AnsiblePull = AnsiblePullPip()
+                ansible = AnsiblePullPip()
             else:
                 ansible = AnsiblePullDistro(cloud.distro)
             ansible.install(ansible_cfg["package-name"])

```

aciba90 (approved):
Thanks, @holmanb, LGTM.

I left an extra nit-suggestion about typing.

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,187 @@
...
+    """remove boolean false values"""
+    return {key: value for (key, value) in cfg.items() if value is not False}
+
+
+def run_ansible_pull(pull: AnsiblePull, cfg: dict, log: Logger):
+    playbook_name: str = cfg.pop("playbook-name")
+
+    v = pull.get_version()
+    if not v:
+        log.warn("Cannot parse ansible version")
```
blackboxsw, line 171: We probably want to avoid using the log param from handle(.., log) in general it a design approach  that is antiquated and should eventually be retired from our `handle()` signatures. I added ticket `SC-1219` to capture that tech debt. Instead we want to use module-level loggers `LOG = logging.getLogger(__name__)` which give us a good prefix for the module generating the logs and avoid passing a log param into every separate function which needs it.

tests/integration_tests/modules/test_ansible.py
```diff
@@ -0,0 +1,113 @@
...
+    # These ensure the repo used for ansible-pull works as expected
+    assert my_client.execute("wget http://0.0.0.0:8000").ok
+    assert my_client.execute("git clone http://0.0.0.0:8000/").ok
+    assert "(dead)" not in my_client.execute(
+        "systemctl status repo_server.service"
+    )
+
+    # Following assertions verify ansible behavior itself
+    assert my_client.execute(["which", "ansible-pull"]).ok
+    verify_clean_log(log)
```
blackboxsw, line 97: QUESTION: do we want to also validate that expected number of playbook operations completed in logs of /var/log/cloud-init-output.log?
```
PLAY RECAP *********************************************************************
localhost                  : ok=3   ...
```
or is there an ansible CLI tool that surfaces audited task status instead?
This item could be disregarded or iterated on in future PRs as a known failure would be caught in verify_clean_logs anyway I suppose

tests/integration_tests/modules/test_ansible.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # These ensure the repo used for ansible-pull works as expected
+    assert my_client.execute("wget http://0.0.0.0:8000").ok
+    assert my_client.execute("git clone http://0.0.0.0:8000/").ok
+    assert "(dead)" not in my_client.execute(
+        "systemctl status repo_server.service"
+    )
+
+    # Following assertions verify ansible behavior itself
+    assert my_client.execute(["which", "ansible-pull"]).ok
```
blackboxsw, line 96: This throws an error on `TestAnsiblePullPip` because which doesn't find the .local path of ansible-pull. I think you can drop this because a lack of ansible-pull will raise an error in cloud-init.log and fail the `verify_clean_logs`

blackboxsw (changes requested):
Thanks for architecting this solution @holmanb. This iteration looks good from scale out perspective and we can separately iterate on other ansible deploy approaches in the future.

Here is a consolidated diff of my inline review comments for use depending on whether you find them applicable/agreeable: Minimally I think we need integration test fixes.
```diff
diff --git a/cloudinit/config/cc_ansible.py b/cloudinit/config/cc_ansible.py
index 934c6538f..d38708d6d 100644
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -4,7 +4,7 @@ import os
 import re
 import sys
 from copy import deepcopy
-from logging import Logger
+import logging
 from textwrap import dedent
 from typing import Optional
 
@@ -15,6 +15,8 @@ from cloudinit.settings import PER_INSTANCE
 from cloudinit.subp import subp, which
 from cloudinit.util import Version, get_cfg_by_path
 
+LOG = logging.getLogger(__name__)
+
 meta: MetaSchema = {
     "id": "cc_ansible",
     "name": "Ansible",
@@ -125,7 +127,7 @@ class AnsiblePullDistro(AnsiblePull):
         return bool(which("ansible"))
 
 
-def handle(name: str, cfg: dict, cloud: Cloud, log: Logger, _):
+def handle(name: str, cfg: dict, cloud: Cloud, _log: logging.Logger, _):
     ansible_cfg: dict = cfg.get("ansible", {})
     if ansible_cfg:
         validate_config(ansible_cfg)
@@ -139,7 +141,7 @@ def handle(name: str, cfg: dict, cloud: Cloud, log: Logger, _):
                 ansible = AnsiblePullDistro(cloud.distro)
             ansible.install(ansible_cfg["package-name"])
             ansible.check_deps()
-            run_ansible_pull(ansible, deepcopy(pull_cfg), log)
+            run_ansible_pull(ansible, deepcopy(pull_cfg))
 
 
 def validate_config(cfg: dict):
@@ -163,12 +165,12 @@ def filter_args(cfg: dict) -> dict:
     return {key: value for (key, value) in cfg.items() if value is not False}
 
 
-def run_ansible_pull(pull: AnsiblePull, cfg: dict, log: Logger):
+def run_ansible_pull(pull: AnsiblePull, cfg: dict):
     playbook_name: str = cfg.pop("playbook-name")
 
     v = pull.get_version()
     if not v:
-        log.warn("Cannot parse ansible version")
+        LOG.warn("Cannot parse ansible version")
     elif v < Version(2, 7, 0):
         # diff was added in commit edaa0b52450ade9b86b5f63097ce18ebb147f46f
         if cfg.get("diff"):
diff --git a/tests/integration_tests/modules/test_ansible.py b/tests/integration_tests/modules/test_ansible.py
index 5241d2385..08c774e92 100644
--- a/tests/integration_tests/modules/test_ansible.py
+++ b/tests/integration_tests/modules/test_ansible.py
@@ -61,8 +61,8 @@ write_files:
 """
 INSTALL_METHOD = """
 ansible:
-  install-method: {}
-  package-name: {}
+  install-method: {method}
+  package-name: {package}
   pull:
     url: "http://0.0.0.0:8000/"
     playbook-name: ubuntu.yml
@@ -93,21 +93,21 @@ def _test_ansible_pull_from_local_server(my_client):
     )
 
     # Following assertions verify ansible behavior itself
-    assert my_client.execute(["which", "ansible-pull"]).ok
     verify_clean_log(log)
-    assert "cc_ansible.py[WARNING]: Error executing" not in log
     assert "SUCCESS: config-ansible ran successfully" in log
 
 
 @pytest.mark.user_data(
-    USER_DATA + INSTALL_METHOD.format("ansible-core", "pip")
+    USER_DATA + INSTALL_METHOD.format(package="ansible-core", method="pip")
 )
 class TestAnsiblePullPip:
     def test_ansible_pull_pip(self, class_client):
         _test_ansible_pull_from_local_server(class_client)
 
 
-@pytest.mark.user_data(USER_DATA + INSTALL_METHOD.format("ansible", "distro"))
+@pytest.mark.user_data(
+    USER_DATA + INSTALL_METHOD.format(package="ansible", method="distro")
+)
 class TestAnsiblePullDistro:
     def test_ansible_pull_distro(self, class_client):
         _test_ansible_pull_from_local_server(class_client)
diff --git a/tests/unittests/config/test_cc_ansible.py b/tests/unittests/config/test_cc_ansible.py
index d40dac662..6f71add3b 100644
--- a/tests/unittests/config/test_cc_ansible.py
+++ b/tests/unittests/config/test_cc_ansible.py
@@ -1,6 +1,5 @@
 import re
 from copy import deepcopy
-from logging import getLogger
 from textwrap import dedent
 from unittest import mock
 from unittest.mock import call
@@ -239,7 +238,7 @@ class TestAnsible:
             cloud = get_cloud(mocked_distro=True)
             print(cfg)
             install = cfg["ansible"]["install-method"]
-            cc_ansible.handle("", cfg, cloud, getLogger(), None)
+            cc_ansible.handle("", cfg, cloud, None, None)
             if install == "distro":
                 cloud.distro.install_packages.assert_called_once()
                 cloud.distro.install_packages.assert_called_with(
@@ -327,7 +326,7 @@ class TestAnsible:
             else cc_ansible.AnsiblePullDistro(get_cloud().distro)
         )
         cc_ansible.run_ansible_pull(
-            ansible_pull, deepcopy(cfg["ansible"]["pull"]), getLogger()
+            ansible_pull, deepcopy(cfg["ansible"]["pull"])
         )
         assert m_subp.call_args[0][0] == expected
 
```

tests/integration_tests/modules/test_ansible.py
holmanb: > or is there an ansible CLI tool that surfaces audited task status instead?

Not that I know of.

> do we want to also validate that expected number of playbook operations completed in logs of /var/log/cloud-init-output.log?

Sounds good to me, I can add something.

holmanb: > validate that expected number of playbook operations completed

@blackboxsw I added `    assert "ok=3" in log`. I'm trying to keep the match resilient to Ansible output format changes by keeping the match minimal. Let me know your thoughts.

cloudinit/config/cc_ansible.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(self):
+        self.cmd_pull = ["ansible-pull"]
+        self.cmd_version = ["ansible-pull", "--version"]
+        self.env["PATH"] = ":".join([self.env["PATH"], "/root/.local/bin/"])
+
+    def install(self, pkg_name: str):
+        """should cloud-init grow an interface for non-distro package
+        managers? this seems reusable
+        """
+        if not self.is_installed():
```
blackboxsw, line 107: @holmanb I could see `AnsiblePullPip.install` growing the smarts to install python3-pip package if not `which["pip3"]`. This can be taken to a later PR if we so desire. Maybe there are complexities to where/how  pip is installed in a system that we don't specifically want to support here?

blackboxsw: Agreed to carry as a todo item in next PR iterations.

tests/integration_tests/modules/test_ansible.py
blackboxsw: we should make this out of cloud-init-output.log instead, the ansible stdout/stderr goes there, not cloud-init.log

blackboxsw (approved):
good work here. LGTM

tests/integration_tests/modules/test_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+@pytest.mark.user_data(
+    USER_DATA + INSTALL_METHOD.format(method="ansible-core", package="pip")
+)
+class TestAnsiblePullPip:
+    def test_ansible_pull_pip(self, class_client):
+        _test_ansible_pull_from_local_server(class_client)
+
+
+@pytest.mark.user_data(
+    USER_DATA + INSTALL_METHOD.format(method="ansible", package="distro")
```
blackboxsw, line 111:
```suggestion
    USER_DATA + INSTALL_METHOD.format(package="ansible", method="distro")
```

tests/integration_tests/modules/test_ansible.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # Following assertions verify ansible behavior itself
+    verify_clean_log(log)
+    output_log = my_client.read_from_file("/var/log/cloud-init-output.log")
+    assert "ok=3" in output_log
+    assert "SUCCESS: config-ansible ran successfully" in log
+
+
+@pytest.mark.user_data(
+    USER_DATA + INSTALL_METHOD.format(method="ansible-core", package="pip")
```
blackboxsw, line 103:
```suggestion
    USER_DATA + INSTALL_METHOD.format(package="ansible-core", method="pip")
```

chifac08:
Fixes Bug LP #[1740937](https://bugs.launchpad.net/cloud-init/+bug/1740937)

PR #1583, opened 2022-07-12
cc_ubuntu_advantage: Implement custom auto-attach behaviors

aciba90:
## Proposed Commit Message

```
cc_ubuntu_advantage: Implement custom auto-attach behaviors

Previously, ua auto-attach was performed in Ubuntu Pro images
transparently for cloud-init. Offering no configurability options
for cloud-init users.

Extend the `ubuntu_advantage` cloud-cfg section to include
`features.disable_auto_attach` and `enable_beta`.

Integrate new UA apis to detect pro instances and perform 
auto-attach if ua config is given.

The client can configure what non-beta and beta services will be enabled.

This config can be provided via:
- user-data allowing customers to configure auto-attach.
- A conf file under /etc/cloud/cloud.cfg.d allowing resellers to create 
ua-preconfigured golden images.

In the case that the ubuntu-advantage-client version is older than
the one containing the required apis, cc_ubuntu_advantage will fallback
to normal attach.
```

## Additional Context

[SC-1145](https://warthogs.atlassian.net/browse/SC-1145)
[SC-1149](https://warthogs.atlassian.net/browse/SC-1149)
https://github.com/canonical/ubuntu-advantage-client/pull/2166
https://github.com/canonical/ubuntu-advantage-client/pull/2171

## Test Steps

```bash
# Run it in gce, azure or ec2 (only clouds providing Pro images)
tox -e integration-test -- tests/integration_tests/modules/test_ubuntu_advantage.py::TestUbuntuAdvantagePro
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -264,21 +264,16 @@ def maybe_install_ua_tools(cloud: Cloud):
         raise
 
 
-def _is_pro(config: dict) -> bool:
+def _is_pro() -> bool:
     try:
         from uaclient.api.u.pro.attach.auto.should_auto_attach.v1 import (
             should_auto_attach,
         )
-        from uaclient.config import UAConfig
     except ImportError as ex:
         LOG.debug("Unable to import `uaclient`: %s", ex)
         return False
-
-    ua_config = UAConfig(
-        **{"cfg": {"ua_config": config}} if config is not None else {}
-    )
     try:
-        result = should_auto_attach(cfg=ua_config)
+        result = should_auto_attach()
     except Exception as ex:
         LOG.debug("Error during `should_auto_attach`: %s", ex)
         LOG.warning(
@@ -292,9 +287,7 @@ def _is_pro(config: dict) -> bool:
 def _attach(ua_section: dict, config: dict):
     token = ua_section.get("token")
     if not token:
-        msg = (
-            "`ubuntu-advantage.token` required in non-Pro Ubuntu" " instances."
-        )
+        msg = "`ubuntu-advantage.token` required in non-Pro Ubuntu instances."
         LOG.error(msg)
         raise RuntimeError(msg)
     configure_ua(
@@ -370,7 +363,7 @@ def handle(
     disable_auto_attach = bool(
         ua_section.get("features", {}).get("disable_auto_attach", False)
     )
-    if not disable_auto_attach and _is_pro(config):
+    if not disable_auto_attach and _is_pro():
         _auto_attach(ua_section, config)
     else:
         _attach(ua_section, config)
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -33,9 +33,11 @@ meta: MetaSchema = {
         'enable' list is not present, any named service will supplement
         contract-default enabled services.
 
-        In Ubuntu Pro instances, `token` will be ignored. `enable` and
-        `enable_beta` will fully determine what services will be enabled,
-        ignoring contract defaults.
+        On Pro instances, when ``ubuntu_advantage`` config is provided to
+        cloud-init, Pro's auto-attach feature will be disabled and cloud-init
+        will perform the Pro auto-attach ignoring the ``token`` key.
+        The ``enable`` and ``enable_beta`` values will strictly determine what
+        services will be enabled, ignoring contract defaults.
 
         Note that when enabling FIPS or FIPS updates you will need to schedule
         a reboot to ensure the machine is running the FIPS-compliant kernel.
@@ -132,6 +134,10 @@ __doc__ = get_meta_doc(meta)
 
 LOG = logging.getLogger(__name__)
 REDACTED = "REDACTED"
+ERROR_MSG_SHOULD_AUTO_ATTACH = (
+    "Unable to determine if this is an Ubuntu Pro instance."
+    " Fallback to normal UA attach."
+)
 
 
 def supplemental_schema_validation(ua_config):
@@ -294,22 +300,20 @@ def maybe_install_ua_tools(cloud: Cloud):
         raise
 
 
-def _is_pro() -> bool:
+def _should_auto_attach() -> bool:
     try:
         from uaclient.api.u.pro.attach.auto.should_auto_attach.v1 import (
             should_auto_attach,
         )
     except ImportError as ex:
         LOG.debug("Unable to import `uaclient`: %s", ex)
+        LOG.warning(ERROR_MSG_SHOULD_AUTO_ATTACH)
         return False
     try:
         result = should_auto_attach()
     except Exception as ex:
         LOG.debug("Error during `should_auto_attach`: %s", ex)
-        LOG.warning(
-            "Unable to determine if this is an Ubuntu Pro instance."
-            " Fallback to normal UA attach."
-        )
+        LOG.warning(ERROR_MSG_SHOULD_AUTO_ATTACH)
         return False
     return result.should_auto_attach
 
@@ -389,7 +393,7 @@ def handle(
     disable_auto_attach = bool(
         ua_section.get("features", {}).get("disable_auto_attach", False)
     )
-    if not disable_auto_attach and _is_pro():
+    if not disable_auto_attach and _should_auto_attach():
         _auto_attach(ua_section)
     elif not ua_section.keys() <= {"features"}:
         _attach(ua_section, config)
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -153,14 +153,6 @@ def validate_schema_features(ua_section: dict):
         LOG.error(msg)
         raise RuntimeError(msg)
 
-    # Validate extra keys under ubuntu_advantage.features
-    extra_keys = set(features.keys()) - {"disable_auto_attach"}
-    if extra_keys:
-        LOG.debug(
-            "Ignoring unknown 'ubuntu_advantage.features': %s",
-            ", ".join(sorted(extra_keys)),
-        )
-
     # Validate ubuntu_advantage.features.disable_auto_attach
     if "disable_auto_attach" not in features:
         return
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -328,7 +328,7 @@ def handle(
 
     # ua-auto-attach.service had noop-ed as ua_section is not empty
     features: dict = ua_section.get("features", {})
-    disable_auto_attach = bool(features.get("disable_auto_attach", True))
+    disable_auto_attach = bool(features.get("disable_auto_attach", False))
     is_pro_cfg = not disable_auto_attach
 
     is_pro = None
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -329,11 +329,11 @@ def _attach(ua_section: dict, config: Optional[dict] = None):
 
 def _auto_attach(ua_section: dict):
     try:
+        from uaclient.api.exceptions import UserFacingError
         from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
             FullAutoAttachOptions,
             full_auto_attach,
         )
-        from uaclient.exceptions import UserFacingError
     except ImportError as ex:
         msg = f"Unable to import `uaclient`: {ex}"
         LOG.error(msg)
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -5,6 +5,7 @@
 import re
 from logging import Logger
 from textwrap import dedent
+from typing import Optional
 from urllib.parse import urlparse
 
 from cloudinit import log as logging
@@ -284,7 +285,7 @@ def _is_pro() -> bool:
     return result.should_auto_attach
 
 
-def _attach(ua_section: dict, config: dict):
+def _attach(ua_section: dict, config: Optional[dict] = None):
     token = ua_section.get("token")
     if not token:
         msg = "`ubuntu-advantage.token` required in non-Pro Ubuntu instances."
@@ -297,9 +298,9 @@ def _attach(ua_section: dict, config: dict):
     )
 
 
-def _auto_attach(ua_section: dict, config: dict):
+def _auto_attach(ua_section: dict, config: Optional[dict] = None):
     try:
-        from uaclient.api.u.pro.attach.auto.full_auto_attach.vi import (
+        from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
             FullAutoAttachOptions,
             full_auto_attach,
         )
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -366,7 +366,7 @@ def handle(
     )
     if not disable_auto_attach and _is_pro():
         _auto_attach(ua_section, config)
-    else:
+    elif not ua_section.keys() <= {"features"}:
         _attach(ua_section, config)
 
 
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -5,6 +5,7 @@
 import re
 from logging import Logger
 from textwrap import dedent
+from typing import Optional
 from urllib.parse import urlparse
 
 from cloudinit import log as logging
@@ -13,6 +14,12 @@ from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema, get_meta_doc
 from cloudinit.settings import PER_INSTANCE
+
+# TODO move following import to a non-global scope
+from uaclient.api.api import call_api
+from uaclient.config import UAConfig
+from uaclient import AutoAttachWithShortRetryOptions
+from uaclient.api.u.pro.detect.should_auto_attach.v1 import should_auto_attach_on_machine
 
 UA_URL = "https://ubuntu.com/advantage"
 
@@ -327,30 +334,45 @@ def handle(
     maybe_install_ua_tools(cloud)
 
     # ua-auto-attach.service had noop-ed as ua_section is not empty
-    features: dict = ua_section.get("features", {})
-    disable_auto_attach = bool(features.get("disable_auto_attach", False))
+    disable_auto_attach = bool(
+        ua_section.get("features", {}).get("disable_auto_attach", False)
+    )
     is_pro_cfg = not disable_auto_attach
 
-    is_pro = None
-    if is_pro_cfg:
-        try:
-            auto_attach_short(features)
-        except NotAProInstance:
-            is_pro = False
-        except RetryError:
-            is_pro = True
-        else:
-            return  # Successful Pro auto attach
-        if is_pro:
-            try:
-                auto_attach_long(features)
-            except Exception as ex:
-                msg = f"Error setting up long auto-attach: \n{ex}"
-                LOG.error(msg)
-                raise RuntimeError(msg) from ex
-            return  # Hand over to long-retry auto attach
+    ua_config: Optional[UAConfig] = None
+    if config is not None:
+        ua_config = UAConfig(cfg={"ua_config": config})
 
-    if not is_pro:
+    is_pro = should_auto_attach_on_machine(cfg=ua_config).should_auto_attach
+    if is_pro_cfg and is_pro:
+        short_retry_kwargs = {}
+
+        enable = ua_section.get("enable")
+        if enable is not None:
+            short_retry_kwargs["enable"] = enable
+
+        enable_beta = ua_section.get("enable_beta")
+        if enable_beta is not None:
+            short_retry_kwargs["enable_beta"] = enable_beta
+
+        options = AutoAttachWithShortRetryOptions(**short_retry_kwargs)
+
+        try:
+            successful = call_api(
+                "u.pro.auto_attach.with_short_retry.v1",
+                options=options,
+                cfg=ua_config,
+            ).successful
+        except Exception:
+            pass  # TODO log exceptions
+        else:
+            if not successful:
+                LOG.warning(
+                    "Ubuntu Advantage will try to auto-attach as Pro instance"
+                    " with a long retry strategy."
+                    " (This could take days or more)"
+                )
+    else:
         # Fallback to normal attach
         token = ua_section.get("token")
         if not token:
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -16,7 +16,6 @@ from cloudinit.config.schema import MetaSchema, get_meta_doc
 from cloudinit.settings import PER_INSTANCE
 
 UA_URL = "https://ubuntu.com/advantage"
-_AUTO_ATTACH_RETRIES = 3
 
 distros = ["ubuntu"]
 
@@ -328,29 +327,25 @@ def _attach(ua_section: dict, config: Optional[dict] = None):
     )
 
 
-def _auto_attach(ua_section: dict, config: Optional[dict] = None):
+def _auto_attach(ua_section: dict):
     try:
         from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
             FullAutoAttachOptions,
             full_auto_attach,
         )
-        from uaclient.config import UAConfig
+        from uaclient.exceptions import UserFacingError
     except ImportError as ex:
         msg = f"Unable to import `uaclient`: {ex}"
         LOG.error(msg)
         raise RuntimeError(msg) from ex
 
-    ua_config = UAConfig(
-        **{"cfg": {"ua_config": config}} if config is not None else {}
-    )
     options = FullAutoAttachOptions(
         enable=ua_section.get("enable"),
         enable_beta=ua_section.get("enable_beta"),
-        retries=_AUTO_ATTACH_RETRIES,
     )
     try:
-        full_auto_attach(options=options, cfg=ua_config)
-    except Exception as ex:
+        full_auto_attach(options=options)
+    except UserFacingError as ex:
         msg = f"Error during `full_auto_attach`: {ex}"
         LOG.error(msg)
         raise RuntimeError(msg) from ex
@@ -395,7 +390,7 @@ def handle(
         ua_section.get("features", {}).get("disable_auto_attach", False)
     )
     if not disable_auto_attach and _is_pro():
-        _auto_attach(ua_section, config)
+        _auto_attach(ua_section)
     elif not ua_section.keys() <= {"features"}:
         _attach(ua_section, config)
 
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -302,6 +302,7 @@ def maybe_install_ua_tools(cloud: Cloud):
 
 def _should_auto_attach() -> bool:
     try:
+        from uaclient.api.exceptions import UserFacingError
         from uaclient.api.u.pro.attach.auto.should_auto_attach.v1 import (
             should_auto_attach,
         )
@@ -311,7 +312,7 @@ def _should_auto_attach() -> bool:
         return False
     try:
         result = should_auto_attach()
-    except Exception as ex:
+    except UserFacingError as ex:
         LOG.debug("Error during `should_auto_attach`: %s", ex)
         LOG.warning(ERROR_MSG_SHOULD_AUTO_ATTACH)
         return False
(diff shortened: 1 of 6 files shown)


cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -273,16 +304,50 @@ def handle(name, cfg, cloud, log, args):
...
+    is_pro_cfg = not disable_auto_attach
+
+    is_pro = None
+    if is_pro_cfg:
+        try:
+            auto_attach_short(features)
+        except NotAProInstance:
+            is_pro = False
+        except RetryError:
+            is_pro = True
```
blackboxsw, line 324: If http_proxy is setup invalid on a system: is it possible we would hit a RetryError even if we are on a non-pro environment (because we never got ahold of the contracts.canonical.com to confirm whether PRO or not)?

If this is a possibility, I wonder if we really should be assuming `is_pro = True` just on a retry failure.

blackboxsw (commented):
Thanks for pulling a sample PR together for discussion @aciba90 I think we need a bit of UA-dev team input on these new API functions/signature and behavior to make progress on this.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -246,7 +248,36 @@ def maybe_install_ua_tools(cloud):
...
-def handle(name, cfg, cloud, log, args):
+class NotAProInstance(Exception):
+    pass
+
+
+class RetryError(Exception):
+    pass
+
+
+def auto_attach_short(*args, **kwargs):
```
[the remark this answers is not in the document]
CalvoM: @orndorffgrant Will have most of the implementation details on this following the spec US011 he is working on.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -246,7 +248,36 @@ def maybe_install_ua_tools(cloud):
         raise
 
 
-def handle(name, cfg, cloud, log, args):
+class NotAProInstance(Exception):
```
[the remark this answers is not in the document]
aciba90: We agreed to use the `uaclient.api.u.pro.detect.should_auto_attach.v1 import should_auto_attach_on_machine` new API function to check if we are in a Pro instance.

Should we also catch `NonAutoAttachImageError` in the call to `with_short_retry`?

@CalvoM

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -246,7 +248,36 @@ def maybe_install_ua_tools(cloud):
...
+    pass
+
+
+class RetryError(Exception):
```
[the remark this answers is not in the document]
aciba90: Exactly, I have provisionally integrated the new functions in https://github.com/canonical/cloud-init/pull/1583/commits/6154d4d6a3ebe680f5111d69ea3a7a77598b8077

cloudinit/config/cc_ubuntu_advantage.py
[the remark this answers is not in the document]
aciba90: Updated in https://github.com/canonical/cloud-init/pull/1583/commits/6154d4d6a3ebe680f5111d69ea3a7a77598b8077

cloudinit/config/cc_ubuntu_advantage.py
aciba90: After https://github.com/canonical/cloud-init/pull/1583/commits/6154d4d6a3ebe680f5111d69ea3a7a77598b8077, we are first check if we are in a Pro instance and if so we pass the http proxy config in https://github.com/canonical/cloud-init/pull/1583/files#diff-e12ed348cca4b27232907df295029b93b250cec08d438f0ab37e498a6df3a263R318

I guess `with_short_retry` will handle those possible errors, and we should just log any error that they raise.

Or do you, @CalvoM, think we should extra error handling?

aciba90:
This PR would benefit from the inclusion of https://github.com/canonical/cloud-init/pull/1604 in order to easily allocate Ubuntu Pro instances for integration-testing.

cloudinit/config/cc_ubuntu_advantage.py
CalvoM: There is more changes we are making to the function for short retrial. I will fully update you, after some changes. e.g if we need to check if we are on a pro instance or we just use `should_auto_attach` function.

cloudinit/config/cc_ubuntu_advantage.py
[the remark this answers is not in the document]
CalvoM: Yes you could catch it to be thorough, but since `should_auto_attach` checks for it as well, it might not be necessary.
`should_auto_attach` endpoint performs an extra check which is checking for the installation of `ua pro` which cannot be done in `full_auto_attach` (updated name for `with_short_retry`). Furthermore, we do the checking in `full_auto_attach` as side effect of trying to get the cloud instance which we use later on in the function.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -247,7 +281,68 @@ def maybe_install_ua_tools(cloud):
...
+        **{"cfg": {"ua_config": config}} if config is not None else {}
+    )
+    options = FullAutoAttachOptions(
+        enable=ua_section.get("enable"),
+        enable_beta=ua_section.get("enable_beta"),
+        retries=_AUTO_ATTACH_RETRIES,
+    )
+    try:
+        full_auto_attach(options=options, cfg=ua_config)
+    except Exception as ex:
```
CalvoM, line 339: @aciba90 I would suggest that you use exceptions from [the api exceptions file](https://github.com/canonical/ubuntu-advantage-client/blob/main/uaclient/api/exceptions.py). We are doing this to control the documented exceptions raised by the api endpoints. Incase you run to any exception that is not covered in the file, please inform me so that I can update the exceptions file.

aciba90: Done. As we do not perform any specific action depending on a concrete exception, I am catching any `UserFacingError`, which should be the parent of any exception raised here, right?

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2199,21 +2199,37 @@
...
+                  "type": "boolean",
+                  "description": "Optional boolean for controlling if auto attach (in Ubuntu Pro instances) will be performed. Default: ``false``",
+                  "default": false
+                }
+              }
             },
             "config": {
               "type": "object",
-              "description": "Configuration settings or override Ubuntu Advantage config",
+              "description": "Configuration settings or override Ubuntu Advantage config. Only used in non Ubuntu Pro instances.",
```
aciba90, line 2232: During the implementation of https://github.com/canonical/ubuntu-advantage-client/pull/2171, it was possible to pass an `UAConfig` instance (containing this config) to [full_auto_attach](https://github.com/canonical/ubuntu-advantage-client/pull/2171/files#diff-4fa660a53fa2584a90fbec8daf677a41bea763aed263b1cf98390656405c850eR65). But this did not get to the merged commit. Is there an alternative way to pass the config? Or is it not necessary to do so in Pro instances? @CalvoM

aciba90:
I have integrated the future UA API to execute `auto-attach` programmatically as we agreed. The implementation is backwards compatible if `ubuntu-advantage-client`'s version is < `v28.0`, that means we could merge this PR even without the UA release.

I have executed the following test:

```sh
tox -e integration-test -- tests/integration_tests/modules/test_ubuntu_advantage.py::TestUbuntuAdvantagePro
```

and verified that `full_auto_attach` is called correctly from `cloud-init`.

This PR is ready to be reviewed. @blackboxsw, @CalvoM.

cloudinit/config/cc_ubuntu_advantage.py
CalvoM: Yes, that is correct.

cloudinit/config/schemas/schema-cloud-config-v1.json
CalvoM: We have a way of passing the UAConfig to the endpoint via [_full_auto_attach](https://github.com/canonical/ubuntu-advantage-client/blob/e2e9fe94d619f2522ee4b18d411d8efe52fc4844/uaclient/api/u/pro/attach/auto/full_auto_attach/v1.py#L66). However, we recommend that you use the endpoint not requiring the UAConfig parameter, since we create the object internally. You can use the other endpoint, when you have some custom configuration for UAConfig default object.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -88,6 +95,32 @@
...
+            """\
+        # Do not enable any service in Ubuntu Pro instances.
+        ubuntu_advantage:
+          enable: []
+          enable_beta: []
+        """
+        ),
+        dedent(
+            """\
+        # Enable esm and beta realtime-kernel services in Ubuntu Pro instances.
```
CalvoM, line 108: Is it possible to inform the users, that including beta services in the `enable` list will cause an error?

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -247,7 +280,64 @@ def maybe_install_ua_tools(cloud):
...
+    )
+
+
+def _auto_attach(ua_section: dict):
+    try:
+        from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
+            FullAutoAttachOptions,
+            full_auto_attach,
+        )
+        from uaclient.exceptions import UserFacingError
```
CalvoM, line 322: Just a recommendation, you could import it from `uaclient.api.exceptions`; just as stated before, it is so that we maintain a list of exceptions for the api endpoints.

aciba90:
This PR, if accepted, could be merged before https://github.com/canonical/cloud-init/pull/1583 and I will solve possible conflicts there.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2199,21 +2199,37 @@
...
+              "description": "Contract token obtained from https://ubuntu.com/advantage to attach (in non-Pro instances)."
+            },
+            "features": {
+              "type": "object",
+              "description": "Ubuntu Advantage features.",
+              "additionalProperties": false,
+              "properties": {
+                "disable_auto_attach": {
+                  "type": "boolean",
+                  "description": "Optional boolean for controlling if auto attach (in Ubuntu Pro instances) will be performed. Default: ``false``",
```
blackboxsw, line 2225:
```suggestion
                  "description": "Optional boolean for controlling if ua-auto-attach.service (in Ubuntu Pro instances) will be performed during boot. Default: ``false``",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2199,21 +2199,37 @@
...
+            },
+            "enable_beta": {
+              "type": "array",
+              "items": {"type": "string"},
+              "description": "Optional list of ubuntu-advantage beta services to enable. By default, a given contract token will automatically enable a number of services, use this list to supplement which services should additionally be enabled. Any service unavailable on a given Ubuntu release or unentitled in a given contract will remain disabled. In Ubuntu Pro instances, if this list is given, then only those services will be enabled, ignoring contract defaults."
             },
             "token": {
               "type": "string",
-              "description": "Required contract token obtained from https://ubuntu.com/advantage to attach."
+              "description": "Contract token obtained from https://ubuntu.com/advantage to attach (in non-Pro instances)."
```
blackboxsw, line 2216:
```suggestion
              "description": "Contract token obtained from https://ubuntu.com/advantage to attach. Required for non-Pro instances."
```

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -28,6 +31,10 @@
         list is present, any named service will supplement the contract-default
         enabled services.
 
+        In Ubuntu Pro instances, `token` will be ignored. `enable` and
+        `enable_beta` will fully determine what services will be enabled,
```
blackboxsw, line 35: Better readthedocs markdown rendering for static cloud-config keys

We might want to say something about ubuntu-advantage-tool's auto-attach being disabled
if config is provided to cloud-init.
 ```suggestion
        On Ubuntu Pro instances, when ``ubuntu_advantage`` config
        is provided to cloud-init, Ubuntu Advantage Pro's auto-attach feature will
        be disabled and cloud-init will perform the PRO auto-attach, ignoring ``token``.
        The ``enable`` and  ``enable_beta`` values will strictly determine what
        services will be enabled,
```

File: cloudinit/config/cc_ubuntu_advantage.py [hunk 12]
```diff
@@ -247,7 +280,64 @@ def maybe_install_ua_tools(cloud):
         raise
 
 
-def handle(name, cfg, cloud, log, args):
+def _is_pro() -> bool:
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -247,7 +280,64 @@ def maybe_install_ua_tools(cloud):
...
+    try:
+        from uaclient.api.u.pro.attach.auto.should_auto_attach.v1 import (
+            should_auto_attach,
+        )
+    except ImportError as ex:
+        LOG.debug("Unable to import `uaclient`: %s", ex)
+        return False
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/schemas/schema-cloud-config-v1.json
aciba90: Done.

cloudinit/config/schemas/schema-cloud-config-v1.json
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done.

aciba90:
Thanks for the review, @blackboxsw. I have addressed your comments and refactored `_should_auto_attach` to except `UserFacingError` exceptions instead of the generic `Exception`.

This PR is ready to be reviewed after the build passes.

aciba90:
Force-pushing to retrigger CI.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -279,16 +372,19 @@ def handle(
...
-        enable=ua_section.get("enable"),
-        config=config,
+
+    # ua-auto-attach.service had noop-ed as ua_section is not empty
+    disable_auto_attach = bool(
+        ua_section.get("features", {}).get("disable_auto_attach", False)
     )
+    if not disable_auto_attach and _should_auto_attach():
+        _auto_attach(ua_section)
+    elif not ua_section.keys() <= {"features"}:
```
blackboxsw, line 386: Something seems amiss here. Please clarify the logic here for me.

The paths written seem to be:
1. call _auto_attach if  not `ubuntu_advantage: {features: {disable_auto_attach: true}}` AND on PRO image
2. call _attach only if (disable_auto_attach OR not PRO) AND ua_section has top-level keys besides "features"

So, if non-PRO images provide "ubuntu_advantage: features" beyond the `disable_auto_attach` allowed in our schema, they'll get a schema warning `Invalid cloud-config provided: Please run 'sudo cloud-init schema --system'` because of denied features properties.

Did we decide that we will be waiting on UA python api to support for passing any feature flag from cloud-init? 

Maybe cloud-init schema allowing for direct passthrough of any uaclient.conf `features` can wait the presence of UA's python API with support for setting supplemental cloud-init requested features before trying to permit that in `#cloud-config` schema.

I think for customers that want to set random UA client features today via `#cloud-config` user-data, can alternatively provide a `write_files` section like the following and cloud-init schema may not necessarily have to define and support to passthrough any uaclient.conf features flags via cloud-init's schema:
```
# Set UA feature flag for UA client before attach or auto-attach is performed by cloud-init
#cloud-config  
write_files:
- path: /etc/ubuntu-advantage/uaclient.conf
  append: true
  content: |
    features:
      some_feature_name: true
```


Maybe it's worth documenting this type of write_files example in our cc_ubuntu_advantage docs to give folks an example of trying to augment feature flags in uaclient.conf

@CalvoM do we have any backlog items in ua-client API that would eventually surface the ability to set `features` overrides for uaclient via cloud-init or are we always expecting that any behavior overrides are provided directly to the UA api operation like auto_attach_full vs short. Right now only `enable` and `enable_beta` options.

@aciba90 what do you think of in the short-term just documenting the above example in cc_ubuntu_advantage.meta.examples just so folks looking to further customize UA behavior have a starting point to work with.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -279,16 +372,19 @@ def handle(
...
         supplemental_schema_validation(config)
 
     maybe_install_ua_tools(cloud)
-    configure_ua(
-        token=ua_section.get("token"),
-        enable=ua_section.get("enable"),
-        config=config,
+
+    # ua-auto-attach.service had noop-ed as ua_section is not empty
+    disable_auto_attach = bool(
```
blackboxsw, line 381: 1. let's be definsive and check initial type of ua_section.get("features") and alert the user if the value is not a dict otherwise we get a nasty traceback if they provided a string or something.
2. Since the only key under ubuntu_advantage.features we honor is "disable_auto_attach" it may be nice to log.debug("Ignoring unknown ubuntu_advantage.features: x, y, z") to give a breadcrumb in logs about what's being disregarded as the schema validation warning is fairly inscrutable. As the schema warning will just be the inscrutable message: `Invalid cloud-config provided: Please run 'sudo cloud-init schema --system'`.

blackboxsw (changes requested):
1. minor comment on your existing PR related to how we redact tokens in our integration logs.
2. some comments on how to better log unexpected types or values for ubuntu_advantage.features values to avoid tracebacks and provide more context on cloud-init's behavior when facing unexpected values

While thinking about redact behavior, I think we have 2 more things I'd like us to address: 
3. Make sure we redact the token value from the logged message `    LOG.debug("Attaching to Ubuntu Advantage. %s", " ".join(attach_cmd))`
4. Make sure our invocation  of `subp.subp also redacts what command it logs by using logstring
   `subp.subp(attach_cmd, logstring=redacted_cmd)`

tests/integration_tests/clouds.py
```diff
@@ -161,9 +161,15 @@ def launch(
             "user_data": user_data,
         }
         launch_kwargs = {**default_launch_kwargs, **launch_kwargs}
+        display_launch_kwargs = deepcopy(launch_kwargs)
+        if display_launch_kwargs.get("user_data") is not None:
+            if "token" in display_launch_kwargs.get("user_data"):
+                display_launch_kwargs["user_data"] = "REDACTED"
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -91,6 +98,32 @@
           - fips
         """
         ),
+        dedent(
+            """\
+        # Do not enable any service in Ubuntu Pro instances.
```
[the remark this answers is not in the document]
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: > Something seems amiss here. Please clarify the logic here for me.
>
> The paths written seem to be:
>
>   1. call _auto_attach if not ubuntu_advantage: {features: {disable_auto_attach: true}} AND on PRO image
>   2. call _attach only if (disable_auto_attach OR not PRO) AND ua_section has top-level keys besides "features"

The first branch is for UA Pro configurations, and the second branch supports the following cases:

1. Attach configurations. i.e. a configuration with `{"token": "<token>", "enable": ["esm"]}`
2. If the configuration only contains the sub-property `features` or it is empty, then do not perform `attach`. It could sound redundant, but a user could want just to disable auto-attach and not perform an attach with the following configuration: `{"features": {"disable_auto_attach": True}}`.

I have written a comment explaining it and renamed plus extended the unit-test [test_cc_ubuntu_advantage.py::TestHandle::test_handle_auto_attach_vs_attach](https://github.com/canonical/cloud-init/pull/1583/files#diff-c588a33656e0018aed4e9aa9b4e90815b6dcd762595ecaf6da132cf9ed681382R681) which cover these cases.

> So, if non-PRO images provide "ubuntu_advantage: features" beyond the disable_auto_attach allowed in our schema, they'll get a schema warning Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' because of denied features properties.

I have defined the `features` object as `additionalProperties: true` to allow that case and we will log a debug message from [cc_ubuntu_advantage::validate_schema_features](https://github.com/canonical/cloud-init/pull/1583/files#diff-e12ed348cca4b27232907df295029b93b250cec08d438f0ab37e498a6df3a263R141) if extra features are given.

> Did we decide that we will be waiting on UA python api to support for passing any feature flag from cloud-init?

I think we concluded that UA features' configuration from cloud-init was out of the scope of this roadmap item.

> @aciba90 what do you think of in the short-term just documenting the above example in cc_ubuntu_advantage.meta.examples just so folks looking to further customize UA behavior have a starting point to work with.

An argument against this was exposed during our [discussions](https://docs.google.com/document/d/1tc3ZJC4wD6iyMWCFXC_rxWzbjrl0mN4J_YbF2iFoiRY/edit?pli=1&disco=AAAAbOEBxfY). Changes to this file could provoke conflicts during the upgrade of `ubuntu-advantage-tools`.

Another argument against it is that those features are not documented by UA. That makes me think whether UA is going to maintain backwards compatibility of them or not.

I think it would not hurt to add the example, but making clear that those features are an advanced topic and that modifying that file could cause friction during system upgrades.

aciba90:
Thanks for the review, @blackboxsw!

> 1. minor comment on your existing PR related to how we redact tokens in our integration logs.
> 
>     2. some comments on how to better log unexpected types or values for ubuntu_advantage.features values to avoid tracebacks and provide more context on cloud-init's behavior when facing unexpected values

I think I have addressed all the comments and left open the decision about the UA features' example.

> While thinking about redact behavior, I think we have 2 more things I'd like us to address: 3. Make sure we redact the token value from the logged message ` LOG.debug("Attaching to Ubuntu Advantage. %s", " ".join(attach_cmd))` 4. Make sure our invocation of `subp.subp also redacts what command it logs by using logstring `subp.subp(attach_cmd, logstring=redacted_cmd)`

Nice idea. I'm going to create a separated PR addressing this, as it is kind of unrelated to this PR.

cloudinit/config/cc_ubuntu_advantage.py
blackboxsw: > I think we concluded that UA features' configuration from cloud-init was out of the scope of this roadmap item.

Great thanks for the clarifications

>  I have defined the features object as additionalProperties: true to allow that case and we will log a debug message from [cc_ubuntu_advantage::validate_schema_features](https://github.com/canonical/cloud-init/pull/1583/files#diff-e12ed348cca4b27232907df295029b93b250cec08d438f0ab37e498a6df3a263R141) if extra features are given.

@aciba90 please set back the  [`additionalProperties: false`](https://github.com/canonical/cloud-init/pull/1583/files#diff-5092062d78cf6dbc9133596b51c11590df10b46d68c7257257486b2911f05b6dR2221) for features and drop the "[Ignoring unknown  'ubuntu_advantage.features loggiing](https://github.com/canonical/cloud-init/pull/1583/files#diff-e12ed348cca4b27232907df295029b93b250cec08d438f0ab37e498a6df3a263R159). The schema error messaging in this case is clear enough and it just doesn't make sense to allow providing unsupported `features` in cloud-config only to log that we are ignoring them.

```bash
$ PYTHONPATH=. python3 -m cloudinit.cmd.main schema --annotate -c ua.yaml

#cloud-config
ubuntu_advantage:
 token: Redacted
 enable_beta: ['str']
 features:		# E1
   a: nope

# Errors: -------------
# E1: Additional properties are not allowed ('a' was unexpected)
```


> An argument against this was exposed during our [discussions](https://docs.google.com/document/d/1tc3ZJC4wD6iyMWCFXC_rxWzbjrl0mN4J_YbF2iFoiRY/edit?pli=1&disco=AAAAbOEBxfY). Changes to this file could provoke conflicts during the upgrade of ubuntu-advantage-tools.

Good point. Ok, best not to publicly document something which could be hard to maintain from cloud-init and or UA in the future. 

> I think it would not hurt to add the example, but making clear that those features are an advanced topic and that modifying that file could cause friction during system upgrades.

Your points carry good weight here. Let's avoid the extra documentation on setting other features with `write_files`. Any customer that needs these features will be told specifically about them and we don't want to get into a support/upgrade debacle here with something that isn't a primary mechanism to configure/run UA.

blackboxsw (changes requested):
@aciba90 2 things left I think:
1. please set back `additionalProperties: false` on `features` schema as it doesn't make sense to allow properties that we just log that we will ignore.
2. Please make sure `make deb` works. I think we may be hitting a pytest exception format error in trying to match handling of "Unable to import `uaclient`: 
```
        >       with pytest.raises(RuntimeError, match=re.escape(expected_msg)):
        E       AssertionError: Regex pattern "Unable\\ to\\ import\\ `uaclient`:\\ No\\ module\\ named\\ 'uaclient'" does not match "Unable to import `uaclient`: No module named 'uaclient.api'".
        
        tests/unittests/config/test_cc_ubuntu_advantage.py:889: AssertionError
```

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done, thanks.

aciba90:
> 1. please set back `additionalProperties: false` on `features` schema as it doesn't make sense to allow properties that we just log that we will ignore.

Done.

> 2. Please make sure `make deb` works. I think we may be hitting a pytest exception format error in trying to match handling of "Unable to import `uaclient`:

Thanks for pointing out this one. The problem was that my way to simulate an import error was not correct. I was removing `uaclient` from `sys.modules` and this was not enough because `_should_auto_attach` and `_auto_attach` perform the imports locally. So when those imports were reached, python tried to import again as it did not find them within `sys.modules`.

The reason only `make deb` detected the issue was that `./packages/bddeb` and `debuild` run with the system python3 interpreter, and `uaclient` is available there.


I have decorated `builtins.__import__` to simulate an import error, avoiding previous difficulties.

This is ready to be reviewed after CI passes, @blackboxsw.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -263,6 +326,75 @@ def maybe_install_ua_tools(cloud):
...
+    if enable_beta:
+        LOG.debug(
+            "Ignoring `ubuntu-advantage.enable_beta` services in UA attach:"
+            " %s",
+            ", ".join(enable_beta),
+        )
+    configure_ua(
+        token=token,
+        enable=ua_section.get("enable"),
+        config=ua_section.get("config"),
```
blackboxsw, line 370: @aciba90 I just realized one big gap in this implementation is that we are only invoking `ua config *http_proxy*` values on non-PRO instances. I think we actually need this feature to override proxy config for both PRO and non-PRO instances, so this may involve a bit of a refactor to ensure both PRO instances and can still manipulate proxy values prior to auto-attach of plain attach.

blackboxsw (commented):
I do think the one final review blocking this PR is that auto_attach (PRO) images still need to be able to specify 'config' settings. Can you please factor that out to a function that both auto_attach and `_attach` can separately call before performing their attach logic? Without this refactor, this represents a regression in current behavior of PRO images which can currently specify `ubuntu_advantage:config` settings to configure *proxy* vars.

If you wish to take that refactor as a separate PR that follows this one I'm fine with that. We just need to make sure we don't release until PRO can support `ubuntu_advantage.config`

aciba90:
> I do think the one final review blocking this PR is that auto_attach (PRO) images still need to be able to specify 'config' settings. Can you please factor that out to a function that both auto_attach and `_attach` can separately call before performing their attach logic? Without this refactor, this represents a regression in current behavior of PRO images which can currently specify `ubuntu_advantage:config` settings to configure _proxy_ vars.
> 
> If you wish to take that refactor as a separate PR that follows this one I'm fine with that. We just need to make sure we don't release until PRO can support `ubuntu_advantage.config`

Per https://github.com/canonical/cloud-init/pull/1583#discussion_r957565379 I did wrongly understand there was not an alternative way to set the UA config up for `full_auto_attach`. Thanks, @blackboxsw, for pointing out that it is possible to set it up via cli.

I am working on it and I have found some bugs related to the config validation and processing, therefore I would suggest including it as a follow-up PR, as this one is getting considerable big.

blackboxsw (approved):
Thanks @aciba90 let's then land this and iterate on the config handling before pushing a release into kinetic.

PR #1585, opened 2022-07-12
Add btrfs and lvm support to lxd storage options (SC-1026)

holmanb:
```
cc_lxd: Add btrfs and lvm lxd storage options

Also add mocked_distro flag to get_cloud() test function
```

### DIFF
--- a/tests/unittests/config/test_cc_lxd.py
+++ b/tests/unittests/config/test_cc_lxd.py
@@ -41,7 +41,7 @@ class TestLxd(t_help.CiTestCase):
         "cloudinit.config.cc_lxd.maybe_cleanup_default", return_value=None
     )
     def test_lxd_init(self, m_maybe_clean, m_which, m_subp):
-        cc = get_cloud()
+        cc = get_cloud(mocked_distro=True)
         m_install = cc.distro.install_packages
 
         for backend, cmd, package in self.backend_def:
--- a/tests/unittests/config/test_cc_lxd.py
+++ b/tests/unittests/config/test_cc_lxd.py
@@ -1,5 +1,6 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 import re
+from copy import deepcopy
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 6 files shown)


TheRealFalcon:
Do you have plans to add ceph too? IIRC, ceph was going to be one of LXDs main storage backends. It doesn't have to be this PR, but I think it should be part of the overall effort here.

blackboxsw:
> Do you have plans to add ceph too? IIRC, ceph was going to be one of LXDs main storage backends. It doesn't have to be this PR, but I think it should be part of the overall effort here.

I think we had a planing discussion with @stgraber about ceph vs btfs and lvm and I thought I recall ceph being deprioritized as part of that discussion.

blackboxsw (commented):
Would we be able to integration tests add storage_backends btfs and lvm?

tests/unittests/config/test_cc_lxd.py
```diff
@@ -27,33 +28,42 @@ class TestLxd(t_help.CiTestCase):
...
-            [
-                mock.call(["lxd", "waitready", "--timeout=300"]),
-                mock.call(
+        for backend in self.backends:
+            with mock.patch("cloudinit.config.cc_lxd.subp") as mock_subp:
+                mock_subp.which.return_value = True
+                lxd_cfg = deepcopy(self.lxd_cfg)
+                lxd_cfg["lxd"]["init"]["storage_backend"] = backend
+                cc_lxd.handle("cc_lxd", lxd_cfg, cc, self.logger, [])
+                self.assertTrue(mock_subp.which.called)
```
blackboxsw, line 48: This assertion should be more specific to the backend or maybe more generally we should have a class TestGetRequiredPackages which asserts the package string returned from get_required_packages and the specific "which" calls when the package is "present" vs absent.

blackboxsw (changes requested):
Thanks @holmanb. I think we need a bit more test coverage. at least unittests covering the logic under `get_required_packages` and the mocked calls to `cloud.distro.install_packages(packages)`

Ideally we'd get an integration test or two around for this, but maybe that needs a separate PR?

stgraber:
> I think we had a planing discussion with @stgraber about ceph vs btfs and lvm and I thought I recall ceph being deprioritized as part of that discussion.

Yeah, basically the conclusion was that dir, btrfs and zfs are pretty common on standalone servers and may make sense to directly support. For other setups, you're likely better off just using a full LXD init preseed.

That's particularly true of Ceph which isn't exactly something trivial to setup on a single server. You're usually dealing with a cluster of server, need keyrings to be issued and deployed, ... Not something that's particularly likely to be done entirely through cloud-init.

holmanb:
> at least unittests covering the logic under get_required_packages and the mocked calls to cloud.distro.install_packages(packages)

After implementing tests for get_required_packages and the mocked calls to `cloud.distro.install_packages()`, I realized that implementing the `install_packages()` tests renders `get_required_packages()` as an implementation detail that is already covered. I'll leave it up to a round 2 review to decide whether or not to drop `get_required_packages()` tests.

holmanb:
Oh that's interesting. `black` is complaining about a file I didn't touch (`cloudinit/sources/helpers/azure.py`), despite `black` passing locally. I'll rebase on main to force a re-run.

blackboxsw:
> Oh that's interesting. `black` is complaining about a file I didn't touch (`cloudinit/sources/helpers/azure.py`), despite `black` passing locally. I'll rebase on main to force a re-run.

I saw this on my local PR too. Somehow we managed to merge a PR that should have failed the black check in github actions. not sure how. I had opted to fix format this locally in my PR. The alternative I was looking at was whether tox could support a subshell call like the following for our black formatting:

`tox/black/bin/python3 -m black  $(git diff --diff-filter=d --name-only upstream/main | grep py$)`

blackboxsw (approved):
Looks good @holmanb,

Let's just tweak our existing integration test from the default `storage_backend: dir` to storage_backend: btrfs` as it'll also let us assert that the btrfs utils are installed

```diff
diff --git a/tests/integration_tests/modules/test_lxd_bridge.py b/tests/integration_tests/modules/test_lxd_bridge.py
index 3292a833d..735b9cc9a 100644
--- a/tests/integration_tests/modules/test_lxd_bridge.py
+++ b/tests/integration_tests/modules/test_lxd_bridge.py
@@ -12,7 +12,7 @@ USER_DATA = """\
 #cloud-config
bootcmd:
  
 lxd:
   init:
-    storage_backend: dir
+    storage_backend: btrfs
   bridge:
     mode: new
     name: lxdbr0
```

holmanb:
> We could optionally apt remove btrfs-progs in a bootcmd if we wanted to assert it was installed in the process

I took your suggestion on this, thanks!

@blackboxsw If the extra line looks good to you I think I'm ready to merge this.

PR #1587, opened 2022-07-12
freebsd: Don't assume that root user is in root group

holmanb:
```
Don't assume root user is in root group

BSD's root user has group name "wheel" for GID 0

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265177
```

Rather than assuming that "root:root" is default user:group everywhere, move this value into the distro definitions as the default, and override this to "root:wheel" for ~~freebsd~~ BSDs.

I also added some type annotations along the way, which help us catch/prevent bugs thanks to mypy and friends.

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -26,6 +26,7 @@ class Distro(cloudinit.distros.bsd.BSD):
     """
 
     networking_cls = FreeBSDNetworking
+    default_owner = "root:wheel"
     usr_lib_exec = "/usr/local/lib"
     login_conf_fn = "/etc/login.conf"
     login_conf_fn_bak = "/etc/login.conf.orig"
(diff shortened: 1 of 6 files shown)


holmanb:
Ah, it looks like I have freebsd source local from the last time I looked at a freebsd bug. I see:
```
arc~/workspace/freebsd-src(main|✔) grep root etc/group
wheel:*:0:root
operator:*:5:root
```

holmanb:
@igalic do you have time for a quick review? I'd like a review from a freebsd user or developer.

cloudinit/distros/freebsd.py
```diff
@@ -26,6 +26,7 @@ class Distro(cloudinit.distros.bsd.BSD):
     """
 
     networking_cls = FreeBSDNetworking
+    default_owner = "root:wheel"
```
igalic, line 29: this is the case for all BSDs, so this can go into bsd.py

igalic (changes requested):
minor change for a big impact ;)

holmanb:
@igalic Thanks for the review! It should be good now I think.

TheRealFalcon (approved):
LGTM, though I won't merge right now to see if we have anymore comments.

holmanb:
Thanks for the reviews @igalic and @TheRealFalcon! Merging now.

PR #1589, opened 2022-07-13
typing: Use TypedDict backport in python < 3.8

aciba90:
## Proposed Commit Message

```
typing: Use TypedDict backport in python < 3.8
```

## Additional Context

I think it would be nice to make use of `type_extensions` in order to have type-annotation back-ports.

Sadly, in order to use it normally, we need to have it as a dependency, but `python3-typing-extensions` is only available in series >= focal. That's why I imported it only if we are type-checking.

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -8,12 +8,11 @@ import os
 import re
 import sys
 import textwrap
-import typing
 from collections import defaultdict
 from copy import deepcopy
 from functools import partial
 from itertools import chain
-from typing import List, NamedTuple, Optional, Type, Union, cast
+from typing import TYPE_CHECKING, List, NamedTuple, Optional, Type, Union, cast
 
 import yaml
 
@@ -66,12 +65,14 @@ SCHEMA_EXAMPLES_SPACER_TEMPLATE = "\n    # --- Example{0} ---"
 DEPRECATED_KEY = "deprecated"
 
 
-# annotations add value for development, but don't break old versions
-# pyver: 3.6 -> 3.8
-# pylint: disable=E1101
-if sys.version_info >= (3, 8):
+# type-annotate only if type-checking.
+# Consider to add `type_extensions` as a dependency when Bionic is EOL.
+if TYPE_CHECKING:
+    import typing
 
-    class MetaSchema(typing.TypedDict):
+    from typing_extensions import TypedDict
+
+    class MetaSchema(TypedDict):
         name: str
         id: str
         title: str
@@ -82,7 +83,6 @@ if sys.version_info >= (3, 8):
 
 else:
     MetaSchema = dict
-# pylint: enable=E1101
 
 
 class SchemaDeprecationError(ValidationError):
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
While I'm a big fan of the typing_extensions library, I'm a bit wary of pulling it in here. I don't like the idea of relying on dependencies that don't actually make it into the build, as it can become to easy to accidentally include something that works locally but will blow up later. I'm also not sure I want to proliferate this type of code where we have `if TYPE_CHECKING`, do something complicated, else, fallback to simple. I think it could wind up bloating the code and making it harder to follow in some ways.

I could be convinced otherwise though. I think it's worth bringing up to the rest of the team in a meeting.

aciba90:
> While I'm a big fan of the typing_extensions library, I'm a bit wary of pulling it in here. I don't like the idea of relying on dependencies that don't actually make it into the build, as it can become to easy to accidentally include something that works locally but will blow up later. I'm also not sure I want to proliferate this type of code where we have `if TYPE_CHECKING`, do something complicated, else, fallback to simple. I think it could wind up bloating the code and making it harder to follow in some ways.
> 
> I could be convinced otherwise though. I think it's worth bringing up to the rest of the team in a meeting.

I agree, it is not a clean solution. 

In [schema: boot speed skip inapplicable modules](https://warthogs.atlassian.net/browse/SC-1021) we want to introduce a new meta-schema key to define how to skip innaplicable modules. In order to not break backwards compat in cloud-init's downstreams:

1. We can make the new key not required and assign it a default.
2. Type checking is going to be a problem as we type MetaSchema with TypeDict and we cannot define not required properties.

To solve 2, I wanted to introduce [NotRequired](https://docs.python.org/3.11/library/typing.html#typing.NotRequired), which is only available in Python 3.11 or through typing_extensions. For reference: https://github.com/aciba90/cloud-init/blob/332b93905136ce9a3f4766f3ae179c3c26b4ecc4/cloudinit/config/schema.py#L83

But yes, introducing `typing_extension` as a typing-only requirement implies a risk. 
I am going to think about an alternative solution.

holmanb:
> In order to not break backwards compat in cloud-init's downstreams:

Just to make sure I understand correctly: how does this break backwards compatibility? Do we get new mypy warnings somewhere without in your linked commit if the new MetaSchema attribute doesn't use NotRequired?

aciba90:
> > In order to not break backwards compat in cloud-init's downstreams:
> 
> Just to make sure I understand correctly: how does this break backwards compatibility? Do we get new mypy warnings somewhere without in your linked commit if the new MetaSchema attribute doesn't use NotRequired?

Exactly. If we do not have access to `NotRequired`, and we include a new prop under `MetaSchema`

> Just to make sure I understand correctly: how does this break backwards compatibility? Do we get new mypy warnings somewhere without in your linked commit if the new MetaSchema attribute doesn't use NotRequired?

Exactly, answer here: https://github.com/canonical/cloud-init/pull/1597/files#r922082086

aciba90:
Closing this as https://github.com/canonical/cloud-init/pull/1597 has superseded it.

PR #1590, opened 2022-07-13
cc_set_passwords fixes (SC-1183)

TheRealFalcon:
## Proposed Commit Message

```
cc_set_passwords fixes

Includes:
* Fix a case where 'is' instead of '==' resulted in incorrectly set
  passwords
* Fix RANDOM password getting logged as RANDOM rather than the
  password
* A new feature flag to allow disabling the fix in e0416cd6 on older
  releases
* Updated deprecated schema text so docs don't say
  "DEPRECATED. DEPRECATED"
* New integration test for new user schema which is now the CI default
* Unit test conversion from CiTestCase to pytest
* New unit tests covering new schema, password expiration, random
  password creation and logging
```

## Additional Context
For Ubuntu, will require follow-up PRs with quilt patches for Jammy, Focal, and Bionic settings the new feature flag False.


cjp256:
The recent changes affects all Azure users with a password.  SSH commands do not work for the affected user until the password is updated :-(

$ ssh cpatterson@test-ubuntu2204.eastus.cloudapp.azure.com uptime
WARNING: Your password has expired.
Password change required but no TTY available.

cjp256:
Perhaps `expire` can be configured per-user, or maybe don't expire the default user if unset?

cloudinit/features.py
```diff
@@ -49,6 +49,16 @@
...
 """
 
+
+EXPIRE_APPLIES_TO_HASHED_USERS = True
+"""
+If ``EXPIRE_APPLIES_TO_HASHED_USERS`` is True, then when expire is set true
+in cc_set_passwords, hashed passwords will be expired. Previous to 22.3,
+only non-hashed passwords were expired.
+
+(This flag can be removed after Jammy is no longer supported.)
```
cjp256, line 59: Is the intention that this flag will be disabled by distro patch <= 22.04?

TheRealFalcon:
> The recent changes affects all Azure users with a password. SSH commands do not work for the affected user until the password is updated :-(

When you say most recent changes, are you referring to 22.2, or whats in kinetic, or something else? Any changes in this file should ONLY affect users providing configuration to the [cc_set_passwords module](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#set-passwords), so if you're seeing something more widespread than that, it may be a different problem.

cloudinit/features.py
TheRealFalcon: Yes

TheRealFalcon:
Once #1592 merges, then this should be ready to review.

holmanb (approved):
LGTM, thanks @TheRealFalcon!

I'll leave this unmerged until Monday in case @blackboxsw has anything to add.

blackboxsw (approved):
LGTM test improvemts to assert coverage works well and random passwords generated to console work in live manual and automated integration testing

PR #1592, opened 2022-07-14
sources/azure: don't set cfg["password"] for default user pw

cjp256:
The password is still set for the default (admin) user but isn't
immediately expired as a result of this change:
https://github.com/canonical/cloud-init/pull/1577

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1836,7 +1836,7 @@ def read_azure_ovf(contents):
     if ovf_env.password:
         defuser["lock_passwd"] = False
         if DEF_PASSWD_REDACTION != ovf_env.password:
-            defuser["passwd"] = encrypt_pass(ovf_env.password)
+            defuser["hashed_passwd"] = encrypt_pass(ovf_env.password)
 
     if defuser:
         cfg["system_info"] = {"default_user": defuser}
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1836,9 +1836,7 @@ def read_azure_ovf(contents):
     if ovf_env.password:
         defuser["lock_passwd"] = False
         if DEF_PASSWD_REDACTION != ovf_env.password:
-            defuser["passwd"] = cfg["password"] = encrypt_pass(
-                ovf_env.password
-            )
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
@cjp256 I believe if you use `defuser[plain_text_passwd]` rather than `passwd`, it will work if the user already exists as well, but it will not expire it.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1836,9 +1836,7 @@ def read_azure_ovf(contents):
     if ovf_env.password:
         defuser["lock_passwd"] = False
         if DEF_PASSWD_REDACTION != ovf_env.password:
-            defuser["passwd"] = cfg["password"] = encrypt_pass(
-                ovf_env.password
-            )
+            defuser["passwd"] = encrypt_pass(ovf_env.password)
```
blackboxsw, line 1839: @TheRealFalcon so is this really just the suggestion then below?
```suggestion
            defuser["hashed_passwd"] = encrypt_pass(ovf_env.password)
```

TheRealFalcon:
I took the liberty of pushing the suggested change to this branch.

@anhvoms , do you happen to have any context for these changes. Particular [this comment](https://github.com/canonical/cloud-init/pull/1590#issuecomment-1183590745) related to [this PR](https://github.com/canonical/cloud-init/pull/1577)?

We're thinking the changes here should fix the issue mentioned by cjp256. Is there any way you can confirm so we don't continue to break Azure functionality?

anhvoms:
> I took the liberty of pushing the suggested change to this branch.
> 
> @anhvoms , do you happen to have any context for these changes. Particular [this comment](https://github.com/canonical/cloud-init/pull/1590#issuecomment-1183590745) related to [this PR](https://github.com/canonical/cloud-init/pull/1577)?
> 
> We're thinking the changes here should fix the issue mentioned by cjp256. Is there any way you can confirm so we don't continue to break Azure functionality?

I had a discussion last week with Christopher and I am aware of the issue. I will get this tested and update back next week

anhvoms:
@TheRealFalcon @blackboxsw  I have tested this change and it's working as expected for both scenarios (new user with password, existing user with password)

blackboxsw (approved):
Thanks @anhvoms confirmed on our side too this appears to work as expected using hashed_passwd instead of passwd.

PR #1593, opened 2022-07-14
docs: clarify user password purposes

TheRealFalcon:
## Proposed Commit Message

```
docs: clarify user password purposes

Also undeprecate "user" key as we currently have no userdata replacement
```


holmanb (approved):
Good clarification. Thanks!

PR #1598, opened 2022-07-15
Format after rebase

holmanb:
```
ci: rebase prior to format checking
```

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -32,7 +32,7 @@ jobs:
           git fetch github
           git config --global user.name "Gitbot"
           git config --global user.email "dev@null.com"
-          git merge github/main
+          git rebase github/main
 
       - name: Dependencies
         run: |
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -27,7 +27,12 @@ jobs:
       - name: "Checkout #2 (for tools/read-version)"
         run: |
           git fetch --unshallow
-          git remote add upstream https://git.launchpad.net/cloud-init
+          git remote add launchpad https://git.launchpad.net/cloud-init
+          git remote add -t main github https://github.com/canonical/cloud-init.git
+          git fetch github
+          git config --global user.name "Gitbot"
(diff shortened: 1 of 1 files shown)


paride:
If I get it right this is about the issue of `main` moving and CI results for PRs becoming invalid as with what landed in `main` CI wouldn't pass anymore. With the rebase proposed in this branch the idea is to make CI run with the updated `main`, *when CI is manually retriggered*.

I don't think these changes are actually needed: note that on `pull_request` CI already runs on the merge between the proposed branch and the target branch. See `GITHUB_REF` here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request. So manually retriggering CI *may* be sufficient. Honestly I don't think it's going to work very well.

A better solution is https://github.com/bors-ng/bors-ng, which I already mentioned a few times in the past. One of the reasons why I didn't try to setup bors yet is that GitHub is rolling out a native (and generally better) solution for this issue:

https://github.blog/changelog/2021-10-27-pull-request-merge-queue-limited-beta/

but it's still in limited beta at the moment.

I hope I've got the context right :-)

holmanb:
> Maybe that's fine and we just say rebase before we run any checks, but I also know it's not always convenient to rebase.

Our formatting CI jobs already don't run if there is a merge conflict.

Example:
https://github.com/canonical/cloud-init/pull/1599

holmanb:
@paride @TheRealFalcon Thanks for feedback on this.

> I don't think these changes are actually needed

At this point I agree that this PR should be dropped. If no objections from @aciba90 or anyone else I'll close it.

aciba90:
> @paride @TheRealFalcon Thanks for feedback on this.
> 
> > I don't think these changes are actually needed
> 
> At this point I agree that this PR should be dropped. If no objections from @aciba90 or anyone else I'll close it.

Yes, please, sorry for the delayed answer. I guess the best solution atm is to handle the problems manually (which do not appear to be very frequent) and wait for GH native solution.

Thanks for the PR!

holmanb:
> delayed answer

No problem :)

Thanks

PR #1603, opened 2022-07-19
Holmanb/util fix

holmanb:
```
util: Fix error path in get_proc_ppid()
```

## Additional Context
If IOError is thrown, contents is unbound and will throw a name error exception.

## Test Steps
covered in tests/unittests/test_util.py

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2895,7 +2895,7 @@ def get_proc_ppid(pid):
     try:
         contents = load_file("/proc/%s/stat" % pid, quiet=True)
         if contents:
-            parts = contents.split(" ", 4)
+            parts = str(contents).split(" ", 4)
             # man proc says
             #  ppid %d     (4) The PID of the parent.
             ppid = int(parts[3])
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2894,13 +2894,13 @@ def get_proc_ppid(pid):
     ppid = 0
     try:
         contents = load_file("/proc/%s/stat" % pid, quiet=True)
+        if contents:
+            parts = contents.split(" ", 4)
+            # man proc says
+            #  ppid %d     (4) The PID of the parent.
+            ppid = int(parts[3])
     except IOError as e:
         LOG.warning("Failed to load /proc/%s/stat. %s", pid, e)
-    if contents:
-        parts = contents.split(" ", 4)
-        # man proc says
-        #  ppid %d     (4) The PID of the parent.
-        ppid = int(parts[3])
     return ppid
 
 


aciba90 (approved):
LGTM. Thanks for squashing this one!

holmanb:
@aciba90 Thanks for the review.

I just pushed two more commits that fix another bug I just stumbled across in this code. tl;dr split()[4] works in the common case that a command name doesn't have spaces, but fails in scenarios that involve extra spaces. `man 5 proc` states that commands are surrounded in parenthesis (for disambiguation, presumably).

cloudinit/util.py
```diff
@@ -2894,13 +2894,13 @@ def get_proc_ppid(pid):
     ppid = 0
     try:
         contents = load_file("/proc/%s/stat" % pid, quiet=True)
+        if contents:
+            # see proc.5 for format
+            m = re.search(r"^\d+ \(.+\) [RSDZTtWXxKWP] (\d+)", str(contents))
```
blackboxsw, line 2899: Looks like the man page is incomplete as I see an `I` state on my kinetic system. Als you have a duplicate `W` we can drop.

```suggestion
            m = re.search(r"^\d+ \(.+\) [RSDZTtWXxKPI] (\d+)", str(contents))```
Testing on my system:

**Review by blackboxsw — CHANGES_REQUESTED** [review]
Testing this with the following script I could see failures on my side for parsing an `I` state type.
Tested on centos and debian as well. Didn't get into a BSD system though to spot check.

Here's my test script the function from your PR which helped show me re.match failures:
```bash
echo > testit.py <<EOF
#!/usr/bin/python3
import glob
import re

from cloudinit.util import load_file
def get_proc_ppid(pid):
    """
    Return the parent pid of a process.
    """
    ppid = 0
    try:
        contents = load_file("/proc/%s/stat" % pid, quiet=True)
        if contents:
            # see proc.5 for format
            m = re.search(r"^\d+ \(.+\) [RSDZITtWXxKP] (\d+)", str(contents))
            if m:
                ppid = int(m.group(1))
    except IOError as e:
        LOG.warning("Failed to load /proc/%s/stat. %s", pid, e)
    return ppid


for fname in glob.glob("/proc/*/stat"):
  try:
      pid = int(fname.split('/')[2])
  except ValueError:
      continue
  if 0 == get_proc_ppid(pid):
     print(fname)
     with open(fname, "r") as stream:
        print(fname)
        print(stream.read())
EOF
python3 ./testit.py
```
Without the `I` for "Idle" there we lots of pids that didn't get parsed for parent ID

cloudinit/util.py
```diff
@@ -2894,13 +2894,13 @@ def get_proc_ppid(pid):
...
+            if m:
+                ppid = int(m.group(1))
```
blackboxsw, line 2901: I think we probably should also WARN if we are unable to match here so we don't silently succeed returning 0 as parent process on unexpected format

cloudinit/util.py
holmanb: Nice catch :)

Looks like that was introduced in https://github.com/torvalds/linux/commit/06eb61844d841d0032a9950ce7f8e783ee49c0d0, and this hasn't made its way into the upstream man pages project yet. I just submitted a [patch](https://lore.kernel.org/linux-man/20220805044812.glotze62j2dldrdw@isa/T/#u) to add it.

I don't see any states in the array that aren't in our list (besides the one you found), so I think that covers our bases on the linux front (no clue about the BSDs).

blackboxsw: Thanks for the due diligence to check on other states in the kernel and finding the source related to that diff and fixing upstream for it :) All good here!

blackboxsw: ahh right and BSD doesn't surface this info in `/proc` in many cases due to /proc deprecation I believe.

blackboxsw (approved):
Changeset looks good

PR #1605, opened 2022-07-20
Modernise importer.py and type ModuleDetails

aciba90:
## Proposed Commit Message

```
Modernise importer.py and type ModuleDetails
```

## Additional Context

Minor refactoring extracted from https://github.com/canonical/cloud-init/pull/1597

## Checklist:

### DIFF
--- a/cloudinit/importer.py
+++ b/cloudinit/importer.py
@@ -51,6 +51,7 @@ def find_module(
     for full_path in lookup_paths:
         if not importlib.util.find_spec(full_path):
             continue
+        # Check that required_attrs are all present within the module.
         if _count_attrs(full_path, required_attrs) == len(required_attrs):
             found_paths.append(full_path)
     return (found_paths, lookup_paths)
--- a/cloudinit/importer.py
+++ b/cloudinit/importer.py
@@ -8,16 +8,34 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import sys
+import importlib
+from types import ModuleType
+from typing import Optional, Sequence
 
 
-def import_module(module_name):
-    __import__(module_name)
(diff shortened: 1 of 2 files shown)


aciba90:
Force pushed to resolve conflicts.

cloudinit/importer.py
```diff
@@ -8,16 +8,34 @@
...
+def _count_attrs(
+    module_name: str, attrs: Optional[Sequence[str]] = None
+) -> int:
+    found_attrs = 0
+    if not attrs:
+        return found_attrs
+    mod = importlib.import_module(module_name)
+    for attr in attrs:
+        if hasattr(mod, attr):
+            found_attrs += 1
```
holmanb, line 29: Is the number of found attributes really the important thing to test in `find_module()`? In other words, if the module has every required attribute, why should it not be imported?

If the module has every required attribute, but has additional attributes that are not included in the required_attrs argument, won't this counting test cause import to fail? I agree that this looks functionally similar to what we're already doing, I wonder whether it is unnecessarily brittle. If I add an attribute to a module and now the runtime importer fails, I'd be confused - and having to dig into the magical import machinery to figure out what's going on is less than ideal for anyone writing a module.

holmanb (approved):
I left a question inline, but since it appears to match current behavior I'm not going to block this work on my comment/question. Let me know if you want to do anything further, otherwise I'll merge. Thanks for this!

aciba90:
> I left a question inline, but since it appears to match current behavior I'm not going to block this work on my comment/question. Let me know if you want to do anything further, otherwise I'll merge. Thanks for this!

Thanks, @holmanb, for the review. I have answered you and added a comment within the code to make it 99% clear.

PR #1607, opened 2022-07-20
cloud-init clean --machine-id

blackboxsw:
## Proposed Commit Message
```
clean: add param to remove /etc/machine-id for golden image creation

Typically when a cloud-image boots /etc/machine-id should be
auto-generated if /etc/machine-id does not exist. Some cloud images
appropriately generate /etc/machine-id if the image is recognized
as a clone, in most other cases it is a best practice to remove
/etc/machine-id and let DBUS or systemd do the work of generating
a new UUID on next boot so the cloned VMs are represented
uniquely to any monitoring/management software.
```

## Additional Context
US015.UC01.2: subiquity/cloud-init spec, golden image support: ability to clean machine-id from golden image

## Checklist:

### DIFF
--- a/cloudinit/cmd/clean.py
+++ b/cloudinit/cmd/clean.py
@@ -22,6 +22,8 @@ from cloudinit.util import (
     is_link,
 )
 
+ETC_MACHINE_ID = "/etc/machine-id"
+
 
 def get_parser(parser=None):
     """Build or extend an arg parser for clean utility.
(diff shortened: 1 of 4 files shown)


aciba90 (approved):
LGTM. Thanks!

holmanb (approved):
I was considering requesting we document what a machine ID is, but since this is a concept [external to cloud-init](https://www.freedesktop.org/software/systemd/man/machine-id.html), it's probably not necessary. 

It's up to you if you want to add an external link in the docs. Either way, approved. Thanks!

blackboxsw:
Thanks @holmanb added 

> I was considering requesting we document what a machine ID is, but since this is a concept [external to cloud-init](https://www.freedesktop.org/software/systemd/man/machine-id.html), it's probably not necessary.
> 
> It's up to you if you want to add an external link in the docs. Either way, approved. Thanks!

+1 added that URL reference to the CLI docs

PR #1609, opened 2022-07-21
testing: Cover active modules

aciba90:
## Proposed Commit Message

```
testing: Cover active modules
```

## Additional Context

It helps to test SC-1133.
SC-1204

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #1610, opened 2022-07-21
config: Def activate_by_schema_keys for g-n mods

aciba90:
## Proposed Commit Message

```
config: Def activate_by_schema_keys for g-n mods
```

## Additional Context

SC-1133

## Test Steps

Merge #1609

## Checklist:

### DIFF
--- a/cloudinit/config/cc_install_hotplug.py
+++ b/cloudinit/config/cc_install_hotplug.py
@@ -49,6 +49,7 @@ meta: MetaSchema = {
         """
         ),
     ],
+    "activate_by_schema_keys": [],
 }
 
 __doc__ = get_meta_doc(meta)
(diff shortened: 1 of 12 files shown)


cloudinit/config/cc_install_hotplug.py
```diff
@@ -49,6 +49,7 @@
         """
         ),
     ],
+    "activate_by_schema_keys": [],
```
blackboxsw, line 52: good as we potentially need to uninstall hotplug config files if not enabled or available regardless of config

blackboxsw (approved):
LGTM, we can follow up with a single integration test after these all land to validate the collective "skip" message and/or cloud-init analyze show output to confirm modules were skipped w/out "errors"

PR #1611, opened 2022-07-21
config: Def activate_by_schema_keys for p-r mods

aciba90:
## Proposed Commit Message

```
config: Def activate_by_schema_keys for p-r mods
```

## Additional Context

SC-1133
SC-1205

## Test Steps

Merge #1609

## Checklist:

### DIFF
--- a/cloudinit/config/cc_phone_home.py
+++ b/cloudinit/config/cc_phone_home.py
@@ -88,7 +88,7 @@ meta: MetaSchema = {
             """
         ),
     ],
-    "activate_by_schema_keys": [],
+    "activate_by_schema_keys": ["phone_home"],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_phone_home.py
+++ b/cloudinit/config/cc_phone_home.py
@@ -88,6 +88,7 @@ meta: MetaSchema = {
             """
         ),
     ],
+    "activate_by_schema_keys": [],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_rightscale_userdata.py
+++ b/cloudinit/config/cc_rightscale_userdata.py
@@ -44,6 +44,7 @@ meta: MetaSchema = {
     "distros": [ALL_DISTROS],
     "frequency": PER_INSTANCE,
     "examples": [],
+    "activate_by_schema_keys": [],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_package_update_upgrade_install.py
+++ b/cloudinit/config/cc_package_update_upgrade_install.py
@@ -47,6 +47,13 @@ meta: MetaSchema = {
             """
         )
     ],
+    "activate_by_schema_keys": [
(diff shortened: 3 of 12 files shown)


cloudinit/config/cc_rightscale_userdata.py
```diff
@@ -44,6 +44,7 @@
     "distros": [ALL_DISTROS],
     "frequency": PER_INSTANCE,
     "examples": [],
+    "activate_by_schema_keys": [],
```
blackboxsw, line 47: So, in reviewing this. This module does not work at all currently. It calls `[cloud.get_userdata_raw()` which is a method that doesn't exist and then drops into a very generic, unhelpful exception message.](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_rightscale_userdata.py#L73)   The reason this errors always is
```
2022-07-21 21:10:30,402 - cc_rightscale_userdata.py[DEBUG]: Failed to get raw userdata in module cc_rightscale_userdata 'Cloud' object has no attribute 'get_userdata_raw'
```

So,  this looks like it broke in commit 32362c0ac63f10d7a33e3b95dd91a544a1cbdf5 . I've contacted support at Flexera (which now owns RightScale) to ensure they have no use for this module anymore so that we can remove it instead of fixing it if unneeded. That'd be a separate PR if they concur their products don't use cloud-init.

cloudinit/config/cc_phone_home.py
```diff
@@ -88,6 +88,7 @@
             """
         ),
     ],
+    "activate_by_schema_keys": [],
```
blackboxsw, line 91:
```suggestion
    "activate_by_schema_keys": ["phone_home"],
```

blackboxsw (changes requested):
Minor miss on `phone_home` otherwise +1. 
We need to talk as a team about the potential dropping rightscale_userdata. I'll await feedback from support@flexera.com on this though.

cloudinit/config/cc_phone_home.py
aciba90: Fixed. The reason to not added in the first place was that this module makes use of the `args` argument, and I wasn't sure which entry-point passes `args` through.

cloudinit/config/cc_rightscale_userdata.py
aciba90: Thanks for pointing this out. I did diagonally read the modules and did not notice it.

aciba90:
Thanks, @blackboxsw, for the review. Ready to be reviewed after CI.

cloudinit/config/cc_phone_home.py
blackboxsw: +1 @aciba90 the only way to pass through args is via `cloud-init single -c cc_phone_home <some_yaml_file>` which avoids your skip logic over in cloudinit.config.modules anyway. I too had to spend time digging that path up yesterday too.

blackboxsw: not a common use case for sure

blackboxsw (approved):
LGTM!

PR #1612, opened 2022-07-21
config: Def activate_by_schema_keys for s mods

aciba90:
## Proposed Commit Message

```
config: Def activate_by_schema_keys for s mods
```

## Additional Context

SC-1133

## Test Steps

Merge #1609

## Checklist:

### DIFF
--- a/cloudinit/config/cc_seed_random.py
+++ b/cloudinit/config/cc_seed_random.py
@@ -69,6 +69,7 @@ meta: MetaSchema = {
             """
         ),
     ],
+    "activate_by_schema_keys": [],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_ssh_import_id.py
+++ b/cloudinit/config/cc_ssh_import_id.py
@@ -43,6 +43,7 @@ meta: MetaSchema = {
             """
         )
     ],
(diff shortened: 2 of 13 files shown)


cloudinit/config/cc_seed_random.py
```diff
@@ -69,6 +69,7 @@
             """
         ),
     ],
+    "activate_by_schema_keys": [],
```
blackboxsw, line 72: Yes because we fall through to do work related to datasource.metadata too on Azure.

cloudinit/config/cc_ssh_import_id.py
```diff
@@ -43,6 +43,7 @@
             """
         )
     ],
+    "activate_by_schema_keys": [],
```
blackboxsw, line 46: The only reason we can't provide `ssh_import_id` here is because we have a nested check for ssh_import_id under users definitions via is_key_in_nested_dict. If we grew support for callables we'd be able to disregard if is_key_in_nested_dict("ssh_import_id") but I don't think this one case is worth growing that feature. If we have multiple cases that align, it may make sense to do something a bit more complex.

blackboxsw (approved):
LGTM!

PR #1613, opened 2022-07-21
config: Def activate_by_schema_keys for t-z

aciba90:
## Proposed Commit Message

```
config: Def activate_by_schema_keys for t-z mods
```

## Additional Context

SC-1133
SC-1203

## Test Steps

Merge #1609

## Checklist:

### DIFF
--- a/cloudinit/config/cc_update_etc_hosts.py
+++ b/cloudinit/config/cc_update_etc_hosts.py
@@ -88,7 +88,7 @@ meta: MetaSchema = {
         ),
     ],
     "frequency": PER_ALWAYS,
-    "activate_by_schema_keys": [],
+    "activate_by_schema_keys": ["manage_etc_hosts"],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_update_etc_hosts.py
+++ b/cloudinit/config/cc_update_etc_hosts.py
@@ -88,6 +88,7 @@ meta: MetaSchema = {
         ),
     ],
     "frequency": PER_ALWAYS,
+    "activate_by_schema_keys": [],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_timezone.py
+++ b/cloudinit/config/cc_timezone.py
@@ -26,6 +26,7 @@ meta: MetaSchema = {
     "examples": [
         "timezone: US/Eastern",
     ],
+    "activate_by_schema_keys": ["timezone"],
 }
 
 __doc__ = get_meta_doc(meta)
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -90,6 +90,7 @@ meta: MetaSchema = {
         ),
     ],
     "frequency": PER_INSTANCE,
+    "activate_by_schema_keys": ["ubuntu_advantage", "ubuntu-advantage"],
 }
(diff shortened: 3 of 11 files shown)


cloudinit/config/cc_update_etc_hosts.py
```diff
@@ -88,6 +88,7 @@
         ),
     ],
     "frequency": PER_ALWAYS,
+    "activate_by_schema_keys": [],
```
blackboxsw, line 91: @aciba90 shouldn't this be `manage_etc_hosts`? As I'm reading this code it looks like we default to False if not provided, and a False value drops us into [the final else clause which skips](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_update_etc_hosts.py#L138).

```suggestion
    "activate_by_schema_keys": ["manage_etc_hosts"],
```
One thing to consider if if we do skip altogether, we will not get a debug log like:
```
 Configuration option 'manage_etc_hosts' is not set, not managing /etc/hosts in module update_etc_hosts
```

It's dubious the value of such a log, but I'll let you and @holmanb decide before merging.

blackboxsw (approved):
Are we missing cc_timezone.py intentionally?
One minor fix missing cc_timezone. And you and @holmanb can decide on whether or not to handle that manage_etc_hosts comment I had.

```diff
diff --git a/cloudinit/config/cc_timezone.py b/cloudinit/config/cc_timezone.py
index 47da2d065..b9df31af1 100644
--- a/cloudinit/config/cc_timezone.py
+++ b/cloudinit/config/cc_timezone.py
@@ -26,6 +26,7 @@ meta: MetaSchema = {
     "examples": [
         "timezone: US/Eastern",
     ],
+    "activate_by_schema_keys": ["timezone"],
 }
 
 __doc__ = get_meta_doc(meta)
```

cloudinit/config/cc_update_etc_hosts.py
aciba90: I have added as in my opinion that log message is quasi-equivalent to the want we emit while skipping modules:

`Skipping modules 'update_etc_hosts, ...' because no applicable config is provided.`

But if @holmanb thinks the opposite, I will revert it.

aciba90:
> Are we missing cc_timezone.py intentionally? One minor fix missing cc_timezone. And you and @holmanb can decide on whether or not to handle that manage_etc_hosts comment I had.
> 
> ```diff
> diff --git a/cloudinit/config/cc_timezone.py b/cloudinit/config/cc_timezone.py
> index 47da2d065..b9df31af1 100644
> --- a/cloudinit/config/cc_timezone.py
> +++ b/cloudinit/config/cc_timezone.py
> @@ -26,6 +26,7 @@ meta: MetaSchema = {
>      "examples": [
>          "timezone: US/Eastern",
>      ],
> +    "activate_by_schema_keys": ["timezone"],
>  }
>  
>  __doc__ = get_meta_doc(meta)
> ```

The reason to not add it was the same as in https://github.com/canonical/cloud-init/pull/1611#discussion_r927327986. I was not sure about the `args` argument. Fixed.

cloudinit/config/cc_update_etc_hosts.py
holmanb: @aciba90 I'm fine with the "skipping .." message in place of "Configuration option...".

holmanb (approved):
Looks good to me, thanks!

PR #1608, opened 2022-07-21
config: Def `activate_by_schema_keys` for a-f mods

aciba90:
## Proposed Commit Message

```
config: Def `activate_by_schema_keys` for a-f mods
```

## Additional Context

SC-1133

## Test Steps

Merge #1609

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #1614, opened 2022-07-21
cc_debug: Drop this module

aciba90:
## Proposed Commit Message

```
cc_debug: Drop this module
```

## Additional Context

https://github.com/canonical/cloud-init/pull/1608#discussion_r926848487

## Checklist:


cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -589,26 +589,6 @@
         }
       }
     },
-    "cc_debug": {
```
aciba90, line 592: Not sure if it is safe to drop this sub-schema. We could also keep it as deprecated to maintain back-ward compatibility.

blackboxsw (approved):
This is ok to drop. The functionality it originally was intended for is completed replaced by `cloud-init query --all` which is consumed in `cloud-init collect-logs`.

cloudinit/config/schemas/schema-cloud-config-v1.json
blackboxsw: It's ok to drop this. we shouldn't surface schema for modules that no longer exist. I also don't think there are many use-cases where 'debug' makes sense (and it isn't included in default `/etc/cloud/cloud.cfg` module ordering so folks would have to create custom images with modified /etc/cloud/cloud.cfg.d/ or launch with custom vendor-data to add this module into active boot.

PR #1622, opened 2022-07-27
sources/azure: add experimental support for preprovisioned os disks

cjp256:
sources/azure: add experimental support for preprovisioned os disks

    Some pre-provisioning scenarios require that the VM be started and
    shut back down as part of preparing the VM for future use.

    When the PPS type is PreprovisionedOSDisk, report ready and wait for
    host to shut down the VM.

    Provisioning will resume normally on next boot, so do not write
    a reported ready marker file.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -93,6 +93,7 @@ class MetadataType(Enum):
 
 class PPSType(Enum):
     NONE = "None"
+    OS_DISK = "PreprovisionedOSDisk"
     RUNNING = "Running"
     SAVABLE = "Savable"
     UNKNOWN = "Unknown"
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -982,7 +985,30 @@ def _report_ready_for_pps(self) -> None:
...
+        try:
+            subp.subp(shutdown_cmd)
+        except subp.ProcessExecutionError:
+            report_diagnostic_event(
+                "Shutdown command failed, waiting for host to force it...",
+                logger_func=LOG.info,
+            )
+
+        sleep(31536000)
+        raise BrokenAzureDataSource("Shutdown failure for PPS disk.")
```
anhvoms, line 1011: does this exception lead to report failure? If possible, reporting failure would be best if it cannot be shutdown

anhvoms: nvm, can't report failure after reporting ready

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
cjp256: It would go down that code path anyways.

holmanb (approved):
- [x] Tests added for new functionality
- [x] New behavior gated on new config value
- [x] Behavior remains unchanged for current configs

Looks good to me, thanks!

PR #1606, opened 2022-07-20
main: avoid downloading full contents cmdline urls

aciba90:
## Proposed Commit Message

```
main: avoid downloading full contents cmdline urls
    
In the case that the url content does not start with
`#cloud-config`, avoid downloading the full content.
Add some types.

LP: #1937319
```

## Additional Context

SC-1115

## Test Steps

Follow [nocloud.html#datasource-nocloud](https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html#datasource-nocloud) to set up kernel command line to point
to an url without a cloud-config.

url=https://ftp.osuosl.org/pub/ubuntu-releases/18.04/ubuntu-18.04.6-desktop-amd64.iso
and
cloud-config-url=https://ftp.osuosl.org/pub/ubuntu-releases/18.04/ubuntu-18.04.6-desktop-amd64.iso

```sh
$ grep -in 'did not start with' /var/log/cloud-init.log
...contents of 'https://ftp.osuosl.org/pub/ubuntu-releases/18.04/ubuntu-18.04.6-desktop-amd64.iso' did not start with...
```

cloud-init won't download the iso.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #1633, opened 2022-08-02
typing: Type UrlResponse.contents

aciba90:
## Proposed Commit Message

```
typing: Type UrlResponse.contents

As requests.Response.contents returns always an instance of bytes,
type UrlResponse.contents accordingly.

This change undercovered an incorrect typing in
sources/helpers/azure, that was hidden because
ElementTree.fromstring accepts str and bytes.
```

## Additional Context

https://github.com/canonical/cloud-init/pull/1606#discussion_r934791692

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -13,7 +13,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from errno import ENOENT
 from time import sleep, time
-from typing import List, Optional
+from typing import List, Optional, Union
 from xml.etree import ElementTree
 from xml.sax.saxutils import escape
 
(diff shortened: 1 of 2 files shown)


holmanb:
> As requests.Response.contents returns always an instance of bytes,

This isn't [always true](https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/models.py#L897). If `self.status_code == 0 or self.raw is None` and readurl is called with `check_status=False`, won't we have typing that isn't valid?

aciba90:
Reopen to trigger the build.

aciba90:
> > As requests.Response.contents returns always an instance of bytes,
> 
> This isn't [always true](https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/models.py#L897). If `self.status_code == 0 or self.raw is None` and readurl is called with `check_status=False`, won't we have typing that isn't valid?

Good catch, thanks @holmanb. In order to avoid forcing client code to check for a None return, I have extended `UrlResponse.content` to handle that case and return an empty byte-string.

This is ready to be reviewed.

holmanb (approved):
LGTM

PR #1639, opened 2022-08-08
cloud.cfg.tmpl: make sure "centos" settings are identical to "rhel"

esposem:
## Proposed Commit Message
```
We have a couple of bugs where centos does not have the default user as rhel.
This PR makes sure the configuration is exactly the same.

```
RHBZ: 2115565 and 2115576

## Test Steps
1. Boot the latest CentOS Stream cloud image
2. Attempt to log in with the "cloud-user"
3. Attempt to log in with the "centos" user

Actual results:
Login fails with "cloud-user" and succeeds with "centos"

Expected results:
Login succeeds with "cloud-user" and fails with "centos"

Additional info:
Originally, RHEL/CentOS used "cloud-user" with the custom cloud-init configuration files. Too much configuration was removed in the rebase to cloud-init v22.1.

## Checklist:


Conan-Kudo:
Not that this wouldn't help fix the problem for CentOS, but I think the problem is that we're not making the right calls in our spec file in RHEL/CentOS to ensure that the correct configuration is generated?

The Fedora spec ensure the RHEL variant is used for RHEL-like systems, while it ensures the Fedora variant is used for Fedora-like ones.

```rpm-spec
%py3_install -- --init-system=systemd 

# Generate cloud-config file python3 tools/render-cloudcfg --variant %{?rhel:rhel}%{!?rhel:fedora} > $RPM_BUILD_ROOT/%{_sysconfdir}/cloud/cloud.cfg
```

Conan-Kudo:
Not that this wouldn't help fix the problem for CentOS, but I think the problem is that we're not making the right calls in our spec file in RHEL/CentOS to ensure that the correct configuration is generated?

The Fedora spec ensure the RHEL variant is used for RHEL-like systems, while it ensures the Fedora variant is used for Fedora-like ones.

```rpm-spec
%py3_install -- --init-system=systemd 

# Generate cloud-config file 
python3 tools/render-cloudcfg --variant %{?rhel:rhel}%{!?rhel:fedora} > $RPM_BUILD_ROOT/%{_sysconfdir}/cloud/cloud.cfg
```

esposem:
As already explained in the BZ comment, the right call is performed in our spec file:
```
python3 tools/render-cloudcfg --variant rhel systemd/cloud-init-generator.tmpl > $RPM_BUILD_ROOT%{_systemdgeneratordir}/cloud-init-generator
```

Conan-Kudo:
That's not enough.

LaneWolf:
This is not problem of calling but problem of template. Template create cloud-user only for rhel not for centos. Template has to be fixed here.

holmanb:
@LaneWolf @Conan-Kudo @esposem - Thanks for the discussion, and @esposem for a proposed fix.

I'm looking forward to merging a fix, but I'd prefer to see consensus if possible before putting in time on a review. Give a ping if that doesn't seem likely and we'll figure it out one way or another.

In the meantime thanks all for your efforts :)

Conan-Kudo:
The issue with doing it this way is that it means every RHEL derivative needs to be filled in here too. That's why I deliberately have the cloud-init spec file in Fedora calling either the Fedora or RHEL variant for generating the templates. It gets way too complicated and ugly otherwise.

esposem:
@Conan-Kudo  see if it's better now

config/cloud.cfg.tmpl
```diff
@@ -32,9 +33,9 @@ disable_root: false
 disable_root: true
 {% endif %}
 
-{% if variant in ["almalinux", "alpine", "amazon", "centos", "cloudlinux", "eurolinux",
-                  "fedora", "miraclelinux", "openEuler", "openmandriva", "rhel", "rocky", "virtuozzo"] %}
-{% if variant == "rhel" %}
+{% if variant in ["almalinux", "alpine", "amazon", "cloudlinux", "eurolinux",
+                  "fedora", "miraclelinux", "openEuler", "openmandriva", "rocky", "virtuozzo"] or is_rhel %}
```
Conan-Kudo, line 37: `almalinux`, `cloudlinux`, `eurolinux`, `miraclelinux`, `rocky`, and `virtuozzo` are all `is_rhel` variants, so they need to be pulled out too.

Conan-Kudo (changes requested):

config/cloud.cfg.tmpl
LaneWolf: This is question of what we want to have. Originally 'rhel' variant was used exclusively and separated from other variants Now we are making 'centos' equal to it as for CentOS 9 Stream/RHEL 9 it works that way (C9S is synced to RHEL9 as 1:1 copy). Whether this behavior should be applied to other variants is questionable - I do not know what's the general expectation here.

My opinion is: for template treat 'rhel' and 'centos' as equivalent (as they should be from development point of view). Any other variant should use --variant option to say they try to be 'rhel'-like.

However, for RHEL/C9S spec we should probably add this option too to prevent any differences in case centos-rhel equality is not handled properly somewhere in template.

TheRealFalcon:
Any updates here? Do we have a path forward?

esposem:
From what I understand, we (redhat, me and @LaneWolf) agree on keeping it as it is. So from our side, this PR can go ahead

holmanb (approved):
I don't see any more activity on this PR, so I'm going to wrap this up.

Since this forces alignment between CentOS and RHEL and doesn't otherwise break or change other distro definitions, I'm going to merge. 

We can modify other distros if required in subsequent PRs.

Thanks again @Conan-Kudo, @esposem, and @LaneWolf :)

PR #1644, opened 2022-08-09
schema: Resolve user-data if --system given

aciba90:
## Proposed Commit Message

```
schema: Resolve user-data if --system given

Validated the processed/rendered userdata instead of "userdata_raw"
so that `sudo cloud-init schema --system` provides a concise
error about what the procesed invalid user-data is.

LP: #1983306
```

## Additional Context

SC-1215

## Test Steps

```bash
mkdir -p /tmp/instance-data

cat > /tmp/instance-data/user-data <<EOF
#cloud-config
runcmd:
- touch /root/hello.txt
EOF

cat > /tmp/instance-data/script.sh <<EOF
#!/bin/sh

touch /root/bye.txt
EOF
python -m cloudinit.cmd.main devel make-mime -a /tmp/instance-data/user-data:cloud-config -a /tmp/instance-data/script.sh:x-shellscript > /tmp/instance-data/userdata
gzip -f /tmp/instance-data/userdata
cd /tmp/instance-data && python3 -m http.server &

cat > /tmp/include.yaml <<EOF
#include http://0.0.0.0:8000/userdata.gz
EOF

lxc init ubuntu-daily:kinetic include-k -c user.user-data="$(cat /tmp/include.yaml)"
lxc config device add include-k hostbind proxy listen=tcp:0.0.0.0:8000 connect=tcp:0.0.0.0:8000 bind=instance
lxc config device add include-k host-cloud-init disk source="/home/aciba/canonical/cloud-init/cloudinit" path=/usr/lib/python3/dist-packages/cloudinit
lxc start include-k

lxc exec include-k -- cloud-init status --wait

# See raw userdata (this is ok/desired)
$ lxc exec include-k -- cloud-init query userdata
include http://0.0.0.0:8000/userdata.gz

# verify cloud-init schema --system alias for rendered user-data (not correct)
$ lxc exec include-k -- cloud-init schema --system --annotate
Valid cloud-config: system userdata

$ lxc exec include-k -- ls /root/*.txt
bye.txt  hello.txt
```

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -19,7 +19,7 @@ import yaml
 
 from cloudinit import importer, safeyaml
 from cloudinit.stages import Init
-from cloudinit.util import encode_text, error, get_modules_from_dir, load_file
+from cloudinit.util import error, get_modules_from_dir, load_file
 
 try:
     from jsonschema import ValidationError as _ValidationError
@@ -618,16 +618,8 @@ def validate_cloudconfig_file(config_path, schema, annotate=False):
                 " Try using sudo"
             )
         init = Init(ds_deps=[])
-        ds = init.fetch("trust")
-        ud = ds.get_userdata()
-        content = None
-        for part in ud.walk():
-            if part.get_content_type() == "text/cloud-config":
-                content = encode_text(part.get_payload())
-                break
-        if not content:
-            print("No cloud-config userdata found. Skipping verification")
-            return
+        init.consume_data()
+        content = load_file(init.paths.get_ipath("cloud_config"), decode=False)
     else:
         if not os.path.exists(config_path):
             raise RuntimeError(
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -18,8 +18,8 @@ from typing import TYPE_CHECKING, List, NamedTuple, Optional, Type, Union, cast
 import yaml
 
 from cloudinit import importer, safeyaml
-from cloudinit.cmd.devel import read_cfg_paths
-from cloudinit.util import error, get_modules_from_dir, load_file
+from cloudinit.stages import Init
+from cloudinit.util import encode_text, error, get_modules_from_dir, load_file
 
 try:
     from jsonschema import ValidationError as _ValidationError
@@ -617,9 +617,17 @@ def validate_cloudconfig_file(config_path, schema, annotate=False):
                 "Unable to read system userdata as non-root user."
                 " Try using sudo"
             )
-        paths = read_cfg_paths()
-        user_data_file = paths.get_ipath_cur("userdata_raw")
-        content = load_file(user_data_file, decode=False)
+        init = Init(ds_deps=[])
+        ds = init.fetch("trust")
+        ud = ds.get_userdata()
+        content = None
+        for part in ud.walk():
+            if part.get_content_type() == "text/cloud-config":
+                content = encode_text(part.get_payload())
+                break
+        if not content:
+            print("No cloud-config userdata found. Skipping verification")
+            return
     else:
         if not os.path.exists(config_path):
             raise RuntimeError(
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -618,6 +618,7 @@ def validate_cloudconfig_file(config_path, schema, annotate=False):
                 " Try using sudo"
             )
         init = Init(ds_deps=[])
+        init.fetch(existing="trust")
         init.consume_data()
         content = load_file(init.paths.get_ipath("cloud_config"), decode=False)
     else:
--- a/tests/unittests/config/test_schema.py
+++ b/tests/unittests/config/test_schema.py
@@ -742,7 +742,6 @@ class TestValidateCloudConfigFile:
             validate_cloudconfig_file(None, schema, annotate)
 
     @skipUnlessJsonSchema()
-    @httpretty.activate
     @pytest.mark.parametrize("annotate", (True, False))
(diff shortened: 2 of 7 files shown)


tests/unittests/config/test_schema.py
```diff
@@ -16,8 +16,10 @@
 from types import ModuleType
 from typing import List, Optional, Sequence, Set
 
+import httpretty
```
TheRealFalcon, line 19: I think we want to prefer `responses` over `httpretty` now.

Changes to this should just require replacing httpretty with responses everywhere, and use `respones.add` instead of `httpretty.register_uri`, which should take the same arguments.

tests/unittests/config/test_schema.py
```diff
@@ -53,6 +55,17 @@
 
 M_PATH = "cloudinit.config.schema."
 
+INSTANCE_ID = "i-testing"
+
+
+class FakeDataSource(sources.DataSource):
```
TheRealFalcon, line 61: Nit (not something that needs to be changed): I'd like to consolidate this type of helper class for tests into `unittests/util.py`. We already have a `DataSourceTesting` that is more for testing implementations that require using `get_cloud`, but I think the implementation here (or perhaps a helper function) could live there too.

TheRealFalcon (changes requested):
I left some minor comments inline, but otherwise we still don't have all of our cases covered.

I created a multipart archive using the example here:
https://cloudinit.readthedocs.io/en/latest/topics/format.html#examples

I just used a random cloud-config and bash script together. From that I got this:
```
Content-Type: multipart/mixed; boundary="===============6941792458632248426=="
MIME-Version: 1.0

--===============6941792458632248426==
Content-Type: text/x-shellscript-per-boot; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="script.sh"

IyEvYmluL2Jhc2gKCmVjaG8gJ3RoZXJlJyA+IC92YXIvdG1wL3RoZXJlCg==

--===============6941792458632248426==
Content-Type: text/cloud-config; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="cloud-config.yaml"

I2Nsb3VkLWNvbmZpZwpwcmVmZXJfZnFkbl9vdmVyX2hvc3RuYW1lOiB0cnVlCg==

--===============6941792458632248426==--

```

Using that as userdata I get errors (since it's still base64 encoded) when trying to do `cloud-init schema --system`. We should make sure we support that too.

If we really want to go down the rabbit hole, we could test an include file that includes multiple gzipped multipart mime files.

tests/unittests/config/test_schema.py
aciba90: Done.

tests/unittests/config/test_schema.py
aciba90: Done.

aciba90:
> Using that as userdata I get errors (since it's still base64 encoded) when trying to do `cloud-init schema --system`. We should make sure we support that too.

Thanks, @TheRealFalcon, for pointing out this one. Fixed.

> If we really want to go down the rabbit hole, we could test an include file that includes multiple gzipped multipart mime files.

I have updated the Test Steps from the PR description to include the rabbit hole tests, which is supported.

This PR is ready to be reviewed after CI passes.

cloudinit/config/schema.py
```diff
@@ -617,9 +617,10 @@ def validate_cloudconfig_file(config_path, schema, annotate=False):
                 "Unable to read system userdata as non-root user."
                 " Try using sudo"
             )
-        paths = read_cfg_paths()
-        user_data_file = paths.get_ipath_cur("userdata_raw")
-        content = load_file(user_data_file, decode=False)
+        init = Init(ds_deps=[])
+        init.fetch(existing="trust")
```
aciba90, line 621: As this branch is executed when the user requests to check the system config, and we can assume that `cloud-init` did run, then I think it would be safe to remove the following lines:

```python
        init.fetch(existing="trust")
        init.consume_data()
```

as the user-data has already been consumed.

tests/unittests/config/test_schema.py
```diff
@@ -702,6 +706,56 @@ def test_validateconfig_file_strictly_validates_schema(
         with pytest.raises(SchemaValidationError, match=error_msg):
             validate_cloudconfig_file(config_file.strpath, schema, annotate)
 
+    @skipUnlessJsonSchema()
+    @responses.activate
+    @pytest.mark.parametrize("annotate", (True, False))
+    @mock.patch("cloudinit.url_helper.time.sleep")
+    @mock.patch(M_PATH + "os.getuid", return_value=0)
+    def test_validateconfig_file_include_validates_schema(
```
TheRealFalcon, line 714: Can we add one more test case here for checking an `#include` that succeeds?

cloudinit/config/schema.py
TheRealFalcon: I think we should keep it as-is. When using the pickled data, the time to "fetch" is trivial when waiting on the command line.

TheRealFalcon (commented):
Thanks for the changes @aciba90 . PR looks good...just looking for one more test.

tests/unittests/config/test_schema.py
aciba90: Done.

aciba90:
> Thanks for the changes @aciba90 . PR looks good...just looking for one more test.

Thanks, @TheRealFalcon, for the review. I have added the extra test. This is ready to be reviewed after CI passes.

TheRealFalcon (approved):
Looks good!

PR #1646, opened 2022-08-09
Fix sshd tests

holmanb:
```
test: Fix system-dependent sshd tests
    
These tests fail in multiple ways on my system since they
depend on specific directory/file locations of the tests system.
```

## Additional Context

We should probably mock the tests properly rather than just deleting a test call, but (I think) that would require mocking `os.path.exists`, `os.environ.copy`, `os.path.dirname` in addition to `util.del_file` call I added, and I'm wondering if that started to get a little excessive.

Thoughts?

### DIFF
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -284,14 +284,12 @@ class TestHandleSsh:
         )
 
     @pytest.mark.parametrize("with_sshd_dconf", [False, True])
-    @mock.patch(MODPATH + "util.ensure_dir")
     @mock.patch(MODPATH + "ug_util.normalize_users_groups")
     @mock.patch(MODPATH + "util.write_file")
     def test_handle_ssh_keys_in_cfg(
         self,
         m_write_file,
         m_nug,
-        m_ensure_dir,
         m_setup_keys,
         with_sshd_dconf,
         mocker,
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
There's old infrastructure for this. It basically re-routes all those calls to be under /tmp/random_test_dir instead:
https://github.com/canonical/cloud-init/blob/main/tests/unittests/helpers.py#L254

We should really convert that to a (autouse?) pytest fixture so individual unit tests don't have to deal with all that mocking. @aciba90 mentioned looking for new work to do :wink:

aciba90:
> There's old infrastructure for this. It basically re-routes all those calls to be under /tmp/random_test_dir instead: https://github.com/canonical/cloud-init/blob/main/tests/unittests/helpers.py#L254
> 
> We should really convert that to a (autouse?) pytest fixture so individual unit tests don't have to deal with all that mocking. @aciba90 mentioned looking for new work to do wink

Let me take over this one, as I probably introduced the problem too. @holmanb, @TheRealFalcon.

aciba90:
> > There's old infrastructure for this. It basically re-routes all those calls to be under /tmp/random_test_dir instead: https://github.com/canonical/cloud-init/blob/main/tests/unittests/helpers.py#L254
> > We should really convert that to a (autouse?) pytest fixture so individual unit tests don't have to deal with all that mocking. @aciba90 mentioned looking for new work to do wink
> 
> Let me take over this one, as I probably introduced the problem too. @holmanb, @TheRealFalcon.

It should be fixed in this one: https://github.com/canonical/cloud-init/pull/1647
Could you please, @holmanb, verify so? Thanks!

PR #1650, opened 2022-08-12
net: Passthough v2 netconfigs in netplan systems

aciba90:
## Proposed Commit Message

```
net: Passthough v2 raw net_conf in netplan systems

Adhere to Netplan Passthrough documented behavior,
not limiting v2 netplan configs to the subset of
props that cloud-init supports.

LP: #1978543
```

## Additional Context

SC-1120
https://bugs.launchpad.net/cloud-init/+bug/1978543
[Netplan Passthrough](https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v2.html#netplan-passthrough)

## Test Steps

```bash
cat > /tmp/lp1978543.yaml <<EOF
network:
  version: 2
  ethernets:
    enp0s3:
      dhcp4: false
      addresses: [10.0.4.10/24]
      nameservers:
        addresses: [10.0.4.1]
      routes:
      - to: default
        via: 10.0.4.1
        metric: 100
EOF

lxc init ubuntu-daily:jammy lp1978543
lxc file push /tmp/lp1978543.yaml lp1978543/etc/cloud/cloud.cfg.d/custom-networking.cfg
lxc file push /tmp/lp1978543.yaml lp1978543/root/
lxc start lp1978543
lxc exec lp1978543 -- cloud-init status --wait

lxc exec lp1978543 -- grep -in valueerror /var/log/cloud-init.log
# No errors found
$ lxc exec lp1978543 -- cat /etc/netplan/50-cloud-init.yaml
...
network:
    ethernets:
        enp0s3:
            addresses:
            - 10.0.4.10/24
            dhcp4: false
            nameservers:
                addresses:
                - 10.0.4.1
            routes:
            -   metric: 100
                to: default
                via: 10.0.4.1
    version: 2

lxc exec lp1978543 -- python3 -m cloudinit.cmd.main devel net-convert -m enp0s3,AA -O netplan --k yaml -p lp1978543.yaml -D ubuntu -d out
# No errors found

$ lxc exec lp1978543 -- cat out/etc/netplan/50-cloud-init.yaml
...
network:
    ethernets:
        enp0s3:
            addresses:
            - 10.0.4.10/24
            dhcp4: false
            nameservers:
                addresses:
                - 10.0.4.1
            routes:
            -   metric: 100
                to: default
                via: 10.0.4.1
    version: 2
```

## Checklist:

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -23,8 +23,14 @@ from cloudinit import log as logging
 from cloudinit import net, persistence, ssh_util, subp, type_utils, util
 from cloudinit.distros.parsers import hosts
 from cloudinit.features import ALLOW_EC2_MIRRORS_ON_NON_AWS_INSTANCE_TYPES
-from cloudinit.net import activators, eni, network_state, renderers
-from cloudinit.net.network_state import parse_net_config_data
+from cloudinit.net import (
+    RendererNotFoundError,
+    activators,
+    eni,
+    network_state,
+    renderers,
+)
+from cloudinit.net.network_state import NetworkState, parse_net_config_data
 
 from .networking import LinuxNetworking, Networking
 
@@ -230,6 +236,16 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     def generate_fallback_config(self):
         return net.generate_fallback_config()
 
+    def _is_netplan(self) -> bool:
+        priority = util.get_cfg_by_path(
+            self._cfg, ("network", "renderers"), None
+        )
+        try:
+            name, *_ = renderers.select(priority=priority)
+        except RendererNotFoundError:
+            return False
+        return name == "netplan"
+
     def apply_network_config(self, netconfig, bring_up=False) -> bool:
         """Apply the network config.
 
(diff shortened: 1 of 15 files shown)


cloudinit/distros/__init__.py
```diff
@@ -230,6 +236,16 @@ def _apply_network_from_network_config(self, netconfig, bring_up=True):
     def generate_fallback_config(self):
         return net.generate_fallback_config()
 
+    def _is_netplan(self) -> bool:
```
TheRealFalcon, line 239: Since most of this method is doing the same thing as the beginnning of `_write_network_state`, I think it might make more sense to fetch the renderer one time at the beginning of `apply_network_config` and pass it where it's needed.

cloudinit/distros/__init__.py
```diff
@@ -241,7 +257,11 @@ def apply_network_config(self, netconfig, bring_up=False) -> bool:
         """
         # This method is preferred to apply_network which only takes
         # a much less complete network config format (interfaces(5)).
-        network_state = parse_net_config_data(netconfig)
+        if self._is_netplan() and netconfig.get("version") == 2:
```
TheRealFalcon, line 260: I don't like how the network rendering abstraction is leaking into the distro code here. Do you think it's possible to push this section of code down into either `parse_net_config_data` or even into `NetworkStateInterpreter`?

TheRealFalcon (commented):
I left two comments inline. Neither are strongly held though. Otherwise, looks good.

cloudinit/distros/__init__.py
aciba90: Done.

cloudinit/distros/__init__.py
aciba90: Logic moved to `NetworkStateInterpreter`. The downside is that we have to pass the renderer around, but it looks more cohesive in this way. Thanks for the comment.

aciba90:
Thanks, @TheRealFalcon, for the review. I have addressed the comments and this PR is ready to be reviewed.

TheRealFalcon (approved):
Thanks @aciba90 , nice work here.

The `isinstance` checks feel a little awkward. Long term, it might make sense for us to add a name property to the renderers and check that instead, but not required for this PR.

PR #1651, opened 2022-08-12
Return cc_set_hostname to PER_INSTANCE frequency (SC-1222)

TheRealFalcon:
## Proposed Commit Message

```
Return cc_set_hostname to PER_INSTANCE frequency

In 96eb95a the frequency was inadvertently changed to ALWAYS due to
the documentation being incorrect. The code and documentation should
now align to be PER_INSTANCE.

LP: #1983811

```


sshedi:
Thanks @TheRealFalcon.

holmanb (approved):

PR #1653, opened 2022-08-12
Ubuntu/focal: add expire-on-hashed-users.patch

TheRealFalcon:
This also includes a new upstream snapshot as we need the code we're patching in order to patch it.

The actual patch is represented by the last two commits.


blackboxsw (approved):
- local changeset good except authoriship
- buildpackage /sbuild success
- versioning good, LP bugs redacted from d/changelog

PR #1657, opened 2022-08-15
Holmanb/lxd ig test

holmanb:
```
test: drop erroneous lxd assertion, ensure that result is ok
```

## Additional Context
```
    def validate_storage(validate_client, pkg_name, command):
        log = validate_client.read_from_file("/var/log/cloud-init.log")
>       assert re.search(f"apt-get.*install.*{pkg_name}", log) is not None
E       assert None is not None
```

Since we dropped the forced install test logic in https://github.com/canonical/cloud-init/commit/92c43874cc35e66a240d41cda7ce63c3b427e898, we should not assert that this was done.

https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-gce/79/testReport/junit/tests.integration_tests.modules/test_lxd/test_storage_lvm/

### DIFF
--- a/tests/integration_tests/modules/test_lxd.py
+++ b/tests/integration_tests/modules/test_lxd.py
@@ -50,7 +50,7 @@ class TestLxdBridge:
         verify_clean_log(cloud_init_log)
 
         # The bridge should exist
-        assert class_client.execute("ip addr show lxdbr0")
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
Good after CI pass

PR #1663, opened 2022-08-16
cli: status machine-readable output --format yaml/json

blackboxsw:
## Proposed Commit Message

```
    cli: status machine-readable output --format yaml/json
    
    Provide --format option for cloud-init status and cloud-id to
    surface machine-readable output and detailed boot_status_codes.
    Add a `_schema_version` key to the JSON status output to represent
    whether a schema version has been incremented due to incompatibility.
    
    Preserve a static `schemas`:`1` key which contains a copy of status
    version 1 output which will be preserved even if the default
    `_schema_version` is incremented.
    
    Provide new boot_code_code values in json/yaml output
    which will surface unique codes for the current disabled
    status reasons which gives visibility to whether cloud-init
    may become active on next boot or later in the current boot.
    
    The boot-status-code can be:
     - 'unknown': systemd generators and ds-identify haven't run yet to
                  determine if cloud-init should be run during this boot
     - 'disabled-by-marker-file': /etc/cloud/cloud-init.disabled exists
                  which prevents cloud-init from ever running
     - 'disabled-by-generator': systemd generator ran ds-identify and
                  determined no applicable cloud-init datasources
     - 'disabled-by-kernel-cmdline': kernel cmdline contained
                  cloud-init=disabled
     - 'enabled-by-kernel-cmdline': kernel cmdline contained
                  cloud-init=enabled
     - 'enabled-by-generator': ds-identify detected possible cloud-init
                  datasources
     - 'enabled-by-sysvinit': enabled by default in SysV init environment

    Also deprecate `cloud-id --json` in favor of `cloud-init query v1`.
    
    LP: #1883122
```

## Additional Context
General improvements to machine-readable output to aid in ubuntu live installer (subiquity) and snappy /cloud-init  integration

Expect a separate PR to follow this which will standardize how errors are represented in status JSON output

## Checklist:

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -154,7 +154,7 @@ def handle_status_args(name, args) -> int:
     elif args.format == "json":
         print(
             json.dumps(  # Pretty, sorted json
-                details_dict, indent=1, sort_keys=True, separators=(",", ": ")
+                details_dict, indent=2, sort_keys=True, separators=(",", ": ")
             )
         )
     elif args.format == "yaml":
@@ -207,13 +207,12 @@ def get_status_details(paths=None) -> StatusDetails:
     status = UXAppStatus.NOT_RUN
     errors = []
     datasource = ""
-    description = ""
     status_v1 = {}
 
     status_file = os.path.join(paths.run_dir, "status.json")
     result_file = os.path.join(paths.run_dir, "result.json")
 
-    (boot_status_code, description) = get_bootstatus(
+    boot_status_code, description = get_bootstatus(
         CLOUDINIT_DISABLED_FILE, paths
     )
     if boot_status_code in DISABLED_BOOT_CODES:
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -7,12 +7,15 @@
 """Define 'status' utility and handler as part of cloud-init commandline."""
 
 import argparse
+import copy
 import enum
+import json
 import os
 import sys
 from time import gmtime, sleep, strftime
-from typing import Tuple
+from typing import Any, Dict, List, NamedTuple, Tuple, Union
 
+from cloudinit import safeyaml
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.distros import uses_systemd
 from cloudinit.util import get_cmdline, load_file, load_json
@@ -32,6 +35,43 @@ class UXAppStatus(enum.Enum):
     DISABLED = "disabled"
 
 
+@enum.unique
+class UXAppBootStatusCode(enum.Enum):
+    """Enum representing user-visible cloud-init boot status codes."""
+
+    DISABLED_BY_GENERATOR = "disabled-by-generator"
+    DISABLED_BY_KERNEL_CMDLINE = "disabled-by-kernel-cmdline"
+    DISABLED_BY_MARKER_FILE = "disabled-by-marker-file"
+    ENABLED_BY_GENERATOR = "enabled-by-generator"
+    ENABLED_BY_KERNEL_CMDLINE = "enabled-by-kernel-cmdline"
+    ENABLED_BY_SYSVINIT = "enabled-by-sysvinit"
+    UNKNOWN = "unknown"
+
+
+DISABLED_BOOT_CODES = frozenset(
+    [
+        UXAppBootStatusCode.DISABLED_BY_GENERATOR,
+        UXAppBootStatusCode.DISABLED_BY_KERNEL_CMDLINE,
+        UXAppBootStatusCode.DISABLED_BY_MARKER_FILE,
+    ]
+)
+
+
+class StatusDetails(NamedTuple):
+    status: UXAppStatus
+    boot_status_code: UXAppBootStatusCode
+    description: str
+    errors: List[str]
+    last_update: str
+    datasource: str
+
+
+TABULAR_LONG_TMPL = """\
+boot_status_code: {boot_code}
+{last_update}detail:
+{description}"""
+
+
 def get_parser(parser=None):
     """Build or extend an arg parser for status utility.
 
@@ -45,6 +85,13 @@ def get_parser(parser=None):
         parser = argparse.ArgumentParser(
             prog="status", description="Report run status of cloud init"
         )
+    parser.add_argument(
+        "--format",
+        type=str,
+        choices=["json", "tabular", "yaml"],
+        default="tabular",
+        help="Specify output format for cloud-id (default: tabular)",
+    )
     parser.add_argument(
         "-l",
         "--long",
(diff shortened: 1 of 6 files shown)


TheRealFalcon (commented):
A few drive by comments:

`cloud-init status --format json` is giving me:
```
  File "/usr/bin/cloud-init", line 33, in <module>
    sys.exit(load_entry_point('cloud-init==22.2', 'console_scripts', 'cloud-init')())
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 1088, in main
    retval = util.log_time(
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 2621, in log_time
    ret = func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/status.py", line 127, in handle_status_args
    json.dumps(  # Pretty, sorted json
  File "/usr/lib/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib/python3.10/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/usr/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type UXAppBootStatusCode is not JSON serializable
```

`cloud-id --format json` gives the whole metadata:
```
root@me:~# cloud-id --format json
{
 "_beta_keys": [
  "subplatform"
 ],
 "availability-zone": null,
 "availability_zone": null,
 "cloud-name": "lxd",
 "cloud_id": "lxd",
 "cloud_name": "lxd",
 "distro": "ubuntu",
 "distro_release": "jammy",
 "distro_version": "22.04",
 "instance-id": "0d567396-cd02-484a-ae1a-1ff7dbe93275",
 "instance_id": "0d567396-cd02-484a-ae1a-1ff7dbe93275",
 "kernel_release": "5.15.0-35-generic",
 "local-hostname": "me",
 "local_hostname": "me",
 "machine": "x86_64",
 "platform": "lxd",
 "public_ssh_keys": [],
 "python_version": "3.10.4",
 "region": null,
 "subplatform": "LXD socket API v. 1.0 (/dev/lxd/sock)",
 "system_platform": "Linux-5.15.0-35-generic-x86_64-with-glibc2.35",
 "variant": "ubuntu"
}
```
Is that intentional? Does `cloud-id` need any separate output since it should just be the single name which shouldn't be hard to parse?

Personal opinion, but I think yaml is a bit weird to include. Json seems the goto machine readable format these days whereas yaml is for something that needs to be hand formatted. Any yaml library should be able to read json just fine too.

aciba90 (commented):
> Is that intentional? Does cloud-id need any separate output since it should just be the single name which shouldn't be hard to parse?

@TheRealFalcon: I think the reason is to maintain backward-compatibility with `cloud-id --json`.

In [US015.UC05](https://docs.google.com/document/d/13tWGyheWRQp1LqCwYUUZ-H_JikUx4AMd6VhaLXQwrbs/edit#), it is specified to include versioning and a deprecation warning for `cloud-init status`. What's the strategy about that? Do we plan to include it in a follow-up PR?

blackboxsw:
> > Is that intentional? Does cloud-id need any separate output since it should just be the single name which shouldn't be hard to parse?
> 
> @TheRealFalcon: I think the reason is to maintain backward-compatibility with `cloud-id --json`.
> 
> In [US015.UC05](https://docs.google.com/document/d/13tWGyheWRQp1LqCwYUUZ-H_JikUx4AMd6VhaLXQwrbs/edit#), it is specified to include versioning and a deprecation warning for `cloud-init status`. What's the strategy about that? Do we plan to include it in a follow-up PR?

Good point, deprecation should follow as a separate PR. But, I did want to include versioning scheme in this output. And to @TheRealFalcon's point, yes the `cloud-id --format json` was to align with existing behavior which was surfaced by `cloud-id --json` as Alberto mentioned. Having JSON output for cloud-id is a bit "over the top" and maybe that generally should be deprecated, but there are cases where `cloud-id --long` is not satisfactory for format/white-space parsing the two fields. That said most of what folks should be looking to is using `cloud-init query v1` instead of `cloud-id --json` or `cloud-id --format json` so maybe it's best not to put "good new code" after "bad" in this case and avoid building up the cloud-id utility when cloud-init query should be our primary entry point into looking at cloud-init config and environmental artifacts in a machine-readable format.

aciba90 (approved):
LGTM. Thanks, @blackboxsw, for the clarification.

cloudinit/cmd/status.py
```diff
@@ -69,54 +116,87 @@ def handle_status_args(name, args) -> int:
...
+                    prefix=prefix,
+                    boot_code=details.boot_status_code.value,
+                    description=details.description,
+                    last_update=last_update,
+                )
+            )
+    elif args.format == "json":
+        print(
+            json.dumps(  # Pretty, sorted json
+                details_dict, indent=1, sort_keys=True, separators=(",", ": ")
```
TheRealFalcon, line 157: Why 1? If we want it pretty, I think we should do at least 2 (if not 4).

cloudinit/cmd/status.py
```diff
@@ -125,31 +205,33 @@ def get_status_details(paths=None) -> Tuple[UXAppStatus, str, str]:
...
+    errors = []
+    datasource = ""
+    description = ""
     status_v1 = {}
 
     status_file = os.path.join(paths.run_dir, "status.json")
     result_file = os.path.join(paths.run_dir, "result.json")
 
-    (is_disabled, reason) = _is_cloudinit_disabled(
+    (boot_status_code, description) = get_bootstatus(
```
TheRealFalcon, line 216: Nit: parens aren't needed here on the left side.

cloudinit/cmd/status.py
```diff
@@ -125,31 +205,33 @@ def get_status_details(paths=None) -> Tuple[UXAppStatus, str, str]:
     paths = paths or read_cfg_paths()
 
     status = UXAppStatus.NOT_RUN
-    status_detail = ""
+    errors = []
+    datasource = ""
+    description = ""
```
TheRealFalcon, line 210: Nit: Is there a reason to initialize description here?

TheRealFalcon (approved):
Overall looks good. I left a few inline nits. Only thing I think really needs to be changed is the docs:
```diff
diff --git a/doc/rtd/topics/cli.rst b/doc/rtd/topics/cli.rst
index 83c852759..d9b57acf1 100644
--- a/doc/rtd/topics/cli.rst
+++ b/doc/rtd/topics/cli.rst
@@ -304,7 +304,7 @@ non-zero if an error is detected in cloud-init.
 * ``--long``: detailed status information
 * ``--wait``: block until cloud-init completes
 * ``--format [yaml|json|tabular]``: machine-readable JSON or YAML detailed
-                                    output
+  output
 
 Below are examples of output when cloud-init is running, showing status and
 the currently running modules, as well as when it is done.
```

Without this it renders a little weird. I'll leave it to you to fix and then you can merge.

cloudinit/cmd/status.py
blackboxsw: let's make it an arbitrary 2. easier for those humans to see, and avoids burning too many horizontal white space characters for small terminals.

PR #1664, opened 2022-08-16
add new release versions to schema version file

holmanb:
```
schema: drop release number from version file
```


TheRealFalcon (approved):
Don't forget to update the commit msg

PR #1667, opened 2022-08-16
stages: sync datasource cache after updating metadata

cjp256:
The datasource's caches are set on first boot in init-local.  However,
subsequent boots where metadata is updated for network config, changes
do not persist and the original cache is continued to be used for
applying net names.

Ensure the caches are synced when metadata is refreshed.

### DIFF
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -914,6 +914,9 @@ class Init(object):
             self._apply_netcfg_names(netcfg)
             return
 
+        # sync cache after update
+        self._write_to_cache()
+
         # refresh netcfg after update
         netcfg, src = self._find_networking_config()
 


cjp256:
I'll update the tests when we're sure this is the right fix. :)

cjp256:
I think the fix may warrant inclusion in 22.3 even though it is last minute.

TheRealFalcon:
@cjp256 I'm curious what this does for you. I see that on [line 884](https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L884), we obtain the networking config. In the next too many lines we determine whether we should update metadata or not. If we do [wind up updating metadata](https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L895), that call [should eventually write its received config to the cache](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L373). 

If we didn't update metadata, we shouldn't get to the line your adding. If we did update metadata, writing back to the cache is part of the update.

Is there a case we have missed there that we're not covering?

cjp256:
(this fix would also be useful for 

> @cjp256 I'm curious what this does for you. I see that on [line 884](https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L884), we obtain the networking config. In the next too many lines we determine whether we should update metadata or not. If we do [wind up updating metadata](https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L895), that call [should eventually write its received data back to the cache](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L373).
> 
> If we didn't update metadata, we shouldn't get to the line your adding. If we did update metadata, writing back to the cache is part of the update.
> 
> Is there a case we have missed there that we're not covering?

That netcfg is already stale.  Metadata gets updated immediately afterwards:
https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L895

And then netcfg is refreshed:
https://github.com/canonical/cloud-init/blob/main/cloudinit/stages.py#L918

At this point in time in `init-local`, netcfg is synced, but the cached datasource metadata is stale with the old metadata.  On the subsequent `init` phase, the stale cache is used in the second round of "applying net config names".

```
$ grep -e get_data -e applying -e pkl -e Cloud-init /var/log/cloud-init.log
2022-08-16 19:59:29,057 - util.py[DEBUG]: Cloud-init v. 22.2-0ubuntu1~22.04.3 running 'init-local' at Tue, 16 Aug 2022 19:59:29 +0000. Up 4.95 seconds.
2022-08-16 19:59:29,060 - util.py[DEBUG]: Reading from /var/lib/cloud/instance/obj.pkl (quiet=False)
2022-08-16 19:59:29,061 - util.py[DEBUG]: Read 11368 bytes from /var/lib/cloud/instance/obj.pkl
2022-08-16 19:59:29,162 - util.py[DEBUG]: Writing to /var/lib/cloud/instance/obj.pkl - wb: [400] 11368 bytes
2022-08-16 19:59:29,216 - handlers.py[DEBUG]: start: azure-ds/_get_data: _get_data
2022-08-16 19:59:29,616 - handlers.py[DEBUG]: finish: azure-ds/_get_data: SUCCESS: _get_data
2022-08-16 19:59:29,769 - stages.py[DEBUG]: applying net config names for {'version': 2, 'ethernets': {'eth0': {'dhcp4': True, 'dhcp4-overrides': {'route-metric': 100}, 'dhcp6': False, 'match': {'macaddress': '00:0d:3a:98:8a:e6', 'driver': 'hv_netvsc'}, 'set-name': 'eth0'}, 'eth1': {'dhcp4': True, 'dhcp4-overrides': {'route-metric': 200}, 'dhcp6': False, 'match': {'macaddress': '00:0d:3a:98:8f:15', 'driver': 'hv_netvsc'}, 'set-name': 'eth1'}}}
2022-08-16 19:59:32,831 - util.py[DEBUG]: Cloud-init v. 22.2-0ubuntu1~22.04.3 running 'init' at Tue, 16 Aug 2022 19:59:32 +0000. Up 8.73 seconds.
2022-08-16 19:59:32,863 - util.py[DEBUG]: Reading from /var/lib/cloud/instance/obj.pkl (quiet=False)
2022-08-16 19:59:32,870 - util.py[DEBUG]: Read 11368 bytes from /var/lib/cloud/instance/obj.pkl
2022-08-16 19:59:32,993 - util.py[DEBUG]: Writing to /var/lib/cloud/instance/obj.pkl - wb: [400] 11368 bytes
2022-08-16 19:59:33,043 - stages.py[DEBUG]: applying net config names for {'version': 2, 'ethernets': {'eth0': {'dhcp4': True, 'dhcp4-overrides': {'route-metric': 100}, 'dhcp6': False, 'match': {'macaddress': '00:0d:3a:98:8a:e6', 'driver': 'hv_netvsc'}, 'set-name': 'eth0'}}}
2022-08-16 19:59:33,054 - handlers.py[DEBUG]: start: init-network/consume-user-data: reading and applying user-data
```

First round it found both interfaces in new metadata during `init-local`, applying the correct network names and netplan config.  But cloud-init lost this knowledge in the `init` phase because the cache was never updated.

TheRealFalcon:
> At this point in time in init-local, netcfg is synced, but the cached datasource metadata is stale with the old metadata. On the subsequent init phase, the stale cache is used in the second round of "applying net config names".

Are you saying that the actual metadata being retrieved from the IMDS changes between init-local and init timeframes? If not, I'm still not sure I'm following. AFAIK, `self._apply_netcfg_names(netcfg)` is always preceeded by a `netcfg = self._find_networking_config()`, so I'm not sure how config used to apply net config names can be stale.

cjp256:
> > At this point in time in init-local, netcfg is synced, but the cached datasource metadata is stale with the old metadata. On the subsequent init phase, the stale cache is used in the second round of "applying net config names".
> 
> Are you saying that the actual metadata being retrieved from the IMDS changes between init-local and init timeframes? If not, I'm still not sure I'm following. AFAIK, `self._apply_netcfg_names(netcfg)` is always preceeded by a `netcfg = self._find_networking_config()`, so I'm not sure how config used to apply net config names can be stale.

IIUC the metadata isn't updated (i.e. _get_data() is invoked) until the call to self.datasource.update_metadata_if_supported([event_type]).

This happens after the first call to _find_networking_config() (line 884) but before the second call to _find_networking_config() (line 918).

For the same reason _find_networking_config() had to be invoked again after the metadata update, the cache is now out of date.

Because the cache is not updated, a call to netcfg = self._find_networking_config() in later stages (`init`) will use the stale cached data, not the latest.

blackboxsw:
Thanks for this discussion and proposal. A couple of us have been poking at this today. I agree it looks like the failure path is the following on an Azure system which boots after having a new NIC attached

1. init-local stage on Azure sees networking not yet up and gets into the `event_enabled_and_metadata_updated(EventType.BOOT)` which updates metadata via another get_data crawl which clears the `DSAzure._network_config` attribute
2. init-local stage then subsequently calls `_find_networking_config()` which calls `DSAzure.network_config` which sees the `_unset`  value for _network_config and reconstitutes that value from updated metadata...
3. but init-local doesn't persist that update at that point
4. init-net reloads the cache and still automatically applies the renames using stale data as you said.

What I was initiall concerned with is only trying to _write_to_cache the obj.pkl when metadata updates via something like this:
```diff
diff --git a/cloudinit/stages.py b/cloudinit/stages.py
index 66e12eed2..5bed4621c 100644
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -887,12 +887,19 @@ class Init(object):
             return
 
         def event_enabled_and_metadata_updated(event_type):
-            return update_event_enabled(
+            metadata_updated = False
+            if update_event_enabled(
                 datasource=self.datasource,
                 cfg=self.cfg,
                 event_source_type=event_type,
                 scope=EventScope.NETWORK,
-            ) and self.datasource.update_metadata_if_supported([event_type])
+            ):
+                metadata_updated = (
+                    self.datasource.update_metadata_if_supported([event_type])
+                )
+                if metadata_updated:
+                    self._write_to_cache()
+            return metadata_updated
 
         def should_run_on_boot_event():
             return (
```


Because we don't want to always continue to re-write obj.pkl when nothing has changed.
James thought of a good idea to actually pull this into the datasource which we are talking about in IRC nhow

cjp256:
thanks! Closing PR in favor of alternative proposal.

PR #1671, opened 2022-08-18
sources/azure: don't enforce nic ordering

cjp256:
The cloud-init generated netplan config persists between boots and the
names are applied early in boot by the netplan generator.  This may
result in the primary NIC being renamed away from "eth0", causing
cloud-init local to fail.

The ordering of NICs provided by IMDS may not match the order
enumerated by kernel, this will cause additional renaming thrashing
which tends to result in different types of failures with networkd
on Ubuntu 18.04, 20.04, and 22.04.

Stop setting device names until we can guarantee that:
(a) we can find the primary NIC regardless if it is eth0
(b) Ubuntu doesn't fail to bring up networking

Without ordering reliability, we cannot rely on checking the interface
device driver as named, so add a helper to reliably identify the most
appropriate matching.

LP #1958280

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -22,7 +22,6 @@ from cloudinit import dmi
 from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
-from cloudinit.net import device_driver
 from cloudinit.net.dhcp import (
     NoDHCPLeaseError,
     NoDHCPLeaseInterfaceError,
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -208,6 +207,33 @@ def get_hv_netvsc_macs_normalized() -> List[str]:
     ]
 
 
+@azure_ds_telemetry_reporter
+def determine_device_driver_for_mac(mac: str) -> Optional[str]:
+    """Determine the device driver to match on, if any."""
+    interfaces = [
+        i for i in net.get_interfaces() if mac == normalize_mac_address(i[1])
+    ]
+    drivers = [i[2] for i in interfaces if i[2] not in BLACKLIST_DRIVERS]
```
holmanb, line 216: Driveby comment: Could we use the `blacklist_drivers` kwarg in `get_interfaces()` instead of manually filtering?

holmanb: Also I'd like to see `get_interfaces()` return a named tuple at some point so this is all more readable, but that's obviously not your problem or something that needs to be addressed in this PR.

cjp256: > Driveby nitpick: Could we use the `blacklist_drivers` kwarg in `get_interfaces()` instead of manually filtering?

I had done that originally but then I preferred the clarity in the unittest scenarios when mocking return data for get_interfaces().  But now that i've rounded out the scenarios I can go back and clean it up.

cjp256:
Rebased/requires #1672

cloudinit/sources/DataSourceAzure.py
cjp256: Fixed over in #1672

PR #1672, opened 2022-08-18
sources/azure: fix device driver matching for net config

cjp256:
The ordering of NICs provided by IMDS may not match the order enumerated
by kernel.  As such, we do not have any guarantee that the nic we're
checking the driver for is the nic we think it is.

Instead of making any assumptions about how the nics are named, check
all interfaces by mac address.  If there is an interface using
"hv_netvsc", match against that.  If there is only one interface driver
that is not blacklisted, use that (in case it is not "hv_netvsc"), but
log a debug event.  If there are multiple hits, don't match against any
of the names and report a warning.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -22,7 +22,6 @@ from cloudinit import dmi
 from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
-from cloudinit.net import device_driver
 from cloudinit.net.dhcp import (
     NoDHCPLeaseError,
     NoDHCPLeaseInterfaceError,
@@ -206,6 +205,33 @@ def get_hv_netvsc_macs_normalized() -> List[str]:
         for n in net.get_interfaces()
         if n[2] == "hv_netvsc"
     ]
+
+
+@azure_ds_telemetry_reporter
+def determine_device_driver_for_mac(mac: str) -> Optional[str]:
+    """Determine the device driver to match on, if any."""
+    drivers = [
+        i[2]
+        for i in net.get_interfaces(blacklist_drivers=BLACKLIST_DRIVERS)
+        if mac == normalize_mac_address(i[1])
+    ]
+    if "hv_netvsc" in drivers:
+        return "hv_netvsc"
+
+    if len(drivers) == 1:
+        report_diagnostic_event(
+            "Assuming driver for interface with mac=%s drivers=%r"
+            % (mac, drivers),
+            logger_func=LOG.debug,
+        )
+        return drivers[0]
+
+    report_diagnostic_event(
+        "Unable to specify driver for interface with mac=%s drivers=%r"
+        % (mac, drivers),
+        logger_func=LOG.warning,
+    )
+    return None
 
 
 def execute_or_debug(cmd, fail_ret=None) -> str:
@@ -2038,11 +2064,8 @@ def generate_network_config_from_instance_network_metadata(
             dev_config.update(
                 {"match": {"macaddress": mac.lower()}, "set-name": nicname}
             )
-            # With netvsc, we can get two interfaces that
-            # share the same MAC, so we need to make sure
-            # our match condition also contains the driver
-            driver = device_driver(nicname)
-            if driver and driver == "hv_netvsc":
+            driver = determine_device_driver_for_mac(mac)
+            if driver:
                 dev_config["match"]["driver"] = driver
             netconfig["ethernets"][nicname] = dev_config
             continue
--- a/tests/unittests/sources/test_azure.py
+++ b/tests/unittests/sources/test_azure.py
@@ -93,16 +93,6 @@ def mock_chassis_asset_tag():
 
 
 @pytest.fixture
-def mock_device_driver():
-    with mock.patch(
-        MOCKPATH + "device_driver",
-        autospec=True,
-        return_value=None,
-    ) as m:
-        yield m
-
-
-@pytest.fixture
 def mock_generate_fallback_config():
     with mock.patch(
         MOCKPATH + "net.generate_fallback_config",
@@ -173,9 +163,19 @@ def mock_net_dhcp_EphemeralIPv4Network():
         yield m
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 def mock_get_interfaces():
-    with mock.patch(MOCKPATH + "net.get_interfaces", return_value=[]) as m:
+    with mock.patch(
+        MOCKPATH + "net.get_interfaces",
+        return_value=[
+            ("dummy0", "9e:65:d6:19:19:01", None, None),
+            ("enP3", "00:11:22:33:44:02", "unknown_accel", "0x3"),
+            ("eth0", "00:11:22:33:44:00", "hv_netvsc", "0x3"),
+            ("eth2", "00:11:22:33:44:01", "unknown", "0x3"),
+            ("eth3", "00:11:22:33:44:02", "unknown_with_unknown_vf", "0x3"),
+            ("lo", "00:00:00:00:00:00", None, None),
+        ],
+    ) as m:
         yield m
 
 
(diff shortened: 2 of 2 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -208,6 +207,33 @@ def get_hv_netvsc_macs_normalized() -> List[str]:
     ]
 
 
+@azure_ds_telemetry_reporter
+def determine_device_driver_for_mac(mac: str) -> Optional[str]:
+    """Determine the device driver to match on, if any."""
+    drivers = [
+        i[2]
+        for i in net.get_interfaces(blacklist_drivers=BLACKLIST_DRIVERS)
```
blackboxsw, line 215: Is there some way we can hang this off of DataSourceAzure and call `self.distro.networking.get_interfaces_by_mac()` here which should already account for blacklist_drivers? We are trying to move general networking operations into the `distro.networking` as the primary entry point as different distros will have unique APIs or filesystems to get at that device content. Eventually I think we want to retire cloudinit.net functions in favor of distro.networking driven API.

blackboxsw: A note as well, we will be modernizing our variable names in the future to be something more inclusive and blacklist will likely be replaced with `denylist` throughout cloud-init in the future. so expect a couple of merge conflicts if that PR lands before yours.

cjp256: Will fix & revisit after 22.3

cjp256: > Is there some way we can hang this off of DataSourceAzure and call `self.distro.networking.get_interfaces_by_mac()` here which should already account for blacklist_drivers? We are trying to move general networking operations into the `distro.networking` as the primary entry point as different distros will have unique APIs or filesystems to get at that device content. Eventually I think we want to retire cloudinit.net functions in favor of distro.networking driven API.

I looked at get_interfaces_by_mac, but it doesn't include the driver afaict.

blackboxsw:
I meant distros.networking.Networking.get_interfaces_by_mac, which does
include Azure's blacklist_drivers
https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/networking.py#L74-L78

cloudinit/sources/DataSourceAzure.py
blackboxsw: I meant [cloudinit.distros.networking.Networking.get_interfaes_by_mac](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/networking.py#L74-L78) which includes [Azure blacklist_drivers settings](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L712)

cjp256: > I meant [cloudinit.distros.networking.Networking.get_interfaes_by_mac](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/networking.py#L74-L78) which observes [Azure blacklist_drivers settings](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L712)

Yeah, it filters, but I need the actual driver for the interface too, not just the name (which is all that is contained within the dict).  If we can get it to return that namedtuple or whatever you were thinking of that would include the driver name, that'd work perfectly.

holmanb (approved):
I'll leave unmerged in case @blackboxsw has more comments. Looks good to me.

cjp256:
@blackboxsw are you ok with it as-is?  I'll leave the denylist renaming for the project-wide update.

cjp256:
With the promise that I'll come back and use networking.get_interfaces_by_mac() once it supplies the driver names :D

cloudinit/sources/DataSourceAzure.py
blackboxsw: @cjp256 thanks I had misinterpreted the output format from get_interfaces_by_mac. You are correct in this PR, there is no better place at the moment than to call the `net.get_interfaces()`. I was going to pick at the repeated calls to `net.get_interfaces()` here per device representing in IMDS as "costly" to walk /sys/class/net devices on bigger systems. But, from a wall clock time cost, it seems to be ~4 milliseconds per call. so it's not going to be noticeable in the grand scheme of things. Note, this call-site may migrate into the distro.networking classes in the future just like [cloudinit.distros.networking.Networking.get_interfaces_by_mac](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/networking.py#L74-L78)  per the [network/distro refactor work](https://github.com/canonical/cloud-init/blob/main/WIP-ONGOING-REFACTORIZATION.rst).

cloudinit/sources/DataSourceAzure.py
```diff
@@ -208,6 +207,33 @@ def get_hv_netvsc_macs_normalized() -> List[str]:
...
+        if mac == normalize_mac_address(i[1])
```
blackboxsw, line 216: @cjp256 I don't think we need to call normalize_mac_address here on the return value from net.get_interfaces() right?
The output comes from /sys/class/net/<device>/address which should already be lowercase colon-delimited format.
```bash
$ python3 -c 'from cloudinit.net import get_interfaces; print(get_interfaces())'
[('lo', '00:00:00:00:00:00', None, None), ('eth0', '00:22:48:a6:0a:5c', 'hv_netvsc', '0x3')]
```


```suggestion
        if mac == i[1]
```

blackboxsw (changes requested):
@cjp256 Totally minor change:
 - drop the `normalize_mac_address calls` from determine_device_driver_for_mac

At some point in the future, I'd like us to think about lru_cache of `net.get_interfaces`, but there are implications for hotplug clearing that cache if we have a persistent python daemon around fielding device add/removes so I don't want to add that in this PR.

cloudinit/sources/DataSourceAzure.py
blackboxsw: As you mentioned a query in BSD over IRC. I poked around at a BSD 13.1 instance, and making the call to get_interfaces() may not have BSD support.
```bash
freebsd@bsd-131:~/cloud-init $ /usr/local/bin/python3 -c 'from cloudinit.net import get_interfaces; print(get_interfaces())'
[]
freebsd@bsd-131:~/cloud-init $ /usr/local/bin/python3 -c 'from cloudinit.net import get_interfaces; print(get_interfaces())'
```

Checking alternatives

cjp256:
Can this PR be re-opened @blackboxsw @holmanb?  Slipped through the cracks here.

TheRealFalcon:
@cjp256  It actually can't. Github has the reopen button greyed out. Want to open a new one and reference this one?

cloudinit/sources/DataSourceAzure.py
cjp256: I agree get_interfaces() does not work on BSD but in this case it doesn't matter much.

(a) the blacklist_drivers used are very specific to Linux
(b) get_interfaces_by_mac() is already broken on BSD for Azure accelerated networking The mac of the accelerated interfaces clobbers the "master" interface, so networking never comes up properly.

```
>>> net.get_devicelist()
['mce0']
>>> net.get_interfaces_by_mac()
{'00:0d:3a:54:ad:1e': 'mce0'}
>>>
[root@test-freebsd131-x1 /usr/home/cpatterson]# ifconfig -a
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> metric 0 mtu 16384
        options=680003<RXCSUM,TXCSUM,LINKSTATE,RXCSUM_IPV6,TXCSUM_IPV6>
        inet6 ::1 prefixlen 128
        inet6 fe80::1%lo0 prefixlen 64 scopeid 0x1
        groups: lo
        nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
hn0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=8051b<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,TSO4,LRO,LINKSTATE>
        ether 00:0d:3a:54:ad:1e
        inet 10.0.0.35 netmask 0xffffff00 broadcast 10.0.0.255
        media: Ethernet 100GBase-CR4 <full-duplex,rxpause,txpause>
        status: active
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
mce0: flags=8a43<UP,BROADCAST,RUNNING,ALLMULTI,SIMPLEX,MULTICAST> metric 0 mtu 1500
        options=8805bb<RXCSUM,TXCSUM,VLAN_MTU,VLAN_HWTAGGING,JUMBO_MTU,VLAN_HWCSUM,TSO4,LRO,LINKSTATE>
        ether 00:0d:3a:54:ad:1e
        media: Ethernet 100GBase-CR4 <full-duplex,rxpause,txpause>
        status: active
        nd6 options=29<PERFORMNUD,IFDISABLED,AUTO_LINKLOCAL>
```

I intend to purge blacklist_drivers from net/ and constrain this logic to Azure's data source.  Beyond that, to get it right, I think we'll need the BSD networking refactor to be done?

cjp256: Thoughts @igalic on the BSD angle?

igalic: @cjp256 I've been working on getting blacklist driver support into BSD… 
but
are you saying that the blacklisted drivers are only relevant on Linux and their equivalent works on BSDs?

i'm also working on better get_interfaces and get_interfaces_by_mac into BSD, since obviously, that first shot using dictionaries wasn't great

the code you're demoing here isn't mine tho, and doesn't work much better

mine can be found at distros.bsd.networking.get_interfaces_by_mac()

PR #1685, opened 2022-08-23
Networkd multi-address support/fix

garzdin:
Networkd multi-address support/fix

```
Given a cloud-init network config with multiple addresses for an interface the networkd backend outputs an invalid networkd service config. According to the `man` page of `systemd-networkd` only one `Address=` field per `[Address]` section is allowed.
```

## Additional Context
[Ref: https://www.freedesktop.org/software/systemd/man/systemd.network.html#Address=1](https://www.freedesktop.org/software/systemd/man/systemd.network.html#Address=1)
[Ref: https://man.archlinux.org/man/systemd.network.5#%5BADDRESS%5D_SECTION_OPTIONS](https://man.archlinux.org/man/systemd.network.5#%5BADDRESS%5D_SECTION_OPTIONS)

Older versions of `cloud-init` (e.g 19.4) output a valid config.

File location: `/etc/systemd/network/`

The invalid config leads to only one of the IPs being configured for the interface, since the second one is ignored. This can lead to communication issues between systems (e.g Kubernetes).

## Test Steps
Provision two systems that use `systemd-networkd` to configure network interfaces. Install 19.4 on one and 22.3 on the other. In the `cloud-init` network configuration specify 2 IPs for an interface. On the machine provisioned by `cloud-init` 19.4 the interface is correctly configured. On the machine provisioned by `cloud-init` 22.3 the interface only gets assigned the first IP from the `[Address]` section.

## Checklist:

### DIFF
--- a/cloudinit/net/networkd.py
+++ b/cloudinit/net/networkd.py
@@ -44,10 +44,16 @@ class CfgParser:
         for k, v in sorted(self.conf_dict.items()):
             if not v:
                 continue
-            contents += "[" + k + "]\n"
-            for e in sorted(v):
-                contents += e + "\n"
-            contents += "\n"
+            if k == "Address":
+                for e in sorted(v):
+                    contents += "[" + k + "]\n"
+                    contents += e + "\n"
+                    contents += "\n"
+            else:
+                contents += "[" + k + "]\n"
+                for e in sorted(v):
+                    contents += e + "\n"
+                contents += "\n"
 
         return contents
 
--- a/tests/unittests/net/test_networkd.py
+++ b/tests/unittests/net/test_networkd.py
@@ -12,6 +12,7 @@ network:
     eth0:
       match:
         macaddress: '00:11:22:33:44:55'
+      addresses["172.16.10.2/12" "172.16.10.3/12"]
       nameservers:
         search: [spam.local, eggs.local]
         addresses: [8.8.8.8]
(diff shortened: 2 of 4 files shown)


TheRealFalcon:
Closing and re-opening to retrigger CI

TheRealFalcon:
Thanks @garzdin  for this fix. It looks good to merge and I will merge if/when CI passes.

I'm a little confused though:
> Provision two systems that use systemd-networkd to configure network interfaces. Install 19.4 on one and 22.3 on the other.

Explicit support for systemd-networkd landed in cloud-init in 21.3. Prior to 21.3, the only way networkd was used was via netplan. If you aren't using netplan, I'm curious as to how you were using networkd in cloud-init 19.4. If you are using netplan (default on Ubuntu), your system should continue to use netplan and should not go down this code path. If this is the case, this is likely a larger bug in how we handle choosing a network renderer. It would helpful if you could file a bug report at https://bugs.launchpad.net/cloud-init and provide the tarball provided by running `cloud-init collect-logs` on your affected 19.4 vs 22.3 systems.

TheRealFalcon (approved):

garzdin:
@TheRealFalcon Sorry for the confusion. I didn't specify that, yes - in fact older versions used netplan, which is what I meant was working fine. Then when the change over to the networkd backend happened it seems like it broke this specific functionality.

Thank you for the approval!

PR #1686, opened 2022-08-23
Add 3DS Outscale Datasource

outscale-mdr:
Hello 👋,

This PR is to add the DataSource of 3DS Outscale. To help you review this PR here what we have done. 

The datasource is based EC2 datasource but with one big difference : the retrieval of ` tags`. Indeed, in the meta-data server, in the tags section, we (3DS Outscale) accept all characters and this can lead to cloud-init crash during the retrieval of meta-data. AWS does not have this problem because if the tags are in the meta-data server, only a set of [characters is available ](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions).

Instead of duplicating the code and fix the behavior for our need, we think it would be better to handle this directly in the ec2 helpers. Therefore we have added the ability to skip some element in the meta-data retrieval if necessary.

This can be achieved with this code:
```python
def _skip_tags(exception):
    if exception.code == 404:
        if "meta-data/tags/" in exception.url:
            print(exception.url)
            return True
    return False
data = ec2.get_instance_metadata(
                headers_cb=None,
                skip_cb=_skip_tags
            )
```

Tell me what do you think and I hope our datasource will be merged.
Thank you,
Best regards,
Maxime

### DIFF
--- a/cloudinit/sources/helpers/ec2.py
+++ b/cloudinit/sources/helpers/ec2.py
@@ -190,6 +190,7 @@ def _get_instance_metadata(
     headers_cb=None,
     headers_redact=None,
     exception_cb=None,
+    retrieval_exception_ignore_cb=None,
 ):
     md_url = url_helper.combine_url(metadata_address, api_version, tree)
     caller = functools.partial(
@@ -203,7 +204,17 @@ def _get_instance_metadata(
     )
 
     def mcaller(url):
-        return caller(url).contents
+        try:
+            return caller(url).contents
+        except url_helper.UrlError as e:
+            if (
+                not retrieval_exception_ignore_cb
+                or not retrieval_exception_ignore_cb(e)
+            ):
+                raise
+            else:
+                LOG.warning("Skipped retrieval of the content of %s", base_url)
+                return "(skipped)"
 
     try:
         response = caller(md_url)
@@ -229,6 +240,7 @@ def get_instance_metadata(
     headers_cb=None,
     headers_redact=None,
     exception_cb=None,
+    retrieval_exception_ignore_cb=None,
 ):
     # Note, 'meta-data' explicitly has trailing /.
     # this is required for CloudStack (LP: #1356855)
@@ -243,6 +255,7 @@ def get_instance_metadata(
         headers_redact=headers_redact,
         headers_cb=headers_cb,
         exception_cb=exception_cb,
+        retrieval_exception_ignore_cb=retrieval_exception_ignore_cb,
     )
 
 
--- a/cloudinit/sources/helpers/ec2.py
+++ b/cloudinit/sources/helpers/ec2.py
@@ -213,7 +213,7 @@ def _get_instance_metadata(
             ):
                 raise
             else:
-                LOG.warning("Skipped retrieval of the content of %s", base_url)
+                LOG.warning("Skipped retrieval of the content of %s", url)
                 return "(skipped)"
 
     try:
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1135,6 +1135,13 @@ ec2_identify_platform() {
         e24cloud) _RET="E24cloud"; return 0;;
     esac
 
+
+    local product_name="${DI_DMI_PRODUCT_NAME}"
+    if [ "${product_name}" = "3DS Outscale VM" ]  && \
+       [ "${vendor}" = "3DS Outscale" ]; then
+        _RET="Outscale"; return 0
+    fi
+
     # AWS http://docs.aws.amazon.com/AWSEC2/
     #     latest/UserGuide/identify_ec2_instances.html
     local uuid="" hvuuid="${PATH_SYS_HYPERVISOR}/uuid"
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -48,6 +48,7 @@ KNOWN_CLOUD_NAMES = [
     "VMware",
     "Vultr",
     "ZStack",
+    "Outscale",
     "Other",
 ]
 
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -43,12 +43,18 @@ class CloudNames(object):
     BRIGHTBOX = "brightbox"
     ZSTACK = "zstack"
     E24CLOUD = "e24cloud"
+    OUTSCALE = "outscale"
     # UNKNOWN indicates no positive id.  If strict_id is 'warn' or 'false',
     # then an attempt at the Ec2 Metadata service will be made.
     UNKNOWN = "unknown"
     # NO_EC2_METADATA indicates this platform does not have a Ec2 metadata
     # service available. No attempt at the Ec2 Metadata service will be made.
     NO_EC2_METADATA = "no-ec2-metadata"
+
+
+# Drop when LP: #1988157 tag handling is fixed
+def skip_404_tag_errors(exception):
+    return exception.code == 404 and "meta-data/tags/" in exception.url
 
 
 class DataSourceEc2(sources.DataSource):
@@ -527,8 +533,12 @@ class DataSourceEc2(sources.DataSource):
         if self.cloud_name == CloudNames.AWS:
             exc_cb = self._refresh_stale_aws_token_cb
             exc_cb_ud = self._skip_or_refresh_stale_aws_token_cb
-        else:
+            skip_cb = None
+        elif self.cloud_name == CloudNames.OUTSCALE:
             exc_cb = exc_cb_ud = None
+            skip_cb = skip_404_tag_errors
+        else:
+            exc_cb = exc_cb_ud = skip_cb = None
         try:
             crawled_metadata["user-data"] = ec2.get_instance_userdata(
                 api_version,
@@ -543,6 +553,7 @@ class DataSourceEc2(sources.DataSource):
                 headers_cb=self._get_headers,
                 headers_redact=redact,
                 exception_cb=exc_cb,
+                retrieval_exception_ignore_cb=skip_cb,
             )
             if self.cloud_name == CloudNames.AWS:
                 identity = ec2.get_instance_identity(
@@ -667,7 +678,7 @@ class DataSourceEc2Local(DataSourceEc2):
     perform_dhcp_setup = True  # Use dhcp before querying metadata
 
     def get_data(self):
-        supported_platforms = (CloudNames.AWS,)
+        supported_platforms = (CloudNames.AWS, CloudNames.OUTSCALE)
         if self.cloud_name not in supported_platforms:
             LOG.debug(
                 "Local Ec2 mode only supported on %s, not %s",
@@ -757,6 +768,14 @@ def identify_e24cloud(data):
         return CloudNames.E24CLOUD
 
 
+def identify_outscale(data):
+    if (
+        data["product_name"] == "3DS Outscale VM".lower()
+        and data["vendor"] == "3DS Outscale".lower()
+    ):
+        return CloudNames.OUTSCALE
+
+
 def identify_platform():
     # identify the platform and return an entry in CloudNames.
     data = _collect_platform_data()
@@ -765,6 +784,7 @@ def identify_platform():
         identify_brightbox,
         identify_zstack,
         identify_e24cloud,
+        identify_outscale,
         lambda x: CloudNames.UNKNOWN,
     )
     for checker in checks:
@@ -787,6 +807,7 @@ def _collect_platform_data():
        serial: dmi 'system-serial-number' (/sys/.../product_serial)
        asset_tag: 'dmidecode -s chassis-asset-tag'
        vendor: dmi 'system-manufacturer' (/sys/.../sys_vendor)
+       product_name: dmi 'system-product-name' (/sys/.../system-manufacturer)
 
     On Ec2 instances experimentation is that product_serial is upper case,
     and product_uuid is lower case.  This returns lower case values for both.
@@ -817,6 +838,9 @@ def _collect_platform_data():
 
     vendor = dmi.read_dmi_data("system-manufacturer")
     data["vendor"] = (vendor if vendor else "").lower()
+
+    product_name = dmi.read_dmi_data("system-product-name")
+    data["product_name"] = (product_name if product_name else "").lower()
 
     return data
 
--- a/doc/rtd/topics/availability.rst
+++ b/doc/rtd/topics/availability.rst
@@ -60,6 +60,7 @@ environments in the public cloud:
 - UpCloud
 - Vultr
 - Zadara Edge Cloud Platform
+- 3DS Outscale
 
 Additionally, cloud-init is supported on these private clouds:
 
--- a/tests/unittests/sources/helpers/test_ec2.py
+++ b/tests/unittests/sources/helpers/test_ec2.py
@@ -273,5 +273,55 @@ class TestEc2Util(helpers.HttprettyTestCase):
         self.assertEqual(iam["info"]["LastUpdated"], "2016-10-27T17:29:39Z")
         self.assertNotIn("security-credentials", iam)
 
+    def test_metadata_children_with_invalid_character(self):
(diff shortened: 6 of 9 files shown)


doc/rtd/topics/datasources/outscale.rst
```diff
@@ -0,0 +1,101 @@
+.. _datasource_outscale:
+
+3DS Outscale (Outscale)
+======================
```
TheRealFalcon, line 4: Needs an extra `=`

doc/rtd/topics/datasources/outscale.rst
```diff
@@ -0,0 +1,101 @@
...
+    iam/
+    services/
+
+Userdata
+^^^^^^^^
+If provided, user-data will appear at
+``http://169.254.169.254/2016-09-02/user-data``.
+If no user-data is provided, this will return an empty string.
+
+.. code-block:: shell-session
```
TheRealFalcon, line 95: Does this block belong here? I don't understand what it's trying to demonstrate.

tools/ds-identify
```diff
@@ -1487,6 +1487,13 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+dscheck_Outscale() {
+    local product_name="${DI_DMI_PRODUCT_NAME}"
+    local sys_vendor="${DI_DMI_SYS_VENDOR}"
+    test "${product_name}" = "3DS Outscale VM"  &&  test "${sys_vendor}" = "3DS Outscale" && return ${DS_FOUND}
```
TheRealFalcon, line 1493: Just for consistency's sake, can we use `if` instead of test here? Also quote `${NOT_FOUND}` as well as `${DS_NOT_FOUND}` on the next line.

TheRealFalcon (dismissed):
Thanks for the PR @outscale-mdr . New datasources are always welcome into cloud-init.

I left some inline comments. Additionally, the documentation you added looks good, but you should add your datasource to `doc/rtd/topcis/availability.rst` and a link to the page you added in `doc/rtd/topics/datasources.rst`.

We enforce some style and flaking rules. You can check the Github Actions failures or run the default tox env (`tox` without arguments) to run all the checks locally. Those will need to be fixed up before this PR can be merged.

Finally, before we can accept your change, you need to sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . The full details are described [in the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html#submitting-your-first-pull-request)

outscale-mdr:
Thank you @TheRealFalcon for the review, I will make the corrections soon but before I need internal approvals for the CLA :)

cloudinit/sources/helpers/ec2.py
```diff
@@ -190,6 +190,7 @@ def _get_instance_metadata(
     headers_cb=None,
     headers_redact=None,
     exception_cb=None,
+    skip_cb=None
```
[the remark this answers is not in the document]
outscale-mdr: Indeed, I changed to `retrieval_exception_ignore_cb` to explain that it is a more generic cb not only for tags but for whatever the user want

cloudinit/sources/helpers/ec2.py
```diff
@@ -203,7 +204,14 @@ def _get_instance_metadata(
     )
 
     def mcaller(url):
-        return caller(url).contents
+        try:
+            return caller(url).contents
+        except Exception as e:
```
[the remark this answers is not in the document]
outscale-mdr: Indeed, I will change

doc/rtd/topics/datasources/outscale.rst
```diff
@@ -0,0 +1,101 @@
+.. _datasource_outscale:
+
+3DS Outscale (Outscale)
+======================
+The ``Outscale`` datasource reads data from 3DS Outscale.  Support is
+included in cloud-init since 0.7.9.
```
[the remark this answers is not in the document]
outscale-mdr: A mistake, 22.4 it is :)

doc/rtd/topics/datasources/outscale.rst
outscale-mdr: Done

doc/rtd/topics/datasources/outscale.rst
outscale-mdr: Yes, it just show how to retrieve the `user-data` from the meta-data server

tools/ds-identify
outscale-mdr: You're right.

doc/rtd/topics/datasources/outscale.rst
TheRealFalcon: Oh, so the user data is a bash script? Would it make more sense to show a curl command retrieving it?

blackboxsw:
@outscale-mdr hanks for the PR. Sorry for the delay here... before we proceed, I want to also double check that we really need a specialized/separate datasource in this case vs just fixing behavior in the DataSourceEC2 to be more resilient to handling your tag-related differentiation.  EC2 and related subclasses of the datasource falls into this camp of sometimes it makes sense to separate and sometimes not.

Anytime we add a separate datasource in cloud-init there is a tension because some environments don't have access to use `ds-identify` at systemd generator timeframe.  On platforms without ds-identify `cloud-init init --local` will serially walk through all possible python datasources configured in order, calling the respective `DatasourceX.get_data` methods to determine whether or not the datasource is valid for the platform. This ends up being more costly than if we have a single Ec2 datasource that is flexible to "rule them all" and properly identify separate clouds by the right name.

It may make sense to keep your current approach, but I wanted to give you a heads up in case you were actively pushing changes to this branch that I'll look at whether this can be integrated into the current Ec2 behavior and comment within an hour here.

blackboxsw:
@outscale-mdr @TheRealFalcon does the following approach seem reasonable for your cloud (applying this diff to this current PR)
```diff
diff --git a/cloudinit/sources/DataSourceEc2.py b/cloudinit/sources/DataSourceEc2.py
index 32659848a..577ef5a38 100644
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -43,6 +43,7 @@ class CloudNames(object):
     BRIGHTBOX = "brightbox"
     ZSTACK = "zstack"
     E24CLOUD = "e24cloud"
+    OUTSCALE = "outscale"
     # UNKNOWN indicates no positive id.  If strict_id is 'warn' or 'false',
     # then an attempt at the Ec2 Metadata service will be made.
     UNKNOWN = "unknown"
@@ -51,6 +52,10 @@ class CloudNames(object):
     NO_EC2_METADATA = "no-ec2-metadata"
 
 
+def skip_404_tag_errors(exception):
+    return exception.code == 404 and "meta-data/tags/" in exception.url
+
+
 class DataSourceEc2(sources.DataSource):
 
     dsname = "Ec2"
@@ -527,8 +532,12 @@ class DataSourceEc2(sources.DataSource):
         if self.cloud_name == CloudNames.AWS:
             exc_cb = self._refresh_stale_aws_token_cb
             exc_cb_ud = self._skip_or_refresh_stale_aws_token_cb
-        else:
+            skip_cb = None
+        elif self.cloud_name == CloudNames.Outscale:
             exc_cb = exc_cb_ud = None
+            skip_cb = skip_404_tag_errors
+        else:
+            exc_cb = exc_cb_ud = skip_cb = None
         try:
             crawled_metadata["user-data"] = ec2.get_instance_userdata(
                 api_version,
@@ -543,6 +552,7 @@ class DataSourceEc2(sources.DataSource):
                 headers_cb=self._get_headers,
                 headers_redact=redact,
                 exception_cb=exc_cb,
+                retrieval_exception_ignore_cb=skip_cb,
             )
             if self.cloud_name == CloudNames.AWS:
                 identity = ec2.get_instance_identity(
@@ -667,7 +677,7 @@ class DataSourceEc2Local(DataSourceEc2):
     perform_dhcp_setup = True  # Use dhcp before querying metadata
 
     def get_data(self):
-        supported_platforms = (CloudNames.AWS,)
+        supported_platforms = (CloudNames.AWS, CloudNames.OUTSCALE)
         if self.cloud_name not in supported_platforms:
             LOG.debug(
                 "Local Ec2 mode only supported on %s, not %s",
@@ -757,6 +767,14 @@ def identify_e24cloud(data):
         return CloudNames.E24CLOUD
 
 
+def identify_outscale(data):
+    if (
+        data["product_name"] == "3DS Outscale VM"
+        and data["vendor"] == "3DS Outscale"
+    ):
+        return CloudNames.OUTSCALE
+
+
 def identify_platform():
     # identify the platform and return an entry in CloudNames.
     data = _collect_platform_data()
@@ -765,6 +783,7 @@ def identify_platform():
         identify_brightbox,
         identify_zstack,
         identify_e24cloud,
+        identify_outscale,
         lambda x: CloudNames.UNKNOWN,
     )
     for checker in checks:
@@ -818,6 +837,9 @@ def _collect_platform_data():
     vendor = dmi.read_dmi_data("system-manufacturer")
     data["vendor"] = (vendor if vendor else "").lower()
 
+    product_name = dmi.read_dmi_data("system-product-name")
+    data["prouct_name"] = (product_name if product_name else "").lower()
+
     return data
 
 
diff --git a/tools/ds-identify b/tools/ds-identify
index 3f4965629..acc2c5b37 100755
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1135,6 +1135,13 @@ ec2_identify_platform() {
         e24cloud) _RET="E24cloud"; return 0;;
     esac
 
+
+    local product_name="${DI_DMI_PRODUCT_NAME}"
+    if [ "${product_name}" = "3DS Outscale VM" ]  && \
+       [ "${vendor}" = "3DS Outscale" ]; then
+        _RET="Outscale"; return 0
+    fi
+
     # AWS http://docs.aws.amazon.com/AWSEC2/
     #     latest/UserGuide/identify_ec2_instances.html
     local uuid="" hvuuid="${PATH_SYS_HYPERVISOR}/uuid"
@@ -1487,16 +1494,6 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
-dscheck_Outscale() {
-    local product_name="${DI_DMI_PRODUCT_NAME}"
-    local sys_vendor="${DI_DMI_SYS_VENDOR}"
-    if [ "${product_name}" = "3DS Outscale VM" ]  && \
-       [ "${sys_vendor}" = "3DS Outscale" ]; then
-        return "${DS_FOUND}"
-    fi
-    return "${DS_NOT_FOUND}"
-}
-
 collect_info() {
     read_uname_info
     read_virt

```
??? 
Something like this should give you the ability to use the current DataSourceEc2 (maybe even extend it to allow your detecting in init-local stage instead of init-network too)

blackboxsw:
@outscale-mdr one question for you.
```
The datasource is based EC2 datasource but with one big difference : the retrieval of tags. Indeed, in the meta-data server, in the tags section, we (3DS Outscale) accept all characters and this can lead to cloud-init crash during the retrieval of meta-data. AWS does not have this problem because if the tags are in the meta-data server, only a set of [characters is available ](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions).
```

Maybe I'm not understanding the failure mode here. It sounds like meta-data service (IMDS) implementation is publishing routes that get crawled but end up returning 404s. Any chance IMDS can be fixed to not publish those routes if you know they'll return a 404? Or any chance that the IMDS could be setup to return character sets broader than EC2 allows?

outscale-mdr:
Hi @blackboxsw,  thank you for the expplanation !

> On platforms without ds-identify cloud-init init --local will serially walk through all possible python datasources configured in order, calling the respective DatasourceX.get_data methods to determine whether or not the datasource is valid for the platform. This ends up being more costly than if we have a single Ec2 datasource that is flexible to "rule them all" and properly identify separate clouds by the right name.

Ok, good to know this !

outscale-mdr:
> @outscale-mdr one question for you.
> 
> ```
> The datasource is based EC2 datasource but with one big difference : the retrieval of tags. Indeed, in the meta-data server, in the tags section, we (3DS Outscale) accept all characters and this can lead to cloud-init crash during the retrieval of meta-data. AWS does not have this problem because if the tags are in the meta-data server, only a set of [characters is available ](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions).
> ```
> 
> Maybe I'm not understanding the failure mode here. It sounds like meta-data service (IMDS) implementation is publishing routes that get crawled but end up returning 404s. Any chance IMDS can be fixed to not publish those routes if you know they'll return a 404? Or any chance that the IMDS could be setup to return character sets broader than EC2 allows so that the 404 doesn't happen?

I double check with our team today and answer you about this.

outscale-mdr:
Ok so right now, we cannot internally fix IMDS to get around this issue before 6-10 months. 

At the same time, cloud-init crashes with our use case is quite blocking for us. 
Therefore would you accept to add this fix in cloud-init and if we have an IMDS fix we will reach you again?

blackboxsw:
> 

Thanks for the checking in on this @outscale-mdr with your team. I then think it makes sense cloud-init upstream could release support for a known Outscale platform bug and resolve that long-term once the IMDS implementation is confirmed fixed on your end.

I'd prefer then to file a launchpad bug in cloud-init first to capture this known failure mode for tags on Outscale at  https://bugs.launchpad.net/cloud-init/+filebug  and then reference in comments near the `skip_404_tag_errors` or whatever callback function the existing bug that is being worked around. Then we can track that bug long-term in the future and determine when it is safe to remove that skip callback on Outscale datasource.

Given that this is a short-term solution I would suggest more strongly that we also integrate that within the ec2 datasource if we can as I don't want cloud-init to create a new DatasourceOutscale that may be short-lived if the IMDS ends up fixing it's behavior because "making separate datasources disappear" from detection logic and ordering in ds-identify and python detection logic feels a bit more risky that just dropping a skip_callback once the functionality is present.

outscale-mdr:
Hello @blackboxsw,

It is ok for our team to add the detection of Outscale VMs in the EC2 Datasource as well as the short-lived fix.

Do you give the go for filing the launchpad bug this week ?

blackboxsw:
> Hello @blackboxsw,
> 
> It is ok for our team to add the detection of Outscale VMs in the EC2 Datasource as well as the short-lived fix.
> 
> Do you give the go for filing the launchpad bug this week ?

Please go ahead and put this changesset into the Ec2 datasource. I think that functionality makes sense contained within the Ec2 datasource making it a bit more flexible to handle the current state of your instance metadata server with regard to custom tag handling. Plus one on filing an upstream bug to report that tag errors on Outscale VMs result in 404s from the Ec2 datasource, which includes a snippet of the cloud-init.log failures should be enough context on the upstream bug. Then we can just comment `# Drop when LP: #XYZ tag handling is fixed` near the ec2 helper function to inform "future us" to checkin w/ your platform status occasionally to see when that function can be removed.

outscale-mdr:
Ok, I have sign the CLA on behalf of my company (3DS Outscale). Tell if it is all good for you ?

blackboxsw:
Looking good! I confirm signing of the Conributor license agreement for @outscale-mdr. I'm kicking of CI tests on this PR and will get a final review in tomorrow

outscale-mdr:
Fixing linting problem, my bad..

blackboxsw (approved):
Manually validated behavior as well on Ec2 to ensure no regressions or degradation in behavior on that platform. 

Also ran suite of Ec2 integration tests against EC2 ad confirm no failures.

blackboxsw:
Welcome to cloud-init @outscale-mdr . Expect that this commit will minimally be released  in cloud-init upstream release 22.4  in November https://discourse.ubuntu.com/t/cloud-init-2022-release-schedule/25413   Ubuntu team will then publish stable release updates of 22.4 into Ubuntu 18.04, 20.04 and 22.04 shortly thereafter. 22.10 (Kinetic) devel release may have an upload sooner than November for testing.

PR #1687, opened 2022-08-23
Ubuntu/devel

blackboxsw:
ubuntu/devel release so we can SRU to ubuntu/bionic|focal|jammy series

## Proposed Commit Message

```

```

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,15 @@
+cloud-init (22.3-3-g9f0efc47-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
+
+  * New upstream snapshot.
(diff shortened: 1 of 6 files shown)


debian/changelog
```diff
@@ -1,3 +1,15 @@
+cloud-init (22.3-3-g9f0efc47-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
```
TheRealFalcon: This should be kinetic, right?

TheRealFalcon: Doh, nevermind. Looking at a single commit and not all the files.

TheRealFalcon (approved):

PR #1690, opened 2022-08-25
net: Ensure a tmp with exec permissions for dhcp

aciba90:
## Proposed Commit Message

```
net: Ensure a tmp with exec permissions for dhcp

In the case cloudinit.temp_utils points to a fs mounted as noexec
and needs_exe=True, fallback to use
os.join.path(Distro.usr_lib_exec, "cloud-init/clouddir) that
will be mounted with exec perms.

LP: #1962343
```

## Additional Context

https://bugs.launchpad.net/cloud-init/+bug/1962343
https://warthogs.atlassian.net/browse/SC-958

## Test Steps

Execute `tests/integration_tests/datasources/test_tmp_noexec.py`

## Checklist:

### DIFF
--- a/cloudinit/config/cc_chef.py
+++ b/cloudinit/config/cc_chef.py
@@ -14,7 +14,9 @@ import os
 from textwrap import dedent
 
 from cloudinit import subp, temp_utils, templater, url_helper, util
+from cloudinit.cloud import Cloud
 from cloudinit.config.schema import MetaSchema, get_meta_doc
+from cloudinit.distros import Distro
 from cloudinit.settings import PER_ALWAYS
 
 RUBY_VERSION_DEFAULT = "1.8"
@@ -185,7 +187,7 @@ def get_template_params(iid, chef_cfg, log):
     return params
 
 
-def handle(name, cfg, cloud, log, _args):
+def handle(name: str, cfg: dict, cloud: Cloud, log, _args):
     """Handler method activated by cloud-init."""
 
     # If there isn't a chef key in the configuration don't do anything
@@ -289,7 +291,7 @@ def run_chef(chef_cfg, log):
     subp.subp(cmd, capture=False)
 
 
-def subp_blob_in_tempfile(blob, *args, **kwargs):
+def subp_blob_in_tempfile(blob, distro: Distro, *args, **kwargs):
     """Write blob to a tempfile, and call subp with args, kwargs. Then cleanup.
 
     'basename' as a kwarg allows providing the basename for the file.
@@ -298,23 +300,27 @@ def subp_blob_in_tempfile(blob, *args, **kwargs):
     """
     basename = kwargs.pop("basename", "subp_blob")
 
-    if len(args) == 0 and "args" not in kwargs:
-        args = [tuple()]
+    _args = list(args)
+    if len(_args) == 0 and "args" not in kwargs:
+        _args.append(tuple())
 
     # Use tmpdir over tmpfile to avoid 'text file busy' on execute
-    with temp_utils.tempdir(needs_exe=True) as tmpd:
+    with temp_utils.tempdir(
+        dir=distro.get_tmp_exec_path(), needs_exe=True
+    ) as tmpd:
         tmpf = os.path.join(tmpd, basename)
         if "args" in kwargs:
             kwargs["args"] = [tmpf] + list(kwargs["args"])
         else:
-            args = list(args)
-            args[0] = [tmpf] + args[0]
+            _args[0] = [tmpf] + _args[0]
 
         util.write_file(tmpf, blob, mode=0o700)
-        return subp.subp(*args, **kwargs)
+        return subp.subp(*_args, **kwargs)
 
 
-def install_chef_from_omnibus(url=None, retries=None, omnibus_version=None):
+def install_chef_from_omnibus(
+    distro: Distro, url=None, retries=None, omnibus_version=None
+):
     """Install an omnibus unified package from url.
 
     @param url: URL where blob of chef content may be downloaded. Defaults to
@@ -335,11 +341,15 @@ def install_chef_from_omnibus(url=None, retries=None, omnibus_version=None):
         args = ["-v", omnibus_version]
     content = url_helper.readurl(url=url, retries=retries).contents
     return subp_blob_in_tempfile(
-        blob=content, args=args, basename="chef-omnibus-install", capture=False
+        distro=distro,
+        blob=content,
+        args=args,
+        basename="chef-omnibus-install",
+        capture=False,
     )
 
 
-def install_chef(cloud, chef_cfg, log):
+def install_chef(cloud: Cloud, chef_cfg, log):
     # If chef is not installed, we install chef based on 'install_type'
     install_type = util.get_cfg_option_str(
         chef_cfg, "install_type", "packages"
@@ -361,6 +371,7 @@ def install_chef(cloud, chef_cfg, log):
     elif install_type == "omnibus":
         omnibus_version = util.get_cfg_option_str(chef_cfg, "omnibus_version")
         install_chef_from_omnibus(
+            distro=cloud.distro,
             url=util.get_cfg_option_str(chef_cfg, "omnibus_url"),
             retries=util.get_cfg_option_int(chef_cfg, "omnibus_url_retries"),
             omnibus_version=omnibus_version,
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -20,7 +20,15 @@ from typing import Any, Mapping, Optional, Type
 
 from cloudinit import importer
 from cloudinit import log as logging
-from cloudinit import net, persistence, ssh_util, subp, type_utils, util
+from cloudinit import (
+    net,
+    persistence,
+    ssh_util,
+    subp,
+    temp_utils,
+    type_utils,
+    util,
+)
 from cloudinit.distros.parsers import hosts
 from cloudinit.features import ALLOW_EC2_MIRRORS_ON_NON_AWS_INSTANCE_TYPES
 from cloudinit.net import activators, eni, network_state, renderers
@@ -942,7 +950,10 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         else:
             raise NotImplementedError()
 
-    def _get_tmp_exec_path(self) -> str:
+    def get_tmp_exec_path(self) -> str:
+        tmp_dir = temp_utils.get_tmp_ancestor(needs_exe=True)
+        if not util.has_mount_opt(tmp_dir, "noexec"):
+            return tmp_dir
         return os.path.join(self.usr_lib_exec, "cloud-init", "clouddir")
 
 
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -942,6 +942,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         else:
             raise NotImplementedError()
 
+    def _get_tmp_exec_path(self) -> str:
+        return os.path.join(self.usr_lib_exec, "cloud-init", "clouddir")
+
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
     """
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -41,9 +41,7 @@ class NoDHCPLeaseMissingDhclientError(NoDHCPLeaseError):
     """Raised when unable to find dhclient."""
 
 
-def maybe_perform_dhcp_discovery(
-    nic=None, dhcp_log_func=None, alt_exe_dir=None
-):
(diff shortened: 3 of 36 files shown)


cloudinit/net/ephemeral.py
```diff
@@ -417,21 +421,32 @@ def get_first_option_value(
...
         self.state_msg: str = ""
+        self.alt_tmp_dir = alt_tmp_dir
 
     def __enter__(self):
         # ipv6 dualstack might succeed when dhcp4 fails
         # therefore catch exception unless only v4 is used
         try:
             if self.ipv4:
-                self.stack.enter_context(EphemeralIPv6Network(self.interface))
+                self.stack.enter_context(
```
holmanb, line 443: Fixing the reversed ipv4/ipv6 context managers should probably be a separate PR. I suspect this technically breaks ephemeral ipv4 on ec2 currently (which is not currently in any releases).

cloudinit/distros/__init__.py
```diff
@@ -936,6 +936,9 @@ def set_keymap(self, layout, model, variant, options):
         else:
             raise NotImplementedError()
 
+    def _get_tmp_exec_path(self) -> str:
```
cjp256, line 939: Given that this interface is exposed to other packages, consider dropping the leading `_`?

cloudinit/net/dhcp.py
```diff
@@ -41,7 +41,9 @@ class NoDHCPLeaseMissingDhclientError(NoDHCPLeaseError):
     """Raised when unable to find dhclient."""
 
 
-def maybe_perform_dhcp_discovery(nic=None, dhcp_log_func=None):
+def maybe_perform_dhcp_discovery(
+    nic=None, dhcp_log_func=None, alt_exe_dir=None
```
cjp256, line 45: Perhaps  needs_exe=True to temp_utils.tmpdir() should just do the right thing and avoid requiring alt_exe_dir passed around?

cjp256:
Thank you for fixing this!! :D

holmanb (changes requested):
Thanks for this @aciba90!

I have one comment about a debug log that I think should be a warning, and a request/suggestion to check that the alt_exe_dir is itself executable. Once that and @cjp256's comments are addressed, I think this will be ready to merge.

cloudinit/net/ephemeral.py
aciba90: Extracted here: https://github.com/canonical/cloud-init/pull/1697

cloudinit/net/dhcp.py
holmanb: I was wondering about that too (or some other way to reduce the number of repeated changes), however doing that would couple `temp_utils` to `distro`. Is maybe that why @aciba90 didn't do this originally? I don't know off the top of my head if such a change would cause circular import issues, but it might.

cloudinit/distros/__init__.py
aciba90: Done.

cloudinit/net/dhcp.py
aciba90: Refactored to expose the target tmp_dir and let callers (datasources, modules) determine if the dir is mounted in a noexec point and give an exec dir via the distro in use.

cloudinit/temp_utils.py
```diff
@@ -24,12 +33,22 @@ def _tempfile_dir_arg(odir=None, needs_exe=False):
...
         tdir = _EXE_ROOT_TMPDIR
+        if _is_noexec_mount(tdir):
+            msg = (
+                f"Requested temporal dir with exe permission `{tdir}` is"
+                " mounted as noexec"
+            )
+            if alt_exe_dir:
+                msg += f". Falling back to `{alt_exe_dir}`"
+                tdir = alt_exe_dir
+            LOG.debug(msg)
```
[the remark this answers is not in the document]
aciba90: Changed to warning. This exposed other uses of `needs_exe=True` without handling the `noexec` condition in:

- cc_chef
- cc_growpart
- cc_puppet
- cc_ubuntu_drivers

aciba90:
Thanks, @cjp256 and @holmanb, for the constructive comments!

I have addressed your comments by exposing a function to let the callers of `cloudinit.temp_utils` know in advance what will be a tmp mount-point, to let them select the correct one.

This is ready to be reviewed after CI passes. Please, let me know if there are more concerns.

holmanb (changes requested):
I really like the integration test addition.

I assume fstab failing to mount on boot would cause the test to fail, but do you think we should add an assertion that ensures the test can't pass if not mounted noexec? If you think this is completely redundant, feel free to push back on this.

tests/integration_tests/datasources/test_tmp_noexec.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.adhoc
+@pytest.mark.azure
+@pytest.mark.ec2
+@pytest.mark.gce
+@pytest.mark.oci
+@pytest.mark.openstack
+def test_dhcp_tmp_noexec(client: IntegrationInstance):
+    customize_client(client)
```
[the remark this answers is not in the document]
aciba90: Good idea! Added and retested.

aciba90:
> I really like the integration test addition.
> 
> I assume fstab failing to mount on boot would cause the test to fail, but do you think we should add an assertion that ensures the test can't pass if not mounted noexec? If you think this is completely redundant, feel free to push back on this.

Thanks, @holmanb, for the review and suggestion. I think it is a great idea to assert test preconditions. I have done so and retested it.

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -210,7 +210,10 @@ def _get_data(self):
         if self._fallback_interface is None:
             self._fallback_interface = net.find_fallback_nic()
         try:
-            with EphemeralDHCPv4(self._fallback_interface):
+            with EphemeralDHCPv4(
+                self._fallback_interface,
+                tmp_dir=self.distro.get_tmp_exec_path(),
```
cjp256, line 215: I think having each caller sort out tmp_dir is going to be error prone.  Even at the lowest level, `tempdir(needs_exe=True) ` now has a hidden dependency on `dir=self.distro.get_tmp_exec_path()`.

I've got two thoughts:
- figure out how to tell temp_utils where the executable tmp dir is without a distro reference on init?
- add a distro.executable_tmpdir() which handles the needs_exe=True and executable dir requirements, dropping those aspects from tmpdir().   this adds a requirement for net to become distro-aware, which I'm actually surprised isn't already the case, but suspect it'll happen eventually as bringing up ephemeral dhcp is likely to have distro-specific aspects (e.g. distros removing dhclient at some point).

aciba90: Thanks, @cjp256, for the detailed and interesting discussion.

> * figure out how to tell temp_utils where the executable tmp dir is without a distro reference on init?

In the current state of the PR (commit 29833a75611f67d0f86b954bb78f53c17f14eba5), there is no reference to `distro` within `temp_utils`.

> [0]: add a distro.executable_tmpdir() which handles the needs_exe=True and executable dir requirements, dropping those aspects from tmpdir().

There is a new Distro method [get_tmp_exec_path](https://github.com/canonical/cloud-init/pull/1690/files#diff-2902e7f3123dd1e701482d6c60d6ca2e51d2dcbe3f643cb30be70727fde3fbfbR953) which handles the executable dir requirements. It checks if temp_dir with need_exe target tmp dir is in a noexec mount point (via a new function [temp_utils.get_tmp_ancestor](https://github.com/canonical/cloud-init/pull/1690/files#diff-6179e8b68a1cb97cb60b99f1c269a896292fc6dec97cb450cec13886a0101e80R18)) and if so fallback to `{distro.usr_lib_exec}/cloud-init/clouddir`.

We could move the `need_exe` from `temp_utils` to `distro.get_tmp_exec_path` simplifying things a bit.

> [1]: this adds a requirement for net to become distro-aware, which I'm actually surprised isn't already the case, but suspect it'll happen eventually as bringing up ephemeral dhcp is likely to have distro-specific aspects (e.g. distros removing dhclient at some point).

I do not see how [0] implies [1], but assuming it does, I had some arguments against tightly coupling `distro` with `net.dchp` and/or `net.ephemeral`:

1. The necessity of copying and running `dchp` from a tmp dir in order to avoid apparmor restrictions is going to be eliminated (@holmanb can point to links about this, I believe). This implies the dependency will disappear.
2. After [WIP-ONGOING-REFACTORIZATION.rst](https://github.com/canonical/cloud-init/blob/main/WIP-ONGOING-REFACTORIZATION.rst), `net.dchp` and `net.ephemeral` are going to belong to a `cloudinit.distros.networking` hierarchy of classes mapped to the distro hierarchy. If the number of distro-specific aspects (to bring up ephemeral dchp) increases, we could encode the dependencies via this hierarchy.
3. I remember a discussion about dchp. I think `dchp` is going to be discontinued and we talk about the possibility of including a new dchp python-dependency and other alternatives. This will also remove the dependency. (But this belongs to the more obscure future)

Summarizing, I do not think that to only have a distro's exec tmp dir and net justifies to tightly couple them, but I am open to hear more arguments. Maybe @blackboxsw, @holmanb and @TheRealFalcon have opinions/arguments about this.

holmanb: @aciba90 @cjp256 I'm glad we're thinking about architecture, refactoring, and future direction here. Good points made all around.

The topic of dhcp (both sandboxing issues and the looming issue of dhclient's support status) has come up on irc periodically and maybe also in PR conversations. This topic probably deserves better community coordination and communication, since the problem has implications for clouds and distros (i.e. if we wanted to use a dedicated or auto-assigned link local ipv4 address, dhcp probably wouldn't be necessary, but does that work for all clouds?). This probably deserves a mailing list discussion. Or maybe we should find a way to get contributors in the same room to talk about this kind of stuff :)

> in order to avoid apparmor restrictions is going to be eliminated

This would eliminate some headaches and one historical source of bugs. However we might have to push on this one ourselves.

> Even at the lowest level, tempdir(needs_exe=True) now has a hidden dependency on dir=self.distro.get_tmp_exec_path().

Agreed, we already have implicit coupling.
It seems like more and more code has been shifting into the `distro` implementations recently. I expect this trend will likely continue, which suggests that `distro` may become required by more modules over time.

> 1. The necessity of copying and running dchp from a tmp dir in order to avoid apparmor restrictions is going to be eliminated (@holmanb can point to links about this, I believe). This implies the dependency will disappear.
> ...
> 3. I remember a discussion about dchp. I think dchp is going to be discontinued and we talk about the possibility of including a new dchp python-dependency and other alternatives. This will also remove the dependency. (But this belongs to the more obscure future)

Either of these solutions could potentially remove the requirement for a writable executable tempdir for networking. It does not, however, remove all dependency in cloud-init on `needs_exe`. I see at least a couple of other calls (though the debconf one does look unnecessary at a glance).

All in all, I'm not opposed to this approach. I do think that since we currently need distro input in this patch, it might be worth just creating a distro wrapper function to make the callsites less error prone and more DRY. Whether to do that before this PR lands or as a refactor after, I don't feel strongly about.

That's a lot of words to say basically "I don't have strong feelings on this. I could still be swayed either way." I'll wait in case @cjp256 @blackboxsw @TheRealFalcon or anyone else has more thoughts.

cjp256: To clarify, I'm not demanding any changes - I'm just driving by and sharing my thoughts fwiw. :D  I'm happy to have this fixed in whatever way you think is best at this time!

holmanb: @cjp256 Thanks much for sharing. Sounds good :)

cloudinit/config/cc_chef.py
```diff
@@ -298,23 +300,27 @@ def subp_blob_in_tempfile(blob, *args, **kwargs):
     """
     basename = kwargs.pop("basename", "subp_blob")
 
-    if len(args) == 0 and "args" not in kwargs:
-        args = [tuple()]
+    _args = list(args)
+    if len(_args) == 0 and "args" not in kwargs:
+        _args.append(tuple())
```
holmanb, line 305: If this branch is taken then on line 315 we will be concatenating a list and a tuple, which will cause a typerror.
```python
_args[0] = [tmpf] + _args[0]
```

holmanb (changes requested):
I just noticed one last nit. It's not something you caused, but since we're modifying this chunk of code I think we should fix it. Pyright picks up error on main (maybe should we supplement mypy), but not on your branch. The problem still exists, however, it just gets hidden from type analyzers in the changes here. I suspect a fix should be as simple as `s/tuple()/list()/g` on line highlighted in the comment.

I'll merge once that issue is fixed (I'd merge now but I want a second opinion on my suggestion).

holmanb (approved):
Actually it has only one call sight, and we can see that kwargs contains args, so we can simplify a bit. I'm going to merge this and we'll do a little cleanup PR for that shortly.

LGTM. Thanks!

cjp256:
Thanks Alberto @aciba90!

aciba90:
> Thanks Alberto @aciba90!

Thank you, @cjp256, for the reviews and feedback!

PR #1698, opened 2022-08-29
doc: add reporting suggestion to docs (SC-1236)

holmanb:
```
doc: add reporting suggestion to docs
```

## Additional Context
External scripts waiting on cloud-init leads to bug reports that relate to cloud-init only symptomatically. Lets document how users can make better reports in the faq.

(open to suggestions if we want to list more commands to report on)

### DIFF
--- a/doc/rtd/topics/faq.rst
+++ b/doc/rtd/topics/faq.rst
@@ -166,21 +166,6 @@ To wait for clous-init to complete, run:
 
         $ cloud-init status --wait
 
-Failing to Complete on Systemd:
--------------------------------
-
-Cloud-init consists of multiple services on systemd. If a service
-that cloud-init depends on stalls, cloud-init will not continue.
-If reporting a bug related to cloud-init failing to complete on
-systemd, please make sure to include the following logs.
-
-.. code-block:: shell-session
-
-        $ systemd-analyze critical-chain cloud-init.target
-        $ journalctl --boot=-1
-        $ systemctl --failed
-
-
 There are a number of reasons that cloud-init might never complete. This list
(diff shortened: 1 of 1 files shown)


doc/rtd/topics/faq.rst
```diff
@@ -166,12 +166,27 @@ To wait for clous-init to complete, run:
...
+-------------------------------
+
+Cloud-init consists of multiple services on systemd. If a service
+that cloud-init depends on stalls, cloud-init will not continue.
+If reporting a bug related to cloud-init failing to complete on
+systemd, please make sure to include the following logs.
+
+.. code-block:: shell-session
+
+        $ systemd-analyze critical-chain cloud-init.target
```
aciba90, line 179: Not sure if this would add a lot of info but maybe is a good thing:

```suggestion
        $ systemd-analyze critical-chain cloud-init.target
        $ journalctl --boot=-1
```

aciba90 (approved):
LGTM, thanks. I added a journalctl suggestion to collect last boot logs.

doc/rtd/topics/faq.rst
holmanb: Good idea. On long-running systems with limited storage there is a chance that logs from former boots will be garbage collected by systemd, but worth suggesting anyhow I think.

TheRealFalcon (approved):
I would maybe move the
```
There are a number of reasons that cloud-init might never complete. This list
is not exhaustive, but attempts to enumerate potential causes:
```
section above `Failing to complete on systemd`, or perhaps move `Failing to complete on systemd` below `Internal reasons`, but +1 regardless.

PR #1699, opened 2022-08-30
Fix v2 interface matching when no MAC

TheRealFalcon:
Feel free to make any changes necessary to bring this PR across the finish line.

## Proposed Commit Message

```
Fix v2 interface matching when no MAC

Commit af40478 added a regression. When parsing v2 config, matching can
fail when no MAC is available. This commit fixes that behavior.

LP: #1986551
```

## Additional Context
https://bugs.launchpad.net/subiquity/+bug/1986551
caused by https://github.com/canonical/cloud-init/pull/1552

## Test Steps
It seems we don't really have an easy way to add/edit/update v1/v2 config -> renderer "end-to-end" tests. test_net.py exists but is a nightmare to edit. I created a new test file that can hopefully be the place we can put such tests. The implementation is still incomplete as far as supporting all the renderers, but it works for this specific test case and shouldn't be hard to add support to.


tests/unittests/net/test_net_rendering.py
```diff
@@ -0,0 +1,101 @@
...
+        ), f"Expected {actual_path} to exist, but it does not"
+        actual_contents = actual_path.read_text()
+        assert expected_contents.strip() == actual_contents.strip()
+
+
+@pytest.mark.parametrize(
+    "test_name, renderers",
+    [("no_matching_mac_v2", Renderer.Netplan | Renderer.NetworkManager)],
+)
+def test_convert(test_name, renderers, tmp_path):
```
aciba90, line 91: `tmp_path` fixture is not available in `TOXENV=lowest-supported`. A possible solution could be:

```diff
diff --git a/tests/unittests/conftest.py b/tests/unittests/conftest.py
index e265a285..1ab17e8b 100644
--- a/tests/unittests/conftest.py
+++ b/tests/unittests/conftest.py
@@ -1,6 +1,7 @@
 import builtins
 import glob
 import os
+from pathlib import Path
 
 import pytest
 
@@ -55,3 +56,12 @@ def fake_filesystem(mocker, tmpdir):
             func = getattr(mod, f)
             trap_func = retarget_many_wrapper(str(tmpdir), nargs, func)
             mocker.patch.object(mod, f, trap_func)
+
+
+PYTEST_VERSION_TUPLE = tuple(map(int, pytest.__version__.split(".")))
+
+if PYTEST_VERSION_TUPLE < (3, 9, 0):
+
+    @pytest.fixture
+    def tmp_path(tmpdir):
+        return Path(tmpdir)
```

aciba90 (changes requested):
Thanks, @TheRealFalcon.

I have commented a possible solution for the lack of `tmp_path` in older versions of `pytest`.

tests/unittests/net/test_net_rendering.py
cjp256: Interesting!

tests/unittests/conftest.py
```diff
@@ -55,3 +56,12 @@ def fake_filesystem(mocker, tmpdir):
             func = getattr(mod, f)
             trap_func = retarget_many_wrapper(str(tmpdir), nargs, func)
             mocker.patch.object(mod, f, trap_func)
+
+
+PYTEST_VERSION_TUPLE = tuple(map(int, pytest.__version__.split(".")))
+
+if PYTEST_VERSION_TUPLE < (3, 9, 0):
```
blackboxsw, line 63: confirmed I see the pytest commit https://github.com/pytest-dev/pytest/commit/ab3637d486 as defined/included in release 3.9.0. Thanks for the fix

tests/unittests/net/test_net_rendering.py
blackboxsw: Confirmed I see the upstream pytest commit https://github.com/pytest-dev/pytest/commit/ab3637d486 as defining the tmp_path fixture and see it's release was 3.9.0

holmanb (approved):
LGTM thanks! I'll leave this unmerged since I think @blackboxsw is looking at it.

tests/unittests/net/test_net_rendering.py
```diff
@@ -0,0 +1,101 @@
...
+def _check_network_manager(network_state: NetworkState, tmp_path: Path):
+    renderer = NetworkManagerRenderer()
+    renderer.render_network_state(
+        network_state, target=str(tmp_path / "no_matching_mac")
+    )
+    expected_paths = glob.glob(
+        str(ARTIFACT_DIR / "no_matching_mac" / "**/*.nmconnection"),
+        recursive=True,
+    )
+    for expected_path in expected_paths:
```
blackboxsw, line 75: What this testing doesn't cover is whether we've generated config content we didn't expect to see.  Since we were refactoring in this branch we could address render-based abstractions in this test and do full dict comparisons to ensure we aren't unexpected files by a full directory comparison [like this](https://github.com/TheRealFalcon/cloud-init/compare/fix-match...blackboxsw:cloud-init:fix-match) But we will take that as a followup after this branch lands.

blackboxsw (approved):
will followup with more test refactors later. content works well and a good improvement in general from what we had.

PR #1707, opened 2022-08-31
config: enable bootstrapping pip in ansible

holmanb:
```
config: enable bootstrapping pip in ansible
```

### DIFF
--- a/tests/unittests/config/test_cc_ansible.py
+++ b/tests/unittests/config/test_cc_ansible.py
@@ -269,22 +269,18 @@ class TestAnsible:
     def test_deps_not_installed(self, m_which):
         """assert exception raised if package not installed"""
         with raises(ValueError):
-            cc_ansible.AnsiblePullDistro(
-                get_cloud().distro.install_packages
-            ).check_deps()
+            cc_ansible.AnsiblePullDistro(get_cloud().distro).check_deps()
 
     @mock.patch(M_PATH + "which", return_value=True)
     def test_deps(self, m_which):
         """assert exception not raised if package installed"""
-        cc_ansible.AnsiblePullDistro(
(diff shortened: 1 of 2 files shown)


tests/unittests/config/test_cc_ansible.py
```diff
@@ -16,6 +16,7 @@
 from tests.unittests.helpers import skipUnlessJsonSchema
 from tests.unittests.util import get_cloud
 
+MOD = "cloudinit.config.cc_ansible."
```
aciba90, line 19: Nit: According to our [development documentation](https://cloudinit.readthedocs.io/en/latest/topics/testing.html), it would be nice to name this as `M_PATH`.

aciba90 (approved):
LGTM! Thanks, @holmanb. I have added a nit comment only, feel free to ignore it :)

aciba90 (approved):
LGTM, thanks!

TheRealFalcon (approved):
Personally, I find it more preferable to pass `cloud.distro` to the respective class and then call `distro.install_packages`, even if your way is slightly better abstracted. The typing lets you specify the concrete class you can jump back to, whereas Callable is much more generic and requires first going to the call point to see where the callable variable was set.

+1 either way though. Feel free to merge as-is.

holmanb:
> Personally, I find it more preferable to pass `cloud.distro` to the respective class and then call `distro.install_packages`, even if your way is slightly better abstracted. The typing lets you specify the concrete class you can jump back to, whereas Callable is much more generic and requires first going to the call point to see where the callable variable was set.

Agreed. `Distro` is also a common building block in the cloud-init codebase so passing around `Distros` improves recognition. +1 I'll do that.

PR #1709, opened 2022-08-31
Refactor instance json files to use Paths (SC-1238)

TheRealFalcon:
## Proposed Commit Message

```
Refactor instance json files to use Paths

We have multiple areas of the codebase where we fetch the rundir path
and join it with the filename for instance-data.json or
instance-data-sensitive.json. This commit refactors those calls to use a
'Paths' lookup instead.
```

## Additional Context
In a separate task I was adding a call to instance-data-sensitive.json and decided to refactor this first.

### DIFF
--- a/cloudinit/helpers.py
+++ b/cloudinit/helpers.py
@@ -382,6 +382,12 @@ class Paths(persistence.CloudInitPickleMixin):
             self.run_dir = Paths(
                 path_cfgs=self.cfgs, ds=self.datasource
             ).run_dir
+        if "instance_data" not in self.lookups:
+            self.lookups["instance_data"] = "instance-data.json"
+        if "instance_data_sensitive" not in self.lookups:
+            self.lookups[
+                "instance_data_sensitive"
+            ] = "instance-data-sensitive.json"
 
     # get_ipath_cur: get the current instance path for an item
     def get_ipath_cur(self, name=None):
--- a/cloudinit/helpers.py
+++ b/cloudinit/helpers.py
@@ -350,7 +350,12 @@ class Paths(persistence.CloudInitPickleMixin):
             "cloud_config": "cloud-config.txt",
             "data": "data",
             "handlers": "handlers",
+            # File in which public available instance meta-data is written
+            # security-sensitive key values are redacted from this
+            # world-readable file
             "instance_data": "instance-data.json",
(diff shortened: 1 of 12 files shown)


cloudinit/helpers.py
```diff
@@ -333,33 +333,39 @@ class Paths(persistence.CloudInitPickleMixin):
...
+        template_dir: str = path_cfgs.get(
+            "templates_dir", "/etc/cloud/templates/"
+        )
+        self.template_tpl: str = os.path.join(template_dir, "%s.tmpl")
         self.lookups = {
+            "boothooks": "boothooks",
+            "cloud_config": "cloud-config.txt",
+            "data": "data",
             "handlers": "handlers",
+            "instance_data": "instance-data.json",
```
TheRealFalcon, line 353: Only "real" change to this file is these two lines. Everything else is either ordering this list or adding types.

aciba90, line 353: Nit: Maybe it's helpful to carry over the comments from `cloudinit.sources.__init__`:

```suggestion
            # File in which public available instance meta-data is written
            # security-sensitive key values are redacted from this
            # world-readable file
            "instance_data": "instance-data.json",
            # security-sensitive key values are present in this
            # root-readable file
```

aciba90 (approved):
LGTM. Thanks for the nice refactoring!

cloudinit/helpers.py
```diff
@@ -376,6 +382,12 @@ def _unpickle(self, ci_pkl_version: int) -> None:
             self.run_dir = Paths(
                 path_cfgs=self.cfgs, ds=self.datasource
             ).run_dir
+        if "instance_data" not in self.lookups:
```
holmanb, line 385: These lines are a bit confusing at a glance, since in this copy of the code, `instance_data` _is_ in `self.lookups`. This is for loading old pickles that will not contain this data, right? Can we add some comments mentioning that, please?

A separate method `_add_new_fields_to_old_pickles()` to delimit the start and end of upgrade path code might help make this really obvious to future readers, but I'll leave it up to you whether that's necessary.

holmanb (approved):
LGTM. Leaving unmerged so you can address @aciba90's suggestion.

TheRealFalcon:
Done

holmanb (approved):

PR #1710, opened 2022-08-31
support dhcp{4,6}-overrides in networkd renderer

adobley:
## Proposed Commit Message

```
support dhcp{4,6}-overrides in networkd renderer
```

## Test Steps

Create the metadata file for the VM `metadata.yaml`:
```
instance-id: cloud-vm
local-hostname: cloud-vm
network:
  version: 2
  ethernets:
    nics:
      match:
        name: ens*
      dhcp4: true
      dhcp4-overrides:
        use-dns: false
      nameservers:
        addresses: [8.8.8.8, 8.8.4.4]
```

Create the userdata file `userdata.yaml`:
```
#cloud-config

users:
- default
- name: ${YOUR_USER}
    primary_group: ${YOUR GROUP}
    sudo: ALL=(ALL) NOPASSWD:ALL
    groups: sudo, wheel
    lock_passwd: true
    ssh_authorized_keys:
    - ${YOUR_SSH_PUB_KEY}
```

Create a powered-off Photon VM in vSphere:
```
govc import.ova photon-3-kube.ova
govc vm.markastemplate photon-3-kube
govc vm.clone -on=false -vm /dc0/vm/photon-3-kube cloud-vm
```

Set the guestinfo on the VM while it is powered off:
```
export GUESTINFO_USERDATA="$(cat userdata.yaml | base64 -w0)"
export GUESTINFO_METADATA="$(cat metadata.yaml | base64 -w0)"

govc vm.change -vm /dc0/vm/cloud-vm \
	-e guestinfo.userdata="${GUESTINFO_USERDATA}" \
	-e guestinfo.userdata.encoding="base64" \
	-e guestinfo.metadata="${GUESTINFO_METADATA}" \
	-e guestinfo.metadata.encoding="base64"
```

Power on the VM:
```
govc vm.power -on /dc0/vm/humalog
```

SSH to the VM and check the resulting network config is as expected:
```
ssh ${YOUR_USER}@${CLOUD_VM_IP}
networkctl status
```

See that only the configured DNS servers are in the list and no DNS servers came from DHCP.
This is just one of the several flags that could be toggled but gives an idea if it's wired up correctly.

## Checklist:

### DIFF
--- a/cloudinit/net/networkd.py
+++ b/cloudinit/net/networkd.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # vi: ts=4 expandtab
 #
-# Copyright (C) 2021 VMware Inc.
+# Copyright (C) 2021-2022 VMware Inc.
 #
 # Author: Shreenidhi Shedi <yesshedi@gmail.com>
 #
@@ -175,6 +175,8 @@ class Renderer(renderer.Renderer):
         ):
             cfg.update_section(sec, "IPv6AcceptRA", iface["accept-ra"])
 
+        return dhcp
+
     # This is to accommodate extra keys present in VMware config
     def dhcp_domain(self, d, cfg):
         for item in ["dhcp4domain", "dhcp6domain"]:
@@ -216,6 +218,34 @@ class Renderer(renderer.Renderer):
             if k in dns and dns[k]:
                 cfg.update_section(sec, v, " ".join(dns[k]))
 
+    def parse_dhcp_overrides(self, cfg, device, dhcp, version):
+        dhcp_config_maps = {
+            "UseDNS": "use-dns",
+            "UseDomains": "use-domains",
+            "UseHostname": "use-hostname",
+            "UseNTP": "use-ntp",
+        }
+
+        if version == "4":
+            dhcp_config_maps.update(
+                {
+                    "SendHostname": "send-hostname",
+                    "Hostname": "hostname",
+                    "RouteMetric": "route-metric",
+                    "UseMTU": "use-mtu",
+                    "UseRoutes": "use-routes",
+                }
+            )
+
+        if f"dhcp{version}-overrides" in device and dhcp in [
+            "yes",
+            f"ipv{version}",
+        ]:
+            dhcp_overrides = device[f"dhcp{version}-overrides"]
+            for k, v in dhcp_config_maps.items():
+                if v in dhcp_overrides:
+                    cfg.update_section(f"DHCPv{version}", k, dhcp_overrides[v])
+
     def create_network_file(self, link, conf, nwk_dir):
         net_fn_owner = "systemd-network"
 
@@ -248,7 +278,7 @@ class Renderer(renderer.Renderer):
 
             link = self.generate_match_section(iface, cfg)
             self.generate_link_section(iface, cfg)
-            self.parse_subnets(iface, cfg)
+            dhcp = self.parse_subnets(iface, cfg)
             self.parse_dns(iface, cfg, ns)
 
             for route in ns.iter_routes():
@@ -271,7 +301,25 @@ class Renderer(renderer.Renderer):
                             name = dev_name
                             break
                 if name in ns.config["ethernets"]:
-                    self.dhcp_domain(ns.config["ethernets"][name], cfg)
+                    device = ns.config["ethernets"][name]
+
+                    # dhcp{version}domain are extra keys only present in
+                    # VMware config
+                    self.dhcp_domain(device, cfg)
+                    for version in ["4", "6"]:
+                        if (
+                            f"dhcp{version}domain" in device
+                            and "use-domains"
+                            in device.get(f"dhcp{version}-overrides", {})
+                        ):
+                            exception = (
+                                f"{name} has both dhcp{version}domain"
+                                f" and dhcp{version}-overrides.use-domains"
+                                f" configured. Use one"
+                            )
+                            raise Exception(exception)
+
+                        self.parse_dhcp_overrides(cfg, device, dhcp, version)
 
             ret_dict.update({link: cfg.get_final_conf()})
 
--- a/doc/rtd/topics/network-config-format-v2.rst
+++ b/doc/rtd/topics/network-config-format-v2.rst
@@ -8,6 +8,8 @@ version 2 format defined for the `netplan`_ tool.  Cloud-init supports
 both reading and writing of Version 2; the latter support requires a
 distro with `netplan`_ present.
 
+.. _Netplan Passthrough:
+
 Netplan Passthrough
 -------------------
 
@@ -176,6 +178,48 @@ Enable DHCP for IPv4. Off by default.
 **dhcp6**: *<(bool)>*
 
 Enable DHCP for IPv6. Off by default.
+
+**dhcp4-overrides** and **dhcp6-overrides**: *<(mapping)>*
+
+DHCP behavior overrides. Overrides will only have an effect if
+the corresponding DHCP type is enabled. Refer to `netplan#dhcp-overrides`_
+for more documentation.
+
+.. note::
+
+  These properties are only consumed on ``netplan`` and ``networkd``
+  renderers.
+
+The ``netplan`` renderer :ref:`passes through <Netplan Passthrough>`
+everything and the ``networkd`` renderer consumes the following sub-properties:
+
+* ``hostname`` *
+* ``route-metric`` *
+* ``send-hostname`` *
+* ``use-dns``
+* ``use-domains``
+* ``use-hostname``
+* ``use-mtu`` *
+* ``use-ntp``
+* ``use-routes`` *
+
+.. note::
+
+   Sub-properties marked with a ``*`` are unsupported for ``dhcp6-overrides``
+   when used with the ``networkd`` renderer.
+
+Example: ::
+
+  dhcp4-overrides:
+    hostname: hal
+    route-metric: 1100
+    send-hostname: false
+    use-dns: false
+    use-domains: false
+    use-hostname: false
+    use-mtu: false
+    use-ntp: false
+    use-routes: false
 
 **addresses**: *<(sequence of scalars)>*
 
@@ -527,4 +571,5 @@ This is a complex example which shows most available features: ::
         dhcp4: yes
 
 .. _netplan: https://netplan.io
+.. _netplan#dhcp-overrides: https://netplan.io/reference#dhcp-overrides
 .. vi: textwidth=79
--- a/tests/unittests/net/test_networkd.py
+++ b/tests/unittests/net/test_networkd.py
@@ -1,6 +1,9 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
+from string import Template
 from unittest import mock
+
+import pytest
 
 from cloudinit import safeyaml
 from cloudinit.net import network_state, networkd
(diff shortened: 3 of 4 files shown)


cloudinit/net/networkd.py
```diff
@@ -271,7 +298,23 @@ def _render_content(self, ns):
                             name = dev_name
                             break
                 if name in ns.config["ethernets"]:
-                    self.dhcp_domain(ns.config["ethernets"][name], cfg)
+                    device = ns.config["ethernets"][name]
+                    for version in ["4", "6"]:
```
aciba90, line 302: This check is specific for VMware config. The following diff adds a comment underlying that and simplifying the if-condition. Feel free to rephrase it.

```diff
diff --git a/cloudinit/net/networkd.py b/cloudinit/net/networkd.py
index d9ad1fc8c..c54352956 100644
--- a/cloudinit/net/networkd.py
+++ b/cloudinit/net/networkd.py
@@ -299,12 +299,13 @@ class Renderer(renderer.Renderer):
                             break
                 if name in ns.config["ethernets"]:
                     device = ns.config["ethernets"][name]
+                    # dhcp{version}domain are extra keys only present in
+                    # VMware config.
                     for version in ["4", "6"]:
                         if (
                             f"dhcp{version}domain" in device
-                            and f"dhcp{version}-overrides" in device
                             and "use-domains"
-                            in device[f"dhcp{version}-overrides"]
+                            in device.get(f"dhcp{version}-overrides", {})
                         ):
                             exception = (
                                 f"{name} has both dhcp{version}domain"

```

tests/unittests/net/test_networkd.py
```diff
@@ -71,5 +183,89 @@ def test_networkd_render_with_set_name(self):
...
+            renderer = networkd.Renderer()
+            rendered_content = renderer._render_content(ns)
+
+        assert (
+            rendered_content["eth0"] == V2_CONFIG_DHCP_YES_OVERRIDES_RENDERED
+        )
+
+    @pytest.mark.parametrize("dhcp_version", [("4"), ("6")])
+    def test_networkd_render_dhcp_domains_vs_overrides(self, dhcp_version):
+        expectedException = (
```
aciba90, line 200: In general, we use snake-case to name vars. Could we rename `expectedException` to `expected_exception`?

tests/unittests/net/test_networkd.py
```diff
@@ -53,6 +56,115 @@
...
+        use-dns: false
+        use-domains: false
+        use-hostname: false
+        use-mtu: false
+        use-ntp: false
+        use-routes: false
+      match:
+        macaddress: '00:11:22:33:44:55'
+      nameservers:
+        addresses: [8.8.8.8,2001:4860:4860::8888]
```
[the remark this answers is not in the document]
adobley: Fixed, thanks!

cloudinit/net/networkd.py
```diff
@@ -216,6 +218,31 @@ def parse_dns(self, iface, cfg, ns):
...
+            "SendHostname": "send-hostname",
+            "UseHostname": "use-hostname",
+            "Hostname": "hostname",
+            "RouteMetric": "route-metric",
+            "UseDomains": "use-domains",
+            "UseMTU": "use-mtu",
+            "UseRoutes": "use-routes",
+        }
+
+        if "dhcp4-overrides" in device and dhcp in ["yes", "ipv4"]:
```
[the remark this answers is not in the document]
adobley: We took a slightly different approach of making this function parameterized with that data and using the existing for loop we had to do the validation on the conflicting keys. Let me know if there are any concerns :)

cloudinit/net/networkd.py
adobley: Thanks for the tip! I haven't written python in a while. Glad there was a better way :)

tests/unittests/net/test_networkd.py
adobley: Whoops! Thanks for catching that mistake.

adobley:
Hi @aciba90! Thanks for all your feedback! We fixed up the suggestions and added some docs on the fields. I'd love to hear any thoughts you have.

Would it be better to just link out to the netplan docs for the explanations of some of the properties?
Is there a better way we can show that note/warning in the docs for the renderer limitations?

We weren't too sure where a good place to add a log that alerts users when `dhcp{4,6}-overrides` are set and they are not using the `networkd` renderer. I didn't see any obvious place doing a high level validation like that, but maybe I missed it. We could add something just before the render gets called, if we have enough information then.

doc/rtd/topics/network-config-format-v2.rst
```diff
@@ -177,6 +177,72 @@ Enable DHCP for IPv4. Off by default.
 
 Enable DHCP for IPv6. Off by default.
 
+**dhcp4-overrides**: *<(mapping)>*
```
aciba90, line 180: The following diff simplifies and improves the documentation. By:

* Not documenting the sub-properties, but referring to netplan's doc.
* Grouping dhcp4-overrides and dhcp6-overrides in one section.
* Clarify that with v2 netconfs and netplan as renderer, the overrides are also applied.

Please, feel free to modify/rephrase it.

```diff
diff --git a/doc/rtd/topics/network-config-format-v2.rst b/doc/rtd/topics/network-config-format-v2.rst
index 8597cc32b..b00d4110e 100644
--- a/doc/rtd/topics/network-config-format-v2.rst
+++ b/doc/rtd/topics/network-config-format-v2.rst
@@ -8,6 +8,8 @@ version 2 format defined for the `netplan`_ tool.  Cloud-init supports
 both reading and writing of Version 2; the latter support requires a
 distro with `netplan`_ present.
 
+.. _Netplan Passthrough:
+
 Netplan Passthrough
 -------------------
 
@@ -177,62 +179,32 @@ Enable DHCP for IPv4. Off by default.
 
 Enable DHCP for IPv6. Off by default.
 
-**dhcp4-overrides**: *<(mapping)>*
-
-Some DHCP behaviors can be overridden. Overrides will only have an effect if
-the corresponding DHCP type is enabled. Optional.
-
-``hostname`` *<(scalar)>* sets the hostname which is sent to the DHCP server,
-instead of the machine's hostname. Optional.
-``route-metric`` *<(scalar)>* sets the default metric for routes that are
-automatically added. This allows prioritizing routes for devices. Optional.
-``send-hostname`` *<(bool)>* when ``true``, the machine's hostname will be sent
-to the DHCP server. Default: ``true``. Optional.
-``use-dns`` *<(bool)>* when ``true``, the DNS servers received from DHCP will
-be used and take precedence over configured ones. Default: ``true``. Optional.
-``use-hostname`` *<(bool)>* when ``true``,  the hostname received from DHCP
-will be set as the transient hostname of the machine. Default: ``true``. Optional.
-``use-mtu`` *<(bool)>* when ``true``, the MTU received from DHCP will be used
-as the MTU for the network interface. Default: ``true``. Optional.
-``use-ntp`` *<(bool)>* when ``true``, the NTP servers received from DHCP will
-be used by systemd-timesyncd and take precendence over configured ones.
-Default: ``true``. Optional.
-``use-domains`` *<(scalar)>* takes a boolean or the value ``route``. When
-``true``, the domain name received from DHCP will be used as the DNS search
-domain for this network interface. When ``route``, the domain name received
-from DHCP will be used for routing DNS only, not for searching. Optional.
-
-NOTE: This is currently only supported for the ``networkd`` renderer. As of
-writing this the only known usage of this renderer is with the VMware data
-source.
+**dhcp4-overrides** and **dhcp6-overrides**: *<(mapping)>*
 
-Example: ::
+DHCP behavior overrides. Overrides will only have an effect if
+the corresponding DHCP type is enabled. Refer to `netplan#dhcp-overrides`_
+for more documentation.
 
-  dhcp4-overrides:
-    hostname: hal
-    route-metric: 1100
-    send-hostname: false
-    use-dns: false
-    use-domains: false
-    use-hostname: false
-    use-mtu: false
-    use-ntp: false
-    use-routes: false
-
-**dhcp6-overrides**: *<(mapping)>*
+.. note::
 
-Some DHCP behaviors can be overridden. Overrides will only have an effect if
-the corresponding DHCP type is enabled. Optional.
+  These properties are only consumed on ``netplan`` and ``networkd``
+  renderers.
 
-See ``dhcp4-overrides`` above for an explanation of the fields.
+The ``netplan`` renderer :ref:`passes through <Netplan Passthrough>`
+everything and the ``networkd`` renderer consumes the following sub-properties:
 
-NOTE: This is currently only supported for the ``networkd`` renderer. As of
-writing this the only known usage of this renderer is with the VMware data
-source.
+* ``hostname``
+* ``route-metric``
+* ``send-hostname``
+* ``use-dns``
+* ``use-domains``
+* ``use-hostname``
+* ``use-mtu``
+* ``use-ntp``
 
 Example: ::
 
-  dhcp6-overrides:
+  dhcp4-overrides:
     hostname: hal
     route-metric: 1100
     send-hostname: false
@@ -593,4 +565,5 @@ This is a complex example which shows most available features: ::
         dhcp4: yes
 
 .. _netplan: https://netplan.io
+.. _netplan#dhcp-overrides: https://netplan.io/reference#dhcp-overrides
 .. vi: textwidth=79

```

aciba90 (changes requested):
Thanks, @adobley, for the improvements and documentation.

> Would it be better to just link out to the netplan docs for the explanations of some of the properties?

I left a comment suggesting changes in the documentation. I think is better to refer to netplan's doc, rather than duplicating sub-props documentation.

> Is there a better way we can show that note/warning in the docs for the renderer limitations?

Furthermore, I included a `note` sphinx directive to better signalize that fact.

> We weren't too sure where a good place to add a log that alerts users when dhcp{4,6}-overrides are set and they are not using the networkd renderer. I didn't see any obvious place doing a high level validation like that, but maybe I missed it. We could add something just before the render gets called, if we have enough information then.

Maybe it's enough with the outlined docs, and we can leave the warning as a future improvement.

doc/rtd/topics/network-config-format-v2.rst
adobley: Thanks! This patch looks great!

adobley:
Thanks for all your help @aciba90! I think we've got everything. :)

aciba90 (approved):
LGTM! Thanks, @adobley, for this addition!

TheRealFalcon (commented):
If I use this config:
```
network:
  version: 2
  ethernets:
    eth0:
      dhcp4: true
      dhcp4-overrides:
        hostname: hal
        route-metric: 1100
        send-hostname: false
        use-dns: false
        use-domains: false
        use-hostname: false
        use-mtu: false
        use-ntp: false
        use-routes: false
      dhcp6: true
      dhcp6-overrides:
        hostname: hal
        route-metric: 1100
        send-hostname: false
        use-dns: false
        use-domains: false
        use-hostname: false
        use-mtu: false
        use-ntp: false
        use-routes: false
      match:
        macaddress: "00:11:22:33:44:55"
      nameservers:
        addresses: ["8.8.8.8", "2001:4860:4860::8888"]
```

I get a network file with this:
```
[DHCPv4]
Hostname=hal
RouteMetric=1100
SendHostname=False
UseDNS=False
UseDomains=False
UseHostname=False
UseMTU=False
UseNTP=False
UseRoutes=False

[DHCPv6]
Hostname=hal
RouteMetric=1100
SendHostname=False
UseDNS=False
UseDomains=False
UseHostname=False
UseMTU=False
UseNTP=False
UseRoutes=False

[Match]
MACAddress=00:11:22:33:44:55
Name=eth0

[Network]
DHCP=yes
DNS=8.8.8.8 2001:4860:4860::8888
```

But looking at [the networkd documentation](https://www.freedesktop.org/software/systemd/man/systemd.network.html#%5BDHCPv6%5D%20Section%20Options), it appears that most of those options aren't valid for `[DHCPv6]` using networkd.

Is the resulting `[DHCPv6]` section actually valid?

doc/rtd/topics/network-config-format-v2.rst
```diff
@@ -177,6 +179,42 @@ Enable DHCP for IPv4. Off by default.
...
+
+* ``hostname``
+* ``route-metric``
+* ``send-hostname``
+* ``use-dns``
+* ``use-domains``
+* ``use-hostname``
+* ``use-mtu``
+* ``use-ntp``
+
```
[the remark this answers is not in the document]
adobley: Thanks! We made some broader changes to the docs to cover the difference in configuration options between netplan and networkd. We were sure to add this when we made that change. :)

adobley:
@TheRealFalcon Thanks for catching that mistake! It does look like those properties are ignored for DHCPv6. We updated the code to reflect this and took a first pass at some tweaks to the documentation. Any feedback you have is appreciated!

cloudinit/net/networkd.py
```diff
@@ -271,7 +299,26 @@ def _render_content(self, ns):
...
+                            f"dhcp{version}domain" in device
+                            and "use-domains"
+                            in device.get(f"dhcp{version}-overrides", {})
+                        ):
+                            exception = (
+                                f"{name} has both dhcp{version}domain"
+                                f" and dhcp{version}-overrides.use-domains"
+                                f" configured. Use one"
+                            )
+                            LOG.critical(exception)
```
TheRealFalcon, line 318: Nit: An exception raised from here will make it into the logs, so we don't need an additional log here too.

TheRealFalcon (approved):
LGTM! One nit inline and we need to fix the black failure. Once all the checks pass, I can merge.

cloudinit/net/networkd.py
adobley: Fixed. Thanks!

TheRealFalcon:
Closing and re-opening to trigger Travis

TheRealFalcon (approved):
This looks good now. Thanks for the contribution!

PR #1712, opened 2022-09-01
lxd: fetch 1.0/devices content

aciba90:
## Proposed Commit Message

```
lxd: fetch 1.0/devices content

cloud-init awareness of 1.0/devices content will allow
DataSourceLXD to properly limit opt-in hotplug event-handling to
specific udev events based on device instead of just network type.

As devices content is not used atm, a param was added to
read_metadata to control its fetching (defaulting to False).

Note: read_metadata was refactored to try to reduce cyclotomatic
complexity.
```

## Additional Context

SC-608

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -224,6 +224,21 @@ def is_platform_viable() -> bool:
     return False
 
 
+def _get_json_response(
+    session: requests.Session, url: str, do_raise: bool = True
+):
+    url_response = _do_request(session, url, do_raise)
+    try:
+        return url_response.json()
+    except JSONDecodeError as exc:
+        raise sources.InvalidMetaDataException(
+            "Unable to process LXD config at {url}."
+            " Expected JSON but found: {resp}".format(
+                url=url, resp=url_response.text
+            )
+        ) from exc
+
+
 def _do_request(
     session: requests.Session, url: str, do_raise: bool = True
 ) -> requests.Response:
@@ -245,26 +260,21 @@ class _MetaDataReader:
         self.api_version = api_version
         self._version_url = url_helper.combine_url(LXD_URL, self.api_version)
 
-    def _get_meta_data(self, session: requests.Session) -> dict:
-        # Raw meta-data as text
-        md_route = url_helper.combine_url(self._version_url, "meta-data")
-        return {"meta-data": _do_request(session, md_route).text}
+    def _process_config(self, session: requests.Session) -> dict:
+        """Iterate on LXD API config items. Promoting CONFIG_KEY_ALIASES
 
-    def _get_config(self, session: requests.Session) -> dict:
+        Any CONFIG_KEY_ALIASES which affect cloud-init behavior are promoted
+        as top-level configuration keys: user-data, network-data, vendor-data.
+
+        LXD's cloud-init.* config keys override any user.* config keys.
+        Log debug messages if any user.* keys are overridden by the related
+        cloud-init.* key.
+        """
         config: dict = {"config": {}}
         config_url = url_helper.combine_url(self._version_url, "config")
         # Represent all advertized/available config routes under
         # the dict path {LXD_SOCKET_API_VERSION: {config: {...}}.
-        config_response = _do_request(session, config_url)
-        try:
-            config_routes = config_response.json()
-        except JSONDecodeError as exc:
-            raise sources.InvalidMetaDataException(
-                "Unable to determine cloud-init config from {route}."
-                " Expected JSON but found: {resp}".format(
-                    route=config_url, resp=config_response.text
-                )
-            ) from exc
+        config_routes = _get_json_response(session, config_url)
 
         # Sorting keys to ensure we always process in alphabetical order.
         # cloud-init.* keys will sort before user.* keys which is preferred
@@ -274,7 +284,6 @@ class _MetaDataReader:
             config_route_response = _do_request(
                 session, config_route_url, do_raise=False
             )
-
             if not config_route_response.ok:
                 LOG.debug(
                     "Skipping %s on [HTTP:%d]:%s",
@@ -306,34 +315,21 @@ class _MetaDataReader:
                     )
         return config
 
-    def _get_devices(self, session: requests.Session) -> dict:
-        url = url_helper.combine_url(self._version_url, "devices")
-        response = _do_request(session, url)
-        try:
-            devices = response.json()
-        except JSONDecodeError as exc:
-            raise sources.InvalidMetaDataException(
-                "Unable to determine LXD devices from {route}."
-                " Expected JSON but found: {resp}".format(
-                    route=url, resp=response.text
-                )
-            ) from exc
-        return {"devices": devices}
-
     def __call__(
         self, *, metadata_only: bool = False, devices: bool = False
     ) -> dict:
         with requests.Session() as session:
             session.mount(self._version_url, LXDSocketAdapter())
-            md: dict = {}
-            md.update(self._get_meta_data(session))
+            # Document API version read
+            md: dict = {"_metadata_api_version": self.api_version}
+            md_route = url_helper.combine_url(self._version_url, "meta-data")
+            md["meta-data"] = _do_request(session, md_route).text
             if metadata_only:
                 return md  # Skip network-data, vendor-data, user-data
-            # Document API version read
-            md["_metadata_api_version"] = self.api_version
-            md.update(self._get_config(session))
+            md.update(self._process_config(session))
             if devices:
-                md.update(self._get_devices(session))
+                url = url_helper.combine_url(self._version_url, "devices")
+                md.update({"devices": _get_json_response(session, url)})
             return md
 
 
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceLXD.py
```diff
@@ -223,14 +224,129 @@ def is_platform_viable() -> bool:
...
+        config_response = _do_request(session, config_url)
+        try:
+            config_routes = config_response.json()
+        except JSONDecodeError as exc:
+            raise sources.InvalidMetaDataException(
+                "Unable to determine cloud-init config from {route}."
+                " Expected JSON but found: {resp}".format(
+                    route=config_url, resp=config_response.text
+                )
+            ) from exc
```
blackboxsw, line 267: Maybe this is worth encapsulation as a function too. It'd allow us to reuse this behavior for any routes (like devices) where we expect JSON output and get something else. Then we could drop `_get_devices` and use a `_get_json_response(session, url)`  in both places with a parameterize message.

The diff looks like this. What do you think? better/worse?
```diff
diff --git a/cloudinit/sources/DataSourceLXD.py b/cloudinit/sources/DataSourceLXD.py
index 02c80b2ab..ad62fa1b9 100644
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -224,6 +224,21 @@ def is_platform_viable() -> bool:
     return False
 
 
+def _get_json_response(
+    session: requests.Session, url: str, do_raise: bool = True
+):
+    url_response = _do_request(session, url, do_raise)
+    try:
+        return url_response.json()
+    except JSONDecodeError as exc:
+        raise sources.InvalidMetaDataException(
+            "Unable to process LXD config at {url}."
+            " Expected JSON but found: {resp}".format(
+                url=url, resp=url_response.text
+            )
+        ) from exc
+
+
 def _do_request(
     session: requests.Session, url: str, do_raise: bool = True
 ) -> requests.Response:
@@ -245,26 +260,21 @@ class _MetaDataReader:
         self.api_version = api_version
         self._version_url = url_helper.combine_url(LXD_URL, self.api_version)
 
-    def _get_meta_data(self, session: requests.Session) -> dict:
-        # Raw meta-data as text
-        md_route = url_helper.combine_url(self._version_url, "meta-data")
-        return {"meta-data": _do_request(session, md_route).text}
+    def _process_config(self, session: requests.Session) -> dict:
+        """Iterate on LXD API config items. Promoting CONFIG_KEY_ALIASES
 
-    def _get_config(self, session: requests.Session) -> dict:
+        Any CONFIG_KEY_ALIASES which affect cloud-init behavior are promoted
+        as top-level configuration keys: user-data, network-data, vendor-data.
+
+        LXD's cloud-init.* config keys override any user.* config keys.
+        Log debug messages if any user.* keys are overridden by the related
+        cloud-init.* key.
+        """
         config: dict = {"config": {}}
         config_url = url_helper.combine_url(self._version_url, "config")
         # Represent all advertized/available config routes under
         # the dict path {LXD_SOCKET_API_VERSION: {config: {...}}.
-        config_response = _do_request(session, config_url)
-        try:
-            config_routes = config_response.json()
-        except JSONDecodeError as exc:
-            raise sources.InvalidMetaDataException(
-                "Unable to determine cloud-init config from {route}."
-                " Expected JSON but found: {resp}".format(
-                    route=config_url, resp=config_response.text
-                )
-            ) from exc
+        config_routes = _get_json_response(session, config_url)
 
         # Sorting keys to ensure we always process in alphabetical order.
         # cloud-init.* keys will sort before user.* keys which is preferred
@@ -274,7 +284,6 @@ class _MetaDataReader:
             config_route_response = _do_request(
                 session, config_route_url, do_raise=False
             )
-
             if not config_route_response.ok:
                 LOG.debug(
                     "Skipping %s on [HTTP:%d]:%s",
@@ -306,34 +315,21 @@ class _MetaDataReader:
                     )
         return config
 
-    def _get_devices(self, session: requests.Session) -> dict:
-        url = url_helper.combine_url(self._version_url, "devices")
-        response = _do_request(session, url)
-        try:
-            devices = response.json()
-        except JSONDecodeError as exc:
-            raise sources.InvalidMetaDataException(
-                "Unable to determine LXD devices from {route}."
-                " Expected JSON but found: {resp}".format(
-                    route=url, resp=response.text
-                )
-            ) from exc
-        return {"devices": devices}
-
     def __call__(
         self, *, metadata_only: bool = False, devices: bool = False
     ) -> dict:
         with requests.Session() as session:
             session.mount(self._version_url, LXDSocketAdapter())
-            md: dict = {}
-            md.update(self._get_meta_data(session))
+            # Document API version read
+            md: dict = {"_metadata_api_version": self.api_version}
+            md_route = url_helper.combine_url(self._version_url, "meta-data")
+            md["meta-data"] = _do_request(session, md_route).text
             if metadata_only:
                 return md  # Skip network-data, vendor-data, user-data
-            # Document API version read
-            md["_metadata_api_version"] = self.api_version
-            md.update(self._get_config(session))
+            md.update(self._process_config(session))
             if devices:
-                md.update(self._get_devices(session))
+                url = url_helper.combine_url(self._version_url, "devices")
+                md.update({"devices": _get_json_response(session, url)})
             return md
 
 
diff --git a/tests/unittests/sources/test_lxd.py b/tests/unittests/sources/test_lxd.py
index 8a8d58e03..c98593950 100644
--- a/tests/unittests/sources/test_lxd.py
+++ b/tests/unittests/sources/test_lxd.py
@@ -246,7 +246,7 @@ class TestReadMetadata:
                     "http://lxd/1.0/config": "[NOT_JSON",
                 },
                 InvalidMetaDataException(
-                    "Unable to determine cloud-init config from"
+                    "Unable to process LXD config at"
                     " http://lxd/1.0/config. Expected JSON but found:"
                     " [NOT_JSON"
                 ),
@@ -320,7 +320,7 @@ class TestReadMetadata:
                     "http://lxd/1.0/devices": '{"root"',
                 },
                 InvalidMetaDataException(
-                    "Unable to determine LXD devices from"
+                    "Unable to process LXD config at"
                     ' http://lxd/1.0/devices. Expected JSON but found: {"root"'
                 ),
                 [
```

cloudinit/sources/DataSourceLXD.py
```diff
@@ -223,14 +224,129 @@ def is_platform_viable() -> bool:
...
+        )
+    return response
+
+
+class _MetaDataReader:
+    def __init__(self, api_version: str = LXD_SOCKET_API_VERSION):
+        self.api_version = api_version
+        self._version_url = url_helper.combine_url(LXD_URL, self.api_version)
+
+    def _get_meta_data(self, session: requests.Session) -> dict:
```
blackboxsw, line 248: _get_meta_data doesn't really have a lot to it. Can we just avoid this method definition and call it inline.

cloudinit/sources/DataSourceLXD.py
```diff
@@ -223,14 +224,129 @@ def is_platform_viable() -> bool:
...
+            session.mount(self._version_url, LXDSocketAdapter())
+            md: dict = {}
+            md.update(self._get_meta_data(session))
+            if metadata_only:
+                return md  # Skip network-data, vendor-data, user-data
+            # Document API version read
+            md["_metadata_api_version"] = self.api_version
+            md.update(self._get_config(session))
+            if devices:
+                md.update(self._get_devices(session))
```
blackboxsw, line 336: For the hotplug case where we potentially trigger python invocation from a udev event, we'd want this operation to only touch the devices route I presume to avoid two hits on the metadata service for every udev event. I wonder if we might want to consolidate `metadata_only` and `devices` bools into a single parameter `fitler_keys: List[str]` or `request_keys` or `filter_routes` or some other parameter name which would limit the API routes contacted to just the desired keys. Then we have flexibility to get all, some or only one API route depending on our needs

blackboxsw (changes requested):
Looks really good @aciba90 great cleanup. One thing I really want to think about is how to specifically get only devices when needed (and not involve getting metadata too). That will improve performance of the optional hotplug behavior in the future. The other "diff" take what you this is reasonable.

cloudinit/sources/DataSourceLXD.py
aciba90: Nice suggestion applied. Thanks for putting together the diff.

cloudinit/sources/DataSourceLXD.py
aciba90: Done.

cloudinit/sources/DataSourceLXD.py
aciba90: Good point. I have implemented an `enum.Flag` to control which keys to retrieve.

aciba90:
Thanks for the review and constructive comments, @blackboxsw.

I have applied the diff and implemented an enum.Flag to control what meta-data keys to retrieve.

This PR is ready to be reviewed after CI passes.

blackboxsw (approved):
Thanks for this addition. I think MetaDataKeys(Flag) makes a lot of sense here and opens up our hotplug possibilities.

Ran this through both `user.user-data` and `cloud-init.user-data` settings to ensure proper behavior is retained and skipping `user.*` settings  when `cloud-init.*` config is provided.

PR #1716, opened 2022-09-02
Fix cc_chef typing issue (SC-1244)

TheRealFalcon:
## Proposed Commit Message

```
Fix cc_chef typing issue

The function 'subp_blob_in_tempfile' had an if branch that is currently
unused, but attempts to invalidly append a list to a tuple.
```


holmanb (approved):
Thanks for the added tests, and `subp_blob_in_tempfile()` is much easier to read than before. 

LGMT

PR #1721, opened 2022-09-08
tests: Fix ntp installation

aciba90:
## Proposed Commit Message

```
tests: Fix ntp installation

On daily images, the installation of ntp might fail due unmet deps.
Add update and upgrade to these test to avoid that.
```

## Additional Context

https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-kinetic-lxd_vm/81
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-kinetic-lxd_container/77

## Test Steps

```bash
CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_OS_IMAGE=kinetic tox -e integration-tests -- -vv tests/integration_tests/modules/test_ntp_servers.py
```

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_ntp_servers.py
+++ b/tests/integration_tests/modules/test_ntp_servers.py
@@ -16,6 +16,8 @@ from tests.integration_tests.instances import IntegrationInstance
 
 USER_DATA = """\
 #cloud-config
+package_update: true
+package_upgrade: true
 ntp:
   ntp_client: ntp
   servers:
@@ -74,6 +76,8 @@ class TestNtpServers:
 
 CHRONY_DATA = """\
 #cloud-config
+package_update: true
+package_upgrade: true
 ntp:
   enabled: true
   ntp_client: chrony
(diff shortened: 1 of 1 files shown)


aciba90:
After a team discussion, this would add significant overload of resources, therefore closed. 
We will make sure Jenkins jobs show the simplestream associated to the run.

blackboxsw:
Confirmed that our integration test runs will emit /etc/cloud/build.info on all jenkins runs, so we can confirm we are using the appropriate daily build stream for images.

 On development releases, we expect occasional breakage in when apt package dependencies and conflicts are added to the deb control files. Those issues should be resolved once the next daily build is published. I believe the low frequency of that type of breakage affecting our tests doesn't outweigh the cost of continuously performing an apt update, apt upgrade in this test every single integration run.

If that frequency of breaking due to packaging changes increases and is unsustainable on the devel-series giving us a poor test result success vs failure signal, then we can look at instrumenting this type of approach on the base image we create and test specifically on the devel release only.

PR #1722, opened 2022-09-08
Allow jinja templating in /etc/cloud (SC-1170)

TheRealFalcon:
## Proposed Commit Message

```
Allow jinja templating in /etc/cloud

There have been multiple requests to allow jinja templating in
/etc/cloud configs the same way we allow jinja templating in vendordata
and userdata. This commit allows for templating both
/etc/cloud/cloud.cfg and any file in /etc/cloud/cloud.cfg.d. The same
instance data used for substitution in vendordata and userdata will be
used here.

Note that these configs get loaded multiple times during the lifetime
of cloud-init, and during cloud-init's earlier loads, instance data
is not yet available.

LP: #1913461
```

## Additional Context
There's little documentation about how our configuration gets loaded and overlaid, so there's really any documentation to be updated. We have a separate ticket for clarifying docs around /etc/cloud, and I think the changes here can be reflected there.

## Test Steps
See new integration tests

## Checklist:

### DIFF
--- a/cloudinit/handlers/jinja_template.py
+++ b/cloudinit/handlers/jinja_template.py
@@ -10,7 +10,11 @@ from cloudinit import handlers
 from cloudinit import log as logging
 from cloudinit.helpers import Paths
 from cloudinit.settings import PER_ALWAYS
-from cloudinit.templater import MISSING_JINJA_PREFIX, render_string
+from cloudinit.templater import (
+    MISSING_JINJA_PREFIX,
+    detect_template,
+    render_string,
+)
 from cloudinit.util import b64d, json_dumps, load_file, load_json
 
 JUndefinedError: Type[Exception]
@@ -23,6 +27,14 @@ except ImportError:
     operator_re = re.compile(r"[-.]")
 
 LOG = logging.getLogger(__name__)
+
+
+class JinjaLoadError(Exception):
+    pass
+
+
+class NotJinjaError(Exception):
+    pass
 
 
 class JinjaTemplatePartHandler(handlers.Handler):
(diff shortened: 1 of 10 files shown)


cloudinit/stages.py
```diff
@@ -526,11 +529,10 @@ def _default_handlers(self, opts=None):
             ShellScriptByFreqPartHandler(PER_INSTANCE, **opts),
             ShellScriptByFreqPartHandler(PER_ONCE, **opts),
             BootHookPartHandler(**opts),
+            JinjaTemplatePartHandler(
```
holmanb, line 532: +1 Nice cleanup

cloudinit/handlers/jinja_template.py
```diff
@@ -82,21 +94,26 @@ def render_jinja_payload_from_file(
     @return: A string of jinja-rendered content with the jinja header removed.
         Returns None on error.
     """
+    if detect_template(payload)[0] != "jinja":
+        raise NotJinjaError("Payload is not a jinja template")
```
aciba90, line 98: I think this exception should be handled in [cloudinit/cmd/devel/render.py::handle_args](https://github.com/canonical/cloud-init/pull/1722/files#diff-900ac554685fc30e9e71da127c37f66d08e6b7d10a64aa50c4dc42b71d6a68b3R56), otherwise the `render` command will raise it and exit with non 0.

Additionally, I think we should also unit-test cover the interaction between `util.read_conf`, `cmd.devel.render` and this exception.

cloudinit/util.py
```diff
@@ -287,15 +287,50 @@ def rand_dict_key(dictionary, postfix=None):
...
+                "configuration loaded from '%s'",
+                instance_data_file,
+                fname,
+            )
+        except NotJinjaError:
+            # A log isn't appropriate here as we generally expect most
+            # cloud.cfgs to not be templated. The other path is logged
+            pass
+        except JinjaLoadError:
+            LOG.warning(
```
aciba90, line 325: Nit: I would include here the exception message, as `render_jinja_payload_from_file` is handling [OSError](https://github.com/canonical/cloud-init/pull/1722/files#diff-fb04a669500887728f7bbca305f11ba9727ad3ca8adab8e02f2cdddf11058984R113) exceptions and otherwise that info would not reach the log.

I realize this fact (insance_data_file not readable) is very unlikely, as generally this execution path is executed as root, but `render_jinja_payload_from_file`'s exception handling could grow over time, and this log msg won't inform users of those different msgs.

aciba90 (changes requested):
Looks good overall. Thanks, @TheRealFalcon, for this addition.

I left a couple of comments. One related to an unhandled exception during `cloud-init devel render` and another nit one related to logging.

TheRealFalcon:
Updated based on comments. Also refactored some of the render.py code.

cloudinit/util.py
```diff
@@ -1001,18 +1043,29 @@ def read_conf_d(confd):
...
-        cfg = read_conf(cfgfile)
+        cfg = read_conf(cfgfile, instance_data_file=instance_data_file)
     except OSError as e:
         if e.errno == EACCES:
             LOG.warning("REDACTED config part %s for non-root user", cfgfile)
     else:
         cfgs.append(cfg)
 
-    confd = False
+    confd = ""
```
holmanb, line 1068: :D

holmanb (approved):
LGTM. Leaving open in case @aciba90 has anything else to add. I'll merge this tomorrow if not.

aciba90 (approved):
LGTM! Thanks, @TheRealFalcon and @holmanb.

PR #1725, opened 2022-09-09
define new attribute for pre-22.3 pickles

holmanb:
```
azure: Properly unpickle pre-22.3 pickles.

A new attribute was added to DataSourceAzure[1].
Since the base class uses CloudInitPickleMixin,
we need to define this new attribute in _unpickle()

Add multiple tests to improve pickle coverage.

[1] https://github.com/canonical/cloud-init/pull/1523
```

Error logs:
```python
> grep -A5 -B5 ERROR log
2022-09-09 17:28:03,582 - util.py[DEBUG]: Crawl of Azure Instance Metadata Service (IMDS) took 0.023 seconds
2022-09-09 17:28:03,582 - handlers.py[DEBUG]: finish: azure-ds/get_metadata_from_imds: SUCCESS: get_metadata_from_imds
2022-09-09 17:28:03,583 - handlers.py[DEBUG]: finish: azure-ds/get_imds_data_with_api_fallback: SUCCESS: get_imds_data_with_api_fallback
2022-09-09 17:28:03,583 - handlers.py[DEBUG]: finish: azure-ds/crawl_metadata: FAIL: crawl_metadata
2022-09-09 17:28:03,583 - util.py[DEBUG]: Crawl of metadata service took 0.568 seconds
2022-09-09 17:28:03,583 - azure.py[ERROR]: Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
2022-09-09 17:28:03,584 - handlers.py[DEBUG]: start: azure-ds/_report_failure: _report_failure
2022-09-09 17:28:03,584 - azure.py[DEBUG]: Using cached ephemeral dhcp context to report failure to Azure
2022-09-09 17:28:03,585 - handlers.py[DEBUG]: start: azure-ds/report_failure_to_fabric: report_failure_to_fabric
2022-09-09 17:28:03,585 - handlers.py[DEBUG]: start: azure-ds/register_with_azure_and_report_failure: register_with_azure_and_report_failure
2022-09-09 17:28:03,585 - handlers.py[DEBUG]: start: azure-ds/_fetch_goal_state_from_azure: _fetch_goal_state_from_azure
--
2022-09-09 17:28:06,398 - util.py[DEBUG]: Crawl of Azure Instance Metadata Service (IMDS) took 0.024 seconds
2022-09-09 17:28:06,399 - handlers.py[DEBUG]: finish: azure-ds/get_metadata_from_imds: SUCCESS: get_metadata_from_imds
2022-09-09 17:28:06,399 - handlers.py[DEBUG]: finish: azure-ds/get_imds_data_with_api_fallback: SUCCESS: get_imds_data_with_api_fallback
2022-09-09 17:28:06,399 - handlers.py[DEBUG]: finish: azure-ds/crawl_metadata: FAIL: crawl_metadata
2022-09-09 17:28:06,399 - util.py[DEBUG]: Crawl of metadata service took 0.542 seconds
2022-09-09 17:28:06,399 - azure.py[ERROR]: Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
2022-09-09 17:28:06,410 - handlers.py[DEBUG]: start: azure-ds/_report_failure: _report_failure
2022-09-09 17:28:06,410 - azure.py[DEBUG]: Using cached ephemeral dhcp context to report failure to Azure
2022-09-09 17:28:06,410 - handlers.py[DEBUG]: start: azure-ds/report_failure_to_fabric: report_failure_to_fabric
2022-09-09 17:28:06,410 - handlers.py[DEBUG]: start: azure-ds/register_with_azure_and_report_failure: register_with_azure_and_report_failure
2022-09-09 17:28:06,410 - handlers.py[DEBUG]: start: azure-ds/_fetch_goal_state_from_azure: _fetch_goal_state_from_azure
```

Do I understand correctly that all new attributes to these pickle mixin classes will require a definition in `_unpickle()`? If my last statement is correct, I think this fact could maybe be better documented and/or tested.


TheRealFalcon:
> Do I understand correctly that all new attributes to these pickle mixin classes
will require a definition in _unpickle()?

Yes. The real question is why didn't the upgrade test catch it. It should be installing the latest daily cloud image, and then upgrading to proposed. Part of the test is checking instance id.

holmanb:
>  The real question is why didn't the upgrade test catch it

I'm not 100% sure I have an answer for this, but I have a theory and an EOD approaching so I'll share what I have so far. 


Notes so far:

1. test_upgrade.py calls `cloud-init init`
2. based on log analysis, I don't see any calls to `get_data()` during init-network stage (`get_data()` is where the exception is thrown in the logs)
3. supporting this theory:

`_reported_ready_marker_file` is referenced 4 times:

```
_create_report_ready_marker()
_poll_imds()
_determine_pps_type()
_cleanup_markers()
```

all of these functions are only used in DataSourceAzure.py and only appear to be called in codepaths originating from `get_data()`

4. stage 6 calls `init.instancify()`

To summarize:
=============
`cloud-init init` doesn't actually reference this variable, so no exeption will be thrown in this stage. `cloud-init init` _does_ write out state, including instance-id

This doesn't explain why the reboot in the test didn't hit a variable reference. This is the only hole in this theory I have this far.

Thoughts?

TheRealFalcon:
Could you replace `cloud-init init` with `cloud-init init --local; cloud-init init` to test your idea?

blackboxsw:
Here's an additioanl PR that could supplement unittests. Ideally I think we probably want to inspect the attributes that __init__ sets on a current datasource and compare against the unpickled datasource to make sure all initialized attributes from __init__ happen to also be defined on the unpickled old cache 
https://github.com/holmanb/cloud-init/pull/45

holmanb:
> Here's an additioanl PR that could supplement unittests. Ideally I think we probably want to inspect the attributes that **init** sets on a current datasource and compare against the unpickled datasource to make sure all initialized attributes from **init** happen to also be defined on the unpickled old cache [holmanb#45](https://github.com/holmanb/cloud-init/pull/45)

Thanks @blackboxsw, I just left a question on the PR. The first half looks good, but I don't understand the second part.

holmanb:
Here are a couple of other thoughts on how we might avoid future pickling issues:

1. `__init__()` is not called during unpickle, but `__enter__()` is. Could we use this to our advantage and either convert `__init__()` calls to `__enter__()` wherever possible so this doesn't happen again?

2. assuming 1. isn't an option, we could have a simple test in `tests/unittests/test_upgrade.py` that just registers a serialized version of each type of pickled object, and then have the test fail when the serialized object doesn't match the saved key. Each change to attributes would then have a "manual registration" of an attribute change where the dev/reviewer sees a big comment block reminding them/us to add the attribute to `_dump()`.

I'd like to see option 1. if possible. Option 2. would require more ongoing upkeep which is not a joy to maintain.

TheRealFalcon:
The integration test usually finds these issues. Regardless of what we do, we should update that one to ensure it founds the issue we discovered here.

> __init__() is not called during unpickle, but __enter__() is. Could we use this to our advantage and either convert __init__() calls to __enter__() wherever possible so this doesn't happen again?

Unless I'm misunderstanding you, `__enter__` isn't part of object initialization. It is called when constructing a context manager. We can't convert `__init__`s to `__enter__`s

```python
>>> class Test:
...     def __init__(self):
...             print("in init")
...     def __enter__(self):
...             print("in enter")
... 
>>> Test()
in init
<__main__.Test object at 0x7f21e63c7910>
>>>
>>> class Test:
...     def __enter__(self):
...             print("in enter")
... 
>>> 
>>> Test()
<__main__.Test object at 0x7f21e62f1a20>
```

2 is a good idea, but our pickling can also break across python versions ( #857 ). We'd just need to double check that the same pickle can work across all of our supported versions.

holmanb:
> Unless I'm misunderstanding you, `__enter__` isn't part of object initialization. It is called when constructing a context manager. We can't convert `__init__`s to `__enter__`s

You're right on __enter__.  I mixed up my dunders, my apologies. I meant to say __new__, not __enter__.

blackboxsw:
Here's the other issue with the integraiton test, the ERROR raised by azure didn't result in a cloud-init status error.
 Here's a diff that'll help us assert we don't have warnings errors or traces and that we also invoke verify_clean_logs across the reboot.
```diffdiff --git a/tests/integration_tests/test_upgrade.py b/tests/integration_tests/test_upgrade.py
index b13d4703..1ea50926 100644
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -185,4 +185,6 @@ def test_subsequent_boot_of_upgraded_package(session_cloud: IntegrationCloud):
             source, take_snapshot=False, clean=False
         )
         instance.restart()
+        log = instance.read_from_file("/var/log/cloud-init.log")
+        verify_clean_log(log)
         assert instance.execute("cloud-init status --wait --long").ok
diff --git a/tests/integration_tests/util.py b/tests/integration_tests/util.py
index 18ca1917..489fe622 100644
--- a/tests/integration_tests/util.py
+++ b/tests/integration_tests/util.py
@@ -47,6 +47,12 @@ def verify_clean_log(log: str, ignore_deprecations: bool = True):
         )
         log = "\n".join(log_lines)
 
+    error_logs = re.findall("ERROR.*", log)
+    if error_logs:
+        raise AssertionError(
+            "Found unexpected errors: %s" % '\n'.join(error_logs)
+        )
+
     warning_count = log.count("WARN")
     expected_warnings = 0
     traceback_count = log.count("Traceback")
```

holmanb:
@TheRealFalcon Regarding using `__new__` for initialization rather than `__init__`.

The big downside to this angle is that attributes would be class attributes, which may not work in all cases.

```python
#!/usr/bin/env python3

class Base(object):
    def __new__(cls, **kwargs):
        obj = object().__new__(cls)
        return cls.assign(obj, **kwargs)

    @staticmethod
    def assign(obj, **kwargs):
        for key, value in kwargs.items():
            obj.__dict__[key] = value
        return obj

    def __str__(self):
        rep = ''
        for key, value in self.__dict__.items():
            rep += f"{key}: {value}\n"
        return rep


class Thing(Base):
    _attrs = {
        "age": 92,
        "weight": 2
    }

    def __new__(cls, **kwargs):
        obj = super().__new__(Base)
        return cls.assign(obj, **kwargs, **(cls._attrs))


if __name__ == "__main__":
    p = Thing(uuid=1)
    print(p)
```

Note that `uuid` isn't predefined in the instance attributes, but using `__new__` for instantiation enables this new attribute to be inserted into `__dict__`. In our case we could put vars we would typically initialize in `__init__()` into the class variable `_attrs`, which would get inserted into `__dict__` during unpickle (even when the pickle is missing those attributes).

This does allow any attribute to be assigned as an attribute, which was useful for this demonstration but we wouldn't actually want, so we could always add a little check to ensure that there aren't any attributes in `__dict__` that weren't in `_attr`. This could be done, for example, by changing the `Thing` class to look like:
```python
class Thing(Base):
    _attrs = {
        "age": 92,
        "weight": 2
    }

    def __new__(cls, **kwargs):
        obj = super().__new__(Base)
        obj = cls.assign(obj, **kwargs, **(cls._attrs))
        diff = set(obj.__dict__.keys()) - set(cls._attrs.keys())
        if diff:
            raise AttributeError(f"Unexpected keys {diff}")
```

blackboxsw:
@holmanb I refreshed the pkl from azure to be 20.1 based which didn't contain a vendordata2 attribute, Then no need to check on Azure vs non-Azure of pkls as all pkls are old enough in this case  to warrant an __init__ value of None instead of the MIME data. Also we can test the intialized new class `__dict__` vs the loaded pickle to see if any differences exist showing a lack of setup of newly added attributes across the pickle load. This shows too that NoCloud datasource also needs a seed_dir definition as well that went missed by unittests and integration tests.

cjp256:
@holmanb @blackboxsw @TheRealFalcon THANK YOU for following up on this so quicky!

In addition to the patch in this PR, I'd like to address the underlying failure mode (perhaps before merging this PR while we have an obvious flaw to validate against).
- When returning `False` on `_get_data()`, as is done here by the datasource,  it would seem to me that there should be no dependency on `ds.metadata`, here or elsewhere.  Would raising `sources.InvalidMetaDataException` instead of returning `False` make more sense?  Or would this cause it to fall back to `DataSourceNone` with a similar trajectory?
- I can see why `metadata` wasn't updated during the failure, but why was it not already present from the previous run?  Is it not unpickled?
- Is falling back to `"iid-datasource"` a good idea for any datasource?
- Perhaps the data sources should implement `get_instance_id()` instead of relying on populated metadata?

A few related thoughts on testing on our side:
- I usually do ad-hoc testing of upgrade paths when I change the pickled attributes, but I did not notice the cloud_dir PR making this change.
- It doesn't negatively affect provisioning, which is the focus of most of our own testing.
- There is a lot to learn from here that I will be incorporating into our own test coverage.

TheRealFalcon:
I think we should also incorporate the diff from https://github.com/canonical/cloud-init/pull/1725#issuecomment-1242764331

blackboxsw:
@holmanb the "seed_dirs" check failure I mentioned in the NoCloud datasource is a known/acceptable way for a datasource to approach ensuring that the attribute gets set across unpickle. So our unit test needs a little tweak

```diff
diff --git a/tests/integration_tests/util.py b/tests/integration_tests/util.py
index 6b09f26c2..f955e94ec 100644
--- a/tests/integration_tests/util.py
+++ b/tests/integration_tests/util.py
@@ -50,7 +50,7 @@ def verify_clean_log(log: str, ignore_deprecations: bool = True):
     error_logs = re.findall("ERROR.*", log)
     if error_logs:
         raise AssertionError(
-            "Found unexpected errors: %s" % '\n'.join(error_logs)
+            "Found unexpected errors: %s" % "\n".join(error_logs)
         )
 
     warning_count = log.count("WARN")
diff --git a/tests/unittests/test_upgrade.py b/tests/unittests/test_upgrade.py
index 28f2cc343..218915ed6 100644
--- a/tests/unittests/test_upgrade.py
+++ b/tests/unittests/test_upgrade.py
@@ -50,7 +50,15 @@ class TestUpgrade:
         distro = previous_obj_pkl.distro
         paths = previous_obj_pkl.paths
         ds = ds_class(sys_cfg, distro, paths)
-        assert set() == ds.__dict__.keys() - previous_obj_pkl.__dict__.keys()
+        if ds.dsname == "NoCloud" and previous_obj_pkl.__dict__:
+            expected = (
+                set({"seed_dirs"}),  # LP: #1568150 handled with getattr checks
+                set(),
+            )
+        else:
+            expected = (set(),)
+        missing_attrs = ds.__dict__.keys() - previous_obj_pkl.__dict__.keys()
+        assert missing_attrs in expected
 
     def test_networking_set_on_distro(self, previous_obj_pkl):
         """We always expect to have ``.networking`` on ``Distro`` objects."""
```

blackboxsw:
> @holmanb @blackboxsw @TheRealFalcon THANK YOU for following up on this so quicky!

Thanks for bringing this to our attention last week @cjp256 . Good to improve our gaps in both integration and unit testing here.
I think a lot of the problem we have was due to our integration test not validating full logs on the upgrade reboot case, nor triggering the `cloud-init init --local` on the one case where we did validate logs. 



> 
> In addition to the patch in this PR, I'd like to address the underlying failure mode (perhaps before merging this PR while we have an obvious flaw to validate against).

I think we've now addressed a number of gaps in this current PR that should remedy most of the errors missed.
Also on Jammy we can easily reproduce the current failure path by:
```bash
# launch jammy on Azure 
$ sudo add-apt-repository ppa:chad.smith/azure-upgrade-test   # add visibility to upgrade to reproducer version
$ sudo apt install cloud-init=22.1-14-g2e17a0d6-0ubuntu1~22.04.5  # downgrade to exposed version 
$ sudo cloud-init clean --logs --reboot
$ sudo cloud-init status --wait --long
$ sudo apt install cloud-init=22.3-3-g9f0efc47-0ubuntu1~22.04.1  # install reproducer
$ sudo reboot
$ grep ERROR /var/log/cloud-init.log
$ sudo cloud-init status --long # note no error
```

The remaining missed failure modes that are now addressed:

* integration test log validation expected to see top-level "errors" surfaced by `cloud-init status` output, but azure error handling resulted in only LOG.error in logs without raising official Exceptions/Tracebacks which left cloud-init status in working state
  * ✔add full log validation looking for any `ERROR` logs inline which may not bubble up to full cloud-init failure ([we already checked `Tracebacks/WARNINGS`](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/util.py#L91-L95))
* integration test_upgrade only validating logs for invoking `cloud-init init` on the live system.
    * ✔invoke all cloud-init stages on live system before log validation
    * ✔added full log validation after reboot too 

> 
>     * When returning `False` on `_get_data()`, as is done here by the datasource,  it would seem to me that there should be no dependency on `ds.metadata`, here or elsewhere.

The datasource didn't actually return False here, it returned True based on finding the cache on Jammy.  it just reported errors  when it tried to use that cache and something in Azure ephemeral dhcp setup/retries ended up swallowing that error to some level which didn't bubble up to something more critical such as DS Not found. This needs investigation from us to see what we can do to better raise that type or error condition from boot on Azure.

```
2022-09-12 17:05:44,281 - stages.py[DEBUG]: restored from checked cache: DataSourceAzure [seed=/var/lib/waagent]2022-09-12 17:05:44,281 - handlers.py[DEBUG]: finish: init-local/check-cache: SUCCESS: restored from checked cache: DataSourceAzure [seed=/var/lib/waagent]
```



  Would raising `sources.InvalidMetaDataException` instead of returning `False` make more sense?  Or would this cause it to fall back to `DataSourceNone` with a similar trajectory?

This is tricky, something in Azures PPS early boot and error reporting is actually hitting a solid error that should raise a more significant problem, but we are retrying or ignoring that error and trying to use ephemerral dhcp cached information to report that failure and silently move on beyond that. I'd say from a datasource perspective, if we know we are on a datasource that "had" instance-id properly defined, we probably don't ever want to silently "fallback" to `iid-datasource` and claim success there.  If a cloud has known needs to quarantine a VM to make sure they can no longer see IMDS for one reason or another, cloud-init should probably better define explicit behavior in this case to avoid reprovisioning the hostkeys on the system as a side-effect. 



> 
>     * I can see why `metadata` wasn't updated during the failure, but why was it not already present from the previous run?  Is it not unpickled?

The reason metadata is `None` is because Azure tried to made a call to `clear_cached_attrs` due to the event ` Update datasource metadata and network config` which also clears out the metadata/userdata/vendordata attributes prior to the _get_data run. But, that run failed to actually resent `self.metadata` due to the traceback.



> 
>     * Is falling back to `"iid-datasource"` a good idea for any datasource?
As far as I recall, `iid-datasource` is typical/historically used in the VMWare/OVF platforms. Falling back from some other instance-id UUID -> `iid-datasource` could mean someone has tried creating a golden image from some cloud and then moved it to another cloud platform which doesn't define instance-id or that the VM has been moved into quarantine without access to the originally detected IMDS. While this is certainly very uncommon a migration path, it is something folks **may** do.


> 
>     * Perhaps the data sources should implement `get_instance_id()` instead of relying on populated metadata?

Yes that would work. Relying on populated metadata is just a common pattern for looking for metadata['instance-id'] if it exists.

> 
> 
> A few related thoughts on testing on our side:
> 
>     * I usually do ad-hoc testing of upgrade paths when I change the pickled attributes, but I did not notice the cloud_dir PR making this change.
> 
>     * It doesn't negatively affect provisioning, which is the focus of most of our own testing.
> 
>     * There is a lot to learn from here that I will be incorporating into our own test coverage.

Same here; and the unpickling of the cached datasource is the source of most of our upgrade problems we've seen historically. We tried to add adequate coverage for the upgrade path issues we've seen, but missed a couple of corner cases that I think this PR shores up. We are also tracking the following additional (separate) work items.

- check for unexpected SSH host key change across upgraded reboot (indicating cloud-init re-provisioned because it lost the instance-id somehow)
- add more datasource stale pkls beyond just Azure/NoCloud to ensure other datsource custom upgrade paths don't get broken in unittests (earlier exposure to this problem than SRU validation testing)

blackboxsw:
> @blackboxsw Looks good, I'll apply that one too. The use of `getattr` in the commit for that LP issue seems like a smell, but that's not something to resolve today, I guess.
Yeah I agree, it is a smell, and makes the code a bit busier to maintain, this getattr check in NoCloud datasource pre-exists the better unpickle handling. I'd be game for use moving that logic into one place and simplifying runtime (non-init) code.

TheRealFalcon:
> @TheRealFalcon Up to you if we include this bit. The comment directly above your change implies that we're working around internal behavior to ensure that we hit dirty cache values, which makes it sound like running all stages in order would undo some of the pre-existing stages.

We were previously working around the Azure pickle remover. The point is to re-run cloud-init without a reboot. I don't think running all stages would undo anything. IIRC, the pickle remover is still present on older releases, so we should keep this around.

> I'm sure this test has caught bugs, but it seems fragile and I'm not sure how best to improve it without undoing previous test coverage. Let me know what you think, maybe I'm wrong.

What do you think is fragile about it?

holmanb:
> What do you think is fragile about it?

The attribute has to referenced in the codepath taken by this test for the exception to throw, but this test doesn't cover all codepaths.

blackboxsw (approved):
LGTM. This branch fixes the specific upgrade issue, shores up some of the unit and integration testing gaps to ensure we don't hit this pickling type problem in the future for Azure. We also have the specific PPA `sudo add-apt-repository ppa:chad.smith/azure-upgrade-test` to allow us to test upgrade path failures in the future to ensure Azure's early boot config can fail more noisily in the future. Ideally, I think we want Azure datasource to emit a hard failure if we are receiving unexpected exceptions  in `_get_data` in early boot. It seems like the report failure to backplane somehow let's us limp past critical failures if we happen to be already using a cached datasource.

blackboxsw:
@holmanb @TheRealFalcon @cjp256 I think the actions we need to take after this PR are the following:
1. [update_metdata_if_supported is too permissive of failures](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L822-L824) when run on an already discovered datasource. We should make this a hard failure or at least a warning if metadata disappears because DataSource.clear_cached_attrs will have blown away local metadata, vendordata, user-data attribute values and that cleared instance will soon be persisted to disk if we ignore the get_data failure.
2. Azure probably wants to [raise a hard-error if crawl_metata  fails after we confirm `is_platform_viable(Path(self.seed_dir)`](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L732) because we know we are on azure and expect that IMDS will just work. Returning False here just lets cloud-init plug along without error when trying to update metadata. One thing we need check with @cjp256 and @anhvoms are wether there are expected conditions where a VM might be migrated to a quarantine/sandbox with IMDS access turned off. If, so, then Azure datasource might have to handle a different type or remediation for the scenario where IMDS is expected to by 'down'

cjp256:
So IIUC we are returning False, which works for the normal case, but doesn't help the update_metadata_is_supported() case after it purged the cache?

Should raising sources.InvalidMetaDataException address the update_metadata_is_supported() case without additional changes?  Should this cause cloud-init to fail, or should we add some exception handling here?

> 6:01:12 PM <@blackboxsw> cjp256: open question for you as to how Azure wants to handle potential absence of IMDS

Ideally, we would rely on the previously fetched metadata to ensure we don't trigger any unintended side-effects.  The networking config should remain consistent (e.g. previously written config remains and cloud-init bails, or the _network_config cached attr is used, or cached metadata is re-used to generate the network config).

Separately, I'm planning to PR a change to override get_instance_id() for Azure datasource, because it's not the first time a failure led to the "iid-datasource" path.  I'd rather we just avoid that possibility altogether.

blackboxsw:
> So IIUC we are returning False, which works for the normal case, but doesn't help the update_metadata_is_supported() case after it purged the cache?
> 
> Should raising sources.InvalidMetaDataException address the update_metadata_is_supported() case without additional changes? Should this cause cloud-init to fail, or should we add some exception handling here?



> 
> > 6:01:12 PM <@blackboxsw> cjp256: open question for you as to how Azure wants to handle potential absence of IMDS
> 
> Ideally, we would rely on the previously fetched metadata to ensure we don't trigger any unintended side-effects. The networking config should remain consistent (e.g. previously written config remains and cloud-init bails, or the _network_config cached attr is used, or cached metadata is re-used to generate the network config).

I agree here. we should be able to fallback to previously crawled metadata without regression if an error occurs when trying to refresh metadata due to `update_metadata_is_supported`.This does require significant changes to how we invoke `_get_data` broadly  so I'd like to avoid instrumenting such a change in the immediate SRU just due to potential risk. 
Below is a diff I think might give us an easy fallback in the event of unexpected failure[1]


> Separately, I'm planning to PR a change to override get_instance_id() for Azure datasource, because it's not the first time a failure led to the "iid-datasource" path. I'd rather we just avoid that possibility altogether.

Good on https://github.com/canonical/cloud-init/pull/1727 . looks like it merged yesterday.


references:
 - [1]  diff for unexpected update_is_suppported 
 ```diff
diff --git a/cloudinit/sources/__init__.py b/cloudinit/sources/__init__.py
index 3a483c26c..52122ba68 100644
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -339,12 +339,15 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
             }
         }
 
-    def clear_cached_attrs(self, attr_defaults=()):
+    def clear_cached_attrs(self, attr_defaults=()) -> dict:
         """Reset any cached metadata attributes to datasource defaults.
 
-        @param attr_defaults: Optional tuple of (attr, value) pairs to
+        :param attr_defaults: Optional tuple of (attr, value) pairs to
            set instead of cached_attr_defaults.
+
+        :return: dict of attributes cleared and cleared values
         """
+        cleared_attrs = {}
         if not self._dirty_cache:
             return
         if attr_defaults:
@@ -354,9 +357,13 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
         for attribute, value in attr_values:
             if hasattr(self, attribute):
-                setattr(self, attribute, value)
+                # Preserve previous value if cleared
+                if getattr(self, attribute) != value:
+                    cleared_attrs[attribute] = getattr(self, attribute)
+                    setattr(self, attribute, value)
         if not attr_defaults:
             self._dirty_cache = False
+        return cleared_attrs
 
     def get_data(self) -> bool:
         """Datasources implement _get_data to setup metadata and userdata_raw.
@@ -806,6 +813,7 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
         @return True if the datasource did successfully update cached metadata
             due to source_event_type.
         """
+        cleared_attrs = {}
         supported_events = self.get_supported_events(source_event_types)
         for scope, matched_events in supported_events.items():
             LOG.debug(
@@ -816,10 +824,35 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
             # Each datasource has a cached config property which needs clearing
             # Once cleared that config property will be regenerated from
             # current metadata.
-            self.clear_cached_attrs((("_%s_config" % scope, UNSET),))
+            cleared_attrs.update(
+                self.clear_cached_attrs(
+                    (("_%s_config" % scope, UNSET),)
+                )
+            )
         if supported_events:
-            self.clear_cached_attrs()
+            cleared_attrs.update(self.clear_cached_attrs())
             result = self.get_data()
+            if all([
+                EventType.BOOT_NEW_INSTANCE not in source_event_types,
+               not result
+            ]):
+                # Then we can no longer can crawl previous datasource.
+                #  One of the following may be True:
+                #  1. Upgrade of cloud-init introduced unexpected IMDS get_data
+                #     issues or exceptions resulting in get_data False
+                #  2. VM is not in a quarantine network without access to IMDS
+                #  3. Expected IMDS service is temporarily down during this
+                #     event
+                #  4. This VM has migrated to a new datasource environment
+                LOG.warning(
+                    "Update datasource metadata: get_data unexpectedly"
+                    " returned False for %s on events: %s. Preserving"
+                    " previous datasource metadata values.",
+                    self.dsname, source_vent_types
+                )
+                for attr, value in cleared_attrs.items():
+                    setattr(seld, attr, value)
+
             if result:
                 return True
         LOG.debug(
```

**Comment by blackboxsw** [comment]
@cjp256 per your question: `Should raising sources.InvalidMetaDataException address the update_metadata_is_supported() case without additional changes? Should this cause cloud-init to fail, or should we add some exception handling here?`


I also think that (raising an InvalidMetadataError on failure to crawl Azure's IMDS)[https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L732] instead of `return False` is the "correct" behavior for your datasource. On any non-azure platform, they would have failed the  [is_platform_viable(Path(self.seed_dir)) test](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L705)

Any other cases, we know we are on Azure and something "bad" happened so we want that error to get raised and visible (not return False there)

An  exception being raise or unexpected traceback will leave old datasource values intact and leave cloud-init in `error` state which I think is what we 'want' from any subsequent 'metadata updates' due to reboot event or hotplug event after datasource was previously detected. The exception in this secondary `_get_data` run ensures we don't try to react to changed metadata (or the cleared instance attributes on the Datasource in memory) as well as ensuring we don't try to persist that updated data to the obj pkl.

```bash
ubuntu@azure-0914-193330-vm:~$ cloud-id
azure
ubuntu@azure-0914-193330-vm:~$ cloud-init status --long
status: error
time: Wed, 14 Sep 2022 19:46:24 +0000
detail:
Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
ubuntu@azure-0914-193330-vm:~$ egrep "Trace|ERR|WARN|Warn" /var/log/cloud-init.log 
2022-09-14 19:46:12,625 - azure.py[ERROR]: Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
2022-09-14 19:46:12,657 - azure.py[WARNING]: Reported failure to Azure fabric.
2022-09-14 19:46:12,676 - util.py[WARNING]: failed stage init-local
2022-09-14 19:46:15,104 - azure.py[ERROR]: Could not crawl Azure metadata: 'DataSourceAzure' object has no attribute '_reported_ready_marker_file'
2022-09-14 19:46:15,132 - azure.py[WARNING]: Reported failure to Azure fabric.
2022-09-14 19:46:15,141 - util.py[WARNING]: failed stage init

# obj.pkl attributes are still unchanged
ubuntu@azure-0914-193330-vm:~$ sudo python3 -c 'from cloudinit.sources import pkl_load; print(pkl_load("/var/lib/cloud/instance/obj.pkl").metadata)'
{'azure_data': {'configurationsettype': 'LinuxProvisioningConfiguration'}, 'local-hostname': 'azure-0914-193330-vm', 'imds': {'compute': {'azEnvironment': 'AzurePublicCloud', 'customData': '', 'evictionPolicy': '', 'extendedLocation': {'name': '', 'type': ''}, 'isHostCompatibilityLayerVm': 'false', 'licenseType': '', 'location': 'centralus', 'name': 'azure-0914-193330-vm', 'offer': '0001-com-ubuntu-server-jammy-daily', 'osProfile': {'adminUsername': 'ubuntu', 'computerName': 'azure-0914-193330-vm', 'disablePasswordAuthentication': 'true'}, 'osType': 'Linux', 'placementGroupId': '', 'plan': {'name': '', 'product': '', 'publisher': ''}, 'platformFaultDomain': '0', 'platformUpdateDomain': '0', 'priority': '', 'provider': 'Microsoft.Compute', 'publicKeys': [{'keyData': 'ssh-rsa <REDACTED>
....
```

PR #1726, opened 2022-09-13
cc_ubuntu_advantage: Redact token

aciba90:
## Proposed Commit Message

```
cc_ubuntu_advantage: Redact token
```

## Additional Context

https://github.com/canonical/cloud-init/pull/1583#issuecomment-1245294551

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -100,6 +100,7 @@ meta: MetaSchema = {
 __doc__ = get_meta_doc(meta)
 
 LOG = logging.getLogger(__name__)
+REDACTED = "REDACTED"
 
 
 def supplemental_schema_validation(ua_config):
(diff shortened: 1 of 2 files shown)


holmanb (approved):
I don't see any other uses of the token in this module and the tests look good to me.

I assume that this work is being done as a courtesy to users and not because the key could be used to subvert the security of the operating system, correct? I reviewed this PR assuming that this isn't a CVE[1], but just want to be extra sure.

[1] based on a read of some UA docs (didn't find anything security relevant, but I may have missed something) and the fact that this isn't following the security reporting process

aciba90:
> I assume that this work is being done as a courtesy to users and not because the key could be used to subvert the security of the operating system, correct? I reviewed this PR assuming that this isn't a CVE[1], but just want to be extra sure.

Yes, I think so. Plus, additionally, this make sure no token is leaked through the logs. There is not a CVE going on.

PR #1727, opened 2022-09-13
sources/azure: ensure instance id is always correct

cjp256:
Currently, get_instance_id() assumes that the instance ID is in the metadata.  If not found, it falls back to a hardcoded string "iid-datasource".

Override this behavior to query the instance id as needed.


anhvoms (approved):

holmanb (approved):
Looks good, thanks.

PR #1729, opened 2022-09-14
refactor: unify aliased OSError exceptions

aciba90:
## Proposed Commit Message

```
refactor: unify aliased OSError exceptions

Since python3.3 the following exceptions are just alias of OSError and some of them are deprecated:

- EnvironmentError
- IOError
- WindowsError
- socket.error
- select.error
- mmap.error

This commit substitute appearances of them with OSError.
```

## Additional Context

https://docs.python.org/3/library/exceptions.html#OSError

## Checklist:

### DIFF
--- a/cloudinit/cmd/cloud_id.py
+++ b/cloudinit/cmd/cloud_id.py
@@ -76,7 +76,7 @@ def handle_args(name, args):
     try:
         with open(args.instance_data) as file:
             instance_data = json.load(file)
-    except IOError:
+    except OSError:
         return error(
             "File not found '%s'. Provide a path to instance data json file"
             " using --instance-data" % args.instance_data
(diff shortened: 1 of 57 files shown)


TheRealFalcon:
I'm slightly -1 on this change. I think currently seeing `IOError` is a hint or smell that we should change the exception to something more specific than `OSError`. Prior to PEP 3151, attempting to open a file that didn't exist would throw an `IOError`. After, attempting to open a file that doesn't exist will throw `FileNotFoundError` (which is a subclass of `OSError`). Catching `OSError` is more broad of an exception than we should be catching. I know that's technically what is already happening now, but that's due to backwards compatibility in the language. We should use `IOError` as an indication to update the exception accordingly rather than shift to catching OSError everywhere. If we were up against a deprecation deadline I'd feel differently, but for now I'd rather keep it as is.

Do you agree? If not, we can get opinions from others on the team too.

aciba90:
That makes sense, @TheRealFalcon. I was mainly concerned about clarity in the code, but taking into account we could potentially lose information, I close this ticket.

PR #1736, opened 2022-09-16
cc_ansible: add support for galaxy install

holmanb:
```
modules: add ansible-galaxy support to ansible

This is intended to allow users to install roles
and collections directly with the user-data
cloud-config for use in ansible playbooks during boot.
```

## Additional Context
This allows community reuse of roles and collections, which users
may use in their own ansible playbooks.

There is [more work](https://docs.google.com/document/d/1hHEijcgxZNbWZvbVlL5YkDCDBv52Sb49nEJ7FDU5LcI/edit#) to do with this module, but since this is usable with ansible-pull as well, I'm submitting it as is to keep reviews smaller.

https://galaxy.ansible.com/
https://docs.ansible.com/ansible/latest/cli/ansible-galaxy.html

## Test Steps
integration test included to install from galaxy

### DIFF
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -90,7 +90,7 @@ class AnsiblePull(abc.ABC):
             raise ValueError("command: ansible is not installed")
 
     def subp(self, command, **kwargs):
-        return subp(command, **kwargs)
+        return subp(command, env=self.env, **kwargs)
 
     @abc.abstractmethod
     def is_installed(self):
@@ -160,6 +160,10 @@ def handle(
             ansible = AnsiblePullDistro(distro)
         ansible.install(package_name)
         ansible.check_deps()
+        ansible_config = ansible_cfg.get("ansible_config", "")
+
+        if ansible_config:
+            ansible.env[CFG_OVERRIDE] = ansible_config
 
         if galaxy_cfg:
             ansible_galaxy(galaxy_cfg, ansible)
@@ -215,9 +219,6 @@ def run_ansible_pull(pull: AnsiblePull, cfg: dict):
 
 def ansible_galaxy(cfg: dict, ansible: AnsiblePull):
     actions = cfg.get("actions", [])
-    ansible_config = cfg.get(CFG_OVERRIDE, "")
-    if ansible_config:
-        ansible.env[CFG_OVERRIDE] = ansible_config
 
     if not actions:
         LOG.warning("Invalid config: %s", cfg)
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -1,6 +1,6 @@
 """ansible enables running on first boot either ansible-pull"""
 import abc
-import logging
+from logging import Logger, getLogger
 import os
 import re
 import sys
@@ -61,7 +61,7 @@ meta: MetaSchema = {
 }
 
 __doc__ = get_meta_doc(meta)
-LOG = logging.getLogger(__name__)
+LOG = getLogger(__name__)
 PIP_PKG = "python3-pip"
 CFG_OVERRIDE = "ANSIBLE_CONFIG"
 
@@ -139,7 +139,9 @@ class AnsiblePullDistro(AnsiblePull):
         return bool(which("ansible"))
 
 
-def handle(name: str, cfg: Config, cloud: Cloud, _, __) -> None:
+def handle(
+    name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
+) -> None:
 
     ansible_cfg: dict = cfg.get("ansible", {})
     install_method = ansible_cfg.get("install-method")
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -119,13 +119,11 @@ class AnsiblePullPip(AnsiblePull):
         if not self.is_installed():
             # bootstrap pip if required
             try:
-                import pip
+                import pip  # noqa: F401
             except ImportError:
                 self.distro.install_packages(self.distro.pip_package_name)
 
-            self.subp(
-                [sys.executable, "-m", "pip", "install", "--user", pkg_name]
-            )
+            self.subp([sys.executable, "-m", "pip", "install", pkg_name])
 
     def is_installed(self) -> bool:
         stdout, _ = self.subp([sys.executable, "-m", "pip", "list"])
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -1,10 +1,10 @@
 """ansible enables running on first boot either ansible-pull"""
 import abc
-from logging import Logger, getLogger
 import os
 import re
 import sys
 from copy import deepcopy
+from logging import Logger, getLogger
 from textwrap import dedent
 from typing import Optional
 
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -62,7 +62,6 @@ meta: MetaSchema = {
 
 __doc__ = get_meta_doc(meta)
 LOG = getLogger(__name__)
-PIP_PKG = "python3-pip"
 CFG_OVERRIDE = "ANSIBLE_CONFIG"
 
 
@@ -119,7 +118,7 @@ class AnsiblePullPip(AnsiblePull):
         if not self.is_installed():
             # bootstrap pip if required
             if not which("pip3"):
-                self.distro.install_packages(PIP_PKG)
+                self.distro.install_packages(self.distro.pip_package_name)
             self.subp(["python3", "-m", "pip", "install", "--user", pkg_name])
 
     def is_installed(self) -> bool:
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -67,9 +67,11 @@ CFG_OVERRIDE = "ANSIBLE_CONFIG"
 
 
 class AnsiblePull(abc.ABC):
-    cmd_version: list = []
-    cmd_pull: list = []
-    env: dict = {}
+    def __init__(self, distro: Distro):
+        self.cmd_pull = ["ansible-pull"]
+        self.cmd_version = ["ansible-pull", "--version"]
+        self.distro = distro
+        self.env = os.environ.copy()
 
     def get_version(self) -> Optional[Version]:
         stdout, _ = self.subp(self.cmd_version)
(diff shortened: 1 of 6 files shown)


cloudinit/config/cc_ansible.py
```diff
@@ -112,18 +121,22 @@ def install(self, pkg_name: str):
...
+        stdout, _ = self.subp(["python3", "-m", "pip", "list"])
         return "ansible" in stdout
 
+    def subp(self, command, **kwargs):
+        return subp(args=command, env=self.env, **kwargs)
+
 
 class AnsiblePullDistro(AnsiblePull):
     def __init__(self, distro: Distro):
         self.cmd_pull = ["ansible-pull"]
```
aciba90, line 136: Nit: `AnsiblePullDistro` and `AnsiblePullPip`'s `__init__` methods have duplicated set-up, what do you think about abstracting it in the base-class constructor? In that way, it would be more clear what each class does different:

```diff
diff --git a/cloudinit/config/cc_ansible.py b/cloudinit/config/cc_ansible.py
index 0687ec624..d5282bf56 100644
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -67,9 +67,11 @@ CFG_OVERRIDE = "ANSIBLE_CONFIG"
 
 
 class AnsiblePull(abc.ABC):
-    cmd_version: list = []
-    cmd_pull: list = []
-    env: dict = {}
+    def __init__(self, distro: Distro):
+        self.cmd_pull = ["ansible-pull"]
+        self.cmd_version = ["ansible-pull", "--version"]
+        self.distro = distro
+        self.env = os.environ.copy()
 
     def get_version(self) -> Optional[Version]:
         stdout, _ = self.subp(self.cmd_version)
@@ -102,11 +104,8 @@ class AnsiblePull(abc.ABC):
 
 class AnsiblePullPip(AnsiblePull):
     def __init__(self, distro: Distro):
-        self.cmd_pull = ["ansible-pull"]
-        self.cmd_version = ["ansible-pull", "--version"]
-        self.distro = distro
+        super().__init__(distro)
 
-        self.env = os.environ.copy()
         old_path = self.env.get("PATH")
         if old_path:
             self.env["PATH"] = ":".join([old_path, "/root/.local/bin/"])
@@ -132,12 +131,6 @@ class AnsiblePullPip(AnsiblePull):
 
 
 class AnsiblePullDistro(AnsiblePull):
-    def __init__(self, distro: Distro):
-        self.cmd_pull = ["ansible-pull"]
-        self.cmd_version = ["ansible-pull", "--version"]
-        self.distro = distro
-        self.env = os.environ.copy()
-
     def install(self, pkg_name: str):
         if not self.is_installed():
             self.distro.install_packages(pkg_name)
```

cloudinit/config/cc_ansible.py
```diff
@@ -196,3 +216,15 @@ def run_ansible_pull(pull: AnsiblePull, cfg: dict):
...
+def ansible_galaxy(cfg: dict, ansible: AnsiblePull):
+    actions = cfg.get("actions", [])
+    ansible_config = cfg.get(CFG_OVERRIDE, "")
+    if ansible_config:
+        ansible.env[CFG_OVERRIDE] = ansible_config
+
+    if not actions:
+        LOG.warning("Invalid config: %s", cfg)
+    for command in actions:
+        ansible.subp(command)
```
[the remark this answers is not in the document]
holmanb: > Otherwise, a user could insert any arbitrary command here

In general, I would argue that if config writers (who have superuser on their machine) want to `rm -rf /etc`, I don't see a reason to stop them. They can already do that in `bootcmd` / `runcmd` if they would like to.

That said, this is a specialized command where running commands besides `ansible-galaxy` doesn't make sense. If `ansible-galaxy` had names on other distros, or different names based which install type it had, I'd say that would be a fairly strong reason to leave the flexibility as is.

I'm fairly indifferent, I can make the suggested change.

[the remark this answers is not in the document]
aciba90: It was an open question, maybe exacerbated by the leaking to the logs' discussion, ignore my comment please.

cloudinit/config/cc_ansible.py
```diff
@@ -133,23 +146,30 @@ def is_installed(self) -> bool:
         return bool(which("ansible"))
 
 
-def handle(
-    name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
-) -> None:
+def handle(name: str, cfg: Config, cloud: Cloud, _, __) -> None:
```
TheRealFalcon, line 149: For consistency, let's not change this function signature.

holmanb: I typically avoid churn like this, but in this case I was trying to dispose of the unused `Logger` annotation. Either way, fine by me. +1

cloudinit/config/cc_ansible.py
```diff
@@ -64,15 +63,16 @@
 __doc__ = get_meta_doc(meta)
 LOG = logging.getLogger(__name__)
 PIP_PKG = "python3-pip"
```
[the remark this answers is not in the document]
holmanb: How do you feel about a89034c6a35ece91530b143f9c6eacf8d8e5ec23 as a solution to this?

cloudinit/config/cc_ansible.py
```diff
@@ -101,9 +104,15 @@ class AnsiblePullPip(AnsiblePull):
     def __init__(self, distro: Distro):
         self.cmd_pull = ["ansible-pull"]
         self.cmd_version = ["ansible-pull", "--version"]
-        self.env["PATH"] = ":".join([self.env["PATH"], "/root/.local/bin/"])
         self.distro = distro
 
+        self.env = os.environ.copy()
+        old_path = self.env.get("PATH")
+        if old_path:
```
TheRealFalcon, line 109: Is this right? If a PATH existed previous, add `/root/.local/bin` to it, but if not, set the PATH to empty? Why don't we want `/root/.local/bin/` in the else case?

TheRealFalcon (commented):
Just a few questions inline

cloudinit/config/cc_ansible.py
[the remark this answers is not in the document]
aciba90: I think it's fine.

aciba90 (approved):
LGTM aside James' comments.

holmanb:
@TheRealFalcon @aciba90 Thanks for the reviews. I think I've addressed all comments.

cloudinit/config/cc_ansible.py
```diff
@@ -62,17 +61,19 @@
 }
 
 __doc__ = get_meta_doc(meta)
-LOG = logging.getLogger(__name__)
-PIP_PKG = "python3-pip"
+LOG = getLogger(__name__)
+CFG_OVERRIDE = "ANSIBLE_CONFIG"
```
TheRealFalcon, line 65: This needs to be lowercase now.

cloudinit/config/cc_ansible.py
```diff
@@ -196,3 +211,15 @@ def run_ansible_pull(pull: AnsiblePull, cfg: dict):
     )
     if stdout:
         sys.stdout.write(f"{stdout}")
+
+
+def ansible_galaxy(cfg: dict, ansible: AnsiblePull):
+    actions = cfg.get("actions", [])
+    ansible_config = cfg.get(CFG_OVERRIDE, "")
```
TheRealFalcon, line 218: cfg here refers to galaxy, right? Galaxy doesn't have an 'ansible_config' key under it, so I don't think this will work.

TheRealFalcon (changes requested):
I think `CFG_OVERRIDE` could use some testing

holmanb:
> I think `CFG_OVERRIDE` could use some testing

@TheRealFalcon agreed, thanks. I added a test and rectified the behavior.

cloudinit/config/cc_ansible.py
TheRealFalcon: Correct?

cloudinit/config/cc_ansible.py
```diff
@@ -111,20 +119,18 @@ def install(self, pkg_name: str):
         if not self.is_installed():
             # bootstrap pip if required
             if not which("pip3"):
-                self.distro.install_packages(PIP_PKG)
-            subp(["python3", "-m", "pip", "install", "--user", pkg_name])
+                self.distro.install_packages(self.distro.pip_package_name)
+            self.subp(["python3", "-m", "pip", "install", "--user", pkg_name])
```
igalic, line 123: pip3 and python3 might not be available on every platform.
can't we use the python we are executed under as starting off point?

igalic (commented):
❓

cloudinit/config/cc_ansible.py
holmanb: Thanks @igalic for the feedback. How about I replace `python3` with `sys.executable`?

igalic: 👍

holmanb: @igalic: Regarding pip3, I'm not certain what the best approach would be. Maybe something like this?

```python
-            if not which("pip3"):
+            _, stderr = self.subp(
+                [sys.executable, "-m", "pip"], rcs=[0, 1]
+            )
+            if stderr.strip():
```

igalic: we could do a scoped import, and catch the error, then, based on that install

cloudinit/config/cc_ansible.py
```diff
@@ -99,32 +103,39 @@ def install(self, pkg_name: str):
...
-            if not which("pip3"):
-                self.distro.install_packages(PIP_PKG)
-            subp(["python3", "-m", "pip", "install", "--user", pkg_name])
+            try:
+                import pip
+            except ImportError:
+                self.distro.install_packages(self.distro.pip_package_name)
+
+            self.subp(
+                [sys.executable, "-m", "pip", "install", "--user", pkg_name]
```
igalic, line 127: cloud-init is running as root.
how sensible is `--user` here?

igalic (commented):
❓

cloudinit/config/cc_ansible.py
holmanb: Hey @igalic. Interesting question. It actually raises a couple of other questions that I'll get to in a moment.

To directly answer the question:
```
  --user                      Install to the Python user install directory for
                              your platform. Typically ~/.local/, or
                              %APPDATA%\Python on Windows. (See the Python
                              documentation for site.USER_BASE for full
                              details.)
```

I don't think it necessarily hurts to install in `/root/.local`, but also doesn't really add anything of value either.

I'm fine with dropping it when installing as root, however as you said:

> cloud-init is running as root.

And installing with pip as an elevated user is generally not considered best practice. This begets change to support installing as non-root.

To enable users to install and run ansible without running pip as root (and also to run ansible as an unprivileged user if they wish), I'd like to add a couple of keys to this module: `install-user`, and `run-user`. This would allow cloud-init to drop permissions and run as a user already configured via cloud-init's `cc_users_groups` module.

When running as an unprivileged user, I think that the `--user` flag makes more sense and should be used in this case.

#### To Summarize

I'll drop the `--user` flag here. In a follow-up PR, I'll add some new keys that allow installing and running ansible as alternate users to support best practices.

@igalic - Thoughts?

TheRealFalcon (approved):
LGTM!

PR #1737, opened 2022-09-17
Feature: create machine-id

chifac08:
## Proposed Commit Message
```
cc_create_machine_id:
This module is responsible for creating a new machine-id during cloud-init's init stage.

How this module works:
  1. Remove the following files and links:
      - /etc/machine-id
      - /var/lib/dbus/machine-id
  2. Create new machine-id with "systemd-machine-id-setup"

Attention, this module will only work with systemd installed.
```

## Additional Context
Implements a feature from wishlist.

LP: [#1563951](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1563951)

## Test Steps
integration test included
Only tested for Ubuntu.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_create_machine_id.py
+++ b/cloudinit/config/cc_create_machine_id.py
@@ -1,7 +1,7 @@
 # Author: Fabian Lichtenegger-Lukas <fabian.lichtenegger-lukas@nts.eu>
 # This file is part of cloud-init. See LICENSE file for license information.
 
-"""create-machine-id"""
+"""create_machine_id"""
 from logging import Logger
 
 from cloudinit import log as logging
@@ -19,14 +19,14 @@ This will likely take multiple lines.
 
 meta: MetaSchema = {
     "id": "cc_create_machine_id",
-    "name": "create-machine-id",
+    "name": "create_machine_id",
     "title": "Re/creates machine-id",
     "description": MODULE_DESCRIPTION,
     "distros": ["ubuntu"],
     "frequency": PER_ONCE,
-    "activate_by_schema_keys": ["create-machine-id"],
+    "activate_by_schema_keys": ["create_machine_id"],
     "examples": [
-        "create-machine-id: true",
+        "create_machine_id: true",
     ],
 }
 
@@ -50,13 +50,13 @@ def supplemental_schema_validation(mid: dict):
     """
     errors = []
     for key, value in sorted(mid.items()):
-        if key == "create-machine-id":
+        if key == "create_machine_id":
             if not isinstance(value, bool):
                 errors.append(f"Expected a bool for {key}. Found {value}")
 
     if errors:
         raise ValueError(
-            f"Invalid 'create-machine-id' configuration:{NL}{NL.join(errors)}"
+            f"Invalid 'create_machine_id' configuration:{NL}{NL.join(errors)}"
         )
 
 
@@ -69,6 +69,7 @@ def remove_machine_id(delFiles: frozenset):
     """
     try:
         for file in delFiles:
+            LOG.info("Removing file %s", file)
             util.del_file(file)
     except OSError as e:
         raise RuntimeError(f"Failed to remove file '{file}'") from e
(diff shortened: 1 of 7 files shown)


mwhudson:
I'm not a cloud-init developer, but I don't see how generating a machine-id at the point this code will run does anything useful. Won't systemd have generated an id for itself long before this? If you want to change the id, don't you have to do more than this? I might be completely mistaken of course.

aciba90:
I agree with @mwhudson.

I don't see the point of letting the users control whether the machine-id should be created or not, as an instance must always have a machine-id. Furthermore, if we would assume it, it could not be done in a config module, as config module execution happens very late in the boot process. Init systems would have already a machine-id determined and configured for the system, and:

1. bad things can happen if we change it in a running system without a reboot,
2. clouds could potentially identify the same machine twice.

For golden image creation, providers should leave /etc/machine-id empty or not created to let the init systems create a unique one. We also offer the opportunity to use cloud-init to create golden images as `cloud-init clean --machine-id` removes /etc/machine-id.

More info:

https://www.freedesktop.org/software/systemd/man/machine-id.html#
https://www.freedesktop.org/software/systemd/man/systemd-machine-id-setup.html
https://www.freedesktop.org/software/systemd/man/systemd-firstboot.html#
https://www.freedesktop.org/software/systemd/man/systemd-machine-id-commit.service.html#
https://cloudinit.readthedocs.io/en/latest/topics/boot.html

I think cloud-init's unique responsibility in respect to this matter is to ensure that an instance has a machine-id.

In systemd systems, this is automatically handled by systemd itself [1]: 

> When a machine is booted with systemd(1) the ID of the machine will be established. If systemd.machine_id= or --machine-id= options (see first section) are specified, this value will be used. Otherwise, the value in /etc/machine-id will be used. If this file is empty or missing, systemd will attempt to use the D-Bus machine ID from /var/lib/dbus/machine-id, the value of the kernel command line option container_uuid, the KVM DMI product_uuid or the devicetree vm,uuid (on KVM systems), and finally a randomly generated UUID.

In respect to other init systems, I didn't find documentation about how the machine-id is generated. But I would assume they do the right thing.

[1] https://www.freedesktop.org/software/systemd/man/machine-id.html#Initialization

chifac08:
My intention behind this module is to give users the opportunity to create a new machine-id for images with an already configured machine-id. I'm totally aware that an golden image should not have an already configured machine-id and that cloud-init creates the machine-id, via that module, too late (after systemd-firstboot.service) but i thought it would be nice if you could reset machine-id via a simple command in cloud-init.  All in all, this module could reset machine-id of "dirty" images.

I hope you could understand my thoughts and motivations to write this module. I wasn't aware that a recreation of machine-id would need an reboot and i didn't find a hint in documentation. That implies cloud-init have to do a reboot like "Package Update Upgrade Install" modules does. 

Is there any chance this module could make it into upstream or should i close this PR?

Nevertheless, thank you for reviewing.

aciba90:
Thanks, @chifac08, for the good intentions and motivations.

I still think that `machine-id` handling belongs to cloud vendors, not to consumers of cloud images.

Changing the machine-id and rebooting at that point in boot process could potentially cause side effects that cloud-init should fix and maintain along the time across at least clouds and maybe init-systems and distros, therefore I think in this case it does not pay off the pros vs the cons.

In the case that an advanced user would really want to do that, they could do it via the following user-data:

```yaml
runcmd:
- [ "cloud-init", "clean", "--logs", "--machine-id"]
```

or any other custom command.

Thanks again for your work here.

chifac08:
> Thanks, @chifac08, for the good intentions and motivations.
> 
> I still think that `machine-id` handling belongs to cloud vendors, not to consumers of cloud images.
> 
> Changing the machine-id and rebooting at that point in boot process could potentially cause side effects that cloud-init should fix and maintain along the time across at least clouds and maybe init-systems and distros, therefore I think in this case it does not pay off the pros vs the cons.
> 
> In the case that an advanced user would really want to do that, they could do it via the following user-data:
> 
> ```yaml
> runcmd:
> - [ "cloud-init", "clean", "--logs", "--machine-id"]
> ```
> 
> or any other custom command.
> 
> Thanks again for your work here.

I agree and close this PR. Thanks for your time and consideration.

PR #1739, opened 2022-09-19
cc_ubuntu_advantage: add ua_config in auto-attach

aciba90:
## Proposed Commit Message

```
cc_ubuntu_advantage: add ua_config in auto-attach

- Fix and document how to unset ua_config props.
- Redact ua_config values as they could contain sensitive auth info.
- Increase unit test coverage.
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-1257
https://github.com/canonical/cloud-init/pull/1583#pullrequestreview-1110047636

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -5,6 +5,7 @@
 import re
 from logging import Logger
 from textwrap import dedent
+from typing import Any
 from urllib.parse import urlparse
 
 from cloudinit import log as logging
@@ -29,8 +30,8 @@ meta: MetaSchema = {
         FIPS and FIPS Updates. When attaching a machine to Ubuntu Advantage,
         one can also specify services to enable. When the 'enable'
         list is present, only named services will be activated. Whereas
-        'enable' list is not present, any named service will supplement
-        contract-default enabled services.
+        'enable' list is not present, the contract-default services will be
+        enabled.
 
         On Pro instances, when ``ubuntu_advantage`` config is provided to
         cloud-init, Pro's auto-attach feature will be disabled and cloud-init
@@ -90,8 +91,8 @@ meta: MetaSchema = {
           config:
             http_proxy: 'http://some-proxy:8088'
             https_proxy: 'https://some-proxy:8088'
-            global_apt_https_proxy: 'http://some-global-apt-proxy:8088/'
-            global_apt_http_proxy: 'https://some-global-apt-proxy:8088/'
+            global_apt_https_proxy: 'https://some-global-apt-proxy:8088/'
+            global_apt_http_proxy: 'http://some-global-apt-proxy:8088/'
             ua_apt_http_proxy: 'http://10.0.10.10:3128'
             ua_apt_https_proxy: 'https://10.0.10.10:3128'
           enable:
@@ -137,6 +138,14 @@ ERROR_MSG_SHOULD_AUTO_ATTACH = (
     "Unable to determine if this is an Ubuntu Pro instance."
     " Fallback to normal UA attach."
 )
+UA_CONFIG_PROPS = (
+    "http_proxy",
+    "https_proxy",
+    "global_apt_http_proxy",
+    "global_apt_https_proxy",
+    "ua_apt_http_proxy",
+    "ua_apt_https_proxy",
+)
 
 
 def validate_schema_features(ua_section: dict):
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -29,8 +30,8 @@
         FIPS and FIPS Updates. When attaching a machine to Ubuntu Advantage,
         one can also specify services to enable. When the 'enable'
         list is present, only named services will be activated. Whereas
-        'enable' list is not present, any named service will supplement
-        contract-default enabled services.
+        'enable' list is not present, the contract-default services will be
+        enabled.
```
blackboxsw, line 35:
```suggestion
        if the 'enable' list is not present, the contract's default
        services will be enabled.

```

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -172,102 +181,103 @@ def supplemental_schema_validation(ua_config: dict):
...
-                    errors.append(
-                        f"Expected URL scheme http/https for ua:config:{key}."
-                        f" Found: {value}"
-                    )
-            except (AttributeError, ValueError):
+        if key not in UA_CONFIG_PROPS:
+            LOG.warning(
+                "Ignoring unknown ubuntu_advantage.config.%s property", key
+            )
+            continue
```
blackboxsw, line 197: While this is a nice sentiment, this now tightly couples `ubuntu_advantage: config` to whatever cloud-init releases. UA currently has more config options beyond the *proxy variables, such as update_status_timer, update_messaging_timer and metering_timer. I think we still want to allow setting those values, but maybe we just log  a warning and say, "Attempting to set unknown ua config option X".  If we are too strict here, customers which may want to immediately test ua config options that cloud-init doesn't know about will get blocked and have to wait for a cloud-init SRU to support such config settings.   And our worst-case scenario is that we can provide warnings for keys which may not be supported by the underlying ubuntu-advantage-tools version, and the ua config set <invalid_name> <value> will result in a traceback, which we want to fail cloud-init in error state if unsupported.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -172,102 +181,103 @@ def supplemental_schema_validation(ua_config: dict):
...
-        config = dict()
+    supplemental_schema_validation(ua_config)
 
     enable_errors = []
-
-    # UA Config
-    for key, value in sorted(config.items()):
+    for key, value in sorted(ua_config.items()):
+        if key not in UA_CONFIG_PROPS:
+            continue  # Already logged in `supplemental_schema_validation`
```
blackboxsw, line 231: I don't think we want to be so strict here (per my point above) there are some config settings already supported by UA that we might want to allow. This initial validation we were doing was to perform additional level of validation of known URL-type *proxy ua config values. If we want we can add the *_timer config options to cloud-init's schema, and assert that they are integers. But I still think we want to add an `additionalProperties: true`  under `ubuntu_advantage: config` in our schema-cloud-config-v1.json.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2229,37 +2229,38 @@
             },
             "config": {
               "type": "object",
-              "description": "Configuration settings or override Ubuntu Advantage config. Only used in non Ubuntu Pro instances.",
+              "description": "Configuration settings or override Ubuntu Advantage config.",
+              "additionalProperties": false,
```
blackboxsw, line 2233: I actually think we want this to be true to allow for flexibility since UA currently has more config options than cloud-inti schema reports (since we only wanted to validate URL *proxy type keys). We also don't want to disallow valid confg keys as UA grows more because that would require cloud-init to continue to SRU to enable any new config keys that are supported.

blackboxsw (changes requested):
Thanks @aciba90 . I hope we can allow cloud-init to be more premissive of additionalProperties, but I otherwise think the additional explicit logging about unknown config keys and handling is an improvement.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: done

cloudinit/config/cc_ubuntu_advantage.py
aciba90: done

cloudinit/config/schemas/schema-cloud-config-v1.json
aciba90: done

aciba90:
> Thanks @aciba90 . I hope we can allow cloud-init to be more premissive of additionalProperties, but I otherwise think the additional explicit logging about unknown config keys and handling is an improvement.

Make sense. I have addressed the comments. Thanks, @blackboxsw, for the review.

blackboxsw (approved):
LGTM. The logging and handling is indeed much cleaner. Thanks for additionalProperties:true to allow us that flexibility to decouple UA releases from cloud-init SRUs.

PR #1741, opened 2022-09-19
22.3.3 Release Into Kinetic

holmanb:
## Additional Context

Looks like the dpkg version is good:
```
arc~ dpkg --compare-versions 22.3.3-0ubuntu1~22.10.1 gt 22.3-13-g70ce6442-0ubuntu1~22.10.1 | echo $?
0
```

### Notes

Starting the changelog with the first entry being `Release 22.3.1` is a little weird, but this is because we released directly from a commit hash in 22.3.1 and added a release tag (with changelog & version bump) after releasing into Kinetic. 

- I uploaded a [new tarball here](https://launchpad.net/cloud-init/trunk/22.3.3)
- Generating the release notes for the Launchpad release was done [using my unmerged branch](https://github.com/canonical/uss-tableflip/pull/99). This is required because `upstream-release` assumes that release happens from tip of main, and that is not the case now. This release is from 22.3.3.


debian/changelog
```diff
@@ -1,3 +1,17 @@
+cloud-init (22.3.3-0ubuntu1~22.10.1) kinetic; urgency=medium
+
+  * New upstream release.
```
blackboxsw, line 3:
```suggestion
  * New upstream bug fix release.
```

debian/changelog
```diff
@@ -1,3 +1,17 @@
...
+    + Release 22.3.3 (LP: #1986703)
+    + Fix Oracle DS not setting subnet when using IMDS (#1735)
+      (LP: #1989686)
+    + Release 22.3.2 (LP: #1986703)
```
blackboxsw, line 7:
```suggestion
```

debian/changelog
```diff
@@ -1,3 +1,17 @@
...
+    + azure: define new attribute for pre-22.3 pickles (#1725)
+    + sources/azure: ensure instance id is always correct (#1727)
+      [Chris Patterson]
+    + Release 22.3.1 (LP: #1986703)
```
blackboxsw, line 11:
```suggestion
```

debian/changelog
```diff
@@ -1,3 +1,17 @@
+cloud-init (22.3.3-0ubuntu1~22.10.1) kinetic; urgency=medium
```
blackboxsw, line 1: I think we have two choices here to maintain upgrade path on kinetic to next new-upstream-snapshot:

  1. manually set this packaged release version `22.3-19....` so that any subsequent `new-upstream-snapshots` before we publish `22.4`  published to kinetic are > `22.3.3`
  2. add a commit to main to set version.py to 22.3.3 so any subsequent new-upstream-snapshot into kinetic will be `22.3.3-X...`

If we don't do either, our next new-upstream-snapshot into kinetic will be:
`22.3-46-g9aa4ec69-0ubuntu1~22.10.1`  which will be less than a published `22.3.3-0ubuntu1~22.10.1`

This version comparison will break deb package upgrade path logic and not allow us to upload such a version to PPAs or -updates/-proposed pockets.

blackboxsw (changes requested):
This looks good once we redact the d/changelog content a bit. And we need to define the typical process path we want to follow in upstream for subrevisions such as 22.3.3. Do we want to represent the subrevision it in `upstream/main:cloudinit/version.py` or supplement `new-upstream-snapshot` tooling to improve become aware of annotated major.minor.subrevisions when creating the debian/changelog version.

debian/changelog
holmanb: I lean away from manually setting versions if there are alternative options, though I'm not sure how the other option would work.

holmanb: If we desire to be able to snapshot main into ubuntu/devel but then update to, for example, a package built from 22.3.4, then I think that the easiest way might be to bump `ubuntu1` to `ubuntu2` for the snapshots from main. That will ensure that we can always "update" back to the next release version. It is worth noting that this version scheme is a little misleading, however, because doing a snapshot from main and leaving the version starting with `22.3.3` implies that the package is based on `22.3.3` when it is not. Maybe we don't care about that for devel?

On the contrary, if we don't believe that we'll ever want to do a snapshot from main followed by a version bump back down to a minor release (22.3.4), then just pre-releasing the devel package under the next release name `22.4-<commit number>` makes sense I think, since main is the branch that the next release will be based on, and number of commits will always increase so I don't think there should ever be an upgrade issue.

I'm fine to just manually bump for now if that's what you'd prefer, but I'd like to think that with this versioning issue that we can remove the human from the loop.

Thoughts?

holmanb: My above comment ignores how our current tooling works and just tries to make sense of upgrade paths and versions for ubuntu/devel.

blackboxsw: Ok we decided as a squad in standup we would like to preserve upstream release prefixes in our deb packaging releases in Ubuntu so the version prefix will be 22.3.3-0ubuntu1~22.10.1.   Any interim releases which land in future devel series will likely have package major.minor version prefix related to the next release under development, but a deminished version string including `~pre.#-` to ensure the interim release version `22.4~pre.x-0ubuntu1~23.04.1` sorts lower than the official upstream release version of `22.4-ubuntu1~23.04.1`

blackboxsw (approved):
- Changeset deltas are fine vs local run, only authorship diffs in d/changelog
- build-package. sbuild and upload to PPA work without unexpected errors
- CI runs fine and not specific quilt patches to review here

I changed the merge target to ubuntu/kinetic-22.3-hotfix for two reasons:
- we don't want to merge changes to `cloudinit/version.py` into the clean `ubuntu/devel` branch as we'll get merge conflicts on both daily recipe and/or next `new-upstream-snapshot` the moment we add a commit to main moving cloudinit/version.py to `VERSION = 22.4` `CONFLICT (content): Merge conflict in cloudinit/version.py`

If we think this merge conflict is reasonable and that we will new-upstream-snapshot and manually resolve conflicts in each ubuntu/bionic|focal|jammy|kinetic release branch once main bumps to version.py to `VERSION = 22.4`. Then we can also merge this content into `upstream/ubuntu/devel` tomorrow. Currently I've treated this PR as a hotfix type release and as such. We'll have a followup changelog documentation PR into `upstream/ubuntu/devel` to ensure our documented `debian/changelog` published files are preserved for whenever our next ubuntu/devel release is performed.

blackboxsw:
I also added the tag `ubuntu/22.3.3-0ubuntu1_22.10.1` to `upstream/ubuntu/kinetic-22.3-hotfix` as a packaging reference.

PR #1742, opened 2022-09-20
Ubuntu/bionic

aciba90:
do not squash
SC-1258


TheRealFalcon (commented):
Even though it's likely to change next release, we should bump the version number here so it's not confused with what has already been released. I think we can use the same number but `18.04.2` instead of `18.04.1`.

blackboxsw:
> Even though the format is likely to change next release, we should bump the version number here so it's not confused with what has already been released. I think we can use the same number but `18.04.2` instead of `18.04.1`.

+1. Let's just change and merge w/ `.2`. Even though this whole version line will change anyway at next `new-upstream-snapshot` because this top section of d/changelog is in `UNRELEASED` state which will be aggregated and re-versioned with any subsequent commits pulled in on next SRU.

blackboxsw (approved):
LGTM

PR #1743, opened 2022-09-20
Update #include example with real URLs

TheRealFalcon:
## Proposed Commit Message
```
Update #include example with real URLs

LP: #1307667
```

## Test Steps
Launch and LXD container with the include example and verify files written in both cloud-configs are written to their respective locations.


aciba90 (approved):
+1 thanks

holmanb (approved):

PR #1749, opened 2022-09-22
doc: Add configuration explanation (SC-1169)

TheRealFalcon:
Tbh, when I started this document, I thought I'd have more content for it. My goals are:
1. Document our base configuration, how it's generated, and how the pieces get layered on. We had documentation for user data and vendor data independently, but not the config as a whole
2. Help in understanding the difference between /etc/cloud/cloud.cfg and #cloud-config user data
3. Answer some common questions we get around configuration...in particular how do I modify my network or user config

I'm open to updates or suggestions for additional information.

Also, I'm trying to fit new documents into a diataxis category, even if they stick out like a sore thumb. This does indeed stick out like a sore thumb, but I think it's better to start that way and migrate other documents accordingly rather than try to use our outdated categories.

### DIFF
--- a/doc/rtd/topics/configuration.rst
+++ b/doc/rtd/topics/configuration.rst
@@ -13,7 +13,7 @@ From lowest priority to highest, configuration sources are:
 * **Hardcoded config**: Config_ that lives within the source of cloud-init
   and cannot be changed.
 * **Configuration directory**: Anything defined in ``/etc/cloud/cloud.cfg`` and
-  ``/etc/cloud/cloud.cfg.d``
+  ``/etc/cloud/cloud.cfg.d``.
 * **Runtime config**: Anything defined in ``/run/cloud-init/cloud.cfg``
 * **Kernel command line**: On the kernel command line, anything found between
   ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
--- a/doc/rtd/topics/configuration.rst
+++ b/doc/rtd/topics/configuration.rst
@@ -1,12 +1,15 @@
 .. _configuration:
 
-Configuration
-*************
+Configuration Sources
+*********************
 
 Internally, cloud-init builds a single configuration that is then referenced
 throughout the life of cloud-init. The configuration is built from multiple
 sources such that if a key is defined in multiple sources, the higher priority
 source overwrites the lower priority source.
+
+Base Configuration
+==================
 
 From lowest priority to highest, configuration sources are:
 
@@ -18,17 +21,22 @@ From lowest priority to highest, configuration sources are:
 * **Kernel command line**: On the kernel command line, anything found between
   ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
 
-These four sources make up the **base configuration**. Added to these
-to provide the full configuration are:
+These four sources make up the base configuration.
+
+Vendor and User Data
+====================
+Added to the base configuration are:
 
 * **Vendor data**: :ref:`Data<vendordata>` provided by the datasource
 * **User data**: :ref:`Data<user_data_formats>` also provided by
   the datasource
 
+These get fetched from the datasource and are defined at instance launch.
+
 .. note::
-  Even though baseconfiguration is found in ``/etc/cloud/cloud.cfg``, it is
-  distinct from :ref:`#cloud-config<topics/format:Cloud Config Data>`, which
-  is a specific user data format.
+  While much of what is defined in the base configuration can be overridden by
+  vendor data and user data, base configuration sources do not conform to
+  :ref:`#cloud-config<topics/format:Cloud Config Data>`
 
 Network Configuration
 =====================
--- a/doc/rtd/topics/configuration.rst
+++ b/doc/rtd/topics/configuration.rst
@@ -1,0 +1,81 @@
+.. _configuration:
+
+Configuration
+*************
+
+Internally, cloud-init builds a single configuration that is then referenced
+throughout the life of cloud-init. The configuration is built from multiple
+sources such that if a key is defined in multiple sources, the higher priority
+source overwrites the lower priority source.
+
+From lowest priority to highest, configuration sources are:
+
+* **Hardcoded config**: Config_ that lives within the source of cloud-init
+  and cannot be changed.
+* **Configuration directory**: Anything defined in ``/etc/cloud/cloud.cfg`` and
+  ``/etc/cloud/cloud.cfg.d``
+* **Runtime config**: Anything defined in ``/run/cloud-init/cloud.cfg``
+* **Kernel command line**: On the kernel command line, anything found between
+  ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
+
+These four sources make up the **base configuration**. Added to these
+to provide the full configuration are:
+
+* **Vendor data**: :ref:`Data<vendordata>` provided by the datasource
+* **User data**: :ref:`Data<user_data_formats>` also provided by
+  the datasource
+
+.. note::
+  Even though baseconfiguration is found in ``/etc/cloud/cloud.cfg``, it is
+  distinct from :ref:`#cloud-config<topics/format:Cloud Config Data>`, which
+  is a specific user data format.
+
+Network Configuration
+=====================
+Network configuration happens independently from other cloud-init
+configuration. It is sourced from the cloud provided metadata servce and is
+not user editable. If you need to change your network configuration,
+your options are:
+
+* Use the cloud-provided tools to update your network topology in the cloud.
+* :ref:`Disable cloud-init network configuration entirely
+  <topics/network-config:Disabling Network Configuration>`.
+  This leaves network configuration entirely up to you.
+
+Specifying Configuration
+==========================
+
+End users
+---------
+Pass :ref:`user data<user_data_formats>` to the cloud provider.
+Every platform supporting cloud-init will provide a method of supplying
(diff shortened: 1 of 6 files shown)


doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. note::
+  Even though baseconfiguration is found in ``/etc/cloud/cloud.cfg``, it is
+  distinct from :ref:`#cloud-config<topics/format:Cloud Config Data>`, which
+  is a specific user data format.
+
+Network Configuration
+=====================
+Network configuration happens independently from other cloud-init
+configuration. It is sourced from the cloud provided metadata servce and is
```
aciba90, line 36:
```suggestion
configuration. It is sourced from the cloud provided metadata service and is
```

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,81 @@
...
+throughout the life of cloud-init. The configuration is built from multiple
+sources such that if a key is defined in multiple sources, the higher priority
+source overwrites the lower priority source.
+
+From lowest priority to highest, configuration sources are:
+
+* **Hardcoded config**: Config_ that lives within the source of cloud-init
+  and cannot be changed.
+* **Configuration directory**: Anything defined in ``/etc/cloud/cloud.cfg`` and
+  ``/etc/cloud/cloud.cfg.d``
```
aciba90, line 16:
```suggestion
  ``/etc/cloud/cloud.cfg.d``.
```

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,81 @@
...
+* **Runtime config**: Anything defined in ``/run/cloud-init/cloud.cfg``
```
aciba90, line 17:
```suggestion
* **Runtime config**: Anything defined in ``/run/cloud-init/cloud.cfg``.
```

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,81 @@
...
+  ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
+
+These four sources make up the **base configuration**. Added to these
+to provide the full configuration are:
+
+* **Vendor data**: :ref:`Data<vendordata>` provided by the datasource
+* **User data**: :ref:`Data<user_data_formats>` also provided by
+  the datasource
+
+.. note::
```
aciba90, line 28: I found this note a bit difficult to understand. 

As said in the previous paragraph, base configuration is merged from 4 sources, not only `/etc/cloud/cloud.cfg`. Reading this, I ask my self, have the other 3 sources to follow _the specific user data format_?

Maybe it's worth rephrasing this to something in the lines of:

base-configuration sources do not conform to :ref:`#cloud-config<topics/format:Cloud Config Data>`, ...

aciba90 (changes requested):
Thanks for this doc, @TheRealFalcon. I think it helps definitely to understand cloud-init's configuration from a high level point of view.

I left inline-comments to fix a typo, punctuation consistency and suggestion to improve a note.

doc/rtd/topics/configuration.rst
TheRealFalcon: Thanks. I didn't love the wording either, so you make a good point. The impetus for it is that before we had jinja templating in /etc/cloud*, multiple people have confused the #cloud-config mentioned [here](https://cloudinit.readthedocs.io/en/latest/topics/instancedata.html#using-instance-data) with /etc/cloud/cloud.cfg. Part of that confusion probably has something to do with them both being casually referred to as "cloud config".

I update that section as well.

TheRealFalcon:
I started adding some context to the instance-data page and got a bit carried away. I re-structured the whole thing and re-wrote much of it. :smile: 

@blackboxsw , I know you wrote/added a lot of info to the instance-data page, so do you mind reviewing to make sure you're ok with the result? I removed a lot of info that felt irrelevant to the average user trying to figure out what instance data is and how to use it,  and I pushed reference details to the bottom.

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Network Configuration
+=====================
+Network configuration happens independently from other cloud-init
+configuration. It is sourced from the cloud provided metadata service and is
+not user editable. If you need to change your network configuration,
+your options are:
+
+* Use the cloud-provided tools to update your network topology in the cloud.
+* :ref:`Disable cloud-init network configuration entirely
```
blackboxsw, line 49: A third option here is to provide network configuration in a customized image directly in `/etc/cloud/cloud.cfg.d/*.cfg` with a network: <<config>> definition.

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,89 @@
...
+* **Kernel command line**: On the kernel command line, anything found between
+  ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
+
+These four sources make up the base configuration.
+
+Vendor and User Data
+====================
+Added to the base configuration are:
+
+* **Vendor data**: :ref:`Data<vendordata>` provided by the datasource
```
blackboxsw, line 30: Should we also reference metadata here too?

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,89 @@
...
+Distro Providers
+----------------
+Modify the base config. This often involves submitting a PR to modify
+the base `cloud.cfg template`_, which is used to customize
+`/etc/cloud/cloud.cfg` per distro. Additionally, a file can be added to
+``/etc/cloud/cloud.cfg.d`` to override a piece of the base configuration.
+
+Cloud Providers
+---------------
+Pass vendor data. This is the preferred method for clouds to provide
```
blackboxsw, line 82: worth also mentioning meta-data provided by clouds here?

doc/rtd/topics/configuration.rst
```diff
@@ -0,0 +1,89 @@
...
+related cloud-init documentation in the :ref:`datasource<datasources>`
+section.
+
+Once an instance has been initialized, the user data may not be edited.
+It is sourced directly from the cloud, so even if you find a local file
+that contains user data, it will likely be overwritten next boot.
+
+To modify the run frequency of a particular cloud-init module,
+:ref:`cloud-init single<cli_single>` may be used. This can be used to
+disable a module you no longer with to run.
```
blackboxsw, line 71: This is not true is it?  The CLI doesn't modify the run frequency of a module, it allows you to override that frequency manually on individual calls.

Maybe something like:
```suggestion
To override the default run frequency of a particular cloud-init module,
:ref:`cloud-init single<cli_single>` may be invoked from scripts, cloud-config `runcmd` or `boodcmd` sections .
 ```

Not sure we want to document the following but:


To modify the default frequency of a particular cloud-init module to `once-per-instance`, `always`, or `once`. Base config in /etc/cloud/cloud.cfg(.d) is needed. To override default frequency for a particular module, provide a list of values for the specific module name of the format `[<module_name>, <frequency>]` in any of the cloud-init boot stages `cloud_init_modules`, `cloud_config_modules` or `cloud_final_modules`:

```
cloud_init_modules:
 - [ write-files, once-per-instance]
....
```

blackboxsw (commented):
Looks great. a couple questions about whether/where we want to document metadata in config sources and prioritization and whether we should be referencing network v1 vs v2 inputs types when telling folks about ways to provide network config.

doc/rtd/topics/configuration.rst
TheRealFalcon: Doh, good catch. I misread the documentation for cloud-init single, so I've also updated that documentation to be slightly clearer. I also removed this bit for reasons I'll explain in a larger comment on the PR.

doc/rtd/topics/configuration.rst
TheRealFalcon: Is it? What metadata do we fetch from clouds that can be overridden by vendor data?

doc/rtd/topics/configuration.rst
TheRealFalcon: Same comment as above. I'm not sure how other metadata is relevant here.

TheRealFalcon:
I addressed the comments made so far. In addition to what I have here, I have a separate ticket to provide a schema for, and document the type of base configuration that would go into /etc/cloud/cloud.cfg . That should be more of a reference-style documentation, and I think some decisions I make there might dictate what I include here.

If you think this module is valuable as-is (or maybe with another change or two), we can merge it now, but I'm also ok setting it WIP and waiting to see what the reference documentation looks like to influence what's happening here.

Any strong opinions?

doc/rtd/topics/configuration.rst
blackboxsw: I was thinking specifically of `local_hostname` in metadata is the most likely case where metadata gets overridden by a #cloud-config `hostname` value in either vendor-data or user-data. I can't think of more explicit examples off the top of my head but DataSource does provide a get_config_obj which some datasources (Azure) provide which returns ds.cfg and will participate in the merged config via [ConfigMerger._read_cfg](https://github.com/canonical/cloud-init/blob/main/cloudinit/helpers.py#L263-L282). I think generally it's not frequently a case where that ds.cfg data participates, but there is a mechanism that sets priority at datasource metadata config, vendor data then user data. Not worth holding up this PR as-is, but something we can iterate on more in the future if needed.

blackboxsw (approved):
This is really great James. Thanks for cleaning this up.

PR #1751, opened 2022-09-26
cc_ubuntu_advantage: Handle already attached on Pro

aciba90:
## Proposed Commit Message

```
cc_ubuntu_advantage: Handle already attached on Pro

On Pro instances, UA could raise an exception signalizing already
attached instances. Catch it and warn if some service is given in
the cloud-config to be configured.

Improve integration test:
- mask ua services on first boot to upgrade UA on Pro
  instances and let the second boot be clean
- assert the correct services were enabled
- detect instances where cloud-init does yet support custom
  auto-attach and fail the test.
```

## Test Steps

```sh
CLOUD_INIT_PLATFORM=<gce|azure|ec2>  tox -e integration-tests -- tests/integration_tests/modules/test_ubuntu_advantage.py::TestUbuntuAdvantagePro
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -398,8 +398,9 @@ def _auto_attach(ua_section: dict):
         if enable_beta is not None or enable is not None:
             # Only warn if the user defined some service to enable/disable.
             LOG.warning(
-                "The instance is already attached."
-                " Leaving Pro services untouched."
+                "The instance is already attached to Pro. Leaving enabled"
+                " services untouched. Ignoring config directives"
+                " ubuntu_advantage: enable and enable_beta"
             )
     except UserFacingError as ex:
         msg = f"Error during `full_auto_attach`: {ex}"
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -395,7 +395,7 @@ def _auto_attach(ua_section: dict):
     try:
         full_auto_attach(options=options)
     except AlreadyAttachedError:
-        if enable_beta is None or enable is None:
+        if enable_beta is not None or enable is not None:
             # Only warn if the user defined some service to enable/disable.
             LOG.warning(
                 "The instance is already attached."
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -373,7 +373,10 @@ def _attach(ua_section: dict):
 
 def _auto_attach(ua_section: dict):
     try:
-        from uaclient.api.exceptions import UserFacingError
+        from uaclient.api.exceptions import (
+            AlreadyAttachedError,
+            UserFacingError,
+        )
         from uaclient.api.u.pro.attach.auto.full_auto_attach.v1 import (
             FullAutoAttachOptions,
             full_auto_attach,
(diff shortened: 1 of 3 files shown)


tests/integration_tests/modules/test_ubuntu_advantage.py
```diff
@@ -31,6 +31,8 @@
 package_upgrade: true
 packages:
 - ubuntu-advantage-tools
+bootcmd:
+- systemctl mask ua-auto-attach.service
```
orndorffgrant, line 35: `ua-auto-attach.service` will check for UA cloud-init data and noop
`ubuntu-advantage.service` is the GCE daemon that will attach anyway - I think that is the one you need to mask for this test to work

orndorffgrant: Now that I wrote that, `ubuntu-advantage.service` is `After: cloud-config.service` so it shouldn't be attaching before cloud-init would. 

It will automatically re-auto-attach if you run `ua detach` on GCE though

aciba90: I have disabled both units and performed a `pro detach` to be sure it is really detached on second boot. Please, @orndorffgrant, validate this is a correct set-up on first boot.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -383,12 +386,21 @@ def _auto_attach(ua_section: dict):
...
-        enable=ua_section.get("enable"),
-        enable_beta=ua_section.get("enable_beta"),
+        enable=enable,
+        enable_beta=enable_beta,
     )
     try:
         full_auto_attach(options=options)
+    except AlreadyAttachedError:
+        if enable_beta is None or enable is None:
+            # Only warn if the user defined some service to enable/disable.
```
orndorffgrant, line 399: I think for the condition to match the comment, this should be
```suggestion
        if enable_beta is not None or enable is not None:
            # Only warn if the user defined some service to enable/disable.
```

tests/integration_tests/modules/test_ubuntu_advantage.py
```diff
@@ -31,6 +33,9 @@
 package_upgrade: true
 packages:
 - ubuntu-advantage-tools
+bootcmd:
+- sudo systemctl mask ubuntu-advantage-pro.service
+- sudo systemctl mask ubuntu-advantage.service
```
orndorffgrant, line 38: `ubuntu-advantage-pro.service` doesn't exist :). Sorry if my earlier messages were confusing. For context, `ubuntu-advantage-pro` is the name of a separate package that delivers a systemd unit called `ua-auto-attach.service` (this is confusing tbh haha).

But I think this will be good just disabling `ubuntu-advantage.service`.
```suggestion
bootcmd:
- sudo systemctl mask ubuntu-advantage.service
```

aciba90: Thanks for the context there. Fixed and retested.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Oopsie, fixed.

blackboxsw (approved):
Other than the minor logging request. This looks good. Adapt the logging a bit and let's land it.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -383,12 +386,21 @@ def _auto_attach(ua_section: dict):
...
+        enable_beta=enable_beta,
     )
     try:
         full_auto_attach(options=options)
+    except AlreadyAttachedError:
+        if enable_beta is not None or enable is not None:
+            # Only warn if the user defined some service to enable/disable.
+            LOG.warning(
+                "The instance is already attached."
+                " Leaving Pro services untouched."
```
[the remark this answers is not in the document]
aciba90: Done

blackboxsw (approved):

PR #1756, opened 2022-09-27
Feature: Kernel Modules

chifac08:
## Proposed Commit Message

```
cc_kernel_modules: Module to manage kernel modules.

This module is capable of loading a kernel module `/etc/modules-load.d` at boot and enhance 
a kernel module with modprobe parameters (https://manpages.ubuntu.com/manpages/trusty/man5/modprobe.d.5.html). 
Beside applying settings during runtime it will also persist all settings in `/etc/modules-load.d` 
as well as `/etc/modprobe.d`.
```

## Additional Context
For examples please have look at `MetaSchema`.

## Checklist:


chifac08:
Hi cloud-init community and maintainers,

What do you think about such a feature? Any input is much appreciated!

thx

TheRealFalcon:
@chifac08 I'm assuming we'd add/remove any of the modules from `/etc/modules-load.d` depending on the `load` key, and that anything under the `modprobe` key would go into `/etc/modprobe.d`?

The idea sounds good to me!

holmanb:
Hi @chifac08,

Thanks for the proposal. 

A couple thoughts:

I expect the common cases for early boot to be handled during early boot by initrd/initramfs, with udev handling other common drivers as well.

I could see this being helpful on a scientific computing cloud perhaps, where you have some special hardware (gpu/accelerator/rdma nic/etc).

In the case where your hardware driver isn't upstream or in your image already, manually loading modules downloaded from a repo is arguably less complicated than writing udev rules[1] and more elegant than `runcmd`.

No objections from me for the idea, +1

chifac08:
> @chifac08 I'm assuming we'd add/remove any of the modules from `/etc/modules-load.d` depending on the `load` key, and that anything under the `modprobe` key would go into `/etc/modprobe.d`?
> 
> The idea sounds good to me!

@TheRealFalcon @holmanb 
Thanks for your input. Yes, your assumption is correct. Maybe i should explain my idea a bit more detailed. Let's assume we have a config like this:

```
kernel_modules: 
  - name: v4l2loopback
    load: true
    modprobe:
      options: "devices=1 video_nr=20 card_label=fakecam exclusive_caps=1"
```
The module will create following files...

- /etc/modules-load.d/cloud-init.conf:
  `v4l2loopback`
- /etc/modprobe.d/cloud-init.conf:
  `options v4l2loopback devices=1 video_nr=20 card_label=fakecam exclusive_caps=1`

and reload kernel modules by restarting `systemd-modules-load.service` and update `initramfs` by issuing `update-initramfs -u -k all`.
Module will run `PER INSTANCE` and always overwrites config files with given elements (if there are no elements, both config files will be deleted) to ensure idempotence.

igalic:
If we can come up with a better name for the `modprobe` key, we can make it platform independent from the get-go.

How about `persist` ?

TheRealFalcon:
@chifac08 , I noticed you added a lot of functionality here. Is this still intended to be in draft state? Are you looking for more review before proceeding?

chifac08:
@TheRealFalcon I was very busy the last weeks and wanted to make sure to have a stable version of my PR including unit and integration tests.
Finally, I found time and here it is ready to review :)

Be aware, I am still working on documentation.

thx

tests/integration_tests/modules/test_kernel_modules.py
```diff
@@ -0,0 +1,120 @@
...
+            instance.name, KERNEL_MODULES_LXD
+        ).split()
+    )
+
+
+@pytest.mark.user_data(USER_DATA)
+@pytest.mark.ci
+@pytest.mark.lxd_vm
+@pytest.mark.ubuntu
+class TestKernelModules:
```
aciba90, line 44: I think we could verify that there are no errors/warnings in the log. That will catch errors easily:

```suggestion
class TestKernelModules:
   def test_clean_log(self, class_client: IntegrationInstance):
       verify_clean_log(class_client)
```

Same for the other test class.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,399 @@
...
+            f"Failed to update initial ramdisk:{NL}{str(e)}"
+        ) from e
+
+
+def handle(
+    name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
+) -> None:
+    kernel_modules_section = None
+
+    if "kernel_modules" in cfg:
```
aciba90, line 346: Having a look at this if-else block, if this module receives a config as:

```
cc_kernel_modules:
  kernel_modules:
```

Then the flow of execution is going to pass through the `if` block and `handle` will remove any existent configuration files but won't unload the modules, which sounds like an inconsistent state.

Unless I miss something, I think this case should also hit the else block and noop to avoid surprises. Something in the lines of:

```diff
diff --git a/cloudinit/config/cc_kernel_modules.py b/cloudinit/config/cc_kernel_modules.py
index 48499bbf9..01140ad1e 100644
--- a/cloudinit/config/cc_kernel_modules.py
+++ b/cloudinit/config/cc_kernel_modules.py
@@ -341,14 +341,11 @@ def update_initial_ramdisk(distro_cfg: dict):
 def handle(
     name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
 ) -> None:
-    kernel_modules_section = None
-
-    if "kernel_modules" in cfg:
-        LOG.debug("Found kernel_modules section in config")
-        kernel_modules_section = util.get_cfg_option_list(
+    kernel_modules_section = util.get_cfg_option_list(
             cfg, "kernel_modules", []
         )
-    else:
+
+    if not kernel_modules_section:
         LOG.debug(
             "Skipping module named %s, no "
             "'kernel_modules' configuration found",
@@ -356,6 +353,8 @@ def handle(
         )
         return
 
+    LOG.debug("Found kernel_modules section in config")
+
     # check systemd usage
     if not uses_systemd:
         LOG.debug(
```

aciba90 (changes requested):
Thanks, @chifac08, for improving cloud-init.

The approach looks good overall. I left some inline comments, and I have a couple of generic ones:

1) At the end of this module, after the desired config is set we restart `systemd-modules-load` to let it modify the kernel modules. We verify that the exit code of `systemd-modules-load` is 0 and continue, so if some error happens during `systemd-modules-load` execution, `cloud-init` would not see it and report a correct execution. I think `cloud-init` should report back to the user if the system is misconfigured, but at the same time I do not see a way to do so without blocking `cc_kernel_modules` until `systemd-modules-load` finishes. Thoughts?

2) This module can load/unload/tweak kernel modules that are already present on the system. I think a common question from consumers' perspective is going to be: how do I work with a module that is not present on the image? Maybe we could underline this in the documentation and hint users that they could inject new kernel modules by using `bootcmd` or other means.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,399 @@
...
+        else:
+            UNLOAD_MODULES.append(module["name"])
+
+        # Enhance module
+        if module.get("persist"):
+            enhance_module(distro_cfg, module["name"], module["persist"])
+
+    # rebuild initial ramdisk
+    log.info("Update initramfs")
+    update_initial_ramdisk(distro_cfg)
```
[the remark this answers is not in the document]
chifac08: I think we can ignore the order of `reload_modules` and `update_initial_ramdisk` functions as documentation for[ modules-load.d](https://manpages.ubuntu.com/manpages/bionic/man5/modules-load.d.5.html) states the following:
"
If the administrator wants to disable a configuration file supplied by the vendor, the
recommended way is to place a symlink to /dev/null in the configuration directory in 
[/etc/](file:///etc/), with the same filename as the vendor configuration file. If the vendor
configuration file is included in the initrd image, the image has to be regenerated.
"
For me this concludes that, as long as configuration files were written in `/etc`, `update-initramfs` will respect it. This state has already been reached when executing `update_initial_ramdisk` function.

Does that sound good to you?

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,399 @@
...
+            "path": "/etc/modules-load.d/cloud-init.conf",
+            "permissions": 0o600,
+        },
+        "persist": {
+            "path": "/etc/modprobe.d/cloud-init.conf",
+            "permissions": 0o600,
+        },
+    },
+}
+DISTRO_OVERRIDES = {}  # type: dict
```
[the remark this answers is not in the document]
chifac08: Thank you for your explanation. I did not know that such functionality exists and copied the current idea from `cc_ca_certs` module. Nevertheless, my main focus for this module was to make it available for `debian` based distros with an option to easy adapt it for other distributions. I'll remove that variable.

cloudinit/config/cc_kernel_modules.py
chifac08: Agree

chifac08:
> Thanks, @chifac08, for improving cloud-init.
> 
> The approach looks good overall. I left some inline comments, and I have a couple of generic ones:
> 
> 1. At the end of this module, after the desired config is set we restart `systemd-modules-load` to let it modify the kernel modules. We verify that the exit code of `systemd-modules-load` is 0 and continue, so if some error happens during `systemd-modules-load` execution, `cloud-init` would not see it and report a correct execution. I think `cloud-init` should report back to the user if the system is misconfigured, but at the same time I do not see a way to do so without blocking `cc_kernel_modules` until `systemd-modules-load` finishes. Thoughts?
> 2. This module can load/unload/tweak kernel modules that are already present on the system. I think a common question from consumers' perspective is going to be: how do I work with a module that is not present on the image? Maybe we could underline this in the documentation and hint users that they could inject new kernel modules by using `bootcmd` or other means.

Thank you for reviewing. I implemented all suggestions you addressed. Concerning your points:

1. I deeply looked into documentation for `systemd-modules-load.service` again and made some tests. Unfortunately, `/lib/systemd/systemd-modules-load` executable does not return an error code unequal to zero when a module can't be loaded due to an error (only exits unequal zero if a path does not exist - [Link](https://github.com/systemd/systemd/blob/dcebf1d87efea7b0016e0997b0bffc2c3fbd6c1a/src/modules-load/modules-load.c#L98)). Furthermore, calling the unit file will always reload all config files which should not be the way to go as we only wan't to ensure that `/etc/modules-load./cloud-init.conf` gets loaded. Finally, I decided to use `/lib/systemd/systemd-modules-load` executable directly to explicitly load kernel modules from `/etc/modules-load./cloud-init.conf` and parse output to handle possible errors (e.g. when a module does not exist). Let me know what you think
2. I am still writing on documentation but you are totally right. I intentionally placed the module in `config_final_modules` section, directly after `package-update-upgrade-install`, to give the user the possibility to install an kernel module an load it afterwards. For example:

```
#cloud-config
packages:
  - wireguard
kernel_modules:
  - name: wireguard
    load: true
```

Furthermore, I want to discuss the usage of `TypedDict` in cloud-init. `TypedDict` simplifies reading `DEFAULT_CONFIG` by defining a custom definition of an dictionary and stops `mypy` from complaining about collections. Unfortunately, `TypedDict` was moved from `typing-extensions` to `typing` lib in Python 3.8. This means, we have to check what python version is installed and import it from different libs:
```
if sys.version_info >= (3, 8):
    from typing import TypedDict
else:
    from typing_extensions import TypedDict
```

`typing-extensions` lib is not defined as an requirement for cloud-init, so this module will currently fail when using a Python version lower 3.8.

My questions..

- Is the usage of `TypedDict` prohibited? Does someone see a better approach?
- If this approach is ok, should i add `typing-extensions` to requirements.txt?


Last but not least, i want to talk about integration tests. I do not see a way to implement meaningful integration tests for LXD containers as loading kernel modules won't work in a container. Thoughts?

Thanks

cloudinit/config/cc_kernel_modules.py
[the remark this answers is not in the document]
aciba90: It sounds good to me, thanks for the context there.

aciba90 (changes requested):
Thanks, @chifac08, for the improvements.

> 1. I deeply looked into documentation for systemd-modules-load.service again and made some tests. Unfortunately, /lib/systemd/systemd-modules-load executable does not return an error code unequal to zero when a module can't be loaded due to an error (only exits unequal zero if a path does not exist - Link). Furthermore, calling the unit file will always reload all config files which should not be the way to go as we only wan't to ensure that /etc/modules-load.d/cloud-init.conf gets loaded. Finally, I decided to use /lib/systemd/systemd-modules-load executable directly to explicitly load kernel modules from /etc/modules-load.d/cloud-init.conf and parse output to handle possible errors (e.g. when a module does not exist). Let me know what you think

Having a deeper look here, `sudo systemctl restart systemd-modules-load` does actually block and provide feedback to the user.
I think the previous approach was more resilient than calling directly 
`/lib/systemd/systemd-modules-load cloud-init.conf` as:

1. The binary path could change more often than the service name.
2. We bypass the service configuration. For example, at least in Ubuntu, the service has a timeout of 90s that we would ignore.
3. If the user performs a `systemctl status systemd-modules-load` in the booted instance, then we will get the status of the 
first execution of the service and not the status of our call.
4. Loading only the modules defined in cloud-init.conf could be faster but we would ignore the full `.d` config resolution.
Imagine that we shadow the module config of other files or the other way around. In that case, if machine is rebooted and all conf
files are taken into account by `system-modules-load`, then the instance will end up in a different state (no idempotency).

Therefore, I think the initial approach (calling `cloud.distro.manage_service("restart", "systemd-modules-load")`) is more robust than the current one. What do you think?

> My questions..
> - Is the usage of TypedDict prohibited? Does someone see a better approach?
> - If this approach is ok, should i add typing-extensions to requirements.txt?

I think `typing-extensions` would be very useful in cloud-init to fastly adopt newer type hints but, unfortunately, there are some dificulties to do so.
The main issue is that it is required at runtime, therefore every cloud-init downstream would have to include that dependency.
For instance, in the Ubuntu case, `python3-typing-exentions` is not available in Bionic, thus we should package it for bionic or wait for bionic EOL.
Similarly, other distros could lack the `typing-extensions` package.

[Here](https://github.com/canonical/cloud-init/blob/12df62554e0229d697238f868ec9ff3e212d011b/cloudinit/config/schema.py#L70-L92) we have an example (the only one I think) where we use `typing-extensions` only at static analysis time. In this case we are typing `MetaSchema`,
which touches every config module and more places of the code-base, that's why we decided the additional
complexity required to check if we are at typing time or run time was worth it. A bit more context on that: https://github.com/canonical/cloud-init/pull/1589.

We decided to evaluate further imports from `typing-extensions` in a case by case manner. In our current case, we could either:

1. Take a similar approach as the one taken for `MetaSchema`.
2. Restrict ourselves to types supported in python 3.6.

Taking into account that the entity that we are discussing is constant and local to `cc_kernel_modules`, I lean more towards 2. What do you think?

> Last but not least, i want to talk about integration tests.
> I do not see a way to implement meaningful integration tests for LXD containers as loading kernel modules won't work in a container. Thoughts?

In the case where we need to interact with kernel modules we can use `lxd_vm` and/or cloud instances.
The tests marked with `ci` in a non-container instance are executed as part of an internal daily CI.
To run them locally execute `CLOUD_INIT_PLATFORM=lxd_vm tox -e integration-tests -- -vv tests/integration_tests/modules/test_kernel_modules.py`.
I found some errors while doing so: https://pastebin.ubuntu.com/p/mbsZXqZxGd/

In the [Travis CI run](https://app.travis-ci.com/github/canonical/cloud-init/jobs/587049868) (`lxd_container`), I noticed there are some
unrelated tests failing, I think a rebase of `main` could solve them.

heldenhelard (commented):
#1826 #1563 @

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,395 @@
...
+
+LOG = logging.getLogger(__name__)
+NL = "\n"
+REQUIRED_KERNEL_MODULES_KEYS = frozenset(["name"])
+
+DefaultConfigType = Dict[
+    str, Union[Dict[str, List[str]], Dict[str, Dict[str, Union[str, int]]]]
+]
+
+DEFAULT_CONFIG: DefaultConfigType = {
```
aciba90, line 61: What about the following? I do not see a lot of value in this complex type, even mypy cannot properly analyze the code: https://github.com/canonical/cloud-init/actions/runs/3465208769/jobs/5787668073

```suggestion
DEFAULT_CONFIG: dict = {
```

aciba90 (changes requested):
Thanks, @chifac08, for the work here.

Having a look into the failing integration tests, I have noticed some schema validation errors.

tldr: 

```diff
diff --git a/cloudinit/config/schemas/schema-cloud-config-v1.json b/cloudinit/config/schemas/schema-cloud-config-v1.json
index 378f0f941..42e3482fe 100644
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -27,6 +27,7 @@
         "grub_dpkg",
         "install-hotplug",
         "install_hotplug",
+        "kernel_modules",
         "keyboard",
         "keys-to-console",
         "keys_to_console",
```

In https://github.com/canonical/cloud-init/pull/1770, we included validation of the [base config](https://cloudinit.readthedocs.io/en/latest/topics/base_config_reference.html) modules lists that can be
overridden with user data. After that change, we have to declare the new module under `all_modules`. 

This should be documented in [module_creation](https://cloudinit.readthedocs.io/en/latest/topics/module_creation.html), but it is not. (We will fix that).

cloudinit/config/cc_kernel_modules.py
chifac08: Agreed :)

chifac08:
> Thanks, @chifac08, for the work here.
> 
> Having a look into the failing integration tests, I have noticed some schema validation errors.
> 
> tldr:
> 
> ```diff
> diff --git a/cloudinit/config/schemas/schema-cloud-config-v1.json b/cloudinit/config/schemas/schema-cloud-config-v1.json
> index 378f0f941..42e3482fe 100644
> --- a/cloudinit/config/schemas/schema-cloud-config-v1.json
> +++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
> @@ -27,6 +27,7 @@
>          "grub_dpkg",
>          "install-hotplug",
>          "install_hotplug",
> +        "kernel_modules",
>          "keyboard",
>          "keys-to-console",
>          "keys_to_console",
> ```
> 
> In #1770, we included validation of the [base config](https://cloudinit.readthedocs.io/en/latest/topics/base_config_reference.html) modules lists that can be overridden with user data. After that change, we have to declare the new module under `all_modules`.
> 
> This should be documented in [module_creation](https://cloudinit.readthedocs.io/en/latest/topics/module_creation.html), but it is not. (We will fix that).

Implemented changes you requested. Furthermore, I added an short module description. Let me know what you think. 
Thanks

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+from cloudinit.config.schema import MetaSchema, get_meta_doc
+from cloudinit.settings import PER_INSTANCE
+
+MODULE_DESCRIPTION = dedent(
+    """\
+Manages and enhances kernel modules on a systemd based system.
+This module is capable of loading kernel modules at boot as well as
+enhance it with parameters.
+Beside applying settings during runtime it will also persist all
+settings in ``/etc/modules-load.d`` and ``/etc/modprobe.d`.
```
aciba90, line 24:
```suggestion
settings in ``/etc/modules-load.d`` and ``/etc/modprobe.d``.
```

aciba90 (changes requested):
It looks great, @chifac08. A couple of minor things:

1. To render the new module doc, we need to include it in [doc/rtd/topics/modules.rst](https://github.com/canonical/cloud-init/blob/main/doc/rtd/topics/modules.rst). This is not documented in https://cloudinit.readthedocs.io/en/latest/topics/module_creation.html , we will fix it.
2. About the integration tests, the two classes only differ on the required LXD setup for containers. I would suggest to abstract that difference using a base class.

The following patch addresses both things and enables the tests to be executed on clouds. Feel free to modify it.

```diff
diff --git a/doc/rtd/topics/modules.rst b/doc/rtd/topics/modules.rst
index b0ad83e49..58bed3f02 100644
--- a/doc/rtd/topics/modules.rst
+++ b/doc/rtd/topics/modules.rst
@@ -19,6 +19,7 @@ Module Reference
 .. automodule:: cloudinit.config.cc_growpart
 .. automodule:: cloudinit.config.cc_grub_dpkg
 .. automodule:: cloudinit.config.cc_install_hotplug
+.. automodule:: cloudinit.config.cc_kernel_modules
 .. automodule:: cloudinit.config.cc_keyboard
 .. automodule:: cloudinit.config.cc_keys_to_console
 .. automodule:: cloudinit.config.cc_landscape
diff --git a/tests/integration_tests/modules/test_kernel_modules.py b/tests/integration_tests/modules/test_kernel_modules.py
index efeae8c0c..fe02bdbb7 100644
--- a/tests/integration_tests/modules/test_kernel_modules.py
+++ b/tests/integration_tests/modules/test_kernel_modules.py
@@ -37,42 +37,49 @@ def load_kernel_modules_lxd(instance: LXDInstance):
     )


-@pytest.mark.user_data(USER_DATA)
 @pytest.mark.ci
-@pytest.mark.lxd_vm
+@pytest.mark.user_data(USER_DATA)
 @pytest.mark.ubuntu
-class TestKernelModules:
+class BaseTest:
     @pytest.mark.parametrize(
         "cmd,expected_out",
         (
-            # check permissions
-            (
+            pytest.param(
                 "stat -c '%U %a' /etc/modules-load.d/50-cloud-init.conf",
                 r"root 600",
+                id="check-permissions",
             ),
-            (
+            pytest.param(
                 "stat -c '%U %a' /etc/modprobe.d/50-cloud-init.conf",
                 r"root 600",
+                id="check-permissions-2",
+            ),
+            pytest.param(
+                "file /etc/modules-load.d/50-cloud-init.conf",
+                ASCII_TEXT,
+                id="ASCII-check",
             ),
-            # ASCII check
-            ("file /etc/modules-load.d/50-cloud-init.conf", ASCII_TEXT),
-            ("file /etc/modprobe.d/50-cloud-init.conf", ASCII_TEXT),
-            # check loaded modules
-            (
+            pytest.param(
+                "file /etc/modprobe.d/50-cloud-init.conf",
+                ASCII_TEXT,
+                id="ASCII-check-2",
+            ),
+            pytest.param(
                 "lsmod | grep -e '^lockd\\|^ip_tables\\|^wireguard' | wc -l",
                 "3",
+                id="check-loaded-modules",
             ),
-            # sha256sum check modul
-            (
+            pytest.param(
                 "sha256sum </etc/modules-load.d/50-cloud-init.conf",
                 "9d14d5d585dd3e5e9a3c414b5b7af7ed"
                 "9d44e7ee3584652fbf388cad455b5053",
+                id="sha256sum-check-modules",
             ),
-            # sha256sum check modprobe
-            (
+            pytest.param(
                 "sha256sum   </etc/modprobe.d/50-cloud-init.conf",
                 "229ccc941ec34fc8c49bf14285ffeb65"
                 "ea2796c4840f9377d6df76bda42c878e",
+                id="sha256sum-check-modprobe",
             ),
         ),
     )
@@ -88,53 +95,12 @@ class TestKernelModules:
         verify_clean_log(log)


-@pytest.mark.ci
+@pytest.mark.no_container
+class TestKernelModules(BaseTest):
+    pass
+
+
 @pytest.mark.lxd_container
-@pytest.mark.user_data(USER_DATA)
 @pytest.mark.lxd_setup.with_args(load_kernel_modules_lxd)
-@pytest.mark.ubuntu
-class TestKernelModulesWithoutKmod:
-    @pytest.mark.parametrize(
-        "cmd,expected_out",
-        (
-            # check permissions
-            (
-                "stat -c '%U %a' /etc/modules-load.d/50-cloud-init.conf",
-                r"root 600",
-            ),
-            (
-                "stat -c '%U %a' /etc/modprobe.d/50-cloud-init.conf",
-                r"root 600",
-            ),
-            # ASCII check
-            ("file /etc/modules-load.d/50-cloud-init.conf", ASCII_TEXT),
-            ("file /etc/modprobe.d/50-cloud-init.conf", ASCII_TEXT),
-            # check loaded modules
-            (
-                "lsmod | grep -e '^lockd\\|^ip_tables\\|^wireguard' | wc -l",
-                "3",
-            ),
-            # sha256sum check modul
-            (
-                "sha256sum </etc/modules-load.d/50-cloud-init.conf",
-                "9d14d5d585dd3e5e9a3c414b5b7af7ed"
-                "9d44e7ee3584652fbf388cad455b5053",
-            ),
-            # sha256sum check modprobe
-            (
-                "sha256sum   </etc/modprobe.d/50-cloud-init.conf",
-                "229ccc941ec34fc8c49bf14285ffeb65"
-                "ea2796c4840f9377d6df76bda42c878e",
-            ),
-        ),
-    )
-    def test_kernel_modules(
-        self, cmd, expected_out, class_client: IntegrationInstance
-    ):
-        result = class_client.execute(cmd)
-        assert result.ok
-        assert expected_out in result.stdout
-
-    def test_clean_log(self, class_client: IntegrationInstance):
-        log = class_client.read_from_file("/var/log/cloud-init.log")
-        verify_clean_log(log)
+class TestKernelModulesWithoutKmod(BaseTest):
+    pass
```

aciba90:
As this PR includes the documentation of a new module, I added @s-makin as a reviewer.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+from cloudinit.cloud import Cloud
+from cloudinit.config import Config
+from cloudinit.config.schema import MetaSchema, get_meta_doc
+from cloudinit.settings import PER_INSTANCE
+
+MODULE_DESCRIPTION = dedent(
+    """\
+Manages and enhances kernel modules on a systemd based system.
+This module is capable of loading kernel modules at boot as well as
+enhance it with parameters.
```
s-makin, line 22:
```suggestion
enhancing it with parameters.
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+    except Exception as e:
+        raise RuntimeError(
+            f"Failure appending kernel module '{module_name}' to file "
+            f'{DEFAULT_CONFIG["km_files"]["load"]["path"]}:{NL}{str(e)}'
+        ) from e
+
+
+def enhance_module(module_name: str, persist: dict, unload_modules: list):
+    """Enhances a kernel modules behavior
```
s-makin, line 196:
```suggestion
    """Enhances a kernel module's behavior
```

s-makin, line 196:
```suggestion
    """Enhances a kernel modules behaviour
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+
+    This function appends specific settings and options
+    for a kernel module, which will be applied when kernel
```
s-makin, line 199:
```suggestion
    for a kernel module, which will be applied when the kernel
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+    module get's loaded.
```
s-makin, line 200:
```suggestion
    module gets loaded.
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+        except Exception as e:
+            raise RuntimeError(
+                f"Failure enhancing kernel module '{module_name}':{NL}{str(e)}"
+            ) from e
+
+
+def cleanup():
+    """Clean up all kernel specific files
+
+    This function removes all files, which are
```
s-makin, line 239:
```suggestion
    This function removes all files that are
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+        except Exception as e:
+            raise RuntimeError(
+                f"Could not delete file {file_path}:{NL}{str(e)}"
+            ) from e
+
+
+def reload_modules(cloud: Cloud):
+    """Reload kernel modules
+
+    This function reload modules in /etc/modules-load.d/cloud-init.conf
```
s-makin, line 259:
```suggestion
    This function reloads modules in /etc/modules-load.d/cloud-init.conf
```

s-makin (commented):
I've done a sweep for spelling etc. Only a few little nits. Looks good otherwise.

chifac08:
> It looks great, @chifac08. A couple of minor things:
> 
> 1. To render the new module doc, we need to include it in [doc/rtd/topics/modules.rst](https://github.com/canonical/cloud-init/blob/main/doc/rtd/topics/modules.rst). This is not documented in https://cloudinit.readthedocs.io/en/latest/topics/module_creation.html , we will fix it.
> 2. About the integration tests, the two classes only differ on the required LXD setup for containers. I would suggest to abstract that difference using a base class.
> 
> The following patch addresses both things and enables the tests to be executed on clouds. Feel free to modify it.
> 
> ```diff
> diff --git a/doc/rtd/topics/modules.rst b/doc/rtd/topics/modules.rst
> index b0ad83e49..58bed3f02 100644
> --- a/doc/rtd/topics/modules.rst
> +++ b/doc/rtd/topics/modules.rst
> @@ -19,6 +19,7 @@ Module Reference
>  .. automodule:: cloudinit.config.cc_growpart
>  .. automodule:: cloudinit.config.cc_grub_dpkg
>  .. automodule:: cloudinit.config.cc_install_hotplug
> +.. automodule:: cloudinit.config.cc_kernel_modules
>  .. automodule:: cloudinit.config.cc_keyboard
>  .. automodule:: cloudinit.config.cc_keys_to_console
>  .. automodule:: cloudinit.config.cc_landscape
> diff --git a/tests/integration_tests/modules/test_kernel_modules.py b/tests/integration_tests/modules/test_kernel_modules.py
> index efeae8c0c..fe02bdbb7 100644
> --- a/tests/integration_tests/modules/test_kernel_modules.py
> +++ b/tests/integration_tests/modules/test_kernel_modules.py
> @@ -37,42 +37,49 @@ def load_kernel_modules_lxd(instance: LXDInstance):
>      )
> 
> 
> -@pytest.mark.user_data(USER_DATA)
>  @pytest.mark.ci
> -@pytest.mark.lxd_vm
> +@pytest.mark.user_data(USER_DATA)
>  @pytest.mark.ubuntu
> -class TestKernelModules:
> +class BaseTest:
>      @pytest.mark.parametrize(
>          "cmd,expected_out",
>          (
> -            # check permissions
> -            (
> +            pytest.param(
>                  "stat -c '%U %a' /etc/modules-load.d/50-cloud-init.conf",
>                  r"root 600",
> +                id="check-permissions",
>              ),
> -            (
> +            pytest.param(
>                  "stat -c '%U %a' /etc/modprobe.d/50-cloud-init.conf",
>                  r"root 600",
> +                id="check-permissions-2",
> +            ),
> +            pytest.param(
> +                "file /etc/modules-load.d/50-cloud-init.conf",
> +                ASCII_TEXT,
> +                id="ASCII-check",
>              ),
> -            # ASCII check
> -            ("file /etc/modules-load.d/50-cloud-init.conf", ASCII_TEXT),
> -            ("file /etc/modprobe.d/50-cloud-init.conf", ASCII_TEXT),
> -            # check loaded modules
> -            (
> +            pytest.param(
> +                "file /etc/modprobe.d/50-cloud-init.conf",
> +                ASCII_TEXT,
> +                id="ASCII-check-2",
> +            ),
> +            pytest.param(
>                  "lsmod | grep -e '^lockd\\|^ip_tables\\|^wireguard' | wc -l",
>                  "3",
> +                id="check-loaded-modules",
>              ),
> -            # sha256sum check modul
> -            (
> +            pytest.param(
>                  "sha256sum </etc/modules-load.d/50-cloud-init.conf",
>                  "9d14d5d585dd3e5e9a3c414b5b7af7ed"
>                  "9d44e7ee3584652fbf388cad455b5053",
> +                id="sha256sum-check-modules",
>              ),
> -            # sha256sum check modprobe
> -            (
> +            pytest.param(
>                  "sha256sum   </etc/modprobe.d/50-cloud-init.conf",
>                  "229ccc941ec34fc8c49bf14285ffeb65"
>                  "ea2796c4840f9377d6df76bda42c878e",
> +                id="sha256sum-check-modprobe",
>              ),
>          ),
>      )
> @@ -88,53 +95,12 @@ class TestKernelModules:
>          verify_clean_log(log)
> 
> 
> -@pytest.mark.ci
> +@pytest.mark.no_container
> +class TestKernelModules(BaseTest):
> +    pass
> +
> +
>  @pytest.mark.lxd_container
> -@pytest.mark.user_data(USER_DATA)
>  @pytest.mark.lxd_setup.with_args(load_kernel_modules_lxd)
> -@pytest.mark.ubuntu
> -class TestKernelModulesWithoutKmod:
> -    @pytest.mark.parametrize(
> -        "cmd,expected_out",
> -        (
> -            # check permissions
> -            (
> -                "stat -c '%U %a' /etc/modules-load.d/50-cloud-init.conf",
> -                r"root 600",
> -            ),
> -            (
> -                "stat -c '%U %a' /etc/modprobe.d/50-cloud-init.conf",
> -                r"root 600",
> -            ),
> -            # ASCII check
> -            ("file /etc/modules-load.d/50-cloud-init.conf", ASCII_TEXT),
> -            ("file /etc/modprobe.d/50-cloud-init.conf", ASCII_TEXT),
> -            # check loaded modules
> -            (
> -                "lsmod | grep -e '^lockd\\|^ip_tables\\|^wireguard' | wc -l",
> -                "3",
> -            ),
> -            # sha256sum check modul
> -            (
> -                "sha256sum </etc/modules-load.d/50-cloud-init.conf",
> -                "9d14d5d585dd3e5e9a3c414b5b7af7ed"
> -                "9d44e7ee3584652fbf388cad455b5053",
> -            ),
> -            # sha256sum check modprobe
> -            (
> -                "sha256sum   </etc/modprobe.d/50-cloud-init.conf",
> -                "229ccc941ec34fc8c49bf14285ffeb65"
> -                "ea2796c4840f9377d6df76bda42c878e",
> -            ),
> -        ),
> -    )
> -    def test_kernel_modules(
> -        self, cmd, expected_out, class_client: IntegrationInstance
> -    ):
> -        result = class_client.execute(cmd)
> -        assert result.ok
> -        assert expected_out in result.stdout
> -
> -    def test_clean_log(self, class_client: IntegrationInstance):
> -        log = class_client.read_from_file("/var/log/cloud-init.log")
> -        verify_clean_log(log)
> +class TestKernelModulesWithoutKmod(BaseTest):
> +    pass
> ```

@aciba90 @s-makin 
Thank you for reviewing. Implemented the changes you requested, please review again.

s-makin (approved):
Looks good!

aciba90 (approved):
LGTM. Thanks for the work here, @chifac08.

I leave it open, just in case other maintainers want to have a look.

chifac08:
@TheRealFalcon I would be delighted if you could review my PR. Thank you!

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+
+    This function appends a kernel module to a file
+    (depends on operating system) for loading a module on boot.
+
+    @param module_name: string.
+
+    @raises: RuntimeError when write operation fails.
+    """
+
+    LOG.info(
```
blackboxsw, line 175: We should only use info level logs for something we really want users to see and react to. I don't think we need this log as `util.write_file` below should tell us `Writing %s with X bytes` anyway.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+        except Exception as e:
+            raise RuntimeError(
+                f"Could not delete file {file_path}:{NL}{str(e)}"
+            ) from e
+
+
+def reload_modules(cloud: Cloud):
+    """Reload kernel modules
+
+    This function reloads modules in /etc/modules-load.d/cloud-init.conf
```
blackboxsw, line 259:
```suggestion
    This function reloads modules in /etc/modules-load.d/50-cloud-init.conf
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+LOG = logging.getLogger(__name__)
+NL = "\n"
+REQUIRED_KERNEL_MODULES_KEYS = frozenset(["name"])
+
+DEFAULT_CONFIG: dict = {
+    "km_cmd": {
+        "update": ["update-initramfs", "-u", "-k", "all"],
+        "unload": ["rmmod"],
+        "is_loaded": ["lsmod"],
+    },
```
blackboxsw, line 67: These commands are distribution-specific.  FreeBSD will probably have to use something like `kldload` NetBSD `modload`etc.

Maybe this is too early to think about distro-specific implementations...

What do think about hanging them this config off of `cloudinint.distro.Distro` as class attrs and methods:
 -  `kernel_module_cmds`  = {'load': ['"systemctl", "restart", "systemd-modules-load"], 'update' ... 'unload'.. 'is_loaded', 'persist'...}
 -   `kernel_module_fns` = {'load': ..., 'persist': ...}
 -  `def manage_kernel_modules`:  takes params `action: str, module_name: Optional[str]` if self.kernel_module_cmds.get(action) is None raise NotImplementedError. module_name could be optional for "load" action to  
 -  *BSD classes could set kernel_module_cmds = {} to ensure NotImplementedError is raised if called

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+
+    LOG.info(
+        "Writing %s file for loading kernel modules on boot",
+        DEFAULT_CONFIG["km_files"]["load"].get("path"),
+    )
+
+    try:
+        LOG.debug("Appending kernel module %s", module_name)
+        util.write_file(
+            DEFAULT_CONFIG["km_files"]["load"]["path"],
```
blackboxsw, line 183: In theory we could pull this attribute off of the distro class so that other distros could override that behavior appropriately if available.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+    # check systemd usage
+    if not cloud.distro.uses_systemd():
+        LOG.debug(
+            "Skipping module named %s, due to " "no systemd installed",
+            name,
+        )
+        return
+
+    # cleanup
+    LOG.info("Cleaning up kernel modules")
```
blackboxsw, line 368: This should be LOG.debug and the log logic should live inside cleanup() function

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+
+    if not kernel_modules_section:
+        LOG.debug(
+            "Skipping module named %s, no "
+            "'kernel_modules' configuration found",
+            name,
+        )
+        return
+
+    LOG.debug("Found kernel_modules section in config")
```
blackboxsw, line 357: Probably can redact this log, cloud-init logs are really noisy as it is, also we know we've found kernel_module section if we have any logs, command execution or file writes happening between the `start: modules-final/kernel-modules and `finish: modules-final/kernel-modules` bookend logs

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+            prepare_module(module["name"])
+        else:
+            unload_modules.append(module["name"])
+
+        # Enhance module
+        if module.get("persist"):
+            enhance_module(module["name"], module["persist"], unload_modules)
+
+    # rebuild initial ramdisk
+    log.info("Update initramfs")
```
blackboxsw, line 387: let's make this `LOG.debug` instead of `log.info` and bring that inside `update_initial_ramdisk` function.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,396 @@
...
+    update_initial_ramdisk()
+
+    # Unload modules (blacklisted or 'load' is false)
+    log.info("Unloading kernel modules")
+    unload(unload_modules)
```
blackboxsw, line 392: Same `LOG.debug` and pull it into `unload` and sam with the Reloading kernel modules into reload_modules. or `Distro.manage_kernel_modules(action="load")` if we agree it's worth pulling this into the Distro class.

blackboxsw (commented):
Really good PR here @chifac08 thanks for the continued contributions to cloud-init. I still need a thorough review of schema validation and logic throughout but I had a 'big' question about whether we should move some of this into the Distro class given that we know BSD has separate tooling for managing kernel modules (and parts of this could be extended for BSD in the future).  I'll continue this review in earnest tomorrow but wanted to pass the questions and nits folks way first so we can discuss.

cloudinit/config/cc_kernel_modules.py
chifac08: Agreed.

cloudinit/config/cc_kernel_modules.py
chifac08: To be honest, I already had such an implementation in mind and tried to implement such functionality the early stages of development for this module. After some failing unit tests I overruled my plan as my main intention was to develop this module for Debian based systems. Nevertheless, I really like the idea of enhancing this module to become functional for other distros too. I am going to implement it. Stay tuned :)

chifac08: > These commands are distribution-specific. FreeBSD will probably have to use something like `kldload` NetBSD `modload`etc.
> 
> Maybe this is too early to think about distro-specific implementations...
> 
> What do think about hanging them this config off of `cloudinint.distro.Distro` as class attrs and methods:
> 
> * `kernel_module_cmds`  = {'load': ['"systemctl", "restart", "systemd-modules-load"], 'update' ... 'unload'.. 'is_loaded', 'persist'...}
> * `kernel_module_fns` = {'load': ..., 'persist': ...}
> * `def manage_kernel_modules`:  takes params `action: str, module_name: Optional[str]` if self.kernel_module_cmds.get(action) is None raise NotImplementedError. module_name could be optional for "load" action to
> * *BSD classes could set kernel_module_cmds = {} to ensure NotImplementedError is raised if called

Implemented improvements you requested. Concerning the distro-specific implementation I would be delighted if you could help me to implement it. I already tried creating the above methods and class attrs you mentioned but I can't get it to work. Maybe you could give me an more detailed and deep dive example? thx For example: what value does belong to "load" key from `kernel_module_fns` dict?

TheRealFalcon:
@blackboxsw and @chifac08 , I have a feeling you're both waiting for the other. @chifac08 , since your last push, is this ready for re-review?

chifac08:
@TheRealFalcon @blackboxsw 
Current implementation is ready to re-review as i only implemented addressed changes. As already mentioned above, i would need some help if we want to make this module usable for other distros as i have no idea how to implement it.

thank you for your help

cloudinit/config/cc_kernel_modules.py
blackboxsw: @chifac08 sorry I missed the response here. I'm working on crafting the distro class specific extensions for both manage_kernel_modules. update_initramfs  and persist_module file paths. I've attached a patch suggestion this is work in progress so we can discuss further. I'll update this comment tomorrow with a more complete suggestion

The diff below is currently missing the following, but I will update this comment/diff with minimally config file specialization for distro classes.:
 * updated unittests to mock distro.manage_kernel_modules
 * added config file specialization for persisting module loads across rebootfor debian vs NotImplemented for non-debian-based systems
 * established new unit tests for the distro-class extensions 

```diff
diff --git a/cloudinit/config/cc_kernel_modules.py b/cloudinit/config/cc_kernel_modules.py
index 8eee08da1..07c5f285a 100644
--- a/cloudinit/config/cc_kernel_modules.py
+++ b/cloudinit/config/cc_kernel_modules.py
@@ -257,7 +257,7 @@ def reload_modules(cloud: Cloud):
 
     @raises RuntimeError
     """
-
+    LOG.debug("Reloading kernel modules")
     try:
         out = cloud.distro.manage_service("restart", "systemd-modules-load")
         if re.search("Failed", out.stdout.strip()):
@@ -268,32 +268,7 @@ def reload_modules(cloud: Cloud):
         ) from e
 
 
-def is_loaded(module_name: str) -> bool:
-    """
-    Checks if a kernel module is already loaded
-
-    @param module_name: name of kernel module
-
-    @return bool
-    """
-    loaded = False
-    try:
-        out = subp.subp(
-            DEFAULT_CONFIG["km_cmd"]["is_loaded"], capture=True, shell=True
-        )
-        if re.search("^" + module_name, out.stdout.strip()):
-            LOG.debug("Kernel module %s is loaded", module_name)
-            loaded = True
-    except subp.ProcessExecutionError as e:
-        util.logexc(
-            LOG,
-            f"Could not determine status of module {module_name}:{NL}{str(e)}",
-        )
-
-    return loaded
-
-
-def unload(modules: list):
+def unload(cloud: Cloud, modules: list):
     """Unloads a list of kernel module
 
     This function unloads a list of kernel modules.
@@ -305,28 +280,34 @@ def unload(modules: list):
     """
 
     cmd = copy.copy(DEFAULT_CONFIG["km_cmd"]["unload"])
+    loaded_modules = cloud.distro.manage_kernel_modules("list")
     for module in set(modules):
         cmd.append(module)
-        try:
-            if is_loaded(module):
-                LOG.debug("Unloading kernel module %s", module)
-                subp.subp(cmd)
-        except subp.ProcessExecutionError as e:
-            raise RuntimeError(
-                f"Could not unload kernel module {module}:{NL}{str(e)}"
-            ) from e
+        if re.search("^" + module_name + "$", loaded_modules):
+            try:
+                cloud.distro.manage_kernel_modules("unload", module)
+            except subp.ProcessExecutionError as e:
+                raise RuntimeError(
+                    f"Could not unload kernel module {module}:{NL}{str(e)}"
+                ) from e
 
 
-def update_initial_ramdisk():
+def update_initial_ramdisk(cloud: Cloud):
     """Update initramfs for all installed kernels
 
-    @param distro_cfg: Distro config
+    :param distro_cfg: Distro config
 
-    @raises RuntimeError
+    :raises: RuntimeError on command failure to update initramfs
+             NotImplementedError on lack of distro support
     """
-    LOG.debug("Update initramfs")
+    if not cloud.distro.update_initramfs_cmd:
+        raise NotImplementedError(
+            "Unable to update initramfs for %s. Kernel module changes will not"
+            " persist across reboot",
+            cloud.distro.name,
+        )
     try:
-        subp.subp(DEFAULT_CONFIG["km_cmd"]["update"])
+        subp.subp(cloud.distro.update_initramfs_cmd)
     except subp.ProcessExecutionError as e:
         raise RuntimeError(
             f"Failed to update initial ramdisk:{NL}{str(e)}"
@@ -378,12 +359,10 @@ def handle(
             enhance_module(module["name"], module["persist"], unload_modules)
 
     # rebuild initial ramdisk
-    update_initial_ramdisk()
+    update_initial_ramdisk(cloud)
 
     # Unload modules (blacklisted or 'load' is false)
-    log.info("Unloading kernel modules")
-    unload(unload_modules)
+    unload(cloud, unload_modules)
 
     # reload kernel modules
-    log.info("Reloading kernel modules")
     reload_modules(cloud)
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index 735a78327..23a9104f0 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -16,7 +16,7 @@ import stat
 import string
 import urllib.parse
 from io import StringIO
-from typing import Any, Mapping, MutableMapping, Optional, Type
+from typing import Any, List, Mapping, MutableMapping, Optional, Type
 
 from cloudinit import importer
 from cloudinit import log as logging
@@ -88,7 +88,14 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     hostname_conf_fn = "/etc/hostname"
     tz_zone_dir = "/usr/share/zoneinfo"
     default_owner = "root:root"
-    init_cmd = ["service"]  # systemctl, service etc
+    init_cmd: List[str] = ["service"]  # systemctl, service etc
+
+    kernel_module_cmd_map: Mapping[str, str] = {
+        "list": ["lsmod"],
+        "load": ["insmod"],
+        "unload": ["rmmod"],
+    }
+    update_initramfs_cmd: List[str] = ["update-initramfs", "-u", "-k", "all"]
     renderer_configs: Mapping[str, MutableMapping[str, Any]] = {}
     _preferred_ntp_clients = None
     networking_cls: Type[Networking] = LinuxNetworking
@@ -907,6 +914,35 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             args.append(message)
         return args
 
+    def manage_kernel_module(
+        self, action: str, kernel_module: Optional[str] = None
+    ):
+        """
+        Perform the requested load, unload, persist for a kernel module.
+
+        :param action: String of operation to perform, one of the following:
+            - load: load the kernel module in the running environment
+            - unload: remove kernel module from the running environment
+            - list: List all loaded kernel modules
+
+        :raises: NotImplementedError on distros without specific support
+                 ProcessExecutionError on failure on non-zero exit for command
+        """
+        if action not in self.kernel_module_cmd_map:
+            raise NotImplementedError(
+                "Unable to %s kernel module %s on %s. Not implemented.",
+                action,
+                kernel_module,
+                distro,
+            )
+        cmd = self.kernel_module_cmd_map[action]
+        if action != "list":
+            cmd += kernel_module
+            LOG.debug(
+                "%sing kernel module %s", action, kernel_module.capitalize()
+            )
+        return subp.subp(cmd, capture=True)
+
     def manage_service(self, action: str, service: str):
         """
         Perform the requested action on a service. This handles the common
diff --git a/cloudinit/distros/alpine.py b/cloudinit/distros/alpine.py
index 4a23fe079..4cb85b17f 100644
--- a/cloudinit/distros/alpine.py
+++ b/cloudinit/distros/alpine.py
@@ -24,6 +24,7 @@ class Distro(distros.Distro):
     pip_package_name = "py3-pip"
     locale_conf_fn = "/etc/profile.d/locale.sh"
     network_conf_fn = "/etc/network/interfaces"
+    update_initramfs_cmd = []  # TODO(define initramfs support)
     renderer_configs = {
         "eni": {"eni_path": network_conf_fn, "eni_header": NETWORK_FILE_HEADER}
     }
diff --git a/cloudinit/distros/arch.py b/cloudinit/distros/arch.py
index 2d5cfbf65..98ba1d74e 100644
--- a/cloudinit/distros/arch.py
+++ b/cloudinit/distros/arch.py
@@ -22,6 +22,7 @@ class Distro(distros.Distro):
     locale_gen_fn = "/etc/locale.gen"
     network_conf_dir = "/etc/netctl"
     init_cmd = ["systemctl"]  # init scripts
+    update_initramfs_cmd = []  # TODO(define mkinicpio support)
     renderer_configs = {
         "netplan": {
             "netplan_path": "/etc/netplan/50-cloud-init.yaml",
diff --git a/cloudinit/distros/bsd.py b/cloudinit/distros/bsd.py
index c4e1e15c6..d0188d18a 100644
--- a/cloudinit/distros/bsd.py
+++ b/cloudinit/distros/bsd.py
@@ -20,6 +20,9 @@ class BSD(distros.Distro):
     # poweroff.
     shutdown_options_map = {"halt": "-H", "poweroff": "-p", "reboot": "-r"}
 
+    kernel_module_cmd_map = {}  # TODO(define modunload/modload/modstat)
+    update_initramfs_cmd = []  # TODO(define initramfs support)
+
     # Set in BSD distro subclasses
     group_add_cmd_prefix: List[str] = []
     pkg_cmd_install_prefix: List[str] = []
diff --git a/cloudinit/distros/gentoo.py b/cloudinit/distros/gentoo.py
index 37217fe43..6b502fae8 100644
--- a/cloudinit/distros/gentoo.py
+++ b/cloudinit/distros/gentoo.py
@@ -22,6 +22,12 @@ class Distro(distros.Distro):
     network_conf_fn = "/etc/conf.d/net"
     hostname_conf_fn = "/etc/conf.d/hostname"
     init_cmd = ["rc-service"]  # init scripts
+    kernel_module_cmd_map = {
+        "is_loaded": ["lsmod"],
+        "load": ["modprobe"],
+        "unload": ["modprobe", "-r"],
+    }
+    update_initramfs = ["update-initramfs", "-u", "-k", "all"]
     default_locale = "en_US.UTF-8"
 
     # C.UTF8 makes sense to generate, but is not selected
diff --git a/cloudinit/distros/opensuse.py b/cloudinit/distros/opensuse.py
index 00ed15142..d4f21a460 100644
--- a/cloudinit/distros/opensuse.py
+++ b/cloudinit/distros/opensuse.py
@@ -25,6 +25,7 @@ class Distro(distros.Distro):
     systemd_hostname_conf_fn = "/etc/hostname"
     systemd_locale_conf_fn = "/etc/locale.conf"
     tz_local_fn = "/etc/localtime"
+    update_initramfs = []  # TODO(dracut support to regen initramfs)
     renderer_configs = {
         "sysconfig": {
             "control": "etc/sysconfig/network/config",
diff --git a/cloudinit/distros/photon.py b/cloudinit/distros/photon.py
index 14cefe90f..9dd6b7116 100644
--- a/cloudinit/distros/photon.py
+++ b/cloudinit/distros/photon.py
@@ -19,6 +19,10 @@ class Distro(distros.Distro):
     network_conf_dir = "/etc/systemd/network/"
     systemd_locale_conf_fn = "/etc/locale.conf"
     resolve_conf_fn = "/etc/systemd/resolved.conf"
+    kernel_module_cmd_map = {
+        "list": ["lsmod"],  # TODO(sort load and unload for kernel modules)
+    }
+    update_initramfs_cmd = []
 
     renderer_configs = {
         "networkd": {
diff --git a/cloudinit/distros/rhel.py b/cloudinit/distros/rhel.py
index 320f4ba1d..c52b48d52 100644
--- a/cloudinit/distros/rhel.py
+++ b/cloudinit/distros/rhel.py
@@ -36,6 +36,8 @@ class Distro(distros.Distro):
     network_script_tpl = "/etc/sysconfig/network-scripts/ifcfg-%s"
     tz_local_fn = "/etc/localtime"
     usr_lib_exec = "/usr/libexec"
+
+    update_initramfs_cmd = []  # TODO(dracut support to regen initramfs)
     renderer_configs = {
         "sysconfig": {
             "control": "etc/sysconfig/network",
diff --git a/tests/unittests/config/test_cc_kernel_modules.py b/tests/unittests/config/test_cc_kernel_modules.py
index 62a87e921..4bbf54064 100644
--- a/tests/unittests/config/test_cc_kernel_modules.py
+++ b/tests/unittests/config/test_cc_kernel_modules.py
@@ -124,8 +124,10 @@ class TestKernelModules(CiTestCase):
             "update-initramfs: execution error"
         )
 
+        distro = mock.MagicMock()  # No errors raised
+        mycloud = FakeCloud(distro)
         with self.assertRaises(RuntimeError) as context_mgr:
-            cc_kernel_modules.update_initial_ramdisk()
+            cc_kernel_modules.update_initial_ramdisk(mycloud)
         self.assertIn(
             "Failed to update initial ramdisk:\n"
             "Unexpected error while running command.\n"
```

igalic:
BSDs don't have a concept of an initramfs, except as a special kernel that runs before the real system to do tasks like growfs on OpenBSD and NetBSD. The only thing that comes close is regenerating the hints files on FreeBSD using [`kldxref(8)`]( https://man.freebsd.org/kldxref(8)) (or running the service) after installing new kernels (not really a thing, unless you're using PkgBase, or just rebuilt the system, or are sharing a build between machine), or new kernel modules (mostly useful for desktop systems)

OpenBSD got rid of loadable kernel modules about ten years ago

chifac08:
@blackboxsw thank you for your help and detailed explanation. I got an good overview about Distro classes and how to implement such a feature next time :). Would you mind applying your diff? Otherwise i could do it. Pls let me know.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+
+    for (key, value) in persist.items():
+        LOG.debug(
+            "Enhancing kernel module %s with %s:%s", module_name, key, value
+        )
+        entry = f"{key} {module_name} {value}"
+        # softdep special case
+        if key == "softdep":
+            entry += " pre:".join({key["pre"]}) + " post: ".join({key["post"]})
```
blackboxsw, line 213: I think we need happy path/success testing here in unittests validatng expected output for the softdep section as I don't think this does what you are hoping here. I expect this code to fall over. I think we need the join to be a " ".join(key["pre"]) right? Same for post.

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,370 @@
...
+    "name": "Kernel Modules",
+    "title": "Manage and enhance kernel modules",
+    "description": MODULE_DESCRIPTION,
+    "distros": DISTROS,
+    "frequency": PER_INSTANCE,
+    "activate_by_schema_keys": ["kernel_modules"],
+    "examples": [
+        dedent(
+            """\
+    kernel_modules:
```
blackboxsw, line 40: Please add other viable examples using various persist keys beyond `options` to show folks how to use such config keys,

We also need happy path unittest coverage of each key to make sure it's rendered properly to config files.

blackboxsw (changes requested):
Thanks @chifac08.
I have pushed my suggested changes shuffling implementation into distros. I hink we need a bit more unittest coverage of cc_kernel_modules happy path config writing out appropriate formats to modprobe.d conf files.

I probably also should have added tests/unittests/distros/test_manage_kernel_modules.py to cover the new distro  method. I wanted to push what I had so potentially you could work on example extension and unittest coverage for cc_kernel_modules. If needed I can also add supplemental unittest coverage for distros.manage_kernel_module

chifac08:
Hi maintainers,

Would you mind re-opening this PR? I really want to finish my work here and make sure it make it's way in the upstream. 

thank your for time!

cloudinit/config/cc_kernel_modules.py
chifac08: Agreed and implemented. I would be delighted if you could have look at failing unittests. It states that schema validation is failing...

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+
+    for (key, value) in persist.items():
+        LOG.debug(
+            "Enhancing kernel module %s with %s:%s", module_name, key, value
+        )
+        entry = f"{key} {module_name} {value}"
+        # softdep special case
+        if key == "softdep":
+            entry += " ".join({key["pre"]}) + " ".join({key["post"]})
```
blackboxsw, line 219: This entry doesn't appear to be constructed in the format expected by modprobe.d.
Shouldn't it be of the format: `softdep wireguard pre: <all_pre_module_name> post: <all_post_module_names>` ?

The way this entry is appended , I don't see either `pre:` or `post:` definitions in the output.



you have done a fair job establishing failure conditions and error handling in the tests. But, I think we need  to see a happy path test exercising each function and asserting the correct output is being written to the right DEFAULT_CONFIG...<path>.

blackboxsw (changes requested):
Thank you @chifac08 for resurrecting this PR and giving it attention again.


I think the biggest issues is that we need to address are:
- a git rebase against tip of main because there is significant drift in config/cloud.config.tmpl that will likely produce merge conflicts with your PR due to the branches original creation date.
-  specific unittest coverage for correct working behavior and checking the contents of the util.write_file called to assert the correct full content expected in the modprobe.d or modules-load.d file for the functions  prepare_module, enhance_module, cleanup, reload_modules, unload  and update_initial_ramdisk before this PR is ready to land.
- integration tests need updating because the pytest.marks are no longer available


Here's a minimal diff to get existing unittests 'passing' and dropping the `array` import
```
--- a/cloudinit/config/cc_kernel_modules.py
+++ b/cloudinit/config/cc_kernel_modules.py
@@ -3,7 +3,6 @@
 """Kernel Modules"""
 import logging
 import re
-from array import array
 from textwrap import dedent
 from typing import List
 
@@ -50,7 +49,7 @@ meta: MetaSchema = {
         load: true
         persist:
           softdep:
-            pre: ["nf_conntrack" "nf_tables"]
+            pre: ["nf_conntrack", "nf_tables"]
     """
         ),
     ],
@@ -113,9 +112,9 @@ def persist_schema_validation(persist: dict) -> List[str]:
                         " Should be one of: pre, post"
                     )
                 else:
-                    if not isinstance(sdvalues, array):
+                    if not isinstance(sdvalues, list):
                         errors.append(
-                            "Expected an array for"
+                            "Expected a list for"
                             f" kernel_modules:persist:softdep:{sdkey}."
                             f" Found {sdvalues}."
                         )
index d1cf61885..13aae7ada 100644
@@ -3895,6 +3895,7 @@
     "grub_dpkg": {},
     "hostname": {},
     "keyboard": {},
+    "kernel_modules": {},
     "landscape": {},
     "launch-index": {},
     "locale": {},
diff --git a/tests/integration_tests/modules/test_kernel_modules.py b/tests/integration_tests/modules/test_kernel_modules.py
index eadaa8a95..a59225d38 100644
--- a/tests/integration_tests/modules/test_kernel_modules.py
+++ b/tests/integration_tests/modules/test_kernel_modules.py
@@ -4,6 +4,8 @@ from pycloudlib.lxd.instance import LXDInstance
 
 from cloudinit.subp import subp
 from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.integration_settings import PLATFORM
+from tests.integration_tests.releases import IS_UBUNTU
 from tests.integration_tests.util import verify_clean_log
 
 ASCII_TEXT = "ASCII text"
@@ -28,7 +30,7 @@ kernel_modules:
     load: true
     persist:
       softdep:
-        pre: ["nf_conntrack" "nf_tables"]
+        pre: ["nf_conntrack", "nf_tables"]
 """
 
 KERNEL_MODULES_LXD = "lockd,zfs,wireguard"
@@ -43,8 +45,10 @@ def load_kernel_modules_lxd(instance: LXDInstance):
 
 
 @pytest.mark.ci
+@pytest.mark.skipif(
+    not IS_UBUNTU, reason="kernel_module support currently limited to Ubuntu"
+)
 @pytest.mark.user_data(USER_DATA)
-@pytest.mark.ubuntu
 class BaseTest:
     @pytest.mark.parametrize(
         "cmd,expected_out",
@@ -101,12 +105,18 @@ class BaseTest:
         verify_clean_log(log)
 
 
-@pytest.mark.no_container
+@pytest.mark.skipif(
+    PLATFORM != "lxd_vm",
+    reason="Test requires a VM to load additional kernel modules in LXD",
+)
 class TestKernelModules(BaseTest):
     pass
 
 
-@pytest.mark.lxd_container
+@pytest.mark.skipif(
+    PLATFORM != "lxd_container",
+    reason="Uses lxd command-line to set kernel-module config prior to launch",
+)
 @pytest.mark.lxd_setup.with_args(load_kernel_modules_lxd)
 class TestKernelModulesWithoutKmod(BaseTest):
     pass
diff --git a/tests/unittests/config/test_cc_kernel_modules.py b/tests/unittests/config/test_cc_kernel_modules.py
index c040788c7..dea33bab4 100644
--- a/tests/unittests/config/test_cc_kernel_modules.py
+++ b/tests/unittests/config/test_cc_kernel_modules.py
@@ -48,7 +48,7 @@ class TestKernelModules:
             "Expected a string for kernel_modules:persist:install. Found 123.",
             "Expected a boolean for kernel_modules:persist:blacklist. "
             "Found no-bool.",
-            "Expected an array for kernel_modules:persist:softdep:pre. "
+            "Expected a list for kernel_modules:persist:softdep:pre. "
             "Found not-an-array.",
         ]
         with pytest.raises(ValueError) as context_mgr:
```

cloudinit/config/cc_kernel_modules.py
```diff
@@ -0,0 +1,373 @@
...
+        persist:
+          options: "devices=1 video_nr=20 card_label=fakecam exclusive_caps=1"
+      - name: zfs
+        persist:
+          blacklist: true
+      - name: btrfs
+        load: true
+        persist:
+          softdep:
+            pre: ["nf_conntrack" "nf_tables"]
```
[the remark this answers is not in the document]
chifac08: Fixed

config/cloud.cfg.tmpl
```diff
@@ -187,11 +187,13 @@ cloud_config_modules:
...
+ - package-update-upgrade-install
+ - fan
+ - landscape
+ - lxd
+ - kernel_modules
 {% if variant in ["debian", "ubuntu", "unknown"] %}
-  - fan
-  - landscape
-  - lxd
+  - package_update_upgrade_install
```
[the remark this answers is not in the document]
chifac08: Successfully rebased on current main :)

blackboxsw:
Sorry for delay here, while to updates look reasonable. I don't want to land this in tip of main until after our 24.1 release is cut to avoid risk on this upcoming release. Expectation is we should be cutting 24.1 this week. We can finish this review and merge after that occurs.

PR #1757, opened 2022-09-27
Fix Oracle DS not setting default route when using IMDS

TheRealFalcon:
## Proposed Commit Message

```
Fix Oracle DS primary interface when using IMDS

If /run/net* files aren't available, we use IMDS for configuring the
default interface. Rather than attempt a static configuration, grab
the IP and let DHCP do the rest (as it does when /run/net*
is available).

LP: #1989686
```

## Additional Context
https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1989686

## Test Steps
On my instance, `curl --header "Authorization: Bearer Oracle" http://169.254.169.254/opc/v2/vnics` gives:
```
[
  {
    "macAddr": "02:00:17:08:BB:E1",
    "privateIp": "10.0.2.60",
    "subnetCidrBlock": "10.0.2.0/24",
    "virtualRouterIp": "10.0.2.1",
    "vlanTag": 1320,
    "vnicId": "ocid1.vnic.oc1.phx.abyhqljrmna323fjjscndxh2xfa3rkhhrgdwelqhdl6jumhqr3arm44rhuyq"
  }
]
```

Generated deb and installed on Oracle. Moved `/run/net-ens3.conf` to `/run/net-ens3.conf.bak`. Ran `cloud-init clean --logs; cloud-init init --local`. This gives the following netplan config at `/etc/netplan/50-cloud-init.yaml`:
```
network:
    version: 2
    ethernets:
        ens3:
            addresses:
            - 10.0.2.60/24
            match:
                macaddress: 02:00:17:08:bb:e1
            mtu: 9000
            routes:
            -   to: 10.0.2.0/24
                via: 10.0.2.1
            set-name: ens3
```

This was generated from the v1 config. Not sure a great way to show the generation from v2, but it should be fairly straightforward in the code.

Also, not sure a great way to test the generation of routes because even after netplan apply, the routes that the system booted with are still there.

`ip r` before netplan apply:
```
default via 10.0.2.1 dev ens3 proto dhcp src 10.0.2.60 metric 100 
10.0.2.0/24 dev ens3 proto kernel scope link src 10.0.2.60 metric 100 
10.0.2.1 dev ens3 proto dhcp scope link src 10.0.2.60 metric 100 
169.254.169.254 via 10.0.2.1 dev ens3 proto dhcp src 10.0.2.60 metric 100 
```

`ip r` after netplan apply:
```
default via 10.0.2.1 dev ens3 proto dhcp src 10.0.2.60 metric 100 
10.0.2.0/24 via 10.0.2.1 dev ens3 proto static 
10.0.2.0/24 dev ens3 proto kernel scope link src 10.0.2.60 metric 100 
10.0.2.1 dev ens3 proto dhcp scope link src 10.0.2.60 metric 100 
169.254.169.254 via 10.0.2.1 dev ens3 proto dhcp src 10.0.2.60 metric 100 
```

# Other considerations
The default configuration (from /run/net-ens3.conf) is:
```
root@instance20220927160159:/run# cat net-ens3.conf.bak 
DEVICE='ens3'
PROTO='dhcp'
IPV4PROTO='dhcp'
IPV4ADDR='10.0.2.60'
IPV4NETMASK='255.255.255.0'
IPV4BROADCAST='10.0.2.255'
IPV4GATEWAY='10.0.2.1'
IPV4DNS0='169.254.169.254'
ROOTSERVER='10.0.2.1'
HOSTNAME=''
DNSDOMAIN='default.oraclevcn.com'
DOMAINSEARCH='default.oraclevcn.com'
```
The 'dhcp' option makes the resulting netplan config much simpler:
```
network:
    version: 2
    ethernets:
        ens3:
            dhcp4: true
            match:
                macaddress: 02:00:17:08:bb:e1
            set-name: ens3
```
Such a 'dhcp' key isn't present when querying IMDS, but I do wonder if things would be less error-prone if we relied on DHCP for the primary interface. Things "sorta" worked ok back when we relied on the ephemeral configuration to be the configuration for the primary interface. The problem came when we chose the wrong interface as primary, but if we could match the primary mac  and then use DHCP for only that interface, we shouldn't have that problem nor any of the problems we're having with static config.

### DIFF
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -307,7 +307,7 @@ class DataSourceOracle(sources.DataSource):
                     "routes": [
                         {
                             "network": str(network.network_address),
-                            "netmask": network.netmask,
+                            "netmask": str(network.netmask),
                             "gateway": gateway,
                         },
                     ],
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -297,14 +297,20 @@ class DataSourceOracle(sources.DataSource):
                 )
                 continue
             name = interfaces_by_mac[mac_address]
-            prefix = ipaddress.ip_network(
-                vnic_dict["subnetCidrBlock"]
-            ).prefixlen
+            network = ipaddress.ip_network(vnic_dict["subnetCidrBlock"])
+            gateway = vnic_dict["virtualRouterIp"]
 
             if self._network_config["version"] == 1:
                 subnet = {
                     "type": "static",
-                    "address": f"{vnic_dict['privateIp']}/{prefix}",
+                    "address": f"{vnic_dict['privateIp']}/{network.prefixlen}",
+                    "routes": [
+                        {
+                            "network": str(network.network_address),
+                            "prefix": network.prefixlen,
+                            "gateway": gateway,
+                        },
+                    ],
                 }
                 self._network_config["config"].append(
                     {
@@ -317,11 +323,19 @@ class DataSourceOracle(sources.DataSource):
                 )
             elif self._network_config["version"] == 2:
                 self._network_config["ethernets"][name] = {
-                    "addresses": [f"{vnic_dict['privateIp']}/{prefix}"],
+                    "addresses": [
+                        f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                    ],
                     "mtu": MTU,
                     "dhcp4": False,
                     "dhcp6": False,
                     "match": {"macaddress": mac_address},
+                    "routes": [
+                        {
+                            "to": str(network.network_address),
+                            "via": gateway,
+                        },
+                    ],
                 }
 
 
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -306,7 +306,7 @@ class DataSourceOracle(sources.DataSource):
                     "address": f"{vnic_dict['privateIp']}/{network.prefixlen}",
                     "routes": [
                         {
-                            "network": str(network.network_address),
+                            "network": "0.0.0.0/0",
                             "netmask": str(network.netmask),
                             "gateway": gateway,
                         },
@@ -332,7 +332,7 @@ class DataSourceOracle(sources.DataSource):
                     "match": {"macaddress": mac_address},
                     "routes": [
                         {
-                            "to": str(network.network_address),
+                            "to": "default",
                             "via": gateway,
                         },
                     ],
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -301,36 +301,36 @@ class DataSourceOracle(sources.DataSource):
             network = ipaddress.ip_network(vnic_dict["subnetCidrBlock"])
 
             if self._network_config["version"] == 1:
+                if is_primary:
+                    subnet = {"type": "dhcp"}
+                else:
+                    subnet = {
+                        "type": "static",
+                        "address": (
+                            f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                        ),
+                    }
                 interface_config = {
                     "name": name,
                     "type": "physical",
                     "mac_address": mac_address,
                     "mtu": MTU,
-                    "subnets": [
-                        {
-                            "address": (
-                                f"{vnic_dict['privateIp']}/{network.prefixlen}"
-                            )
-                        }
-                    ],
+                    "subnets": [subnet],
                 }
-                # Primary interface uses DHCP as is our first entry
-                interface_config["subnets"][0]["type"] = (
-                    "dhcp" if is_primary else "static"
-                )
                 self._network_config["config"].append(interface_config)
             elif self._network_config["version"] == 2:
                 # Why does this elif exist???
                 # Are there plans to switch to v2?
                 interface_config = {
-                    "addresses": [
-                        f"{vnic_dict['privateIp']}/{network.prefixlen}"
-                    ],
                     "mtu": MTU,
                     "match": {"macaddress": mac_address},
                     "dhcp6": False,
                     "dhcp4": is_primary,
                 }
+                if not is_primary:
+                    interface_config["addresses"] = [
+                        f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                    ]
                 self._network_config["ethernets"][name] = interface_config
 
 
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -289,6 +289,7 @@ class DataSourceOracle(sources.DataSource):
         vnics_data = self._vnics_data if set_primary else self._vnics_data[1:]
 
         for index, vnic_dict in enumerate(vnics_data):
+            is_primary = set_primary and index == 0
             mac_address = vnic_dict["macAddr"].lower()
             if mac_address not in interfaces_by_mac:
                 LOG.warning(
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceOracle.py
```diff
@@ -317,11 +323,19 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
+                        f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                    ],
                     "mtu": MTU,
                     "dhcp4": False,
                     "dhcp6": False,
                     "match": {"macaddress": mac_address},
+                    "routes": [
+                        {
+                            "to": str(network.network_address),
+                            "via": gateway,
```
blackboxsw, line 336: Is it worth us being explicit here with a 'metric': 100 since this is a default route and we can increment v2 route metric definitions in the future if we have additional subordinate routes that should be de-prioritized ?

cloudinit/sources/DataSourceOracle.py
```diff
@@ -297,14 +297,20 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
             if self._network_config["version"] == 1:
                 subnet = {
                     "type": "static",
-                    "address": f"{vnic_dict['privateIp']}/{prefix}",
+                    "address": f"{vnic_dict['privateIp']}/{network.prefixlen}",
+                    "routes": [
+                        {
+                            "network": str(network.network_address),
+                            "prefix": network.prefixlen,
+                            "gateway": gateway,
```
blackboxsw, line 311: Same "metric': 100 question as below since this is the default route on the primary nic

cloudinit/sources/DataSourceOracle.py
```diff
@@ -297,14 +297,20 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
 
             if self._network_config["version"] == 1:
                 subnet = {
                     "type": "static",
-                    "address": f"{vnic_dict['privateIp']}/{prefix}",
+                    "address": f"{vnic_dict['privateIp']}/{network.prefixlen}",
+                    "routes": [
+                        {
+                            "network": str(network.network_address),
+                            "prefix": network.prefixlen,
```
blackboxsw, line 310: Is this prefix key valid for v1? From example docs it looks like  it should be netmask https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v1.html#subnet-ip

TheRealFalcon:
The code I had here previously wasn't allowing me to SSH. I pushed an update to set the default route, and that allows me to SSH. I still don't think this is correct though. I think we need subnet-specific routes per interface, and then a default route for the primary interface. I pushed what I have here so we can at least SSH to an instance.

TheRealFalcon:
> 1. Wouldn't that go in the opposite direction of [lp1967942](https://bugs.launchpad.net/cloud-init/+bug/1967942) ?

Not really. The major point there was to ensure we're using the correct primary nic. Using DHCP doesn't change that. We can create a netplan config to match in MAC, specify the IP address, then use DHCP for the rest (routes and DNS). Chad found that some of that information (i.e., DNS) isn't even provided by the IMDS.

> 2. The logic for translating IMDS to network config is the same for primary and secondary vnics. We would ease the primary vnic translation but, we would still have potential misconfigurations if the instance have secondary vnics, wouldn't we? If this is true, we would have to correctly support IMDS translation in any case.

We would still use IMDS for all of them. Primary nic would use DHCP along with the IMDS matching (as the /run/net* files currently do), while the secondary nics would use a static configuration based off IMDS.

> 3. By doing the dhcp "trick", couldn't we potentially lose IMDS network configuration details and leave the instance in a non-desired state (from the IMDS point of view)?

I don't know the implementation details of Oracle's cloud, but clouds generally use the same source of truth for DHCP and their IMDS. If DHCP is giving us wrong data, then the cloud is broken. 

> What about introducing a config flag like force_primary_dchp_discovery defaulting to false, which would keep the current behavior and if set to true we would apply the dhcp "trick"? In that way, we could have the best of both things.

I'm not sure what the use case for this would be. We just need to do the "right thing". Usage of this flag would probably mean we have a bug somewhere that we need to fix.

aciba90:
That makes totally sense. Thanks for the context there, @TheRealFalcon.

cloudinit/sources/DataSourceOracle.py
blackboxsw: We can take metric setup and auto-increment on secondary routes to a followup effort on oracle secondary NIC configuration. I think there are a couple of inconsistencies with how secondary nics are setup in cloud-init vs Oracle's documented https://docs.oracle.com/en-us/iaas/Content/Resources/Assets/secondary_vnic_all_configure.sh

cloudinit/sources/DataSourceOracle.py
blackboxsw: work item for future.

cloudinit/sources/DataSourceOracle.py
```diff
@@ -297,32 +298,40 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
+                    "mac_address": mac_address,
+                    "mtu": MTU,
+                    "subnets": [
+                        {
+                            "address": (
+                                f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                            )
+                        }
+                    ],
                 }
```
blackboxsw, line 316: This leads to us providing invalid network v1 on the primary nic which would have the following config:
```
subnets:
    address: <addr>/<prefix>
    type: dhcp
```
cloud-init currenty permits and silently ignores the address line on `type: dhcp` which results in no functional diff in the rendered netplan yaml, but eventually our schema validation for network config would warn/error on this.

What do you think about defining subnet in one place to avoid the supplemental mutation of interface_config down below so we have a cohesive view of primary vs not-primary device config?

```diff
diff --git a/cloudinit/sources/DataSourceOracle.py b/cloudinit/sources/DataSourceOracle.py
index ab51732f8..0b22baee2 100644
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -301,23 +301,23 @@ class DataSourceOracle(sources.DataSource):
             network = ipaddress.ip_network(vnic_dict["subnetCidrBlock"])
 
             if self._network_config["version"] == 1:
+                if is_primary:
+                    # Primary interface uses DHCP as is our first entry
+                    subnet = {"type": "dhcp"}
+                else:
+                    subnet = {
+                        "type": "static",
+                        "address": (
+                            f"{vnic_dict['privateIp']}/{network.prefixlen}"
+                        ),
+                    }
                 interface_config = {
                     "name": name,
                     "type": "physical",
                     "mac_address": mac_address,
                     "mtu": MTU,
-                    "subnets": [
-                        {
-                            "address": (
-                                f"{vnic_dict['privateIp']}/{network.prefixlen}"
-                            )
-                        }
-                    ],
+                    "subnets": [subnet],
                 }
-                # Primary interface uses DHCP as is our first entry
-                interface_config["subnets"][0]["type"] = (
-                    "dhcp" if is_primary else "static"
-                )
                 self._network_config["config"].append(interface_config)
             elif self._network_config["version"] == 2:
                 # Why does this elif exist???
diff --git a/tests/unittests/sources/test_oracle.py b/tests/unittests/sources/test_oracle.py
index 274a2e282..c2be3a952 100644
--- a/tests/unittests/sources/test_oracle.py
+++ b/tests/unittests/sources/test_oracle.py
@@ -307,7 +307,6 @@ class TestNetworkConfigFromOpcImds:
             assert "02:00:17:05:d1:db" == primary_cfg["mac_address"]
             assert 9000 == primary_cfg["mtu"]
             assert 1 == len(primary_cfg["subnets"])
-            assert "10.0.0.230/24" == primary_cfg["subnets"][0]["address"]
             assert "dhcp" == primary_cfg["subnets"][0]["type"]
         secondary_cfg = nic_cfg[secondary_nic_index]
         assert "ens4" == secondary_cfg["name"]
```

cloudinit/sources/DataSourceOracle.py
```diff
@@ -297,32 +298,40 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
+                # Primary interface uses DHCP as is our first entry
+                interface_config["subnets"][0]["type"] = (
+                    "dhcp" if is_primary else "static"
                 )
+                self._network_config["config"].append(interface_config)
             elif self._network_config["version"] == 2:
-                self._network_config["ethernets"][name] = {
-                    "addresses": [f"{vnic_dict['privateIp']}/{prefix}"],
+                # Why does this elif exist???
+                # Are there plans to switch to v2?
```
blackboxsw, line 324: Good question on the elif here. This is precisely the reason. e7881d5c62e68e4962e2fcd3d12089e5a3b94dc9

Generally, yes we want clouds to migrate towards v2 for better expression of complex networks/tags/metrics/policy etc. I think that is a long-term goal for any cloud IMDS to move toward v2 and eventually retire our v1 support.

It was only paving the way to move toward v2, some discussions were had a cloud-init summit to try to move this direction with ORacle's IMDS. It'd be nice to continue that work, but as it stands, I don't think any path can actually get to this currently.

blackboxsw (changes requested):
fix needed for network v1 on primary, which doesn't actually result in invalid netplan artifacts, but good to cleanup.

blackboxsw:
`The problem came when we chose the wrong interface as primary, but if we could match the primary mac and then use DHCP for only that interface, we shouldn't have that problem nor any of the problems we're having with static config.` 

Right we decided as a team that due to Oracle's `v2/vnics` IMDS being incomplete as far as DNS definitions, it forces cloud-init to still rely on DHCP for the primary interface in order to get the `  option domain-name-servers 169.254.169.254;` and `option domain-name "default.oraclevcn.com";`   DHCP options. All secondary NIC setup can be handled appropriately with static ip/routes. 


In testing a system with 2 VNICs attached (and the suggested diff above) I see the expectd netplan config defined by cloud-init which provides us with a properly working ip route table and address config.

```
# network: {config: disabled}
network:
    version: 2
    ethernets:
        ens3:
            dhcp4: true
            match:
                macaddress: 02:00:17:0b:a3:a6
            mtu: 9000
            set-name: ens3
        ens5:
            addresses:
            - 10.0.0.114/24
            match:
                macaddress: 02:00:17:0c:ee:28
            mtu: 9000
            set-name: ens5

# ip r
default via 10.0.0.1 dev ens3 proto dhcp src 10.0.0.4 metric 100 
10.0.0.0/24 dev ens5 proto kernel scope link src 10.0.0.114 
10.0.0.0/24 dev ens3 proto kernel scope link src 10.0.0.4 metric 100 
10.0.0.1 dev ens3 proto dhcp scope link src 10.0.0.4 metric 100 
169.254.169.254 via 10.0.0.1 dev ens3 proto dhcp src 10.0.0.4 metric 100 

# ip a
    inet6 ::1/128 scope host 
2: ens3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 02:00:17:0b:a3:a6 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet 10.0.0.4/24 metric 100 brd 10.0.0.255 scope global dynamic ens3
       valid_lft 84578sec preferred_lft 84578sec
    inet6 fe80::17ff:fe0b:a3a6/64 scope link 
3: ens5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 02:00:17:0c:ee:28 brd ff:ff:ff:ff:ff:ff
    altname enp0s5
    inet 10.0.0.114/24 brd 10.0.0.255 scope global ens5
    inet6 fe80::17ff:fe0c:ee28/64 scope link 

```

cloudinit/sources/DataSourceOracle.py
```diff
@@ -301,36 +301,36 @@ def _add_network_config_from_opc_imds(self, set_primary: bool = False):
...
                 interface_config = {
-                    "addresses": [
-                        f"{vnic_dict['privateIp']}/{network.prefixlen}"
-                    ],
                     "mtu": MTU,
                     "match": {"macaddress": mac_address},
                     "dhcp6": False,
                     "dhcp4": is_primary,
                 }
+                if not is_primary:
```
blackboxsw, line 330: @TheRealFalcon good catch and +1 on only setting addresses when we are non-dhcp. we could use the addresses declaration only if we have secondary 'static' IPs set on the device from `v2/vnics` in the future beyond what DHCP is already giving us.

blackboxsw (approved):
LGTM and good catch on v2 dropping addresses on primary NIC as well.

PR #1758, opened 2022-09-27
Distro manage service: Improve BSD support

igalic:
## Proposed Commit Message

```
Distro manage service: Improve BSD support

OpenBSD has wrapper called `rcctl` that does the same thing as 'service'.
FreeBSD's 'service' knows about 'enable', so we specialize that, too.
Add support and tests.
NetBSD's service does about as much as the base-case, so no need to do anything there.
To mark the occasion, we also add all other BSD `distros` to `OSFAMILIES`.

LP: #1990070
```

## Test Steps
- update cloud-init on *BSD
- test:
```python
>>> from cloudinit.distros.freebsd import Distro
>>> f = Distro('FreeBSD', cfg={}, paths={})
>>> f.manage_service('enable', 'ntpd')
SubpResult(stdout='ntpd enabled in /etc/rc.conf\n', stderr='')
>>> f.manage_service('restart', 'ntpd')
SubpResult(stdout='Starting ntpd.\n', stderr='ntpd not running? (check /var/db/ntp/ntpd.pid).\n')
>>> f.manage_service('restart', 'ntpd')
SubpResult(stdout='Stopping ntpd.\nWaiting for PIDS: 1127.\nStarting ntpd.\n', stderr='')
>>> 
```
## Checklist:

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -37,6 +37,26 @@ class Distro(cloudinit.distros.bsd.BSD):
     prefer_fqdn = True  # See rc.conf(5) in FreeBSD
     home_dir = "/usr/home"
 
+    def manage_service(self, action: str, service: str):
+        """
+        Perform the requested action on a service. This handles FreeBSD's
+        'service' case as necessary. The FreeBSD 'service' is closer in
+        features systemctl than SysV init's service.
+        May raise ProcessExecutionError
+        """
+        init_cmd = self.init_cmd
+        cmds = {
+            "stop": [service, "stop"],
+            "start": [service, "start"],
+            "enable": [service, "enable"],
+            "restart": [service, "restart"],
+            "reload": [service, "restart"],
+            "try-reload": [service, "restart"],
+            "status": [service, "status"],
+        }
+        cmd = list(init_cmd) + list(cmds[action])
+        return subp.subp(cmd, capture=True)
+
     def _get_add_member_to_group_cmd(self, member_name, group_name):
         return ["pw", "usermod", "-n", member_name, "-G", group_name]
 
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -40,8 +40,8 @@ class Distro(cloudinit.distros.bsd.BSD):
     def manage_service(self, action: str, service: str):
         """
         Perform the requested action on a service. This handles FreeBSD's
-        'service' case as necessary. The FreeBSD 'service' is closer in
-        features systemctl than SysV init's service.
+        'service' case. The FreeBSD 'service' is closer in features to
+        'systemctl' than SysV init's 'service', so we override it.
         May raise ProcessExecutionError
         """
         init_cmd = self.init_cmd
--- a/tests/unittests/distros/test_manage_service.py
+++ b/tests/unittests/distros/test_manage_service.py
@@ -3,6 +3,8 @@
 from tests.unittests.helpers import CiTestCase, mock
 from tests.unittests.util import MockDistro
 
+from . import _get_distro
+
 
 class TestManageService(CiTestCase):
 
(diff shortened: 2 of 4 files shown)


aciba90 (approved):
+1. Thanks, @igalic, for the addition.

igalic:
i should add
```python
"openbsd": ["openbsd"],
````
and perhaps
```python
"bsd": ["freebsd", "openbsd", "netbsd", ],
```

igalic:
>```python
> "bsd": ["freebsd", "openbsd", "netbsd", ],
>
This only makes sense if we also add

```python
"linux": [...],
```

igalic (commented):
I don't think this is the correct way to do this.

tests/unittests/distros/test_manage_service.py
```diff
@@ -28,6 +30,14 @@ def test_manage_service_service_initcmd(self, m_subp, m_sysd):
         self.dist.manage_service("start", "myssh")
         m_subp.assert_called_with(["service", "myssh", "start"], capture=True)
 
+    @mock.patch.object(MockDistro, "uses_systemd", return_value=False)
+    @mock.patch("cloudinit.distros.subp.subp")
+    def test_manage_service_rcctl_initcmd(self, m_subp, m_sysd):
+        self.dist = _get_distro("openbsd")
```
igalic, line 36: i honestly did not expect this to work.
but it does…??

igalic (commented):
works as unexpected.

tests/unittests/distros/test_manage_service.py
```diff
@@ -3,6 +3,8 @@
 from tests.unittests.helpers import CiTestCase, mock
 from tests.unittests.util import MockDistro
 
+from . import _get_distro
```
TheRealFalcon, line 6: Nit: I think we prefer absolute imports these days.

tests/unittests/distros/test_manage_service.py
TheRealFalcon, line 36: Nit: since you're not using the one created by setup, we don't really need the self here (though it's not hurting anything).

tests/unittests/distros/test_manage_service.py
```diff
@@ -28,6 +30,14 @@ def test_manage_service_service_initcmd(self, m_subp, m_sysd):
         self.dist.manage_service("start", "myssh")
         m_subp.assert_called_with(["service", "myssh", "start"], capture=True)
 
+    @mock.patch.object(MockDistro, "uses_systemd", return_value=False)
```
TheRealFalcon, line 33: This mock can be removed since you're not using it in this test.

TheRealFalcon (commented):
Left a few minor comments, but overall looks good!

With regard to the OSFAMILIES question you had, yeah, it's a bit ambiguous. The general thinking is "is something the list based on the parent?" For modules that's usually related to package management and/or networking tools. I.e., I can expect some apt module to work across all debian families or some NetworkManager module to work across all redhat ones. I don't know enough about the BSDs to know which makes more sense, but I'll trust that what you provided makes the most sense here.

tests/unittests/distros/test_manage_service.py
igalic: most other files in that directory import it that way ;)

but I'll happily fix that (and the others if desired)

igalic: see #1760

cloudinit/distros/freebsd.py
```diff
@@ -37,6 +37,26 @@ class Distro(cloudinit.distros.bsd.BSD):
     prefer_fqdn = True  # See rc.conf(5) in FreeBSD
     home_dir = "/usr/home"
 
+    def manage_service(self, action: str, service: str):
+        """
+        Perform the requested action on a service. This handles FreeBSD's
+        'service' case as necessary. The FreeBSD 'service' is closer in
+        features systemctl than SysV init's service.
```
igalic, line 44: in features *to* systemctl

igalic (commented):
typo

TheRealFalcon (approved):
LGTM! Thanks @igalic

PR #1761, opened 2022-09-30
Release 22.3.4

holmanb:
## Proposed Commit Message

```

```

## Checklist:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+22.3.4
+ - Fix Oracle DS primary interface when using IMDS (LP: #1989686)
+
 22.3.3
  - Fix Oracle DS not setting subnet when using IMDS (LP: #1989686)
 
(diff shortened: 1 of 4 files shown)


ChangeLog
```diff
@@ -1,3 +1,6 @@
+22.3.4
+ - Fix Oracle DS primary interface when using IMDS (#1757)
```
blackboxsw, line 2: Let's annotate this documenting the original bug too
```suggestion
 - Fix Oracle DS primary interface when using IMDS (LP: #1989686)
```

Then let's squash merge that change into https://github.com/canonical/cloud-init/pull/1761/commits/bcf94ba35d6e87186cef27b76b0724b061725657

blackboxsw (approved):
LGTM!

ChangeLog
holmanb: Done (tag updated as well)

PR #1765, opened 2022-10-03
testing: check logs for critical errors

cjp256:
Having recently noticed that LOG.critical() is used in cloudinit, update integration tests to check for CRITICAL.


vmarcella (approved):

blackboxsw (approved):
+1, thanks @cjp256 should have had this earlier in integration test validation, but not a very frequent use of this log level. Thank you for the improvement.

PR #1767, opened 2022-10-03
analyze: use init-local as start of boot record

cjp256:
When using the Azure data source, frames are not guaranteed to be unique to each boot.  Instead of splitting boot records when any event is seen more than once, do so whenever the start of `init-local` stage is found (and records have already been processed).


anhvoms (approved):

TheRealFalcon (approved):
Good find and nice change.

PR #1759, opened 2022-09-28
cc_ntp: add support for BSDs

igalic:
## Proposed Commit Message

```
cc_ntp: add support for BSDs

*BSDs have ntpd installed in base the base system
This PR extends cc_ntp to add support for ntpd,
openntpd, and chrony on the FreeBSD, and OpenBSD.

To make tests pass, we ensure that we are mocking functions,
not entire classes.

Co-authored-by: Ryan Harper <rharper@woxford.com>
Co-authored-by: James Falcon <james.falcon@canonical.com>

Sponsored by: FreeBSD Foundation

LP: #1990041
```

## Additional Context
This PR depends on #1758 

## Test Steps
- start a cloud-init vm with *BSD
- configure ntp
- observe results

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [x] I have updated or added any documentation accordingly

PR #1772, opened 2022-10-04
distros/manage_services: add support to disable service

igalic:
## Proposed Commit Message

```
distros/manage_services: add support to disable service

in order to support some modules on BSD, which could replace
base-services, we'll need to be able to disable those base-services.
This PR adds support for the "disable" command to 'systemct', BSD
'service', and OpenBSD 'rcctl', which currently suppor that.
For SysV init 'service', we follow the 'enable' path, and 'stop'
the service instead.


LP: #1991024

```

## Additional Context
#1759 needs support for this.

## Test Steps
The tests currently don't test every possible command.
As long as we have no typos, we can just try this in a REPL.

## Checklist:


holmanb:
I think this looks good. Could we add a test?


Maybe something like this? (feel free to use)
```suggestion
  diff --git a/tests/unittests/distros/test_manage_service.py b/tests/unittests/distros/test_manage_service.py
  index e53baf9a0..a6b8a3dce 100644
  --- a/tests/unittests/distros/test_manage_service.py
  +++ b/tests/unittests/distros/test_manage_service.py
  @@ -52,5 +52,11 @@ class TestManageService(CiTestCase):
               ["systemctl", "start", "myssh"], capture=True
           )
   
  -
  -# vi: ts=4 sw=4 expandtab
  +    @mock.patch.object(MockDistro, "uses_systemd", return_value=True)
  +    @mock.patch("cloudinit.distros.subp.subp")
  +    def test_manage_service_disable_systemctl(self, m_subp, m_sysd):
  +        self.dist.init_cmd = ["ignore"]
  +        self.dist.manage_service("disable", "myssh")
  +        m_subp.assert_called_with(
  +            ["systemctl", "disable", "myssh"], capture=True
  +        )
```

igalic:
n.b.: I didn't add that vi line…
I'm using emacs (as vim ;)

holmanb (changes requested):
Could we switch the sysvinit implementation to use `chkconfig` on/off? I doubt that code gets much mileage, but I think that's probably more "correct" than start/stop.

igalic:
on whichever Linux still has that thing: i guess.
on NetBSD, i'd have to reimplement the whole thing in… better.

the first one is easy, the latter might take some hackery

holmanb:
> the first one is easy, the latter might take some hackery

I was only meant to ask about sysV Linux, I should have been more explicit.

However, since you pointed out that this codepath is shared between sysV and openrc (in irc), this request bigger than I thought and would require changes to codepaths that don't have great test coverage at the moment, so maybe the request doesn't make as much sense as I thought it did.

igalic:
and there might be less benefit: on Linux, once you install competing systems services, the other things are usually disabled by the package manager, uninstalled and then the actual thing you're trying to install is enabled

on BSD, there's a number of systems in the base system which you don't have to / can't uninstall. (FreeBSD is trying to go a different path with PkgBase, but after many years that's still far from complete).
for this reason, i suppose, when something is installed from ports / packages / pkgsrc, nothing is enabled or is disabled.

PR #1773, opened 2022-10-04
OpenBSD: remove pkg_cmd_environ function

igalic:
## Proposed Commit Message

```
OpenBSD: remove pkg_cmd_environ function

This environment variable is being set to the main mirror, which is
heavily under load. It also sets it to a deprecated and highly insecure
protocol.

OpenBSD's default behaviour is to read a list of mirrors from
installurl(5): https://man.openbsd.org/installurl.5

We can restore this behaviour by removing the contents of this function.
Note that the Equivalent NetBSD function makes vastly more sense.

LP: 1991567

```

## Test Steps
Try install a package thru a REPL.
If this code ever did anything, it should now be faster.

## Checklist:

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -176,7 +176,8 @@ class Distro(cloudinit.distros.bsd.BSD):
                 )
 
     def _get_pkg_cmd_environ(self):
-        """Return environment vars used in FreeBSD package_command operations"""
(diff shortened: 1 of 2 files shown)


aciba90 (changes requested):
+1, thanks. Could you, @igalic, please fix the linting errors?

goneri (approved):

holmanb (approved):
Thanks!

PR #1673, opened 2022-08-19
Improve Ubuntu Advantage module's idempotency

chifac08:
## Proposed Commit Message

```
Ubuntu Advantage: Improved idempotency

Ubuntu Advantage module is now capable of detecting already
activated subscriptions or/and enabled services. Furthermore,
when "enable" parameter is set, no "default" services
will be enabled anymore.
```

## Additional Context
If `instance-id` changes due to updated user or meta-data, cloud-init
will run again and UA causes an error when a machine has an already
attached subscription and/or service.

Ubuntu Advantage is now capable of ignoring the following messages:

```
Stderr: This machine is already attached to 'X'
        To use a different subscription first run: sudo ua detach.

Stdout: One moment, checking your subscription first
        <Service> is already enabled.
        See: sudo ua status
```
        
The module won't set cloud-init status to error when 
detecting those messages. Nevertheless, it will print error
messages in cloud-init log for easier troubleshooting.

Furthermore, when "enable" parameter is set, the module
only activates the list of services given rather than 
activating default services + list of enabled.

## Test Steps
Tested with the following user-data:

1. user-data with no "enable":
```
ubuntu_advantage:
  token: "{{ ds.meta_data.ua_token }}"
```

3. cloud-init clean --logs --reboot

4. cloud-init status should not state "error" and "default services" should be enabled:
```
root@test:/home/ubuntu# cloud-init status -l
status: done
time: Fri, 19 Aug 2022 06:49:50 +0000
detail:
DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud,/dev/sdc][dsmode=net]

SERVICE          ENTITLED  STATUS    DESCRIPTION
cc-eal           yes       n/a       Common Criteria EAL2 Provisioning Packages
esm-infra        yes       enabled   UA Infra: Extended Security Maintenance (ESM)
fips             yes       disabled  NIST-certified core packages
fips-updates     yes       disabled  NIST-certified core packages with priority security updates
livepatch        yes       enabled   Canonical Livepatch service
usg              yes       disabled  Security compliance and audit tools
```

5. ua detach

6. user-data with "enable":
```
ubuntu_advantage:
  token: "{{ ds.meta_data.ua_token }}"
  enable:
    - livepatch
```

7. cloud-init clean --logs --reboot

8. cloud-init status should not state "error" and only "livepatch" service should be enabled:
```
root@test:/home/ubuntu# cloud-init status -l
status: done
time: Fri, 19 Aug 2022 06:49:50 +0000
[...]

PR #1775, opened 2022-10-05
cc_ua: consume ua json api for enable commands

aciba90:
## Proposed Commit Message

```
cc_ua: consume ua json api for enable commands

Remove safeguard uaclient imports as uaclient has already been released.
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-1253
https://github.com/canonical/cloud-init/pull/1673#discussion_r965352866

Example the UA json api for enable:

```bash
$ sudo ua enable --assume-yes --format json asdf livepatch esm-apps esm-apps | jq
{
  "_schema_version": "0.1",
  "errors": [
    {
      "message": "Cannot install Livepatch on a container.",
      "message_code": "livepatch-error-install-on-container",
      "service": "livepatch",
      "type": "service"
    },
    {
      "message": "UA Apps: ESM is already enabled.\nSee: sudo ua status",
      "message_code": "service-already-enabled",
      "service": "esm-apps",
      "type": "service"
    },
    {
      "message": "UA Apps: ESM is already enabled.\nSee: sudo ua status",
      "message_code": "service-already-enabled",
      "service": "esm-apps",
      "type": "service"
    },
    {
      "message": "Cannot enable unknown service 'asdf'.\nTry cc-eal, cis, esm-infra, fips, fips-updates, livepatch.",
      "message_code": "invalid-service-or-failure",
      "service": null,
      "type": "system"
    }
  ],
  "failed_services": [
    "asdf",
    "esm-apps",
    "livepatch"
  ],
  "needs_reboot": false,
  "processed_services": [],
  "result": "failure",
  "warnings": []
}
```

## Test Steps

```bash
CLOUD_INIT_UA_TOKEN="<token>" tox -e integration-tests -- tests/integration_tests/modules/test_ubuntu_advantage.py::TestUbuntuAdvantage::test_idempotency
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -6,7 +6,7 @@ import json
 import re
 from logging import Logger
 from textwrap import dedent
-from typing import Any, List, Tuple
+from typing import Any, List, Sequence
 from urllib.parse import urlparse
 
 from cloudinit import log as logging
@@ -148,8 +148,9 @@ KNOWN_UA_CONFIG_PROPS = (
     "ua_apt_https_proxy",
 )
 
-# Pro cli message codes
-_PRO_MC_SERVICE_ALREADY_ENABLED = "service-already-enabled"
+
+def _format_services(services: Sequence[str]) -> str:
+    return ", ".join(services)
 
 
 def validate_schema_features(ua_section: dict):
@@ -282,6 +283,7 @@ def configure_ua(token, enable=None):
         )
         enable = []
 
+    # Perform attach
     if enable:
         attach_cmd = ["ua", "attach", "--no-auto-enable", token]
     else:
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -147,6 +148,9 @@
     "ua_apt_https_proxy",
 )
 
+# Pro cli message codes
+_PRO_MC_SERVICE_ALREADY_ENABLED = "service-already-enabled"
```
blackboxsw, line 152: Maybe it's worth us `from uaclient.messages impor ALREADY_ENABLED` and matching `ALREADY_ENABLED.name` in our errors checks?

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -293,21 +297,25 @@ def configure_ua(token, enable=None):
...
         raise RuntimeError(msg) from e
 
-    enable_errors = []
+    enable_errors: List[Tuple[str, str]] = []
     for service in enable:
         try:
-            cmd = ["ua", "enable", "--assume-yes", service]
-            subp.subp(cmd, capture=True)
+            cmd = ["ua", "enable", "--assume-yes", "--format", "json", service]
+            enable_cmd_result = subp.subp(cmd, capture=True)
```
blackboxsw, line 304: I think we minimally have to have an rcs to allow for exit 1 here otherwise enable_cmd_result will not be set resulting in UnboundLocalErrors on enable_cmd_result if the first item in an enable: [esm-apps] is already enabled.
```suggestion
            enable_cmd_result = subp.subp(cmd, capture=True, rcs={0,1})
```
By providing for rcs={0,1} we may not actually need the except ProcessExecutionError handling, but I think we should leave it in in case UA adds more return codes from this CLI call besides 0 and 1. We should catch any errors which exit 1 in your following for loop anyway.

tests/unittests/config/test_cc_ubuntu_advantage.py
```diff
@@ -251,13 +269,21 @@ def test_configure_ua_attach_on_service_enabled(self, m_subp, caplog):
...
+                ["ua", "enable", "--assume-yes", "--format", "json", svc]
+                for svc in ["livepatch"]
             ]
             if cmd in fail_cmds and capture:
-                svc = cmd[-1]
-                raise subp.ProcessExecutionError(
-                    'Service "{}" is already enabled.'.format(svc)
-                )
+                response = {
+                    "errors": [
```
blackboxsw, line 277: I think we need to set the exit code 1 on this response and assert handling that failure path

blackboxsw (changes requested):
Thanks for this @aciba90 . I think we minimally need a unittest that'll exercise enable response and exit 1  when a service is already enabled as that should have caught a traceback on  "local variable 'enable_cmd_result' referenced before assignment" when subp exits 1 on service-already-enabled.

Sorry I didn't get a full review yet, but wanted to give you something to work with in the morn.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Done, and unit-covered.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -288,35 +294,78 @@ def configure_ua(token, enable=None):
...
+    # related. We can distinguish them by checking if `service` is non-null
+    # or null respectively.
+
+    try:
+        from uaclient.messages import ALREADY_ENABLED  # noqa
+
+        UA_MC_ALREADY_ENABLED = ALREADY_ENABLED.name
+    except ImportError:
+        UA_MC_ALREADY_ENABLED = "service-already-enabled"
+
```
blackboxsw, line 348: I don't think we need/want this ImportError handling anymore.... `27.11.2` is released anywhere cloud-init will be published. So we should have this import problem.

blackboxsw: While we are touching this file, we can also drop the try/ImportErrors around `uaclient.api.u.pro.attach.auto.should_auto_attach.v1` too as that's also published now.
Might as well just pull these imports up out of the functions and up to the top of the cc_ubuntu_advantage module.

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -288,35 +294,78 @@ def configure_ua(token, enable=None):
...
+        enable_errors.append(err)
 
     if enable_errors:
-        for service, error in enable_errors:
-            msg = 'Failure enabling "{service}":\n{error}'.format(
-                service=service, error=str(error)
-            )
+        error_services: List[str] = []
+        for err in enable_errors:
+            if err["service"] is not None:
```
blackboxsw, line 359: Let's be definsive here and use `err.get("service")` in case the error does not contain a "service" key at some point in the future.

tests/integration_tests/modules/test_ubuntu_advantage.py
```diff
@@ -107,6 +115,25 @@ def test_valid_token(self, client: IntegrationInstance):
...
         assert "This machine is not attached" not in status.stdout
 
+    @pytest.mark.user_data(ATTACH.format(token=CLOUD_INIT_UA_TOKEN))
+    def test_idempotency(self, client: IntegrationInstance):
+        assert CLOUD_INIT_UA_TOKEN, "CLOUD_INIT_UA_TOKEN env var not provided"
+        log = client.read_from_file("/var/log/cloud-init.log")
+        verify_clean_log(log)
+        status = client.execute("ua status")
+        assert status.ok
+        assert "This machine is not attached" not in status.stdout
```
blackboxsw, line 125: Just to avoid fragility in the event of internationalization of messages at some point. Let's use `ua status --format json` here and look at `{"enabled": True|False}` in the response in this case and below

blackboxsw (changes requested):
Thanks for this @aciba90 looks really good. A couple of points to bring this home and let's land it Monday.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: Safe imports removed. I keep them within the functions to avoid import errors in instances without `uaclient` installed. For example, someone building the deb package on a debian machine would get import errors at autopkgtest-time.

cloudinit/config/cc_ubuntu_advantage.py
aciba90: done

tests/integration_tests/modules/test_ubuntu_advantage.py
aciba90: done

aciba90:
Thanks, @blackboxsw, for the comments. This PR is ready to be reviewed.

cloudinit/config/cc_ubuntu_advantage.py
blackboxsw: Oops. I stand corrected. Totally agree with you on that, then downstreams without ``uaclient` don't have to worry about not packaging it. +1

cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -288,35 +290,79 @@ def configure_ua(token, enable=None):
...
+    #   }
+    #
+    # From our pov there are two type of errors, service and non-service
+    # related. We can distinguish them by checking if `service` is non-null
+    # or null respectively.
+
+    # pylint: disable=import-error
+    from uaclient.messages import ALREADY_ENABLED
+
+    # pylint: enable=import-error
```
blackboxsw, line 341: I added this closing pylint disable/enable comment here. Ultimately, I think we may want to add something like the following to test-requirements.txt:
```diff
diff --git a/test-requirements.txt b/test-requirements.txt
index 739d8ebb4..fb77c867a 100644
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -7,3 +7,6 @@ pytest-mock
 setuptools
 jsonschema
 responses
+
+# Ubuntu-specific support for cc_ubuntu_advantage module in unittests
+uaclient @ git+https://github.com/canonical/ubuntu-advantage-client.git@b5ed2599917423d2633ce6101eb47f17b9f63426
```

But, this needs more discussion on the approach we agree to take as we'd need to think more about impact to non-ubuntu unittest runners.

Currently, the above would break non-ubuntu testing as seen if we were to run `./tools/run-container --package --source-package --unittest centos/7`. But, having the actual library imported for uaclient will provide is with more thorough unittests for Ubuntu and I think eventually e want to get to a unittest path where we don't have to ignore 
pylint warnings on import-error as well as eventually drop the [fake_uaclient fixture](https://github.com/canonical/cloud-init/blob/main/tests/unittests/config/test_cc_ubuntu_advantage.py#L57) for our ubuntu-based testing.   This comment doesn't need to block this current PR though.

blackboxsw (approved):
LGTM. We can iterate on ubuntu-specific testing w/ `uaclient` library imports in separate PRs.

PR #1776, opened 2022-10-05
tests: Use LXD metadata to determine NoCloud status

TheRealFalcon:
## Proposed Commit Message

```
tests: Use LXD metadata to determine NoCloud status

Rather than hardcoding a series name, use the LXD metadata to determine
if the current instance should use NoCloud rather than LXD datasource.
```

### DIFF
--- a/tests/integration_tests/util.py
+++ b/tests/integration_tests/util.py
@@ -5,12 +5,14 @@ import re
 import time
 from collections import namedtuple
 from contextlib import contextmanager
+from functools import lru_cache
 from itertools import chain
 from pathlib import Path
 from typing import Set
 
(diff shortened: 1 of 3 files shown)


tests/integration_tests/util.py
```diff
@@ -161,3 +163,12 @@ def get_console_log(client: IntegrationInstance):
     if console_log.lower().startswith("no console output"):
         pytest.fail("no console output")
     return console_log
+
+
+@lru_cache()
```
blackboxsw, line 168: Nice idea caching this +1.

holmanb (approved):
LGTM

PR #1779, opened 2022-10-10
Net: add BSD ifconfig(8) parser and state class

igalic:
## Proposed Commit Message

```
Net: add BSD ifconfig(8) parser and state class

This parser covers all 3 main BSDs' `ifconfig(8)` output and will serve
as jumping off point to replicating the `/sys` reads on Linux.

The parser class consists of three public methods: `parse()`, returns a
`Dict` of `Ifstate` objects, addressed by name and MAC address.
`ifs_by_name()` returns a `Dict` of `Ifstate` objects by device name,
and `ifs_by_mac()` returns them by MAC address.
The `parse()` method is LRU cached, based on input.

The `Ifstate` class holds all *parsed*, as well as some easy to compute
state of a BSD network interface.
It is, and should remain entirely side-effect-free.

```

## Test Steps
So far, testing has bee in REPL:

```python
>>> from cloudinit.distros import subp
>>> from cloudinit.distros.parsers import ifconfig
>>> ifc = ifconfig.Ifconfig()
>>> ifc_txt, _err = subp.subp(['ifconfig', '-a'])
>>> ifs =  ifc.parse(ifc_txt)
```

## Checklist:


igalic:
I don't think it's ready to merge yet, but it's ready for review.

I'd love to hear some feedback on this.
Where to integrate it first, and what else to build in.

It would be good to get a driver mapping; the BSDs don't implement the blacklist_driver parameter of various networking functions
Tho that might be tricky if the interface has been renamed:
Usually it's really simple: `re0` → [`re(4)`](https://man.freebsd.org/re(4)) (RealTek)
But if the interface has been renamed, we'll have to consult /etc/rc.conf for entries of the form:
``` 
ifconfig_re0_name=eth0
```

I think the most interesting function to build on top of this will be `get_interfaces_by_mac()`, because right now, everything is based on name.
A simple hack might be to put all `macs` as dict keys referencing the same `Ifstate` object.

igalic:
a simpler way to verify if a device is renamed is to query [`devinfo(8)`](https://man.freebsd.org/devinfo(8)): If a ***physical*** device returns an error when queried with devinfo:

```
freebsd@fbsd14-amd64 ~> sudo -H grep eth0 /var/log/messages 
Oct 18 22:36:33 fbsd14-amd64 kernel: vtnet0: changing name to 'eth0'
Oct 18 22:36:33 fbsd14-amd64 kernel: eth0: link state changed to UP
freebsd@fbsd14-amd64 ~> sudo -H grep eth0 /etc/rc.conf
ifconfig_vtnet0_name="eth0"
freebsd@fbsd14-amd64 ~> devinfo -p eth0
devinfo: eth0: Not found
freebsd@fbsd14-amd64 ~ [1]> devinfo -p vtnet0
vtnet0 virtio_pci0 pci0 pcib0 acpi0 nexus0
freebsd@fbsd14-amd64 ~> 
```

tho we still cannot tell what it was renamed from, and thus what its driver is, without looking at rc.conf

I'm now wondering if we could just scrape `/var/log/messages` for `changing name to '<new-name0>'`, which is probably a good point to stop, and reconsider my life choices.

TheRealFalcon (changes requested):
Overall, the approach looks good. I left a few inline comments.

I do also wonder how consistent this output will look across BSDs and their versions. It's already decently different than Linux.

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,245 @@
...
+        # OpenBSD makes this very easy:
+        if "egress" in self.groups:
+            return True
+        if self.groups == [] and self.media and "Ethernet" in self.media:
+            return True
+        return False
+
+    @property
+    def is_bridge(self) -> bool:
+        if "bridge" in self.groups:
```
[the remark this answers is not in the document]
igalic: I know
but there are exceptions to this rule, so i thought I'd keep them all looking the same

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,245 @@
...
+            ip = IPv6Interface("%s/%s" % (toks[1], prefixlen))
+
+        if not scope and ip.is_link_local:
+            scope = "link-local"
+        elif not scope and ip.is_site_local:
+            scope = "site-local"
+
+        return (str(ip.ip), {"prefixlen": prefixlen, "scope": scope})
+
+    def _parse_flags(self, toks: list) -> dict:
```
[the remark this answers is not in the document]
igalic: the issue you're seeing is definitely due to differences in ifconfig output between Linux and BSD, but there's no harm making this more resilient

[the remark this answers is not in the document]
igalic: note also that this section of the code is almost verbatim lifted from the netinfo code

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,257 @@
...
+from functools import lru_cache
+from ipaddress import IPv4Address, IPv4Interface, IPv6Interface
+from typing import Dict, Optional, Tuple
+
+from cloudinit import log as logging
+
+LOG = logging.getLogger(__name__)
+
+
+class Ifstate:
```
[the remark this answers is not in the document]
igalic: this whole file is BSD specific, and the reason is because Linux has long since deprecated ifconfig
but also, neither Linux' ifconfig nor its ip provide the depth of information that the BSD utilities do.

ifconfig on BSD is a userland tool to expose kernel information, and is treated to require the same kind of stability as an ABI. new things might be added, but things are changed or removed with great care.

my overconfident guess is that the general form of this output hasn't changed in a decade on each of the BSSs

cloudinit/distros/networking.py
```diff
@@ -185,18 +186,47 @@ def try_set_link_up(self, devname: DeviceName) -> bool:
 class BSDNetworking(Networking):
     """Implementation of networking functionality shared across BSDs."""
 
+    def __init__(self):
+        self.ifc = ifconfig.Ifconfig()
+        self.ifs = {}
+
+    def _update_ifs(self):
```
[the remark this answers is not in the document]
igalic: I've head this initially, and made lots of tests fail 😅 
so, yes, i think it would make sense… it'll just be more work.
but, yeah, we probably do not need to call `ifconfig -a` seventeen thousand times.
so, yes, a good idea.

cloudinit/distros/parsers/ifconfig.py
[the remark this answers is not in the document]
igalic: added a check if the dict is empty;

TheRealFalcon:
@igalic Is this ready for re-review?

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        parse returns a dict of Ifstates
+        This dict will always be the same, given the same input, so we can
+        lru_cache() it. note that lru_cache() takes only the __hash__() of the
+        input (text), so it should be fairly quick, despite our giant inputs.
+        """
+        for line in text.splitlines():
+            if len(line) == 0:
+                continue
+            if line[0] not in ("\t", " "):
```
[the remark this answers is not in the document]
igalic: maybe, but: What is the default value for `curif` and `dev`?

`curif` could be `''`, which is a valid key in a `dict`, so I'd have to check that…

but for `dev`, the only sensible default is None, and then `mypy` gets really petty about all the assignments to `dev[x]`, because they are now all `Optional`.

igalic:
> @igalic Is this ready for re-review?

nope, still have to fix tests… which only seem to be failing on my machine??
are we not running unit tests anywhere?

cloudinit/distros/parsers/ifconfig.py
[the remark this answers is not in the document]
igalic: I've added a comment to explain what's going on here, and hope it helps
because initializing these at the top of the loop is impossible.

TheRealFalcon:
> nope, still have to fix tests… which only seem to be failing on my machine??
are we not running unit tests anywhere?

Unit tests are run on every PR push. If you look at the Travis runs for the PR, you can click through to the unit test runs. `main` is currently passing.

igalic:
> > nope, still have to fix tests… which only seem to be failing on my machine??
> > are we not running unit tests anywhere?
> 
> Unit tests are run on every PR push. If you look at the Travis runs for the PR, you can click through to the unit test runs. `main` is currently passing.

I'm not seeing any of that.
Only Lint Tests.

Is this perhaps a permission issue?

cloudinit/net/__init__.py
```diff
@@ -372,9 +372,6 @@ def device_devid(devname):
 
 
 def get_devicelist():
-    if util.is_FreeBSD() or util.is_DragonFlyBSD():
```
TheRealFalcon, line 375: Why was this section removed?

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,252 @@
...
+    def is_bridge(self) -> bool:
+        return "bridge" in self.groups
+
+    @property
+    def is_bond(self) -> bool:
+        return "lagg" in self.groups
+
+    @property
+    def is_vlan(self) -> bool:
+        return "vlan" in self.groups or self.vlan
```
TheRealFalcon, line 68: Is this supposed to be
`return "vlan" in self.groups or bool(self.vlan)`

As-is, it will return `True` or the `self.vlan` dict

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -0,0 +1,252 @@
...
+
+    @property
+    def is_wlan(self) -> bool:
+        return "wlan" in self.groups or self.vlan
```
TheRealFalcon, line 72: Same question here.

TheRealFalcon (commented):
I think we're really close! Test changes look good overall. I left a couple inline comments.

Additionally, the following example you gave me is failing to parse:
```
lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> mtu 32768
	index 3 priority 0 llprio 3
	groups: lo
	inet6 ::1 prefixlen 128
	inet6 fe80::1%lo0 prefixlen 64 scopeid 0x3
vio0: flags=a48843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST,AUTOCONF6TEMP,AUTOCONF6,AUTOCONF4> mtu 1500
	lladdr 52:54:00:35:4f:b8
	index 1 priority 0 llprio 3
	groups: egress
	media: Ethernet autoselect
	status: active
	inet6 fe80::5054:ff:fe35:4fb8%vio0 prefixlen 64 scopeid 0x1
	inet 192.168.122.81 netmask 0xffffff00 broadcast 192.168.122.255
enc0: flags=0<>
	index 2 priority 0 llprio 3
	groups: enc
	status: active
pflog0: flags=141<UP,RUNNING,PROMISC> mtu 33136
	index 4 priority 0 llprio 3
	groups: pflog
```

```
  File "/home/james/unrelated_python.py", line 30, in <module>
    foo.parse(output)
  File "/home/james/cloud-init/cloudinit/distros/parsers/ifconfig.py", line 121, in parse
    flags = self._parse_flags(toks)
  File "/home/james/cloud-init/cloudinit/distros/parsers/ifconfig.py", line 250, in _parse_flags
    if toks[4] == "mtu":
IndexError: list index out of range
```

cloudinit/net/__init__.py
igalic: because there's now a get_devicelist in networking.py

cloudinit/distros/parsers/ifconfig.py
igalic: this should only return True if we have a vlan device on our hands

usually vlan would be set in groups
but if for some reason it isn't, i meant to check that the vlan dict isn't empty. maybe i should've checked for that, exactly

cloudinit/distros/parsers/ifconfig.py
igalic: this shouldn't be checking for vlan at all
only wlan

cloudinit/net/__init__.py
igalic: however, this will need an actual refactor, so I'm adding that line back in.

cloudinit/distros/networking.py
```diff
@@ -185,17 +186,51 @@ def try_set_link_up(self, devname: DeviceName) -> bool:
...
+        Not guaranteed to bring the interface up. The caller is expected to
+        add wait times before retrying."""
+        subp.subp(["ifconfig", devname, "up"])
+        return self.is_up(devname)
+
+    def get_devicelist(self) -> list:
+        return list(self.ifc.ifs_by_name().keys())
+
+    def get_interface_mac(self, devname: DeviceName):
+        return self.ifs[devname].mac
```
igalic, line 233: Weirdly, this isn't being called:
```
  File "/usr/local/lib/python3.9/site-packages/cloudinit/cmd/main.py", line 767, in status_wrapper
    ret = functor(name, args)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/cmd/main.py", line 433, in main_init
    init.apply_network_config(bring_up=bring_up_interfaces)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/stages.py", line 924, in apply_network_config
    self.distro.networking.wait_for_physdevs(netcfg)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/distros/networking.py", line 149, in wait_for_physdevs
    present_macs = self.get_interfaces_by_mac().keys()
  File "/usr/local/lib/python3.9/site-packages/cloudinit/distros/networking.py", line 77, in get_interfaces_by_mac
    blacklist_drivers=self.blacklist_drivers
AttributeError: 'FreeBSDNetworking' object has no attribute 'blacklist_drivers'
```

it looks like what is being called is the one from `Networking`

igalic (commented):
testing this on a FreeBSD 14.0-CURRENT VM with two nics (re0, and vtnet0 renamed to eth0)

according to logs, `ifconfig -a` is called 8 times, so I'm pretty glad I added the cache.

cloudinit/distros/networking.py
igalic: it looks like the most sensible approach is to backout this change, and do a proper refactor for everything that's not an `@abc.abstractmethod` yet in `Networking`.

igalic: done.

cloudinit/distros/networking.py
```diff
@@ -185,17 +186,51 @@ def try_set_link_up(self, devname: DeviceName) -> bool:
...
+        """Try setting the link to up explicitly and return if it is up.
+        Not guaranteed to bring the interface up. The caller is expected to
+        add wait times before retrying."""
+        subp.subp(["ifconfig", devname, "up"])
+        return self.is_up(devname)
+
+    def get_devicelist(self) -> list:
+        return list(self.ifc.ifs_by_name().keys())
+
+    def get_interface_mac(self, devname: DeviceName):
```
igalic, line 232: also, remove this one…

TheRealFalcon (approved):
Thanks for all of the work here @igalic . This looks good to me now!

PR #1785, opened 2022-10-14
collect logs apport subiquity support (SC-1279)

blackboxsw:
## Proposed Commit Message
```
cli/apport: collect-logs include subiquity logs config when present

Add support for both cloud-init collect-logs and apport to
include subiquity live installer artifacts if present to aid in
bug triage.

Apport integration to attach subiquity, curtin and
ubuntu-desktop-installer report keys when logs or
config files are present.

ubuntu-bug cloud-init will also automatically tag the
bug as curtin, subiquity or ubuntu-desktop-installer if
related logs are present in the bug report.

Additional collect-logs support to collect
/var/lib/cloud/data in the event that cloud-init is disabled
by systemd generator with /run/cloud-init/disabled flag.
In these situations we want to collect /var/lib/cloud/data
dir for more context on prior cloud-init behavior.
```

## Additional Context
[LP: #1993011](http://pad.lv/1993011) using `sudo ubuntu-bug cloud-init` on server live installed system which attached subiquitylogs and adds `subiquity` tag

## Test Steps
```bash
make deb
CLOUD_INIT_CLOUD_INIT_SOURCE=./cloud-init_22.3-72-g760fd1b9-1~bddeb_all.deb tox -e integration-tests   tests/integration_tests/test_paths.py
```

## Checklist:

### DIFF
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -3,11 +3,17 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 """Cloud-init apport interface"""
+
 from cloudinit.cmd.devel import read_cfg_paths
+from cloudinit.cmd.devel.logs import (
+    INSTALLER_APPORT_FILES,
+    INSTALLER_APPORT_SENSITIVE_FILES,
+)
 
 try:
     from apport.hookutils import (
         attach_file,
+        attach_file_if_exists,
         attach_root_command_outputs,
         root_command_output,
     )
@@ -110,30 +116,57 @@ def attach_cloud_info(report, ui=None):
                 report["CloudName"] = "None"
 
 
+def attach_installer_files(report, ui=None):
+    """Attach any subiquity installer logs config.
+
+    To support decoupling apport integration from installer config/logs,
+    we eventually want to either source this function or APPORT_FILES
+    attribute from subiquity  and/or ubuntu-desktop-installer package-hooks
+    python modules.
+    """
+    for apport_file in INSTALLER_APPORT_FILES:
+        attach_file_if_exists(report, apport_file.path, apport_file.label)
+
+
 def attach_user_data(report, ui=None):
     """Optionally provide user-data if desired."""
     if ui:
         user_data_file = _get_user_data_file()
         prompt = (
-            "Your user-data or cloud-config file can optionally be provided"
-            " from {0} and could be useful to developers when addressing this"
-            " bug. Do you wish to attach user-data to this bug?".format(
-                user_data_file
-            )
+            "Your user-data, cloud-config or autoinstall files can optionally "
+            " be provided from {0} and could be useful to developers when"
+            " addressing this bug. Do you wish to attach user-data to this"
+            " bug?".format(user_data_file)
         )
         response = ui.yesno(prompt)
         if response is None:
             raise StopIteration  # User cancelled
         if response:
             attach_file(report, user_data_file, "user_data.txt")
+            for apport_file in INSTALLER_APPORT_SENSITIVE_FILES:
+                attach_file_if_exists(
+                    report, apport_file.path, apport_file.label
+                )
 
 
 def add_bug_tags(report):
     """Add any appropriate tags to the bug."""
+    new_tags = []
+    if report.get("CurtinError"):
+        new_tags.append("curtin")
+    if report.get("SubiquityLog"):
+        new_tags.append("subiquity")
     if "JournalErrors" in report.keys():
         errors = report["JournalErrors"]
         if "Breaking ordering cycle" in errors:
-            report["Tags"] = "systemd-ordering"
+            new_tags.append("systemd-ordering")
+    if report.get("UdiLog"):
+        new_tags.append("ubuntu-desktop-installer")
+    if new_tags:
+        report.setdefault("Tags", "")
+        if report["Tags"]:
+            report["Tags"] += " "
+        report["Tags"] += " ".join(new_tags)
 
 
 def add_info(report, ui):
@@ -151,6 +184,7 @@ def add_info(report, ui):
     attach_hwinfo(report, ui)
     attach_cloud_info(report, ui)
     attach_user_data(report, ui)
+    attach_installer_files(report, ui)
     add_bug_tags(report)
     return True
 
--- a/cloudinit/cmd/devel/logs.py
+++ b/cloudinit/cmd/devel/logs.py
@@ -45,9 +45,11 @@ INSTALLER_APPORT_FILES = [
     ApportFile(
         "/var/log/installer/subiquity-client-debug.log", "SubiquityClientLog"
     ),
-    ApportFile("/var/log/curtin/install.log", "CurtinLog"),
-    ApportFile("/var/log/curtin-install.log", "CurtinLogPostReboot"),
-    ApportFile("/var/log/curtin/curtin-error-logs.tar", "CurtinError"),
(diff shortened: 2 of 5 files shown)


cloudinit/apport.py
```diff
@@ -110,30 +116,57 @@ def attach_cloud_info(report, ui=None):
                 report["CloudName"] = "None"
 
 
+def attach_installer_files(report, ui=None):
```
blackboxsw, line 119: @dbungert in the future we might lifyt this type of function into `/usr/share/apport/package-hooks/subiquity.py`  or maybe surfacing an APPORT_FILES module-level attribute which could be a dict keyed on "<path>" with <apport_label> as the value. In either approach subiquity could control what files is cares to attach to a report. We can talk about this approach for next cycle if it sounds compelling.

cloudinit/cmd/devel/logs.py
```diff
@@ -22,11 +24,44 @@
 CLOUDINIT_RUN_DIR = "/run/cloud-init"
 
 
+class ApportFile(NamedTuple):
+    path: str
+    label: str
+
+
+INSTALLER_APPORT_SENSITIVE_FILES = [
```
blackboxsw, line 32: @dbungert please review these potential files and Apport bug labels to see if they make sense or align w/ current expectations for subiquity apport bugs. I've scraped them from latest server live installer and desktop canary  `/usr/share/apport/package-hooks/(subiquity|ubuntu-desktop-installer).py`

aciba90 (approved):
LGTM. I left only a typo correction suggestion.

cloudinit/apport.py
```diff
@@ -110,30 +116,57 @@ def attach_cloud_info(report, ui=None):
...
+                attach_file_if_exists(
+                    report, apport_file.path, apport_file.label
+                )
 
 
 def add_bug_tags(report):
     """Add any appropriate tags to the bug."""
+    new_tags = []
+    if report.get("CurtinError"):
+        new_tags.append("curtin")
```
TheRealFalcon, line 156: Should I be expecting to see these additional tags in LP: #1993011 ?

cloudinit/cmd/devel/logs.py
```diff
@@ -179,6 +232,14 @@ def collect_logs(tarfile, include_userdata: bool, verbosity=0):
                 1,
                 verbosity,
             )
+        if os.path.exists(os.path.join(CLOUDINIT_RUN_DIR, "disabled")):
```
TheRealFalcon, line 235: Is there a reason we don't always want whats in this dir?

TheRealFalcon (approved):
LGTM, but we should wait for other comments

cloudinit/apport.py
blackboxsw: Added context too on the test bug filed.

As this bug was filed from a "live server" install, and there were no curtin-specific errors, we only expect to see the `subiquity` tag.

`curtin` tag would be added if there were a top-level `CurtinError` report key due to curtin specific errors and the presence of `/var/log/installer/curtin-error-logs.tar`).

The `ubuntu-desktop-installer` tag would be added only if ubuntu-bug cloud-init were called from a live desktop installer environment which will add a report key of `UdiLog`. This key is added if the following log exists on the installed system: `/var/log/installer/ubuntu_desktop_installer.log`

cloudinit/cmd/devel/logs.py
blackboxsw: Only reason is that those files we care about are already symlinked into /run/cloud-init when cloud-init is enabled. So, it'd be duplication in the typical enabled case.
```bash
lrwxrwxrwx 1 root root   36 Oct 17 20:19 result.json -> ../../var/lib/cloud/data/result.json
....
lrwxrwxrwx 1 root root   36 Oct 17 20:19 status.json -> ../../var/lib/cloud/data/status.json
```

cloudinit/cmd/devel/logs.py
```diff
@@ -110,6 +147,21 @@ def _collect_file(path, out_dir, verbosity):
         _debug("file %s did not exist\n" % path, 2, verbosity)
 
 
+def collect_installer_logs(log_dir, include_userdata, verbosity):
+    """Obtain subiquity logs and config flies."""
```
aciba90, line 151:
```suggestion
    """Obtain subiquity logs and config files."""
```

blackboxsw (commented):
We'll likely get a review on this next week from subiquity team. They are aware of this approach and we can iterate on additional needs if need be to improve integration with subiquity if there are other files we need to grab.

dbungert:
Sorry that I was delayed in getting a review going.
I did spot some things, so I've sent #1797 about it.
The rest is LGTM.

cloudinit/cmd/devel/logs.py
```diff
@@ -22,11 +24,46 @@
...
+        "/var/log/installer/subiquity-server-debug.log", "SubiquityLog"
+    ),
+    ApportFile(
+        "/var/log/installer/subiquity-client-debug.log", "SubiquityClientLog"
+    ),
+    ApportFile("/var/log/installer/curtin-install.log", "CurtinLog"),
+    ApportFile(
+        "/var/log/installer/subiquity-curtin-install.cfg", "CurtinConfig"
+    ),
+    ApportFile("/var/log/installer/curtin-error-logs.tar", "CurtinError"),
```
ogayot, line 52: There have been some back and forth changes with this file recently. I first split it into multiple tar files (one for each curtin stage) and then eventually we decided to merge them back into a single file.

Sadly, we did not preserve the original filename.
The right path for the subiquity version that was shipped with the kinetic ISO would be `curtin-errors.tar` - which in Subiquity we tag as `CurtinErrors` for apport.

It's probably a good thing to expect either `curtin-errors.tar` or `curtin-error-logs.tar` to be present since the desktop canary (for instance) does not yet use the updated subiquity version.

cloudinit/cmd/devel/logs.py
```diff
@@ -22,11 +24,46 @@
...
+    ApportFile("/var/log/installer/ubuntu_desktop_installer.log", "UdiLog"),
+    ApportFile(
+        "/var/log/installer/subiquity-server-debug.log", "SubiquityLog"
+    ),
+    ApportFile(
+        "/var/log/installer/subiquity-client-debug.log", "SubiquityClientLog"
+    ),
+    ApportFile("/var/log/installer/curtin-install.log", "CurtinLog"),
+    ApportFile(
+        "/var/log/installer/subiquity-curtin-install.cfg", "CurtinConfig"
```
ogayot, line 50: With the version of Subiquity released for kinetic, this file was split into different files:

`/var/log/installer/curtin-install/subiquity-initial.conf`
`/var/log/installer/curtin-install/subiquity-partitioning.conf`
`/var/log/installer/curtin-install/subiquity-extract.conf`
`/var/log/installer/curtin-install/subiquity-curthooks.conf`

ogayot: If you would like to keep compatibility with older versions, it seems that the right name for this file was `subiquity-curtin-install.conf` instead of `.cfg` :)

cloudinit/cmd/devel/logs.py
blackboxsw: Thanks @ogayot and @dbungert for the reviews here (and the #1797) . I've addressed the remaining curtin review comments and put up the following #1812  to round out @ogayot's comments.

PR #1787, opened 2022-10-17
Enable hotplug for LXD datasource (SC-1107)

TheRealFalcon:
## Proposed Commit Message

```
Enable hotplug for LXD datasource

When a NIC appears, check for a cloud-init.network-config and apply it.
```

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -14,7 +14,7 @@ import socket
 import stat
 from enum import Flag, auto
 from json.decoder import JSONDecodeError
-from typing import Any, Dict, Union, cast
+from typing import Any, Dict, List, Optional, Union, cast
 
 import requests
 from requests.adapters import HTTPAdapter
@@ -25,6 +25,8 @@ from urllib3.connectionpool import HTTPConnectionPool
 
 from cloudinit import log as logging
 from cloudinit import sources, subp, url_helper, util
+from cloudinit.event import EventScope, EventType
+from cloudinit.net import find_fallback_nic
 
 LOG = logging.getLogger(__name__)
 
@@ -43,18 +45,8 @@ CONFIG_KEY_ALIASES = {
 }
 
 
-def generate_fallback_network_config() -> dict:
-    """Return network config V1 dict representing instance network config."""
-    network_v1: Dict[str, Any] = {
-        "version": 1,
-        "config": [
-            {
-                "type": "physical",
-                "name": "eth0",
-                "subnets": [{"type": "dhcp", "control": "auto"}],
-            }
-        ],
-    }
+def _get_fallback_interface_name() -> str:
+    default_name = "eth0"
     if subp.which("systemd-detect-virt"):
         try:
             virt_type, _ = subp.subp(["systemd-detect-virt"])
@@ -64,19 +56,44 @@ def generate_fallback_network_config() -> dict:
                 " Rendering default network config.",
                 err,
             )
-            return network_v1
+            return default_name
         if virt_type.strip() in (
             "kvm",
             "qemu",
         ):  # instance.type VIRTUAL-MACHINE
             arch = util.system_info()["uname"][4]
             if arch == "ppc64le":
-                network_v1["config"][0]["name"] = "enp0s5"
+                return "enp0s5"
             elif arch == "s390x":
-                network_v1["config"][0]["name"] = "enc9"
+                return "enc9"
             else:
-                network_v1["config"][0]["name"] = "enp5s0"
-    return network_v1
+                return "enp5s0"
+    return default_name
+
+
+def generate_network_config(
+    nics: Optional[List[str]] = None,
+) -> Dict[str, Any]:
+    """Return network config V1 dict representing instance network config."""
+    if not nics:
+        primary_nic = _get_fallback_interface_name()
+    elif len(nics) > 1:
+        primary_nic = find_fallback_nic()
+        if primary_nic not in nics:
+            primary_nic = nics[0]
+    else:
+        primary_nic = nics[0]
+
+    return {
+        "version": 1,
+        "config": [
+            {
+                "type": "physical",
+                "name": primary_nic,
+                "subnets": [{"type": "dhcp", "control": "auto"}],
+            }
+        ],
+    }
 
 
 class SocketHTTPConnection(HTTPConnection):
@@ -146,6 +163,8 @@ class DataSourceLXD(sources.DataSource):
         "user.user-data",
     )
 
+    skip_hotplug_detect = True
+
     def _is_platform_viable(self) -> bool:
         """Check platform environment to report if this datasource may run."""
         return is_platform_viable()
@@ -207,14 +226,30 @@ class DataSourceLXD(sources.DataSource):
         if self._network_config == sources.UNSET:
             if self._crawled_metadata == sources.UNSET:
                 self._get_data()
-            if isinstance(
-                self._crawled_metadata, dict
-            ) and self._crawled_metadata.get("network-config"):
-                self._network_config = self._crawled_metadata.get(
-                    "network-config", {}
-                )
-            else:
-                self._network_config = generate_fallback_network_config()
+            if isinstance(self._crawled_metadata, dict):
+                if self._crawled_metadata.get("network-config"):
+                    LOG.debug("LXD datasource using provided network config")
+                    self._network_config = self._crawled_metadata.get(
+                        "network-config", {}
+                    )
+                elif self._crawled_metadata.get("devices"):
+                    # If no explicit network config, but we have net devices
+                    # available to us, find the primary and set it up.
+                    devices: List[str] = self._crawled_metadata[
+                        "devices"
+                    ].keys()
+                    LOG.debug(
+                        "LXD datasource generating network config using "
+                        f"{len(devices)} detected devices"
+                    )
+                    self._network_config = generate_network_config(devices)
+        if self._network_config == sources.UNSET:
+            # We know nothing about network, so setup fallback
+            LOG.debug(
+                "LXD datasource generating network config using fallback."
+            )
+            self._network_config = generate_network_config()
+
         return cast(dict, self._network_config)
 
 
@@ -338,13 +373,15 @@ class _MetaDataReader:
                 md.update(self._process_config(session))
             if MetaDataKeys.DEVICES in metadata_keys:
                 url = url_helper.combine_url(self._version_url, "devices")
-                md.update({"devices": _get_json_response(session, url)})
+                md["devices"] = _get_json_response(session, url)
             return md
 
 
 def read_metadata(
     api_version: str = LXD_SOCKET_API_VERSION,
-    metadata_keys: MetaDataKeys = MetaDataKeys.CONFIG | MetaDataKeys.META_DATA,
+    metadata_keys: MetaDataKeys = MetaDataKeys.CONFIG
+    | MetaDataKeys.META_DATA
+    | MetaDataKeys.DEVICES,
 ) -> dict:
     """Fetch metadata from the /dev/lxd/socket routes.
 
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -235,9 +235,11 @@ class DataSourceLXD(sources.DataSource):
                 elif self._crawled_metadata.get("devices"):
                     # If no explicit network config, but we have net devices
                     # available to us, find the primary and set it up.
-                    devices: List[str] = self._crawled_metadata[
-                        "devices"
-                    ].keys()
+                    devices: List[str] = [
+                        k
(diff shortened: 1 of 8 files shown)


TheRealFalcon:
I removed WIP status. Assuming CI checks pass, this should be fully ready for review.

tests/unittests/cmd/devel/test_hotplug_hook.py
```diff
@@ -122,6 +124,18 @@ def test_successful_remove(self, mocks):
         mocks.m_activator.bring_up_interface.assert_not_called()
         init._write_to_cache.assert_called_once_with()
 
+    @pytest.mark.parametrize("skip", [True, False])
+    def test_skip_detected(self, skip, mocks):
```
aciba90, line 128: A super nit comment (feel free to ignore):

If I understand correctly, this relies on the fact that `event_handler.detect_hotplugged_device` raises a `ValueError` if `udevaction` is unknown and that function is only called if `skip_hotplug_detect` is false.

While that works at the moment, I find it a bit too complex for the reader to realize and also a bit fragile, as independent changes in `event_handler.detect_hotplugged_device` could break this test, e.g:

- Changing the raised exception
- The inclusion of `asdf` as a supported action (I know, very unlikely...)

Why not just mocking `event_handler.detect_hotplugged_device` and assert it is called depending on `skip_hotplug_detect`? In that way, this test would be easier to understand and more resilient.

aciba90 (approved):
Thanks, @TheRealFalcon. I left a nit comment. Aside that, LGTM.

cloudinit/sources/DataSourceLXD.py
```diff
@@ -64,19 +55,46 @@ def generate_fallback_network_config() -> dict:
...
+) -> Dict[str, Any]:
+    """Return network config V1 dict representing instance network config."""
+    if not nics:
+        primary_nic = _get_fallback_interface_name()
+    elif len(nics) > 1:
+        fallback_nic = find_fallback_nic()
+        if not fallback_nic or fallback_nic not in nics:
+            primary_nic = nics[0]
+        else:
+            primary_nic = fallback_nic
```
blackboxsw, line 84: Would this be the right place to raise awareness about needing to set `user.network-config` or `cloud-init.network-config` and cloud-init only setting up network on primary_nic?
```suggestion
            primary_nic = fallback_nic
            LOG.warning("Found multiple NICs at lxd/1.0/devices and no applicable `user.network-config` or `cloud-init.network-config` set. Only configuring DHCP on %s", primary_nic)
```

blackboxsw (changes requested):
Thanks James, this looks really good and is handles our primary deployment use-case well:
* Opt-in to hotplug w/ user-data or /etc/cloud/cloud.cfg.d
* prerequisite for `user.network-config` or `cloud-init.network-config` setting for complex multi-nic configuration

I think we need to sort:
* the ability in integration tests to add a second vNic to the bridge as it seems it is disallowed on tip of LXD main repo.
* `user.network-config` option names on bionic/focal
* maybe the unpickling handler for DataSourceLXD.skip_hottplug_detect

cloudinit/sources/DataSourceLXD.py
```diff
@@ -14,7 +14,7 @@
 import stat
 from enum import Flag, auto
 from json.decoder import JSONDecodeError
-from typing import Any, Dict, Union, cast
```
blackboxsw, line 17: I think you can also remove the `TODO ( Hotplug support using websockets API) comment in the docstr too.

cloudinit/sources/DataSourceLXD.py
TheRealFalcon: I don't think so. I'm guessing the more common LXD use case will be configuring it yourself.

TheRealFalcon:
Thanks for review. I pushed my most recent changes. We still need to sort out the LXD issues and how we're going to deal with that. Additionally, I think earlier you mentioned something about removing devices not working as expected. It'll be probably be easier to talk side-channel for those.

tests/integration_tests/datasources/test_lxd_hotplug.py
```diff
@@ -0,0 +1,108 @@
...
+                f"cloud-init.network-config={UPDATED_NETWORK_CONFIG}",
+            ]
+        )
+        assert (
+            client.read_from_file("/etc/netplan/50-cloud-init.yaml")
+            == pre_netplan
+        )
+        assert subp(
+            f"lxc config device add {client.instance.name} eth2 nic name=eth2 "
+            f"nictype=bridged parent={get_parent_network()}".split()
```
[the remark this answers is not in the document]
blackboxsw: The sentiment here is we'll comment in this integration test the related upstream commit. There is still more discussion on whether this is a viable approach or what desired alernatives are.

cloudinit/sources/DataSourceLXD.py
blackboxsw: We decided in standup: it's best really to add documentation for hotplug enablement of LXD datasource informing the user that desired network config should be set via `lxc config set user.network-config` or `cloud-init.network-config` prior to attaching/detaching multiple NICS. And absence of lxc network-config will result in dhcp on primary NIC.

blackboxsw:
> Thanks for review. I pushed my most recent changes. ... Additionally, I think earlier you mentioned something about removing devices not working as expected. It'll be probably be easier to talk side-channel for those.

We are ok here. On both device addition or removal of nics, the network-config surfaced from /dev/lxd/sock 1.0/config is what is applied in both cases. The expectation in docs that individuals should set  `(user|cloud-init).network-config` the capture their htoplug config needs and desired config prior to triggering hotplug events. LXD will document the same expectation.

blackboxsw (approved):
This changeset looks good to me. We can followup with an RTD documentation element around hotplug or LXD datasource that specifically notes that we need to set user.network-config or cloud-init.network-config when using multi-nic hot add/remove configuration.

TheRealFalcon:
@blackboxsw Thanks! I think I'll push one more comment/doc-only commit and then merge, but that shouldn't need any additional input from you.

tests/integration_tests/datasources/test_lxd_hotplug.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_no_network_change_default(
+        self, class_client: IntegrationInstance
+    ):
+        client = class_client
+        assert "eth1" not in client.execute("ip address")
+        pre_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
+        parent_network = get_parent_network(client.instance.name)
+        assert subp(
+            f"lxc config device add {client.instance.name} eth1 nic name=eth1 "
```
blackboxsw, line 68: This will break in LXD 5.7 due to commit  https://github.com/lxc/lxd/commit/56f29b48c2. 
This is because dnsmasq setups try to give out hostnames to dhcp requesters on managed networks. Each of thes now has a strict requirement to only allow on such device attached to the same hostname on that network so there aren't DNS collisions for mutliple hosts.


I think our solution here is to create a separate bridge network first for this integration test.
f`lxc network create {client.instance.name}-net --type=bridge ipv4.address=10.10.40.1/24 ipv4.nat="true"` and then attaching to that network. I'm confirming this right now.

blackboxsw: But again, let's avoid this until we have a discussion with lxd folks next week on the appropriate expectations w/ managed bridges.

tests/integration_tests/datasources/test_lxd_hotplug.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        ensure_hotplug_exited(client)
+        post_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
+        assert pre_netplan == post_netplan
+        ip_info = json.loads(client.execute("ip --json address"))
+        eth1s = [i for i in ip_info if i["ifname"] == "eth1"]
+        assert len(eth1s) == 1
+        assert eth1s[0]["operstate"] == "DOWN"
+
+    def test_network_config_applied(self, class_client: IntegrationInstance):
```
blackboxsw, line 79: These tests will fail on bionic and focal because NoCloud seed files are still written to `/var/lib/cloud/nocloud-seed` for Bionic and Focal.
To avoid that we need to create a /etc/cloud/cloud.cfg,d dropin that'll force LXD detection before NoCloud. Something like the following diff should get that in place (also not this diff tries to avoid the duplicate nics from one host attached to same managed bridge error on LXD 5.7:
```diff
diff --git a/tests/integration_tests/datasources/test_lxd_hotplug.py b/tests/integration_tests/datasources/test_lxd_hotplug.py
index b026dcc8b..6ec8d96d0 100644
--- a/tests/integration_tests/datasources/test_lxd_hotplug.py
+++ b/tests/integration_tests/datasources/test_lxd_hotplug.py
@@ -41,6 +41,20 @@ def get_parent_network(instance_name: str):
     return "lxdbr0"
 
 
+def _prefer_lxd_datasource_over_nocloud(client: IntegrationInstance):
+    """For hotplug support we need LXD datasource detected instead of NoCloud
+
+    Bionic and Focal still deliver nocloud-net seed files so override it
+    with /etc/cloud/cloud.cfg.d/99-detect-lxd-first.cfg
+    """
+    client.write_to_file(
+        "/etc/cloud/cloud.cfg.d/99-detect-lxd-first.cfg",
+        "datasource_list: [LXD, NoCloud]\n",
+    )
+    client.execute("cloud-init clean --logs")
+    client.restart()
+
+
 @pytest.mark.lxd_container
 @pytest.mark.lxd_vm
 @pytest.mark.user_data(USER_DATA)
@@ -56,20 +70,30 @@ class TestLxdHotplug:
         name = class_client.instance.name
         subp(f"lxc config device remove {name} eth1".split())
         subp(f"lxc config device remove {name} eth2".split())
+        subp(f"lxc network delete ci-test-br-eth1".split())
+        subp(f"lxc network delete ci-test-br-eth2".split())
 
     def test_no_network_change_default(
         self, class_client: IntegrationInstance
     ):
         client = class_client
+        if ImageSpecification.from_os_image().release in [
+            "bionic",
+            "focal",
+        ]:  # pyright: ignore
+            _prefer_lxd_datasource_over_nocloud(client)
         assert "eth1" not in client.execute("ip address")
         pre_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
-        parent_network = get_parent_network(client.instance.name)
-        assert subp(
-            f"lxc network create {client.instance.name}-net --type=bridge ipv4.address=10.10.41.1/24 ipv4.nat=true".split()
-        )
+
+        networks = subp("lxc network list".split())
+        if "ci-test-br-eth1" not in networks.stdout:
+            assert subp(
+                f"lxc network create ci-test-br-eth1 --type=bridge"
+                " ipv4.address=10.10.41.1/24 ipv4.nat=true".split()
+            )
         assert subp(
             f"lxc config device add {client.instance.name} eth1 nic name=eth1 "
-            f"nictype=bridged parent={parent_network}".split()
+            f"nictype=bridged parent=ci-test-br-eth1".split()
         )
         ensure_hotplug_exited(client)
         post_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
@@ -81,16 +105,17 @@ class TestLxdHotplug:
 
     def test_network_config_applied(self, class_client: IntegrationInstance):
         client = class_client
-        assert "eth2" not in client.execute("ip address")
-        pre_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
-        assert "eth2" not in pre_netplan
         if ImageSpecification.from_os_image().release in [
             "bionic",
             "focal",
         ]:  # pyright: ignore
+            _prefer_lxd_datasource_over_nocloud(client)
             top_key = "user"
         else:
             top_key = "cloud-init"
+        assert "eth2" not in client.execute("ip address")
+        pre_netplan = client.read_from_file("/etc/netplan/50-cloud-init.yaml")
+        assert "eth2" not in pre_netplan
         assert subp(
             [
                 "lxc",
@@ -104,10 +129,15 @@ class TestLxdHotplug:
             client.read_from_file("/etc/netplan/50-cloud-init.yaml")
             == pre_netplan
         )
-        parent_network = get_parent_network(client.instance.name)
+        networks = subp("lxc network list".split())
+        if "ci-test-br-eth2" not in networks.stdout:
+            assert subp(
+                f"lxc network create ci-test-br-eth2 --type=bridge"
+                " ipv4.address=10.10.42.1/24 ipv4.nat=true".split()
+            )
         assert subp(
             f"lxc config device add {client.instance.name} eth2 nic name=eth2 "
-            f"nictype=bridged parent={parent_network}".split()
+            f"nictype=bridged parent=ci-test-br-eth2".split()
         )
         ensure_hotplug_exited(client)
         post_netplan = safeyaml.load(
```

The above changeset gets a pass for me on LXD 5.7 *edge" stream and focal

blackboxsw (approved):
Integration tests needs a minor tweak for bionic focal LXD detection. other than that +1.

TheRealFalcon:
Thanks @blackboxsw . I (mostly) applied your changes along with a doc update. I'll merge if/when CI passes.

PR #1782, opened 2022-10-13
Rewrite the doc for final_message so it works for multi-line output

jherrman:
Signed-off-by: Jiří Herrmann <jherrman@redhat.com>

## Proposed Commit Message
Rewriting the doc for final_message so it works for multi-line output

```
Summary: Rewriting the foc final_message so it works for multi-line output

```

## Additional Context

If the current syntax of the final_message is followed, it ends up displaying as a single line. Adding `\n`s changes to output to multiple lines as intended.

For more info, see See also https://access.redhat.com/solutions/6184961

PR #1788, opened 2022-10-17
Add clarity to cc_final_message docs

TheRealFalcon:
## Proposed Commit Message

```
Add clarity to cc_final_message docs
```

### DIFF
--- a/cloudinit/config/cc_final_message.py
+++ b/cloudinit/config/cc_final_message.py
@@ -31,7 +31,7 @@ as well as stderr (which usually redirects to /var/log/cloud-init-output.log).
 
 Upon exit, this module writes the system uptime, timestamp, and cloud-init
 version to ``/var/lib/cloud/instance/boot-finished`` independent of any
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Old docs make it seems like `/var/lib/cloud/instance/boot-finished` is written based on this configuration, and this is not true. See #1782

holmanb (approved):
> independent of any configuration here.

Minor nit[1]/question regarding the surrounding context:

This statement is rendered into the module reference page, which contains reference to various modules that use the userdata cloud-config format. However, we don't explicitly state which config format is used on that page, though it is implied in wording of a couple of the modules.

Do you think it would make sense to further clarify what "configuration here" means for users in this context?

[1] This behavior is not independent of cloud.cfg (it can be disabled by removing this module from `cloud_final_modules` in cloud.cfg).

Either way this PR is an improvement. +1 / LGTM!

holmanb (approved):

PR #1789, opened 2022-10-17
lxd: add support for lxd preseed config (SC-1025)

blackboxsw:
## Proposed Commit Message
```
User-data now supports opaque lxd.preseed YAML
which can be provided directly to the command:
  lxd init --preseed

Complex LXD host, network, profile and storage setup
and configuration can be provided via this config entry.

For details on available configuration options and examples
https://linuxcontainers.org/lxd/docs/master/preseed/
- or -
lxd init --dump
```

## Test Steps
```bash
make deb
CLOUD_INIT_KEEP_INSTANCE=true CLOUD_INIT_PLATFORM=lxd_vm CLOUD_INIT_OS_IMAGE=jammy CLOUD_INIT_SOURCE=./cloud-init_all.deb  tox -e integration-tests tests/integration_tests/test_lxd.py::test_storage_preseed_btrfs

# also check docs
make doc
xdg-open doc/rtd_html/topics/modules.html
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -91,65 +91,65 @@ meta: MetaSchema = {
               preseed:
                 config:
                   core.https_address: 192.168.1.1:9999
-                  preseed:
                 networks:
-                - config:
-                    ipv4.address: 10.42.42.1/24
-                    ipv4.nat: true
-                    ipv6.address: fd42:4242:4242:4242::1/64
-                    ipv6.nat: true
-                  description: ""
-                  name: lxdbr0
-                  type: bridge
-                  project: default
-                  storage_pools:
+                  - config:
+                      ipv4.address: 10.42.42.1/24
+                      ipv4.nat: true
+                      ipv6.address: fd42:4242:4242:4242::1/64
+                      ipv6.nat: true
+                    description: ""
+                    name: lxdbr0
+                    type: bridge
+                    project: default
+                storage_pools:
                   - config:
                       size: 5GiB
                       source: /var/snap/lxd/common/lxd/disks/default.img
                     description: ""
                     name: default
-                    driver: {}
+                    driver: zfs
                 profiles:
-                - config: {}
-                  description: Default LXD profile
-                  devices:
-                    eth0:
-                      name: eth0
-                      network: lxdbr0
-                      type: nic
-                    root:
-                      path: /
-                      pool: default
-                      type: disk
-                  name: default
-                - config: {security.nesting: true}
-                  devices:
-                    eth0:
-                      name: eth0
-                      network: lxdbr0
-                      type: nic
-                    root:
-                      eath: /
-                      pool: default
-                      type: disk
-                  name: nested
+                  - config: {}
+                    description: Default LXD profile
+                    devices:
+                      eth0:
+                        name: eth0
+                        network: lxdbr0
+                        type: nic
+                      root:
+                        path: /
+                        pool: default
+                        type: disk
+                    name: default
+                  - config: {}
+                    security.nesting: true
+                    devices:
+                      eth0:
+                        name: eth0
+                        network: lxdbr0
+                        type: nic
+                      root:
+                        path: /
+                        pool: default
+                        type: disk
+                    name: nested
                 projects:
-                - config:
-                    features.images: true
-                    features.networks: true
-                    features.profiles: true
-                    features.storage.buckets: true
-                    features.storage.volumes: true
-                  description: Default LXD project
-                  name: default
-                - config:
-                    features.images: false
-                    features.networks: true
-                    features.profiles: false
-                    features.storage.buckets: false
-                    features.storage.volumes: false
-                  description: Limited Access LXD project
-                  name: limited
+                  - config:
+                      features.images: true
+                      features.networks: true
+                      features.profiles: true
+                      features.storage.volumes: true
+                    description: Default LXD project
+                    name: default
+                  - config:
+                      features.images: false
+                      features.networks: true
+                      features.profiles: false
+                      features.storage.volumes: false
+                    description: Limited Access LXD project
+                    name: limited
+
+
             """
         ),
     ],
@@ -227,8 +227,8 @@ def handle(
             log.warning("failed to install packages %s: %s", packages, exc)
             return
 
+    subp.subp(["lxd", "waitready", "--timeout=300"])
     if preseed_cfg:
-        subp.subp(["lxd", "waitready", "--timeout=300"])
         subp.subp(
             ["lxd", "init", "--preseed"], data=safeyaml.dumps(preseed_cfg)
         )
@@ -247,8 +247,6 @@ def handle(
             "storage_pool",
             "trust_password",
         )
-
-        subp.subp(["lxd", "waitready", "--timeout=300"])
 
         # Bug https://bugs.launchpad.net/ubuntu/+source/linux-kvm/+bug/1982780
         kernel = util.system_info()["uname"][2]
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -12,7 +12,7 @@ from textwrap import dedent
 from typing import List, Tuple
 
 from cloudinit import log as logging
-from cloudinit import subp, util
+from cloudinit import safeyaml, subp, util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema, get_meta_doc
@@ -50,6 +50,7 @@ meta: MetaSchema = {
         ),
         dedent(
             """\
+            # LXD init showcasing cloud-init's LXD config options
             lxd:
               init:
                 network_address: 0.0.0.0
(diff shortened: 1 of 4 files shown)


tests/integration_tests/modules/test_lxd.py
```diff
@@ -62,27 +137,81 @@ def validate_storage(validate_client, pkg_name, command):
...
+@pytest.mark.user_data(STORAGE_PRESEED_USER_DATA.format("btrfs"))
+def test_storage_preseed_btrfs(client):
+    validate_storage(client, "btrfs-progs", "mkfs.btrfs")
+    src_cfg = yaml.safe_load(STORAGE_PRESEED_USER_DATA.format("btrfs"))
+    preseed_cfg = src_cfg["lxd"]["preseed"]
+    validate_preseed_profiles(client, preseed_cfg)
+    validate_preseed_storage_pools(client, preseed_cfg)
+    validate_preseed_projects(client, preseed_cfg)
+
+    @pytest.mark.no_container
```
TheRealFalcon, line 189: The rest of the tests in the file aren't actually getting run because of the indentation here.

cloudinit/config/cc_lxd.py
```diff
@@ -73,6 +74,84 @@
...
+            # `lxd:preseed` config can be passed directly to the command:
+            #  lxd init --preseed
+            # See https://linuxcontainers.org/lxd/docs/master/preseed/ or
+            # run: lxd init --dump to see viable preseed YAML allowed.
+            #
+            # Preseed settings configuring the LXD daemon for HTTPS connections
+            # on 192.168.1.1 port 9999, a nested profile which allows for
+            # LXD nesting on containers and a limited project allowing for
+            # RBAC approach when defining behavior for sub projects.
+            lxd:
```
TheRealFalcon, line 90: This example has a few different errors. Can we apply a diff like this?
```diff
diff --git a/cloudinit/config/cc_lxd.py b/cloudinit/config/cc_lxd.py
index a8c137d34..d48045229 100644
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -91,65 +91,65 @@ meta: MetaSchema = {
               preseed:
                 config:
                   core.https_address: 192.168.1.1:9999
-                  preseed:
                 networks:
-                - config:
-                    ipv4.address: 10.42.42.1/24
-                    ipv4.nat: true
-                    ipv6.address: fd42:4242:4242:4242::1/64
-                    ipv6.nat: true
-                  description: ""
-                  name: lxdbr0
-                  type: bridge
-                  project: default
-                  storage_pools:
+                  - config:
+                      ipv4.address: 10.42.42.1/24
+                      ipv4.nat: true
+                      ipv6.address: fd42:4242:4242:4242::1/64
+                      ipv6.nat: true
+                    description: ""
+                    name: lxdbr0
+                    type: bridge
+                    project: default
+                storage_pools:
                   - config:
                       size: 5GiB
                       source: /var/snap/lxd/common/lxd/disks/default.img
                     description: ""
                     name: default
-                    driver: {}
+                    driver: zfs
                 profiles:
-                - config: {}
-                  description: Default LXD profile
-                  devices:
-                    eth0:
-                      name: eth0
-                      network: lxdbr0
-                      type: nic
-                    root:
-                      path: /
-                      pool: default
-                      type: disk
-                  name: default
-                - config: {security.nesting: true}
-                  devices:
-                    eth0:
-                      name: eth0
-                      network: lxdbr0
-                      type: nic
-                    root:
-                      eath: /
-                      pool: default
-                      type: disk
-                  name: nested
+                  - config: {}
+                    description: Default LXD profile
+                    devices:
+                      eth0:
+                        name: eth0
+                        network: lxdbr0
+                        type: nic
+                      root:
+                        path: /
+                        pool: default
+                        type: disk
+                    name: default
+                  - config: {}
+                    security.nesting: true
+                    devices:
+                      eth0:
+                        name: eth0
+                        network: lxdbr0
+                        type: nic
+                      root:
+                        path: /
+                        pool: default
+                        type: disk
+                    name: nested
                 projects:
-                - config:
-                    features.images: true
-                    features.networks: true
-                    features.profiles: true
-                    features.storage.buckets: true
-                    features.storage.volumes: true
-                  description: Default LXD project
-                  name: default
-                - config:
-                    features.images: false
-                    features.networks: true
-                    features.profiles: false
-                    features.storage.buckets: false
-                    features.storage.volumes: false
-                  description: Limited Access LXD project
-                  name: limited
+                  - config:
+                      features.images: true
+                      features.networks: true
+                      features.profiles: true
+                      features.storage.volumes: true
+                    description: Default LXD project
+                    name: default
+                  - config:
+                      features.images: false
+                      features.networks: true
+                      features.profiles: false
+                      features.storage.volumes: false
+                    description: Limited Access LXD project
+                    name: limited
+
+
             """
         ),
     ],
```

TheRealFalcon (changes requested):
Overall looks good. Beyond the inline comments, it might be worth adding a unit test to test that when preseed is specified the preseed subp call gets made...though...I'm not sure there's a ton of value there. Your chocie.

cloudinit/config/cc_lxd.py
blackboxsw: Thanks. Applied!

cloudinit/config/cc_lxd.py
```diff
@@ -73,6 +74,84 @@
                 domain: lxd
             """
         ),
+        dedent(
+            """\
+            # For more complex non-iteractive LXD configuration of networks,
+            # storage_pools, profiles, projects, clusters and core config,
+            # `lxd:preseed` config can be passed directly to the command:
```
holmanb, line 81: Nit: Could we clarify that this is getting passed to `stdin` of `lxd init --preseed`? In the doc here, the command `lxd init --preseed` without the pipe in the example leaves that ambiguous.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1356,6 +1358,11 @@
                   "description": "Domain to advertise to DHCP clients and use for DNS resolution."
                 }
               }
+            },
+            "preseed": {
+              "type": "object",
+              "minProperties": 1,
+              "description": "Opaque LXD preseed YAML config passed directly to lxd init --preseed. See: https://linuxcontainers.org/lxd/docs/master/preseed/ of lxd init --dump for viable config. Can not be combined with either ``lxd.init`` or ``lxd.bridge``."
```
holmanb, line 1365: Nit: "of lxd init" should be "or lxd init"





```json
"description": "Opaque LXD preseed YAML config passed directly to lxd init --preseed. See: https://linuxcontainers.org/lxd/docs/master/preseed/ of lxd init --dump for viable config. Can not be combined with either ``lxd.init`` or ``lxd.bridge``."
 ```

**Review by holmanb — COMMENTED** [review]
A couple of nits, I'll look more tomorrow.

File: cloudinit/config/schemas/schema-cloud-config-v1.json [hunk 5]
```diff
@@ -1356,6 +1358,11 @@
                   "description": "Domain to advertise to DHCP clients and use for DNS resolution."
                 }
               }
+            },
+            "preseed": {
+              "type": "object",
```
holmanb, line 1363: Invalid yaml under this key will cause failure in cloud-init schema validation (rather than in lxd) if / when we move to strict validation, which seems doesn't seem very opaque.

I think I lean towards treating opaque objects as strings for this reason. Is there some reason that we need this as a dict that I'm not thinking of?

TheRealFalcon: Multiline strings are uglier and more error prone...especially if it's yaml embedded in the string it makes it much harder to know if you got a bit of indentation off. I think I'd lean towards keeping it an object.

holmanb: I'm a little conflicted about saying we treat preseed as opaque when we actually parse it as yaml, use it, and create the config that gets passed to lxd.

I see we're inspecting storage information to ensure storage drivers are available for users, but since we are taking on responsibility for parsing and turning back into valid lxd config, I wonder if we're taking on some risk that we might not need to.

Would it make more sense to make this key a string and then parse the string into json inside the module for driver install? That would allow us to:

1. pass to lxd exactly the preseed string that was passed to us
2. avoid the schema validation issue mentioned above

Recently I was working on [something else](https://github.com/holmanb/ansible-lxd-public/blob/1062d79baa8fe0061a864752ff0783b98258d2d1/start-lxd.yml#L21) that involves an passing cloud-config value to lxd, and I personally found that using a string helps make it clear that this information is not a part of the surrounding config (via syntax highlighting and context).

holmanb (dismissed):
I left a couple of in-line comments regarding using type object vs type string in the config. Please let me know your thoughts.

cloudinit/config/schemas/schema-cloud-config-v1.json
TheRealFalcon: Ok...looking at that example, not uglier. So maybe that is the better way to go

blackboxsw: I was on the fence about this too... and we talked side-channel on this. 
One benefit of having this YAML inline as an object:
* cloud-init schema can error early on stuff that doesn't look like it'll pass minimal lxd init --preseed verification 
* cloud-init control the schema validation errors and can point people to the problem to the right place in user-data where the invalid value type occurred

The downsides of defining a very shallow schema verification as "object" in cloud-init:
* there is an implication that cloud-init is validating this schema more than it really is, any YAML object keys will pass schema validation in cloud-init, but we will still rely on schema validation and errors from the underlying lxd init --preseed cmdline to bubble up to cloud-init anyway.
* by forcing YAML object in lxd.preseed cloud-config, we tightly couplescloud-init's support to only one format of preseed input (a yaml object). If we instead decouple cloud-init completely from whatver `lxd init --preseed` expects on STDIN, cloud-init doesn't have to change our schema definition if `--preseed` grows support for non-yaml binary data or external urls or some other datastructure from stdin.

For those reasons, we agree it makes sense at the moment to make preseed a multi-line string value. If stronger use-cases are raised for an alternative or more comprehensive schema validation for lxd preseed content, we can grow them in cloud-init's schema to support either type.

TheRealFalcon:
I'm getting some Bionic failures on the integration test. I think the fact that LXD is shipped as a snap in Jammy but not Bionic is the issue:
```
2022-10-21 05:04:35,848 - util.py[DEBUG]: Running module lxd (<module 'cloudinit.config.cc_lxd' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_lxd.py'>) failed
  File "/usr/lib/python3/dist-packages/cloudinit/config/modules.py", line 247, in _run_modules
    run_name, mod.handle, func_args, freq=freq
  File "/usr/lib/python3/dist-packages/cloudinit/cloud.py", line 67, in run
  File "/usr/lib/python3/dist-packages/cloudinit/helpers.py", line 185, in run
    results = functor(*args)
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_lxd.py", line 232, in handle
    subp.subp(["lxd", "init", "--preseed"], data=preseed_str)
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 336, in subp
    stdout=out, stderr=err, exit_code=rc, cmd=args
Command: ['lxd', 'init', '--preseed']
Exit code: 1
Reason: -
Stdout: 
Stderr: Error: Failed to create storage pool 'default': Custom loop file locations are not supported
```

The btrfs version of the test has the same problem with a slightly different error:
```
Stderr: Error: Failed to create storage pool 'default': Failed to run: btrfs subvolume create /var/snap/lxd/common/lxd/disks/default.img: ERROR: not a btrfs filesystem: /var/snap/lxd/common/lxd/disks
```

tests/integration_tests/modules/test_lxd.py
```diff
@@ -22,7 +22,7 @@
     ipv4_netmask: 24
     ipv4_dhcp_first: 10.100.100.100
     ipv4_dhcp_last: 10.100.100.200
-    ipv4_nat: true
+    ipv4_nat: "true"
```
TheRealFalcon, line 25: Why the change? `cloud-init schema --system` complains about this not being a boolean (causing the log check to fail 
 in test_lxd.py::TestLxdBridge::test_bridge)

tests/integration_tests/modules/test_lxd.py
```diff
@@ -62,12 +137,52 @@ def validate_storage(validate_client, pkg_name, command):
...
+
+def validate_preseed_storage_pools(client, preseed_cfg):
+    for src_storage in preseed_cfg["storage_pools"]:
+        storage_pool = yaml.safe_load(
+            client.execute(f"lxc storage show {src_storage['name']}")
+        )
+        assert storage_pool["config"]["source"] == storage_pool["config"].pop(
+            "volatile.initial_source"
+        )
+        assert storage_pool["config"] == src_storage["config"]
```
aciba90, line 156: While testing in `kinetic`, the config for the `test_storage_preseed_zfs` case has an extra key:

```yaml
config:
  size: 5GiB
  source: /var/snap/lxd/common/lxd/disks/default.img
  volatile.initial_source: /var/snap/lxd/common/lxd/disks/default.img
  zfs.pool_name: default  # <---

```

aciba90 (changes requested):
Looks great overall, thanks. I agree with the approach of defining `preseed` as an opaque string.

cloudinit/config/cc_lxd.py
```diff
@@ -81,6 +160,43 @@
...
+            f"lxd.bridge config must be a dictionary. found a"
+            f" '{type(bridge_cfg).__name__}'",
+        )
+
+    if not isinstance(preseed_str, str):
+        errors.append(
+            f"lxd.preseed config must be a string. found a"
+            f" '{type(preseed_str).__name__}'",
+        )
+    if preseed_str and (init_cfg or bridge_cfg):
```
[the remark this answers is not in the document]
blackboxsw: Yes I think mutual exclusion in our schema is something we need to better support in schema and erroring/warnings. But, I do think we need to deal with this in a separate PR because there are a couple of places where we need to think about how we approach the schema definition and the handling or errors in that regard.

Mutual exclusion cases:
- users:ssh_authorized_keys/ssh_import_id and ssh_redirect_user
- users:snapuser vs name
- lxd:init|network vs lxd:preseed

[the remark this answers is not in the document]
blackboxsw: Added https://warthogs.atlassian.net/browse/SC-1288 for tracking

tests/unittests/config/test_cc_lxd.py
```diff
@@ -139,6 +128,12 @@ def test_no_lxd_does_nothing(self, mock_subp, m_maybe_clean):
         self.assertFalse(mock_subp.subp.called)
         self.assertFalse(m_maybe_clean.called)
 
+    @mock.patch("cloudinit.config.cc_lxd.subp")
+    def test_lxd_preseed(self, mock_subp):
+        cc = get_cloud()
+        cc.distro = mock.MagicMock()
+        cc_lxd.handle("cc_lxd", {"chad": True}, cc, self.logger, [])
```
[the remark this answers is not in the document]
blackboxsw: thanks forgot to follow through on that test.

TheRealFalcon:
@blackboxsw , I took the liberty of fixing one more thing. All tests pass for me now, so I'm going to go ahead and merge this.

TheRealFalcon (approved):

PR #1794, opened 2022-10-21
Netplan deprecated keys (SC-1050)

aciba90:
## Proposed Commit Message

```
network: Deprecate gateway{4,6} keys in network config v2

- Do not render gateway{4,6} when transforming from network config v1 to v2.
- Issue a warning if gateway{4,6} is present in network config v2.
  This warning is not issued if a passthrough to netplan is performed.

LP: #1992512
```

## Additional Context

https://netplan.io/reference#default-routes

## Checklist:

## TODO:

- [x] Increase unit test coverage of v1 configs with different device types

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -118,7 +118,6 @@ def _extract_addresses(config: dict, entry: dict, ifname, features=None):
                 new_route = {
                     "via": subnet.get("gateway"),
                     "to": "default",
-                    "metric": 100,  # do not deprioritize this route
                 }
                 routes.append(new_route)
             if "dns_nameservers" in subnet:
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -46,7 +46,7 @@ def _get_params_dict_by_match(config, match):
 def _extract_addresses(config: dict, entry: dict, ifname, features=None):
     """This method parse a cloudinit.net.network_state dictionary (config) and
        maps netstate keys/values into a dictionary (entry) to represent
-       netplan yaml. (network config v1 -> v2)
+       netplan yaml. (config v1 -> netplan)
 
     An example config dictionary might look like:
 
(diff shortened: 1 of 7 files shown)


TheRealFalcon:
@aciba90 Is draft state still accurate here? Is this PR ready for review?

aciba90:
> @aciba90 Is draft state still accurate here? Is this PR ready for review?

@TheRealFalcon, I have added extra unit-tests covering v1 netconfs with all device types. This is ready for review.

cloudinit/net/netplan.py
```diff
@@ -43,10 +43,10 @@ def _get_params_dict_by_match(config, match):
     )
 
 
-def _extract_addresses(config, entry, ifname, features=None):
+def _extract_addresses(config: dict, entry: dict, ifname, features=None):
     """This method parse a cloudinit.net.network_state dictionary (config) and
        maps netstate keys/values into a dictionary (entry) to represent
-       netplan yaml.
+       netplan yaml. (network config v1 -> v2)
```
TheRealFalcon, line 49: nit: config v1 -> netplan

TheRealFalcon (approved):
Nit: Both in the commit message and in my comment inline, I think it makes more sense to talk about v1 to netplan, rather than v1 to v2. Even though v2 is a subset of netplan, the goal is to render v1 config to netplan config, not v1 to v2.

Otherwise, looks good here! Thanks!

aciba90:
Thanks, @TheRealFalcon, for the feedback. I have corrected the commit and doc messages.

TheRealFalcon:
LGTM now...

However, I was just trying to think through the implications of the change and backwards incompatibility. While I think the risk is low, I'm wondering if it makes sense to put this change behind a feature flag and keep the old behavior for released distros. I'm wondering if there could be weird changes in behavior if somebody had already set a default route.

This might be overkill too, but I think it's at least a question worth bringing to the team.

TheRealFalcon:
@aciba90 , If I recall correctly, during the engineering sprint we discussed doing some additional testing to ensure the new default routes don't conflict or do anything unexpected if there are already existing default routes defined. Is that something you can do for this PR?

aciba90:
Definitely, I mark this PR as draft and will work on that one-off test.

TheRealFalcon:
Thanks for the additional coverage @aciba90 . The only other thing I'd like to check is if specifying a gateway results in the exact same netplan configuration pre and post this change. I think we should do that for when a default route is already specified. This doesn't have to be a test in the codebase, just an ad hoc check.

TheRealFalcon (approved):
Thanks!

PR #1796, opened 2022-10-25
Fix last reported event possibly not being sent

TheRealFalcon:
## Proposed Commit Message

```
Fix last reported event possibly not being sent

The run of every cloud-init mode is wrapped in a reporting
context manager. The final flush of events before the process exits was
happening within the context manager, however, one final event is sent
when the context manager exits. Since this event isn't subject to
waiting for event flush, cloud-init can exit before this event gets
sent.

This commit fixes this issue and also adds logging of POST data when
POSTING to a URL.
```

## Additional Context
New log format actually includes useful information now:
> 2022-10-25 16:37:56,759 - url_helper.py[DEBUG]: [0/4] open 'http://10.20.30.1:55555/new_stuff' with {'url': 'http://10.20.30.1:55555/new_stuff', 'stream': False, 'allow_redirects': True, 'method': 'POST', 'timeout': 5.0, 'headers': {'User-Agent': 'Cloud-Init/22.3'}} configuration and POST data: {"name": "init-network/config-set_hostname", "description": "config-set_hostname ran successfully", "event_type": "finish", "origin": "cloudinit", "timestamp": 1666715876.7045417, "result": "SUCCESS"}

## Checklist:


cloudinit/url_helper.py
```diff
@@ -303,11 +303,12 @@ def _cb(url):
 
             if log_req_resp:
                 LOG.debug(
-                    "[%s/%s] open '%s' with %s configuration",
+                    "[%s/%s] open '%s' with %s configuration%s",
                     i,
                     "infinite" if infinite else manual_tries,
                     url,
                     filtered_req_args,
+                    f" and POST data: {data}" if data else "",
```
blackboxsw, line 311: We may want to scrub this post data to redact secrets right this could be a security problem as we'd log all potential post `data` provided to this common function to /var/log/cloud-init.log which is world-readable.

Two known areas that may cause problems, or lead to sensitive data leaks:
Azure and GCP both send data in their requests.
cloudinit/sources/helpers/azure.py: http_with_retries 
cloudinit/sources/DataSourceGCE.py:_write_host_key_to_guest_attributes maybe?

TheRealFalcon:
I pushed an alternative for logging. Log the event data rather than the `readurl` data directly. Looks like this:
> 2022-10-27 21:17:24,062 - handlers.py[DEBUG]: Queuing POST to http://127.0.0.1:55555/, data: {'name': 'init-network/activate-datasource', 'description': 'activating datasource', 'event_type': 'start', 'origin': 'cloudinit', 'timestamp': 1666905444.0623877}

tests/integration_tests/reporting/test_webhook_reporting.py
```diff
@@ -0,0 +1,66 @@
...
+    events = [json.loads(line) for line in server_output]
+
+    # Only time this should be less is if we remove modules
+    assert len(events) > 58, events
+
+    # Assert our first and last expected messages exist
+    ds_events = [
+        e for e in events if e["name"] == "init-network/activate-datasource"
+    ]
+    assert len(ds_events) == 2  # 1 for start, 1 for stop
```
blackboxsw, line 63: Can we assert there are paired  start and finish events for each scope. I'm thinking we are occasionally seeing starts or finishes getting swallowed across bring up of a stage and teardown.
Minimally I want to assert each high level boot stage scope has bookend start/finish events:
As it is currently, it seems like we may be missing a matching start event for
```
  "name": "modules-final",
  "event_type": "finish",
```

blackboxsw (changes requested):
Also something seems to be falling over as far as network being brought up as I run through this integration test on a lxd_vm platform. It is likely something unrelated that landed in tip of main, but I haven't been able to put my finger on it yet. So, a bit more testing is needed to tease out whether network not being brought up is related to this branch  or something else in main.

I'd uploaded the failing bits (tip of main plus your branch for kinetic) to ppa:chad.smith/maas-testing 
So, we should be able to peek at VM integration test failures:
```
CLOUD_INIT_KEEP_INSTANCE=1 CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_OS_IMAGE=kinetic CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:chad.smith/maas-testing tox -e integration-tests tests/integration_tests/reporting/test_webhook_reporting.py
```

tests/integration_tests/assets/echo_server.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=echo_server
+Before=cloud-init-local.service
```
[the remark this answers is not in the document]
TheRealFalcon: Good idea, but technically the dependencies shouldn't even matter because I'm not even doing a reboot in the test and this file isn't pushed to the instance until after it is up. The only reason this file exists is because it's the only way to run a process in the background that doesn't involve more substantial changes to pycloudlib.

tests/integration_tests/reporting/test_webhook_reporting.py
TheRealFalcon: We can't assert this, because we still don't get every start event. Once that's fixed, we can make the change to this test.

blackboxsw (approved):
LGTM. I think we can iterate on the start event issues separately.
This commit message should probably also capture that it fixes:

`LP: #1993836`
in the footer of the message.

blackboxsw:
We can address https://bugs.launchpad.net/cloud-init/+bug/1992711 separately for squelched "start" events.

PR #1797, opened 2022-10-26
Realpath logfiles

dbungert:
## Proposed Commit Message
```
apport: fix some data collection failures due to symlinks

Apport screens out symlinks, so we should send realpath() normalized
paths instead.
```

## Additional Context
In the LP: #[1993011](http://pad.lv/1993011) test bug, some failures related to symlinks can be seen.  Fix that.

## Test Steps
Run `ubuntu-bug` in a Subiquity live-server ephemeral environment with the affected version of cloud-init installed.

## Checklist:

### DIFF
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -4,6 +4,8 @@
 
 """Cloud-init apport interface"""
 
+import os
+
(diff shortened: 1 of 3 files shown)


TheRealFalcon (approved):
LGTM, thanks!

PR #1806, opened 2022-10-30
udevadm_settle should handle non-udev system grace…

dermotbradley:
## Proposed Commit Message

```
udevadm_settle should handle non-udev system gracefully

Not all Linux systems use udev (and therefore udevadm). Specifically
an Alpine Linux system may have either udev (via "eudev" package)
or mdev, or mdevd installed.

Change the udev_settle function to check for the presence of the
udevadm binary before calling it - if it is not present then
silently exit the function.

This change will enable cloud-init to run on Alpine systems using
mdev/mdevd rather than udev.
```

## Checklist:

 - [ x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2933,6 +2933,10 @@ def mount_is_read_write(mount_point):
 
 def udevadm_settle(exists=None, timeout=None):
     """Invoke udevadm settle with optional exists and timeout parameters"""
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Thanks!

PR #1810, opened 2022-11-03
Auto-format schema

holmanb:
```
Add schema auto-format to tox and a CI test
```

Our schema format is currently inconsistent. Use json.tool to auto-format.

I didn't include `tox -e check_format` integration for this tool, but it would be possible if desired to use the current CI workflow instead of what I did.


TheRealFalcon (approved):
LGTM

PR #1817, opened 2022-11-08
tests: Ignore unsupported lxd project keys

aciba90:
## Proposed Commit Message

```
tests: Restrict lxd tests to applicable series

Some lxd init --preseed features are only available in specific 
lxd versions.
The use of projects within the preseed is lxd >= 4.12
and `features.storage.buckets` is lxd >= 5.5 

Adapt the tests to perform the assertions only in series that
contain the applicable lxd versions.
```

## Additional Context

https://github.com/canonical/cloud-init/pull/1789

## Test Steps

```bash
#!/bin/bash -eux

function run_test {
	local SERIES=$1
	local PLATFORM=$2
	LOG_FN="test_preseed_${PLATFORM}_${SERIES}.log"
	CLOUD_INIT_SOURCE=IN_PLACE CLOUD_INIT_PLATFORM=$PLATFORM CLOUD_INIT_OS_IMAGE=$SERIES tox -e integration-tests -- -vv tests/integration_tests/modules/test_lxd.py | tee $LOG_FN
}

for PLATFORM in lxd_container lxd_vm
do
	for SERIES in bionic focal jammy kinetic
	do
		run_test $SERIES $PLATFORM
	done
done
```

## Checklist:

### DIFF
--- a/tests/integration_tests/modules/test_lxd.py
+++ b/tests/integration_tests/modules/test_lxd.py
@@ -201,6 +201,10 @@ def validate_preseed_storage_pools(client, preseed_cfg):
 
 
 def validate_preseed_projects(client: IntegrationInstance, preseed_cfg):
+    # Support for projects by lxd init --preseed was added in lxd 4.12
+    # https://discuss.linuxcontainers.org/t/lxd-4-12-has-been-released/10424#projects-now-supported-by-lxd-init-dump-and-preseed-9
+    if client.instance.series in {"bionic", "focal"}:
+        return
     for src_project in preseed_cfg.get("projects", []):
         proj_name = src_project["name"]
         proj_result = client.execute(f"lxc project show {proj_name}")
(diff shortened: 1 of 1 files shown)


tests/integration_tests/modules/test_lxd.py
```diff
@@ -199,12 +200,28 @@ def validate_preseed_storage_pools(client, preseed_cfg):
         assert storage_pool["driver"] == src_storage["driver"]
 
 
-def validate_preseed_projects(client, preseed_cfg):
+def validate_preseed_projects(client: IntegrationInstance, preseed_cfg):
+    # Support for projects by lxd init --preseed was added in lxd 4.12
```
aciba90, line 204: Being this true, I wonder if `cloud-init` should issue some kind of warning to better inform users about this fact.

So if the user data contains a lxd preseed with project config and lxd < 4.12, then the project config is ignored by lxd and cloud-init does not notify the user.

blackboxsw: I'd lean toward filing a bug agains LXD for this. While I expect this is going to be outside of support for features LXD will grow for LXD < 4.12, it'd be good to track this shortfall in preseed validation in LXD. 

We walk a fine line in cloud-init when we "support" opaque keys that pass through direct to external software/services. I'd prefer if cloud-init doesn't grow this static knowledge about whether or not 'projects' or any other top-level config key is supported by a particular version of LXD preseed. This support of "projects" config on a the LXD 4.0-stream can change over time and cloud-init hard-coding expectations and warnings about those cases seems a bit out of scope.

TheRealFalcon (approved):
LGTM, thanks!

tests/integration_tests/modules/test_lxd.py
aciba90: Agreed, I created an LXD issue representing this: https://github.com/lxc/lxd/issues/11106

PR #1820, opened 2022-11-10
testing: Fix preseed test

TheRealFalcon:
## Proposed Commit Message

```
testing: Fix preseed test

The wrong method of determining the current Ubuntu release was used,
causing tests to fail.
```


holmanb (approved):
Nice catch.

PR #1821, opened 2022-11-10
docs: switch sphinx theme to furo (SC-1327)

aciba90:
## Proposed Commit Message

```
docs: switch sphinx theme to furo
```

## Checklist:


holmanb:
I think we need a little more than just this. I see:

image

aciba90:
Good catch, tocs are shown in a right-side panel with furo. State after dropping the toc:

image

s-makin (approved):

blackboxsw (changes requested):
Changeset look good, but before we land this PR I think we need to sort a couple of things related to style that is rendered on a few pages.

I see unterminated notes and warnings that end up categorizing unrelated content as a note or warning
: rtd/topics/format.html and rtd/topics/base_config_reference.html are the most visible examples. We might need to change our RST a bit to work better with the rendered style in furo.

blackboxsw:
Bero

blackboxsw:
unterminated_warning_style
unterminated_notes_styling

blackboxsw:
Also we may want to provide a dark-mode themed image and set configuration to switch images between themes.
```diff
diff --git a/doc/rtd/conf.py b/doc/rtd/conf.py
index d81cca3b2..7975c4ae8 100644
--- a/doc/rtd/conf.py
+++ b/doc/rtd/conf.py
@@ -64,9 +64,14 @@ show_authors = False
 # a list of builtin themes.
 html_theme = "furo"
 
+
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
-html_logo = "static/logo.png"
+html_static_path = ["static"]
+html_theme_options = {
+    "light_logo": "logo.png",
+    "dark_logo": "logo-dark-mode.png",
+}
 
```

 here's a 5 minute attempt with gimp to get a white background optioon for dark-mode theme
https://github.com/aciba90/cloud-init/compare/rdt_furo...blackboxsw:cloud-init:rdt_furo

blackboxsw:
Note this rendering problem seems to be a version problem when built on focal versus jammy. I can see clean rendering of Notes/Warnings etc on jammy. I can also see it on focal if I adapt some of the doc-requirements.txt to align with what black project uses https://github.com/psf/black/blob/main/docs/requirements.txt

aciba90:
Thanks, @blackboxsw, for the review.

I have migrated the deprecated [sphinx-panels](https://sphinx-panels.readthedocs.io/en/latest/) to [sphinx-design](https://sphinx-design.readthedocs.io/en/latest/get_started.html#migrating-from-sphinx-panels) because pining that dependency forced us to pin others.

After that, we can unpin every doc dependency and let pip select versions across Ubuntu series. I have tested it in Focal and Jammy.

Additionally, I have included your dark logo.

aciba90:
> A quick grep of the codebase for directives that `sphinx-design` doesn't support shows that we weren't using any of them.

Yes, AFAICT, the only `sphinx-panels` | `sphinx-design` directive that we use is `dropdown` under [doc/rtd/topics/datasources/configdrive.rst](https://github.com/canonical/cloud-init/blob/461d2631010e54cab1ccf26a331c708796dae06a/doc/rtd/topics/datasources/configdrive.rst).

blackboxsw:
>> I'm tempted to merge it since I think it's ready, but I'll leave that for @blackboxsw since he has a -1 revew on it currently.

I'm also tempted on the PR merge. I regenerated docs on my side on focal and see the correct rendering without a CSS failure as before.
This looks great. thanks for fixing the rendering issues. We may have to pin versions again at somepoint in doc-requirements.txt to avoid being broken by some new version of a dependency on the internet. But, let's deal with that only if we encounter this problem again.

updated_screenshot

blackboxsw (approved):
+1. We can pin versions in doc-requirements.txt if needed in the future. Thanks for the cleanup

PR #1822, opened 2022-11-10
ansible: s/-/_/g

holmanb:
```
ansible: standardize schema keys

The cloud-config schema definition standardizes on underscores
instead of hyphenated key names. This makes for standard config keys
and reduces the burden of supporting aliases for either hyphenated or
underscore-delimited keys.
```
Note:
-----
The `.replace("_", "-")` is because there was previously almost[1] a 1:1 mapping between schema keys and ansible arguments.

[1] "almost" because boolean values are indicated be the absence or presence of a `--$FLAGNAME`

### DIFF
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -224,7 +224,11 @@ def validate_config(cfg: dict):
 
 def filter_args(cfg: dict) -> dict:
     """remove boolean false values"""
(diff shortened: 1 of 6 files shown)


blackboxsw (approved):
Approved though you have a couple of conflicts to resolve in our static schema now. I've updated the PR description to represent the reason we want to standardize JSON schema on underscores.

PR #1824, opened 2022-11-10
tests: replace missed ansible install-method with underscore

blackboxsw:
## Proposed Commit Message

```
tests: replace ansible install-method with underscore

Also support test runs in environments without pip
```

## Checklist:

### DIFF
--- a/tests/unittests/config/test_cc_ansible.py
+++ b/tests/unittests/config/test_cc_ansible.py
@@ -17,6 +17,13 @@ from cloudinit.config.schema import (
 from tests.unittests.helpers import skipUnlessJsonSchema
 from tests.unittests.util import get_cloud
 
+try:
+    import pip as _pip  # type: ignore # noqa: F401
+
+    HAS_PIP = True
+except ImportError:
+    HAS_PIP = False
+
 M_PATH = "cloudinit.config.cc_ansible."
 distro_version = dedent(
     """ansible 2.10.8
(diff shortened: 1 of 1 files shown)


holmanb:
I think that the failure that this attempts to address is due to `cc_ansible.py:135` driving bootstrap behavior, which means that a test environment without `pip` will have more calls (1) than an environment with pip (no bootstrap required).

We could do something like this to test according to the test environment: 
```diff
  diff --git a/tests/unittests/config/test_cc_ansible.py b/tests/unittests/config/test_cc_ansible.py
  index 3048d9e57..6f840658b 100644
  --- a/tests/unittests/config/test_cc_ansible.py
  +++ b/tests/unittests/config/test_cc_ansible.py
  @@ -1,4 +1,5 @@
   import re
  +import importlib
   from copy import deepcopy
   from os import environ
   from textwrap import dedent
  @@ -298,6 +299,7 @@ class TestAnsible:
           else:
               cloud = get_cloud(mocked_distro=True)
               install = cfg["ansible"]["install_method"]
  +            has_pip = importlib.find_loader("pip")
               cc_ansible.handle("", cfg, cloud, None, None)
               if install == "distro":
                   cloud.distro.install_packages.assert_called_once()
  @@ -305,7 +307,10 @@ class TestAnsible:
                       "ansible-core"
                   )
               elif install == "pip":
  -                assert 0 == cloud.distro.install_packages.call_count
  +                if has_pip:
  +                    assert 0 == cloud.distro.install_packages.call_count
  +                else:
  +                    assert 1 == cloud.distro.install_packages.call_count
   
       @mock.patch(M_PATH + "which", return_value=False)
       def test_deps_not_installed(self, m_which):
```

Perhaps a better fix would be to mock the import, though this test is already heavily mocked.

Thoughts?

blackboxsw:
@holmanb thanks the addressing this. I took you suggestion just a bit farther to assert that we are actually trying to install the distro.pip_package_name for that install_packages call.

holmanb (approved):
Looks like we need to drop the unused import for flake8, but otherwise looks good to me.

Thanks!

PR #1832, opened 2022-11-15
Release 22.4

TheRealFalcon:


holmanb (commented):
Since 22.3+ changes never got merged back into the Changelog on main,  we have a few commits that have already been included in 22.3.1-22.3.4, which might cause confusion:
```
  diff --git a/ChangeLog b/ChangeLog
  index 3f6419990..16d58d3a8 100644
  --- a/ChangeLog
  +++ b/ChangeLog
  @@ -116,23 +116,6 @@
    - testing: Fix test regression in test_combined (#1713) [Alberto Contreras]
    - Identify Huawei Cloud as OpenStack (#1689) [huang xinjie]
    - doc: add reporting suggestion to FAQ (SC-1236) (#1698)
  - - Fix v2 interface matching when no MAC (LP: #1986551)
  - - test: reduce number of network dependencies in flaky test (#1702)
  - - docs: publish cc_ubuntu_autoinstall docs to rtd (#1696)
  - - net: Fix EphemeraIPNetwork (#1697) [Alberto Contreras]
  - - test: make ansible test work across older versions (#1691)
  - - Networkd multi-address support/fix (#1685) [Teodor Garzdin]
  - - make: drop broken targets (#1688)
  - - net: Passthough v2 netconfigs in netplan systems (#1650)
  -   [Alberto Contreras] (LP: #1978543)
  - - NM ipv6 connection does not work on Azure and Openstack (#1616)
  -   [Emanuele Giuseppe Esposito]
  - - Fix check_format_tip (#1679) [Alberto Contreras]
  - - DataSourceVMware: fix var use before init (#1674)
  -   [Andrew Kutz] (LP: #1987005)
  - - rpm/copr: ensure RPM represents new clean.d dir artifacts (#1680)
  - - test: avoid centos leaked check of /etc/yum.repos.d/epel-testing.repo
  -   (#1676)
   
   22.3
    - sources: obj.pkl cache should be written anyime get_data is run (#1669)
  diff --git a/cloudinit/version.py b/cloudinit/version.py
  index b9b42af33..ec2621f75 100644
```

TheRealFalcon:
Thanks @holmanb , updated accordingly.

TheRealFalcon:
@holmanb , updated for real this time :smile:

blackboxsw:
> Since 22.3+ changes never got merged back into the Changelog on main, we have a few commits that have already been included in 22.3.1-22.3.4, which might cause confusion:

Yeah I had tossed around the idea of representing 22.3.3 vs 22.3.4 public security releases but that's not something we can specifically capture unless we manually tease them out of the inline time-ordered commits on main. I think given that we also released a 22.3-13 with a number of other commits that will make this changelog hard to track.  I think generally we shouldn't be representing interim releases between our major.minor in this Changelog as it does add that complexity of 'what did we release from side-branches or hot-fixes type question' which I think it better represented to hotfix bugs or d/changelog of those releases.

blackboxsw (approved):
LGTM, assuming we agree on avoiding manual documentation of the individual hot-fixes that went into 22.3.3 and 22.3.4

holmanb (approved):
Send it :)

PR #1853, opened 2022-11-16
net: skip duplicate mac check for netvsc nic and its VF

anhvoms:
## Proposed Commit Message
net: Skip duplicate mac check for netvsc nic and its VF
```
summary: 
When accelerated network is enabled on Azure, the host presents
two network interfaces with the same mac address to the VM: 
a synthetic nic (netvsc) and a VF nic, which is enslaved to the synthetic 
nic. 

The net module is already excluding slave nics when enumerating
interfaces. However, if cloud-init starts enumerating after the kernel
makes the VF visible to userspace, but before the enslaving has finished,
cloud-init will see two nics with duplicate mac.

This change will skip the duplicate mac error if one of the two nics 
with duplicate mac is a netvsc nic

LP: #1844191

## Checklist:
that apply. -->

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1000,15 +1000,32 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
 
     Bridges and any devices that have a 'stolen' mac are excluded."""
     ret: dict = {}
+    driver_map: dict = {}
     for name, mac, _driver, _devid in get_interfaces(
         blacklist_drivers=blacklist_drivers
     ):
         if mac in ret:
-            raise RuntimeError(
-                "duplicate mac found! both '%s' and '%s' have mac '%s'"
-                % (name, ret[mac], mac)
-            )
+            raise_error = True
+            # Hyper-V netvsc driver will register a VF with the same mac
+            #
+            # The VF will be enslaved to the master nic shortly after
+            # registration. If cloud-init starts enumerating the interfaces
+            # before the completion of the enslaving process, it will see
+            # two different nics with duplicate mac. Cloud-init should ignore
+            # the slave nic (which does not have hv_netvsc driver).
+            if _driver != driver_map[mac]:
+                if driver_map[mac] == "hv_netvsc":
+                    continue
+                if _driver == "hv_netvsc":
+                    raise_error = False
+
+            if raise_error:
+                raise RuntimeError(
+                    "duplicate mac found! both '%s' and '%s' have mac '%s'"
+                    % (name, ret[mac], mac)
+                )
         ret[mac] = name
+        driver_map[mac] = _driver
 
         # Pretend that an Infiniband GUID is an ethernet address for Openstack
         # configuration purposes
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1001,11 +1001,16 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
     Bridges and any devices that have a 'stolen' mac are excluded."""
     ret: dict = {}
     driver_map: dict = {}
-    for name, mac, _driver, _devid in get_interfaces(
+    for name, mac, driver, _devid in get_interfaces(
         blacklist_drivers=blacklist_drivers
     ):
         if mac in ret:
-            raise_error = True
+            raise_duplicate_mac_error = True
+            msg = "duplicate mac found! both '%s' and '%s' have mac '%s'." % (
+                name,
+                ret[mac],
+                mac,
+            )
             # Hyper-V netvsc driver will register a VF with the same mac
             #
             # The VF will be enslaved to the master nic shortly after
@@ -1013,19 +1018,27 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
             # before the completion of the enslaving process, it will see
             # two different nics with duplicate mac. Cloud-init should ignore
             # the slave nic (which does not have hv_netvsc driver).
-            if _driver != driver_map[mac]:
+            if driver != driver_map[mac]:
                 if driver_map[mac] == "hv_netvsc":
+                    LOG.warning(
+                        msg + " Ignoring '%s' due to driver '%s' and "
+                        "'%s' having driver hv_netvsc."
+                        % (name, driver, ret[mac])
+                    )
                     continue
-                if _driver == "hv_netvsc":
-                    raise_error = False
+                if driver == "hv_netvsc":
+                    raise_duplicate_mac_error = False
+                    LOG.warning(
+                        msg + " Ignoring '%s' due to driver '%s' and "
+                        "'%s' having driver hv_netvsc."
+                        % (ret[mac], driver_map[mac], name)
+                    )
 
-            if raise_error:
-                raise RuntimeError(
-                    "duplicate mac found! both '%s' and '%s' have mac '%s'"
-                    % (name, ret[mac], mac)
-                )
+            if raise_duplicate_mac_error:
+                raise RuntimeError(msg)
+
         ret[mac] = name
-        driver_map[mac] = _driver
+        driver_map[mac] = driver
 
         # Pretend that an Infiniband GUID is an ethernet address for Openstack
         # configuration purposes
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cloudinit/net/__init__.py [hunk 1]
```diff
@@ -1000,15 +1000,32 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
...
+            # Hyper-V netvsc driver will register a VF with the same mac
+            #
+            # The VF will be enslaved to the master nic shortly after
+            # registration. If cloud-init starts enumerating the interfaces
+            # before the completion of the enslaving process, it will see
+            # two different nics with duplicate mac. Cloud-init should ignore
+            # the slave nic (which does not have hv_netvsc driver).
+            if _driver != driver_map[mac]:
+                if driver_map[mac] == "hv_netvsc":
+                    continue
```
cjp256, line 1018: consider scenario wif we're considering hv_netvsc after mlx, we would want to overwrite the mlx

tests/unittests/test_net.py
```diff
@@ -7693,14 +7695,23 @@ class TestGetInterfacesByMac(CiTestCase):
             "bridge1-nic": "aa:aa:aa:aa:aa:03",
             "lo": "00:00:00:00:00:00",
             "greptap0": "00:00:00:00:00:00",
+            "netvsc0-vf": "aa:aa:aa:aa:aa::04",
```
cjp256, line 7698:
```suggestion
            "netvsc0-vf": "aa:aa:aa:aa:aa:04",
            "netvsc0": "aa:aa:aa:aa:aa:04",
```

tests/unittests/test_net.py
```diff
@@ -7759,6 +7771,7 @@ def test_excludes_stolen_macs(self):
                 "aa:aa:aa:aa:aa:02": "enp0s2",
                 "aa:aa:aa:aa:aa:03": "bridge1-nic",
                 "00:00:00:00:00:00": "lo",
+                "aa:aa:aa:aa:aa::04": "netvsc0",
```
cjp256: Perhaps add a test case for the reverse ordering (device list with hv_netvsc first, and another with it second)?

anhvoms: Good idea. Added the suggested test. I also added another test for two netvsc nics with same mac would still raise an error.

blackboxsw (changes requested):
Thanks for this @anhvoms and @cjp256 . While I don't really prefer us surfacing driver-specifics to handle this race condition in the common net code. I think you've contained it well to the device driver for which we are exposed to this race. Also, this sort of aligns with complexities of infiniband device subordinate interfaces have to be ignored with duplicate MACs are  discovered. I'm poking around at whether there is anything else we can look at in sysfs files to understand if there might be some information exposed for a device is targeted for a given driver, but the driver has not yet finished setup as this may also give us an opportunity to ignore the subordinate device.   For the moment, here's a diff I was thinking would add a bit more logging for this failure path so we can at least track that it happened when ignoring the duplicate MACs.

```diff
diff --git a/cloudinit/net/__init__.py b/cloudinit/net/__init__.py
index 353e4ab57..82181a4ca 100644
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1001,11 +1001,16 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
     Bridges and any devices that have a 'stolen' mac are excluded."""
     ret: dict = {}
     driver_map: dict = {}
-    for name, mac, _driver, _devid in get_interfaces(
+    for name, mac, driver, _devid in get_interfaces(
         blacklist_drivers=blacklist_drivers
     ):
         if mac in ret:
-            raise_error = True
+            raise_duplicate_mac_error = True
+            msg = "duplicate mac found! both '%s' and '%s' have mac '%s'." % (
+                name,
+                ret[mac],
+                mac,
+            )
             # Hyper-V netvsc driver will register a VF with the same mac
             #
             # The VF will be enslaved to the master nic shortly after
@@ -1013,19 +1018,25 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
             # before the completion of the enslaving process, it will see
             # two different nics with duplicate mac. Cloud-init should ignore
             # the slave nic (which does not have hv_netvsc driver).
-            if _driver != driver_map[mac]:
+            if driver != driver_map[mac]:
                 if driver_map[mac] == "hv_netvsc":
+                    LOG.warning(
+                        msg
+                        + " Ignoring '%s' due to driver '%s'." % (name, driver)
+                    )
                     continue
-                if _driver == "hv_netvsc":
-                    raise_error = False
-
-            if raise_error:
-                raise RuntimeError(
-                    "duplicate mac found! both '%s' and '%s' have mac '%s'"
-                    % (name, ret[mac], mac)
-                )
+                if driver == "hv_netvsc":
+                    raise_duplicate_mac_error = False
+                    LOG.warning(
+                        msg
+                        + " Ignoring '%s' due to driver '%s'."
+                        % (ret[mac], driver_map[mac])
+                    )
+
+            if raise_duplicate_mac_error:
+                raise RuntimeError(msg)
         ret[mac] = name
-        driver_map[mac] = _driver
+        driver_map[mac] = driver
 
         # Pretend that an Infiniband GUID is an ethernet address for Openstack
         # configuration purposes
```

If you think there is a good way to add more comprehensive logging of this event, please adapt this as you see fit.

tests/unittests/test_net.py
```diff
@@ -7693,14 +7697,27 @@ class TestGetInterfacesByMac(CiTestCase):
...
+            "netvsc0": "aa:aa:aa:aa:aa:04",
+            "netvsc1-vf": "aa:aa:aa:aa:aa:05",
+            "netvsc1": "aa:aa:aa:aa:aa:05",
             "tun0": None,
         },
+        "drivers": {
+            "netvsc0": "hv_netvsc",
+            "netvsc0-vf": "foo",
+            "netvsc1": "hv_netvsc",
+            "netvsc1-vf": "bar",
```
holmanb, line 7710: Is this verifying that the `-vf` interfaces doesn't cause existing calls to `get_interfaces_by_mac()` to throw an exception?

holmanb: Initially I assumed you forgot to push another test commit.

anhvoms: @holmanb  Yep, that is correct. The existing tests will provide coverage with the added test data.

blackboxsw (commented):
My initial comment about avoiding a specific driver check here was because I was thinking about the potential of searching for the presence of something like `/sys/class/net/<devname>/master` link to indicate that we are a subordinate device, and if so, emit a duplicate mac warning instead of an error, ignoring it is a subordinate device.   But, given that this is a race condition in early boot, I inspected the presence of said "master" link and that does seem to get created by the kernel later than the device/driver file. So, it's probably best to stick with driver check as we know that'll show up a couple seconds earlier in boot.
```
lrwxrwxrwx  1 root root    0 2022-11-18 01:46:17.767825388 +0000 device -> ../../../5007:00:02.0
...
lrwxrwxrwx  1 root root    0 2022-11-18 01:46:19.599825444 +0000 master -> ../../../../../00224847-0e64-0022-4847-0e6400224847/net/eth0
```

So, all said. Let's keep this specific check to ensure we are including the hv_netvsc device and ignoring the mellanox devices in this timeframe.


I've pushed this content to a ppa:chad.smith/cloud-init-uploads and I'm running through tests on this to ensure we don't have degraded behavior.

blackboxsw:
Changeset looks good to me as-is. Initial testing on Azure trying to hit a reproducer on this race has been hard. But running through general cloud integration test run and other cloud's full integration test suites don't seem to show any degradation in behavior here. I'll await the full integraion test run with this PPA I mentioned above before approving tomorrow morn.

blackboxsw:
Integration test runs on Azure/GCP/Ec2 succeed on my side without degreding behavior. I can +1 this content and our team can pull together a release into Ubuntu Lunar for testing. Plan is the sync this into Bionic, Focal, Jammy and Kinetic in our next viable SRU window. This content will be available for testing in the meantime via `add-apt-repository ppa:cloud-init-dev/daily -y` on stable Ubuntu releases or on latest Ubuntu Lunar images built after Nov 18th.

blackboxsw (approved):
Integration test runs on Azure/GCP/Ec2 succeed on my side without degreding behavior. I can +1 this content and our team can pull together a release into Ubuntu Lunar for testing. Plan is the sync this into Bionic, Focal, Jammy and Kinetic in our next viable SRU window. This content will be available for testing in the meantime via `add-apt-repository ppa:cloud-init-dev/daily -y` on stable Ubuntu releases or on latest Ubuntu Lunar images built after Nov 18th.

PR #1858, opened 2022-11-18
Ubuntu/kinetic 22.4.x

TheRealFalcon:
Bug template has not been updated yet, so do not request review from SRU vanguard yet.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-cloud-init (22.4.2-0ubuntu0~22.10.1) UNRELEASED; urgency=medium
+cloud-init (22.4.2-0ubuntu0~22.10.1) kinetic; urgency=medium
 
   * Upstream snapshot based on 22.4.2 upstream release. (LP: #1996645)
     - List of changes from upstream can be found at
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (22.4.2-0ubuntu0~22.10.1) UNRELEASED; urgency=medium
+
+  * Upstream snapshot based on 22.4.2 upstream release. (LP: #1996645)
+    - List of changes from upstream can be found at
+      https://raw.githubusercontent.com/canonical/cloud-init/22.4.2/ChangeLog
+
+ -- James Falcon <james.falcon@canonical.com>  Mon, 28 Nov 2022 10:45:45 -0600
+
 cloud-init (22.4-0ubuntu0~22.10.1) kinetic; urgency=medium
 
   * d/control: drop python3-httpretty from Build-Depends
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+22.4.2
+ - status: handle ds not defined in status.json (#1876) (LP: #1997559)
+
 22.4.1
  - net: skip duplicate mac check for netvsc nic and its VF (#1853)
    [Anh Vo] (LP: #1844191)
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+22.4.1
+ - net: skip duplicate mac check for netvsc nic and its VF (#1853)
+   [Anh Vo] (LP: #1844191)
+ - ChangeLog: whitespace cleanup (#1850)
+ - changelog: capture 22.3.1-4 releases
+
 22.4
  - test: fix pro integration test [Alberto Contreras]
  - cc_disk_setup: pass options in correct order to utils (#1829)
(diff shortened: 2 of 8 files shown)


debian/changelog
```diff
@@ -1,3 +1,11 @@
+cloud-init (22.4.1-0ubuntu0~22.10.1) kinetic; urgency=medium
```
blackboxsw, line 1: We'll need to sort the actual package version. In this case, we released `22.4-0ubuntu4` to lunar, so we can't release a version `22.4.1-0ubuntu0~22.10.1` to kinetic because  pkg versioning won't be an upgrade path
```bash
dpkg --compare-versions 22.4-0ubuntu4 gt 22.4.1-0ubuntu0~22.10.1 || echo "NOT A VALID UPGRADE"
NOT A VALID UPGRADE
```

blackboxsw: Release of 22.4.1-0ubuntu1 has been published to lunar. So we are good on versioning now.

blackboxsw (approved):
LGTM: 
- package version is good in d/changelog for kinetic 22.4.1-0ubuntu0~22.10.1
- content only contains 22.4.1 commit/changes
- build-package and sbuild are happy w/out new lintian errors/warnings

Holding on landing this PR until our SRU of 22.4 is approved and published.

debian/changelog
```diff
@@ -1,3 +1,19 @@
...
+
+  * Upstream snapshot based on 22.4.2 upstream release. (LP: #1997559)
+    - List of changes from upstream can be found at
+      https://raw.githubusercontent.com/canonical/cloud-init/22.4.2/ChangeLog
+
+ -- James Falcon <james.falcon@canonical.com>  Mon, 28 Nov 2022 10:45:45 -0600
+
+cloud-init (22.4.1-0ubuntu0~22.10.1) kinetic; urgency=medium
+
+  * Upstream snapshot based on 22.4.1 upstream release. (LP: #1844191)
```
blackboxsw, line 11: I think this bug should be LP: #1996645 for both of these d/changelog entries (22.4.2 and 22.4.1)  since we have an SRU process bug. Because we didn't officially release beyond -proposed. We don't need a separate process LP bug for 22.4.2

blackboxsw: Because of our SRU exception, we don't generally track individual bugs in d/changelog but rather our SRU process bug LP: #1996645. This bug should be reflected in both 22.4.1 and 22.4.2 or just in 22.4.2, but it's your call. Maybe in this case (because of regression detection during SRU verification it is worth itemizing the unique Bugs fixed in this release).

Here's a diff that aligns with our upstream publish in lunar that captures both our SRU process bug and the unique bugs fixed for which we can perform supplemental validation if needed.

```diff
 diff --git a/debian/changelog b/debian/changelog
index 5f2c734d..28e24c2a 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,16 +1,19 @@
 cloud-init (22.4.2-0ubuntu0~22.04.1) jammy; urgency=medium
 
-  * Upstream snapshot based on 22.4.2 upstream release. (LP: #1997559)
+  * Upstream snapshot based on 22.4.2 upstream release. (LP: #1996645)
     - List of changes from upstream can be found at
       https://raw.githubusercontent.com/canonical/cloud-init/22.4.2/ChangeLog
+    - Bugs fixed in this snapshot: (LP: #1997559)
+
 
  -- James Falcon <james.falcon@canonical.com>  Mon, 28 Nov 2022 10:32:24 -0600
 
 cloud-init (22.4.1-0ubuntu0~22.04.1) jammy; urgency=medium
 
-  * Upstream snapshot based on 22.4.1 upstream release. (LP: #1844191)
+  * Upstream snapshot based on 22.4.1 upstream release. (LP: #1996645)
     - List of changes from upstream can be found at
       https://raw.githubusercontent.com/canonical/cloud-init/22.4.1/ChangeLog
+    - Bugs fixed in this snapshot: (LP: #1844191)
 
  -- James Falcon <james.falcon@canonical.com>  Fri, 18 Nov 2022 20:22:28 -0600
 
```

blackboxsw (approved):
LGTM:
 - change set in this PR is representative of only 22.4.1 and 22.4.2 content
 - only authorshi
 - d/changelog version good for kinetic `(22.4.2-0ubuntu0~22.10.1) kinetic`
 - d/changelog content points to proper upstream tagged ChangeLog for release 
 - build-package sbuild success without any new lintian warnings/errors
 - no d/patches to validate

PR #1865, opened 2022-11-21
feat: add support for aliyun metadata security harden mode

ManassehZhou:
## Proposed Commit Message

```
feat: add support Aliyun metadata security hardening mode

currently, Alibaba cloud provides a security hardening mode for its metadata server, which is alike IMDSv2, and we should support it.

Detailed information be found here: https://www.alibabacloud.com/help/en/elastic-compute-service/latest/view-instance-metadata#concept-dwj-y1x-wgb
```

## Checklist:

 - [x] I have updated or added any documentation accordingly (not suitable, since no document need to be changed)

### DIFF
--- a/cloudinit/sources/DataSourceAliYun.py
+++ b/cloudinit/sources/DataSourceAliYun.py
@@ -18,6 +18,11 @@ class DataSourceAliYun(EC2.DataSourceEc2):
     min_metadata_version = "2016-01-01"
     extended_metadata_versions: List[str] = []
 
+    # Aliyun metadata server security enhanced mode overwrite
+    @property
+    def imdsv2_token_put_header(self):
+        return "X-aliyun-ecs-metadata-token"
+
     def get_hostname(self, fqdn=False, resolve_ip=False, metadata_only=False):
         hostname = self.metadata.get("hostname")
         is_default = False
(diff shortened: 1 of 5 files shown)


TheRealFalcon (changes requested):
@ManassehZhou , Thanks for the PR. I don't like the idea of re-implementing entire methods only to change 1 or 2 lines.

One of the only changes in each method is:
```python
        if self.cloud_name != EC2.CloudNames.ALIYUN:
            return {}
```

It seems this check exists to ensure these methods don't run on EC2 derivatives that don't support them. Since Aliyun now does support this, can we instead change it to:
```python
        if self.cloud_name not in [CloudNames.AWS, CloudNames.ALIYUN]:
            return {}
```

Similarly, for `AWS_TOKEN_PUT_HEADER` vs  `ALIYUN_TOKEN_PUT_HEADER`, can we instead move these token related constants into the DataSourceEc2 class and make the name slightly more generic? Then DataSourceAliYun can override them instead of having to create entirely new method definitions around them.

Finally, the CLA checker is case sensitive. Instead of `manassehzhou`, please add `ManassehZhou` to the `.github-cla-signers` file.

ManassehZhou:
ok, let me change that

TheRealFalcon (approved):
LGTM, thanks for the contribution!

PR #1870, opened 2022-11-21
Update read-version

TheRealFalcon:
## Proposed Commit Message

```
Update read-version

Use 'git describe <branch>' as the version number, even if the version
differs from what is in version.py. This means that if the most recent
commit is a tag, we'll get the tag number, otherwise we'll also show
the number of commits since the last tag on the branch.

Additionally, did some cleanup and typing fixes on the script.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,13 @@ def in_virtualenv():
 def get_version():
     cmd = [sys.executable, "tools/read-version"]
     ver = subprocess.check_output(cmd)
-    return ver.decode("utf-8").strip()
+    version = ver.decode("utf-8").strip()
+    # read-version can spit out something like 22.4-15-g7f97aee24
+    # which is invalid under PEP440. If we replace the first - with a +
+    # that should give us a valid version.
+    if "-" in version:
+        version = version.replace("-", "+", 1)
+    return version
 
 
 def read_requires():
--- a/tools/read-version
+++ b/tools/read-version
@@ -5,10 +5,10 @@ import json
 import subprocess
 import sys
 
-if "avoid-pep8-E402-import-not-top-of-file":
-    _tdir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
-    sys.path.insert(0, _tdir)
-    from cloudinit import version as ci_version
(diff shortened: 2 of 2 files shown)


tools/read-version
```diff
@@ -133,8 +119,6 @@ data = {
 if output_json:
     sys.stdout.write(json.dumps(data, indent=1) + "\n")
 else:
-    sys.stdout.write(release + "\n")
+    sys.stdout.write(version + "\n")
```
blackboxsw, line 122: I think we want to leave this as release (which will be Major.Minor) so that we don't break python sdist builds, which call tools/read-version directly as well.

Otherwise we hit the sdist wheel errors as seen in CI on this branch


```
  WARNING: Built wheel for cloud-init is invalid: Metadata 1.2 mandates PEP 440 version, but '22.4-16-g4f7e8b9a' is not
```

blackboxsw (changes requested):
Needs fixing to handle python sdist versioning issues and generating proper MM.NN.[1-
9] on upstream/22.4.x branches before annotated tags are pushed. Diff supplied that I think will work for us.

tools/read-version
TheRealFalcon: I pushed another possibility for for the sdist builds to replace the first '-' with a '+'. Do you think that would work?

setup.py
```diff
@@ -73,7 +73,13 @@ def in_virtualenv():
...
     cmd = [sys.executable, "tools/read-version"]
     ver = subprocess.check_output(cmd)
-    return ver.decode("utf-8").strip()
+    version = ver.decode("utf-8").strip()
+    # read-version can spit out something like 22.4-15-g7f97aee24
+    # which is invalid under PEP440. If we replace the first - with a +
+    # that should give us a valid version.
+    if "-" in version:
+        version = version.replace("-", "+", 1)
+    return version
```
blackboxsw, line 82: Good add. python sdist should know how to parse our local version output and your "+" aligns with [PEP 440 [+<local version label>]](https://peps.python.org/pep-0440/#local-version-identifiers)

blackboxsw (approved):
LGTM

PR #161, opened 2020-01-09
import sysvinit patches from freebsd-ports tree

igalic:
bugzilla ref: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=224361
svn rev: https://svnweb.freebsd.org/ports?view=revision&revision=457768

PR #1875, opened 2022-11-23
FreeBSD init: use cloudinit_enable as only rcvar

igalic:
## Proposed Commit Message

```
FreeBSD init: use cloudinit_enable as only rcvar

all components of cloudinit need to run, and in a specific order.
If cloudinit is to be enabled, it should only rely on one variable.
This change better encodes that, than #161 

```

## Additional Context
Currently, when cloud-init isn't enabled, some of its components issue rc warnings.
Previously, in https://github.com/canonical/cloud-init/pull/161 we went overboard, by giving each its own variable

## Test Steps
install this patch, reboot, see no more warnings.

## Checklist:

### DIFF
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -21,8 +21,8 @@ cloudconfig_start()
 	${command} modules --mode config
 }
 
-load_rc_config $name
+load_rc_config 'cloudinit'
 
-: ${cloudconfig_enable="NO"}
+: ${cloudinit_enable="NO"}
 
 run_rc_command "$1"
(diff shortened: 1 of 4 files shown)


igalic:
cc @Jehops

goneri (approved):

Jehops:
We don't want `load_rc_config $rcvar`.  This will look to source files from places like `/etc/rc.conf.d/cloudinit_enable/*`.

igalic:
> We don't want `load_rc_config $rcvar`. This will look to source files from places like `/etc/rc.conf.d/cloudinit_enable/*`.

oh, right, so, `load_rc_config $name` or `load_rc_config cloudinit` would be better here

igalic:
updated the PR now.

TheRealFalcon:
LGTM, but I'll wait for an approval from @Jehops since BSD isn't my wheelhouse.

Jehops:
LGTM.  Thanks @igalic.

TheRealFalcon (approved):
Thanks!

PR #1876, opened 2022-11-23
status: handle ds not defined in status.json

aciba90:
## Proposed Commit Message

```
status: handle ds not defined in status.json

Handles any situation where `status.json` does not yet contain
datasource information, by gracefully fulfilling a 
`datasource=None` in `StatusDetails`.

LP: #1997559
```

## Additional Context

https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1997559

## Test Steps

Execute:

```sh
lxc launch --ephemeral ubuntu-daily:lunar test
lxc exec test -- cloud-init status --wait
```

I noticed the error does not arise 100% of times, therefore there is some kind of race condition or non-deterministic behavior happening when using the ephemeral option on LXD.

This PR handles any situation where `status.json` does contain `{"v1": {"datasource": null}}`, in general, by gracefully fulfilling a `datasource=None` in `StatusDetails`.

## Checklist:

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -13,11 +13,12 @@ import json
 import os
 import sys
 from time import gmtime, sleep, strftime
-from typing import Any, Dict, List, NamedTuple, Tuple, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
 
 from cloudinit import safeyaml
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.distros import uses_systemd
+from cloudinit.helpers import Paths
 from cloudinit.util import get_cmdline, load_file, load_json
 
 CLOUDINIT_DISABLED_FILE = "/etc/cloud/cloud-init.disabled"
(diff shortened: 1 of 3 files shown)


blackboxsw (approved):
+1 LGTM.

And yes this is a small race window when `/run/cloud-init/status.json` does not yet exist that the status command generates it's own [empty status dict with datasource: None](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/main.py#L751)
So, this only happens very early in boot, and we also need to include this fix into our pending 22.4 SRU because we can't risk regression of installers that use this on  bionic ,focal, jammy or kinetic.

blackboxsw:
@ogayot thx for the bug. FYI this should get in shortly. We can fix lunar with a quick upload for this changeset and we'll make sure to also pull this into our current pending 22.4 SRU to avoid regression there.

ogayot:
Thank you for the quick fix!

aciba90:
Thank you, @ogayot, for finding the regression. We are preparing a release for lunar: https://github.com/canonical/cloud-init/pull/1880 .

PR #1808, opened 2022-11-01
Support string replacement for nocloud seed path

ITJamie:
## Proposed Commit Message

```
summary: Support string replacement for nocloud seed path

allows the use of serial numbers / asset tags as a variable in seed 
paths .It will replace `_chassis-asset-tag_` or 
`_system-serial-number_` with the related system data.

LP: #1994980 
```

## Checklist:

 - [ ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #1879, opened 2022-11-24
nocloud: add support for dmi variable expansion for seedfrom URL

blackboxsw:
## Proposed Commit Message
```
nocloud: add support for dmi variable expansion for seedfrom URL

NoCloud meta-data seedfrom (or kernel commandline seedfrom) URL can
now provide variable expansion for system-specific DMI values as seen
in /sys/class/dmi/id on Linux or kenv on FreeBSD platforms.

Variable names of the format __dmi.SOME_VAR__ will be replaced when
determining the URL from which NoCloud datasource GETs its user-data
and meta-data.

This allows for a common templated seedfrom URL which can be reused
for mass deployments, but can allow for unique URLs based on classes
of DMI system characteristics such as chassis serial, product name,
UUID etc.

LP: #1994980
```

## Checklist:

### DIFF
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -183,22 +183,13 @@ def read_dmi_data(key: str) -> Optional[str]:
     return None
 
 
-def _get_dmi_keys():
-    """Return a list of valid DMI keys which can be used by sub_dmi_vars"""
-    attr_name = "freebsd" if is_FreeBSD() else "linux"
-    return list(DMIDECODE_TO_KERNEL.keys()) + [
-        getattr(kernelname, attr_name)
-        for kernelname in DMIDECODE_TO_KERNEL.values()
-    ]
-
-
 def sub_dmi_vars(src: str) -> str:
     """Replace %dmi.VARNAME% with DMI values from either sysfs or kenv."""
     if "%" not in src:
         return src
-    matches = re.findall(r"\%dmi\.([-_.\w]+)\%", src)
+    matches = re.findall(r"\%dmi\.([-.\w]+)\%", src)
+    valid_dmi_keys = DMIDECODE_TO_KERNEL.keys()
     if matches:
-        valid_dmi_keys = _get_dmi_keys()
         for match in matches:
             if match not in valid_dmi_keys:
                 LOG.warning(
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -202,12 +202,12 @@ def sub_dmi_vars(src: str) -> str:
             dmi_value = read_dmi_data(match)
             if not dmi_value:
                 dmi_value = ""
-                LOG.debug(
-                    "Replacing $dmi.%s in '%s' with '%s'.",
-                    match,
-                    src,
-                    dmi_value,
-                )
+            LOG.debug(
+                "Replacing $dmi.%s in '%s' with '%s'.",
+                match,
+                src,
+                dmi_value,
+            )
             src = src.replace(f"%dmi.{match}%", dmi_value)
     return src
 
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -184,16 +184,16 @@ def read_dmi_data(key: str) -> Optional[str]:
 
 
 def sub_dmi_vars(src: str) -> str:
-    """Replace %dmi.VARNAME% with DMI values from either sysfs or kenv."""
-    if "%" not in src:
+    """Replace __dmi.VARNAME__ with DMI values from either sysfs or kenv."""
+    if "__" not in src:
         return src
-    matches = re.findall(r"\%dmi\.([-.\w]+)\%", src)
+    matches = re.findall(r"__dmi\.([^_]+)__", src)
     valid_dmi_keys = DMIDECODE_TO_KERNEL.keys()
     if matches:
         for match in matches:
             if match not in valid_dmi_keys:
                 LOG.warning(
-                    "Ignoring invalid %%dmi.%s%% in %s. Expected one of: %s.",
+                    "Ignoring invalid __dmi.%s__ in %s. Expected one of: %s.",
                     match,
                     src,
                     valid_dmi_keys,
@@ -203,12 +203,12 @@ def sub_dmi_vars(src: str) -> str:
             if not dmi_value:
                 dmi_value = ""
             LOG.debug(
-                "Replacing $dmi.%s in '%s' with '%s'.",
+                "Replacing __dmi.%s__ in '%s' with '%s'.",
                 match,
                 src,
                 dmi_value,
             )
-            src = src.replace(f"%dmi.{match}%", dmi_value)
+            src = src.replace(f"__dmi.{match}__", dmi_value)
     return src
 
 
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -187,28 +187,26 @@ def sub_dmi_vars(src: str) -> str:
     """Replace __dmi.VARNAME__ with DMI values from either sysfs or kenv."""
     if "__" not in src:
         return src
-    matches = re.findall(r"__dmi\.([^_]+)__", src)
     valid_dmi_keys = DMIDECODE_TO_KERNEL.keys()
-    if matches:
-        for match in matches:
-            if match not in valid_dmi_keys:
-                LOG.warning(
-                    "Ignoring invalid __dmi.%s__ in %s. Expected one of: %s.",
-                    match,
-                    src,
-                    valid_dmi_keys,
-                )
-                continue
-            dmi_value = read_dmi_data(match)
-            if not dmi_value:
-                dmi_value = ""
-            LOG.debug(
-                "Replacing __dmi.%s__ in '%s' with '%s'.",
+    for match in re.findall(r"__dmi\.([^_]+)__", src):
+        if match not in valid_dmi_keys:
+            LOG.warning(
+                "Ignoring invalid __dmi.%s__ in %s. Expected one of: %s.",
                 match,
                 src,
-                dmi_value,
+                valid_dmi_keys,
             )
-            src = src.replace(f"__dmi.{match}__", dmi_value)
+            continue
+        dmi_value = read_dmi_data(match)
+        if not dmi_value:
+            dmi_value = ""
+        LOG.debug(
+            "Replacing __dmi.%s__ in '%s' with '%s'.",
+            match,
+            src,
+            dmi_value,
+        )
+        src = src.replace(f"__dmi.{match}__", dmi_value)
     return src
 
 
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -1,5 +1,6 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 import os
+import re
 from collections import namedtuple
 from typing import Optional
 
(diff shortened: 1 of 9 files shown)


blackboxsw:
Based on the original intent and discussion brought up by @ITJamie in #1808 .  Added some unittests and restructured logic to live over in cloudinit.dmi

ITJamie:
One thing i noticed, We probably want to add a note about when this feature was added in docs. Since it might cause confusion when people use older images that dont have support for it. 

Otherwise LGTM

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
 ``file://``.  With ``ds=nocloud-net``, the ``seedfrom`` value must start
-with ``http://`` or ``https://``.
+with ``http://`` or ``https://`` and end with a trailing ``/``.
+
+Cloud-init performs variable expansion of the seedfrom URL for any DMI kernel
+variables present in ``/sys/class/dmi/id`` or FreeBSD's kenv.
+Your ``seedfrom`` URL can contain variable names of the format `:dmi.varname:`
+to indicate to cloud-init NoCloud datasource that variable expansion to the
+actual value of the DMI system attribute is wanted. The leading and trailing
+colon `:` before and after any `:dmi.<varname>:` are used because other
```
TheRealFalcon, line 42: I think we can remove this sentence. It's a design detail that isn't really relevant to using it. If you decide to keep this sentence, `have` is misspelled on the following line and we have the same issue of docs say `:` but implementation uses `%`

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
 With ``ds=nocloud``, the ``seedfrom`` value must start with ``/`` or
 ``file://``.  With ``ds=nocloud-net``, the ``seedfrom`` value must start
-with ``http://`` or ``https://``.
+with ``http://`` or ``https://`` and end with a trailing ``/``.
+
+Cloud-init performs variable expansion of the seedfrom URL for any DMI kernel
+variables present in ``/sys/class/dmi/id`` or FreeBSD's kenv.
+Your ``seedfrom`` URL can contain variable names of the format `:dmi.varname:`
+to indicate to cloud-init NoCloud datasource that variable expansion to the
+actual value of the DMI system attribute is wanted. The leading and trailing
```
TheRealFalcon, line 41: nit: `actual` does add anything here.

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -167,6 +167,9 @@ def _pp2d_callback(mp, data):
             if not seedfound:
                 LOG.debug("Seed from %s not supported by %s", seedfrom, self)
                 return False
+            # check and replace instances of known dmi.<dmi_keys> such as
+            # chass-board, serial number etc.
```
TheRealFalcon, line 171: nit: chassis-board?

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
 
 With ``ds=nocloud``, the ``seedfrom`` value must start with ``/`` or
 ``file://``.  With ``ds=nocloud-net``, the ``seedfrom`` value must start
-with ``http://`` or ``https://``.
+with ``http://`` or ``https://`` and end with a trailing ``/``.
```
TheRealFalcon, line 35: Is this true? Why do we need a trailing '/'?

If so, we should update https://github.com/canonical/cloud-init/blob/main/doc/examples/cloud-config-datasources.txt#L34 accordingly.

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
+
+Cloud-init performs variable expansion of the seedfrom URL for any DMI kernel
+variables present in ``/sys/class/dmi/id`` or FreeBSD's kenv.
+Your ``seedfrom`` URL can contain variable names of the format `:dmi.varname:`
```
TheRealFalcon, line 39: Docs say `:` but implementation uses `%`

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
+
+or append the following GRUB kernel commandline on your Ubuntu Live server/desktop installer
 
 ::
 
-  -smbios type=1,serial=ds=nocloud-net;s=http://10.10.0.1:8000/
+  autostart ds=nocloud-net;s=http://10.10.0.1:8000/%dmi.chassis-serial-number%/
 
-to cause NoCloud to fetch the full meta-data from http://10.10.0.1:8000/meta-data
+to cause NoCloud to fetch the full meta-data for YOUR_SERIAL_NUMBER as seen in `/sys/class/dmi/id/chassis_serial_number` or kenv on FreeBSD from http://10.10.0.1:8000/YOUR_SERIAL_NUMBER/meta-data
```
TheRealFalcon, line 112: I'd put (or kenv on FreeBSD) in parentheses.

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
+
+
+e.g. you can pass this option to QEMU
 
-e.g. you can pass this option to QEMU:
+::
+
+  -smbios type=1,serial=ds=nocloud-net;s=http://10.10.0.1:8000/%dmi.chassis-serial-number%/
+
+or append the following GRUB kernel commandline on your Ubuntu Live server/desktop installer
```
TheRealFalcon, line 106: I'm wary of putting installer-specific info in cloud-init docs

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
+    - dmi.smbios.system.serial
+  * - dmi.system-uuid
+    - dmi.product_uuid
+    - dmi.smbios.system.uuid
+  * - dmi.system-version
+    - dmi.product_version
+    - dmi.smbios.system.version
+
+
+e.g. you can pass this option to QEMU
```
TheRealFalcon, line 100: nit: Not your doing, but I think `For example` would be better here.

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,15 +32,84 @@ The permitted keys are:
...
+actual value of the DMI system attribute is wanted. The leading and trailing
+colon `:` before and after any `:dmi.<varname>:` are used because other
+metacharacters would hae to be escaped to avoid Grub interpreting those chars.
+Typically you want to use cloud-init's distro-agnostic alias
+
+.. list-table:: Available DMI variables for expansion in ``seedfrom`` URL
+  :widths: 25 25 50
+  :header-rows: 1
+
+  * - Distro-agnostic DMI string
```
TheRealFalcon, line 50: Nice formatting.

What does `Distro-agnostic` mean in this context? Linux distro? If it's linux distro, then what's the difference between this column and the `Linux-specific` column? If it's encompassing of FreeBSD, then why do we need multiple columns instead of just the `Distro-agnostic` list?

cloudinit/dmi.py
```diff
@@ -182,4 +183,42 @@ def is_x86(arch):
     return None
 
 
+def _get_dmi_keys():
```
TheRealFalcon, line 186: Related to my doc table comment, is there a reason we can't just match on keys and skip the values entirely?

cloudinit/dmi.py
```diff
@@ -182,4 +183,42 @@ def is_x86(arch):
...
+                    "Ignoring invalid %%dmi.%s%% in %s. Expected one of: %s.",
+                    match,
+                    src,
+                    valid_dmi_keys,
+                )
+                continue
+            dmi_value = read_dmi_data(match)
+            if not dmi_value:
+                dmi_value = ""
+                LOG.debug(
```
TheRealFalcon, line 214: I think this LOG is supposed to be outdented a level.

TheRealFalcon (changes requested):
Nice work. I left some comments inline.

TheRealFalcon:
Oh, also docs linting doesn't like a few line lengths.

doc/rtd/topics/datasources/nocloud.rst
ITJamie: This is true. Its caught me out before

doc/rtd/topics/datasources/nocloud.rst
blackboxsw: dropped.

doc/rtd/topics/datasources/nocloud.rst
blackboxsw: @TheRealFalcon we probably don't. I was documenting them for reference. All aliases are allowed for those familiar with either the naming of the sysfs dmi file IDs or the kenv vars. I didn't want to preclude use of those keys because they will inevitably may be more familiar than the DMI standard strings presented by `dmidecode -s`.  Maybe this is too much to document though or maybe we should actually limit the aliases supported to only the standardized DMI string names.
What do you think?

holmanb:
There is a possibility that this breaks someone out there using a similar URL scheme in their current deployments, right?

For example if someone is currently expecting that passing in `http://10.10.0.1:8000/%dmi.chassis-serial-number%/` will cause a request to `/%dmi.chassis-serial-number%` and not whatever that expands to under this new scheme, then they will have been broken in a way that would require server-side changes I think.

It would be a weird coincidence, but we have no way of knowing someone isn't doing this.

There is the case for encoding / special characters in URLs and the fact that this PR hasn't settled on whether to use `%` or `:`, each of which has special meaning in URLs, but I do think that even with encoding and limitations that a collision is still possible.

doc/rtd/topics/datasources/nocloud.rst
blackboxsw: As discussed in IRC. We don't really need/want distro-specific options as DMI standard strings will provide an alias that works on both FreeBSD and Linux. Folks can look to `dmidecode -s` for references of keys on their systems regardless of *BSD or Linux.

doc/rtd/topics/datasources/nocloud.rst
blackboxsw: I've added a commit to this PR to auto-append the missing / if absent in seedfrom urls (which don't already contain a `%s`).

cloudinit/dmi.py
blackboxsw: Dropped, we'll just use DMI standard keys.

igalic:
in an environment like libvirt, how would i set these variables for testing this code?

ITJamie:
So i just spoke with a friend in another company today about this. Auto adding the / is going to break some existing usecases. 

His seedfrom url is something like
`Http://some.tld/seedmaker?template=somename&filename=`

In this case adding a slash automatically is going to at minimum need him to update his code to strip a prefixing slash from the filename but is effectively a breaking change.

igalic:
> I guess we could add a check to see if the last character is an `=` and if so dont add the slash?

or more generally: is there's a query string in the URI, there's no need to append a `/`

blackboxsw:
> There is a possibility that this breaks someone out there using a similar URL scheme in their current deployments, right?
> 
> For example if someone has deployments that pass in `http://10.10.0.1:8000/%dmi.chassis-serial-number%/` with the expectation that this requests to `http://10.10.0.1:8000/%dmi.chassis-serial-number%/meta-data` and not whatever that expands to under this new scheme, then they will have been broken in a way that would require server-side changes I think.
> 
> It would be a weird coincidence, but we have no way of knowing someone isn't doing this.
> There is the case for encoding / special characters in URLs and the fact that this PR hasn't settled on whether to use `%` or `:`, each of which has special meaning in URLs, but I do think that even with encoding and limitations that a collision is still possible.

Thanks @holmanb for the thoughtful suggestion here. 

Through the course of this PR, we ended up dropping support for other kenv keys which happened to contain underscores `_` in the key names, I think I can revert the bookend variable expansion characters from `%` to `__` which was closer to @ITJamie 's first suggestion because the supported characters of  our dmi variable are now limited to `[.-a-z]` which means:
  1. `__` is a both URL-safe and a more unique unique marker as the end or beginning and end of a `__dmi.varname__`.
  2.  It can then be used without any negative impact to url escaping or unescaping, python's %-formatting in cloud-init functions and GRUB-related custom variable expansion

By making this a dunder marker instead of just a single character. I think we've probably insulated ourselves from most usefuls corner cases for use of a `seedfrom` URL. But, if we still think this is a potential regression risk, we can add  RELEASE_BLOCKER comment and avoid this nocloud behavior on stable releases. I have added the RELEASE_BLOCKER comment around the automatic append of a trailing forward slash because I feel that's probably more likely to have more cases in the wild that we have to be wary of.

doc/rtd/topics/datasources/nocloud.rst
```diff
@@ -32,18 +32,50 @@ The permitted keys are:
...
+    - ``dmi.system-uuid``
+    - ``dmi.system-version``
+
+
+For example, passing this option to QEMU
 
 ::
 
-  -smbios type=1,serial=ds=nocloud-net;s=http://10.10.0.1:8000/
+  -smbios type=1,serial=ds=nocloud-net;s=http://10.10.0.1:8000/_dmi.chassis-serial-number_/
```
TheRealFalcon, line 72: This example won't actually work, right? Where's the double underscore?

TheRealFalcon:
Left two inline comments, but overall LGTM. If we're going to patch out the trailing slash in our SRUs, I'll plug going the feature flag route again, which I think would make the most sense to add in this PR.

TheRealFalcon (commented):
Since rtd is faling anyway, one small nit left that somehow didn't get posted yesterday.

TheRealFalcon (approved):

PR #1886, opened 2022-11-29
Autoinstall clarification

holmanb:
```
autoinstall: clarify docs for users
```
Add autoinstall pointer to subiquity docs to FAQ.

Remove cc_ubuntu_autoinstall from docs module definitions, since it doesn't give users actionable information.

Leave the module definition intact in the code, since it does add context that may be valuable for developers.

The net affect of this is PR that future searches of the website for "autoinstall" will lead users to the FAQ (and then to subiquity docs), which makes sense because we don't have autoinstall-specific documentation in the cloud-init docs.

### DIFF
--- a/doc/rtd/topics/faq.rst
+++ b/doc/rtd/topics/faq.rst
@@ -206,12 +206,12 @@ Since cloud-init ignores top level userdata cloud-config keys, other projects
 such as Juju and subiquity_ autoinstaller use a yaml-formatted config
 that combines cloud-init's userdata cloud-config yaml format with their custom
 yaml keys. These combined yaml configurations may be valid cloud-config files,
-however keys such as autoinstall, preruncmd, and postruncmd are ignored by
-cloud-init. If you are unsure whether cloud-init uses a specific key,
-you may view cloud-init's schema_ or documentation_.
+however keys such as autoinstall, preruncmd, and postruncmd are not used by
+cloud-init to configure anything.
 
 Please direct bugs and questions related to the autoinstaller, Juju, and
 other similar projects to their respective support channels.
+
 
 How can I make a module run on every boot?
 ==========================================
@@ -442,5 +442,3 @@ Whitepapers:
 .. _cloud-init Summit 2018: https://powersj.io/post/cloud-init-summit18/
 .. _cloud-init Summit 2017: https://powersj.io/post/cloud-init-summit17/
 .. _subiquity: https://ubuntu.com/server/docs/install/autoinstall
-.. _schema: https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schemas/schema-cloud-config-v1.json
-.. _documentation: https://cloudinit.readthedocs.io/en/latest/topics/modules.html
--- a/doc/rtd/topics/faq.rst
+++ b/doc/rtd/topics/faq.rst
@@ -206,7 +206,7 @@ Since cloud-init ignores top level userdata cloud-config keys, other projects
 such as Juju and subiquity_ autoinstaller use a yaml-formatted config
 that combines cloud-init's userdata cloud-config yaml format with their custom
 yaml keys. These combined yaml configurations may be valid cloud-config files,
-however keys such as autoinstall, preruncmd, and postruncmd are not used by
+however keys such as ``autoinstall``, ``preruncmd``, and ``postruncmd`` are not used by
 cloud-init to configure anything.
 
 Please direct bugs and questions related to the autoinstaller, Juju, and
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
LGTM, but lets get a +1 from @blackboxsw too

doc/rtd/topics/faq.rst
```diff
@@ -200,6 +200,19 @@ systemd, please make sure to include the following logs.
         $ systemctl --failed
 
 
+Autoinstall, Preruncmd, Postruncmd
+==================================
+Since cloud-init ignores top level userdata cloud-config keys, other projects
+such as Juju and subiquity_ autoinstaller use a yaml-formatted config
+that combines cloud-init's userdata cloud-config yaml format with their custom
+yaml keys. These combined yaml configurations may be valid cloud-config files,
```
blackboxsw, line 208: YAML should be caps in docs I think.

doc/rtd/topics/faq.rst
```diff
@@ -200,6 +200,19 @@ systemd, please make sure to include the following logs.
...
+however keys such as autoinstall, preruncmd, and postruncmd are not used by
```
blackboxsw, line 209: Let's decorate these a bit as they are reserved keys:
```suggestion
however keys such as ``autoinstall``, ``preruncmd``, and ``postruncmd`` are not used by
```

doc/rtd/topics/faq.rst
```diff
@@ -200,6 +200,19 @@ systemd, please make sure to include the following logs.
         $ systemctl --failed
 
 
+Autoinstall, Preruncmd, Postruncmd
+==================================
+Since cloud-init ignores top level userdata cloud-config keys, other projects
+such as Juju and subiquity_ autoinstaller use a yaml-formatted config
```
blackboxsw, line 206: It it worth us linking to Juju's model config docs too? https://discourse.charmhub.io/t/model-config-cloudinit-userdata/512

doc/rtd/topics/faq.rst
```diff
@@ -200,6 +200,19 @@ systemd, please make sure to include the following logs.
...
+that combines cloud-init's userdata cloud-config yaml format with their custom
+yaml keys. These combined yaml configurations may be valid cloud-config files,
+however keys such as autoinstall, preruncmd, and postruncmd are not used by
+cloud-init to configure anything.
+
+Please direct bugs and questions related to the autoinstaller, Juju, and
```
blackboxsw, line 212: Shall we link to the support channels here? For Juju maybe https://discourse.charmhub.io/  and for autoinstaller either #subiquity channel on IRC or #ubuntu-server maybe. We should check on that with subiquity devs.
@dbungert any preferences on what mediums autoinstall questions get steered?

blackboxsw (changes requested):
Thanks @holmanb for making this a clearer line in documentation. Let's round out the references to other projects so folks go to the right place/channel to seek answers.
minor nits otherwise

doc/rtd/topics/faq.rst
s-makin: +1 to that

doc/rtd/topics/faq.rst
s-makin: +1 to linking the support channels

s-makin (approved):
Overall LGTM, I'm assuming that Chad will add his approval after the links are added, but I'll add mine now.

doc/rtd/topics/faq.rst
dbungert: There isn't a #subiquity channel today on IRC and based on the average feedback I have seen on IRC I don't plan to start one.  Others are certainly welcome.

Discourse and #ubuntu-server are the way to go today.

holmanb: +1 Thanks, I've added that to this PR.

doc/rtd/topics/faq.rst
```diff
@@ -151,9 +151,6 @@ provided to the system:
 
     $ cloud-init schema --system --annotate
 
-As launching instances in the cloud can cost money and take a bit longer,
```
holmanb, line 154: I removed this since it doesn't make sense where it is. I think this may have previously preceded the section titled "How can I test cloud-init locally before deploying to the cloud?", but I don't think it adds anything there, so I propose deleting it.

holmanb:
Thanks @blackboxsw and @s-makin for the reviews. I updated the links, and deleted some lost text.

Ready for review.

blackboxsw (approved):
+1 thanks for this @holmanb

PR #1887, opened 2022-11-30
Append derivatives to is_rhel list in cloud.cfg.tmpl

nazunalika:
## Proposed Commit Message
```
summary: Append derivatives to is_rhel list in cloud.cfg.tmpl

This commit adds Rocky Linux, AlmaLinux, CloudLinux, EuroLinux, Miracle
Linux, and Virtuozzo to the is_rhel list. Recent downstream patch from
Red Hat causes issues with RHEL derivatives with the cloud.cfg template,
which leads to derivatives having to make small changes to bring back
expected functionality.
```

## Additional Context
In RHEL 8.7 and 9.1, and thus Rocky Linux 8.7 and 9.1, cloud-init was rebased to 22.1. In this update, Red Hat backported #1431. In CentOS Stream 8 and 9, they also backported #1639. Unfortunately, this caused some issues with our cloud images as the cloud configuration is now missing some pieces that were once there that would work the same RHEL or CentOS Stream. We had to create a downstream patch in our own package to address a bug report we received for the issue. This PR addresses not just Rocky Linux, but other derivatives as they are likely to run into the same issues.

We received a bug report for this at our bug tracker: https://bugs.rockylinux.org/view.php?id=1189

## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -260,7 +260,7 @@ system_info:
      name: ec2-user
      lock_passwd: True
      gecos: EC2 Default User
-{% elif is_rhel and variant not in ["eurolinux"] %}
+{% elif variant in ["rhel", "centos"] %}
      name: cloud-user
      lock_passwd: true
      gecos: Cloud User
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -3,7 +3,8 @@
 # The top level settings are used as module
 # and base configuration.
 {% set is_bsd = variant in ["dragonfly", "freebsd", "netbsd", "openbsd"] %}
-{% set is_rhel = variant in ["rhel", "centos"] %}
+{% set is_rhel = variant in ["almalinux", "centos", "cloudlinux", "eurolinux",
+                             "miraclelinux", "rhel", "rocky", "virtuozzo" ] %}
 {% if is_bsd %}
 syslog_fix_perms: root:wheel
 {% elif variant in ["suse"] %}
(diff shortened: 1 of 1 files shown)


config/cloud.cfg.tmpl
```diff
@@ -252,15 +252,15 @@ system_info:
          primary: http://ports.ubuntu.com/ubuntu-ports
          security: http://ports.ubuntu.com/ubuntu-ports
    ssh_svcname: ssh
-{% elif variant in ["almalinux", "alpine", "amazon", "arch", "cloudlinux", "eurolinux",
-                    "fedora", "gentoo", "miraclelinux", "openEuler", "openmandriva", "rocky", "suse", "virtuozzo"] or is_rhel %}
+{% elif variant in ["alpine", "amazon", "arch", "fedora",
```
TheRealFalcon, line 255: By changing this, aren't we changing the default user for all of these distros? Should we instead just change line 263 to say `{% elif variant in ["rhel", "centos"] %}`?

TheRealFalcon (changes requested):
Left a comment inline, but otherwise looks good.

config/cloud.cfg.tmpl
nazunalika: For line 255 and 256, I only moved the Enterprise Linux distributions from the list and tacked them on to the `is_rhel` list at the top. So that should be operating basically the same just as how it is now.

For line 263: I think you're right. Setting it to `{% elif variant in ["rhel", "centos"] %}` is probably cleaner here.

nazunalika:
Thank you so much for the feedback! I made a change based on your suggestion to line 263 and it seems all tests have passed. On the changes for line 255/256, we actually did this on our downstream package and it presented no issues (we had to rebuild our cloud images the other day which prompted this PR, we've not seen any issues and our bug report was addressed).

However, I am open to undoing those lines. Please let me know what you think! Thank you again!

config/cloud.cfg.tmpl
TheRealFalcon: Ah right, this line was a red herring. Thanks for the update on 263.

TheRealFalcon (approved):
LGTM, thanks!

PR #1890, opened 2022-12-05
Docs copy button

s-makin:
## Proposed Commit Message
```
Docs: Adding "copy-code" functionality

- Uses the sphinx-copybutton extension
- I have reviewed existing code blocks to separate out
  input and output lines for user convenience (so they 
  don't end up copying output as well). Didn't work in 
  every case.

```

### DIFF
--- a/doc/rtd/topics/analyze.rst
+++ b/doc/rtd/topics/analyze.rst
@@ -19,10 +19,10 @@ The analyze command requires one of the four subcommands:
 
 .. code-block:: shell-session
 
-  $ cloud-init analyze blame
-  $ cloud-init analyze show
-  $ cloud-init analyze dump
-  $ cloud-init analyze boot
+   $ cloud-init analyze blame
+   $ cloud-init analyze show
+   $ cloud-init analyze dump
+   $ cloud-init analyze boot
 
 Availability
 ============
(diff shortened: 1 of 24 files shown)


doc/rtd/topics/analyze.rst
```diff
@@ -43,117 +43,121 @@ execution.
...
-      00.00000s (modules-final/config-landscape)
-      00.00000s (modules-final/config-chef)
-      00.00000s (modules-config/config-snap_config)
-      00.00000s (modules-config/config-ntp)
-      00.00000s (modules-config/config-disable-ec2-metadata)
-      00.00000s (init-network/setup-datasource)
-
-  1 boot records analyzed
+   $ cloud-init analyze blame
+
```
holmanb, line 47: Back to back codeblocks with a command and example output looks a bit odd. Would it make sense to add the following text in between to help make it clear?

```
Example Output:
```

doc/rtd/topics/analyze.rst
```diff
@@ -43,117 +43,121 @@ execution.
...
-  |`->config-package-update-upgrade-install ran successfully @16.87600s +00.00100s
-  ...
-  |`->config-final-message ran successfully @16.93700s +00.00500s
-  |`->config-power-state-change ran successfully @16.94300s +00.00000s
-  Finished stage: (modules-final) 00.10300 seconds
-
-  Total Time: 4.60100 seconds
-
-  1 boot records analyzed
+    $ cloud-init analyze show
```
holmanb, line 125: All other subcommands in this section have separate example output and command. For consistency could we do the same here?

doc/rtd/topics/analyze.rst
```diff
@@ -167,85 +171,88 @@ consumed for other reporting needs. Each element in the list is a boot entry.
...
-  {
-    "description": "running modules for final",
-    "event_type": "finish",
-    "name": "modules-final",
-    "origin": "cloudinit",
-    "result": "SUCCESS",
-    "timestamp": 1567057594.982
-  }
-  ]
+    $ cloud-init analyze dump
```
holmanb, line 174: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/analyze.rst
```diff
@@ -265,15 +272,18 @@ to cloud-init startup, and tracking regression.
...
-  $ cloud-init analyze boot
-  -- Most Recent Boot Record --
-      Kernel Started at: 2019-08-29 01:35:37.753790
-      Kernel ended boot at: 2019-08-29 01:35:38.807407
-      Kernel time to boot (seconds): 1.053617000579834
-      Cloud-init activated by systemd at: 2019-08-29 01:35:43.992460
-      Time between Kernel end boot and Cloud-init activation (seconds): 5.185053110122681
-      Cloud-init start: 2019-08-29 08:35:45.867000
-  successful
+    $ cloud-init analyze boot
```
holmanb, line 275: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/analyze.rst
```diff
@@ -292,10 +302,19 @@ commands below will gather the UserspaceTimestamp and InactiveExitTimestamp:
 
 .. code-block:: shell-session
 
-  $ systemctl show -p UserspaceTimestampMonotonic
-  UserspaceTimestampMonotonic=989279
-  $ systemctl show cloud-init-local -p InactiveExitTimestampMonotonic
-  InactiveExitTimestampMonotonic=4493126
+   $ systemctl show -p UserspaceTimestampMonotonic
+
```
holmanb, line 306: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/analyze.rst
```diff
@@ -292,10 +302,19 @@ commands below will gather the UserspaceTimestamp and InactiveExitTimestamp:
...
+.. code-block::
+
+   UserspaceTimestampMonotonic=989279
+
+.. code-block:: shell-session
+
+   $ systemctl show cloud-init-local -p InactiveExitTimestampMonotonic
+
+.. code-block::
```
holmanb, line 315: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/bugs.rst
```diff
@@ -17,17 +17,20 @@ To aid in debugging, please collect the necessary logs. To do so, run the
 
 .. code-block:: shell-session
 
-  $ sudo cloud-init collect-logs
-  Wrote /home/ubuntu/cloud-init.tar.gz
+   $ sudo cloud-init collect-logs
+
```
holmanb, line 21: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/cli.rst
```diff
@@ -133,9 +138,12 @@ this document.
 
 .. code-block:: shell-session
 
-  $ cloud-init features
-  NETWORK_CONFIG_V1
-  NETWORK_CONFIG_V2
+   $ cloud-init features
+
```
holmanb, line 142: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/cli.rst
```diff
@@ -199,6 +207,9 @@ aliases:
 .. code-block:: shell-session
 
     $ cloud-init query --list-keys
+
```
holmanb, line 210: Same comment about `Example Output:` line between command and example output.

doc/rtd/topics/cli.rst
```diff
@@ -199,6 +207,9 @@ aliases:
...
+.. code-block::
+
```
holmanb, line 212: Some of the examples in the query section (starting at line 222, and the section after that one) have a couple of issues:

1. Examples start with `%`, which I think aught to be standardized on `$` (not your doing).
2. They also strip the comment (`#`) in copying, which is not desirable.
3. Mixed command and output.

More mixed command and output in the status section of this page (again, since no changes there I can't put the comment there otherwise I would).

doc/rtd/topics/debugging.rst
```diff
@@ -45,6 +45,9 @@ subcommands default to reading /var/log/cloud-init.log.
 .. code-block:: shell-session
 
     $ cloud-init analyze dump -i my-cloud-init.log
+
```
holmanb, line 48: Mixed command and output on line 29.

doc/rtd/topics/debugging.rst
```diff
@@ -45,6 +45,9 @@ subcommands default to reading /var/log/cloud-init.log.
...
+.. code-block::
```
holmanb, line 49: Same comment about `Example Output:`.

doc/rtd/topics/debugging.rst
```diff
@@ -61,6 +64,9 @@ subcommands default to reading /var/log/cloud-init.log.
 .. code-block:: shell-session
 
     $ cloud-init analyze blame -i my-cloud-init.log
+
```
holmanb, line 67: Same comment about `Example Output:`.

doc/rtd/topics/debugging.rst
```diff
@@ -73,6 +79,9 @@ subcommands default to reading /var/log/cloud-init.log.
 .. code-block:: shell-session
 
     $ cloud-init analyze boot
+
```
holmanb, line 82: Same comment about `Example Output:`.

doc/rtd/topics/debugging.rst
```diff
@@ -155,10 +164,13 @@ commandline:
 
 .. code-block:: shell-session
 
-  $ sudo cloud-init single --name cc_ssh --frequency always
-  ...
-  Generating public/private ed25519 key pair
-  ...
+   $ sudo cloud-init single --name cc_ssh --frequency always
```
holmanb, line 167: Same comment about `Example Output:`.

doc/rtd/topics/debugging.rst
```diff
@@ -155,10 +164,13 @@ commandline:
...
+
+.. code-block::
+
+   ...
+   Generating public/private ed25519 key pair
+   ...
```
holmanb, line 173: The codeblocks starting on line 225 through the end of the document (in your copy of this file) has commands that are missing the `$` (the line containing `ssh ubuntu@$CI_VM_IP -- cloud-init status --wait --long`)

doc/rtd/topics/format.rst
```diff
@@ -54,12 +54,16 @@ Example
 
 .. code-block:: shell-session
 
-  $ cat myscript.sh
+   $ cat myscript.sh
 
-  #!/bin/sh
-  echo "Hello World.  The time is now $(date -R)!" | tee /root/output.txt
+.. code-block::
```
holmanb, line 59: Could we add an `Example output:` line after the `cat` command?

doc/rtd/topics/format.rst
```diff
@@ -89,6 +93,9 @@ Supported content-types are listed from the cloud-init subcommand make-mime:
 .. code-block:: shell-session
 
     $ cloud-init devel make-mime --list-types
+
```
holmanb, line 96: Could we add an `Example output:` line after the `cloud-init devel ...` command?

doc/rtd/topics/tutorial.rst
```diff
@@ -87,6 +87,11 @@ We can now verify that cloud-init received the expected user data:
 .. code-block:: shell-session
 
     $ cloud-init query userdata
+
+Which should print the following to the terminal window:
+
+.. code-block::
+
```
holmanb, line 94: Line 103 mixes commands and output, and also has an unwanted trailing `$` (not your doing).

Line 118 has a trailing `$` (no your doing).

holmanb (changes requested):
Thanks for this @s-makin! This looks like a really great start and should help with ease of use for our docs. +1 to the effort.

I think that in the places that we separate a command from its output, a line of text between the two commands that states `Example ouput:` or similar would be helpful for context and visual appeal. A couple of spots already have it, but I added comments for those places I think it would be helpful.

I left some comments inline (including a couple of small fix requests to existing docs where things look a little odd).

I haven't had a chance to check files that aren't touched here - this review is only of the files edited.

Please let me know if you have any questions.

doc/rtd/topics/analyze.rst
s-makin: Tox had a bit of a problem with this block specifically - when I tried to break up the block it started complaining about the lines inside it being too long - even though when they were part of one block, it happily accepted them (also, they weren't even too long). I can make the change, though.

doc/rtd/topics/analyze.rst
s-makin: +1

doc/rtd/topics/analyze.rst
s-makin: +1

doc/rtd/topics/analyze.rst
s-makin: +1 - sorry for making you write the same thing over and over :D

doc/rtd/topics/debugging.rst
s-makin: That line doesn't have a $ in it on the live version of the docs either - if it should have one, I can add it in np

doc/rtd/topics/format.rst
s-makin: +1

doc/rtd/topics/tutorial.rst
s-makin: Yeah I wasn't quite sure exactly what to do with this block so I left it alone. I'll fix it up :)

doc/rtd/topics/analyze.rst
holmanb: > when I tried to break up the block it started complaining about the lines inside it being too long - even though when they were part of one block, it happily accepted them 

Hrm, that's a bit annoying. Doing a thing that seemingly shouldn't break anything that causes breakage is never fun.

> (also, they weren't even too long)

This sounds like potentially a report-able bug. If it's still doing that with this version I can take a look at the specific warning and check if we should report to whichever project is causing the inaccurate warning.

doc/rtd/topics/debugging.rst
```diff
@@ -228,30 +253,39 @@ from **-proposed**
...
+    $ ssh ubuntu@$CI_VM_IP -- sudo hostname something-else
+    $ ssh ubuntu@$CI_VM_IP -- sudo cloud-init clean --logs --reboot
 
-5. Validate **-proposed** cloud-init came up without error
+5. Validate **-proposed** cloud-init came up without error. First, we block
+   until cloud-init completes and verify from --long that the datasource is
+   the same as the one picked up from step 1. Errors will show up in --long:
 
-.. code-block:: bash
+.. code-block:: shell-script
```
holmanb, line 273: It looks like we don't have any code-block type `shell-script`, which is causing the [build to fail](https://app.travis-ci.com/github/canonical/cloud-init/jobs/590436459).

That should be I think this is supposed to be `shell-session`?

s-makin: Yes, you're absolutely right - I don't know what my brain was up to last night :joy:

doc/rtd/topics/debugging.rst
```diff
@@ -203,22 +228,22 @@ from **-proposed**
     hostname: SRU-worked-{{v1.cloud_name}}
 
 2. Wait for current cloud-init to complete, replace `<YOUR_VM_IP>` with the IP
-   address of the VM that you launched in step 1:
+   address of the VM that you launched in step 1. Be sure to make a note of the
+   datasource cloud-init detected in --long output. You will need this during
```
s-makin, line 232:
```suggestion
   datasource cloud-init detected in --long output. You will need this during
```

doc/rtd/topics/debugging.rst
```diff
@@ -203,22 +228,22 @@ from **-proposed**
     hostname: SRU-worked-{{v1.cloud_name}}
 
 2. Wait for current cloud-init to complete, replace `<YOUR_VM_IP>` with the IP
-   address of the VM that you launched in step 1:
+   address of the VM that you launched in step 1. Be sure to make a note of the
+   datasource cloud-init detected in --long output. You will need this during 
+   step 5, where you will use it to confirm the same datasource is detected after
+   the upgrade:
```
s-makin, line 234:
```suggestion
   step 5, where you will use it to confirm the same datasource is
   detected after the upgrade:
```

holmanb:
Rebase on `upstream/main` & force push to resolve merge conflict.

doc/rtd/topics/analyze.rst
```diff
@@ -292,10 +313,23 @@ commands below will gather the UserspaceTimestamp and InactiveExitTimestamp:
...
-  $ systemctl show cloud-init-local -p InactiveExitTimestampMonotonic
-  InactiveExitTimestampMonotonic=4493126
+   $ systemctl show -p UserspaceTimestampMonotonic
+
+Example output:
+
+.. code-block::
+
+   UserspaceTimestampMonotonic=989279
+
```
holmanb, line 323: This now looks cluttered:

image

Maybe we could split up the line from before: "will gather the UserspaceTimestamp and InactiveExitTimestamp:" into two sentances so that one can go between the example and the command?

holmanb (changes requested):
Hey @s-makin, thanks for the updates! This is looking good!

I just did a scan of all the docs and found a couple of codeblocks on other pages that may warrant some changes with this change, in addition to a couple of changes in this PR that may need a little touch up. 

### FAQ Page:
1. Under "What datasource am I using?", the `cloud-id` command mixes command and output.

### Various datasource pages

Scanning through the datasource pages, I see several mixed command/output examples - I think these just got missed.

### Network Configuration Page

The section at the bottom titled "Network Configuration Tools"


One last question I had is whether we want to update the [docs](https://cloudinit.readthedocs.io/en/latest/topics/docs.html) page for contributors regarding the copy button?

doc/rtd/topics/cli.rst
```diff
@@ -312,33 +349,66 @@ the currently running modules, as well as when it is done.
...
-  $ cloud-init status --format=json
-  {
-   "boot_status_code": "enabled-by-generator",
-   "datasource": "nocloud",
-   "detail": "DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud-net][dsmode=net]",
-   "errors": [],
-   "last_update": "Tue, 16 Aug 2022 19:12:58 +0000",
-   "status": "done"
-  }
+   $ cloud-init status
```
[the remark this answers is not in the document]
holmanb: Maybe we just add a line of text before the command codeblocks? I'm open to suggestions. The highlighting does make it nicer to visually parse.

doc/rtd/topics/cli.rst
```diff
@@ -215,35 +230,57 @@ aliases:
...
-Below demonstrates how to query standardized metadata from clouds:
+Here are a few examples of how to query standardized metadata from clouds:
+
+.. code-block:: shell-session
+
+   $ cloud-init query v1.cloud_name
+
+Example output:
+
+.. code-block::
```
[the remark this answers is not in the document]
s-makin: I'm not 100% sure exactly what the output for this one should look like. Is it another one of those cases where the comment isn't part of the output and should be moved outside the code block? If so I might be able to rework the text so that we don't even need the code block.

doc/rtd/topics/cli.rst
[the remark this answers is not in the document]
s-makin: I think the reason it looks odd is because there's no context around the commands, so you have the previous output immediately followed by the next command. According to my spreadsheet this page will go under "Explanation", so we can go one of two ways with it (depending on how users would be most likely to want to use the content here). 

(1) We can add context to explain each command before presenting it. In an explanation page we have the leeway to do that in a way that we probably wouldn't in e.g. a how-to, because it's a "learning oriented" page. In this way, if users want to copy the commands without the output examples, they can. ++ convenience.

(2) We can make the assumption that someone who is reading an "explanation" page is doing so from the POV of learning, and that they won't necessarily want to copy each command like they would in a tutorial or how-to. In this case, we can revert the change so it's a single code block again. Unfortunately this would also revert the changes to the highlighting (which I agree is nicer to look at). 

Personally, I would tend towards (1), because it would be more convenient for the users. I'm happy to add the text needed to break up the blocks and make the page look nicer, but equally happy to take advice if this isn't the preferred option.

doc/rtd/topics/analyze.rst
s-makin: Yes, I think that would improve the overall look of it - I'll do that!

s-makin:
> Hey @s-makin, thanks for the updates! This is looking good!
> 
> I just did a scan of all the docs and found a couple of codeblocks on other pages that may warrant some changes with this change, in addition to a couple of changes in this PR that may need a little touch up.
> [snip]

Thanks for the feedback, I'll update those and do another sweep on RTD as well just to make sure I haven't missed any other pages.

I was wondering if I ought to add something to the docs page, yeah, so thanks for the +1 to my thoughts on that - I'll update it now rather than as a separate thing.

doc/rtd/topics/cli.rst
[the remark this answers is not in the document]
s-makin: Nevermind, there was a space missing between the .. and code-block...d'oh!

s-makin:
> > I was wondering if I ought to add something to the docs page
> 
> If you see anything missing or unnecessary, don't hesitate to propose changes. I'm not sure if rtd syntax information really belongs there (the headers), but that's just a passing thought - I haven't put serious thought into what should be on the page.

Based on that, and the questions I have re: standard spellings for different terms, I'm going to leave the devdocs "doc" page alone for now, and update it once I know what's going on there. 

Also, some of the code in the datasource pages was quite hefty, and I wasn't sure what it's doing - so a couple of them I've left alone for now as being "out of scope" for this PR.

doc/rtd/topics/cli.rst
```diff
@@ -307,38 +344,62 @@ non-zero if an error is detected in cloud-init.
...
+
+   status: done
+   boot_status_code: enabled-by-generator
+   last_update: Tue, 16 Aug 2022 19:12:58 +0000
+   detail:
+   DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud-net][dsmode=net]
+
+The detailed output can be shown in machine-readable JSON or YAML with the
```
s-makin, line 385:
```suggestion
The detailed output can be shown in machine-readable JSON or YAML with the
```

doc/rtd/topics/debugging.rst
```diff
@@ -203,22 +228,23 @@ from **-proposed**
...
-   cloud-init:
+   cloud-init. To do this, create the following bash script, which will
+   add the -proposed pocket to APT's sources and install cloud-init
+   from that pocket:
 
 .. code-block:: bash
 
-    # Create a script that will add the -proposed pocket to APT's sources
-    # and install cloud-init from that pocket
     cat > setup_proposed.sh <<EOF
```
holmanb, line 248: This code block can be merged with the lines below.

doc/rtd/topics/tutorials/lxd.rst
```diff
@@ -87,15 +87,24 @@ successfully:
 .. code-block:: shell-session
 
     $ cloud-init status --wait
+
+Which provides the following output:
+
+.. code-block::
+
```
holmanb, line 94: I also noticed this output is incorrect, it should be `status: done` here.

holmanb (changes requested):
Nicely done! I see a couple more nitpicks to fix and then this one should be ready to merge.

holmanb (approved):
I to get this unblocked I fixed up the last couple of comments I had myself. This looks good to me.

Thanks for all the effort on this @s-makin!

PR #1891, opened 2022-12-05
net: netplan config root read-only as wifi config  can contain creds

blackboxsw:
## Proposed Commit Message
```
Restrict read access of /etc/netplan/50-cloud-init.yaml.

On netplan systems, network v2 is passed directly though and written
to /etc/netplan/50-cloud-init.yaml without validation. Current netplan
configuration provides the ability to configure sensitive information
such as `wifi:access-points:password`.

Limit permissions for /etc/network/50-cloud-init.yaml as read-only
for root. Since configuration or modification or netplan config
needs to be performed by an admin user this permission restriction
aligns with netplan tooling.

Set root read-only only always and not just 'if' sensitive material
exists within custom config because it will add confusion to have
two expected modes for this file based on external conditions.
```

## Additional Context
feature tracking improved netplan config security & validation phase 1
https://warthogs.atlassian.net/browse/SC-1365

## Checklist:

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -260,7 +260,8 @@ class Renderer(renderer.Renderer):
 
         if not header.endswith("\n"):
             header += "\n"
-        util.write_file(fpnplan, header + content)
+
+        util.write_file(fpnplan, header + content, mode=0o600)
 
         if self.clean_default:
             _clean_default(target=target)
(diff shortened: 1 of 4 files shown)


TheRealFalcon (approved):
Might be worth an additional check in `test_combined.py`  (and cloud-init.log while we're at it), but not necessary for this PR.

I think we need follow-up PRs for patch these behavior out downstream. I'll leave it to you whether you want to merge now and add those later, or wait to merge this until those are up.

PR #1892, opened 2022-12-05
Networking Clarification

holmanb:
```
net: clarify network function name and docstring
```

## Additional Context
It doesn't seem intuitive to me to describe a function as "removing" a key, when it returns the value in the dict referenced by that key, since I would typically remove a key from a dict using `pop()`, which modifies the dict. 

I propose as alternative `dereference`, `find`, or simply `get`. Thoughts?

```python
pop(...)
    D.pop(k[,d]) -> v, remove specified key and return the corresponding value.
    
    If the key is not found, return the default if given; otherwise,
    raise a KeyError.
```

### DIFF
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -801,11 +801,10 @@ class Init:
         # Run the handlers
         self._do_handlers(user_data_msg, c_handlers_list, frequency)
 
-    def _remove_top_level_network_key(self, cfg):
-        """If network-config contains top level 'network' key, then remove it.
-
-        Some providers of network configuration skip the top-level network
-        key, so ensure both methods works.
+    def _get_network_config(self, cfg) -> dict:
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
Eh...I want something to indicate we're converting to the format that the rest of the code requires. "get_network_config" feels like we're fetching it from somewhere. I'm not really coming up with much right now, but something like `convert_to_local` or `get_network_key_contents` might be better?

If what's here still sounds best to you, I'm fine with it as-is.

TheRealFalcon:
@holmanb thoughts? Also, black complained.

holmanb:
> @holmanb thoughts? Also, black complained.

`get_network_key_contents()` works for me.

I'll update that (and the black complaint) shortly.

Thanks!

PR #1895, opened 2022-12-07
Fix the distro.osfamily output problem in the openEuler system.

xiaoge1001:
fix https://bugs.launchpad.net/cloud-init/+bug/1999042

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -57,7 +57,6 @@ OSFAMILIES = {
         "fedora",
         "mariner",
         "miraclelinux",
-        "openEuler",
         "openmandriva",
         "photon",
         "rhel",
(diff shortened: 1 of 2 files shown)


xiaoge1001:
https://stackoverflow.com/questions/28159539/how-to-fix-the-w293-blank-line-contains-whitespace

I didn't find a problem using the above method.

xiaoge1001:
Lint Tests / Check black  failed, I use the following command to check：
python -m black . --check

TheRealFalcon:
@xiaoge1001 do you work on openEuler? I was under the impression it is a derivative of Ret Hat. Why do we need to distinguish it with a separate osfamily here?

If this change is needed, I think it also makes sense to remove it from here: https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/__init__.py#L60

xiaoge1001:
I think OSFAMILIES needs to be modified, I'll deal with it later.

xiaoge1001:
I've discussed it with my team and we all agree that this is really wrong and needs to be changed.

xiaoge1001:
@TheRealFalcon Hello, any more questions?

TheRealFalcon (approved):
LGTM, thanks!

PR #1897, opened 2022-12-07
sources/azure: encode health report as utf-8

cjp256:
If utf-8 characters are used in the report, it will fail to encode:

```
azure.py[ERROR]: exception while reporting ready: 'latin-1' codec can't
encode characters in position 392-397:
Body ('乱写一些单词') is not valid Latin-1.
Use body.encode('utf-8') if you want to send it encoded in UTF-8.
```

Explicitly encode document as utf-8.

### DIFF
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -13,7 +13,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from errno import ENOENT
 from time import sleep, time
-from typing import Callable, List, Optional, ParamSpec, TypeVar, Union
+from typing import Callable, List, Optional, TypeVar, Union
 from xml.etree import ElementTree
 from xml.sax.saxutils import escape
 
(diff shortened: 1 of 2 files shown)


cloudinit/sources/helpers/azure.py
```diff
@@ -752,7 +752,7 @@ def build_report(
         status: str,
         substatus=None,
         description=None,
-    ) -> str:
```
anhvoms, line 755: there's a cascading change that should follow this change. post_health_report should now take bytes instead of str, the http_with_retries should also switch data from str to bytes

anhvoms (changes requested):

cloudinit/sources/helpers/azure.py
cjp256: Turns out the decorator prevented mypy from doing it's job.  Added type info for the decorator to fix it.

anhvoms (approved):

TheRealFalcon (approved):
LGTM!

PR #1904, opened 2022-12-09
Ubuntu/bionic

blackboxsw:
## Do not squash merge

Upstream commit bbf200f3b2f89ab540a31fb61cf0e6a1bc1cfd07 changes netplan/50-cloud-init.yaml to be root read-only. 
We don't want a breaking change in downstream stable releases. Add a quilt patch to retain original behavior on bionic.

## Additional Context

Note that landing the quilt patch now means that:
 1. our [daily build recipes](https://code.launchpad.net/~cloud-init-dev/+recipe/cloud-init-daily-bionic) will continue to work because they first merge the ubuntu/bionic branch into main, and only then try to apply quilt patches.
 2. people trying to debuild, sbuild or dpkg-buildpackage directly from ubuntu/bionic will not be able to apply quilt patches because this new quilt patch will not apply unless all upstream commits from main are already merged
 
Historically we used to new-upstream-snapshot into ubuntu/bionic prior to introducing the quilt patches, but I'm not certain we have a good use-case or direct consumers of our `upstream/ubuntu/*` branches. And git-ubuntu clone cloud-init; git checkout pkg/ubuntu/bionic-updates would get a buildable source tree with the appropriate patches for anything that we officially released.

## Test Steps

git fetch upstream;
git checkout upstream/main
git merge <this_branch>
quilt push -a # make sure all patches apply
tox -e py3
tox -e do_format
quilt pop -a # make sure all patches can be removed

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+cloud-init (22.4-0ubuntu0~18.04.2) UNRELEASED; urgency=medium
+
+  * d/patches/retain-netplan.patch: netplan-retain-world-readable.patch
+    + Leave 50-cloud-init.yaml permis 644 instead of setting it root read-only
+
+ -- Chad Smith <chad.smith@canonical.com>  Thu, 08 Dec 2022 22:33:02 -0700
+
 cloud-init (22.4-0ubuntu0~18.04.1) bionic; urgency=medium
 
   * d/control: drop python3-httpretty from Build-Depends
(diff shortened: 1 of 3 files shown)


TheRealFalcon (commented):
I'm going to be Mr Annoying with this comment, and I should have thought of this before we merged the upstream PR, but I'm wondering if it makes more sense to introduce a feature flag upstream. If we add a flag [here](https://github.com/canonical/cloud-init/blob/main/cloudinit/features.py#L1) and then update the code and tests to take the flag value into account, then the only thing we need in this patch is changing a single line from True to False (or vice versa) in the features.py file. Future us will be much happier because it'll mean a much lower likelihood of quilt patches failing to apply and/or needing to be refreshed because somebody changed an unrelated test in one of these files.

blackboxsw:
Closing this for two reasons:
  We decided to approach this with a feature flag per @TheRealFalcon's review comment above. It'll make downstream patch management easier.
I put up the following PR to change this into a proper cloudinit.feature: https://github.com/canonical/cloud-init/pull/1905

PR #1907, opened 2022-12-12
Remove distutils dependency

TheRealFalcon:
## Proposed Commit Message

```
Remove distutils dependency

Distutils is gone entirely in Python 3.12. If setuptools isn't
available, it makes no sense to fallback to distutils.

See https://github.com/pypa/setuptools/blob/main/setuptools/errors.py#L29
```


TheRealFalcon:
Nevermind. Looks like we still need the distutils package on older versions, and newer versions will use the setuptools package.

PR #1909, opened 2022-12-12
cc_set_passwords: Move ssh status checking later (SC-1368)

TheRealFalcon:
## Proposed Commit Message

```
cc_set_passwords: Move ssh status checking later

If the service starts between checking status and writing config, then
the new config will no longer be applied. Move the status checking later
so the config is already written before we check if restart needed.

LP: #1998526
```

## Additional Context
I'm not sure this is the best way to handle this. Previously, if the service response indicates ssh won't run at all, we wouldn't write the sshd config at all. With this new way, we write the config regardless, but then don't attempt restart it'll fail. We could instead check early, then check again late, but that seemed like needless additional status checking. Writing out the config shouldn't hurt anything and will take less time than shelling out to systemctl.

## Test Steps
Tests still need to be updated. More looking to see if the idea is sound.

### DIFF
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -117,47 +117,6 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
 
     @return: None"""
     service = distro.get_option("ssh_svcname", "ssh")
-    restart_ssh = True
-    try:
-        distro.manage_service("status", service)
-    except subp.ProcessExecutionError as e:
-        uses_systemd = distro.uses_systemd()
-        if not uses_systemd:
-            LOG.debug(
-                "Writing config 'ssh_pwauth: %s'. SSH service '%s'"
-                " will not be restarted because it is not running or not"
-                " available.",
-                pw_auth,
-                service,
-            )
-            restart_ssh = False
-        elif e.exit_code == 3:
-            # Service is not running. Write ssh config.
-            LOG.debug(
-                "Writing config 'ssh_pwauth: %s'. SSH service '%s'"
-                " will not be restarted because it is stopped.",
-                pw_auth,
-                service,
-            )
-            restart_ssh = False
-        elif e.exit_code == 4:
-            # Service status is unknown
-            LOG.warning(
-                "Ignoring config 'ssh_pwauth: %s'."
-                " SSH service '%s' is not installed.",
-                pw_auth,
-                service,
-            )
-            return
-        else:
-            LOG.warning(
-                "Ignoring config 'ssh_pwauth: %s'."
-                " SSH service '%s' is not available. Error: %s.",
-                pw_auth,
-                service,
-                e,
-            )
-            return
 
     cfg_name = "PasswordAuthentication"
 
@@ -184,11 +143,25 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
         LOG.debug("No need to restart SSH service, %s not updated.", cfg_name)
         return
 
-    if restart_ssh:
-        distro.manage_service("restart", service)
-        LOG.debug("Restarted the SSH daemon.")
+    if distro.uses_systemd():
+        state = subp.subp(
+            f"systemctl show --property ActiveState --value {service}"
+        ).stdout
+        if state.lower() in ["active", "activating", "reloading"]:
+            distro.manage_service("restart", service)
+            LOG.debug("Restarted the SSH daemon.")
+        else:
+            LOG.debug("Not restarting SSH service: service is stopped.")
     else:
-        LOG.debug("Not restarting SSH service: service is stopped.")
+        try:
+            distro.manage_service("restart", service)
+            LOG.debug("Restarted the SSH daemon.")
+        except subp.ProcessExecutionError:
+            util.logexc(
+                LOG,
+                "Cloud-init was unable to restart SSH daemon. "
+                "'ssh_pwauth' configuration may not be applied.",
+            )
 
 
 def handle(
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -2519,7 +2519,7 @@
               "deprecated": true
             }
           ],
-          "description": "Sets whether or not to accept password authentication. ``true`` will enable password auth. ``false`` will disable. Default is to leave the value unchanged."
+          "description": "Sets whether or not to accept password authentication. ``true`` will enable password auth. ``false`` will disable. Default is to leave the value unchanged. In order for this config to be applied, SSH may need to be restarted. On systemd systems, this restart will only happen if the SSH service has already been started. On non-systemd systems, a restart will be attempted regardless of the service state."
         },
         "chpasswd": {
           "type": "object",
--- a/tests/unittests/config/test_cc_set_passwords.py
+++ b/tests/unittests/config/test_cc_set_passwords.py
@@ -102,7 +102,7 @@ class TestHandleSSHPwauth:
         cloud = get_cloud("ubuntu")
         upname = f"{MODPATH}update_ssh_config"
         optname = "PasswordAuthentication"
-        for n, value in enumerate(util.FALSE_STRINGS + util.TRUE_STRINGS, 1):
+        for _, value in enumerate(util.FALSE_STRINGS + util.TRUE_STRINGS, 1):
             optval = "yes" if value in util.TRUE_STRINGS else "no"
             with mock.patch(upname, return_value=False) as m_update:
                 setpass.handle_ssh_pwauth(value, cloud.distro)
--- a/tests/unittests/config/test_cc_set_passwords.py
+++ b/tests/unittests/config/test_cc_set_passwords.py
@@ -17,6 +17,11 @@ from tests.unittests.util import get_cloud
 
 MODPATH = "cloudinit.config.cc_set_passwords."
 LOG = logging.getLogger(__name__)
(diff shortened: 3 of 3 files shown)


holmanb (commented):
This would introduce a different race, wouldn't it? 

If `update_ssh_config()` executes, then the ssh service starts, then we run `distro.manage_service("status", service)` -> in this case we have raced the other way. The ssh service is running, but not on the config we wrote.

The problem is that we are checking the service status and assuming that the state will not change, which is not safe thing to do.

holmanb (changes requested):
To prevent this race, I think we need to start the service if `e.exit_code == 3`. If not using systemd, or it isn't installed or some other return code, then we don't need to start the service.

holmanb:
Oops, just noticed the `wip` label, sorry for the early review.

TheRealFalcon:
I'm not sure I follow. 

> If update_ssh_config() executes, then the ssh service starts, then we run distro.manage_service("status", service) -> in this case we have raced the other way.

In this case the config is already written, so when SSH comes up why wouldn't it be using our config? Note that in this case we will actually restart SSH again. `restart_ssh = True` is the default case and we only don't restart if ssh has been stopped or is otherwise unavailable. It'll be a needless restart of the service if we hit the race condition, but other than adding some time to boot, what else would be the problem?

holmanb:
> we only don't restart if ssh has been stopped or... 

`rc=3` could also mean "has not yet reported that it is running", right?

`systemctl(1)`:
```
       Table 3. LSB return codes
       ┌──────┬──────────────────────────────────────────────────┬─────────────────────────────────────┐
       │Value │ Description in LSB                               │ Use in systemd                      │
       ├──────┼──────────────────────────────────────────────────┼─────────────────────────────────────┤
       │0     │ "program is running or service is OK"            │ unit is active                      │
       ├──────┼──────────────────────────────────────────────────┼─────────────────────────────────────┤
       │1     │ "program is dead and /var/run pid file exists"   │ unit not failed (used by is-failed) │
       ├──────┼──────────────────────────────────────────────────┼─────────────────────────────────────┤
       │2     │ "program is dead and /var/lock lock file exists" │ unused                              │
       ├──────┼──────────────────────────────────────────────────┼─────────────────────────────────────┤
       │3     │ "program is not running"                         │ unit is not active                  │
       ├──────┼──────────────────────────────────────────────────┼─────────────────────────────────────┤
       │4     │ "program or service status is unknown"           │ no such unit                        │
       └──────┴──────────────────────────────────────────────────┴─────────────────────────────────────┘

```

holmanb:
With the current PR consider the following order:

1. sshd service starts and reads the config
2. update_ssh_config() executes
3. distro.manage_service("status", service) (returns `3`)
4. ssh service *reports that it has started*

In this scenario, sshd is running with a stale config.

TheRealFalcon:
I didn't realize activating state would be exit code 3. I think the problem here is that we'll also have exit code 3 if the service has been intentionally stopped or disabled. Also in the case of Ubuntu socket activation, we'll see exit code 3 if there has been no connection yet. In these cases we don't want to (re)start the service.

Maybe we should be using `systemctl show` instead and filtering on `ActiveState`?

holmanb:
> Also in the case of Ubuntu socket activation, we'll see exit code 3 if there has been no connection yet. In these cases we don't want to (re)start the service.

+1 Agreed

> Maybe we should be using `systemctl show` instead and filtering on `ActiveState`?

How do you mean?

holmanb:
My first glance at activestate yields the following:
```bash
$ systemctl show - sshd  | grep -i activestate
ActiveState=inactive
ActiveState=active
```

Maybe I'm not getting it the right way?

TheRealFalcon:
Heh, I think the `-` isn't being treated like most commands. It's interpreting it as an service name. I.e., show me all the details of the `-` service and the `sshd` service.
```
root@asdf:~# systemctl show - sshd ssh | grep -i loaderr
LoadError=org.freedesktop.systemd1.NoSuchUnit "Unit -.service not found."
LoadError=org.freedesktop.systemd1.NoSuchUnit "Unit sshd.service not found."
```

holmanb:
I don't think that there is a race-free way to query systemd state at runtime and act upon it. As soon as we have made our query the state information could be outdated.

Systemd provides well-defined ordering mechanisms via service files, but I assume that such a proposed change would not be well received with good reason (big change, would break /etc/cloud/cloud.cfg module ordering expectations, etc, etc). I'm not going to propose service ordering changes, but it does feel a bit like we're holding the tool by the wrong end with our current implementation.

TheRealFalcon:
Note that `ActiveState` can be `activating`. That may be relevant to how you're thinking about this.

> I don't think that there is a race-free way to query systemd state at runtime and act upon it. 

I'm not sure that actually matters here. We're writing a config and need to make sure the config is respected. By checking the status after we've written the config, all we really need to know is was this service attempted to be started at some point on this boot **before we finished writing the config**. If so, issue a restart. If not, do nothing. If the service starts or is activating after we've written the config but before we do our service check, issuing a restart will be harmless (I'm assuming) other than boot time. If the service starts or is activating after we do a service check, it doesn't matter because we've already written the config. Either way the config gets applied.

holmanb:
> I think the problem here is that we'll also have exit code 3 if the service has been intentionally stopped or disabled.

Prior to https://github.com/canonical/cloud-init/commit/5054ffe1884896bca9250a3c30678e78d228df1a it looks like we restarted sshd for both `pw_auth: true` and `pw_auth: false`, which would have violated the "intentionally changed" case you mention. I agree with the spirit of not surprising the users by "undoing what they did", however I don't think we have enough (non-raced) information available to deliver on this.

It looks like https://github.com/canonical/cloud-init/pull/1422 did a bit more than required to resolve LP: #1969526. Wouldn't simply checking for the existence of sshd before proceeding have been enough?

Aside: this key should really have documented that it is doing service management, which would preclude us from having to guess what they want and deliver on it.

One path forward may be to revert current restart behavior to something more like pre-22.2 (restart for valid ssh_pw key, plus fix LP: #1969526 by checking that sshd is installed). We could round it out by documenting that use of this key means that the service is restarted, which would make it clear to users for both the common case and whatever unexpected use cases users find themselves in (cloud-init singleshot / install / running on a raspi,network device,arduino,whatever) so that we don't have to guess in every possible case what the user wants.

Thoughts?

holmanb:
`systemctl(1)`:
```
       restart PATTERN...
           Stop and then start one or more units specified on the command
           line. If the units are not running yet, they will be started.
```

Pre-22.2 images may have depended on cloud-init to start sshd in the first place, since the call to `restart` would have started the service if it was not started/enabled. Dropping this is a pretty big behavior change for anybody that might have been depending on this in their images.

holmanb:
> Note that ActiveState can be activating. That may be relevant to how you're thinking about this.

+1 Thanks

>> I don't think that there is a race-free way to query systemd state at runtime and act upon it.

> By checking the status after we've written the config, all we really need to know is was this service attempted to be started at some point on this boot before we finished writing the config. If so, issue a restart. If not, do nothing. If the service starts or is activating after we've written the config but before we do our service check, issuing a restart will be harmless (I'm assuming) other than boot time. If the service starts or is activating after we do a service check, it doesn't matter because we've already written the config. Either way the config gets applied.

1. write config
2. check the status
3. service starts after our query
4. restart if already started (could be `SIGHUP`?)

I think this works.

Want to mock it up?

TheRealFalcon:
Still WIP and untested and undocumented...just looking for feedback on the idea. I decided not to go with the SIGHUP version because we still have a possible race condition if the service is activating, it may not be ready to accept a SIGHUP signal.

This version brings back the relatively wide hammer of "restart regardless" on a non-systemd system (which was the case on pre-22.2) with error checking in place to not die if the restart fails for some reason (i.e., service not found).

cloudinit/config/cc_set_passwords.py
```diff
@@ -184,11 +143,22 @@ def handle_ssh_pwauth(pw_auth, distro: Distro):
...
         return
 
-    if restart_ssh:
-        distro.manage_service("restart", service)
-        LOG.debug("Restarted the SSH daemon.")
+    if distro.uses_systemd():
+        state = subp.subp(
+            f"systemctl show --property ActiveState --value {service}"
+        ).stdout
+        if state in ["active", "activating"]:
```
holmanb, line 150: Think we should restart if in the `reloading` state?

holmanb (commented):
> just looking for feedback on the idea

I like the idea so far. One question, and no real objections.

> it may not be ready to accept a SIGHUP signal.

Good point

cloudinit/config/cc_set_passwords.py
holmanb: Properties [list](https://www.freedesktop.org/software/systemd/man/org.freedesktop.systemd1.html#)

TheRealFalcon:
Since it's been a few weeks...no longer WIP and should be ready for full review.

cjp256:
LGTM, but a looksy indicates that sshd should be ready to accept SIGHUP prior to writing out the pid file. Not that looking for the pid file is much of an improvement, but possibly an option if you want to avoid the restart?
https://github.com/openssh/openssh-portable/blob/master/sshd.c#L2070

TheRealFalcon:
> Not that looking for the pid file is much of an improvement

Yeah...I didn't mention it, but that was another reason I didn't love that option. Systemd makes it easy for systemd-based services, but for non systemd, it's not very pretty.

holmanb (approved):
Thanks for putting this together @TheRealFalcon!

PR #1912, opened 2022-12-14
read-version: When insufficient tags, use cloudinit.version.get_version

blackboxsw:
## Proposed Commit Message

```
read-version: When insufficient tags, use cloudinit.version.get_version
```

## Test Steps

```
git clone --no-single-branch --depth 50 https://github.com/canonical/cloud-init.git
cd cloud-init
git fetch origin --force --tags --prune --prune-tags --depth 50 pull/1879/head:external-1879
git checkout --force ac0795a9ff76c48ca74ae23b5766543ef6785387
git apply <this patch>
./tools/read-version
```

## Checklist:

### DIFF
--- a/tools/read-version
+++ b/tools/read-version
@@ -90,7 +90,16 @@ if is_gitdir(_tdir) and which("git") and not is_release_branch_ci:
             branch_name,
         ] + flags
 
-        version = tiny_p(cmd).strip()
+        try:
+            version = tiny_p(cmd).strip()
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Can we instead use what's defined in version.py?

blackboxsw:
> Can we instead use what's defined in version.py?

I think that'll work too. Our only goal with this is to make tools/read-version more resilient in the face of a missing annotated tag.
I've updated the PR

TheRealFalcon (approved):

PR #1913, opened 2022-12-14
Document Incorrect Prefix For cc_mounts

holmanb:
```
mounts: document weird prefix in schema

Add test and support for parsing IEC prefix format.
```

Top search engine link for me:
https://inst.eecs.berkeley.edu/~cs61c/su14/labs/02/iec.pdf

Feel free to suggest better wording/placement. I do think that a warning somewhere is probably warranted, since we're doing something explicitly incorrect without telling the user.

Note: There was a time when these prefixes were less well-defined, but I think that point in time has since passed.

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -1764,7 +1764,7 @@
               "description": "Path to the swap file to create"
             },
             "size": {
-              "description": "The size in bytes of the swap file, 'auto' or a human-readable size abbreviation of the format <float_size><units> where units are one of B, K, M, G or T. WARNING: Attempts to use IEC prefixes in your configuration will result in unexpected behavior. SI prefixes names (KB, MB) are currently required in your configuration, despite the fact that IEC values are used. In summary, assume 1KB == 1024B, not 1000B.",
+              "description": "The size in bytes of the swap file, 'auto' or a human-readable size abbreviation of the format <float_size><units> where units are one of B, K, M, G or T. **WARNING: Attempts to use IEC prefixes in your configuration will result in unexpected behavior. SI prefixes names (KB, MB) are currently required, despite the fact that IEC values are used. In summary, assume 1KB == 1024B, not 1000B**",
               "oneOf": [
                 {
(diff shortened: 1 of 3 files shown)


s-makin (commented):
I think as well as where you've put it, it's probably a good idea to insert a warning note next to swap: (object) or after the box that follows it - if that's even something you can do in this file?

If not, then I think the note you propose is clear enough, it just needs to be a bit more visible (bold face, maybe, or italic?)

image

holmanb:
> if that's even something you can do in this file?

Good question - as one of those magically generated bits I'm not sure how we would do that.

holmanb:
@xiaoge1001 While reviewing your PR https://github.com/canonical/cloud-init/pull/1910 I noticed this issue. Since you most recently touched this code, could I get a review from you please?

holmanb:
@s-makin Thanks for your earlier review. I just updated the PR to allow support for IEC prefixes as well going forward. Supporting both SI and IEC for IEC values is a little weird, but less weird than what we currently do (using SI prefix notation to configure IEC values).

I'm assuming [8 years of this behavior](https://github.com/canonical/cloud-init/commit/9736b260434af860c6ec81776f4278640f1fa9be) is a bit long to start changing behavior when `MB` is configured, so I'm not trying to fix it, merely maintain current behavior while also supporting the prefixes that actually correspond to the numbers.

xiaoge1001:
@holmanb Thanks for your invitation. I think it's good to keep the current behavior, after all, it's been used for 8 years.
  I think this is an issue with the util.human2bytes() implementation, which currently only affects the cc_mounts module, is it possible to affect other modules in the future? If possible, can we add a description for this issue to the function comments?

holmanb:
> is it possible to affect other modules in the future?

Yes

> If possible, can we add a description for this issue to the function comments?

Good suggestion, thanks!

holmanb:
@xiaoge1001 Thanks for the review. I've updated per your comments in commit https://github.com/canonical/cloud-init/pull/1913/commits/1ed130c9c8761a1ca459367a88c9fc0155e620e0.

s-makin (approved):

PR #1914, opened 2022-12-14
sources/azure: fix device driver matching for net config

cjp256:
The ordering of NICs provided by IMDS may not match the order enumerated by kernel.  As such, we do not have any guarantee that the nic we're checking the driver for is the nic we think it is.

Instead of making any assumptions about how the nics are named, check all interfaces by mac address.  If there is an interface using "hv_netvsc", match against that.  If there is only one interface driver that is not blacklisted, use that (in case it is not "hv_netvsc"), but log a debug event.  If there are multiple hits, don't match against any of the names and report a warning.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -22,7 +22,6 @@ from cloudinit import dmi
 from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
-from cloudinit.net import device_driver
 from cloudinit.net.dhcp import (
     NoDHCPLeaseError,
     NoDHCPLeaseInterfaceError,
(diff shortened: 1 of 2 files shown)


cjp256:
Moved from #1672

cjp256:
Question is what to do about BSD until the refactor by @igalic lands.

- I don't expect the driver matching bits in the network config to be relevant on BSD?
- get_interfaces() returns an empty list on BSD today, so perhaps this change is harmless as-is?
- mlx4/mlx5 accelerated networking scenarios on Azure breaks today in cloud-init, so I don't believe there is a risk of regression (though we do need to fix that separately)

igalic:
@cjp256 basically, `get_interfaces()` and `get_interface_by_mac()` would be prime candidates for moving into `cloudinit.distris.networking`, especially given that `get_interfaces()` is currently broken 

but they have a lot of call-sites, some of which don't have a distro object, currently. so it's gonna be a lot of work.

I would like to coordinate this work with y'all, so we don't actually get into each other's way

cjp256:
Sounds great @igalic!

Given that get_interfaces() returns an empty list, this PR shouldn't regress BSD in any way that I can see.  Once the distro support is firmed up, I can rework this (and other areas) to use the distro APIs.

TheRealFalcon (approved):
LGTM!

PR #1915, opened 2022-12-15
integration tests: add  IBM VPC support (SC-1352)

aciba90:
## Proposed Commit Message

```
integration tests: add IBM VPC support
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-1352
https://github.com/canonical/pycloudlib/pull/225

## Test Steps

```sh
CLOUD_INIT_PLATFORM=ibm tox -re integration-tests -- tests/integration_tests/<some-test>
```

## Checklist:


tests/integration_tests/clouds.py
```diff
@@ -408,3 +409,14 @@ def _get_initial_image(self, **kwargs):
...
+
+
+class IbmCloud(IntegrationCloud):
+    datasource = "ibm"
+    cloud_instance: IBM
+
+    def _get_cloud_instance(self) -> IBM:
+        # Note: IBM image names starting with `ibm` are reserved.
+        return IBM(
+            tag="cloud-init-integration-test-ibm",
```
TheRealFalcon, line 421: Let's have this follow the <name>-integration-test convention. I.e., `ibm-integration-test`

TheRealFalcon (approved):
One comment inline. Also, can we add ibm to the list here:
https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/integration_settings.py#L18

Those are fairly trivial changes, so once you make feel free to merge.

tests/integration_tests/clouds.py
aciba90: I forgot to mention that instance/image names starting with `ibm` are reserved in IBM VPC.

Taking into account that, do you think `cloud-init-integration-test-ibm` is okay?

TheRealFalcon: Ah, sad. That is fine then. It might be worth double checking there's no length restriction in IBM. The name plus the timeestamp could get long. There is a regex here used to validate instance names:
https://github.com/aciba90/pycloudlib/blob/main/pycloudlib/util.py#L322

aciba90: * Renamed the tag to `integration-test-ibm`
* Launched an integration test and passes, thus `validate_tag` accepts a tag with timestamp
* The max length of instance / snapshot names in IBM VPC is 63 (see attached image), thus we do not need to restrict the regex within `validate_tag`.

image

TheRealFalcon:
Your changes LGTM. Since this is already approved, feel free to merge when ready.

PR #1917, opened 2022-12-15
BSD: fix duplicate macs in Ifconfig parser

igalic:
## Proposed Commit Message

```
BSD: fix duplicate macs in Ifconfig parser

Some cloud providers can have more than one device with the same MAC
address. This PR allows parsing and storing and retrieving such
configurations.

We now use `defaultdict` to retrieve `ifs_by_mac`, before converting it
all to a `dict`. We also store the two in separate variables.

Add test case from Azure to verify, and test data in a new file, since
our old cloudinit.net functions can't handle it.

```

## Checklist:

### DIFF
--- a/cloudinit/distros/parsers/ifconfig.py
+++ b/cloudinit/distros/parsers/ifconfig.py
@@ -6,6 +6,7 @@
 
 import copy
 import re
+from collections import defaultdict
 from functools import lru_cache
 from ipaddress import IPv4Address, IPv4Interface, IPv6Interface
 from typing import Dict, Optional, Tuple
@@ -87,10 +88,10 @@ class Ifconfig:
     """
 
     def __init__(self):
-        self._ifs = {}
+        self._ifs = defaultdict(list)
 
     @lru_cache()
-    def parse(self, text: str) -> Dict[str, Ifstate]:
+    def parse(self, text: str) -> defaultdict[(str, Ifstate)]:
         """
         Parse the ``ifconfig -a`` output ``text``, into a dict of ``Ifstate``
         objects, referenced by ``name`` *and* by ``mac`` address.
@@ -119,7 +120,7 @@ class Ifconfig:
                     curif = curif[:-1]
                 dev = Ifstate(curif)
                 dev.index = ifindex
-                self._ifs[curif] = dev
+                self._ifs.append([curif, dev])
 
             toks = line.lower().strip().split()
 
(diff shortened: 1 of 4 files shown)


tests/unittests/distros/test_ifconfig.py
```diff
@@ -38,6 +38,13 @@ def test_vtnet_options(self):
         ifs = Ifconfig().parse(self.ifs_txt)
         assert "txcsum" in ifs["vtnet0"].options
 
+    def test_duplicate_mac(self):
+        """assert that we can have duplicate macs"""
+        ifc = Ifconfig()
+        ifc.parse(self.ifs_txt)
+        ifs_by_mac = ifc.ifs_by_mac()
+        assert len(ifs_by_mac["00:0d:3a:54:ad:1e"]) == 2
```
igalic, line 46: we can do another assert here, that the two devices are different ones, and this isn't an accident or bug in the parser

igalic (commented):
improvement potential

tests/data/netinfo/freebsd-ifconfig-output
```diff
@@ -39,3 +39,16 @@ lo0: flags=8049<UP,LOOPBACK,RUNNING,MULTICAST> metric 0 mtu 16384
 	inet 127.0.0.1 netmask 0xff000000 
 	groups: lo
 	nd6 options=21<PERFORMNUD,AUTO_LINKLOCAL>
+hn0: flags=8843<UP,BROADCAST,RUNNING,SIMPLEX,MULTICAST> metric 0 mtu 1500
```
aciba90, line 42: Modifying this data file provokes other unit tests to fail.

aciba90 (changes requested):
Thanks, @igalic, for the fix! The change set LGTM. 

Unfortunately, after modifying the data file representing the `ifconfig` output, other test cases break.

cloudinit/distros/parsers/ifconfig.py
```diff
@@ -87,10 +88,11 @@ class Ifconfig:
     """
 
     def __init__(self):
-        self._ifs = {}
+        self._ifs_by_name = {}
+        self._ifs_by_mac = {}
 
     @lru_cache()
-    def parse(self, text: str) -> Dict[str, Ifstate]:
+    def parse(self, text: str) -> Dict[str, Union[Ifstate, List[Ifstate]]]:
```
aciba90, line 95: I guess the type change is not needed as we do not return a list, right?

```suggestion
    def parse(self, text: str) -> Dict[str, Ifstate]:
```

aciba90 (changes requested):
Thanks, @igalic, for the test fixes.

Aside from the type change, it looks great to me.

cloudinit/distros/parsers/ifconfig.py
igalic: we do return a str: List type dict in the case of duplicate mac addresses tho

aciba90: Indeed, still booting up from holidays :)

aciba90 (approved):
LGTM, thanks!

PR #1923, opened 2022-12-16
Added skip_detect_openstack option for OpenStack

BirknerAlex:
## Added "skip_detect_openstack" for OpenStack datasource

```
summary: Fixing an issue when running the OpenStack datasource on Bare Metal machines

LP: #1815990 (might be fixed as well)
```

## Additional Context
Running CloudInit with the OpenStack datasource configured on a Bare Metal machine is resulting in an issue since the DataSource itself also checks if it's running on OpenStack inside a virtual machine using dmi decode results. 

Maybe fixes a 2019 issue or at least gives an workaround for it https://bugs.launchpad.net/cloud-init/+bug/1815990

## Test Steps
Removing that check allows to run the OpenStack data source also on bare metal provisioned machines. I have verified that Cloud Init executes on Bare Metal now and if the default still behaves the same.

## Checklist:


BirknerAlex:
Something to discuss, might been changing the default setting for skipping based on if the system is bare metal or not, e.g. use `systemd-detect-virt` (if systemd available) output to see if it's virtual machine or bare metal.

This would allow to use existing OpenStack images with CloudInit without further modifications on bare metal machines. Currently you need to specify the OpenStack data source and adding the new config option `skip_detect_openstack` when building images.

igalic:
> Something to discuss, might been changing the default setting for skipping based on if the system is bare metal or not, e.g. use `systemd-detect-virt` (if systemd available) output to see if it's virtual machine or bare metal.
> 
> This would allow to use existing OpenStack images with CloudInit without further modifications on bare metal machines. Currently you need to specify the OpenStack data source and adding the new config option `skip_detect_openstack` when building images.
> 
> See [BirknerAlex@86993cf](https://github.com/BirknerAlex/cloud-init/commit/86993cffa1e5dd1946a07484ae89cbdaf1142ae6) as example.

the `is_virtual` helper looks like a candidate to put into the distro class.

The distros will know best whether something is virtual, and how to figure that out. it also does not restrict us to systemd Linux systems

(there's currently a project to get [OpenStack running natively running on FreeBSD](https://freebsdfoundation.org/project/openstack-on-freebsd/))

holmanb:
This doesn't fix LP: #1815990, does it? Passing in a kernel commandline option to force datasource will still get ignored. This allows a workaround, but I think that the issue of that bug remains.

I agree that autodetection would be nice for users (+1 to @igalic's suggestion), and this cloud config option is also a nice way to force the desired behavior, but I don't think we can call this bug fixed with this change.

BirknerAlex:
First of all, I wish you all a happy new year 🎉 . Thank you for all that input on my PR.

> the `is_virtual` helper looks like a candidate to put into the distro class.

Yeah I can implement that for sure, what should be the default behaviour for distros that have not implemented a own implementation? Fallback to systemd (which should match the most distros)? 

I am not familiar with all your supported distros so I would be very happy to get more input on that implementation. 

Maybe to explain: The reason why I did go with the systemd utility was I thought most of your distros are systemd based (correct me if I am wrong) and I also added fallback to the old behaviour for does who doesn't. Just to prevent something to break after my changes. But for sure a proper solution would be better but requires more knowlege on all other distros for me.

> This doesn't fix LP: #1815990, does it? Passing in a kernel commandline option to force datasource will still get ignored. This allows a workaround, but I think that the issue of that bug remains.

Good question, that change on my personal branch allowed me to run CloudInit on my bare metal machines, the only thing that prevented me from doing so was that internal data source check which aborted without any log message at all, was a little bit hard to find. After putting my 2 patched files into my Ubuntu test image it worked. But I need to admit that I put the following config as well to that image:

```
datasource_list: [ OpenStack, None ]
datasource:
    OpenStack:
        skip_detect_openstack: true
```

I also added `ci.ds=OpenStack` to the grub kernel parameters - but I never tested only using the kernel parameter and removing `datasource_list` from the config.

BirknerAlex:
In the meantime I added this as property to the base distro class now. So it can be overwritten on other distro classes if needed. I added a fallback to `True` since I don't wanna break something else, currently this method is also only used on the OpenStack datasource which should be 99% (just guessing) virtual machines.

cloudinit/distros/__init__.py
```diff
@@ -983,6 +984,40 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
             **kwargs,
         )
 
+    @property
+    def is_virtual(self) -> bool:
+        """Detect if is running as a virtual machine or bare metal.
+        If fails, it returns True to avoid breaking changes.
```
aciba90, line 990: IMO this is going to be confusing and misleading. What about encoding the unknown case as None?

cloudinit/distros/__init__.py
```diff
@@ -983,6 +984,39 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+    def is_virtual(self) -> bool:
+        """Detect if running on a virtual machine or bare metal.
+
+        If the detection fails, it returns None.
+        """
+        if not uses_systemd():
+            # For non systemd systems the method should be
+            # implemented in the distro class.
+            LOG.debug("is_virtual should be implemented on distro class")
+            return True
```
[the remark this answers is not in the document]
BirknerAlex: done, thanks

cloudinit/distros/__init__.py
```diff
@@ -983,6 +984,39 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
             **kwargs,
         )
 
+    @property
+    def is_virtual(self) -> bool:
```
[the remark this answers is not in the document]
BirknerAlex: Oops. Sorry for that, I am really not an python dev, trying my best here and learning. Thanks for your feedback.

cloudinit/distros/__init__.py
```diff
@@ -983,6 +984,39 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+        If the detection fails, it returns None.
+        """
+        if not uses_systemd():
+            # For non systemd systems the method should be
+            # implemented in the distro class.
+            LOG.debug("is_virtual should be implemented on distro class")
+            return True
+
+        try:
+            proc = subprocess.run(
```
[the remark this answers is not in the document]
BirknerAlex: Changed.

cloudinit/distros/__init__.py
```diff
@@ -994,23 +993,17 @@ def is_virtual(self) -> bool:
...
         try:
-            proc = subprocess.run(
-                "systemd-detect-virt",
-                stdin=None,
-                shell=True,
-                capture_output=True,
-                check=False,
-                text=True,
+            (out, _) = subp.subp(
+                ["systemd-detect-virt"], shell=True, capture=True, rcs=[0, 1]
```
igalic, line 1000: why do we need the shell here?

igalic (changes requested):
could you add a test (or a few…) for this in the existing test file?

cloudinit/distros/__init__.py
BirknerAlex: The path to the binary differs from distribution to distribution, my understanding is that shell loads the PATH environment variable correctly. Am I wrong?

igalic:
I've opened a draft pull request for implementing is_virtual() in FreeBSD: #1957 

this is based on the general hierarchy that Puppet's facter figures out if a machine is a container, virtual is physical.

BirknerAlex:
Just letting you know I am still on it but currently I'm very busy with other tasks, I will take care about the feedback asap.

holmanb:
> Just letting you know I am still on it but currently I'm very busy with other tasks, I will take care about the feedback asap.

Thanks, sounds good!

dermotbradley:
> I am not familiar with all your supported distros so I would be very happy to get more input on that implementation.
> 
> Maybe to explain: The reason why I did go with the systemd utility was I thought most of your distros are systemd based (correct me if I am wrong)

Apart from the BSD distros supported, only Alpine Linux and (potentially Gentoo) doesn't use systemd - it uses OpenRC.

Similar to @igalic is_virtual for FreeBSD I can implement such a function for Alpine, probably running the ```virt-what``` program.

BirknerAlex:
Thanks again for feedback and waiting. I did some reworks and added some unit tests. Please let me know if you having more feedback. The pipeline seems to be successful now too. 🎉

holmanb (commented):
Hi @BirknerAlex,

Thanks for following up on this, and good work so far!

Thinking a bit about cloud-init datasource detection, I'm hesitant to require cloud-init users to specify a cmdline or config value in the default case on Openstack bare metal.

I think that cloud-init should automatically detect its platform, and if that doesn't work it is a bug.

Rather than inject a new key that allows overriding detection, I think it would be better to detect that we're in an Openstack environment.

I don't know the best way to do this, but one approach would be to modify the OpenStack datasource such that `detect_openstack()` returns true if it's not a virtual machine, implementation suggestions here [1]. This would allow cloud-init to query for the datasource when on bare-metal, and if none is found continue with datasource detection on other datasources. I think Ec2 might do something similar to this.

[1] Since this function isn't used outside of DataSourceOpenstack.py except for testing, this could just be made an instance method of `DataSourceOpenStack` to make use of the `is_virtual()` method you've added to the distro class).

cloudinit/distros/__init__.py
```diff
@@ -983,6 +983,39 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+            LOG.warning("is_virtual should be implemented on distro class")
+            return None
+
+        try:
+            detect_virt_path = subp.which("systemd-detect-virt")
+            if detect_virt_path:
+                (out, _) = subp.subp(
+                    [detect_virt_path], capture=True, rcs=[0, 1]
+                )
+                if out.strip() == "none":
```
holmanb, line 1004: Nit:
```python
                if out.strip() == "none":
                    return False
                else:
                    return True
```
could just be:
```python
                return not out.strip() == "none":
```

cloudinit/distros/__init__.py
```diff
@@ -983,6 +983,39 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+        if not uses_systemd():
+            # For non systemd systems the method should be
+            # implemented in the distro class.
+            LOG.warning("is_virtual should be implemented on distro class")
+            return None
+
+        try:
+            detect_virt_path = subp.which("systemd-detect-virt")
+            if detect_virt_path:
+                (out, _) = subp.subp(
```
holmanb, line 1001: Nit: parenthesis not needed in assignment
```python
               out, _ = subp.subp(
```

holmanb:
> [1] Since this function isn't used outside of DataSourceOpenstack.py except for testing, this could just be made an instance method of `DataSourceOpenStack` to make use of the `is_virtual()` method you've added to the distro class).

This patch (on top of your latest commit) is what I generally had in mind: 
```diff
diff --git a/cloudinit/sources/DataSourceOpenStack.py b/cloudinit/sources/DataSourceOpenStack.py
index 46368d1b7..cbf106747 100644
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -150,19 +150,8 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
             False when unable to contact metadata service or when metadata
             format is invalid or disabled.
         """
-        is_bare_metal = self.distro.is_virtual is False
-
-        LOG.debug(
-            "OpenStack is running on %s.",
-            "bare metal" if is_bare_metal else "virtual machine",
-        )
-        skip_detect = self.ds_cfg.get("skip_detect_openstack", is_bare_metal)
-
         oracle_considered = "Oracle" in self.sys_cfg.get("datasource_list")
-        if not skip_detect and not detect_openstack(
-            accept_oracle=not oracle_considered
-        ):
-            LOG.debug("OpenStack datasource Not running on OpenStack")
+        if not self.detect_openstack(accept_oracle=not oracle_considered):
             return False
 
         if self.perform_dhcp_setup:  # Setup networking in init-local stage.
@@ -258,6 +247,25 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
             raise sources.InvalidMetaDataException(msg) from e
         return result
 
+    def detect_openstack(self, accept_oracle=False):
+        """Return True when a potential OpenStack platform is detected."""
+        if not util.is_x86():
+            return True  # Non-Intel cpus don't properly report dmi product names
+        product_name = dmi.read_dmi_data("system-product-name")
+        if product_name in VALID_DMI_PRODUCT_NAMES:
+            return True
+        elif dmi.read_dmi_data("chassis-asset-tag") in VALID_DMI_ASSET_TAGS:
+            return True
+        elif accept_oracle and oracle._is_platform_viable():
+            return True
+        elif util.get_proc_env(1).get("product_name") == DMI_PRODUCT_NOVA:
+            return True
+        elif not self.distro.is_virtual:
+            return True
+        return False
+
+
+
 
 class DataSourceOpenStackLocal(DataSourceOpenStack):
     """Run in init-local using a dhcp discovery prior to metadata crawl.
@@ -278,22 +286,6 @@ def read_metadata_service(base_url, ssl_details=None, timeout=5, retries=5):
     return reader.read_v2()
 
 
-def detect_openstack(accept_oracle=False):
-    """Return True when a potential OpenStack platform is detected."""
-    if not util.is_x86():
-        return True  # Non-Intel cpus don't properly report dmi product names
-    product_name = dmi.read_dmi_data("system-product-name")
-    if product_name in VALID_DMI_PRODUCT_NAMES:
-        return True
-    elif dmi.read_dmi_data("chassis-asset-tag") in VALID_DMI_ASSET_TAGS:
-        return True
-    elif accept_oracle and oracle._is_platform_viable():
-        return True
-    elif util.get_proc_env(1).get("product_name") == DMI_PRODUCT_NOVA:
-        return True
-    return False
-
-
 # Used to match classes to dependencies
 datasources = [
     (DataSourceOpenStackLocal, (sources.DEP_FILESYSTEM,)),
```

Thoughts?

BirknerAlex:
Hi @holmanb,

Thanks for your patch, I took it and migrated all tests to the new interface. I agree that this looks more clean now after migrating it inside the class.

Please let me know if that PR is fine now, I would love to see this merged soon. 🥳 

Maybe some "stupid question". What is your personal experience how long it normally would take until the change is included on the most major linux distributions if using their pre-build cloud images?

Thanks in advance.

holmanb:
> Thanks for your patch, I took it and migrated all tests to the new interface. I agree that this looks more clean now after migrating it inside the class.
> 
> Please let me know if that PR is fine now, I would love to see this merged soon. 

+1 Sounds good, I'll give another review.

> What is your personal experience how long it normally would take until the change is included on the most major linux distributions if using their pre-build cloud images?

Good question. Assuming that this gets merged before the February 13th release, then it should be available for distros to start testing and including in their repos.

For Ubuntu this typically takes a couple of weeks or longer to ensure that the package is working as expected on all supported clouds. Bugs found during this period require rework and re-release. Typically Ubuntu, Debian, Alpine, Arch, Freebsd, etc have the latest version of cloud-init in their repository within a release cycle. See [this link](https://repology.org/project/cloud-init/versions) for current package versions on various distros. For any distro that isn't keeping up with the latest version, you might be able to reach out to them to request either get the latest version or a backported patch with the fix.

Once the packages are available then cloud images can start incorporating them. I don't know how long that cycle typically takes.

holmanb:
Hi @BirknerAlex,

Thanks for all the work on this! This is shaping up nicely.

Limiting detection to bare-metal machines does a lot to mitigate risk, but I think that since other clouds might have bare-metal instances, the full timeout of `_crawl_metadata()` would be excessive in the case that we got a false positive and need to carry on to detect a different datasource. This is mitigated in some other ways with Ubuntu images (and possibly other distros) with per-cloud image configs that limit checking to individual clouds, but to be a better upstream I think we should include a URL check that doesn't require timing out on crawling the whole imds - preferably under 30s total (a little wait is desirable in case imds isn't ready on boot).

For implementation, the check should probably be something like this (with tweaked timeout/max_wait settings).
```python
                url, _ = url_helper.wait_for_url(
                    urls=[f"{url}/openstack/" for url in md_urls],
                    max_wait=url_params.max_wait_seconds,
                    timeout=url_params.timeout_seconds,
                    connect_synchronously=False,
                )
```
If `url` is truthy, the datasource should be OpenStack.

Note: Since openstack's ipv4 IMDS address is used by other clouds, the URL checked will have to be an openstack-specific URL, otherwise we'll certainly get false positives on just `http://169.254.169.254`.

However, if `self.perform_dhcp_setup`, this call will need to be called within the `EphemeralDHCPv4` context manager, since host networking won't be setup yet. That is to say that we can't just put this inside the `self.distro.is_virtual` check in `detect_openstack()`. 

Let me know if you have any thoughts or questions on this.

holmanb (commented):
Hi @BirknerAlex,

Thanks for all the work on this! This is shaping up nicely.

Limiting detection to bare-metal machines does a lot to mitigate risk, but I think that since other clouds might have bare-metal instances, the full timeout of `_crawl_metadata()` would be excessive in the case that we got a false positive and need to carry on to detect a different datasource. This is mitigated in some other ways with Ubuntu images (and possibly other distros) with per-cloud image configs that limit checking to individual clouds, but to be a better upstream I think we should include a URL check that doesn't require timing out on crawling the whole imds - preferably under 30s total (a little wait is desirable in case imds isn't ready on boot).

For implementation, the check should probably be something like this (with tweaked timeout/max_wait settings).
```python
                url, _ = url_helper.wait_for_url(
                    urls=[f"{url}/openstack/" for url in md_urls],
                    max_wait=url_params.max_wait_seconds,
                    timeout=url_params.timeout_seconds,
                    connect_synchronously=False,
                )
```
If url is truthy, the datasource should be OpenStack.

Note: Since openstack's ipv4 IMDS address is used by other clouds, the URL checked will have to be an openstack-specific URL, otherwise we'll certainly get false positives on just `http://169.254.169.254`.

However, if `self.perform_dhcp_setup`, this call will need to be called within the EphemeralDHCPv4 context manager, since host networking won't be setup yet. That is to say that we can't just put this inside the `self.distro.is_virtual` check in `detect_openstack()`.

Let me know if you have any thoughts or questions on this.

holmanb (changes requested):
See my previous comment, requesting changes.

holmanb (commented):
Interesting. I was just testing this code, and it appears that `test_ds_identify.py` tests pass in containers (and therefore our CI tests), but not on bare metal.

```python
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_bobrightbox_is_not_brightbox - AssertionError: Lists differ: ['Ec2', 'None'] != ['OpenStack', 'None']
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_e24cloud_not_active - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_ibmcloud_template_no_userdata - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_ibmcloud_template_no_userdata_in_provisioning - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_ibmcloud_template_userdata_in_provisioning - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_openstack_on_non_intel_is_maybe - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_outscale_not_active_productname - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_outscale_not_active_sysvendor - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_envvar_no_data - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_guestinfo_no_data - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_guestinfo_no_virt_id - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_no_valid_transports - AssertionError: 1 != 0
FAILED tests/unittests/test_ds_identify.py::TestOracle::test_not_found - AssertionError: 1 != 0
==================================================================================== 13 failed, 75 passed, 1 warning in 1.67s =====================================================================================
ERROR: InvocationError for command /home/holmanb/cloud-init-h/.tox/py3/bin/python -m pytest --durations 10 -m 'not hypothesis_slow' tests/unittests/test_ds_identify.py (exited with code 1)
```

For non-systemd portability, and also to make use of the shell mock magic in `test_ds_identify.py`, we should be using ds-identify's `detect_virt` function. Using this patch on your branch, it gets us down to 6 failures:
```diff
  diff --git a/tools/ds-identify b/tools/ds-identify
  index da1399767..d4e78e09b 100755
  --- a/tools/ds-identify
  +++ b/tools/ds-identify
  @@ -1264,8 +1264,9 @@ dscheck_OpenStack() {
   
       # If we are on bare metal, then we maybe are on a
       # bare metal Ironic environment.
  -    if [ -d /run/systemd ] &&
  -        [ "$(systemd-detect-virt)" = "none" ]; then
  +
  +    detect_virt
  +    if [ "${_RET}" = "none" ]; then
           return ${DS_MAYBE}
       fi
   
```

I'm nearing EOD now, but I'll check back tomorrow to look into the other failing tests to see why they are failing.

holmanb:
The remaining tests use `_test_ds_not_found()`/`_check_via_dict()` with the default ds-identify mock (which returns "none") to assert that no datasource is found. Since this PR changes the behavior such that openstack is a "maybe" in these scenarios, I think we just need to update these tests to verify new behavior.

Let me know if you agree with the following approach:

I force `dscheck_OpenStack` to not return maybe by mocking the failing tests with the KVM mock. This should give the same coverage that existed previously.

The alternative (possibly cleaner) way would be to add a `dslist` parameter to `_test_ds_found` so we can pass `["openstack", "none"]` to these tests and change the failing tests to `_test_ds_found()`. Since this patch does something similar for the failing oracle test, I don't think there is much of a difference in coverage and I would be fine with either approach.

Combine the following with the patch in the last comment.
```diff
diff --git a/tests/unittests/test_ds_identify.py b/tests/unittests/test_ds_identify.py
index cc75209e2..8065a92f0 100644
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -950,7 +950,9 @@ class TestOracle(DsIdentifyBase):
         """Simple negative test of Oracle."""
         mycfg = copy.deepcopy(VALID_CFG["Oracle"])
         mycfg["files"][P_CHASSIS_ASSET_TAG] = "Not Oracle"
-        self._check_via_dict(mycfg, rc=RC_NOT_FOUND)
+
+        # on bare metal, openstack is "maybe"
+        self._check_via_dict(mycfg, ds=["openstack", "none"], rc=RC_FOUND)
 
 
 def blkid_out(disks=None):
@@ -1056,6 +1058,7 @@ VALID_CFG = {
     "Ec2-brightbox-negative": {
         "ds": "Ec2",
         "files": {P_PRODUCT_SERIAL: "tricky-host.bobrightbox.com\n"},
+        "mocks": [MOCK_VIRT_IS_KVM],
     },
     "GCE": {
         "ds": "GCE",
@@ -1597,6 +1600,7 @@ VALID_CFG = {
     "Ec2-E24Cloud-negative": {
         "ds": "Ec2",
         "files": {P_SYS_VENDOR: "e24cloudyday\n"},
+        "mocks": [MOCK_VIRT_IS_KVM],
     },
     "VMware-NoValidTransports": {
         "ds": "VMware",
@@ -1755,6 +1759,7 @@ VALID_CFG = {
     "VMware-GuestInfo-NoVirtID": {
         "ds": "VMware",
         "mocks": [
+            MOCK_VIRT_IS_KVM,
             {
                 "name": "vmware_has_rpctool",
                 "ret": 0,
@@ -1860,6 +1865,7 @@ VALID_CFG = {
             P_PRODUCT_NAME: "3DS Outscale VM\n",
             P_SYS_VENDOR: "Not 3DS Outscale\n",
         },
+        "mocks": [MOCK_VIRT_IS_KVM],
     },
     "Ec2-Outscale-negative-productname": {
         "ds": "Ec2",
@@ -1867,6 +1873,7 @@ VALID_CFG = {
             P_PRODUCT_NAME: "Not 3DS Outscale VM\n",
             P_SYS_VENDOR: "3DS Outscale\n",
         },
+        "mocks": [MOCK_VIRT_IS_KVM],
     },
 }
 
```

holmanb (changes requested):
See last two comments and suggested changes.

BirknerAlex:
I executed it on a Ubuntu bare metal node. looking good now. Thanks for the help!

```root@linux-test-node:~/cloud-init# tox -e py3 -- tests/unittests/test_ds_identify.py
py3: commands[0]> .tox/py3/bin/python -m pytest --durations 10 -m 'not hypothesis_slow' tests/unittests/test_ds_identify.py
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.8.10, pytest-7.2.1, pluggy-1.0.0
cachedir: .tox/py3/.pytest_cache
rootdir: /root/cloud-init, configfile: tox.ini
plugins: cov-4.0.0, mock-3.10.0
collected 88 items                                                                                                                                                                     

tests/unittests/test_ds_identify.py ........................................................................................                                                     [100%]

=================================================================================== warnings summary ===================================================================================
.tox/py3/lib/python3.8/site-packages/_pytest/config/__init__.py:1240
  /root/cloud-init/.tox/py3/lib/python3.8/site-packages/_pytest/config/__init__.py:1240: PytestRemovedIn8Warning: The --strict option is deprecated, use --strict-markers instead.
    self.issue_config_time_warning(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================= slowest 10 durations =================================================================================
0.05s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_ovf_on_vmware_iso_found_by_cdrom_with_matching_fs_label
0.02s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_ovf_on_vmware_iso_found_by_cdrom_with_different_size
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_aliyun_over_ec2
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_openstack_on_non_intel_is_maybe
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_config_drive_interacts_with_ibmcloud_config_disk
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_ovf_on_vmware_iso_found_by_cdrom_with_ovf_schema_match
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_on_vmware_open_vm_tools_64
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_on_vmware_open_vm_tools_aarch64_linux_gnu
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_vmware_on_vmware_when_vmware_customization_is_enabled
0.01s call     tests/unittests/test_ds_identify.py::TestDsIdentify::test_default_ovf_returns_not_found_on_azure
============================================================================ 88 passed, 1 warning in 0.73s =============================================================================
.pkg: _exit> python /usr/local/lib/python3.8/dist-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  py3: OK (9.24=setup[8.23]+cmd[1.00] seconds)
  congratulations :) (9.30 seconds)
```

holmanb (approved):
Thanks, @BirknerAlex, for this effort! I did some more testing on non-openstack platforms to ensure no unwanted surprises, it looks good to me.

Regarding LP 1815990: Since it has multiple components (which I enumerated in [a comment earlier](https://bugs.launchpad.net/cloud-init/+bug/1815990/comments/4)), I think that it would be fair to call the core requirement "properly support openstack bare metal" and set it to "fix committed". I'll open a separate bug to track allowing users to force datasource detection using commandline parameters.

BirknerAlex:
@holmanb That won't work, we do not have a PXE environment, Ironic boots the live iso via ILO or IDRAC and then copies the OS image to the disk (dd). After this the bare metal node gets rebooted and is just booting from disk, without involving PXE at all. 

It's really required that the OpenStack datasource gets loaded or at least tried when booting on a bare metal hardware otherwise public images won't work and we still need to patch the existing images.

PR #1924, opened 2022-12-19
make the same authentication behavior for arm and x86 machine

xiaoge1001:
fix https://bugs.launchpad.net/cloud-init/+bug/1999762

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1258,7 +1258,7 @@ dscheck_OpenStack() {
 
     # LP: #1715241 : arch other than intel are not identified properly.
     case "$DI_UNAME_MACHINE" in
-        i?86|x86_64) :;;
+        i?86|x86_64|aarch64) :;;
         *) return ${DS_MAYBE};;
     esac
 
@@ -1685,9 +1685,7 @@ parse_policy() {
     local def=""
     case "$DI_UNAME_MACHINE" in
         # these have dmi data
-        i?86|x86_64) def=${DI_DEFAULT_POLICY};;
-        # aarch64 has dmi, but not currently used (LP: #1663304)
-        aarch64) def=${DI_DEFAULT_POLICY_NO_DMI};;
+        i?86|x86_64|aarch64) def=${DI_DEFAULT_POLICY};;
         *) def=${DI_DEFAULT_POLICY_NO_DMI};;
     esac
     local policy="$1"


xiaoge1001:
Can someone review the code?

xiaoge1001:
> Hello! Thank you for this proposed change to cloud-init. This pull request is now marked as stale as it has not seen any activity in 14 days. If no activity occurs within the next 7 days, this pull request will automatically close.
> 
> If you are waiting for code review and you are seeing this message, apologies! Please reply, tagging TheRealFalcon, and he will ensure that someone takes a look soon.
> 
> (If the pull request is closed and you would like to continue working on it, please do tag TheRealFalcon to reopen it.)

What should I do to put TheRealFalcon tag？

TheRealFalcon:
@xiaoge1001 , sorry about that. The reviewers have been on holiday. We should be getting back to reviews shortly.

tools/ds-identify
```diff
@@ -1685,9 +1685,7 @@ parse_policy() {
     local def=""
     case "$DI_UNAME_MACHINE" in
         # these have dmi data
-        i?86|x86_64) def=${DI_DEFAULT_POLICY};;
-        # aarch64 has dmi, but not currently used (LP: #1663304)
```
igalic, line 1689: if this comment is wrong now

tools/ds-identify
```diff
@@ -1258,7 +1258,7 @@ dscheck_OpenStack() {
 
     # LP: #1715241 : arch other than intel are not identified properly.
     case "$DI_UNAME_MACHINE" in
-        i?86|x86_64) :;;
```
igalic, line 1261: what about the one up here?

igalic (commented):
❓

TheRealFalcon (commented):
@xiaoge1001 , can you help me understand what you're trying to accomplish here? These aarch64 cases are here because DMI isn't available on ARM. They basically say "If we're on an ARM machine and can't detect our datasource via DMI, barrel ahead anyway and let cloud-init figure it out in a later stage". Your changes here will cause ds-identify to indicate "no datasource found, cloud-init shouldn't run at all" on an ARM instance with no DMI.  Maybe DMI is available on some ARM platforms now (I honestly don't know and haven't checked), but
1. I can't see how this would fix a problem of not being able to login.
2. It won't be available everywhere, so we can't make this change universally.

I added a comment to LP: #1999762. I think we should confirm  the root cause there first.

xiaoge1001:
> @xiaoge1001 , sorry about that. Nothing else needs to be done. The reviewers have been on holiday. We should be getting back to reviews shortly.

Have a good holiday, everyone.

tools/ds-identify
xiaoge1001: Sorry, I didn't notice that before. I'll check again.

igalic:
unfortunately, mostly by testing.
if dmidecode (doesn't) work on that specific hardware, then that's your answer — for that hardware/kernel/libc/compiler, at that point in time.

this is the only open bug @ dmidecode: http://savannah.nongnu.org/bugs/index.php?55026

TheRealFalcon:
> How do I know if DMI is available?

Aspects of the hardware should be exposed under `/sys/class/dmi`. Cloud-init can use these values to quickly detect which platform/datasource it is being run on.

xiaoge1001:
> > How do I know if DMI is available?
> 
> Aspects of the hardware should be exposed under `/sys/class/dmi`. Cloud-init can use these values to quickly detect which platform/datasource it is being run on.

I check my  the ARM machine. The /sys/class/dmi/id/ directory exists. The directory contains the chassis_type, product_sku, product_uuid, modalias ..., which are the same as those in the x86 environment.

TheRealFalcon:
> I check my the ARM machine. The /sys/class/dmi/id/ directory exists. The directory contains the chassis_type, product_sku, product_uuid, modalias ..., which are the same as those in the x86 environment.

Yes, but I'm still not sure this is true everywhere yet, including on openstack where it was initially reported. Do you have reason to believe that the changes represented here will fix your problem?

xiaoge1001:
> > I check my the ARM machine. The /sys/class/dmi/id/ directory exists. The directory contains the chassis_type, product_sku, product_uuid, modalias ..., which are the same as those in the x86 environment.
> 
> Yes, but I'm still not sure this is true everywhere yet, including on openstack where it was initially reported. Do you have reason to believe that the changes represented here will fix your problem?

After verification, this modification can solve my problem.

[root@localhost ~]# cat /sys/class/dmi/id/product_name
KVM Virtual Machine

xiaoge1001:
After installing the cloud-init package and reboot the machine, I find that the value of PasswordAuthentication in /etc/ssh/sshd_config is changed to no.

xiaoge1001:
https://github.com/canonical/cloud-init/blob/main/config/cloud.cfg.tmpl#L49  
`ssh_pwauth:   false`

xiaoge1001:
Compare the ds-identify.log files on the x86 and ARM machine and find the following key differences:

x86:
ON_NOTFOUND=disabled
No ds found [mode=search, notfound=disabled]. Disabled cloud-init [1]

arm:
ON_NOTFOUND=enabled
check for 'OpenStack' returned maybe
1 datasources returned maybe: OpenStack

PR #1926, opened 2022-12-22
Use btrfs enquque when available

rjschwei:
## Proposed Commit Message
enqueue the btrfs resize operation when supported

```
Use btrfs enquque when available

btrfs has operations that are blocking and when we try to resize a btrfs filesystem we may be in a race condition with blocking operations. Use the enqueue feature introduced in btrfs 5.10 to queue our resize request until resize if possible to avoid the race condition.
```

## Test Steps
This is a race condition and as such difficult to reproduce. Introduce the need for btrfs to do a resize blocking operation, such as rebalancing. Then trigger the resize request, it will fail. With a sufficiently new version of btrfs and this new code the resize will succeed.

## Checklist:

 - [x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ x] I have updated or added any unit tests accordingly

### DIFF
--- a/cloudinit/config/cc_resizefs.py
+++ b/cloudinit/config/cc_resizefs.py
@@ -60,15 +60,28 @@ def _resize_btrfs(mount_point, devpth):
     if not util.mount_is_read_write(mount_point) and os.path.isdir(
         "%s/.snapshots" % mount_point
     ):
-        return (
+        cmd = [
             "btrfs",
             "filesystem",
             "resize",
             "max",
             "%s/.snapshots" % mount_point,
-        )
+        ]
     else:
-        return ("btrfs", "filesystem", "resize", "max", mount_point)
+        cmd = ["btrfs", "filesystem", "resize", "max", mount_point]
+
+    # btrfs has exclusive operations and resize may fail if btrfs is busy
+    # doing one of the operations that prevents resize. As of btrfs 5.10
+    # the resize operation can be queued
+    btrfs_with_queue = util.Version().from_str("5.10")
+    system_btrfs_ver = util.Version().from_str(
+        subp.subp(["btrfs", "--version"])[0].split("v")[-1].strip()
+    )
+    if system_btrfs_ver >= btrfs_with_queue:
+        idx = cmd.index("resize")
+        cmd.insert(idx + 1, "--enqueue")
+
+    return tuple(cmd)
 
 
 def _resize_ext(mount_point, devpth):
--- a/cloudinit/config/cc_resizefs.py
+++ b/cloudinit/config/cc_resizefs.py
@@ -73,8 +73,8 @@ def _resize_btrfs(mount_point, devpth):
     # btrfs has exclusive operations and resize may fail if btrfs is busy
     # doing one of the operations that prevents resize. As of btrfs 5.10
     # the resize operation can be queued
-    btrfs_with_queue = util.Version().from_str("5.10")
-    system_btrfs_ver = util.Version().from_str(
+    btrfs_with_queue = util.Version.from_str("5.10")
+    system_btrfs_ver = util.Version.from_str(
         subp.subp(["btrfs", "--version"])[0].split("v")[-1].strip()
     )
     if system_btrfs_ver >= btrfs_with_queue:
--- a/tests/unittests/config/test_cc_resizefs.py
+++ b/tests/unittests/config/test_cc_resizefs.py
@@ -444,10 +444,12 @@ class TestMaybeGetDevicePathAsWritableBlock(CiTestCase):
 
     @mock.patch("cloudinit.util.mount_is_read_write")
     @mock.patch("cloudinit.config.cc_resizefs.os.path.isdir")
-    def test_resize_btrfs_mount_is_ro(self, m_is_dir, m_is_rw):
+    @mock.patch("cloudinit.subp.subp")
+    def test_resize_btrfs_mount_is_ro(self, m_subp, m_is_dir, m_is_rw):
         """Do not resize / directly if it is read-only. (LP: #1734787)."""
         m_is_rw.return_value = False
         m_is_dir.return_value = True
+        m_subp.return_value = ("btrfs-progs v4.19 \n", "")
         self.assertEqual(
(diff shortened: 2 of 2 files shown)


cloudinit/config/cc_resizefs.py
```diff
@@ -60,15 +60,28 @@ def _resize_btrfs(mount_point, devpth):
...
+        cmd = ["btrfs", "filesystem", "resize", "max", mount_point]
+
+    # btrfs has exclusive operations and resize may fail if btrfs is busy
+    # doing one of the operations that prevents resize. As of btrfs 5.10
+    # the resize operation can be queued
+    btrfs_with_queue = util.Version().from_str("5.10")
+    system_btrfs_ver = util.Version().from_str(
+        subp.subp(["btrfs", "--version"])[0].split("v")[-1].strip()
+    )
+    if system_btrfs_ver >= btrfs_with_queue:
```
igalic, line 80: maybe put the messy stuff into a function, `btrfs_ver_gte(str: cmp) -> bool`

igalic (commented):
looks good, minor suggestion

cloudinit/config/cc_resizefs.py
rjschwei: Sorry what is messy about the `>=` operator?

igalic: i was thinking that from Transforming the string all the way to the comparison could be put into its own function

rjschwei: Well, then `btrfs_ver_gte(str: cmp) -> bool` as a signature would be somewhat confusing IMHO. `btrfs_has_queue_resize() -> bool` would match that idea better. Then again from my perspective the not so nice part is really the indexing into the command list.

holmanb (commented):
Note that the blocking poll loop is [1 minute](https://github.com/kdave/btrfs-progs/blob/1169f4ee63d900b25d9828a539cee4f59f8e9ad7/common/utils.c#L1289
).
 
This is better than current state however, which is that the command [will immediately fail](https://github.com/kdave/btrfs-progs/commit/e198c6674a440ec699e068c338e66b2632962197).

igalic:
> Note that the blocking poll loop is [1 minute](https://github.com/kdave/btrfs-progs/blob/1169f4ee63d900b25d9828a539cee4f59f8e9ad7/common/utils.c#L1289).
> 
> This is better than current state however, which is that the command [will immediately fail](https://github.com/kdave/btrfs-progs/commit/e198c6674a440ec699e068c338e66b2632962197).

might be worth adding this information as comment or as part of the commit message

igalic:
> Note that the blocking poll loop is [1 minute](https://github.com/kdave/btrfs-progs/blob/1169f4ee63d900b25d9828a539cee4f59f8e9ad7/common/utils.c#L1289).
> 
> This is better than current state however, which is that the command [will immediately fail](https://github.com/kdave/btrfs-progs/commit/e198c6674a440ec699e068c338e66b2632962197).

might be worth adding this information as comment or as part of the commit message

cloudinit/config/cc_resizefs.py
```diff
@@ -60,15 +60,28 @@ def _resize_btrfs(mount_point, devpth):
...
-        )
+        ]
     else:
-        return ("btrfs", "filesystem", "resize", "max", mount_point)
+        cmd = ["btrfs", "filesystem", "resize", "max", mount_point]
+
+    # btrfs has exclusive operations and resize may fail if btrfs is busy
+    # doing one of the operations that prevents resize. As of btrfs 5.10
+    # the resize operation can be queued
+    btrfs_with_queue = util.Version().from_str("5.10")
```
holmanb, line 76: fyi: Since `from_str()` is a class method, creating Version object instances isn't necessary. It could be this instead:
```python
    btrfs_with_queue = util.Version.from_str("5.10")
```

This line and the next would save us a couple of object instantiations, which I would expect to be insignificant but measurable.

cloudinit/config/cc_resizefs.py
```diff
@@ -60,15 +60,28 @@ def _resize_btrfs(mount_point, devpth):
...
+    system_btrfs_ver = util.Version().from_str(
+        subp.subp(["btrfs", "--version"])[0].split("v")[-1].strip()
```
holmanb, line 78: fyi: `subp` has a named tuple return value for readability:

```python
subp.subp(["btrfs", "--version"]).stdout.split("v")[-1].strip()
```

The mocked return values would have to use `subp.SubpResult()` to pass, if you make this change.

holmanb (approved):
I left a couple of nits, but nothing that I'm adamant about changing.

5.10rc1 was the first to contain this change, so blocking on 5.10 makes sense to me. I'm a little concerned that the version parsing might be fragile should the version format string change, but spot checked a few different versions and haven't seen anything that would be an issue today, so I think current implementation is probably good.

The included test coverage looks good.

@rjschwei I'll leave this open in case you want to make the changes I suggested, otherwise I'll merge this in a couple of days if I don't hear back.

TheRealFalcon:
> Note that the blocking poll loop is 1 minute

If the resize is queued but not finished, could that have negative impacts on disk setup or mount modules? I realized we will have failed previously anyway, but is that something we can easily account for here?

rjschwei:
> I left a couple of nits, but nothing that I'm adamant about changing.
> 
> 5.10rc1 was the first to contain this change, so blocking on 5.10 makes sense to me. I'm a little concerned that the version parsing might be fragile should the version format string change, but spot checked a few different versions and haven't seen anything that would be an issue today, so I think current implementation is probably good.
> 
> The included test coverage looks good.
> 
> @rjschwei I'll leave this open in case you want to make the changes I suggested, otherwise I'll merge this in a couple of days if I don't hear back.

Saved us the Version object instantiation by switching to `Version.from_str()` I'd like to leave everything else as is.

holmanb:
> Saved us the Version object instantiation by switching to Version.from_str() I'd like to leave everything else as is.

That works for me. Thanks for this!

I'll update the commit message to wrap the length, and include the aforementioned details about blocking / queueing behavior to the commit log.

PR #1931, opened 2022-12-28
network/netplan: add gateways as on-link when necessary

sbraz:
## Proposed Commit Message

```
network/netplan: add gateways as on-link when necessary

When the gateway isn't part of the subnet's network, the "on-link" flag is required for the route to work.

LP: #2000596
```

## Additional Context
This is a fix for https://bugs.launchpad.net/cloud-init/+bug/2000596
I don't know if I should also do this for additional `routes`, please let me know.

## Test Steps

I have added a unit test which adds an on-link IPv4 and IPv6 gateways.

## Checklist:

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -1,6 +1,7 @@
 # This file is part of cloud-init.  See LICENSE file ...
 
 import copy
+import ipaddress
 import os
 import textwrap
 from typing import Optional, cast
@@ -120,6 +121,28 @@ def _extract_addresses(config: dict, entry: dict, ifname, features=None):
                     "via": subnet.get("gateway"),
                     "to": "default",
                 }
+                try:
+                    subnet_gateway = ipaddress.ip_address(subnet["gateway"])
+                    subnet_network = ipaddress.ip_network(addr, strict=False)
+                    # If the gateway is not contained within the subnet's
+                    # network, mark it as on-link so that it can still be
+                    # reached.
+                    if subnet_gateway not in subnet_network:
+                        LOG.debug(
+                            "Gateway %s is not contained within subnet %s,"
+                            " adding on-link flag",
+                            subnet["gateway"],
+                            addr,
+                        )
+                        new_route["on-link"] = True
+                except ValueError as e:
+                    LOG.warning(
+                        "Failed to check whether gateway %s"
+                        " is contained within subnet %s: %s",
+                        subnet["gateway"],
+                        addr,
+                        e,
+                    )
                 routes.append(new_route)
             if "dns_nameservers" in subnet:
                 nameservers += _listify(subnet.get("dns_nameservers", []))
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -6289,6 +6289,44 @@ class TestNetplanNetRendering:
                 """,
                 id="one_subnet_old_new_gateway46",
             ),
+            # Assert gateways outside of the subnet's network are added with
+            # the on-link flag
+            pytest.param(
+                """
+                version: 1
+                config:
+                  - type: physical
+                    name: interface0
+                    mac_address: '00:11:22:33:44:55'
+                    subnets:
+                      - type: static
+                        address: 192.168.23.14/24
+                        gateway: 192.168.255.1
+                      - type: static
+                        address: 2001:cafe::/64
+                        gateway: 2001:ffff::1
+                """,
+                """
+                network:
+                  version: 2
+                  ethernets:
+                    interface0:
+                      addresses:
+                      - 192.168.23.14/24
+                      - 2001:cafe::/64
+                      match:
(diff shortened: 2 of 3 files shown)


sbraz:
I'm also wondering if I should add a log in case users wonder why the `on-link` flag is present.

igalic (approved):
I think this looks good as is, but if you think a debug log would be good, you can add that too

sbraz:
@igalic I've added the debug log. We also discussed on IRC the possibility of adding the flag unconditionally, as ifupdown does, so it might be worth waiting a bit before merging this.

cloudinit/net/netplan.py
```diff
@@ -120,6 +121,17 @@ def _listify(obj, token=" "):
                     "via": subnet.get("gateway"),
                     "to": "default",
                 }
+                subnet_gateway = ipaddress.ip_address(subnet["gateway"])
+                subnet_network = ipaddress.ip_network(addr, strict=False)
+                # If the gateway is not contained within the subnet's network, mark it as
```
igalic, line 126: this line is too long

igalic (changes requested):
please fix these flake8 issues

sbraz:
Sorry about that, I forgot to re-run the format checks locally. I hope I used the preferred way to split the long string.

cloudinit/net/netplan.py
```diff
@@ -120,6 +121,18 @@ def _listify(obj, token=" "):
                     "via": subnet.get("gateway"),
                     "to": "default",
                 }
+                subnet_gateway = ipaddress.ip_address(subnet["gateway"])
```
TheRealFalcon, line 124: I'm being overly cautious, but can we wrap this in a try/except? If we get an exception here (i.e., a bad value passed to us), I don't think it gets caught anywhere, and  I don't think we should crash the entire process over it.

TheRealFalcon (changes requested):
Thanks for adding this, and thanks for the test.

I left one request inline, but otherwise LGTM.

cloudinit/net/netplan.py
sbraz: OK I wrapped everything around a try/except. `addr` seems to get before though, should I move it outside of the block?

TheRealFalcon: Sorry, not sure I follow the question. What do you want to move outside of what block?

sbraz: My fault, I missed a word. I meant that `addr` gets checked before `netplan.py` is called:
```
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 767, in status_wrapper
    ret = functor(name, args)
  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 433, in main_init
    init.apply_network_config(bring_up=bring_up_interfaces)
  File "/usr/lib/python3/dist-packages/cloudinit/stages.py", line 940, in apply_network_config
    return self.distro.apply_network_config(
  File "/usr/lib/python3/dist-packages/cloudinit/distros/__init__.py", line 268, in apply_network_config
    network_state = parse_net_config_data(netconfig, renderer=renderer)
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 1103, in parse_net_config_data
    nsi.parse_config(skip_broken=skip_broken)
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 303, in parse_config
    self.parse_config_v1(skip_broken=skip_broken)
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 319, in parse_config_v1
    handler(self, command)
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 115, in decorator
    return func(self, command, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 396, in handle_physical
    subnets = _normalize_subnets(command.get("subnets"))
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 1079, in _normalize_subnets
    return [_normalize_subnet(s) for s in subnets]
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 1079, in <listcomp>
    return [_normalize_subnet(s) for s in subnets]
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 944, in _normalize_subnet
    _normalize_net_keys(
  File "/usr/lib/python3/dist-packages/cloudinit/net/network_state.py", line 994, in _normalize_net_keys
    raise ValueError(f"Address {addr} is not a valid ip address")
ValueError: Address p is not a valid ip address
```

So I thought maybe we could move `subnet_network = ipaddress.ip_network(addr, strict=False)` outside of the `try` block because it should not fail. But we can also leave it for consistency.

TheRealFalcon: Gotcha. I think we can keep it as-is for consistency.

TheRealFalcon (approved):
Thanks!

PR #1935, opened 2023-01-06
RFC: Option to warn on duplicate mac

TheRealFalcon:
## Proposed Commit Message

```
TODO
```

WIP...looking for feedback. Tests currently fail.

This feels hacky to me, but I can't think of a better way of doing this that doesn't involve sending the config through multiple levels of networking functions that currently don't take the config. That's certainly doable but would be a much more significant refactor. I'm open to any better ideas.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -999,17 +999,21 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
     """Build a dictionary of tuples {mac: name}.
 
     Bridges and any devices that have a 'stolen' mac are excluded."""
+    from cloudinit.stages import fetch_base_config
+
     ret: dict = {}
     driver_map: dict = {}
+    warn_on_duplicate_mac = util.get_cfg_option_bool(
+        fetch_base_config(), "warn_on_duplicate_mac", False
+    )
     for name, mac, driver, _devid in get_interfaces(
         blacklist_drivers=blacklist_drivers
     ):
         if mac in ret:
-            raise_duplicate_mac_error = True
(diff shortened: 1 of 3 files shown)


cjp256:
The detection of "duplicate mac" raises an error because cloud-init's data structures / APIs that will behave improperly due to the assumption that there are no duplicate macs.  It makes more sense to me to fix the APIs & callers than add a toggle to conditionally warn about it?  I'm hoping with the eventual refactor this becomes a non-issue?

TheRealFalcon:
@cjp256 , we got a request from somebody using a switch where all ports on the switch intentionally have the same mac address. I agree that this is something we'd ideally fix with the refactor, but in the mean time, we're looking for a quicker solution.

igalic:
what do we do with the quickfix after the refactoring?
I suppose we can just ignore the option…

TheRealFalcon:
If at that point it's no longer relevant, we can just remove it.

blackboxsw:
We discussed this PR side-channel and reach consensus. 

Since this PR proposes a short term fix in lieu of the more involved network refactor to cope with duplicate mac addresses, it is best to just extend the static driver definitions for drivers we know we have to handle both "hv_netvsc" and "mscc_felix" use-cases in this pass.

In a separate effort, as part of the network refactor work,  we can spend more effort on the long-term, resilient network refactor that can better handle configuration of virtual devices/VF (virtual functions) where duplicate MACs are expected early in boot if cloud-init happens to enumerate network interfaces before drivers have fully initialized the leader/subordinate relationships for those devices or switches.

TheRealFalcon:
Closing in favor of #1988

PR #1936, opened 2023-01-06
sources/azure: minor refactor for metadata source detection logic

cjp256:
- Initialize md and cfg to the fallback used when no OVF is found and IMDS is required.

- Rename metadata_source -> ovf_source and drop usage of "IMDS" as a valid value.

- Set `self.seed` to "IMDS" when ovf_source is unset.

- Remove late check for metadata source.  This is already done by the earlier check where we'll fail with "No OVF or IMDS available".

- Move "Found provisioning metadata" diagnostic up to where we read OVF.  Suggesting it was "IMDS" prior to querying IMDS is misleading.

- Add warning when falling back to IMDS-only provisioning.


TheRealFalcon (approved):
LGTM

PR #1937, opened 2023-01-06
Order cloud-init.service after dbus.socket on Ubuntu

TheRealFalcon:
## Proposed Commit Message

```
Order cloud-init.service after dbus.socket on Ubuntu

Calling "netplan apply" results in an error if cloud-init runs before 
dbus is ready, so ensure cloud-init.service starts after dbus.socket.

LP: #1997124
```

## Additional Context
This will increase boot time for launches that would have otherwise hit the race.

## Test Steps
Our current regression tests found this, but since there is a race involved, it won't be deterministic.

### DIFF
--- a/systemd/cloud-init.service.tmpl
+++ b/systemd/cloud-init.service.tmpl
@@ -12,6 +12,9 @@ After=systemd-networkd-wait-online.service
 {% if variant in ["ubuntu", "unknown", "debian"] %}
 After=networking.service
 {% endif %}
+{% if variant == "ubuntu" %}
+After=dbus.socket
+{% endif %}
 {% if variant in ["almalinux", "centos", "cloudlinux", "eurolinux", "fedora",
                   "miraclelinux", "openEuler", "openmandriva", "rhel", "rocky", "virtuozzo"] %}
 After=network.service


TheRealFalcon:
I set this to WIP due to failures. It's transient, but sometimes cloud-init.service doesn't appear to run at all. I haven't gotten to the root cause but I'm wondering if it's related to the timeout in the spec file.

systemd/cloud-init.service.tmpl
```diff
@@ -12,6 +12,9 @@ After=systemd-networkd-wait-online.service
 {% if variant in ["ubuntu", "unknown", "debian"] %}
 After=networking.service
 {% endif %}
+{% if variant == "ubuntu" %}
+After=dbus.socket
+{% endif %}
```
blackboxsw, line 17: @TheRealFalcon I actually think this should be placed in cloud-init-local.service as with a `Requires=dbus.socket` to make sure that it is started and running by the time `cloud-init.service` is started. .... In testing it seems to work, but, I'm poking around to prove to  myself why this ordering works as well as discovering why cloud-init.service is getting punted otherwise.

TheRealFalcon: As mentioned by Brett, this could add several seconds to boot time:
```
holmanb@elm:~$ systemd-analyze critical-chain dbus.socket
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

dbus.socket @5.474s
└─sysinit.target @5.377s
  └─cloud-init.service @4.647s +692ms
    └─cloud-init-local.service @1.129s +2.149s
      └─systemd-remount-fs.service @1.095s +31ms
        └─systemd-fsck-root.service @971ms +75ms
          └─systemd-journald.socket @754ms
            └─system.slice @533ms
              └─-.slice @533ms
```

TheRealFalcon:
Based on the analysis by @blackboxsw (thanks!) in https://bugs.launchpad.net/cloud-init/+bug/1997124 , it's clear this PR isn't the correct solution, so I'm going to close this until we come up with a better solution.

PR #1938, opened 2023-01-06
Fix issue regarding cache and region codes

eb3095:
## Proposed Commit Message

```
Fix a bug related to the region being reassigned and breaking init-local when cache is used. 
```

## Additional Context
image

## Test Steps
* Deploy vultr instance
* Reboot

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceVultr.py
+++ b/cloudinit/sources/DataSourceVultr.py
@@ -50,10 +50,12 @@ class DataSourceVultr(sources.DataSource):
         self.metadata = self.get_metadata()
         self.metadata["instance-id"] = self.metadata["instance-v2-id"]
         self.metadata["local-hostname"] = self.metadata["hostname"]
-        region = self.metadata["region"]["regioncode"]
-        if "countrycode" in self.metadata["region"]:
-            region = self.metadata["region"]["countrycode"]
-        self.metadata["region"] = region.lower()
+        # init-local uses a cache, region wont be a dict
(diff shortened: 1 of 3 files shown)


holmanb:
I see this bug was introduced in https://github.com/canonical/cloud-init/commit/69f9a7888ea770f4dc0454be91dea225bdd5957. This datastructure is pickled, so we probably shouldn't be modifying the type of the data structure. What was wrong with the region code handling before 69f9a7888ea770f4? I don't see any explanation given in [that PR](https://github.com/canonical/cloud-init/pull/1192), and with a quick scan I'm not seeing where that gets used.

eb3095:
The data structure we used for metadata prior to that commit changed. Before we assigned a string with the VM's region's airport code, but this broke several distros that use that to setup the repos such as Ubuntu, something we were requested to resolve. However the metadata is now provided as such,

```
 "region": {
  "countrycode": "US",
  "regioncode": "EWR"
 },
 ```
 
region however is the metadata portion used so it cant just be dropped into the region variable like that. Note this fix has been tested and is working as expected and we are likely moving this patch into production for the time being.

eb3095:
Eh, I went ahead and just did that. This is cleaner and feels better overall. I can revert however if its not ideal.

holmanb (approved):
> This is cleaner and feels better overall.

Agreed. Putting the transformations in a function seems cleaner, and shifting the callsite to within `get_metadata()` avoids applying the transformation multiple times.

PR #1941, opened 2023-01-08
distros/rhel.py: _read_hostname() missing strip on "hostname"

MarkMielke:
Debugging a related issue, I found that the determination of the "previous-hostname" was wrong an RHEL-based distributions.

In `distros/__init__.py`: update_hostname() has this code:

```
            prev_hostname = self._read_hostname(prev_hostname_fn)
...
        (sys_fn, sys_hostname) = self._read_system_hostname()
...
        if sys_hostname and prev_hostname and sys_hostname != prev_hostname:
            LOG.info(
                "%s differs from %s, assuming user maintained hostname.",
                prev_hostname_fn,
                sys_fn,
            )
            return
```

sys_hostname was found to contain a trailing newline, while prev_hostname did not. `sys_hostname != prev_hostname` is therefore always true. The check effectively reduces to:

```
        if sys_hostname and prev_hostname:
```

Evidence can be found in the log file, as "assuming user maintained hostname", even when this is not true.

prev_hostname is determined using _read_hostname().

sys_hostname is determined using _read_system_hostname() which also calls _read_hostname().

However, only the prev_hostname is being stripped by _read_hostname():

```
        if self.uses_systemd() and filename.endswith("/previous-hostname"):
            return util.load_file(filename).strip()
        elif self.uses_systemd():
            (out, _err) = subp.subp(["hostname"])
            if len(out):
                return out
```

I was unable to determine when this problem was introduced. It may
exist as far back as since "self.uses_systemd()" has been returning true?

The fix is to add a call to strip on the result of the call to command "hostname" the same way the strip is done on the file containing the "previous-hostname".

## Proposed Commit Message

```

```

## Checklist:

### DIFF
--- a/cloudinit/distros/rhel.py
+++ b/cloudinit/distros/rhel.py
@@ -132,6 +132,7 @@ class Distro(distros.Distro):
             return util.load_file(filename).strip()
         elif self.uses_systemd():
             (out, _err) = subp.subp(["hostname"])
+            out = out.strip()
             if len(out):
                 return out
             else:
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -75,6 +75,7 @@ mal
 ManassehZhou
 mamercad
 manuelisimo
+MarkMielke
 marlluslustosa
 matthewruffell
 maxnet


MarkMielke:
I was unable to find an distros/rhel.py related unit tests to use as a model to change. Also, unable to sign up for Canonical contributor. So, please feel free to make a new commit of your own that has the same effect and meets whatever requirements are intended. I am only trying to help solve a problem I encountered. :-)

TheRealFalcon:
@MarkMielke , thanks for the contribution!

> Also, unable to sign up for Canonical contributor without a sponsor apparently? 

Are you referring to the part of the CLA that says 'Please add the Canonical Project Manager or contact'? If so, you can enter my name, James Falcon.

Beyond that, you would also need to add your name (alphabetically) to [the CLA signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) (as part of this PR is fine). The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html#submitting-your-first-pull-request).

MarkMielke:
Thanks. That worked.

TheRealFalcon (approved):
Thanks!

PR #1946, opened 2023-01-09
netplan: keep original permissions when 50-cloud-init.yaml exists

blackboxsw:
## Proposed Commit Message

```
netplan: keep original permissions when 50-cloud-init.yaml exists

To avoid limiting permissions of netplan config across package
upgrade, pass preserve_mode=True when writing 50-cloud-init.yaml.

Ensure any custom permissions are retained across package upgrade
even when netplan config is rewritten across system reboot
```

## Additional Context
https://warthogs.atlassian.net/browse/SC-1375

## Checklist:

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -263,9 +263,12 @@ class Renderer(renderer.Renderer):
             header += "\n"
 
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
-        util.write_file(
-            fpnplan, header + content, mode=mode, preserve_mode=True
-        )
+        if os.path.exists(fpnplan):
+            current_mode = util.get_permissions(fpnplan)
+            if current_mode & mode == current_mode:
+                # preserve mode if existing perms are more strict than default
+                mode = current_mode
+        util.write_file(fpnplan, header + content, mode=mode)
 
         if self.clean_default:
             _clean_default(target=target)
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -263,7 +263,9 @@ class Renderer(renderer.Renderer):
             header += "\n"
 
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
-        util.write_file(fpnplan, header + content, mode=mode)
+        util.write_file(
+            fpnplan, header + content, mode=mode, preserve_mode=True
+        )
 
         if self.clean_default:
             _clean_default(target=target)
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -265,7 +265,8 @@ class Renderer(renderer.Renderer):
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
         if os.path.exists(fpnplan):
             current_mode = util.get_permissions(fpnplan)
-            if current_mode & mode == current_mode:
+            current_suid_sticky_bits = current_mode & 0o7000
+            if current_mode & mode | current_suid_sticky_bits == current_mode:
                 # preserve mode if existing perms are more strict than default
                 mode = current_mode
         util.write_file(fpnplan, header + content, mode=mode)
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -265,8 +265,7 @@ class Renderer(renderer.Renderer):
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
         if os.path.exists(fpnplan):
             current_mode = util.get_permissions(fpnplan)
-            current_suid_sticky_bits = current_mode & 0o7000
-            if current_mode & mode | current_suid_sticky_bits == current_mode:
+            if current_mode & mode == current_mode:
                 # preserve mode if existing perms are more strict than default
                 mode = current_mode
         util.write_file(fpnplan, header + content, mode=mode)
--- a/tests/unittests/distros/test_netconfig.py
+++ b/tests/unittests/distros/test_netconfig.py
@@ -622,20 +622,20 @@ class TestNetCfgDistroUbuntuNetplan(TestNetCfgDistroBase):
     def test_apply_network_config_v2_passthrough_retain_orig_perms(self):
         """Custom permissions on existing netplan is kept when more strict."""
         expected_cfgs = (
-            (self.netplan_path(), V2_TO_V2_NET_CFG_OUTPUT, 0o640),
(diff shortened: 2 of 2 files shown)


cloudinit/net/netplan.py
```diff
@@ -263,7 +263,9 @@ def render_network_state(
             header += "\n"
 
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
-        util.write_file(fpnplan, header + content, mode=mode)
+        util.write_file(
+            fpnplan, header + content, mode=mode, preserve_mode=True
```
holmanb, line 267: If `preserve_mode=True` and the file exists, we might not be honoring the behavior that we describe for [NETPLAN_CONFIG_ROOT_READ_ONLY](https://github.com/canonical/cloud-init/blob/786ca97d2972c4e0555a826a81ecaf3ae40071cd/cloudinit/features.py#L64), since the existing mode might not be root read-only.

tests/unittests/distros/test_netconfig.py
```diff
@@ -619,6 +625,20 @@ def test_apply_network_config_v2_passthrough_ub(self):
             expected_cfgs=expected_cfgs.copy(),
         )
 
+    def test_apply_network_config_v2_passthrough_retain_orig_perms(self):
+        """Custom permissions on existing netplan config is kept on rewrite."""
+        expected_cfgs = {
+            self.netplan_path(): V2_TO_V2_NET_CFG_OUTPUT,
+        }
+        # ub_distro.apply_network_config(V2_NET_CFG, False)
```
aciba90, line 633:
```suggestion
```

cloudinit/net/netplan.py
```diff
@@ -263,6 +263,11 @@ def render_network_state(
             header += "\n"
 
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
+        if os.path.exists(fpnplan):
+            current_mode = util.get_permissions(fpnplan)
+            if current_mode & mode == current_mode:
```
holmanb, line 268: This bitwise test won't apply the change to files with suid/sgid/sticky bit set. I think we should be consistent in behavior for those cases as well.

blackboxsw: @holmanb Thanks adapted this logic and the unit test to ensure sticky bit is preserved across updates to the file if perms are 'more strict' than cloud-init's defaults.

holmanb: I guess suid/guid/sticky shouldn't matter for a config file.

tests/unittests/distros/test_netconfig.py
```diff
@@ -1298,7 +1315,7 @@ def test_mariner_network_config_v1_with_duplicates(self):
 
 
 def get_mode(path, target=None):
-    return os.stat(subp.target_path(target, path)).st_mode & 0o777
+    return os.stat(subp.target_path(target, path)).st_mode & 0o7777
```
holmanb, line 1318: `st_mode & 0o7777` is a noop right?

holmanb (approved):
I left a couple of comments/nits for you, but it looks fine as is if you'd rather not make suggested changes.

some extra tests I checked with:
```python3
#!/usr/bin/env python3
import pytest


def get_strictest_mode(current_mode, mode=0o600):
    current_suid_sticky_bits = current_mode & 0o7000
    if current_mode & mode | current_suid_sticky_bits == current_mode:
        # preserve mode if existing perms are more strict than default
        mode = current_mode
    return mode


@pytest.mark.parametrize(
    ("input", "expected"),
    [
        (0o400, 0o400),
        (0o700, 0o600),
        (0o644, 0o600),
        (0o200, 0o200),
        (0o400, 0o400),
    ]
)
def test_strict(input, expected):
    a = get_strictest_mode(input)
    print(f"{oct(expected)} == {oct(a)}")
    assert expected == a


@pytest.mark.parametrize(
    ("input", "expected"),
    [
        (0o400, 0o400),
        (0o700, 0o640),
        (0o644, 0o640),
        (0o200, 0o200),
        (0o400, 0o400),
        (0, 0),
    ]
)
def test_strict_0o640(input, expected):
    a = get_strictest_mode(input, mode=0o640)
    print(f"{oct(expected)} == {oct(a)}")
    assert expected == a
```

tests/unittests/distros/test_netconfig.py
blackboxsw: ohhh right. yeah this'll just return the st_mode and any sticky bits etc.... should be dropped as some of our tests want those sticky bits.

blackboxsw: I forgot about the need to mask st_mode flags like [S_IFREG  = 0o100000](https://github.com/python/cpython/blob/3.11/Lib/stat.py#L39). Leaving the mask intact.

cloudinit/net/netplan.py
blackboxsw: Probably more a misconfiguration if those bits are set on a flat config file in /etc/netplan, I agree.  I could see use-cases for shared folders with group/world write perms like /tmp, but /etc/netplan is root:root 0xo644. Even the sticky bit, if set, and non-root perm on /etc/netplan/50-cloud-init.yaml will still prevent the owning non-root user from moving, deleting or modifying that file.
```
$ ls -l /etc/netplan/
total 1
-rw------T 1 chad chad 390 Feb  4 14:14 50-cloud-init.yaml

$ mv /etc/netplan/50-cloud-init.yaml  .
mv: cannot move '/etc/netplan/50-cloud-init.yaml' to './50-cloud-init.yaml': Permission denied
```
Yeah reverting this aspect of support. We want to blow away antiquated or irrelevant bits on these conf files. Not preserve them unnecessarily.

cloudinit/net/netplan.py
```diff
@@ -263,6 +263,12 @@ def render_network_state(
             header += "\n"
 
         mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
+        if os.path.exists(fpnplan):
```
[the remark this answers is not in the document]
blackboxsw: Let's hold off on placing more functionality into util until we have other callers. util is already a very sprawling module at more than 3000 lines which adds an element of time cost/memory cost for loading irrelevant function definitions for most consuming use-cases.  

That said, maybe when we do have other callers which need to reuse the  type of behavior which preserves mode if "at least as strict", we can tackle generalizing this in a util or atomic_helper function. 

Where I can see this being defined generally is an extension or change in behavior of `util.write_file` or `atomic_helper.write_file` to either grow a new param to scope `preserve_if_more_strict` or maybe even change default behavior of `preserve_mode` to only preserve if at least as strict as the request.  Because that will involve a potential change in behavior broadly for write_file

I'll leave that discussion to a separate PR as that could represent change in behavior which affects cloudinit.stages and ssh_util modules I definitely want to defer that effort to a separate PR when we deem it necessary.

PR #1947, opened 2023-01-10
Fix cloud-user for is_rhel derivatives in cloud.cfg.tmpl

nazunalika:
## Proposed Commit Message

```
summary: Fix cloud-user for is_rhel derivatives in cloud.cfg.tmpl

The previous change in PR #1887 attempted to fix some issues created by
PR's #1431 and #1639. Unfortunately this change wrongly assumed all
derivatives were not expecting cloud-user. This attempts to correct that.
```

## Additional Context
In current installations of cloud-init in rhel/centos/other derivatives, cloud-user has generally been expected. Downstream patches generally don't deviate from cloud-user as well. This PR is simply a minor correction to an assumption in #1887.

## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -260,7 +260,7 @@ system_info:
      name: ec2-user
      lock_passwd: True
      gecos: EC2 Default User
-{% elif variant in ["rhel", "centos"] %}
+{% elif is_rhel and variant != "eurolinux" %}
      name: cloud-user
      lock_passwd: true
      gecos: Cloud User


TheRealFalcon:
Typing this out to make sure I understand properly. Prior to this (and #1887):
```
for rhel and centos:
default user = cloud-user

for almalinux, cloudlinux, eurolinux, miraclelinux, rocky, virtuozzo:
default user = <distro_name>
```
After this change:
```
For rhel, centos, almalinux, cloudlinux, miraclelinux, rocky, virtuozzo:
default user = cloud-user

for eurolinux:
default user = <distro_name>
```

Is this correct? If so, how are you able to speak on behalf of so many downstreams? Or are there downstream patches you can point to that change this to be `cloud-user` rather than the distro name?

nazunalika:
After some thought, you are correct in this. We actually had a bug report of a user wondering why we changed "cloud-user" to "rocky" in Rocky 8 and 9. In error, I made the assumption that everyone is used to "cloud-user". I imagine that this isn't really the case anyway. In fact even in our own cloud images we create, we change cloud-user to rocky anyway. So this change doesn't really make sense.

I'll close this up. Have a good weekend!

PR #1951, opened 2023-01-10
ci: doc to gh actions

holmanb:
```
test: move doc test from travis to actions

Changes:
- drop unnecessary setuptools pin
- use current RTD Python version 3.10.8
```

## Additional Context
Makes doc failures visible sooner via gh actions, plus reduce CI bill. This is mostly just copy / paste from existing gh actions tests and the current travis test - low impact / low effort.

FYI @s-makin I mentioned this would be possible today while we were wrapping up #1933.

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -79,7 +79,7 @@ jobs:
       - name: "Install Python 3.10"
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.10.8'
       - name: "Install dependencies"
         run: |
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -76,7 +76,10 @@ jobs:
         run: |
           git fetch --unshallow
           git remote add upstream https://git.launchpad.net/cloud-init
-
+      - name: "Install Python 3.10"
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
       - name: "Install dependencies"
         run: |
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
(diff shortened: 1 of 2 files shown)


.github/workflows/check_format.yml
```diff
@@ -62,3 +62,26 @@ jobs:
...
+          git fetch --unshallow
+          git remote add upstream https://git.launchpad.net/cloud-init
+
+      - name: "1. Install dependencies 2. Spellcheck 3. Build docs"
+        env:
+          TOXENV: doc
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install tox lintian
+          make check_spelling
```
TheRealFalcon, line 86: Nit: these last two lines would be better as their own run commands.

TheRealFalcon (changes requested):
This might be a little pedantic, but I think we should ensure we're using the same python version that rtd uses to build the docs. In the past, because of dependency issues, we've had issues where local/CI passes but rtd fails on a different python version. That's probably much less likely to happen after removing the old sphinx pins, but I still think it'd make our CI better.

s-makin (approved):
Definitely a nice idea, avoiding the "install python" step that it does every time. Anything that makes the process faster or more efficient gets a thumbs up from me.

holmanb:
> This might be a little pedantic, but I think we should ensure we're using the same python version that rtd uses to build the docs. In the past, because of dependency issues, we've had issues where local/CI passes but rtd fails on a different python version. That's probably much less likely to happen after removing the old sphinx pins, but I still think it'd make our CI better.

I will do that, but it may drift since RTD only allows configuring [Major.Minor version](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python). I see the most recent [build used 3.10.8](https://readthedocs.org/projects/cloudinit/builds/19129720/).

TheRealFalcon (approved):
LGTM. With the python version I'm mostly concerned about major.minor.

holmanb:
> LGTM. With the python version I'm mostly concerned about major.minor

Ah, I guess I was confused about that then, since Jammy already had the same major.minor as RTD.

PR #1953, opened 2023-01-12
machine-id: set to uninitilized to trigger regeneration on clones

blackboxsw:
Proposal for further discussion on expected behavior,
 tested on Bionic kvm and container  in LXC will successful regeeration of /etc/machine-id across first reboot
Also tested subiquity-based installs to assert regeneration of /etc/machine-id on bare-metal installs.

## Proposed Commit Message
```
machine-id: set to uninitilized to trigger regeneration on clones

In many cases, the removal or /etc/machine-id from systemd
environments results in an unbootable system.

Adapt cloud-init clean --machine-id to be systemd-aware.

systemd v. 237 and later will trigger /etc/machine-id regeneration
on next boot when the value of /etc/machine-id file is
'uninitialized\n'.

On systems without systemd, continue to remove /etc/machine-id when
cloud-init clean --machine-id is provided as there are triggers in
place to regenerate on absence of this file.

LP: #1999680
```

## Checklist:


cloudinit/cmd/clean.py
```diff
@@ -120,7 +123,12 @@ def handle_clean_args(name, args):
     """Handle calls to 'cloud-init clean' as a subcommand."""
     exit_code = remove_artifacts(args.remove_logs, args.remove_seed)
     if args.machine_id:
-        del_file(ETC_MACHINE_ID)
+        if uses_systemd():
+            # Systemd v237 and later will create a new machine-id on next boot
+            write_file(ETC_MACHINE_ID, "uninitialized\n", mode=0o444)
```
TheRealFalcon, line 128: Should we be changing permissions if the file already exists?

TheRealFalcon (approved):
One question, but +1 if it's ok.

doc/rtd/reference/cli.rst
```diff
@@ -74,9 +74,11 @@ first boot.
 * :command:`--logs`: Optionally remove all ``cloud-init`` log files in
   :file:`/var/log/`.
 * :command:`--reboot`: Reboot the system after removing artifacts.
-* :command:`--machine-id`: Remove :file:`/etc/machine-id` on this image. Best
-  practice when cloning a golden image, to ensure that the next boot of that
-  image auto-generates a unique machine ID. `More details on machine-id`_.
+* :command:`--machine-id`: Set :file:`/etc/machine-id` to ``uninitialized\n``
+  on this image for systemd environemnts. On distributions without systemd,
```
fliiiix, line 78:
```suggestion
  on this image for systemd environments. On distributions without systemd,
```

fliiiix:
Also small detail in the commit message 

```
In many cases, the removal or /etc/machine-id from systemd
environments results in an unbootable system.
```

as far as i can tell it does not actually prevent you from booting but the system is in a degenerated state for example systemd-network fails to bring up dhcp interfaces

cloudinit/cmd/clean.py
```diff
@@ -55,8 +57,9 @@ def get_parser(parser=None):
         action="store_true",
         default=False,
         help=(
-            "Remove /etc/machine-id for golden image creation."
-            " Next boot generates a new machine-id."
+            "Set /etc/machine-id to 'uninitialized\n' for golden image"
+            "creation. On next boot, systemd generates a new machine-id."
+            " Remove /etc/machine-id on non-systemd environments like FreeBSD."
```
holmanb, line 62: Nit: Why include " like FreeBSD" in the `--help` output?

PR #1957, opened 2023-01-12
add function is_virtual to distro/FreeBSD

igalic:
## Proposed Commit Message

```
add function is_virtual to distro/FreeBSD

- is_virtual property dentifies identify if the thing we're running
  is any kind of virtualization
- virtual() identifies what kind of virtualisation we're dealing with
- is_container() tells us if we're running in a container, or in FreeBSD's
  case, in a jail.
- the helper functions are @lru_cached, since this is very unlikely to
  change

Co-authored-by: Brett Holman <brett.holman@canonical.com>
```

## Additional Context
see discussions in #1923 
We can consider merging this after that PR.

## Checklist:

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -6,7 +6,9 @@
 
 import os
 import re
+from functools import lru_cache
 from io import StringIO
+from typing import Optional
 
 import cloudinit.distros.bsd
 from cloudinit import log as logging
@@ -192,5 +194,40 @@ class Distro(cloudinit.distros.bsd.BSD):
             freq=PER_INSTANCE,
         )
 
+    @lru_cache()
+    def is_container(self) -> bool:
+        """return whether we're running in a container.
+        Cached, because it's unlikely to change."""
+        jailed, _ = subp.subp(["sysctl", "-n", "security.jail.jailed"])
+        if jailed.strip() == "0":
+            return False
+        return True
 
-# vi: ts=4 expandtab
+    @lru_cache()
+    def virtual(self) -> str:
+        """return the kind of virtualisation system we're running under.
+        Cached, because it's unlikely to change."""
+        if self.is_container():
+            return "jail"
+        # map FreeBSD's kern.vm_guest to systemd-detect-virt, just like we do
+        # in ds-identify
+        VM_GUEST_TO_SYSTEMD = {
+            "hv": "microsoft",
+            "vbox": "oracle",
+            "generic": "vm-other",
+        }
+        vm, _ = subp.subp(["sysctl", "-n", "kern.vm_guest"])
+        vm = vm.strip()
+        if vm in VM_GUEST_TO_SYSTEMD:
+            return VM_GUEST_TO_SYSTEMD[vm]
+        return vm
+
+    @property
+    def is_virtual(self) -> Optional[bool]:
+        """Detect if running on a virtual machine or bare metal.
+
+        This can fail on some platforms, so the signature is Optional[bool]
+        """
+        if self.virtual() == "none":
+            return False
+        return True
--- a/tests/unittests/distros/test__init__.py
+++ b/tests/unittests/distros/test__init__.py
@@ -275,6 +275,45 @@ class TestGenericDistro(helpers.FilesystemMockingTestCase):
                 d.is_virtual, "Reflect unknown state on ProcessExecutionError"
             )
 
+    def test_virtualization_on_freebsd(self):
+        # This test function is a crime.
+        # We're using the fact that is_container() and virtual() are @lru_cache
+        # this way, we can nest the various mocks to subp.subp to target the
+        # correct function's subp call.
+
+        cls = distros.fetch("freebsd")
+        with mock.patch(
+            "cloudinit.distros.subp.subp", return_value=("", None)
+        ):
+            d = cls("freebsd", {}, None)
+        # This mock is called by `sysctl -n security.jail.jailed`
+        with mock.patch(
+            "cloudinit.distros.subp.subp",
+            side_effect=[("0\n", None), ("literaly any truthy value", None)],
+        ):
+            self.assertFalse(d.is_container())
+            d.is_container.cache_clear()
+            self.assertTrue(d.is_container())
+            d.is_container.cache_clear()
+
+        # This mock is called by `sysctl -n kern.vm_guest`
+        with mock.patch(
+            "cloudinit.distros.subp.subp",
+            side_effect=[
+                ("0\n", None), ("hv\n", None),
+                ("0\n", None), ("none\n", None),
+                ("0\n", None), ("hv\n", None),
+            ],
+        ):
+            self.assertEqual(d.virtual(), "microsoft")
+            d.is_container.cache_clear()
+            d.virtual.cache_clear()
+            self.assertEqual(d.virtual(), "none")
+            d.is_container.cache_clear()
+            d.virtual.cache_clear()
+
+            self.assertTrue(d.is_virtual)
+
 
 class TestGetPackageMirrors:
     def return_first(self, mlist):
--- a/tests/unittests/distros/test__init__.py
+++ b/tests/unittests/distros/test__init__.py
@@ -299,11 +299,13 @@ class TestGenericDistro(helpers.FilesystemMockingTestCase):
         # This mock is called by `sysctl -n kern.vm_guest`
         with mock.patch(
             "cloudinit.distros.subp.subp",
+            # fmt: off
             side_effect=[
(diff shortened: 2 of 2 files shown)


cloudinit/distros/freebsd.py
```diff
@@ -192,5 +192,29 @@ def update_package_sources(self):
...
 
-
-# vi: ts=4 expandtab
+    def is_container(self) -> bool:
+        jailed, _ = subp.subp(["sysctl", "-n", "security.jail.jailed"])
+        if jailed.strip() == "0":
+            return False
+        return True
+
+    def virtual(self) -> str:
```
igalic, line 201: I think `virtual()` could be better named `virtualization()`. or `virtualisation()`?
whatever the canonical form

which I guess is a pretty good argument for calling it `virtual()`

igalic (commented):
🤔

f-andrey:
May be add bhyve detection too `kern.vm_guest: bhyve`

igalic:
@f-andrey `bhyve` is already covered

cloudinit/distros/freebsd.py
```diff
@@ -192,5 +192,29 @@ def update_package_sources(self):
...
+        if self.is_container():
+            return "jail"
+        # map FreeBSD's kern.vm_guest to systemd-detect-virt, just like we do
+        # in ds-identify
+        VM_GUEST_TO_SYSTEMD = {
```
f-andrey, line 206: no bhyve or not this place?

igalic: 'bhyve' is reported by both `kern.vm_guest` and `systemd-detect-virt`, so no *translation* is necessary.

that's what that table is for

holmanb:
@igalic Is this ready for review? I see it's in "Draft" state but hasn't seen updates in a few days.

igalic:
@holmanb it's still in draft stage, because I don't want to interfere with https://github.com/canonical/cloud-init/pull/1923, which introduces this function to Linux

igalic:
this is pending a merge of https://github.com/canonical/cloud-init/pull/1923

tests/unittests/distros/test__init__.py
```diff
@@ -275,6 +275,32 @@ def test_virtualization_unknown(self, m_uses_systemd):
                 d.is_virtual, "Reflect unknown state on ProcessExecutionError"
             )
 
+    def test_virtualization_on_freebsd(self):
+        # This test function is a crime.
```
igalic, line 279: I'd like to draw any reviewer's attention to this crime.

igalic (commented):
Dear Reviewer…

cloudinit/distros/freebsd.py
```diff
@@ -192,5 +194,40 @@ def update_package_sources(self):
             freq=PER_INSTANCE,
         )
 
+    @lru_cache()
```
blackboxsw, line 197: When working with testing functions with lru_cache you may run into intermittent problems with unittests receiving the cached response for that function and not a parametrized response based on the unittest setup. You may have to call  the clear_cache method on that function. [See the following example](https://github.com/canonical/cloud-init/blob/main/tests/unittests/net/test_init.py#L1640-L1646)

igalic: honestly, it seems that adding lru cache helped fix the tests, so I'm completely at a loss at what's going on

cloudinit/distros/freebsd.py
```diff
@@ -192,5 +194,40 @@ def update_package_sources(self):
...
+        if vm in VM_GUEST_TO_SYSTEMD:
+            return VM_GUEST_TO_SYSTEMD[vm]
+        return vm
+
+    @property
+    def is_virtual(self) -> Optional[bool]:
+        """Detect if running on a virtual machine or bare metal.
 
-# vi: ts=4 expandtab
+        This can fail on some platforms, so the signature is Optional[bool]
```
holmanb, line 229: How would this fail in a way that returns other than `bool`?

holmanb (changes requested):
@igalic Since the cached values in the current implementation return the same values in subsequent tests, getting coverage of the non-cached value requires clearing the cache (per @blackboxsw's comment). See the following example which increases the coverage using this approach.

Also note that if we didn't use `@lru_cache`, this would be a bit simpler, because then we could just use `side_effect` to return values from mocked `subp` based on the call order and wouldn't need to manually clear the cache. Either way, we don't need to nest each mock further - we can mock the call sites which gives us a bit cleaner (less indented) test. 

```
diff --git a/tests/unittests/distros/test__init__.py b/tests/unittests/distros/test__init__.py
index 04ba4d034..b2c61dd25 100644
--- a/tests/unittests/distros/test__init__.py
+++ b/tests/unittests/distros/test__init__.py
@@ -281,25 +281,38 @@ class TestGenericDistro(helpers.FilesystemMockingTestCase):
         # this way, we can nest the various mocks to subp.subp to target the
         # correct function's subp call.
 
-        # This first mock is swallowed by initialization
+        cls = distros.fetch("freebsd")
         with mock.patch(
             "cloudinit.distros.subp.subp", return_value=("", None)
         ):
-            cls = distros.fetch("freebsd")
             d = cls("freebsd", {}, None)
-            # This mock is called by `sysctl -n security.jail.jailed`
-            with mock.patch(
-                "cloudinit.distros.subp.subp",
-                return_value=("0\n", None),
-            ):
-                self.assertFalse(d.is_container())
-                # This mock is called by `sysctl -n kern.vm_guest`
-                with mock.patch(
-                    "cloudinit.distros.subp.subp",
-                    return_value=("hv\n", None),
-                ):
-                    self.assertEqual(d.virtual(), "microsoft")
-                    self.assertTrue(d.is_virtual)
+        # This mock is called by `sysctl -n security.jail.jailed`
+        with mock.patch(
+            "cloudinit.distros.subp.subp",
+            side_effect=[("0\n", None), ("literaly any truthy value", None)],
+        ):
+            self.assertFalse(d.is_container())
+            d.is_container.cache_clear()
+            self.assertTrue(d.is_container())
+            d.is_container.cache_clear()
+
+        # This mock is called by `sysctl -n kern.vm_guest`
+        with mock.patch(
+            "cloudinit.distros.subp.subp",
+            side_effect=[
+                ("0\n", None), ("hv\n", None),
+                ("0\n", None), ("none\n", None),
+                ("0\n", None), ("hv\n", None),
+            ],
+        ):
+            self.assertEqual(d.virtual(), "microsoft")
+            d.is_container.cache_clear()
+            d.virtual.cache_clear()
+            self.assertEqual(d.virtual(), "none")
+            d.is_container.cache_clear()
+            d.virtual.cache_clear()
+
+            self.assertTrue(d.is_virtual)
 
 
 class TestGetPackageMirrors:
```

cloudinit/distros/freebsd.py
holmanb: If referring to the possibility of thrown exceptions, that shouldn't be part of the return type annotation.

igalic: it can fail on Linux.
as in, it can fail to detect what type of platform its on. we're making the return type the same as the base implementation, which is Linux 

but i haven't observed this behaviour on FreeBSD

igalic:
not sure why the Co-authored-by line doesn't show your face, @holmanb 
but, either way, thanks for the help with these tests.

I'm on the fence about leaving @lru_cache in here, it's not as useful as in places where this is called much more often.

holmanb:
> thanks for the help with these tests.

No problem.

> I'm on the fence about leaving @lru_cache in here, it's not as useful as in places where this is called much more often.

Also on the fence about it, but I'm fine with it as is. Caching increases complexity with testing and debugging (and of course memory usage), but for a distro property like this I don't expect it to cause issues.

If you're fine with the way it is now, let me know and I'll merge it.

igalic:
@holmanb  I've opted to not format this section (with `# fmt: off`/`# fmt: on`), because your code better expresses the intent behind it.

tests/unittests/distros/test__init__.py
```diff
@@ -275,6 +275,48 @@ def test_virtualization_unknown(self, m_uses_systemd):
...
+            d.is_container.cache_clear()
+            self.assertTrue(d.is_container())
+            d.is_container.cache_clear()
+
+        # This mock is called by `sysctl -n kern.vm_guest`
+        with mock.patch(
+            "cloudinit.distros.subp.subp",
+            # fmt: off
+            side_effect=[
+                ("0\n", None), ("hv\n", None),  # virtual
```
holmanb, line 305: we typically try to avoid inline comments as a point of style, but since it isn't documented anywhere and it does add value I'm not going to push back on this

holmanb (approved):
LGTM, thanks!

PR #1958, opened 2023-01-13
docs: update the link to readthedocs

CalvoM:
## Proposed Commit Message

```
docs: update the link to the readthedocs

Changing the link url to the correct url.
```

PR #1960, opened 2023-01-13
link is working because of the redirects, but fix it anyways

holmanb:
```
doc: fix link
```

## Additional Context
@CalvoM noted a broken link in https://github.com/canonical/cloud-init/pull/1958 (thanks!). A grep for "topics" in the docs showed another issue like it. This one is currently not "broken" due to the redirects, but assuming we don't want to maintain those forever lets fix this one as well.


s-makin (commented):
Actually, we probably want to avoid this issue completely by using an internal reference to the correct doc (rather than a link). I must admit, this one totally escaped my notice!

holmanb:
@s-makin Good point. Updated!

s-makin (approved):

TheRealFalcon (approved):
LGTM!

PR #1961, opened 2023-01-13
Fix wrong subp syntax in cc_set_passwords.py

TheRealFalcon:
Well this one is embarrassing. Added an integration test to check the logs.

## Proposed Commit Message

```
Fix wrong subp syntax in cc_set_passwords.py

In #1909, I added a subp call with incorrect syntax. Fixing that here.
```


holmanb (approved):
My bad on the review. Looks good now, thanks for the fix. Integration test passes locally. +1

PR #1962, opened 2023-01-14
overhaul cc_ca_certs functionality

dermotbradley:
## Proposed Commit Message

```
summary: overhaul cc_ca_certs functionality

Change "ca-certs" references to "ca_certs".

New certificates are written to individual files, with an incrementing
number as part of their filename, rather than all being placed in a
single file. This resolves issues caused when certificate files
containing more than a single certificate are placed in /etc/ssl/certs
(by utilities such as "update-ca-certificates" run by ca_certs).

Alpine / Debian / Ubuntu:

The current behaviour, whilst it works, is incorrect with regard to
the design of the underlying OS utilities for managing certificates.
For "remove_defaults" the system-installed certificate files should
not be actually deleted (otherwise it becomes problematic
if someone wishes to later re-enable one or more of them), rather
they should be deactivated and these OSes already provide the
means to do so - this MR modifies the certificate entries in the
/etc/ca-certificates.conf file by prefixing them with "!" - when the
update-ca-certificate utility is then run it will *not* place such
delimited certificates into either the /etc/ssl/certs/ directory
(via symlinks) nor add them to the (re)generated certificates
bundle file.

Additionally it is incorrect for added certificates to be placed in
the /usr/share/ca-certificates directory - this location is intended
for standard/"official" certificates, the
/usr/local/share/ca-certificates directory is intended for "local"
or "site-specific" certificates and so this PR adds them there
instead - for certs in /usr/local/share/ca-certificates the
update-ca-certificates utility will automatically use them,
there is *no* need to add their filenames to the
/etc/ca-certificates.conf file.

LP: #1931174
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ca_certs.py
+++ b/cloudinit/config/cc_ca_certs.py
@@ -8,45 +8,47 @@ import os
 from logging import Logger
 from textwrap import dedent
 
+from cloudinit import log as logging
 from cloudinit import subp, util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema, get_meta_doc
 from cloudinit.settings import PER_INSTANCE
 
+LOG = logging.getLogger(__name__)
+
 DEFAULT_CONFIG = {
-    "ca_cert_path": "/usr/share/ca-certificates/",
-    "ca_cert_filename": "cloud-init-ca-certs.crt",
+    "ca_cert_path": None,
+    "ca_cert_local_path": "/usr/local/share/ca-certificates/",
+    "ca_cert_filename": "cloud-init-ca-cert-{cert_index}.crt",
     "ca_cert_config": "/etc/ca-certificates.conf",
-    "ca_cert_system_path": "/etc/ssl/certs/",
     "ca_cert_update_cmd": ["update-ca-certificates"],
 }
 DISTRO_OVERRIDES = {
     "rhel": {
-        "ca_cert_path": "/usr/share/pki/ca-trust-source/",
-        "ca_cert_filename": "anchors/cloud-init-ca-certs.crt",
+        "ca_cert_path": "/etc/pki/ca-trust/",
+        "ca_cert_local_path": "/usr/share/pki/ca-trust-source/",
+        "ca_cert_filename": "anchors/cloud-init-ca-cert-{cert_index}.crt",
         "ca_cert_config": None,
-        "ca_cert_system_path": "/etc/pki/ca-trust/",
         "ca_cert_update_cmd": ["update-ca-trust"],
-    }
+    },
 }
 
 MODULE_DESCRIPTION = """\
-This module adds CA certificates to ``/etc/ca-certificates.conf`` and updates
-the ssl cert cache using ``update-ca-certificates``. The default certificates
-can be removed from the system with the configuration option
-``remove_defaults``.
+This module adds CA certificates to the system's CA store and updates any
+related files using the appropriate OS-specific utility. The default CA
+certificates can be disabled/deleted from use by the system with the
+configuration option ``remove_defaults``.
 
 .. note::
     certificates must be specified using valid yaml. in order to specify a
     multiline certificate, the yaml multiline list syntax must be used
 
 .. note::
-    For Alpine Linux the "remove_defaults" functionality works if the
-    ca-certificates package is installed but not if the
-    ca-certificates-bundle package is installed.
+    Alpine Linux requires the ca-certificates package to be installed in
+    order to provide the ``update-ca-certificates`` command.
 """
-distros = ["alpine", "debian", "ubuntu", "rhel"]
+distros = ["alpine", "debian", "rhel", "ubuntu"]
 
 meta: MetaSchema = {
     "id": "cc_ca_certs",
@@ -79,11 +81,11 @@ def _distro_ca_certs_configs(distro_name):
     """Return a distro-specific ca_certs config dictionary
 
     @param distro_name: String providing the distro class name.
-    @returns: Dict of distro configurations for ca-cert.
+    @returns: Dict of distro configurations for ca_cert.
     """
     cfg = DISTRO_OVERRIDES.get(distro_name, DEFAULT_CONFIG)
     cfg["ca_cert_full_path"] = os.path.join(
-        cfg["ca_cert_path"], cfg["ca_cert_filename"]
+        cfg["ca_cert_local_path"], cfg["ca_cert_filename"]
     )
     return cfg
 
@@ -100,124 +102,145 @@ def update_ca_certs(distro_cfg):
 def add_ca_certs(distro_cfg, certs):
     """
     Adds certificates to the system. To actually apply the new certificates
-    you must also call L{update_ca_certs}.
+    you must also call the appropriate distro-specific utility such as
+    L{update_ca_certs}.
 
     @param distro_cfg: A hash providing _distro_ca_certs_configs function.
     @param certs: A list of certificate strings.
     """
     if not certs:
         return
-    # First ensure they are strings...
-    cert_file_contents = "\n".join([str(c) for c in certs])
-    util.write_file(
-        distro_cfg["ca_cert_full_path"], cert_file_contents, mode=0o644
-    )
-    update_cert_config(distro_cfg)
+    # Write each certificate to a separate file.
+    for cert_index, c in enumerate(certs, 1):
+        # First ensure they are strings...
+        cert_file_contents = str(c)
+        cert_file_name = distro_cfg["ca_cert_full_path"].format(
+            cert_index=cert_index
+        )
+        util.write_file(cert_file_name, cert_file_contents, mode=0o644)
 
 
-def update_cert_config(distro_cfg):
+def disable_default_ca_certs(distro_name, distro_cfg):
     """
-    Update Certificate config file to add the file path managed cloud-init
+    Disables all default trusted CA certificates. For Alpine, Debian and
+    Ubuntu to actually apply the changes you must also call
+    L{update_ca_certs}.
+
+    @param distro_name: String providing the distro class name.
+    @param distro_cfg: A hash providing _distro_ca_certs_configs function.
+    """
+    if distro_name == "rhel":
+        remove_default_ca_certs(distro_cfg)
+    elif distro_name in ["alpine", "debian", "ubuntu"]:
+        disable_system_ca_certs(distro_cfg)
+
+        if distro_name in ["debian", "ubuntu"]:
+            debconf_sel = (
+                "ca-certificates ca-certificates/trust_new_crts " + "select no"
+            )
+            subp.subp(("debconf-set-selections", "-"), debconf_sel)
+
+
+def disable_system_ca_certs(distro_cfg):
+    """
+    For every entry in the CA_CERT_CONFIG file prefix the entry with a "!"
+    in order to disable it.
 
     @param distro_cfg: A hash providing _distro_ca_certs_configs function.
     """
     if distro_cfg["ca_cert_config"] is None:
         return
-    if os.stat(distro_cfg["ca_cert_config"]).st_size == 0:
-        # If the CA_CERT_CONFIG file is empty (i.e. all existing
-        # CA certs have been deleted) then simply output a single
-        # line with the cloud-init cert filename.
-        out = "%s\n" % distro_cfg["ca_cert_filename"]
-    else:
-        # Append cert filename to CA_CERT_CONFIG file.
-        # We have to strip the content because blank lines in the file
-        # causes subsequent entries to be ignored. (LP: #1077020)
+    header_comment = (
+        "# Modified by cloud-init to deselect certs due to user-data"
+    )
+    added_header = False
+    if os.stat(distro_cfg["ca_cert_config"]).st_size != 0:
         orig = util.load_file(distro_cfg["ca_cert_config"])
-        cr_cont = "\n".join(
-            [
-                line
-                for line in orig.splitlines()
-                if line != distro_cfg["ca_cert_filename"]
-            ]
-        )
-        out = "%s\n%s\n" % (cr_cont.rstrip(), distro_cfg["ca_cert_filename"])
-    util.write_file(distro_cfg["ca_cert_config"], out, omode="wb")
+        out_lines = []
+        for line in orig.splitlines():
+            if line == header_comment:
+                added_header = True
+                out_lines.append(line)
+            elif line == "" or line[0] in ("#", "!"):
+                out_lines.append(line)
+            else:
+                if not added_header:
+                    out_lines.append(header_comment)
+                    added_header = True
+                out_lines.append("!" + line)
+    util.write_file(
+        distro_cfg["ca_cert_config"], "\n".join(out_lines) + "\n", omode="wb"
+    )
 
 
-def remove_default_ca_certs(distro_name, distro_cfg):
+def remove_default_ca_certs(distro_cfg):
     """
-    Removes all default trusted CA certificates from the system. To actually
-    apply the change you must also call L{update_ca_certs}.
+    Removes all default trusted CA certificates from the system.
 
-    @param distro_name: String providing the distro class name.
     @param distro_cfg: A hash providing _distro_ca_certs_configs function.
     """
-    util.delete_dir_contents(distro_cfg["ca_cert_path"])
-    util.delete_dir_contents(distro_cfg["ca_cert_system_path"])
-    util.write_file(distro_cfg["ca_cert_config"], "", mode=0o644)
+    if distro_cfg["ca_cert_path"] is None:
+        return
 
-    if distro_name in ["debian", "ubuntu"]:
-        debconf_sel = (
-            "ca-certificates ca-certificates/trust_new_crts " + "select no"
-        )
-        subp.subp(("debconf-set-selections", "-"), debconf_sel)
+    LOG.debug("Deleting system CA certificates")
+    util.delete_dir_contents(distro_cfg["ca_cert_path"])
+    util.delete_dir_contents(distro_cfg["ca_cert_local_path"])
 
 
 def handle(
     name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
 ) -> None:
     """
-    Call to handle ca-cert sections in cloud-config file.
+    Call to handle ca_cert sections in cloud-config file.
 
-    @param name: The module name "ca-cert" from cloud.cfg
+    @param name: The module name "ca_cert" from cloud.cfg
     @param cfg: A nested dict containing the entire cloud config contents.
     @param cloud: The L{CloudInit} object in use.
     @param log: Pre-initialized Python logger object to use for logging.
     @param args: Any module arguments from cloud.cfg
     """
     if "ca-certs" in cfg:
-        log.warning(
+        LOG.warning(
             "DEPRECATION: key 'ca-certs' is now deprecated. Use 'ca_certs'"
             " instead."
         )
     elif "ca_certs" not in cfg:
-        log.debug(
+        LOG.debug(
             "Skipping module named %s, no 'ca_certs' key in configuration",
             name,
         )
         return
 
     if "ca-certs" in cfg and "ca_certs" in cfg:
-        log.warning(
+        LOG.warning(
             "Found both ca-certs (deprecated) and ca_certs config keys."
             " Ignoring ca-certs."
         )
     ca_cert_cfg = cfg.get("ca_certs", cfg.get("ca-certs"))
     distro_cfg = _distro_ca_certs_configs(cloud.distro.name)
 
-    # If there is a remove_defaults option set to true, remove the system
+    # If there is a remove_defaults option set to true, disable the system
     # default trusted CA certs first.
     if "remove-defaults" in ca_cert_cfg:
-        log.warning(
+        LOG.warning(
             "DEPRECATION: key 'ca-certs.remove-defaults' is now deprecated."
             " Use 'ca_certs.remove_defaults' instead."
         )
-        if ca_cert_cfg.get("remove-defaults", False):
-            log.debug("Removing default certificates")
-            remove_default_ca_certs(cloud.distro.name, distro_cfg)
-    elif ca_cert_cfg.get("remove_defaults", False):
-        log.debug("Removing default certificates")
-        remove_default_ca_certs(cloud.distro.name, distro_cfg)
+    if ca_cert_cfg.get(
+        "remove_defaults", ca_cert_cfg.get("remove-defaults", False)
+    ):
+        LOG.debug("Disabling/removing default certificates")
+        disable_default_ca_certs(cloud.distro.name, distro_cfg)
 
     # If we are given any new trusted CA certs to add, add them.
     if "trusted" in ca_cert_cfg:
         trusted_certs = util.get_cfg_option_list(ca_cert_cfg, "trusted")
         if trusted_certs:
-            log.debug("Adding %d certificates" % len(trusted_certs))
+            LOG.debug("Adding %d certificates", len(trusted_certs))
             add_ca_certs(distro_cfg, trusted_certs)
 
     # Update the system with the new cert configuration.
-    log.debug("Updating certificates")
+    LOG.debug("Updating certificates")
     update_ca_certs(distro_cfg)
 
 
--- a/tests/unittests/config/test_cc_ca_certs.py
+++ b/tests/unittests/config/test_cc_ca_certs.py
@@ -74,15 +74,19 @@ class TestConfig(TestCase):
             mock.patch.object(cc_ca_certs, "update_ca_certs")
         )
         self.mock_remove = self.mocks.enter_context(
-            mock.patch.object(cc_ca_certs, "remove_default_ca_certs")
+            mock.patch.object(cc_ca_certs, "disable_default_ca_certs")
         )
 
-    def test_no_trusted_list(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_no_trusted_list(self, _):
         """
         Test that no certificates are written if the 'trusted' key is not
         present.
         """
-        config = {"ca-certs": {}}
+        config = {"ca_certs": {}}
 
         for distro_name in cc_ca_certs.distros:
             self._mock_init()
@@ -93,9 +97,13 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 0)
 
-    def test_empty_trusted_list(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_empty_trusted_list(self, _):
         """Test that no certificate are written if 'trusted' list is empty."""
-        config = {"ca-certs": {"trusted": []}}
+        config = {"ca_certs": {"trusted": []}}
 
         for distro_name in cc_ca_certs.distros:
             self._mock_init()
@@ -106,9 +114,13 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 0)
 
-    def test_single_trusted(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_single_trusted(self, _):
         """Test that a single cert gets passed to add_ca_certs."""
-        config = {"ca-certs": {"trusted": ["CERT1"]}}
+        config = {"ca_certs": {"trusted": ["CERT1"]}}
 
         for distro_name in cc_ca_certs.distros:
             self._mock_init()
@@ -120,9 +132,13 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 0)
 
-    def test_multiple_trusted(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_multiple_trusted(self, _):
         """Test that multiple certs get passed to add_ca_certs."""
-        config = {"ca-certs": {"trusted": ["CERT1", "CERT2"]}}
+        config = {"ca_certs": {"trusted": ["CERT1", "CERT2"]}}
 
         for distro_name in cc_ca_certs.distros:
             self._mock_init()
@@ -134,7 +150,11 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 0)
 
-    def test_remove_default_ca_certs(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_remove_default_ca_certs(self, _):
         """Test remove_defaults works as expected."""
         config = {"ca_certs": {"remove_defaults": True}}
 
@@ -147,7 +167,11 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 1)
 
-    def test_no_remove_defaults_if_false(self):
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_no_remove_defaults_if_false(self, _):
         """Test remove_defaults is not called when config value is False."""
         config = {"ca_certs": {"remove_defaults": False}}
 
@@ -160,8 +184,14 @@ class TestConfig(TestCase):
             self.assertEqual(self.mock_update.call_count, 1)
             self.assertEqual(self.mock_remove.call_count, 0)
 
-    def test_correct_order_for_remove_then_add(self):
-        """Test remove_defaults is not called when config value is False."""
+    @mock.patch(
+        "cloudinit.distros.networking.subp.subp",
+        return_value=("", None),
+    )
+    def test_correct_order_for_remove_then_add(self, _):
+        """
+        Test remove_defaults is called before add.
+        """
         config = {"ca_certs": {"remove_defaults": True, "trusted": ["CERT1"]}}
 
         for distro_name in cc_ca_certs.distros:
@@ -170,9 +200,9 @@ class TestConfig(TestCase):
             conf = cc_ca_certs._distro_ca_certs_configs(distro_name)
             cc_ca_certs.handle(self.name, config, cloud, self.log, self.args)
 
+            self.assertEqual(self.mock_remove.call_count, 1)
             self.mock_add.assert_called_once_with(conf, ["CERT1"])
             self.assertEqual(self.mock_update.call_count, 1)
-            self.assertEqual(self.mock_remove.call_count, 1)
 
 
 class TestAddCaCerts(TestCase):
@@ -200,13 +230,9 @@ class TestAddCaCerts(TestCase):
                 cc_ca_certs.add_ca_certs(conf, [])
             self.assertEqual(mockobj.call_count, 0)
 
-    def test_single_cert_trailing_cr(self):
-        """Test adding a single certificate to the trusted CAs
-        when existing ca-certificates has trailing newline"""
+    def test_single_cert(self):
+        """Test adding a single certificate to the trusted CAs."""
         cert = "CERT1\nLINE2\nLINE3"
-
-        ca_certs_content = "line1\nline2\ncloud-init-ca-certs.crt\nline3\n"
-        expected = "line1\nline2\nline3\ncloud-init-ca-certs.crt\n"
 
         self.m_stat.return_value.st_size = 1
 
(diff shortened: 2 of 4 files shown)


dermotbradley:
The testcases fail because FreeBSD does a subp for "ifconfig" whenever get_cloud() is called. I tried to handle this in _mock_init as suggested by @igalic but it did not work.

The FreeBSD functionality was developed by reading documentation online - @igalic could you test this for real on FreeBSD?

Regarding RHEL/Fedora, I have been unable to determine the correct method to disable/delete system certs. I thought I could use ```trust list``` to get IDs for each cert and then for each run ```trust anchor remove "<id>"``` but it turns out all the standard certs are marked as "read only" inside the cert store and so that command will not remove them. The old behaviour of deleting files from /etc/pki/ca-trust/ and /usr/share/pki/ca-trust-source/ is wrong, especially as, for example, there are JVM and UEFI related cert stuff in /etc/pki/ca-trust/extracted/edk2 and /etc/pki/ca-trust/extracted/java

cloudinit/config/cc_ca_certs.py
```diff
@@ -107,88 +135,108 @@ def add_ca_certs(distro_cfg, certs):
...
+
+    @param distro_name: String providing the distro class name.
+    @param distro_cfg: A hash providing _distro_ca_certs_configs function.
+    """
+    if distro_name == "freebsd":
+        move_freebsd_ca_certs_to_blacklist_dir(distro_cfg)
+    elif distro_name in ["alpine", "debian", "ubuntu"]:
+        disable_system_ca_certs(distro_cfg)
 
+        if distro_name in ["debian", "ubuntu"]:
```
igalic, line 170: we could use the family here

cloudinit/config/cc_ca_certs.py
```diff
@@ -107,88 +135,108 @@ def add_ca_certs(distro_cfg, certs):
...
+
+def disable_default_ca_certs(distro_name, distro_cfg):
+    """
+    Disables all default trusted CA certificates. To actually apply the
+    changes you must also call L{update_ca_certs}.
+
+    @param distro_name: String providing the distro class name.
+    @param distro_cfg: A hash providing _distro_ca_certs_configs function.
+    """
+    if distro_name == "freebsd":
```
igalic, line 165: and here

cloudinit/config/cc_ca_certs.py
```diff
@@ -198,26 +246,37 @@ def handle(
...
     if "remove-defaults" in ca_cert_cfg:
-        log.warning(
+        LOG.warning(
             "DEPRECATION: key 'ca-certs.remove-defaults' is now deprecated."
             " Use 'ca_certs.remove_defaults' instead."
         )
-        if ca_cert_cfg.get("remove-defaults", False):
-            log.debug("Removing default certificates")
-            remove_default_ca_certs(cloud.distro.name, distro_cfg)
+        if cloud.distro.name in ["fedora", "rhel"]:
```
igalic, line 253: again, i would use family here

tests/unittests/config/test_cc_ca_certs.py
```diff
@@ -62,30 +62,35 @@ def _fetch_distro(self, kind):
...
+            mock.patch.object(cc_ca_certs, "disable_default_ca_certs")
+        )
         self.mock_update = self.mocks.enter_context(
             mock.patch.object(cc_ca_certs, "update_ca_certs")
         )
-        self.mock_remove = self.mocks.enter_context(
-            mock.patch.object(cc_ca_certs, "remove_default_ca_certs")
-        )
+        if distro_name == "freebsd":
+            mock.patch(
```
igalic, line 80: it doesn't make sense to add this mock here

tests/unittests/config/test_cc_ca_certs.py
```diff
@@ -62,30 +62,35 @@ def _fetch_distro(self, kind):
...
         Test that no certificates are written if the 'trusted' key is not
         present.
         """
-        config = {"ca-certs": {}}
+        config = {"ca_certs": {}}
 
         for distro_name in cc_ca_certs.distros:
-            self._mock_init()
+            self._mock_init(distro_name)
             cloud = get_cloud(distro_name)
```
igalic, line 94: the instanciation happens here, so this is where the mock needs to go

tests/unittests/config/test_cc_ca_certs.py
```diff
@@ -122,10 +127,10 @@ def test_single_trusted(self):
 
     def test_multiple_trusted(self):
         """Test that multiple certs get passed to add_ca_certs."""
-        config = {"ca-certs": {"trusted": ["CERT1", "CERT2"]}}
+        config = {"ca_certs": {"trusted": ["CERT1", "CERT2"]}}
 
         for distro_name in cc_ca_certs.distros:
-            self._mock_init()
+            self._mock_init(distro_name)
             cloud = get_cloud(distro_name)
```
igalic, line 134: and here,

tests/unittests/config/test_cc_ca_certs.py
```diff
@@ -139,20 +144,23 @@ def test_remove_default_ca_certs(self):
...
+                self.assertEqual(self.mock_remove.call_count, 1)
 
     def test_no_remove_defaults_if_false(self):
         """Test remove_defaults is not called when config value is False."""
         config = {"ca_certs": {"remove_defaults": False}}
 
         for distro_name in cc_ca_certs.distros:
-            self._mock_init()
+            self._mock_init(distro_name)
             cloud = get_cloud(distro_name)
```
igalic, line 164: and so forth

igalic (commented):
first look. i haven't tested it yet.
it would be cool, if someone knowledgeable could help us out the FreeBSD mocks into one place, instead of every place

cloudinit/config/cc_ca_certs.py
```diff
@@ -5,48 +5,76 @@
...
+    },
 }
 
 MODULE_DESCRIPTION = """\
-This module adds CA certificates to ``/etc/ca-certificates.conf`` and updates
-the ssl cert cache using ``update-ca-certificates``. The default certificates
-can be removed from the system with the configuration option
-``remove_defaults``.
+This module adds CA certificates to the system's OpenSSL CA directory and
+updates the OpenSSL CA cache using the appropriate OS-specific utility. The
```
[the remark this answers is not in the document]
dermotbradley: True. I'll modify that text to be more generic.

cloudinit/config/cc_ca_certs.py
```diff
@@ -107,88 +135,108 @@ def add_ca_certs(distro_cfg, certs):
     """
     if not certs:
         return
-    # First ensure they are strings...
-    cert_file_contents = "\n".join([str(c) for c in certs])
-    util.write_file(
-        distro_cfg["ca_cert_full_path"], cert_file_contents, mode=0o644
-    )
-    update_cert_config(distro_cfg)
+    # Write each certificate to a separate file, maximum 99 certificates.
```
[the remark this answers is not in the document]
dermotbradley: Are you asking why seperate file per certificate? or why maximum 99 certs?

If regarding 99 max certs, that's just an arbitary limit as I am naming files with a 2 digit value.

If it is regarding why 1 cert per file, see LP #1931174. Basically c_rehash (in Alpine, Debian, Ubuntu) creates a hash for each individual cert in /etc/ssl/certs/ but this does not work in more than a single cert present in a file.

cloudinit/config/cc_ca_certs.py
dermotbradley: I didn't look at any of the other BSDs, do they use the same mechanism/utilities for managing certificates?

cloudinit/config/cc_ca_certs.py
dermotbradley: Again, not sure if all the various RHEL derivatives use the same NSS utils for managing certificates. If so, then yes in general cc_ca_certs could have all the rhel family added as supported.

tests/unittests/config/test_cc_ca_certs.py
dermotbradley: That's right after where _mock_init is called (in several testcases) so it made sense to put the mock in a single place in _mock_init. I had previously tried putting it here but it also had no effect.

tests/unittests/config/test_cc_ca_certs.py
dermotbradley: This function is common to all the testcases that are giving the subp error and so it made sense to put the mock in a single (shared) place.

tests/unittests/config/test_cc_ca_certs.py
dermotbradley: again, this is after _mock_init is called where the mock is defined.

tests/unittests/config/test_cc_ca_certs.py
dermotbradley: again, this is after _mock_init is called where the mock is defined.

cloudinit/config/cc_ca_certs.py
dermotbradley: True. I had problems figuring out how to use family in the equivalent testcases (which loop through each distro and so need to derive osfamily from each distro name)...

cloudinit/config/cc_ca_certs.py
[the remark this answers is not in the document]
igalic: it was, indeed, about the limit
tho, i suppose most people are adding two (standard) to six (excessive) certificates, not two to six hundred certificates.

cloudinit/config/cc_ca_certs.py
igalic: Good point. dragonfly is one major release behind. but that still means it should be available.

cloudinit/config/cc_ca_certs.py
igalic: aye

tests/unittests/config/test_cc_ca_certs.py
igalic: it doesn't make sense, because it's not consumed here.
the patch goes out of scope by the time and is destroyed

tests/unittests/config/test_cc_ca_certs.py
igalic: that's interesting! that should not be happening!

holmanb:
@esposem Requesting a review for the RHEL-specific bits, if you get a chance.

holmanb (commented):
First pass, still grokking.

> I thought I could use trust list to get IDs for each cert and then for each run trust anchor remove "<id>" but it turns out all the standard certs are marked as "read only" inside the cert store and so that command will not remove them.

I poked at Fedora a little myself. I think this is a bug, since behavior doesn't match [the docs](https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/).

I've filed here:
https://bugzilla.redhat.com/show_bug.cgi?id=2163554

cloudinit/config/cc_ca_certs.py
```diff
@@ -5,48 +5,76 @@
...
     "ca_cert_update_cmd": ["update-ca-certificates"],
 }
 DISTRO_OVERRIDES = {
+    "fedora": {
+        "ca_cert_path": None,
+        "ca_cert_blacklist_path": None,
+        "ca_cert_local_path": "/usr/share/pki/ca-trust-source/anchors/",
+        "ca_cert_filename": "cloud-init-ca-cert-%(number)s.crt",
+        "ca_cert_config": None,
+        "ca_cert_update_cmd": ["update-ca-trust"],
```
esposem, line 36: Why making a "fedora" object here if it's the same (as it should be) to "rhel"?

cloudinit/config/cc_ca_certs.py
```diff
@@ -107,88 +135,108 @@ def add_ca_certs(distro_cfg, certs):
...
             "Skipping module named %s, no 'ca_certs' key in configuration",
             name,
         )
         return
 
     if "ca-certs" in cfg and "ca_certs" in cfg:
-        log.warning(
+        LOG.warning(
             "Found both ca-certs (deprecated) and ca_certs config keys."
             " Ignoring ca-certs."
```
esposem, line 241: Has nothing to do with this PR, but the warning says "ignoring ca-certs" and then below we have `ca_cert_cfg = cfg.get("ca_certs", cfg.get("ca-certs"))`.

cloudinit/config/cc_ca_certs.py
```diff
@@ -198,26 +246,37 @@ def handle(
...
+                " this distro."
+            )
+        elif ca_cert_cfg.get("remove-defaults", False):
+            LOG.debug("Disabling default certificates")
+            disable_default_ca_certs(cloud.distro.name, distro_cfg)
     elif ca_cert_cfg.get("remove_defaults", False):
-        log.debug("Removing default certificates")
-        remove_default_ca_certs(cloud.distro.name, distro_cfg)
+        if cloud.distro.name in ["fedora", "rhel"]:
+            LOG.error(
```
esposem, line 263: So basically fedora and rhel won't support this option? I didn't understand why

cloudinit/config/cc_ca_certs.py
```diff
@@ -198,26 +246,37 @@ def handle(
...
+                "Disabling default certificates is not supported for"
+                " this distro."
+            )
+        elif ca_cert_cfg.get("remove-defaults", False):
+            LOG.debug("Disabling default certificates")
+            disable_default_ca_certs(cloud.distro.name, distro_cfg)
     elif ca_cert_cfg.get("remove_defaults", False):
-        log.debug("Removing default certificates")
-        remove_default_ca_certs(cloud.distro.name, distro_cfg)
+        if cloud.distro.name in ["fedora", "rhel"]:
```
esposem, line 262: This whole `if cloud.distro.name in ["fedora", "rhel"]:` and `else` could be put in a separate function instead of duplicating the code, it seems exactly the same to me.

cloudinit/config/cc_ca_certs.py
esposem: ah ok now I got it. I would wait for https://bugzilla.redhat.com/show_bug.cgi?id=2163554 opened by @holmanb before disabling it to fedora and rhel. I will also ask if someone from our side can help fixing that in RHEL.

esposem:
Thanks for pinging me @holmanb!
I had minor comments, but I would like to have this rhel/fedora behavior fixed and correctly implemented in this PR before merging it. Or if this is really urgent, we will for sure need a new PR where we enable this feature for RHEL too.
I would wait for comments in the fedora BZ, and in the meanwhile I will look for someone at RH that can fix/backport the eventual fix for us too.

holmanb:
> Thanks for pinging me @holmanb!

No problem @esposem!

> I would wait for comments in the fedora BZ, and in the meanwhile I will look for someone at RH that can fix/backport the eventual fix for us too.

Based on the comment in the BZ, it looks like they aren't going to try to support deleting system-certs. Granted, I didn't bother explaining the use case in the BZ, but it would be nice if RHEL/Fedora had more dynamic/configurable CA management so that we could have a nicer approach than "delete them", since we're concerned that deleting them all might not always be the best thing for this feature. See the [conversation between minimal, meena and I in IRC](https://irclogs.ubuntu.com/2023/01/24/%23cloud-init.txt) for some of the concerns that have been raised.

cloudinit/config/cc_ca_certs.py
dermotbradley: I'm assuming as that is a "fall through" for either (a) ca_certs section only, or (b) ca-certs section only, or (c) both ca_certs and ca-certs sections that the behaviour is that only the ca_certs settings are used (as it comes first) when both are defined.

cloudinit/config/cc_ca_certs.py
dermotbradley: @esposem unfortunately the response in that Redhat bug referred to "local" certificates which was not what the question actually as regarding - it was regarding what is the official/"proper" way in RHEL/Fedora to disable/delete the system CA certificates.

Anyway as @holmanb suggested I have reintroduced the previous behaviour - though only for the /etc/pki/ca-trust/ directory and not for the /usr/share/pki/ca-trust-source/ directory.

cloudinit/config/cc_ca_certs.py
dermotbradley: I could change these definitions to use families instead, but that assumes all members of family "rhel" act in the same way - something that I'm not sure about.

cloudinit/config/cc_ca_certs.py
dermotbradley: ah, "freebsd" is a family, I'd assumed there was a single "bsd" family defined. Yes this could use family rather than distro_name, though when I checked in the tests thereis currrently nothing in place to handle osfamily lookups so I'd have to figure that out.

cloudinit/config/cc_ca_certs.py
dermotbradley: Not related to this PR, so resolving.

cloudinit/config/cc_ca_certs.py
dermotbradley: That code is gone now anyway as part of my reverting to old behaviour.

cloudinit/config/cc_ca_certs.py
esposem: Not sure either, I guess it will be ok as it is for now

cloudinit/config/cc_ca_certs.py
esposem: I also tried to ping the person internally, no answer so far. I guess if this is the only blocker of this PR, just go ahead, we'll introduce the proper behaviour later. If I find something I will let you know.

igalic (changes requested):
okay, first test on FreeBSD yielded nothing, because we're not enabling the module for FreeBSD in cloud.cfg.tmpl

dermotbradley:
> okay, first test on FreeBSD yielded nothing, because we're not enabling the module for FreeBSD in cloud.cfg.tmpl

Fixed now.

config/cloud.cfg.tmpl
```diff
@@ -102,10 +102,10 @@ cloud_init_modules:
 {% if variant in ["alpine", "photon"] %}
  - resolv_conf
 {% endif %}
-{% if not variant.endswith("bsd") %}
-{% if variant not in ["photon"] %}
+{% if variant in ["alpine", "debian", "fedora", "freebsd", "rhel", "ubuntu"] %}
  - ca-certs
 {% endif %}
+{% if not variant.endswith("bsd") %}
  - rsyslog
```
igalic, line 109: really need to fix this one…

dermotbradley: what's the issue?

igalic: time, and brain
https://bugs.launchpad.net/cloud-init/+bug/1798055

dermotbradley: OK but whether or not rsyslog should be enabled for any BSD is unrelated to this PR, I kept the existing logic in this regard.

igalic: yeah, sorry.
that comment was for me, really, not for you.

cloudinit/config/cc_ca_certs.py
```diff
@@ -100,95 +121,133 @@ def update_ca_certs(distro_cfg):
...
+
+    @param distro_cfg: A hash providing _distro_ca_certs_configs function.
+    """
+    if (
+        distro_cfg["ca_cert_path"] is None
+        or distro_cfg["ca_cert_blacklist_path"] is None
+    ):
+        return
+    LOG.debug("Moving system CA certificates to blacklist directory")
+    shutil.move(
```
igalic, line 220: this is, unfortunately, not idempotent:

```python-traceback
2023-02-01 00:10:59,257 - cc_ca_certs.py[DEBUG]: Moving system CA certificates to blacklist directory
2023-02-01 00:10:59,260 - handlers.py[DEBUG]: finish: init-network/config-ca-certs: FAIL: running config-ca-certs with frequency once-per-instance
2023-02-01 00:10:59,260 - util.py[WARNING]: Running module ca-certs (<module 'cloudinit.config.cc_ca_certs' from '/usr/local/lib/python3.9/site-packages/cloudinit/config/cc_ca_certs.py'>) failed
2023-02-01 00:10:59,261 - util.py[DEBUG]: Running module ca-certs (<module 'cloudinit.config.cc_ca_certs' from '/usr/local/lib/python3.9/site-packages/cloudinit/config/cc_ca_certs.py'>) failed
  File "/usr/local/lib/python3.9/shutil.py", line 825, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/certs/trusted/' -> '/usr/share/certs/blacklisted/trusted'

  File "/usr/local/lib/python3.9/site-packages/cloudinit/config/modules.py", line 246, in _run_modules
    ran, _r = cc.run(
  File "/usr/local/lib/python3.9/site-packages/cloudinit/cloud.py", line 67, in run
  File "/usr/local/lib/python3.9/site-packages/cloudinit/helpers.py", line 185, in run
    results = functor(*args)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/config/cc_ca_certs.py", line 269, in handle
    disable_default_ca_certs(cloud.distro.name, distro_cfg)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/config/cc_ca_certs.py", line 161, in disable_default_ca_certs
    move_freebsd_ca_certs_to_blacklist_dir(distro_cfg)
  File "/usr/local/lib/python3.9/site-packages/cloudinit/config/cc_ca_certs.py", line 220, in move_freebsd_ca_certs_to_blacklist_dir
    shutil.move(
  File "/usr/local/lib/python3.9/shutil.py", line 845, in move
    copy_function(src, real_dst)
  File "/usr/local/lib/python3.9/shutil.py", line 444, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/lib/python3.9/shutil.py", line 264, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/certs/trusted/'
```

let's pre-condition this `shutil.move()` on `distro_cfg["ca_cert_path"]` actually existing.

igalic: at the very least, we need to catch an exceptions here.

igalic (commented):
works pretty well, aside from the idempotency, if the module runs more than once on the same system.

cloudinit/config/cc_ca_certs.py
```diff
@@ -5,48 +5,69 @@
...
+        "ca_cert_path": "/etc/pki/ca-trust/",
+        "ca_cert_blacklist_path": None,
+        "ca_cert_local_path": "/usr/share/pki/ca-trust-source/anchors/",
+        "ca_cert_filename": "cloud-init-ca-cert-%(number)s.crt",
+        "ca_cert_config": None,
+        "ca_cert_update_cmd": ["update-ca-trust"],
+    },
+    "freebsd": {
+        "ca_cert_path": "/usr/share/certs/trusted/",
+        "ca_cert_blacklist_path": "/usr/share/certs/blacklisted/",
```
igalic, line 40: This is different between 13 and 14. In 14 it's `/usr/share/certs/untrusted`
We need to find a good way to map this

igalic (commented):
dunno what to do about this, but otherwise this looks okay.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed FreeBSD-related changes.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed FreeBSD related changes.

dermotbradley:
I have removed FreeBSD and Fedora/RHEL related changes from this PR.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed Fedora and RHEL related changes from this PR.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed Fedora and RHEL related changes from this PR.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed Fedora and RHEL related changes from this PR.

cloudinit/config/cc_ca_certs.py
dermotbradley: Removed FreeBSD related changes from this PR.

cloudinit/config/cc_ca_certs.py
```diff
@@ -8,45 +8,47 @@
...
 from cloudinit.config.schema import MetaSchema, get_meta_doc
 from cloudinit.settings import PER_INSTANCE
 
+LOG = logging.getLogger(__name__)
+
 DEFAULT_CONFIG = {
-    "ca_cert_path": "/usr/share/ca-certificates/",
-    "ca_cert_filename": "cloud-init-ca-certs.crt",
+    "ca_cert_path": None,
+    "ca_cert_local_path": "/usr/local/share/ca-certificates/",
```
holmanb, line 22: +1 to changing to /usr/local/share/ca-certificates/ per [the docs](https://ubuntu.com/server/docs/security-trust-store)

holmanb (changes requested):
Tested on ubuntu 22.04 with:
```
root@jammy-cloud1:~# cloud-init query userdata
#cloud-config
ca_certs:
  remove_defaults: true
```
Https request to google fails, as expected.
```bash
root@jammy-cloud1:~# wget https://www.google.com
--2023-02-06 19:25:19--  https://www.google.com/
Resolving www.google.com (www.google.com)... 142.250.191.132, 2607:f8b0:4009:807::2004
Connecting to www.google.com (www.google.com)|142.250.191.132|:443... connected.
ERROR: cannot verify www.google.com's certificate, issued by ‘CN=GTS CA 1C3,O=Google Trust Services LLC,C=US’:
  Unable to locally verify the issuer's authority.
To connect to www.google.com insecurely, use `--no-check-certificate'.

```
all certs are disabled after run
```
root@jammy-cloud1:~# grep -v -e '^#' -e '^!' /etc/ca-certificates.conf | wc -l
0
```

Logs look clean:
[ca.log](https://github.com/canonical/cloud-init/files/10668545/ca.log)

One difference from the old implementation in the logs is that the old implementation spammed per-cert that was disabled. Someone used to that might be surprised not to see that, but I'd rather not do that anymore anyways - it was a bit excessive and the release notes should tell users that something changed should they want to understand what changed.

I left a couple of minor change requests inline, but the rest looks good to me.

Given the security aspect of this change I'd prefer to get a review from @blackboxsw or @aciba90 before merging.

cloudinit/config/cc_ca_certs.py
```diff
@@ -8,45 +8,47 @@
...
     "ca_cert_config": "/etc/ca-certificates.conf",
-    "ca_cert_system_path": "/etc/ssl/certs/",
     "ca_cert_update_cmd": ["update-ca-certificates"],
 }
 DISTRO_OVERRIDES = {
     "rhel": {
-        "ca_cert_path": "/usr/share/pki/ca-trust-source/",
-        "ca_cert_filename": "anchors/cloud-init-ca-certs.crt",
+        "ca_cert_path": "/etc/pki/ca-trust/",
+        "ca_cert_local_path": "/usr/share/pki/ca-trust-source/anchors/",
```
holmanb, line 30: +1 [per docs](https://www.redhat.com/sysadmin/ca-certificates-cli)

cloudinit/config/cc_ca_certs.py
```diff
@@ -100,95 +102,114 @@ def update_ca_certs(distro_cfg):
...
-    )
-    update_cert_config(distro_cfg)
+    # Write each certificate to a separate file, maximum 99 certificates.
+    cert_number = 0
+    for c in certs:
+        # First ensure they are strings...
+        cert_file_contents = str(c)
+        cert_number += 1
+        if cert_number == 100:
+            LOG.error(
```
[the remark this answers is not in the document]
holmanb: actually zero padding the cert name isn't necessary, is it? If not I guess I'd prefer dropping the zero padding so we don't add a new requirement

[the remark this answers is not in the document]
dermotbradley: Ok, have removed the padding.

cloudinit/config/cc_ca_certs.py
```diff
@@ -100,95 +102,114 @@ def update_ca_certs(distro_cfg):
...
-            [
-                line
-                for line in orig.splitlines()
-                if line != distro_cfg["ca_cert_filename"]
-            ]
-        )
-        out = "%s\n%s\n" % (cr_cont.rstrip(), distro_cfg["ca_cert_filename"])
+        out = ""
+        for line in orig.splitlines():
+            if line.startswith("#") or line == "":
```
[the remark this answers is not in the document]
dermotbradley: Done.

esposem:
Fedora docs were updated: https://docs.fedoraproject.org/en-US/quick-docs/using-shared-system-certificates/#proc_adding-new-certificates
When adding a package with `trust anchor --store`, you can remove it with `trust anchor --remove`. If you add it manually by copying it in the folder and calling `update-ca-trust`, then you can't remove it with `trust`.

mdeslaur:
ACK from the Ubuntu security team on switching to this approach.

blackboxsw (changes requested):
Thanks for this PR! Looks really good. A couple of nits, minor unittest addition and integration test fixes. Otherwise +1.


1. Let's get the example docs fixed to document the deselect/disable instead of just remove.
2. might be nice to use `enumerate` for the certificate file naming
3. integration test fixes for the new cert file names/links

Combined diff  suggestion, take what you will. throw out the rest :)
```diff
diff --git a/cloudinit/config/cc_ca_certs.py b/cloudinit/config/cc_ca_certs.py
index 2cd49eede..578dd4fbb 100644
--- a/cloudinit/config/cc_ca_certs.py
+++ b/cloudinit/config/cc_ca_certs.py
@@ -20,7 +20,7 @@ LOG = logging.getLogger(__name__)
 DEFAULT_CONFIG = {
     "ca_cert_path": None,
     "ca_cert_local_path": "/usr/local/share/ca-certificates/",
-    "ca_cert_filename": "cloud-init-ca-cert-%(number)s.crt",
+    "ca_cert_filename": "cloud-init-ca-cert-{cert_index}.crt",
     "ca_cert_config": "/etc/ca-certificates.conf",
     "ca_cert_update_cmd": ["update-ca-certificates"],
 }
@@ -28,7 +28,7 @@ DISTRO_OVERRIDES = {
     "rhel": {
         "ca_cert_path": "/etc/pki/ca-trust/",
         "ca_cert_local_path": "/usr/share/pki/ca-trust-source/anchors/",
-        "ca_cert_filename": "cloud-init-ca-cert-%(number)s.crt",
+        "ca_cert_filename": "cloud-init-ca-cert-{cert_index}.crt",
         "ca_cert_config": None,
         "ca_cert_update_cmd": ["update-ca-trust"],
     },
@@ -111,15 +111,12 @@ def add_ca_certs(distro_cfg, certs):
     if not certs:
         return
     # Write each certificate to a separate file.
-    cert_number = 0
-    for c in certs:
+    for cert_index, c in enumerate(certs, 1):
         # First ensure they are strings...
         cert_file_contents = str(c)
-        cert_number += 1
-
-        cert_file_name = distro_cfg["ca_cert_full_path"] % {
-            "number": str(cert_number)
-        }
+        cert_file_name = distro_cfg["ca_cert_full_path"].format(
+            cert_index=cert_index
+        )
         util.write_file(cert_file_name, cert_file_contents, mode=0o644)
 
 
@@ -153,15 +150,27 @@ def disable_system_ca_certs(distro_cfg):
     """
     if distro_cfg["ca_cert_config"] is None:
         return
+    header_comment = (
+        "# Modified by cloud-init to deselect certs due to user-data"
+    )
+    added_header = False
     if os.stat(distro_cfg["ca_cert_config"]).st_size != 0:
         orig = util.load_file(distro_cfg["ca_cert_config"])
-        out = ""
+        out_lines = []
         for line in orig.splitlines():
-            if line.startswith("#") or line.startswith("!") or line == "":
-                out += line + "\n"
+            if line == header_comment:
+                added_header = True
+                out_lines.append(line)
+            elif line == "" or line[0] in ("#", "!"):
+                out_lines.append(line)
             else:
-                out += "!" + line + "\n"
-    util.write_file(distro_cfg["ca_cert_config"], out, omode="wb")
+                if not added_header:
+                    out_lines.append(header_comment)
+                    added_header = True
+                out_lines.append("!" + line)
+    util.write_file(
+        distro_cfg["ca_cert_config"], "\n".join(out_lines) + "\n", omode="wb"
+    )
 
 
 def remove_default_ca_certs(distro_cfg):
@@ -210,17 +219,16 @@ def handle(
     ca_cert_cfg = cfg.get("ca_certs", cfg.get("ca-certs"))
     distro_cfg = _distro_ca_certs_configs(cloud.distro.name)
 
-    # If there is a remove_defaults option set to true, remove the system
+    # If there is a remove_defaults option set to true, disable the system
     # default trusted CA certs first.
     if "remove-defaults" in ca_cert_cfg:
         LOG.warning(
             "DEPRECATION: key 'ca-certs.remove-defaults' is now deprecated."
             " Use 'ca_certs.remove_defaults' instead."
         )
-        if ca_cert_cfg.get("remove-defaults", False):
-            LOG.debug("Disabling/removing default certificates")
-            disable_default_ca_certs(cloud.distro.name, distro_cfg)
-    elif ca_cert_cfg.get("remove_defaults", False):
+    if ca_cert_cfg.get(
+        "remove_defaults", ca_cert_cfg.get("remove-defaults", False)
+    ):
         LOG.debug("Disabling/removing default certificates")
         disable_default_ca_certs(cloud.distro.name, distro_cfg)
 
diff --git a/doc/examples/cloud-config-ca-certs.txt b/doc/examples/cloud-config-ca-certs.txt
index 9f7beb054..3c1c47fb6 100644
--- a/doc/examples/cloud-config-ca-certs.txt
+++ b/doc/examples/cloud-config-ca-certs.txt
@@ -8,11 +8,12 @@
 # It should be passed as user-data when starting the instance.
 
 ca_certs:
-  # If present and set to True, the 'remove_defaults' parameter will remove
-  # all the default trusted CA certificates that are normally shipped with
-  # Ubuntu.
-  # This is mainly for paranoid admins - most users will not need this
-  # functionality.
+  # If present and set to True, the 'remove_defaults' parameter will either
+  # disable all the trusted CA certifications  normally shipped with
+  # Alpine, Debian or Ubuntu. On RedHat, this action will delete those
+  # certificates.
+  # This is mainly for very security-sensitive use cases - most users will not
+  # need this functionality.
   remove_defaults: true
 
   # If present, the 'trusted' parameter should contain a certificate (or list
diff --git a/tests/integration_tests/modules/test_ca_certs.py b/tests/integration_tests/modules/test_ca_certs.py
index 8d18fb765..2baedda9c 100644
--- a/tests/integration_tests/modules/test_ca_certs.py
+++ b/tests/integration_tests/modules/test_ca_certs.py
@@ -76,10 +76,10 @@ class TestCaCerts:
                 unlinked_files.append(filename)
 
         assert ["ca-certificates.crt"] == unlinked_files
-        assert "cloud-init-ca-certs.pem" == links["a535c1f3.0"]
+        assert "cloud-init-ca-cert-1.pem" == links["a535c1f3.0"]
         assert (
-            "/usr/share/ca-certificates/cloud-init-ca-certs.crt"
-            == links["cloud-init-ca-certs.pem"]
+            "/usr/local/share/ca-certificates/cloud-init-ca-cert-1.crt"
+            == links["cloud-init-ca-cert-1.pem"]
         )
 
     def test_cert_installed(self, class_client: IntegrationInstance):
diff --git a/tests/unittests/config/test_cc_ca_certs.py b/tests/unittests/config/test_cc_ca_certs.py
index f17f549cd..3c83bf2a7 100644
--- a/tests/unittests/config/test_cc_ca_certs.py
+++ b/tests/unittests/config/test_cc_ca_certs.py
@@ -249,7 +249,7 @@ class TestAddCaCerts(TestCase):
                 mock_write.assert_has_calls(
                     [
                         mock.call(
-                            conf["ca_cert_full_path"] % {"number": "1"},
+                            conf["ca_cert_full_path"].format(cert_index=1),
                             cert,
                             mode=0o644,
                         )
@@ -277,12 +277,12 @@ class TestAddCaCerts(TestCase):
                 mock_write.assert_has_calls(
                     [
                         mock.call(
-                            conf["ca_cert_full_path"] % {"number": "1"},
+                            conf["ca_cert_full_path"].format(cert_index=1),
                             expected_cert_1_file,
                             mode=0o644,
                         ),
                         mock.call(
-                            conf["ca_cert_full_path"] % {"number": "2"},
+                            conf["ca_cert_full_path"].format(cert_index=2),
                             expected_cert_2_file,
                             mode=0o644,
                         ),
@@ -314,7 +314,10 @@ class TestRemoveDefaultCaCerts(TestCase):
 
     def test_commands(self):
         ca_certs_content = "# line1\nline2\nline3\n"
-        expected = "# line1\n!line2\n!line3\n"
+        expected = (
+            "# line1\n# Modified by cloud-init to deselect certs due to"
+            " user-data\n!line2\n!line3\n"
+        )
 
         for distro_name in cc_ca_certs.distros:
             conf = cc_ca_certs._distro_ca_certs_configs(distro_name)
@@ -344,6 +347,7 @@ class TestRemoveDefaultCaCerts(TestCase):
                             mock.call(conf["ca_cert_local_path"]),
                         ]
                     )
+                    self.assertEqual([], mock_subp.call_args_list)
                 elif distro_name in ["alpine", "debian", "ubuntu"]:
                     mock_load.assert_called_once_with(conf["ca_cert_config"])
                     mock_write.assert_called_once_with(
```

igalic:
after the FreeBSD changes were dropped, I hadn't invested much more time in reviewing this PR, but i do really like @blackboxsw' suggestions here. they're feel a lot easier on the pythonic eye

cloudinit/config/cc_ca_certs.py
```diff
@@ -8,45 +8,47 @@
...
 from cloudinit.settings import PER_INSTANCE
 
+LOG = logging.getLogger(__name__)
+
 DEFAULT_CONFIG = {
-    "ca_cert_path": "/usr/share/ca-certificates/",
-    "ca_cert_filename": "cloud-init-ca-certs.crt",
+    "ca_cert_path": None,
+    "ca_cert_local_path": "/usr/local/share/ca-certificates/",
+    "ca_cert_filename": "cloud-init-ca-cert-%(number)s.crt",
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/config/cc_ca_certs.py
```diff
@@ -100,95 +102,108 @@ def update_ca_certs(distro_cfg):
...
-            [
-                line
-                for line in orig.splitlines()
-                if line != distro_cfg["ca_cert_filename"]
-            ]
-        )
-        out = "%s\n%s\n" % (cr_cont.rstrip(), distro_cfg["ca_cert_filename"])
+        out = ""
+        for line in orig.splitlines():
+            if line.startswith("#") or line.startswith("!") or line == "":
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/config/cc_ca_certs.py
```diff
@@ -100,95 +102,108 @@ def update_ca_certs(distro_cfg):
...
         return
-    # First ensure they are strings...
-    cert_file_contents = "\n".join([str(c) for c in certs])
-    util.write_file(
-        distro_cfg["ca_cert_full_path"], cert_file_contents, mode=0o644
-    )
-    update_cert_config(distro_cfg)
+    # Write each certificate to a separate file.
+    cert_number = 0
+    for c in certs:
```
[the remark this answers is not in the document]
dermotbradley: Done

tests/unittests/config/test_cc_ca_certs.py
```diff
@@ -378,40 +313,49 @@ def setUp(self):
...
-                        ("debconf-set-selections", "-"),
-                        "ca-certificates ca-certificates/trust_new_crts"
-                        " select no",
-                    )
+                    if distro_name in ["debian", "ubuntu"]:
+                        mock_subp.assert_called_once_with(
+                            ("debconf-set-selections", "-"),
+                            "ca-certificates ca-certificates/trust_new_crts"
+                            " select no",
+                        )
```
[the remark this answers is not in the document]
dermotbradley: Done

dermotbradley:
@blackboxsw I also noticed a subtle change had been introduced (compared to original behaviour) after rolling back the RHEL changes - I've corrected the values of ``ca_cert_local_path`` and ``ca_cert_filename`` for RHEL so that the deletion occurs the level above the "anchors" directory.

holmanb (approved):
Thanks for this @dermotbradley. LGTM.

blackboxsw (approved):
Thank you for the updates and effort here minimal. +1 let's merge this and get it into 23.1

PR #1963, opened 2023-01-16
build: support newer releases

CalvoM:
## Proposed Commit Message

```
build: support newer releases
make the package build possible on newer releases after focal.
```

## Test Steps

Start a jammy lxd container and run the make command.
```
make ci-deps-ubuntu
```

## Checklist:

### DIFF
--- a/packages/bddeb
+++ b/packages/bddeb
@@ -100,8 +100,8 @@ def write_debian_folder(root, templ_data, cloud_util_deps):
     requires.extend(['python3'] + reqs + test_reqs)
     if templ_data['debian_release'] == 'xenial':
         requires.append('python3-pytest-catchlog')
-    elif templ_data['debian_release'] == 'impish':
-        requires.remove('dh-systemd')
+    elif templ_data['debian_release'] in ('xenial', 'bionic', 'focal'):
+        requires.append('dh-systemd')
     templater.render_to_file(util.abs_join(find_root(),
                                            'packages', 'debian', 'control.in'),
                              util.abs_join(deb_dir, 'control'),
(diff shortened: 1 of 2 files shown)


packages/bddeb
```diff
@@ -100,8 +100,8 @@ def write_debian_folder(root, templ_data, cloud_util_deps):
     requires.extend(['python3'] + reqs + test_reqs)
     if templ_data['debian_release'] == 'xenial':
         requires.append('python3-pytest-catchlog')
-    elif templ_data['debian_release'] == 'impish':
-        requires.remove('dh-systemd')
+    elif templ_data['debian_release'] in ('xenial', 'bionic', 'focal'):
```
blackboxsw, line 103: Let's include "stretch" & "buster" so this still works on Debian-based images since you have removed dh-systemd from the common "debian" key in pkg-deps.json. I tested w/ LXC and bullseye and later don't have dh-systemd.

```suggestion
    elif templ_data['debian_release'] in ('stretch', 'buster', 'xenial', 'bionic', 'focal'):
```

blackboxsw (approved):
Minor fix for debian then +1

packages/bddeb
blackboxsw: No need for stretch as it's python3.5 and cloud-init tip of main [doesn't support python3.5 anymore anyway](https://lists.launchpad.net/cloud-init/msg00377.html)

blackboxsw:
Also ran this through `./tools/run-container debian/10 --unittest --package` and `./tools/run-container debian/11 --unittest --package` asserting that the deb package is buildable on buster and bullseye debian.

PR #1965, opened 2023-01-16
add proxy support for ansible galaxy

holmanb:
```
ansible: add proxy support and test
```

## Additional Context
Some environments, like our own Jenkins environment, require proxy support to reach Ansible Galaxy. Implement it, and add integration test coverage.

## Test

Local Machine:
```
tox -e integration-tests -- tests/integration_tests/modules/test_ansible.py
```

Jenkins:
```bash
mkdir /jenkins/tmp
cd /jenkins/tmp
git clone https://github.com/holmanb/cloud-init.git
cd cloud-init
git checkout holmanb/ansible-proxy
sed -i 's/NONE/IN_PLACE/' tests/integration_tests/integration_settings.py
tox -e integration-tests -- tests/integration_tests/modules/test_ansible.py
```

### DIFF
--- a/tests/integration_tests/modules/test_ansible.py
+++ b/tests/integration_tests/modules/test_ansible.py
@@ -13,7 +13,7 @@ from tests.integration_tests.util import verify_clean_log
 
 proxy_name = "squid.internal:3128"
 try:
-    socket.gethostbyname(proxy_name.split(':')[0])
+    socket.gethostbyname(proxy_name.split(":")[0])
     if os.getcwd().startswith("/jenkins"):
         proxy = f"""
     proxy:
@@ -101,12 +101,15 @@ runcmd:
   - [systemctl, enable, repo_waiter.service]
 """
 
-INSTALL_METHOD = """
+INSTALL_METHOD = (
+    """
 ansible:
   ansible_config: /etc/ansible/ansible.cfg
   install_method: {method}
   package_name: {package}
-  galaxy:""" + proxy + """
+  galaxy:"""
+    + proxy
+    + """
     actions:
      - ["ansible-galaxy", "collection", "install", "community.grafana"]
   pull:
@@ -114,6 +117,7 @@ ansible:
     playbook_name: ubuntu.yml
     full: true
 """
+)
 
 SETUP_REPO = f"cd {REPO_D}                                    &&\
 git config --global user.name auto                            &&\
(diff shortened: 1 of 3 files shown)


tests/integration_tests/modules/test_ansible.py
```diff
@@ -8,6 +10,20 @@
 # This instrumentation allows the test to run self-contained
 # without network access or external git repos.
 
+proxy_name = 'squid.internal:3128'
+try:
+    socket.gethostbyname(proxy_name)
+    if os.environ["PATH"].startswith("/jenkins"):
```
holmanb, line 16: Open to ideas if someone has a better suggestion than checking if squid.internal is resolvable and checking the `/jenkins` path. This seems to work locally, but there may be better ways.

aciba90: What about checking if some Jenkins' environment variable is set, e.g., `JENKINS_HOME` or `JOB_NAME`?

TheRealFalcon:
After a brief google, it looks like `ansible-galaxy` supports the `http(s)_proxy` env variables. It'd be nice if we rely on those instead of hardcoding proxy into the config. I'd rather not have any module that can reach out to the internet need to grow a proxy key. We already added a "short term" hack for it in cc_ssh_import_id:
https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_ssh_import_id.py#L131

That sort of example is probably a bit too hacky for an integration test, so I think what's proposed here is ok for now. Long term, I think it's worth growing some kind of env module that'll set subp's default env.

TheRealFalcon:
Eh...but once we add it we can't really remove it. I would be good to find a nicer solution now.

holmanb:
> After a brief google, it looks like ansible-galaxy supports the http(s)_proxy env variables.

That's actually how this PR works. `proxy`'s subkeys would be `http_proxy` and `https_proxy`. I should have clarified that in the schema description for this key.

> Long term, I think it's worth growing some kind of env module that'll set subp's default env.

Since this test failure isn't blocking anything I could make this PR into a `cc_env` or something like that, if that's the approach we'd prefer.

tests/integration_tests/modules/test_ansible.py
aciba90: It looks great, thank you.

igalic:
> Since this test failure isn't blocking anything I could make this PR into a `cc_env` or something like that, if that's the approach we'd prefer.

🤔 a global state module

holmanb:
> a global state module

For something like http_proxy/https_proxy, the ability to set this globally somehow makes sense in most cases. I'm trying to think of an example where this would be undesirable - where any environment variable should be set or have a value in one module, where it would be expected to be unset or have a different value in another. I think that the most likely (runcmd/bootcmd/etc) candidates have enough flexibility to modify their own run environments, so those aren't worth worrying too much about.

TheRealFalcon:
> Since this test failure isn't blocking anything I could make this PR into a cc_env or something like that, if that's the approach we'd prefer.

I think it's worth a shot.

PR #1968, opened 2023-01-17
sources/azure: retry on connection error when fetching metdata

cjp256:
Early attempts to fetch metadata on Azure may fail with connection errors.  While this class of errors is not ideal to retry on, the impact is minimal given that:

1. retries are fairly limited (10)

2. Persistent connection errors would indicate that cloud-init is using a non-primary NIC which is a rare case of failure that will be addressed in the future.


anhvoms (approved):

holmanb (approved):
LGTM!

PR #1969, opened 2023-01-18
CI: migrate integration-test to GH actions

aciba90:
## Proposed Commit Message

```
ci: migrate integration-test to gh actions

changes:
- drop caching machinery for LXD images and chroots
- divide the workflow in two steps, one for building packages
  and other for running the tests so that the second one can
  be retried in case of error
```

## Additional Context

I have not ported the caching machinery for LXD images and chroots, as in my opinion it is not worth the complexity vs the gains, but I am open to other opinions.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: "Checkout"
-        uses: actions/checkout@v3.0.0
+        uses: actions/checkout@v3
         with:
           # Fetch all tags for tools/read-version
           fetch-depth: 0
@@ -43,7 +43,7 @@ jobs:
           sudo -E su $USER -c 'mk-sbuild ${{ env.RELEASE }}'
           sudo -E su $USER -c 'DEB_BUILD_OPTIONS=nocheck sbuild --nolog --no-run-lintian --no-run-autopkgtest --verbose --dist=${{ env.RELEASE }} --build-dir=${{ runner.temp }} cloud-init_*.dsc'
       - name: Archive debs as artifacts
-        uses: actions/upload-artifact@v3.1.2
+        uses: actions/upload-artifact@v3
         with:
           name: 'cloud-init-${{ env.RELEASE }}-deb'
           path: '${{ runner.temp }}/cloud-init*.deb'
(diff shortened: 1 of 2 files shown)


paride:
Nice, I agree with not implementing caching for now, that's something that can be added later. I'll have a better look later today. @aciba90 did you manage to test this on a fork or something?

aciba90:
> @aciba90 did you manage to test this on a fork or something?

Thanks, @paride. The new workflow is being executed as part of the checks of this PR. See the `Checks` tab or https://github.com/canonical/cloud-init/actions/runs/3994716010/jobs/6852827704.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,86 @@
...
+jobs:
+  package-build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: "Checkout #1"
+        uses: actions/checkout@v3.0.0
+      - name: "Checkout #2 (for tools/read-version)"
+        run: |
+          git fetch --unshallow
+          git remote add upstream https://git.launchpad.net/cloud-init
```
paride, line 28: Do we actually need to add LP as a remote, especially given that it is added but never fetched?

paride (approved):
Looks great! Just one minor inline question.

.github/workflows/integration.yml
aciba90: That was probably a relic / copy-paste, I have removed it and unified Checkout #1 and #2 in a single one.

paride: Nice and TIL about `fetch-depth`, thanks!

.github/workflows/integration.yml
```diff
@@ -0,0 +1,84 @@
...
+
+env:
+  RELEASE: bionic
+
+jobs:
+  package-build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v3.0.0
```
paride, line 24: Here (and below, also for other actions), is there a reason for not going with `@v3` instead of `v3.x.y`, relying on the fact that compatibility shouldn't be broken in a given major version?

The official docs (e.g. https://github.com/actions/checkout) only specify the major.

paride (commented):
One more inline comment/question.

.github/workflows/integration.yml
aciba90: The only reason was to keep it consistent with the rest of workflows, but taking into account that they are official GitHub actions, I agree it is better to pin only major versions. Updated, thanks for pointing out that.

I will align the other workflows to this approach in a follow-up PR.

paride (approved):
Looks good, thanks!

TheRealFalcon (approved):
Looks good! Just curious, is this split into 2 separate jobs with the upload and download so that we can download the built deb when investigating failures? Or is there different reason for the split?

aciba90:
> Looks good! Just curious, is this split into 2 separate jobs with the upload and download so that we can download the built deb when investigating failures? Or is there different reason for the split?

Thanks, @TheRealFalcon. That's one of the reason plus:

* If the `integration-test` step fails, we can retry it without executing the `package-build` step.
* Separation of concerns. It is more clear what dependencies are required for each step.

PR #1970, opened 2023-01-19
cc_puppet: Update puppet service name

CalvoM:
```
cc_puppet: Update puppet service name

From Lunar, we see that the default puppet version is 7.20 which replaces `puppet.service` 
with `puppet-agent.service`. Thus, we need to have a way of calling the appropriate service 
depending with the ubuntu release.


LP: #2002969 
```

## Test Steps

Run the tests at `tests/integration_tests/modules/test_puppet.py` on `Lunar`

## Checklist:

### DIFF
--- a/cloudinit/config/cc_puppet.py
+++ b/cloudinit/config/cc_puppet.py
@@ -25,6 +25,7 @@ from cloudinit.settings import PER_INSTANCE
 
 AIO_INSTALL_URL = "https://raw.githubusercontent.com/puppetlabs/install-puppet/main/install.sh"  # noqa: E501
 PUPPET_AGENT_DEFAULT_ARGS = ["--test"]
+PREFERRED_PUPPET_PACKAGE_NAMES = ("puppet-agent", "puppet")
 
 MODULE_DESCRIPTION = """\
 This module handles puppet installation and configuration. If the ``puppet``
@@ -118,26 +119,27 @@ class PuppetConstants:
         self.csr_attributes_path = csr_attributes_path
 
 
-def _autostart_puppet(log):
+def _autostart_puppet(log, cloud: Cloud, service_name: str):
     # Set puppet to automatically start
-    if os.path.exists("/etc/default/puppet"):
+    if os.path.exists(f"/etc/default/{service_name}"):
         subp.subp(
             [
                 "sed",
                 "-i",
                 "-e",
                 "s/^START=.*/START=yes/",
-                "/etc/default/puppet",
+                f"/etc/default/{service_name}",
             ],
             capture=False,
         )
     elif subp.which("systemctl"):
-        subp.subp(["systemctl", "enable", "puppet.service"], capture=False)
+        cloud.distro.manage_service("enable", f"{service_name}.service")
     elif os.path.exists("/sbin/chkconfig"):
-        subp.subp(["/sbin/chkconfig", "puppet", "on"], capture=False)
+        subp.subp(["/sbin/chkconfig", service_name, "on"], capture=False)
     else:
         log.warning(
-            "Sorry we do not know how to enable puppet services on this system"
+            f"Sorry we do not know how to enable {service_name} services on \
+              this system"
         )
 
 
@@ -221,11 +223,12 @@ def handle(
     else:  # default to 'packages'
         puppet_user = "puppet"
         puppet_bin = "puppet"
-        puppet_package = "puppet"
+        puppet_package = None  # changes with distro
 
     package_name = util.get_cfg_option_str(
         puppet_cfg, "package_name", puppet_package
     )
+    service_names = ["puppet-agent"]
     if not install and version:
         log.warning(
             "Puppet install set to false but version supplied, doing nothing."
@@ -238,7 +241,24 @@ def handle(
         )
 
         if install_type == "packages":
-            cloud.distro.install_packages((package_name, version))
+            if package_name is None:  # conf has no package_name
+                try:
+                    log.debug("Trying to install puppet-agent")
+                    cloud.distro.install_packages(
+                        (PREFERRED_PUPPET_PACKAGE_NAMES[0], version)
+                    )
+                    service_names = ["puppet-agent"]
+                except subp.ProcessExecutionError:
+                    log.debug("Trying to install puppet")
+                    cloud.distro.install_packages(
+                        (PREFERRED_PUPPET_PACKAGE_NAMES[1], version)
+                    )
+                    service_names = ["puppet"]
+            else:
+                cloud.distro.install_packages((package_name, version))
+                # install both services to find the correct one
+                service_names = ["puppet-agent", "puppet"]
+
         elif install_type == "aio":
             install_puppet_aio(
                 cloud.distro, aio_install_url, version, collection, cleanup
@@ -318,7 +338,12 @@ def handle(
 
     # Set it up so it autostarts
     if start_puppetd:
-        _autostart_puppet(log)
+        for service_name in service_names:
+            try:
+                _autostart_puppet(log, cloud, service_name)
+                break
+            except subp.ProcessExecutionError:
+                log.debug(f"Could not auto enable {service_name}")
 
     # Run the agent if needed
     if run:
@@ -344,7 +369,12 @@ def handle(
 
     if start_puppetd:
         # Start puppetd
-        subp.subp(["service", "puppet", "start"], capture=False)
+        for service_name in service_names:
+            try:
+                cloud.distro.manage_service("start", service_name)
+                break
+            except subp.ProcessExecutionError:
+                log.debug(f"Could not auto start {service_name}")
 
 
 # vi: ts=4 expandtab
--- a/cloudinit/config/cc_puppet.py
+++ b/cloudinit/config/cc_puppet.py
@@ -25,7 +25,7 @@ from cloudinit.settings import PER_INSTANCE
 
 AIO_INSTALL_URL = "https://raw.githubusercontent.com/puppetlabs/install-puppet/main/install.sh"  # noqa: E501
 PUPPET_AGENT_DEFAULT_ARGS = ["--test"]
-PREFERRED_PUPPET_PACKAGE_NAMES = ("puppet-agent", "puppet")
+PUPPET_PACKAGE_NAMES = ("puppet-agent", "puppet")
 
 MODULE_DESCRIPTION = """\
 This module handles puppet installation and configuration. If the ``puppet``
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_puppet.py
```diff
@@ -132,7 +140,8 @@ def _autostart_puppet(log):
             capture=False,
         )
     elif subp.which("systemctl"):
-        subp.subp(["systemctl", "enable", "puppet.service"], capture=False)
+        service = f"{get_service_name()}.service"
+        subp.subp(["systemctl", "enable", service], capture=False)
```
igalic, line 144: why don't we call `distro.service` here, that's what it's for

cloudinit/config/cc_puppet.py
```diff
@@ -344,7 +353,8 @@ def handle(
 
     if start_puppetd:
         # Start puppetd
-        subp.subp(["service", "puppet", "start"], capture=False)
+        service = get_service_name()
+        subp.subp(["service", service, "start"], capture=False)
```
igalic, line 357: and here, and everywhere else.

igalic (changes requested):
We can do one better

cloudinit/config/cc_puppet.py
CalvoM: Makes sense. Updated.

cloudinit/config/cc_puppet.py
CalvoM: I see this will require us to pass a cloud instance to the function. Can we have a separate PR for such and other refactors?

igalic: fair.

cloudinit/config/cc_puppet.py
```diff
@@ -118,6 +118,14 @@ def __init__(
         self.csr_attributes_path = csr_attributes_path
 
 
+def get_service_name():
+    release = util.lsb_release()["codename"]
+    service = "puppet-agent"
+    if release in ("bionic", "focal", "jammy", "kinetic"):
+        service = "puppet"
+    return service
```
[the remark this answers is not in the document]
igalic: other modules have a hash like this: https://github.com/canonical/cloud-init/blob/009377174f9a5224218f24c37036ff226dffa08d/cloudinit/config/cc_ntp.py#L50-L91

followed by distro specific overrides:

https://github.com/canonical/cloud-init/blob/009377174f9a5224218f24c37036ff226dffa08d/cloudinit/config/cc_ntp.py#L94-L216

this general approach might make some sense here: we could default to the new method, and allow a config selection to choose the old method

this would also allow us to encode different config paths, like FreeBSD for example uses.

cloudinit/config/cc_puppet.py
```diff
@@ -118,26 +119,27 @@ def __init__(
         self.csr_attributes_path = csr_attributes_path
 
 
-def _autostart_puppet(log):
+def _autostart_puppet(log, cloud: Cloud, service_name: str):
     # Set puppet to automatically start
-    if os.path.exists("/etc/default/puppet"):
+    if os.path.exists(f"/etc/default/{service_name}"):
```
igalic, line 124: shouldn't this be using `distro.manage_service('enable', service_name)`, or is there a reason you're avoiding that?

igalic (commented):
❓

cloudinit/config/cc_puppet.py
CalvoM: @igalic, no specific reason. I am still improving the PR. Thank you for the suggestion.

cloudinit/config/cc_puppet.py
```diff
@@ -118,26 +119,17 @@ def __init__(
...
-            capture=False,
-        )
-    elif subp.which("systemctl"):
-        subp.subp(["systemctl", "enable", "puppet.service"], capture=False)
-    elif os.path.exists("/sbin/chkconfig"):
-        subp.subp(["/sbin/chkconfig", "puppet", "on"], capture=False)
-    else:
+    try:
+        if os.path.exists("/sbin/chkconfig"):
+            subp.subp(["/sbin/chkconfig", service_name, "on"], capture=False)
```
igalic, line 126: how many systems out there still have `chkconfig`?

i think i would leave this out, if there's need support those systems, i would push that into `distro.manage_service`

igalic (commented):
another suggestion…

cloudinit/config/cc_puppet.py
CalvoM: I had the same concern. I noticed that we could push it into chkconfig for mostly rhel systems. but if we have reason for it.

blackboxsw: I believe `chkconfig` was deprecated in favor of systemd's `systemctl`(systemd-based) or `service`(opensuse), `rcctl`(openbsd) or `rc-service`(alpine|gentoo) commands to manage services The original commit was > 10 years ago when that support was necessary to have a fallback in old sysvinit distros. I don't think we need this fallback anymore in tip of main and I don't want to pull that into cloudinit.distros if we can avoid it.  Let's actually drop this try/except handling on chkconfig as I don't think there are any images where want to support it anymore.   

Side note: When looking at 'old way' of doing things vs new way, we generally want to organize our code to prefer the new way and only fallback to the old way when new way doesn't work. Otherwise, we waste CPU cycles at boot for 99% or recent distributions unnecessarily checking for a file or behavior that no longer should exist.

cloudinit/config/cc_puppet.py
```diff
@@ -344,7 +359,12 @@ def handle(
 
     if start_puppetd:
         # Start puppetd
-        subp.subp(["service", "puppet", "start"], capture=False)
+        for service_name in service_names:
+            try:
+                cloud.distro.manage_service("start", service_name)
+                break
+            except subp.ProcessExecutionError:
+                log.debug(f"Could not auto start {service_name}")
```
blackboxsw, line 367: Given that both the "enable" and "start" logic should probably be the same. It may be useful to generalize a function like _manage_one_of_puppet_services  to deal with both enable and start. Since we likely want to fallback to puppet if puppet-agent service isn't available, let's handle all fallback options in one place and only log.warning if we can't find a viable option


Here's a diff for you to consider if you think it is applicable/functional for this case. It let's us consolidate the _autostart_puppet for the "enable" case and the semi-duplication of the if start_puppetd checks below during "start"


```diff
diff --git a/cloudinit/config/cc_puppet.py b/cloudinit/config/cc_puppet.py
index d19eb43db..7cbe2c2e3 100644
--- a/cloudinit/config/cc_puppet.py
+++ b/cloudinit/config/cc_puppet.py
@@ -25,7 +25,7 @@ from cloudinit.settings import PER_INSTANCE
 
 AIO_INSTALL_URL = "https://raw.githubusercontent.com/puppetlabs/install-puppet/main/install.sh"  # noqa: E501
 PUPPET_AGENT_DEFAULT_ARGS = ["--test"]
-PREFERRED_PUPPET_PACKAGE_NAMES = ("puppet-agent", "puppet")
+PREFERRED_PUPPET_NAMES = ("puppet-agent", "puppet")
 
 MODULE_DESCRIPTION = """\
 This module handles puppet installation and configuration. If the ``puppet``
@@ -119,20 +119,6 @@ class PuppetConstants:
         self.csr_attributes_path = csr_attributes_path
 
 
-def _autostart_puppet(log, cloud: Cloud, service_name: str):
-    # Set puppet to automatically start
-    try:
-        if os.path.exists("/sbin/chkconfig"):
-            subp.subp(["/sbin/chkconfig", service_name, "on"], capture=False)
-        else:
-            cloud.distro.manage_service("enable", f"{service_name}.service")
-    except subp.ProcessExecutionError:
-        log.warning(
-            f"Sorry we do not know how to enable {service_name} services on \
-              this system"
-        )
-
-
 def get_config_value(puppet_bin, setting):
     """Get the config value for a given setting using `puppet config print`
     :param puppet_bin: path to puppet binary
@@ -178,6 +164,24 @@ def install_puppet_aio(
         return subp.subp([tmpf] + args, capture=False)
 
 
+def _manage_one_of_puppet_services(action: str):
+    """Attempt to perform action on one of puppet services."""
+    service_managed: str = ""
+    for puppet_name in PREFERRED_PUPPET_NAMES:
+        try:
+            cloud.distro.manage_service("enable", f"{puppet_name}.service")
+            service_managed = puppet_name
+            break
+        except subp.ProcessExecutionError:
+            pass
+    if not service_managed:
+        log.warning(
+            "Could not '%s' any of the following services: %s",
+            action,
+            ", ".join(PUPPET_PREFERRED_NAMES)
+        )
+
+
 def handle(
     name: str, cfg: Config, cloud: Cloud, log: Logger, args: list
 ) -> None:
@@ -218,7 +222,6 @@ def handle(
     package_name = util.get_cfg_option_str(
         puppet_cfg, "package_name", puppet_package
     )
-    service_names = ["puppet-agent"]
     if not install and version:
         log.warning(
             "Puppet install set to false but version supplied, doing nothing."
@@ -232,23 +235,17 @@ def handle(
 
         if install_type == "packages":
             if package_name is None:  # conf has no package_name
-                try:
-                    log.debug("Trying to install puppet-agent")
-                    cloud.distro.install_packages(
-                        (PREFERRED_PUPPET_PACKAGE_NAMES[0], version)
+                for puppet_name in PREFERRED_PUPPET_NAMES:
+                    try:
+                        cloud.distro.install_packages(puppet_name, version)
+                        package_name = puppet_name
+                    except subp.ProcessExecutionError:
+                        pass
+                if not package_name:
+                    log.warning(
+                        "No installable puppet package in any of: %s",
+                        ", ".join(PREFERRED_PUPPET_NAMES)
                     )
-                    service_names = ["puppet-agent"]
-                except subp.ProcessExecutionError:
-                    log.debug("Trying to install puppet")
-                    cloud.distro.install_packages(
-                        (PREFERRED_PUPPET_PACKAGE_NAMES[1], version)
-                    )
-                    service_names = ["puppet"]
-            else:
-                cloud.distro.install_packages((package_name, version))
-                # install both services to find the correct one
-                service_names = ["puppet-agent", "puppet"]
-
         elif install_type == "aio":
             install_puppet_aio(
                 cloud.distro, aio_install_url, version, collection, cleanup
@@ -326,14 +323,9 @@ def handle(
             yaml.dump(puppet_cfg["csr_attributes"], default_flow_style=False),
         )
 
-    # Set it up so it autostarts
     if start_puppetd:
-        for service_name in service_names:
-            try:
-                _autostart_puppet(log, cloud, service_name)
-                break
-            except subp.ProcessExecutionError:
-                log.debug(f"Could not auto enable {service_name}")
+        # Enable puppet auto-start on boot
+        _manage_one_of_puppet_services("enable")
 
     # Run the agent if needed
     if run:
@@ -358,13 +350,8 @@ def handle(
         subp.subp(cmd, capture=False)
 
     if start_puppetd:
-        # Start puppetd
-        for service_name in service_names:
-            try:
-                cloud.distro.manage_service("start", service_name)
-                break
-            except subp.ProcessExecutionError:
-                log.debug(f"Could not auto start {service_name}")
+        # Start puppetd now
+        _manage_one_of_puppet_services("start")
 
 
 # vi: ts=4 expandtab
```

CalvoM: This is a goodt suggestion and improvement.

cloudinit/config/cc_puppet.py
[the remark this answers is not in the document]
blackboxsw: @igalic good suggestion on teasing out distribution config path defaults generally speaking. But, the case in the puppet module looks like it actually shells out to discover the configured defaults on disk, so the separate distro config file paths are unnecessary as the installed puppet (wherever you are) will tell us on the cmdline via a call to: [puppet config print <setting>](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_puppet.py#L149)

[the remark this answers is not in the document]
blackboxsw: Also, given that the binary in question is either puppet-agent (forward-looking) vs puppet for this module, I don't think we want to hardcode expected puppet* binary names per distro as pip installed vs packaged puppet binaries may change/evolve over time. So baking in the try <new_puppet>/except <old_puppet> seems to cover the primary/secondary use-cases without having to keep cloud-init implementation in step with each distro's update timeline for puppet package updates.

blackboxsw (approved):
Thanks Calvin for rework on this. It looks good to me. When Cloud-init SRUs, we may want to add a patch to the debian directory to only prefer puppet above puppet-agent since default packaged version there is likely going to stay at 'puppet'. But beyond that this works fine for tip of main.

PR #1971, opened 2023-01-19
Fix permission of SSH host keys

Mazorius:
private keys should have 0600: https://www.tenable.com/audits/items/CIS_Google_Container_Optimized_OS_v1.0.0_L1_Server.audit:6225b8224fbd4f360ebdc72c56f3eae9
public keys should have 0644: https://www.tenable.com/audits/items/CIS_Google_Container_Optimized_OS_v1.0.0_L1_Server.audit:7f016cd406100a1ee2ad94834111f005

## Proposed Commit Message

```
fix: permissions for SSH host keys

If the host-keys are automatically generated the private key has 0640 which should be 0600 so no group or other can see the private key. The public key is correctly set to 0644.

If the host-keys are provided the script generated the private key with 0600 which is indeed correct. But the public key is also 0600 which should be instead 0644.

With this change the private key is always 0600 and the public key is always 0644.
```

## Additional Context

During the execution of Terraform the keys permission were not set correctly.
The application could not generate the public fingerprint as it could not read the public host key.

After changing the private key to 0600 and the public to 0644 everything works as expected.

## Test Steps

Install a tool like GitLab via Omnibus that require to generate the public host key fingerprint.
Open this page after setup: https://gitlab.example.com/help/configuration .

Which the correct permission the page should be open. With the wrong permission the page should throw a 500 HTTP Code.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -188,7 +188,7 @@ for k in GENERATE_KEY_NAMES:
         {
             f"{k}_private": (KEY_FILE_TPL % k, 0o600),
             f"{k}_public": (f"{KEY_FILE_TPL % k}.pub", 0o644),
-            f"{k}_certificate": (f"{KEY_FILE_TPL % k}-cert.pub", 0o600),
+            f"{k}_certificate": (f"{KEY_FILE_TPL % k}-cert.pub", 0o644),
         }
     )
     PRIV_TO_PUB[f"{k}_private"] = f"{k}_public"
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -187,7 +187,7 @@ for k in GENERATE_KEY_NAMES:
     CONFIG_KEY_TO_FILE.update(
         {
             f"{k}_private": (KEY_FILE_TPL % k, 0o600),
-            f"{k}_public": (f"{KEY_FILE_TPL % k}.pub", 0o600),
+            f"{k}_public": (f"{KEY_FILE_TPL % k}.pub", 0o644),
             f"{k}_certificate": (f"{KEY_FILE_TPL % k}-cert.pub", 0o600),
         }
     )
@@ -278,7 +278,7 @@ def handle(
                     if gid != -1:
                         # perform same "sanitize permissions" as sshd-keygen
                         os.chown(keyfile, -1, gid)
-                        os.chmod(keyfile, 0o640)
+                        os.chmod(keyfile, 0o600)
                         os.chmod(keyfile + ".pub", 0o644)
                 except subp.ProcessExecutionError as e:
                     err = util.decode_binary(e.stderr).lower()
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -278,7 +278,7 @@ def handle(
                     if gid != -1:
                         # perform same "sanitize permissions" as sshd-keygen
                         os.chown(keyfile, -1, gid)
-                        os.chmod(keyfile, 0o600)
+                        os.chmod(keyfile, 0o640)
                         os.chmod(keyfile + ".pub", 0o644)
                 except subp.ProcessExecutionError as e:
                     err = util.decode_binary(e.stderr).lower()
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -330,17 +330,17 @@ class TestHandleSsh:
                     mock.call(
                         "/etc/ssh/ssh_host_{}_key".format(key_type),
                         private_value,
-                        384,
+                        0o600,
                     ),
                     mock.call(
                         "/etc/ssh/ssh_host_{}_key.pub".format(key_type),
                         public_value,
-                        384,
+                        0o644,
                     ),
                     mock.call(
                         "/etc/ssh/ssh_host_{}_key-cert.pub".format(key_type),
                         cert_value,
-                        384,
+                        0o600,
                     ),
                     mock.call(
                         sshd_conf_fname,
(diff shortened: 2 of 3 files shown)


Mazorius:
I not really get the issue in the test. Anyone an idea what the issue is? Cannot see something checks the permission.

cloudinit/config/cc_ssh.py
```diff
@@ -278,7 +278,7 @@ def handle(
                     if gid != -1:
                         # perform same "sanitize permissions" as sshd-keygen
                         os.chown(keyfile, -1, gid)
-                        os.chmod(keyfile, 0o640)
+                        os.chmod(keyfile, 0o600)
```
igalic, line 281: I think this is the private key

cloudinit/config/cc_ssh.py
```diff
@@ -278,7 +278,7 @@ def handle(
...
                         os.chmod(keyfile + ".pub", 0o644)
```
igalic, line 282: whereas this is the public key

igalic (commented):
thank you for this PR!

you'll need to sign the CLA, as per the Contribution guidelines

Mazorius:
@microsoft-github-policy-service agree

Mazorius:
> thank you for this PR!
> 
> you'll need to sign the CLA, as per the Contribution guidelines

Ok. Need to do that tomorrow. The contribution link inside the readme.md is a deadlink.

Mazorius:
@igalic I cannot as I have not contact:

image

cloudinit/config/cc_ssh.py
Mazorius: yes that I in fact correct.

cloudinit/config/cc_ssh.py
Mazorius: of course but you also have a configuration in the beginning of the file: https://github.com/canonical/cloud-init/blob/9aa2e8a408ce61e2d2ccc1dbd5f9c9b97848d2c9/cloudinit/config/cc_ssh.py#L190

Mazorius: And the lines 28x are only valid if ssh-keys attribute is empty. As soon as you provide a key with that attribute the permission got be set as defined in line 189,190 and 191 ;)

igalic:
from https://cloudinit.readthedocs.io/en/latest/index.html

> When signing it:
>
> -  ensure that you fill in the GitHub username field,
> - when prompted for ‘Project contact’ or ‘Canonical Project Manager’, enter ‘James Falcon’.

Mazorius:
Thanks @igalic 
image

cloudinit/config/cc_ssh.py
Mazorius: I found the issue. In the test it was decimal used instead of octal.

I changed it to octal and adapt the value as it should be. Now the pipeline should successfully run.

igalic:
you also need to add your name to this file tools/.lp-to-git-user

Mazorius:
I checked a bit in the network and find that the cert.pub should also has 644: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-creating_ssh_ca_certificate_signing-keys#doc-wrapper

Therefore I changed this also to 644.

cloudinit/config/cc_ssh.py
aciba90, line 281: This was introduced in https://github.com/canonical/cloud-init/commit/ee296ced9c0a61b1484d850b807c601bcd670ec1, and the read group permissions are specific to the `ssh_keys` group in system with that group set (RHEL-like OSes).

I would assume that that fact is a requirement needed for `openscap` to function properly. Maybe @esposem can clarify that.

In any case, I think it is safe to keep this line as it was (`0o640`) as only users belonging to the `ssh_keys` group would be able to read private keys, which makes sense.

aciba90 (commented):
Thanks, @Mazorius, for the improvements.

I am only concerned about the inline comment related to the change in group permissions on systems with `ssh_keys` group. Otherwise, looks good to me.

- [x] CLA signed

cloudinit/config/cc_ssh.py
esposem: The issue is that we are replacing sshd-keygen. So cloud-init should do the same exact thing that sshd-keygen does.
And when reading `/usr/libexec/openssh/sshd-keygen` this is what we find (also posted in my PR):

```
# create new keys
if ! $KEYGEN -q -t $KEYTYPE -f $KEY -C '' -N '' >&/dev/null; then
	exit 1
fi

# sanitize permissions
/usr/bin/chgrp ssh_keys $KEY
/usr/bin/chmod 640 $KEY
/usr/bin/chmod 644 $KEY.pub
if [[ -x /usr/sbin/restorecon ]]; then
	/usr/sbin/restorecon $KEY{,.pub}
fi
```
That's why I wrote `640` and not `600`.

Mazorius: good to know.

that is also fine for me. Will change this back to 0640.

Should we then also change this to 640?

https://github.com/canonical/cloud-init/blob/e49852d44aba7e32e763244af2cb152b214b2306/cloudinit/config/cc_ssh.py#L189

aciba90: Thanks @Mazorius. I think line 189 does not interfere with the `ssh_keys` case, and it is right as-is.

aciba90 (changes requested):
Marked as request changes until the `ssh_keys` block is reverted.

cloudinit/config/cc_ssh.py
esposem: Agree

Mazorius: changed back to 640

aciba90 (approved):
LGTM. Thanks, @Mazorius for improving cloud-init!

Being this change a security related one, I will let it unmerged for a while to let others have a look.

TheRealFalcon (approved):
LGTM!

PR #1973, opened 2023-01-23
Fix default route rendering on v2 ipv6

TheRealFalcon:
## Proposed Commit Message

```
Fix default route rendering on v2 ipv6

::/0 would get rendered as ::/64 rather than ::/0 across all renderers
using ipv6 in a v2 config.

LP: #2003562
```

### DIFF
--- a/cloudinit/net/network_state.py
+++ b/cloudinit/net/network_state.py
@@ -989,7 +989,7 @@ def _normalize_net_keys(network, address_keys=()):
 
     @returns: A dict containing normalized prefix and matching addr_key.
     """
-    net = dict((k, v) for k, v in network.items() if v)
+    net = {k: v for k, v in network.items() if v or v == 0}
     addr_key = None
     for key in address_keys:
         if net.get(key):
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -5225,7 +5225,7 @@ USERCTL=no
                 # Created by cloud-init on instance boot automatically, do not edit.
                 #
                 2a00:1730:fff9:100::1/128 via ::0  dev eth0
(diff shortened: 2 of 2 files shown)


cloudinit/net/network_state.py
```diff
@@ -989,7 +989,7 @@ def _normalize_net_keys(network, address_keys=()):
 
     @returns: A dict containing normalized prefix and matching addr_key.
     """
-    net = dict((k, v) for k, v in network.items() if v)
+    net = {k: v for k, v in network.items() if v or v == 0}
```
aciba90, line 992: This would capture other falsey values as `"0"`.

```suggestion
    net = {k: v for k, v in network.items() if v is not None}
```

aciba90 (changes requested):
Nice catch here, @TheRealFalcon.

I have suggested an improvement for the patch. There is also a similar construct in https://github.com/canonical/cloud-init/blob/16cb229b8d2ae5fe8b49fd4df57d203ccbcadbbe/cloudinit/net/network_state.py#L955 that we could probably fix in this PR.

cloudinit/net/network_state.py
TheRealFalcon: "0" isn't truthy in Python. I think we do want to capture other falsey values here. 0 is the only one we want to pass through.

TheRealFalcon:
> There is also a similar construct that we could probably fix in this PR.

I don't think that this one has a prefix key, so I don't think the current way does any damage.

cloudinit/net/network_state.py
aciba90: My bad here.

aciba90 (approved):
LGTM, thanks.

PR #1976, opened 2023-01-24
CI: add dependabot to auto-update gh actions

aciba90:
## Proposed Commit Message

```
CI: add dependabot to auto-update gh actions
```

## Additional Context

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
We've had a few conversations about dependabot before. Since we can be deployed to multiple different environments, I'm not sure that dependabot makes a lot of sense for us. We generally need to keep our version ranges wide and not pinned to a specific version. Is there more that dependabot can do for us?

PR #1977, opened 2023-01-24
sources/azure: refactor imds handler into own module

cjp256:
Create new azure package for better organization and move IMDS logic for fetching into it.

Functional changes include comprehensive retry logic:

- Re-use metadata retry codes for reprovision data, even if these codes are unused.  There shouldn't be any regressive behavior as these codes currently would result in a retry after re-performing DHCP by the caller.

- Use 10 retries for reprovision data instead of infinite and custom exception callback.

Future work will clean up the test_azure.py tests a little further thanks to these changes, but wanted to minimize churn here to make changes fairly visible.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -6,7 +6,6 @@
 
 import base64
 import crypt
-import functools
 import os
 import os.path
 import re
@@ -15,8 +14,6 @@ from enum import Enum
 from pathlib import Path
 from time import sleep, time
 from typing import Any, Dict, List, Optional
-
-import requests
 
 from cloudinit import dmi
 from cloudinit import log as logging
@@ -29,6 +26,7 @@ from cloudinit.net.dhcp import (
 )
 from cloudinit.net.ephemeral import EphemeralDHCPv4
 from cloudinit.reporting import events
+from cloudinit.sources.azure import imds
 from cloudinit.sources.helpers import netlink
 from cloudinit.sources.helpers.azure import (
     DEFAULT_WIRESERVER_ENDPOINT,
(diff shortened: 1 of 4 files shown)


blackboxsw:
Assigning @TheRealFalcon as he's offered here. thanks James

TheRealFalcon (commented):
Overall LGTM.

My only comment is previously, two of the exception callbacks had a logging threshold with backoff to avoid diagnostic event / log spamming, but looks like they've been removed here. Is that spamming no longer a concern? Without the history, it's hard to know how big of a problem this was, but I'm assuming it was added for a reason at the time.

cjp256:
@TheRealFalcon good question.  With pre-provisioning, a VM may be created and held for a long time to be re-used.  In normal operation, we wait on a netlink event for the network interface to signal to us that re-use is occurring.  Should that fail for whatever reason, we'll ultimately keep polling IMDS for reprovisiondata to get the re-use configuration.  The concern was that we would be spamming the logs for whatever duration the VM is polling, which could be days or weeks.

Azure previously supported a "reprovisioning" mode which may cause the VM to reboot and force us into polling mode as we cannot rely on netlink events at that point.  Since this mode has been dropped, I don't think it's still a concern for long-term spamming.  Regardless, log_req_resp=False for this scenario and only reports the final result which is reasonable as it shouldn't happen more than a few times.

@anhvoms can tell me if I'm wrong :D

cjp256:
Upon further consideration, I decided to roll back the change in behavior for reprovisiondata (aside from some cleanup).  I shouldn't have changed the behavior in this PR.  We can revisit in the future, if needed.

TheRealFalcon (approved):
LGTM!

PR #1979, opened 2023-01-26
docs: minor polishing

s-makin:
```
Minor changes requested by Daniele for Diataxis:

1) Added "reverse" to front page section blocks so they display
   correctly on mobile/smaller screens.
2) Added links to security/performance pages from Explanation
   section. These are currently linking directly to the pages in
   the Development section and will eventually become more like
   signposts as more content is added/revised.
```


TheRealFalcon (commented):
Do we also want to remove those links/pages from the development section?

s-makin:
> Do we also want to remove those links/pages from the development section?

I'd recommend not -- as we discussed on Wednesday, these pages are currently intended for developers and so they should stay there as they are. The intention (following discussion with Daniele) is that the ones in the Explanation section will become like landing pages pointing to different aspects of security/performance and that every team will have these pages in their Explanation section. However, that will require substantial re-writing, and we don't yet have enough content on these topics to justify that approach. So, for the time being I'd rather have the links point to the same page as a placeholder (so at least it's useful info and not "coming soon"), pending future rewrites.

TheRealFalcon (approved):
LGTM!

PR #1981, opened 2023-01-27
cc_set_passwords: Prevent traceback when restarting ssh

TheRealFalcon:
## Proposed Commit Message

```
Prevent traceback when restarting ssh

On Bionic in GCE, when restarting ssh, it's not uncommon to see a
traceback with `ssh.service is not active, cannot reload.`. Instead log
a warning.
```

## Additional Context
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-bionic-gce/171/testReport/junit/tests.integration_tests.modules.test_set_password/TestPasswordListString/test_check_ssh_service/

### DIFF
--- a/cloudinit/config/cc_set_passwords.py
+++ b/cloudinit/config/cc_set_passwords.py
@@ -108,6 +108,18 @@ def get_users_by_type(users_list: list, pw_type: str) -> list:
     )
 
 
+def _restart_ssh_daemon(distro, service):
+    try:
+        distro.manage_service("restart", service)
+        LOG.debug("Restarted the SSH daemon.")
+    except subp.ProcessExecutionError as e:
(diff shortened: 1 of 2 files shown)


igalic (approved):
that seems sensible

TheRealFalcon:
I used warn because if restart fails, it could also be after the ssh module is running, which would mean the configuration never got applied. Most of the time it's likely to be harmless, but when it's not, I think the warn is appropriate.

holmanb (approved):
> I used warn because if restart fails, it could also be after the ssh module is running, which would mean the configuration never got applied. Most of the time it's likely to be harmless, but when it's not, I think the warn is appropriate.

+1 LGTM.

PR #1983, opened 2023-01-27
Wait for NetworkManager

rjschwei:
## Proposed Commit Message

Newer SUSE distributions are switching to NetworkManager from wicked. For those distributions we need to wait for NetworkManager before starting the cloud-init service.

## Checklist:

 - [x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)


TheRealFalcon (commented):
Would it be better to  add `suse` to the list in line 15?

rjschwei:
> Would it be better to add `suse` to the list in line 15?

Yes, fixed. Thanks for catching this.

TheRealFalcon (approved):
LGTM!

PR #1986, opened 2023-01-31
Fix Vultr cloud_interfaces usage

eb3095:
## Proposed Commit Message
```
cloud_interfaces is intended to be a netcfg. Change the helper function to reflect this.
```

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -280,7 +280,7 @@ def add_interface_names(netcfg):
     for interface in netcfg["config"]:
         if interface["type"] != "physical":
             continue
-        interface_name = get_interface_name(interface["mac"])
+        interface_name = get_interface_name(interface["mac_address"])
         if not interface_name:
             raise RuntimeError(
                 "Interface: %s could not be found on the system"
--- a/cloudinit/sources/helpers/vultr.py
+++ b/cloudinit/sources/helpers/vultr.py
@@ -284,10 +284,9 @@ def add_interface_names(netcfg):
         if not interface_name:
             raise RuntimeError(
                 "Interface: %s could not be found on the system"
-                % interface["mac"]
+                % interface["mac_address"]
             )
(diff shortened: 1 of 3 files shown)


cloudinit/sources/helpers/vultr.py
```diff
@@ -276,8 +276,10 @@ def generate_interface_additional_addresses(interface, netcfg):
 
 
 # Make required adjustments to the network configs provided
-def add_interface_names(interfaces):
-    for interface in interfaces:
+def add_interface_names(netcfg):
+    for interface in netcfg["config"]:
```
aciba90, line 280: This change implies that either:

1. This function has never been functional, or
2. There was a change in the instance metadata schema under `cloud_interfaces`. See https://github.com/canonical/cloud-init/blob/aacdff5b9298db2acbb5d10f0a42cb06d2c0a745/cloudinit/sources/DataSourceVultr.py#L73

aciba90 (commented):
Thanks, @eb3095.

I have been trying to look for [documentation](https://www.vultr.com/metadata/) about the `cloud_interfaces` key without success. Could you please help us to understand why this change is required?

Additionally, this is not covered by any unit tests. Could we please minimally cover it?

cloudinit/sources/helpers/vultr.py
eb3095: Its never been functional nor has it ever been used. This is our way of dropping in a fully configured, outside of interface names, cloud-init netcfg file in. Its not currently documented anywhere. As for unit tests, I will add that now.

aciba90: Thanks for the context and the unit test.

eb3095:
@aciba90 fixed the unit test and got this passing. Sorry ran into some python weirdness with the dict I had to work around. Looks to be acting correctly now however.

aciba90 (approved):
LGTM. Thanks, @eb3095.

PR #1987, opened 2023-01-31
improve resolvconf and network docs

holmanb:
```
docs: Clarify networking docs

- Add stronger warnings for resolv_conf module.
- Add link to lxd network config docs.
```

## Additional Context
I really don't like the current [networking docs](https://cloudinit.readthedocs.io/en/latest/reference/network-config.html), which has just a list of links to datasources and which formats they support. I think we could do a lot to improve user discovery of information for "how to use the format", however I don't want so do that in this PR so I'm just adding the LXD external link and kicking the "refactor network docs" can down the road.

Additionally I think stronger wording around the resolv_conf module is warranted, and some outdated statements removed.

[LP: #2004131](https://bugs.launchpad.net/cloud-init/+bug/2004131)

### DIFF
--- a/cloudinit/config/cc_resolv_conf.py
+++ b/cloudinit/config/cc_resolv_conf.py
@@ -26,12 +26,18 @@ RESOLVE_CONFIG_TEMPLATE_MAP = {
 }
 
 MODULE_DESCRIPTION = """\
+Unless manually editing :file:`/etc/resolv.conf` is the correct way to manage
+nameserver information on your operating system, you do not want to use
+this module. Many distros have moved away from manually editing ``resolv.conf``
+so please verify that this is the preferred nameserver management method for
+your distro :file:`/etc/resolv.conf` before using this module.
+
+Note that using :ref:`network_config` is preferred to this module, when
+possible.
+
 This module is intended to manage resolv.conf in environments where early
 configuration of resolv.conf is necessary for further bootstrapping and/or
 where configuration management such as puppet or chef own DNS configuration.
-As Debian/Ubuntu will, by default, utilize resolvconf, and similarly Red Hat
-will use sysconfig, this module is likely to be of little use unless those
-are configured correctly.
 
 When using a :ref:`datasource_config_drive` and a RHEL-like system,
 resolv.conf will also be managed automatically due to the available
@@ -44,10 +50,6 @@ must be set ``true``.
 .. note::
     For Red Hat with sysconfig, be sure to set PEERDNS=no for all DHCP
     enabled NICs.
-
-.. note::
-    And, in Ubuntu/Debian it is recommended that DNS be configured via the
-    standard /etc/network/interfaces configuration file.
 """
 
 meta: MetaSchema = {
--- a/cloudinit/config/cc_resolv_conf.py
+++ b/cloudinit/config/cc_resolv_conf.py
@@ -32,8 +32,8 @@ this module. Many distros have moved away from manually editing ``resolv.conf``
 so please verify that this is the preferred nameserver management method for
 your distro before using this module.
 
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):

cloudinit/config/cc_resolv_conf.py
```diff
@@ -26,12 +26,18 @@
...
+so please verify that this is the preferred nameserver management method for
+your distro :file:`/etc/resolv.conf` before using this module.
+
+Note that using :ref:`network_config` is preferred to this module, when
+possible.
+
 This module is intended to manage resolv.conf in environments where early
 configuration of resolv.conf is necessary for further bootstrapping and/or
 where configuration management such as puppet or chef own DNS configuration.
-As Debian/Ubuntu will, by default, utilize resolvconf, and similarly Red Hat
```
holmanb, line 32: Debian/Ubuntu aren't supported in this module, Red Hat no longer defaults to sysconfig (also not sure how that was ever relevant - on rhel 7 default method was always to edit /etc/resolv.conf ifaik)

cloudinit/config/cc_resolv_conf.py
```diff
@@ -44,10 +50,6 @@
 .. note::
     For Red Hat with sysconfig, be sure to set PEERDNS=no for all DHCP
     enabled NICs.
-
-.. note::
```
holmanb, line 48: Again, Ubuntu/Debian are irrelevant here.

cloudinit/config/cc_resolv_conf.py
```diff
@@ -26,12 +26,18 @@
 }
 
 MODULE_DESCRIPTION = """\
+Unless manually editing :file:`/etc/resolv.conf` is the correct way to manage
+nameserver information on your operating system, you do not want to use
+this module. Many distros have moved away from manually editing ``resolv.conf``
+so please verify that this is the preferred nameserver management method for
+your distro :file:`/etc/resolv.conf` before using this module.
```
s-makin, line 33:
```suggestion
your distro before using this module.
```

s-makin: Reading this sentence aloud, I got a bit lost in it, and I think this is the offending party

cloudinit/config/cc_resolv_conf.py
```diff
@@ -26,12 +26,18 @@
...
+
+Note that using :ref:`network_config` is preferred to this module, when
```
s-makin, line 35:
```suggestion
Note that using :ref:`network_config` is preferred, rather than using this module, when
```

cloudinit/config/cc_resolv_conf.py
```diff
@@ -26,12 +26,18 @@
...
+
 This module is intended to manage resolv.conf in environments where early
 configuration of resolv.conf is necessary for further bootstrapping and/or
 where configuration management such as puppet or chef own DNS configuration.
-As Debian/Ubuntu will, by default, utilize resolvconf, and similarly Red Hat
-will use sysconfig, this module is likely to be of little use unless those
-are configured correctly.
 
 When using a :ref:`datasource_config_drive` and a RHEL-like system,
 resolv.conf will also be managed automatically due to the available
```
s-makin, line 43: Can't leave a comment there, but line 45 has a confusing wording ("for those that with to have different settings")

s-makin (approved):
LGTM, apart from those couple of little nits

cloudinit/config/cc_resolv_conf.py
holmanb: My guess is that this was attempting to say "for those that **wish** to have different settings".

s-makin: I suspect so, but I didn't want to assume.

PR #1988, opened 2023-01-31
Ignore duplicate macs from mscc_felix and fsl_enetc

TheRealFalcon:
## Proposed Commit Message

```
Ignore duplicate macs from mscc_felix and fsl_enetc

mscc_felix and fsl_enetc are drivers representing a switch that is
expected to have duplicate macs. If we encounter either of these
drivers, we should not raise the duplicate mac exception.

LP: #1997922
```

## Additional Context
https://bugs.launchpad.net/cloud-init/+bug/1997922

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1034,6 +1034,16 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
                         % (ret[mac], driver_map[mac], name)
                     )
 
+            # This is intended to be a short-term fix of LP: #1997922
+            # Long term, we should better handle configuration of virtual
+            # devices where duplicate MACs are expected early in boot if
+            # cloud-init happens to enumerate network interfaces before drivers
+            # have fully initialized the leader/subordinate relationships for
+            # those devices or switches.
+            if driver == "mscc_felix" or driver == "fsl_enetc":
+                raise_duplicate_mac_error = False
+                continue
+
             if raise_duplicate_mac_error:
                 raise RuntimeError(msg)
 
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -8220,6 +8220,19 @@ class TestGetInterfacesByMac(CiTestCase):
         }
         self.assertEqual(expected, result)
 
+    def test_duplicate_ignored_macs(self):
+        # LP: #199792
+        self._data = copy.deepcopy(self._data)
+        self._data["macs"]["swp0"] = "9a:57:7d:78:47:c0"
+        self._data["macs"]["swp1"] = "9a:57:7d:78:47:c0"
+        self._data["own_macs"].append("swp0")
+        self._data["own_macs"].append("swp1")
+        self._data["drivers"]["swp0"] = "mscc_felix"
+        self._data["drivers"]["swp1"] = "mscc_felix"
+        self._mock_setup()
+        net.get_interfaces_by_mac()
+        # If the previous statement didn't Traceback, then test passed
+
 
 class TestInterfacesSorting(CiTestCase):
     def test_natural_order(self):


holmanb (approved):
The test, code, and comments look good to me. Lets let @blackboxsw take a look before merging, however.

cloudinit/net/__init__.py
```diff
@@ -1034,6 +1034,16 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
...
 
+            # This is intended to be a short-term fix of LP: #1997922
+            # Long term, we should better handle configuration of virtual
+            # devices where duplicate MACs are expected early in boot if
+            # cloud-init happens to enumerate network interfaces before drivers
+            # have fully initialized the leader/subordinate relationships for
+            # those devices or switches.
+            if driver == "mscc_felix" or driver == "fsl_enetc":
+                raise_duplicate_mac_error = False
+                continue
```
blackboxsw, line 1045: Since we are issuing`continue` in this case we actually don't need to set `raise_duplicate_mac_error` as we won't consume that anywhere. How about if we just log a similar warning breadcrumb that we do in Azure. As this is currently written, we silently ignore the duplicate macs and don't log any warnings.

```suggestion
+            if driver in ("mscc_felix", "fsl_enetc"):
                LOG.warning(
                    msg + " Ignoring '%s' due to driver '%s' and "
                    "'%s' having driver '%s'."
                    % (name, driver, ret[mac], driver_map[mac])
                )
                continue
```

tests/unittests/test_net.py
```diff
@@ -8220,6 +8220,19 @@ def test_ib(self):
...
+        # LP: #199792
+        self._data = copy.deepcopy(self._data)
+        self._data["macs"]["swp0"] = "9a:57:7d:78:47:c0"
+        self._data["macs"]["swp1"] = "9a:57:7d:78:47:c0"
+        self._data["own_macs"].append("swp0")
+        self._data["own_macs"].append("swp1")
+        self._data["drivers"]["swp0"] = "mscc_felix"
+        self._data["drivers"]["swp1"] = "mscc_felix"
+        self._mock_setup()
+        net.get_interfaces_by_mac()
```
blackboxsw, line 8233: We could use  does_not_raise and could check logs
```suggestion
        with does_not_raise():
            net.get_interfaces_by_mac()
        msg = (
            "both 'swp1' and 'swp0' have mac '9a:57:7d:78:47:c0'."
            " Ignoring 'swp1' due to driver 'mscc_felix' and 'swp0' having"
            " driver 'mscc_felix'."
        )
        self.assertIn(msg, self.logs.getvalue())
```

blackboxsw (approved):
Looks good, let's also add a log and validate that behavior.
combined diff
```diff
diff --git a/cloudinit/net/__init__.py b/cloudinit/net/__init__.py
index 487742e6f..6f9387ce6 100644
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1040,8 +1040,12 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
             # cloud-init happens to enumerate network interfaces before drivers
             # have fully initialized the leader/subordinate relationships for
             # those devices or switches.
-            if driver == "mscc_felix" or driver == "fsl_enetc":
-                raise_duplicate_mac_error = False
+            if driver in ("mscc_felix", "fsl_enetc"):
+                LOG.warning(
+                    msg + " Ignoring '%s' due to driver '%s' and "
+                    "'%s' having driver '%s'."
+                    % (name, driver, ret[mac], driver_map[mac])
+                )
                 continue
 
             if raise_duplicate_mac_error:
diff --git a/tests/unittests/test_net.py b/tests/unittests/test_net.py
index 40baddc88..c64715b11 100644
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -34,6 +34,7 @@ from tests.unittests.helpers import (
     CiTestCase,
     FilesystemMockingTestCase,
     dir2dict,
+    does_not_raise,
     mock,
     populate_dir,
 )
@@ -8008,6 +8009,7 @@ class TestInterfaceHasOwnMac(CiTestCase):
     mock.Mock(return_value=False),
 )
 class TestGetInterfacesByMac(CiTestCase):
+    with_logs = True
     _data = {
         "bonds": ["bond1"],
         "bridges": ["bridge1"],
@@ -8230,7 +8232,14 @@ class TestGetInterfacesByMac(CiTestCase):
         self._data["drivers"]["swp0"] = "mscc_felix"
         self._data["drivers"]["swp1"] = "mscc_felix"
         self._mock_setup()
-        net.get_interfaces_by_mac()
+        with does_not_raise():
+            net.get_interfaces_by_mac()
+        msg = (
+            "both 'swp1' and 'swp0' have mac '9a:57:7d:78:47:c0'."
+            " Ignoring 'swp1' due to driver 'mscc_felix' and 'swp0' having"
+            " driver 'mscc_felix'."
+        )
+        self.assertIn(msg, self.logs.getvalue())
         # If the previous statement didn't Traceback, then test passed
 
 
```

TheRealFalcon:
Thanks @blackboxsw . I modified the log somewhat. Since this behavior is expected for this device, I don't think we should be warning.

tests/unittests/test_net.py
```diff
@@ -8234,10 +8234,11 @@ def test_duplicate_ignored_macs(self):
         self._mock_setup()
         with does_not_raise():
             net.get_interfaces_by_mac()
-        assert (
-            "Ignoring duplicate macs from 'swp0' and 'swp1' due to "
+        pattern = (
+            "Ignoring duplicate macs from 'swp[0-1]' and 'swp[0-1]' due to "
```
blackboxsw, line 8238: @TheRealFalcon I think this is python-version-specific for how `list`s are generated from `set`s?  Given that _mock_setup is creating a set from a list of mac,keys() without explicitly calling sorted, I think that's what we are hitting here.

Let's try this diff instead.
```diff
diff --git a/tests/unittests/test_net.py b/tests/unittests/test_net.py
index 056aaeb6f..2e58e685e 100644
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -7862,7 +7862,7 @@ class TestGetInterfaces(CiTestCase):
     data: dict = {}
 
     def _se_get_devicelist(self):
-        return list(self.data["devices"])
+        return sorted(list(self.data["devices"]))
 
     def _se_device_driver(self, name):
         return self.data["drivers"][name]
@@ -8234,11 +8234,10 @@ class TestGetInterfacesByMac(CiTestCase):
         self._mock_setup()
         with does_not_raise():
             net.get_interfaces_by_mac()
-        pattern = (
-            "Ignoring duplicate macs from 'swp[0-1]' and 'swp[0-1]' due to "
+        assert (
+            "Ignoring duplicate macs from 'swp0' and 'swp1' due to "
             "driver 'mscc_felix'."
-        )
-        assert re.search(pattern, self.logs.getvalue())
+        ) in self.logs.getvalue()
 
 
 class TestInterfacesSorting(CiTestCase):
```

blackboxsw: No such luck. py 3.7 then failed. :/

PR #1989, opened 2023-02-01
integration tests: Refactor instance checking

TheRealFalcon:
## Proposed Commit Message

```
integration tests: Refactor instance checking

Using invidual release and platform  marks to specify our test support
matrix was leading to too many marks specifying different combinations
of things. Rather, we can rely on the "skipif" mark to perform any
needed release or platform checks.
```

## Additional Context
~~I want to get buy-in on the idea before converting all the tests. I converted 2 tests to show what it would look like (one for release, one for platform), but other tests haven't been modified yet.~~

~~This should also replace [the image specification code](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/clouds.py/#L37-L93) completely, but I didn't want to break working tests yet.~~

I intentionally kept things flat. It feels a bit weird importing module-level variables, and enums could possibly help discoverability, but I really didn't like the lengths when throwing in enum/class names everywhere. I'm open to changing it though.

### DIFF
--- a/tests/integration_tests/bugs/test_gh868.py
+++ b/tests/integration_tests/bugs/test_gh868.py
@@ -2,6 +2,7 @@
 import pytest
 
 from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.integration_settings import PLATFORM
 from tests.integration_tests.util import verify_clean_log
 
 USERDATA = """\
@@ -15,12 +16,9 @@ chef:
 
 
 @pytest.mark.adhoc  # Can't be regularly reaching out to chef install script
-@pytest.mark.ec2
-@pytest.mark.gce
-@pytest.mark.azure
-@pytest.mark.oci
-@pytest.mark.lxd_container
-@pytest.mark.lxd_vm
+@pytest.mark.skipif(
+    "openstack" == PLATFORM, reason="Firewall preventing openstack run"
+)
 @pytest.mark.user_data(USERDATA)
 def test_chef_license(client: IntegrationInstance):
     log = client.read_from_file("/var/log/cloud-init.log")
--- a/tests/integration_tests/bugs/test_lp1898997.py
+++ b/tests/integration_tests/bugs/test_lp1898997.py
@@ -12,6 +12,8 @@ default gateway.
 import pytest
 
 from tests.integration_tests import random_mac_address
+from tests.integration_tests.integration_settings import PLATFORM
+from tests.integration_tests.releases import CURRENT_RELEASE, FOCAL
 from tests.integration_tests.util import verify_clean_log
 
 MAC_ADDRESS = random_mac_address()
@@ -44,10 +46,12 @@ version: 2
         "volatile.eth0.hwaddr": MAC_ADDRESS,
     }
 )
-@pytest.mark.lxd_vm
+@pytest.mark.skipif(
+    PLATFORM != "lxd_vm",
+    reason="Test requires custom networking provided by LXD",
+)
+@pytest.mark.skipif(CURRENT_RELEASE < FOCAL)
 @pytest.mark.lxd_use_exec
-@pytest.mark.not_bionic
-@pytest.mark.ubuntu
 class TestInterfaceListingWithOpenvSwitch:
     def test_ovs_member_interfaces_not_excluded(self, client):
         # We need to install openvswitch for our provided network configuration
--- a/tests/integration_tests/bugs/test_lp1898997.py
+++ b/tests/integration_tests/bugs/test_lp1898997.py
@@ -50,7 +50,9 @@ version: 2
     PLATFORM != "lxd_vm",
     reason="Test requires custom networking provided by LXD",
 )
-@pytest.mark.skipif(CURRENT_RELEASE < FOCAL)
+@pytest.mark.skipif(
+    CURRENT_RELEASE < FOCAL, reason="Tested on Focal and above"
+)
 @pytest.mark.lxd_use_exec
 class TestInterfaceListingWithOpenvSwitch:
     def test_ovs_member_interfaces_not_excluded(self, client):
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -7,7 +7,7 @@ import re
 import string
 from abc import ABC, abstractmethod
 from copy import deepcopy
-from typing import Optional, Type
+from typing import Type
 from uuid import UUID
 
 from pycloudlib import (
--- a/tests/integration_tests/clouds.py
+++ b/tests/integration_tests/clouds.py
@@ -28,6 +28,7 @@ import cloudinit
 from cloudinit.subp import ProcessExecutionError, subp
 from tests.integration_tests import integration_settings
 from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.releases import CURRENT_RELEASE
 from tests.integration_tests.util import emit_dots_on_travis
 
 log = logging.getLogger("integration_testing")
@@ -127,12 +128,9 @@ class IntegrationCloud(ABC):
         raise NotImplementedError
 
     def _get_initial_image(self, **kwargs) -> str:
-        image = ImageSpecification.from_os_image()
-        try:
-            return self.cloud_instance.daily_image(image.image_id, **kwargs)
-        except (ValueError, IndexError) as ex:
-            log.debug("Exception while executing `daily_image`: %s", ex)
-            return image.image_id
+        return CURRENT_RELEASE.image_id or self.cloud_instance.daily_image(
+            CURRENT_RELEASE.series, **kwargs
+        )
 
     def _perform_launch(self, launch_kwargs, **kwargs) -> BaseInstance:
         pycloudlib_instance = self.cloud_instance.launch(**launch_kwargs)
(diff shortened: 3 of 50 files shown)


tests/integration_tests/releases.py
```diff
@@ -0,0 +1,93 @@
+import functools
+import logging
+from typing import Optional
+
+from packaging import version
```
holmanb, line 5: Is packaging an indirect dependency? I think we should probably make this an explicit requirement if we're going to explicitly use it.

tests/integration_tests/bugs/test_gh868.py
```diff
@@ -15,12 +16,9 @@
...
 @pytest.mark.adhoc  # Can't be regularly reaching out to chef install script
-@pytest.mark.ec2
-@pytest.mark.gce
-@pytest.mark.azure
-@pytest.mark.oci
-@pytest.mark.lxd_container
-@pytest.mark.lxd_vm
+@pytest.mark.skipif(
+    "openstack" == PLATFORM, reason="Firewall preventing openstack run"
+)
```
holmanb, line 21: brilliant

tests/integration_tests/clouds.py
```diff
@@ -127,12 +128,9 @@ def _get_cloud_instance(self):
...
 
     def _get_initial_image(self, **kwargs) -> str:
-        image = ImageSpecification.from_os_image()
-        try:
-            return self.cloud_instance.daily_image(image.image_id, **kwargs)
-        except (ValueError, IndexError) as ex:
-            log.debug("Exception while executing `daily_image`: %s", ex)
-            return image.image_id
+        return CURRENT_RELEASE.image_id or self.cloud_instance.daily_image(
+            CURRENT_RELEASE.series, **kwargs
```
holmanb, line 132: much cleaner

tests/integration_tests/test_paths.py
```diff
@@ -111,7 +112,10 @@ def collect_logs(self, custom_client: IntegrationInstance):
     # because the test ensures nothing is running under /var/lib/cloud.
     # Since LXD is doing this and not cloud-init, we should just not run
     # on Bionic to avoid it.
-    @pytest.mark.not_bionic
+    @pytest.mark.skipif(
+        CURRENT_RELEASE < FOCAL,
+        reason="LXD inserts conflicting setup on releases prior to focal",
```
holmanb, line 117: love it

tests/integration_tests/releases.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+XENIAL = Release("ubuntu", "xenial", "16.04")
+BIONIC = Release("ubuntu", "bionic", "18.04")
+FOCAL = Release("ubuntu", "focal", "20.04")
+GROOVY = Release("ubuntu", "jammy", "22.04")
+HIRSUTE = Release("ubuntu", "kinetic", "22.10")
+
+CURRENT_RELEASE = Release.from_os_image()
+RELEASE_IS_UBUNTU = CURRENT_RELEASE.os == "ubuntu"
```
holmanb, line 92: Could we maybe do `IS_UBUNTU` and `IS_NOT_UBUNTU`? In pytest markers I think this will read cleaner.

holmanb (commented):
Nice direction, it's very clean.

> I really didn't like the lengths when throwing in enum/class names everywhere.

This reads really nice in tests; I'm +1 for the way you have it.

> I want to get buy-in on the idea before converting all the tests.

+1 Lets do it

Nit: I think that the current summary line in the proposed commit message makes sense, but could be more descriptive. Out of context it could mean a bunch of different things.

tests/integration_tests/releases.py
```diff
@@ -0,0 +1,93 @@
...
+            ["ubuntu-distro-info", "--release", "--series", series]
+        )
+    except subp.ProcessExecutionError as e:
+        raise ValueError(
+            f"'{series}' is not a recognized Ubuntu release"
+        ) from e
+    return out.rstrip(" LTS")
+
+
+@functools.total_ordering
```
blackboxsw, line 39: Doing comparisons of series is great. How do you envision supporting multiple distribution comparisons for tests that may be applicable to both (debian < bookworm) or (ubuntu < lunar) etc

blackboxsw (commented):
driveby comments I like the general direction and I think it simplifies our tox marks to have the specific operator-based comparisons listed in the specific tests (as well as skip reasons).

tests/integration_tests/releases.py
```diff
@@ -0,0 +1,93 @@
...
+                "of <image_id>[::<os>[::<release>[::<version>]]]"
+            )
+        return cls(os, series, version, image_id)
+
+
+XENIAL = Release("ubuntu", "xenial", "16.04")
+BIONIC = Release("ubuntu", "bionic", "18.04")
+FOCAL = Release("ubuntu", "focal", "20.04")
+GROOVY = Release("ubuntu", "jammy", "22.04")
+HIRSUTE = Release("ubuntu", "kinetic", "22.10")
```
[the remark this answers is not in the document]
TheRealFalcon: > What will we do for non-Ubuntu?

Basically the same. Define the releases we'll be testing in one big section. Do you think that'll be problematic?

tests/integration_tests/releases.py
TheRealFalcon: It's slightly more verbose, but something like this should work, right?
```python
# Ubuntu
BIONIC = Release("ubuntu", "bionic", "18.04")
FOCAL = Release("ubuntu", "focal", "20.04")
JAMMY = Release("ubuntu", "jammy", "22.04")
KINETIC = Release("ubuntu", "kinetic", "22.10")
LUNAR = Release("ubuntu", "lunar", "23.04")

# Debian
BULLSEYE = Release("debian", "bullseye", "11")
BOOKWORM = Release("debian", "bookworm", "12")

CURRENT_RELEASE = Release.from_os_image()
IS_UBUNTU = CURRENT_RELEASE.os == "ubuntu"
IS_DEBIAN = CURRENT_RELEASE.os == "debian"
```

Then:
```python
@pytest.mark.skipif(
    (IS_UBUNTU and CURRENT_RELEASE < LUNAR)
    or (IS_DEBIAN and CURRENT_RELEASE < BOOKWORM),
    reason="The turbo encabulator doesn't exist in older versions",
)
```
for more than 2, using `any()` might look nicer.

tests/integration_tests/releases.py
```diff
@@ -0,0 +1,93 @@
...
+        self.series = release
+        self.version = version
+        self.image_id = image_id
+
+    def __repr__(self):
+        return f"Release({self.os}, {self.version})"
+
+    def __lt__(self, other: "Release"):
+        if self.os != other.os:
+            return False
```
holmanb, line 58: From [the docs](https://docs.python.org/3/reference/datamodel.html#object.__lt__):

> A rich comparison method may return the singleton NotImplemented if it does not implement the operation for a given pair of arguments. By convention, False and True are returned for a successful comparison.

I think we should probably shouldn't return `False` when comparing `ubuntu` to `debian`.

holmanb (changes requested):

tests/integration_tests/releases.py
TheRealFalcon: Good point, I was thinking more equality. An exception seems more appropriate here.

TheRealFalcon: Thanks. I think it actually makes more sense to raise an exception here rather than return NotImplemented:
```
  File "/home/james/unrelated_python.py", line 3, in <module>
    print(releases.LUNAR > releases.SOMETHING)
TypeError: '>' not supported between instances of 'Release' and 'Release'
```

vs

```
  File "/home/james/unrelated_python.py", line 3, in <module>
    print(releases.LUNAR > releases.SOMETHING)
  File "/home/james/.pyenv/versions/3.7.13/lib/python3.7/functools.py", line 93, in _gt_from_lt
    op_result = self.__lt__(other)
  File "/home/james/c/cloud-init/tests/integration_tests/releases.py", line 58, in __lt__
    raise ValueError(f"{self.os} cannot be compared to {other.os}!")
ValueError: ubuntu cannot be compared to debian!
```

TheRealFalcon:
Latest push updates all call points. Not tested yet so likely more changes still to come. I just wanted to push what I had.

tests/integration_tests/releases.py
holmanb: Agreed, I think that looks better.

holmanb (approved):
This looks really nice. I left some inline nitpicks and comments - feel free to take what you want and ditch the rest.

tests/integration_tests/bugs/test_lp1898997.py
```diff
@@ -44,10 +46,14 @@
...
+@pytest.mark.skipif(
+    PLATFORM != "lxd_vm",
+    reason="Test requires custom networking provided by LXD",
+)
+@pytest.mark.skipif(
+    CURRENT_RELEASE < FOCAL, reason="Tested on Focal and above"
+)
 @pytest.mark.lxd_use_exec
-@pytest.mark.not_bionic
-@pytest.mark.ubuntu
```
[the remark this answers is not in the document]
TheRealFalcon: It's implicit in the `CURRENT_RELEASE < FOCAL` check. Do you think it'd be better to have an explicit ubuntu check here too?

PR #1991, opened 2023-02-02
tests: Remove restart check from test

TheRealFalcon:
## Proposed Commit Message

```
tests: Remove restart check from test

If no SSH restart happens, we get no log. Since this is perfectly
normal, we shouldn't be checking for a log.
```

## Checklist:


blackboxsw (approved):
LGTM

PR #1994, opened 2023-02-03
Ubuntu/devel: new upstream snapshot

TheRealFalcon:


debian/changelog
```diff
@@ -1,10 +1,13 @@
-cloud-init (22.4.2-0ubuntu3) UNRELEASED; urgency=medium
+cloud-init (22.4.2-0ubuntu3) lunar; urgency=medium
 
   * Upstream snapshot based on upstream/main at bdc7b040.
     - Bugs fixed in this snapshot: (LP: #2001737, #1998526, #2000596)
       (LP: #1851438)
+  * Upstream snapshot based on upstream/main at 3b8b4692.
+    - Bugs fixed in this snapshot: (LP: #1998998, #1997922, #2002969)
+      (LP: #2003562, #1999680)
```
blackboxsw, line 8: @TheRealFalcon given that we have pre-released content here, I think we actually want a minor manual fix here (not needed in tooling). To consolidate both sets of entries into a single:
```
  Upstream snapshot based on upstream/main at 3b8b4692.
  - Bugs fixed in this snapshot: (LP: #2001737, #1998526, #2000596)
     (LP: #1851438, #1998998, #1997922, #2002969, #2003562)
     (LP: #1999680)
```

blackboxsw: ` cloud-init | 22.4.2-0ubuntu2                    | lunar           | source, all`

blackboxsw: As we also discussed, this does represent that internally (before publication) that we did perform a snapshot to fix internal PPA build recipes. So, no harm in preserving that content.

blackboxsw (approved):
* only authorship changes in PR (neglecting the suggestion for consolidated single d/changelog snapshot message
* SUCCESS build-package; sbuild package on lunar no new lintian warnings
* no d/patches to cope with, ignoring quilt push/pop -a validation
* version number looks good for interim release in lunar: `(22.4.2-0ubuntu3) lunar`

PR #1995, opened 2023-02-04
schema: fix gpt labels, use type string for GUID

holmanb:
```
schema: fix gpt labels, support type string for GUID

LP #2004599
```

## Additional Context
https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2004599


aciba90 (approved):
LGTM, thanks.

PR #1997, opened 2023-02-04
Support transactional-updates for SUSE based distros

rjschwei:
## Proposed Commit Message

openSUSE/SUSE has distros that use read only root and btrfs. To update a running system in such a setup the transactional-update command needs to be used. This change implements support for use of the transactional-update commend when appropriate.
```

Support transactional-updates for SUSE based distros
```

## Checklist:

 - [ x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [x ] I have updated or added any unit tests accordingly

### DIFF
--- a/cloudinit/distros/opensuse.py
+++ b/cloudinit/distros/opensuse.py
@@ -74,39 +74,10 @@ class Distro(distros.Distro):
         )
 
     def package_command(self, command, args=None, pkgs=None):
-        pckg_op = command
         if pkgs is None:
             pkgs = []
 
-        if self.update_method is None:
-            result = util.get_mount_info("/")
-            fs_type = ""
-            if result:
-                (devpth, fs_type, mount_point) = result
-                # Check if the file system is read only
-                mounts = util.load_file("/proc/mounts").split("\n")
-                for mount in mounts:
-                    if mount.startswith(devpth):
-                        mount_info = mount.split()
-                        if mount_info[1] != mount_point:
-                            continue
-                        if mount_info[3].startswith("ro"):
-                            self.read_only_root = True
-                        else:
-                            self.read_only_root = False
-                        break
-                if fs_type.lower() == "btrfs" and os.path.exists(
-                    "/usr/sbin/transactional-update"
-                ):
-                    self.update_method = "transactional"
-                else:
-                    self.update_method = "zypper"
-            else:
-                LOG.info(
-                    "Could not determine filesystem type of '/' using zypper"
-                )
-                self.update_method = "zypper"
-
+        self._set_update_method()
         if self.read_only_root and not self.update_method == "transactional":
             LOG.error(
                 "Package operation requested but read only root "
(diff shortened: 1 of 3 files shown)


cloudinit/distros/opensuse.py
```diff
@@ -66,15 +74,72 @@ def install_packages(self, pkglist):
         )
 
     def package_command(self, command, args=None, pkgs=None):
+        pckg_op = command
         if pkgs is None:
             pkgs = []
 
+        if self.update_method is None:
```
igalic, line 81: this whole if could be a helper function to make this function easier to read

cloudinit/distros/opensuse.py
```diff
@@ -66,15 +74,72 @@ def install_packages(self, pkglist):
...
+            if result:
+                (devpth, fs_type, mount_point) = result
+                # Check if the file system is read only
+                mounts = util.load_file("/proc/mounts").split("\n")
+                for mount in mounts:
+                    if mount.startswith(devpth):
+                        mount_info = mount.split()
+                        if mount_info[1] != mount_point:
+                            continue
+                        if mount_info[3].startswith("ro"):
```
TheRealFalcon, line 93: Nit: `self.read_only_root = mount_info[3].startswith("ro")`

cloudinit/distros/opensuse.py
```diff
@@ -66,15 +74,72 @@ def install_packages(self, pkglist):
...
+                "--non-interactive",
+                "--drop-if-no-change",
+                "pkg",
+            ]
 
         # Command is the operation, such as install
-        if command == "upgrade":
+        if pckg_op == "upgrade":
             command = "update"
+        if self.update_method == "transactional" and not pkgs:
```
TheRealFalcon, line 131: Nit: It might be a little cleaner to move this into the else above (with a subcondition checking for pkgs). Up to you.

cloudinit/distros/opensuse.py
```diff
@@ -66,15 +74,72 @@ def install_packages(self, pkglist):
         )
 
     def package_command(self, command, args=None, pkgs=None):
+        pckg_op = command
```
TheRealFalcon, line 77: I'm not quite understanding the difference between `pckg_op` and `command`. Here we set `pckg_op` to `command`, but then later down we set `command` based on `pckg_op`. Is there a way to simplify that?

cloudinit/distros/opensuse.py
TheRealFalcon: I agree. Something like a `set_update_method` would make it easier to follow.

TheRealFalcon (changes requested):
Overall looks good. Inline comments are mostly nits, but I think it'd be good to separate out most of the new functionality into a separate function.

cloudinit/distros/opensuse.py
rjschwei: This was need, I thought because, because I was basically checking for "command=='upgrade'" twice but in the second check it may have already been set to "update". This was a logic bug.  No longer needed.

cloudinit/distros/opensuse.py
rjschwei: This has to be here as the command could be "update" or "upgrade".

cloudinit/distros/opensuse.py
rjschwei: Done

cloudinit/distros/opensuse.py
rjschwei: Done

TheRealFalcon (approved):
LGTM!

PR #2000, opened 2023-02-07
docs: add google search verification

s-makin:
```
Added google verification file to ensure that the 
Google bots can find and index the site. Should
speed up the listing of our docs in Google's 
search results. 
```


aciba90 (approved):
LGTM, thanks.

- [x] verified the `google<id>.html` is generated in the root of the html site (`doc/rtd_html`).

PR #2001, opened 2023-02-07
readme: add reference to packages dir

aciba90:
## Proposed Commit Message

```
readme: add reference to packages dir
```

## Additional Context

[LP: #2006065](https://bugs.launchpad.net/cloud-init/+bug/2006065)

### DIFF
--- a/packages/README.md
+++ b/packages/README.md
@@ -2,7 +2,18 @@
 
 Package builders under this folder are development only templates. Do not rely on them.
 
-Downstream packaging resources:
+## Build/Install
+
+Cloud-init's build/install procedure is not OS/Distro independent as cloud-init
+is tightly couple to OS implementation details, as for example,
+the init units' definitions, see [systemd/](systemd/) and [sysvinit/](sysvinit/).
+
+For users interested into trying out cloud-init, a pre-built image is the easiest option.
+
+For users interested into packaging cloud-init, see the reference implementations under this folder
+and official packages in the following section.
+
+## Downstream packaging resources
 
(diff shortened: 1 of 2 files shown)


holmanb:
I think that Håvard is getting at something important. Most projects have a **Build, Test, Install** section, which also doubles as a starting point for anyone trying to use the project. Since our **Build/Install** isn't really independent of OS/Distro packaging and implementation details, we should probably add something to the packaging page explaining this - it is a bit unique for a project. This is due to the code being tightly coupled to OS implementation, but the typical user might not realize this.

Also, I didn't realize this packages readme existed. I have a list of links to some other downstream resources that I'll include a PR to shortly.

aciba90:
@holmanb, thanks for pointing out that. I have added documentation about it.

packages/README.md
```diff
@@ -2,7 +2,18 @@
...
-Downstream packaging resources:
+## Build/Install
+
+Cloud-init's build/install procedure is not OS/Distro independent as cloud-init
+is tightly couple to OS implementation details, as for example,
+the init units' definitions, see [systemd/](systemd/) and [sysvinit/](sysvinit/).
+
+For users interested into trying out cloud-init, a pre-built image is the easiest option.
+
+For users interested into packaging cloud-init, see the reference implementations under this folder
```
holmanb, line 13:
```suggestion
For users interested in packaging cloud-init, see the reference implementations under this folder
```

holmanb (approved):
I have one small wording change request. Otherwise it looks good to me.

PR #2012, opened 2023-02-08
disk_setup: use byte string when purging the partition table

ederst:
## disk_setup: use byte string when purging the partition table

```
disk_setup: use byte string when purging the partition table

This writes a byte string to the device instead of a string when
purging the partition table.

Essentially, this will prevent the error "a bytes-like object is
required, not 'str'" from happening.
```

## Additional Context

This will prevent the error "a bytes-like object is required, not 'str'" from happening.

## Test Steps

Using the following disk setup in the user-data:

```yaml
device_aliases:
  my_disk: /dev/vdb

disk_setup:
  my_disk:
    layout: remove
    overwrite: true
```

Results in an error/warning in the cloud-init log:

```
2023-02-08 20:48:55,158 - util.py[WARNING]: Failed partitioning operation
a bytes-like object is required, not 'str'
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_disk_setup.py
+++ b/cloudinit/config/cc_disk_setup.py
@@ -657,7 +657,7 @@ def get_partition_gpt_layout(size, layout):
 def purge_disk_ptable(device):
     # wipe the first and last megabyte of a disk (or file)
     # gpt stores partition table both at front and at end.
-    null = "\0"
+    null = b"\0"
     start_len = 1024 * 1024
     end_len = 1024 * 1024
     with open(device, "rb+") as fp:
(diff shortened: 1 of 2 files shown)


ederst:
@holmanb i have opted to stay with the add `b` to the string instead of using `wipefs`, and have added a simple unittest for it too, so the PR can be reviewed now

holmanb (approved):
LGTM, thanks!

```bash
$ python3 -c 'open("./a.txt", "rb+").write("\0")'                    
TypeError: a bytes-like object is required, not 'str'
$ python3 -c 'open("./a.txt", "rb+").write(b"\0")'
```

PR #2013, opened 2023-02-08
docs: deprecation generation support [1/2]

holmanb:
```
docs: deprecation version generation support

schema: Add the following metadata keys:
 - changed_version
 - deprecated_version
 - new_version
 - changed_description
 - new_description
 - deprecation_description
 - changed
 - new

Generate formatted docs from this new metadata.
Update current descriptions containing unstructured data.
Switch current deprecation from a date to deprecated version.
Ensure changed/new/deprecated keys have an associated version key.
```

## Additional Context
The "changed" attribute is added to accommodate for keys which are not themselves deprecated, but which may have possible values that are deprecated. The `changed_description` field is especially valuable in this scenario to tell the user _what_ about the key changed. This PR intends to make clearer the distinction between a key's purpose/usage, and metadata describing the state of the key.

Inspiration for this model comes directly from the upstream Python language documentation. See examples of [new](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup), [changed](https://docs.python.org/3/library/asyncio-task.html#sleeping), and [deprecated](https://docs.python.org/3/library/ssl.html#ssl.RAND_pseudo_bytes) attributes.

Example:

Note that since the changed/deprecation message is italicized, which currently means that RST in the description doesn't currently get rendered. This needs to be resolved before merging[1].

Format tests may fail, but I want to get reviews before fixing that up.

[1] @s-makin Do you know if there is a way to render nested RST? If not, do you have an opinion about what would be best here?


s-makin:
> [1] @s-makin Do you know if there is a way to render nested RST? If not, do you have an opinion about what would be best here?

I'm not 100% sure I've properly understood what you mean by nested RST. It might be worth us having a chat later to clarify how you want the page to look.

aciba90:
Thanks, @holmanb. I like this idea a lot. A couple of questions:

1. Have you thought about the possibility of modeling the deprecation/change/new info as a class/object? To have more structured and extensible entities. Something in the lines of:

```json
{
...
  "<deprecated | changed | new>": true,
  "<deprecated | changed | new>_info": {
    "version": "22.3",
    "description": "Use ``lock_passwd`` instead."
  }
}
```

2. We could have objects that are deprecated, changed and new at the same time, right?

holmanb (commented):
> 1. Have you thought about the possibility of modeling the deprecation/change/new info as a class/object? To have more structured and extensible entities. Something in the lines of:

No. However, I like the idea.

> ```json
> {
> ...
>   "<deprecated | changed | new>": true,
>   "<deprecated | changed | new>_info": {
>     "version": "22.3",
>     "description": "Use ``lock_passwd`` instead."
>   }
> }
> ```

Perhaps while we're making this change, we also change the existing `deprecated`/`changed`/`new` boolean to be an object key, since it is now redundant due to requiring the version key:

> ```json
> {
> ...
>   "new": {
>     "version": "19.3",
>   },
>   "changed": {
>     "version": "22.3",
>     "description": "Use ``lock_passwd`` instead."
>   }
> }
> ```
What do you think about that?

>     2. We could have objects that are deprecated, changed and new at the same time, right?

I wasn't thinking about this possibility during implementation, but yes that could (and will) certainly happen. I'll work on support for this functionality.

aciba90:
> Perhaps while we're making this change, we also change the existing `deprecated`/`changed`/`new` boolean to be an object key, since it is now redundant due to requiring the version key:
> 
> > ```json
> > {
> > ...
> >   "new": {
> >     "version": "19.3",
> >   },
> >   "changed": {
> >     "version": "22.3",
> >     "description": "Use ``lock_passwd`` instead."
> >   }
> > }
> > ```
> 
> What do you think about that?

That's even more compact. My only concern is that [deprecated](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) was added as an official key in [2019-09](https://json-schema.org/draft/2019-09/release-notes.html#meta-data-vocabulary) jsonschema version as a boolean type. We could break application tooling consuming this schema.

A solution could be to keep a `"deprecated": true` and use an alternative name for the deprecated object:

```json
{
...
  "new": {
    "version": "19.3",
  },
  "changed": {
    "version": "22.3",
    "description": "Use ``lock_passwd`` instead."
  },
  "deprecated": true,
  "deprecated_": {
    "version": "20.1"
  }
}
```

holmanb:
> We could have objects that are deprecated, changed and new at the same time, right?

Thanks for the review @aciba90. I updated the tests for this functionality and updated the validator as well to log for `"change"` keys, since they indicate a change in functionality that the user may need to be warned of in the same way that we warn of deprecated keys.

> That's even more compact. My only concern is that [deprecated](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.3) was added as an official key in [2019-09](https://json-schema.org/draft/2019-09/release-notes.html#meta-data-vocabulary) jsonschema version as a boolean type. We could break application tooling consuming this schema.

In light of that I think I prefer your original suggestion over the other possibilities we've discussed.

Since this seems preferential and doesn't change anything functional, I'll leave it as is in case any other reviewers have a preference they'd like to voice. I can change implementation to whichever format we all prefer later.

Requesting re-review @aciba90.

doc/rtd/reference/modules.rst
```diff
@@ -3,6 +3,21 @@
 Module reference
 ****************
 
+Deprecation Schedule and Versions
```
s-makin, line 6:
```suggestion
Deprecation schedule and versions
```

tests/unittests/config/test_schema.py
```diff
@@ -1806,6 +1924,6 @@ def test_handle_schema_args_annotate_deprecated_config(
         )
         handle_schema_args("unused", args)
         out, err = capsys.readouterr()
-        assert expected_output.format(cfg_file=user_data_fn) == out
+        assert expected_output.format(user_data_fn).split() == out.split()
         assert not err
         assert "deprec" not in caplog.text
```
aciba90, line 1929: I know this is not part of this PR, but just for the record.

This check might not be sufficient, as cloud-init logs the following line, containing a full path to the jsonschema and potentially any valid string for a path:

```log
2023-02-15 11:54:06 DEBUG     cloudinit.util:util.py:1531 Reading from /home/aciba/canonical/worktrees/cloud-init/holmanb/deprecation-schema/cloudinit/config/schemas/schema-cloud-config-v1.json (quiet=False)\n
```

cloudinit/config/schema.py
```diff
@@ -843,32 +879,36 @@ def _get_property_description(prop_config: dict) -> str:
     Order and deprecated property description after active descriptions.
     Add a trailing stop "." to any description not ending with ":".
     """
-    prop_descr = prop_config.get("description", "")
+
+    def assign_descriptions(
+        config: dict, descriptions: list, deprecated_descriptions: list
+    ):
+        if any(
+            [
```
aciba90, line 887: Nit: what do you think about the following change?

```diff
diff --git a/cloudinit/config/schema.py b/cloudinit/config/schema.py
index ffc3dfe91..acab72806 100644
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -884,11 +884,10 @@ def _get_property_description(prop_config: dict) -> str:
         config: dict, descriptions: list, deprecated_descriptions: list
     ):
         if any(
-            [
-                config.get("deprecated_version"),
-                config.get("changed_version"),
-                config.get("new_version"),
-            ]
+            map(
+                config.get,
+                ("deprecated_version", "changed_version", "new_version"),
+            )
         ):
             deprecated_descriptions.append(
                 _add_deprecated_changed_or_new_msg(config)
```

aciba90 (changes requested):
It looks great, thanks @holmanb.

I left some minor comments and suggestions.

Additionally, I saw a possible schema improvement:

```diff
diff --git a/cloudinit/config/schemas/schema-cloud-config-v1.json b/cloudinit/config/schemas/schema-cloud-config-v1.json
index 431bd30f0..f0785d71e 100644
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -2557,7 +2557,7 @@
               "description": "Whether to expire all user passwords such that a password will need to be reset on the user's next login. Default: ``true``"
             },
             "users": {
-              "description": "Replaces the deprecated ``list`` key. This key represents a list of existing users to set passwords for. Each item under users contains the following required keys: ``name`` and ``password`` or in the case of a randomly generated password, ``name`` and ``type``. The ``type`` key has a default value of ``hash``, and may alternatively be set to ``text`` or ``RANDOM``.",
+              "description": "This key represents a list of existing users to set passwords for. Each item under users contains the following required keys: ``name`` and ``password`` or in the case of a randomly generated password, ``name`` and ``type``. The ``type`` key has a default value of ``hash``, and may alternatively be set to ``text`` or ``RANDOM``.",
               "type": "array",
               "items": {
                 "minItems": 1,
@@ -2623,8 +2623,8 @@
               "minItems": 1,
               "description": "List of ``username:password`` pairs. Each user will have the corresponding password set. A password can be randomly generated by specifying ``RANDOM`` or ``R`` as a user's password. A hashed password, created by a tool like ``mkpasswd``, can be specified. A regex (``r'\\$(1|2a|2y|5|6)(\\$.+){2}'``) is used to determine if a password value should be treated as a hash.\n\n",
               "deprecated": true,
-              "deprecated_version": "22.2"
-            }
+              "deprecated_version": "22.2",
+              "deprecated_description": "Use ``users`` instead."
           }
         },
         "password": {
```

cloudinit/config/schema.py
holmanb: I like it

holmanb:
> Additionally, I saw a possible schema improvement:

Much cleaner, agreed.

Thanks for the review @aciba90. I think I've addressed all requested changes.

PR #1998, opened 2023-02-06
oracle-ds: prefer system_cfg over ds network config source (SC-1414)

aciba90:
## Proposed Commit Message

```
oracle-ds: prefer system_cfg over ds network config source

Bump system_cfg over ds network_config_source for Oracle DS,
so that if network config is defined under /etc/cloud,
it will be honored.

In a previous change, we moved the initramfs and system_cfg
bellow ds to favor ds, but this implied system-wide configs
were always not honored.

LP: #1956788
```

## Additional Context

https://bugs.launchpad.net/cloud-init/+bug/1956788
https://warthogs.atlassian.net/browse/SC-1414
https://warthogs.atlassian.net/browse/CPC-1975

## Test Steps

```sh
CLOUD_INIT_PLATFORM="oci" CLOUD_INIT_CLOUD_INIT_SOURCE=$(ls cloud*.deb) tox -e integration-tests -- -vv tests/integration_tests/datasources/test_oci_networking.py::test_oci_networking_system_cfg
```

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #2014, opened 2023-02-10
d/cloud-init.preinst: drop 99-disable-network-config.cfg in Oracle instances (SC-1414)

aciba90:
## Proposed Commit Message

```
d/cloud-init.preinst: drop 99-disable-network-config.cfg in Oracle instances
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-1414
https://bugs.launchpad.net/cloud-init/+bug/1956788

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (22.4.2-0ubuntu4) UNRELEASED; urgency=medium
+
+  * d/cloud-init.preinst: Oracle to remove vestigial /etc/cloud.cloud.cfg.d/
+    99-disable-network-config.cfg because system config is now honored before
+    datasource config (LP: #1956788)
+
+ -- Alberto Contreras <alberto.contreras@canonical.com>  Fri, 17 Feb 2023 11:40:55 +0100
+
 cloud-init (22.4.2-0ubuntu3) lunar; urgency=medium
 
   * Upstream snapshot based on upstream/main at bdc7b040.


aciba90:
After approved, I will replicate this change in b, f, j series.

Per https://github.com/canonical/cloud-init/pull/2009#issuecomment-1421968790, I am assuming we do not need to update `debian/changelog`.

TheRealFalcon:
I could be wrong, but I think the meaning behind that comment was that because Chad was changing something that only ever appeared in devel and never got released, there was no need to update the changelog because there was no meaningful difference to what was already mentioned in the changelog. I think this would need a changelog entry.

blackboxsw (changes requested):
Yes @aciba90, @TheRealFalcon is correct here, This is a debian packaging preinst behavior change that represents a new change in behavior which we will release (and need to account for in d/changelog). The prior PR you referenced was one which already had a line in d/changelog that told of updates to `d/patches/retain-netplan-world-readable.patch` which accounted for the minor quilt patch fixes we needed to actually properly build. Since nothing was built or published to the daily PPA, there was no 'public' d/changelog entry needed to account for a public change in behavior.

In this PR, I think we need something like the following:
```diff
diff --git a/debian/changelog b/debian/changelog
index 535749cbd..0a93c1b36 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (22.4.2-0ubuntu4) UNRELEASED; urgency=medium
+
+  * d/cloud-init.preinst: Oracle to remove vestigial /etc/cloud.cloud.cfg.d/
+    99-disable-network-config.cfg because system config is now honored before
+    datasource config: (LP: #1956788)
+
+ -- Chad Smith <chad.smith@canonical.com>  Mon, 13 Feb 2023 17:27:04 -0700
+
 cloud-init (22.4.2-0ubuntu3) lunar; urgency=medium
 
   * Upstream snapshot based on upstream/main at bdc7b040.
diff --git a/debian/cloud-init.preinst b/debian/cloud-init.preinst
index cb2ffd97c..0dcef7c4f 100644
--- a/debian/cloud-init.preinst
+++ b/debian/cloud-init.preinst
@@ -182,8 +182,8 @@ cleanup_lp1552999() {
 }
 
 cleanup_oci_network_lp1956788() {
-   # Remove a previously ignored and not needed any more file to disable network config
-   # in Oracle instances.
+   # Remove vestigial /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg
+   # from Oracle now that datasource honors system cfg above datasource cfg.
    grep DataSourceOracle /var/lib/cloud/instance/datasource > /dev/null 2>&1 || return 0
    rm -f /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg
 }
```

I think we want to wait to land this  officially until we can validate the related PR #1998.
Let's gate landing this until 1998 is approved/merged.

request changes for d/changelog

TheRealFalcon:
@blackboxsw, actually I think this one needs to land first. If #1998 lands first, then the file being removed in this PR to disable networking config will be honored over the DS config, so networking will be disabled on every launch. As is, this file is basically ignored because of the datasource ordering. Removing this file should have no effect because the datasource config (currently) comes first.

TheRealFalcon:
We should wait until after the 23.1 release to merge this.

blackboxsw:
> We should wait until after the 23.1 release to merge this.
Agreed, I'll finish review on #1998  and we'll just keep the PR unmerged until we clear 23.1 SRU

PR #2016, opened 2023-02-13
refactor: stop passing log instances to cc_* handlers

CalvoM:
## Proposed Commit Message

```
refactor: stop passing log to cc_* handlers

Use the module level Log instances instead of passing log instances
to the cc_* handlers
```

## Checklist:

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -26,8 +26,9 @@ jobs=4
 # W1401(anomalous-backslash-in-string)
 # W1514(unspecified-encoding)
 # E0012(bad-option-value)
+# W1203(logging-fstring-interpolation)
 
-disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401, W1514, E0012
+disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401, W1514, E0012, W1203
 
 
 [REPORTS]
--- a/.pylintrc
+++ b/.pylintrc
@@ -26,9 +26,8 @@ jobs=4
 # W1401(anomalous-backslash-in-string)
 # W1514(unspecified-encoding)
 # E0012(bad-option-value)
-# W1203(logging-fstring-interpolation)
 
-disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401, W1514, E0012, W1203
+disable=C, F, I, R, W0201, W0212, W0221, W0222, W0223, W0231, W0311, W0511, W0602, W0603, W0611, W0613, W0621, W0622, W0631, W0703, W1401, W1514, E0012
 
 
 [REPORTS]
--- a/cloudinit/config/cc_grub_dpkg.py
+++ b/cloudinit/config/cc_grub_dpkg.py
@@ -141,7 +141,6 @@ def is_efi_booted() -> bool:
 
 
 def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
-
     mycfg = cfg.get("grub_dpkg", cfg.get("grub-dpkg", {}))
     if not mycfg:
         mycfg = {}
@@ -153,31 +152,6 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
 
     dconf_sel = get_debconf_config(mycfg)
     LOG.debug("Setting grub debconf-set-selections with '%s'", dconf_sel)
-    idevs = util.get_cfg_option_str(mycfg, "grub-pc/install_devices", None)
-    if idevs is None:
-        idevs = fetch_idevs()
-
-    idevs_empty = mycfg.get("grub-pc/install_devices_empty")
-    if idevs_empty is None:
-        idevs_empty = not idevs
-    elif not isinstance(idevs_empty, bool):
-        idevs_empty = util.translate_bool(idevs_empty)
-    idevs_empty = str(idevs_empty).lower()
-
-    # now idevs and idevs_empty are set to determined values
-    # or, those set by user
-
-    dconf_sel = (
-        "grub-pc grub-pc/install_devices string %s\n"
-        "grub-pc grub-pc/install_devices_empty boolean %s\n"
-        % (idevs, idevs_empty)
-    )
-
-    LOG.debug(
-        "Setting grub debconf-set-selections with '%s','%s'",
-        idevs,
-        idevs_empty,
-    )
 
     try:
         subp.subp(["debconf-set-selections"], dconf_sel)
--- a/cloudinit/config/cc_grub_dpkg.py
+++ b/cloudinit/config/cc_grub_dpkg.py
@@ -8,8 +8,8 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 """Grub Dpkg: Configure grub debconf installation device"""
 
+import logging
 import os
-from logging import Logger
 from textwrap import dedent
 
 from cloudinit import subp, util
@@ -57,9 +57,10 @@ meta: MetaSchema = {
 }
 
 __doc__ = get_meta_doc(meta)
+LOG = logging.getLogger(__name__)
 
 
-def fetch_idevs(log: Logger):
+def fetch_idevs():
     """
     Fetches the /dev/disk/by-id device grub is installed to.
     Falls back to plain disk name if no by-id entry is present.
@@ -71,7 +72,7 @@ def fetch_idevs(log: Logger):
     # EFI mode systems use /boot/efi and the partition path.
     probe_target = "disk"
     probe_mount = "/boot"
-    if is_efi_booted(log):
+    if is_efi_booted():
         probe_target = "device"
         probe_mount = "/boot/efi"
 
@@ -84,18 +85,18 @@ def fetch_idevs(log: Logger):
         # grub-common may not be installed, especially on containers
         # FileNotFoundError is a nested exception of ProcessExecutionError
         if isinstance(e.reason, FileNotFoundError):
-            log.debug("'grub-probe' not found in $PATH")
+            LOG.debug("'grub-probe' not found in $PATH")
         # disks from the container host are present in /proc and /sys
         # which is where grub-probe determines where /boot is.
         # it then checks for existence in /dev, which fails as host disks
         # are not exposed to the container.
         elif "failed to get canonical path" in e.stderr:
-            log.debug("grub-probe 'failed to get canonical path'")
+            LOG.debug("grub-probe 'failed to get canonical path'")
         else:
             # something bad has happened, continue to log the error
             raise
     except Exception:
-        util.logexc(log, "grub-probe failed to execute for grub-dpkg")
+        util.logexc(LOG, "grub-probe failed to execute for grub-dpkg")
 
     if not disk or not os.path.exists(disk):
         # If we failed to detect a disk, we can return early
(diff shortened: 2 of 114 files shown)


igalic (commented):
thank you for unifying this!

blackboxsw:
@CalvoM thank you for bringing this discussion to standup. @aciba90's point is valid here. We do have a couple of customers in the wild with private modules delivered in their custom images.

If we are changing the signature to avoid the unnecessary log parameter in upstream modules, I think we agreed we still need to support older signatures which still require the `log` parameter.

I think the most likely place is as Alberto mentioned changing [the func_args param](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/modules.py#L234)  Most, likely I think we may want to dry to use a dict kwargs format with named params and values because the gets expanded when calling the 'handle' functor in [cloudinit.helper.Runners.run](https://github.com/canonical/cloud-init/blob/main/cloudinit/helpers.py#L182-L183).

Ideally, we'd try/except around the call site in config/modules.py#L246, but there may be some unfriendly side-effects due to our `sem.lock` context manager possibly creating a semaphore. If we fail due to function signature requiring a madatory `log` param, we may not be able to re-enter into `run` function again because the semaphore file already exists.

I'll let you look into this a bit more to see what's a viable option.

aciba90:
Another approach could be to [dynamically inspect](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object) the `handle`'s signature and decide how many and what arguments to pass in.

CalvoM:
@aciba90 Thanks for this suggestion, from a general skimming, it might help with the sem lock issue.

.pylintrc
```diff
@@ -26,8 +26,9 @@ jobs=4
 # W1401(anomalous-backslash-in-string)
 # W1514(unspecified-encoding)
 # E0012(bad-option-value)
+# W1203(logging-fstring-interpolation)
```
aciba90, line 29: I think it makes sense to not disable this warning as:

1. We can save some cycles by taking advantage of the [lazy string formatting optimization](https://docs.python.org/3/howto/logging.html#optimization).
2. Keep it simple and consistent.

cloudinit/config/modules.py
```diff
@@ -241,8 +237,18 @@ def _run_modules(self, mostly_mods: List[ModuleDetails]):
...
+                    "cfg": self.cfg,
+                    "cloud": cc,
+                    "args": args,
+                }
 
                 with myrep:
+                    func_signature = signature(mod.handle)
+                    func_params = func_signature.parameters
+                    if len(func_params) == 5:
+                        func_args.update({"log": LOG})
```
aciba90, line 251: I think we must warn users about the fact that we will drop support for modules with the log parameter. Something in the lines of:

```suggestion
                        LOG.warning("DEPRECATED: The use of modules with a `log` parameter is deprecated.")
                        func_args.update({"log": LOG})
```

aciba90 (changes requested):
It looks very good, thanks @CalvoM.

I have left a couple of concerns. Additionally, I think we should minimally unit-test the changes made to `cloudinit/config/modules.py`.

cloudinit/config/modules.py
aciba90: After 6100fda632f63605981636c0900e1e0b8b354979 being merged, I think we might want to use the standard machinery to deprecated things.

aciba90 (changes requested):
Thanks, @CalvoM, for addressing the comments and adding unit testing.

I attach a diff that uses a new standard function to deprecate things, included in 2ee614aa6431f4339f48ebabdf2de4ebcc2401ab.

Additionally, I have reworked the unit test to remove the need to have an extra module and add some extra checks. If you agree with the approach, feel free to modify it.

```diff
diff --git a/cloudinit/config/modules.py b/cloudinit/config/modules.py
index 75efd151e..89be21be0 100644
--- a/cloudinit/config/modules.py
+++ b/cloudinit/config/modules.py
@@ -248,9 +248,9 @@ class Modules:
                     func_signature = signature(mod.handle)
                     func_params = func_signature.parameters
                     if len(func_params) == 5:
-                        LOG.warning(
-                            "DEPRECATED: The use of modules with a `log` "
-                            "parameter is depecrated"
+                        util.deprecate(
+                            deprecated="Config modules with a `log` parameter",
+                            deprecated_version="23.2",
                         )
                         func_args.update({"log": LOG})
                     ran, _r = cc.run(
diff --git a/tests/unittests/config/test_modules.py b/tests/unittests/config/test_modules.py
index aea205a3f..7980b2013 100644
--- a/tests/unittests/config/test_modules.py
+++ b/tests/unittests/config/test_modules.py
@@ -2,15 +2,15 @@
 
 
 import importlib
+import inspect
 import logging
-from inspect import signature
 from pathlib import Path
 from typing import List
 
 import pytest
 
 from cloudinit import util
-from cloudinit.config.modules import ModuleDetails, Modules, _is_active
+from cloudinit.config.modules import Init, ModuleDetails, Modules, _is_active
 from cloudinit.config.schema import MetaSchema
 from cloudinit.distros import ALL_DISTROS
 from cloudinit.settings import FREQUENCIES
@@ -22,10 +22,7 @@ M_PATH = "cloudinit.config.modules."
 def get_module_names() -> List[str]:
     """Return list of module names in cloudinit/config"""
     files = list(
-        Path(cloud_init_project_dir("cloudinit/config/")).glob("cc_*.py"),
-    )
-    files.append(
-        Path(cloud_init_project_dir("tests/unittests/config/mock_cc_ssh.py"))
+        Path(cloud_init_project_dir("cloudinit/config/")).glob("cc_*.py")
     )
 
     return [mod.stem for mod in files]
@@ -34,11 +31,7 @@ def get_module_names() -> List[str]:
 def get_modules():
     examples = []
     for mod_name in get_module_names():
-        mod_path = f"cloudinit.config.{mod_name}"
-        # TODO: We could make it generic if needed ie mock_cc*
-        if "mock_cc_ssh" == mod_name:
-            mod_path = "tests.unittests.config.mock_cc_ssh"
-        module = importlib.import_module(mod_path)
+        module = importlib.import_module(f"cloudinit.config.{mod_name}")
         for i, example in enumerate(module.meta.get("examples", [])):
             examples.append(
                 pytest.param(
@@ -181,35 +174,56 @@ class TestModules:
         ] == m_run_modules.call_args_list
         assert "Skipping" not in caplog.text
 
-    @pytest.mark.parametrize("mod_name, module, example", get_modules())
+    @mock.patch(M_PATH + "signature")
     @mock.patch("cloudinit.config.modules.ReportEventStack")
-    def test_args_in_handle(
-        self, event, mod_name, module, example, caplog, mocker
+    def test_module_with_old_handle_signature(
+        self, event, m_signature, caplog
     ):
-        def run_side_effect(name, handle, func_args, freq):
-            return (1, 2)
+        """Asserts that a module with an old handle signature receives
+        its arguments and a deprecation msg is issued.
+        """
+
+        def handle(name, cfg, cloud, log, args):
+            pass
+
+        m_signature.return_value = inspect.signature(handle)
+        module = mock.Mock()
+        module.handle.side_effect = handle
 
-        init_mock = mock.Mock()
-        # init_mock.cloudify = mock.Mock()
-        # init_mock.cloudify.run = mock.Mock(side_effect=run_side_effect)
         mods = Modules(
-            init=init_mock,
+            init=mock.Mock(spec=Init),
             cfg_files=mock.Mock(),
             reporter=mock.Mock(),
         )
         mods._cached_cfg = {}
         module_details = ModuleDetails(
             module=module,
-            name=mod_name,
+            name="mod_name",
             frequency=["always"],
             run_args=[],
         )
-        with mock.patch.object(init_mock, "cloudify") as m_cloudify:
-            m_cc = mock.Mock()
-            m_cloudify.return_value = m_cc
-            m_cc.run.side_effect = run_side_effect
-            mods._run_modules([module_details])
-            func_args_len = len(m_cc.run.call_args_list[0][0][2])
-            func_signature = signature(module.handle)
-            expected_len = len(func_signature.parameters)
-            assert func_args_len == expected_len
+
+        m_cc = mods.init.cloudify.return_value
+        m_cc.run.return_value = (1, "doesnotmatter")
+
+        mods._run_modules([module_details])
+
+        assert [
+            mock.call(
+                mock.ANY,
+                mock.ANY,
+                {
+                    "name": "mod_name",
+                    "cfg": {},
+                    "cloud": mock.ANY,
+                    "args": [],
+                    "log": mock.ANY,
+                },
+                freq=["always"],
+            )
+        ] == m_cc.run.call_args_list
+
+        assert (
+            "Config modules with a `log` parameter is deprecated in 23.2"
+            in caplog.text
+        )

```

aciba90 (changes requested):
We must also update https://github.com/canonical/cloud-init/blob/main/doc/rtd/development/module_creation.rst reflecting the new `handle`'s signature and probably adding a deprecation warning about the old one.

doc/rtd/development/module_creation.rst
```diff
@@ -128,3 +128,65 @@ in the ``cloud_final_modules`` section before the ``final-message`` module.
 .. _cloud_init_modules: https://github.com/canonical/cloud-init/blob/b4746b6aed7660510071395e70b2d6233fbdc3ab/config/cloud.cfg.tmpl#L70
 .. _cloud_config_modules: https://github.com/canonical/cloud-init/blob/b4746b6aed7660510071395e70b2d6233fbdc3ab/config/cloud.cfg.tmpl#L101
 .. _cloud_final_modules: https://github.com/canonical/cloud-init/blob/b4746b6aed7660510071395e70b2d6233fbdc3ab/config/cloud.cfg.tmpl#L144
+
```
aciba90, line 131: Taking into account:

1. The old signature does not offer any advantage over the new one.
2. Existing external modules already know the old signature.
 
Maybe it is better to not advertise the old signature and remove this whole block from the docs.

Thoughts?

aciba90 (changes requested):

doc/rtd/development/module_creation.rst
CalvoM: Since we support it in the code, I thought it would just put a note on it. I could remove the example, and just leave a note for anyone with questions; when we later drop the support, we can remove the documentation as well.

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -113,73 +114,97 @@ def fetch_idevs(log: Logger):
...
 
     enabled = mycfg.get("enabled", True)
     if util.is_false(enabled):
-        log.debug("%s disabled by config grub_dpkg/enabled=%s", name, enabled)
+        LOG.debug("%s disabled by config grub_dpkg/enabled=%s", name, enabled)
         return
 
-    dconf_sel = get_debconf_config(mycfg, log)
-    log.debug("Setting grub debconf-set-selections with '%s'" % dconf_sel)
+    dconf_sel = get_debconf_config(mycfg)
```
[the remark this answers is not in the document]
CalvoM: Yes, I ran into alot of rebase issues. Let me check this one as well.

doc/rtd/development/module_creation.rst
aciba90: Yes, my way of reasoning is: as the log parameter is a feature that we don't want users to use and they can see the documentation of older version, we should probably not document it and simplify the docs.

I think the example should be removed and I'll leave it open to you to decide if a note should be included.

TheRealFalcon: +1 to removing the log parameter from the docs for the reasons @aciba90 mentioned.

CalvoM: Updated.

doc/rtd/development/module_creation.rst
```diff
@@ -28,6 +27,8 @@ Example
 
     This will likely take multiple lines.
     """
+
```
aciba90, line 30: There are some trailing whitespaces that Sphinx does not like, (`tox -e doc` to reproduce it):

```suggestion

```

aciba90 (changes requested):
Thanks for the changes, apart from the linting error, LGTM.

PR #2017, opened 2023-02-13
Fix minor schema validation regression and fixup typing

TheRealFalcon:
~~Changing to WIP as I just noticed https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-bionic-azure/192/testReport/junit/tests.integration_tests.modules/test_cli/test_valid_userdata/ , and this can probably also be rolled in here.~~

## Proposed Commit Message

```
Fix minor schema validation regression and fixup typing

When annotating a schema, we should print "Valid cloud-config" along
with the filename if a file has been used rather than just the config
type.
```

## Additional Context
Fixes https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-bionic-azure/192/testReport/junit/tests.integration_tests.cmd.test_schema/TestSchemaDeprecations/test_schema_deprecations/

Also, pyright was complaining because `config_files` could not be set, so I changed the `elif` to and `else` (we have a guard further up). Also, `get_ipath` can return `None`, so I added checks for that as well.

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -1231,6 +1231,7 @@ def handle_schema_args(name, args):
         error(
             ", ".join(error_type for error_type in error_types),
             fmt="Error: Invalid cloud-config schema: {}\n",
+            sys_exit=True,
         )
 
 
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -1166,7 +1166,7 @@ def handle_schema_args(name, args):
         return
     if args.config_file:
         config_files = (("user-data", args.config_file),)
-    elif args.system:
+    else:
         if os.getuid() != 0:
             error(
                 "Unable to read system userdata or vendordata as non-root"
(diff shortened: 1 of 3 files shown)


blackboxsw (changes requested):
Looks good thanks @TheRealFalcon. 
Unrelated to this PR, but there was an omission in my previous branch that doesn't exit 1.
We also need the following patch (which is already validated and [fails integration tests here](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_cli.py#L59)

```diff
diff --git a/cloudinit/config/schema.py b/cloudinit/config/schema.py
index de025c64f..dcaf13afc 100644
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -1231,6 +1231,7 @@ def handle_schema_args(name, args):
         error(
             ", ".join(error_type for error_type in error_types),
             fmt="Error: Invalid cloud-config schema: {}\n",
+            sys_exit=True,
         )
 
 ```

blackboxsw:
With the above changes: ================== 3 passed, 2 warnings in 159.49s (0:02:39) ===================

blackboxsw (approved):

PR #2024, opened 2023-02-17
test: switch pycloudlib source to pypi

holmanb:
```
test: switch pycloudlib source to pypi
```

## Additional Context
This PR tests this change with CI.


TheRealFalcon:
I'm thinking it might make more sense to pin minor or even major version here as worst case scenario somebody adding a bug causing some jenkins tests fail.

TheRealFalcon (approved):

PR #2026, opened 2023-02-19
log: Add standardized deprecation tooling (SC-1312)

holmanb:
```
log: Add standardized deprecation tooling

- Add deprecation log level
- Add deprecation utilities for structured format and messaging
- Update existing deprecation log sites, add deprecated versions
```

## Additional Context
Barring any objections to the approach, dedup and integration tests will follow.

This change allows us to consistently make clear deprecation expectations with specific versions in all cases. This messaging is required to ensure community members have sufficient time to react to impending change - to either update their configs or make a case for the necessity of the deprecated feature.

### DIFF
--- a/cloudinit/log.py
+++ b/cloudinit/log.py
@@ -63,6 +63,16 @@ def flushLoggers(root):
     flushLoggers(root.parent)
 
 
+def defineDeprecationLogger(lvl=25):
+    logging.addLevelName(lvl, "DEPRECATED")
+
+    def deprecated(self, message, *args, **kwargs):
+        if self.isEnabledFor(lvl):
+            self._log(lvl, message, args, **kwargs)
+
+    logging.Logger.deprecated = deprecated
+
+
 def setupLogging(cfg=None):
     # See if the config provides any logging conf...
     if not cfg:
@@ -83,6 +93,7 @@ def setupLogging(cfg=None):
                 log_cfgs.append("\n".join(cfg_str))
             else:
                 log_cfgs.append(str(a_cfg))
+    defineDeprecationLogger()
 
     # See if any of them actually load...
     am_tried = 0
--- a/cloudinit/log.py
+++ b/cloudinit/log.py
@@ -63,7 +63,7 @@ def flushLoggers(root):
     flushLoggers(root.parent)
 
 
-def defineDeprecationLogger(lvl=25):
+def defineDeprecationLogger(lvl=35):
     logging.addLevelName(lvl, "DEPRECATED")
 
     def deprecated(self, message, *args, **kwargs):
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3111,7 +3111,7 @@ def deprecate(
     logged.
 
     @param deprecated: Noun to be deprecated. Write this as the start
-        of a sentance, with no period. Version and extra message will
+        of a sentence, with no period. Version and extra message will
         be appended.
     @param deprecated_version: The version in which the thing was
         deprecated
@@ -3138,7 +3138,10 @@ def deprecate(
             f"{deprecated_version} and scheduled to be removed in "
             f"{version_removed}. {message}"
         ).rstrip()
-        LOG.deprecated(deprecate_msg)
+        if hasattr(LOG, "deprecated"):
+            LOG.deprecated(deprecate_msg)
+        else:
+            LOG.warning(deprecate_msg)
 
 
 def deprecate_call(
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3126,13 +3126,13 @@ def deprecate(
     Note: uses keyword-only arguments to improve legibility
     """
     if not hasattr(deprecate, "_log"):
-        deprecate._log = set()
+        deprecate._log = set()  # type: ignore
     message = extra_message or ""
     dedup = hash(deprecated + message + deprecated_version + str(schedule))
     version = Version.from_str(deprecated_version)
     version_removed = Version(version.major + schedule, version.minor)
-    if dedup not in deprecate._log:
-        deprecate._log.add(dedup)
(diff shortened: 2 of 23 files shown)


cloudinit/util.py
```diff
@@ -3061,3 +3063,78 @@ def _compare_version(self, other) -> int:
         if self.rev > other.rev:
             return 1
         return -1
+
+
+def deprecate(
```
aciba90, line 3068: What do you think about the following diff?

The intention is to remove module global state and make this function more modular:

```diff
diff --git a/cloudinit/util.py b/cloudinit/util.py
index 8cd28e642..23fbe67e0 100644
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -54,7 +54,6 @@ from cloudinit.settings import CFG_BUILTIN

 _DNS_REDIRECT_IP = None
 LOG = logging.getLogger(__name__)
-DEPRECATION_LOG = set()

 # Helps cleanup filenames to ensure they aren't FS incompatible
 FN_REPLACEMENTS = {
@@ -3090,18 +3089,20 @@ def deprecate(

     Note: uses keyword-only arguments to improve legibility
     """
+    if not hasattr(deprecate, "_log"):
+        deprecate._log = set()
     message = extra_message or ""
     dedup = hash(deprecated + message + deprecated_version + str(schedule))
     version = Version.from_str(deprecated_version)
     version_removed = Version(version.major + schedule, version.minor)
-    if dedup not in DEPRECATION_LOG:
-        DEPRECATION_LOG.add(dedup)
-        deprecate = (
+    if dedup not in deprecate._log:
+        deprecate._log.add(dedup)
+        deprecate_msg = (
             f"{deprecated} is deprecated in "
             f"{deprecated_version} and scheduled to be removed in "
             f"{version_removed}. {message}"
         ).rstrip()
-        LOG.deprecated(deprecate)
+        LOG.deprecated(deprecate_msg)


 def deprecate_call(
```

cloudinit/util.py
```diff
@@ -3061,3 +3063,78 @@ def _compare_version(self, other) -> int:
...
+    version = Version.from_str(deprecated_version)
+    version_removed = Version(version.major + schedule, version.minor)
+    if dedup not in DEPRECATION_LOG:
+        DEPRECATION_LOG.add(dedup)
+        deprecate = (
+            f"{deprecated} is deprecated in "
+            f"{deprecated_version} and scheduled to be removed in "
+            f"{version_removed}. {message}"
+        ).rstrip()
+        LOG.deprecated(deprecate)
```
aciba90, line 3104: This is call ties this function to a call to [log.setupLogging](https://github.com/canonical/cloud-init/pull/2026/files#diff-ddaf8784109f41cf2bd02421e7f7a5f37d9bdef3bed9c6bb866b3c5fc116f123R96), as that custom function will only be present if that call is executed.

I know this is not a common situation, but it is something we should maintain, as for example [ua](https://github.com/canonical/ubuntu-advantage-client/blob/9b46480b9e4b88e918bac5ced0d4b8edb3cbbeab/lib/auto_attach.py#L35) uses our code as a library, without setting up the log.

They will get a: `AttributeError: 'Logger' object has no attribute 'deprecated'`.

Another place where we can see this dependency in action is the unit test framework. Depending on the test execution order, we could see the previous error surfacing. For example, if we execute `tox -e py3 -- tests/unittests/test_net.py`.

What about checking for the existence of the new method and fallback to a normal logging mode if not present? Something like:

```suggestion
        if hasattr(LOG, "deprecated"):
            LOG.deprecated(deprecate)
        else:
            LOG.debug(deprecate)
```

aciba90 (changes requested):
Overall looks good, thanks @holmanb.

I left a nit comment and one concern about the `logger.deprecated` function set up.

cloudinit/util.py
holmanb: > I know this is not a common sit[ua](https://github.com/canonical/ubuntu-advantage-client/blob/9b46480b9e4b88e918bac5ced0d4b8edb3cbbeab/lib/auto_attach.py#L35)tion, but it is something we should maintain, as for example ua uses our code as a library, without setting up the log.

Good point, we shouldn't break them. Updated accordingly.

holmanb: Off topic: note the github UA bug seen in the above pasted quote. The "ua" link got shifted to the first matching text in the string rather than matching the text that it originally quoted. Copy and paste for yourself and see if it is reproducible for you (it is for me) :)

cloudinit/log.py
```diff
@@ -63,6 +63,16 @@ def flushLoggers(root):
     flushLoggers(root.parent)
 
 
+def defineDeprecationLogger(lvl=25):
+    logging.addLevelName(lvl, "DEPRECATED")
```
aciba90, line 67: Correct me if I am wrong, but I think we do not refer to this level by the name anywhere.

Taking into account that and https://docs.python.org/3/howto/logging.html#custom-levels I would suggest removing this line.

aciba90 (approved):
LGTM!

Thanks, @holmanb. Very nice deprecate function and decorator!

cloudinit/log.py
holmanb: Since this string is printed in the logs, removal of it causes `DEPRECATED` to be replaced with the much less useful `Level 25`, so I lean towards keeping it as is.

If desired I can perhaps add to the documentation a section about using cloud-init as a library, to communicate to users logging expectations.

holmanb: @aciba90 I just updated this PR with updated docs per this comment, as well as some extra tests. Requesting re-review.

TheRealFalcon: We talked it over a bit and I'm ok with these changes. Since cloud-init isn't _really_ used as a library in too many places, I don't think we need to worry about conflicts.

aciba90 (approved):
LGTM, thanks.

PR #2027, opened 2023-02-21
sources/vmware/imc: fix-missing-catch-few-negtive-scenarios

PengpengSun:
## Proposed Commit Message

```
summary: sources/vmware/imc: fix-missing-catch-few-negtive-scenarios

1. When metadata is invalid, should catch exception and report imc failure event.
2. When pre/post customization script is invalid, should catch exception and report imc failure event.

```

## Checklist:

 - [] I have updated or added any unit tests accordingly

### DIFF
--- a/cloudinit/sources/helpers/vmware/imc/guestcust_util.py
+++ b/cloudinit/sources/helpers/vmware/imc/guestcust_util.py
@@ -1,5 +1,5 @@
 # Copyright (C) 2016 Canonical Ltd.
-# Copyright (C) 2016-2022 VMware Inc.
+# Copyright (C) 2016-2023 VMware Inc.
 #
 # Author: Sankar Tanguturi <stanguturi@vmware.com>
 #         Pengpeng Sun <pegnpengs@vmware.com>
@@ -11,7 +11,7 @@ import os
 import re
 import time
 
-from cloudinit import subp, util
+from cloudinit import safeyaml, subp, util
 
 from .config import Config
 from .config_custom_script import (
@@ -264,6 +264,17 @@ def get_data_from_imc_raw_data_cust_cfg(cust_cfg):
                 cust_cfg,
             )
             return (None, None, None)
+
+        try:
+            logger.debug("Safeyaml loads md tocheck if it's valid")
+            md = safeyaml.load(md)
+        except safeyaml.YAMLError as e:
+            set_cust_error_status(
+                "Error parsing the cloud-init meta data",
+                str(e),
+                GuestCustErrorEnum.GUESTCUST_ERROR_WRONG_META_FORMAT,
+                cust_cfg,
+            )
 
         ud_file = cust_cfg.user_data_name
         if ud_file:
@@ -520,6 +531,14 @@ def do_pre_custom_script(cust_cfg, custom_script, cust_cfg_dir):
             cust_cfg,
         )
         return False
+    except Exception as e:
+        set_cust_error_status(
+            "Error executing pre-customization script",
+            str(e),
+            GuestCustEventEnum.GUESTCUST_EVENT_CUSTOMIZE_FAILED,
+            cust_cfg,
+        )
+        return False
     return True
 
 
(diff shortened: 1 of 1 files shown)


cloudinit/sources/helpers/vmware/imc/guestcust_util.py
```diff
@@ -265,6 +265,17 @@ def get_data_from_imc_raw_data_cust_cfg(cust_cfg):
             )
             return (None, None, None)
 
+        try:
+            logger.debug("Safeyaml loads md tocheck if it's valid")
```
igalic, line 269: usually you would log something like: "Validating meta data"

igalic (changes requested):
the format of the pull request description / commit message is broken

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
PengpengSun: And corrected the Proposed Commit Message format.

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
```diff
@@ -265,6 +265,17 @@ def get_data_from_imc_raw_data_cust_cfg(cust_cfg):
             )
             return (None, None, None)
 
+        try:
+            logger.debug("Validating if meta data is valid or not")
+            md = safeyaml.load(md)
```
TheRealFalcon, line 270: This is translating a string to some other structure...likely a dict or list. This means the return type of this function will be different than before. Is that what you're expecting?

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
```diff
@@ -520,6 +531,14 @@ def do_pre_custom_script(cust_cfg, custom_script, cust_cfg_dir):
             cust_cfg,
         )
         return False
+    except Exception as e:
```
TheRealFalcon, line 534: Since this is a copy of the `except` block above and `CustomScriptNotFound` is a subclass of exception, we should be just able to replace `CustomScriptNotFound` with `Exception`.

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
```diff
@@ -535,6 +554,14 @@ def do_post_custom_script(cust_cfg, custom_script, cust_cfg_dir, ccScriptsDir):
...
```
TheRealFalcon, line 557: Since this is a copy of the `except` block above and `CustomScriptNotFound` is a subclass of exception, we should be just able to replace `CustomScriptNotFound` with `Exception`.

TheRealFalcon (changes requested):
Left some inline comments/questions.

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
PengpengSun: Many thanks for the review. Yes, this is expected.  Metadata got from different transports in DS VMware will be loaded at https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceVMware.py#L626,  in this load_json_or_yaml() function, it will return data directly if it's already a dict.  Now this change is to validate the metadata from imc transport early so that we get a chance to report customization error status to vSphere, this is the same behavior when imc data transport was in DS OVF.

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
PengpengSun: Thanks, code updated with only replacing `CustomScriptNotFound` with `Exception`

cloudinit/sources/helpers/vmware/imc/guestcust_util.py
PengpengSun: Ditto.

TheRealFalcon (approved):
Thanks!

PR #2029, opened 2023-02-22
cc_grub_dpkg: Added UEFI support

BirknerAlex:
## Proposed Commit Message

```
summary: Fixed cc_grub_dpkg on UEFI systems

On Debian and Ubuntu based systems the cc_grub_dpkg module
handles the needed change of the disk device name / path between
the pre created image and the real hardware system.

Currently it seems only BIOS mode is supported. This adds UEFI support as
well to change the configuration keys for UEFI.
```

## Test Steps
- Unit tests fixed
- Tested on different bare metal UEFI machines
- Tested on a legacy BIOS machine

## Checklist:


BirknerAlex:
@holmanb Is there any chance to get this inside the upcoming version or too late? This bricks UEFI systems after a new `grub-efi-amd64-signed` package has been released and apt-get tries to upgrade it.

BirknerAlex:
I just found another issue with NVME disks. While SATA works fine the NVME and UEFI combination seems to fail. Fix is coming.

holmanb:
> @holmanb Is there any chance to get this inside the upcoming version or too late?

It is too late, but if something is truly broken we can re-release / SRU.

> This bricks UEFI systems after a new grub-efi-amd64-signed package has been released and apt-get tries to upgrade it.

By this do you mean that cloud-init is rendering the hardware unusable? So re-install fails after this occurs?

BirknerAlex:
@holmanb If it's an EFI machine, the BIOS `grub-pc/install_devices` keys are definitely wrong here, cloud-init itself does not break the machine configuration (it's already broken before booting the first time). But cloud-init creates unused config entries (which are intended for BIOS mode). 

So `cc_grub_dpkg` should fix the debconf settings but can't since it has no UEFI support and it seems the way how apt-get requires these settings are different between BIOS and UEFI. 

But maybe I am completely wrong and I misunderstood the whole thing here, at least with this change the errors disappearing on a fresh install in my case.

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -130,35 +151,55 @@ def handle(
...
+            idevs_empty = util.translate_bool(idevs_empty)
+        idevs_empty = str(idevs_empty).lower()
+
+        # now idevs and idevs_empty are set to determined values
+        # or, those set by user
+        return (
+            "grub-pc grub-pc/install_devices string %s\n"
+            "grub-pc grub-pc/install_devices_empty boolean %s\n"
+            % (idevs, idevs_empty)
+        )
```
holmanb, line 203: While we're making changes here, lets drop the footer, we don't need those.
```
# vi: ts=4 expandtab
```

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -130,35 +151,55 @@ def handle(
         log.debug("%s disabled by config grub_dpkg/enabled=%s", name, enabled)
         return
 
-    idevs = util.get_cfg_option_str(mycfg, "grub-pc/install_devices", None)
-    if idevs is None:
-        idevs = fetch_idevs(log)
+    dconf_sel = get_debconf_config(mycfg, log)
+    if dconf_sel is None:
```
holmanb, line 155: I don't see a way that this path could be hit. In `get_debconf_config()`, both branches return a truthy string.

We can remove this check, I think.

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -130,35 +151,55 @@ def handle(
...
-    idevs_empty = str(idevs_empty).lower()
+    log.debug("Setting grub debconf-set-selections with '%s'" % dconf_sel)
 
-    # now idevs and idevs_empty are set to determined values
-    # or, those set by user
+    try:
+        subp.subp(["debconf-set-selections"], dconf_sel)
+    except Exception as e:
+        util.logexc(
+            log, "Failed to run debconf-set-selections for grub-dpkg: %s", e
```
holmanb, line 168: Since grub-dpkg is deprecated in favor of grub_dpkg, can we change this error to reflect either the preferred key, or whichever was actually passed?

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -130,35 +151,55 @@ def handle(
...
-        "grub-pc grub-pc/install_devices string %s\n"
-        "grub-pc grub-pc/install_devices_empty boolean %s\n"
-        % (idevs, idevs_empty)
-    )
 
-    log.debug(
-        "Setting grub debconf-set-selections with '%s','%s'"
-        % (idevs, idevs_empty)
-    )
+# Returns the debconf config for grub-pc or grub-efi depending on the
```
holmanb, line 172: Can we make this comment a function docstring?

cloudinit/config/cc_grub_dpkg.py
```diff
@@ -117,10 +129,19 @@ def fetch_idevs(log):
     return idevs
 
 
+# Check if the system is booted in EFI mode.
```
holmanb, line 132: Can we make this a function docstring instead please?

holmanb (approved):
Thanks for putting this together @BirknerAlex, overall this looks really good.

I reproduced this issue and verified the solution today in a VM with OVMF.

I left a few minor comments inline. Once they're fixed, I'll merge.

BirknerAlex:
@holmanb Thank you for the review. I am really happy that you were able to reproduce the issue on UEFI.

I did fix all your comments and I'm thankful tor the feedback, let me know if you find anything else.

PR #2020, opened 2023-02-16
Add Apport general hook

bdrung:
## Proposed Commit Message

```
Add Apport general hook

Apport 2.25.0-0ubuntu1 removed the buggy `add_cloud_info` from the
ubuntu general hook. Add an Apport general hook that reads
`/run/cloud-init/instance-data.json` if present.

LP: #1724623
Thanks for the initial version: John Chittum
```

## Checklist:

 - [ ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #2030, opened 2023-02-22
apport: only prompt for cloud_name when instance-data.json is absent

blackboxsw:
Do not prompt for CloudName when instance-data.json exists.

When instance-data.json exists, general-hooks/cloud_init.py will add the following fields to bug reports:
   CloudName, CloudID, CloudPlatform and CloudSubplatform.

## Proposed Commit Message

```
Do not prompt for CloudName when instance-data.json exists.

When instance-data.json exists, general-hooks/cloud_init.py will add the
following fields to bug reports:
   CloudName, CloudID, CloudPlatform and CloudSubplatform.
```

## Checklist:

### DIFF
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -4,6 +4,7 @@
 
 """Cloud-init apport interface"""
 
+import json
 import os
 
 from cloudinit.cmd.devel import read_cfg_paths
@@ -72,6 +73,15 @@ def _get_user_data_file() -> str:
     return paths.get_ipath_cur("userdata_raw")
 
 
+def _get_instance_data() -> dict:
+    paths = read_cfg_paths()
+    try:
+        with open(paths.get_runpath("instance_data")) as file:
+            return json.load(file)
+    except IOError:
+        return {}
+
+
 def attach_cloud_init_logs(report, ui=None):
     """Attach cloud-init logs and tarfile from 'cloud-init collect-logs'."""
     attach_root_command_outputs(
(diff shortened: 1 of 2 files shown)


cloudinit/apport.py
```diff
@@ -101,8 +102,25 @@ def attach_hwinfo(report, ui=None):
 
 
 def attach_cloud_info(report, ui=None):
-    """Prompt for cloud details if available."""
+    """Prompt for cloud details if instance-data unavailable.
+
+    When we have valid _get_instance_data, apport/generic-hooks/cloud_init.py
```
TheRealFalcon, line 107: Where does this `apport/generic_hooks/cloud_init.py` live?

blackboxsw: This is delivered by our downstream packaging branches per the (yet to be landed) https://github.com/canonical/cloud-init/pull/2020 from @bdrung . We'll reflect the generic hooks delivery into each ubuntu/* packaging branch. Since apport is only ubuntu, we don't really have to cope with this in other distributions. Ideally, I'd like to move even this apport integration down into packaging branches too. But, not for this PR

blackboxsw: Also note, we will reflect #2020 to each ubuntu/(bionic|focal|jammy|kinetic) once this PR and 2020 land

TheRealFalcon: Makes sense. I think that could be a little more clear in the docstring (or comment near the change) as it's a little strange having a function obtain a value and then not do anything with it.

TheRealFalcon (approved):
Would still prefer some clarity in the docstring but otherwise LGTM.

PR #2033, opened 2023-02-23
Adapt DataSourceScaleway to IPv6

karibou:
## Adapt DataSourceScaleway to IPv6

```
summary: Adapt DataSourceScaleway to upcoming IPv6 support

Add support for IPv4 and/or IPv6 provided metadata. Upcoming
features will allow for Instances to be started with only IPv6
connectivity. The datasource must be able to fetch its metadata from
the following endpoints : 

* IPv4 on http://169.264.42.42/conf
* IPv6 on http://[fd00:42::42]/conf

This URL may eventually be overridden in the future by a DNS
resolvable URL defined in /etc/cloud/config.d/scaleway.conf

Add support for the configuration of one or many IP addresses when
provided by the `public_ips` metadata key.  In such configurations,
instances no longer have a `private_ip`. The previous configuration
method is kept for backward compatibility.

This commit also fixes a bug in the definition of ssh keys from tags.

```

## Test Steps
Utilization of this DataSource in the Scaleway infrastructure will
produce identical results as with the previous DataSource for
Instances with IPv4 addresses. As the new functionalities get
rolled out in Scaleway's infrastructure, creation of instances with
IPv6 only addresses will allow for cloud-init to continue to 
function as expected.

## Checklist:
 - N/A : I have updated or added any documentation accordingly

### DIFF
--- a/cloudinit/sources/DataSourceScaleway.py
+++ b/cloudinit/sources/DataSourceScaleway.py
@@ -9,8 +9,10 @@ import json
 import os
 import socket
 import time
+from urllib.parse import urlparse
 
 import requests
+from requests.exceptions import ConnectionError
 
 # Note: `urllib3` is transitively installed by `requests`
 from urllib3.connection import HTTPConnection
@@ -21,20 +23,16 @@ from cloudinit import log as logging
 from cloudinit import net, sources, url_helper, util
 from cloudinit.event import EventScope, EventType
 from cloudinit.net.dhcp import NoDHCPLeaseError
-from cloudinit.net.ephemeral import EphemeralDHCPv4
+from cloudinit.net.ephemeral import EphemeralDHCPv4, EphemeralIPv6Network
 from cloudinit.sources import DataSourceHostname
+from cloudinit.subp import ProcessExecutionError
 
 LOG = logging.getLogger(__name__)
 
-DS_BASE_URL = "http://169.254.42.42"
+DS_BASE_URLS = ["http://169.254.42.42", "http://[fd00:42::42]"]
 
-BUILTIN_DS_CONFIG = {
-    "metadata_url": DS_BASE_URL + "/conf?format=json",
-    "userdata_url": DS_BASE_URL + "/user_data/cloud-init",
-    "vendordata_url": DS_BASE_URL + "/vendor_data/cloud-init",
-}
-
-DEF_MD_RETRIES = 5
+DEF_MD_RETRIES = 3
+DEF_MD_ALIVE = 2
 DEF_MD_TIMEOUT = 10
 
 
@@ -117,9 +115,20 @@ def query_data_api(api_type, api_address, retries, timeout):
                 "Trying to get %s data (bind on port %d)...", api_type, port
             )
             requests_session = requests.Session()
+            # Adapt Session.mount to IPv4/IPv6 context
+            localhost = "0.0.0.0"
+            try:
+                url_address = urlparse(api_address).netloc
+                address = url_address
+                if url_address[0] == "[":
+                    address = url_address[1:-1]
+                if net.is_ipv6_address(address):
+                    localhost = "0::"
+            except ValueError:
+                pass
             requests_session.mount(
                 "http://",
-                SourceAddressAdapter(source_address=("0.0.0.0", port)),
+                SourceAddressAdapter(source_address=(localhost, port)),
             )
             data = query_data_api_once(
                 api_address, timeout=timeout, requests_session=requests_session
@@ -154,30 +163,64 @@ class DataSourceScaleway(sources.DataSource):
         self.ds_cfg = util.mergemanydict(
             [
                 util.get_cfg_by_path(sys_cfg, ["datasource", "Scaleway"], {}),
-                BUILTIN_DS_CONFIG,
             ]
         )
 
-        self.metadata_address = self.ds_cfg["metadata_url"]
-        self.userdata_address = self.ds_cfg["userdata_url"]
-        self.vendordata_address = self.ds_cfg["vendordata_url"]
-
         self.retries = int(self.ds_cfg.get("retries", DEF_MD_RETRIES))
         self.timeout = int(self.ds_cfg.get("timeout", DEF_MD_TIMEOUT))
+        self.alive = int(self.ds_cfg.get("alive", DEF_MD_ALIVE))
         self._fallback_interface = None
         self._network_config = sources.UNSET
+        self.metadata_urls = DS_BASE_URLS
+        self.userdata_url = None
+        self.vendordata_url = None
+        self.ephemeral_fixed_address = None
+        self.has_ipv4 = True
+        if "metadata_urls" in self.ds_cfg.keys():
+            # Merge and deduplicate lists
+            self.metadata_urls = list(
+                set(self.metadata_urls) | set(self.ds_cfg["metadata_urls"])
+            )
+
+    def _set_metadata_url(self, urls):
+        """
+        Define metadata_url based upon api-metadata URL availability.
+        """
+
+        for url in urls:
+            resp = None
+            for _ in range(0, self.retries):
+                try:
+                    LOG.debug("Trying to reach %s", url)
+                    resp = requests.get(
+                        url + "/conf?format=json",
+                        timeout=self.alive,
+                    )
+                except ConnectionError:
+                    continue
+                break
+            if not resp:
+                continue
+            if resp.status_code == 200:
+                LOG.debug("%s is reachable", url)
+                self.metadata_url = f"{url}/conf?format=json"
+                self.userdata_url = f"{url}/user_data/cloud-init"
+                self.vendordata_url = f"{url}/vendor_data/cloud-init"
+                return
+        LOG.debug("Unable to reach api-metadata at %s", urls)
+        raise ConnectionError
 
     def _crawl_metadata(self):
         resp = url_helper.readurl(
-            self.metadata_address, timeout=self.timeout, retries=self.retries
+            self.metadata_url, timeout=self.timeout, retries=self.retries
         )
         self.metadata = json.loads(util.decode_binary(resp.contents))
 
         self.userdata_raw = query_data_api(
-            "user-data", self.userdata_address, self.retries, self.timeout
+            "user-data", self.userdata_url, self.retries, self.timeout
         )
         self.vendordata_raw = query_data_api(
-            "vendor-data", self.vendordata_address, self.retries, self.timeout
+            "vendor-data", self.vendordata_url, self.retries, self.timeout
         )
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


cloudinit/sources/DataSourceScaleway.py
```diff
@@ -21,20 +23,16 @@
...
 from cloudinit.net.dhcp import NoDHCPLeaseError
-from cloudinit.net.ephemeral import EphemeralDHCPv4
+from cloudinit.net.ephemeral import EphemeralDHCPv4, EphemeralIPv6Network
 from cloudinit.sources import DataSourceHostname
+from cloudinit.subp import ProcessExecutionError
 
 LOG = logging.getLogger(__name__)
 
-DS_BASE_URL = "http://169.254.42.42"
+DS_BASE_URLS = ["http://169.254.42.42", "http://[fd00:42::42]"]
```
igalic, line 32: what if we made this into a hash with the protocol as key?

igalic (commented):
some comments after a casual review

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
         self._network_config = sources.UNSET
+        self.metadata_urls = DS_BASE_URLS
+        self.userdata_url = None
+        self.vendordata_url = None
+        self.ephemeral_fixed_address = None
+        self.has_ipv4 = True
+        if "metadata_urls" in self.ds_cfg.keys():
+            # Merge and deduplicate lists
+            self.metadata_urls = list(
+                set(self.metadata_urls) | set(self.ds_cfg["metadata_urls"])
```
holmanb, line 205: Sets are, by definition, unordered. Since the list that gets created with the combined set defines the order that urls are checked, we introduce uncertainty into the order that addresses are tried.

Do we really have no preference which address is first? I would think this is undesirable for debugging errors at the very least, and we may actually want to have ipv6 addresses if no other reason than because of its continued adoption / future.

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
-            util.logexc(LOG, str(e))
-            return False
+        if self.has_ipv4:
+            try:
+                # DHCPv4 waits for timeout defined in /etc/dhcp/dhclient.conf
+                # before giving up. Lower it in config file and try it first as
+                # it will only reach timeout on VMs with only IPv6 addresses.
+                with EphemeralDHCPv4(
+                    self._fallback_interface,
+                    tmp_dir=self.distro.get_tmp_exec_path(),
```
holmanb, line 286: This line can be dropped. `tmp_dir` isn't actually used anymore - that kwarg needs to get removed from the definition

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
+                    self.ephemeral_fixed_address = ipv4["fixed-address"]
+                    self.metadata["net_in_use"] = "ipv4"
+            except (
+                NoDHCPLeaseError,
+                ConnectionError,
+                ProcessExecutionError,
+            ) as e:
+                util.logexc(LOG, str(e))
+                self.has_ipv4 = False
+        if not self.has_ipv4:
```
holmanb, line 312: Using `self.has_ipv4 = False` to define control flow feels weird. Why not just return `true` at the end of the `try` block and allow flow to continue if an exception gets thrown?

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
+            "user-data", self.userdata_url, self.retries, self.timeout
         )
         self.vendordata_raw = query_data_api(
-            "vendor-data", self.vendordata_address, self.retries, self.timeout
+            "vendor-data", self.vendordata_url, self.retries, self.timeout
         )
 
+    def _set_urls_on_ip_version(self, proto, urls):
+
+        if proto not in ["ipv4", "ipv6"]:
```
holmanb, line 251: Why? Since all of the call sites currently have known values for `proto`, this won't ever get hit (except in the unit test).

holmanb (changes requested):
Thanks for this PR @karibou. This looks pretty good for the most part. I haven't looked at tests yet, but I have a couple of questions and  requested changes inline.

cloudinit/sources/DataSourceScaleway.py
holmanb: Why do that?

Most (if not all) of the datasources that use http have some sort of list of endpoints, and address attempt order needs to be preserved - some data sources may have reason to prefer different address class orders for different reasons.

igalic: the main reason was because later on we parse the addresses to figure out which one is IPv4 and which one is IPv6. We do this twice.

karibou: This would also complexify the structure as we may eventually add a DNS resolved endpoint that would resolve in both IPv4 and IPv6.

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -140,9 +138,20 @@ def query_data_api(api_type, api_address, retries, timeout):
...
+            # Adapt Session.mount to IPv4/IPv6 context
+            localhost = "0.0.0.0"
+            try:
+                url_address = urlparse(api_address).netloc
+                address = url_address
+                if url_address[0] == "[":
+                    address = url_address[1:-1]
+                if net.is_ipv6_address(address):
+                    localhost = "0::"
+            except ValueError:
```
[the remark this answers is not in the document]
karibou: Yes, indeed. Like passing an IPv6 URL with a missing square bracket as in `"http://[fd00:42::42"`

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -140,9 +138,20 @@ def query_data_api(api_type, api_address, retries, timeout):
                 "Trying to get %s data (bind on port %d)...", api_type, port
             )
             requests_session = requests.Session()
+            # Adapt Session.mount to IPv4/IPv6 context
+            localhost = "0.0.0.0"
```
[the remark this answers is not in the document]
karibou: I may change the variable name to something more obvious if you can suggest something

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
+            self.metadata_urls = list(
+                set(self.metadata_urls) | set(self.ds_cfg["metadata_urls"])
+            )
+
+    def _set_metadata_url(self, urls):
+        """
+        Define metadata_url based upon api-metadata URL availability.
+        """
+
+        for url in urls:
```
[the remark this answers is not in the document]
karibou: I did check `url_helper.wait_for_url()` late in the work, after creating `_set_urls_on_ip_version()` so it did not seem so useful since we already knew that was the proper IP version.
But after you mentioning it, I went back and its usage makes the code less complex by removing the need for  `_set_urls_on_ip_version()` so I went ahead and used it. 
Thanks for the tip

cloudinit/sources/DataSourceScaleway.py
karibou: Removed.

cloudinit/sources/DataSourceScaleway.py
karibou: I do agree with you that keeping the `metadata_urls` list ordered is most important. The idea behind this was to avoid trying an unreachable url more than once. But now that `url_helper.wait_for_url` is being used, it is no longer a problem since they are tried in parallel.
I'll change this to simply add to the end of the list.

cloudinit/sources/DataSourceScaleway.py
karibou: No longer relevant as this method has been dropped.

cloudinit/sources/DataSourceScaleway.py
karibou: This has been put there to avoid a second DHCPv4 timeout when the `init` phase runs as it also calls `_get_data` and when executed with IPv6 only, we hit the 5 second dhcp timeout twice.
There maybe a better way to handle this though

cloudinit/sources/DataSourceScaleway.py
```diff
@@ -177,51 +186,152 @@ def __init__(self, sys_cfg, distro, paths):
...
+            resp = None
+            for _ in range(0, self.retries):
+                try:
+                    LOG.debug("Trying to reach %s", url)
+                    resp = requests.get(
+                        url + "/conf?format=json",
+                        timeout=self.alive,
+                    )
+                except ConnectionError:
```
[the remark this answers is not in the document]
karibou: This has been removed with the use of `url_helper.wait_for_url`

cloudinit/sources/DataSourceScaleway.py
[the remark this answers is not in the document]
holmanb: > Thanks for the tip

No problem, and I'm glad this works for you. Delegating dualstack logic in shared code is a win.

cloudinit/sources/DataSourceScaleway.py
holmanb: Are you saying that with this change that `has_ipv4` will be set during Local stage so that it doesn't have to timeout again in Network stage? If so I don't believe that's how this works, since this shouldn't run at Network, but maybe I'm missing something.

karibou: I have sent you a copy of the `cloud-init.log` without the `if not self.has_ipv4` test where we clearly see the dhcp timeout appearing twice.
I believe that it is caused by the use of `EventType.BOOT` in `EventScope.NETWORK` that we use as we want to do the network configuration on each boot as the IP address may change b/w reboots.
Let me know if you need access to a test instance.

holmanb: Thanks! I'll take a look.

holmanb: Does ScaleWay use a dhcp server in the cloud environment? I wasn't expecting to see the exception:
```
cloudinit.net.dhcp.InvalidDHCPLeaseFileError: Cannot parse empty dhcp lease file /run/dhclient.lease
```

Is this is currently happening on all cloud-init instances on Scaleway (before this PR)? If so, that's a bug.

Since cloud-init is clearly not using the dhcp lease (it never got one), we could probably just bring up the interface with `EphemeralIPv4Network` to connect over the link local address, rather than using `EphemeralDHCPv4`. This will avoid the timeout issues we are seeing here.

karibou: Let me be a bit clearer on the current and upcoming behavior.
Up until now, IPv4 was configured using DHCPv4, hence the use of EphemeralDHCPv4 in the previous version. The optional IPv6 address was statically defined. With the upcoming changes to our infrastructure, we can have 3 different situations : 

- DHCPv4 only
- DHCPv4 and SLAAC IPv6
- SLAAC IPv6 only

Since DHCPv4 is and will remain most of the cases we start by trying EphemeralDHCPv4 as previously.

Since we have no way of knowing beforehand which IP version will be available, we try EphemeralDHCPv4 even if it is expected to fail when SLAAC IPv6 only is available. In that case only, the EphemeralDHCPv4 will timeout and we will continue with EphemeralIPv6Network for IPv6 configuration.
 
The issue here is not knowing which of IPv4 and/or IPv6 is available before the start of the DataSource so we must try both cases, starting with EphemeralDHCPv4 which will work in the vast majority.

holmanb: Thanks for the context @karibou, this is much clearer now. Can we add this explanation to the comments please? This answers many of the questions I've had while reading this code, so it should help future readers as well.

karibou: ok, will do

karibou: Added comment to the code & commit log.

holmanb (approved):
Thanks for this @karibou! This change looks good to me. Thanks for making the requested changes.

PR #2038, opened 2023-02-24
Bin unused bits

holmanb:
```
Remove dead code
```

## Additional Context
Give a shout if you think there is anything in here worth keeping. I'm of that opinion that if it isn't used, then git history is a good place to find it should we ever need it.


igalic (approved):
👍

dermotbradley (commented):
parse_deb_config is used by _ifaces_to_net_config_data function several lines below...

cloudinit/net/eni.py
```diff
@@ -308,18 +308,6 @@ def _parse_deb_config_data(ifaces, contents, src_dir, src_path):
             ifaces[iface]["auto"] = False
 
 
-def parse_deb_config(path):
```
dermotbradley, line 311: parse_deb_config is used by _ifaces_to_net_config_data function several lines below...

dermotbradley: Doh! I didn't spot that line was inside a comment block.

holmanb: All good. Updating the comments is important too :)

TheRealFalcon:
I'm a little nervous to do something like this when this project has historically embraced some of the more dynamic features of python. For example, at first glance one could be forgiven for thinking [this method](https://github.com/canonical/cloud-init/blob/ubuntu/jammy/cloudinit/net/network_state.py/#L383) is dead code (the decorator isn't required), but [because of some metaclass magic](https://github.com/canonical/cloud-init/blob/ubuntu/jammy/cloudinit/net/network_state.py/#L140), it actually gets used [here](https://github.com/canonical/cloud-init/blob/ubuntu/jammy/cloudinit/net/network_state.py/#L323).

Keeping obviously dead code around forever is bad too, but I guess I'm curious how the dead bits here were found. I think it might require some deeper inspection rather than simple greps of function names.

holmanb:
> I'm a little nervous to do something like this when this project has historically embraced some of the more dynamic features of python.

That's a valid concern. However, doesn't this argument apply to deleting _any_ function (ignoring first class functions) or method definition in this project? As a matter of practice, how well do we scrutinize every function deletion for errors like missing a dynamic call somewhere out in the code base?

This concern makes me think that maybe those dynamic call sites are more of a liability to keep around than they are worth. Readability, static analyzer defeat, and refactor risk are all worsened by their existence in the codebase.

> Keeping obviously dead code around forever is bad too, but I guess I'm curious how the dead bits here were found.

I used [vulture](https://pypi.org/project/vulture/) to get a list of potentially unused functions, and then checked manually (a bunch of grepping, basically) whether these were in use.

You'll note that if you run Vulture you'll see I sorted out a bunch of those dynamic functions (and therefore still show up as false positives). There are also some small functions in `util.py` that appear to only called by `tools/` functions, which doesn't pass the sniff test for me, but I didn't bother to include that refactor here.

I don't guarantee that none of these functions are called dynamically. I'm curious how one could do that for even one function, other than reading the whole code base.

This is pretty low priority. I don't have any issues with dropping this PR. Or maybe we just put it on hold until dynamic function risk is mitigated. Let me know your thoughts.

igalic:
> This is pretty low priority. I don't have any issues with dropping this PR. Or maybe we just put it on hold until dynamic function risk is mitigated. Let me know your thoughts.

how do we mitigate that?

TheRealFalcon:
I think maybe it's the all-at-once approach that concerns me. For any suspected piece of dead code, we could look at git history, search for metaclasses/globals()/locals(), look for importlib calls, etc and be reasonably sure it's not being used anywhere else. Have the changes here been that detailed though?

> This concern makes me think that maybe those dynamic call sites are more of a liability to keep around than they are worth. Readability, static analyzer defeat, and refactor risk are all worsened by their existence in the codebase.

Agree 100%

> This is pretty low priority. I don't have any issues with dropping this PR. Or maybe we just put it on hold until dynamic function risk is mitigated. Let me know your thoughts.

It's also pretty low risk. I'm being extra careful, but I think it'd be really bad to break a workload because a function just disappeared. Could we do it more piecemeal? I think some functions will be a lot more obvious than others based on where they are in the code base.

@aciba90 or @blackboxsw any strong opinions here? If others think I'm being too cautious I can let it go.

holmanb:
> For any suspected piece of dead code, we could look at git history, search for metaclasses/globals()/locals(), look for importlib calls, etc and be reasonably sure it's not being used anywhere else. Have the changes here been that detailed though?

I did most of the suggested in various forms before submitting this PR.

As for `globals()`, looks like I dropped the last usage of it from our non-test code in https://github.com/canonical/cloud-init/commit/51fef28bc561ca688308ac6ae060fa315e1c2d90.

Same goes for `locals()`, no access except in tests.

Metaclasses are tricky, since I don't know the best way to identify them. Any suggestions for locating metaclasses would be helpful. 

Importlib is only used in `importer.py` for module loading.

I checked git history in several of the changes, but not all of them. This could probably have been more thorough.

> Could we do it more piecemeal?

Sure. I suppose smaller bytes would be more palatable.

TheRealFalcon:
> Any suggestions for locating metaclasses would be helpful.

Anything involving `__new__` or `__init_subclass__` I think.

> Could we do it more piecemeal?
Sure. I suppose smaller bytes would be more palatable.

Eh...based on your findings I think I'm changing my mind. There's probably less dynamic stuff than I was thinking. I'll try to do a more thorough review soon.

TheRealFalcon (approved):
Thanks, LGTM!

PR #2040, opened 2023-02-27
do not attempt dns resolution on ip addresses

holmanb:
```
util: Skip dns resolution check for ip addresses
```

A slight optimization in the default case.

### DIFF
--- a/tests/unittests/test_util.py
+++ b/tests/unittests/test_util.py
@@ -3030,11 +3030,14 @@ class TestVersion:
 
 @pytest.mark.allow_dns_lookup
 class TestResolvable:
-    def test_ips_need_not_be_resolved(self):
+    @mock.patch.object(util, "_DNS_REDIRECT_IP", return_value=True)
+    @mock.patch.object(util.socket, "getaddrinfo")
+    def test_ips_need_not_be_resolved(self, m_getaddr, m_dns):
         """Optimization test: dns resolution may timeout during early boot, and
         often the urls being checked use IP addresses rather than dns names.
         Therefore, the fast path checks if the address contains an IP and exits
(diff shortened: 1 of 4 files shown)


tests/unittests/test_util.py
```diff
@@ -3026,3 +3026,15 @@ def test_to_version_and_back_to_str(self, version):
...
+
+@pytest.mark.allow_dns_lookup
+class TestResolvable:
+    def test_ips_need_not_be_resolved(self):
+        """Optimization test: dns resolution may timeout during early boot, and
+        often the urls being checked use IP addresses rather than dns names.
+        Therefore, the fast path checks if the address contains an IP and exits
+        early if the path is a valid IP.
+        """
+        assert util.is_resolvable("http://169.254.169.254/")
```
TheRealFalcon, line 3039: Nit: We should probably add an `is True`  to these asserts so we're more specific than just asserting truthiness.

tests/unittests/test_util.py
```diff
@@ -3026,3 +3026,15 @@ def test_to_version_and_back_to_str(self, version):
...
+        assert util.is_resolvable("http://[fd00:ec2::254]/")
```
TheRealFalcon, line 3040: I think we should be checking that `socket.getaddrinfo` isn't called here.

TheRealFalcon (approved):
Change looks good. I think we should probably add a mock to the unit test to ensure we're not attempting to resolve, but that's minor enough that I'm gonna approve anyway.

tests/unittests/test_util.py
holmanb: It gets used higher up in the function, but I can mock the global variable to skip that code.

PR #2041, opened 2023-03-01
sources/azure: fix regressions in IMDS behavior

cjp256:
sources/azure: fix regressions in IMDS behavior

    There are effectively two regressions in the recent IMDS refactor:

    1. The metadata check len(imds_md["interface"]) in
       _check_if_nic_is_primary() is no longer correct as the refactor
       switched URLs and did not update this call to account for the
       fact that this metadata now lives under "network".

    2. Network metadata was fetched with infinite=True and is now limited
       to ten retries.  This callback had the twist of only allowing up to
       ten connection errors but otherwise would retry indefinetely.

    For check_if_nic_is_primary():

    - Drop the interface count check for _check_if_nic_is_primary(),
      we don't need it anyways.

    - Fix/update the unit tests mocks that allowed the tests to pass,
      adding another test to verify max retries for http and connection
      errors.

    - Use 300 retries. We do want to hit a case where we spin forever,
      but this should be more than enough time for IMDS to respond in
      the Savable PPS case (~5 minutes).

    For IMDS:

    - Consolidate IMDS retry handlers into a new ReadUrlRetryHandler class
      that supports the options required for each variant of request.

    - Minor tweaks to log and expand logging checks in unit tests.

    - Move all unit tests to mocking via mock_requests_session_request
      and replace mock_readurl fixture with wrapped_readurl to improve
      consistency between tests.

    Note that this change drops usage of `retry_on_url_exc` and can probably
    be removed altogether as it is no longer used AFAICT.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -680,9 +680,9 @@ class DataSourceAzure(sources.DataSource):
         return crawled_data
 
     @azure_ds_telemetry_reporter
-    def get_metadata_from_imds(self) -> Dict:
+    def get_metadata_from_imds(self, retries: int = 10) -> Dict:
         try:
-            return imds.fetch_metadata_with_api_fallback()
+            return imds.fetch_metadata_with_api_fallback(retries=retries)
         except (UrlError, ValueError) as error:
             report_diagnostic_event(
                 "Ignoring IMDS metadata due to: %s" % error,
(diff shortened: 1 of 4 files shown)


anhvoms (approved):
LGTM

holmanb:
> I apologize for failing to catch this prior to tagging, but this would need to be included with any upcoming release.

Thanks for catching this! I think we'll be working on a fix tomorrow for a separate issue we found. Regardless, it'll be good to get this in. I'll review tomorrow unless someone else beats me to it.

cjp256:
This PR still needs further work.  I'll push out another round today.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -979,7 +979,7 @@ def _wait_for_pps_os_disk_shutdown(self):
         raise BrokenAzureDataSource("Shutdown failure for PPS disk.")
 
     @azure_ds_telemetry_reporter
-    def _check_if_nic_is_primary(self, ifname):
+    def _check_if_nic_is_primary(self, ifname: str) -> bool:
         """Check if a given interface is the primary nic or not. If it is the
         primary nic, then we also get the expected total nic count from IMDS.
```
anhvoms, line 984: update comment - the code is no longer getting the total nic count

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
cjp256: done!

TheRealFalcon:
@cjp256 , I've seen some additional pushes. Can you ping here when you think this is ready for review?

cjp256:
@TheRealFalcon it should be good to go and is now under heavy testing.  I failed to catch it earlier because the last few rounds of testing had a dist-ugprade in the setup script (added while doing the dhclient fix testing).  It caused an "upgrade" over the custom deb under test.  Whoops🤦 

`dpkg: warning: downgrading cloud-init from 22.4.2-0ubuntu0~22.04.1 to 22.4-114-g1bff600d-1~bddeb`

TheRealFalcon (approved):
Thanks!

PR #2043, opened 2023-03-01
send dhcp-client-identifier with dhclient command for InfiniBand ports

wmousa:
Sending dhclient command failed for InfiniBand ports because dhcp-client-identifier is not specified.

So, providing this patch to allow send dhcp-client-identifier hardware with the dhclient command for InfiniBand ports.

## Proposed Commit Message

```

```

## Checklist:

### DIFF
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -14,8 +14,13 @@ from io import StringIO
 
 import configobj
 
-from cloudinit import subp, util
-from cloudinit.net import find_fallback_nic, get_devicelist
+from cloudinit import subp, temp_utils, util
+from cloudinit.net import (
+    find_fallback_nic,
+    get_devicelist,
+    get_interface_mac,
+    is_ib_interface,
+)
 
 LOG = logging.getLogger(__name__)
 
@@ -136,6 +141,9 @@ def dhcp_discovery(dhclient_cmd_path, interface, dhcp_log_func=None):
     # link up before attempting discovery. Since we are using -sf /bin/true,
     # we need to do that "link up" ourselves first.
     subp.subp(["ip", "link", "set", "dev", interface, "up"], capture=True)
+    # For INFINIBAND port the dhlient must be sent with dhcp-client-identifier.
+    # So here we are checking if the interface is INFINIBAND or not.
+    # If yes, we are generating the the client-id to be used with the dhclient
     cmd = [
         dhclient_cmd_path,
         "-1",
@@ -148,6 +156,18 @@ def dhcp_discovery(dhclient_cmd_path, interface, dhcp_log_func=None):
         "-sf",
         "/bin/true",
     ]
+    if is_ib_interface(interface):
+        dhcp_client_identifier = "20:%s" % get_interface_mac(interface)[36:]
+        interface_dhclient_content = (
+            'interface "%s" '
+            "{send dhcp-client-identifier %s;}"
+            % (interface, dhcp_client_identifier)
+        )
+        tmp_dir = temp_utils.get_tmp_ancestor(needs_exe=True)
+        file_name = os.path.join(tmp_dir, interface + "-dhclient.conf")
+        util.write_file(file_name, interface_dhclient_content)
+        cmd.append("-cf")
+        cmd.append(file_name)
     out, err = subp.subp(cmd, capture=True)
 
     # Wait for pid file and lease file to appear, and for the process
--- a/tests/unittests/net/test_dhcp.py
+++ b/tests/unittests/net/test_dhcp.py
@@ -221,7 +221,6 @@ class TestDHCPRFC3442(CiTestCase):
 
 
 class TestDHCPParseStaticRoutes(CiTestCase):
-
     with_logs = True
 
     def parse_static_routes_empty_string(self):
@@ -336,6 +335,7 @@ class TestDHCPParseStaticRoutes(CiTestCase):
 
 class TestDHCPDiscoveryClean(CiTestCase):
     with_logs = True
+    ib_address_prefix = "00:00:00:00:00:00:00:00:00:00:00:00"
 
     @mock.patch("cloudinit.net.dhcp.find_fallback_nic")
     def test_no_fallback_nic_found(self, m_fallback_nic):
@@ -450,12 +450,21 @@ class TestDHCPDiscoveryClean(CiTestCase):
         )
         m_kill.assert_not_called()
 
+    @mock.patch("cloudinit.net.dhcp.is_ib_interface", return_value=False)
     @mock.patch("cloudinit.net.dhcp.os.remove")
     @mock.patch("cloudinit.net.dhcp.util.get_proc_ppid")
     @mock.patch("cloudinit.net.dhcp.os.kill")
     @mock.patch("cloudinit.net.dhcp.subp.subp")
     @mock.patch("cloudinit.util.wait_for_files", return_value=False)
-    def test_dhcp_discovery(self, m_wait, m_subp, m_kill, m_getppid, m_remove):
+    def test_dhcp_discovery(
+        self,
+        m_wait,
+        m_subp,
+        m_kill,
+        m_getppid,
+        m_remove,
+        mocked_is_ib_interface,
+    ):
         """dhcp_discovery brings up the interface and runs dhclient.
 
         It also returns the parsed dhcp.leases file.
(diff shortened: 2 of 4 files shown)


cloudinit/net/dhcp.py
```diff
@@ -149,6 +157,19 @@ def dhcp_discovery(dhclient_cmd_path, interface, dhcp_log_func=None):
         "-sf",
         "/bin/true",
     ]
+    is_ib_interface = read_sys_net_safe(interface, "type") == "32"
```
igalic, line 160: this should be a function

igalic (changes requested):
is this an infiniband device is a question we ask more than once, but we have yet to put it into a function, so it can be abstracted and more easily answered for more systems than just Linux

cloudinit/net/dhcp.py
```diff
@@ -149,6 +157,19 @@ def dhcp_discovery(dhclient_cmd_path, interface, dhcp_log_func=None):
...
+    if is_ib_interface:
+        dhcp_client_identifier = "20:%s" % get_interface_mac(interface)[36:]
+        interface_dhclient_content = (
+            'interface "%s" '
+            "{send dhcp-client-identifier %s;}"
+            % (interface, dhcp_client_identifier)
+        )
+        tmp_dir = temp_utils.get_tmp_ancestor(needs_exe=True)
+        file_name = os.path.join(tmp_dir, interface + "-dhclient.conf")
```
holmanb, line 169: I'd really rather not go down the path of dealing with temp configs. Why not just use the flag?
```
-I <dhcp-client-identifier>
    Specify the dhcp-client-identifier option to send to the DHCP server.
```

holmanb (changes requested):

holmanb (changes requested):
Thanks for this PR @wmousa.

Please [sign the CLA](https://cloudinit.readthedocs.io/en/latest/development/contributing.html#submitting-your-first-pull-request).

Can you please share a bit about your use-case? Do you actually expect to be using an InfiniBand device as your primary interface, or are you just trying to work around a bug you're seeing?

moshe010:
@holmanb In Openstack Nvidia has support for InfiniBand with virtual machine [1] . VMs can be defined with one interface which is InfiniBand. We want to fix cloud-init to be able to work also in this cases. 

[1] - https://docs.openstack.org/neutron/wallaby/admin/config-sriov.html under SR-IOV with InfiniBand[¶](https://docs.openstack.org/neutron/wallaby/admin/config-sriov.html#sr-iov-with-infiniband)

cloudinit/net/dhcp.py
```diff
@@ -149,6 +157,19 @@ def dhcp_discovery(dhclient_cmd_path, interface, dhcp_log_func=None):
         "-sf",
         "/bin/true",
     ]
+    is_ib_interface = read_sys_net_safe(interface, "type") == "32"
+    if is_ib_interface:
+        dhcp_client_identifier = "20:%s" % get_interface_mac(interface)[36:]
```
holmanb, line 162: Why not use the whole GUID? What's the point of this? ~~Mellanox~~ Nvidia's docs don't suggest that in the example I saw, so I'm curious why this approach here.

moshe010: the dhcp over ipoib spec [1] doesn't dictate what should be the client-id value.  redhat change the client-id generation in newer dhcp client. The only way it working consistant across the centos7/8/9 is using the following flag send dhcp-client-identifier = hardware; which send the client-id in format of <hw_type>+ <hw address>. 
So for ib it will be hardware-type 32 + <guid>


[1]  - https://www.rfc-editor.org/rfc/rfc4390.html
[2] - https://bugzilla.redhat.com/show_bug.cgi?id=1814386

cloudinit/net/dhcp.py
moshe010: so using -I will be send string and not as hex as we would like.  we didn't find away to do it without file and it seem to be consistent with how NetowrkManager is doing it.

cloudinit/net/dhcp.py
wmousa: Done

cloudinit/net/dhcp.py
holmanb: @esposem Apologies for the direct ping - can get access to [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1814386)? I tried requesting group membership, but that failed.

moshe010: what is you email address I can try to add you?

moshe010: try now I remove it from being private

holmanb: > try now I remove it from being private

Excellent, thank you. Do I understand correctly that this truncated hwaddr format is required for backwards compatibility with centos7?

moshe010: so according to redhat if you don't specify "send dhcp-client-identifier = hardware" the won't guarantee how client-id is generated between major version.
the client-id specified in Nvidia documentation [1] will work for CentOS 7, but for CentOS 8 it won't work it will be generated differently.
The only why to enforce it to be compatible in all versions is using "send dhcp-client-identifier = hardware" which for ethernet means 01-<mac> and for IB means 20-<GUID>. because we don't want to enforce the "send dhcp-client-identifier = hardware" on all interfaces (some may be ethernet) we are generating client-id just for IB interface according to 20-<GUID>. 

by the way the we are not truncated GUID.  
the hardware address exposed for ipoib is 20 bytes [2]
1. 1 byte reserved
2. 4 bytes for QP
3. 16 bytes for GID

GID (Global Identifier)  contains the GUID (Global Unique Identifiers) [3] which are the last 8 bytes of the GID.

[1] - https://docs.nvidia.com/networking/pages/viewpage.action?pageId=49158775#IPoverInfiniBand(IPoIB)-DHCPClient(Optional)
[2] - https://www.rdmamojo.com/2015/02/16/ip-infiniband-ipoib-architecture/ 
[3] - chrome-extension://efaidnbmnnnibpcajpcglclefindmkaj/https://network.nvidia.com/related-docs/whitepapers/InfiniBandFAQ_FQ_100.pdf

holmanb: Thanks for the info.

I spent some time reading systemd-networkd and NetworkManager's implementations, as well as the relevant specs and docs. I agree that this is the right approach.

If I understand correctly, this way the dhcp server will see the same dhcp-client-identifier later in boot (when systemd/NetworkManager provision network) and should issue the same IP as it did to dhclient.

moshe010: yes that correct

cloudinit/net/dhcp.py
holmanb: Unfortunately it looks like `dhclient` changed the meaning of this flag at some point:

From the latest man page:
```
       -I     Use the standard DDNS scheme from RFCs 4701 & 4702.
```

I guess this means that your current approach is the most cross-platform one at the moment, even if that's less ideal than a commandline argument.

tests/unittests/net/test_dhcp.py
```diff
@@ -512,6 +521,100 @@ def test_dhcp_discovery(self, m_wait, m_subp, m_kill, m_getppid, m_remove):
             ]
         )
         m_kill.assert_has_calls([mock.call(my_pid, signal.SIGKILL)])
+        mocked_is_ib_interface.assert_called_once_with("eth9")
+
+    @mock.patch("cloudinit.temp_utils.get_tmp_ancestor")
```
holmanb, line 526: Style nitpick: Mocks are usually easier to read when the return value is defined with the mock declaration like this:
```
    @mock.patch("cloudinit.temp_utils.get_tmp_ancestor", return_value="/tmp")
```

For this new one could we make that change to these mocks please?

holmanb (changes requested):
Thanks @wmousa. I see that you've signed the CLA, and this PR looks good, besides a small style nit below.

If you don't mind making the suggested change I'll get this merged.

cloudinit/net/dhcp.py
holmanb: Hmmm, actually the codepath exists in systemd-networkd, but using hardware address (`ClientIdentifier=mac`) is not the default setting. By default systemd-networkd uses `ClientIdentifier=duid`, which means that different IP addresses will be assigned to  dhclient and systemd-networkd.

Think we can/should try to add systemd support too?

wmousa: So, as I saw that IPOIB is not supported in cloud-init for ubuntu at all, there is no implementation here [1]
Now as conclusion, we have 3 issues here

1. It sends a wrong dhcp-client-identifier for sysconfig implementation (redhat based) which this patch fixes it
2. When metadata service is not reachable for any reason, cloud-init calls fallback data source which:

- For redhat-based, it creates ifcfg files with ethernet type and so it failed to get IP address. (We need to provide a fix for it)
- For ubuntu-based, it creates ethernet configuration without configuring dhcp-identifier and so it used the default ClientIdentifier=duid and it failed to get IP address, (Should be included in the last fix above)

3. We don't have netplan implementation in cloud-init for ubuntu-based, so in case metadata service is reachable and it applied the network configuration from DS, it will skip IB devices because it won't match any case in rendered objects in netplan (We need to add support for this)

[1] https://github.com/canonical/cloud-init/blob/main/cloudinit/net/netplan.py

wmousa:
> Thanks @wmousa. I see that you've signed the CLA.
> 
> Per [systemd/systemd#9609 (comment)](https://github.com/systemd/systemd/issues/9609#issuecomment-776277655), I don't see systemd-network switching to the way NetworkManager does it anytime soon.
> 
> Ideally I think we would be able support both duid and mac methods. What do you think would be the best approach for systemd-networkd support?

So duid sends different hardware-type and dhcp server won't reply to it, then it must be the mac

cloudinit/net/dhcp.py
holmanb: Thanks, good point on netplan support being required. Since this PR is an improvement in current state, I don't think we need to hold this PR up any further - netplan IPoIB support can be implemented in later change.

holmanb (approved):
Thanks for this contribution!

PR #2045, opened 2023-03-01
Force OpenStack When It Is Only Datasource in datasource_list

holmanb:
```
source: Force OpenStack when it is only option

Running on OpenStack Ironic was broken in 1efa8a0a0,
which prevented a system configured to run on only
Openstack from actually running this ds. This change
also prevents the kernel commandline override from
working. This change was required to prevent
unnecessarily probing OpenStack on Ec2, and is
therefore still required. 

This commit reverts an earlier attempt[1][2] to
automatically detect OpenStack, due to regression
it caused. Additionally, this change allows a
system that defines a datasource list containing
only [OpenStack] or [OpenStack, None] to attempt
running on OpenStack, overriding ds_detect(). A
datasource list that defines [OpenStack, None]
still falls back to DataSourceNone if OpenStack
fails to reach the IMDS.

This change also lays groundwork for the following
future work:

1. Add support for other datasources
2. Also override datasource checking when the kernel
   command line defines a datasource. This work needs
   to be done manually to support non-systemd systems.

Besides forcing OpenStack to run when it is the only
datasource in the datasource list, this commit also:

[1] 0220295 (it breaks some use cases)
[2] 29faf66 (no longer used)

LP: #2008727
```

## Additional Context
It may be easier to review this commit-by-commit. The first two commits reverted 29faf66 and 0220295, respectively

### Sample Logs

Log snippet on non-openstack (last line is new):
```
2023-03-01 21:45:56,065 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceNoCloud.DataSourceNoCloud'>
2023-03-01 21:45:56,066 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: boot-new-instance
2023-03-01 21:45:56,066 - __init__.py[DEBUG]: Machine is running on DataSourceNoCloud [seed=None][dsmode=net].
```

Log snippet on lxd, with this config:
```
$ cat /etc/cloud/cloud.cfg.d/90_dpkg.cfg 
datasource_list: [ OpenStack ]
```

Forced OpenStack on non-openstack:
```
2023-03-02 05:12:34,084 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceO
penStack.DataSourceOpenStackLocal'>
2023-03-02 05:12:34,084 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: boot-new-in
stance
2023-03-02 05:12:34,084 - __init__.py[DEBUG]: Machine is configured to run on single datasource DataSourceOpenStackLocal [net,ver=None].
```

Forced OpenStack on OpenStack Ironic:
```
2023-03-02 12:10:13,375 - stages.py[DEBUG]: Using distro class <class 'cloudinit.distros.ubuntu.Distro'>
2023-03-02 12:10:13,375 - __init__.py[DEBUG]: Looking for data source in: ['OpenStack'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM', 'NETWORK']
2023-03-02 12:10:13,377 - __init__.py[DEBUG]: Searching for network data source in: ['DataSourceOpenStack']
2023-03-02 12:10:13,377 - handlers.py[DEBUG]: start: init-network/search-OpenStack: searching for network data from DataSourceOpenStack
2023-03-02 12:10:13,377 - handlers.py[DEBUG]: Queuing POST to http://169.254.169.254/reporting/cloud-init, data: {'name': 'init-network/search-OpenStack', 'description': 'searching for network data from DataSourceOpenStack', 'event_type': 'start', 'origin': 'cloudinit', 'timestamp': 1677759013.3775737}
2023-03-02 12:10:13,377 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceOpenStack.DataSourceOpenStack'>
2023-03-02 12:10:13,377 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: boot-new-instance
2023-03-02 12:10:13,377 - __init__.py[DEBUG]: Machine is configured to run on single datasource DataSourceOpenStack [net,ver=None].
```

ref: https://github.com/canonical/cloud-init/pull/1923

### DIFF
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -280,22 +280,6 @@ def read_metadata_service(base_url, ssl_details=None, timeout=5, retries=5):
     return reader.read_v2()
 
 
-def detect_openstack(accept_oracle=False):
-    """Return True when a potential OpenStack platform is detected."""
-    if not util.is_x86():
-        return True  # Non-Intel cpus don't properly report dmi product names
-    product_name = dmi.read_dmi_data("system-product-name")
-    if product_name in VALID_DMI_PRODUCT_NAMES:
-        return True
-    elif dmi.read_dmi_data("chassis-asset-tag") in VALID_DMI_ASSET_TAGS:
-        return True
-    elif accept_oracle and oracle._is_platform_viable():
-        return True
-    elif util.get_proc_env(1).get("product_name") == DMI_PRODUCT_NOVA:
-        return True
-    return False
-
-
 # Used to match classes to dependencies
 datasources = [
     (DataSourceOpenStackLocal, (sources.DEP_FILESYSTEM,)),
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -150,9 +150,6 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
             False when unable to contact metadata service or when metadata
             format is invalid or disabled.
         """
-        oracle_considered = "Oracle" in self.sys_cfg.get("datasource_list")
-        if not detect_openstack(accept_oracle=not oracle_considered):
-            return False
 
         if self.perform_dhcp_setup:  # Setup networking in init-local stage.
             try:
@@ -246,6 +243,24 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
             )
             raise sources.InvalidMetaDataException(msg) from e
         return result
+
+    def ds_detect(self):
+        """Return True when a potential OpenStack platform is detected."""
+        accept_oracle = "Oracle" in self.sys_cfg.get("datasource_list")
+        if not util.is_x86():
+            # Non-Intel cpus don't properly report dmi product names
+            return True
+
+        product_name = dmi.read_dmi_data("system-product-name")
+        if product_name in VALID_DMI_PRODUCT_NAMES:
+            return True
+        elif dmi.read_dmi_data("chassis-asset-tag") in VALID_DMI_ASSET_TAGS:
+            return True
+        elif accept_oracle and oracle._is_platform_viable():
+            return True
+        elif util.get_proc_env(1).get("product_name") == DMI_PRODUCT_NOVA:
+            return True
+        return False
 
 
 class DataSourceOpenStackLocal(DataSourceOpenStack):
--- a/cloudinit/sources/DataSourceOpenStack.py
+++ b/cloudinit/sources/DataSourceOpenStack.py
@@ -73,7 +73,7 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
         mstr = "%s [%s,ver=%s]" % (root, self.dsmode, self.version)
         return mstr
 
-    def wait_for_metadata_service(self, max_wait=None, timeout=None):
+    def wait_for_metadata_service(self):
         urls = self.ds_cfg.get("metadata_urls", DEF_MD_URLS)
         filtered = [x for x in urls if util.is_resolvable_url(x)]
         if set(filtered) != set(urls):
@@ -90,23 +90,16 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
         md_urls = []
         url2base = {}
         for url in urls:
-            # Wait for a specific openstack metadata url
             md_url = url_helper.combine_url(url, "openstack")
             md_urls.append(md_url)
             url2base[md_url] = url
 
         url_params = self.get_url_params()
-        if max_wait is None:
-            max_wait = url_params.max_wait_seconds
-
-        if timeout is None:
-            timeout = url_params.timeout_seconds
-
         start_time = time.time()
         avail_url, _response = url_helper.wait_for_url(
             urls=md_urls,
-            max_wait=max_wait,
-            timeout=timeout,
+            max_wait=url_params.max_wait_seconds,
+            timeout=url_params.timeout_seconds,
             connect_synchronously=False,
         )
         if avail_url:
@@ -158,6 +151,8 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
             format is invalid or disabled.
         """
         oracle_considered = "Oracle" in self.sys_cfg.get("datasource_list")
+        if not detect_openstack(accept_oracle=not oracle_considered):
+            return False
 
         if self.perform_dhcp_setup:  # Setup networking in init-local stage.
             try:
@@ -165,15 +160,6 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
                     self.fallback_interface,
                     tmp_dir=self.distro.get_tmp_exec_path(),
                 ):
-                    if not self.detect_openstack(
-                        accept_oracle=not oracle_considered
-                    ):
-                        LOG.debug(
-                            "OpenStack datasource not running"
-                            " on OpenStack (dhcp)"
-                        )
-                        return False
-
                     results = util.log_time(
                         logfunc=LOG.debug,
                         msg="Crawl of metadata service",
@@ -183,13 +169,6 @@ class DataSourceOpenStack(openstack.SourceMixin, sources.DataSource):
                 util.logexc(LOG, str(e))
                 return False
         else:
-            if not self.detect_openstack(accept_oracle=not oracle_considered):
-                LOG.debug(
-                    "OpenStack datasource not running"
-                    " on OpenStack (non-dhcp)"
-                )
-                return False
-
             try:
                 results = self._crawl_metadata()
             except sources.InvalidMetaDataException as e:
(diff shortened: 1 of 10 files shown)


cloudinit/sources/__init__.py
```diff
@@ -317,18 +317,18 @@ def override_ds_detect(self):
         - TODO: commandline argument is used (ci.ds=OpenStack)
         """
         return self.sys_cfg.get("datasource_list", []) in (
-            [self.dslist],
-            [self.dslist, "None"],
+            [self.dsname],
```
TheRealFalcon, line 320: What is None comes first?

Would something like this be better?
```python
ds_list = self.sys_cfg.get("datasource_list", [])
return self.dsname in ds_list and any(
    [
        len(ds_list) == 1,
        "None" in ds_list and len(ds_list) == 2,
    ]
)
```

holmanb: > What is None comes first?

If that can happen, this improvement looks good to me.

TheRealFalcon: since it's user-modifyable, it's unlikely, but it can.

TheRealFalcon: I mean...we specifically document a datasource alongside none, so I don't think it was an unreasonable thought. However, I wasn't thinking that the datasource list is ordered, so `None` coming at front doesn't actually work. So....nevermind.

cloudinit/sources/__init__.py
```diff
@@ -307,6 +307,29 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
         return type_utils.obj_name(self)
 
+    def ds_detect(self) -> bool:
+        """Check if running on this datasource"""
+        return True
+
+    def override_ds_detect(self):
+        """Override if either:
+        - only a single datasource defined (nothing to fall back to)
+        - TODO: commandline argument is used (ci.ds=OpenStack)
```
blackboxsw, line 317: Either `ci.ds=OpenStack` or `ci.datasource=OpenStack` result in `sys_cfg["datasource_list] == ["OpenStack", "None"]`

cloudinit/sources/__init__.py
```diff
@@ -307,6 +307,29 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+        """
+        return self.sys_cfg.get("datasource_list", []) in (
+            [self.dsname],
+            [self.dsname, "None"],
+        )
+
+    def _check_and_get_data(self):
+        """Overrides runtime datasource detection"""
+        if self.override_ds_detect() or self.ds_detect():
+            LOG.debug("Machine is running on %s.", self)
```
blackboxsw, line 327: ??? WDYT
```suggestion
            LOG.debug("Machine configured to run on single datasource %s.", self)
```

cloudinit/sources/__init__.py
blackboxsw: For future, not this PR. I don't like that NoCloud datasource is unique in that it [accepts `ds=nocloud-*` params](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L838-L841) but every other datasource has to rely on ci.ds or ci.datasource. That only adds to confusion.

blackboxsw (approved):
minor nits, refactors look good.

cloudinit/sources/__init__.py
holmanb: As written, this is incorrect. I think you mean for that log to be written only if `self.override_ds_detect()`, but not if `self.ds_detect()`?

I can rework the logic to make that happen; I can see that being a useful distinction to log.

cloudinit/sources/__init__.py
holmanb: > but maybe cloud-init it's a bit presumptive to assume we can own the ds= scope of kernel parameter

Agreed. That, and the fact that `ci.ds=` and `ci.datasource=` have parity, I don't see any reason not to deprecate `ci.ds=` (and only documenting `ci.datasource=`) when we get around to making this changes.

holmanb:
cc @BirknerAlex 

The idea here is:

If `/etc/cloud/cloud.cfg` is defined with a single datasource configured, ds-identify will select that datasource.

This change tweaks logic such that if only one datasource is defined, then runtime checking for that datasource is disabled. We think that this is sensible, since no other datasources will be attempted, so forcing the datasource to be tried should be okay.

cloudinit/sources/__init__.py
blackboxsw: to clarify, I mean I think it'd be ideal to support `ds=openstack`  because that means typing less on the grub menu  when manually extending supplemental kernel params.... so I'd want to avoid dropping `ci.ds=` because less character is better where possible. I was just uncertain if we'd want to support even shorter alias `ds=` which would be 'outside` the top-level `ci` (cloud-init) scope.

BirknerAlex:
I will test your changes on bare metal in a few hours, let you know if everything is working. 👍

cloudinit/sources/__init__.py
holmanb: Eh, not sure I'm on board with this angle.

I lean towards the safety of longer & namespaced over shaving a few keystrokes, and less supported argument formats rather than more. More means more current and future potential collisions, more things to maintain, etc.

As we've seen with Ironic, different datasources make different usage of the kernel commandline, so we might not know of collisions on other datasources until we enable it on those datasources.

I'm willing to be convinced, my gut reaction is just the opposite of the proposed.

holmanb: > For future, not this PR. I don't like that NoCloud datasource is unique in that it [accepts ds=nocloud-* params](https://github.com/canonical/cloud-init/blob/main/tools/ds-identify#L838-L841) but every other datasource has to rely on ci.ds or ci.datasource. That only adds to confusion.

As for this earlier comment, I 100% agree. I'd just love to see NoCloud change to line up with the rest of cloud-init rather than move the rest of the datasources to line up with it.

BirknerAlex:
I can confirm it working on Bare Metal after building the deb file and installing it on the system and creating the config `/etc/cloud.cfg.d/91-cloud-init-datasource.cfg` like:

```
#cloud-config
datasource_list: [ OpenStack ]
```

aciba90 (approved):
LGTM, thanks!

cloudinit/sources/__init__.py
TheRealFalcon: > I lean towards the safety of longer & namespaced over shaving a few keystrokes

I would normally agree, but I'm pretty sure there can still be fairly low character limits for the kernel command line. I think it's one of the few places that terseness should be the norm. I agree that collisions might be an issue though.

cloudinit/sources/DataSourceOpenStack.py
```diff
@@ -280,13 +258,6 @@ def detect_openstack(self, accept_oracle=False):
             return True
         elif util.get_proc_env(1).get("product_name") == DMI_PRODUCT_NOVA:
```
igalic, line 259: I have to make a mental note of this:

is this another way of reading kernel command line parameters?

igalic (commented):
❓

blackboxsw (approved):
This looks good Performed additional integration test runs on Azure LXD Container and LXD VM. and Openstack manual upgrades successfully with this changeset.  Only found an unrelated rsystlog config bug that affects lunar.. will put up a separate PR for that.

```
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: warning during parsing file /etc/rsyslog.conf, on or before line 54: STOP is followed by unreachable stateme>
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: imuxsock: Acquired UNIX socket '/run/systemd/journal/syslog' (fd 3) from systemd.  [v8.2302.0]
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: imklog: cannot open kernel log (/proc/kmsg): Permission denied.
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: activation of module imklog failed [v8.2302.0 try https://www.rsyslog.com/e/2145 ]
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: rsyslogd's groupid changed to 107
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: rsyslogd's userid changed to 101
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: [origin software="rsyslogd" swVersion="8.2302.0" x-pid="242" x-info="https://www.rsyslog.com"] start
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: rsyslogd: file '/var/tmp/rsyslog.log': open error: Permission denied [v8.2302.0 try https://www.rsyslog.com/>
Mar 02 06:32:55 cloudinit-0302-162033fwxto24o rsyslogd[242]: file '/var/tmp/rsyslog.log': open error: Permission denied [v8.2302.0 try https://www.rsyslog.com/e/2433 ]
```

cloudinit/sources/__init__.py
holmanb: > I would normally agree, but I'm pretty sure there can still be fairly low character limits for the kernel command line. I think it's one of the few places that terseness should be the norm. I agree that collisions might be an issue though.

Fair point. [On x86 the buffer is 2048 bytes](https://elixir.bootlin.com/linux/v6.2.1/source/arch/x86/include/asm/setup.h#L7), and poking around it looks like any architecture we should care about has a size of >=1024 bytes.

@blackboxsw @TheRealFalcon This is going to be documented in a followup PR to document datasource selection behavior. Do we have strong feelings about shaving those first three bytes (`ci.`)? Or does using `ci.ds=` as a tradeoff seem reasonable - shorter, but keep the name-spacing for more uniqueness?

TheRealFalcon: `ci.ds=` sounds good to me

holmanb: @TheRealFalcon @blackboxsw While documenting this change, I notice that we recently deprecated `url` in favor of `cloud-config-url` in 8e26b2f8a9ce8c1186a6e5f2e9bd90a984fde52a. For consistency, I'm thinking we should maybe revert this, and possibly switch to `ci.url=` while deprecating the other two.

PR #2047, opened 2023-03-02
get an container shell with tox -e boot

holmanb:
```
tox: add boot environment

Uses test infrastructure to start an lxc container
from current directory.
```

User / dev tool improvement. There is nothing novel here, just a bit of glue for nicer UX. I assume most devs have their own magic to do effectively this quickly, no reason not to give it to users. It would be trivial to add a dedicated user/meta/vendor filepath to source as well.

## Test Steps
```bash
$ git clone https://github.com/canonical/cloud-init.git
$ cd cloud-init
$ tox -e boot
root@boot:~#
```

### DIFF
--- a/tests/integration_tests/test_utility.py
+++ b/tests/integration_tests/test_utility.py
@@ -1,0 +1,13 @@
+import logging
+
+import pytest
+
+LOG = logging.getLogger("integration_testing.test_utility")
+
+
+class TestUtilities:
+    @pytest.mark.instance_name("boot")
+    def test_boot(self, client):
+        """Boot an instance. That's all it does."""
+        assert client.execute("whoami").ok
+        LOG.info('Instance name "boot" started')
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
I think it might make more sense to run a bash script:
```bash
NAME=boot

lxc rm -f $NAME
lxc init ubuntu:jammy $NAME
lxc config device add $NAME host-cloud-init disk source=<current_directory> path=/usr/lib/python3/dist-packages/cloudinit
lxc start $NAME
lxc exec $NAME -- /bin/bash -c ":"
while [ $? -ne 0 ]; do
    lxc exec $NAME -- /bin/bash -c ":" &> /dev/null
done

lxc shell $NAME
```

It doesn't require adding an otherwise useless test and it'd be easier to configure with env variables (e.g., how do you pass a cloud-config using the pytest version?). It's also a little more straightforward for newcomers as it doesn't require understanding our integration test setup.

PR #2055, opened 2023-03-03
Ubuntu/bionic 23.1.x

aciba90:
do not squash


debian/changelog
```diff
@@ -1,9 +1,12 @@
...
   * d/patches/netplan99-cannot-use-default.patch:
     - Retain routes' definitions compatible with netplan 0.99
+  * Upstream snapshot based on 23.1.1. (LP: #2008230).
+    List of changes from upstream can be found at
+    https://raw.githubusercontent.com/canonical/cloud-init/23.1.1/ChangeLog
 
- -- Alberto Contreras <alberto.contreras@canonical.com>  Mon, 27 Feb 2023 20:41:44 +0100
+ -- Alberto Contreras <alberto.contreras@canonical.com>  Fri, 03 Mar 2023 10:03:11 +0100
 
 cloud-init (23.1-0ubuntu0~18.04.1) bionic; urgency=medium
```
blackboxsw, line 11: @aciba90 since we never released 23.1....~18.04.1 I think we probably should collapse the content from even this entry below into one d/changelog entry

What do you think about something  like this?
```
cloud-init (23.1.1-0ubuntu0~18.04.1) bionic; urgency=medium

  * d/patches/netplan99-cannot-use-default.patch:
    - Retain routes' definitions compatible with netplan 0.99
  * d/patches/retain-netplan-world-readable.patch:
    - Retain original world-readable perms of /etc/netplan/50-cloud-init.yaml.
      Lunar made the config root read-only.
  * refresh patches:
    + debian/patches/expire-on-hashed-users.patch
  * Upstream snapshot based on 23.1.1. (LP: #2008230).
    List of changes from upstream can be found at
    https://raw.githubusercontent.com/canonical/cloud-init/23.1.1/ChangeLog

 -- Alberto Contreras <alberto.contreras@canonical.com>   ......  +0100
```

aciba90: That's right, done.

blackboxsw (approved):
- included changeset about 23.1 -> 23.1.1 looks correct
- only authorship diffs in d/changelog between local branch and this 
- build-package ; sbuild happy with no new lintian warnings/errors
- good consolidation of unlreased d/changelog entries

PR #2057, opened 2023-03-03
docs: cosmetic improvements

s-makin:
Docs: cosmetic improvements and styling

```
A series of cosmetic docs changes that will make them look more like the
documentation produced on Ubuntu.com. It's not *exactly* the same, but
close enough to make it look recognisable as Canonical docs.

Although I tried to test this on a variety of screen sizes, if you spot
anything that doesn't render right or looks weird, please let me know. The
default font sizes were problematic from an accessibility pov so I made
them larger. They are still variable so they should render right on smaller
screens but it's possible I missed some.

Code literals are now the same size as regular font for the same reason,
but made slightly bolded to help pick them out from text (which should
help e.g. when commands are included as inline text and not a code block.

For easier navigation, I've given the sidebar headings a higher weight, and
as an added bonus, when scrolling through a long page the right-hand side bar
also shows what part you're on.

For accessibility, I changed the weight and spacing between topics in the
side bars to help separate them and make it easier to tell where one link
ends and the next begins.

As always, let me know if there's any other improvements I can make or any
things I've done that don't work!  
```

### DIFF
--- a/doc/rtd/conf.py
+++ b/doc/rtd/conf.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import datetime
 
 from cloudinit import version
 
@@ -18,7 +19,8 @@ sys.path.insert(0, os.path.abspath("."))
 
 # General information about the project.
 project = "cloud-init"
-copyright = "Canonical Ltd."
+author = "Canonical Group Ltd."
+copyright = "%s, %s" % (datetime.date.today().year, author)
 
 # -- General configuration ----------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


doc/rtd/conf.py
```diff
@@ -18,7 +19,8 @@
 
 # General information about the project.
 project = "cloud-init"
-copyright = "Canonical Ltd."
+author = "Canonical Group Ltd."
+copyright = "%s, %s" % (datetime.date.today().year, author)
```
holmanb, line 23: How do we feel about f-strings?
```python
copyright = f"Canonical Group Ltd., {datetime.date.today().year}"
```

holmanb (approved):
I don't know css, but I spot checked several pages and it all looked "good", so +1 from me.

I'll leave it open in case you'd like to apply the nitpick suggestion.

doc/rtd/conf.py
s-makin: I'm FOR 'em! (My Python is just a bit rusty)

s-makin: Thanks for the review, I'll wait til Monday before applying it though, just in case anyone hates f-strings with a burning passion.

holmanb (commented):
It looks like your import order is off - we use black/isort to automatically keep formatting and import order looking nice.

Running the following and commiting the changes should most likely fix the failing tests.
```bash
tox -e do_format
```

s-makin:
> It looks like your import order is off - we use black/isort to automatically keep formatting and import order looking nice.
> 
> Running the following and commiting the changes should most likely fix the failing tests.
> 
> ```shell
> tox -e do_format
> ```

Thanks Brett, I'll give it a go and see what happens!

doc/rtd/conf.py
holmanb: > Thanks for the review, I'll wait til Monday before applying it though, just in case anyone hates f-strings with a burning passion.

It's up to you, but in cloud-init the current devs prefer f-strings in most places (except in [logging messages as an optimization](https://docs.python.org/3/howto/logging.html#optimization)), so I don't think you'll get any push back on that part.

s-makin: Good to know! :D Thanks, I'll get those changes pushed over

PR #2058, opened 2023-03-03
update grub-dpkg docs

holmanb:
```
config: Update grub-dpkg docs
```

## Additional Context
We should always describe config key metadata in the appropriate schema field rather than in the module description.

### DIFF
--- a/cloudinit/config/cc_grub_dpkg.py
+++ b/cloudinit/config/cc_grub_dpkg.py
@@ -22,7 +22,7 @@ from cloudinit.subp import ProcessExecutionError
 MODULE_DESCRIPTION = """\
 Configure which device is used as the target for grub installation. This module
 can be enabled/disabled using the ``enabled`` config key in the ``grub_dpkg``
-config dict. This module automatically selects a disk with ``grub-probe`` if no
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_grub_dpkg.py
```diff
@@ -22,9 +22,8 @@
 MODULE_DESCRIPTION = """\
 Configure which device is used as the target for grub installation. This module
 can be enabled/disabled using the ``enabled`` config key in the ``grub_dpkg``
-config dict. The global config key ``grub-dpkg`` is an alias for ``grub_dpkg``.
-If no installation device is specified this module will execute grub-probe to
-determine which disk the /boot directory is associated with.
+config dict. This module automatically selects a disk with ``grub-probe`` if no
```
s-makin, line 25:
```suggestion
config dict. This module automatically selects a disk using ``grub-probe`` if no
```

s-makin: This read a little ambiguously at first. I think "using" is a more accurate reflection of what's happening (happy to be corrected though if I'm off base)

aciba90 (approved):
LGTM, thanks!

PR #2062, opened 2023-03-07
Log deprecated Python features

holmanb:
```
Log deprecated Python features
      
- Add deduplicating Python deprecation utility
- Only log for versions of Python which have features that have been deprecated
- Also fix bug in deprecate_call()
```

## Additional Context
Fixing all of these issues would be better than just logging them, but this is better than nothing - this makes Python deprecated features discoverable for future maintenance, and a minor nuisance for those that are affected until they are resolved.

### DIFF
--- a/cloudinit/distros/netbsd.py
+++ b/cloudinit/distros/netbsd.py
@@ -90,6 +90,11 @@ class NetBSD(cloudinit.distros.bsd.BSD):
         if hashed:
             hashed_pw = passwd
         else:
+            util.deprecate_python(
+                deprecated="Standard library module 'crypt'",
+                deprecated_version="3.11",
+                removed_version="3.13",
+            )
             method = crypt.METHOD_BLOWFISH  # pylint: disable=E1101
             hashed_pw = crypt.crypt(passwd, crypt.mksalt(method))
 
@@ -136,6 +141,3 @@ class NetBSD(cloudinit.distros.bsd.BSD):
 
 class Distro(NetBSD):
     pass
-
-
-# vi: ts=4 expandtab
--- a/cloudinit/distros/parsers/sys_conf.py
+++ b/cloudinit/distros/parsers/sys_conf.py
@@ -17,6 +17,8 @@ from io import StringIO
 # bash scripts...
 import configobj
 
+from cloudinit import util
+
 # See: http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html
 # or look at the 'param_expand()' function in the subst.c file in the bash
 # source tarball...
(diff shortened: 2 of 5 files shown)


TheRealFalcon:
Unfortunately, I'm -1 on this idea for a few reasons:
* I don't really understand the benefit. If we know something is deprecated and needs to be fixed, why not just file a launchpad or jira issue? If that's not a loud enough signal then something in our team process is broken. We should fix that rather than add code to scream even louder at us.
* I don't think "build time" issues should go into a runtime log.
* Saying "Please file a bug report" puts burden on the users that is really entirely our responsibility.
* It adds a type of deprecation warning in the logs that the user can't do anything about. That probably means they'll be more likely to ignore the schema warnings that they do need to do something about.
* pytest gives us this for free (as long as the code is covered). It has a warnings section at the bottom that as of 3.11 includes
```
cloudinit/sources/DataSourceAzure.py:8
  /home/james/c/cloud-init/cloudinit/sources/DataSourceAzure.py:8: DeprecationWarning: 'crypt' is deprecated and slated for removal in Python 3.13
    import crypt

cloudinit/distros/parsers/sys_conf.py:7
  /home/james/c/cloud-init/cloudinit/distros/parsers/sys_conf.py:7: DeprecationWarning: 'pipes' is deprecated and slated for removal in Python 3.13
    import pipes
```
Pytest also allows treating warnings as errors. Not something I think we should enable everywhere, but perhaps as a non-voting run on our latest supported version.

Please let me know if I'm missing something!

holmanb:
> why not just file a launchpad or jira issue?

The work to fix these items feel like something that would would be great for new contributors. For that, however, Launchpad feels like a black hole in a lot of ways, and JIRA (which does have an item for this) isn't user-facing to solicit contributions. I guess yelling at all users in the logs probably isn't the right way to communicate this either. The upcoming Github Issues support will help with this.

>  I don't think "build time" issues should go into a runtime log.
>  Saying "Please file a bug report" puts burden on the users that is really entirely our responsibility.

Fair, and agreed. Not all Python deprecations can be discovered at build time, but I guess all of the ones in this PR can be, and most that can't be discovered at build time can probably be trivially mitigated.

>    It adds a type of deprecation warning in the logs that the user can't do anything about. That probably means they'll be more likely to ignore the schema warnings that they do need to do something about.

Hopefully seeing a deprecation log wouldn't throw them off from their troubleshooting, but I see your point about adding noise.

> pytest gives us this for free (as long as the code is covered). It has a warnings section at the bottom that as of 3.11 includes

A limitation of our coverage, but yes, it picked up two of the deprecations I documented here.

> Please let me know if I'm missing something!

Nope, I think you're right. This doesn't make sense. I'll reopen a different PR to fix the bug separate.

PR #2075, opened 2023-03-17
sources/azure: move pps handling out of _poll_imds()

cjp256:
Pull out remaining PPS handling bits from _poll_imds() and add two explicit methods for the overloaded path:

- _wait_for_pps_running_reuse() for running PPS logic.

- _wait_for_pps_unknown_reuse() for unknown and recovery PPS logic.

For consistency:

- Rename _wait_for_all_nics_ready() -> _wait_for_pps_savable_reuse().

- Move reporting ready logic into _wait_for_pps_os_disk_shutdown().

Drop several impacted tests as coverage already exists in TestProvisioning, and update the rest to handle the +/- 1 DHCP attempt due to varying assumptions around PPS state and DHCP.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -9,6 +9,7 @@ import crypt
 import os
 import os.path
 import re
+import socket
 import xml.etree.ElementTree as ET
 from enum import Enum
 from pathlib import Path
@@ -571,11 +572,14 @@ class DataSourceAzure(sources.DataSource):
                 report_diagnostic_event(msg, logger_func=LOG.error)
                 raise sources.InvalidMetaDataException(msg)
 
+            if pps_type == PPSType.RUNNING:
+                self._wait_for_pps_running_reuse()
             if pps_type == PPSType.SAVABLE:
-                self._wait_for_all_nics_ready()
+                self._wait_for_pps_savable_reuse()
             elif pps_type == PPSType.OS_DISK:
-                self._report_ready_for_pps(create_marker=False)
                 self._wait_for_pps_os_disk_shutdown()
+            else:
+                self._wait_for_pps_unknown_reuse()
 
             md, userdata_raw, cfg, files = self._reprovision()
             # fetch metadata again as it has changed after reprovisioning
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1061,16 +1056,71 @@ def _wait_for_hot_attached_primary_nic(self, nl_sock):
...
+
+        try:
+            if (
+                self._ephemeral_dhcp_ctx is None
+                or self._ephemeral_dhcp_ctx.iface is None
+            ):
+                raise RuntimeError("missing ephemeral context")
+
+            iface = self._ephemeral_dhcp_ctx.iface
+            self._report_ready_for_pps()
```
anhvoms, line 1094: might be better to call it while explicitly specifying create_marker=True, this will help distinguishing between the different PPS scenarios during code review
do we want some unit tests covering the behavior of different wait_for_pps_* wrt to creating marker file?

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1061,16 +1056,71 @@ def _wait_for_hot_attached_primary_nic(self, nl_sock):
...
+        raise BrokenAzureDataSource("Shutdown failure for PPS disk.")
+
+    @azure_ds_telemetry_reporter
+    def _wait_for_pps_running_reuse(self) -> None:
+        """Report ready and wait for nic link to switch upon re-use."""
+        nl_sock = self._create_bound_netlink_socket()
+
+        try:
+            if (
+                self._ephemeral_dhcp_ctx is None
```
anhvoms, line 1088: I'm curious if these conditions can actually happen. With the current code setup the ephemeral_dhcp_ctx can't be None because setup_ephemeral_network is called in a loop before this function is called. Perhaps this is more of an assert

cjp256: They won't but they are checked for mypy.  The alternative is an `assert` in code to keep it more concise.  Otherwise, I do have a planned refactor in the distant future which should address this.

cloudinit/sources/DataSourceAzure.py
cjp256: the exceptional one is osdisk as it doesn't create a marker.

There was some coverage for ready_ready marker, but I expanded it for all PPS scenarios.

TheRealFalcon (approved):
LGTM

PR #2076, opened 2023-03-20
Don't change permissions of netrules target

TheRealFalcon:
## Proposed Commit Message

```
Don't change permissions of netrules target

Set permissions if file doesn't exist. Leave them if it does.

LP: #2011783
```

## Additional Context
https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2011783


blackboxsw (approved):
Changeset good for me. I was hoping for a bit of unittest coverage on this. WDYT of the pushed additional commit to at least get eni coverage of pre-existing perms being preserved?

PR #2077, opened 2023-03-20
Ubuntu/focal

blackboxsw:
## do not squash merge
Reflecting @bdrung's #2020 to all `ubuntu/*` downstream branches we support 

```
    Add Apport general hook
    
    Apport 2.25.0-0ubuntu1 removed the buggy `add_cloud_info` from the
    ubuntu general hook. Add an Apport general hook that reads
    `/run/cloud-init/instance-data.json` if present.
    
    LP: #1724623
    Thanks for the initial version: John Chittum
```

## Checklist:

### DIFF
--- a/debian/apport-general-hook.py
+++ b/debian/apport-general-hook.py
@@ -1,0 +1,99 @@
+"""General Apport hook for all reports that are using cloud-init."""
+
+import json
+import logging
+
+
+def _get_azure_data(ds_data) -> dict[str, str]:
+    compute = ds_data.get("meta_data", {}).get("imds", {}).get("compute")
+    if not compute:
+        return {}
+    name_to_report_map = {
+        "publisher": "ImagePublisher",
+        "offer": "ImageOffer",
+        "sku": "ImageSKU",
+        "version": "ImageVersion",
+        "vmSize": "VMSize",
+    }
+    azure_data = {}
+    for src_key, report_key_name in name_to_report_map.items():
+        azure_data[report_key_name] = compute[src_key]
+    return azure_data
+
+
+def _get_ec2_data(ds_data) -> dict[str, str]:
+    document = (
+        ds_data.get("dynamic", {}).get("instance-identity", {}).get("document")
+    )
+    if not document:
+        return {}
+    wanted_keys = {
+        "architecture",
+        "billingProducts",
+        "imageId",
+        "instanceType",
+        "region",
+    }
+    return {
+        key: value for key, value in document.items() if key in wanted_keys
+    }
+
+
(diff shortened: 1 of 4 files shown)


debian/apport-general-hook.py
```diff
@@ -0,0 +1,99 @@
...
+def add_info(report, ui) -> None:
+    """Entry point for Apport.
+
+    Add a subset of non-sensitive cloud-init data from
+    /run/cloud/instance-data.json that will be helpful for debugging.
+    """
+    try:
+        with open("/run/cloud-init/instance-data.json", "r") as fopen:
+            instance_data = json.load(fopen)
+    except FileNotFoundError:
```
TheRealFalcon, line 70: Is it worth adding a `JSONDecodeError` catch here too? How bad is it if we exit with exception?

debian/apport-general-hook.py
```diff
@@ -0,0 +1,99 @@
...
+    /run/cloud/instance-data.json.
+    """
+    platform_info = PLATFORM_SPECIFIC_INFO.get(platform)
+    if not platform_info:
+        return
+    retrieved_data = platform_info(ds_data)
+    for key, value in retrieved_data.items():
+        if not value:
+            continue
+        report[platform.capitalize() + key.capitalize()] = value
```
TheRealFalcon, line 58: Should we have an underscore separating these? Otherwise wouildn't we get keys like "AZUREPUBLISHER"?

TheRealFalcon (commented):
Two inline questions

debian/apport-general-hook.py
blackboxsw: We could catch that, tracebacks from apport on not critical as it just prevented you from creating and uploading an official bug report automatically.  In those cases, I'd expect we'd see another bug report manually filed about inability to create a bug using apport.
 Generally though, I don't expect we'd receive apport bugs with this issue as we'd probably get a bigger manually filed issue of "nothing works in cloud-init and my datasource wasn't detected".

debian/apport-general-hook.py
TheRealFalcon: Oooh, right. Sorry, I mistook capitalize to mean make the whole word uppercase.

debian/apport-general-hook.py
TheRealFalcon: Yeah, I think it's fine as-is

TheRealFalcon (approved):

debian/apport-general-hook.py
bdrung: If `add_info` fails, apport will include the stack trace in the report. So IMO this should be fine.

PR #2088, opened 2023-03-22
test-requirements: add pytest-antilru plugin to manage lru caches

cjp256:
lru caches may be populated on module loads and not cleared for tests (or between tests).

Ensure all tests run with a clean cache with pytest-antilru.  It wraps lru_cache() to track each user and clear them for each test.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -3,6 +3,9 @@ pytest
 pytest-cov
 pytest-mock
 
+# Clear lru caches on every test
+pytest-antilru
+
 # Only really needed on older versions of python
 setuptools
 jsonschema
--- a/tox.ini
+++ b/tox.ini
@@ -191,6 +191,7 @@ deps =
     netifaces==0.10.4
     # test-requirements
     pytest==3.3.2
+    pytest-antilru==1.1.1
     pytest-cov==2.5.1
     pytest-mock==1.7.1
     # Needed by pytest and default causes failures


holmanb:
I think that this is cleaner and less error prone than [manually calling `cache_clear()`](https://github.com/canonical/cloud-init/blob/055395d38a4ec0cc87b005e93110e6d06419f43d/tests/unittests/test_util.py#L931) in various places. 

However, I think we would either need to get pytest-antilru packaged in distros, or vendor it into the codebase. Any distro that doesn't have antilru as a build-time dependency may bump into caching issues when tests run during build.

Given the permissive license, small project size, and test-only usage, this seems like a possible candidate for vendoring.

+1 if we can somehow use this.

holmanb (commented):
I think that this is cleaner and less error prone than [manually calling cache_clear()](https://github.com/canonical/cloud-init/blob/055395d38a4ec0cc87b005e93110e6d06419f43d/tests/unittests/test_util.py#L931) in various places.

However, I think we would either need to get pytest-antilru packaged in distros, or vendor it into the codebase. Any distro that doesn't have antilru as a build-time dependency may bump into caching issues when tests run during build.

Given the permissive license, small project size, and test-only usage, this seems like a possible candidate for vendoring.

+1 if we can somehow use this.

cjp256:
Do we have any examples of vendoring to follow?  It wouldn't be too bad to write our own fixture either.

holmanb:
> Do we have any examples of vendoring to follow?

None in the codebase that I'm aware of.

> It wouldn't be too bad to write our own fixture either.

+1 I'm in favor of it.

PR #2093, opened 2023-03-23
Standardize NoCloud Command Line Parsing (SC-1457)

holmanb:
```
Standardize kernel commandline user interface

- deprecate ci.ds= and ci.datasource= in favor of ds=
- enable semi-colon-delimited datasource everywhere
- add support for case-insensitive datasource match
- add integration tests
```

## Additional Context
Currently NoCloud's kernel commandline format is an outlier in several of ways. 

```
ds=nocloud[;key=val;key=val]
```

Unique NoCloud behaviors:
1. NoCloud matches the lowercase of the datasource name (`nocloud` [rather than `NoCloud`](https://github.com/canonical/cloud-init/blob/56c88cafd1b3606e814069a79f4ec265fc427c87/cloudinit/sources/__init__.py#LL324))
2. In the case of NoCloudNet, the match is on `nocloud-net`, rather than the dsname (`NoCloud`).
3. It matches on un-namespaced `ds=<name>` rather than `ci.ds=<name>`
4. Semi-colon delimiter is allowed

## Proposal

Don't remove existing behavior, for backwards compatibility, but better standardize a single format and deprecate / undocument the the others formats. NoCloud has several unique values it gets from the commandline, standardize shared the elements (listed above). 

This will allow us to have a standardized interface, which is good for clarity, documentation, adoption, usage, testing, shared code, etc. This way there will be less peculiarities to document.

#### Proposed UI Changes to NoCloud:

1. Add support for the namespaced `ci.ds=` in addition to `ds=`
2. Set `NoCloudNet.dsname` to `NoCloud-Net`

#### Proposed UI Changes to All Other Datasources:
1. Enable semi-colon delimiter
2. Make datasource names case insensitive, document the lower-cased version.
3. Add support for the namespaced `ds=` in addition to `ci.ds=`

Feel free to bikeshed the details, as long as we can find a consistent format that works for all datasources I'll be happy.


cloudinit/sources/__init__.py
```diff
@@ -321,7 +321,7 @@ def override_ds_detect(self) -> bool:
         does not run, _something_ needs to detect the kernel command line
         definition.
         """
-        if self.dsname == parse_cmdline():
+        if self.dsname.lower() == parse_cmdline():
```
blackboxsw, line 324: +1 on standardizing on lower(), but for backward compat, I think we may want to support Case insensitive or at least the slightly more strict `parse_cmdline() in (self.dsname, self.dsname.lower())`

tests/unittests/sources/test___init__.py
```diff
@@ -13,11 +13,13 @@
         "ci.ds=OpenStack",
         "aosiejfoij ci.ds=OpenStack blah",
         "aosiejfoij ci.ds=OpenStack faljskebflk",
+        "ci.ds=OpenStack;",
```
blackboxsw, line 16: Probably want the unit test extension for `ci.ds=openstack;` or something similar if we don't already have it

cloudinit/sources/__init__.py
```diff
@@ -1154,8 +1154,8 @@ def parse_cmdline():
     Passing by command line overrides runtime datasource detection
     """
     cmdline = util.get_cmdline()
-    ds_parse_1 = re.search(r"ci\.ds=([a-zA-Z]+)(\s|$)", cmdline)
-    ds_parse_2 = re.search(r"ci\.datasource=([a-zA-Z]+)(\s|$)", cmdline)
+    ds_parse_1 = re.search(r"ci\.ds=([a-zA-Z]+)(\s|$|;)", cmdline)
```
blackboxsw, line 1157: +1  on supporting that ';` delimiter more generally, it'll give an option to start shoving in datasource specific config options on the cmdline in the future like metadata_urls,url_retries,url_timeout etc that seem to be fairly well used in multiple datasources.

blackboxsw (commented):
This looks like a good generalization for other datasources. I agree with all points here with the exception of migrating away from `ds=nocloud-net`. I realize it's an outlier, but having a two character option for kernel commandline parameter declarations is a feature that it highly valued due to character limited on kernel cmdlines. This short character desire is also what motivated the feature for the nocloud datasource to define [one-character aliases such as h, i and s](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html#permitted-keys) for the more lengthy alternative config option variable names.

holmanb:
> Ideally, I'd vote to see a followup PR (which may be more contentious) that allows a shortened ds=openstack or any other datasouce.

Lets do it in this PR, whichever way we prefer; the code changes are minimal. That said, I'm actually in agreement here (changed my mind from before). If an accidental collision happens, such as `cds=Thriller`, the failure path is `parse_cmdline()` would return `Thriller`, which will harmlessly not match any `dsname`s. We can add an additional test to ensure that this failure path doesn't break with future changes, if we want to be extra safe.

An exact match on a `ds=datasource` would have to happen for an accidental override, and if someone somehow passes `ds=openstack` and _doesn't_ want to force openstack, I'd be fine arguing that such a scenario is user error.

* Note: One more change here, I think we should probably add tests to verify that the last match in the commandline is the one to match (`ds=openstack ds=ec2` should match on ec2) - this is expected of the kernel commandline.

tests/unittests/sources/test___init__.py
holmanb: +1 done

tools/ds-identify
```diff
@@ -836,9 +836,11 @@ dscheck_NoCloud() {
     local fslabel="cidata CIDATA" d=""
     case " ${DI_KERNEL_CMDLINE} " in
         *\ ds=nocloud*) return ${DS_FOUND};;
+        *\ ci.ds=nocloud*) return ${DS_FOUND};;
```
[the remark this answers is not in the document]
blackboxsw: I know ds-identify will early exist with  `debug 1 "datasource '$DI_DSNAME' specified."`                       
 it sees `ci.ds=openstack` and our `cloudinit/sources/__init__py: override_ds_detect` in python will return True for _check_and_get_data. Are there any other elements we need to cover to ensure that the processed cfg_list of potential datasources from `find_source` is limited to just a single entry, instead walking through potentially 20 datasources?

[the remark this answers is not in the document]
holmanb: > so we've added ci.ds support ci.ds=nocloud but aren't we instead missing an element to allow ds-identify to generally handle lowercase datasource name on /proc/cmdline?

Good point, I don't think we need to add this here if we're going to prefer `ds=nocloud` (since it isn't already supported by ds-id), and lowercase match is required since we want parity between ds-id and the Python parsing code, so I'll implement that.

holmanb:
Thanks for the review @blackboxsw. I've updated all tests and added support for case-insensitive datasource matching.

> Do we want something general to perform read_config and handle lowercase versions of the known DataSource names or is this already handled in the python cloudinit/stages during find_source

To implement this it required some additional usage of `importlib`. Formerly there was a 1:1 mapping between possible arguments - supporting lower case means that this is no longer true and now requires an importlib lookup to match the maybe-lowercase dsname against.

> Are there any other elements we need to cover to ensure that the processed cfg_list of potential datasources from find_source is limited to just a single entry, instead walking through potentially 20 datasources?

The override should force the matching datasource from the kernel commandline even after this change. Is that what you are asking?

Ready for re-review.

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -353,10 +353,18 @@ def _merge_new_seed(cur, seeded):
 
 
 class DataSourceNoCloudNet(DataSourceNoCloud):
+    dsname = "NoCloud-Net"
```
holmanb, line 356: Note: this will allow matching on nocloud-net

cloudinit/importer.py
```diff
@@ -38,23 +57,16 @@ def find_module(
     """Finds specified modules"""
     if not required_attrs:
         required_attrs = []
-    # NOTE(harlowja): translate the search paths to include the base name.
```
holmanb, line 41: a `filter` + `join` accomplishes the same thing, and is less verbose and requires less data manipulation - clean it up while we're here

holmanb:
@blackboxsw - Question / proposal:

Is there any reason we shouldn't allow the `seedfrom` value to determine whether we're using NoCloud or NoCloudNet? It feels really awkward that DataSourceNoCloud is a single datasource that requires a unique key to differentiate between `DEP_FILESYSTEM` and `DEP_NETWORK`. No other datasources do this, and from what I can tell it isn't actually required. I don't see a reason why we shouldn't just consolidate on `ds=nocloud` and allow the `seedfrom` to indicate which one we're using. This would give users less characters in the kernel commandline, give users one less UI detail to have to remember, and better align NoCloud with the rest of the datasources with a more standardized user interface and cleaner documentation.

We would have to support the nocloud-net for backwards compatibility, of course.

Thoughts? I'm going to hold off on actually implementing this detail until I've heard some feedback on the idea.

blackboxsw:
> Is there any reason we shouldn't allow the `seedfrom` value to determine whether we're using NoCloud or NoCloudNet? It feels really awkward that DataSourceNoCloud is a single datasource that requires a unique key to differentiate between `DEP_FILESYSTEM` and `DEP_NETWORK`. 

I think that makes sense for nocloud datasource and kernel cmdline params to be simplified to a single alias which would interpret the correct cloud-init boot stage based on seedfrom urls that appear external vs local filesystem paths.

The only potential wrinkle I see would be seedfrom urls that is a file path which may actually be backed by NAS-based filesystems or network storage (maybe NFS, samba, sshfs, glusterfs etc). If we automatically determine protocol http(s)? == network-based url and choose datasource == NoCloudNet, we need to make sure the plumbing of kernel command line options still exists to allow explicit setting of the boot stage. I think the [dsmode logic here](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceNoCloud.py#L286-L287) still allows for explicit setting of dsmode, but we need to test and better document that option in nocloud datasource docs for your followup PR.

cloudinit/sources/__init__.py
```diff
@@ -1149,13 +1150,25 @@ def pkl_load(fname: str) -> Optional[DataSource]:
...
     Passing by command line overrides runtime datasource detection
     """
     cmdline = util.get_cmdline()
-    ds_parse_1 = re.search(r"ci\.ds=([a-zA-Z]+)(\s|$)", cmdline)
-    ds_parse_2 = re.search(r"ci\.datasource=([a-zA-Z]+)(\s|$)", cmdline)
-    ds = ds_parse_1 or ds_parse_2
-    if ds:
+    ds_parse_0 = re.search(r"ds=([a-zA-Z]+)(\s|$|;)", cmdline)
+    ds_parse_1 = re.search(r"ci\.ds=([a-zA-Z]+)(\s|$|;)", cmdline)
+    ds_parse_2 = re.search(r"ci\.datasource=([a-zA-Z]+)(\s|$|;)", cmdline)
```
blackboxsw, line 1160: All three of these searches should only match with a leading space. We always want to lead with cloud-init cmdline params with our ds config first, and follow with other semi-colon delimited modifiers afterward so we don't have to worry about other applicable configuration characters in front of ds or ci.ds.   Also we don't want to accidentally catch any longer kernelcmdline params that show up in the wild that happen to end with `ds=` such as 
        nfs.callback_nr_threads=
        rcutree.rcu_kick_kthreads= [KNL]

Which would inadvertently cause cloud-init a bit of wasted effort matching those values and finding out that there are no datasourse matching "4" or something.

I think a leading space is safe because every kernel param has to be space-separated so our regex shouldn't expect to extract our leading ds= from a substring of other non-whitespace character strings. Also there are typically other leading params provided before we get to cloud-init extensions like `BOOT_IMAGE` and `root` etc.

because that would affect cloud-init's ability to detect a datasource.
```suggestion
    ds_parse_0 = re.search(r" ds=([a-zA-Z]+)(\s|$|;)", cmdline)
    ds_parse_1 = re.search(r" ci\.ds=([a-zA-Z]+)(\s|$|;)", cmdline)
    ds_parse_2 = re.search(r" ci\.datasource=([a-zA-Z]+)(\s|$|;)", cmdline)
```

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -357,6 +358,30 @@ def __init__(self, sys_cfg, distro, paths):
...
+        ds_parse_0 = re.search(r"ds=([a-zA-Z]+)(\s|$)", cmdline)
+        ds_parse_1 = re.search(r"ci\.ds=([a-zA-Z]+)(\s|$)", cmdline)
+        ds_parse_2 = re.search(r"ci\.datasource=([a-zA-Z]+)(\s|$)", cmdline)
+        ds = ds_parse_0 or ds_parse_1 or ds_parse_2
+        deprecated = ds_parse_1 or ds_parse_2
+        if deprecated:
+            util.deprecate(
+                deprecated=(
+                    f"Defining the datasource on the commandline using "
+                    f"ci.ds={ds} or ci.datasource={ds}"
```
blackboxsw, line 377: This is actually printing the re.search results, not the datasource name.
Let's just tell them what they sent. no guesswork needed.
```suggestion
                    f"Defining the datasource on the commandline using "
                    f"{ds.group().strip()}"
```

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -357,6 +358,30 @@ def __init__(self, sys_cfg, distro, paths):
...
+                ),
+                deprecated_version="23.2",
+                extra_message=f"Use ds={ds} instead",
```
blackboxsw, line 380: Again printing re.search results.
```suggestion
                extra_message=f"Use ds={ds.group(1)} instead",
```

blackboxsw (changes requested):
@holmanb excellent work on this. I think it's also worth a minor doc update somewhere 

not sure if it belongs in doc/rtd/explanation/kernel-cmdline.rst or doc/rtd/reference/datasources.rst  or cloudinit/sources/DataSourceNoCloud.py   we can take the doc aspect of this to a separate PR if we want too.

A couple needs fixings that I saw in the meantime

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -357,6 +358,30 @@ def __init__(self, sys_cfg, distro, paths):
...
+            )
+        if ds:
+            return "nocloud-net" == ds.group(1).lower()
+
```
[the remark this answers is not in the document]
holmanb: Oops, nice catch, thanks.

cloudinit/sources/__init__.py
holmanb: > I think a leading space is safe because every kernel param has to be space-separated so our regex shouldn't expect to extract our leading ds= from a substring of other non-whitespace character strings. Also there are typically other leading params provided before we get to cloud-init extensions like BOOT_IMAGE and root etc.

Can we guarantee that ds= will never start at the beginning of `/proc/cmdline`? I see that starting the kernel cmdline with `BOOT_IMAGE` is a [grub behavior](https://git.savannah.gnu.org/cgit/grub.git/commit/?id=16ccb8b138218d56875051d547af84410d18f9aa)[1], however, can we rely on this in all cases? It looks like `initrd=` is a [possible argument](https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html) on the commandline, so it makes me think that maybe the initramfs possibly doesn't need to be the first argument on the kernel commandline, however, I haven't been able to verify this. I went back and forth about a leading whitespace, but was never able to convince myself that this would never happen (this could be violated by different bootloaders, future changes to grub, etc).

> Also we don't want to accidentally catch any longer kernelcmdline params that show up in the wild that happen to end with ds= such as
> nfs.callback_nr_threads=
> rcutree.rcu_kick_kthreads= [KNL]
>
> Which would inadvertently cause cloud-init a bit of wasted effort matching those . values and finding out that there are no datasourse matching "4" or something.

Agreed, unwanted work is never ideal. However, I made the choice in favor of matching ds=name at the front of the commandline since breaking that case would be a problem, if it is possible. I'd rather get the behavior right in all cases rather than an optimization to avoid string matching in a corner case. If we are confident that this will never happen, I'm fine with making this change, but I'd like to push back in favor of safety here. Thoughts @blackboxsw?

[1] the other args, such as `root=` can be trivially reordered in grub config

holmanb:
Really good review @blackboxsw, thanks for the review.

> excellent work on this

Thanks :-)

> I think it's also worth a minor doc update somewhere

Currently the [docs branch](https://github.com/canonical/cloud-init/compare/main...holmanb:cloud-init:holmanb/datasource-docs) I have for [this change](https://warthogs.atlassian.net/browse/SC-1457) consists of mostly TODOs. I've added notes to it on a few different things that need updating, including this change and the OpenStack baremetal documentation. I was hoping to land this one and a PR for [this nocloud idea](https://github.com/canonical/cloud-init/pull/2093#issuecomment-1488969266) prior to landing the docs PR.


> A couple needs fixings that I saw in the meantime
Thanks for the suggestions. I'll take most of them, but I'm not sure that it is safe to add whitespace to the regex match. I've explained my reasoning inline. Please let me know what you think.

cloudinit/sources/__init__.py
blackboxsw: Works for me. let's leave it as you have it

cloudinit/sources/DataSourceNoCloud.py
holmanb: thanks! done

cloudinit/sources/DataSourceNoCloud.py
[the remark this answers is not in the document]
holmanb: done

holmanb:
@blackboxsw I think I've addressed all of your remaining comments in the latest commit.

blackboxsw (approved):
LGTM. Behavior is flexible, original behavior is retained for kernel cmdline handling and improved messaging related to deprecations. LGTM. We'll sort docs changes and any followup as separate PRS.

PR #2098, opened 2023-03-27
schema: do not manipulate draft4 metaschema on jsonschema 2.6.0

blackboxsw:
## Proposed Commit Message
```
    schema: do not manipulate draft4 metaschema for jsonschema 2.6.0
    
    Only set additionalProperties = False in unittest
    draft4 schema definition because cloud-init globally
    registers its draft4 extensions as the primary validator
    for any draft4 schemas in the python process.
    
    This affects solutions such as subiquity and
    ubuntu-desktop-installer which invoke jsonschema.validate
    in the same process at runtime just after calling
    cloudinit.schema.get_jsonschema_validator.
    
    The resulting Tracebacks are seen as something like:
     jsonschema.exceptions.SchemaError:
     {'$ref': '#/definitions/ref_id'} is not valid under any of the
     given schema
    
    Background:
    cloud-init needs to extend draft4 schema to better
    validate and warn 'deprecated' properties in draft4-based
    cloud-init schema definitions. Our unittests also attempt
    to strictly validate any meta schema definitions for the
    cc_* config modules.
    
    To accomplish strict meta schema validation cloud-init makes
    a copy of the draft4 meta schema and adds an
    'additionalProperties' = True to that schema to raise specific
    errors and catch typos in cc_ module schema definitions.
    
    Given that cloud-init at runtime extends and registers
    a draft4 schema validator, any external consumers
    of jsonschema.validate with draft4-base schemas are
    exposed to cloud-init's validator extentions so let's
    limit our risk exposure.
    
    For python 2.6.0, we cannot specify make draft4 schema
    strict because any "$ref" keys are not yet resolved
    to their actual #/defintions/<id> values so the traceback above
    will always be generated in 'strict' mode for complex schemas.
    
    This does not affect jsonschema 3.0+ which appears to resolve
    schema $refs values before schema validation.
```

## Additional Context

python3-jsonschema 2.6.0 in ubuntu-desktop-installer causing probs for cloud-init/curtin https://github.com/canonical/ubuntu-desktop-installer/issues/1714

## Checklist:

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -319,8 +319,7 @@ def get_jsonschema_validator():
     # type jsonschema attributes in cloud-init's schema.
     # This allows #cloud-config to provide valid yaml "content: !!binary | ..."
 
-    strict_metaschema = deepcopy(Draft4Validator.META_SCHEMA)
-    strict_metaschema["additionalProperties"] = False
+    meta_schema = deepcopy(Draft4Validator.META_SCHEMA)
 
     # This additional label allows us to specify a different name
     # than the property key when generating docs.
@@ -328,10 +327,11 @@ def get_jsonschema_validator():
     # otherwise the property label in the generated docs will be a
     # regular expression.
     # http://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties
-    strict_metaschema["properties"]["label"] = {"type": "string"}
+    meta_schema["properties"]["label"] = {"type": "string"}
 
     validator_kwargs = {}
     if hasattr(Draft4Validator, "TYPE_CHECKER"):  # jsonschema 3.0+
+        meta_schema["additionalProperties"] = False  # Unsupported in 2.6.0
         type_checker = Draft4Validator.TYPE_CHECKER.redefine(
             "string", is_schema_byte_string
         )
@@ -339,6 +339,7 @@ def get_jsonschema_validator():
             "type_checker": type_checker,
         }
     else:  # jsonschema 2.6 workaround
+        # pylint:disable-next=no-member
         types = Draft4Validator.DEFAULT_TYPES  # pylint: disable=E1101
         # Allow bytes as well as string (and disable a spurious unsupported
         # assignment-operation pylint warning which appears because this
@@ -354,7 +355,7 @@ def get_jsonschema_validator():
     validators["anyOf"] = _anyOf
 
     cloudinitValidator = create(
-        meta_schema=strict_metaschema,
+        meta_schema=meta_schema,
         validators=validators,
         version="draft4",
         **validator_kwargs,
--- a/tests/unittests/config/test_schema.py
+++ b/tests/unittests/config/test_schema.py
@@ -51,6 +51,7 @@ from tests.unittests.helpers import (
     mock,
     skipUnlessHypothesisJsonSchema,
     skipUnlessJsonSchema,
+    skipUnlessJsonSchemaVersionGreaterThan,
 )
 from tests.unittests.util import FakeDataSource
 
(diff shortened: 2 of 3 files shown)


TheRealFalcon (commented):
Unit tests need to be updated to say `strict_metaschema=True`, correct?

blackboxsw:
> Unit tests need to be updated to say `strict_metaschema=True`, correct?

Nope the unittests already provided that [strict_metaschema = True here](https://github.com/canonical/cloud-init/blob/2f9e958ec12643457b500f494f127249fd1e2291/tests/unittests/config/test_schema.py#L419)   I just plumbed that param into   `get_jsonschema_validator` to finally determine whether to set `additionalProperties = False`

blackboxsw:
rebased for mypy/lint fixes

TheRealFalcon (approved):

holmanb (commented):
@blackboxsw Thanks for this fix.

+1 to get this fix in quick, I guess, but long term I think we need to remove the metaschema manipulation or fix it - it no longer works as designed.

The whole point of the metaschema changes (as implemented in bedac77e934), was to prevent invalid schema entries, such as the various useless `required: []`'s that littered the schema.

Today, the metaschema no longer catches invalid entries. So what value does it add? Does it actually catch anything anymore?

With the following, `test_schema.py` passes, which it shouldn't.
```json
index 2a2d8631c..929f196e9 100644
@@ -696,6 +696,7 @@
         "apk_repos": {
           "type": "object",
           "minProperties": 1,
+          "required": [],
           "properties": {
             "preserve_repositories": {

```

I can't find a way that it helps, and it adds a lot of complexity now, so I think this is a net negative at this point.

holmanb:
> I can't find a way that it helps, and it adds a lot of complexity now, so I think this is a net negative at this point.

I haven't bothered to bisect to find what broke the current metaschema implementation, but much has changed in the codebase wrt schema since that was implemented.

TheRealFalcon:
I was assuming the unit tests would still be catching our  metaschema changes...but that's not true anymore?

aciba90 (commented):
If we are sure that the problem only happens with `jsonschema==2.6.0`, I think we should leave a note explaining it to make it easy to revert/refactor this workaround.

Additionally, a minimized reproducible example would be a nice breadcrumb for future maintainers (it could be just a comment on the PR).

An idea that I think it is be worth investigating is:

If we do not want our schema to conflict with the global draft4 validator, as we explicitly create and consume our custom validator

https://github.com/canonical/cloud-init/blob/e3f1ec3f57cc1f5eacff9506d285007966414481/cloudinit/config/schema.py#L457

and we do not need it to be global, then maybe a possible better solution could be to register it for a custom version in:

https://github.com/canonical/cloud-init/blob/e3f1ec3f57cc1f5eacff9506d285007966414481/cloudinit/config/schema.py#L359

something like:

```python
    cloudinitValidator = create(
        meta_schema=strict_metaschema,
        validators=validators,
        version="cloud-init-draft4",
        **validator_kwargs,
    )
```

could potentially solve the global state issue for `jsonschema==2.6.0` but I did not manage to reproduce it.

Other jsonschema things that we modify globally are:

https://github.com/canonical/cloud-init/blob/e3f1ec3f57cc1f5eacff9506d285007966414481/cloudinit/config/schema.py#L335

https://github.com/canonical/cloud-init/blob/e3f1ec3f57cc1f5eacff9506d285007966414481/cloudinit/config/schema.py#L342

https://github.com/canonical/cloud-init/blob/e3f1ec3f57cc1f5eacff9506d285007966414481/cloudinit/config/schema.py#L350

aciba90:
Thanks @blackboxsw, I do agree, it sounds good to me to try that approach as a follow-up.

PR #2101, opened 2023-03-28
Ubuntu/bionic

TheRealFalcon:
```
new_upstream_snapshot.py
# Merge fails
git mergetool
# Select use deleted file
git commit
new_upstream_snapshot.py --post-stage merge
# Update changelog accordingly
```


debian/changelog
```diff
@@ -8,9 +8,9 @@ cloud-init (23.1-0ubuntu0~18.04.3) UNRELEASED; urgency=medium
     - d/p/openstack-no-network-patch.patch
     - d/p/netplan99-cannot-use-default.patch
     - d/p/retain-netplan-world-readable.patch
-  * Upstream snapshot based on upstream/main at 4b6e4e13.
```
blackboxsw, line 11: Since this previous snapshot wasn't released. Let's just manually redact this line

TheRealFalcon: I already got rid of the old one. Are you saying I should get rid of the new one too?

blackboxsw (commented):
Our daily build recipe merges `ubuntu/bionic` into `main` and tries to build from that. With the branch it doesn't look like we can build from main with the following steps:
```
git checkout upstream/main -B main
git merge james/ubuntu/bionic
build-package
# I'm seeing package errors and quilt patch conflicts 
patching file cloudinit/features.py
patching file tests/unittests/config/test_cc_set_passwords.py
Reversed (or previously applied) patch detected!  Skipping patch.
2 out of 2 hunks ignored
dpkg-source: info: the patch has fuzz which is not allowed, or is malformed
dpkg-source: info: if patch 'expire-on-hashed-users.patch' is correctly applied by quilt, use 'quilt refresh' to update it
dpkg-source: info: if the file is present in the unpacked source, make sure it is also present in the orig tarball
dpkg-source: info: restoring quilt backup files for expire-on-hashed-users.patch
dpkg-source: error: LC_ALL=C patch -t -F 0 -N -p1 -u -V never -E -b -B .pc/expire-on-hashed-users.patch/ --reject-file=- < wtd/debian/patches/expire-on-hashed-users.patch subprocess returned exit status 1
dpkg-buildpackage: error: dpkg-source --before-build . subprocess returned exit status 2
debuild: fatal error at line 1182:
dpkg-buildpackage -d -us -uc -ui -S -nc failed
FAILED: debuild -d -S -nc

```

PR #2103, opened 2023-03-30
azure/errors: introduce reportable errors

cjp256:
When provisioning failures occur an Azure, a generic description is used in the report and ultimately is returned to the user.  To improve the user experience, report details of the failure in a manner that is parsable, readable and succinct.  The current approach is to use csv with a custom delimiter ("|") and quote character ("'").  This format may change in the future.

Gracefully handle reportable errors thrown while crawling metadata and treat other exceptions as ReportableErrorUnhandledException.  Future work will introduce more reportable errors to handle the expected cases.

Note that not all reportable errors are guaranteed to be fatal and that and the datasource may recover after reporting a failure.  However, due to the use of a variable timeout for provisioning, there are cases we will report failure because of an issue likely to affect provisioning within a reasonable timeframe (e.g. DHCP is very delayed, IMDS metadata is unavailable, etc.).

### DIFF
--- a/cloudinit/sources/azure/dmi.py
+++ b/cloudinit/sources/azure/dmi.py
@@ -1,0 +1,36 @@
+# Copyright (C) 2023 Microsoft Corporation.
+#
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import logging
+from typing import Optional
+
+from cloudinit import dmi
+
+LOG = logging.getLogger(__name__)
+
+
+def query_vm_id() -> Optional[str]:
+    system_uuid = dmi.read_dmi_data("system-uuid")
+    LOG.debug("Read product uuid: %s", system_uuid)
+    if not system_uuid:
+        return None
+
+    system_uuid = system_uuid.lower()
+    parts = system_uuid.split("-")
+
+    # Swap endianess for first three parts.
+    for i in [0, 1, 2]:
+        try:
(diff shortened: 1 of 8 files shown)


cloudinit/sources/azure/dmi.py
```diff
@@ -0,0 +1,36 @@
...
+def query_vm_id() -> Optional[str]:
+    system_uuid = dmi.read_dmi_data("system-uuid")
+    LOG.debug("Read product uuid: %s", system_uuid)
+    if not system_uuid:
+        return None
+
+    system_uuid = system_uuid.lower()
+    parts = system_uuid.split("-")
+
+    # Swap endianess for first three parts.
```
anhvoms, line 22: endianness?

cloudinit/sources/azure/errors.py
```diff
@@ -48,10 +50,9 @@ def as_description(
             f"timestamp={self.timestamp.isoformat()}",
         ]
 
-        system_uuid = dmi.read_dmi_data("system-uuid")
-        if system_uuid:
-            system_uuid = system_uuid.lower()
-            error.append(f"vm_id={system_uuid}")
+        vm_id = query_vm_id()
```
anhvoms, line 53: it will be useful to report this via diagnostic KVP, even for the successful case

cjp256: I will add a diag in a separate PR.

PR #2104, opened 2023-03-30
tox.ini: set -vvvv --showlocals for pytest

cjp256:
Test failures may have truncated assertions using `...` in some output.

Increase verbosity and print locals to help with triaging failures caught in CI, particularly as some environments may not be readily testable locally.


holmanb (approved):
Looks like a useful addition. Thanks!

PR #2109, opened 2023-04-03
ubuntu/kinetic: gce drop disable network activation kinetic (SC-1440)

aciba90:
# do not squash

## Test steps

```bash
build-package
sbuild
CLOUD_INIT_PLATFORM=gce CLOUD_INIT_CLOUD_INIT_SOURCE=$(ls cloud-init_*.deb) tox -e integration-tests -- tests/integration_tests/modules/test_combined.py
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -8,16 +8,16 @@ cloud-init (23.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
 
  -- Alberto Contreras <alberto.contreras@canonical.com>  Thu, 09 Mar 2023 11:49:48 +0100
 
-cloud-init (23.1-0ubuntu0~22.10.1) kinetic; urgency=medium
+cloud-init (23.1.1-0ubuntu0~22.10.1) kinetic; urgency=medium
 
   * d/patches/retain-netplan-world-readable.patch:
     - Retain original world-readable perms of /etc/netplan/50-cloud-init.yaml.
       Lunar made the config root read-only.
-  * Upstream snapshot based on 23.1. (LP: #2008230).
+  * Upstream snapshot based on 23.1.1. (LP: #2008230).
     List of changes from upstream can be found at
-    https://raw.githubusercontent.com/canonical/cloud-init/23.1/ChangeLog
+    https://raw.githubusercontent.com/canonical/cloud-init/23.1.1/ChangeLog
 
- -- Chad Smith <chad.smith@canonical.com>  Thu, 23 Feb 2023 22:37:55 -0700
+ -- James Falcon <james.falcon@canonical.com>  Thu, 02 Mar 2023 12:02:34 -0600
 
 cloud-init (22.4.2-0ubuntu0~22.10.1) kinetic; urgency=medium
 
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,12 +1,15 @@
-cloud-init (23.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
+cloud-init (23.1.1-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
 
   * d/apport-general-hook.py: Add general apport hook to append cloud type,
     image and instance size information to bug reports (LP: #1724623)
   * d/cloud-init.preinst: Oracle to remove vestigial /etc/cloud.cloud.cfg.d/
     99-disable-network-config.cfg because system config is now honored before
     datasource config (LP: #1956788)
+  * d/cloud-init.preinst: Clean up vestigial
+    /etc/cloud/cloud.cfg.d/99-disable-network-activation.cfg on GCE instances
+    after fix in upstream google-guest-agent.
 
- -- Alberto Contreras <alberto.contreras@canonical.com>  Thu, 09 Mar 2023 11:49:48 +0100
+ -- Alberto Contreras <alberto.contreras@canonical.com>  Mon, 03 Apr 2023 13:39:38 +0200
 
 cloud-init (23.1.1-0ubuntu0~22.10.1) kinetic; urgency=medium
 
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,12 +1,15 @@
-cloud-init (23.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
+cloud-init (23.1-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
```
blackboxsw, line 1: No Need to bump this d/changelog version number because it's UNRELEASED. We only care to bump this number when it is officially published.

Also this changelog version will cause issues with build-package because our upstream orig.tar.gz is 23.1.1 currently because of a hotfix release. This 23.1 major.minor version won't allow `build-package` script to download the official 23.1.1 upstream and it'll fall back to trying to create a local orig.tar.gz from source with `./tools/make-tarfile`. 

What I think we need to do is the following.:
 1. Bump this d/changelog version number to 23.1.1 based so build-package will properly find and download the upstream cloud-init_23.1.1.orig.tar.gz
```suggestion
cloud-init (23.1.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
```

 2. sync our official released Changelog from `upstream/ubuntu/kinetic-23.1.x` branch 
```diff
diff --git a/ChangeLog b/ChangeLog
index b1157b2b..ecf177d6 100644
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+23.1.1
+ - source: Force OpenStack when it is only option (#2045)
+ - sources/azure: fix regressions in IMDS behavior (#2041)
+   [Chris Patterson]
+
 23.1
  - Support transactional-updates for SUSE based distros (#1997)
    [Robert Schweikert]
```
3. sync our official released debian/changelog from `upstream/ubuntu/kinetic-23.1.x` branch 
```diff
diff --git a/debian/changelog b/debian/changelog
index 46f2f2b8..2cd22f9f 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-cloud-init (23.1-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
+cloud-init (23.1.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
 
   * d/apport-general-hook.py: Add general apport hook to append cloud type,
     image and instance size information to bug reports (LP: #1724623)
@@ -11,16 +11,16 @@ cloud-init (23.1-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
 
  -- Alberto Contreras <alberto.contreras@canonical.com>  Mon, 03 Apr 2023 13:39:38 +0200
 
-cloud-init (23.1-0ubuntu0~22.10.1) kinetic; urgency=medium
+cloud-init (23.1.1-0ubuntu0~22.10.1) kinetic; urgency=medium
 
   * d/patches/retain-netplan-world-readable.patch:
     - Retain original world-readable perms of /etc/netplan/50-cloud-init.yaml.
       Lunar made the config root read-only.
-  * Upstream snapshot based on 23.1. (LP: #2008230).
+  * Upstream snapshot based on 23.1.1. (LP: #2008230).
     List of changes from upstream can be found at
-    https://raw.githubusercontent.com/canonical/cloud-init/23.1/ChangeLog
+    https://raw.githubusercontent.com/canonical/cloud-init/23.1.1/ChangeLog
 
- -- Chad Smith <chad.smith@canonical.com>  Thu, 23 Feb 2023 22:37:55 -0700
+ -- James Falcon <james.falcon@canonical.com>  Thu, 02 Mar 2023 12:02:34 -0600
 
 cloud-init (22.4.2-0ubuntu0~22.10.1) kinetic; urgency=medium
 
```


What also may be helpful is to use this [WIP branch of uss-tableflip which tells build-package to go download the upstream orig.tar.gz for you if you don't already have it locally]( https://github.com/blackboxsw/uss-tableflip/tree/build-package-fixes).

blackboxsw: If you're using the uss-tableflip WIP branch, you should be able to run build-package from ubuntu/kinetic branch and it'll download and verify upstream vs local orig.tar.gz. What we should be expecting is a `cloud-init_23.1.1.orig.tar.gz` from upstream not `23.1`. Note that our current unreleased debian/changelog version is 'wrong' and only represents `23.1` instead of `23.1.1` which breaks build-package assumptions when trying to find an upstream tarball as we never officially released a `23.1` orig.tar.gz.

blackboxsw (changes requested):
We can't build-package this PR as-is because we left the wrong version number `23.1` instead of `23.1.1` in the UNRELEASED changelog entries once we hotfixed a `23.1.1` release in the `upstream/ubuntu/kinetic-23.1.x` branch. Let's sync that Changelog and debian/changelog now (and drop the previous incorrect top-most kinetic changelog entry because it wasn't what we officially released)

aciba90:
Thanks for the detailed explanation! `d/changelog` and `Changelog` synched with `ubuntu/kinetic-23.1.x`.

blackboxsw (approved):
LGTM! land without squash merge looks good here.

debian/changelog
```diff
@@ -1,23 +1,26 @@
-cloud-init (23.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
+cloud-init (23.1.1-0ubuntu1~22.10.1) UNRELEASED; urgency=medium
```
blackboxsw, line 1: Although it doesn't really matter as we will replace this UNRELEASED version before actual SRU upload anyway with somethng like 23.1.2-0ubuntu-~22.10.1), we probably should at least use a version that is at least > the previous version in this changelog `23.1.1-0ubuntu1~22.10.1`

```suggestion
cloud-init (23.1.1-0ubuntu1~22.10.2) UNRELEASED; urgency=medium
```

PR #2112, opened 2023-04-03
Standardize kernel commandline disable

holmanb:
```
Standardize disabling cloud-init on non-systemd

Some distros support disabling cloud-init using the 
kernel argument cloud-init=disabled. Standardize it
across non-systemd distros. Skip NetBSD, which
doesn't appear to support passing external
arguments to the kernel.

Add support for disabling cloud-init using
/etc/cloud/cloud-init.disabled to non-systemd
distros.
```

## Additional Context
Support disabling cloud-init via kernel commandline using `cloud-init=disabled`

Note: suse and rhel shouldn't need these changes per #2115 and #2114

Docs to follow in another PR.

### DIFF
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -18,7 +18,7 @@ start_cmd="cloudconfig_start"
 cloudconfig_start()
 {
 	kenv -q kernel_options | grep -q 'cloud-init=disabled' \
-      || -e /etc/cloud/cloud-init.disabled \
+      || test -e /etc/cloud/cloud-init.disabled \
       && exit 0
 	echo "${command} starting"
 	${command} modules --mode config
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -17,6 +17,7 @@ start_cmd="cloudconfig_start"
 
 cloudconfig_start()
 {
+	kenv -q kernel_options | grep 'cloud-init=disabled' && exit 0
 	echo "${command} starting"
 	${command} modules --mode config
 }
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -18,8 +18,8 @@ start_cmd="cloudconfig_start"
 cloudconfig_start()
 {
 	kenv -q kernel_options | grep -q 'cloud-init=disabled' \
-      || test -e /etc/cloud/cloud-init.disabled \
-      && exit 0
+		|| test -e /etc/cloud/cloud-init.disabled \
+		&& exit 0
 	echo "${command} starting"
 	${command} modules --mode config
 }
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -17,7 +17,9 @@ start_cmd="cloudconfig_start"
 
 cloudconfig_start()
 {
-	kenv -q kernel_options | grep -q 'cloud-init=disabled' && exit 0
+	kenv -q kernel_options | grep -q 'cloud-init=disabled' \
+      || -e /etc/cloud/cloud-init.disabled \
+      && exit 0
 	echo "${command} starting"
 	${command} modules --mode config
 }
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -17,7 +17,7 @@ start_cmd="cloudconfig_start"
 
 cloudconfig_start()
 {
-	kenv -q kernel_options | grep 'cloud-init=disabled' && exit 0
+	kenv -q kernel_options | grep -q 'cloud-init=disabled' && exit 0
 	echo "${command} starting"
 	${command} modules --mode config
 }
--- a/sysvinit/freebsd/cloudconfig
+++ b/sysvinit/freebsd/cloudconfig
@@ -17,11 +17,14 @@ start_cmd="cloudconfig_start"
 
 cloudconfig_start()
 {
-	kenv -q kernel_options | grep -q 'cloud-init=disabled' \
-		|| test -e /etc/cloud/cloud-init.disabled \
-		&& exit 0
-	echo "${command} starting"
-	${command} modules --mode config
+    echo "${command} starting"
+    if kenv -q kernel_options | grep -q 'cloud-init=disabled'; then
+        warn "cloud-init is disabled via kernel_options."
+    elif test -e /etc/cloud/cloud-init.disabled; then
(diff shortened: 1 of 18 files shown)


sysvinit/freebsd/cloudconfig
```diff
@@ -17,6 +17,7 @@ start_cmd="cloudconfig_start"
 
 cloudconfig_start()
 {
+	grep 'cloud-init=disabled' /proc/cmdline && exit 0
```
igalic, line 20: given that we don't have /proc/cmdline on FreeBSD, this should be:

```
kenv -q kernel_options | grep -q 'cloud-init=disabled' && exit 0
```

igalic (commented):
given the high chance of failure, i would `grep -q`

in the slim chance that this succeeds, maybe a warning of message should be issued that cloud-init has been disabled on the kernel command line

holmanb:
> in the slim chance that this succeeds, maybe a warning of message should be issued that cloud-init has been disabled on the kernel command line

Agreed. Do you think it would make more sense to do that in each of the 4 services? Or would it make more sense to only print that once?

holmanb:
@dermotbradley @prometheanfire If you wouldn't mind reviewing the change to the gentoo sysvinit scripts (which are also used by Alpine) I'd much appreciate it.

I tried looking for the best place to drop a log about skipping cloud-init startup, but didn't see how to do that on OpenRC. I suppose we could always just `echo "Cloud-init is disabled via /proc/cmdline." > /var/log/cloud-init.log` or similar. Thoughts?

sysvinit/freebsd/cloudinitlocal
```diff
@@ -18,6 +18,11 @@ start_cmd="cloudlocal_start"
 cloudlocal_start()
 {
 	echo -n "${command} starting"
+	if kenv -q kernel_options | grep 'cloud-init=disabled'; then
+		echo "cloud-init is disabled via /proc/cmdline."
```
igalic, line 22: instead of echo, this should be "warn"

and i should change the other echos in these files to be "info"

igalic: Also i would say kernel_options instead of /proc/cmdline here

igalic (commented):
looks almost prefect

BTW, what's the reason you're not using "grep -q" instead of grep?

sysvinit/freebsd/cloudinitlocal
holmanb: > instead of echo, this should be "warn"

does this apply on the netbsd side too?

sysvinit/gentoo/cloud-config
```diff
@@ -8,6 +8,7 @@ depend() {
 }
 
 start() {
+  grep 'cloud-init=disabled' /proc/cmdline && exit 0
```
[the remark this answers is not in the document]
holmanb: Thanks for the review!

sysvinit/netbsd/cloudconfig
```diff
@@ -10,6 +10,7 @@ name="cloudinit"
 start_cmd="start_cloud_init"
 start_cloud_init()
 {
+    kenv -q kernel_options | grep -q 'cloud-init=disabled' && exit 0
```
igalic, line 13: this won't work on NetBSD. there's no kenv, and no way to pass or read out kernel options

igalic (changes requested):
we can nix the NetBSD changes again

sysvinit/freebsd/cloudinitlocal
igalic: warn, yes.
kenv and kernel_options: nope

igalic (approved):
👍 from FreeBSD

holmanb:
Thanks for the review @dermotbradley.

> For the gentoo/openrc scripts can they use the following instead?

Nice suggestion - done.

> Secondly, don't you also need to make related changes to systemd/cloud-init-hotplugd.service as well so it doesn't run when cloud-init is disabled?

Good point, that probably makes sense.

> BTW I have a openrc equivalent of cloud-init-hotplugd.service in Alpine using [cloud-init-hotplugd.initd](https://git.alpinelinux.org/aports/tree/community/cloud-init/cloud-init-hotplugd.initd) and [cloud-init-hotplugd](https://git.alpinelinux.org/aports/tree/community/cloud-init/cloud-init-hotplugd), which perhaps should be upstreamed?

+1 I'm in favor

holmanb:
I'm going to increase scope on this one a little bit and also introduce the disable file to non-systemd while we're making this change.

holmanb:
@dermotbradley @prometheanfire @igalic Sorry for the re-ping on this PR, and thanks for the initial reviews.

I've updated this PR to additionally include support for `/etc/cloud/cloud-init.disabled` as an on-image file to disable cloud-init. This should bring freebsd/alpine/gentoo/debian/netbsd up to parity with systemd-based cloud-init (minus netbsd, which doesn't support kernel commandline arguments).

sysvinit/netbsd/cloudinitlocal
```diff
@@ -11,6 +11,9 @@ name="cloudinitlocal"
 start_cmd="start_cloud_init_local"
 start_cloud_init_local()
 {
+    test -e /etc/cloud/cloud-init.disabled \
+      && warn "cloud-init disabled by cloud-init.disabled file" \
```
TheRealFalcon, line 15: Is there a reason to warn on this file but not the other 3 netbsd files?

TheRealFalcon (approved):
Nits inline but LGTM!

sysvinit/netbsd/cloudinitlocal
holmanb: In my mind once per boot was enough to warn that cloud-init is disabled. I guess since this PR doesn't standardize on that I can add it into the other 3.

sysvinit/freebsd/cloudinitlocal
```diff
@@ -17,8 +17,14 @@ start_cmd="cloudlocal_start"
 
 cloudlocal_start()
 {
-	echo -n "${command} starting"
-	${command} init --local
+    echo -n "${command} starting"
+    if kenv -q kernel_options | grep -q 'cloud-init=disabled'; then
+        warn "cloud-init is disabled via kernel_options."
+	elif test -e /etc/cloud/cloud-init.disabled; then
```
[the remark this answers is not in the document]
holmanb: fixed

sysvinit/netbsd/cloudinitlocal
holmanb: fixed (both netbsd and freebsd)

PR #2114, opened 2023-04-04
rhel: Remove sysvinit files.

holmanb:
```
rhel: Remove sysvinit files.
```

## Additional Context
In https://github.com/canonical/cloud-init/pull/2112, we would like to add kernel commandline disable of cloud-init to the non-systemd distros. It is believed that supported RHEL-family distros no longer use sysvinit, and can therefore be removed.


holmanb:
We could just as easily add support to these files in #2112 if these files are still used by anything. That said, if not used we should just remove them.

I suspect this is safe; the last known sysvinit release was RHEL 6 and even RHEL 7 doesn't get cloud-init updates anymore.

@Conan-Kudo could you please review?

Conan-Kudo:
Red Hat / Fedora doesn't use them, but I believe a number of offshoots of the Red Hat family do use them: ALT Linux and PLD Linux are both in the greater Red Hat family (through Mandrake) and still offer SysVinit mode.

That said, I'm fine with them being removed and only having them re-added if someone cares.

Conan-Kudo (approved):

holmanb:
Thanks for the review @Conan-Kudo.

>  ALT Linux and PLD Linux are both in the greater Red Hat family (through Mandrake) and still offer SysVinit mode

Since Mandrake

PR #2115, opened 2023-04-04
suse: Remove sysvinit files. 

holmanb:
suse: Remove sysvinit files.

Additional Context

In https://github.com/canonical/cloud-init/pull/2112, we would like to add kernel commandline disable of cloud-init to the non-systemd distros. It is believed that supported SUSE-family distros no longer use sysvinit, and can therefore be removed.


holmanb:
We could just as easily add support to these files in https://github.com/canonical/cloud-init/pull/2112 if these files are still used by anything. That said, if not used we should just remove them.

I suspect this is safe; the last known sysvinit release was SLES 11.

@rjschwei could you please review?

rjschwei (approved):

aciba90 (approved):
LGTM, thanks!

PR #2121, opened 2023-04-11
azure/imds: retry fetching metadata up to 300 seconds

cjp256:
Instead of a fixed number of retries, allow up to 5 minutes to fetch metadata from IMDS.  The current approach allows for up to 11 attempts depending on the path.  Given the timeout setting, this can vary from ~11 seconds up to ~32 seconds depending on whether or not read/connection timeouts are encountered.

Delaying boot on the rare occasion that IMDS is delayed is better than ignoring the metadata as it ensures the VM is configured as expected.

This is a very conservative timeout and may be reduced in the future.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -673,10 +673,10 @@ class DataSourceAzure(sources.DataSource):
 
     @azure_ds_telemetry_reporter
     def get_metadata_from_imds(self) -> Dict:
-        retry_timeout = time() + 300
+        retry_deadline = time() + 300
         try:
             return imds.fetch_metadata_with_api_fallback(
-                retry_timeout=retry_timeout
+                retry_deadline=retry_deadline
             )
         except (UrlError, ValueError) as error:
             report_diagnostic_event(
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -672,9 +672,12 @@ class DataSourceAzure(sources.DataSource):
         return crawled_data
 
     @azure_ds_telemetry_reporter
-    def get_metadata_from_imds(self, retries: int = 10) -> Dict:
+    def get_metadata_from_imds(self) -> Dict:
+        retry_timeout = time() + 300
         try:
-            return imds.fetch_metadata_with_api_fallback(retries=retries)
+            return imds.fetch_metadata_with_api_fallback(
+                retry_timeout=retry_timeout
+            )
         except (UrlError, ValueError) as error:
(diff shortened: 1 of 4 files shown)


cloudinit/sources/azure/imds.py
```diff
@@ -46,7 +49,10 @@ def exception_callback(self, req_args, exception) -> bool:
             return False
 
         log = True
-        retry = True
+        if self.retry_timeout is not None and time() >= self.retry_timeout:
```
TheRealFalcon, line 52: Won't this always return True? `time()` returns seconds since epoch, but `self.retry_timeout` will be something like 300.

TheRealFalcon (changes requested):
I think the `time()` call needs to be changed, otherwise LGTM.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -675,9 +675,12 @@ def crawl_metadata(self):
         return crawled_data
 
     @azure_ds_telemetry_reporter
-    def get_metadata_from_imds(self, retries: int = 10) -> Dict:
+    def get_metadata_from_imds(self) -> Dict:
```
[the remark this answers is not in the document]
cjp256: Yeah, it's effectively the greater of the two.  The reason the one path was set to 300 was because it was treated as a hard provisioning failure and errors were apparent.  300 retries was long enough to address the tail issues and now we're just regrouping and treating both paths the same (300 retries @ 1 second sleep is in the ballpark of 300 seconds).

cloudinit/sources/azure/imds.py
cjp256: I don't think so? If you look at the source of retry_timeout in get_metadata_from_imds(), it's configured to be `retry_timeout = time() + 300`

cjp256:
> I think the `time()` call needs to be changed, otherwise LGTM.

As in `retry_timeout = 300` instead of `retry_timeout = time() + 300`.  I probably should have used "deadline" instead of "timeout".

TheRealFalcon:
Sorry, as I commented inline I missed that line when reviewing. I do think the name `retry_timeout` could be named differently in `imds.py` as I think timeout is usually expressed as number of seconds to wait rather than the unix time to wait until.

cjp256:
> Sorry, as I commented inline I missed that line when reviewing. I do think the name `retry_timeout` could be named differently in `imds.py` as I think timeout is usually expressed as number of seconds to wait rather than the unix time to wait until.

Agreed, done.  Added some pydocs too.

TheRealFalcon (approved):
LGTM, thanks!

PR #2122, opened 2023-04-13
[1/2] DHCP: Refactor dhcp client code 

holmanb:
```
Move isc-dhclient code to dhcp.py

In support of the upcoming deprecation of
isc-dhcp-client, this code refactors current
dhcp code into classes in dhcp.py. The
primary user-visible change should be the
addition of the following log:

dhcp.py[DEBUG]: DHCP client selected: dhclient

This code lays groundwork to enable
alternate implementations to live side by
side in the codebase to be selected with
distro-defined priority fallback. Note that
maybe_perform_dhcp_discovery() now selects
which dhcp client to call, and then runs the
corresponding client's dhcp_discovery()
method. Currently only class IscDhclient is
implemented, however a yet-to-be-implemented
class Dhcpcd exists to test fallback behavior
and this will be implemented in part two of
this series.

Part of this refactor includes shifting
dhclient service management from hardcoded
calls to the distro-defined manage_service()
method in the *BSDs. Future work is required
in this area to support multiple clients via
select_dhcp_client().
```

## Additional Context
Reviewers may find this pull request easiest to review commit by commit rather than all at once, since this change spans several files. Each commit is logically grouped - code changes and corresponding test updates share commits.

## Test Steps
Boot cloud-init on a datasource that uses ephemeral network setup. 

```bash
tox -e integration-test -- tests/integration_tests/test_logging.py
```

New Log:
```
2023-04-13 18:00:28,773 - util.py[DEBUG]: Read 5 bytes from /sys/class/net/ens5/operstate
2023-04-13 18:00:28,773 - dhcp.py[DEBUG]: DHCP client selected: dhclient                                                                                                                  
2023-04-13 18:00:28,773 - dhcp.py[DEBUG]: Performing a dhcp discovery on ens5  
```

[Full log](https://github.com/canonical/cloud-init/files/11225377/cloud-init.log)

## Checklist:

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -32,7 +32,7 @@ from cloudinit import (
 from cloudinit.distros.networking import LinuxNetworking, Networking
 from cloudinit.distros.parsers import hosts
 from cloudinit.features import ALLOW_EC2_MIRRORS_ON_NON_AWS_INSTANCE_TYPES
-from cloudinit.net import activators, eni, network_state, renderers
+from cloudinit.net import activators, eni, network_state, renderers, dhcp
 from cloudinit.net.network_state import parse_net_config_data
 from cloudinit.net.renderer import Renderer
 
@@ -110,12 +110,14 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     resolve_conf_fn = "/etc/resolv.conf"
 
     osfamily: str
+    dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd]
 
     def __init__(self, name, cfg, paths):
         self._paths = paths
         self._cfg = cfg
         self.name = name
         self.networking: Networking = self.networking_cls()
+        self.dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd]
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         """Perform deserialization fixes for Distro."""
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -32,7 +32,7 @@ from cloudinit import (
 from cloudinit.distros.networking import LinuxNetworking, Networking
 from cloudinit.distros.parsers import hosts
 from cloudinit.features import ALLOW_EC2_MIRRORS_ON_NON_AWS_INSTANCE_TYPES
-from cloudinit.net import activators, eni, network_state, renderers, dhcp
+from cloudinit.net import activators, dhcp, eni, network_state, renderers
 from cloudinit.net.network_state import parse_net_config_data
 from cloudinit.net.renderer import Renderer
 
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -185,7 +185,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         self._write_hostname(writeable_hostname, self.hostname_conf_fn)
         self._apply_hostname(writeable_hostname)
 
-    def uses_systemd(self):
+    @staticmethod
+    def uses_systemd():
         """Wrapper to report whether this distro uses systemd or sysvinit."""
         return uses_systemd()
 
@@ -916,15 +917,16 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             args.append(message)
         return args
 
-    def manage_service(self, action: str, service: str):
+    @classmethod
+    def manage_service(cls, action: str, service: str, rcs=None):
         """
         Perform the requested action on a service. This handles the common
         'systemctl' and 'service' cases and may be overridden in subclasses
         as necessary.
         May raise ProcessExecutionError
         """
-        init_cmd = self.init_cmd
-        if self.uses_systemd() or "systemctl" in init_cmd:
+        init_cmd = cls.init_cmd
+        if cls.uses_systemd() or "systemctl" in init_cmd:
             init_cmd = ["systemctl"]
             cmds = {
                 "stop": ["stop", service],
@@ -948,7 +950,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 "status": [service, "status"],
             }
         cmd = list(init_cmd) + list(cmds[action])
-        return subp.subp(cmd, capture=True)
+        return subp.subp(cmd, capture=True, rcs=rcs)
 
     def set_keymap(self, layout, model, variant, options):
         if self.uses_systemd():
@@ -1193,6 +1195,3 @@ def uses_systemd():
         return stat.S_ISDIR(res.st_mode)
     except Exception:
         return False
-
-
-# vi: ts=4 expandtab
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -920,7 +920,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         return args
 
     @classmethod
-    def manage_service(cls, action: str, service: str, rcs=None):
+    def manage_service(
+        cls, action: str, service: str, *extra_args: str, rcs=None
+    ):
         """
         Perform the requested action on a service. This handles the common
         'systemctl' and 'service' cases and may be overridden in subclasses
--- a/cloudinit/distros/openbsd.py
+++ b/cloudinit/distros/openbsd.py
@@ -26,7 +26,7 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
         return ["usermod", "-G", group_name, member_name]
 
     @classmethod
-    def manage_service(cls, action: str, service: str, rcs=None):
+    def manage_service(cls, action: str, service: str, *extra_args, rcs=None):
         """
         Perform the requested action on a service. This handles OpenBSD's
         'rcctl'.
--- a/cloudinit/distros/openbsd.py
+++ b/cloudinit/distros/openbsd.py
@@ -25,13 +25,14 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
     def _get_add_member_to_group_cmd(self, member_name, group_name):
         return ["usermod", "-G", group_name, member_name]
 
-    def manage_service(self, action: str, service: str, rcs=None):
+    @classmethod
+    def manage_service(cls, action: str, service: str, rcs=None):
         """
         Perform the requested action on a service. This handles OpenBSD's
         'rcctl'.
         May raise ProcessExecutionError
         """
-        init_cmd = self.init_cmd
+        init_cmd = cls.init_cmd
         cmds = {
             "stop": ["stop", service],
             "start": ["start", service],
--- a/cloudinit/distros/openbsd.py
+++ b/cloudinit/distros/openbsd.py
@@ -25,7 +25,7 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
     def _get_add_member_to_group_cmd(self, member_name, group_name):
         return ["usermod", "-G", group_name, member_name]
 
-    def manage_service(self, action: str, service: str):
+    def manage_service(self, action: str, service: str, rcs=None):
         """
         Perform the requested action on a service. This handles OpenBSD's
         'rcctl'.
@@ -43,7 +43,7 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
             "status": ["check", service],
         }
         cmd = list(init_cmd) + list(cmds[action])
-        return subp.subp(cmd, capture=True)
+        return subp.subp(cmd, capture=True, rcs=rcs)
 
     def lock_passwd(self, name):
         try:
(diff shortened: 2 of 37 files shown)


igalic (commented):
first look

cloudinit/sources/DataSourceVultr.py
```diff
@@ -136,7 +137,16 @@ def get_datasource_list(depends):
         print("Machine is not a Vultr instance")
         sys.exit(1)
 
+    # It should probably be safe to try to detect distro via stages.Init(),
```
holmanb, line 140: @eb3095 I'm not quite sure what `__main__` gets used for, or if it is even still used at all. Would you mind looking at this section if you get a chance?

Basically piping distros throughout the dhcp and ephemeral setup code means that `vultr.get_metadata()` will now require a distro object. This approach uses the same strategy that cloud-init normally uses. Will that work here? Based on the `is_vultr()` check above, I assume that this should be running on an image with cloud-init. Therefore, I assume that this approach would work, but I want to double check my assumption.

cloudinit/net/dhcp.py
```diff
@@ -267,101 +132,366 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
...
+        # to avoid false positive in wait_for_files
+        with contextlib.suppress(FileNotFoundError):
+            os.remove(pid_file)
+            os.remove(lease_file)
+
+        # ISC dhclient needs the interface up to send initial discovery packets
+        # Generally dhclient relies on dhclient-script PREINIT action to bring
+        # the link up before attempting discovery. Since we are using
+        # -sf /bin/true, we need to do that "link up" ourselves first.
+        subp.subp(["ip", "link", "set", "dev", interface, "up"], capture=True)
```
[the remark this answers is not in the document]
holmanb: Good point. This is how it works currently on main, but you're right, this isn't going to work without iproute2. I think that means that much of `ephemeral.py` needs to be reworked too (which should finally be reasonable to do now that we're getting distros plumbed through that module).

cloudinit/net/dhcp.py
```diff
@@ -267,101 +132,366 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
...
+            self.dhclient_path,
+            "-1",
+            "-v",
+            "-lf",
+            lease_file,
+            "-pf",
+            pid_file,
+            interface,
+            "-sf",
+            "/bin/true",
```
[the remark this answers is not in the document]
holmanb: I'll propose something that uses the equivalents available in freebsd

cloudinit/net/dhcp.py
```diff
@@ -267,101 +132,366 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
...
-    return static_routes
+            static_routes.append(
+                ("%s/%s" % (net_address, net_length), gateway)
+            )
 
+        return static_routes
+
+    @staticmethod
+    def get_dhclient_d():
+        # find lease files directory
```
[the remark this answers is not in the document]
holmanb: Good question. This is basically a direct copy/paste from DataSourceCloudStack, but modified to be in a class.

From the git history I see that:

- [6 years ago](https://github.com/canonical/cloud-init/commit/33816e96d8981918f734dab3ee1a967bce85451a) the network manager directory was added. I suspect that this is no longer required, since recently when digging around in upstream networkmanager code I recall seeing that isc-dhclient was dropped in favor of an internal implementation. A hasty grep of my local (ubuntu) NetworkManager-based system sees nothing that matches `dhcp-server-identifier`, which is the whole point of this function.
- [12 years ago](https://github.com/canonical/cloud-init/commit/4fde399a38765fa9641b3177b966ad6c8ec9750f) Ubuntu support was added, Fedora was already supported

I'm a bit hesitant to split this up, however, because of the risk of accidentally breaking one distro or another by making this more exact. Unless someone wants to audit all of the distros, I'd be more comfortable leaving it as is for now.

cloudinit/net/dhcp.py
```diff
@@ -267,101 +132,366 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
...
+
+        # lease files are named inconsistently across distros.
+        # We assume that 'dhclient6' indicates ipv6 and ignore it.
+        # ubuntu:
+        #   dhclient.<iface>.leases, dhclient.leases, dhclient6.leases
+        # centos6:
+        #   dhclient-<iface>.leases, dhclient6.leases
+        # centos7: ('--' is not a typo)
+        #   dhclient--<iface>.lease, dhclient6.leases
+        for fname in lease_files:
```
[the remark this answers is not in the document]
holmanb: It could be, but I'm not going to prioritize it right now. Right now my focus is getting functionality correct without risking breakage of things that currently work.

cloudinit/sources/DataSourceVultr.py
eb3095: Yeah this should work fine. __main__ i believe was an early discussion about how to debug. I dont think we have ever really used it, but that was the intended purpose.

eb3095 (approved):

cloudinit/net/dhcp.py
```diff
@@ -364,4 +369,86 @@ def _trunc_error(cidr, required, remain):
...
+
+        abs_path = os.path.join(lease_d, fname)
+        mtime = os.path.getmtime(abs_path)
+        if mtime > latest_mtime:
+            latest_mtime = mtime
+            latest_file = abs_path
+    return latest_file
+
+
+def parse_dhcp_server_from_lease_file(lease_file):
```
blackboxsw: Is there opportunity to leverage `parse_dhcp_lease_file` here? 

We could also speed up this code potentially by reversing the order of leases processed to break on the first case of `dhcp-server-identifier` instead of processing the whole file and keeping the last case. It may not matter much in practice as the lease file processed in most cases may only have 1 lease in it though.

cloudinit/distros/openbsd.py
```diff
@@ -25,7 +25,7 @@ def _write_hostname(self, hostname, filename):
     def _get_add_member_to_group_cmd(self, member_name, group_name):
         return ["usermod", "-G", group_name, member_name]
 
-    def manage_service(self, action: str, service: str):
+    def manage_service(self, action: str, service: str, rcs=None):
```
blackboxsw: Forgot the `@classmethod`

cloudinit/net/freebsd.py
```diff
@@ -50,10 +51,8 @@ def start_services(self, run=False):
         for dhcp_interface in self.dhcp_interfaces():
             # Observed on DragonFlyBSD 6. If we use the "restart" parameter,
             # the routes are not recreated.
-            subp.subp(
-                ["service", "dhclient", "stop", dhcp_interface],
-                rcs=[0, 1],
-                capture=True,
+            net.dhcp.IscDhclient.start_service(
+                dhcp_interface, distros.freebsd.Distro
             )
```
blackboxsw, line 56: I don't think we can make this shift as-written. These definitions are no longer the same, the new `IscDhclient.start_service` method completely ignores the dhcp_interface because we are using `distro.manage_interface` and dropping the supplemental `dhcp_interface` argument in the refactor.

Instead of running:
`service dhclient stop eth0` 
we now run only the following on FreeBSD
`service dhclient stop`

cloudinit/net/dhcp.py
```diff
@@ -267,101 +131,366 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
...
+        self.dhclient_path = subp.which("dhclient")
+        if not self.dhclient_path:
+            LOG.debug(
+                "Skip dhclient configuration: No dhclient command found."
+            )
+            raise NoDHCPLeaseMissingDhclientError()
+
+    @staticmethod
+    def parse_dhcp_lease_file(lease_file):
+        """Parse the given dhcp lease file for the most recent lease.
```
blackboxsw, line 170: This returns a list of all leases. Let's use type annotations on the function definitions here to guide our expectations about params and return types
And let's also either:
* fix the logic to return only one lease and correct the one call-site in ephemeral.py to assume most recent lease
-- or --
* Fix the docstr header

blackboxsw (commented):
Quick first pass. Wanted to get my BSD manage_service question out there as it may involve a bit of refactor to retain previous behavior. I'll continue the review today.

cloudinit/distros/openbsd.py
holmanb: I see it in the latest version of this branch. I'm guessing this got fixed in a later commit than the one that you reviewed.

cloudinit/net/dhcp.py
holmanb: > Is there opportunity to leverage parse_dhcp_lease_file here?

It looks like it. This code came from a different file and looks like a case of partial code duplication.

> We could also speed up this code potentially...

Agreed. I'd like for that to be a separate PR for a cleaner git history, since it doesn't fit the scope of this change.

cloudinit/net/freebsd.py
holmanb: Nice catch, thanks. This should be fixed in [f1eb8ac](https://github.com/canonical/cloud-init/pull/2122/commits/f1eb8acc98b0995cfb7f00c0c24adfa7f951b604).

cloudinit/net/dhcp.py
holmanb: fixed in 5cd048f6311ae4c509d890d3dfb0320913b6252c

holmanb:
> Quick first pass. Wanted to get my BSD manage_service question out there as it may involve a bit of refactor to retain previous behavior. I'll continue the review today.

Thanks for the initial review @blackboxsw. I think that I have addressed all of your comments. 

Additionally I just pushed up the refactor (commit 9d16846c2b8061bf5c6bd06f413d9419ce729101) that makes network ops in ephemeral.py distro-agnostic (cc @igalic).

cloudinit/net/dhcp.py
```diff
@@ -167,7 +168,7 @@ def __init__(self):
             raise NoDHCPLeaseMissingDhclientError()
 
     @staticmethod
-    def parse_dhcp_lease_file(lease_file):
+    def parse_dhcp_lease_file(lease_file: str) -> list:
```
blackboxsw, line 171: I think we may want to `from typing import Any, Dict, List` and have to following:
```suggestion
    def parse_dhcp_lease_file(lease_file: str) -> List[Dict[str, Any]]:
```

cloudinit/distros/__init__.py
```diff
@@ -104,18 +111,22 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     # This is used by self.shutdown_command(), and can be overridden in
     # subclasses
     shutdown_options_map = {"halt": "-H", "poweroff": "-P", "reboot": "-r"}
+    net_ops = iproute2
```
blackboxsw, line 114: This net_ops work feels like a bolt-on ideal that represents a bit of scope creep for this PR. Would it be okay to separate this to another PR to ease review? Otherwise, we can spend a bit more time going through this in this PR but I worry a bit about missing something here in the refactor dance.

cloudinit/net/freebsd.py
```diff
@@ -50,10 +51,8 @@ def start_services(self, run=False):
         for dhcp_interface in self.dhcp_interfaces():
             # Observed on DragonFlyBSD 6. If we use the "restart" parameter,
             # the routes are not recreated.
-            subp.subp(
-                ["service", "dhclient", "stop", dhcp_interface],
-                rcs=[0, 1],
-                capture=True,
+            net.dhcp.IscDhclient.start_service(
```
blackboxsw, line 54: @holmanb it looks like this PR unintentionally inverts the operations, the intent here was to stop any existing dhcp services on any dhcp interfaces before starting them again (instead of just using the systemctl restart operation). But it looks like we are now starting and then stopping. Let's flip them back.

blackboxsw (changes requested):
Looks good @holman. Just to make sure I'm not lost in refactor intents, if we can separate net_ops work from the IscDhclient work it'd make me more certain we aren't missing something in review.

cloudinit/distros/__init__.py
holmanb: That's fair. I'll turn this into a 3 part series.

cloudinit/net/dhcp.py
holmanb: Sorry, I must have accidentally dropped that in a rebase. Thanks!

holmanb:
> Looks good @holman. 

Thanks! However, Zach might be confused by the Github notification from your comment ;) 

> Just to make sure I'm not lost in refactor intents, if we can separate net_ops work from the IscDhclient work it'd make me more certain we aren't missing something in review.

Nice attention to detail in the review. And I agree - breaking that out into a separate commit will make this a more coherent commit, in addition to helping review burden. I've moved that work [here](https://github.com/canonical/cloud-init/pull/2127).

cloudinit/net/freebsd.py
holmanb: Nice catch. Fixed.

blackboxsw (approved):
@holmanb LGTM, thanks for separating the PRs into something easier to review. Refactor looks good.

blackboxsw (changes requested):
hold merge on this for a moment, I want to check that Vultr invocation of `Init` I had missed.

blackboxsw (approved):
Clear, If we have to call `stages.Init()` to get the distro object for callers of `__main__`. I think that's ok. We could source this from the `sources.pkl_load(/var/lib/cloud/instance/obj.pkl)`, but I think stages.Init() is probably an easier entry point.

PR #2124, opened 2023-04-14
users: schema permit empty list to indicate create no users

blackboxsw:
```
users: schema permit empty list to indicate create no users

When defining json schema in a63f45f7, we were a bit too strict by
setting `minItems: 1` on the `users:` list.

This schema definition regressed the ability to prevent default_user
creation with user-data.  Remove that schema constraint because the
code already supports this case.
```

## Additional Context

This hits desktop and subiquity autoinstall deployment needs.

## Checklist:

### DIFF
--- a/tests/unittests/distros/test_user_data_normalize.py
+++ b/tests/unittests/distros/test_user_data_normalize.py
@@ -44,7 +44,7 @@ class TestUGNormalize(TestCase):
                 {"bob": "users2"},
             ]
         }
-        (_users, groups) = self._norm(g, distro)
+        _users, groups = self._norm(g, distro)
         self.assertIn("ubuntu", groups)
         ub_members = groups["ubuntu"]
         self.assertEqual(sorted(["foo", "bar"]), sorted(ub_members))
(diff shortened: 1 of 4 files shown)


tests/unittests/distros/test_user_data_normalize.py
```diff
@@ -160,6 +160,8 @@ def test_users_simple(self):
         self.assertIn("bob", users)
         self.assertEqual({"default": False}, users["joe"])
         self.assertEqual({"default": False}, users["bob"])
+        (users, _groups) = self._norm({"users": []}, distro)
```
holmanb, line 163: parenthesis are not necessary here

holmanb (commented):
Since there are no code changes here, I assume that this is fixing a regression? If so it would be good to include in the commit message which commit introduced the bug that is being fixed.

Nit in the commit message: the word "customer" feels like an odd choice for cloud-init. Maybe "user" would make more sense here?

tests/unittests/distros/test_user_data_normalize.py
blackboxsw: Changed. I was adhering to the style within the test module, but we might as well just fix the style within the module.

TheRealFalcon (approved):

blackboxsw:
Fixes: https://bugs.launchpad.net/cloud-init/+bug/1999476

PR #2127, opened 2023-04-18
Support Ephemeral Networking for BSD

holmanb:
```
Support ephemeral network for BSDs

Currently ephemeral.py uses hardcoded iproute2
calls to setup ephemeral networking, which is
incompatible with BSDs. Refactor to support
pluggable network interface operations.
```

## Additional Context
Blocked by https://github.com/canonical/cloud-init/pull/2130 ~~and https://github.com/canonical/cloud-init/pull/2122 ([context](https://github.com/canonical/cloud-init/pull/2122#pullrequestreview-1389048745))~~.

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -18,6 +18,7 @@ import urllib.parse
 from io import StringIO
 from typing import Any, Mapping, MutableMapping, Optional, Type
 
+import cloudinit.net.netops.iproute2 as iproute2
 from cloudinit import importer
 from cloudinit import log as logging
 from cloudinit import (
@@ -104,6 +105,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     # This is used by self.shutdown_command(), and can be overridden in
     # subclasses
     shutdown_options_map = {"halt": "-H", "poweroff": "-P", "reboot": "-r"}
+    net_ops = iproute2.Iproute2
 
     _ci_pkl_version = 1
     prefer_fqdn = False
@@ -118,6 +120,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         self.name = name
         self.networking: Networking = self.networking_cls()
         self.dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd]
+        self.net_ops = iproute2.Iproute2
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         """Perform deserialization fixes for Distro."""
@@ -997,6 +1000,26 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             ],
             **kwargs,
         )
+
+    @staticmethod
+    def build_dhclient_cmd(
+        path: str,
+        lease_file: str,
+        pid_file: str,
+        interface: str,
+        config_file: str,
+    ) -> list:
+        return [
+            path,
+            "-1",
+            "-v",
+            "-lf",
+            lease_file,
+            "-pf",
+            pid_file,
+            "-sf",
+            "/bin/true",
+        ] + (["-cf", config_file, interface] if config_file else [interface])
 
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -203,6 +203,6 @@ class Distro(cloudinit.distros.bsd.BSD):
         interface: str,
         config_file: str,
     ) -> list:
-        return ["-l", lease_file, "-p", pid_file] + (
+        return [path, "-l", lease_file, "-p", pid_file] + (
             ["-c", config_file, interface] if config_file else [interface]
         )
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -194,3 +194,15 @@ class Distro(cloudinit.distros.bsd.BSD):
             ["update"],
             freq=PER_INSTANCE,
         )
+
+    @staticmethod
+    def build_dhclient_cmd(
+        path: str,
+        lease_file: str,
+        pid_file: str,
+        interface: str,
+        config_file: str,
+    ) -> list:
+        return ["-l", lease_file, "-p", pid_file] + (
+            ["-c", config_file, interface] if config_file else [interface]
+        )
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -4,7 +4,8 @@
 """
 import contextlib
 import logging
-from typing import Any, Dict, List, Optional
+from functools import partial
+from typing import Any, Callable, Dict, List, Optional
 
 import cloudinit.net as net
 from cloudinit import subp
@@ -30,6 +31,7 @@ class EphemeralIPv4Network:
 
     def __init__(
         self,
+        distro,
         interface,
         ip,
         prefix_or_mask,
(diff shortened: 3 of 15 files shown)


holmanb:
> Perhaps netinfo.py (`netdev_info()`, `_netdev_route_info_iproute()`, `route_info()`, `_netdev_route_info_netstat()`) should also get support of this interface.

This will be left for a separate pull request.

cloudinit/distros/__init__.py
```diff
@@ -998,6 +1001,26 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+            path,
+            "-1",
+            "-v",
+            "-lf",
+            lease_file,
+            "-pf",
+            pid_file,
+            "-sf",
+            "/bin/true",
+        ] + (["-cf", config_file, interface] if config_file else [interface])
```
blackboxsw, line 1022: No need for parens in this line
```suggestion
        ] + ["-cf", config_file, interface] if config_file else [interface]
```

cloudinit/distros/freebsd.py
```diff
@@ -194,3 +194,15 @@ def update_package_sources(self):
...
+
+    @staticmethod
+    def build_dhclient_cmd(
+        path: str,
+        lease_file: str,
+        pid_file: str,
+        interface: str,
+        config_file: str,
+    ) -> list:
+        return ["-l", lease_file, "-p", pid_file] + (
```
blackboxsw, line 206: We are missing a `path` var here that probably breaks things
```suggestion
        return [path, "-l", lease_file, "-p", pid_file] + (
```

cloudinit/net/netops/iproute2.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class Iproute2(netops.NetOps):
+    @staticmethod
+    def link_up(interface: str, family: Optional[str] = None):
+        subp.subp(
+            ["ip"]
+            + (["-family", family] if family else [])
+            + ["link", "set", "dev", interface, "up"]
+        )
```
blackboxsw, line 14: If we are manipulating network config, I think it may be ideal to capture=True on these commands to aid in debugging, What do you think @holmanb? Too noisy?

blackboxsw (changes requested):
Overall looks pretty good. Take the rest of my nits with a grain of salt, but we need minimally to fix the `freebsd.py:build_dhclient_cmd`.

Here's a consolidated diff with review comments:
 - dropping unneeded parens
 - adding capture=True to network config cmds
 - adding `path` in bsd dhclient cmd

```diff
diff --git a/cloudinit/distros/freebsd.py b/cloudinit/distros/freebsd.py
index a4c8097e4..4ed1356e9 100644
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -203,6 +203,6 @@ class Distro(cloudinit.distros.bsd.BSD):
         interface: str,
         config_file: str,
     ) -> list:
-        return ["-l", lease_file, "-p", pid_file] + (
+        return [path, "-l", lease_file, "-p", pid_file] + (
             ["-c", config_file, interface] if config_file else [interface]
         )
diff --git a/cloudinit/net/netops/iproute2.py b/cloudinit/net/netops/iproute2.py
index 08d79b187..be2fa1055 100644
--- a/cloudinit/net/netops/iproute2.py
+++ b/cloudinit/net/netops/iproute2.py
@@ -10,7 +10,8 @@ class Iproute2(netops.NetOps):
         subp.subp(
             ["ip"]
             + (["-family", family] if family else [])
-            + ["link", "set", "dev", interface, "up"]
+            + ["link", "set", "dev", interface, "up"],
+            capture=True,
         )
 
     @staticmethod
@@ -18,7 +19,8 @@ class Iproute2(netops.NetOps):
         subp.subp(
             ["ip"]
             + (["-family", family] if family else [])
-            + ["link", "set", "dev", interface, "down"]
+            + ["link", "set", "dev", interface, "down"],
+            capture=True,
         )
 
     @staticmethod
@@ -31,20 +33,19 @@ class Iproute2(netops.NetOps):
     ):
         subp.subp(
             ["ip", "-4", "route", "add", route]
-            + (["via", gateway] if gateway and gateway != "0.0.0.0" else [])
-            + [
-                "dev",
-                interface,
-            ]
-            + (["src", source_address] if source_address else []),
+            + ["via", gateway] if gateway and gateway != "0.0.0.0" else []
+            + ["dev", interface]
+            + ["src", source_address] if source_address else [],
+            capture=True,
         )
 
     @staticmethod
     def append_route(interface: str, address: str, gateway: str):
         subp.subp(
             ["ip", "-4", "route", "append", address]
-            + (["via", gateway] if gateway and gateway != "0.0.0.0" else [])
-            + ["dev", interface]
+            + ["via", gateway] if gateway and gateway != "0.0.0.0" else []
+            + ["dev", interface],
+            capture=True,
         )
 
     @staticmethod
@@ -57,9 +58,10 @@ class Iproute2(netops.NetOps):
     ):
         subp.subp(
             ["ip", "-4", "route", "del", address]
-            + (["via", gateway] if gateway and gateway != "0.0.0.0" else [])
+            + ["via", gateway] if gateway and gateway != "0.0.0.0" else []
             + ["dev", interface]
-            + (["src", source_address] if source_address else [])
+            + ["src", source_address] if source_address else [],
+            capture=True,
         )
 
     @staticmethod
@@ -83,11 +85,13 @@ class Iproute2(netops.NetOps):
                 "dev",
                 interface,
             ],
+            capture=True,
             update_env={"LANG": "C"},
         )
 
     @staticmethod
     def del_addr(interface: str, address: str):
         subp.subp(
-            ["ip", "-family", "inet", "addr", "del", address, "dev", interface]
+            ["ip", "-family", "inet", "addr", "del", address, "dev", interface],
+            capture=True,
         )
```

igalic:
thanks for the review, @blackboxsw
i have started working on bsd side of this change

cloudinit/net/ephemeral.py
```diff
@@ -152,127 +140,71 @@ def _bringup_device(self):
...
-                "add",
-                self.router,
-                "dev",
-                self.interface,
-                "src",
-                self.ip,
-            ],
-            capture=True,
+        self.distro.net_ops.add_route(
+            self.interface, self.router, source_address=self.ip
```
igalic, line 191: what exactly does this do?
how is this route different from one without `src self.ip`?

igalic (commented):
❓

cloudinit/net/netops/iproute2.py
holmanb: Isn't this already [the default behavior](https://github.com/canonical/cloud-init/blob/e8ef58b9c07a851d62777c68c34d26b8088597ff/cloudinit/subp.py#L164)? I removed it because it looks like a noop.

blackboxsw: Dumb, yes thanks for the correction there. My fault for not looking at the function sig when I made this comment. Please disregard.

cloudinit/distros/__init__.py
holmanb: I disagree. That would cause an invalid command to be returned when `config_file` is passed. Note the `if`/`else` logic.

blackboxsw: Man, note to self, do not review on a plane w/ lack of sleep. Thanks precedence. disregard please.

cloudinit/net/ephemeral.py
holmanb: I think ading `src self.ip` allows the route to force outgoing traffic to use the specified source address, which is otherwise unopinionated by a route that only specifies an interface.

blackboxsw (approved):
This changeset looks good to me, it fixes the missing `path` issue on freebsd.py.'s `build_dhclient_cmd`. Are we are waiting on a potential failure path seen on BSD outside of this?

PR #2128, opened 2023-04-18
gce: activate network discovery on every boot (cpc-2351)

aciba90:
## Proposed Commit Message

```
gce: activate network discovery on every boot

Google wants to allow users make changes on nics while the instance
is stopped. Activate network discovery on every boot.

Additionally, skip the call to `netplan generate` if the rendered
config is the same on subsequent boots.
```

## Additional Context

- https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L329
- [Google wants users to be able to reconfigure the NIC after instance creation](https://cloud.google.com/compute/docs/instances/migrating-interfaces-between-networks)
- More context on architectural relationship between cloud-init and netplan: https://bugs.launchpad.net/cloud-init/+bug/1737630

## Test Steps

1. Create an instance
2. Stop the instance
3. Move the instance to another VPC
4. Start the instance
5. Verify the instance is reachable

## Checklist:

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -1725,7 +1725,7 @@ def logexc(log, msg, *args):
     log.debug(msg, exc_info=exc_info, *args)
 
 
-def hash_blob(blob, routine, mlen=None):
+def hash_blob(blob, routine: str, mlen=None) -> str:
     hasher = hashlib.new(routine)
     hasher.update(encode_text(blob))
     digest = hasher.hexdigest()
@@ -1734,6 +1734,13 @@ def hash_blob(blob, routine, mlen=None):
         return digest[0:mlen]
     else:
         return digest
+
+
+def hash_buffer(f: io.BufferedIOBase) -> bytes:
+    hasher = hashlib.sha1()
+    for chunk in iter(lambda: f.read(io.DEFAULT_BUFFER_SIZE), b""):
+        hasher.update(chunk)
+    return hasher.digest()
 
 
 def is_user(name):
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -1737,6 +1737,11 @@ def hash_blob(blob, routine: str, mlen=None) -> str:
 
 
 def hash_buffer(f: io.BufferedIOBase) -> bytes:
+    """Hash the content of a binary buffer using SHA1.
+
+    @param f: buffered binary stream to hash.
+    @return: digested data as bytes.
+    """
     hasher = hashlib.sha1()
     for chunk in iter(lambda: f.read(io.DEFAULT_BUFFER_SIZE), b""):
         hasher.update(chunk)
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -1,7 +1,6 @@
 # This file is part of cloud-init.  See LICENSE file ...
 
 import copy
-import hashlib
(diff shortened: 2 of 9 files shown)


cloudinit/util.py
```diff
@@ -1736,6 +1736,13 @@ def hash_blob(blob, routine, mlen=None):
         return digest
 
 
+def hash_buffer(f: io.BufferedIOBase) -> bytes:
```
TheRealFalcon, line 1739: Can we get a docstring for this? In general I prefer having docstrings for any new utility functions.

Is the purpose here over calling `hashlib.sha1(f.read()).digest()` so that we don't crash on very large files? I'm not criticizing, just making sure I understand the purpose.

TheRealFalcon (changes requested):
Just one comment/question inline.

I do also wonder if it's worth adding an integration test for the netplan behavior. Part of that would rely on trusting log messages, so it may be of limited value, but I still lean towards having one.

aciba90:
Thanks, @TheRealFalcon, for the review. I have improved the docs. About the integration test:

> Just one comment/question inline.
> 
> I do also wonder if it's worth adding an integration test for the netplan behavior. Part of that would rely on trusting log messages, so it may be of limited value, but I still lean towards having one.

Yes, I thought about adding an integration test, but if we do so in GCE we need the ability to force a nic modification. The way I manually did it was to move the instance to another VPC, so we would need:

1. GCE VPC lifecycle support in pycloudlib.
2. Be able to move an instance to another VPC:
   1. We could add this functionality to pycloudlib, but I doubt is useful enough to add it to pycloudlib's public API, or
   2. Use the underlying GCE compute client in the integration test.

Maybe there is an easier way to force a nic modification that I am missing, any thoughts on it?

TheRealFalcon:
I wasn't really thinking about a GCE specific test. Your manual coverage should be good enough for this specific issue. Do you agree?

Longer term though, I was thinking we could more generally test the new "don't call netplan generate" behavior. The current user events test combined with a new test specific to the skipping on netplan generate should be good overall coverage.

Now that I look at the [current user events test](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_user_events.py#L35), I'm pretty sure that it will fail unless it is updated. GCE should get moved from `test_boot_event_disabled_by_default` to `test_boot_event_enabled_by_default`

Beyond that, I was thinking about another test(s) where we
1. Boot, change the network config, reboot and ensure netplan generate was called and that we have different configs
2. Boot, don't change the config, reboot and ensure netplan generate was not called and we have the same config

As I said in an earlier comment, I think some of this will rely on trusting log messages, so maybe not the most valuable test for the amount of time it would take. I was also thinking more about using LXD where it'd be easier to change something in the network config (e.g., MTU) without having to mess with multiple NICs. I'm not sure how well a test like that would translate to other clouds though. Do you think an integration test like that would be worthwhile?

aciba90:
I agree, @TheRealFalcon, long term we should think how to test this reliably on clouds.

I have:

* fixed `tests/integration_tests/modules/test_user_events.py`
* added lxd tests for the cases where netplan generate should and shouldn't be called

Let me know if more coverage is needed in this PR.

TheRealFalcon (approved):
LGTM!

PR #2129, opened 2023-04-18
azure/errors: introduce reportable errors

cjp256:
When provisioning failures occur an Azure, a generic description is used in the report and ultimately returned to the user.  To improve the user experience, report details of the failure in a manner that is parsable, readable and succinct.  The current approach is to use csv with a custom delimiter ("|") and quote character ("'").  This format may change in the future.

Gracefully handle reportable errors thrown while crawling metadata and treat other exceptions as ReportableErrorUnhandledException.  Future work will introduce more reportable errors to handle the expected failure cases.


aciba90 (approved):
+1. Nice touch on `ReportableError`, thanks!

I left an inline comment showing how could we solve the circular dependency issue and make the `errors` module more cohesive. I leave it open for a while, just in case you want to include it.

PR #2130, opened 2023-04-19
net: Document use of `ip route append` to add routes

holmanb:
```
net: Document use of `ip route append` to add routes
```

**Update**
I am changing this PR to document the reason that the current code is needed for better context.

### DIFF
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -197,6 +197,14 @@ class EphemeralIPv4Network:
             via_arg = []
             if gateway != "0.0.0.0":
                 via_arg = ["via", gateway]
+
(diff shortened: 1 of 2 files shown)


cjp256 (approved):
LGTM

TheRealFalcon:
This was intentionally changed in #1124, though the code has moved files.

holmanb:
> This was intentionally changed in #1124, though the code has moved files. I can't say I 100% understand the use case, but I think changing this could break some (or at least one) use cases.

Great context, thanks for pointing out the requirement for this behavior @TheRealFalcon.

I just reverted the previous commit and updated this PR to include context alongside the code, since that is currently not obvious.

TheRealFalcon (approved):
Thanks!

PR #2133, opened 2023-04-19
Update kernel commandline docs (SC-1457)

holmanb:
```
Update kernel commandline docs

- describe kernel commandline override
- add OpenStack Ironic selection instructions
- describe datasource selection
- move datasource creation doc to dev section
- remove generator stage from boot stages
- document how to disable cloud-init in howtos
- fix missing kernel-cmdline.rst link
- update outdated security advice

Document the following commits:

 34e8c914df666c937e48f5d1c3add0bd47e4e7eb
 f146fe71733e72b94fad525b8cc9988b1405e760
 250280ada67995a8449b64027b879d01939d2729
 612b4de892d19333c33276d541fed99fd16d3998
 a60c0845806baff72c74603286d048efbafab664
 d1ffbea556a06105d1ade88b4143ad43f53692c4
 02202954c65a7a1cdb9b28703bd0af01edd0e091
```

## Additional Context
This documentation serves to document recent work in the following commits:

https://github.com/canonical/cloud-init/commit/34e8c914df666c937e48f5d1c3add0bd47e4e7eb https://github.com/canonical/cloud-init/commit/f146fe71733e72b94fad525b8cc9988b1405e760 https://github.com/canonical/cloud-init/commit/250280ada67995a8449b64027b879d01939d2729 https://github.com/canonical/cloud-init/commit/612b4de892d19333c33276d541fed99fd16d3998 https://github.com/canonical/cloud-init/commit/a60c0845806baff72c74603286d048efbafab664 https://github.com/canonical/cloud-init/commit/d1ffbea556a06105d1ade88b4143ad43f53692c4 https://github.com/canonical/cloud-init/commit/02202954c65a7a1cdb9b28703bd0af01edd0e091

### DIFF
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -3,7 +3,7 @@
 Datasource creation
 *******************
 
-There are multiple ways to provide `user data`, `meta data`, and
+There are multiple ways to provide `user data`, `metadata`, and
 `vendor data`, and each cloud solution prefers its own way. A datasource
 abstract class was internally created to allow for a single way to access the
 different cloud systems methods, providing this data through the typical usage
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -1,0 +1,89 @@
+.. _datasource_creation:
+
+Datasource creation
+*******************
+
+Since there are multiple ways to provide `user data`, `meta data`, and
+`vendor data` (each cloud solution seems to prefer its own way), a datasource
+abstract class was internally created to allow for a single way to access the
+different cloud systems methods, providing this data through the typical usage
+of subclasses. See :file:`cloud-init/sources/__init__.py` as a starting
+point for understanding the datastructures used.
+
+The datasource objects have a few touch points with ``cloud-init``. If you
+are interested in adding a new datasource for your cloud platform you will
+need to take care of the following items:
+
+Identify a mechanism for positive identification of the platform
+================================================================
+
+It is good practice for a cloud platform to positively identify itself to
+the guest. This allows the guest to make educated decisions based on the
+platform on which it is running. On the x86 and arm64 architectures, many
+clouds identify themselves through DMI data. For example, Oracle's public
+cloud provides the string ``'OracleCloud.com'`` in the DMI chassis-asset
+field.
+
+``Cloud-init``-enabled images produce a log file with details about the
+platform. Reading through this log in :file:`/run/cloud-init/ds-identify.log`
+may provide the information needed to uniquely identify the platform.
+If the log is not present, you can generate it by running from source
+:file:`./tools/ds-identify` or the installed location
+:file:`/usr/lib/cloud-init/ds-identify`.
+
+The mechanism used to identify the platform will be required for the
+``ds-identify`` and datasource module sections below.
+
+Add datasource module cloudinit/sources/DataSource<CloudPlatform>.py
+====================================================================
+
+It is suggested that you start by copying one of the simpler datasources
+such as ``DataSourceHetzner``.
+
+Add tests for datasource module
+===============================
+
+Add a new file with some tests for the module to
+:file:`cloudinit/sources/test_<yourplatform>.py`. For example, see
+:file:`cloudinit/sources/tests/test_oracle.py`
+
+Update ``ds-identify``
+======================
+
+In ``systemd`` systems, ``ds-identify`` is used to detect which datasource
+should be enabled, or if ``cloud-init`` should run at all. You'll need to
+make changes to :file:`tools/ds-identify`.
+
+Add tests for ``ds-identify``
+=============================
+
+Add relevant tests in a new class to
+:file:`tests/unittests/test_ds_identify.py`. You can use ``TestOracle`` as
+an example.
+
+Add your datasource name to the built-in list of datasources
+============================================================
+
+Add your datasource module name to the end of the ``datasource_list``
+entry in :file:`cloudinit/settings.py`.
+
+Add your cloud platform to apport collection prompts
+====================================================
+
+Update the list of cloud platforms in :file:`cloudinit/apport.py`. This list
+will be provided to the user who invokes :command:`ubuntu-bug cloud-init`.
+
+Enable datasource by default in Ubuntu packaging branches
+=========================================================
+
+Ubuntu packaging branches contain a template file,
+:file:`debian/cloud-init.templates`, which ultimately sets the default
+``datasource_list`` when installed via package. This file needs updating when
+the commit gets into a package.
+
+Add documentation for your datasource
+=====================================
+
+You should add a new file in :file:`doc/datasources/<cloudplatform>.rst`.
+
+.. _make-mime: https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#storage-locations
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -37,7 +37,7 @@ The mechanism used to identify the platform will be required for the
 Add datasource module cloudinit/sources/DataSource<CloudPlatform>.py
 ====================================================================
 
-It is suggested that you start by copying one of the simpler datasources
+We suggest you start by copying one of the simpler datasources
 such as ``DataSourceHetzner``.
 
 Add tests for datasource module
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -5,14 +5,13 @@ Datasource creation
 
 There are multiple ways to provide `user data`, `metadata`, and
 `vendor data`, and each cloud solution prefers its own way. A datasource
-abstract class was internally created to allow for a single way to access the
-different cloud systems methods, providing this data through the typical usage
-of subclasses. See :file:`cloud-init/sources/__init__.py` as a starting
-point for understanding the datastructures used.
+abstract base class defines a single interface to interact with the different
+clouds. Each cloud implementation must inherit from this base class to use this
+shared functionality and interface. See :file:`cloud-init/sources/__init__.py`
+to see this class.
 
-The datasource objects have a few touch points with ``cloud-init``. If you
-are interested in adding a new datasource for your cloud platform you will
-need to take care of the following items:
+If you are interested in adding a new datasource for your cloud platform you
+will need to do all of the following:
 
 Identify a mechanism for positive identification of the platform
 ================================================================
@@ -20,7 +19,7 @@ Identify a mechanism for positive identification of the platform
 It is good practice for a cloud platform to positively identify itself to
 the guest. This allows the guest to make educated decisions based on the
 platform on which it is running. On the x86 and arm64 architectures, many
-clouds identify themselves through DMI data. For example, Oracle's public
+clouds identify themselves through `DMI`_ data. For example, Oracle's public
 cloud provides the string ``'OracleCloud.com'`` in the DMI chassis-asset
 field.
 
@@ -87,3 +86,4 @@ Add documentation for your datasource
 You should add a new file in :file:`doc/datasources/<cloudplatform>.rst`.
 
 .. _make-mime: https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#storage-locations
+.. _DMI: https://www.dmtf.org/sites/default/files/standards/documents/DSP0005.pdf
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -4,7 +4,7 @@ Datasource creation
 *******************
 
 Since there are multiple ways to provide `user data`, `meta data`, and
-`vendor data` (each cloud solution seems to prefer its own way), a datasource
+`vendor data`, and each cloud solution prefers its own way. A datasource
 abstract class was internally created to allow for a single way to access the
 different cloud systems methods, providing this data through the typical usage
 of subclasses. See :file:`cloud-init/sources/__init__.py` as a starting
--- a/doc/rtd/development/datasource_creation.rst
+++ b/doc/rtd/development/datasource_creation.rst
@@ -3,7 +3,7 @@
 Datasource creation
 *******************
 
-Since there are multiple ways to provide `user data`, `meta data`, and
+There are multiple ways to provide `user data`, `meta data`, and
 `vendor data`, and each cloud solution prefers its own way. A datasource
 abstract class was internally created to allow for a single way to access the
 different cloud systems methods, providing this data through the typical usage
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -17,7 +17,7 @@ stages to boot:
 
 Generator
 =========
-When booting under systemd, a generator will run that determines if
+When booting under ``systemd``, a generator will run that determines if
 cloud-init.target should be included in the boot goals. ``ds-identify``
 runs at this stage.
 
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -7,36 +7,11 @@ To be able to provide the functionality that it does, ``cloud-init`` must be
 integrated into the boot in a fairly controlled way. There are five
 stages to boot:
 
-1. Generator
-2. Local
-3. Network
-4. Config
-5. Final
+1. Local
+2. Network
+3. Config
+4. Final
 
-.. _boot-Generator:
-
-Generator
-=========
-
-# TODO: This entire stage is really just an implementation detail of systemd
-# systems - as long as we document how to disable cloud-init somewhere else
-# I think we can probably remove this information - unless we think it is
-# somehow actionable for the average user?
-
-When booting under ``systemd``, a `generator`_ will run that determines if
-``cloud-init.target`` should be included in the boot goals. By default, this
-generator will enable ``cloud-init``. It will not enable ``cloud-init``
-if either:
-
-- The file :file:`/etc/cloud/cloud-init.disabled` exists, or
-- The kernel command line as found in :file:`/proc/cmdline` contains
-  ``cloud-init=disabled``. When running in a container, the kernel command
-  line is not honored, but ``cloud-init`` will read an environment
-  variable named ``KERNEL_CMDLINE`` in its place.
-
-.. note::
-   These mechanisms for disabling ``cloud-init`` at runtime currently only
-   exist in ``systemd``.
 
 .. _boot-Local:
 
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -18,6 +18,11 @@ stages to boot:
 Generator
 =========
 
+# TODO: This entire stage is really just an implementation detail of systemd
+# systems - as long as we document how to disable cloud-init somewhere else
+# I think we can probably remove this information - unless we think it is
+# somehow actionable for the average user?
+
 When booting under ``systemd``, a `generator`_ will run that determines if
 ``cloud-init.target`` should be included in the boot goals. By default, this
 generator will enable ``cloud-init``. It will not enable ``cloud-init``
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -7,11 +7,19 @@ To be able to provide the functionality that it does, ``cloud-init`` must be
 integrated into the boot in a fairly controlled way. There are five
 stages to boot:
 
-1. Local
-2. Network
-3. Config
-4. Final
+1. Generator
+2. Local
+3. Network
+4. Config
+5. Final
 
+.. _boot-Generator:
+
+Generator
+=========
+When booting under systemd, a generator will run that determines if
+cloud-init.target should be included in the boot goals. ``ds-identify``
+runs at this stage.
 
 .. _boot-Local:
 
--- a/doc/rtd/explanation/format.rst
+++ b/doc/rtd/explanation/format.rst
@@ -6,6 +6,8 @@ User data formats
 User data that will be acted upon by ``cloud-init`` must be in one of the
 following types.
 
+.. _user_data_cloud_config:
+
 Cloud config data
 =================
 
--- a/doc/rtd/explanation/format.rst
+++ b/doc/rtd/explanation/format.rst
@@ -6,7 +6,7 @@ User data formats
 User data that will be acted upon by ``cloud-init`` must be in one of the
 following types.
 
-.. _user_data_cloud_config:
+.. _user_data_formats-cloud_config:
 
 Cloud config data
 =================
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -36,8 +36,8 @@ Then, when the datasource runs, it will find that config already available.
 # TODO: document cloud-init=disabled (also, does this work on non-systemd)
 
 
-# TODO: Since cloud-config-url in the commandline tells ds-identify that this
-# is MASS, this belongs in MAAS-specific documentation
+# TODO: make this not-maas-specific example
+# TODO: add a test for this behavior
 So, to be able to configure the MAAS datasource by controlling the
 kernel command line from outside the image, you can append:
 
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -9,12 +9,12 @@ resort, since this method supports only a subset of the
 do not support injecting kernel commandline arguments without modifying
 the bootloader.
 
-Despite the limitations of using the kernel commandline, cloud-init supports
+Despite the limitations of using the kernel command line, cloud-init supports
 some use-cases.
 
-Note that this page describes kernel commandline behavior that applies
+Note that this page describes kernel command line behavior that applies
 to all clouds. To provide a local configuration with an image using kernel
-commandline, see :ref:`datasource NoCloud<datasource_nocloud>` which provides
+command line, see :ref:`datasource NoCloud<datasource_nocloud>` which provides
 more configuration options.
 
 .. _kernel_datasource_override:
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -5,9 +5,9 @@ Kernel command line
 
 Providing configuration data via the kernel command line is somewhat of a last
 resort, since this method supports only a subset of the
-:ref:`cloud config<user_data_formats-cloud_config>` format, and many datasources
-do not support injecting kernel commandline arguments without modifying
-the bootloader.
+:ref:`cloud config<user_data_formats-cloud_config>` format, and many
+datasources do not support injecting kernel commandline arguments without
+modifying the bootloader.
 
 Despite the limitations of using the kernel command line, cloud-init supports
 some use-cases.
@@ -88,8 +88,8 @@ Then, have the following content at that url:
 .. note::
 
    Since ``cloud-config-url=`` is so generic, in order to avoid false
-   positives, only :ref:`cloud config<user_data_cloud_config>` user data
-   starting with ``#cloud-config`` is supported.
+   positives, only :ref:`cloud config<user_data_formats-cloud_config>` user
+   data starting with ``#cloud-config`` is supported.
 
 
 .. note::
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -32,6 +32,8 @@ If that file exists already, it will not be overwritten, and the
 
 Then, when the datasource runs, it will find that config already available.
 
+# TODO say something about datasource detection override
+
 So, to be able to configure the MAAS datasource by controlling the
 kernel command line from outside the image, you can append:
 
@@ -61,6 +63,7 @@ Then, have the following content at that url:
    Since ``cloud-config-url=`` is so generic, in order to avoid false
    positives, ``cloud-init`` requires the content to start with
    ``#cloud-config`` for it to be considered.
+
 
 .. note::
 
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -32,7 +32,8 @@ If that file exists already, it will not be overwritten, and the
 
 Then, when the datasource runs, it will find that config already available.
 
-# TODO say something about datasource detection override
+# TODO: say something about datasource detection override
+# TODO: document cloud-init=disabled (also, does this work on non-systemd)
 
 So, to be able to configure the MAAS datasource by controlling the
 kernel command line from outside the image, you can append:
--- a/doc/rtd/explanation/kernel-cmdline.rst
+++ b/doc/rtd/explanation/kernel-cmdline.rst
@@ -35,6 +35,9 @@ Then, when the datasource runs, it will find that config already available.
 # TODO: say something about datasource detection override
 # TODO: document cloud-init=disabled (also, does this work on non-systemd)
 
+
+# TODO: Since cloud-config-url in the commandline tells ds-identify that this
+# is MASS, this belongs in MAAS-specific documentation
 So, to be able to configure the MAAS datasource by controlling the
 kernel command line from outside the image, you can append:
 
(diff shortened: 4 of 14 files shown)


doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -62,8 +87,9 @@ Then, have the following content at that url:
    positives, ``cloud-init`` requires the content to start with
    ``#cloud-config`` for it to be considered.
 
+
 .. note::
 
-   The ``cloud-config-url=`` is un-authed http GET, and contains credentials.
```
holmanb, line 67: Requesting a security review of this change from here to the end of this file, please.

I propose a statement that should hopefully last the test of time a bit better than the current docs.

eslerm: could an encrypted alternative to `cloud-config-url=` be suggested here?

instead of "un-authed", I would say "unencrypted"

*if* credentials are needed, the channel is unlikely to be trusted. perhaps a strongly worded id est could be added to the last sentence, like: `... via trusted channels (i.e., within a closed system).`

eslerm (changes requested):

doc/rtd/explanation/kernel-cmdline.rst
holmanb: > instead of "un-authed", I would say "unencrypted"

+1 thanks

> could an encrypted alternative to cloud-config-url= be suggested here?

Unfortunately, this feature doesn't have an encrypted alternative to suggest.

> ... via trusted channels (i.e., within a closed system).

Thanks for the suggestions. I've added them in the latest commit.

eslerm (approved):
thanks :+1:

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -29,10 +47,42 @@ The permitted keys are:
...
+
+These user data and metadata files are required as separate files at the
+same base URL: ::
+
+  /user-data
+  /meta-data
+
+Both files must be present for it to be considered a valid seed ISO.
+
+Basically, ``user-data`` is simply :ref:`user data<user_data_formats>` and
```
TheRealFalcon, line 74: I know you didn't initially write it, but sentence is better without `Basically`. Also `what you'd find in the EC2 metadata service` isn't very intuitive. It'd be nice to fix here, but doesn't have to be.

TheRealFalcon (commented):
I left some inline comments, but great changes and reorganization overall.

doc/rtd/explanation/boot.rst
```diff
@@ -7,31 +7,11 @@ To be able to provide the functionality that it does, ``cloud-init`` must be
 integrated into the boot in a fairly controlled way. There are five
 stages to boot:
 
-1. Generator
-2. Local
-3. Network
-4. Config
-5. Final
+1. Local
```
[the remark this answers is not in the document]
holmanb: This stage is systemd-only, and the information that was contained here only described how to disable cloud-init. Since disabling cloud-init got moved to the new howto page, I didn't see any reason to keep this information here, since what value does that provide to a user? The other bootstages need to be understood sometimes for a user to know how to configure cloud-init, since cloud-init is not declarative. The fact that a systemd generator is used on systemd distros seems more like implementation detail than  actionable information to me. Why do you think this stage is just as important as the others? I can see this maybe being worth storing in a devdoc, but I'm not sure how it makes sense in the explanation category.

Thoughts?

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -29,10 +47,42 @@ The permitted keys are:
 * ``i`` or ``instance-id``
 * ``s`` or ``seedfrom``
 
-With ``ds=nocloud``, the ``seedfrom`` value must start with ``/`` or
-``file://``.  With ``ds=nocloud-net``, the ``seedfrom`` value must start
-with ``http://`` or ``https://`` and end with a trailing ``/``.
+A valid ``seedfrom`` values consist of:
```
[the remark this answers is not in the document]
holmanb: Not sure I follow, is this statement confusing as written? I was trying to say that one of the following seedfrom values consists of the following formats (filesystem or http server)

doc/rtd/reference/datasources.rst
```diff
@@ -9,20 +9,29 @@ configuration drive (i.e., metadata). Typical user data includes files,
...
+
+By default ``cloud-init`` should automatically determine which datasource it is
+running on. Therefore, in most cases, users of ``cloud-init`` should not
+have to configure ``cloud-init`` to specify which datasource cloud-init is
+running on; ``cloud-init`` should "figure it out".
+
+There are exceptions, however, when the :ref:`datasource does not
+identify<datasource_ironic>` itself to ``cloud-init``.
+
+- mention datasource_list in /etc/cloud/cloud.cfg.d/
```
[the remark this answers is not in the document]
holmanb: *facepalm*

doc/rtd/reference/datasources/nocloud.rst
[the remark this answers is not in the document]
TheRealFalcon: Sorry, grammatically, wouldn't you say `A valid ``seedfrom`` value consists of` rather than `A valid ``seedfrom`` values consists of`?

doc/rtd/explanation/boot.rst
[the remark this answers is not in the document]
holmanb: > I did a quick search through our docs, and we mention ds-identify a few times, but we don't really ever explain what it is or what it is doing.

100% agreed. The only "real" documentation of `ds-identify` imo is the source code.... which really sucks.

> What if we left

>> When booting under systemd, a generator_ will run that determines if cloud-init.target should be included in the boot goals.`

I wonder if it wouldn't make more sense to have a dedicated explanation page that describes details specific to each init system, rather than interspersing systemd-specific details throughout the docs, but I'm fine with the suggested approach for now.

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,44 @@
...
+Datasource discovery override
+=============================
+
+During boot, cloud-init must identify which datasource it is running on
+(OpenStack, AWS, Azure, GCP, etc). This discovery step can be optionally
+overriden by specifying the datasource name, such as:
+
+.. code-block:: text
+
+   root=/dev/sda ro ds=openstack
```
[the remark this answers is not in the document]
holmanb: Which part, exactly? Are you saying that want this example/section duplicated there? Or do you mean something different?

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,44 @@
 Kernel command line
 *******************
 
+Providing configuration data via the kernel command line is somewhat of a last
+resort, as it often requires modifying the boot loader to work. Despite the
```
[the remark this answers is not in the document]
holmanb: I agree about the difficulty of kernel commandline injection when launching a VM.

However, I think that there is value in encouraging users to use dedicated datasource 
methods when possible instead of this user interface.

How about this:
```
 Providing configuration data via the kernel command line is somewhat of a last
-resort, as it often requires modifying the boot loader to work. Despite the
-limitations of using the kernel commandline, cloud-init supports some
-use-cases.
+resort, since this method supports only a subset of the
+:ref:`cloud config<user_data_cloud_config>` format, and many datasources
+do not support injecting kernel commandline arguments.
+
+Despite the limitations of using the kernel commandline, cloud-init supports
+some use-cases.
 
```

doc/rtd/reference/datasources/nocloud.rst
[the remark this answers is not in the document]
holmanb: +1 I see what you're saying. Thanks.

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
TheRealFalcon: Works for me!

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
holmanb: See the new file `datasource_dsname_map.rst` that I just added. Maybe that's what you were asking for?

doc/rtd/reference/datasources/nocloud.rst
holmanb: +1

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -3,22 +3,40 @@
...
+The data source ``NoCloud`` is a flexible datasource that can be used in
+multiple different ways. With NoCloud, the user can provide user data and
+metadata to the instance without running a network service (or even without
+having a network at all). Alternatively, one may use a custom webserver to
+provide configurations.
 
-You can provide metadata and user data to a local VM boot via files on a
-`vfat`_ or `iso9660`_ filesystem. The filesystem volume label must be
-``cidata`` or ``CIDATA``.
+Configuration Methods:
```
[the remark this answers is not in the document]
holmanb: Thanks :)

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
TheRealFalcon: Sorry, I wasn't explaining myself very well last night.

I'm thinking more along the lines of a place where we list everything you can put on the kernel cmdline and adding it there, though, now that I think more about it, so much of what can be provided is datasource specific that it probably doesn't make sense to have something like that.

So...nevermind :smile:

holmanb:
Thanks for the feedback @TheRealFalcon. I think that I have addressed all requested changes.

TheRealFalcon (approved):
LGTM!

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -3,22 +3,40 @@
...
+======================
 
-Alternatively, you can provide metadata via the kernel command line or SMBIOS
-"serial number" option. The data must be passed in the form of a string: ::
+Method 1: Local filesystem, labeled filesystem
+----------------------------------------------
+To provide cloud-init configurations from the local filesystem, a labeled
+`vfat`_ or `iso9660`_ filesystem containing user data and metadata may
+be used. For this method to work, the filesystem volume must be labeled
+``CIDATA``.
```
dermotbradley, line 20: Is that correct?

cloud-utils' cloud-localds creates both [ISO](https://github.com/canonical/cloud-utils/blob/main/bin/cloud-localds#L235) and [VFAT](https://github.com/canonical/cloud-utils/blob/main/bin/cloud-localds#L241) filesystems with label ``cidata``, not ``CIDATA``.

holmanb: Both `cidata` and `CIDATA` are supported. I'd rather document one or the other, I don't feel strongly about which one to document.

doc/rtd/development/datasource_creation.rst
```diff
@@ -0,0 +1,89 @@
...
+*******************
+
+Since there are multiple ways to provide `user data`, `meta data`, and
+`vendor data` (each cloud solution seems to prefer its own way), a datasource
+abstract class was internally created to allow for a single way to access the
+different cloud systems methods, providing this data through the typical usage
+of subclasses. See :file:`cloud-init/sources/__init__.py` as a starting
+point for understanding the datastructures used.
+
+The datasource objects have a few touch points with ``cloud-init``. If you
```
s-makin, line 13: Is a datasource object the same thing as the datasource class mentioned above?

doc/rtd/development/datasource_creation.rst
```diff
@@ -0,0 +1,89 @@
+.. _datasource_creation:
+
+Datasource creation
+*******************
+
+Since there are multiple ways to provide `user data`, `meta data`, and
+`vendor data` (each cloud solution seems to prefer its own way), a datasource
```
s-makin, line 7:
```suggestion
`vendor data`, and each cloud solution prefers its own way, a datasource
```

doc/rtd/development/datasource_creation.rst
```diff
@@ -0,0 +1,89 @@
+.. _datasource_creation:
+
+Datasource creation
+*******************
+
+Since there are multiple ways to provide `user data`, `meta data`, and
```
s-makin, line 6:
```suggestion
There are multiple ways to provide `user data`, `meta data`, and
```

doc/rtd/development/datasource_creation.rst
```diff
@@ -0,0 +1,89 @@
...
+:file:`./tools/ds-identify` or the installed location
+:file:`/usr/lib/cloud-init/ds-identify`.
+
+The mechanism used to identify the platform will be required for the
+``ds-identify`` and datasource module sections below.
+
+Add datasource module cloudinit/sources/DataSource<CloudPlatform>.py
+====================================================================
+
+It is suggested that you start by copying one of the simpler datasources
```
s-makin, line 40:
```suggestion
We suggest you start by copying one of the simpler datasources
```

doc/rtd/explanation/boot.rst
```diff
@@ -17,21 +17,9 @@ stages to boot:
...
-- The file :file:`/etc/cloud/cloud-init.disabled` exists, or
-- The kernel command line as found in :file:`/proc/cmdline` contains
-  ``cloud-init=disabled``. When running in a container, the kernel command
-  line is not honored, but ``cloud-init`` will read an environment
-  variable named ``KERNEL_CMDLINE`` in its place.
-
-.. note::
-   These mechanisms for disabling ``cloud-init`` at runtime currently only
-   exist in ``systemd``.
+When booting under systemd, a generator will run that determines if
```
s-makin, line 20:
```suggestion
When booting under ``systemd``, a generator will run that determines if
```

doc/rtd/explanation/format.rst
```diff
@@ -6,6 +6,8 @@ User data formats
 User data that will be acted upon by ``cloud-init`` must be in one of the
 following types.
 
+.. _user_data_cloud_config:
```
s-makin, line 9:
```suggestion
.. _user_data_formats-cloud_config:
```

doc/rtd/development/datasource_creation.rst
holmanb: Good question.

There is a relationship between the [(abstract) parent class](https://github.com/canonical/cloud-init/blob/34637a49df13a054c4355de675b7cf286450b2f5/cloudinit/sources/__init__.py#L165) and the datasource object. The various child classes are defined in `cloudinit/sources/DataSource*.py` and inherit from `cloudinit.sources.DataSource`

As for this statement:

> The datasource objects have a few touch points with ``cloud-init``.

It does feel odd that we're trying to say that "there are several things that need to be updated when adding a new datasource", yet we use the word `object`, which only refers to only an instantance of the child class, not the creation of the child class itself. This wording seems odd to me, since adding a new datasource will always involve defining a new class, and is unlikely to require manipulating how the datasource object gets used.  Perhaps this would be more clear if it just said:

> Multiple parts of the codebase require an update when adding a new datasource.

Thoughts?

doc/rtd/explanation/format.rst
s-makin, line 9:
```suggestion
.. _user_data_formats-cloud_config:
```

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,48 @@
 Kernel command line
 *******************
 
+Providing configuration data via the kernel command line is somewhat of a last
+resort, since this method supports only a subset of the
```
s-makin, line 7: "subset of the cloud config format" reads a bit weirdly. What does it mean?

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,48 @@
 Kernel command line
 *******************
 
+Providing configuration data via the kernel command line is somewhat of a last
```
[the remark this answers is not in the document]
s-makin: Also - which configuration data are we referring to here?

doc/rtd/howto/disable_cloud_init.rst
```diff
@@ -0,0 +1,34 @@
+.. _disable-Cloud_init:
+
+Disable cloud-init
+******************
+
+There are two cross-platform methods of disabling ``cloud-init``.
+
+Method 1: text file
+====================
+To disable cloud-init, create the empty file
```
s-makin, line 10:
```suggestion
Method 1: text file
====================

To disable cloud-init, create the empty file
```

doc/rtd/howto/disable_cloud_init.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Example:
+
+.. code-block::
+
+    $ touch /etc/cloud/cloud-init.disabled
+
+Method 2: kernel commandline
+============================
+To disable cloud-init, add ``cloud-init=disabled`` to the kernel commandline.
```
s-makin, line 23:
```suggestion
Method 2: kernel commandline
============================

To disable cloud-init, add ``cloud-init=disabled`` to the kernel commandline.
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -3,22 +3,40 @@
...
+Configuration Methods:
+======================
 
-Alternatively, you can provide metadata via the kernel command line or SMBIOS
-"serial number" option. The data must be passed in the form of a string: ::
+Method 1: Local filesystem, labeled filesystem
+----------------------------------------------
+To provide cloud-init configurations from the local filesystem, a labeled
+`vfat`_ or `iso9660`_ filesystem containing user data and metadata may
+be used. For this method to work, the filesystem volume must be labeled
```
s-makin, line 19:
```suggestion
be used. For this method to work, the filesystem volume must be labelled
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -3,22 +3,40 @@
...
-or, ::
+Alternatively, one can provide metadata via the kernel command line or SMBIOS
+"serial number" option. This argument might look like: ::
 
-  ds=nocloud-net[;key=val;key=val]
+  ds=nocloud s=file://path/to/directory/;h=node-42
+
+Method 3: Custom webserver: kernel commandline or SMBIOS
+--------------------------------------------------------
+In a similar fashion, configuration files can be provided to cloud-init using a
```
s-makin, line 35:
```suggestion
Method 3: Custom webserver: kernel commandline or SMBIOS
--------------------------------------------------------

In a similar fashion, configuration files can be provided to cloud-init using a
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -29,10 +47,42 @@ The permitted keys are:
...
 * ``s`` or ``seedfrom``
 
-With ``ds=nocloud``, the ``seedfrom`` value must start with ``/`` or
-``file://``.  With ``ds=nocloud-net``, the ``seedfrom`` value must start
-with ``http://`` or ``https://`` and end with a trailing ``/``.
+A valid ``seedfrom`` value consists of:
+
+Filesystem
+----------
+A filesystem path starting with ``/`` or ``file://`` that points to a directory
```
s-makin, line 54:
```suggestion
Filesystem
----------

A filesystem path starting with ``/`` or ``file://`` that points to a directory
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -29,10 +47,42 @@ The permitted keys are:
...
+containing files: ``user-data``, ``meta-data``, and (optionally)
+``vendor-data``
+
+Http server
+-----------
+An ``http`` or ``https`` url (a trailing ``/`` is required)
```
s-makin, line 60:
```suggestion
HTTP server
-----------

An ``http`` or ``https`` url (a trailing ``/`` is required)
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -29,10 +47,42 @@ The permitted keys are:
...
+
+You may also optionally provide a vendor data file adhering to
+:ref:`user data formats<user_data_formats>` at the same base URL: ::
 
+  /vendor-data
+
+
+DMI-Specific Kernel Commandline
+===============================
 Cloud-init performs variable expansion of the ``seedfrom`` URL for any DMI
```
s-makin, line 86:
```suggestion
DMI-specific kernel commandline
===============================

Cloud-init performs variable expansion of the ``seedfrom`` URL for any DMI
```

doc/rtd/reference/datasources/openstack.rst
```diff
@@ -125,4 +125,35 @@ path; if found, settings are applied after (and, hence, overriding) the
 settings from static vendor data. Both sets of vendor data can be overridden
 by user data.
 
+.. _datasource_ironic:
+
+OpenStack Ironic Bare Metal
+===========================
+During boot, cloud-init typically has to identify which platform it is running
```
s-makin, line 132:
```suggestion
===========================

During boot, cloud-init typically has to identify which platform it is running
```

doc/rtd/reference/datasources/openstack.rst
```diff
@@ -125,4 +125,35 @@ path; if found, settings are applied after (and, hence, overriding) the
...
+on. Since OpenStack Ironic bare metal doesn't provide a method for cloud-init
```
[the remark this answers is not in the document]
s-makin: I had a look on their website, they use capitalisation in the whole name

doc/rtd/reference/datasources/openstack.rst
```diff
@@ -125,4 +125,35 @@ path; if found, settings are applied after (and, hence, overriding) the
...
+to discover that it is running on Ironic, extra user configuration is required.
+
+Cloud-init provides two method to do this:
```
s-makin, line 136:
```suggestion
Cloud-init provides two methods to do this:
```

doc/rtd/reference/datasources/openstack.rst
```diff
@@ -125,4 +125,35 @@ path; if found, settings are applied after (and, hence, overriding) the
...
+
+Method 1:
+---------
+Explicitly set ``datasource_list`` to only ``openstack``, such as:
```
s-makin, line 140:
```suggestion
---------

Explicitly set ``datasource_list`` to only ``openstack``, such as:
```

doc/rtd/reference/datasources/openstack.rst
```diff
@@ -125,4 +125,35 @@ path; if found, settings are applied after (and, hence, overriding) the
...
+
+OpenStack Ironic Bare Metal
+===========================
+During boot, cloud-init typically has to identify which platform it is running
+on. Since OpenStack Ironic bare metal doesn't provide a method for cloud-init
+to discover that it is running on Ironic, extra user configuration is required.
+
+Cloud-init provides two method to do this:
+
+Method 1:
```
s-makin, line 138: Having "method 1" and "method 2" as the headers is not ideal, as when it comes up in the RHS navigation menu it's not very descriptive. I'd recommend that instead, we make these into a numbered list.

s-makin (commented):
Mostly just minor nits, although there were a few places where I got a bit confused - not sure if that's a "me" problem, so if you wouldn't mind clarifying on those, that'd be great! Overall really nice work, makes some of these pages much more consumable. Thanks for doing it :)

doc/rtd/reference/datasources/openstack.rst
holmanb: Agreed, the navigation menu with this state is not ideal. I'm having a hard time picturing a numbered list looking good, however, since each method has some text and an example in it. Could you share an example of what you mean?

Alternatively I could improve the navigation by adding context to the title:
```
Method 1: Configuration file
...
Method 2: Kernel commandline
```

doc/rtd/howto/disable_cloud_init.rst
```diff
@@ -0,0 +1,34 @@
+.. _disable-Cloud_init:
+
+Disable cloud-init
```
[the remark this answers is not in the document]
holmanb: One may wish to disable cloud-init to ensure that it doesn't do anything on subsequent boots. Some parts of cloud-init run once per boot.

doc/rtd/development/datasource_creation.rst
```diff
@@ -0,0 +1,89 @@
+.. _datasource_creation:
+
+Datasource creation
+*******************
+
+Since there are multiple ways to provide `user data`, `meta data`, and
+`vendor data` (each cloud solution seems to prefer its own way), a datasource
+abstract class was internally created to allow for a single way to access the
+different cloud systems methods, providing this data through the typical usage
+of subclasses. See :file:`cloud-init/sources/__init__.py` as a starting
```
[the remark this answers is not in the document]
holmanb: That sentence is a bunch of fluff and jargon.

How about:
```
An abstract base class defines a single interface to interact with the different clouds. Each cloud implementation must inherit from this base class to utilize this shared functionality and interface.```

File: doc/rtd/development/datasource_creation.rst [hunk 32]
```diff
@@ -0,0 +1,89 @@
...
+are interested in adding a new datasource for your cloud platform you will
+need to take care of the following items:
+
+Identify a mechanism for positive identification of the platform
+================================================================
+
+It is good practice for a cloud platform to positively identify itself to
+the guest. This allows the guest to make educated decisions based on the
+platform on which it is running. On the x86 and arm64 architectures, many
+clouds identify themselves through DMI data. For example, Oracle's public
```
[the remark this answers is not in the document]
holmanb: It wouldn't hurt to link to something external. Not everyone will know.

That said, I'm not sure which link might be best:
[The standard would be most authoritative](https://www.dmtf.org/sites/default/files/standards/documents/DSP0005.pdf).
[Wikipedia might be most accessible](https://en.wikipedia.org/wiki/Desktop_Management_Interface)
[dmidecode may be most familiar to some](https://linux.die.net/man/8/dmidecode)

doc/rtd/explanation/kernel-cmdline.rst
holmanb: It means that this doesn't actually support all cloud config user data formats. Some valid user data cloud configurations may start with `Content-Type: text/x-shellscript` in the case of multipart MIME archive, however this isn't supported at this time via this interface - therefore only a subset of the cloud config format is supported.

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,48 @@
...
-   Usage of the kernel command line is somewhat of a last resort,
-   as it requires knowing in advance the correct command line or modifying
-   the boot loader to append data.
+This method is different from the standard way that most datasources receive
+configuration.
 
-For example, when :command:`cloud-init init --local` runs, it will check to
-see if ``cloud-config-url`` appears in key/value fashion in the kernel command
-line, as in:
+When :ref:`the local stage<boot-Local>` runs, it will check to see if
```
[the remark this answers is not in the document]
holmanb: No

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,48 @@
...
+configuration, ``cloud-init`` can read a URL directed by the kernel command
 line and proceed as if its data had previously existed.
 
 This allows for configuring a metadata service, or some other data.
 
-.. note::
-   Usage of the kernel command line is somewhat of a last resort,
-   as it requires knowing in advance the correct command line or modifying
-   the boot loader to append data.
+This method is different from the standard way that most datasources receive
```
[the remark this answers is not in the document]
holmanb: Do we? I guess we could link to the `datasources.rst` file, is that what you mean?

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
holmanb: > Should it be command line or commandline? (we've used both in the page)

The [kernel docs](https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html#the-kernel-s-command-line-parameters) use both `command-line` and `command line`. Lets standardize on `command line` I guess?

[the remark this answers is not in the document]
holmanb: Fixed

doc/rtd/reference/datasource_dsname_map.rst
```diff
@@ -0,0 +1,51 @@
...
+..
+    generate the following map with the following one-liner:
+
+    find cloudinit/sources -name 'DataSource*.py' \
+       | xargs grep 'dsname =' \
+       | awk -F '[/:"]' '{print $3 ", " $5}'
+
+
+.. csv-table::
+   :align: left
```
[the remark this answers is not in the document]
holmanb: I'll poke at it. The example in the docs shows a different default alignment than we have, so I'm curious if this is possibly theme-caused.

holmanb:
> Mostly just minor nits, although there were a few places where I got a bit confused - not sure if that's a "me" problem, so if you wouldn't mind clarifying on those, that'd be great! Overall really nice work, makes some of these pages much more consumable. Thanks for doing it :)

Thanks for the review @s-makin! I have responded with either fixes, answers, or solutions to the questions and issues raised. I think it's ready for another round of feedback.

doc/rtd/reference/datasources/nocloud.rst
holmanb: Thanks for the review @dermotbradley!

doc/rtd/reference/datasource_dsname_map.rst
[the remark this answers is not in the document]
s-makin: It probably is, because the Pro docs have the same

doc/rtd/development/datasource_creation.rst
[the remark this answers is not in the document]
s-makin: Definitely better. I'd say "use" instead of "utilize", and I'd keep "datasource" with the "abstract base class" bit (so that we're being specific about which abstract base class we're talking about), but otherwise, yeah, much improved.

doc/rtd/development/datasource_creation.rst
s-makin: The reason why I asked is because it reads like we're introducing a new term or concept (in "database objects") that hasn't been defined in the page. I know objects usually have a specific meaning, so I don't think it hurts us to explicitly state that relationship you just described, unless (since it's a devdoc page) you think our readers will already understand all of that?

I do agree that stating "Multiple parts of the codebase require an update when adding a new datasource." is clearer than what was there originally, but should we also state explicitly which parts of the codebase? (or is that fully taken care of in the list that follows?)

doc/rtd/development/datasource_creation.rst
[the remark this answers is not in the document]
s-makin: I think probably the source is the best choice. If it were a public-facing doc, rather than a devdoc, I might be tempted to go with a more accessible option, but I think in this case it's better to link to the standard. It's the most likely to be kept up to date.

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
s-makin: In that case I guess I've seen it three ways now lol. I don't have strong feelings about which one is best, as long as we use the same one consistently. Happy to go with your pick :)

doc/rtd/explanation/kernel-cmdline.rst
```diff
@@ -3,20 +3,48 @@
...
+resort, since this method supports only a subset of the
+:ref:`cloud config<user_data_cloud_config>` format, and many datasources
+do not support injecting kernel commandline arguments without modifying
+the bootloader.
+
+Despite the limitations of using the kernel commandline, cloud-init supports
+some use-cases.
+
+Note that this page describes kernel commandline behavior that applies
+to all clouds. To provide a local configuration with an image using kernel
```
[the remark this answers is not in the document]
s-makin: No need. Since the way I understood it is the way it was meant to be understood we can leave it as is.

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
s-makin: I can't find the place I thought it was. It might have been datasources.rst to be fair - or it might have been the instancedata.rst file in explanation. I'm not sure which is most appropriate so I'll leave that decision to you, but I do think we should link to somewhere where the standard way is outlined (however briefly).

s-makin (commented):
Thanks for the explanations and clarifications! Added a few more comments, but I think it's nearly ready to go.

doc/rtd/explanation/kernel-cmdline.rst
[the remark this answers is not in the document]
holmanb: @s-makin - I'll just drop this sentence, since it's basically a duplication of the opening paragraph.

s-makin (approved):
LGTM! Thanks for all your work on this :)

PR #2134, opened 2023-04-20
Remove mount NTFS error message

KsenijaS:
Provide an option to suppress error logging from _mount_cb_ as some errors can be expected error and handled appropriately by DataSources. For example: failure to mount NTFS volumes on VMs that do not have NTFS drivers.

### DIFF
--- a/tests/unittests/test_util.py
+++ b/tests/unittests/test_util.py
@@ -1583,7 +1583,7 @@ class TestMountCb:
         callback = mock.Mock(autospec=True)
         with pytest.raises(Exception):
             util.mount_cb(
-                "/dev/sda1",
+                "/dev/fake0",
                 callback,
                 mtype="ntfs",
                 update_env_for_mount={"LANG": "C"},
--- a/tests/unittests/test_util.py
+++ b/tests/unittests/test_util.py
@@ -1572,23 +1572,16 @@ class TestMountCb:
         ] == callback.call_args_list
 
     @mock.patch(M_PATH + "subp.subp")
-    @mock.patch("cloudinit.temp_utils.tempdir", autospec=True)
-    def test_mount_cb_does_not_log(self, m_subp, m_tmpdir, caplog):
+    def test_mount_cb_does_not_log(self, m_subp, caplog):
         log_msg = (
             "Failed to mount device: '/dev/sda1' with type: "
             "'ntfs' using mount command:"
         )
         m_subp.side_effect = subp.ProcessExecutionError(
-            "unknown filesystem type 'ntfs'"
-        )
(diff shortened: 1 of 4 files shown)


cjp256:
When opening a PR, I would suggest squashing the branch to start the PR with the minimum number of commits (usually one) with a descriptive commit message(s).  It will reduce the work for the maintainer to finalize the commit message when merging.

TheRealFalcon:
@cjp256 , do these changes look reasonable to you as they're Azure-specific? Is there any time where we would want the log but won't see it?

anhvoms:
> Remove error message from cloud-init log when trying to mount ntfs on Redhat/CentOS. This is an expected behavior.

Suggesting change to the description to provide better clarity:

Provide an option to suppress error logging from mount_cb as some errors can be expected error and handled appropriately by datasources. For example: failure to mount ntfs volumes on VMs that do not have ntfs drivers.

tests/unittests/test_util.py
```diff
@@ -1571,6 +1571,46 @@ def test_already_mounted_calls_callback_with_data(
...
+            "Failed to mount device: '/dev/sda1' with type: "
+            "'ntfs' using mount command:"
+        )
+        m_subp.side_effect = subp.ProcessExecutionError(
+            "", "unknown filesystem type 'ntfs'"
+        )
+        callback = mock.Mock(autospec=True)
+        with pytest.raises(Exception):
+            util.mount_cb(
+                "/dev/sda1",
```
cjp256, line 1586: the tests may be failing if the system mounts are leaking and /dev/sda1 is already mounted?

KsenijaS:
LGTM!

anhvoms (approved):

holmanb (approved):
@KsenijaS Thanks for proposing this change. This looks good to me.

> The only suggestion I have is to update test_mount_cb_does_not_log() to parametrize both cases (log_error=True and False) to ensure the log check is working as expected.
>
> Would be nice to add a test case for Azure as well to vet the expected behavior.

Do you plan on implementing the suggested changes by @cjp256?

KsenijaS:
> @KsenijaS Thanks for proposing this change. This looks good to me.
> 
> > The only suggestion I have is to update test_mount_cb_does_not_log() to parametrize both cases (log_error=True and False) to ensure the log check is working as expected.
> > Would be nice to add a test case for Azure as well to vet the expected behavior.
> 
> Do you plan on implementing the suggested changes by @cjp256?

The first suggestion to parametrize both cases is done now.
For second one, I don't think there is a need for a test case for Azure, since the logs that we are trying to suppress are visible only from utils file.

PR #2136, opened 2023-04-21
sources/azure: report failures to host via kvp

cjp256:
Azure can report provisioning failures via the Wireserver health endpoint.  However, in the event of networking failures or Wireserver issues, this report cannot be made and the VM will result in an OS provisioning timeout and a generic error is presented to the user.

Report the failure via KVP using the "PROVISIONING_REPORT" key so that the host can relay the provisioning error report to the user when the VM fails to provision.

The format used is subject to change and/or removal.


TheRealFalcon (approved):
LGTM!

PR #1513, opened 2022-06-13
setup.py: adjust udev/rules default path

esposem:
```
RHEL must put cloudinit .rules files in /usr/lib/udev/rules.d
This place is a rhel standard and since it is used by all packages
cannot be modified.

Signed-off-by: Emanuele Giuseppe Esposito <eesposit@redhat.com>
```
## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #2137, opened 2023-04-21
setup.py: use pkg-config for udev/rules path

dankm:
Distributions other than RHEL also use /usr/lib/udev for the rules path. Instead of hardcoding the udev rules path for RedHat, check pkg-config for the proper location.

## Proposed Commit Message

```
setup.py: use pkg-config for udev/rules path

Distributions other than RHEL also use /usr/lib/udev for the rules
path. Instead of hardcoding the udev rules path for RedHat, check
pkg-config for the proper location.
```

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,10 @@ def pkg_config_read(library, var):
             "systemdsystemconfdir": "/etc/systemd/system",
             "systemdsystemunitdir": "/lib/systemd/system",
             "systemdsystemgeneratordir": "/lib/systemd/system-generators",
-        }
+        },
+        "udev": {
+            "udevdir": "/lib/udev",
+        },
     }
     cmd = ["pkg-config", "--variable=%s" % var, library]
     try:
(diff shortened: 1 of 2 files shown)


dankm:
This relates to my comment on #1513, I ran into this with a Yocto build on a non-Redhat host.

setup.py
```diff
@@ -307,14 +310,11 @@ def finalize_options(self):
     ),
 ]
 if not platform.system().endswith("BSD"):
-
-    RULES_PATH = LIB
-    if os.path.isfile("/etc/redhat-release"):
-        RULES_PATH = "/usr/lib"
+    RULES_PATH = pkg_config_read("udev", "udevdir")
```
TheRealFalcon, line 313: Shouldn't this be `"/" + pkg_config_read("udev", "udevdir")` since `pkg_config_read` strips off the leading `/` ?

dankm: Not necessarily, you're right that there's an issue here, but it shouldn't be `"/" + ...` unconditionally. I need to check if we're in a virtualenv first, like in the function on line 276. I'll update the pull request shortly.

TheRealFalcon (approved):
Thanks!

PR #2138, opened 2023-04-22
openstack/static: honor the DNS servers associated with a network

goneri:
The `network_data.json` allows the definition of the DNS through the `services` list at the network level.

See:
- https://opendev.org/openstack/nova/src/commit/700db274c613d6f8f30e5cdc3462beaeb0fda456/nova/tests/unit/network/test_network_info.py#L979-L980
- https://opendev.org/openstack/metalsmith/src/commit/f98dfa61c1d7475b81c20dabbf2c74198c38c793/metalsmith/test/test_network_metadata.py#L52-L90
- https://opendev.org/openstack/nova/commit/4b333b989dfc778a8b61db4a1b8552e988a10471

### DIFF
--- a/cloudinit/sources/helpers/openstack.py
+++ b/cloudinit/sources/helpers/openstack.py
@@ -642,6 +642,14 @@ def convert_net_json(network_json=None, known_macs=None):
                     }
                 )
 
+            dns_nameservers = [
+                service["address"]
+                for service in network.get("services", [])
+                if service.get("type") == "dns"
+            ]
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
LGTM, thanks! Also verified against [the schema](https://docs.openstack.org/nova/latest/_downloads/9119ca7ac90aa2990e762c08baea3a36/network_data.json) to ensure we won't get any KeyErrors.

PR #2139, opened 2023-04-23
Remove the crypt dependency

goneri:
crypt is deprecated in Python 3.11 and will be removed in 3.13. This uses the passlib module to replace it. passlib is a pure-Python package.

See: https://passlib.readthedocs.io

### DIFF
--- a/cloudinit/distros/netbsd.py
+++ b/cloudinit/distros/netbsd.py
@@ -2,9 +2,10 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import crypt
 import os
 import platform
+
+from passlib.hash import bcrypt
 
 import cloudinit.distros.bsd
 from cloudinit import log as logging
@@ -90,8 +91,7 @@ class NetBSD(cloudinit.distros.bsd.BSD):
         if hashed:
             hashed_pw = passwd
         else:
-            method = crypt.METHOD_BLOWFISH  # pylint: disable=E1101
-            hashed_pw = crypt.crypt(passwd, crypt.mksalt(method))
+            hashed_pw = bcrypt.hash(passwd)
 
         try:
             subp.subp(["usermod", "-p", hashed_pw, user])
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -5,7 +5,6 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 import base64
-import crypt
 import os
 import os.path
 import re
@@ -15,6 +14,8 @@ from enum import Enum
 from pathlib import Path
 from time import sleep, time
 from typing import Any, Dict, List, Optional
+
+import passlib
 
 from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
@@ -1764,8 +1765,8 @@ def read_azure_ovf(contents):
     return (md, ud, cfg)
 
 
-def encrypt_pass(password, salt_id="$6$"):
-    return crypt.crypt(password, salt_id + util.rand_str(strlen=16))
+def encrypt_pass(password):
+    return passlib.hash.sha512_crypt.encrypt(password)
 
 
 @azure_ds_telemetry_reporter
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1766,7 +1766,7 @@ def read_azure_ovf(contents):
 
 
 def encrypt_pass(password):
-    return passlib.hash.sha512_crypt.encrypt(password)
+    return passlib.hash.sha512_crypt.hash(password)
 
 
 @azure_ds_telemetry_reporter
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -188,7 +188,6 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
-        assert self.execute("apt-get install -qy python3-passlib").ok
         assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
 
     @retry(tries=30, delay=1)
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -188,9 +188,7 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
-        assert self.execute(
-            "apt install -y --allow-downgrades {path}".format(path=remote_path)
-        ).ok
+        assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
 
     @retry(tries=30, delay=1)
     def upgrade_cloud_init(self):
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -188,6 +188,7 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
+        assert self.execute("apt-get install -qy python3-passlib").ok
         assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
 
     @retry(tries=30, delay=1)
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -188,7 +188,9 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
-        assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
+        assert self.execute(
+            "apt install -y --allow-downgrades {path}".format(path=remote_path)
+        ).ok
 
     @retry(tries=30, delay=1)
     def upgrade_cloud_init(self):
--- a/doc/examples/cloud-config.txt
+++ b/doc/examples/cloud-config.txt
@@ -425,8 +425,8 @@ syslog_fix_perms: syslog:root
 #   expire: True
 # ssh_pwauth: [ True, False, "" or "unchanged" ]
 #
-# Hashed passwords can be generated in multiple ways, example with python3:
-# python3 -c 'import crypt,getpass; print(crypt.crypt(getpass.getpass(), crypt.mksalt(crypt.METHOD_SHA512)))'
+# Hashed passwords can be generated in multiple ways, example with python3 and the passlib module:
+# python3 -c 'import passlib.hash;print(passlib.hash.sha512_crypt.hash("password"))'
 # Newer versions of 'mkpasswd' will also work: mkpasswd -m sha-512 password
 #
 # So, a simple working example to allow login via ssh, and not expire
--- a/integration-requirements.txt
+++ b/integration-requirements.txt
@@ -10,4 +10,3 @@ pycloudlib==1!3.0.2
 pytest<=7.3.1
 
 packaging
-passlib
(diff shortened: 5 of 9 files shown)


goneri:
Before I fix the integration tests, is adding a new dependency to replace crypt an acceptable option?

holmanb (commented):
> Before I fix the integration tests, is adding a new dependency to replace crypt an acceptable option?

Good question. I think that adding new dependencies is typically not ideal. However for something security-focused like `crypt` / `passlib`, my sentiment is that this approach likely makes more sense than rolling our own code.

### Context on Ubuntu:

This library isn't currently included in the base images, so at an installed size of 2096 kB, we would be increasing the base image a bit. This library is already in the main repo. See the following packages that depend on this library.

```bash
$ apt-cache rdepends python3-passlib
python3-passlib
Reverse Depends:
  mitmproxy
  tryton-server
  radicale
  python3-trove
  python3-radicale
  python3-flask-security
  python3-autobahn
  prewikka
  odoo-14
  python3-keystone
  mailman3
  bundlewrap
```

cloudinit/distros/netbsd.py
```diff
@@ -90,8 +90,7 @@ def set_passwd(self, user, passwd, hashed=False):
         if hashed:
             hashed_pw = passwd
         else:
-            method = crypt.METHOD_BLOWFISH  # pylint: disable=E1101
-            hashed_pw = crypt.crypt(passwd, crypt.mksalt(method))
+            hashed_pw = bcrypt.hash("password")
```
tobhe, line 93: Is there a reason the password is now a static string and no salt is used? Would `bcrypt.hashpw(passwd, bcrypt.gensalt())` work?

smoser:
It looks to me like cloud-init uses crypt library in these places:
1. Azure - if the ovf input from Azure platform had a plaintext password.  Is this a actual runtime path that is still used and needs to continue working?
2. netbsd - if cloud-config contained a plaintext password 
3. tests or doc

Ultimately the strong suggestion should be to *not*  provide use plaintext passwords.

Rather than picking up 2MB of additional python dependency, would it be possible to deprecate this functionality or remove it entirely.  In the deprecation path, you could dynamically load the bcrypt library to avoid an actual dependency.

cjp256:
> 1. Azure - if the ovf input from Azure platform had a plaintext password.  Is this a actual runtime path that is still used and needs to continue working?

The admin password is still supported as plaintext and fairly widely used, unfortunately. :(

cloudinit/distros/netbsd.py
holmanb: @goneri Could you please update this? As @tobhe pointed out, we need to pass in the variable `passwd`, rather that using the static password `"password"`, as you do in this PR.

setharnold (commented):
It makes sense to me, I've got a quick question on one change, though. Thanks

holmanb:
Thanks for the input @smoser.

> Ultimately the strong suggestion should be to not provide use plaintext passwords.

100% agreed. I wish this wasn't a behavior that people now depend on. Going forward hopefully we can avoid introducing additional poor security practices as supported features in cloud-init.

> Rather than picking up 2MB of additional python dependency, would it be possible to deprecate this functionality or remove it entirely.

The upstream cloud-init team would like to see this deprecated, but it will require working with Azure.

Thankfully since this is limited to netbsd and Azure, distro providers can opine images to limit the scope of this additional dependency to Azure images. We currently plan to do this with Ubuntu until we can remove the Azure dependency completely.

holmanb:
@goneri In order to allow Python < 3.13 to continue to use crypt without the additional dependency, and to not require the dependency where it isn't used, could we add something like the following patch?

Note: This includes changing to `passlib.hash.sha512_crypt.hash` from `passlib.hash.sha512_crypt.encrypt` as @setharnold pointed out that `.encrypt` is deprecated. 

```suggestion
diff --git a/cloudinit/distros/netbsd.py b/cloudinit/distros/netbsd.py
index bcd518779..d8654210a 100644
--- a/cloudinit/distros/netbsd.py
+++ b/cloudinit/distros/netbsd.py
@@ -2,15 +2,37 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
+import functools
 import os
 import platform
 
-from passlib.hash import bcrypt
-
 import cloudinit.distros.bsd
 from cloudinit import log as logging
 from cloudinit import subp, util
 
+try:
+    import crypt
+
+    hash = functools.partial(
+        crypt.crypt, salt=crypt.mksalt(crypt.METHOD_BLOWFISH)
+    )
+except ImportError:
+    try:
+        from passlib.hash import bcrypt
+
+        hash = bcrypt.hash
+    except ImportError:
+
+        def hash(_):
+            """Raise when called so that importing this module doesn't throw
+            ImportError when ds_detect() returns false. In this case, crypt
+            and passlib are not needed.
+            """
+            raise ImportError(
+                "crypt and passlib not found, missing dependency"
+            )
+
+
 LOG = logging.getLogger(__name__)
 
 
@@ -91,7 +113,7 @@ class NetBSD(cloudinit.distros.bsd.BSD):
         if hashed:
             hashed_pw = passwd
         else:
-            hashed_pw = bcrypt.hash(passwd)
+            hashed_pw = hash(passwd)
 
         try:
             subp.subp(["usermod", "-p", hashed_pw, user])
diff --git a/cloudinit/sources/DataSourceAzure.py b/cloudinit/sources/DataSourceAzure.py
index e5419769c..df6191fd4 100644
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -5,6 +5,7 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 import base64
+import functools
 import os
 import os.path
 import re
@@ -15,8 +16,6 @@ from pathlib import Path
 from time import sleep, time
 from typing import Any, Dict, List, Optional
 
-import passlib
-
 from cloudinit import log as logging
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
@@ -48,6 +47,29 @@ from cloudinit.sources.helpers.azure import (
 )
 from cloudinit.url_helper import UrlError
 
+try:
+    import crypt
+
+    hash = functools.partial(
+        crypt.crypt, salt=f"$6${util.rand_str(strlen=16)}"
+    )
+except ImportError:
+    try:
+        import passlib
+
+        hash = passlib.hash.sha512_crypt.hash
+    except ImportError:
+
+        def hash(_):
+            """Raise when called so that importing this module doesn't throw
+            ImportError when ds_detect() returns false. In this case, crypt
+            and passlib are not needed.
+            """
+            raise ImportError(
+                "crypt and passlib not found, missing dependency"
+            )
+
+
 LOG = logging.getLogger(__name__)
 
 DS_NAME = "Azure"
@@ -1766,7 +1788,7 @@ def read_azure_ovf(contents):
 
 
 def encrypt_pass(password):
-    return passlib.hash.sha512_crypt.encrypt(password)
+    return hash(password)
 
 
 @azure_ds_telemetry_reporter
```

holmanb (changes requested):
See previous comment

goneri:
Hi @holmanb! I won't have time to work on this during the 3 coming weeks. Feel free to take this over. Also, I'm not sure it's a good idea to dynamically switch between different kind of crypto lib. I'm afraid this will likely increase the testing matrix and delay potential early feedback regarding passlib.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1764,8 +1765,8 @@ def read_azure_ovf(contents):
     return (md, ud, cfg)
 
 
-def encrypt_pass(password, salt_id="$6$"):
-    return crypt.crypt(password, salt_id + util.rand_str(strlen=16))
+def encrypt_pass(password):
+    return passlib.hash.sha512_crypt.encrypt(password)
```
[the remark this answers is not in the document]
TheRealFalcon: This still needs to be updated

tests/integration_tests/instances.py
```diff
@@ -186,6 +186,7 @@ def install_deb(self):
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
+        assert self.execute("apt-get install -qy python3-passlib").ok
```
TheRealFalcon, line 189: This isn't the right place to put this. It should probably go in integration_tests/conftest.py/setup_image or perhaps the test itself.

TheRealFalcon (changes requested):
Comments are for @holmanb as he is taking over.

In addition to inline, for the patch you suggested, we shouldn't be defining `hash`. It is a built-in function in python already. Let's choose a different word like `crypt_hash` or something.

Additionally, I'm not sure what we need to do to ensure Azure and netbsd know of this change and get the dependency added to their images...but we probably need to do something.

tests/integration_tests/instances.py
holmanb: I updated this to be an integration test requirement.

holmanb:
> Comments are for @holmanb as he is taking over.
> 
> In addition to inline, for the patch you suggested, we shouldn't be defining hash. It is a built-in function in python already. Let's choose a different word like crypt_hash or something.

Oops, right.

> Additionally, I'm not sure what we need to do to ensure Azure and netbsd know of this change and get the dependency added to their images...but we probably need to do something.

Agreed. Initially this shouldn't affect them but I can put together an email to the mailing list to broadcast this information.

TheRealFalcon (approved):
LGTM. Integration tests failed, but it might be Github related. I'm gonna trigger a re-run.

cjp256 (approved):

holmanb:
@TheRealFalcon I know you already approved this PR, but could you please take a look at the additions I made, specifically https://github.com/canonical/cloud-init/pull/2139/commits/772d492cbc8b343495f3fea2011555e0c2462d0f? This change is currently required to pull in the dependency, but it's a change to the way integration tests get setup so if we'd rather not go that route I can just go back to the manual package install Goneri included which I reverted in https://github.com/canonical/cloud-init/pull/2139/commits/cdc0cc6ad934b4a27e752ab3516328d0147ad52d.

Thoughts?

TheRealFalcon:
@holmanb , yeah, `apt install`ing the package can work. I'm wondering if we should just remove the library from `requirements.txt` though. [CI calls bddeb](https://github.com/canonical/cloud-init/blob/main/.github/workflows/integration.yml/#L41) and [bddeb calls read-dependencies](https://github.com/canonical/cloud-init/blob/main/packages/bddeb/#L99), so we wind up with a `python3-passlib` requirement in the `control` file of the built package.

If it's a runtime dependency expected to already exist on the system, should it be listed in the requirements for the package?

If we did remove it from the requirements file, we'd have to make sure it gets installed in the test that needs in though. Looking again, `integration-requirements.txt` doesn't do that, and I think we should remove it from that file, as those are requirements to run the integration tests on the host machine, not requirements inside the tests themselves).

holmanb:
> @holmanb , yeah, apt installing the package can work. I'm wondering if we should just remove the library from requirements.txt though. [CI calls bddeb](https://github.com/canonical/cloud-init/blob/main/.github/workflows/integration.yml/#L41) and [bddeb calls read-dependencies](https://github.com/canonical/cloud-init/blob/main/packages/bddeb/#L99), so we wind up with a python3-passlib requirement in the control file of the built package.

I'll revert my changes then. I don't feel strongly about which approach to do, since they both get the same result.

> If it's a runtime dependency expected to already exist on the system, should it be listed in the requirements for the package?

No, but it being a runtime dependency isn't yet implemented in this PR (which is why tests fail when it isn't manually installed). I suppose that since the next PR will be back to back it won't matter that much.

> If we did remove it from the requirements file, we'd have to make sure it gets installed in the test that needs in though. Looking again, integration-requirements.txt doesn't do that, and I think we should remove it from that file, as those are requirements to run the integration tests on the host machine, not requirements inside the tests themselves).

+1 good point

TheRealFalcon (approved):

holmanb (approved):

PR #2141, opened 2023-04-25
sources/azure: report success to host and introduce kvp module

cjp256:
Add success reporting to the host via KVP.

- Move _report_failure_to_host() into kvp module.

- Tweak error description to use result=error instead of PROVISIONING_ERROR: ...

- Use result=success for the successful ("ready") reports.

The format used is subject to change and/or removal.

### DIFF
--- a/cloudinit/sources/azure/kvp.py
+++ b/cloudinit/sources/azure/kvp.py
@@ -35,11 +35,11 @@ def report_via_kvp(report: str) -> bool:
     return True
 
 
-def report_failure_via_kvp(error: errors.ReportableError) -> bool:
-    return report_via_kvp(error.as_description())
+def report_failure_to_host(error: errors.ReportableError) -> bool:
+    return report_via_kvp(error.as_encoded_report())
 
 
-def report_success_via_kvp() -> bool:
+def report_success_to_host() -> bool:
     try:
         vm_id = identity.query_vm_id()
     except Exception as id_error:
--- a/cloudinit/sources/azure/kvp.py
+++ b/cloudinit/sources/azure/kvp.py
@@ -1,0 +1,42 @@
+# Copyright (C) 2022 Microsoft Corporation.
+#
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import logging
+from typing import Optional
+
+from cloudinit.reporting import handlers, instantiated_handler_registry
+from cloudinit.sources.azure import errors
+
+LOG = logging.getLogger(__name__)
+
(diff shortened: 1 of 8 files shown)


cloudinit/sources/azure/kvp.py
```diff
@@ -0,0 +1,42 @@
...
+LOG = logging.getLogger(__name__)
+
+
+def get_kvp_handler() -> Optional[handlers.HyperVKvpReportingHandler]:
+    """Get instantiated KVP telemetry handler."""
+    kvp_handler = instantiated_handler_registry.registered_items.get(
+        "telemetry"
+    )
+    if not isinstance(kvp_handler, handlers.HyperVKvpReportingHandler):
+        LOG.debug("KVP handler not enabled, skipping host report.")
```
anhvoms, line 20: Remove this debug print. Caller of this method can decice whether to print any debug or not. Currently report_via_kvp will also print a message if kvp is not enabled which means the same message is printed twice

cloudinit/sources/azure/kvp.py
```diff
@@ -0,0 +1,42 @@
...
+    kvp_handler.write_key("PROVISIONING_REPORT", report)
+    return True
+
+
+def report_failure_via_kvp(error: errors.ReportableError) -> bool:
+    return report_via_kvp(error.as_description())
+
+
+def report_success_via_kvp() -> bool:
+    return report_via_kvp("result=success")
```
anhvoms, line 42: Do you need to report the vm_id and timestamp here as well? The same reason you're reporting supporting_data for result=failure? (i.e., the result=success could be from stale KVP such as the ones from source PPS VM)

cloudinit/sources/azure/kvp.py
cjp256: whoops, that was not meant to be there. thanks!

cloudinit/sources/azure/kvp.py
cjp256: excellent point! will include

cloudinit/sources/azure/kvp.py
cjp256: fixed

cloudinit/sources/azure/kvp.py
cjp256: done

anhvoms (approved):
LGTM

TheRealFalcon (approved):
LGTM!

PR #2142, opened 2023-04-26
Do not generate dsa and ed25519 key types when crypto FIPS mode is enabled

ani-sinha:
DSA and ED25519 key types are not supported when FIPS is enabled in crypto. Check if FIPS has been enabled on the system and if so, do not generate those key types. Presently the check is only available on Linux systems.

Test Steps:
Steps to Reproduce:
Manual:
1. Boot into an RHEL-9.1 system with fips enabled
2. Try to clean and init cloud-init again
$sudo cloud-init clean
$sudo cloud-init init

LP: 2017761
RHBZ: 2187164

### DIFF
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -173,6 +173,8 @@ __doc__ = get_meta_doc(meta)
 LOG = logging.getLogger(__name__)
 
 GENERATE_KEY_NAMES = ["rsa", "dsa", "ecdsa", "ed25519"]
+FIPS_UNSUPPORTED_KEY_NAMES = ["dsa", "ed25519"]
+
 pattern_unsupported_config_keys = re.compile(
     "^(ecdsa-sk|ed25519-sk)_(private|public|certificate)$"
 )
@@ -258,9 +260,26 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         genkeys = util.get_cfg_option_list(
             cfg, "ssh_genkeytypes", GENERATE_KEY_NAMES
         )
+        # remove keys that are not supported in fips mode if its enabled
+        key_names = (
+            genkeys
+            if not util.fips_enabled()
+            else [
+                names
+                for names in genkeys
+                if names not in FIPS_UNSUPPORTED_KEY_NAMES
+            ]
+        )
+        skipped_keys = set(genkeys).difference(key_names)
+        if skipped_keys:
+            LOG.debug(
+                "skipping keys that are not supported in fips mode: %s",
+                ",".join(skipped_keys),
+            )
+
         lang_c = os.environ.copy()
         lang_c["LANG"] = "C"
-        for keytype in genkeys:
+        for keytype in key_names:
             keyfile = KEY_FILE_TPL % (keytype)
             if os.path.exists(keyfile):
                 continue
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -1578,6 +1578,18 @@ def get_cmdline():
     return _get_cmdline()
 
 
+def fips_enabled() -> bool:
+    fips_proc = "/proc/sys/crypto/fips_enabled"
+    try:
+        contents = load_file(fips_proc).strip()
+        return contents == "1"
+    except (IOError, OSError):
+        # for BSD systems and Linux systems where the proc entry is not
+        # available, we assume FIPS is disabled to retain the old behavior
+        # for now.
+        return False
+
+
 def pipe_in_out(in_fh, out_fh, chunk_size=1024, chunk_cb=None):
     bytes_piped = 0
     while True:
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -101,11 +101,16 @@ class TestHandleSsh:
             expected_calls = [mock.call(set(keys), user)] + expected_calls
         assert expected_calls == m_setup_keys.call_args_list
 
+    @pytest.mark.parametrize("fips_enabled", (True, False))
     @mock.patch(MODPATH + "glob.glob")
     @mock.patch(MODPATH + "ug_util.normalize_users_groups")
     @mock.patch(MODPATH + "os.path.exists")
-    def test_handle_no_cfg(self, m_path_exists, m_nug, m_glob, m_setup_keys):
+    @mock.patch(MODPATH + "util.fips_enabled")
+    def test_handle_no_cfg(
+        self, m_fips, m_path_exists, m_nug, m_glob, m_setup_keys, fips_enabled
+    ):
         """Test handle with no config ignores generating existing keyfiles."""
+        m_fips.return_value = fips_enabled
         cfg = {}
         keys = ["key1"]
         m_glob.return_value = []  # Return no matching keys to prevent removal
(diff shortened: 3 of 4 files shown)


cloudinit/util.py
```diff
@@ -1578,6 +1578,17 @@ def get_cmdline():
     return _get_cmdline()
 
 
+def fips_enabled():
+    fips_proc = "/proc/sys/crypto/fips_enabled"
+    if is_Linux() and os.path.exists(fips_proc):
+        contents = load_file(fips_proc).strip()
+        return True if contents == "1" else False
```
blackboxsw, line 1585: We probably don't need to check `is_Linux` first as fallback to `False` is the same in all cases if we don't have a proc file. We could just handle this w/ try/except around the load_file attempt if we wanted.
```suggestion
    try:                                                                        
        contents = load_file(fn)
        return contents.strip() == "1"
    except (IOError, OSError):
        # On systems where the proc entry is not available,
        # we assume FIPS is disabled to retain the old behavior.
        return False
```

cloudinit/config/cc_ssh.py
```diff
@@ -258,9 +260,19 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         genkeys = util.get_cfg_option_list(
             cfg, "ssh_genkeytypes", GENERATE_KEY_NAMES
         )
+        # remove keys that are not supported in fips mode if its enabled
+        key_names = (
```
blackboxsw, line 264: It may be worth a LOG.debug mentioning the matched and skipped  FIPS_UNSUPPORTED_KEY_NAMES.

We could perform a `set(genkeys).difference(key_names)` to get that value for the log

ani-sinha: done.

cloudinit/util.py
ani-sinha: done.

cloudinit/config/cc_ssh.py
```diff
@@ -258,9 +260,26 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
...
+                for names in genkeys
+                if names not in FIPS_UNSUPPORTED_KEY_NAMES
+            ]
+        )
+        skipped_keys = set(genkeys).difference(key_names)
+        if skipped_keys:
+            LOG.debug(
+                "skipping keys that are not supported in fips mode: %s",
+                ",".join(skipped_keys),
+            )
```
blackboxsw, line 278: On Ubuntu FIPS (18.04) 4.15 based kernel I'm seeing that ssh-keygen of rsa, ecdsa and ed25519 succeed, and only dsa type fails(See log below).

 I'm wondering if, instead of avoiding the attempt to generate certain keys, we instead want to conditionally process our exception handling `if (fips_enabled and keytype in FIPS_UNSUPPORTED_KEY_NAMES)` avoid calling `log.exec(... Failed generating key type`  but instead just `LOG.debug('Unable to generate key type {keytype} because FIPS is enabled')`


```
Ubuntu FIPS 4.15 based on Ubuntu 18.04 
$ egrep 'WARN|ssh-keygen|Stderr:' /var/log/cloud-init.log
Stderr: mount: /run/cloud-init/tmp/tmpenfrf09s: no medium found on /dev/sr0.
2023-04-28 03:53:29,950 - subp.py[DEBUG]: Running command ['ssh-keygen', '-t', 'rsa', '-N', '', '-f', '/etc/ssh/ssh_host_rsa_key'] with allowed return codes [0] (shell=False, capture=True)
2023-04-28 03:53:30,114 - subp.py[DEBUG]: Running command ['ssh-keygen', '-t', 'dsa', '-N', '', '-f', '/etc/ssh/ssh_host_dsa_key'] with allowed return codes [0] (shell=False, capture=True)
2023-04-28 03:53:30,141 - util.py[WARNING]: Failed generating key type dsa to file /etc/ssh/ssh_host_dsa_key
Command: ['ssh-keygen', '-t', 'dsa', '-N', '', '-f', '/etc/ssh/ssh_host_dsa_key']
Stderr: DSA keys are not allowed in FIPS mode
2023-04-28 03:53:30,148 - subp.py[DEBUG]: Running command ['ssh-keygen', '-t', 'ecdsa', '-N', '', '-f', '/etc/ssh/ssh_host_ecdsa_key'] with allowed return codes [0] (shell=False, capture=True)
2023-04-28 03:53:30,177 - subp.py[DEBUG]: Running command ['ssh-keygen', '-t', 'ed25519', '-N', '', '-f', '/etc/ssh/ssh_host_ed25519_key'] with allowed return codes [0] (shell=False, capture=True)
```

blackboxsw (commented):
Thank you for this changeset and proposal for cloud-init upstream.

I have reflected this Pull request as well to Canonical's security team for additional review as there may additionally be something to look at with regard to whether the behavior is strict for supported key generation types on all distros for both dsa and ed25519 key types.

cloudinit/config/cc_ssh.py
ani-sinha: As per our doc here: https://access.redhat.com/solutions/3643252 both dsa and ed25519 keys are not allowed in fips mode. I will check if there is a more standard spec on this.

ani-sinha: https://csrc.nist.gov/CSRC/media/projects/cryptographic-module-validation-program/documents/security-policies/140sp2472.pdf I found this in section 9.1 under "In addition to the configuration of the kernel, the OpenSSH client configuration should contain the following rules:"

Remove or do not configure Ed25519 host keys as Ed25519 keys are not allowed in FIPS Approved
mode. The crypto office should inform the user to not use Ed25519 keys for user authentication.

ani-sinha: In the bug report I have added the specific commit in fedora that adds the error message in ssh-keygen. So all recent redhat distros are affected. Ubuntu may not be affected by the same way.

blackboxsw: @ani-sinha I got confirmation that minimally we don't want cloud-init broadly to generate keys unless permitted due to FIPS either being disabled or not present as generating those keys results in bringing that system into non-compliance with FIPS anyway (which you already explained). Aside from cloud-init behavior, we'll take homework to look at what else may need to be in place to ensure a more noisy indication on Ubuntu if entering into non-FIPS compliant mode due to key generation. Ideally I'd like to see something from FIPS solutions that would advertize what certification/policy of FIPS a platform adheres to if there is possibility that newer certifications have change in behavior for cloud-init in terms of key type generation.

tests/unittests/config/test_cc_ssh.py
```diff
@@ -131,8 +135,9 @@ def test_handle_no_cfg(self, m_path_exists, m_nug, m_glob, m_setup_keys):
     @mock.patch(MODPATH + "glob.glob")
     @mock.patch(MODPATH + "ug_util.normalize_users_groups")
     @mock.patch(MODPATH + "os.path.exists")
+    @mock.patch(MODPATH + "util.fips_enabled", return_value=False)
```
blackboxsw, line 138: Let's add a positive and negative test case here too to show avoidance of key creation
```diff
diff --git a/tests/unittests/config/test_cc_ssh.py b/tests/unittests/config/test_cc_ssh.py
index 902dea8ce..83bfd6543 100644
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -122,12 +124,21 @@ class TestHandleSsh:
         options = options.replace("$DISABLE_USER", "root")
         m_glob.assert_called_once_with("/etc/ssh/ssh_host_*key*")
         m_fips.assert_called_once()
-        assert [
-            mock.call("/etc/ssh/ssh_host_rsa_key"),
-            mock.call("/etc/ssh/ssh_host_dsa_key"),
-            mock.call("/etc/ssh/ssh_host_ecdsa_key"),
-            mock.call("/etc/ssh/ssh_host_ed25519_key"),
-        ] in m_path_exists.call_args_list
+
+        if not m_fips():
+            expected_calls = [
+                mock.call("/etc/ssh/ssh_host_rsa_key"),
+                mock.call("/etc/ssh/ssh_host_dsa_key"),
+                mock.call("/etc/ssh/ssh_host_ecdsa_key"),
+                mock.call("/etc/ssh/ssh_host_ed25519_key"),
+            ]
+        else:
+            # Enabled fips doesn't generate dsa or ed25519
+            expected_calls = [
+                mock.call("/etc/ssh/ssh_host_rsa_key"),
+                mock.call("/etc/ssh/ssh_host_ecdsa_key"),
+            ]
+        assert expected_calls in m_path_exists.call_args_list
         assert [
             mock.call(set(keys), "root", options=options)
         ] == m_setup_keys.call_args_list
```

blackboxsw (changes requested):
Just a couple unit tests away I think

tests/unittests/config/test_cc_ssh.py
ani-sinha: applied.

cloudinit/util.py
```diff
@@ -1578,6 +1578,18 @@ def get_cmdline():
     return _get_cmdline()
 
 
+def fips_enabled():
```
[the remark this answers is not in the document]
ani-sinha: applied.

ani-sinha:
> Just a couple unit tests away I think

done.

blackboxsw (approved):
Thank you for these contributions @ani-sinha this changeset makes sense for cloud-init and it improves the behaviors related to FIPS-enforcing systems broadly. This will make it into our 23.2 upstream release in a couple weeks (~ May 15th)

PR #2143, opened 2023-04-26
Fix linux-specific call on BSD.

holmanb:
```
Fix linux-specific call on BSDs.

util.get_mount_info() which is called from the
distro base class is linux-specific
```

## Additional Context
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/cloudinit/sources/__init__.py", line 946, in find_source
    if s.update_metadata_if_supported(
  File "/usr/local/lib/python3.9/site-packages/cloudinit/sources/__init__.py", line 828, in update_metadata_if_supported
    result = self.get_data()
  File "/usr/local/lib/python3.9/site-packages/cloudinit/sources/__init__.py", line 373, in get_data
    return_value = self._get_data()
  File "/usr/local/lib/python3.9/site-packages/cloudinit/sources/DataSourceOpenStack.py", line 161, in _get_data
    tmp_dir=self.distro.get_tmp_exec_path(),
  File "/usr/local/lib/python3.9/site-packages/cloudinit/distros/__init__.py", line 961, in get_tmp_exec_path
    if not util.has_mount_opt(tmp_dir, "noexec"):
  File "/usr/local/lib/python3.9/site-packages/cloudinit/util.py", line 2650, in has_mount_opt
    *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
TypeError: cannot unpack non-iterable NoneType object
```

Traceback doesn't occur after this change.

PR #2146, opened 2023-04-26
FreeBSD fix parsing of mount and mount options

igalic:
## Proposed Commit Message

```
FreeBSD fix parsing of mount and mount options

The regular expressions used to parse BSD mount(8) are both
overly complicated and limited.

- Combine the two regular expressions into one
- clarify regular expressions with named groups
- add parsing of mount options
- Attempt to extract mount-parent sub-path finding code from  parse_mount_info()
- Failing that, duplicate it in parse_mount()

This leaves parse_mtab() as only mount info provider without options parsing, for now.

```

## Additional Context

a hopefully better fix for #2143 

## Checklist:

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2626,38 +2626,36 @@ def get_device_info_from_zpool(zpool):
             return disk
 
 
-def parse_mount(path):
+def parse_mount(path, get_mnt_opts=False):
     (mountoutput, _err) = subp.subp(["mount"])
     mount_locs = mountoutput.splitlines()
     # there are 2 types of mount outputs we have to parse therefore
     # the regex is a bit complex. to better understand this regex see:
-    # https://regex101.com/r/2F6c1k/1
-    # https://regex101.com/r/T2en7a/1
+    # https://regex101.com/r/L51Td8/1
     regex = (
-        r"^(/dev/[\S]+|.*zroot\S*?) on (/[\S]*) "
-        r"(?=(?:type)[\s]+([\S]+)|\(([^,]*))"
+        r"^(?P<devpth>[\S]+?) on (?P<mountpoint>[\S]+?) "
+        r"(\(|type )(?P<type>[^,\(\) ]+)( \()?(?P<options>.*?)\)$"
     )
-    if is_DragonFlyBSD():
-        regex = (
-            r"^(/dev/[\S]+|\S*?) on (/[\S]*) "
-            r"(?=(?:type)[\s]+([\S]+)|\(([^,]*))"
-        )
+
+    mountpoints = []
     for line in mount_locs:
         m = re.search(regex, line)
         if not m:
             continue
-        devpth = m.group(1)
-        mount_point = m.group(2)
-        # above regex will either fill the fs_type in group(3)
-        # or group(4) depending on the format we have.
-        fs_type = m.group(3)
-        if fs_type is None:
-            fs_type = m.group(4)
+        devpth = m.group("devpth")
+        mount_point = m.group("mountpoint")
+        mountpoints.append(mount_point)
+        fs_type = m.group("type")
+        options = m.group("options")
+        if options is not None:
+            options = ",".join(m.group("options").strip().split(", "))
         LOG.debug(
-            "found line in mount -> devpth: %s, mount_point: %s, fs_type: %s",
+            "found line in mount -> devpth: %s, mount_point: %s, fs_type: %s"
+            ", options: %s",
             devpth,
             mount_point,
             fs_type,
+            options,
         )
         # check whether the dev refers to a label on FreeBSD
         # for example, if dev is '/dev/label/rootfs', we should
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2592,7 +2592,7 @@ def get_path_dev_freebsd(path, mnt_list):
     return path_found
 
 
-def get_mount_info_freebsd(path):
+def get_freebsd_devpth(path):
     (result, err) = subp.subp(["mount", "-p", path], rcs=[0, 1])
     if len(err):
         # find a path if the input is not a mounting point
@@ -2603,7 +2603,7 @@ def get_mount_info_freebsd(path):
         result = path_found
     ret = result.split()
     label_part = find_freebsd_part(ret[0])
-    return "/dev/" + label_part, ret[2], ret[1]
+    return "/dev/" + label_part
 
 
 def get_device_info_from_zpool(zpool):
@@ -2628,7 +2628,7 @@ def get_device_info_from_zpool(zpool):
 
 def parse_mount(path, get_mnt_opts=False):
     (mountoutput, _err) = subp.subp(["mount"])
-    mount_locs = mountoutput.splitlines()
+
     # there are 2 types of mount outputs we have to parse therefore
     # the regex is a bit complex. to better understand this regex see:
     # https://regex101.com/r/L51Td8/1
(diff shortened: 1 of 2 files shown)


igalic:
Note that two tests are failing, and it's not quite clear why.

CalvoM:
@igalic I have run the tests I think for the `TestMountinfoParsing.test_parse_mount_with_ext` test failure, we fail since the last line of `mount_parse_ext.txt` mock file is 
`10.0.1.1:/backup on /backup type nfs (rw,noexec,nosuid,nodev,bg,nolock,tcp,nfsvers=3,hard,addr=10.0.1.1)` which has valid `devpth`, `fs_type` and `match_mount_point` thus after the loop is done, thus [line](https://github.com/igalic/cloud-init/blob/e0bd0ab6958643f64b7be93ac4eb273fbb265a63/cloudinit/util.py#L2676) will evaluate to true.

cloudinit/util.py
```diff
@@ -2603,38 +2602,60 @@ def get_device_info_from_zpool(zpool):
...
+        # Ignore mounts deeper than the path in question.
+        if len(mount_point_elements) > len(path_elements):
+            continue
+
+        # Ignore mounts where the common path is not the same.
+        x = min(len(mount_point_elements), len(path_elements))
+        if mount_point_elements[0:x] != path_elements[0:x]:
+            continue
+
+        # Ignore mount points higher than an already seen mount
```
CalvoM, line 2636: a small nitpick, I assume you mean `and` instead of `an`.

cloudinit/util.py
```diff
@@ -2603,38 +2602,60 @@ def get_device_info_from_zpool(zpool):
...
+        # point.
+        if match_mount_point_elements is not None and len(
+            match_mount_point_elements
```
CalvoM, line 2639: Quick question on FreeBSD, is the mount output arranged in some sort of hierarchy? e.g. 
```
path on /path/1/2/3
path2 on /path/2/
```

igalic: if anything, it's going the other way around. See https://gist.github.com/igalic/89da01454ad25316b168efbf960b0070

cloudinit/util.py
CalvoM: Oh, makes sense. Sorry I had misread the comment.

igalic:
> @igalic I have run the tests. I think for the test failures, we fail since the last line of `mount_parse_ext.txt` and `mount_parse_zfs.txt` mock files are parsed successfully by the regex. For example `10.0.1.1:/backup on /backup type nfs (rw,noexec,nosuid,nodev,bg,nolock,tcp,nfsvers=3,hard,addr=10.0.1.1)` which has valid `devpth`, `fs_type` and `match_mount_point` thus after the loop is done, thus [line](https://github.com/igalic/cloud-init/blob/e0bd0ab6958643f64b7be93ac4eb273fbb265a63/cloudinit/util.py#L2676) will evaluate to true hence it will not return None.

I think the problem with the tests is that so far, `parse_mount()` was only looking for full matches, and now it does the same thing as `parse_mount_info()`

so, I think with permission, I'd declare the tests are faulty… and say that the `/non/existent/thing` ending up in `/` makes sense.

igalic:
The two test failures are … very interesting:

given

```
vmzroot/ROOT/freebsd on / (zfs, local, nfsv4acls)
devfs on /dev (devfs, local, multilabel)
fdescfs on /dev/fd (fdescfs)
vmzroot/root on /root (zfs, local, nfsv4acls)
vmzroot/tmp on /tmp (zfs, local, nosuid, nfsv4acls)
vmzroot/ROOT/freebsd/usr on /usr (zfs, local, nfsv4acls)
vmzroot/ROOT/freebsd/usr/local on /usr/local (zfs, local, nfsv4acls)
vmzroot/ROOT/freebsd/var on /var (zfs, local, nfsv4acls)
vmzroot/ROOT/freebsd/var/cache on /var/cache (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/ROOT/freebsd/var/crash on /var/crash (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/var/cron on /var/cron (zfs, local, nosuid, nfsv4acls)
vmzroot/ROOT/freebsd/var/db on /var/db (zfs, local, noatime, noexec, nosuid, nfsv4acls)
vmzroot/ROOT/freebsd/var/empty on /var/empty (zfs, local, noexec, nosuid, read-only, nfsv4acls)
vmzroot/var/log on /var/log (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/var/log/pf on /var/log/pf (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/var/mail on /var/mail (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/ROOT/freebsd/var/run on /var/run (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/var/spool on /var/spool (zfs, local, noexec, nosuid, nfsv4acls)
vmzroot/var/tmp on /var/tmp (zfs, local, nosuid, nfsv4acls)
10.0.0.1:/vol/test on /mnt/test (nfs, read-only)
10.0.0.2:/vol/tes2 on /mnt/test2 (nfs, nosuid)
```

we get this failure:

```
        ret = util.parse_mount("/not/existing/mount")
>       self.assertIsNone(ret)
E       AssertionError: ('10.0.0.2:/vol/tes2', 'zfs', '/') is not None
```

I'm not sure how that makes sense
We "match" the last device, but return the root and it's fstype

same in the other case. Given:

```
/dev/mapper/vg00-lv_root on / type ext4 (rw,errors=remount-ro)
proc on /proc type proc (rw,noexec,nosuid,nodev)
sysfs on /sys type sysfs (rw,noexec,nosuid,nodev)
none on /sys/fs/cgroup type tmpfs (rw)
none on /sys/fs/fuse/connections type fusectl (rw)
none on /sys/kernel/debug type debugfs (rw)
none on /sys/kernel/security type securityfs (rw)
udev on /dev type devtmpfs (rw,mode=0755)
devpts on /dev/pts type devpts (rw,noexec,nosuid,gid=5,mode=0620)
none on /tmp type tmpfs (rw)
tmpfs on /run type tmpfs (rw,noexec,nosuid,size=10%,mode=0755)
none on /run/lock type tmpfs (rw,noexec,nosuid,nodev,size=5242880)
none on /run/shm type tmpfs (rw,nosuid,nodev)
none on /run/user type tmpfs (rw,noexec,nosuid,nodev,size=104857600,mode=0755)
none on /sys/fs/pstore type pstore (rw)
/dev/mapper/vg00-lv_var on /var type ext4 (rw)
rpc_pipefs on /run/rpc_pipefs type rpc_pipefs (rw)
systemd on /sys/fs/cgroup/systemd type cgroup (rw,noexec,nosuid,nodev,none,name=systemd)
10.0.1.1:/backup on /backup type nfs (rw,noexec,nosuid,nodev,bg,nolock,tcp,nfsvers=3,hard,addr=10.0.1.1)
```

we get this failure:

```
        # this one does not even exist in mount_parse_ext.txt
        ret = util.parse_mount("/not/existing/mount")
>       self.assertIsNone(ret)
E       AssertionError: ('10.0.1.1:/backup', 'ext4', '/') is not None
```

that looks a bit jumbled up and I wonder where that jumbling is happening

tests/unittests/test_util.py
```diff
@@ -2200,7 +2200,7 @@ def test_parse_mount_with_ext(self, mount_out):
         self.assertEqual(("/dev/mapper/vg00-lv_root", "ext4", "/"), ret)
         # this one exists in mount_parse_ext.txt
         ret = util.parse_mount("/sys/kernel/debug")
-        self.assertIsNone(ret)
+        self.assertEqual(("none", "debugfs", "/sys/kernel/debug"), ret)
```
holmanb, line 2203: why does this one change?

igalic: before it didn't parse, i don't know why.
now it just does.

holmanb (approved):
LGTM

PR #2147, opened 2023-04-27
OpenBSD: Added support for static routes

kadiron:
## Proposed Commit Message
OpenBSD: Added support for static routes

```
This adds support for static routes on OpenBSD.
Note that since BSDs see routes as per-interface-configurations, you have to specify your routes inside the network interface context.
```

### DIFF
--- a/cloudinit/net/openbsd.py
+++ b/cloudinit/net/openbsd.py
@@ -4,7 +4,7 @@ import platform
 
 import cloudinit.net.bsd
 from cloudinit import log as logging
-from cloudinit import net, subp, util
+from cloudinit import subp, util
 
 LOG = logging.getLogger(__name__)
 
@@ -28,7 +28,7 @@ class Renderer(cloudinit.net.bsd.BSDRenderer):
                 mtu = v.get("mtu")
                 if mtu:
                     content += " mtu %d" % mtu
-                content += "\n"
+                content += "\n" + self.interface_routes
             util.write_file(fn, content)
 
     def start_services(self, run=False):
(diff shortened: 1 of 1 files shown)


cloudinit/net/openbsd.py
```diff
@@ -43,7 +43,7 @@ def start_services(self, run=False):
                     ["dhcpleasectl", "-w", "30", interface], capture=True
                 )
         else:
-            net.dhcp.IscDhclient.kill_dhcp_client()
+            subp.subp(["pkill", "dhclient"], capture=True, rcs=[0, 1])
```
goneri, line 46: Hi, I'm not sure this needed. `kill_dhcp_client()` does the same thing already.

cloudinit/net/openbsd.py
```diff
@@ -54,6 +54,12 @@ def set_route(self, network, netmask, gateway):
             fn = subp.target_path(self.target, if_file)
             content = gateway + "\n"
             util.write_file(fn, content)
+        else:
+            try:
```
goneri, line 58: It would be more elegant to define the attribute in the constructor.

goneri (changes requested):

cloudinit/net/openbsd.py
kadiron: Hi and sorry, you're right. I don't know what's happened here, I didn't wanted to change this. It's the line as it is in my clone from main at that day, so just ignore that please.

cloudinit/net/openbsd.py
kadiron: Sure, I'm aware of this being hacky. But tbh, I don't know how to do this right -- I hardly ever had to do anything in python yet.

kadiron:
Here is one version conflict and something hacky that shouldn't be like this. I'm starting over and will make a new pull request.
Thanks so far!

PR #2148, opened 2023-04-27
Switch supported distro setting to RH distros

rjschwei:
## Proposed Commit Message

```

The cc_refresh_rmc_and_interface module has hard coded paths that only exist on RH distributions. Having it enabled for ALL_DISTROS as was the case triggers failures if the module is set in the configuration file.

```

## Checklist:

 - [x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)


blackboxsw (commented):
@rjschwei thanks!. We might want to also tweak cc_reset_rmc.py distros to align?  We'll double-check w/ IBM on support exposure here for other distros and what should remain viable.

blackboxsw:
Awaiting feedback from IBM on the support need here.  Ultimately, if this module is only desired for RHEL7, and latest RedHat and Fedora doesn't need this logic, I'd probably be more interested in us dropping these modules altogether as I don't believe RHEL7 and derivatives have moved past cloud-init 19.4 due to being python 2.7 based. They will never receive the updates which drop rmc* related config in those cases due to incompatibility w/ elements of our py3 feature dependencies.

esposem:
@blackboxsw no, we don't expect rhel-7 to rebase into new cloud-init versions, also due to the python version difference.

blackboxsw:
> 

Thanks @Aman306 && @esposem that helps clarify here. So, @rjschwei I think we can drop cloudinit/config/cc_refresh_rmc_and_interface.py altogether and any unittests in cloud-init/tests/unittests/config/test_cc_refresh_rmc_and_interface.py

blackboxsw:
@rjschwei does this pushed update look okay? I'm dropping refresh_rmc_and_interface altogether as RHEL7 isn't getting updates for cloud-init from tip or main.


IBM Docs state clearly that this was a RHEL7-ism
https://www.ibm.com/docs/en/powervc/2.1.0?topic=linux-powervc-cloud-init-modules

rjschwei (commented):
@blackboxsw LGTM

blackboxsw (approved):
Let's bring it home

PR #2153, opened 2023-05-01
net: refactor hyper-v VF filtering and apply to get_interfaces()

cjp256:
Azure is introducing a new VF ("MANA") that will initially behave
    similarly to mlx4/5 but cannot be denylisted in the same manner.
    This is because the synthetic interface (hv_netvsc) will no longer
    be required to function in the future which means we must
    intelligently filter the VFs out instead of relying solely on the
    driver name.

    - Isolate filtering logic for Hyper-V's SR-IOV/VFs when used
      with synthetic hv_netvsc interfaces.

    - Move the filter up to get_interfaces() from
      get_interfaces_by_mac_on_linux() to increase coverage of the
      filter. With this in place, we should be able to purge the
      "blacklist_drivers" across the codebase as it will no longer be
      necessary unless there are other paths to be considered.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -1127,13 +1127,14 @@ def get_interfaces(
 
     # Last-pass filter(s) which need the full device list to perform properly.
     if filter_hyperv_vf_with_synthetic:
-        filter_hyperv_vf_with_synthetic_interface(ret)
+        filter_hyperv_vf_with_synthetic_interface(filtered_logger, ret)
 
     return ret
 
 
 def filter_hyperv_vf_with_synthetic_interface(
-    interfaces: List[Tuple[str, str, str, str]]
+    filtered_logger: Callable[..., None],
+    interfaces: List[Tuple[str, str, str, str]],
 ) -> None:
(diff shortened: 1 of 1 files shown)


cjp256:
While current tests are passing, I plan to update them if PR is acceptable.

cloudinit/net/__init__.py
```diff
@@ -1004,7 +1004,7 @@ def get_interfaces_by_mac_on_linux(blacklist_drivers=None) -> dict:
 
     Bridges and any devices that have a 'stolen' mac are excluded."""
     ret: dict = {}
-    driver_map: dict = {}
+
```
anhvoms, line 1007: I think we can remove the raise_duplicate_mac_error variable and its check now that the hv_netvsc processing bit is refactored out.
if mac in ret:
    if driver in in ("fsl_enetc", "mscc_felix", "qmi_wwan"):
         ...
         continue
     raise RuntimeError(...)

cjp256:
This would be rebased on #2159

anhvoms (approved):

TheRealFalcon (approved):
LGTM!

cjp256:
rebased and passed filtered_logger

PR #2156, opened 2023-05-02
Add documentation label automatically

holmanb:
```
Add documentation label automatically
```

## Additional Context
All bugs with this label assigned can be viewed at the following url: https://github.com/canonical/cloud-init/labels/Documentation

## Test
See test [here](https://github.com/holmanb/cloud-init/pull/49).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,2 @@
+Documentation:
+- doc/*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,2 +1,2 @@
-Documentation:
+documentation:
 - doc/*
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
LGTM, but since we don't currently have any "Documentation" labeled PRs, can we change it to "documentation"? By default Github labels are lowercase, and for new labels I'd prefer we follow that convention.

s-makin:
> LGTM, but since we don't currently have any "Documentation" labeled PRs, can we change it to "documentation"? By default Github labels are lowercase, and for new labels I'd prefer we follow that convention.

Is it possible to also add the "conf.py" file? Changes in there are more often than not docs-related somehow so it'd be good to also track that

s-makin (approved):

igalic (approved):
that seems very useful!

TheRealFalcon (approved):

PR #2159, opened 2023-05-02
net: refactor find_candidate_nics_on_linux() to use get_interfaces()

cjp256:
Much of the filtering logic used in get_interfaces() is duplicated in find_candidate_nics_on_linux().  Consolidate the two interfaces by updating find_candidate_nics_on_linux() to use get_interfaces() and provide boolean toggles to maintain existing behavior.

The following parameters have been added to get_interfaces():
- log_filtered_reasons: bool = False
- filter_openvswitch_internal: bool = True
- filter_vlan: bool = True
- filter_without_own_mac: bool = True
- filter_zero_mac: bool = True

It may be that these deltas are unwanted or harmless, but the toggles could be removed independently with more deliberate consideration. Similarly, logging could be extended if desired for log_filtered_reasons.

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -519,6 +519,7 @@ def find_candidate_nics_on_linux(
     for interface, _, _, _ in get_interfaces(
         blacklist_drivers=blacklist_drivers,
         filter_openvswitch_internal=False,
+        filter_slave_if_master_not_bridge_bond_openvswitch=False,
         filter_vlan=False,
         filter_without_own_mac=False,
         filter_zero_mac=False,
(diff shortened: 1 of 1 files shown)


anhvoms (approved):
LGTM!

TheRealFalcon (approved):
I think the new arguments feel a bit awkward, but it's certainly an improvement, so LGTM.

TheRealFalcon:
@cjp256 , so do you want me to merge this one, then you'll rebase #2153 and we merge that one? Or something else?

cjp256:
> @cjp256 , so do you want me to merge this one, then you'll rebase #2153 and we merge that one? Or something else?

#2153 first, then i'll rebase this.

cjp256:
> @cjp256 , so do you want me to merge this one, then you'll rebase #2153 and we merge that one? Or something else?

Yeah, merge this first then I'll update 2153.

cjp256:
> I think the new arguments feel a bit awkward, but it's certainly an improvement, so LGTM.

Curious what you would prefer.  I

> I think the new arguments feel a bit awkward, but it's certainly an improvement, so LGTM.

Any thoughts on how to improve it? I can update this or come back to it later

TheRealFalcon:
> Any thoughts on how to improve it? I can update this or come back to it later

I think it's fine as-is. I think it's mainly because we're combining two functions without breaking any behavior, so that gets exposed as some extra options. I'm not sure there's a better way to do it without breaking behavior.

PR #2160, opened 2023-05-02
net: purge blacklist_drivers across net and azure

cjp256:
It was only used by Hyper-V which now has a filtering
    mechanism that does not require the use of a denylist.

    This exposed some issues with tests misspelling "hv_netvsc"
    and using unmatched mac addresses. This fixes those to work
    with the current filter that does not rely on the driver name.

### DIFF
--- a/cloudinit/distros/networking.py
+++ b/cloudinit/distros/networking.py
@@ -1,7 +1,6 @@
 import abc
 import logging
 import os
-from typing import List, Optional
 
 from cloudinit import net, subp, util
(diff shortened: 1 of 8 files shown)


cjp256:
@holmanb @TheRealFalcon @blackboxsw If you can add a quick review when you get a moment would be great.  With the release coming up, I just want to ensure there's nothing problematic with this queue.

TheRealFalcon (approved):
LGTM! It's good this is no longer needed and it cleans things up a lot.

PR #2163, opened 2023-05-03
Fix logger, use instance rather than module function

holmanb:
```
Fix logger calls

Use Logger instance rather than module function.
Add missing logger names.
```


cjp256 (approved):
LGTM

blackboxsw (approved):

PR #2165, opened 2023-05-03
Ephemeral Networking for FreeBSD

igalic:
## Proposed Commit Message

```
implement ephemeral networking for BSD

After #2127 lay the foundation, we now implement the BSD side of this

```

## Additional Context

This is based on #2127 

## Checklist:


cloudinit/distros/__init__.py
```diff
@@ -118,6 +120,7 @@ def __init__(self, name, cfg, paths):
         self.name = name
         self.networking: Networking = self.networking_cls()
         self.dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd]
+        self.net_ops = iproute2.Iproute2
```
cjp256, line 123: is this intended to be configurable?  defined as class variable and instance variable

igalic: that's a question for @holmanb

cloudinit/distros/__init__.py
```diff
@@ -998,6 +1001,26 @@ def do_as(self, command: list, user: str, cwd: str = "", **kwargs):
...
+            "-1",
+            "-v",
+            "-lf",
+            lease_file,
+            "-pf",
+            pid_file,
+            "-sf",
+            "/bin/true",
+        ] + (["-cf", config_file, interface] if config_file else [interface])
+
```
cjp256, line 1023: just a personal preference, but for readability, I think I'd rather see the args built up over a few lines and conditionals than trying to get it all into one statement. e.g.

```
args = [
            path,
            "-1",
            "-v",
            "-lf",
            lease_file,
            "-pf",
            pid_file,
            "-sf",
            "/bin/true",
        ]

if config_file:
    args.extend(["-cf", config_file, interface])

args.append(interface)
```

I may be alone in this preference, so feel free to ignore :D

igalic: i like the idea!
but this, too is @holmanb's code.

cloudinit/net/netops/bsd_netops.py
```diff
@@ -21,11 +21,15 @@ def add_route(
         gateway: Optional[str] = None,
         source_address: Optional[str] = None
     ):
-        pass
+        subp.subp(["route", "add", route, "-interface", interface])
+        if gateway and gateway != "0.0.0.0":
+            subp.subp(
+                ["route", "change", route, gateway],
+            )
```
igalic, line 28: We don't support source selection

and i had to make this addition in two steps, because when i did it in a single step, the result was nonsense.

cloudinit/net/netops/bsd_netops.py
```diff
@@ -21,11 +21,15 @@ def add_route(
...
 
     @staticmethod
     def append_route(interface: str, address: str, gateway: str):
-        pass
+        return BsdNetOps.add_route(interface, address=address, gateway=gateway)
```
igalic, line 32: we don't have append, so just pass thru to add

igalic (commented):
guidelines for reviewers

igalic:
that last commit should probably go into its own PR

goneri (approved):

cloudinit/distros/__init__.py
holmanb: > is this intended to be configurable? defined as class variable and instance variable

This is not intended to be configurable at runtime. This is defined as an instance variable to avoid unpickle issues across the upgrade path. It might not be required as a class variable, I don't recall why it is defined both ways.

holmanb (approved):
Looks good to me. Thanks!

PR #2167, opened 2023-05-05
azure/errors: add host reporting for dhcp errors

cjp256:
- Add host_only flag to _report_failure() to allow caller to only report the failure to host.  This is for cases where we don't want _report_failure() to attempt DHCP or we expect that we may recover from the reported error (there is no issue reporting multiple times to host, whereas fabric reports will immediately fail the VM provisioning).

- Add ReportableErrorDhcpLease() to report lease failures.

- Add ReportableErrorDhcpInterfaceNotFound() to report errors where the DHCP interface hasn't been found yet.

- Add TestReportFailure class with new test coverage.  Will migrate other _report_failure() tests in the future as they currently depend on TestAzureDataSource/CiTestCase.

Future work will add the interface name to supporting data, but as that information is not available with iface=None, another PR will explicitly add a call to net.find_fallback_nic() to specify it.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -390,7 +390,8 @@ class DataSourceAzure(sources.DataSource):
         )
 
         lease = None
-        timeout = timeout_minutes * 60 + time()
+        start_time = time()
+        deadline = start_time + timeout_minutes * 60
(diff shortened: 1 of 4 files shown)


TheRealFalcon (approved):
LGTM!

PR #2168, opened 2023-05-09
FreeBSD (and friends): better identify MBR slices

igalic:
## Proposed Commit Message

```
FreeBSD (and friends): better identify MBR slices

the code is already there, we just need to use it.

LP: #2016350
```

## Checklist:


cloudinit/config/cc_growpart.py
```diff
@@ -265,13 +265,15 @@ def device_part_info(devpath):
 
     # FreeBSD doesn't know of sysfs so just get everything we need from
     # the device, like /dev/vtbd0p2.
-    if util.is_FreeBSD():
-        freebsd_part = "/dev/" + util.find_freebsd_part(devpath)
-        m = re.search("^(/dev/.+)p([0-9])$", freebsd_part)
+    fpart = "/dev/" + util.find_freebsd_part(devpath)
```
TheRealFalcon, line 268: I think we still need a guard under this section otherwise I think linux devices could match here and unexpected return early. Can we use `util.is_BSD()`?

Another option would be to move `if not os.path.exists(syspath):` up here and then do the BSD-related stuff after checking that the Linux-specific path doesn't exist.

cloudinit/config/cc_growpart.py
```diff
@@ -263,16 +263,20 @@ def device_part_info(devpath):
...
-        dragonflybsd_part = "/dev/" + util.find_dragonflybsd_part(devpath)
-        m = re.search("^(/dev/.+)s([0-9])$", dragonflybsd_part)
-        return (m.group(1), m.group(2))
+    if util.is_BSD():
+        # FreeBSD doesn't know of sysfs so just get everything we need from
+        # the device, like /dev/vtbd0p2.
+        fpart = "/dev/" + util.find_freebsd_part(devpath)
+        m = re.search("^(/dev/.+)p([0-9])$", fpart)
+        if m:
+            return (m.group(1), m.group(2))
```
holmanb, line 272: fwiw, this is equivalent to:
```
            return m.group(1), m.group(2)
```

cloudinit/config/cc_growpart.py
```diff
@@ -263,16 +263,20 @@ def device_part_info(devpath):
...
+
+        # alternatively, we could be dealing with MBR slices,
+        # or we're on DragonFly
+        fslice = "/dev/" + util.find_dragonflybsd_part(devpath)
+        m = re.search("^(/dev/.+)s([0-9])$", fslice)
+        if m:
+            return (m.group(1), m.group(2))
```
holmanb, line 279: and again

cloudinit/config/cc_growpart.py
igalic: I would've expected black to take care of that, like it did in the loops

holmanb: Agreed, though I guess it doesn't have an opinion on this one.

igalic: just from looking at the other code in this file, and the codebase in general, `return (foo, bar)` seems to be way more common than `return foo, bar`

igalic: fun fact, and irrelevant aside: FreeBSD's [`style(9)`](https://man.freebsd.org/style(9)) *requires* parenthesis for a `return` statement.

(but that has to do with C's weird precedence rules)

holmanb: > just from looking at the other code in this file, and the codebase in general, return (foo, bar) seems to be slightly more common than return foo, bar

I'm not surprised. Similarly, the codebase also used to have more tuple unpacking using syntax with unnecessary parenthesis, i.e. `(a, b) = some_func()`, which hasn't been the style trend as of late, but which is the dominant style in many files still.

Stylistically, I try to avoid using parenthesis unless they are required or unless the code is confusing without them.

TheRealFalcon (approved):
LGTM!

PR #2334, opened 2023-05-09
FreeBSD: add ResizeGrowFS class to cc_growpart

igalic:
## Proposed Commit Message

```
FreeBSD: add ResizeGrowFS class to cc_growpart

this FreeBSD specific resizer resizes the root partition and grows the Filesystem all in one.
All we have to do is call ``service growfs onestart``
Document behaviour: especially that growfs will insert a swap partition
if none is present, unless instructed otherwise.

```

## Additional Context
This service has existed for a long time in FreeBSD, I think it's time we use it instead of reinventing the wheel.
caveat: This only works for the root device.

I think this should probably be rebased on #2168 .

## Checklist:

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -32,7 +32,9 @@ MODULE_DESCRIPTION = """\
 Growpart resizes partitions to fill the available disk space.
 This is useful for cloud instances with a larger amount of disk space available
 than the pristine image uses, as it allows the instance to automatically make
-use of the extra space.
+use of the extra space. Note that this only works if the partition to be
+resized is the last one on a disk with classic partitioning scheme (MBR, BSD,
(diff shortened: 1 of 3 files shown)


cjp256 (approved):

cjp256 (approved):
lgtm

igalic:
thanks for the review @cjp256!
since your review I've extended the code to only use the `growfs` service when dealing with the root device.
There's now a test, and documentation too.

The FreeBSD `growfs` service adds a swap partition if it finds none on the system (or is instructed *not* to do that).
However: Swap on FreeBSD is… quite important, IMO. Not just from a Virtual Memory management perspective, but also as dump device in case of kernel crashes.

TheRealFalcon (approved):
LGTM!

PR #3625, opened 2023-05-12
schema: read_cfg_paths call init.fetch to lookup /v/l/c/instance

blackboxsw:
## Proposed Commit Message
```
Fix cloud-init schema --system being unable to find merged
userdata stored at /var/lib/cloud/instance/cloud_config.txt.

commit e9d1d3a regressed cloud-init schema --system by dropping
a call to fetch existing cached datasource information before
calling Init.paths.get_ipath.

Init.paths.get_ipath only has visibility to merged cloud config in
/var/lib/cloud/<instance_id>/cloud-config.txt after fetching the
existing cached datasource which provides instance-id from metadata
in order to determine the unique instance-id which represents
the full path to the cloud-config.txt.

To support reuse of read_cfg_paths helper function, add an optional
parameter fetch_existing_datasource which indicates whether reading
the existing datasource is necessary for this helper function.

cloud-init schema --system calls read_cfg_paths providing
fetch_existing_datasource="trust" prior to calls to
paths.get_ipath().
```

## Test Steps

```
make deb
CLOUD_INIT_OS_IMAGE=lunar CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_23.1-98-g689cd92a7-1~bddeb_all.deb tox -e integration-tests tests/integration_tests/modules/test_cli.py
# Assert test_valid_userdata and test_invalid_userdata succeed
```
## Checklist:

### DIFF
--- a/cloudinit/cmd/devel/__init__.py
+++ b/cloudinit/cmd/devel/__init__.py
@@ -17,9 +17,17 @@ def addLogHandlerCLI(logger, log_level):
     return logger
 
 
-def read_cfg_paths() -> Paths:
-    """Return a Paths object based on the system configuration on disk."""
+def read_cfg_paths(fetch_existing_datasource: str = "") -> Paths:
+    """Return a Paths object based on the system configuration on disk.
+
+    :param fetch_existing_datasource: String one of check or trust. Whether to
+        load the pickled datasource before returning Paths. This is necessary
+        when using instance paths via Paths.get_ipath method which are only
+        known from the instance-id metadata in the detected datasource.
+    """
     init = Init(ds_deps=[])
+    if fetch_existing_datasource:
+        init.fetch(existing=fetch_existing_datasource)
     init.read_cfg()
     return init.paths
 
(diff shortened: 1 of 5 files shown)


cloudinit/cmd/devel/__init__.py
```diff
@@ -17,9 +18,17 @@ def addLogHandlerCLI(logger, log_level):
     return logger
 
 
-def read_cfg_paths() -> Paths:
-    """Return a Paths object based on the system configuration on disk."""
+def read_cfg_paths(fetch_existing_datasource: Optional[str] = "") -> Paths:
```
TheRealFalcon, line 21: Why accept None? If we remove the Optional, don't forget to remove the import too.

TheRealFalcon (approved):
One inline nit, but otherwise LGTM!

cloudinit/cmd/devel/__init__.py
blackboxsw: +1 no need at all to allow None, just default to empty string if unspecified any callers expecting to need get_ipath will have to provide a valid `fetch_existing_datasource=("trust"|"check")`

PR #3647, opened 2023-05-12
azure/errors: introduce reportable errors for imds

cjp256:
Always report failure to host, but report failure to fabric only outside of _check_if_nic_is_primary() which is expected to fail if nic is not primary.

Add two types of reportable errors for IMDS metadata:

- add ReportableErrorImdsUrlError() for url errors.

- add ReportableErrorImdsMetadataParsingException() for parsing errors.

Tweak ReportableError repr to be a bit friendlier.

### DIFF
--- a/cloudinit/sources/azure/errors.py
+++ b/cloudinit/sources/azure/errors.py
@@ -10,8 +10,11 @@ from datetime import datetime
 from io import StringIO
 from typing import Any, Dict, List, Optional
 
+import requests
+
 from cloudinit import version
 from cloudinit.sources.azure import identity
+from cloudinit.url_helper import UrlError
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


aciba90 (approved):
+1 thanks

cloudinit/sources/azure/errors.py
```diff
@@ -99,6 +107,37 @@ def __init__(self, duration: float, interface: Optional[str]) -> None:
...
+class ReportableErrorImdsUrlError(ReportableError):
+    def __init__(self, *, exception: UrlError, duration: float) -> None:
+        # ConnectTimeout sub-classes ConnectError so order is important.
+        if isinstance(exception.cause, requests.ConnectTimeout):
+            reason = "connection timeout querying IMDS"
+        elif isinstance(exception.cause, requests.ConnectionError):
+            reason = "connection error querying IMDS"
+        elif isinstance(exception.cause, requests.ReadTimeout):
+            reason = "read timeout querying IMDS"
+        elif exception.code:
```
anhvoms, line 119: Do you want to add other known errors such as connection reset, no route to host, network unreachable ?

cjp256: I didn't want to resort to string matching here, though we could.  The original exception is captured in supporting_data so it will be evident (those should all be a form of connection errors).

cjp256:
> +1 thanks

Thank you for the fast review and merge!  For future reference, we prefer that PRs that affect Azure get a +1 from another Azure reviewer (usually @anhvoms and/or myself).   We don't have an internal review before opening PRs here and I know Anh has saved my bacon more than a handful of times with details I wouldn't expect anyone outside of Azure to know :D

PR #3671, opened 2023-05-12
network-manager: Set higher autoconnect priority for nm keyfiles

ani-sinha:
cloud init generated keyfiles by network manager renderer for network
interfaces can sometimes conflict with existing keyfiles that are left as an
artifact of an upgrade process or are old user generated keyfiles. When two
such keyfiles are present, the existing keyfile can take precedence over the
cloud init generated keyfile making the later ineffective. Removing the old
keyfile blindly by cloud init would also not be correct since there would be
no way to enforce a different interface configuration if one needs it.

This change adds an autoconnect-priority value for cloud init generated keyfile
so that the cloud init configuration takes precedence over the existing old
keyfile configuration in the default case. The priority values range from 0
to 999. We set a value of 120 so that it would be high enough in the default
case and result in cloud init keyfile to take precedence but not too high so
that if the user generated keyfile needs to take precedence, the user can do
so by using a higher value than the one used by cloud init key file, between
the values 121 and 999.

BUG: https://github.com/canonical/cloud-init/issues/4121
RHBZ: 2196231

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -43,6 +43,7 @@ class NMConnection:
         self.config["connection"] = {
             "id": f"cloud-init {con_id}",
             "uuid": str(uuid.uuid5(CI_NM_UUID, con_id)),
+            "autoconnect-priority": "120",
         }
 
         # This is not actually used anywhere, but may be useful in future
--- a/tests/unittests/cmd/devel/test_net_convert.py
+++ b/tests/unittests/cmd/devel/test_net_convert.py
@@ -74,6 +74,7 @@ SAMPLE_NETWORK_MANAGER_CONTENT = """\
 [connection]
 id=cloud-init eth0
 uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
+autoconnect-priority=120
 type=ethernet
 interface-name=eth0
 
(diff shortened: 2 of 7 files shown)


ani-sinha:
@blackboxsw review please!

blackboxsw (commented):
@ani-sinha Thank you for the proposal, from the looks of [docs](https://developer-old.gnome.org/NetworkManager/stable/nm-settings-nmcli.html) and [man pages](https://manpages.ubuntu.com/manpages/impish/en/man5/nm-settings.5.html) this seems like a reasonable setting and provisions for possibly overriding that behavior make sense to me. Any environment with cloud-init enabled, should expect cloud-init's network to take priority over system config snippets on disk.

I have no visibility to RHBZ: 2196231. If you confirm that is the proper bug ID, then I think I'm good on this PR.

ani-sinha:
> @ani-sinha Thank you for the proposal, from the looks of [docs](https://developer-old.gnome.org/NetworkManager/stable/nm-settings-nmcli.html) and [man pages](https://manpages.ubuntu.com/manpages/impish/en/man5/nm-settings.5.html) this seems like a reasonable setting and provisions for possibly overriding that behavior make sense to me. Any environment with cloud-init enabled, should expect cloud-init's network to take priority over system config snippets on disk.
> 
> I have no visibility to RHBZ: 2196231. If you confirm that is the proper bug ID, then I think I'm good on this PR.

I have opened upstream ticket https://github.com/canonical/cloud-init/issues/4121 which has the description. I have also updated the summary above.

ani-sinha:
> > @ani-sinha Thank you for the proposal, from the looks of [docs](https://developer-old.gnome.org/NetworkManager/stable/nm-settings-nmcli.html) and [man pages](https://manpages.ubuntu.com/manpages/impish/en/man5/nm-settings.5.html) this seems like a reasonable setting and provisions for possibly overriding that behavior make sense to me. Any environment with cloud-init enabled, should expect cloud-init's network to take priority over system config snippets on disk.
> > I have no visibility to RHBZ: 2196231. If you confirm that is the proper bug ID, then I think I'm good on this PR.

and yes, that is the proper internal RH BUG ID.

blackboxsw (approved):
LGTM!

PR #93, opened 2019-12-06
fix unlocking method on FreeBSD

igalic:
on FreeBSD, lock_passwd is implemented as `pw usermod <user> -h -`

This does not lock the account. It prompts for a password change on the console during cloud-init run.

To lock an account, we have to execute: `pw lock <name>`.

-----
Fixes LP # 1854594

PR #4114, opened 2023-05-16
FreeBSD: Fix user account locking

igalic:
## Proposed Commit Message
```
FreeBSD: Fix user account locking

On Linux, we only disable login via Password authentication.
On FreeBSD, we were right out locking the account.
This was reported back when #93 was merged,
but back then I didn't understand that.

Newly created FreeBSD users can now login with with SSH keys.

LP: #1854594 (For real this time)
Fixes: #3507

```

## Additional Context

#93

## Checklist:

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -147,9 +147,9 @@ class Distro(cloudinit.distros.bsd.BSD):
 
     def lock_passwd(self, name):
         try:
-            subp.subp(["pw", "usermod", name, "-h", "-"])
+            subp.subp(["pw", "usermod", name, "-w", "no"])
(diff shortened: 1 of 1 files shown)


igalic:
cc @goneri

goneri (approved):

blackboxsw (approved):
Changset looks good to me. I don't find a lot of merit in us adding unittests here to validadate that we are passing the new params we think we are. Ideall we'd like to rely on integration tests on freebsd images with cloud-init included to assert success ... but @igalic is looking at trying to make that more a reality.

Man pages agree with this approach  https://man.freebsd.org/cgi/man.cgi?pw(8)

Issue #3250, opened 2023-05-11
Cloud init rsyslog module will not start on FreeBSD


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1798055](https://bugs.launchpad.net/cloud-init/+bug/1798055)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2018-10-16T09:30:41.283569+00:00
date_fix_committed = None
date_fix_released = None
id = 1798055
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1798055
milestone = None
owner = charliemaiors
owner_name = Carlo Maiorano
private = False
status = triaged
submitter = charliemaiors
submitter_name = Carlo Maiorano
tags = ['freebsd']
duplicates = []
</pre>
</details>

_Launchpad user **Carlo Maiorano(charliemaiors)** wrote on 2018-10-16T09:30:41.283569+00:00_

Cloud Provider: Openstack
Cloud init configuration is embedded in the attached archive
Cloud init platform: FreeBSD
FreeBSD image built using CLOUDWARE.

Cloud init port (net/cloud-init) on FreeBSD will not load rsyslog module because is not in configuration file.

ubuntu-server-builder:
_Launchpad user **Scott Moser(smoser)** wrote on 2018-10-16T13:55:25.133324+00:00_

@Carlo,
Can you verify that the module works on FreeBSD ?

If so we can quite easily add it to the stock upstream config file generated for FreeBSD.

ubuntu-server-builder:
_Launchpad user **Carlo Maiorano(charliemaiors)** wrote on 2018-10-17T12:08:37.230256+00:00_

@Scott, 
I've tried to test the single module but it didn't run because there are no rsyslog in configuration (The configuration is in the archive for the previous submission). I've also attached the dump from cloud-init analyze command.
_Launchpad attachments_: [dumps.tar.gz](https://api.launchpad.net/devel/cloud-init/+bug/1798055/+attachment/5202190/data)

Issue #4118, opened 2023-05-17
Unit tests: get_cloud(distro=FreeBSD) returns a bogus OS Family 


igalic:
```python
In [1]: from tests.unittests.util import get_cloud

In [2]: from unittest import mock

In [3]:         with mock.patch(
   ...:             "cloudinit.distros.networking.subp.subp", return_values=("", None)
   ...:         ):
   ...:             cloud = get_cloud(distro="freebsd", metadata={})
   ...: 

In [4]: cloud.distro.name
Out[4]: 'freebsd'

In [5]: cloud.distro.osfamily
Out[5]: 'linux'

In [6]: 
```

TheRealFalcon:
@igalic want me to assign you to this one?

igalic:
not sure i can figure this out without handholding
I'm finding our test framework very confusing

blackboxsw:
I think the issue is  [cloudinicloudinit/distros/bsd.py:        `self.osfamily = platform.system().lower()`](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/bsd.py#L38).  The tests.unittests.get_cloud helper is leaking through to `cloudinit.util.platform`  on the underlying host system, but I think you want `def get_cloud` to either mock `platform.system()` 


might think about something like:
```diff

diff --git a/tests/unittests/util.py b/tests/unittests/util.py
index 8c6682e18..e8c5c90e5 100644
--- a/tests/unittests/util.py
+++ b/tests/unittests/util.py
@@ -20,8 +20,11 @@ def get_cloud(
     """
     paths = paths or helpers.Paths({})
     sys_cfg = sys_cfg or {}
+    import pdb; pdb.set_trace()
     cls = distros.fetch(distro) if distro else MockDistro
-    mydist = cls(distro, sys_cfg, paths)
+    # *BSD calls platform.system to determine osfamilies
+    with mock.patch('platform.system', return_value=distro.lower()):
+        mydist = cls(distro, sys_cfg, paths)
     if mocked_distro:
         mydist = mock.MagicMock(wraps=mydist)
     myds = DataSourceTesting(sys_cfg, mydist, paths)
```

TheRealFalcon:
Fixed in #4119

PR #4119, opened 2023-05-17
cc_rsyslog: Refactor for better multi-platform support

igalic:
## Proposed Commit Message

```
cc_rsyslog: Refactor for better multi-platform support

- remove Hash key redirection
- copy basic multi-distro config from cc_ntp
- add optional installation
- enable and start services
- on BSD: disable Base system syslog

Co-Authored-by: Chad Smith <chad.smith@canonical.com>
Co-Authored-by: James Falcon <james.falcon@canonical.com>

Fixes: #3250, #4118
LP: #1798055
```

## Additional Context
Fixes: #3250, #4118
LP: #1798055

## Checklist:

### DIFF
--- a/cloudinit/config/cc_rsyslog.py
+++ b/cloudinit/config/cc_rsyslog.py
@@ -1,8 +1,10 @@
 # Copyright (C) 2009-2010 Canonical Ltd.
 # Copyright (C) 2012 Hewlett-Packard Development Company, L.P.
+# Copyright (C) 2023 FreeBSD Foundation
 #
 # Author: Scott Moser <scott.moser@canonical.com>
 # Author: Juerg Haefliger <juerg.haefliger@hp.com>
+# Author: Mina Galić <FreeBSD@igalic.co>
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
--- a/cloudinit/config/cc_rsyslog.py
+++ b/cloudinit/config/cc_rsyslog.py
@@ -8,6 +8,7 @@
 
 """Rsyslog: Configure system logging via rsyslog"""
 
+import copy
 import os
 import re
 from textwrap import dedent
@@ -17,7 +18,7 @@ from cloudinit import subp, util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema, get_meta_doc
-from cloudinit.distros import ALL_DISTROS
+from cloudinit.distros import ALL_DISTROS, Distro
 from cloudinit.settings import PER_INSTANCE
 
 MODULE_DESCRIPTION = """\
@@ -67,12 +68,29 @@ meta: MetaSchema = {
 
 __doc__ = get_meta_doc(meta)
 
-default_config = {
+RSYSLOG_CONFIG = {
     "config_dir": "/etc/rsyslog.d",
     "config_filename": "20-cloud-config.conf",
     "service_reload_command": "auto",
     "remotes": {},
     "configs": {},
+}
+
+DISTRO_OVERRIDES = {
+    "freebsd": {
+        "config_dir": "/usr/local/etc/rsyslog.d",
+    },
+    # this one is only necessary because of
+    # https://github.com/canonical/cloud-init/issues/4118
+    "dragonfly": {
+        "config_dir": "/usr/local/etc/rsyslog.d",
+    },
+    "openbsd": {
+        "config_dir": "/usr/local/etc/rsyslog.d",
+    },
+    "netbsd": {
+        "config_dir": "/usr/pkg/etc/rsyslog.d",
+    },
 }
 
 LOG = logging.getLogger(__name__)
@@ -85,6 +103,20 @@ HOST_PORT_RE = re.compile(
 )
 
 
+def distro_default_rsyslog_config(distro: Distro):
+    """Construct a distro-specific rsyslog config dictionary by merging
+       distro specific changes into base config.
+
+    @param distro: String providing the distro class name.
+    @returns: Dict of distro configurations for ntp clients.
+    """
+    dcfg = DISTRO_OVERRIDES
+    cfg = copy.copy(RSYSLOG_CONFIG)
+    if distro.name in dcfg:
+        cfg = util.mergemanydict([cfg, dcfg[distro.name]], reverse=True)
+    return cfg
+
+
 def reload_syslog(distro, command="auto"):
     if command == "auto":
         service = distro.get_option("rsyslog_svcname", "rsyslog")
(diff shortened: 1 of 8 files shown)


cloudinit/config/cc_rsyslog.py
```diff
@@ -296,41 +351,102 @@ def remotes_to_rsyslog_cfg(remotes, header=None, footer=None):
     return "\n".join(lines) + "\n"
 
 
+def disable_and_stop_bsd_base_syslog(cloud: Cloud) -> None:
```
igalic, line 354: this function is a mess, but it *is* correct.

cloudinit/config/cc_rsyslog.py
```diff
@@ -90,38 +107,76 @@
...
+    @returns: Dict of distro configurations for ntp clients.
+    """
+    dcfg = DISTRO_OVERRIDES
+    cfg = copy.copy(RSYSLOG_CONFIG)
+    if distro.osfamily in dcfg:
+        cfg = util.mergemanydict([cfg, dcfg[distro.name]], reverse=True)
+    return cfg
+
+
+def install_rsyslog(install_func, packages=None, check_exe="rsyslog"):
```
igalic, line 124: stolen from cc_ntp

igalic (commented):
This PR, which was meant as a short distraction, turned into a massive yak shave.

here are some guide posts for reviewers.

TheRealFalcon (changes requested):
Nice change overall. I left some inline comments.

One major concern...the rsyslog module runs early enough in boot that, at least on systemd systems, it's considered to be before network is even online. I don't think we should be reaching out to the internet to do package installs in this time frame. I just tried running cloud-init locally where rsyslog isn't installed, and it appears that apt is hanging trying to install it. I think at the very least we need to make installation a configuration option that defaults to false.

igalic:
> One major concern...the rsyslog module runs early enough in boot that, at least on systemd systems, it's considered to be before network is even online. I don't think we should be reaching out to the internet to do package installs in this time frame. I just tried running cloud-init locally where rsyslog isn't installed, and it appears that apt is hanging trying to install it. I think at the very least we need to make installation a configuration option that defaults to false.

On FreeBSD, cc_rsyslog runs well after we have a network connection, so package installation is not a problem.

cloudinit/config/cc_rsyslog.py
igalic: function now less messy

igalic: (and still correct)

TheRealFalcon (changes requested):
> On FreeBSD, cc_rsyslog runs well after we have a network connection, so package installation is not a problem.

That's fine, but we can't start breaking things on Linux. As-is, the install will run regardless of the OS we're on. Without this change, if rsyslog wasn't already installed, you already had a problem in that you were trying to run a module for a service that wasn't installed, but the change here can compound the problem and can prevent cloud-init from completing. Additionally, if somebody wants to start running rsyslog on a distro that doesn't have it pre-installed, hitting these install issues is not a great experience.

I think we need to add one more config option, something like `install_syslog` that defaults to False along with a documentation note in the module mentioning that due to rsyslog running so early in boot, installation is not supported for every distro.

cloudinit/config/cc_rsyslog.py
```diff
@@ -296,41 +365,79 @@ def remotes_to_rsyslog_cfg(remotes, header=None, footer=None):
...
+    need to tell rc to reload its config
+    """
+    try:
+        cloud.distro.manage_service("enabled", "syslogd")
+    except subp.ProcessExecutionError:
+        return
+    cloud.distro.manage_service("disable", "syslogd")
+    cloud.distro.reload_init()
+
+    try:
```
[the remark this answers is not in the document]
igalic: *\*points at Erlang's and Common Lisp's error handling\**

can we have some of this?

TheRealFalcon (changes requested):
Just a minor doc change request, but other LGTM!

Doc standard these days is to put the default for a config value in the schema rather than in the module description. Also, I think it makes more sense for `install_rsyslog` to sit above the other new options as they both depend on that one. I put these suggestions in the following patch:
```diff
diff --git a/cloudinit/config/cc_rsyslog.py b/cloudinit/config/cc_rsyslog.py
index a61ae943c..5ec99f0b0 100644
--- a/cloudinit/config/cc_rsyslog.py
+++ b/cloudinit/config/cc_rsyslog.py
@@ -30,10 +30,9 @@ This module configures remote system logging using rsyslog.
 Configuration for remote servers can be specified in ``configs``, but for
 convenience it can be specified as key value pairs in ``remotes``.
 
-This module can install rsyslog ``packages`` (Default: ``["rsyslog"]``) if
-``check_exe`` (Default: ``"rsyslogd"``) cannot be found.
-On many systems, this may not work, because networking isn't up yet.
-That's why ``install_rsyslog`` defaults to ``False``.
+This module can install rsyslog if not already present on the system using the
+``install_rsyslog``, ``packages``, and ``check_exe`` options. Installation
+may not work on systems where this module runs before networking is up.
 
 .. note::
     On BSD cloud-init will attempt to disable and stop the base system syslogd.
diff --git a/cloudinit/config/schemas/schema-cloud-config-v1.json b/cloudinit/config/schemas/schema-cloud-config-v1.json
index c6a08b237..6466e541c 100644
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -2373,9 +2373,14 @@
                 }
               ]
             },
+            "install_rsyslog": {
+              "default": false,
+              "description": "Install rsyslog. Default: ``false``",
+              "type": "boolean"
+            },
             "check_exe": {
               "type": "string",
-              "description": "The executable name for the rsyslog daemon.\nFor example, ``rsyslogd``, or ``/opt/sbin/rsyslogd`` if the rsyslog binary is in an unusual path."
+              "description": "The executable name for the rsyslog daemon.\nFor example, ``rsyslogd``, or ``/opt/sbin/rsyslogd`` if the rsyslog binary is in an unusual path. This is only used if ``install_rsyslog`` is ``true``. Default: ``rsyslogd``"
             },
             "packages": {
               "type": "array",
@@ -2383,12 +2388,7 @@
                 "type": "string"
               },
               "uniqueItems": true,
-              "description": "List of packages needed to be installed for rsyslog."
-            },
-            "install_rsyslog": {
-              "default": false,
-              "description": "Install rsyslog. Default: ``false``",
-              "type": "boolean"
+              "description": "List of packages needed to be installed for rsyslog. This is only used if ``install_rsyslog`` is ``true``. Default: ``[rsyslog]``"
             }
           }
         }
```
```

TheRealFalcon (approved):
LGTM!

PR #4122, opened 2023-05-18
cc_ntp: Sync up with current FreeBSD ntp.conf

igalic:
## Proposed Commit Message

```
cc_ntp: Sync up with current FreeBSD ntp.conf

ref: https://reviews.freebsd.org/D39954

```

## Checklist:

### DIFF
--- a/templates/ntp.conf.freebsd.tmpl
+++ b/templates/ntp.conf.freebsd.tmpl
@@ -22,12 +22,13 @@
 tos minclock 3 maxclock 6
 
 #
-# The following pool statement will give you a random set of NTP servers
-# geographically close to you.  A single pool statement adds multiple
-# servers from the pool, according to the tos minclock/maxclock targets.
+# The following pool statements will give you a random set of IPv4 and IPv6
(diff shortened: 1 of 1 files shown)


templates/ntp.conf.freebsd.tmpl
```diff
@@ -40,6 +41,15 @@ pool {{pool}} iburst
 peer {{peer}}
 {% endfor %}
 
+#
+# If you want to pick yourself which country's public NTP server
```
dermotbradley, line 45: I don't understand why you'd want to add this comment in the template file (and therefore into the created ntp.conf file) - the "pools" cc_ntp configuration setting is used to specify pools to use other than the defaults, if someone wants to use country-specific ntp.org pool thenthey can specify it via the "pools" setting.

igalic: I thought it might be good to have this file look the same as upstream.

but I think there's a good reason why i left this particular paragraph out initially

TheRealFalcon (approved):
LGTM!

Issue #4126, opened 2023-05-18
 Cloud-init generated IPv6 NM config not being correct when config static IPv6 address


ani-sinha:
Description of problem:
Cloud-init generated IPv6 NM keyfile is not correct when config static IPv6 address in customization file.

Version-Release number of selected component (if applicable):
rhel-9.3
cloud-init-23.1.1

How reproducible:
100%

Steps to Reproduce:
1. Create VM_template with cloud-init installed
2. Config static IP and static IPv6 in customization file
3. Clone VM_new with above customization file
4. Boot the VM_new and do checks

Actual results:
The method=auto for IPv6

```
# cat /etc/NetworkManager/system-connections/cloud-init-ens33.nmconnection
# Generated by cloud-init. Changes will be lost.

[connection]
id=cloud-init ens33
uuid=d2c9f616-cb0f-5b5a-9bb2-0f77bbf3e06a
type=ethernet

[user]
org.freedesktop.NetworkManager.origin=cloud-init

[ethernet]
mac-address=00:50:56:A7:98:00

[ipv4]
method=manual
may-fail=false
address1=192.168.10.23/24

[ipv6]
method=auto
may-fail=true
address1=2001::1/63
```

Expected results:
IT should be method=manual for IPv6, which is same as IPv4.

PR #4127, opened 2023-05-18
network_manager: add a method for ipv6 static IP configuration

ani-sinha:
The static IP configuration for IPv6 in the method_map is missing for network manager renderer. This is causing cloud-init to generate a keyfile with IPv6 method as "auto" instead of "manual". This fixes this issue.

fixes: https://github.com/canonical/cloud-init/issues/4126
RHBZ: 2196284


ani-sinha:
@blackboxsw review please :-)

PR #4129, opened 2023-05-18
BSD: simplify finding MBR partitions

igalic:
## Proposed Commit Message

```
BSD: simplify finding MBR partitions

by removing duplicate code.

Co-authored-by: Alberto Contreras <aciba90@gmail.com>
Co-authored-by: Chad Smith <chad.smith@canonical.com>

Fixes: #2168 
```

## Additional Context
Fixes: #2168 

## Checklist:

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -305,14 +305,7 @@ def device_part_info(devpath):
         # FreeBSD doesn't know of sysfs so just get everything we need from
         # the device, like /dev/vtbd0p2.
         fpart = "/dev/" + util.find_freebsd_part(devpath)
-        m = re.search("^(/dev/.+)p([0-9])$", fpart)
-        if m:
-            return m.group(1), m.group(2)
-
-        # alternatively, we could be dealing with MBR slices,
-        # or we're on DragonFly
-        fslice = "/dev/" + util.find_dragonflybsd_part(devpath)
-        m = re.search("^(/dev/.+)s([0-9])$", fslice)
+        m = re.search("^(/dev/.+)sp([0-9]+[a-z]*)$", fpart)
         if m:
             return m.group(1), m.group(2)
 
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -305,9 +305,12 @@ def device_part_info(devpath):
         # FreeBSD doesn't know of sysfs so just get everything we need from
         # the device, like /dev/vtbd0p2.
         fpart = "/dev/" + util.find_freebsd_part(devpath)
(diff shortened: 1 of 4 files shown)


tests/unittests/distros/test_dragonflybsd.py
```diff
@@ -1,12 +1,11 @@
...
-
+# This file is part of cloud-init. See LICENSE file for license information.
 
 import cloudinit.util
 from tests.unittests.helpers import mock
 
 
 def test_find_dragonflybsd_part():
-    assert cloudinit.util.find_dragonflybsd_part("/dev/vbd0s3") == "vbd0s3"
+    assert cloudinit.util.find_freebsd_part("/dev/vbd0s3") == "vbd0s3"
```
blackboxsw, line 8: Do we expect to allow trailing a-z beyond 3? If so can we @pytest.mark.parmetrize this test and have the different cases we expect to support to give more context?

cloudinit/config/cc_growpart.py
```diff
@@ -305,14 +305,7 @@ def device_part_info(devpath):
...
         fpart = "/dev/" + util.find_freebsd_part(devpath)
-        m = re.search("^(/dev/.+)p([0-9])$", fpart)
-        if m:
-            return m.group(1), m.group(2)
-
-        # alternatively, we could be dealing with MBR slices,
-        # or we're on DragonFly
-        fslice = "/dev/" + util.find_dragonflybsd_part(devpath)
-        m = re.search("^(/dev/.+)s([0-9])$", fslice)
+        m = re.search("^(/dev/.+)[sp]([0-9]+[a-z]*)$", fpart)
```
blackboxsw, line 308: @igalic Let's leave some comment around the slices vs partitions as future-me won't recall why we have the 's' match.
`# We could be dealing with partition devices 'p' or MBR slices 's'.

Also we've added [a-z] supplements here to the match. This now permits /dev/<anything>s9zzzzzzzzzasdfzzz. What device types require this supplemental slicing id suffix? Can we add a comment  or unit test that exercises the expected /dev/ examples which meet these new matches?

Also since we are touching this regex might it be worth naming the symbolic group matches so we don't have discern below m.group(1) m.group(2)?
```suggestion
        m = re.search("^(?P<dev>/dev/.+)[sp](?P<part_slice>[0-9]+[a-z]*)$", fpart)
        if m:
            return m["dev"], m['part_slice"]
```

blackboxsw (commented):
Looks pretty good, but I'm not sure what cases we are trying to solve with the additional a-z match allowance. Nothing indicates in comments, tests or commit message why that extension is needed.

blackboxsw (approved):
LGTM

PR #4130, opened 2023-05-19
Move DigitalOcean before ConfigDrive datasource in default list

dermotbradley:
## Proposed Commit Message

```
Move DigitalOcean before ConfigDrive datasource in default list

DigitalOcean originally used the ConfigDrive data source before they
wrote and submitted their own DigitalOcean datasource. However it
appears that DO instances are still provided with a ConfigDrive ISO
upon boot - this means that unless the datasource_list is tailored
to specify "DigitalOcean" before "ConfigDrive" or to not mention
"ConfigDrive" at all then ConfigDrive will incorrectly be used.

If no datasource_list is specified in /etc/cloud.cfg or
/etc/cloud.cfg.d/* then the default datasource list is used which
defines ConfigDrive before DigitalOcean - this PR changes the order
of those 2 datasources in the default list.

Note: from searching online it seems that various documents on
DigitalOcean's website, as well as 3rd party article, still only mention
using the ConfigDrive datasource - this makes no sense as the
DigitalOcean datasource has existed since 2014!
```

## Checklist:

 - [ x ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)

### DIFF
--- a/cloudinit/settings.py
+++ b/cloudinit/settings.py
@@ -22,10 +22,10 @@ RUN_CLOUD_CONFIG = "/run/cloud-init/cloud.cfg"
 CFG_BUILTIN = {
     "datasource_list": [
         "NoCloud",
+        "DigitalOcean",
         "ConfigDrive",
         "LXD",
         "OpenNebula",
-        "DigitalOcean",
         "Azure",
         "AltCloud",
         "OVF",
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -124,8 +124,8 @@ DS_MAYBE=2
 DI_DSNAME=""
 # this has to match the builtin list in cloud-init, it is what will
 # be searched if there is no setting found in config.
-DI_DSLIST_DEFAULT="MAAS ConfigDrive NoCloud AltCloud Azure Bigstep \
-CloudSigma CloudStack DigitalOcean Vultr AliYun Ec2 GCE OpenNebula OpenStack \
+DI_DSLIST_DEFAULT="MAAS DigitalOcean ConfigDrive NoCloud AltCloud Azure \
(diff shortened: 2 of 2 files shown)


aciba90 (approved):
+1, thanks. Do we know any contributor from DigitalOcean to double-check this change?

dermotbradley:
> +1, thanks. Do we know any contributor from DigitalOcean to double-check this change?

There are 2 DO email addresses mentioned at the top of DataSourceDigitalOcean.py

aciba90:
Sorry for the ping @andrewsomething. Do you know someone in DigitalOcean to verify this change set is correct?

aciba90:
Thanks, @sutton-digitalocean, for the context.

Then, I believe it is not worth reordering `DatasourceDigitalOcean` as DigitialOcean overrides the datasource list in their images, and it is effectively not used or used as a fallback in the case there is no attached drive at boot time.

> Rather than deal with upstreaming and handling API versioning we have elected to use ConfigDrive since that supports all of the typical workloads for our services.

A wonder if it makes sense to remove `DatasourceDigitalOcean` entirely, as it is not supported and probably not used anywhere.

dermotbradley:
> Rather than deal with upstreaming and handling API versioning we have elected to use ConfigDrive since that supports all of the typical workloads for our services.

So would you suggest that we remove the DigitalOcean datasource completely from cloud-init as you are neither using or maintaining it?

> I wouldn't recommend merging this change, but we do override the DataSource list as part of the image build process so it _should't_ break anything (see below).

That makes sense for DO's own images but the current situation is confusing for someone making their own cloud images for use on DigitalOcean as to which DataSource to use.

aciba90:
@sutton-digitalocean, do you have any insights / recommendations about the previous comments?

Issue #4131, opened 2023-05-19
unable to enable sysconfig renderer by changing the default priorities in cloud-init config


ani-sinha:
The issue is on RHEL8.7 and RHEL 9.2, can't run sysconfig when changing the priority of network renderers to: 

```
   network:
      renderers: ['sysconfig', 'netplan', 'network-manager', 'networkd', 'eni']
```

Version-Release number of selected component (if applicable):
cloud-init-23.1.1-3.el9

How reproducible:
100%

Steps to Reproduce:
1.Changing the priority of network renderers in /etc/cloud/cloud.cfg, for example, 
 'sysconfig' is the first one.
```
   network:
      renderers: ['sysconfig', 'netplan', 'network-manager', 'networkd', 'eni']
```
2.rm -rf /var/lib/cloud /var/log/cloud-init*  and reboot
3.It's network-manager selected not sysconfig

We can see it in /var/log/cloud-init.log
```
----------
2023-05-05 13:20:58,557 - __init__.py[DEBUG]: Selected renderer 'network-manager' from priority list: ['sysconfig', 'netplan', 'network-manager', 'networkd', 'eni']
----------
```

Expected results:
sysconfig is selected when its priority is higher than network-manager.

This is because RHEL 8 and RHEL 9 as far as I can see, does not ship with ifup and ifdown scripts.

TheRealFalcon:
What OS and OS version?

The cloud-init sysconfig renderer will only run on a limited number of distros. In addition, the renderer requires:
1. `ifup` and `ifdown` exist in either `/sbin` or `/usr/sbin`
2. Either `/etc/sysconfig/network-scripts/network-functions` or `/etc/sysconfig/config` exists on the filesystem

If these aren't true, cloud-init won't use the sysoncifg renderer regardless of ordering. Are these cases true for you?

ani-sinha:
> What OS and OS version?

RHEL 9.2 or CenOS 9 stream.

> 
> The cloud-init sysconfig renderer will only run on a limited number of distros. In addition, the renderer requires:
> 
> 1. `ifup` and `ifdown` exist in either `/sbin` or `/usr/sbin`
> 2. Either `/etc/sysconfig/network-scripts/network-functions` or `/etc/sysconfig/config` exists on the filesystem
> 
> If these aren't true, cloud-init won't use the sysoncifg renderer regardless of ordering. Are these cases true for you?

[...]

PR #4132, opened 2023-05-19
net/sysconfig: enable sysconfig renderer if network manager has ifcfg-rh plugin

ani-sinha:
Some distributions like RHEL does not have ifup and ifdown scripts that traditionally handled ifcfg-eth* files. Instead RHEL uses network manager with ifcfg-rh plugin to handle ifcfg scripts. Therefore, the sysconfig should check for the existence of ifcfg-rh plugin in addition to checking for the existence of ifup and ifdown scripts in order to determine if it can handle ifcfg files. If either the plugin or ifup/ifdown scripts are present, sysconfig renderer can be enabled.

fixes: https://github.com/canonical/cloud-init/issues/4131
RHBZ: 2194050

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -1,6 +1,7 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 import copy
+import glob
 import io
 import os
 import re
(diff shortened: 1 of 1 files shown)


ani-sinha:
@blackboxsw review please :-)

holmanb (approved):
LGTM

PR #4134, opened 2023-05-19
ubuntu/focal:cherry pick netplan99-cannot-use-default.patch

TheRealFalcon:
To stay consistent with #4135


blackboxsw (approved):
+1 as with #4135

PR #4136, opened 2023-05-19
docs: Add feedback button to docs

TheRealFalcon:
## Proposed Commit Message

```
docs: Add feedback button to docs

Adds a "Give feedback" button to the docs that links to creating a new
documentation issue when clicked
```

## Additional Context
Docs page should now look like this: image

When clicked, we get a link to [create a new issue](https://github.com/canonical/cloud-init/issues/new?assignees=&labels=documentation%2C+new&projects=&template=documentation.md&title=%5Bdocs%5D%3A+&body=#%20Documentation%0A%0A%0A%0A%0A---%0A*Reported+from%3A+file:///home/james/cloud-init/doc/rtd_html/index.html)

The css and js code was boldly taken from the Ubuntu Pro documentation.

### DIFF
--- a/doc/rtd/static/js/github_issue_links.js
+++ b/doc/rtd/static/js/github_issue_links.js
@@ -6,7 +6,7 @@ window.onload = function () {
   link.href = (
     "https://github.com/canonical/cloud-init/issues/new?"
     + "assignees=&labels=documentation%2C+new&projects=&template=documentation.md&title=%5Bdocs%5D%3A+"
-    + "&body=# Documentation"
+    + "&body=%23 Documentation request"
     + "%0A%0A%0A%0A%0A"
     + "---"
     + "%0A"
--- a/doc/rtd/static/js/github_issue_links.js
+++ b/doc/rtd/static/js/github_issue_links.js
@@ -1,0 +1,23 @@
+window.onload = function () {
+  const link = document.createElement("a");
+  link.classList.add("muted-link");
+  link.classList.add("github-issue-link");
+  link.text = "Give feedback";
+  link.href = (
+    "https://github.com/canonical/cloud-init/issues/new?"
+    + "assignees=&labels=documentation%2C+new&projects=&template=documentation.md&title=%5Bdocs%5D%3A+"
+    + "&body=# Documentation"
+    + "%0A%0A%0A%0A%0A"
+    + "---"
+    + "%0A"
+    + `*Reported+from%3A+${location.href}*`
+  );
+  link.target = "_blank";
+
+  const div = document.createElement("div");
+  div.classList.add("github-issue-link-container");
+  div.append(link)
+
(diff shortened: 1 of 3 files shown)


doc/rtd/static/js/github_issue_links.js
```diff
@@ -0,0 +1,23 @@
...
+  link.classList.add("github-issue-link");
+  link.text = "Give feedback";
+  link.href = (
+    "https://github.com/canonical/cloud-init/issues/new?"
+    + "assignees=&labels=documentation%2C+new&projects=&template=documentation.md&title=%5Bdocs%5D%3A+"
+    + "&body=# Documentation"
+    + "%0A%0A%0A%0A%0A"
+    + "---"
+    + "%0A"
+    + `*Reported+from%3A+${location.href}*`
```
blackboxsw, line 13: I think you may need to urlencode any whitespace in the querystrings provided here James. Also our `.github/ISSUE_TEMPLATE/documentation.md` template seems to override any body you provide in querystring, so we may instead want to just provide the trailing route filename in the title querystring param if we can't sort that template behavior.

blackboxsw: Turns out you only need to urlencode the leading `# Documentation` as ``%23 Documentation`

blackboxsw:
```diff
diff --git a/doc/rtd/static/js/github_issue_links.js b/doc/rtd/static/js/github_issue_links.js
index 6741688cb..4aa54651b 100644
--- a/doc/rtd/static/js/github_issue_links.js
+++ b/doc/rtd/static/js/github_issue_links.js
@@ -6,7 +6,7 @@ window.onload = function () {
   link.href = (
     "https://github.com/canonical/cloud-init/issues/new?"
     + "assignees=&labels=documentation%2C+new&projects=&template=documentation.md&title=%5Bdocs%5D%3A+"
-    + "&body=# Documentation"
+    + "&body=%23 Documentation"
     + "%0A%0A%0A%0A%0A"
     + "---"
     + "%0A"
```
The above works for me. This behavior looks good, but let's let @s-makin review this as she may have other ideas to improve this

blackboxsw (approved):
would want on Sally on this

doc/rtd/static/js/github_issue_links.js
s-makin: It looks like it's set to default to the standard documentation issue

image

If this is what you intended then it all works fine and I'd be happy to sign off. The Pro Client docs button includes the page the user redirected from which negates them having to manually provide us the URL they just came from (which could provide an annoying experience for them if they didn't think to copy it, having to go back and then clicking the link again). If we included that functionality, what would the side effect be on the default docs issue template?

s-makin (approved):

PR #4140, opened 2023-05-22
ubuntu/devel: Remove pep8 dependency

TheRealFalcon:
It is no longer used and no longer exists in mantic.


blackboxsw (approved):
LGTM! paint it yellow and ship it.

blackboxsw (commented):
We should probably reflect this change back to all supported series w/ ubuntu/bionic as well since pep8 deps no longer exist in any of those series.

PR #4144, opened 2023-05-24
Modify PR template so autoclose works

TheRealFalcon:
## Proposed Commit Message
```
Modify PR template so autoclose works
```


blackboxsw (approved):
+1 tested an works

PR #4148, opened 2023-05-24
tests: update test_webhook_reporting

TheRealFalcon:
## Proposed Commit Message

```
tests: update test_webhook_reporting

Since the config-refresh_rmc_and_interface module was removed in #2148,
we also need to update the reporting test.
```


blackboxsw (approved):
Thanks @TheRealFalcon

PR #4155, opened 2023-05-30
add local_hostname to cc_ntp for enhanced template logic

jacobsalmela:
## Proposed Commit Message

```
Adds a simple `local_hostname` variable to the NTP module.  In complex configurations, it can be helpful to key off of the local_hostname as detected by the OS.  I suspect most users will never need it, but in complex and custom templates, it is helpful to have in place, so I did not include any template changes.
```

## Test Steps

Run ntp module.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ntp.py
+++ b/cloudinit/config/cc_ntp.py
@@ -8,6 +8,7 @@
 
 import copy
 import os
+import socket
 from textwrap import dedent
 
 from cloudinit import log as logging
@@ -436,6 +437,7 @@ def generate_server_names(distro):
 def write_ntp_config_template(
     distro_name,
     service_name=None,
+    local_hostname=None,
     servers=None,
     pools=None,
     allow=None,
@@ -448,6 +450,7 @@ def write_ntp_config_template(
 
     @param distro_name: string.  The distro class name.
     @param service_name: string. The name of the NTP client service.
+    @param local_hostname: string. Derived automatically from the hostname.
     @param servers: A list of strings specifying ntp servers. Defaults to empty
     list.
     @param pools: A list of strings specifying ntp pools. Defaults to empty
@@ -465,6 +468,8 @@ def write_ntp_config_template(
     @raises: ValueError when path is None.
     @raises: ValueError when template_fn is None and template is None.
     """
+    if not local_hostname:
+        local_hostname = socket.gethostname()
     if not servers:
         servers = []
     if not pools:
@@ -498,6 +503,7 @@ def write_ntp_config_template(
         raise ValueError("Not template_fn or template provided")
 
     params = {
+        "local_hostname": local_hostname,
         "servers": servers,
         "pools": pools,
         "allow": allow,
@@ -599,6 +605,10 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         LOG.debug("Skipping module named %s, disabled by cfg", name)
         return
 
+    # Set the hostname to the node this is running on if not specified
+    ntp_cfg.setdefault("local_hostname", socket.gethostname())
+    local_hostname = ntp_cfg.get("local_hostname")
+
     # Select which client is going to be used and get the configuration
     ntp_client_config = select_ntp_client(
         ntp_cfg.get("ntp_client"), cloud.distro
@@ -625,6 +635,9 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             raise RuntimeError(msg)
 
     LOG.debug("service_name: %s", ntp_client_config.get("service_name"))
+    LOG.debug(
+        "local_hostname: %s", ntp_cfg.get("local_hostname", local_hostname)
+    )
     LOG.debug("servers: %s", ntp_cfg.get("servers", []))
     LOG.debug("pools: %s", ntp_cfg.get("pools", []))
     LOG.debug("allow: %s", ntp_cfg.get("allow", []))
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -1826,6 +1826,15 @@
           ],
           "additionalProperties": false,
           "properties": {
+            "local_hostname": {
+              "type": "string",
+              "items": {
+                "type": "string",
+                "format": "hostname"
+              },
+              "uniqueItems": true,
+              "description": "Local hostname to use for ntp. If not provided,\nthe system hostname will be used."
+            },
             "pools": {
               "type": "array",
               "items": {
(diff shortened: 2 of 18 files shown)


dermotbradley:
It is not clear exactly what the purpose of this new setting, is it to control which IP address(es) the NTP daemon listens on?

You haven't updated any of the NTP template configuration files to make use of this new setting.

Is this setting specific to only some of the supported NTP daemons? If so then perhaps a daemon-specific configuration subsection should be created for clarify? I am doing this in a PR I expect to submit shortly where I am adding some Chrony-specific settings and so have added a "chrony_settings:" subsection.

jacobsalmela:
We often need node X to have configuration A and node B to have configuration Y.  The template keys off of the local hostname to include/exclude it from peers.  It's a smaller use case, so I intentionally didn't update the templates.  I just need the capability bundled in. 

We use it with chrony but it could be used elsewhere.  As I mentioned though, it would not be used by too many people but it also shouldn't hurt to have it in, sans a small overhead.

I can update the templates if you think this could go in.  Or I can see what you come up with for your other PR.

jacobsalmela:
@dermotbradley I updated the templates.  The use case is small, but an example is that we do not want a node to peer with itself, so keying off the hostname is a simple check for this.  LMK your thought.

aciba90 (commented):
Thanks @Cray-HPE.

If I understand correctly, the main target of this PR is to make NTP templates aware of the instance hostname, to enable them building logic depending on the hostname.

As a use-case example, you are excluding the current instance from `peers` if given in the user-data for chrony confs.

I do not see the point of adding a new config key to specify the hostname, why not just get it from `cloud.get_hostname().hostname`?

> We often need node X to have configuration A and node B to have configuration Y. The template keys off of the local hostname to include/exclude it from peers. It's a smaller use case, so I intentionally didn't update the templates. I just need the capability bundled in.

Ultimately, I think this could be solved without modifying cloud-init, as you know in advance what config you want for each node, don't you?

If the only use-case we have is the one you are exposing, we could probably filter out self from peers in cc_ntp.handle and do not modify the static templates. In that way, all templates (static and given in the user-data) would receive a correct set of peers.

And/or just exposing the hostname to the templates. But then, the users have to build the logic within custom templates.

What do you think?

jacobsalmela:
@aciba90

aciba90:
Thanks, @jacobsalmela, for the context. 

It sounds like this could be solved by using the [instance-metadata](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#instance-metadata) in your global [user-data](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#cloud-config-data).

User-data example:

```yaml
## template: jinja
#cloud-config
ntp:
  config:
     template: |
         ## template:jinja
         # My NTP Client config
         {% raw %}
         {% for peer in peers -%}
         {% endraw %}
         {% raw %}
           {% if {% endraw %} "{{ v1.local_hostname }}" {% raw %} == peer %}{% endraw %}
         {% raw %}
           {# a host should not peer with itself #}
           {% else %}
         peer {{peer}}
           {% endif %}
         {% endfor %}
         {% endraw %}
  peers:
    - self
    - other
```

Note that [v1.local_hostname](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#v1-local-hostname) will contain exactly what you are looking for.

In my example, I escape the inner jinja template using the [escape](https://jinja.palletsprojects.com/en/2.11.x/templates/#escaping) functionality, not sure if there is a simpler way to do so.

One useful command to iterate quickly while testing this is https://cloudinit.readthedocs.io/en/latest/reference/cli.html#render . Use it within an instance, to test what would be the rendered cloud-config/user-data.

Please, let us know if this solution works for you, or if you need any support to test it.

jacobsalmela:
I had originally tried using `v1.local_hostname` as well as the `render` command but it was not working as I expected.  Specifically, I remember the hostname variable not being populated.  It was many years ago now, so I do not remember the exact errors, just that the logic I wrote failed since the that var was not working or available.  

I was also under the impression the NTP module only ran in userdata context, not Global.  Could you expand a bit on how I would implement that?  That does sound somewhat promising.

The `{% raw %}` /`{% raw %}` looks like it could work, though I do not completely understand why it is necessary rather than the template key just containing my existing jinja.  

We use image-based installs and it makes it difficult to change the template, so the PR is part selfish so we can use upstream cloud-init as-is instead of a fork.  We also have two custom implementations of a cloud-init datasource, which further complicates things.

jacobsalmela:
> To check what is available/populated now execute cloud-init query v1 from within an instance.

Image-based installs makes this difficult since the instance is not active until after the node boots.

> Per your previous comment, I assume you are defining the ntp config as a base configuration in static files under /etc/cloud/cloud.cfg.d.

No.  And perhaps this is the caveat.  We are mixing (for better or worse) local files baked into the image at `/etc/cloud/cloud.cfg.d` with other configs/settings in our cloud-init datasources.  

> The idea is, instead of expanding the cc_ntp module to expose the hostname to cc_ntp.config.template, use the global templating functionality to render [instance metadata variables](https://canonical-cloud-init.readthedocs-hosted.com/en/latest/explanation/instancedata.html#instance-metadata) in the cloud-config, in your case the hostname in your ntp cloud-config.

Yes, this makes sense.  I think we sort of dug ourselves into a hole we can't quite get out of in regards to this.

> Just out of curiosity, in what cloud are you running? And why those custom datasources are required?

- https://github.com/Cray-HPE/metal-basecamp during bootstrap
- https://github.com/Cray-HPE/hms-bss during runtime

An unfortunate aspect of this is that we need to transfer the cloud-init data from one service to another, and it can get munged along the way.

> Please, let us know if you need more support and if you have success with this approach.

It sounds like in addition to our own complexities, we may just be under utilizing cloud-init's features and my PR will not be adding much to the product beyond what we would use it for if merged.  As the repo maintainer though, it is your choice 😉.

Thank you kindly for your help @aciba90.  I may be opening another PR soon for the timezone module in regards to setting the hardware clock.  Is there anything I should know (considering I am underutilizing cloud-init functionality) before traveling down that road?

aciba90:
Okay, the argument still applies. Either as a 

- supplemental `/etc/cloud/cloud.cfg.d` config file
- vendor_data
- user_data

within either

- your bootstrap step (maybe here: https://github.com/Cray-HPE/metal-basecamp/blob/main/configs/data.json#L48 ??)
- runtime step

it is possible to define the ntp cloud-config to consume the instance-metada hostname.

And, in theory, you can control the hostnames of your nodes and generate the correct ntp config in one of your steps, not needing to use cloud-init's templating.

I would say that the key is to understand:

- https://canonical-cloud-init.readthedocs-hosted.com/en/latest/explanation/configuration.html
- Your workflow between those two datasources

For this issue and the timezone change, if you have any doubts / questions, please reach us in IRC / [#cloud-init](https://canonical-cloud-init.readthedocs-hosted.com/en/latest/index.html#project-and-community), many of the discussions happen there and the community can help you to fix this issue.

For the moment, I am going to close this PR. Happy to help :)

PR #4159, opened 2023-06-01
BSD: remove datasource_list from cloud.cfg template

igalic:
## Proposed Commit Message

```
BSD: remove datasource_list from cloud.cfg template

This limited list was put into place, because in the past, something
wasn't working right. Whatever it was, seems to be resolved. Removing it
means BSDs have the same chance on all clouds like other distros do.

People can always create images that customize the selection, but the
out of box experience should be that cloud-init works as advertised on
all supported distros and all available clouds.

```

## Additional Context
Found while testing my last change on Hetzner.
ds-identify identifies Hetzner, but cloud-init did not.
Because it wasn't in the list of options.

## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -54,11 +54,6 @@ preserve_hostname: false
 
 # If you use datasource_list array, keep array items in a single line.
 # If you use multi line array, ds-identify script won't read array items.
-{% if variant.endswith("bsd") %}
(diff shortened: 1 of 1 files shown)


igalic:
cc @goneri who added these lines I'm now removing.

goneri (approved):

aciba90 (approved):
LGTM. Thanks.

PR #4163, opened 2023-06-05
gce: improve ephemeral fallback NIC selection (CPC-2578)

aciba90:
## Proposed Commit Message

```
gce: improve ephemeral fallback NIC selection (CPC-2578)

While setting up the ephemeral network config for fetching the IMD at init-local timeframe on multi-NIC instances, fulfilling:

- All NICs without carrier flag or more than one NIC with carrier flag
- systemd's predictable interface names is enabled

net.find_fallback_nic could select a NIC that is not the primary NIC, leaving the instance without access to the network, as on GCE, only the primary NIC can talk to the IMDS.

At this point in time, there is yet not access to
{instance,vendor,user}-data. Thus, it is tricky to dynamically inject a breadcrum pointing to the primary NIC.

Two actions have been taken to fix this situation:

1. Substitute eth0 with ens4 as the default primary NIC candidate on Linux systems.
2. Try through the list of candidate NICs and use the first that can reach the IMDS.

Note that none of these changes do change cloud-init's behavior as:

(1) only applies to instances with more than one NIC. (2) will eventually find the correct NIC as opposed to leaving the instance without network (previous behavior).
```

## Additional Context

[SF: #00359767](https://bootstack.canonical.com/cases/00359767)
[CPC-2578](https://warthogs.atlassian.net/browse/CPC-2578)

## Test Steps

1. Create instance on GCE with two NICs.
2. Ideally, the secondary instance should have a `ensXpY` name, but I wasn't able to achieve this situation (I do not have access to the machine type facing this situation). To have a similar situation, change the [DataSourceGCE.py::DEFAULT_PRIMARY_NIC](https://github.com/canonical/cloud-init/compare/main...aciba90:gce-multinic?expand=1#diff-da4370ff1bd47f364de530e8738ac0906b2664145bbeb23f1344543dbc6909f3R29) to the secondary NIC name on your instance (ens5 in my case).
3. `cloud-init clean --logs --reboot`
4. Verify the instance is reachable.

Log example: https://pastebin.canonical.com/p/7W3td2VttZ/

## Checklist:

[CPC-2578]: https://warthogs.atlassian.net/browse/CPC-2578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -399,9 +399,7 @@ def is_disabled_cfg(cfg):
     return cfg.get("config") == "disabled"
 
 
-def find_candidate_nics(
-    default_primary_interface: Optional[str] = None,
-) -> List[str]:
+def find_candidate_nics() -> List[str]:
     """Get the list of network interfaces viable for networking.
 
     @return List of interfaces, sorted naturally.
@@ -411,7 +409,7 @@ def find_candidate_nics(
     elif util.is_NetBSD() or util.is_OpenBSD():
         return find_candidate_nics_on_netbsd_or_openbsd()
     else:
-        return find_candidate_nics_on_linux(default_primary_interface)
+        return find_candidate_nics_on_linux()
 
 
 def find_fallback_nic() -> Optional[str]:
@@ -471,9 +469,7 @@ def find_fallback_nic_on_freebsd() -> Optional[str]:
     return None
 
 
-def find_candidate_nics_on_linux(
-    default_primary_interface: Optional[str] = None,
-) -> List[str]:
+def find_candidate_nics_on_linux() -> List[str]:
     """Get the names of the candidate network devices on Linux.
 
     @return List of sorted interfaces.
@@ -535,14 +531,12 @@ def find_candidate_nics_on_linux(
     # 2. Remaining connected interfaces, naturally sorted.
     # 3. DEFAULT_PRIMARY_INTERFACE, if possibly connected.
     # 4. Remaining possibly connected interfaces, naturally sorted.
-    if not default_primary_interface:
-        default_primary_interface = DEFAULT_PRIMARY_INTERFACE
     sorted_interfaces = []
     for interfaces in [connected, possibly_connected]:
         interfaces = sorted(interfaces, key=natural_sort_key)
-        if default_primary_interface in interfaces:
-            interfaces.remove(default_primary_interface)
-            interfaces.insert(0, default_primary_interface)
+        if DEFAULT_PRIMARY_INTERFACE in interfaces:
+            interfaces.remove(DEFAULT_PRIMARY_INTERFACE)
+            interfaces.insert(0, DEFAULT_PRIMARY_INTERFACE)
         sorted_interfaces += interfaces
 
     return sorted_interfaces
@@ -670,7 +664,7 @@ def _get_current_rename_info(check_downable=True):
          }}
     """
     cur_info = {}
-    for name, mac, driver, device_id in get_interfaces():
+    for (name, mac, driver, device_id) in get_interfaces():
         cur_info[name] = {
             "downable": None,
             "device_id": device_id,
@@ -703,6 +697,7 @@ def _get_current_rename_info(check_downable=True):
 def _rename_interfaces(
     renames, strict_present=True, strict_busy=True, current_info=None
 ):
+
     if not len(renames):
         LOG.debug("no interfaces to rename")
         return
@@ -998,6 +993,7 @@ def get_interfaces_by_mac_on_linux() -> dict:
         # TODO: move this format to openstack
         ib_mac = get_ib_interface_hwaddr(name, True)
         if ib_mac:
+
             # If an Ethernet mac address happens to collide with a few bits in
             # an IB GUID, prefer the ethernet address.
             #
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -399,7 +399,9 @@ def is_disabled_cfg(cfg):
     return cfg.get("config") == "disabled"
 
 
-def find_candidate_nics() -> List[str]:
+def find_candidate_nics(
+    default_primary_interface: Optional[str] = None,
+) -> List[str]:
     """Get the list of network interfaces viable for networking.
 
     @return List of interfaces, sorted naturally.
@@ -409,7 +411,7 @@ def find_candidate_nics() -> List[str]:
     elif util.is_NetBSD() or util.is_OpenBSD():
         return find_candidate_nics_on_netbsd_or_openbsd()
     else:
-        return find_candidate_nics_on_linux()
+        return find_candidate_nics_on_linux(default_primary_interface)
 
 
 def find_fallback_nic() -> Optional[str]:
@@ -469,7 +471,9 @@ def find_fallback_nic_on_freebsd() -> Optional[str]:
     return None
 
 
-def find_candidate_nics_on_linux() -> List[str]:
+def find_candidate_nics_on_linux(
+    default_primary_interface: Optional[str] = None,
+) -> List[str]:
     """Get the names of the candidate network devices on Linux.
 
     @return List of sorted interfaces.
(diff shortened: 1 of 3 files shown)


cloudinit/net/__init__.py
```diff
@@ -531,12 +535,14 @@ def find_candidate_nics_on_linux() -> List[str]:
     # 2. Remaining connected interfaces, naturally sorted.
     # 3. DEFAULT_PRIMARY_INTERFACE, if possibly connected.
     # 4. Remaining possibly connected interfaces, naturally sorted.
+    if not default_primary_interface:
+        default_primary_interface = DEFAULT_PRIMARY_INTERFACE
```
blackboxsw, line 539: Is it worth a breadcrumb comment referencing # provided by DataSourceGCE?

cloudinit/sources/DataSourceGCE.py
```diff
@@ -88,13 +88,39 @@ def __init__(self, sys_cfg, distro, paths):
 
     def _get_data(self):
         url_params = self.get_url_params()
-        network_context = noop()
         if self.perform_dhcp_setup:
-            network_context = EphemeralDHCPv4(
-                self.distro,
-                self.fallback_interface,
+            candidate_nics = net.find_candidate_nics(
+                default_primary_interface=DEFAULT_PRIMARY_INTERFACE
```
blackboxsw, line 93: Let's debug log the candidate NICs we are going to try iterating through. Just because we may glean something interesting on bigger multi-NIC instances that sheds light on ordering/priority problems past the default ens4 NIC.

Also do we have a cloud-init log of the get_data operation working on the affected system. I'd also be curious if we don't can avoid passing DEFAULT_PRIMARY_INTERFACE on this instance type and `cloud-init clean --logs ---reboot` to see sort of delays we can expect on timeouts for invalid NICs prior to finding the valid ens4.

cloudinit/sources/DataSourceGCE.py
```diff
@@ -88,13 +88,39 @@ def __init__(self, sys_cfg, distro, paths):
...
+                network_context = EphemeralDHCPv4(
+                    self.distro,
+                    iface=candidate_nic,
+                )
+                try:
+                    with network_context:
+                        ret = util.log_time(
+                            LOG.debug,
+                            "Crawl of GCE metadata service",
+                            read_md,
```
blackboxsw, line 108: If `read_md` fails for some reason do we want to catch the exception and continue to try other candidates? If we hit some error here it looks like we are on the uncaught exception path and avoid iterating through any remaining candidates.

cloudinit/net/__init__.py
```diff
@@ -531,12 +535,14 @@ def find_candidate_nics_on_linux() -> List[str]:
...
     sorted_interfaces = []
     for interfaces in [connected, possibly_connected]:
         interfaces = sorted(interfaces, key=natural_sort_key)
-        if DEFAULT_PRIMARY_INTERFACE in interfaces:
-            interfaces.remove(DEFAULT_PRIMARY_INTERFACE)
-            interfaces.insert(0, DEFAULT_PRIMARY_INTERFACE)
+        if default_primary_interface in interfaces:
```
blackboxsw, line 543: What this means now is that if any GCE multi-nic instance type doesn't have an `ens4` NIC in `interfaces` but happens to have `eth0` somwwhere in the sorted list, we will no longer insert `eth0` at index 0  because we will be searching exclusively for `ens4`.  

Did we want to limit the `ens4` default to certain instance types for google? If not, are certain we aren't inadvertently delaying some other google intsance types adversely by leaving `eth0` in a multi nic scenario unsorted as first?

Alternatively, is it worth an `else:` here  to attempt to insert this module's DEFAULT_PRIMARY_INTERFACE at index 0 if the parameter `default_primary_interface` wasn't found in interfaces.

blackboxsw (commented):
This approach looks pretty good. Could or questions inline.

I'm not sure there is an integration test we could provide here that would be viable on a specific multi-NIC instance that we could run as a one-off to validate exercise of the re-ordering logic.

I'd like to also have us confirm that we aren't adversely affecting boot times because we are now shifting the DEFAULT_PRIMARY_INTERFACE on all GCE vms to ens4 instead of eth0. But other than a manual before/after comparison of vm launches w/ `cloud-init analyze show|blame` on multi-NIC instance types, I'm not sure what else we should do here.

cloudinit/net/__init__.py
aciba90: Added.

cloudinit/sources/DataSourceGCE.py
aciba90: Catched and added debug log with the exception.

cloudinit/net/__init__.py
aciba90: > Did we want to limit the ens4 default to certain instance types for google? If not, are certain we aren't inadvertently delaying some other google intsance types adversely by leaving eth0 in a multi nic scenario unsorted as first?

We cannot limit this to certain instance types because we have not yet the instance metadata and there is not enough DMI information to identify those machines. Additionally, they explicitly wanted to change the default globally (see attached SF ticket).

> Alternatively, is it worth an else: here to attempt to insert this module's DEFAULT_PRIMARY_INTERFACE at index 0 if the parameter default_primary_interface wasn't found in interfaces.

I went with this approach, as it is more backwards compatible.

cloudinit/sources/DataSourceGCE.py
```diff
@@ -88,13 +89,50 @@ def __init__(self, sys_cfg, distro, paths):
...
+                                    "url_params": url_params,
+                                },
+                            )
+                        except Exception:
+                            LOG.debug(
+                                "Error fetching IMD with candidate NIC %s",
+                                candidate_nic,
+                                exc_info=True,
+                            )
+                            continue
```
blackboxsw, line 124: Thanks for the supplemental handling.

Let's drop the exc_info=True here. We don't need/want to inject tracebacks in cloud-init logs for this as it'll be hard to read for minimal benefit.

Can we instead just append the exception str(e) here? 

```suggestion
                        except Exception as e:
                            LOG.debug(
                                "Error fetching IMD with candidate NIC %s: %s",
                                candidate_nic,
                                e,
                            )
                            continue
```

Also since we know the likelihood of seeing HTTPConnectionPool type errors for non-primary devices like `2023-06-05 11:30:52,610 - DataSourceGCE.py[DEBUG]: url instance/id raised exception HTTPConnectionPool(host='metadata.google.internal', port=80): Max retries exceeded with url: /computeMetadata/v1/instance/id (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f9169432760>: Failed to establish a new connection: [Errno 101] Network is unreachable'))` , should we `continue` on those exception types instead of logging a duplicated message?

aciba90: Trace-backs removed.

> Also since we know the likelihood of seeing HTTPConnectionPool ... should we continue on those exception types instead of logging a duplicated message?

Those exceptions are [caught, logged and not re-raised](https://github.com/canonical/cloud-init/blob/8615c240c2405165d9755b838bc2be5d82ad0c1d/cloudinit/sources/DataSourceGCE.py#L49-L51), thus there is no need to `continue`. We are going to cleanly pass to the next loop cycle if no other errors happen and the connection could have not been established in subsequent retries.

tests/unittests/sources/test_gce.py
```diff
@@ -412,6 +416,56 @@ def test_local_datasource_uses_ephemeral_dhcp(self, _m_fallback, m_dhcp):
...
+        ]
+        m_dhcp.return_value.__enter__.side_effect = (
+            None,
+            NoDHCPLeaseError,
+            None,
+            None,
+        )
+        m_read_md.side_effect = (
+            {"success": False},
+            Exception("error fetching IMD"),
```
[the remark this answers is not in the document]
aciba90: Added!

aciba90:
> Wondering if there are any implications for OS disk images or distros that use the old-style ethX interface naming together with multiple interfaces on GCE.

@dermotbradley, per https://github.com/canonical/cloud-init/pull/4163#discussion_r1223998611, a system is not expected to have both naming schemes, thus, in the old-style we maintain the previous behavior.

I have verified that

* Ubuntu b,f,k,l, Debian 10, Fedora use predictable naming scheme.
* n1-standard-1, n2-standard-2, n2d-standard-2, t2d-standard-1 instance types have `ens4` and `ens5`.
* Plus https://github.com/canonical/cloud-init/pull/4163#discussion_r1223998611

And I believe that covers the testing we wanted to do. @blackboxsw, let me know if I should do any supplemental test.

aciba90:
@blackboxsw, CPC has created an image including this change-set from https://launchpad.net/~aciba/+archive/ubuntu/sf-00359767 and test. The result was positive.

I have attached the logs and analyze output, here:
https://drive.google.com/drive/folders/1wbYOm0KZHbUkCIVbKhTOOv8c8rLisqhM?usp=sharing

blackboxsw (approved):
Excellent. Thanks @aciba90 for going the extra mile on integration test validation etc. LGTM!

PR #4167, opened 2023-06-06
Added support for Akamai Connected Cloud (formerly Linode)

Dorthu:
## Proposed Commit Message

```
Added support for Akamai Connected Cloud (formerly Linode)

Added a DataSource for the Akamai cloud platform, enabling metadata
and user data retrieval and instance provisioning via cloud-init.
```

## Test Steps

Our metadata service is not yet publicly available, and as such this is being opened as a draft.  We have tested this in
our lab environment, and will work with you to get you access to the service when it becomes available in beta in the
near future.

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceAkamai.py
+++ b/cloudinit/sources/DataSourceAkamai.py
@@ -336,12 +336,16 @@ class DataSourceAkamai(sources.DataSource):
                         local_instance_id,
                     )
                     break
+        else:
+            # even if we failed to reach the metadata service this loop, we
+            # still have the locally-available metadata (namely the instance id
+            # and cloud name), and by accepting just that we ensure that
+            # cloud-init won't run on our next boot
+            LOG.warning(
+                "Failed to contact metadata service, falling back to local "
+                "metadata only."
+            )
 
-        # even if we failed to reach the metadata service in the loop above, we
-        # still have the locally-available metadata (namely the instance id and
-        # cloud name), and by accepting just that we ensure that cloud-init
-        # won't run on our next boot; as such, regardless of the outcome of the
-        # loop above, we should return True here
         return True
 
     def check_instance_id(self, sys_cfg) -> bool:
--- a/cloudinit/sources/DataSourceAkamai.py
+++ b/cloudinit/sources/DataSourceAkamai.py
@@ -48,9 +48,9 @@ class MetadataAvailabilityResult(Enum):
     of metadata to it
     """
 
-    not_available = 0
-    available = 1
-    defer = 2
+    NOT_AVAILABLE = 0
+    AVAILABLE = 1
+    DEFER = 2
 
 
 class DataSourceAkamai(sources.DataSource):
@@ -96,7 +96,8 @@ class DataSourceAkamai(sources.DataSource):
 
     def _should_fetch_data(self) -> MetadataAvailabilityResult:
         """
-        Returns True if we should fetch metadata right now, otherwise false
+        Returns whether metadata should be retrieved at this stage, at the next
+        stage, or never, in the form of a MetadataAvailabilityResult.
         """
         if (
             not self.ds_cfg["allow_ipv4"] and not self.ds_cfg["allow_ipv6"]
@@ -106,7 +107,7 @@ class DataSourceAkamai(sources.DataSource):
         ):
             # if we're not allowed to fetch data, we shouldn't try
             LOG.info("Configuration prohibits fetching metadata.")
-            return MetadataAvailabilityResult.not_available
+            return MetadataAvailabilityResult.NOT_AVAILABLE
 
         if self.local_stage:
             return self._should_fetch_data_local()
@@ -115,13 +116,13 @@ class DataSourceAkamai(sources.DataSource):
 
     def _should_fetch_data_local(self) -> MetadataAvailabilityResult:
         """
-        Returns True if data should be fetched at the local stage (before
-        networking is available), or False otherwise
+        Returns whether metadata should be retrieved during the local stage, or
+        if it should wait for the init stage.
         """
         if not self.ds_cfg["allow_local_stage"]:
             # if this stage is explicitly disabled, don't attempt to fetch here
             LOG.info("Configuration prohibits local stage setup")
-            return MetadataAvailabilityResult.defer
+            return MetadataAvailabilityResult.DEFER
 
         if not self.ds_cfg["allow_dhcp"] and not self.ds_cfg["allow_ipv6"]:
             # without dhcp, we can't fetch during the local stage over IPv4.  If we're
@@ -129,21 +130,20 @@ class DataSourceAkamai(sources.DataSource):
             LOG.info(
                 "Configuration does not allow for ephemeral network setup."
             )
-            return MetadataAvailabilityResult.defer
+            return MetadataAvailabilityResult.DEFER
 
-        return MetadataAvailabilityResult.available
+        return MetadataAvailabilityResult.AVAILABLE
 
     def _should_fetch_data_network(self) -> MetadataAvailabilityResult:
         """
-        Returns True if data should be fetched during the init stage, or False
-        otherwise
+        Returns whether metadata should be fetched during the init stage.
         """
         if not self.ds_cfg["allow_init_stage"]:
             # if this stage is explicitly disabled, don't attempt to fetch here
             LOG.info("Configuration does not allow for init stage setup")
-            return MetadataAvailabilityResult.defer
+            return MetadataAvailabilityResult.DEFER
 
-        return MetadataAvailabilityResult.available
+        return MetadataAvailabilityResult.AVAILABLE
 
     def _get_network_context_managers(
         self,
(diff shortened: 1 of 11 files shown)


aciba90 (commented):
Welcome to cloud-init, @Dorthu!

A couple of pre-steps that we could sort out in advance are:

- [x] Sign the Canonical [contributor license agreement](https://ubuntu.com/legal/contributors).
- [x] Add your GitHub username (alphabetically) to the in-repository list that we use to track CLA signatures: [tools/.github-cla-signers](https://github.com/canonical/cloud-init/blob/main/tools/.github-cla-signers).

More info: https://cloudinit.readthedocs.io/en/latest/development/contributing.html#submitting-your-first-pull-request

Dorthu:
I believe that's already taken care of  - Akamai has already signed the CLA, and @jamesottinger and I have been added to the CLA signatures [here](https://github.com/canonical/cloud-init/blob/5267ebbf37ecb2a44c8823ed73cf6289bacc5488/tools/.github-cla-signers#L43) and [here](https://github.com/canonical/cloud-init/blob/5267ebbf37ecb2a44c8823ed73cf6289bacc5488/tools/.github-cla-signers#L67) through the company.  If there's something else we need to do on that front, we'd be happy to.

Dorthu:
Sorry for the long delay on this.  Our metadata service is now set up and ready for beta, so I'm opening this for review.

If you need access to the beta to review/test this change, please open a ticket through the Linode account that you need access from referencing this PR and post the Ticket ID here so I can verify it's you and grant you access.

igalic:
do you have a canonical link where to do that?
I'm not sure I've signed up correctly

Dorthu:
You can open a ticket in our system at https://cloud.linode.com/support/tickets

igalic:
https://cloud.linode.com/support/tickets/20767878

aciba90:
https://cloud.linode.com/support/tickets/20769719

aciba90:
I have been able to provision an instance with a successful cloud-init execution by following:

1. Install cloud-init from this PR in an instance.
2. Modify datasource_list in `/etc/cloud/cloud.cfg.d/99-linode.cfg` to: `datasource_list: [ NoCloud, Akamai, ConfigDrive, None ]`
3. Create snapshot.
4. Launch instance from the snapshot with:

    ```
    curl -H "Authorization: Bearer $API_TOKEN" -H "Content-type: application/json" https://api.linode.com/v4/linode/instances \
        -d '{"region":"us-iad","type":"g6-nanode1","image":"private/314","root_pass":"something secure","metadata":{"user_data":"I2Nsb3VkLWNvbmZpZwpydW5jbWQ6CiAgLSBlY2hvICdIZWxsbywgV29ybGQhJyA+IC92YXIvdG1wL2hlbGxvLXdvcmxkLnR4dA=="}}'
    ```

Cloud-init logs: [cloud-init-ud.tar.gz](https://github.com/canonical/cloud-init/files/12094153/cloud-init-ud.tar.gz)

cloudinit/sources/DataSourceAkamai.py
```diff
@@ -0,0 +1,372 @@
...
+                network_context_managers.append(
+                    (
+                        noop(),
+                        False,
+                    ),
+                )
+
+        return network_context_managers
+
+    def _make_requests(self, use_v6: bool = False) -> bool:
```
aciba90, line 227: Nit: This method not only make requests, it does change the state of the ds, by setting `self.metadata` and `self.userdata_raw`. Could think a better name?

cloudinit/sources/DataSourceAkamai.py
```diff
@@ -0,0 +1,372 @@
...
+        for manager, use_v6 in network_context_managers:
+            with manager as m:
+                done = self._make_requests(use_v6=use_v6)
+                if done:
+                    # fix up some field names
+                    self.metadata["instance-id"] = self.metadata.get(
+                        "id",
+                        local_instance_id,
+                    )
+                    break
```
aciba90, line 331: If the loop does not break, then the ds has not been update. Should we return false in that case?

```suggestion
                    break
        else:
            return False
```

doc/rtd/reference/datasources/akamai.rst
aciba90, line 1: We need to reference this file in `doc/rtd/reference/datasources.rst` to render it in the docs:

```diff
diff --git a/doc/rtd/reference/datasources.rst b/doc/rtd/reference/datasources.rst
index 04568001e..11e3ea621 100644
--- a/doc/rtd/reference/datasources.rst
+++ b/doc/rtd/reference/datasources.rst
@@ -38,6 +38,7 @@ The following is a list of documentation for each supported datasource:
 .. toctree::
    :titlesonly:
 
+   datasources/akamai.rst
    datasources/aliyun.rst
    datasources/altcloud.rst
    datasources/ec2.rst
```

tools/ds-identify
```diff
@@ -753,6 +753,12 @@ dscheck_CloudSigma() {
     return $DS_NOT_FOUND
 }
 
+dscheck_Akamai() {
```
aciba90, line 756: Could we please add unit-test coverage of this in `tests/unittests/test_ds_identify.py`

https://canonical-cloud-init.readthedocs-hosted.com/en/latest/development/datasource_creation.html#add-tests-for-ds-identify

aciba90 (changes requested):
Thanks, for the clean PR! Overall looks great. I left some inline comments.

cloudinit/sources/DataSourceAkamai.py
Dorthu: I tested this, but I found it didn't behave as expected.

If the metadata service was down when cloud-init ran and DataSourceAkamai returned `False`, cloud-init moved onto DataSourceNone and populated a cached instance-id of `iid-datasource-none`.  This was fine, except that on the next boot cloud-init would ask DataSourceAkamai for the instance id to check that the cache was valid, and our code would return the actual instance id (taken from the `system-serial-number` dmi field).  This, of course, did not match, invalidating the cache and causing cloud-init to reprovision the instance, generating new host keys and possibly applying any user data that was sent up with the instance.

My concern with this behavior is that a user may create an instance without user data during an outage of our metadata service, not notice that DataSourceNone was used, and then at some point in the future reboot and unexpectedly see their host keys change, which would be alarming.

This DataSource will set up metadata [here](https://github.com/canonical/cloud-init/pull/4167/files#diff-20689f9a33920f781eeb32d8802f9f931443049d9dff5c87151f3df00fb3ccadR304-R306) that is used if it's either configured not to run, or if the service is unavailable - if a user really needs cloud-init to re-run, cleaning and rebooting is a simple way to do that without the risk of it happening unexpectedly.

aciba90: Sounds good. Do you think it is worth to leave some kind of breadcrumb in the log for the users to know this happened? Something like:

```python
    else:
        LOG.warning("Unable to reach IMDS.")
```

Dorthu: Yep, that makes sense - I'll push that up alongside the changes to satisfy the linters 👍

Dorthu:
I'm not sure why the "Verify Contributor License Agreement" step failed - I'm pretty sure we're on there.  Is there something else i need to do?

aciba90:
> I'm not sure why the "Verify Contributor License Agreement" step failed - I'm pretty sure we're on there. Is there something else i need to do?

`tools/.github-cla-signers` is case sensitive, could you please `s/dorthu/Dorthu/` ?

tests/unittests/sources/helpers/test_akamai.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Tests that dmi-based configuration overrides parse as expected
+        """
+        read_dmi_data.return_value = dmi_config
+        result = get_dmi_config()
+
+        assert (
+            result == expected_result
+        ), f"Unexpected result parsing dmi config '{dmi_config}'!"
+        assert read_dmi_data.called_with("baseboard-serial-number")
```
aciba90, line 57:
```suggestion
        assert [
            mock.call("baseboard-serial-number") 
        ] == read_dmi_data.call_args_list
```

tests/unittests/sources/helpers/test_akamai.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Tests that is_on_akamai correctly detects if we are on Akama's platform
+        """
+        read_dmi_data.return_value = dmi_data
+        result = is_on_akamai()
+
+        assert (
+            result == expected_result
+        ), f"Unexpected result checking if '{dmi_data}' is Akamai"
+        assert read_dmi_data.called_with("bios-vendor")
```
aciba90, line 82:
```suggestion
        assert [
            mock.call("system-manufacturer")
        ] == read_dmi_data.call_args_list
```

aciba90 (changes requested):
Thanks for addressing the previous comments. I left a couple of comments fixing the two unit tests, with that in place I think we are good to go!

aciba90 (approved):
Thanks for adding support for Akamai and welcome to cloud-init!

Issue #4180, opened 2023-06-12
/run is not ephemeral on *BSD


igalic:
# Bug report

cloud-init sets up a `Paths.run_dir` which on Linux is the ephemeral `/run` directory.
On BSD, the corresponding directory would be `/var/run`, but we don't change that path. `/run` is non-standard. (See [`hier(7)`](https://man.freebsd.org/cgi/man.cgi?query=hier&apropos=0&sektion=7&manpath=FreeBSD+14.0-CURRENT&arch=default&format=html))

That means that cloud-init potentially behaves very differently on BSD

## Steps to reproduce the problem

1. boot a cloud-init enabled BSD system.
2. disable cloud-init.
3. `/run/cloud-init` still exists

## Environment details
- Cloud-init version: all
- Operating System Distribution: *BSD
- Cloud provider, platform or installer type: all

blackboxsw:
Cloud-init will still write each file on reboot for BSD like other distributions.

But, if cloud-init is disabled via `/etc/cloud/cloud-init.disabled` (or by kernel commandline) or can no longer detect any datasource, it's likely *BSD will have stale files around in /run/cloud-init across reboot that Linux distributions would have magically cleared.

loth:
This would fix #4231 if implemented

igalic:
can you try by replacing /run with a link to /var/run ? that one should be ephemeral, in that it's cleared before cloudinitlocal

PR #4182, opened 2023-06-13
BSD: add dsidentify to early startup scripts

igalic:
## Proposed Commit Message

```
in #4159, we removed the artifically restricted datasource_list for BSD.
This can now lead in some cases to very long boot times.
In this patch we add an RC script for ds-identify, similarly to how it's
run in systemd's generator stage.
The script is added in such a way that it will run before cloudinitlocal,
but can easily be removed by people building custom images.

Additionally, the rc scripts are now templated. This makes it now easier
for ports / pkgsrc users to move cloud-init package from the standard
`$LOCALBASE` to another location.

```

## Additional Context

Partially discussed in #4180 

## Test Steps
if this patch reduces start-up time, we can count it as successful 

## Checklist:

### DIFF
--- a/sysvinit/freebsd/dsidentify
+++ b/sysvinit/freebsd/dsidentify
@@ -1,28 +1,28 @@
 #!/bin/sh
 
-# PROVIDE: cloudinit
-# REQUIRE: FILESYSTEMS NETWORKING cloudinitlocal ldconfig devd
-# BEFORE:  LOGIN cloudconfig cloudfinal
+# PROVIDE: dsidentify
+# REQUIRE: mountcritlocal
+# BEFORE:  cloudinitlocal
 
 . /etc/rc.subr
 
 PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 
-name="cloudinit"
-command="/usr/local/bin/cloud-init"
-start_cmd="cloudinit_start"
+name="dsidentify"
+command="/usr/local/lib/cloud-init/ds-identify"
+start_cmd="dsidentify_start"
 stop_cmd=":"
 rcvar="cloudinit_enable"
 
-cloudinit_start()
+dsidentify_start()
 {
-    echo -n "${command} starting"
+    echo "${command} starting"
     if kenv -q kernel_options | grep -q 'cloud-init=disabled'; then
         warn "cloud-init is disabled via kernel_options."
     elif test -e /etc/cloud/cloud-init.disabled; then
         warn "cloud-init is disabled via cloud-init.disabled file."
     else
-        ${command} init
+        ${command}
     fi
 }
 
--- a/sysvinit/freebsd/dsidentify.tmpl
+++ b/sysvinit/freebsd/dsidentify.tmpl
@@ -2,6 +2,12 @@
 #!/bin/sh
 
 # PROVIDE: dsidentify
+{#
+once we are correctly using ``paths.run_dir`` / ``paths.get_runpath()`` in the
+python code-base, we can start thinking about how to bring that into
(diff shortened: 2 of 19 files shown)


sysvinit/freebsd/dsidentify
```diff
@@ -0,0 +1,33 @@
...
+# PROVIDE: dsidentify
+# REQUIRE: mountcritlocal
+# BEFORE:  cloudinitlocal
+
+. /etc/rc.subr
+
+PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+
+name="dsidentify"
+command="/usr/local/lib/cloud-init/ds-identify"
```
igalic, line 12: i notice a lot of fixed paths. It makes me wonder if we can somehow replace them with `@LOCALBASE@` or similar, which the ports system would translate for us.

sysvinit/freebsd/dsidentify
```diff
@@ -0,0 +1,33 @@
...
+start_cmd="dsidentify_start"
+stop_cmd=":"
+rcvar="cloudinit_enable"
+
+dsidentify_start()
+{
+    echo "${command} starting"
+    if kenv -q kernel_options | grep -q 'cloud-init=disabled'; then
+        warn "cloud-init is disabled via kernel_options."
+    elif test -e /etc/cloud/cloud-init.disabled; then
```
igalic, line 22: this might make more sense in `/usr/local/etc/…`

sysvinit/freebsd/dsidentify.tmpl
```diff
@@ -0,0 +1,34 @@
+## template:jinja
+#!/bin/sh
+
+# PROVIDE: dsidentify
+# REQUIRE: mountcritlocal
```
igalic, line 5: once we are correctly using `paths.run_dir` / `paths.get_runpath()` in the python code-base, we can start thinking about how to squeeze that into ds-identify itself, and then!, then we can depend on `var_run` instead of `mountcritlocal` here.

igalic (commented):
some day…

sysvinit/freebsd/dsidentify.tmpl
aciba90: Do you think this would be useful as a comment in the code? If so, we could use a [template comment](https://jinja.palletsprojects.com/en/3.0.x/templates/#comments) to avoid render it in the output file.

aciba90 (approved):
LGTM, thanks!

igalic:
something very odd is happening when i try to package this:

```
% python3 ./setup.py install --distro=freebsd --prefix=/usr/local --init-system=sysvinit_freebsd --root=/tmp/tmp.phnOvze98E
…
copying RENDERED_TEMPD6jnfs41_/cloudinitloca -> /tmp/tmp.phnOvze98E/usr/local/etc/rc.d
copying RENDERED_TEMPDt9kv1mz1/cloudfina -> /tmp/tmp.phnOvze98E/usr/local/etc/rc.d
copying RENDERED_TEMPD9k9k2vog/dsidentify -> /tmp/tmp.phnOvze98E/usr/local/etc/rc.d
copying RENDERED_TEMPD1k4nrvrd/cloudconfig -> /tmp/tmp.phnOvze98E/usr/local/etc/rc.d
copying RENDERED_TEMPDpdicbnjj/cloudini -> /tmp/tmp.phnOvze98E/usr/local/etc/rc.d
```

looks like .rstrip(".tmpl") does more than expected: https://docs.python.org/3/library/stdtypes.html#str.rstrip

igalic:
https://github.com/canonical/cloud-init/pull/4209

PR #4184, opened 2023-06-13
integration-tests: pin pytest 7.3.1 to avoid adverse testpaths behavior

### LINKED ISSUE #4183: integration-test target broken due to pytest 7.3.1 behavior change related to testpaths config
# Bug report

pytest 7.3.2 changed treatment of testpaths to now always load `conftest` from the tox configured testpaths. this breaks our integration-tests because our configured `testpaths` in `tox.ini` points to `tests/unittests` which carries with it a dependency on `responses` that is unneeded for our integration-tests target. This behavior breaks all of our jenkins test runners which don't need to express `responses` as a dependency.

This issue is known in [pytest issue: #11104](https://github.com/pytest-dev/pytest/issues/11104) and will likely be reverted in the coming week and introduced as a separate config feature in a major release.

```
tox -e integration-test  # results in a traceback module not found `reponses`
```

blackboxsw:
pytest 7.3.2 changed treatment of `testpaths` config which forced test/unittest/conftest.py module load in our integration-tests toxenv. Upstream bug is tracking their planned revert/fix for this regression. Pin 7.3.1 in the meantime.

Fixes #4183

## Proposed Commit Message

```

```

## Additional Context

Jenkins job failures: https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-lxd_container/lastSuccessfulBuild/console

## Test Steps
```
tox -re integration-test # should work without traceback in your local env
```

## Checklist:

### DIFF
--- a/integration-requirements.txt
+++ b/integration-requirements.txt
@@ -2,5 +2,11 @@
 # https://cloudinit.readthedocs.io/en/latest/topics/integration_tests.html
 #
 pycloudlib==1!3.0.2
-pytest
+
+# Await pytest > 7.3.2. Breaking change in `testpaths` treatment forced
+# test/unittests/conftest.py to be loaded by our integration-tests tox env
+# resulting in an unmet dependency issue:
+# https://github.com/pytest-dev/pytest/issues/11104
+pytest<=7.3.1
+
 packaging
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,11 @@
 # Needed generally in tests
-pytest
+
+# Await pytest > 7.3.2. Breaking change in `testpaths` treatment forced
+# test/unittests/conftest.py to be loaded by our integration-tests tox env
+# resulting in an unmet dependency issue:
+# https://github.com/pytest-dev/pytest/issues/11104
+pytest<=7.3.1
+
 pytest-cov
 pytest-mock
 


TheRealFalcon (approved):
Nit: Might want `pytest<7.3.2` just in case it matters while running on older/other distros, but since  (afaik) we're the only ones running integration tests and we're running from pip, I can't actually think of a way this would be a problem.

aciba90 (dismissed):
Yeah, a hard pin breaks unit testing. We could go with James' commented approach `pytest<7.3.2`, or if we feel brave enough, with `pytest!=7.3.2`.

blackboxsw:
> Yeah, a hard pin breaks unit testing. We could go with James' commented approach `pytest<7.3.2`, or if we feel brave enough, with `pytest!=7.3.2`.

Thanks @aciba90. I definitely do not feel brave today given how it broke us. I'm not certain whether pytest's  next patch version will contain a fix for this or not. So let's err on the side of conservative for this one as we plan on removing our pin in the future as upstream fix is released.

TheRealFalcon (approved):
LGTM!

blackboxsw:
run in progress utilizing this change working in jenkins https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-lxd_container/274/console

PR #2125, opened 2023-04-14
net/dhcp: add udhcpc support

jfroche:
The currently used dhcp client, dhclient, is coming from the unmaintained package, isc-dhcp-client (refer https://www.isc.org/dhcp/) which ended support in 2022.

This change introduce support for the dhcp client, udhcpc, from the busybox project. Busybox advantages are that it is available across many distributions and comes with lightweight executables.

It might be useful to rebase this PR once https://github.com/canonical/cloud-init/pull/2122 is merged.

## Checklist:

 - [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [X] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #4190, opened 2023-06-19
net/dhcp: add udhcpc support

jfroche:
## Proposed Commit Message

```
The currently used dhcp client, dhclient, is coming from the unmaintained package,
`isc-dhcp-client` (refer https://www.isc.org/dhcp/) which ended support in 2022.

This change introduce support for the dhcp client, udhcpc, from the busybox project.
Busybox advantages are that it is available across many distributions and comes with
lightweight executables.
```

## Additional Context

Already discussed in #2125 

## Checklist:

 - [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -112,14 +112,18 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     resolve_conf_fn = "/etc/resolv.conf"
 
     osfamily: str
-    dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd]
+    dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd, dhcp.Udhcpc]
 
     def __init__(self, name, cfg, paths):
         self._paths = paths
         self._cfg = cfg
         self.name = name
         self.networking: Networking = self.networking_cls()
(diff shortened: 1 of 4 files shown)


holmanb:
@jfroche Thanks for putting this together. Since there hasn't been any further activity I assume this is ready for review now?

- [x] CLA signed, thanks!

holmanb:
I'll test this out, thanks.

At first glance it looks like the generic implementation should work, but I don't see the datasource-specific methods implemented which means that this implementation might not work on all clouds.

@jfroche Are there plans to implement those in a follow up PR?

holmanb (approved):
I tested this out. It looks good to me.

Here is a [log file](https://github.com/canonical/cloud-init/files/12089405/cloud-init.log) for an Ubuntu instance using udhcp to set up an ephemeral network using this PR. Future cloud-specific implementation support would be a nice to have for udhcp, but this is a nice addition as is so I'll merge it now.

Thanks @jfroche!

jfroche:
> datasource-specific methods



> I'll test this out, thanks.
> 
> At first glance it looks like the generic implementation should work, but I don't see the datasource-specific methods implemented which means that this implementation might not work on all clouds.
> 
> @jfroche Are there plans to implement those in a follow up PR?

Yes I will make another PR for that.
Thank you for already merging this part !

PR #4198, opened 2023-06-26
Add new datasource for TencentCloud CVM

wynnfeng:
## Proposed Commit Message

```
summary: Add new datasource for TencentCloud CVM 

Support TencentCloud. This datasource inherits from EC2 and add it to ensure cloud-init acts normally on TencentOS and OpencloudOS
```

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceTencentCloud.py
+++ b/cloudinit/sources/DataSourceTencentCloud.py
@@ -1,0 +1,268 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import json
+import os
+import traceback
+
+from cloudinit import dmi
+from cloudinit import log as logging
+from cloudinit import sources, ssh_util, util
+from cloudinit.sources import DataSourceEc2 as EC2
+from cloudinit.sources import DataSourceHostname
+
+TENCENTCLOUD_PRODUCT = "Tencent Cloud CVM"
+LOG = logging.getLogger(__name__)
+
+
+class DataSourceTencentCloud(EC2.DataSourceEc2):
+
+    metadata_urls = ["http://169.254.0.23", "http://metadata.tencentyun.com"]
+
+    # The minimum supported metadata_version from the ec2 metadata apis
+    min_metadata_version = "2017-09-19"
+    extended_metadata_versions: list[str] = []
+
+    def __init__(self, sys_cfg, distro, paths):
+        super(DataSourceTencentCloud, self).__init__(sys_cfg, distro, paths)
+        self.seed_dir = os.path.join(paths.seed_dir, "TencentCloud")
+
+    def set_config_mod_always(self, mod_name):
+        self.set_mod_always(mod_name, "cloud_config_modules")
+
+    def set_mod_always(self, mod_name, stage):
+        modules = self.sys_cfg.get(stage, [])
+        new_modules = []
+        for mod in modules:
+            if mod == mod_name:
+                new_modules.append([mod_name, "always"])
+            else:
+                new_modules.append(mod)
+        self.cfg[stage] = new_modules
+
+    def add_config_mod_always(self, mod_name):
+        modules = self.sys_cfg.get("cloud_config_modules", [])
+        modules.append([mod_name, "always"])
+        self.cfg["cloud_config_modules"] = modules
+
+    def check_item_change(self, item, value):
+        if not value and item != "ssh_key":
+            return False
+        changed = False
+        d_item = {}
+        item_path = "/var/lib/cloud/tencentcloud.item"
+        try:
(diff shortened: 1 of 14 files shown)


wynnfeng:
pylint and  tip-pylint check failure are due to "503  Service Unavailable" ,how can I re-run the checks

wynnfeng:
> @wynnfeng I'm re-running the failed jobs.
thanks!

cloudinit/sources/DataSourceTencentCloud.py
```diff
@@ -0,0 +1,259 @@
...
+                if d_item.get(item) != value:
+                    changed = True
+            else:
+                changed = True
+            if changed:
+                d_item[item] = value
+                with open(item_path, "w") as f:
+                    json.dump(d_item, f)
+        except Exception as e:
+            LOG.debug(str(e))
```
igalic, line 67: this seems excessive

cloudinit/sources/DataSourceTencentCloud.py
```diff
@@ -0,0 +1,259 @@
...
+            LOG.debug(traceback.format_exc())
```
igalic, line 68: as does this

cloudinit/sources/DataSourceTencentCloud.py
```diff
@@ -0,0 +1,259 @@
...
+        with util.SeLinuxGuard(ssh_dir, recursive=True):
+            content, to_del = self.del_authorized_keys(
+                auth_key_entries, key_entries
+            )
+            if to_del:
+                util.ensure_dir(os.path.dirname(auth_key_fn), mode=0o700)
+                util.write_file(auth_key_fn, content, mode=0o600)
+                util.chownbyid(auth_key_fn, pwent.pw_uid, pwent.pw_gid)
+
+    def get_data(self):
```
igalic, line 120: I'm really confused about what's happening in this function

cloudinit/sources/DataSourceTencentCloud.py
igalic, line 68: this logging seems excessive

igalic (commented):
first look

cloudinit/sources/DataSourceTencentCloud.py
wynnfeng: this function gets metadata of the Tencent Cloud instance,like hostname, SSH keys, passwords, etc.and stores them in "self.cfg" for use by other moules and functions

cloudinit/sources/DataSourceTencentCloud.py
wynnfeng: thanks,I remove this line "LOG.debug(traceback.format_exc())"

cloudinit/sources/DataSourceTencentCloud.py
TheRealFalcon: Most of the things being done in this method shouldn't really be done by a datasource. You're potentially replacing data provided by the user with meta data from the IMDS. Cloud-init's generally philosophy is that the user should always have the last say, and this prevents them from doing so. For example, if the user sets their own runcmd, how can that work if you're overwriting it with runcmd data from the IMDS? For this, I would suggest checking if the key exists in userdata, and if so, use it. Otherwise use what you have in your metadata.

I have the same concern with modifying the module definitions in `/etc/cloud/cloud.cfg`. Having `runcmd` get run every boot is going to be confusing from somebody who has used cloud-init on any other cloud, and they don't really have the option to change it since you're changing it here on every boot.

There are generally two ways for clouds to set defaults:
- Modify your OS images to create config overrides in `/etc/cloud/cloud.cfg.d/`. Much of the changes shown here could instead be put there.
- Provide [vendor data](https://cloudinit.readthedocs.io/en/latest/explanation/vendordata.html) rather than overwriting any provided user data.

TheRealFalcon (changes requested):
Overall the code looks good, but I have some concerns about overwriting user data. See my comments inline.

LavenderQAQ:
Does TecentCloud CVM support configuration data sources now? I see the default configuration as follows:
```
datasource_list: [ ConfigDrive, TencentCloud ]
datasource:
  ConfigDrive:
    dsmode: local
  TencentCloud:
    metadata_urls: ['http://169.254.0.23', 'http://metadata.tencentyun.com']
```
But I looked at the run log and found that the source was not there.
```
2023-12-07 03:01:48,015 - __init__.py[DEBUG]: Looking for data source in: ['ConfigDrive', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM']
```
So I found this pr, what is the situation now?

PR #4201, opened 2023-06-26
docs: Correct some bootcmd example wording

TheRealFalcon:
A user initially found the poor wording of `but possibly with more friendly.`, and I added a few other minor changes to fix up the wording. I don't think we need the extra boothook comparisons or explanation here as boothooks are very rarely used anymore.

### DIFF
--- a/doc/examples/cloud-config-boot-cmds.txt
+++ b/doc/examples/cloud-config-boot-cmds.txt
@@ -5,7 +5,7 @@
 # This is very similar to runcmd, but commands run very early
 # in the boot process, only slightly after a 'boothook' would run.
 # - bootcmd will run on every boot
-# - INSTANCE_ID variable will be set to the current instance id
+# - INSTANCE_ID variable will be set to the current instance ID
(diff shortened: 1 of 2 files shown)


doc/examples/cloud-config-boot-cmds.txt
```diff
@@ -2,14 +2,11 @@
...
-# this is very similar to runcmd, but commands run very early
+# This is very similar to runcmd, but commands run very early
 # in the boot process, only slightly after a 'boothook' would run.
-# bootcmd should really only be used for things that could not be
-# done later in the boot process.  bootcmd is very much like
-# boothook, but possibly with more friendly.
 # - bootcmd will run on every boot
-# - the INSTANCE_ID variable will be set to the current instance id.
-# - you can use 'cloud-init-per' command to help only run once
+# - INSTANCE_ID variable will be set to the current instance id
```
s-makin, line 8:
```suggestion
# - INSTANCE_ID variable will be set to the current instance ID
```

s-makin (approved):

aciba90 (approved):
+1 thanks.

PR #4202, opened 2023-06-26
Package snap

TheRealFalcon:
## Proposed Commit Message

```
Allow installing snaps via package_update_upgrade_install module

This also includes a major refactoring in how the packagement management
code is handled. Changes include:
* Backwards compatible change to cc_package_update_upgrade_install
  schema to allow explicitly specifying the package manager to use
* Create PackageManager base class that new package manager classes
  can inherit from
* Allow distros to specify the package managers they support with
  generic install code to install from any of the supported
  package managers
* Create new snap.py and apt.py classes for anything snap and APT
  related respectively
* Move all APT functionality out of debian.py and into apt.py and update
  callers appropriately
* Pull the packaging related calls out of child distro classes and into
  `distros/__init__.py` so distro code continues to work once package
  management code is factored out.
* Add and update tests

Note that this currently only affects debian and ubuntu, along with apt
and snap. Migrating other package managers should be straightforward
enough, but can be done in follow-on PRs.
```

## Additional Context
Even though `distros/__init__.py` has method implementations where it didn't before, these should be safe because all child classes had to create their own implementations/overrides.


cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,213 @@
+import fcntl
```
TheRealFalcon, line 1: Most of this file is taken directly from debian.py

cloudinit/distros/__init__.py
```diff
@@ -134,9 +151,76 @@ def _unpickle(self, ci_pkl_version: int) -> None:
             # missing expected instance state otherwise.
             self.networking = self.networking_cls()
 
-    @abc.abstractmethod
+    def _extract_package_by_manager(
```
TheRealFalcon, line 154: These next two methods still feel overly complicated to me. I'm open to any ideas.

cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,213 @@
...
+        if apt_get_upgrade_subcommand is None:
+            apt_get_upgrade_subcommand = "dist-upgrade"
+        self.apt_command = tuple(apt_get_wrapper_command) + tuple(apt_get_command)
+
+        self.apt_get_upgrade_subcommand = apt_get_upgrade_subcommand
+        self.environment = os.environ.copy()
+        self.environment["DEBIAN_FRONTEND"] = "noninteractive"
+
+    @classmethod
+    def from_config(cls, runner: helpers.Runners, cfg: Mapping) -> "Apt":
```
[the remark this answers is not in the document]
blackboxsw: Sure, it made me re-read what runners were. I get that we'd want semaphore files around related to some expensive PackageManager interactions. +1 on a clear func signature/factory to represent that.

cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,213 @@
...
+        return Apt(
+            runner,
+            apt_get_wrapper_command=get_apt_wrapper(cfg.get("apt_get_wrapper")),
```
blackboxsw, line 96: Whoa are we missing schema defintions and docs on these config options?

cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,213 @@
...
+        # time, but resulting searches become binary searches and are much
+        # faster
+        return set(resp.splitlines())
+
+    def get_unavailable_packages(self, pkglist: Iterable[str]):
+        return [pkg for pkg in pkglist if pkg not in self.get_all_packages()]
+
+    def install_packages(self, pkglist: Iterable[str]) -> UninstalledPackages:
+        self.update_package_sources()
+        unavailable = self.get_unavailable_packages(pkglist)
```
blackboxsw, line 124: While not thrilled about the cost of this lookup function I get that it improves our error handling and fallback to other package managers...  Given that apt-get install X Y NOT_PRESENT doesn't give us the ability to optionally install X and Y if NOT_PRESENT isn't available, this is the best option to look before we leap.

cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,213 @@
...
+        to_install = [p for p in pkglist if p not in unavailable]
+        if to_install:
+            self.run_package_command("install", pkgs=to_install)
+        LOG.debug("Apt cannot install %s", unavailable)
```
blackboxsw, line 128: all caps on APT in messaging everywhere I think

cloudinit/distros/debian.py
```diff
@@ -181,121 +152,12 @@ def _get_localhost_ip(self):
...
-    def update_package_sources(self):
-        self._runner.run(
-            "update-sources",
-            self.package_command,
-            ["update"],
-            freq=PER_INSTANCE,
-        )
+        # As of this writing, the only use of `package_command` outside of
+        # distros calling it within their own classes is calling "upgrade"
+        if command != "upgrade":
```
blackboxsw, line 158: -missing dist-upgrade too?  ... as well as any config overrides in apt_get_upgrade_subcommand looks like

blackboxsw (commented):
Agreed w/ PackageManager subclasses and registration "global" that gives us the ability to define specialized package managers in the future for flatpak, yum, zypper, dnf apk pkg pacman emerge etc if needed. Given that some of the package managers are distro-specific, it may not make sense to refactor the pkg interaction to this layer in all cases. But, if we are interacting with a package manager that is supported on multiple distributions/platforms I like this common API layer for interacting w/ the available package managers as well as a distributions ability to define which pkg mgr to prioritize by default.



As far as your example yaml I think we'd look to something like the following instead of nested lists of lists. Your code is treating it as a dict where key == pkgmgr so I think it's just a typo.

Do we want the user-data to override default distro package pkg manager priority? Maybe not because folks can explicitly use `packages: <preferred_pkg_mgr>: [list, of, pkgs]` to make that opinionated decisions without any supplemental user-data config options.

Please do document and schemify the undoc'd apt_* config options that seem to have slipped through the cracks.
````
packages:
   apt: 
       - pkg1
       - pkg2
    snap: 
       - snap1
       - snap2
```

cloudinit/distros/package_management/package_manager.py
```diff
@@ -0,0 +1,29 @@
+from abc import ABC, abstractmethod
+from typing import Iterable, List, MutableMapping, Type
+
+from cloudinit import helpers
+
+UninstalledPackages = List[str]
+known_package_managers: MutableMapping[str, Type["PackageManager"]] = {}
```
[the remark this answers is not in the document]
TheRealFalcon: I use it in `distros/__init__.py`, and referencing a class attribute directly from outside of the class just feels a little weirder to me. I think that having it outside the class makes it feel a little more publicly accessible. 

Not a strong opinion though.

cloudinit/distros/debian.py
igalic: isn't dist-upgrade called full-upgrade these days?

TheRealFalcon:
Commenting so the stale-pr bot doesn't get me :smile:

cloudinit/distros/debian.py
TheRealFalcon: > -missing dist-upgrade too?

Not as far as I can see. There's no calling code that calls `package_command()` with anything other than `upgrade`.

cloudinit/distros/package_management/apt.py
TheRealFalcon: On closer insepction, this can actually only be specified in the `system_info` section of /etc/cloud/cloud.cfg, so I updated the docs accordingly.

tests/unittests/distros/package_management/test_apt.py
```diff
@@ -0,0 +1,87 @@
+from itertools import count, cycle
```
TheRealFalcon, line 2: This file is just the tests removed from `test_debian.py`

tests/unittests/distros/test_debian.py
```diff
@@ -1,11 +1,7 @@
 # This file is part of cloud-init. See LICENSE file for license information.
-from itertools import count, cycle
 from unittest import mock
 
-import pytest
-
-from cloudinit import distros, subp, util
-from cloudinit.distros.debian import APT_GET_COMMAND, APT_GET_WRAPPER
+from cloudinit import distros, util
```
TheRealFalcon, line 4: Tests removed here were moved to `test_apt.py`.

cloudinit/distros/__init__.py
```diff
@@ -139,9 +157,88 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+
+        # First install packages using package manager(s)
+        # supported by the distro
+        uninstalled = []
+        for manager in self.package_managers:
+            to_try = (
+                packages_by_manager.get(manager.__class__, set())
+                | generic_packages
+            )
+            uninstalled = manager.install_packages(to_try)
```
blackboxsw, line 214: If we only have generic_packages and len (distro.package_managers) == 1. We could provide a `strict=True` or `force=True` flag to install_packages to force using this package manager instead of try/fallback on unavailable. This could give us the ability to avoid checks on unavailable packages if we think that is something worth avoiding on debian or stable ubuntu.

cloudinit/distros/__init__.py
```diff
@@ -139,9 +157,88 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+            }
+            if failed:
+                LOG.error(error_message, failed)
+            generic_packages = set(uninstalled)
+
+        # Now attempt any specified package managers not explicitly supported
+        # by distro
+        for manager, packages in packages_by_manager.items():
+            if manager.name not in known_package_managers:
+                LOG.error(
```
blackboxsw, line 226: It doesn't look like we have unittest/integration coverage for this error, either would be fine.

I think we actually want to raise an error or continue here otherwise we continue after this and fail below with an ugly traceback on 234 because presumably `manager` is not known/supported in this image.

cloudinit/distros/__init__.py
```diff
@@ -139,9 +157,88 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+                    "Cannot install any packages listed under unknown package "
+                    "manager: %s.",
+                    manager,
+                )
+            if manager.name in [p.name for p in self.package_managers]:
+                # We already installed/attempt these; don't try again
+                continue
+            uninstalled.extend(
```
blackboxsw, line 234: 1. Nothing is covering this in tests either. Maybe we want something like:

```diff
--- a/tests/unittests/config/test_cc_package_update_upgrade_install.py
+++ b/tests/unittests/config/test_cc_package_update_upgrade_install.py
@@ -75,6 +75,17 @@ class TestMultiplePackageManagers:
         assert mock.call(["snap", "install", "pkg1"]) in m_subp.call_args_list
         assert mock.call(["snap", "install", "pkg2"]) in m_subp.call_args_list
 
+    @mock.patch("cloudinit.subp.subp")
+    def test_explicit_snap_when_not_distro_default(self, m_subp, common_mocks):
+        import pdb; pdb.set_trace()
+        cloud = get_cloud("debian")
+        cfg = {"packages": [{"snap": ["pkg1", "pkg2"]}]}
+        handle("", cfg, cloud, [])
+
+        assert len(m_subp.call_args_list) == 2
+        assert mock.call(["snap", "install", "pkg1"]) in m_subp.call_args_list
+        assert mock.call(["snap", "install", "pkg2"]) in m_subp.call_args_list
+
     @mock.patch("cloudinit.subp.subp")
     def test_explicit_snap_version(self, m_subp, common_mocks):
         cloud = get_cloud("ubuntu")
```


2. One thought this brings up is, if debian doesn't have snap command installed should the package_manager have a `available/viable` check to determine if it is even compatible? If not we will traceback with a called process error here on `command not found`. We can either add a `PackageManager.is_applicable/available` check or better handle a ProcessExecutionError when the PackageManager primary command is absent.

blackboxsw (changes requested):
Thanks @TheRealFalcon for pulling this together and patience on the review. I hit a couple of snags will schema annotation on invalid schema now that we are representing nested dicts in lists for this schema. I'll provide a separate PR for the schema annotation fixes, but I wanted to capture the review comments I have so far before pushing a separate PR for schema annotation fixes.

I have a number of questions and suggestion inline. Biggest question though is how we are handling "unknown package managers" from a distro perspective when user-data requests "snap" yet the image itself may not contain such a package manager by default.

cloudinit/distros/package_management/apt.py
```diff
@@ -0,0 +1,227 @@
...
+        self.update_package_sources()
+        pkglist = util.expand_package_list("%s=%s", list(pkglist))
+        unavailable = self.get_unavailable_packages(
+            [x.split("=")[0] for x in pkglist]
+        )
+        LOG.debug(
+            "The following packages were not found by APT so APT will "
+            "not attempt to install them: %s",
+            unavailable,
+        )
```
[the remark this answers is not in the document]
TheRealFalcon: > This changes default behavior in the case where we had packages: [pkg1, pkg2] in that we now pay the cost of get_unavailable_packages for any previous #cloud-config which provides a generic pkg list.

This doesn't actually change **behavior** though, right? It just adds some search time?

> On distributions like debian where only one package_manager is defined, do we want to avoid this optional unavailable check and just attempt to install all packages provided and "fail hard" if some don't exist?

Our inheritance rules make this a little tricky. I'm also not sure I see a super compelling reason to.

> So, maybe that small cost is worth the better handling and support of snaps on stable releases.

This was my thinking. If you're installing packages, you're already adding more than a handful of seconds to your boot, so this should be fairly trivial compared to that. It doesn't impact the "fast boot" use case. Also, package install also comes after ssh, so it shouldn't impact time to ssh.

cloudinit/distros/__init__.py
TheRealFalcon: Per your 2, I think I just missed a `continue`. If you go up two blocks, we have
```
            if manager.name not in known_package_managers:
                LOG.error(
                    "Cannot install any packages listed under unknown package "
                    "manager: %s.",
                    manager,
                )
```
That was meant to cover the case where somebody specifies something like snap but snap isn't installed. If I add continue after the error, that should address the problem,correct?

cloudinit/distros/package_management/apt.py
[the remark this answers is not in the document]
blackboxsw: > This doesn't actually change behavior though, right? It just adds some search time?
 Correct, yes no config changes or artifacts from cloud-init other than potential upfront cost to parse the apt-cache. Which again is not bad relative to the actual apt update we just performed and the software we need to install and configure during first boot anyway.... We've also added more error handling/coping if we want to fallback to snap'd packages in cases where the deb isn't available in apt, so I think it's a defensible cost to pay here.

>  Also, package install also comes after ssh, so it shouldn't impact time to ssh.
+1 to this too

cloudinit/distros/__init__.py
TheRealFalcon: (original comment deleted).

This will be refactored in the next push

TheRealFalcon:
Also, I'll push a rebase to fix the changes

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1924,19 +1939,28 @@
...
+                    "items": {
+                      "$ref": "#/$defs/package_item_definition"
+                    }
+                  },
+                  "snap": {
+                    "type": "array",
+                    "items": {
+                      "$ref": "#/$defs/package_item_definition"
+                    }
+                  }
```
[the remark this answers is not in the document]
blackboxsw: @TheRealFalcon still think we are missing this for strict schema disallowing non-snap/apt package manager keys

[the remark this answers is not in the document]
TheRealFalcon: Whoops, I put it in the wrong place. Updated

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1924,22 +1939,32 @@
...
+                }
               },
               {
-                "type": "string"
+                "$ref": "#/$defs/package_item_definition"
               }
             ]
           },
+          "additionalProperties": false,
```
blackboxsw: @TheRealFalcon I see you placed this additionalProperties:false here this doesn't end up applying to the oneOf object:
With your existing schema we still allow the following:
```
cat > pkg.yaml <<EOF
#cloud-config
packages:
 - apta: ["sl"]
EOF
$ PYTHONPATH=. python3 -m cloudinit.cmd.main schema -c pkg.yaml
Valid cloud-config: pkg.yaml
```

TheRealFalcon: That's not what I'm seeing:
```
root@me:~/a# cat some-config 
#cloud-config
packages:
 - apta: ["sl"]
root@me:~/a# cloud-init schema --config-file some-config
Invalid cloud-config some-config
Error: Cloud config schema errors: packages.0: Additional properties are not allowed ('apta' was unexpected), packages.0: {'apta': ['sl']} is not valid under any of the given schemas

Error: Invalid cloud-config schema: user-data
```
AFAIK, `additionalProperties` isn't valid inside the `properties` definition. It goes outside.

blackboxsw (commented):
Thank you for this @TheRealFalcon, I also evaluated cost of install from generic package list for a known snap-only package and saw that our costs of apt cache check are < 100 ms to determine that we don't want to try/fail to install the uninstallable package before falling back to snap install. This is an ok price to pay when we have to fallback from generic packages that are not found in APT.

One last thing we need to address is only specific secondary package mangement sources are provided, we probably shouldn't be calling the primary PackageManager.install_packages in the first place. This leads to wasted `apt update` being called on user-data like the following:
````
#cloud-config    
packages:
- snap: [mattermost-desktop]
```


to avoid this I think we need a trivial check to avoid wasted calls on primary package_manager:
```diff
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index b272ff73a..67b66940b 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -210,6 +210,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 packages_by_manager.get(manager.__class__, set())
                 | generic_packages
             )
+            if not to_try:
+                continue
             uninstalled = manager.install_packages(to_try)
             failed = {
                 pkg for pkg in uninstalled if pkg not in generic_packages
```
Otherwise +1

blackboxsw (approved):
LGTM. one minor commit added to avoid calling update_package_sources based on a non-empty packages list. We will be calling that anyway for specific package_managers based on processing the packages later. No loss in doing this apt update later if/when needed.

PR #4209, opened 2023-06-28
setup: fix generation of init templates

igalic:
## Proposed Commit Message

```
setup: fix generation of init templates

if an init file ends in any of the letters [tmpl] these will be stripped
in addition to the .tmpl extension.

This is documented behaviour of the str.rstrip() method.
This PR fixes it by using os.path.splitext() (rather than
str.removesuffix which is only available in python 3.9+)

Since BSD sysvinit scripts are now rendered as template,
we need to fix an oversight from #4182 and pass the mode.

```

## Additional Context

https://github.com/canonical/cloud-init/pull/4182

## Test Steps

successfully tested on Hetzner

## Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -167,10 +167,14 @@ for a in sys.argv:
 INITSYS_FILES = {
     "sysvinit": [f for f in glob("sysvinit/redhat/*") if is_f(f)],
     "sysvinit_freebsd": [
-        render_tmpl(f) for f in glob("sysvinit/freebsd/*") if is_f(f)
+        render_tmpl(f, mode=0o755)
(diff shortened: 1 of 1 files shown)


igalic:
cc @aciba90 who reviewed #4182.

aciba90 (approved):
LGTM, thanks for caching and fixing this.

This _footgun_ was introduced in dcbe479, and we did not see it because none of the templates finished with none of the chars in `.tmpl` until [529e108](https://github.com/canonical/cloud-init/commit/529e108ca1b2e839539588c4cea040232440dbeb).

PR #4218, opened 2023-07-03
Specify build-system for pep517 [SC-1492]

TheRealFalcon:
## Proposed Commit Message

```
Specify build-system for pep517

Make it clear to various packaging frontends that we're using
setuptools as our backend.
```

## Test Steps
I created 3 separate builds:
`python setup.py sdist`
`python -m build` without pyproject.toml changes
`python -m build` with pyproject.toml changes

All three builds contained exactly the same contents


aciba90 (approved):
+1, thanks!

Issue #4219, opened 2023-07-04
Network v2 -> ENI yielding unexpected results


TheRealFalcon:
# Bug report
Given the v2 config:
```
version: 2
ethernets:
    eth1:
        match:
            macaddress: cf:d6:af:48:e8:80
        set-name: eth1
    eth99:
        addresses:
        - 192.168.21.3/24
        dhcp4: true
        match:
            macaddress: c0:d6:9f:2c:e8:80
        nameservers:
            addresses:
            - 8.8.8.8
            - 8.8.4.4
            search:
            - barley.maas
            - sach.maas
        routes:
        -   metric: 10000
            to: 0.0.0.0/0
            via: 65.61.151.37
        set-name: eth99
```

Rendering to eni gives us:
```
auto lo
iface lo inet loopback
    dns-nameservers 8.8.8.8 8.8.4.4
    dns-search barley.maas sach.maas

iface eth1 inet manual

auto eth99
iface eth99 inet dhcp
    dns {'nameservers': ['8.8.8.8', '8.8.4.4'], 'search': ['barley.maas', 'sach.maas']}
    post-up route add default gw 65.61.151.37 metric 10000 || true
    pre-down route del default gw 65.61.151.37 metric 10000 || true

# control-alias eth99
iface eth99 inet static
    address 192.168.21.3/24
    dns-nameservers 8.8.8.8 8.8.4.4
    dns-search barley.maas sach.maas
```
In particular, the `dns {'nameservers': ['8.8.8.8', '8.8.4.4'], 'search': ['barley.maas', 'sach.maas']}` line is invalid.

## Steps to reproduce the problem
Put the aforementioned v2 yaml into a file named `tmp.yaml`. Then run:
```
# cloud-init devel net-convert -p tmp.yaml -k yaml -d /tmp -D ubuntu -O eni
```
Results can be found under `/tmp/etc/network`

## Environment details
tip of main branch.

dermotbradley:
Hmm, I never tried testing BOTH dhcp and static IP at the same time.

Related to what you're noticed when there are multiple "iface" stanzas for the **same** ethernet interface routes are always added to the first stanza which in many cases is incorrect.

Take the following for example:

```
version: 2
ethernets:
  eth0:
    dhcp4: true
    addresses:
      - 10.20.1.2/24
    routes:
      - to: 10.20.1.0/24
        via: 10.20.1.1
```

this results in a /e/n/i file:

```
auto lo
iface lo inet loopback

[...]

PR #4220, opened 2023-07-04
Fix network v2 metric rendering

TheRealFalcon:
## Proposed Commit Message

```
Fix network v2 metric rendering

Metric info was not being included in v2-based routes.

Fixes GH-4217
```

## Additional Context
Also found #4219 in fixing this


aciba90 (approved):
+1, thanks!

PR #4223, opened 2023-07-05
Update LXD URL

s-makin:
```
LXD documentation has moved and broken this link as a result.
I believe this is the correct one to redirect to, but please
advise if not.
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -79,7 +79,7 @@ meta: MetaSchema = {
             # storage_pools, profiles, projects, clusters and core config,
             # `lxd:preseed` config will be passed as stdin to the command:
             #  lxd init --preseed
-            # See https://linuxcontainers.org/lxd/docs/master/preseed/ or
+            # See https://documentation.ubuntu.com/lxd/en/latest/howto/initialize/#non-interactive-configuration or
             # run: lxd init --dump to see viable preseed YAML allowed.
             #
             # Preseed settings configuring the LXD daemon for HTTPS connections
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -149,7 +149,7 @@ meta: MetaSchema = {
                     name: limited
 
 
-            """
+            """  # noqa: E501
         ),
     ],
     "frequency": PER_INSTANCE,


cloudinit/config/cc_lxd.py
```diff
@@ -79,7 +79,7 @@
             # storage_pools, profiles, projects, clusters and core config,
             # `lxd:preseed` config will be passed as stdin to the command:
             #  lxd init --preseed
-            # See https://linuxcontainers.org/lxd/docs/master/preseed/ or
+            # See https://documentation.ubuntu.com/lxd/en/latest/howto/initialize/#non-interactive-configuration or
```
aciba90, line 82: I think might fix it:

```suggestion
            # See 
            # https://documentation.ubuntu.com/lxd/en/latest/howto/initialize/#non-interactive-configuration
            # or
```

aciba90 (changes requested):
LGTM but CI complains: https://github.com/canonical/cloud-init/actions/runs/5462632134/jobs/9942252697?pr=4223

cloudinit/config/cc_lxd.py
aciba90: or:

```suggestion
            # See https://documentation.ubuntu.com/lxd/en/latest/howto/initialize/#non-interactive-configuration or  # noqa
```

s-makin: I'm not sure I understand the difference between the two - which one is preferred?

aciba90: Sorry for the lack of explanation. They were clumsy ways to disable the [line too long](https://github.com/canonical/cloud-init/actions/runs/5462632134/jobs/9942252697?pr=42230) complaint from flake8.

The problem is that those lines get rendered in the final docs, so anything we append to the line will get rendered in the docs.

As a probably better way to disable the lint error, I would suggest to apply the following diff:

```diff
diff --git a/cloudinit/config/cc_lxd.py b/cloudinit/config/cc_lxd.py
index 2b8837aac..5dcb620b1 100644
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -149,7 +149,7 @@ meta: MetaSchema = {
                     name: limited
 
 
-            """
+            """  # noqa: E501
         ),
     ],
     "frequency": PER_INSTANCE,

```

This disables the E501 error on the whole string block without affecting the rendered docs.

s-makin: Thanks for the suggestion - and the explanation! :) I've changed that line as per your comment

aciba90 (approved):
+1 thanks!

aciba90:
@s-makin, something weird is going on here, the new link is not in [reference/modules.html#lxd](https://canonical-cloud-init.readthedocs-hosted.com/en/latest/reference/modules.html#lxd), or am I missing something?

PR #4224, opened 2023-07-05
instance-data: add system-info and features to combined-cloud-config

blackboxsw:
```
instance-data: add system-info and features to combined-cloud-config

Some snap environments (subiquity live installer) have different
versions of cloudinit within the snap environment than is installed
on the host system. Invoking cloudinit.stages.Init().read_cfg on an
older version of cloudinit potentially breaks when the host environment
has a newer version of cloud-init because the obj.pkl can not be
deserialized on an older version of cloud-init.

To avoid having to invoke cloudinit.stages.Init().read_cfg() to get
system_info and default_user config, provide this config in:
 - /run/cloud-init/combined-cloud-config.json and
 - /run/cloud-init/instance-data-sensitive.json:merged_system_cfg key

Add a system_info key in combined-cloud-config.json as seen in
/etc/cloud/cloud.cfg*.

Also add a util.get_features function which will seed the "features"
key in combined-cloud-config.json. This also prevents the need for
importing cloudinit.features to read that value or overridden feature
values.
```

## Checklist:

### DIFF
--- a/cloudinit/features.py
+++ b/cloudinit/features.py
@@ -14,6 +14,9 @@ the flag and intended lifetime.
 Tests are required for new feature flags, and tests must verify
 all valid states of a flag, not just the default state.
 """
+import re
+import sys
+from typing import Dict
 
 ERROR_ON_USER_DATA_FAILURE = True
 """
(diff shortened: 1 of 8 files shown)


dbungert:
Thanks Chad.

Using a deb of this branch for cloud-init outside the snap, and sending cloud-config with no default user, I can see a system_info.default_user.gecos == 'Ubuntu' - is that expected?

blackboxsw:
> Thanks Chad.
> 
> Using a deb of this branch for cloud-init outside the snap, and sending cloud-config with no default user, I can see a system_info.default_user.gecos == 'Ubuntu' - is that expected?

@dbungert yep/ Gecos is set by default to just Ubuntu. subiquity or live installer likely will be looking for `system_info.default_user.name` I believe not the GECOS comment/user information field as cloud-init doesn't set that to something super helpful and it really could contain almost anything verbose about the default user, Amazon linux gives that value "EC2 Default User", rhel calls it "Cloud user", open mandriva, "Openmandriva admin" etc.

Issue #3769, opened 2023-05-12
cloud-init status --wait returns before cloud-final has finished executing


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1890528](https://bugs.launchpad.net/cloud-init/+bug/1890528)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = chad.smith
assignee_name = Chad Smith
date_closed = None
date_created = 2020-08-06T04:17:34.808808+00:00
date_fix_committed = None
date_fix_released = None
id = 1890528
importance = high
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1890528
milestone = None
owner = jeffjo
owner_name = Jeff Jo
private = False
status = triaged
submitter = jeffjo
submitter_name = Jeff Jo
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Jeff Jo(jeffjo)** wrote on 2020-08-06T04:17:34.808808+00:00_

I'm instrumenting our EC2 instances to report whether cloud-init succeeded or failed during the  boot process. I've implemented this as a systemd unit that is triggered early in the boot process that runs a script that resembles the following:

cloud_init_status=0
cloud-init status --wait || cloud_init_status=$?

if [ "$cloud_init_status" = "0" ]; then
  report_launch_success
else
  report_launch_failure
fi

I was expecting `cloud-init status --wait` to only return after cloud-final has completed, but I discovered today that it can return early if there is an error encountered during the cloud-config stage.

I reported this in IRC and @blackboxsw thought the issue might be somewhere in this code ( https://github.com/canonical/cloud-init/blob/a13febd286d21f1754e32f4a05e722039eb452b8/cloudinit/cmd/status.py#L133-L144) and suggested I file a bug here.

I am using cloud-init 20.2-45 on Xenial.

ubuntu-server-builder:
_Launchpad user **Dan Watkins(oddbloke)** wrote on 2020-08-10T19:46:23.966132+00:00_

Thanks for filing this bug, Jeff!  I agree with Chad, this is a bug; the help text for the --wait options reads:

  -w, --wait  Block waiting on cloud-init to complete

I can't think of cases where someone might be running `cloud-init status --wait` and relying on the current behaviour of exit-early-in-case-of-error, so I think the correct thing to do here is to modify the behaviour of --wait to match its help text.

ubuntu-server-builder:
[...]

PR #4230, opened 2023-07-07
Inspect systemd state for cloud-init status

TheRealFalcon:
## Proposed Commit Message

```
Inspect systemd state for cloud-init status

Relying solely on status.json and result.json can result in reporting
error before cloud-init has fully completed. Using systemd state can
give us more correct info.

Fixes GH-3769
```

## Additional Context
#3769

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -224,9 +224,8 @@ def _get_systemd_status() -> Optional[UXAppStatus]:
             states["UnitFileState"].startswith("enabled")
             or states["UnitFileState"] == "static"
         ):
-            # Service is not set to run, so don't determine state
-            # based on this service
-            continue
+            # Individual services should not get disabled
+            return UXAppStatus.ERROR
         if (
             states["ActiveState"] == "active"
             and states["SubState"] == "exited"
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -15,7 +15,7 @@ import sys
 from time import gmtime, sleep, strftime
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
 
-from cloudinit import safeyaml
+from cloudinit import safeyaml, subp
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.distros import uses_systemd
 from cloudinit.helpers import Paths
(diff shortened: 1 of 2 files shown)


cloudinit/cmd/status.py
```diff
@@ -196,6 +196,54 @@ def get_bootstatus(disable_file, paths) -> Tuple[UXAppBootStatusCode, str]:
...
+        ):
+            # Service is not set to run, so don't determine state
+            # based on this service
+            continue
+        if (
+            states["ActiveState"] == "active"
+            and states["SubState"] == "exited"
+        ):
+            # Service exited normally, nothing interesting from systemd
+            continue
```
blackboxsw, line 235: If `cloud-init-final.service` exited normally, do we really want to loop through all prior required boot stages? Or can we say, last stage of cloud-init has exited success, so we are done.

cloudinit/cmd/status.py
```diff
@@ -196,6 +196,54 @@ def get_bootstatus(disable_file, paths) -> Tuple[UXAppBootStatusCode, str]:
...
+        states = dict(
+            [[x.strip() for x in r.split("=")] for r in stdout.splitlines()]
+        )
+        if not (
+            states["UnitFileState"].startswith("enabled")
+            or states["UnitFileState"] == "static"
+        ):
+            # Service is not set to run, so don't determine state
+            # based on this service
+            continue
```
blackboxsw, line 229: I think we should consider this an error condition because cloud-init boot stages need to be active and enabled, and run serially from local -> init -> config -> final for us to support cloud-init typical first boot config. The times we've have bugs filed that cloud-init didn't work (and nothing reported an error) was due to systemd ordering cycles resulting in cloud-init systemd units being kicked out of boot targets. I'd suggest if we are not in systemd generator "DISABLED" state, all systemd units need to be enabled for us to claim cloud-init is happy/supportable.

blackboxsw (changes requested):
Request changes pending discussion on how best to handle of disabled services.

cloudinit/cmd/status.py
TheRealFalcon: I realize it's unlikely, but I was thinking it might be possible for an earlier service file to exit with error in a way that might not make it's way to status or result jsons. Not something we should worry about?

holmanb (approved):
Looks good to me.

cloudinit/cmd/status.py
blackboxsw: works for me. I think it's possible that we can have a failed earlier stage and still execute final etc. So I may be incorrect in that assumptions. Ship it

blackboxsw (approved):
LGTM!

PR #4232, opened 2023-07-08
cc_mounts.py: Add note about issue with creating mounts inside mounts

dermotbradley:
## Proposed Commit Message

```
cc_mounts.py: Add note about issue with creating mounts inside mounts

Creating two, or more, mounts where the mountpoint of one is inside the
mountpoint of another does not work correctly.

Added a note detailing how to workaround this on systems using util-linux's
mount utility.

Also rewrite the description of mount_default_fields to clarify how it works.
```

## Checklist:


TheRealFalcon (commented):
Thanks for the additional info here. I modified the text to emphasize certain words/options like we do in other parts of the docs. That also forced me to move some lines around due to line length, but I otherwise kept your changes.

If that sounds good to you, than you can apply the following diff
```diff
diff --git a/cloudinit/config/cc_mounts.py b/cloudinit/config/cc_mounts.py
index da91a6db5..0cd14340a 100644
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -63,15 +63,15 @@ swap file is created.
 .. note::
     If multiple mounts are specified where a subsequent mount's mountpoint is
     inside of a previously declared mount's mountpoint (i.e. the 1st mount has
-    a mountpoint of /abc and the 2nd mount has a mountpoint of /abc/def) then
-    this will not work as expected - ``cc_mounts`` first creates the
-    directories for all the mountpoints **before** it starts to actually
+    a mountpoint of ``/abc`` and the 2nd mount has a mountpoint of
+    ``/abc/def``) then this will not work as expected - ``cc_mounts`` first
+    creates the directories for all the mountpoints **before** it starts to
     perform any mounts and so the sub-mountpoint directory will not be created
     correctly inside the parent mountpoint.
 
-    For systems using util-linux's ```mount``` program this issue can be
-    worked around by specifying "X-mount.mkdir" as part of a fs_mntops value
-    for the subsequent mount entry.
+    For systems using util-linux's ``mount`` program this issue can be
+    worked around by specifying ``X-mount.mkdir`` as part of a ``fs_mntops``
+    value for the subsequent mount entry.
 """
 
 example = dedent(
```

dermotbradley:
> If that sounds good to you, than you can apply the following diff

Done.

dermotbradley:
BTW just to point out that with the above mentioned workaround in place the "def" subdirectory will still be created inside "/abc" even though it is not used.

The only way to prevent this/fix this would be to modify the cc_mounts.py code so that rather than its current approach of (a) create all mount directories and (b) run "mount -a" that it instead did a combination of (a) create mount directory and (b) mount onto that directory individually for each mount.

TheRealFalcon:
Now there's a line too long :smile: 

I had also removed the word "actually" as it seemed unnecessary and I think that's what brought the line length down under 80.

TheRealFalcon:
> BTW just to point out that with the above mentioned workaround in place the "def" subdirectory will still be created inside "/abc" even though it is not used.

That shouldn't really be a problem though, right? We can handle that separately.

dermotbradley:
> That shouldn't really be a problem though, right? We can handle that separately.

Yeah, it's not an operational problem, just unecessary directory/directories being created.

dermotbradley:
> Now there's a line too long smile

Fixed.

TheRealFalcon (approved):
LGTM, thanks!

PR #4233, opened 2023-07-08
landscape: use landscape-config utility

blackboxsw:
## Proposed Commit Message

```
landscape: use landscape-config to write configuration

Avoid writing directly to /etc/landscape/client.conf as formats
or config file location could change.
Invoke the `landscape-config` utility instead to persist landscape
client config options.

No longer restart landscape-client service because
landcape-config starts landscape-client with updated values.

Add doc examples for installing via PPA using apt: config.
```

## Additional Context

SC-1531

## Checklist:

### DIFF
--- a/cloudinit/config/cc_landscape.py
+++ b/cloudinit/config/cc_landscape.py
@@ -153,7 +153,6 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     )
     subp.subp(["landscape-config", "--silent"] + cmd_params)
     util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
-    cloud.distro.manage_service("restart", "landscape-client")
 
 
 def merge_together(objs):
--- a/cloudinit/config/cc_landscape.py
+++ b/cloudinit/config/cc_landscape.py
@@ -9,7 +9,7 @@
 """install and configure landscape client"""
 
 import logging
-import re
+from itertools import chain
 from textwrap import dedent
 
 from configobj import ConfigObj
@@ -140,18 +140,20 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         LSC_CLIENT_CFG_FILE,
         ls_cloudcfg,
     ]
-    cmd_params = []
-    for k, v in sorted(merge_together(merge_data)["client"].items()):
-        if re.search(r"(\"|\s)", v):
-            cmd_params.append(
-                f"--{k.replace('_', '-')}=\"" + v.replace('"', '"') + '"'
-            )
-        else:
-            cmd_params.append(f"--{k.replace('_', '-')}={v}")
-
+    # Flatten dict k,v pairs to [--KEY1, VAL1, --KEY2, VAL2, ...]
+    cmd_params = list(
+        chain(
+            *[
+                [f"--{k.replace('_', '-')}", v]
+                for k, v in sorted(
+                    merge_together(merge_data)["client"].items()
+                )
+            ]
+        )
+    )
     subp.subp(["landscape-config", "--silent"] + cmd_params)
     util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
-    subp.subp(["service", "landscape-client", "restart"])
+    cloud.distro.manage_service("restart", "landscape-client")
 
 
 def merge_together(objs):
--- a/cloudinit/config/cc_landscape.py
+++ b/cloudinit/config/cc_landscape.py
@@ -9,6 +9,7 @@
 """install and configure landscape client"""
 
 import logging
+import re
 from textwrap import dedent
 
 from configobj import ConfigObj
@@ -70,7 +71,6 @@ meta: MetaSchema = {
             # To discover additional supported client keys, run
             # man landscape-config.
             landscape:
-                install_source: distro
                 client:
                     url: "https://landscape.canonical.com/message-system"
                     ping_url: "http://landscape.canonical.com/ping"
@@ -85,11 +85,25 @@ meta: MetaSchema = {
         ),
         dedent(
             """\
-            # Any keys below `client` are optional and the default values will
-            # be used.
+            # Minimum viable config requires account_name and computer_title
             landscape:
-                client: {}
+                client:
+                    computer_title: kiosk 1
+                    account_name: Joe's Biz
             """
+        ),
+        dedent(
+            """\
+           # To install landscape-client from a PPA, specify apt.sources
+           apt:
+               sources:
+                 trunk-testing-ppa:
+                   source: ppa:landscape/self-hosted-beta
+           landscape:
+               client:
+                 account_name: myaccount
+                 computer_title: himom
+           """
         ),
     ],
     "frequency": PER_INSTANCE,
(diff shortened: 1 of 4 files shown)


cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1480,6 +1480,11 @@
           ],
           "additionalProperties": false,
           "properties": {
+            "install_source": {
+              "type": "string",
+              "pattern": "^(distro|ppa:.+)$",
+              "description": "Whether to install landscape-client deb package from the 'distro' -updates pocket or a custom 'ppa:YOUR_PPA'. Default: 'distro'"
```
aciba90, line 1486: Nit: As this used gets rendered in cloud-init docs, we could probably be less specific for the sake of the less ubuntu-versed users.

```suggestion
              "description": "Whether to install landscape-client deb package from the 'distro' repos or a custom 'ppa:YOUR_PPA'. Default: 'distro'"
```

aciba90 (approved):
LGTM! Thanks for the additional migration to pytest.

blackboxsw:
@aciba90 in manually testing this on a lxd VM with landscape-client installed I discovered that the quotag around the  cmd_params was causing landscape to break and not register the VM. I filed https://bugs.launchpad.net/landscape-client/+bug/2027521 to track this in landscape, but for cloud-init's part, we should only quote values which contain whitespace or double-quotes (which we will escape).   I've also backed out the install_source key from landscape, because we should be managing apt PPAs in the `apt:` key and not reflect that config into each module. And apt_configure module and package_update_upgrade_install are both called before landscape... so I added an example to cc_landscape for installing from PPA.

cloudinit/config/cc_landscape.py
```diff
@@ -128,14 +140,16 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
...
-    merged.write(contents)
-
-    util.ensure_dir(os.path.dirname(LSC_CLIENT_CFG_FILE))
-    util.write_file(LSC_CLIENT_CFG_FILE, contents.getvalue())
-    LOG.debug("Wrote landscape config file to %s", LSC_CLIENT_CFG_FILE)
+    cmd_params = []
+    for k, v in sorted(merge_together(merge_data)["client"].items()):
+        if re.search(r"(\"|\s)", v):
+            cmd_params.append(
+                f"--{k.replace('_', '-')}=\"" + v.replace('"', '"') + '"'
```
aciba90, line 147: I believe the intention is to quote the inner double quotes, right ?

```suggestion
                f"--{k.replace('_', '-')}=\"" + v.replace('"', '\\"') + '"'
```

aciba90 (changes requested):
Thanks for not adding the new ppa key and keep it simple. About the escaping issue, I believe we missed it somehow. Here is a diff, adding the escaping:

```diff
diff --git a/cloudinit/config/cc_landscape.py b/cloudinit/config/cc_landscape.py
index 54c177403..91f46b643 100644
--- a/cloudinit/config/cc_landscape.py
+++ b/cloudinit/config/cc_landscape.py
@@ -142,12 +142,10 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     ]
     cmd_params = []
     for k, v in sorted(merge_together(merge_data)["client"].items()):
+        k = k.replace("_", "-")
         if re.search(r"(\"|\s)", v):
-            cmd_params.append(
-                f"--{k.replace('_', '-')}=\"" + v.replace('"', '"') + '"'
-            )
-        else:
-            cmd_params.append(f"--{k.replace('_', '-')}={v}")
+            v = '"{}"'.format(v.replace('"', r"\""))
+        cmd_params.append(f"--{k}={v}")
 
     subp.subp(["landscape-config", "--silent"] + cmd_params)
     util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
diff --git a/tests/unittests/config/test_cc_landscape.py b/tests/unittests/config/test_cc_landscape.py
index 51fa39825..766932361 100644
--- a/tests/unittests/config/test_cc_landscape.py
+++ b/tests/unittests/config/test_cc_landscape.py
@@ -173,7 +173,7 @@ class TestLandscape:
                 [
                     "landscape-config",
                     "--silent",
-                    '--computer-title="My" PC"',
+                    '--computer-title="My\\" PC"',
                     "--data-path=/var/lib/landscape/client",
                     "--log-level=info",
                     "--ping-url=http://landscape.canonical.com/ping",
```

Note that:

```py
>>> subprocess.run('echo \"', shell=True).stdout
/bin/sh: 1: Syntax error: Unterminated quoted string

>>> subprocess.run('echo \\"', shell=True).stdout
"
```

blackboxsw:
> Thanks for not adding the new ppa key and keep it simple. About the escaping issue, I believe we missed it somehow. Here is a diff, adding the escaping:
> 
> ```diff
> diff --git a/cloudinit/config/cc_landscape.py b/cloudinit/config/cc_landscape.py
> index 54c177403..91f46b643 100644
> --- a/cloudinit/config/cc_landscape.py
> +++ b/cloudinit/config/cc_landscape.py
> @@ -142,12 +142,10 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
>      ]
>      cmd_params = []
>      for k, v in sorted(merge_together(merge_data)["client"].items()):
> +        k = k.replace("_", "-")
>          if re.search(r"(\"|\s)", v):
> -            cmd_params.append(
> -                f"--{k.replace('_', '-')}=\"" + v.replace('"', '"') + '"'
> -            )
> -        else:
> -            cmd_params.append(f"--{k.replace('_', '-')}={v}")
> +            v = '"{}"'.format(v.replace('"', r"\""))
> +        cmd_params.append(f"--{k}={v}")
>  
>      subp.subp(["landscape-config", "--silent"] + cmd_params)
>      util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
> diff --git a/tests/unittests/config/test_cc_landscape.py b/tests/unittests/config/test_cc_landscape.py
> index 51fa39825..766932361 100644
> --- a/tests/unittests/config/test_cc_landscape.py
> +++ b/tests/unittests/config/test_cc_landscape.py
> @@ -173,7 +173,7 @@ class TestLandscape:
>                  [
>                      "landscape-config",
>                      "--silent",
> -                    '--computer-title="My" PC"',
> +                    '--computer-title="My\\" PC"',
>                      "--data-path=/var/lib/landscape/client",
>                      "--log-level=info",
>                      "--ping-url=http://landscape.canonical.com/ping",
> ```
> 
> Note that:
> 
> ```python
> >>> subprocess.run('echo \"', shell=True).stdout
> /bin/sh: 1: Syntax error: Unterminated quoted string
> 
> >>> subprocess.run('echo \\"', shell=True).stdout
> "
> ```



Thanks @aciba for the continued review and suggestions. I realized that trying to format cmdline params as a simple combined param of the fomat --k="v" and passing that into subp was adding the complexity of all the escaping we were doing. I've decided in steam to just pass k, v as separate params which allows subp to handle the proper escaping of the simple values as necessary. This way cc_landscape doesn't need to inspect the individual values directly. We can just flatten the k, v, k2, v2 into a single list

blackboxsw:
@aciba90 I additionally migrated to `cloud.distro.manage_service` instead of directly subp(["service", "restart", "landscape-client"]) since we've now grown that API on distro

blackboxsw:
> @aciba90 I additionally migrated to `cloud.distro.manage_service` instead of directly subp(["service", "restart", "landscape-client"]) since we've now grown that API on distro

I've actually noticed we can drop the landscape-client restart because we are calling `landscape-config` which starts the services for us with the updated config values. No need to restart anymore

aciba90 (approved):
+1. Thanks.

PR #4237, opened 2023-07-10
Overhaul module reference page

s-makin:
```
Module reference page was previously long lists of descriptions that were hard to consume.
- Add tabs to split up sections within module reference
- Tab syncing is an option, but may not be helpful here so left it off by default
- Subheadings can't be used within tabs, so changed headers in cc_ssh module description
```

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -88,7 +88,8 @@ SCHEMA_DOC_TMPL = """
 SCHEMA_PROPERTY_HEADER = ""
 SCHEMA_PROPERTY_TMPL = "{prefix}* **{prop_name}:** ({prop_type}){description}"
 SCHEMA_LIST_ITEM_TMPL = (
-   "{prefix}* Each object in **{prop_name}** list supports the following keys:"
+    "{prefix}* Each object in **{prop_name}** list supports "
+    "the following keys:"
 )
 SCHEMA_EXAMPLES_HEADER = ""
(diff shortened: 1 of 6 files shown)


igalic:
is there a way to preview this?

cloudinit/config/schema.py
```diff
@@ -170,11 +182,11 @@ def handle_problems(problems, prefix):
 
         self.schema_errors = handle_problems(
             schema_errors,
-            prefix="Cloud config schema errors: ",
+            prefix=f"Cloud config schema errors: ",
         )
         self.schema_deprecations = handle_problems(
             schema_deprecations,
-            prefix="Cloud config schema deprecations: ",
+            prefix=f"Cloud config schema deprecations: ",
```
TheRealFalcon, line 189: f-string isn't needed here

TheRealFalcon (changes requested):
Fancy! There's a few small inline issues, but otherwise LGTM.

Tests need to be updated. I'm assuming you want some help with that?

s-makin:
> Tests need to be updated. I'm assuming you want some help with that?

Thanks for the review, I'll fix the nits - I made a bunch of experimental changes in trying to solve the "indentation" errors, and definitely forgot to undo those f-strings. 

Yes please on the help with tests!

TheRealFalcon:
@s-makin , I rebased the branch and fixed tests. I took the liberty of pushing directly to your branch. If that's a problem, I can always undo it. If any other CI fails I can take a look.

PR #4249, opened 2023-07-13
testing: Clear all LRU caches after each test

TheRealFalcon:
## Proposed Commit Message

```
testing: Clear all LRU caches after each test

Some unit tests were missing necessary mocks because a previous test
had mocked the behavior and mock stuck around because of an LRU cache.
This means that if tests got run out of a specific order, they would
fail due to missing mocks. Because of this, a new autouse fixture was
added to automatically clear any lru cache after every test.

Additionally, updated unit tests accordingly so that they have the
necessary mocks, and convert `test_create_users.py` to pytest as
I was getting some transient errors from it.
```

## Additional Context
Been hidden forever, but started surfacing in #4237

### DIFF
--- a/tests/unittests/distros/test_create_users.py
+++ b/tests/unittests/distros/test_create_users.py
@@ -1,342 +1,294 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import re
+from typing import List
 
-from cloudinit import distros, log, ssh_util
-from tests.unittests.helpers import CiTestCase, mock
+import pytest
+
+from cloudinit import distros, ssh_util
+from tests.unittests.helpers import mock
 from tests.unittests.util import abstract_to_concrete
 
+USER = "foo_user"
 
-@mock.patch("cloudinit.distros.util.system_is_snappy", return_value=False)
+
+@pytest.fixture(autouse=True)
+def common_mocks(mocker):
+    mocker.patch("cloudinit.distros.util.system_is_snappy", return_value=False)
+
+
+def _useradd2call(args: List[str]):
+    # return a mock call for the useradd command in args
+    # with expected 'logstring'.
+    args = ["useradd"] + args
+    logcmd = list(args)
+    for i in range(len(args)):
+        if args[i] in ("--password",):
+            logcmd[i + 1] = "REDACTED"
+    return mock.call(args, logstring=logcmd)
+
+
 @mock.patch("cloudinit.distros.subp.subp")
-class TestCreateUser(CiTestCase):
-
-    with_logs = True
-
-    def setUp(self):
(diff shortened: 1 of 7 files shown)


tests/unittests/early_patches.py
```diff
@@ -0,0 +1,20 @@
+import functools
+
+old_lru_cache = functools.lru_cache
+cached_functions = []
+
+
+def wrapped_lru_cache(*args, **kwargs):
+    def wrapper(func):
+        new_func = old_lru_cache(*args, **kwargs)(func)
+        cached_functions.append(new_func)
```
blackboxsw, line 10: nifty registration of `lru_cache` callers.... Not critical, but do we need to check for duplicates before appending. running unittests and dumping cached_functions to a file we get to 47 cached_functions.... yet I only see 17 functions decorated with @lru_cache. Is this because we are importing lru_cached functions in other modules maybe?

tests/unittests/distros/test_create_users.py
```diff
@@ -1,342 +1,294 @@
...
-from cloudinit import distros, log, ssh_util
-from tests.unittests.helpers import CiTestCase, mock
-from tests.unittests.util import abstract_to_concrete
+from typing import List
 
+import pytest
 
-@mock.patch("cloudinit.distros.util.system_is_snappy", return_value=False)
-@mock.patch("cloudinit.distros.subp.subp")
-class TestCreateUser(CiTestCase):
```
blackboxsw, line 12: Thank you for for refactoring away from CiTestCase. Yay!! An angel just got its wings.

tests/unittests/distros/test_create_users.py
```diff
@@ -1,342 +1,294 @@
...
-        password = "passfoo"
-        self.dist.create_user(user, passwd=password)
-        self.assertEqual(
-            m_subp.call_args_list,
-            [
-                self._useradd2call([user, "--password", password, "-m"]),
-                mock.call(["passwd", "-l", user]),
-            ],
-        )
+    @pytest.mark.parametrize(
```
blackboxsw, line 38: Excellent refactor. thank you pytest param

blackboxsw (approved):
great improvement. +1 generally one comment on conftest.py that may need clarification, otherwise +1

tests/unittests/early_patches.py
TheRealFalcon: Good catch. We don't need to check for duplicates as imports only happen once, BUT we were storing any import from anywhere that had an `@lru_cache`, including the stdlib. By the end of tests the number of functions we were storing was over 1000! With my last push, it's now 18 the entire test run.

Issue #4243, opened 2023-07-12
Don't change log permissions if they are already more restrictive


TheRealFalcon:
Because #4093, we modified the cloud-init log permissions to always be 640 to prevent logs with more open permissions from leaking sensitive information. This is undesireable for those that want stricter permissions, such as 600. We should enforce 640 or stricter, but if the file is already stricter, don't change back to 640.

holmanb:
> We should enforce 640 or stricter, but if the file is already stricter, don't change back to 640.


Due to a buggy implementation in #4250, we do not enforce 640 or stricter. Reopening.

PR #4250, opened 2023-07-14
logging: keep current file mode of log file if its stricter than the new mode

ani-sinha:
By default, the cloud init log file is created with mode 0o644 with
`preserve_mode` parameter of `write_file()` set to False. This means that when
an existing log file is found, its mode will be unconditionally reset to the
mode 0o644. It is possible that this might cause the change of the mode of the
log file from the current more stricter mode to a less strict mode
(when the new mode 0o644 is less strict than the existing mode of the file).

In order to mitigate the above issue, check the current mode of the log file
and if the current mode is stricter than the default new mode 0o644, then
preserve the current mode of the file.

Fixes GH-4243

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2129,6 +2129,29 @@ def safe_int(possible_int):
         return None
 
 
+def compare_permission(mode1, mode2):
+    """Compare two file modes in octal.
+
+    If mode1 is less restrictive than mode2 return 1
+    If mode1 is more restrictive than mode2 return -1
+    If mode1 is same as mode2, return 0
+
+    The comparison starts from the permission of the
+    set of users in "others" and then works up to the
+    permission of "user" set.
+    """
+    # Convert modes to octal and reverse the last 3 digits
+    # so 0o640 would be become 0o046
+    mode1_oct = oct(mode1)[2:].rjust(3, "0")
+    mode2_oct = oct(mode2)[2:].rjust(3, "0")
+    m1 = int(mode1_oct[:-3] + mode1_oct[-3:][::-1], 8)
+    m2 = int(mode2_oct[:-3] + mode2_oct[-3:][::-1], 8)
+
+    # Then do a traditional cmp()
+    # https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons
+    return (m1 > m2) - (m1 < m2)
+
+
 def chmod(path, mode):
     real_mode = safe_int(mode)
     if path and real_mode:
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -217,12 +217,25 @@ class Init:
         self._initialize_filesystem()
 
     def _initialize_filesystem(self):
+        mode = 0o640
+        fmode = None
+
         util.ensure_dirs(self._initial_subdirs())
         log_file = util.get_cfg_option_str(self.cfg, "def_log_file")
         if log_file:
             # At this point the log file should have already been created
             # in the setupLogging function of log.py
-            util.ensure_file(log_file, mode=0o640, preserve_mode=False)
+
+            try:
+                fmode = util.get_permissions(log_file)
+            except OSError:
+                pass
+
+            # if existing file mode fmode is stricter, do not change it.
+            if fmode and util.compare_permission(fmode, mode) < 0:
+                mode = fmode
+
+            util.ensure_file(log_file, mode, preserve_mode=False)
             perms = self.cfg.get("syslog_fix_perms")
             if not perms:
                 perms = {}
--- a/tests/unittests/test_stages.py
+++ b/tests/unittests/test_stages.py
@@ -606,13 +606,22 @@ class TestInit_InitializeFilesystem:
         # Assert we create it 0o640  by default if it doesn't already exist
         assert 0o640 == stat.S_IMODE(log_file.stat().mode)
 
-    def test_existing_file_permissions(self, init, tmpdir):
+    @pytest.mark.parametrize(
+        "set_perms,expected_perms",
+        [
+            (0o640, 0o640),
+            (0o606, 0o640),
+            (0o600, 0o600),
+        ],
+    )
+    def test_existing_file_permissions(
+        self, init, tmpdir, set_perms, expected_perms
+    ):
         """Test file permissions are set as expected.
 
-        CIS Hardening requires 640 permissions. These permissions are
-        currently hardcoded on every boot, but if there's ever a reason
-        to change this, we need to then ensure that they
-        are *not* set every boot.
+        CIS Hardening requires 640 permissions. If the file has looser
+        permissions, then hard code 640. If the file has tighter
+        permissions, then leave them as they are
 
         See https://bugs.launchpad.net/cloud-init/+bug/1900837.
         """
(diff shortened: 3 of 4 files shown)


cloudinit/util.py
```diff
@@ -2225,11 +2260,18 @@ def write_file(
...
-        try:
-            mode = get_permissions(filename)
-        except OSError:
-            pass
+    try:
+        fmode = get_permissions(filename)
+    except OSError:
+        fmode = None
+
+    if preserve_mode or (
```
TheRealFalcon, line 2268: I don't think this is the right place to do this. `write_file` is a general purpose utility function, and there may be cases we want to write a new file with completely different (and possibly looser) permissions.

I think `_initialize_filesystem` in `stages.py` is probably the place to do this.

TheRealFalcon (changes requested):
Thanks for the PR, but I think we need a slightly different approach. See the inline comment for more details.

cloudinit/util.py
ani-sinha: @TheRealFalcon fixed in v2.

TheRealFalcon (changes requested):
The new changes look good, thanks!

I noticed a few issues with the `compare_permission` function. Directory comparisons won't work. Also, permissions with the first digit being 0 won't work. E.g.,
```
>>> oct(0o007)
'0o7'
```
I updated the function to take these into account and is hopefully a little simpler. I also added/updated some unit tests, and they also work against your implementation (minus the 0o007 issue I mentioned earlier). Let me know if this works for you.

```diff
diff --git a/cloudinit/util.py b/cloudinit/util.py
index 73d7f8b92..6d44c307e 100644
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2130,36 +2130,26 @@ def safe_int(possible_int):
 
 
 def compare_permission(mode1, mode2):
-    """
-    Compare two file modes in octal.
+    """Compare two file modes in octal.
+
     If mode1 is less restrictive than mode2 return 1
     If mode1 is more restrictive than mode2 return -1
     If mode1 is same as mode2, return 0
-
+    
     The comparison starts from the permission of the
     set of users in "others" and then works up to the
     permission of "user" set.
-
     """
-
-    m1 = str(oct(mode1))[2:]  # remove "0o" octal prefix
-    m2 = str(oct(mode2))[2:]
-
-    # start the comparison from permission of "others" because its the widest
-    # set of users. Then work up to "groups" and "user" sets.
-    # For example, between 407 and 600, 407 is less restrictive because
-    # even through user has only read permission, others have all three,
-    # read/write and executable permission and those set of users covers
-    # the widest group.
-    if int(m2[2]) == int(m1[2]):
-        if int(m2[1]) == int(m1[1]):
-            ret = int(m1[0]) - int(m2[0])
-        else:
-            ret = int(m1[1]) - int(m2[1])
-    else:
-        ret = int(m1[2]) - int(m2[2])
-
-    return -1 if ret < 0 else 1 if ret > 0 else 0
+    # Convert modes to octal and reverse the last 3 digits
+    # so 0o640 would be become 0o046
+    mode1_oct = oct(mode1)[2:].rjust(3, "0")
+    mode2_oct = oct(mode2)[2:].rjust(3, "0")
+    m1 = int(mode1_oct[:-3] + mode1_oct[-3:][::-1], 8)
+    m2 = int(mode2_oct[:-3] + mode2_oct[-3:][::-1], 8)
+
+    # Then do a traditional cmp()
+    # https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons
+    return (m1 > m2) - (m1 < m2)
 
 
 def chmod(path, mode):
diff --git a/tests/unittests/test_stages.py b/tests/unittests/test_stages.py
index a61f9df9b..831ea9f2a 100644
--- a/tests/unittests/test_stages.py
+++ b/tests/unittests/test_stages.py
@@ -606,13 +606,22 @@ class TestInit_InitializeFilesystem:
         # Assert we create it 0o640  by default if it doesn't already exist
         assert 0o640 == stat.S_IMODE(log_file.stat().mode)
 
-    def test_existing_file_permissions(self, init, tmpdir):
+    @pytest.mark.parametrize(
+        "set_perms,expected_perms",
+        [
+            (0o640, 0o640),
+            (0o606, 0o640),
+            (0o600, 0o600),
+        ],
+    )
+    def test_existing_file_permissions(
+        self, init, tmpdir, set_perms, expected_perms
+    ):
         """Test file permissions are set as expected.
 
-        CIS Hardening requires 640 permissions. These permissions are
-        currently hardcoded on every boot, but if there's ever a reason
-        to change this, we need to then ensure that they
-        are *not* set every boot.
+        CIS Hardening requires 640 permissions. If the file has looser
+        permissions, then hard code 640. If the file has tighter
+        permissions, then leave them as they are
 
         See https://bugs.launchpad.net/cloud-init/+bug/1900837.
         """
@@ -620,9 +629,9 @@ class TestInit_InitializeFilesystem:
         log_file.ensure()
         # Use a mode that will never be made the default so this test will
         # always be valid
-        log_file.chmod(0o606)
+        log_file.chmod(set_perms)
         init._cfg = {"def_log_file": str(log_file)}
 
         init._initialize_filesystem()
 
-        assert 0o640 == stat.S_IMODE(log_file.stat().mode)
+        assert expected_perms == stat.S_IMODE(log_file.stat().mode)
diff --git a/tests/unittests/test_util.py b/tests/unittests/test_util.py
index bc7e3d412..a1291d38b 100644
--- a/tests/unittests/test_util.py
+++ b/tests/unittests/test_util.py
@@ -3113,3 +3113,27 @@ class TestHashBuffer:
                 util.hash_buffer(f)
                 == b"\x99\x80\x0b\x85\xd38>:/\xb4^\xb7\xd0\x06jHy\xa9\xda\xd0"
             )
+
+
+class TestComparePermissions:
+    @pytest.mark.parametrize(
+        "perm1,perm2,expected",
+        [
+            (0o777, 0o777, 0),
+            (0o000, 0o000, 0),
+            (0o421, 0o421, 0),
+            (0o1640, 0o1640, 0),
+            (0o1407, 0o1600, 1),
+            (0o1600, 0o1407, -1),
+            (0o407, 0o600, 1),
+            (0o600, 0o407, -1),
+            (0o007, 0o700, 1),
+            (0o700, 0o007, -1),
+            (0o077, 0o100, 1),
+            (0o644, 0o640, 1),
+            (0o640, 0o600, 1),
+            (0o600, 0o400, 1),
+        ],
+    )
+    def test_compare_permissions(self, perm1, perm2, expected):
+        assert util.compare_permission(perm1, perm2) == expected

```

ani-sinha:
> The new changes look good, thanks!
> 
> I noticed a few issues with the `compare_permission` function. Directory comparisons won't work. Also, permissions with the first digit being 0 won't work. E.g.,
> 
> ```
> >>> oct(0o007)
> '0o7'
> ```
> 
> I updated the function to take these into account and is hopefully a little simpler. I also added/updated some unit tests, and they also work against your implementation (minus the 0o007 issue I mentioned earlier). Let me know if this works for you.


Thanks! It works and make the function lot simpler. I pushed the changes.

TheRealFalcon (approved):
LGTM, thanks!

PR #4256, opened 2023-07-17
lxd: install lxd from snap, not deb if absent in image

blackboxsw:
## Proposed Commit Message

```
When lxd: cloud-config is provided and lxd command is not present on the image, install lxd from snap instead of deb.

On Ubuntu Focal and later LXD is available as a snap, the transitional lxd deb package was no longer available for install as a deb package as of 22.04 (Jammy) and later.

All ubuntu cloud images contain LXD as snap within the base image manifest. So, this logic generally should not get triggered. But, errors in cloud images build process generated some images without LXD and that triggered the install_packages logic to fail in cloud-init integration tests.
```

## Additional Context

Failing Jenkins builds of ec2 images which temporarily lacked lxd in snap seed.
```
01:57:46   '2023-07-15 06:54:44,336 - cc_lxd.py[WARNING]: failed to install packages '
01:57:46   "['lxd']: Unexpected error while running command.\n"
```
- https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-mantic-azure/31/testReport/junit/tests.integration_tests.modules.test_lxd/TestLxdBridge/test_bridge/

- https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-mantic-azure/31/testReport/junit/tests.integration_tests.modules.test_lxd/TestLxdBridge/test_binaries_installed_lxc_/

## Test Steps

```
CLOUD_INIT_OS_IMAGE=mantic CLOUD_INIT_PLATFORM=ec2 tox -e integration-tests tests/integration_tests/modules/test_lxd.py
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -216,6 +216,13 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     bridge_cfg = lxd_cfg.get("bridge", {})
     supplemental_schema_validation(init_cfg, bridge_cfg, preseed_str)
 
+    if not subp.which("lxd"):
+        try:
+            subp.subp(["snap", "install", "lxd"])
+        except subp.ProcessExecutionError as e:
+            raise RuntimeError(
+                "Failed to install lxd from snap: %s" % e
+            ) from e
     packages = get_required_packages(init_cfg, preseed_str)
     if len(packages):
         try:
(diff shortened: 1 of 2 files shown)


holmanb (commented):
Doesn't this represent a change of behavior in Focal? Jammy current behavior is broken, so there is no problem there.

I recognize that this is an uncommon codepath given cloud image build pipeline, but some users may build their own images which would hit this change in behavior.

blackboxsw:
> Doesn't this represent a change of behavior in Focal? Jammy current behavior is broken, so there is no problem there.
> 
> I recognize that this is an uncommon codepath given cloud image build pipeline, but some users may build their own images which would hit this change in behavior.

Thanks for raising this. it does not represent change in behavior because the lxd deb on focal was only a transitional deb package which also installed the snap under the hood in it's deb presinst script. So we can safely install the snap everywhere we support ubuntu (focal++).

```
root@t-f:~# lsb_release -sc
focal
root@t-f:~# apt-cache show lxd
Package: lxd
Architecture: all
Version: 1:0.10
Priority: optional
....
Description-en: Transitional package - lxd -> snap (lxd)
 This is a transitional dummy package. It can safely be removed.
 .
 lxd is now replaced by the LXD snap.
...
```

PR #4258, opened 2023-07-18
Remove default membership in redundant groups

waveform80:
## Proposed Commit Message
```
Remove default membership in redundant groups

The cloud-init template for Ubuntu currently includes membership in
numerous groups. Many of these are groups that were removed from the
default list over a decade ago on the desktop, and which are better
served by udev rules today. Specifically: audio, dialout, floppy,
netdev, plugdev, video.

LP: #1923363
```

## Additional Context
[LP: #1923363](https://bugs.launchpad.net/ubuntu/mantic/+source/user-setup/+bug/1923363)

## Checklist:
 - [x] I have updated or added any unit tests accordingly (n/a, no tests failed)

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -206,7 +206,7 @@ system_info:
      name: ubuntu
      lock_passwd: True
      gecos: Ubuntu
-     groups: [adm, audio, cdrom, dialout, dip, floppy, lxd, netdev, plugdev, sudo, video]
+     groups: [adm, cdrom, dip, lxd, sudo]
      sudo: ["ALL=(ALL) NOPASSWD:ALL"]
      shell: /bin/bash
 {# SRU_BLOCKER: do not ship network renderers on Xenial, Bionic or Eoan #}
--- a/doc/examples/cloud-config-user-groups.txt
+++ b/doc/examples/cloud-config-user-groups.txt
@@ -146,4 +146,4 @@ users:
 #     shell: /bin/bash
 #     lock_passwd: True
 #     gecos: Ubuntu
-#     groups: [adm, audio, cdrom, dialout, floppy, video, plugdev, dip, netdev]
+#     groups: [adm, cdrom, dip, lxd, sudo]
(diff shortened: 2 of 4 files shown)


aciba90 (approved):
LGTM. Thanks!

Open question: Do we want to gate this change for `<=mantic`, to minimize potential breaks?

holmanb:
> Open question: Do we want to gate this change for `<=mantic`, to minimize potential breaks?

This represents a change in behavior so we shouldn't include this in old Ubuntu releases since it could break users that are expecting the default user to be a member of these groups.

holmanb:
> This represents a change in behavior so we shouldn't include this in old Ubuntu releases since it could break users that are expecting the default user to be a member of these groups.

I think we would need to include the following patch in non-mantic releases:
```
diff --git a/config/cloud.cfg.tmpl b/config/cloud.cfg.tmpl
index e496c62c3..a4a4671a6 100644
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -206,7 +206,7 @@ system_info:
      name: ubuntu
      lock_passwd: True
      gecos: Ubuntu
-     groups: [adm, cdrom, dip, lxd, sudo]
+     groups: [adm, audio, cdrom, dialout, dip, floppy, lxd, netdev, plugdev, sudo, video]
      sudo: ["ALL=(ALL) NOPASSWD:ALL"]
      shell: /bin/bash
 {# SRU_BLOCKER: do not ship network renderers on Xenial, Bionic or Eoan #}
```

waveform80:
Agreed: I wouldn't suggest shipping such changes pre-mantic. I've checked through the udev rules and I'm *reasonably* confident that the removal of these groups shouldn't result in any loss of functionality in mantic, but I can't say the same for jammy (largely because I haven't checked; not sure how much churn there is in the udev rules though).

One notable group I've left alone for now, although it *may* be redundant as well, is the `cdrom` group. This still has several udev rules that rely on membership in this group instead of, say, using the uaccess tag (see `/lib/udev/rules.d/50-udev-default.rules` for examples). This group is also still present on the desktop defaults. Still, it should probably be evaluated for removal at some future point.

PR #4265, opened 2023-07-20
[jammy] patch: Retain old groups in default user template

holmanb:
```
patch: Retain old groups in default user template
```

## Additional Context
I just merged https://github.com/canonical/cloud-init/pull/4258, for which this patch retains old behavior.

https://bugs.launchpad.net/ubuntu/+source/ubiquity/+bug/1923363

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (23.2-0ubuntu0~22.04.1ubuntu2) UNRELEASED; urgency=medium
+
+  * d/patches/retain-old-groups.patch:
+    - Retain original groups in default group for ubuntu.
+      Mantic removed unneeded groups.
+
+ -- Brett Holman <brett.holman@canonical.com>  Thu, 20 Jul 2023 15:07:36 -0600
+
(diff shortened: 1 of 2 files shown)


TheRealFalcon (changes requested):
Don't forget the d/changelog entry

holmanb:
@TheRealFalcon Thanks for the review. I'll open against lunar and focal once this is approved.

TheRealFalcon:
Sorry, didn't really look super thoroughly the last time.

I think we need an upstream snapshot first. How can we have a patch against code that doesn't actually exist yet in the branch? After that, we need to generate a quilt patch, not just a diff. It needs a DEP-3 header and a reference in `d/patches/series`.

See https://github.com/canonical/uss-tableflip/blob/main/doc/ubuntu_release_process.md#adding-a-quilt-patch-to-debianpatches (or https://wiki.debian.org/UsingQuilt).

TheRealFalcon:
If #4286 merges we can close this one

TheRealFalcon:
Closing as it was incorporated in #4286 .

PR #4266, opened 2023-07-20
[bionic] patch: Retain old groups in default user template

holmanb:
```
patch: Retain old groups in default user template
```

## Additional Context
I just merged https://github.com/canonical/cloud-init/pull/4258, for which this patch retains old behavior.

https://bugs.launchpad.net/ubuntu/+source/ubiquity/+bug/1923363


TheRealFalcon:
We're no longer SRUing to Bionic, so this shouldn't be necessary. We do Focal and Lunar though.

PR #4272, opened 2023-07-21
cloud.cfg.tmpl: reorganise, minimise/reduce duplication

dermotbradley:
## Proposed Commit Message

```
cloud.cfg.tmpl: reorganise, minimise/reduce duplication

Simplify the cloud.cfg.tmpl file. There is a lot of duplication (e.g.
the same sudo rule specified multiple times).

This also addresses #4267 and aligns Debian-specific template
configuration with Debian's own packaged cloud.cfg, located here:
https://salsa.debian.org/cloud-team/cloud-init/-/blob/master/debian/cloud.cfg

Fixes GH-4267
```

## Checklist:

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -3,28 +3,51 @@
 # and base configuration.
 {% set is_bsd = variant in ["dragonfly", "freebsd", "netbsd", "openbsd"] %}
 {% set is_rhel = variant in ["almalinux", "centos", "cloudlinux", "eurolinux",
-                             "miraclelinux", "rhel", "rocky", "virtuozzo" ] %}
+                             "miraclelinux", "rhel", "rocky", "virtuozzo"] %}
+{% set gecos = ({"amazon": "EC2 Default User", "centos": "Cloud User",
+                 "debian": "Debian", "dragonfly": "DragonFly",
+                 "freebsd": "FreeBSD", "mariner": "MarinerOS",
+                 "rhel": "Cloud User", "netbsd": "NetBSD",
+                 "openbsd": "openBSD", "openmandriva": "OpenMandriva admin",
+                 "photon": "PhotonOS", "ubuntu": "Ubuntu",
+                 "unknown": "Ubuntu"}) %}
+{% set groups = ({"alpine": "adm, wheel", "arch": "wheel, users",
+                  "debian": "adm, audio, cdrom, dialout, dip, floppy, netdev, plugdev, sudo, video",
+                  "gentoo": "users, wheel", "mariner": "wheel",
+                  "photon": "wheel",
+                  "openmandriva": "wheel, users, systemd-journal",
+                  "suse": "cdrom, users",
+                  "ubuntu": "adm, cdrom, dip, lxd, sudo",
+                  "unknown": "adm, cdrom, dip, lxd, sudo"}) %}
+{% set shells = ({"alpine": "/bin/ash", "dragonfly": "/bin/sh",
+                  "freebsd": "/bin/tcsh", "netbsd": "/bin/sh",
+                  "openbsd": "/bin/ksh"}) %}
+{% set usernames = ({"amazon": "ec2-user", "centos": "cloud-user",
+                     "openmandriva": "omv", "rhel": "cloud-user",
+                     "unknown": "ubuntu"}) %}
 {% if is_bsd %}
 syslog_fix_perms: root:wheel
-{% elif variant in ["suse"] %}
+{% elif variant == "suse" %}
 syslog_fix_perms: root:root
 {% endif %}
+
 # A set of users which may be applied and/or used by various modules
 # when a 'default' entry is found it will reference the 'default_user'
 # from the distro configuration specified below
 users:
-{% if variant in ["photon"] %}
-   - name: root
-     lock_passwd: false
+{% if variant == "photon" %}
+  - name: root
+    lock_passwd: false
 {% else %}
-   - default
+  - default
 {% endif %}
 
-{% if variant in ["photon"] %}
+{% if variant == "photon" %}
 # VMware guest customization.
 disable_vmware_customization: true
 manage_etc_hosts: false
-{% endif %}
+
+{% endif -%}
 
 # If this is set, 'root' will not be able to ssh in and they
 # will get a message to login instead as the default $user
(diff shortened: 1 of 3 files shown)


config/cloud.cfg.tmpl
```diff
@@ -191,34 +192,155 @@ cloud_final_modules:
...
+{% if variant == "alpine" %}
+     shell: /bin/ash
+{% elif variant in ["amazon", "arch", "debian", "fedora", "gentoo", "mariner",
+                    "OpenCloudOS", "openEuler", "openmandriva", "photon",
+                    "suse", "TencentOS", "ubuntu", "unknown"] or is_rhel %}
+     shell: /bin/bash
+{% elif variant in ["dragonfly", "netbsd"] %}
+     shell: /bin/sh
+{% elif variant == "freebsd" %}
+     shell: /bin/tcsh
```
igalic, line 284: didn't realise this was our default shell.
probably because i usually change it to fish…
the "new" default shell for quite some time is "/bin/sh"

(dunno what it is in dragonfly, will try to investigate)

igalic (commented):
👀

config/cloud.cfg.tmpl
```diff
@@ -124,47 +133,44 @@ cloud_config_modules:
  - locale
 {% endif %}
  - set_passwords
+{% if variant == "alpine" %}
```
TheRealFalcon, line 136: Why'd you move this block? In general I don't think we should be reordering modules unless there's a really good reason.

TheRealFalcon (commented):
Good work here. I have a couple thoughts inline that I think we should figure out before we go any further.

config/cloud.cfg.tmpl
dermotbradley: Alphabetising the distro's package managers :-)

Obviously package managers are exclusive of each other (if, elif, elif etc) so reordering does matter.

config/cloud.cfg.tmpl
igalic: let's fold all the BSDs except OpenBSD into one case

TheRealFalcon (commented):
Bah...I had left another comment yesterday that seems to have disappeared. Sorry about that. It was the main thing I was wondering about so I recreated it here.

config/cloud.cfg.tmpl
```diff
@@ -191,34 +197,156 @@ cloud_final_modules:
...
    distro: ubuntu
 {% endif %}
-{% if variant in ["ubuntu", "unknown", "debian"] %}
+{% if variant in ["alpine", "amazon", "arch", "debian", "fedora", "gentoo",
+                  "OpenCloudOS", "openEuler", "openmandriva", "suse",
+                  "TencentOS", "ubuntu", "unknown"] or is_rhel %}
    # Default user name + that default users groups (if added/used)
+{% endif %}
    default_user:
+{% if variant == "amazon" %}
```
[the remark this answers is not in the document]
dermotbradley: Done for gecos, groups, shells and usernames.

dermotbradley:
I made some more changes including tidying up multiple blank lines, making YAML indentation uniform.

I used tools/render-cloudcfg to compare all the distros it supports "before" and "after" and things look good.

I notice that render-cloudcfg does handle some distros as valid parameters though. Should "dragonfly", "cos", and the various Suse variations be valid options for that script?

config/cloud.cfg.tmpl
dermotbradley: the gecos/groups/shells/usernames stuff has been reworked in general.

Is there some other BSD stuff you think should be changed?

dermotbradley:
BTW I happened to notice that for variant "amazon" there is no module enabled in cloud_config_modules section for (RPM) packages, e.g yum_add_repo.

This is not something I've broken, I checked the rendered version of the existing cloud.cfg.tmpl and it is the same situation.

I assume therefore that AWS must be modifying or replacing the cloud.cfg when they package cloud-init for AWS Linux.

TheRealFalcon:
This is looking really good. Thanks @dermotbradley 

> I notice that render-cloudcfg does not permit some distros as valid parameters though. Should "dragonfly", "cos", and the various Suse variations be valid options for that script?

Let's add dragonfly as that is specifically called out in the template, but I think we can leave the others.

I left a few small nits inline. Beyond that:

For Alpine, I noticed a few non-trivial changes. I'm assuming those are intentional? In particular, `groups` changed from `[adm, sudo]` to `[adm, wheel]`. Also, there's an added `eni` network renderer when there was none before.

For Debian, I don't think the added renderers/activators are correct. Bookworm cloud images are [shipping with netplan and systemd-networkd by default](https://blog.slyon.de/2023/07/10/netplan-and-systemd-networkd-on-debian-bookworm/). I would expect a list similar to Ubuntu.

TheRealFalcon (commented):
> I left a few small nits inline

Here's the inline comments.

dermotbradley:
> Let's add dragonfly as that is specifically called out in the template, but I think we can leave the others.

Ok, I'll do that.

> For Alpine, I noticed a few non-trivial changes. I'm assuming those are intentional? In particular, `groups` changed from `[adm, sudo]` to `[adm, wheel]`. Also, there's an added `eni` network renderer when there was none before.

Yes these are intentional.

> For Debian, I don't think the added renderers/activators are correct. Bookworm cloud images are [shipping with netplan and systemd-networkd by default](https://blog.slyon.de/2023/07/10/netplan-and-systemd-networkd-on-debian-bookworm/). I would expect a list similar to Ubuntu.

I wasn't sure what these should be and so assumed only 'eni' (for older Debian releases) and 'networkd' as newer releases seemed to use that. Didn't realise Debian used/supported netplan.

I'll change this as you indicated.

config/cloud.cfg.tmpl
```diff
@@ -45,340 +70,295 @@ resize_rootfs: noblock
...
- - scripts_per_instance
- - scripts_user
- - ssh_authkey_fingerprints
- - keys_to_console
- - install_hotplug
- - phone_home
- - final_message
- - power_state_change
+  - ubuntu_drivers
+{% endif %}
```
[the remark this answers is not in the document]
dermotbradley: Makes sense. Done.

config/cloud.cfg.tmpl
```diff
@@ -45,340 +70,295 @@ resize_rootfs: noblock
...
-     lock_passwd: true
-     gecos: Cloud User
-{% elif variant == "openmandriva" %}
-     name: omv
-     lock_passwd: True
-     gecos: OpenMandriva admin
+  # Unknown/fallback distro.
+  distro: ubuntu
+{% endif %}
+{% if variant in ["alpine", "amazon", "arch", "debian", "fedora", "gentoo",
```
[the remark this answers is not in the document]
dermotbradley: True, removed.

TheRealFalcon (approved):
LGTM!

Final thing is this branch needs a rebase. There's a conflict (I believe due to openeuler changing case).

igalic (approved):
👍

dermotbradley:
> Final thing is this branch needs a rebase. There's a conflict (I believe due to openeuler changing case).

Done.

TheRealFalcon:
Thanks!

PR #4273, opened 2023-07-21
Kernel Commandline: Fix 4271

holmanb:
```
Fix NoCloud kernel commandline key parsing

Currently >2 keys in NoCloud datasource are not
supported. Previously 250280 attempted to correct
no keys being supported, however that fix was only
partial since more than one key breaks datasource
detection.

Additionally add test coverage.

Fixes GH-4271
```

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1742,8 +1742,8 @@ read_config() {
         key=${tok%%=*}
         val=${tok#*=}
 
-        # discard anything after delimiter
-        val=${val%;*}
+        # discard anything after the first delimiter
+        val=${val%%;*}
         case "$key" in
             ds) _rc_dsname="$val";;
             ci.ds) _rc_dsname="$val";;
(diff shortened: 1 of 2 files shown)


tools/ds-identify
```diff
@@ -1742,8 +1742,8 @@ read_config() {
         key=${tok%%=*}
         val=${tok#*=}
 
-        # discard anything after delimiter
-        val=${val%;*}
+        # discard anything after the first delimiter
+        val=${val%%;*}
```
blackboxsw, line 1746: Ugh, excellent, and correct fix. bash variable truncation :/

blackboxsw (approved):
LGTM


```
$ cat  > 1.sh  <<EOF
val="ds=nocloud;s=http://my-url/;h=hostname"
echo -- PRIOR BEHAVIOR --
echo ${val%;*}
echo -- FIXED BEHAVIOR --
echo ${val%%;*}
EOF
$ sh ./1.sh
-- PRIOR BEHAVIOR --
ds=nocloud;s=http://my-url/
-- FIXED BEHAVIOR --
ds=nocloud

```

PR #4274, opened 2023-07-21
Don't require passlib dependency

holmanb:
```
Don't require passlib dependency

Thee crypt module suffices on old versions of Python.
Use it when available. 
```

TODO: tests

### DIFF
--- a/integration-requirements.txt
+++ b/integration-requirements.txt
@@ -10,3 +10,4 @@ pycloudlib==1!3.0.2
 pytest<=7.3.1
 
 packaging
+passlib
--- a/cloudinit/distros/netbsd.py
+++ b/cloudinit/distros/netbsd.py
@@ -2,14 +2,37 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
+import functools
 import os
 import platform
-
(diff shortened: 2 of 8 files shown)


TheRealFalcon:
I know this is still WIP, but we'll want to move `passlib` from `requirements.txt` to `test-requirements.txt`, right?

TheRealFalcon:
Looking at the test failure, looks like the netbsd code never worked on <= 3.6, so looks like the test will need a version gate now.

integration-requirements.txt
```diff
@@ -10,3 +10,4 @@ pycloudlib==1!3.0.2
 pytest<=7.3.1
 
 packaging
+passlib
```
TheRealFalcon, line 13: Nit: Is this actually needed? AFAIK, integration requirements get installed on the host system, not on the target where tests actually run.

TheRealFalcon (approved):
Nit inline but LGTM.

integration-requirements.txt
holmanb: > Is this actually needed?

Yes - `tests/integration_tests/bugs/test_gh671.py` imports it on the host system.

PR #4276, opened 2023-07-21
ec2: Support double encoded userdata

nmeyerhans:
## Proposed Commit Message

```
ec2: Support double encoded userdata

The Amazon APIs expect userdata to be base64 encoded when passed as
input to e.g. RunInstances.  A number of tools, including the AWS CLI,
perform this base64 encoding implicitly, but it's common for users to
base64 encode the data prior to passing it to them. This results in two
layers of base64 encoding and effectively results in a failed EC2
launch.  This change adds the ability to decode the redundant layer of 
encoding.

Fixes https://github.com/amazonlinux/amazon-linux-2023/issues/401
```

## Additional Context

Amazon Linux has long carried a patch to cloud-init to handle this case of double base64 encoding.  This patch broke when it was forward ported to python3 and a more recent cloud-init during the Amazon Linux 2023 development.  By submitting the fixed version here, we hope to remove this custom patch and also make this functionality work for users of other distros.

## Test Steps

Start with any valid cloud-config.  This can be passed to the AWS CLI, which will take care of base64 encoding.  This works as expected.

Base64 encode the userdata, and pass the encoded results to the AWS CLI, which will re-encode it with a second layer of base64 encoding.  Example:

```plain
$ cat double-base64-test.yaml                                      
#cloud-config
write_files:
  - content: |
      This is some content
    path: /test-file

$ base64 -w0 double-base64-test.yaml | tee double-base64-test.yaml.b64
I2Nsb3VkLWNvbmZpZwp3cml0ZV9maWxlczoKICAtIGNvbnRlbnQ6IHwKICAgICAgVGhpcyBpcyBzb21lIGNvbnRlbnQKICAgIHBhdGg6IC90ZXN0LWZpbG

$ aws ec2 run-instances --user-data file://double-base64-test.yaml.b64 ...
```

Launch an instance with the base64 encoded data.  Without this change, cloud-init fails to process userdata and records  a message similar to the following in `cloud-init.log` :

```plain
2023-07-21 21:09:00,336 - __init__.py[WARNING]: Unhandled non-multipart (text/x-not-multipart) userdata: 'b'I2Nsb3VkLWNvbmZpZwp3cml0'...'
```

With this change, the userdata is decoded and executed as intended.

Tested on Amazon Linux 2023 and Debian 12 and unstable.

## Checklist:

Note that AWS has already signed the CLA on behalf of employees, which should cover this contribution. I haven't added my name to `tools/.github-cla-signers`, but can do so if you need it.


TheRealFalcon (commented):
Changes here LGTM! Thanks for the contribution.

> I haven't added my name to tools/.github-cla-signers, but can do so if you need it

Yes please. We'll need that along with fixes to the failing lint job in order for this PR to pass CI. Once those two things happen, we can merge this.

nmeyerhans:
> Changes here LGTM! Thanks for the contribution.
> 
> > I haven't added my name to tools/.github-cla-signers, but can do so if you need it
> 
> Yes please. We'll need that along with fixes to the failing lint job in order for this PR to pass CI. Once those two things happen, we can merge this.

Thanks, updated.

TheRealFalcon (approved):
Thanks!

PR #4279, opened 2023-07-22
Ubuntu/jammy

blackboxsw:
Fix daily jammy build recipes due to feature drift in unit tests
do not squash merge. push separate commits to upstream/ubuntu/jammy  when reviewed

## Additional Context
Daily recipe builds failing on jammy  due to quilt apply debian/patch/series
https://launchpadlibrarian.net/678153853/buildlog_ubuntu-jammy-amd64.cloud-init_23.2.daily-202307201631-392346cc~ubuntu22.04.1_BUILDING.txt.gz

## Test Steps
```
git checkout main
git merge upstream/ubuntu/jammy
quilt push -a
tox -e py3 # expect failures in test_data and test_features for the features that are set False ROOT_READ_ONLY and HASHED_USERS
quilt pop -a


# See fix
git reset --hard upstream/main
git merge blackboxsw/ubuntu/jammy
quilt push -a 
tox -e py3 # expect success
quilt pop -a
```

## Checklist:

 - [ ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #4285, opened 2023-07-24
Ensure get_features() grabs all features

TheRealFalcon:
## Proposed Commit Message

```
Ensure get_features() grabs all features

It was previously ignoring keys with numbers in them.

Additionally, update tests so that they don't have to be patched
downstream any time we add a new feature flag. Since the whole point
of the feature flags is to create overrides, that seems like a lot
of needless churn.
```

## Additional Context
Will replace https://github.com/canonical/cloud-init/pull/4279 if accepted.

I more-or-less duplicated the mocking code and comment. If we find we need it in more places, it should probably be refactored into a common fixture then, but for now I'm ok with 2.


holmanb (approved):
LGTM!

PR #4292, opened 2023-07-25
fixes 4288

holmanb:
```
Fix reference before assignment

Fixes GH-4288 
```

### DIFF
--- a/tests/unittests/test_importer.py
+++ b/tests/unittests/test_importer.py
@@ -1,0 +1,24 @@
+import pytest
+
+from cloudinit.importer import match_case_insensitive_module_name
+
+
+@pytest.mark.parametrize(
+    "m_name,m_match",
+    (
+        pytest.param(
+            "nocloud-net",
+            "DataSourceNoCloud",
+            id="nocloud-net is a special case, make sure it works",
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
Looks good. Can we add a few unit tests checking nocloud-net, nocloud, and DataSource<Somethingelse> at the very least?

tests/unittests/test_importer.py
```diff
@@ -0,0 +1,24 @@
...
+from cloudinit.importer import match_case_insensitive_module_name
+
+
+@pytest.mark.parametrize(
+    "m_name,m_match",
+    (
+        pytest.param(
+            "nocloud-net",
+            "DataSourceNoCloud",
+            id="nocloud-net is a special case, make sure it works",
```
TheRealFalcon, line 12: Not something that needs to change here, but beyond explaining the test case, these id's are convenient for specifying a particular test when running on the cli.

`pytest "tests/unittests/test_importer.py::test_importer[nocloud-net is a special case, make sure it works]"` works but could probably be more concise.

TheRealFalcon (approved):
LGTM!

Issue #4048, opened 2023-05-12
writing of sudoers is not idempotent


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1998539](https://bugs.launchpad.net/cloud-init/+bug/1998539)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2022-12-01T22:07:21.964231+00:00
date_fix_committed = None
date_fix_released = None
id = 1998539
importance = undecided
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1998539
milestone = None
owner = minagalic
owner_name = Mina Galić
private = False
status = triaged
submitter = minagalic
submitter_name = Mina Galić
tags = ['bitesize']
duplicates = []
</pre>
</details>

_Launchpad user **Mina Galić(minagalic)** wrote on 2022-12-01T22:07:21.964231+00:00_

after several (full) re-runs of cloud-init, my /usr/local/etc/sudoers.d/90-cloud-init-users file looks like this:

# Created by cloud-init v. 22.3 on Wed, 05 Oct 2022 21:34:14 +0000

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

# User rules for freebsd
freebsd ALL=(ALL) NOPASSWD:ALL

while this has no affect on sudo's functionality, it's also not deduplicated:

freebsd@fbsd14-amd64 ~> sudo -l
User freebsd may run the following commands on fbsd14-amd64:
    (ALL) NOPASSWD: ALL
    (ALL) NOPASSWD: ALL
    (ALL) NOPASSWD: ALL
    (ALL) NOPASSWD: ALL
[...]

PR #4296, opened 2023-07-26
sudoers not idempotent (SC-1589) 

a-dubs:
## Proposed Commit Message

```
full re-runs will no longer write duplicate rules to sudoers file

Rerunning cloud-init would always append sudoer rules to the
sudoers file without checking if the rules already existed first.
This is no longer the case - only unique rules not already found
in the sudoers file will be added to it.

Fixes GH-4048 
LP: #1998539
```

## Additional Context
This change maintains the current functionality of only adding rules to the sudoers file without checking for old rules that
should be removed.   
Adding functionality for doing so should probably be addressed in a future ticket. 
See issue #4048 for more context on the bug this ticket is addressing.

### To replicate duplication bug: 
- Launch an LXD container using any cloud-init source that isnt this branch with the following cloud-config:
```
#cloud-config for SC-1589
groups:
  - admingroup: [root,sys]
  - cloud-users
users:
  - default
  - name: foobar
    gecos: Foo B. Bar
    primary_group: foobar
    groups: users
    expiredate: '2032-09-01'
    passwd: $6$j212wezy$7H/1LT4f9/N3wpgNunhsIqtMj62OKiS3nyNwuizouQc3u7MbYCarYeAHWYPYb2FT.lbioDm2RrkJPb9BZMN1O/
  - name: barfoo
    gecos: Bar B. Foo
    sudo: ALL=(ALL) NOPASSWD:ALL
    groups: users, admin
    lock_passwd: true
  - name: fizzbuzz
    sudo: false
    shell: /bin/bash
```
- Shell into the container and run the following command to fully rerun cloud init:
```
cloud-init clean --logs; cloud-init init --local; cloud-init init; cloud-init modules --mode=config; cloud-init modules --mode=final; cloud-init status --wait --long;
```
- Look at the sudoers file to see the duplicate rules (`cat /etc/sudoers.d/90-cloud-init-users`) and it should look like:
```
# Created by cloud-init v. ________________________________________

# User rules for barfoo
barfoo ALL=(ALL) NOPASSWD:ALL

# User rules for ubuntu
ubuntu ALL=(ALL) NOPASSWD:ALL

# User rules for barfoo
barfoo ALL=(ALL) NOPASSWD:ALL

# User rules for ubuntu
ubuntu ALL=(ALL) NOPASSWD:ALL
```

### To see new changes that prevent duplication bug from occurring:
- Launch an LXD container using this branch as the source cloud-init with the same cloud-config from above
- Follow all the same steps from above but this time when you peek at the sudoers file, there should be no duplicate rules and it should look like:
```
# Created by cloud-init v. ________________________________________

# User rules for barfoo
barfoo ALL=(ALL) NOPASSWD:ALL

# User rules for ubuntu
ubuntu ALL=(ALL) NOPASSWD:ALL
```

## Checklist:

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -860,6 +860,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         content += "\n"  # trailing newline
 
         self.ensure_sudo_dir(os.path.dirname(sudo_file))
+
         if not os.path.exists(sudo_file):
             contents = [
                 util.make_header(),
@@ -871,11 +872,14 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 util.logexc(LOG, "Failed to write sudoers file %s", sudo_file)
                 raise e
         else:
-            try:
-                util.append_file(sudo_file, content)
-            except IOError as e:
-                util.logexc(LOG, "Failed to append sudoers file %s", sudo_file)
-                raise e
+            if content not in util.load_file(sudo_file):
+                try:
+                    util.append_file(sudo_file, content)
+                except IOError as e:
+                    util.logexc(
+                        LOG, "Failed to append to sudoers file %s", sudo_file
+                    )
+                    raise e
 
     def create_group(self, name, members=None):
         group_add_cmd = ["groupadd", name]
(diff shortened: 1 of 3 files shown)


holmanb (approved):
Nice work! This change makes this idempotent and adds unit test coverage for this behavior. Looks good to me.

CalvoM (approved):
LGTM.

aciba90 (approved):
+1, thanks!

PR #4314, opened 2023-08-03
Replace flake8 with ruff

TheRealFalcon:
## Proposed Commit Message

```
Replace flake8 with ruff

Ruff is much faster and replicates everything flake8 can do.
Additionally, I added a few more checks that seemed relevant to our
project and would not require many changes.

Long term, we should be able to replace isort with ruff as well,
and maybe eventually pylint, though ruff is not yet able to do
everything that pylint can.
```

## Additional Context
Lots of files touched, but changes should be relatively uncontroversial. List of rules ruff can enable are listed here:
https://beta.ruff.rs/docs/rules/


cloudinit/sources/DataSourceAkamai.py
```diff
@@ -275,21 +274,18 @@ def _fetch_metadata(self, use_v6: bool = False) -> bool:
...
                 LOG.warning("Failed to base64 decode userdata due to %s", e)
         except url_helper.UrlError as e:
             # we failed to retrieve data with an exception; log the error and
             # return false, indicating that we should retry using a different
             # network if possible
             LOG.warning(
-                "Failed to retrieve metadata using IPv%s due to %s" "6"
-                if use_v6
-                else "4",
+                "Failed to retrieve metadata using IPv%s due to %s",
```
TheRealFalcon, line 287: Legit bug found here

holmanb: nice

holmanb (approved):
No objections for the code changes, they all look like either a clear improvement or an inconsequential change.

As for the replacement of flake8 with Ruff, I think this is fine. It has broad adoption and support, and as implemented here is stricter than flake8 so no functionality should be lost with this change.

PR #4315, opened 2023-08-03
Fix test_tools.py collection

TheRealFalcon:
## Proposed Commit Message

```
Fix test_tools.py collection

When required setuptools functions are not available, we skip running
the module. Doing this requires specifying `allow_module_level=True`,
which we were not doing before.
```


dermotbradley:
That fixes the problem for me. Thanks.

holmanb (approved):
LGTM

PR #4318, opened 2023-08-04
Undocument nocloud-net

holmanb:
```
Undocument nocloud-net

In f146fe cloud-init developed capability to
automatically determine nocloud mode. Remove
documentation related to nocloud-net since this
is no longer a necessary argument and may
confuse users during debugging.
```

Should probably deprecate use of this key too, to allow eventual code cleanup, but that can be a separate PR.

### DIFF
--- a/doc/examples/seed/README
+++ b/doc/examples/seed/README
@@ -3,11 +3,8 @@ This directory is an example of a 'seed' directory.
 
 copying these files inside an instance's
   /var/lib/cloud/seed/nocloud
-or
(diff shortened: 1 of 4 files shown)


TheRealFalcon (approved):
Should probably also update `doc/examples/seed/README`, `doc/rtd/tutorial/qemu-script.sh`, `doc/rtd/tutorial/qemu.rst`.

I'll leave it up to you whether to do that here or later.

PR #4319, opened 2023-08-04
cc_mounts: Fix swapfile not works on btrfs

bin456789:
## Proposed Commit Message

```
cc_mounts: Fix swapfile not works on btrfs

To make a swapfile works on btrfs
We need to create an empty file, add "no copy-on-write" attribute
before turn it into swapfile

Fixes GH-3713
LP: #1884127
```

## Additional Context
https://btrfs.readthedocs.io/en/latest/Swapfile.html

## Test Steps
    tox  -- tests/unittests/config/test_cc_mounts.py::TestSwapFileCreation::test_swap_creation_method_btrfs
    tests/unittests/config/test_cc_mounts.py::TestSwapFileCreation::test_swap_creation_method_btrfs PASSED [100%]

## Checklist:

### DIFF
--- a/cloudinit/config/cc_mounts.py
+++ b/cloudinit/config/cc_mounts.py
@@ -317,6 +317,10 @@ def create_swapfile(fname: str, size: str) -> None:
 
     fstype = util.get_mount_info(swap_dir)[1]
 
+    if fstype == "btrfs":
+        subp.subp(["truncate", "-s", "0", fname])
+        subp.subp(["chattr", "+C", fname])
+
     if (
         fstype == "xfs" and util.kernel_version() < (4, 18)
     ) or fstype == "btrfs":
--- a/tests/unittests/config/test_cc_mounts.py
+++ b/tests/unittests/config/test_cc_mounts.py
@@ -304,6 +304,8 @@ class TestSwapFileCreation(test_helpers.FilesystemMockingTestCase):
         cc_mounts.handle(None, self.cc, self.mock_cloud, [])
         self.m_subp_subp.assert_has_calls(
             [
+                mock.call(["truncate", "-s", "0", self.swap_path]),
+                mock.call(["chattr", "+C", self.swap_path]),
                 mock.call(
                     [
                         "dd",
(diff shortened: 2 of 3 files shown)


cloudinit/config/cc_mounts.py
```diff
@@ -317,6 +317,13 @@ def create_swap(fname, size, method):
 
     fstype = util.get_mount_info(swap_dir)[1]
 
+    if fstype == "btrfs":
+        try:
+            subp.subp(["truncate", "-s", "0", fname])
+            subp.subp(["chattr", "+C", fname])
+        except subp.ProcessExecutionError:
+            LOG.info("Failed to disable copy-on-write for swapfile on btrfs")
```
igalic, line 325: this log message will be wrong,

cloudinit/config/cc_mounts.py
```diff
@@ -317,6 +317,13 @@ def create_swap(fname, size, method):
 
     fstype = util.get_mount_info(swap_dir)[1]
 
+    if fstype == "btrfs":
+        try:
+            subp.subp(["truncate", "-s", "0", fname])
```
igalic, line 322: if this is the command that fails

bin456789: Is there any suggestion?

igalic: if truncate fails, the log message will say failed to disable copy on write.

despite the order in which i made the comments, they show up in the order of line numbers

bin456789: So we should move first command outside the try block. Or add another try-catch for it?
The first command is likely won't be failed. It just create a new file.

igalic: yes. if `truncate` fails, you don't want to catch the exception, because something is very wrong.

igalic (changes requested):

TheRealFalcon (approved):
LGTM! Thanks @igalic for the additional review.

PR #4320, opened 2023-08-04
Fix KeyError when handling users.0.uid in schema.py

Chavis00:
## Proposed Commit Message

```
Fix KeyError when executing `$ cloud-init schema -c ./ud.yaml --annotate` with a string uid.

The message "Changed in version 22.3. The use of the `string` type is deprecated.
Use an `integer` instead." was not handled for the `--annotate` argv. 
To handle it, now it checks if the `self._schemamarks` dict has the key 
before appending the message to `errors_by_line[self._schemamarks[path]]`.

Fixes GH-4305
```

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -605,7 +605,8 @@ class _Annotator:
                 errors_by_line[int(line)].append(msg)
             else:
                 col = None
-                errors_by_line[self._schemamarks[path]].append(msg)
+                if path in self._schemamarks:
+                    errors_by_line[self._schemamarks[path]].append(msg)
             if col is not None:
                 msg = "Line {line} column {col}: {msg}".format(
                     line=line, col=col, msg=msg


blackboxsw (changes requested):
Thank you @Chavis00 for providing the pull request and making cloud-init better. Unfortunately, the issue referenced by this pull request is looking for a fix in the annotations so that we get a hint inline for the deprecated property value. I'll update the original issue with comments to better suggest desired output. What your fix proposes is to ignore annotating a line if the path is not in _schemamarks, but that basically ignores a failure without addressing the underlying symptom.


Here's a better example of what output we are looking for for deprecation annotations:
```

# launch a container in LXD which has cloud-init in the image
lxc launch ubuntu-daily:mantic m && lxc shell m

# Write valid yaml with a deprecated property
#cloud-config
users:
    - name: a
      uid: "1743"
EOF

# Confirm deprecatation messages exist

$ cloud-init schema -c ./ud.yaml
Cloud config schema deprecations: users.0.uid:  Changed in version 22.3. The use of ``string`` type is deprecated. Use an ``integer`` instead.
Valid cloud-config: ./ud.yaml

#  Expected --annotated output should highlight he deprecated keys
$ cloud-init schema -c ./ud.yaml --annotate
#cloud-config
users:
    - name: -1                            # E1
      uid: "1743"                        # D1

# Errors: -------------
# E1: '-1' is not of type 'string'
# D1: Changed in version 22.3. The use of ``string`` type is deprecated. Use an ``integer`` instead.
```

But, alas the sympom we have is that the self._schemamarks is being initialized with the wrong nested path
```
(Pdb) self._schemamarks
{'users': 2, 'users.name.0': 3, 'users.name.0.name': 3, 'users.name.0.uid': 4}
```

and the valid path we are searching to match is `users.0` not `users.0.name`. 
I think the fix here is minimally to ensure schemamarks is properly generated for lists of dicts. This schema mark construction happens over in `cloudinit.safeyaml._CustomSafeLoaderWithMarks`.

I realize this is a bit more complex than just avoiding the traceback, if you wanted to get started on this I'd suggest looking over at the `tests/unittests/test_safeyaml.py` which exercises some of the schema mark validation.


For this to properly match the annotated path on failure, we probably want to following new unit test to pass:
```diff
diff --git a/tests/unittests/test_safeyaml.py b/tests/unittests/test_safeyaml.py
index 8713546f5..6178c2ed8 100644
--- a/tests/unittests/test_safeyaml.py
+++ b/tests/unittests/test_safeyaml.py
@@ -65,7 +65,13 @@ class TestLoadWithMarks:
                 b"a:\n- a1\n- a2\nb:\n- b3",
                 {"a": ["a1", "a2"], "b": ["b3"]},
                 {"a": 1, "a.0": 2, "a.1": 3, "b": 4, "b.0": 5},
-                id="separate_dicts_nestes_lists",
+                id="separate_dicts_nested_lists",
+            ),
+            pytest.param(
+                b"users:\n- name: -1\n  uid: '1234'\n",
+                {'users': [{'name': -1, 'uid': '1234'}]},
+                {'users': 1, 'users.0': 2, 'users.0.name': 2, 'users.0.uid': 3},
+                id="lists_nested_dicts",
             ),
         ),
     )
```

You can test this with tox -e py3 tests/unittests/test_safeyaml.py -vv

Issue #4321, opened 2023-08-07
TypeError: cannot unpack non-iterable NoneType object in has_mount_opt()


xiaoge1001:
# Bug report
If the get_mount_info() function returns None, TypeError is triggered.
https://github.com/canonical/cloud-init/blame/main/cloudinit/util.py#L2785

## Steps to reproduce the problem
python3 -m pytest tests/unittests/

## Environment details
- Cloud-init version:23.2.2
- Operating System Distribution:openEuler

## cloud-init logs
Sorry, I can't log in to the problem environment. This is a CI environment and I can't get some useful information. But the analysis code does have this problem.

## Additional information
```
[  123s] path = '/var/tmp/cloud-init', opt = 'noexec'
[  123s] 
[  123s]     def has_mount_opt(path, opt: str) -> bool:
[  123s] >       *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
[  123s] E       TypeError: cannot unpack non-iterable NoneType object
```

igalic:
what Version of cloud-init is this?
what OS is it running on?

xiaoge1001:
> what Version of cloud-init is this? what OS is it running on?

Cloud-init version:23.2.2
Operating System Distribution:openEuler release 22.03 (LTS-SP1)

xiaoge1001:
@igalic I'm sorry. The version information above is wrong. I tested two versions (21.4 and 23.2.2). 23.2.2 is fine, but 21.4+patch 919e22dc1d779f58f03dfd1cfde5ad9dc33a60d1 has problems.
However, analyzing the code shows that get_mount_info may indeed return None.

xiaoge1001:
I can't log into the environment and it's not clear what triggered the p

blackboxsw:
This feels a bit to me like unittests not mocking something it should have and through into the underlying test host/environment and which doesn't behave the way we expect. Is it possible to add your pytest logs or minimally the full failed log of the unit test names that are failing? That would help us narrow down which tests are problematic on openEuler

xiaoge1001:
@igalic @blackboxsw 
In my CI environment, the /etc/fstab file exists but the size is 0.

xiaoge1001:
I've learned that openeuler build environment is that there is no disk to mount in the chroot environment.

igalic:
So my analysis here, https://github.com/canonical/cloud-init/pull/4322#discussion_r1286482441 was correct.
[...]

PR #4322, opened 2023-08-07
handle TypeError when get_mount_info function get None

xiaoge1001:
## Proposed Commit Message
handle TypeError when get_mount_info function get None

Fixes GH-4321

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2782,7 +2782,14 @@ def get_mount_info(path, log=LOG, get_mnt_opts=False):
 
 
 def has_mount_opt(path, opt: str) -> bool:
-    *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
+    try:
+        *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
+    except TypeError:
+        LOG.warning(
+            "Get Nothing mount info from %s",
+            path,
+        )
+        return False
     return opt in mnt_opts.split(",")
 
 
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2786,7 +2786,7 @@ def has_mount_opt(path, opt: str) -> bool:
         *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
     except TypeError:
         LOG.warning(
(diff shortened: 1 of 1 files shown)


cloudinit/util.py
```diff
@@ -2782,7 +2782,14 @@ def get_mount_info(path, log=LOG, get_mnt_opts=False):
 
 
 def has_mount_opt(path, opt: str) -> bool:
-    *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
+    try:
+        *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
+    except TypeError:
+        LOG.warning(
+            "Get Nothing mount info from %s",
```
igalic, line 2789: this message reads really confusing

cloudinit/util.py
```diff
@@ -2782,7 +2782,14 @@ def get_mount_info(path, log=LOG, get_mnt_opts=False):
 
 
 def has_mount_opt(path, opt: str) -> bool:
-    *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
+    try:
+        *_, mnt_opts = get_mount_info(path, get_mnt_opts=True)
```
igalic, line 2786: why would this function return `None` to begin with?

xiaoge1001: > why would this function return `None` to begin with?

The reason still needs to be further identified. However, when we look at the implementation code of this function, there is some cases of returning None.

igalic: of all of those, only the first and the last would leave no trace in the log

the first one would mean your mountinfo file is in the wrong, or a new unknown, different, format.

the last one would mean no match was found, including no match for `/`!
I'm not sure there are scenarios where that makes sense (at least as a system to run cloud-init on)

cloudinit/util.py
igalic: I think this should say: `"Found no mount info for %s",`

xiaoge1001:
> I don't think that the current approach is desired to resolve the issue.

Yes, we may need to fundamentally address why function get_mount_info() returns None.

PR #4325, opened 2023-08-08
check whether old version is empty in postinst

jinkkkang:
## check whether old version is empty in postinst

```
summary: check whether old version is empty in postinst

The /var/lib/cloud/data/upgraded-network file touched after apt install cloudinit in
instance of cloud-init installed by source instead of deb pkg, add check  to determine whether the old version is empty during installation

LP: #2012044
```

### DIFF
--- a/debian/cloud-init.postinst
+++ b/debian/cloud-init.postinst
@@ -230,6 +230,10 @@ disable_network_config_on_upgrade() {
         # this is a fresh system not one that has been booted.
         return 0
     fi
+    if [ -z $oldver ];then
+        # if the previous cloudinit was installed by source instead of a deb pkg, oldver would be empty
+        return 0
+    fi
     if dpkg --compare-versions "$oldver" le "$last_without_net"; then
         echo "dpkg upgrade from $oldver" > /var/lib/cloud/data/upgraded-network
     fi
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -76,6 +76,7 @@ jordimassaguerpla
 jqueuniet
 jsf9k
 jshen28
+jinkkkang
 kadiron
 kaiwalyakoparkar
 kallioli


igalic:
> ```
> LP: #2012044
> ```

This LP issue id seems to be invalid.
Can you replace this with the correct GH issue id, please?

blackboxsw:
The issue LP: #2012044 is a downstream Ubuntu issue because is affects only Ubuntu's debian package postinstall script  which can be seen https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2012044.  Generally for downstream issues in RedHat or Ubuntu we expect those downstreams to carry bugs related to their packaging so it's ok that we don't have an upstream GH issue for this.

blackboxsw (changes requested):
@jinkkkang thank you for providing a Pull request to make cloud-init better. 

Re-reading the original bug discussion,  I'm uneasy about this approach and the original bug having had time to think this over. It is tough to support the transition from a source based install to a deb-based install for the following reasons:
 1. we are going to miss/skip the business/package upgrade logic baked into the debian postinst that assumes behavior in the previous releases of the previous postinst.
 2. A source-based install never invoked any earlier cloud-init.postinst logic for prior versions for setup or teardown and migration from earlier versions, so our postinst probably can't trust that the latest current source version number has all previous postinst functions applied. 
 4. Upgrading from a source-based install may not take into account complex PYTHONPATHs which could leave 'oldversion' of cloud-init modules active in a path before the deb-based deliverable paths under /usr/lib/python3/dist-packages
 5. Skipping the function is not on unset $oldver only hides the problem, we need to define a function/mechanism to truly determine which version of cloud-init was installed locally
 6. Finding the right install path for local cloudinit modules may be complex as different ``install-dir` or `--prefix` could have been used in those images
 7. Should our debian package postinst cleanup the old locally installed cloudinit sources during the upgrade path to prevent oldversions from being invoked inadvertently on the system? If different PYTHONPATHs are present on these custom installs, it's highly likely that an newversion detecting the datasource will cause tracebacks for oldversions trying to query that datasource after the fact due to object serialization issues.
 



Minimally, If we are trying to add support to Ubuntu's `cloud-init.postinst` to allow upgrade from previous source-based install, I think we generally would need the following:
 1.  provide a function or mechanism in postinst to determine what the previous version $oldver of cloud-init is present on the system from either invoking `cloudiinit.version.version_string()` or grepping __VERSION__ out of cloudinit.version module.
 2.  make sure all oldver checks in cloud-init.postinst use the discovered source VERSION from step 1 to determine whether or not to operate
 3. Emit warnings in /var/log/cloud-init..log and /var/log/cloud-init-output.log from postinst saying this is not a fully supported upgrade path. The log should record the source install directory from which we upgraded, the version we thought we upgraded from and informs the user that they should likely remove that source install in favor of deb-packaged cloud-init.


Other general questions:
1. why support moving from python install-based cloudinit to deb-package based cloud-init? Why not stick with the python install-based distribution to upgrade, those images are already off the typical support path for what the distribution tests? As upstream we test and validated upgrade paths from deb-package to deb-package because we don't let "newer" cloud-init into stable releases until it has undergone significant verification and testing. Generally, folks who are installing a python program from source are those that are off-path for typical support and pulling down 'newer' versions of that program than what the distribution says is supported.

PR #4327, opened 2023-08-09
Fix/ds identify with grep

smoser:
## Proposed Commit Message
Use grep for faster parsing of cloud config in ds-identify.

'while read line ... < /file' is very slow with dash, because it
reads character by character.

The change here is to change config_check to incur the cost of a
single 'grep' fork.  That way, grep is able to do the matching.
Grep is much faster than dash (or bash) at filtering the non-matching
lines.

LP: #2030729

## Additional Context
The default cloud.cfg was added because the majority of hte unit tests
did not have one, and thus check_cfg would only read a very small
amount of lines (or none at all).  The realistic config allows the
performance benefit to be shown.

## Test Steps
./tools/tox-venv py3 - tests/unittests/test_ds_identify.py

## Checklist:

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -777,21 +777,24 @@ check_config() {
     if [ "$1" = "$files" -a ! -f "$1" ]; then
         return 1
     fi
-    local fname="" line="" ret="" found=0 found_fn=""
-    # shellcheck disable=2094
-    for fname in "$@"; do
-        [ -f "$fname" ] || continue
-        while read line; do
-            line=${line%%#*}
-            case "$line" in
-                $key:\ *|"${key}":)
-                    ret=${line#*:};
-                    ret=${ret# };
-                    found=$((found+1))
-                    found_fn="$fname";;
-            esac
-        done <"$fname"
+    local line="" ret="" found=0 found_fn="" oifs="$IFS" out=""
+    out=$(grep "$key\"\?:" "$@" 2>/dev/null)
+    IFS=${CR}
+    for line in $out; do
+        # drop '# comment'
+        line=${line%%#*}
+        # if more than one file was 'grep'ed, then grep will output filename:
+        # but if only one file, line will not be prefixed.
+        if [ $# -eq 1 ]; then
+            found_fn="$1"
+        else
+            found_fn="${line%%:*}"
+            line=${line#*:}
+        fi
+        ret=${line#*: };
+        found=$((found+1))
     done
+    IFS="$oifs"
     if [ $found -ne 0 ]; then
         _RET="$ret"
         _RET_fname="$found_fn"
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -57,6 +57,146 @@ BLKID_UEFI_UBUNTU = [
 ]
 
 
+DEFAULT_CLOUD_CONFIG = """\
+# The top level settings are used as module
+# and base configuration.
+# A set of users which may be applied and/or used by various modules
+# when a 'default' entry is found it will reference the 'default_user'
+# from the distro configuration specified below
+users:
+   - default
+
+# If this is set, 'root' will not be able to ssh in and they
+# will get a message to login instead as the default $user
+disable_root: true
+
+# This will cause the set+update hostname module to not operate (if true)
+preserve_hostname: false
+
+# If you use datasource_list array, keep array items in a single line.
+# If you use multi line array, ds-identify script won't read array items.
+# Example datasource config
+# datasource:
+#    Ec2:
+#      metadata_urls: [ 'blah.com' ]
+#      timeout: 5 # (defaults to 50 seconds)
+#      max_wait: 10 # (defaults to 120 seconds)
+
+# The modules that run in the 'init' stage
+cloud_init_modules:
+ - migrator
+ - seed_random
+ - bootcmd
+ - write-files
+ - growpart
+ - resizefs
+ - disk_setup
+ - mounts
+ - set_hostname
+ - update_hostname
+ - update_etc_hosts
+ - ca-certs
+ - rsyslog
+ - users-groups
+ - ssh
+
+# The modules that run in the 'config' stage
+cloud_config_modules:
+ - wireguard
+ - snap
+ - ubuntu_autoinstall
+ - ssh-import-id
+ - keyboard
+ - locale
+ - set-passwords
+ - grub-dpkg
+ - apt-pipelining
+ - apt-configure
+ - ubuntu-advantage
+ - ntp
+ - timezone
+ - disable-ec2-metadata
+ - runcmd
+ - byobu
+
+# The modules that run in the 'final' stage
+cloud_final_modules:
+ - package-update-upgrade-install
+ - fan
+ - landscape
+ - lxd
+ - ubuntu-drivers
+ - write-files-deferred
+ - puppet
+ - chef
+ - ansible
+ - mcollective
+ - salt-minion
+ - reset_rmc
+ - refresh_rmc_and_interface
+ - rightscale_userdata
+ - scripts-vendor
+ - scripts-per-once
+ - scripts-per-boot
+ - scripts-per-instance
+ - scripts-user
+ - ssh-authkey-fingerprints
+ - keys-to-console
+ - install-hotplug
+ - phone-home
+ - final-message
+ - power-state-change
+
+# System and/or distro specific settings
+# (not accessible to handlers/transforms)
+system_info:
+   # This will affect which distro class gets used
+   distro: ubuntu
+   # Default user name + that default users groups (if added/used)
+   default_user:
+     name: ubuntu
+     lock_passwd: True
+     gecos: Ubuntu
+     groups: [adm, audio, cdrom, floppy, lxd, netdev, plugdev, sudo, video]
+     sudo: ["ALL=(ALL) NOPASSWD:ALL"]
+     shell: /bin/bash
+   network:
+     renderers: ['netplan', 'eni', 'sysconfig']
+     activators: ['netplan', 'eni', 'network-manager', 'networkd']
+   # Automatically discover the best ntp_client
+   ntp_client: auto
+   # Other config here will be given to the distro class and/or path classes
+   paths:
+      cloud_dir: /var/lib/cloud/
+      templates_dir: /etc/cloud/templates/
+   package_mirrors:
+     - arches: [i386, amd64]
+       failsafe:
+         primary: http://archive.ubuntu.com/ubuntu
+         security: http://security.ubuntu.com/ubuntu
+       search:
+         primary:
+           - http://%(ec2_region)s.ec2.archive.ubuntu.com/ubuntu/
+           - http://%(availability_zone)s.clouds.archive.ubuntu.com/ubuntu/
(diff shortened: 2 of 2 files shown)


tools/ds-identify
```diff
@@ -777,21 +777,24 @@ check_config() {
...
-            case "$line" in
-                $key:\ *|"${key}":)
-                    ret=${line#*:};
-                    ret=${ret# };
-                    found=$((found+1))
-                    found_fn="$fname";;
-            esac
-        done <"$fname"
+    local line="" ret="" found=0 found_fn="" oifs="$IFS" out=""
+    out=$(grep "$key\"\?:" "$@" 2>/dev/null)
```
smoser, line 781: we are ignoring the exit code of grep here. grep will exit 0 on "found", 1 on "not found", 2 on file-not-found. Unfortunately, nothing is checking if the file globs provided to this function exist.  Previously the for loop would stat each expanded glob member. 

When shell expands a glob, any non-matching globs remain in the expanded form.  so if you called with
 
    /path/to/existing.cfg, /path/to/not-existing-glob*, /path/to/not-existing-file

The result "$@" will have

    /path/to/existing.cfg /path/to/not-existing-glob* /path/to/not-existing-file

We're just passing that list right through to grep without checking it.  Checking/filtering it would cost another fork.

bash has an option that is useful for this ('nullglob') but that's not posix /bin/sh.

smoser: If you're interested, this change checks grep exit code and logs an error:

```diff
diff --git a/tools/ds-identify b/tools/ds-identify
index ad028766f..dcf350a8e 100755
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -765,20 +765,35 @@ check_config() {
     # file_globs are expanded via path expansion and
     # default to /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg
     # currently does not respect any hierarchy in searching for key.
-    local key="$1" files=""
+    local key="$1" files="" f=""
     shift
     if [ $# -eq 0 ]; then
         files="${PATH_ETC_CI_CFG} ${PATH_ETC_CI_CFG_D}/*.cfg"
     else
         files="$*"
     fi
-    # shellcheck disable=2086
-    { set +f; set -- $files; set -f; }
-    if [ "$1" = "$files" -a ! -f "$1" ]; then
+
+    # expand the globs and check if files exist.
+    set --
+    set +f
+    for f in $files; do
+       [ -f "$f" ] || continue
+       set -- "$@" "$f"
+    done
+    set -f
+
+    if [ $# -eq 0 ]; then
         return 1
     fi
-    local line="" ret="" found=0 found_fn="" oifs="$IFS" out=""
+    local line="" ret="" found=0 found_fn="" oifs="$IFS" out="" rc=""
     out=$(grep "$key\"\?:" "$@" 2>/dev/null)
+    rc=$?
+    if [ $rc -eq 1 ]; then
+        return 1
+    elif [ $rc -ne 0 ]; then
+        error "Unexpected exit code $rc from grep in check_config"
+    fi
+
     IFS=${CR}
     for line in $out; do
         # drop '# comment'
```

smoser:
@holmanb , any feedback here?

holmanb (commented):
@smoser Thanks for putting this PR together.

We might want to drop a message on the dash mailing list to let them know that performance issues are identified in the current implementation, given that speed is a purported advantage of dash.

From a [somewhat related discussion](https://lore.kernel.org/dash/20221214010539.g3zy2jtzdrhsyrrz@tarta.nabijaczleweli.xyz/T/#t) within the last year related to stdin parsing in dash:

> > Reading the standard input byte-by-byte is the obvious solution to this
> issue, Just Works, and is how all other shells do it (we could,
> theoretically, read regular files block-wise, then seek within them after
> parsing, but the complexity out-weighs the rarity of running
> sh < program; we could also do whole-line reads on teletypes in
> icanon mode, but, again, the gain here is miniscule for an interactive
> session, and the teletype mode can change at any time, so...).
> Naturally, we keep reading block-wise for non-standard-input.
> 
> There are a few things to consider here:
> 
> - Not all shells do it this way. bash does do the block-wise read,
>    followed by a seek, when stdin is seekable. While I agree that it is
>    not necessary and not worth it, you specifically say that other shells
>    do not do this. That's simply not true.

As for your PR, I think it looks good.

@smoser Do you think the extra check of the `grep` exit code is necessary? I could see it helping debugging, I guess, if `/etc/cloud/cloud.cfg` is missing, but since we don't already have it and the purpose of this PR is to deal with performance penalty, I'm curious if you think it's actually required / worth it.

As for the performance I ran checked some performance stats and this looks to be about a 2x improvement:

```
root@me:~# time ./ds-id-current.sh 

real	0m0.010s
user	0m0.005s
sys	0m0.005s
root@me:~# time ./ds-id-proposed.sh 

real	0m0.005s
user	0m0.004s
sys	0m0.000s

root@me:~# perf stat ./ds-id-proposed.sh 

 Performance counter stats for './ds-id-proposed.sh':

              3.73 msec task-clock                       #    0.907 CPUs utilized             
                 2      context-switches                 #  536.695 /sec                      
                 1      cpu-migrations                   #  268.348 /sec                      
               213      page-faults                      #   57.158 K/sec                     
           5678936      cycles                           #    1.524 GHz                         (92.91%)
            413455      stalled-cycles-frontend          #    7.28% frontend cycles idle      
           1066640      stalled-cycles-backend           #   18.78% backend cycles idle       
           5858808      instructions                     #    1.03  insn per cycle            
                                                  #    0.18  stalled cycles per insn   
           1198864      branches                         #  321.712 M/sec                     
             12845      branch-misses                    #    1.07% of all branches             (19.07%)

       0.004109971 seconds time elapsed

       0.002412000 seconds user
       0.001847000 seconds sys


root@me:~# perf stat ./ds-id-current.sh 

 Performance counter stats for './ds-id-current.sh':

              8.43 msec task-clock                       #    0.950 CPUs utilized             
                 0      context-switches                 #    0.000 /sec                      
                 0      cpu-migrations                   #    0.000 /sec                      
                68      page-faults                      #    8.063 K/sec                     
          11251741      cycles                           #    1.334 GHz                         (52.65%)
            329144      stalled-cycles-frontend          #    2.93% frontend cycles idle        (86.94%)
           2462323      stalled-cycles-backend           #   21.88% backend cycles idle       
          18141715      instructions                     #    1.61  insn per cycle            
                                                  #    0.14  stalled cycles per insn   
           3604065      branches                         #  427.359 M/sec                     
             45971      branch-misses                    #    1.28% of all branches             (60.41%)

       0.008881366 seconds time elapsed

       0.004466000 seconds user
       0.004466000 seconds sys

```

for completeness these are the functions I pruned out of ds-identify:

```
root@me:~# cat ds-id-proposed.sh 
#!/bin/sh

proposed_check_config() {
    # check_config(key [,file_globs])
    # somewhat hackily read through file_globs for 'key'
    # file_globs are expanded via pathight expansion and
    # default to /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg
    # currently does not respect any hierarchy in searching for key.
    local key="$1" files=""
    shift
    if [ $# -eq 0 ]; then
        files="${PATH_ETC_CI_CFG} ${PATH_ETC_CI_CFG_D}/*.cfg"
    else
        files="$*"
    fi
    # shellcheck disable=2086
    { set +f; set -- $files; set -f; }
    if [ "$1" = "$files" -a ! -f "$1" ]; then
        return 1
    fi
    local line="" ret="" found=0 found_fn="" oifs="$IFS" out=""
    out=$(grep "$key\"\?:" "$@" 2>/dev/null)
    IFS=${CR}
    for line in $out; do
        # drop '# comment'
        line=${line%%#*}
        # if more than one file was 'grep'ed, then grep will output filename:
        # but if only one file, line will not be prefixed.
        if [ $# -eq 1 ]; then
            found_fn="$1"
        else
            found_fn="${line%%:*}"
            line=${line#*:}
        fi
        ret=${line#*: };
        found=$((found+1))
    done
    IFS="$oifs"
    if [ $found -ne 0 ]; then
        _RET="$ret"
        _RET_fname="$found_fn"
        return 0
    fi
    return 1
}

proposed_check_config datasource_list /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg
root@me:~# cat ds-id-current.sh 
#!/bin/sh

current_check_config() {
    # check_config(key [,file_globs])
    # somewhat hackily read through file_globs for 'key'
    # file_globs are expanded via path expansion and
    # default to /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg
    # currently does not respect any hierarchy in searching for key.
    local key="$1" files=""
    shift
    if [ $# -eq 0 ]; then
        files="${PATH_ETC_CI_CFG} ${PATH_ETC_CI_CFG_D}/*.cfg"
    else
        files="$*"
    fi
    # shellcheck disable=2086
    { set +f; set -- $files; set -f; }
    if [ "$1" = "$files" -a ! -f "$1" ]; then
        return 1
    fi
	local fname="" line="" ret="" found=0 found_fn=""
	# shellcheck disable=2094
	for fname in "$@"; do
	[ -f "$fname" ] || continue
	while read line; do
	    line=${line%%#*}
	    case "$line" in
		$key:\ *|"${key}":)
		    ret=${line#*:};
		    ret=${ret# };
		    found=$((found+1))
		    found_fn="$fname";;
	    esac
	done <"$fname"
    done
    if [ $found -ne 0 ]; then
        _RET="$ret"
        _RET_fname="$found_fn"
        return 0
    fi
    return 1
}
	
current_check_config datasource_list /etc/cloud/cloud.cfg /etc/cloud/cloud.cfg.d/*.cfg
```

This was test was ran on a stock Jammy cloud image.

holmanb (approved):
Per test data in prior comment, this change demonstrates a measurable performance improvement in ds-identify.

Thanks for this change @smoser!

smoser:
(@holmanb , sorry i went ghost on this PR I must have just missed your question.

> Do you think the extra check of the grep exit code is necessary? I could see it helping debugging, I guess, if /etc/cloud/cloud.cfg is missing, but since we don't already have it and the purpose of this PR is to deal with performance penalty, I'm curious if you think it's actually required / worth it.

I kind of expected you to ask for me to change to read the exit code.  I think i'd probably try to go that route if i were doing it again (explicitly checking grep RC).  Swallowing errors just makes things harder to debug when something is going wrong.  Its probably not a big deal here.

Thanks for landing.

Random thought - it would be nice to have subp collect 'duration' of a command.

holmanb:
> Swallowing errors just makes things harder to debug when something is going wrong.

Agreed. I'm just not sure what error conditions that patch would actually catch, besides the one I mentioned above. If you'd rather I can still add it.

> Thanks for landing.

Of course, and thanks for the PR.

> Random thought - it would be nice to have subp collect 'duration' of a command.

Agreed, and that sounds like a simple change to make too. I'll ping you for review on an implementation shortly if you don't mind.

holmanb:
> Swallowing errors just makes things harder to debug when something is going wrong.

Agreed. I'm just not sure what error conditions that patch would actually catch, besides the one I mentioned above. If you'd rather I can still add it.

> Thanks for landing.

Of course, and thanks for the PR.

> Random thought - it would be nice to have subp collect 'duration' of a command.

Agreed. PR here -> https://github.com/canonical/cloud-init/pull/4606

PR #685, opened 2020-11-24
Adding BOOTPROTO = dhcp to render sysconfig dhcp6 stateful on RHEL

otubo:
## Proposed Commit Message
Adding BOOTPROTO = dhcp to render sysconfig dhcp6 stateful on RHEL

BOOTPROTO needs to be set to 'dhcp' on RHEL so NetworkManager can
properly acquire ipv6 address.

rhbz: #1859695

Signed-off-by: Eduardo Otubo <otubo@redhat.com>

## Additional Context
The original [bz 1859695](https://bugzilla.redhat.com/show_bug.cgi?id=1859695) is mostly marked as private but the bug as noticed while configuring an internal IPv6 network on Openstack between the instance and an NFS storage. No patch required or bug noticed if the instance was configured with a regular ipv6/dhcp interface.

## Test Steps
(as described on the above mentioned BZ)
Steps to Reproduce:
1. Create StorageNFS network with IPv6 CIDR and address range defined
2. Create a Nova instance with 2 NICs, 2nd NIC connected to StorageNFS network
3. Login to VM and examin Eth1 IP address

Actual results:
The Nova assigned IPv6 address was not assigned to device Eth1

Expected results:
Expected the Nova assigned IP address to be assigned to Eth1 interface

## Checklist:
 - [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/hacking.html)
 - [X] I have updated or added any unit tests accordingly
 - [X] I have updated or added any documentation accordingly

PR #4329, opened 2023-08-10
net: fix ipv6_dhcpv6_stateful configuration for rhel

ani-sinha:
When network type is ipv6_dhcpv6-stateful, cloud-init seems to enable dhcp for both ipv4 and ipv6. Network manager prefers dhcp over ipv4 and hence dhcp6 is not used to obtain the IP address. This is incorrect. For ipv6_dhcpv6-stateful networks, we should set:

```
ipv4.method = disabled // disables all ipv4 dhcp
ipv6.method = dhcp
ipv6.may-fail = no // dhcp6 must succeed.
```
For ifcfg files (sysconfig renderer), the corresponding changes should be:
```
BOOTPROTO = none // instead of dhcp so that dhcp4 is disabled. 
IPV6_FAILURE_FATAL = yes // so that dhcp6 should succeed.
```
This patch fixes this. This patch has been tested by Red Hat QE and it seems to fix the above issue.

RHBZ: 2227767
fixes: f550c8765ca03d ("Adding BOOTPROTO = dhcp to render sysconfig dhcp6 stateful on RHEL (#685)")

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -36,7 +36,7 @@ LOG = logging.getLogger(__name__)
 class NMConnection:
     """Represents a NetworkManager connection profile."""
 
-    def __init__(self, con_id):
+    def __init__(self, con_id, cloud_config=None):
         """
         Initializes the connection with some very basic properties,
         notably the UUID so that the connection can be referred to.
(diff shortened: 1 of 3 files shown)


ani-sinha:
Seems we have not considered a scenario where both ipv6-stateful and ipv4 dhcp is enabled. Will need to fix it. Closing.

PR #4336, opened 2023-08-11
tests/net: fix TestGetInterfaces' mock coverage for get_master

cjp256:
get_master() was checking against system's sysfs which may cause failures when the mocked interface conflicted with real one.


rmhsawyer (approved):

anhvoms (approved):
LGTM

bos37 (approved):

PR #4343, opened 2023-08-14
Ubuntu/lunar: Fix removing the unused network manager hook

TheRealFalcon:
DO NOT SQUASH

When publishing packaging changes to `ubuntu/devel`, I got sponsor feedback that our `rm_conffile` should target the version being prepared rather than any previous version (see the man page quoted in the commit message). Additionally, we should include the LP number for the change.

If this get's a +1, I'll push the same changes for focal and bionic.


holmanb (approved):
Looks good, thanks!

PR #4346, opened 2023-08-14
cc_wireguard: make tests pass on FreeBSD

igalic:
## Proposed Commit Message

```
cc_wireguard: make tests pass on FreeBSD

These tests relied on the SUT to provide sensible return values for
util.kernel_version(). On FreeBSD, those values are 14.0-CURRENT for my
development machine, which doesn't work well with when split at '.' and
compared against a pair of integers (14, '0-CURRENT') < (5, 4) doesn't
make much sense.

This patch adds `@mock`s for util.kernel_version wherever it is used.
That way, test_cc_wireguard.py tests passes on a FreeBSD system.

```

## Checklist:

### DIFF
--- a/tests/unittests/config/test_cc_wireguard.py
+++ b/tests/unittests/config/test_cc_wireguard.py
@@ -22,7 +22,6 @@ class FakeCloud:
 
 
 class TestWireGuard(CiTestCase):
-
     with_logs = True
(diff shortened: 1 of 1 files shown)


holmanb (approved):
LGTM, thanks!

PR #4347, opened 2023-08-14
analyze: fix (unexpected) timestamp parsing

igalic:
## Proposed Commit Message

```
analyze: fix (unexpected) timestamp parsing

In case of very unexpected timestamps, we pass them on to date.
Unfortunately, only GNU date is able to perform this level of deduction
(guess work).

Rework the code and tests to look for GNU date on non-Linux platforms.
Only fail, and loudly! if GNU date cannot be found.

This patch (partially)

fixes GH-4333

```

## Checklist:

### DIFF
--- a/cloudinit/analyze/dump.py
+++ b/cloudinit/analyze/dump.py
@@ -4,7 +4,7 @@ import calendar
 import sys
 from datetime import datetime
 
-from cloudinit import atomic_helper, subp
+from cloudinit import atomic_helper, subp, util
 
 stage_to_description = {
     "finished": "finished running cloud-init",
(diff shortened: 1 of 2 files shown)


cloudinit/analyze/dump.py
```diff
@@ -44,14 +44,21 @@ def parse_timestamp(timestampstr):
...
 
 
 def parse_timestamp_from_date(timestampstr):
-    out, _ = subp.subp(["date", "+%s.%3N", "-d", timestampstr])
+    cmd = "date"
+    if not util.is_Linux():
+        if subp.which("gdate"):
+            cmd = "gdate"
+        else:
+            raise ValueError("Unable to parse timestamps without GNU date(1)")
```
igalic, line 60: I think we should include the concrete timestamp here, so we can learn and adapt

igalic, line 60: I think we should include the concrete timestamp here, so we can learn and adapt

igalic (commented):
thoughts?

cloudinit/analyze/dump.py
holmanb: By that do you mean something like this?
```
           raise ValueError(f"Unable to parse timestamps without GNU date: [{timestampstr}]")
```

I think it would certainly help debugging to get this information.

holmanb (commented):
Aside from "making it just work" (which this PR does), I'm tempted to say that maybe it would be better if at some point in the future we didn't depend on GNU date and transitionally logged a warning anytime `parse_timestamp_from_date()` is called so that we can collect logs for any formats that could be handled in python code directly. I didn't see this path actually taken on Ubuntu, so I'm curious when (if) this path actually gets taken.

Thoughts?

cloudinit/analyze/dump.py
igalic: yes. that sort of thing.
wanna put that into a "suggestion", so i can merge it from my phone?
(I'm off the laptop)

igalic:
I've only seen this in tests.
I've never seen it hit an exception in real life

cloudinit/analyze/dump.py
holmanb:
```suggestion
            raise ValueError(
                f"Unable to parse timestamps without GNU date: [{timestampstr}]"
            )
```

holmanb: > yes. that sort of thing. wanna put that into a "suggestion", so i can merge it from my phone? (I'm off the laptop)

sure, just did

holmanb:
> I've only seen this in tests. I've never seen it hit an exception in real life

I'll make it a separate bug.

holmanb (approved):

PR #4348, opened 2023-08-15
util.get_proc*: Fix tests on non-Linux systems

igalic:
## Proposed Commit Message

```
util.get_proc*: Fix tests on non-Linux systems

`/proc/$pid/environ` and `/proc/$pid/stat` don't exist on BSD.
procfs on BSDs is highly limited, and its use very discouraged.

While FreeBSD has procstat(1), which could reasonably read a process
environment, that use-case in the code is Linux specific and doesn't map
on FreeBSD.

Guard get_proc_env() tests with is_Linux().
The code already mostly uses this as a fallback, when all else fails.

implement get_proc_ppid() with `ps(1)`.
This should work on most platforms that have `ps(1)`.

Fixes: GH-4332
```

## Checklist:

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3117,9 +3117,9 @@ def udevadm_settle(exists=None, timeout=None):
     return subp.subp(settle_cmd)
 
 
-def get_proc_ppid(pid):
+def get_proc_ppid_linux(pid):
     """
-    Return the parent pid of a process.
+    Return the parent pid of a process by parsing /proc/$pid/stat.
     """
     ppid = 0
     try:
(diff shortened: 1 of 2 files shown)


cjp256 (approved):
lgtm.  fwiw this is where unit testing on different distros would be handy (to verify that ps works as expected) :D

igalic:
the only other distro that needs checking would be Alpine (and soon AIX ;)

according to busybox man page, this should work.
(same for AIX https://www.ibm.com/docs/en/aix/7.1?topic=p-ps-command )

holmanb (approved):
Looks good, thanks!

Note: Since the callsite of `get_proc_ppid()` has a distro object passed in, this might be cleaner implemented as a classmethod of the base class and bsd childclass class, but this is an improvement and looks like it should work as is so Iĺl merge.

PR #4351, opened 2023-08-15
cc_growpart: fix tests to run on FreeBSD

igalic:
## Proposed Commit Message

```
cc_growpart: fix tests to run on FreeBSD

because /etc/rc.d/growfs exists on FreeBSD, we need to be selective
about when to allow tests access to it, so that they will pass.

```

## Checklist:


holmanb (approved):
Thanks!

PR #4361, opened 2023-08-16
Fix cc_keyboard in mantic

TheRealFalcon:
## Proposed Commit Message

```
Fix cc_keyboard in mantic

localectl set-x11-keymap is currently broken for new debian based
installs. Workaround it by writing /etc/default/keyboard directly

LP: #2030788
```

## Additional Context
https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/2030788
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1038762
`man 5 keyboard`

## Test Steps
I haven't been able to verify this manually, so it may be worth asking the submitter to verify first


cloudinit/distros/debian.py
```diff
@@ -300,11 +300,38 @@ def update_package_sources(self):
...
+        contents = "\n".join(
+            [
+                "# This file was generated by cloud-init",
+                "",
+                f'XKBMODEL="{model}"',
+                f'XKBLAYOUT="{layout}"',
+                f'XKBVARIANT="{variant}"',
+                f'XKBOPTIONS="{options}"',
+                "",
+                'BACKSPACE="guess"',  # This is provided on default installs
```
CalvoM, line 320: Might be a nitpick... Is it possible to allow the user to configure this, but still default to "guess" incase they do not configure this option.

blackboxsw: While a good thought, the implementation is highly debian-specific given that this implementation is writing directly to /etc/default/keyboard. Surfacing such and option is something that won't be supportable easily on any other distribution that is calling `localectl` directly, because that command doesn't support setting backspace values.

cloudinit/distros/debian.py
```diff
@@ -300,11 +300,38 @@ def update_package_sources(self):
...
+                "",
+            ]
+        )
+        util.write_file(
+            filename="/etc/default/keyboard",
+            content=contents,
+            mode=644,
```
blackboxsw: Mode here needs to be octal not decimal, this resulted in some very bad behavior on mantic when cloud-init wrote the file with perms 1204 :)
```
root@testlocale-manticdaily:~# localectl status
Could not get properties: Access Denied

root@testlocale-manticdaily:~# ls -l /etc/default/keyboard 
-rw-r----T 1 root root 120 Aug 25 02:31 /etc/default/keyboard
```
More details in [the bug](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2030788/comments/7)

blackboxsw (changes requested):
Needs fixing for 0o644. I'll push a commit to resolve this and test on stable releases.

blackboxsw (approved):
Confirmed behavior on stable releases is retained and keyboard/map is set properly on Focal, Jammy and Lunar as well as mantic  while writing /etc/default/keyboard directly.

PR #4363, opened 2023-08-17
cc_users_groups: Add doas/opendoas support

dermotbradley:
## Proposed Commit Message

```
cc_users_groups: Add doas/opendoas support

Add doas (an alternative to sudo) support for Alpine, Debian, FreeBSD,
OpenBSD, and Ubuntu to cc_users_groups module for user creation/update.
```

## Checklist:


blackboxsw:
Adding priority label so we can get a review on this prior to 23.3 if we can.

holmanb (commented):
Thanks @dermotbradley for this PR. I've taken a cursory glance at the code, but left a comment on the proposed UX.

I like the idea of doas as an alternative to sudo in this module, and I don't see any glaring issue with the code (though I'm curious if it makes sense to add in cloud.cfg.tmpl on distros that don't default to `doas` such as ubuntu).

holmanb:
Also, a rebase on tip of main will fix the failing doc test.

dermotbradley:
> (though I'm curious if it makes sense to add in cloud.cfg.tmpl on distros that don't default to `doas` such as ubuntu).

Ok, I can remove Debian and Ubuntu from the template.

@igalic can you confirm which of BSD distros come with doas installed by default?

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -140,6 +140,24 @@
           "description": "The user's login name. Required otherwise user creation will be skipped for this user.",
           "type": "string"
         },
+        "doas": {
+          "description": "List of doas rules to use. Absence of a doas value or ``null`` will result in no doas rules added for this user.",
```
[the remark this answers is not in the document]
dermotbradley: > I assume this is from copying the sudo schema entry?

Yes that's where it came from.

Ok, I'll change it to just use a list of strings.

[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/__init__.py
```diff
@@ -791,6 +798,89 @@ def chpasswd(self, plist_in: list, hashed: bool):
...
+        if isinstance(rules, (list, tuple)):
+            for rule in rules:
+                if not self.is_doas_rule_valid(user, rule):
+                    msg = (
+                        "Invalid doas rule %r for user '%s',"
+                        " not writing any doas rules for user!" % (rule, user)
+                    )
+                    LOG.error(msg)
+                    return
+        elif isinstance(rules, str):
```
holmanb, line 846: This should never be true with the updated jsonschema, right?

cloudinit/distros/__init__.py
```diff
@@ -791,6 +798,89 @@ def chpasswd(self, plist_in: list, hashed: bool):
...
+            contents = [util.make_header(), content]
+            try:
+                util.write_file(doas_file, "\n".join(contents), mode=0o440)
+            except IOError as e:
+                util.logexc(LOG, "Failed to write doas file %s", doas_file)
+                raise e
+        else:
+            if content not in util.load_file(doas_file):
+                try:
+                    util.append_file(doas_file, content)
```
holmanb, line 877: Think we should append the header before the cloud-init generated content?

holmanb (commented):
@dermotbradley Thanks for making the updates. I got a chance to look more closely at the code today. Altogether it looks really good. I have a few questions inline, but I think this is pretty close to being ready.

cloudinit/distros/__init__.py
```diff
@@ -791,6 +798,89 @@ def chpasswd(self, plist_in: list, hashed: bool):
         cmd = ["chpasswd"] + (["-e"] if hashed else [])
         subp.subp(cmd, payload)
 
+    def is_doas_rule_valid(self, user, rule):
+        rule_pattern = (
+            r"^(?:permit|deny)"
+            r"(?:\s+(?:nolog|nopass|persist|keepenv|setenv \{[^}]+\})+)*"
```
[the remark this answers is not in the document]
dermotbradley: I hadn't considered adding support for doas to the "groups:" section, only for "users:".

cloudinit/distros/__init__.py
dermotbradley: Not sure, the entries added for each individual user have "# cloud-init User rules for <user>" headers added.

cloudinit/distros/__init__.py
dermotbradley: Right I missed that, I fixed this and a similar section just after it.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -140,6 +140,14 @@
           "description": "The user's login name. Required otherwise user creation will be skipped for this user.",
           "type": "string"
         },
+        "doas": {
+          "description": "List of doas rules to add for a user.",
```
holmanb, line 144: Since this key is only supported on distros that have doas in place of sudo, could we document something to say that? That will help avoid user confusion.

Maybe something like:
```
List of doas rules to add for a user. Doas must be installed for rules to take effect.
```

holmanb (changes requested):
@dermotbradley One small change request then I'll merge.

cloudinit/config/schemas/schema-cloud-config-v1.json
dermotbradley: Done

holmanb (approved):
Thanks for putting this together @dermotbradley.

Merging

PR #4366, opened 2023-08-18
read-version: fallback to get_version when git describe fails

blackboxsw:
## Proposed Commit Message
```
read-version: fallback to get_version when git describe fails

Annotated git tags are not present in shallow git clones. In shallow
clones, git describe may fail with exit 128 with one of the following
errors:
 - No tags can describe <hash>
 - No names found, cannot describe anything

If either message is present, fallback to
cloudinit.version.version_string.

This fixes readthedocs CI builds which may use git clone --depth 1 in
Annotated git tags are not present in shallow git clones. In shallow
clones, git describe may fail with exit 128 with one of the following
errors:
 - No tags can describe <hash>
 - No names found, cannot describe anything

If either message is present, fallback to
cloudinit.version.version_string.

Also pin sphinx=7.1.2 as the release 7.2.0 on  August 17 dropped the
extension setup_js_tag_helper used by our sphinx builds.

This fixes readthedocs CI builds which may use git clone --depth 1 in
the doc build process.
```

## Checklist:

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -2,7 +2,7 @@ doc8
 furo
 m2r2
 pyyaml
-sphinx
+sphinx==7.1.2
 sphinx-design
 sphinx-copybutton
 sphinx-notfound-page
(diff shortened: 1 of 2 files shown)


holmanb (approved):
Thanks for this @blackboxsw!

PR #4372, opened 2023-08-21
azure/imds: increase read-timeout to 30s

cjp256:
When fetching metadata or reprovisiondata, allow up to 30 seconds for a read timeout.  For cases where a read request is delayed, this would eliminate extra follow-up request(s).


anhvoms (approved):
LGTM

holmanb (commented):
Looks good, thanks!

holmanb (approved):

PR #4368, opened 2023-08-18
test: provide puppet agent --onetime to avoid blocking daemon

blackboxsw:
## Proposed Commit Message

```
test: provide puppet agent --onetime to avoid blocking daemon

test_puppet invokes puppet agent --noop which spawns puppet agent as a daemon which lives after cloud-final.service exits. Systemd v253 will report SubState=running in this case which indefinitely blocks cloud-init status --wait.

Avoid leaving that daemon around by passing --onetime which will force the puppet agent to exit once it has run.
```

## Additional Context

Jenkins job hitting 10 hour timeout on `cloud-init status --wait` due to puppet agent as background process not exiting
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-mantic-lxd_container/50/console

## Test Steps
```
# make sure the integration test doesn't block indefinitely on cloud-init status --wait
CLOUD_INIT_OS_IMAGE=mantic tox -e integration-tests tests/integration_tests/modules/test_puppet.py
```

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [x] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #4374, opened 2023-08-21
status: treat SubState=running and MainPID=0 as service exited

blackboxsw:
b## Proposed Commit Message
```
status: treat SubState=running and MainPID=0 as service exited

Avoid cloud-init status --wait blocking indefinitely on systemd
status SubState=running and MainPID=0 condition for cloud-init
systemd services.

When daemons are spawned by a systemd unit
in the unified heirarchy, long-lived processes or daemons launched by
the systemd unit will be tracked in the CGroup/slice.
As long as spawned daemons of processes live, the systemd tooling will
report the parent systemd unit as SubState=running instead of
SubState=exited.

This affected Ubuntu across the 23.04 -> 23.10 release boundary
because a long-term patch to ignore cgroup status was dropped at
systemd 252.5 which last existed in Ubuntu 23.04 (Lunar).

Cloud-init status --wait checks MainPID == 0 when SubState=running
as a secondary indicator the the executing process of the boot
stage has completed despite any forked daemons.

Additionally fix typo in puppet integration test name.

Fixes GH-4373
```

## Test Steps

```
# Make pkg from this branch
# Run puppet integration test and expect it to pass with the deb built from this branch
CLOUD_INIT_OS_IMAGE=mantic CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_23.2-109-g259cb588-1~bddeb_all.deb tox -e integration-tests ./tests/integration_tests/modules/test_puppet.py
```

## Checklist:

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -213,7 +213,7 @@ def _get_systemd_status() -> Optional[UXAppStatus]:
             [
                 "systemctl",
                 "show",
-                "--property=ActiveState,UnitFileState,SubState",
+                "--property=ActiveState,UnitFileState,SubState,MainPID",
                 service,
             ],
         ).stdout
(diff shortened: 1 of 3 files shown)


blackboxsw:
a more general approach that #4368 to treating SubState for cloud-init boot stage units which spawn background processes but have finished running themselves and transferred control back to systemd.

holmanb (approved):
@blackboxsw Thanks for putting together this fix. I have a small nitpick for the commit message, but the code change looks good to me. 

This approach is general enough that it should also catch daemons started in `bootcmd`/`runcmd` so I greatly prefer this over https://github.com/canonical/cloud-init/pull/4368.

Nitpick: Both uses of "or background processes" and the one use of are redundant. I think the commit message would read more clearly without them.

Issue #4244, opened 2023-07-13
test_ansible_controller fails to use pip to install packages on Lunar


blackboxsw:
# Bug report

Seen in jenkins daily integration logs on tip of main, affects cloud-init 23.1 and later.

## Environment details
- Cloud-init version: 23.1
- Operating System Distribution: Ubuntu Lunar
- Cloud provider, platform or installer type: lxd_vm

## cloud-init logs
Seen @ https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-lunar-lxd_vm/119/testReport/junit/tests.integration_tests.modules/test_ansible/test_ansible_controller/

### excerpt
```
"'/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py'>) failed\n"
 'Traceback (most recent call last):\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/config/modules.py", line '
 '256, in _run_modules\n'
 '    ran, _r = cc.run(\n'
 '              ^^^^^^^\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/cloud.py", line 67, in run\n'
 '    return self._runners.run(name, functor, args, freq, clear_on_fail)\n'
 '           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/helpers.py", line 172, in '
 'run\n'
 '    results = functor(**args)\n'
 '              ^^^^^^^^^^^^^^^\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line '
 '179, in handle\n'
 '    ansible.install(package_name)\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line '
 '142, in install\n'
 '    self.do_as([*cmd, "--upgrade", "pip"])\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line '
 '99, in do_as\n'
 '    return self.distro.do_as(command, self.run_user, **kwargs)\n'
 '           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n'
 '  File "/usr/lib/python3/dist-packages/cloudinit/distros/__init__.py", line '
 '1003, in do_as\n'
[...]

Issue #4356, opened 2023-08-15
Integration test failure: test_ansible_controller


TheRealFalcon:
`tests/integration_tests/modules/test_ansible.py::test_ansible_controller`
Looks to be a legit failure. Consistently happens on Lunar:
```
2023-08-15 23:11:13,710 - util.py[WARNING]: Running module ansible (<module 'cloudinit.config.cc_ansible' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py'>) failed
2023-08-15 23:11:13,710 - util.py[DEBUG]: Running module ansible (<module 'cloudinit.config.cc_ansible' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py'>) failed
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/cloudinit/config/modules.py", line 256, in _run_modules
    ran, _r = cc.run(
              ^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/cloud.py", line 67, in run
    return self._runners.run(name, functor, args, freq, clear_on_fail)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/helpers.py", line 172, in run
    results = functor(**args)
              ^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line 179, in handle
    ansible.install(package_name)
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line 142, in install
    self.do_as([*cmd, "--upgrade", "pip"])
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_ansible.py", line 99, in do_as
    return self.distro.do_as(command, self.run_user, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/distros/__init__.py", line 1011, in do_as
    return subp.subp(
           ^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 335, in subp
    raise ProcessExecutionError(
cloudinit.subp.ProcessExecutionError: Unexpected error while running command.
[...]

PR #4380, opened 2023-08-23
Fix pip-managed ansible

holmanb:
```
Fix pip-managed ansible

PEP 668 disallows installing packages via pip outside of virtual
environments, to avoid breaking due to conflicting system packages.

Ansible is unlikely to be installed in the base system, and if it is and
the user still chooses to use `install_method: pip`, that would be an
invalid configuration.

This flag exists to allow users to bootstrap ansible to run as a
controller. The alternative solution would be to require virtual
environments, however this is often packaged as a separate dependency,
which further complicates bootstrap. Allow installing ansible outside of
virtual environments via pip's --break-system-packages.

Fixes GH-4244
```

fixes https://github.com/canonical/cloud-init/issues/4244

## Test
```
tox -re integration-tests -- tests/integration_tests/modules/test_ansible.py
```

### DIFF
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -136,7 +136,13 @@ class AnsiblePullPip(AnsiblePull):
                 import pip  # noqa: F401
             except ImportError:
                 self.distro.install_packages(self.distro.pip_package_name)
-            cmd = [sys.executable, "-m", "pip", "install"]
+            cmd = [
+                sys.executable,
+                "-m",
+                "pip",
+                "install",
+                "--break-system-packages",
+            ]
             if self.run_user:
                 cmd.append("--user")
             self.do_as([*cmd, "--upgrade", "pip"])


holmanb:
@blackboxsw #4356 is a duplicate of your older report of the same issue (#4244).

I will just close the new one as dup and mark the old one priority and link them, since we don't need to keep both around.

blackboxsw (approved):
@holmanb as you suggest, this approach is safe from an ansible pip-installed extensions perspective and the typical intent of an Ansible consumer wanting to have fine-grained control of the version of anisble and extensions installed via pip on the system.   Even [PEP 668](https://peps.python.org/pep-0668/#mark-the-installation-as-externally-managed#motivation) refernces Ansible as one of those use-cases where this override is desirable behavior:
```In some cases, however, it’s useful and intentional to install a Python package from outside of the distro that influences the behavior of distro-shipped commands. This is common in the case of software like Sphinx or Ansible which have a mechanism for writing Python-language extensions. A user may want to use their distro’s version of the base software (for reasons of paid support or security updates) but install a small extension from PyPI, and they’d want that extension to be importable by the software in their base system.```

PR #4381, opened 2023-08-23
Ubuntu/jammy fix daily recipe

blackboxsw:


aciba90 (approved):
+1

Thanks, @blackboxsw for the changes and the detailed explanation.

- [x] build_pkg
- [x] sbuild
- [x] no new lintian errors.

aciba90 (changes requested):
Please, we need an extra commit describing what we modified in debian/changelog.

PR #4382, opened 2023-08-23
Ubuntu/focal fix daily recipe

blackboxsw:


aciba90 (changes requested):
Please, we need an extra commit describing what we modified in debian/changelog.

aciba90 (approved):
LGTM.

There is no need to add context to debian/changelog because this quilt patch has not yet been released.

Issue #4386, opened 2023-08-24
ec2: exception when processing empty userdata


nmeyerhans:
# Bug report

#4276 introduced support for decoding base64-encoded userdata retrieved from the EC2 IMDS.  However, it introduced a regression in the case where an instance runs with no userdata.  In that case, cloud-init can fail with messages similar to the following:

```plain
2023-08-24 00:35:59,584 - url_helper.py[DEBUG]: [0/6] open 'http://169.254.169.254:80/2021-03-23/user-data' with {'url': 'http://169.254.169
.254:80/2021-03-23/user-data', 'allow_redirects': True, 'method': 'GET', 'timeout': 5.0, 'headers': {'X-aws-ec2-metadata-token': 'REDACTED'}
} configuration
2023-08-24 00:35:59,590 - util.py[WARNING]: Failed reading from metadata address http://169.254.169.254:80
2023-08-24 00:35:59,590 - util.py[DEBUG]: Failed reading from metadata address http://169.254.169.254:80
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/cloudinit/sources/DataSourceEc2.py", line 538, in crawl_metadata
    crawled_metadata["user-data"] = util.maybe_b64decode(raw_userdata)
  File "/usr/lib/python3.9/site-packages/cloudinit/util.py", line 157, in maybe_b64decode
    raise TypeError("data is '%s', expected bytes" % type(data))
TypeError: data is '<class 'str'>', expected bytes
```
I will follow up with a pull request to address this issue.

## Steps to reproduce the problem

Launch an EC2 instance running the 'main' branch of cloud-init and that does did not have a UserData object passed to the RunInstances API call.  In this case, the IMDS will return a 404 error from the `http://169.254.169.254:80/2021-03-23/user-data` IMDS endpoint and cloud-init will fail with the exception shown above.

## Environment details
- Cloud-init version: main
- Operating System Distribution: Verified on Debian sid and Amazon Linux 2023
- Cloud provider, platform or installer type: Amazon EC2

PR #4387, opened 2023-08-24
ec2: initialize get_instance_userdata return value to bytes

nmeyerhans:
https://github.com/canonical/cloud-init/pull/4276 uncovered an issue with the initialization of the return value for get_instance_userdata().  The return value was initialized with

`user_data = ""`

which is a str class.  It then calls `url_helper.read_file_or_url()`, which attempts to retrieve user-data content from IMDS. `read_file_or_url()` returns its results as a bytes object, which is then passed directly up to the caller.  In the event that `read_file_or_url()` does not successfully retrieve content (e.g. because it was given a `file://` path to a nonexistent file or an `http://` path that generates a 404 code), an exception is raised and `get_instance_userdata()` returns the string object initially stored in user_data.

Rather than make the caller cope with return data potentially encoded as either `bytes` or `str`, this commit changes the initialization of `user_data` to an empty bytes object, ensuring type consistency in `get_instance_userdata()`'s return value.

Fixes https://github.com/canonical/cloud-init/issues/4386

## Proposed Commit Message

```
ec2: initialize get_instance_userdata return value to bytes

https://github.com/canonical/cloud-init/pull/4276 uncovered an issue
with the initialization of the return value for
get_instance_userdata().  The return value was initialized with

user_data = ""

which is a str class.  It then calls url_helper.read_file_or_url(),
which attempts to retrieve user-data content from IMDS.
read_file_or_url() returns its results as a bytes object, which is
then passed directly up to the caller.  In the event that
read_file_or_url() does not successfully retrieve
content (e.g. because it was given a file:// path to a nonexistent
file or an http:// path that generates a 404 code), an exception is
raised and get_instance_userdata() returns the string object initially
stored in user_data.

Rather than make the caller cope with return data potentially encoded
as either bytes or str, this commit changes the initialization of
user_data to an empty bytes object, ensuring type consistency in
get_instance_userdata()'s return value.

Fixes https://github.com/canonical/cloud-init/issues/4386

Signed-off-by: Noah Meyerhans <nmeyerha@amazon.com>
```

## Test Steps

Launch an instance with no userdata, such that the EC2 IMDS returns 404 for the `user-data` key.  Prior to the PR, cloud-init would generate an exception as documented in #4386. With this change, it should properly cope with that scenario and record a zero-length value for userdata.  This will create a zero-length file `/var/lib/cloud/instance/user-data.txt` as expected, etc.

## Checklist:

### DIFF
--- a/cloudinit/sources/helpers/ec2.py
+++ b/cloudinit/sources/helpers/ec2.py
@@ -152,7 +152,7 @@ def get_instance_userdata(
 ):
     ud_url = url_helper.combine_url(metadata_address, api_version)
     ud_url = url_helper.combine_url(ud_url, "user-data")
-    user_data = ""
+    user_data = b""
     try:
         if not exception_cb:
             # It is ok for userdata to not exist (thats why we are stopping if
--- a/tests/unittests/sources/helpers/test_ec2.py
+++ b/tests/unittests/sources/helpers/test_ec2.py
@@ -27,7 +27,7 @@ class TestEc2Util(helpers.ResponsesTestCase):
             status=404,
         )
         userdata = ec2.get_instance_userdata(self.VERSION, retries=0)
-        self.assertEqual("", userdata)
+        self.assertEqual(b"", userdata)
 
     def test_userdata_fetch_fail_server_dead(self):
         self.responses.add(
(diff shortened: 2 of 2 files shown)


holmanb (approved):
Thanks for this fix @nmeyerhans!

PR #4392, opened 2023-08-28
pep-594: drop deprecated pipes module import

blackboxsw:
## Proposed Commit Message

```
pep-594: drop deprecated pipes module import

python3.11 will deprecated pipes module 3.13 will drop it from main.

cloud-init only used the undocumented pipes.quote function, which is actually only wrapper around shlex.quote[1].

Use shlex.quote instead.

[1] https://github.com/python/cpython/blob/3.11/Lib/pipes.py#L64-L66
```

## Checklist:

### DIFF
--- a/cloudinit/distros/parsers/sys_conf.py
+++ b/cloudinit/distros/parsers/sys_conf.py
@@ -4,8 +4,8 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import pipes
 import re
+import shlex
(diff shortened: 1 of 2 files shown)


holmanb (approved):
This is also true all the way back to [Python 3.6](https://github.com/python/cpython/blob/3.6/Lib/pipes.py#L64-L66) so this should be fine.

Just need to fix the isort failure, but then feel free to merge!

Thanks!

PR #4395, opened 2023-08-29
net: fix ipv6_dhcpv6_stateful/stateless/slaac configuration for rhel

ani-sinha:
When network type is ipv6_dhcpv6-stateful/stateless/slaac, cloud-init seems to
enable dhcp for both ipv4 and ipv6. Network manager prefers dhcp over ipv4 and
hence dhcp6 is not used to obtain the IP address. This is incorrect.
For only ipv6_dhcpv6-stateful/stateless/slaac networks, we should set:
    `ipv4.method = disabled // disables all ipv4 dhcp`
    
For ifcfg files (sysconfig renderer), the corresponding changes should be:
    `BOOTPROTO = none // instead of dhcp so that dhcp4 is disabled.`

Additionally, for only ipv6_dhcpv6_stateful, we should set:
    `ipv6.may-fail = no // dhcp6 must succeed.`
    
which translates to the following ifcfg setting:
    `IPV6_FAILURE_FATAL = yes // so that dhcp6 should succeed.`
    
This patch fixes this for rhel. The patch has been tested by Red Hat QE.
    
RHBZ: 2046491
fixes: f550c8765ca03d3 ("Adding BOOTPROTO = dhcp to render sysconfig dhcp6 stateful on RHEL (#685)")

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -105,6 +105,14 @@ class NMConnection:
         if self.config[family]["method"] == "auto" and method == "manual":
             return
 
+        if (
+            subnet_type == "ipv6_dhcpv6-stateful"
+            or subnet_type == "ipv6_dhcpv6-stateless"
+            or subnet_type == "ipv6_slaac"
+        ):
+            # set ipv4 method to 'disabled' to align with sysconfig renderer.
+            self._set_default("ipv4", "method", "disabled")
+
         self.config[family]["method"] = method
         self._set_default(family, "may-fail", "false")
 
@@ -342,6 +350,7 @@ class Renderer(renderer.Renderer):
 
     def __init__(self, config=None):
         self.connections = {}
+        self.config = config
 
     def get_conn(self, con_id):
         return self.connections[con_id]
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -436,13 +436,13 @@ class Renderer(renderer.Renderer):
                         iface_cfg["BOOTPROTO"] = "dhcp6"
                     iface_cfg["DHCLIENT6_MODE"] = "managed"
                 # only if rhel AND dhcpv6 stateful
-                elif (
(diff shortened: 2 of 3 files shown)


cloudinit/net/network_manager.py
```diff
@@ -105,6 +107,15 @@ def _set_ip_method(self, family, subnet_type):
         if self.config[family]["method"] == "auto" and method == "manual":
             return
 
+        flavor = self.cloud_config.get("flavor", "rhel")
```
TheRealFalcon, line 110: I'm not sure that checking for flavor here really makes any sense given that [RHEL doesn't provide any network-manager specific renderer configuration](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/rhel.py#L33) (nor does any other distro ATM). It seems to give the illusion that this would be a RHEL-only change, rather than a change affecting all distros using network-manager. If you want it to be RHEL-only, then you need `network-manager` config in the renderer configuration. However, given the nature of this change, I think it makes sense to apply it universally.

TheRealFalcon (changes requested):
Thanks for the change! I have one inline comment about renderer configuration, but other than that I think this looks good.

cloudinit/net/network_manager.py
ani-sinha: Oh I see. I was looking at [this](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/sysconfig.py#L440) and I thought I add that check because it would simply make it symmetric with sysconfig.

ani-sinha: If you think applying it universally would be safe and we would not break anyone else, I am fine with removing the check.

TheRealFalcon: Yes, in this case I think it should be fine.

ani-sinha: Addressed in latest patch.

TheRealFalcon (approved):
Thanks!

PR #4397, opened 2023-08-29
cc_ubuntu_advantage: do not rely on uaclient.messages module

orndorffgrant:
## Proposed Commit Message

```
cc_ubuntu_advantage: do not rely on uaclient.messages module

Importing a uaclient module that is not under "uaclient.api" is not
supported and likely to break in the future.

The expected message_codes are defined as part of the API of `pro enable
--format=json` and won't change, so we don't lose anything by hardcoding
the value.
```

## Test Steps

No functional change. The ubuntu_advantage `enable` and `enable_beta` options should still work.

## Checklist:

### DIFF
--- a/cloudinit/config/cc_ubuntu_advantage.py
+++ b/cloudinit/config/cc_ubuntu_advantage.py
@@ -334,16 +334,9 @@ def configure_ua(token, enable=None):
     # related. We can distinguish them by checking if `service` is non-null
     # or null respectively.
 
-    # pylint: disable=import-error
-    from uaclient.messages import ALREADY_ENABLED
-
-    # pylint: enable=import-error
-
-    UA_MC_ALREADY_ENABLED = ALREADY_ENABLED.name
-
     enable_errors: List[dict] = []
     for err in enable_resp.get("errors", []):
-        if err["message_code"] == UA_MC_ALREADY_ENABLED:
+        if err["message_code"] == "service-already-enabled":
             LOG.debug("Service `%s` already enabled.", err["service"])
             continue
         enable_errors.append(err)
--- a/tests/integration_tests/modules/test_ubuntu_advantage.py
+++ b/tests/integration_tests/modules/test_ubuntu_advantage.py
@@ -142,6 +142,19 @@ class TestUbuntuAdvantage:
         verify_clean_log(log)
         assert is_attached(client)
 
+        # Assert service-already-enabled handling for esm-infra.
+        # First totally destroy ubuntu-advantage-tools data and state.
+        # This is a hack but results in a system that thinks it
+        # is detached even though esm-infra is still enabled.
+        # When cloud-init runs again, it will successfully re-attach
+        # and then notice that esm-infra is already enabled.
+        client.execute("rm -rf /var/lib/ubuntu-advantage")
+        assert client.execute("cloud-init clean --logs").ok
+        client.restart()
+        log = client.read_from_file("/var/log/cloud-init.log")
+        verify_clean_log(log)
+        assert "Service `esm-infra` already enabled" in log
+
 
 def maybe_install_cloud_init(session_cloud: IntegrationCloud):
     source = get_validated_source(session_cloud)
--- a/tests/integration_tests/modules/test_ubuntu_advantage.py
+++ b/tests/integration_tests/modules/test_ubuntu_advantage.py
@@ -148,7 +148,7 @@ def maybe_install_cloud_init(session_cloud: IntegrationCloud):
 
     launch_kwargs = {
         "image_id": session_cloud.cloud_instance.daily_image(
-            CURRENT_RELEASE.image_id, image_type=ImageType.PRO
+            CURRENT_RELEASE.series, image_type=ImageType.PRO
         )
     }
 
--- a/tests/integration_tests/releases.py
+++ b/tests/integration_tests/releases.py
@@ -53,6 +53,21 @@ class Release:
     def __repr__(self):
         return f"Release({self.os}, {self.version})"
 
+    def __eq__(self, other: object):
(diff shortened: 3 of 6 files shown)


cloudinit/config/cc_ubuntu_advantage.py
```diff
@@ -334,16 +334,9 @@ def configure_ua(token, enable=None):
...
-    from uaclient.messages import ALREADY_ENABLED
-
-    # pylint: enable=import-error
-
-    UA_MC_ALREADY_ENABLED = ALREADY_ENABLED.name
-
     enable_errors: List[dict] = []
     for err in enable_resp.get("errors", []):
-        if err["message_code"] == UA_MC_ALREADY_ENABLED:
+        if err["message_code"] == "service-already-enabled":
```
blackboxsw, line 339: @orndorffgrant Will this hard-coding of this string hit is an an unknown dependency on uaclient if those error codes change in the future, or are we guaranteed backwards-compatibliity with error message_codes in all future releases?

blackboxsw (changes requested):
Since the is an opportunity for drift in the future if uaclient changes message_code values and cloud-init now depends on a hard-coded string for message_code. Let's extend the existing tests/integration_tests/modules/test_ubuntu_advantage.py jobs to attempt to assert that we have a log "Service `livepatch` is already enabled" which shows we properly parsed the message_code value.


Here's the suggested integration test extension it includes as bug fix as it looks like we didn't have an __eq__ defined for Releases which resulted in always skipping `test_custom_services`
```diff
diff --git a/tests/integration_tests/modules/test_ubuntu_advantage.py b/tests/integration_tests/modules/test_ubuntu_advantage.py
index 0b3ed2d7b..acd4589f2 100644
--- a/tests/integration_tests/modules/test_ubuntu_advantage.py
+++ b/tests/integration_tests/modules/test_ubuntu_advantage.py
@@ -142,13 +142,23 @@ class TestUbuntuAdvantage:
         verify_clean_log(log)
         assert is_attached(client)
 
+        # Assert esm-infra already enabled sevice handling
+        # Clear ubuntu-advantage attach status, but don't disable services
+        client.execute("rm -rf /var/lib/ubuntu-advantage")
+        assert client.execute("cloud-init clean --logs").ok
+        client.restart()
+        log = client.read_from_file("/var/log/cloud-init.log")
+        verify_clean_log(log)
+        assert "Service `esm-infra` already enabled" in log
+
+
 
 def maybe_install_cloud_init(session_cloud: IntegrationCloud):
     source = get_validated_source(session_cloud)
 
     launch_kwargs = {
         "image_id": session_cloud.cloud_instance.daily_image(
-            CURRENT_RELEASE.image_id, image_type=ImageType.PRO
+            CURRENT_RELEASE.series, image_type=ImageType.PRO
         )
     }
 
diff --git a/tests/integration_tests/releases.py b/tests/integration_tests/releases.py
index 45b0ffccd..aa8078b0a 100644
--- a/tests/integration_tests/releases.py
+++ b/tests/integration_tests/releases.py
@@ -53,6 +53,13 @@ class Release:
     def __repr__(self):
         return f"Release({self.os}, {self.version})"
 
+    def __eq__(self, other: "Release"):
+        if all([self.os == other.os, self.version == other.version]):
+            if self.image_id and other.image_id:
+                return self.image_id == other.image_id
+            return True  # If either image_id is None match os and version
+        return False
+
     def __lt__(self, other: "Release"):
         if self.os != other.os:
             raise ValueError(f"{self.os} cannot be compared to {other.os}!")
diff --git a/tests/integration_tests/util.py b/tests/integration_tests/util.py
index 857d89cc8..b113ebf26 100644
--- a/tests/integration_tests/util.py
+++ b/tests/integration_tests/util.py
@@ -69,6 +69,10 @@ def verify_clean_log(log: str, ignore_deprecations: bool = True):
         "WARNING]: Could not match supplied host pattern, ignoring:",
     ]
     traceback_texts = []
+    if "install canonical-livepatch" in log:
+        warning_texts.append(
+            "canonical-livepatch returned error when checking status"
+        )
     if "oracle" in log:
         # LP: #1842752
         lease_exists_text = "Stderr: RTNETLINK answers: File exists"

```

test procedure
```
# ensure you have pyproject.toml with the follwing ec2 credentials
cat >> ~/.config/pycloudlib.toml <<EOF
[ec2]
access_key_id = "YOUR_KEY"
secret_access_key = "YOUR_SECRET_KEY"
region = "us-east-2" # or other
public_key_path = "<preferred_ssh_pubkey_or_pem>"
private_key_path = "<preferred_ssh_privatekey_or_pem>"
key_name = "mykey" # can be found with aws ec2 describe-key-pairs
EOF

CLOUD_INIT_UA_TOKEN=<YOUR_PRO_TOKEN> CLOUD_INIT_PLATFORM=ec2 tests/integration_tests/modules/test_ubuntu_advantage.py::TestUbuntuAdvantagePro::test_custom_services
```

cloudinit/config/cc_ubuntu_advantage.py
orndorffgrant: Unless we make a mistake, message codes will not change. So if this were to break it would be an ubuntu-pro-client bug, not a cloud-init bug.

In the future, ubuntu-pro-client should provide an api function for enabling that you can import. That would come with defined exceptions that you can catch instead of looking at json error objects.

orndorffgrant:
Awesome thank you for the detailed suggestion @blackboxsw - This will be a good opportunity to explore cloud-init testing too :)

orndorffgrant:
@blackboxsw Just added your suggestions. I split them up a bit and tested them to make sure I understood and I think they make sense. LMK if we need anything else here

Issue #4359, opened 2023-08-16
Rebooting VM during packer build causes cloud-init to run instance modules again


shaneholder:
# Bug report
I am attempting to use Packer to create a golden image that mounts /var as a partition.  In the Packer build data is copied from several folders to the mounted *alt_* folders and then a sed command is executed against fstab to replace the /alt_ with a blank to make the real mount points.  The VM is then rebooted before continuing with the rest of the golden image configuration, however after the boot the changes made to fstab are replace by cloud-init.  Log files are attached.

Packer build section
``` 
build {
  sources = ["source.azure-arm.linux-image"]

  provisioner "file" {
    source = "${local.scripts_folder}/100_golden.cfg"
    destination = "/tmp/"
  }

  provisioner "shell" {
    inline = ["sudo mv /tmp/100_golden.cfg /etc/cloud/cloud.cfg.d/"]
  }

  # copy var to /alt_var
  provisioner "shell" {
    inline = [
      "sudo sh -c 'cd /var/; tar cf - . | (cd /alt_var; tar xf -)'",
      "sudo sh -c 'cd /home/; tar cf - . | (cd /alt_home; tar xf -)'",
      "sudo sh -c 'cd /tmp/; tar cf - . | (cd /alt_tmp; tar xf -)'"      
    ]
  }

  # reboot vm to mount /var
  provisioner "shell" {
    inline = [
      "sudo sed -i 's#/alt_#/#' /etc/fstab",
    ]
  }

  # reboot vm to mount /var
  provisioner "shell" {
    inline = [
      "sudo reboot"
    ]
    expect_disconnect = true
    pause_after = "60s"
  }
}
```

Cloud Config
``` 
#cloud-config

disk_setup:
  /dev/disk/azure/scsi1/lun0:
    table_type: gpt
    layout: 
      - 15
      - 14
      - 28
      - 14
      - 14
      - 14
    overwrite: True

fs_setup:
  - device: /dev/disk/azure/scsi1/lun0
    partition: 1
    filesystem: ext4
  - device: /dev/disk/azure/scsi1/lun0
    partition: 2
    filesystem: ext4
  - device: /dev/disk/azure/scsi1/lun0
    partition: 3
    filesystem: ext4
  - device: /dev/disk/azure/scsi1/lun0
    partition: 4
    filesystem: ext4
  - device: /dev/disk/azure/scsi1/lun0
    partition: 5
    filesystem: ext4
[...]

PR #4399, opened 2023-08-30
Move disable checks out of systemd generator

holmanb:
#### Please do not squash merge this PR.

The commits in this PR are written separately with the intent of being added to the git tree individually. 

<details>
<summary>Old squashed commit message</summary>

```
Move disable checks out of systemd generator

- Add environment variable check to unit conditionals
- Change unit conditionals from rhel-only to all distros
- Add unit conditionals to cloud-init target
- Makefile: Benchmark generator script
- generator: Remove redundant code
- generator: Eliminate variables, branches and functions
- docs: Describe environment variable disablement method
- cloud-init status: Fix bug in containers where
  disabling cloud-init with KERNEL_CMDLINE causes
  cloud-init status --wait to loop indefinitely
- Add integration test coverage for disablement


- KERNEL_CMDLINE environment variable is now checked on
  all platforms which may simplify testing and presents
  minimal risk of unintended use.
- Kernel cmdline, cloud-init.disabled, and KERNEL_CMDLINE
  environment variable now have higher precedence than
  ds-identify policy in all distros, not just RHEL. 

https://www.freedesktop.org/software/systemd/man/systemd.unit.html
https://www.freedesktop.org/software/systemd/man/systemd-analyze.html
```
<br>
</details>

Behavior changes:

## Additional Details
Motivations: eliminate redundant code, performance improvements (do less in systemd generator, which blocks all of boot), standardize behavior across distros

The current [cloud-init docs](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) that describe how to disable cloud-init are incomplete, since [ds-identify policy](https://github.com/canonical/cloud-init/issues/4359#issuecomment-1699682165)  may take precedence over the disable file, the kernel commandline and environment variable arguments. This is not ideal, as ds-identify.cfg is an undocumented and less user-friendly user interface for a common user operation.

Assuming that we want cloud-init to behave as the docs describe (i.e.: `touch /etc/cloud/cloud-init.disabled` actually disables cloud-init), this PR fixes that discrepancy between docs and behavior.

## Performance Implications
As @blackboxsw mentioned out of band:

> having the disable in generator script timeframe may solve the case earlier than having systemd's ordering logic have to calculate and redact the cloud-init* boot stages from boot. Not sure if that matters in terms of boot time.

and some of my own thoughts:

> The current generator code may allow the systemd ordering solver to run sooner without cloud-init, but it also delays all of boot with checking kernel commandline and cloud-init.disabled file while the generator is running (in a shell script), so I think there are perf costs and benefits with this (potential) change.
>
> Where the benefit of the generator pruning out cloud-init only improves performance in the case that cloud-init isn't running (by skipping ds-identify), and the potential speed benefit is the case where cloud-init is not disabled (by eliminating a redundant `systemd-detect-virt` call, and kernel cmdline and file checking that could be done by systemd later, and potentially faster).

Benchmarking the systemd generator shows a 2-3x performance improvement. However, this appears negligible since `systemd-analyze critical-chain` shows identical times for both (below) - and some work does get shifted from the generator to systemd itself in this change (later in boot).

Generator Benchmark: this branch
```
# make benchmark-generator 
python3 ./tools/render-cloudcfg --variant="benchmark" ./systemd/cloud-init-generator.tmpl ./systemd/cloud-init-generator
./tools/benchmark.sh ./systemd/cloud-init-generator

real	0m0.210s
user	0m0.140s
sys	0m0.072s
root@cloudinit-0901-025531ijbm4hwd:~/cloud-init# make benchmark-generator 
python3 ./tools/render-cloudcfg --variant="benchmark" ./systemd/cloud-init-generator.tmpl ./systemd/cloud-init-generator
./tools/benchmark.sh ./systemd/cloud-init-generator

real	0m0.225s
user	0m0.162s
sys	0m0.068s
```

Generator Benchmark: tip of main with just benchmark commit applied (6cf53172a)
```
# make benchmark-generator 
python3 ./tools/render-cloudcfg --variant="benchmark" ./systemd/cloud-init-generator.tmpl ./systemd/cloud-init-generator
./tools/benchmark.sh ./systemd/cloud-init-generator

real	0m0.631s
user	0m0.361s
sys	0m0.268s
root@cloudinit-0901-025531ijbm4hwd:~/cloud-init# make benchmark-generator 
python3 ./tools/render-cloudcfg --variant="benchmark" ./systemd/cloud-init-generator.tmpl ./systemd/cloud-init-generator
./tools/benchmark.sh ./systemd/cloud-init-generator

real	0m0.710s
user	0m0.353s
sys	0m0.360s
``` 

#### systemd-analyze critical chain

*note: I think that the time is since systemd starts, but if a better methodology exists, feedback is welcome*  

with this change:
```
root@cloudinit-0901-025531ijbm4hwd:~# systemd-analyze critical-chain cloud-init-local.service
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

cloud-init-local.service +749ms
└─systemd-remount-fs.service @185ms +6ms
  └─systemd-journald.socket @176ms
    └─-.mount @147ms
      └─-.slice @147ms
```

from tip of main:
```
root@cloudinit-0901-025531ijbm4hwd:~# systemd-analyze critical-chain cloud-init-local.service 
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

cloud-init-local.service +749ms
└─systemd-remount-fs.service @185ms +6ms
  └─systemd-journald.socket @176ms
    └─-.mount @147ms
      └─-.slice @147ms
```

Since the root slice starts at 147ms for both tip of main and this branch, this change doesn't appear to provide a performance regression.

### Measuring with and without conditionals in .target

From a different (slower) machine (don't compare the times below with the times above), I compare this branch:
```
# systemd-analyze critical-chain cloud-init-local.service
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

cloud-init-local.service +884ms
└─systemd-journald.socket @299ms
  └─-.mount @219ms
    └─system.slice @219ms
      └─-.slice @219ms
```

against an otherwise identical branch that drops f310f9aa68e51636e3266c93b8152b53f6c6708d:
```
# systemd-analyze critical-chain cloud-init-local.service
The time when unit became active or started is printed after the "@" character.
The time the unit took to start is printed after the "+" character.

cloud-init-local.service +1.098s
└─systemd-journald.socket @316ms
  └─-.mount @243ms
    └─system.slice @242ms
      └─-.slice @243ms
```

I'd be surprised if this conditional to the target is actually an optimization that would save 0.2 seconds, so I assume that the measurable difference due to the target conditional is within the noise - not something to be concerned about.

## Test Steps

Systemd environment conditional test:
```
arc~ systemd-analyze condition 'ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled' 
test.service: ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled succeeded.
Conditions succeeded.
arc~ export KERNEL_CMDLINE=cloud-init=disabled                                           
arc~ systemd-analyze condition 'ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled'
test.service: ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled failed.
Conditions failed.
```

Integration Tests:
```
tox -e integration-tests -- tests/integration_tests/test_kernel_commandline_match.py
```

### DIFF
--- a/systemd/cloud-init-generator.tmpl
+++ b/systemd/cloud-init-generator.tmpl
@@ -64,8 +64,12 @@ main() {
     ds=$?
     debug 1 "ds-identify rc=$ds"
 
-    if [ "$ds" = "1" ]; then
-        debug 1 "cloud-init is enabled but no datasource found, disabling"
+    if [ "$ds" = "1" -o "$ds" = "2" ]; then
+        if [ "$ds" = "1" ]; then
+            debug 1 "cloud-init is enabled but no datasource found, disabling"
+        else
+            debug 1 "cloud-init is disabled by kernel commandline or etc_file"
+        fi
         if [ -f "$link_path" ]; then
             if rm -f "$link_path"; then
                 debug 1 "disabled. removed existing $link_path"
--- a/systemd/cloud-init-generator.tmpl
+++ b/systemd/cloud-init-generator.tmpl
@@ -42,105 +42,47 @@ debug() {
     echo "$@" >> "$LOG"
 }
 
-etc_file() {
-    local pprefix="${1:-/etc/cloud/cloud-init.}"
-    _RET="unset"
-    [ -f "${pprefix}$ENABLE" ] && _RET="$ENABLE" && return 0
-    [ -f "${pprefix}$DISABLE" ] && _RET="$DISABLE" && return 0
-    return 0
-}
-
-read_proc_cmdline() {
-    # return /proc/cmdline for non-container, and /proc/1/cmdline for container
-    local ctname="systemd"
-    if [ -n "$CONTAINER" ] && ctname=$CONTAINER ||
-        systemd-detect-virt --container --quiet; then
-        if { _RET=$(tr '\0' ' ' < /proc/1/cmdline); } 2>/dev/null; then
-            _RET_MSG="container[$ctname]: pid 1 cmdline"
-            return
-        fi
-        _RET=""
-        _RET_MSG="container[$ctname]: pid 1 cmdline not available"
-        return 0
-    fi
-
-    _RET_MSG="/proc/cmdline"
-    read _RET < /proc/cmdline
-}
-
-kernel_cmdline() {
-    local cmdline="" tok=""
-    if [ -n "${KERNEL_CMDLINE+x}" ]; then
-        # use KERNEL_CMDLINE if present in environment even if empty
-        cmdline=${KERNEL_CMDLINE}
-        debug 1 "kernel command line from env KERNEL_CMDLINE: $cmdline"
-    else
-        read_proc_cmdline && cmdline="$_RET" &&
-                debug 1 "kernel command line ($_RET_MSG): $cmdline"
-    fi
-    _RET="unset"
-    cmdline=" $cmdline "
-    tok=${cmdline##* cloud-init=}
-    [ "$tok" = "$cmdline" ] && _RET="unset"
-    tok=${tok%% *}
-    [ "$tok" = "$ENABLE" -o "$tok" = "$DISABLE" ] && _RET="$tok"
-    return 0
-}
-
-default() {
-    _RET="$ENABLE"
-}
-
-check_for_datasource() {
-    local ds_rc=""
-    if [ ! -x "$dsidentify" ]; then
-        debug 1 "no ds-identify in $dsidentify"
-        return 0
-    fi
-    $dsidentify
-    ds_rc=$?
-    debug 1 "ds-identify rc=$ds_rc"
-    if [ "$ds_rc" = "0" ]; then
-        return 0
-    fi
-    return 1
-}
-
 main() {
     local normal_d="$1" early_d="$2" late_d="$3"
     local target_name="multi-user.target" gen_d="$early_d"
     local link_path="$gen_d/${target_name}.wants/${CLOUD_TARGET_NAME}"
-    local ds=""
+    local ds="" ret=""
 
     debug 1 "$0 normal=$normal_d early=$early_d late=$late_d"
     debug 2 "$0 $*"
 
-    local search result="error" ret=""
-    for search in kernel_cmdline etc_file default; do
-        if $search; then
-            debug 1 "$search found $_RET"
-            [ "$_RET" = "$ENABLE" -o "$_RET" = "$DISABLE" ] &&
-                result=$_RET && break
-        else
-            ret=$?
-            debug 0 "search $search returned $ret"
-        fi
-    done
+    # ds=found => enable
+    # ds=notfound => disable
+    # <any> => disable
+    debug 1 "checking for datasource"
 
-    # enable AND ds=found == enable
-    # enable AND ds=notfound == disable
-    # disable || <any> == disabled
-    if [ "$result" = "$ENABLE" ]; then
-        debug 1 "checking for datasource"
-        check_for_datasource
-        ds=$?
-        if [ "$ds" = "1" ]; then
-            debug 1 "cloud-init is enabled but no datasource found, disabling"
-            result="$DISABLE"
-        fi
+    if [ ! -x "$dsidentify" ]; then
+        debug 1 "no ds-identify in $dsidentify"
+        ds=0
     fi
+    $dsidentify
+    ds=$?
+    debug 1 "ds-identify rc=$ds"
 
-    if [ "$result" = "$ENABLE" ]; then
+    if [ "$ds" = "1" ]; then
+        debug 1 "cloud-init is enabled but no datasource found, disabling"
+        if [ -f "$link_path" ]; then
+            if rm -f "$link_path"; then
+                debug 1 "disabled. removed existing $link_path"
+            else
+                ret=$?
+                debug 0 "[$ret] disable failed, remove $link_path"
+            fi
+        else
+            debug 1 "already disabled: no change needed [no $link_path]"
+        fi
+        if [ -e "$RUN_ENABLED_FILE" ]; then
+            debug 1 "removing $RUN_ENABLED_FILE and creating $RUN_DISABLED_FILE"
+            rm -f "$RUN_ENABLED_FILE"
+        fi
+        : > "$RUN_DISABLED_FILE"
+
+    elif [ "$ds" = "0" ]; then
         if [ -e "$link_path" ]; then
                 debug 1 "already enabled: no change needed"
         else
@@ -159,22 +101,6 @@ main() {
             rm -f $RUN_DISABLED_FILE
         fi
         : > "$RUN_ENABLED_FILE"
-    elif [ "$result" = "$DISABLE" ]; then
-        if [ -f "$link_path" ]; then
-            if rm -f "$link_path"; then
-                debug 1 "disabled. removed existing $link_path"
-            else
-                ret=$?
-                debug 0 "[$ret] disable failed, remove $link_path"
-            fi
-        else
-            debug 1 "already disabled: no change needed [no $link_path]"
-        fi
-        if [ -e "$RUN_ENABLED_FILE" ]; then
-            debug 1 "removing $RUN_ENABLED_FILE and creating $RUN_DISABLED_FILE"
-            rm -f "$RUN_ENABLED_FILE"
-        fi
-        : > "$RUN_DISABLED_FILE"
     else
         debug 0 "unexpected result '$result' 'ds=$ds'"
         ret=3
--- a/systemd/cloud-init-generator.tmpl
+++ b/systemd/cloud-init-generator.tmpl
@@ -23,6 +23,8 @@ CLOUD_SYSTEM_TARGET="/lib/systemd/system/cloud-init.target"
 {% if variant in ["almalinux", "centos", "cloudlinux", "eurolinux", "fedora",
                   "miraclelinux", "openeuler", "OpenCloudOS", "openmandriva", "rhel", "rocky", "TencentOS", "virtuozzo"] %}
     dsidentify="/usr/libexec/cloud-init/ds-identify"
+{% elif variant == "benchmark" %}
+    dsidentify="/bin/true"
 {% else %}
     dsidentify="/usr/lib/cloud-init/ds-identify"
 {% endif %}
--- a/systemd/cloud-init.target
+++ b/systemd/cloud-init.target
@@ -10,3 +10,6 @@
 [Unit]
 Description=Cloud-init target
 After=multi-user.target
+ConditionPathExists=!/etc/cloud/cloud-init.disabled
+ConditionKernelCommandLine=!cloud-init=disabled
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
--- a/tests/integration_tests/test_kernel_commandline_match.py
+++ b/tests/integration_tests/test_kernel_commandline_match.py
@@ -11,7 +11,24 @@ from tests.integration_tests.util import wait_for_cloud_init
 log = logging.getLogger("integration_testing")
 
 
-def override_kernel_cmdline(ds_str: str, c: IntegrationInstance) -> str:
+def restart_cloud_init(c):
+    client = c
+    client.instance.shutdown(wait=False)
+    try:
+        client.instance.wait_for_state("STOPPED", num_retries=20)
+    except RuntimeError as e:
+        log.warning(
+            "Retrying shutdown due to timeout on initial shutdown request %s",
+            str(e),
+        )
+        client.instance.shutdown()
+
+    client.instance.execute_via_ssh = False
+    client.instance.start()
+    client.execute("cloud-init status --wait")
+
+
+def override_kernel_cmdline(ds_str: str, c: IntegrationInstance):
     """
     Configure grub's kernel command line to tell cloud-init to use OpenStack
     - even though LXD should naturally be detected.
@@ -44,20 +61,7 @@ def override_kernel_cmdline(ds_str: str, c: IntegrationInstance) -> str:
     # most likely be as simple as updating the output path for grub-mkconfig
     client.execute("grub-mkconfig -o /boot/efi/EFI/ubuntu/grub.cfg")
     client.execute("cloud-init clean --logs")
-    client.instance.shutdown(wait=False)
-    try:
-        client.instance.wait_for_state("STOPPED", num_retries=20)
-    except RuntimeError as e:
-        log.warning(
-            "Retrying shutdown due to timeout on initial shutdown request %s",
-            str(e),
-        )
-        client.instance.shutdown()
-
-    client.instance.execute_via_ssh = False
-    client.instance.start()
-    client.execute("cloud-init status --wait")
-    return client.execute("cat /var/log/cloud-init.log")
+    restart_cloud_init(client)
 
 
 @pytest.mark.skipif(PLATFORM != "lxd_vm", reason="Modifies grub config")
@@ -85,10 +89,11 @@ def test_lxd_datasource_kernel_override(
     kernel commandline in Python code is required.
     """
 
+    override_kernel_cmdline(ds_str, client)
     assert (
         "Machine is configured by the kernel commandline to run on single "
         f"datasource {configured}"
-    ) in override_kernel_cmdline(ds_str, client)
+    ) in client.execute("cat /var/log/cloud-init.log")
 
 
 GH_REPO_PATH = "https://raw.githubusercontent.com/canonical/cloud-init/main/"
@@ -129,7 +134,9 @@ def test_lxd_datasource_kernel_override_nocloud_net(
             client.install_new_cloud_init(
                 source, take_snapshot=False, clean=False
             )
-        logs = override_kernel_cmdline(ds_str, client)
+        override_kernel_cmdline(ds_str, client)
+
+        logs = client.execute("cat /var/log/cloud-init.log")
         assert (
             "nocloud"
             == client.execute("cloud-init query platform").stdout.strip()
@@ -139,3 +146,40 @@ def test_lxd_datasource_kernel_override_nocloud_net(
             "Detected platform: DataSourceNoCloudNet [seed=None]"
             "[dsmode=net]. Checking for active instance data"
         ) in logs
+
+
+@pytest.mark.skipif(PLATFORM != "lxd_vm", reason="Modifies grub config")
+@pytest.mark.lxd_use_exec
+def test_lxd_disable_cloud_init_cmdline(client: IntegrationInstance):
+    """Verify cloud-init disablement via kernel commandline works."""
+
+    override_kernel_cmdline("cloud-init=disabled", client)
+    assert "Active: inactive (dead)" in client.execute(
+        "systemctl status cloud-init"
+    )
+
+
+@pytest.mark.lxd_use_exec
+def test_lxd_disable_cloud_init_file(client: IntegrationInstance):
+    """Verify cloud-init disablement via file works."""
+
+    client.execute("touch /etc/cloud/cloud-init.disabled")
+    client.execute("cloud-init --clean")
+    restart_cloud_init(client)
+    assert "Active: inactive (dead)" in client.execute(
+        "systemctl status cloud-init"
+    )
+
+
+@pytest.mark.lxd_use_exec
+def test_lxd_disable_cloud_init_env(client: IntegrationInstance):
+    """Verify cloud-init disablement via environment variable works."""
+    env = """DefaultEnvironment=KERNEL_CMDLINE=cloud-init=disabled"""
+
+    client.execute(f'echo "{env}" >> /etc/systemd/system.conf')
+
+    client.execute("cloud-init --clean")
+    restart_cloud_init(client)
+    assert "Active: inactive (dead)" in client.execute(
+        "systemctl status cloud-init"
+    )
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ render-template:
 # Our generator is a shell script. Make it easy to measure the
 # generator. This should be monitored for performance regressions
 benchmark-generator: FILE=$(GENERATOR_F).tmpl
+benchmark-generator: VARIANT="benchmark"
 benchmark-generator: export ITER=$(NUM_ITER)
 benchmark-generator: render-template
 	$(BENCHMARK) $(GENERATOR_F)
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -43,6 +43,7 @@ class UXAppBootStatusCode(enum.Enum):
     DISABLED_BY_GENERATOR = "disabled-by-generator"
     DISABLED_BY_KERNEL_CMDLINE = "disabled-by-kernel-cmdline"
     DISABLED_BY_MARKER_FILE = "disabled-by-marker-file"
+    DISABLED_BY_ENV_VARIABLE = "disabled-by-environment-variable"
     ENABLED_BY_GENERATOR = "enabled-by-generator"
     ENABLED_BY_KERNEL_CMDLINE = "enabled-by-kernel-cmdline"
     ENABLED_BY_SYSVINIT = "enabled-by-sysvinit"
@@ -54,6 +55,7 @@ DISABLED_BOOT_CODES = frozenset(
         UXAppBootStatusCode.DISABLED_BY_GENERATOR,
         UXAppBootStatusCode.DISABLED_BY_KERNEL_CMDLINE,
         UXAppBootStatusCode.DISABLED_BY_MARKER_FILE,
+        UXAppBootStatusCode.DISABLED_BY_ENV_VARIABLE,
     ]
 )
 
@@ -184,6 +186,16 @@ def get_bootstatus(disable_file, paths) -> Tuple[UXAppBootStatusCode, str]:
     elif "cloud-init=disabled" in cmdline_parts:
         bootstatus_code = UXAppBootStatusCode.DISABLED_BY_KERNEL_CMDLINE
         reason = "Cloud-init disabled by kernel parameter cloud-init=disabled"
+    elif "cloud-init=disabled" in os.environ.get("KERNEL_CMDLINE", "") or (
+        uses_systemd()
+        and "cloud-init=disabled"
+        in subp.subp(["systemctl", "show-environment"]).stdout
+    ):
+        bootstatus_code = UXAppBootStatusCode.DISABLED_BY_ENV_VARIABLE
+        reason = (
+            "Cloud-init disabled by environment variable "
+            "KERNEL_CMDLINE=cloud-init=disabled"
+        )
     elif os.path.exists(os.path.join(paths.run_dir, "disabled")):
         bootstatus_code = UXAppBootStatusCode.DISABLED_BY_GENERATOR
         reason = "Cloud-init disabled by cloud-init-generator"
--- a/doc/rtd/howto/disable_cloud_init.rst
+++ b/doc/rtd/howto/disable_cloud_init.rst
@@ -6,7 +6,7 @@ How to disable cloud-init
 One may wish to disable cloud-init to ensure that it doesn't do anything on
 subsequent boots. Some parts of cloud-init may run once per boot otherwise.
 
-There are two cross-platform methods of disabling ``cloud-init``.
+There are three cross-platform methods of disabling ``cloud-init``.
 
 Method 1: text file
 ====================
@@ -34,6 +34,15 @@ Example (using GRUB2 with Ubuntu):
     $ echo 'GRUB_CMDLINE_LINUX="cloud-init=disabled"' >> /etc/default/grub
     $ grub-mkconfig -o /boot/efi/EFI/ubuntu/grub.cfg
 
-.. note::
-   When running in containers, ``cloud-init`` will read an environment
-   variable named ``KERNEL_CMDLINE`` in place of a kernel commandline.
+Method 3: environment variable
+==============================
+
+To disable cloud-init, pass the environment variable
+``KERNEL_CMDLINE=cloud-init=disabled`` into each of cloud-init's
+processes.
+
+Example (using systemd):
+
+.. code-block::
+
+    $ echo "DefaultEnvironment=KERNEL_CMDLINE=cloud-init=disabled" >> /etc/systemd/system.conf
--- a/systemd/cloud-config.service.tmpl
+++ b/systemd/cloud-config.service.tmpl
@@ -5,10 +5,9 @@ After=network-online.target cloud-config.target
 After=snapd.seeded.service
 Before=systemd-user-sessions.service
 Wants=network-online.target cloud-config.target
-{% if variant == "rhel" %}
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
-{% endif %}
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
 [Service]
 Type=oneshot
--- a/systemd/cloud-final.service.tmpl
+++ b/systemd/cloud-final.service.tmpl
@@ -7,10 +7,9 @@ After=multi-user.target
 Before=apt-daily.service
 {% endif %}
 Wants=network-online.target cloud-config.service
-{% if variant == "rhel" %}
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
-{% endif %}
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
 
 [Service]
--- a/systemd/cloud-init-local.service.tmpl
+++ b/systemd/cloud-init-local.service.tmpl
@@ -26,10 +26,9 @@ Before=sysinit.target
 Conflicts=shutdown.target
 {% endif %}
 RequiresMountsFor=/var/lib/cloud
-{% if variant == "rhel" %}
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
-{% endif %}
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
 [Service]
 Type=oneshot
--- a/systemd/cloud-init.service.tmpl
+++ b/systemd/cloud-init.service.tmpl
@@ -38,10 +38,9 @@ Conflicts=shutdown.target
 Before=shutdown.target
 Conflicts=shutdown.target
 {% endif %}
-{% if variant == "rhel" %}
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
-{% endif %}
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
 [Service]
 Type=oneshot
--- a/tests/unittests/cmd/test_status.py
+++ b/tests/unittests/cmd/test_status.py
@@ -164,16 +164,26 @@ class TestStatus:
     ):
         if ensured_file is not None:
             ensure_file(ensured_file(config))
-        (code, reason) = wrap_and_call(
-            M_NAME,
-            {
-                "uses_systemd": uses_systemd,
-                "get_cmdline": get_cmdline,
-            },
-            status.get_bootstatus,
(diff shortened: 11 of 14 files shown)


holmanb:
@smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from a few people directly.

A few specific things I'd like to draw attention to:

1. With this change, cloud-init now prioritizes the [currently documented methods of disabling cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) over the ds-identify policy in all distros (formerly just RHEL) via systemd's conditional unit execution - is this acceptable behavior change? This should allow existing ds-identify policy to still work, but be overridden in the case that a user disables cloud-init as described in the docs. Are there any use cases that would be broken by this?
2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed. 
3. The new environment variable conditional `ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled`.
4. The changes in `cloud-init-generator.tmpl`

ani-sinha:
> @smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from you directly on this pull request.
> 
> A few specific things I'd like to draw attention to:
> 
> 1. With this change, cloud-init now prioritizes the [currently documented methods of disabling cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) over the ds-identify policy in all distros (formerly just RHEL) via systemd's conditional unit execution - is this acceptable behavior change? This should allow existing ds-identify policy to still work, but be overridden in the case that a user disables cloud-init as described in the docs. Are there any use cases that would be broken by this?
> 2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.

This needs to be tested. We are going to test this internally.

> 3. The new environment variable conditional `ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled`.
> 4. The changes in `cloud-init-generator.tmpl`

raharper:
> @smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from you directly on this pull request.
>
> A few specific things I'd like to draw attention to:
>
>     1. With this change, cloud-init now prioritizes the [currently
>     documented methods of disabling
>     cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html)
>     over the ds-identify policy in all distros (formerly just RHEL) via
>     systemd's conditional unit execution - is this acceptable behavior

Thanks for looking into this; it was certainly non-obvious why adding
documented parameters didn't appear to disable cloud-init in some scenarios.

In the open issue w.r.t Azure+Packer -- I think we've come to the conclusion
(though yet to be verified) that *something* external to cloud-init is the
cause of additional invocations of cloud-init, rather than that the documented
method does not work.

The documentation _works_ AFAICT in the absence of that packer scenario:

Using the touch method, for example:

```
# ls -al /etc/cloud/cloud-init.disabled
-rw-r--r-- 1 root root 0 Sep  1 18:47 /etc/cloud/cloud-init.disabled

# systemctl status cloud-init.target
○ cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; static)
     Active: inactive (dead)

# cat /run/cloud-init/cloud-init-generator.log
/usr/lib/systemd/system-generators/cloud-init-generator normal=/run/systemd/generator early=/run/systemd/generator.early late=/run/systemd/generator.late
kernel command line (container[lxc]: pid 1 cmdline): /sbin/init
kernel_cmdline found unset
etc_file found disabled
already disabled: no change needed [no /run/systemd/generator.early/multi-user.target.wants/cloud-init.target]
```

>     change? This should allow existing ds-identify policy to still work, but
>     be overridden in the case that a user disables cloud-init as described

Yes.

>     in the docs. Are there any use cases that would be broken by this?

I don't think so.  The policy is primarily about controlling *when* cloud-init
should search for datasources.  In the default Ubuntu cloud-image, the
DataSource list has many entries; some of these (like say EC2, but others)
require a DHCP on an interface to find a metadata service -- ds-identify and
the policy control whether an image will attempt to search out on potentially
untrusted networks for something that *looks* like a known datasource.


>
>     2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.

Generally, the additional conditions in the Unit file is a workaround for not
utilzing ds-identify.

Historically there were some Distros which moved to cloud-init releases with
ds-identify but did utilize it for reasons of their choosing.

IIUC, all of the systemd files only become added as systemd Jobs if
cloud-init.target does, which should mean it's far simpler to write the
conditions in *one* place (the target)  rather than maintain it in four
places.

Distro variants may yet still want to render in the conditions into each unit;
but as an upstream, I believe it's fine to keep them in the target.

>
>     3. The new environment variable conditional `ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled`.

+1

>
>     4. The changes in `cloud-init-generator.tmpl`

+1

Generally making the generator take less time is valuable since
`systemctl daemon-reload` happens far more than many realize;
keeping cloud-init's generator as light as possible is good policy.

ani-sinha:
> > @smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from you directly on this pull request.
> > A few specific things I'd like to draw attention to:
> > 
> > 1. With this change, cloud-init now prioritizes the [currently documented methods of disabling cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) over the ds-identify policy in all distros (formerly just RHEL) via systemd's conditional unit execution - is this acceptable behavior change? This should allow existing ds-identify policy to still work, but be overridden in the case that a user disables cloud-init as described in the docs. Are there any use cases that would be broken by this?
> > 2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.
> 
> This needs to be tested. We are going to test this internally.

This change does not work.
1. When touch file /etc/cloud/cloud-init.disabled, the cloud-init status is disabled, but the four cloud-init services are active, and cloud-init still runs
2. It is same when config kernel commandline per doc https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html.
Btw, the command in the doc is wrong, it should be:
`# echo 'GRUB_CMDLINE_LINUX=”cloud-init=disable”’ >> /etc/default/grub`

ani-sinha:
> 

For the doc issue I sent the PR https://github.com/canonical/cloud-init/pull/4406 .

raharper:
> > > @smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from you directly on this pull request.
> > > A few specific things I'd like to draw attention to:
> > > 
> > > 1. With this change, cloud-init now prioritizes the [currently documented methods of disabling cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) over the ds-identify policy in all distros (formerly just RHEL) via systemd's conditional unit execution - is this acceptable behavior change? This should allow existing ds-identify policy to still work, but be overridden in the case that a user disables cloud-init as described in the docs. Are there any use cases that would be broken by this?
> > > 2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.
> > 
> > 
> > This needs to be tested. We are going to test this internally.
> 
> This change does not work. This is the feedback from our QE:
> 
>     1. When touch file /etc/cloud/cloud-init.disabled, the cloud-init status is disabled, but the four cloud-init services are active, and cloud-init still runs

Do you have cloud-init logs from this run? 

> 
>     2. It is same when config kernel commandline per doc https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html.
>        Btw, the command in the doc is wrong, it should be:
>        `# echo 'GRUB_CMDLINE_LINUX=”cloud-init=disable”’ >> /etc/default/grub`

You're suggesting that it  should be 'disable', present tense,   not 'disabled', past tense?

Where in the cloud-init repo does it check for value 'disable' versus 'disabled'?

https://github.com/search?q=repo%3Acanonical%2Fcloud-init+%27disable%27&type=code

suggests only the 'byobu' module looks for `disable` .

A longer look for "disable" has more hits[2] but none of them look like kernel command expects the present tense, 'disable'.


2. https://github.com/search?q=repo%3Acanonical%2Fcloud-init+%22disable%22&type=code&p=4

ani-sinha:
> > > > @smoser @raharper @ani-sinha @esposem Given the invasive nature this change, I am requesting feedback from you directly on this pull request.
> > > > A few specific things I'd like to draw attention to:
> > > > 
> > > > 1. With this change, cloud-init now prioritizes the [currently documented methods of disabling cloud-init](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html) over the ds-identify policy in all distros (formerly just RHEL) via systemd's conditional unit execution - is this acceptable behavior change? This should allow existing ds-identify policy to still work, but be overridden in the case that a user disables cloud-init as described in the docs. Are there any use cases that would be broken by this?
> > > > 2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.
> > > 
> > > 
> > > This needs to be tested. We are going to test this internally.
> > 
> > 
> > This change does not work. This is the feedback from our QE:
> > ```
> > 1. When touch file /etc/cloud/cloud-init.disabled, the cloud-init status is disabled, but the four cloud-init services are active, and cloud-init still runs
> > ```
> 
> Do you have cloud-init logs from this run?
> 
> > ```
> > 2. It is same when config kernel commandline per doc https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html.
> >    Btw, the command in the doc is wrong, it should be:
> >    `# echo 'GRUB_CMDLINE_LINUX=”cloud-init=disable”’ >> /etc/default/grub`
> > ```
> 
> You're suggesting that it should be 'disable', present tense, not 'disabled', past tense?
> 
> Where in the cloud-init repo does it check for value 'disable' versus 'disabled'?

You got it all wrong. Please see my PR https://github.com/canonical/cloud-init/pull/4406

raharper:
> You got it all wrong. Please see my PR #4406

Your comment quoted, 
```
Btw, the command in the doc is wrong, it should be:
# echo 'GRUB_CMDLINE_LINUX=”cloud-init=disable”’ >> /etc/default/grub
```
includes a typo `disable` which is misleading.

W.r.t to the other requests that are relevant to this PR, do you have any logs from those runs?

ani-sinha:
> > You got it all wrong. Please see my PR #4406
> 
> Your comment quoted,
> 
> ```
> Btw, the command in the doc is wrong, it should be:
> # echo 'GRUB_CMDLINE_LINUX=”cloud-init=disable”’ >> /etc/default/grub
> ```
> 
> includes a typo `disable` which is misleading.
> 
> W.r.t to the other requests that are relevant to this PR, do you have any logs from those runs?

https://people.redhat.com/~anisinha/1108859.txt

holmanb:
@raharper @ani-sinha Thanks for the feedback and discussion on this PR. It seems sentiment so far is that the changes are agreeable, with the exception being the relocation of the service conditionals to the target file. Regarding that:

>>     2. Is it safe to shift the conditional execution lines from the four .service files to cloud-init.target? This part of the change can be dropped if needed.
>
> Generally, the additional conditions in the Unit file is a workaround for not
utilzing ds-identify.
>
> Historically there were some Distros which moved to cloud-init releases with
ds-identify but did utilize it for reasons of their choosing.
>
> IIUC, all of the systemd files only become added as systemd Jobs if
cloud-init.target does, which should mean it's far simpler to write the
conditions in one place (the target) rather than maintain it in four
places.
>
> Distro variants may yet still want to render in the conditions into each unit;
but as an upstream, I believe it's fine to keep them in the target.

This was my understanding as well, but testing this now I see that the PR as it is fails to disable, as @ani-sinha reported. Re-adding these lines to the service files disables cloud-init as expected.

I'll re-add the conditions to the service files as distro-independent conditions.

It should be functionally equivalent, so I'll leave the conditions in the target file so that users can see why cloud-init is disabled in the output of `systemctl status cloud-init.target`.

#### Output with conditions in target file:
```
root@cloudinit-0906-163408azrr4c80:~# systemctl status cloud-init.target
○ cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; enabled-runtime; preset: enabled)
     Active: inactive (dead)
  Condition: start condition failed at Wed 2023-09-06 16:48:35 UTC; 5min ago

Sep 06 16:48:35 cloudinit-0906-163408azrr4c80 systemd[1]: cloud-init.target - Cloud-init target was skipp>
```

#### Output without conditions in target file:
```
root@cloudinit-0906-163408azrr4c80:~# systemctl status cloud-init.target
○ cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; enabled-runtime; preset: enabled)
     Active: inactive (dead)
```

raharper:
> I'll re-add the conditions to the service files as distro-independent conditions.
> 
> It should be functionally equivalent, so I'll leave the conditions in the target file so that users can see why cloud-init is disabled in the output of `systemctl status cloud-init.target`.

From what I recall, systems which did not use ds-identify would fully install the services; that is the generator was effectively disabled and cloud-init*.service always ran no matter what; which prompted the inclusion of the conditionals.
Like this:

https://gitlab.com/redhat/centos-stream/src/cloud-init/-/blob/c9s/packages/redhat/cloud-init.spec.in?ref_type=heads#L125

With this style of install, ds-identify/generator doesn't matter as *systemd* itself is going to run these units on its own (`systemctl enable` creates symlinks to /etc/systemd/system/multi-user.target.wants/*) ; in this case, conditionals *per* service unit make sense.

If a distro does not `systemctl enable` cloud-init services; and instead lets the generator dynamically enable cloud-init.target; then no units should run if generator doesn't link cloud-init.target to multi-user.target.wants.

A few things to confirm.  In @ani-sinha test, were the cloud-init units already activated (via the systemctl enable like in the specfile I link)?  If so, then I believe the cloud-init.target won't affect whether the cloud-init services run since in the multi-user.wants dir, it isn't *just* cloud-init.target, but rather all 4 services linked.  

To have cloud-init.target control the execution of the service units, cloud-init.target and _only_ cloud-init target can be enabled (dynamically, or statically).

That is, the upstream spec file can instead do:

```
systemctl enable cloud-init.target

```
Or allow the generator to check if cloud-init.target should be added to multi-user.target.

In either case here, if cloud-init.target conditionals fail, then the other cloud-init services should not run either.
I do have a doubt/concern; it is possible that once a generator enables a target (we add the cloud-init.target symlink to multi-user.target) it could mean systemd enqueues the other targets (per the [Install] wants cloud-init.target in each service file) before systemd evaluates whether the cloud-init.target is active (per the conditionals).

If that is the case, then the conditional tests will need to remain in the generator to avoid adding cloud-init.target to multi-user.target.wants to prevent the services from running.

holmanb:
> I do have a doubt/concern; it is possible that once a generator enables a target (we add the cloud-init.target symlink to multi-user.target) it could mean systemd enqueues the other targets (per the [Install] wants cloud-init.target in each service file) before systemd evaluates whether the cloud-init.target is active (per the conditionals).
> 
> If that is the case, then the conditional tests will need to remain in the generator to avoid adding cloud-init.target to multi-user.target.wants to prevent the services from running.

I don't think this is a concern. See the enabled cloud-init.target is disabled on reboot below due to the condition check.
```
root@cloudinit-0906-23040247wpwk2q:~# systemctl status cloud-init.target
● cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; enabled-runtime; preset: enabled)
     Active: active since Wed 2023-09-06 23:24:52 UTC; 1s ago
      Until: Wed 2023-09-06 23:24:52 UTC; 1s ago

Sep 06 23:24:52 cloudinit-0906-23040247wpwk2q systemd[1]: Reached target cloud-init.target - Cloud-init target.
root@cloudinit-0906-23040247wpwk2q:~# touch /etc/cloud/cloud-init.disabled
root@cloudinit-0906-23040247wpwk2q:~# cloud-init clean --logs --reboot
root@cloudinit-0906-23040247wpwk2q:~# 

Session terminated, killing shell... ...killed.
arc~/cloud-init-d(holmanb/do-less-in-generator|…) lxc shell cloudinit-0906-23040247wpwk2q
root@cloudinit-0906-23040247wpwk2q:~# systemctl status cloud-init.target
○ cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; static)
     Active: inactive (dead)
root@cloudinit-0906-23040247wpwk2q:~# cloud-init status --long
status: disabled
boot_status_code: disabled-by-marker-file
detail:
Cloud-init disabled by /etc/cloud/cloud-init.disabled
root@cloudinit-0906-23040247wpwk2q:~# systemctl status cloud-init
○ cloud-init.service - Initial cloud-init job (metadata service crawler)
     Loaded: loaded (/lib/systemd/system/cloud-init.service; enabled; preset: enabled)
     Active: inactive (dead)
root@cloudinit-0906-23040247wpwk2q:~# systemctl status cloud-init.target
○ cloud-init.target - Cloud-init target
     Loaded: loaded (/lib/systemd/system/cloud-init.target; static)
     Active: inactive (dead)
```

ani-sinha:
> A few things to confirm. In @ani-sinha test, were the cloud-init units already activated (via the systemctl enable like in the specfile I link)? I

Yes. That is correct. My changes are here:
https://gitlab.com/anisinha/cloud-init/-/commits/PR4399/

I think we need to keep the check on individual unit files. Otherwise, with only in `.target` once disabled, it did not get enabled again when the check failed.
If the unit test refectoring changes are not essential, can you please revert them and keep things as they are? Testing this PR multiple times is consuming dev and QE resources - something we can't afford right now. Thanks.

holmanb:
> I think we need to keep the check on individual unit files. 

This is my plan for now with this PR.

> If the unit test refectoring changes are not essential, can you please revert them and keep things as they are?

The failing unit tests are a result of fixing a bug that I uncovered while testing this code out:

Currently cloud-init supports [disabling cloud-init via environment variable `KERNEL_CMDLINE` on containers](https://github.com/canonical/cloud-init/blob/fd214a1243011275c5dffb92b481c235e4c7a1bf/systemd/cloud-init-generator.tmpl#L71), however this method is not captured in the [status subcommand's `get_bootstatus()`](https://github.com/canonical/cloud-init/blob/fd214a1243011275c5dffb92b481c235e4c7a1bf/cloudinit/cmd/status.py#L166), which means that if cloud-init is disabled via this method, a `cloud-init status --wait` will run indefinitely; this is unacceptable. Fixing this broke a couple of unit tests. I plan to fix shortly and then mark this PR as ready for review (no longer WIP).

> Testing this PR multiple times is consuming dev and QE resources - something we can't afford right now. Thanks.

@ani-sinha Many thanks for the engagement and feedback!

holmanb:
> Excellent! That was what I expecting but wasn't 100% sure. Thank you for testing.

@raharper Happy to, thanks for the feedback

holmanb (commented):
Leaving a few comments to help reviewers understand the generator changes.

systemd/cloud-init-generator.tmpl
```diff
@@ -40,105 +42,47 @@ debug() {
...
+    if [ ! -x "$dsidentify" ]; then
+        debug 1 "no ds-identify in $dsidentify"
+        ds=0
+    fi
+    $dsidentify
+    ds=$?
+    debug 1 "ds-identify rc=$ds"
+
+    if [ "$ds" = "1" ]; then
+        debug 1 "cloud-init is enabled but no datasource found, disabling"
```
antcodd, line 68: If a cloud-init is installed but disabled, the misleading message `cloud-init is enabled but no datasource found, disabling` will now be printed by the generator on boot, and ds-identify will output misleading messages attempting to detect clouds. 

ds-identify will likely write incorrect /run/cloud-init/cloud.cfg configuration as if it was enabling datasources. 

$RUN_ENABLED_FILE` (`/run/cloud-init/cloud-init.enabled`/`.disabled`) will also be set to the wrong state. I'm not sure if this file is used for anything since dhclient hooks were removed in #2051, but it could be used as an indicator for other external scripts. cloud-init.enabled also doesn't appear to be created when not using the generator.

antcodd:
Having the cloud-init package installed but disabled will become very expensive, as ds-identify will now still run and attempt to detect clouds. On every boot ds-identify will run expensive hardware DMI detection logic as a systemd generator that synchronously blocks all of boot. This performance impact will likely dwarf any small changes from checking a few flags on each boot.

Did the benchmarking include deleting the cloud-init cached configuration from /run, as it would not be present on boot? ds-identify skips cloud detection checks when the result is cached. Generators also run any time systemctl daemon-reload is triggered, though in that case ds-identify should at least use the cached result.

I think the changes to cloud-init-generator should be reverted, or ds-identify needs to check the disabled state flags before it tries to detect clouds. I also suspect there might be more to be gained by conditionally skipping some of the logic in collect_info from ds_identify than changes to the generator itself.

Another option might be to have cloud-init-generator (and ds-identify) just run as a unit before cloud-init-local.service and check the conditions as part of the unit file? Enqueuing additional services during boot using systemctl appears to work, so long as they aren't waited on. This would also allow ds-identify to read configuration from other filesystems. This wouldn't cover new cloud-init package installation having the generator run by systemctl daemon-reload though.

systemd/cloud-init-generator.tmpl
rjschwei: The message will only be printed if no datasource is found. This is still misleading, no question, but it is not unconditional. In the case of a  "happy path", i.e. dsidentify runs and finds a data source the message will not be printed.

That said if cloud-init is disabled, why run dsitentify at all?

systemd/cloud-init-generator.tmpl
```diff
@@ -42,105 +42,47 @@ debug() {
...
-        return 0
-    fi
-
-    _RET_MSG="/proc/cmdline"
-    read _RET < /proc/cmdline
-}
-
-kernel_cmdline() {
-    local cmdline="" tok=""
-    if [ -n "${KERNEL_CMDLINE+x}" ]; then
```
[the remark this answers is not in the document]
rjschwei: Yes and no. It is not handled from a generator perspective. If `KERNEL_CMDLINE=cloud-init=disabled` exists there is really nothing for the generator to do and it should just exit as systemd will handle the rest via the ConditionEnvironment setting. But this does not apply to the generator and as such dsidentify is still run, which we shouldn't really do if the user disabled cloud-init.

I think the generator should specifically look for `KERNEL_CMDLINE=cloud-init=disabled` and if found exit and do nothing else.

blackboxsw:
Thanks @holmanb for the refactors here and insights.

* I agree that we want to avoid calling systemd-detect-virt multiple time in systemd-generator timeframe.  And reducing some of the calls in cloud-init-generator.tmpl will cut down on the 2 separate `systemd-detect-virt` calls that cloud-init-generator and ds-identify make in happy path case.
* I also agree we should be defensive in both `cloud-init.target` and `cloud-*.service` files to ensure that platforms which only deliver the systemd units instead of the cloud-init.target are able to effectively see those units inactive/disabled if neither ds-identify, nor cloud-init.target are present on the system. Thanks @raharper and @ani-sinha for confirmation about CentOS setup there.

Reducing the number of `systemd-detect-virt` calls as you've done here is good but I share some of the same concerns @rjschwei and @antcodd have raised that we want to NOOP as early as possible in systemd-generator timeframe when we know cloud-init is disabled.  The use-cases I can think of are Live server and live desktop environments which perform installations with cloud-init and leave cloud-init in a disabled state after the post-install system first boot with the `/etc/cloud/cloud-init.disabled` flag file. We don't want those physical systems always re-reading DMI information at boot on every boot when we know it is declared disabled.

What I suggest is the following:
  Since cloud-init-generator and ds-identify both attempt to process the kernel cmdline, and ds-identify needs to process the kernel commandline for much more than just whether cloud-init is disabled, I think it makes sense to avoid calling `systemd-detect-virt` in the generator script. Instead, we can extend ds-identify kernel cmdline parsing it already does to also check for `cloud-init=disabled` state and the /etc/cloud/cloud-init.disabled flag file.

Per @antcodd's suggestion, ds-identify can perform it's initial disabled check earlier than DMI reads. I think we can do that with with a different exit code(2) that will also resolve @antcodd's other concern about misleading messages coming from the generator, as the generator can check exit code 2 and report "cloud-init disabled due to kernel commandline or etc_file".


Here's a suggested proposal for discussion, what his does is gives ds-identify a quick exit when disabled, leaves the generator "simple" without the duplicated kernel cmdline parsing based on container/non-container environments and give avoid seeding systemd units/services in a system when cloud-init disabled by the environment.

```diff
commit b6c7606ec962defdedf29adf0ac369d02bb53fe1 (HEAD)
Author: Chad Smith <chad.smith@canonical.com>
Date:   Mon Sep 25 15:21:49 2023 -0600

    still check disabled in ds-id

diff --git a/systemd/cloud-init-generator.tmpl b/systemd/cloud-init-generator.tmpl
index 5999fc5d2..3c9ca1695 100644
--- a/systemd/cloud-init-generator.tmpl
+++ b/systemd/cloud-init-generator.tmpl
@@ -64,8 +64,12 @@ main() {
     ds=$?
     debug 1 "ds-identify rc=$ds"
 
-    if [ "$ds" = "1" ]; then
-        debug 1 "cloud-init is enabled but no datasource found, disabling"
+    if [ "$ds" = "1" -o "$ds" = "2" ]; then
+        if [ "$ds" = "1" ]; then
+            debug 1 "cloud-init is enabled but no datasource found, disabling"
+        else
+            debug 1 "cloud-init is disabled by kernel commandline or etc_file"
+        fi
         if [ -f "$link_path" ]; then
             if rm -f "$link_path"; then
                 debug 1 "disabled. removed existing $link_path"
diff --git a/tests/unittests/test_ds_identify.py b/tests/unittests/test_ds_identify.py
index 66dae60e4..200c488c6 100644
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -329,6 +329,10 @@ class DsIdentifyBase(CiTestCase):
                 "ret": 1,
                 "err": "No dmidecode program. ERROR.",
             },
+            {
+                "name": "is_disabled",
+                "ret": 1,
+            },
             {
                 "name": "get_kenv_field",
                 "ret": 1,
diff --git a/tools/ds-identify b/tools/ds-identify
index 8b5229d8e..babd1f43b 100755
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1039,6 +1039,23 @@ has_ovf_cdrom() {
     return 1
 }
 
+is_disabled() {
+    if [ -f /etc/cloud/cloud-init.disabled ]; then
+        debug 1 "disabled by marker file /etc/cloud-init.disabled"
+        return 0
+    fi
+    if [ "${KERNEL_CMDLINE:-}" = "cloud-init=disabled" ]; then
+        debug 1 "disabled by KERNEL_CMDLINE environment variable"
+        return 0
+    fi
+    case "$DI_KERNEL_CMDLINE" in
+       *cloud-init=disabled*)
+            debug 1 "disabled by kernel command line cloud-init=disabled"
+            return 0
+    esac
+    return 1
+}
+
 dscheck_OVF() {
     check_seed_dir ovf ovf-env.xml && return "${DS_FOUND}"
 
@@ -1517,10 +1534,7 @@ dscheck_VMware() {
 }
 
 collect_info() {
-    read_uname_info
-    read_virt
     read_pid1_product_name
-    read_kernel_cmdline
     read_config
     read_datasource_list
     read_dmi_sys_vendor
@@ -1795,6 +1809,12 @@ _main() {
 
     read_uptime
     debug 1 "[up ${_RET}s]" "ds-identify $*"
+    read_uname_info
+    read_virt
+    read_kernel_cmdline
+    if is_disabled; then
+        return 2
+    fi
     collect_info
 
     if [ "$DI_LOG" = "stderr" ]; then
```

blackboxsw:
Also note on Ubuntu systems we are generally seeing 15 calls to `systemd-detect-virt` on boot when cloud-init is enabled. Since cloud-init generator + ds-identify perform two of those calls, it'd be good to reduce one of those calls to systemd-detect-virt if both the generator wrapper and ds-identify can rely on the same single call to detect-virt to determine disabled state from the proper kernel command line.  I can see with this changeset + diff we are down to 14 calls instead of 15. In cloud-init disabled case systemd-detect-virt is called 10 times.

systemd/cloud-init.target
```diff
@@ -10,3 +10,6 @@
 [Unit]
 Description=Cloud-init target
 After=multi-user.target
+ConditionPathExists=!/etc/cloud/cloud-init.disabled
+ConditionKernelCommandLine=!cloud-init=disabled
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
```
blackboxsw, line 15: I like these additional condition checks for visibility at systemctl status cloud-init.target level, but this duplicates conditions we've decided to keep on each subordinate `cloud-*.service` file and we don't strictly need them if we are going to preserve the checks on the individual units. Can we measure if there is a cost to these duplicated conditions? It does make systemd dependency resolver do a little but more as far as checks for this target unit file.

systemd/cloud-init-generator.tmpl
blackboxsw: Even though this text in /run/cloud-init/cloud-init-generator.log is not machine readable, and I don't think we have scripts parsing and relying on the output of this file, I'd like us to specialize that message when cloud-init is disabled. Then we better represent  the actual state during generator time-frame when we see cloud-init disabled and don't have to fallback to querying cloud-init status or systemctl status cloud-init.target etc to get that info.

blackboxsw (changes requested):
Marking request changes to discuss whether we can push disable awareness from kernel commandline in ds-identify since it has to parse kernel cmdline anyway in the case that cloud-init is enabled.

blackboxsw:
> Having the cloud-init package installed but disabled will become very expensive, as ds-identify will now still run and attempt to detect clouds. On every boot ds-identify will run expensive hardware DMI detection logic as a systemd generator that synchronously blocks all of boot. This performance impact will likely dwarf any small changes from checking a few flags on each boot.

Agreed, DMI reading is probably a fair amount of cost in ds-identify that we should probably avoid if we know we are disabled.
And, we could even go further and not even attempt to invoke systemd-detect-virt before parsing the kernel cmdline if we first check the /etc/cloud/cloud-config.disabled flag to save even more time in a disabled system.


> 
> Did the benchmarking include deleting the cloud-init cached configuration from /run, as it would not be present on boot? ds-identify skips cloud detection checks when the result is cached. Generators also run any time systemctl daemon-reload is triggered, though in that case ds-identify should at least use the cached result.
> 

Most of ds-identify's caching is really just checks on whether variables have been defined and if so, return the cached var name. It's not disk caching that we are talking about here.  The benchmarking is going to be highly platform dependent as it does go though and source DMI vars on the host in which the benchmark it is running


> I think the changes to cloud-init-generator should be reverted, or ds-identify needs to check the disabled state flags before it tries to detect clouds. I also suspect there might be more to be gained by conditionally skipping some of the logic in collect_info from ds_identify than changes to the generator itself.

I disagree with the first statement as long as we can conditionally skip logic down in `ds-identify.collect_info` if disabled artifacts are present.

systemd/cloud-init.target
holmanb: I'll measure and report back.

holmanb:
> What I suggest is the following: Since cloud-init-generator and ds-identify both attempt to process the kernel cmdline, and ds-identify needs to process the kernel commandline for much more than just whether cloud-init is disabled, I think it makes sense to avoid calling `systemd-detect-virt` in the generator script. Instead, we can extend ds-identify kernel cmdline parsing it already does to also check for `cloud-init=disabled` state and the /etc/cloud/cloud-init.disabled flag file.

This seems like a fair trade-off. Since ds-identify already gathers this information it shouldn't be a measurable extra cost to the non-disabled case, and a clear improvement for the disabled case. This is a reasonable tradeoff.

> Per @antcodd's suggestion, ds-identify can perform it's initial disabled check earlier than DMI reads. I think we can do that with with a different exit code(2) that will also resolve @antcodd's other concern about misleading messages coming from the generator, as the generator can check exit code 2 and report "cloud-init disabled due to kernel commandline or etc_file".

This is a slight UI change to `ds-identify`, but since it's a low level tool that isn't documented I would make the argument that such changes don't count as real behavior changes. Non-zero for an error condition makes sense.

> Here's a suggested proposal for discussion, what his does is gives ds-identify a quick exit when disabled, leaves the generator "simple" without the duplicated kernel cmdline parsing based on container/non-container environments and give avoid enabling systemd units/services in a system when cloud-init disabled by the environment.

@blackboxsw If you can push your branch up to Github or share the formatted patch (`git format-patch -1 HEAD --stdout`) I'll apply this change with authorship preserved, since you did write it and I think it fits this series nicely as its own commit (albeit with a more detailed message, but I can update that if you'd rather not).

blackboxsw:
> @blackboxsw If you can push your branch up to Github or share the formatted patch (`git format-patch -1 HEAD --stdout`) I'll apply this change with authorship preserved, since you did write it and I think it fits this series nicely as its own commit (albeit with a more detailed message, but I can update that if you'd rather not).


I've updated the inline patch above w/ a more specific commit msg and also pushed https://github.com/blackboxsw/cloud-init/tree/holmanb/do-less-in-generator

systemd/cloud-init-generator.tmpl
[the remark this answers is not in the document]
holmanb: This is addressed in the latest commit.

systemd/cloud-init.target
holmanb: @blackboxsw see the section in the PR description titled "Measuring with and without conditionals in .target" for the numbers

blackboxsw: Thanks @holmanb. This looks good to me as well as your assessment above. Also I triggered those compound conditionals directly on the systemd-analyze conditional subcommand with comparable results. The avg time seems to be +- one millisecond with or without the 3 extra checks.
```
root@schema-m:~# time systemd-analyze condition 'ConditionKernelCommandLine=!cloud-init=disabled'
test.service: ConditionKernelCommandLine=!cloud-init=disabled succeeded.
Conditions succeeded.

real	0m0.007s
user	0m0.000s
sys	0m0.007s
root@schema-m:~# time systemd-analyze condition 'ConditionKernelCommandLine=!cloud-init=disabled' 'ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled' 'ConditionPathExists=!/etc/cloud/cloud-init.disabled' 'ConditionKernelCommandLine=!cloud-init=disabled' 'ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled'
test.service: ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled succeeded.
test.service: ConditionKernelCommandLine=!cloud-init=disabled succeeded.
test.service: ConditionPathExists=!/etc/cloud/cloud-init.disabled succeeded.
test.service: ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled succeeded.
test.service: ConditionKernelCommandLine=!cloud-init=disabled succeeded.
Conditions succeeded.

real	0m0.007s
user	0m0.000s
sys	0m0.007s
```

tests/integration_tests/test_kernel_commandline_match.py
```diff
@@ -11,7 +11,24 @@
 log = logging.getLogger("integration_testing")
 
 
-def override_kernel_cmdline(ds_str: str, c: IntegrationInstance) -> str:
+def restart_cloud_init(c):
+    client = c
```
blackboxsw, line 15: nit:non-blocking why not just pass in a variable named client and use that.

blackboxsw (approved):
LGTM, minor nit on integration test but you don't have to hold up the PR on that. Also, please provide the `(#4399)` suffix to each commit if rebase merging.

rjschwei (approved):

tests/integration_tests/test_kernel_commandline_match.py
holmanb: since `client` is a fixture, I didn't want to use that name because 1) I don't know if a fixture would be invoked (non-test functions can get fixtures invoked, for example a fixture that invokes other fixtures) and 2) even if 1 doesn't occur, it might confuse a reader who might assume that it does. This is more clear in my mind, though I guess I could have just picked a different variable name.

PR #4400, opened 2023-08-31
integration tests: don't clean when KEEP_* flags true

TheRealFalcon:
## Proposed Commit Message

```
integration tests: don't clean when KEEP_* flags true

When bumping to pycloudlib version 1!5.0.0, we gained the ability to
clean anything created by the Cloud instance, not just the Instance
instance. This commit ensures we don't do this when KEEP_ settings
are true.
```


holmanb (approved):
Thanks for this fix. Looks good to me!

PR #4401, opened 2023-09-01
Fix typing issues in subp module

TheRealFalcon:
## Proposed Commit Message

```
Fix typing issues in subp module

Add the 'SubpResult' return type from subp and fix some associated
typing issues

Additionally convert to using f-strings, use if expressions rather
than if blocks where applicable, and enforce subp kwargs
```

## Additional Context
Adding `-> SubpResult` to subp's type signature wound up exposing other typing issues from mypy.

### DIFF
--- a/cloudinit/subp.py
+++ b/cloudinit/subp.py
@@ -134,19 +134,13 @@ class ProcessExecutionError(IOError):
     ) -> Union[str, bytes]:
         """
         indent text on all but the first line, allowing for easy to read output
+
+        remove any newlines at end of text first to prevent unneeded blank
+        line in output
         """
-        cr_str = "\n"
-        indent_str = " " * indent_level
-        # if input is bytes, return bytes
-        if isinstance(text, bytes):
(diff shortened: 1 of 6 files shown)


cloudinit/subp.py
```diff
@@ -141,23 +131,29 @@ def _ensure_string(self, text):
         """
         return text.decode() if isinstance(text, bytes) else text
 
-    def _indent_text(self, text, indent_level=8):
+    def _indent_text(
```
TheRealFalcon, line 134: I couldn't find an elegant way to fix this method

holmanb (approved):
See [this proposal](https://github.com/TheRealFalcon/cloud-init/pull/4) for a proposal of a more elegant `_indent_text()`. Feel free to incorporate it if you like it.

This is more readable, less spaghetti. Thanks for this cleanup!

cloudinit/subp.py
TheRealFalcon: Applied, thanks!

PR #4403, opened 2023-09-01
Fix pip-managed ansible on pip < 23.0.1

holmanb:
```
Fix pip-managed ansible on pip < 23.0.1

Commit b417b218 broke pip-managed ansible with
pip < 23.0.1 and fixed pip-managed ansible with
pip >= 23.0.1. Gate change by pip version to
correctly operate on both versions.
```

## Additional Context
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-lxd_vm/303/testReport/junit/tests.integration_tests.modules/test_ansible/test_ansible_controller/

### DIFF
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -136,32 +136,13 @@ class AnsiblePullPip(AnsiblePull):
                 import pip  # noqa: F401
             except ImportError:
                 self.distro.install_packages(self.distro.pip_package_name)
-
-            pip_version_output = self.do_as(
-                [
-                    sys.executable,
-                    "-m",
-                    "pip",
-                    "--version",
-                ]
-            ).stdout
-            pip_version = pip_version_output.split()
-
-            if 2 > len(pip_version):
-                LOG.warning(
-                    "Error parsing pip version string: %s", pip_version_output
-                )
-            version = Version.from_str(pip_version[1])
             cmd = [
                 sys.executable,
                 "-m",
                 "pip",
                 "install",
+                "--break-system-packages",
             ]
-
-            # see pip commit e6deb9b87c18cdd27a9ba27cb7e0670ffb81d45e
-            if version >= Version(23, 0, 1):
-                cmd.append("--break-system-packages")
             if self.run_user:
                 cmd.append("--user")
             self.do_as([*cmd, "--upgrade", "pip"])
--- a/cloudinit/config/cc_ansible.py
+++ b/cloudinit/config/cc_ansible.py
@@ -3,6 +3,7 @@ import abc
 import os
 import re
 import sys
+import sysconfig
 from copy import deepcopy
 from logging import getLogger
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


tests/unittests/util.py
```diff
@@ -153,7 +153,7 @@ def update_package_sources(self):
         return (True, "yay")
 
     def do_as(self, command, args=None, **kwargs):
-        return ("stdout", "stderr")
+        return subp.SubpResult("pip 23.0.1 from /usr/lib/python3/dist-packages/pip (python 3.11)", "stderr")
```
holmanb, line 156: Feels a little funny shoving test-specific data into the distro mock. Suggestions welcome.

cloudinit/config/cc_ansible.py
```diff
@@ -136,13 +136,32 @@ def install(self, pkg_name: str):
                 import pip  # noqa: F401
             except ImportError:
                 self.distro.install_packages(self.distro.pip_package_name)
+
+            pip_version_output = self.do_as(
```
blackboxsw, line 140: Instead of checking pip version to see if we can provide this output, shouldn't we instead be checking if the EXTERNALLY-MANAGED marker file exists in the first place? This could be our indicator that the system is implementing the PEP 668, and therefore probably also supports --break-system-packages param. This would also have the benefit on stable releases of no change in behavior if the distro itself doesn't package the marker file.

[Pip does this check like so](https://github.com/pypa/pip/blob/e6deb9b87c18cdd27a9ba27cb7e0670ffb81d45e/src/pip/_internal/utils/misc.py#L594C1-L595C1)  The additional benefit of doing a file presence check is that it avoids a more costly shell out to the cmdline.

blackboxsw (changes requested):
Hey @holmanb what o you think about an alternative approach to only append --break-system-packages if that EXTERNALLY-MANAGED marker file is present? This would work on all flavors of Ubuntu back to focal as we'd avoid adding the param unless the distro itself implements the PEP668 marker.

cloudinit/config/cc_ansible.py
holmanb: > Instead of checking pip version to see if we can provide this output, shouldn't we instead be checking if the EXTERNALLY-MANAGED marker file exists in the first place?

@blackboxsw This file is not a requirement, but rather a [recommendation for distros to include](https://peps.python.org/pep-0668/#recommendations-for-distros). This comes with distro risk. Your recommendation is cleaner, however, so I'm slightly leaning towards doing that anyway. Thoughts?

cloudinit/config/cc_ansible.py
```diff
@@ -141,8 +142,11 @@ def install(self, pkg_name: str):
                 "-m",
                 "pip",
                 "install",
-                "--break-system-packages",
             ]
+            if os.path.join(
+                sysconfig.get_path("stdlib"), "EXTERNALLY-MANAGED"
+            ):
```
blackboxsw, line 148: I think you meant the following:
```suggestion
            if os.path.exists(
                os.path.join(
                    sysconfig.get_path("stdlib"), "EXTERNALLY-MANAGED"
                )
            ):
```

blackboxsw (changes requested):
Ran through jammy and focal with the suggested changeset.

cloudinit/config/cc_ansible.py
holmanb: Yes, thanks!

holmanb:
Thanks for the review @blackboxsw. Fixed according to your comment.

blackboxsw (approved):
LGTM!

PR #4406, opened 2023-09-04
doc: fix instructions on how to disable cloud-init from kernel commadline

ani-sinha:
There was a typo in the instruction on disabling cloud-init from the kernel command line. The actual command is:
#echo 'GRUB_CMDLINE_LINUX="cloud-init=disabled"' >> /etc/default/grub

Instead of
#echo 'GRUB_CMDLINE_LINUX=cloud-init.disabled' >> /etc/default/grub

Fix it.


holmanb (approved):
Thanks, nice catch!

PR #4409, opened 2023-09-05
Ubuntu/lunar

TheRealFalcon:
I popped off the top commit and added these two. If approved, I'll force push upstream.

### DIFF
--- a/debian/cloud-init.maintscript
+++ b/debian/cloud-init.maintscript
@@ -6,4 +6,4 @@ rm_conffile /etc/init/cloud-init-local.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init-nonet.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-log-shutdown.conf 0.7.9-243-ge74d775-0ubuntu2~
-rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3-0ubuntu0~23.04.1~
+rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3.1-0ubuntu1~


debian/cloud-init.maintscript
```diff
@@ -6,4 +6,4 @@ rm_conffile /etc/init/cloud-init-local.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init-nonet.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-log-shutdown.conf 0.7.9-243-ge74d775-0ubuntu2~
-rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3-0ubuntu0~23.04.1~
+rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3-0ubuntu1~
```
TheRealFalcon, line 9: Now that we're going to SRU a 23.1.1, I'm not sure if this should be updated to `23.3.1-0ubuntu1~` 

In the dpkg-maintscript-helper manpage:
```
    If the conffile has not been shipped for several versions, and you are
    now modifying the maintainer scripts to clean up the obsolete file,
    prior-version should be based on the version of the package that you
    are now preparing, not the first version of the package that lacked the
    conffile
```

debian/changelog
```diff
@@ -1,4 +1,4 @@
-cloud-init (23.3-0ubuntu0~23.04.1) UNRELEASED; urgency=medium
+cloud-init (23.3-0ubuntu1~23.04.1) UNRELEASED; urgency=medium
```
blackboxsw, line 1: Let's drop this changeset as it will be replaced with 23.3.1-0ubuntu1~23.04.1 on our next SRU anyway

debian/cloud-init.maintscript
blackboxsw:
```suggestion
rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3.1-0ubuntu1~
```

blackboxsw: Agreed. we know we are SRUing 23.3.1 not 23.3 so let's plan for that

blackboxsw (changes requested):

TheRealFalcon:
Updated

blackboxsw (approved):

PR #4411, opened 2023-09-05
apt: kill dirmngr/gpg-agent without gpgconf dependency

blackboxsw:
## Proposed Commit Message

```
Upstream commit 842d0452 introduced a dependency on gpgconf
utility without a strict Depends: clause in debian/control.

While gnupg is in Recommends for cloud-init, some images
like Ubuntu minimal will not install the package Recommends
to keep images smaller.

To avoid adding a hard Requires dependency on gpgconf, use os.kill
directly to stop any dirmngr and gpg-agent when spawned by
cloud-config's root ppid due to apt-related user-data.

NOOP If no gpg-agent or dirmngr processes are spawned by PPID 1.

LP: #2034273
```

## Additional Context

As this represents a regression in behavior for Ubuntu minimal images and possibly Debian images without gpgconf, this is going to require a hotfix release of 23.3.1

## Test Steps
```
# Generate local deb from this branch

# test using local deb
CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_23.3-11-gce9b83b2-1~bddeb_all.deb CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_OS_IMAGE=mantic .tox/integration-tests/bin/pytest  tests/integration_tests/bugs/test_lp1813396.py
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -12,6 +12,7 @@ import glob
 import os
 import pathlib
 import re
+import signal
 from textwrap import dedent
 
 from cloudinit import gpg
@@ -245,23 +246,17 @@ def apply_apt(cfg, cloud, target):
     # GH: 4344 - stop gpg-agent/dirmgr daemons spawned by gpg key imports.
     # Daemons spawned by cloud-config.service on systemd v253 report (running)
     gpg_process_out, _err = subp.subp(
-        ["ps", "-o", "pid,args", "-C", "dirmngr", "-C", "gpg-agent"],
+        ["ps", "-o", "ppid,pid", "-C", "dirmngr", "-C", "gpg-agent"],
         target=target,
         capture=True,
         rcs=[0, 1],
     )
-    gpg_pids = re.findall(r"(?P<pid>\d+).*homedir \/root", gpg_process_out)
-    if len(gpg_pids) <= 2:
-        if gpg_pids:
-            LOG.debug("Killing gpg-agent and dirmngr pids: %s", gpg_pids)
-        for gpg_pid in gpg_pids:
-            os.kill(int(gpg_pid))
-    else:
-        LOG.debug(
-            "Leaving all gpg-agent and dirmngr daemons running. Greater than 2"
-            " active daemons present: %s",
-            gpg_process_out,
-        )
+    gpg_pids = re.findall(r"(?P<ppid>\d+)\s+(?P<pid>\d+)", gpg_process_out)
+    root_gpg_pids = [int(pid[1]) for pid in gpg_pids if pid[0] == "1"]
+    if root_gpg_pids:
+        LOG.debug("Killing gpg-agent and dirmngr pids: %s", root_gpg_pids)
+    for gpg_pid in root_gpg_pids:
+        os.kill(gpg_pid, signal.SIGKILL)
 
 
 def debconf_set_selections(selections, target=None):
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -244,11 +244,24 @@ def apply_apt(cfg, cloud, target):
         )
     # GH: 4344 - stop gpg-agent/dirmgr daemons spawned by gpg key imports.
     # Daemons spawned by cloud-config.service on systemd v253 report (running)
-    subp.subp(
-        ["gpgconf", "--kill", "all"],
+    gpg_process_out, _err = subp.subp(
+        ["ps", "-o", "pid,args", "-C", "dirmngr", "-C", "gpg-agent"],
         target=target,
         capture=True,
+        rcs=[0, 1],
     )
+    gpg_pids = re.findall(r"(?P<pid>\d+).*homedir \/root", gpg_process_out)
+    if len(gpg_pids) <= 2:
+        if gpg_pids:
+            LOG.debug("Killing gpg-agent and dirmngr pids: %s", gpg_pids)
+        for gpg_pid in gpg_pids:
+            os.kill(int(gpg_pid))
+    else:
+        LOG.debug(
+            "Leaving all gpg-agent and dirmngr daemons running. Greater than 2"
+            " active daemons present: %s",
+            gpg_process_out,
+        )
 
 
 def debconf_set_selections(selections, target=None):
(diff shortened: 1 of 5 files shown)


cloudinit/config/cc_apt_configure.py
```diff
@@ -244,11 +244,24 @@ def apply_apt(cfg, cloud, target):
...
     # Daemons spawned by cloud-config.service on systemd v253 report (running)
-    subp.subp(
-        ["gpgconf", "--kill", "all"],
+    gpg_process_out, _err = subp.subp(
+        ["ps", "-o", "pid,args", "-C", "dirmngr", "-C", "gpg-agent"],
         target=target,
         capture=True,
+        rcs=[0, 1],
     )
+    gpg_pids = re.findall(r"(?P<pid>\d+).*homedir \/root", gpg_process_out)
```
TheRealFalcon, line 253: I think that relying on the cli args (e.g., `homedir \/root`) could cause problems in the future if they ever happen to change. What do you think about adding `ppid` to the ps options, and then having a regex like the following?
`r"(?P<pid>\d+)\s+1"`

It will capture the PID of every line that also has a PPID of 1, and since we specified `dirmngr` and `gpg-agent`, we shouldn't get any extra processes.

cloudinit/config/cc_apt_configure.py
```diff
@@ -244,11 +244,24 @@ def apply_apt(cfg, cloud, target):
...
+    if len(gpg_pids) <= 2:
+        if gpg_pids:
+            LOG.debug("Killing gpg-agent and dirmngr pids: %s", gpg_pids)
+        for gpg_pid in gpg_pids:
+            os.kill(int(gpg_pid))
```
TheRealFalcon, line 258: `os.kill` requires a signal.

TheRealFalcon (changes requested):

cloudinit/config/cc_apt_configure.py
```diff
@@ -244,11 +244,24 @@ def apply_apt(cfg, cloud, target):
...
+    else:
+        LOG.debug(
+            "Leaving all gpg-agent and dirmngr daemons running. Greater than 2"
```
holmanb, line 261: This is actually a failure path. If >2 are found, then `cloud-init status --wait` will last forever. Are you sure we want that? All it would take to break this logic is some other service starts a daemon prior to cloud-init. It probably fits the default case, but seems fragile.

For something more correct couldn't we do something like check for for `gpg-agent` and `dirmgr` processes inside of cloud-init cgroups and end those? That way we're not broken if some process outside of cloud-init starts an agent?

holmanb (changes requested):
See comment

cloudinit/config/cc_apt_configure.py
blackboxsw: Yes better, I'm using `ps -o ppid,pid` and comparing to PPID==1 now. thanks

cloudinit/config/cc_apt_configure.py
blackboxsw: fixed

cloudinit/config/cc_apt_configure.py
blackboxsw: @holmanb this is no longer a failure path because we landed f7a2c48f23d924bc9a12742983f866387a248e18 which now treats the cloud-config.service as complete when MainPID== 0 and SubState==running. So we aren't blocking indefinitely anymore on subprocesses

cloudinit/config/cc_apt_configure.py
```diff
@@ -244,11 +245,18 @@ def apply_apt(cfg, cloud, target):
         )
     # GH: 4344 - stop gpg-agent/dirmgr daemons spawned by gpg key imports.
     # Daemons spawned by cloud-config.service on systemd v253 report (running)
-    subp.subp(
-        ["gpgconf", "--kill", "all"],
+    gpg_process_out, _err = subp.subp(
+        ["ps", "-o", "ppid,pid", "-C", "dirmngr", "-C", "gpg-agent"],
         target=target,
         capture=True,
+        rcs=[0, 1],
```
TheRealFalcon, line 252: Why is rc 1 needed?

TheRealFalcon (approved):
LGTM! One question inline but not a request/blocker

cloudinit/config/cc_apt_configure.py
blackboxsw: @TheRealFalcon if ps doesn't find any matching processes with the `-c gpg-agent -c dirmngr` it'll return 1 sort of like grep. So, in the condition where user-data didn't specify apt/keyids, we wouldn't have these spawned processes and we want to ignore that rc 1 error code. The stdout in the case of no matching process names is
```
   PPID     PID
```
So our  re.findall will match no ppid/pid lines anyway and we NOOP in that case.

cloudinit/config/cc_apt_configure.py
holmanb: @blackboxsw Right, okay got it not a failure path. But we will still leak daemons in that case, right? If that's the case think that making this new code run conditionally in the case that `gpgconf` doesn't exist on that path would be safer?

Up to you - I'll remove my blocking review.

holmanb (approved):

cloudinit/config/cc_apt_configure.py
blackboxsw: Also, I've dropped this else now anyway since we are more explicitly matching on processed spawned by PPID=1. This prevents cloud-init from accidentally clobbering other user's gpg-agent/dirmngr.

mwhudson:
I happened to be looking at this code recently and wondered if gpgconf was really necessary for this so I'm glad I'm not the only one :wink:

I did have a different idea for how to handle this though that I wonder if you considered: could you use systemd-run to run the gpg guff in a separate cgroup or whatever and then use systemd facilities to make sure all the processes gpg spawned are dead?

Given that you've fixed this already I guess it's not worth doing it again, but I do want to spread the meme of cgroups as a way to manage groups of processes more reliably.

mwhudson:
Here's the sort of thing I mean https://github.com/mwhudson/cloud-init/commit/a960620aabdac9ed1e8766cc41f518764ce1d12d -- untested in the cloud-init code, but I tested commands like this interactively and it worked there...

blackboxsw:
@mwhudson great suggestion. Much cleaner than shelling out after the fact to ps to find and kill spawned processes. I'll take that suggestion and run with it in tip of main, it probably won't make it back into mantic until the first SRU, but it's much more contained the way you suggest it, thanks.

PR #4421, opened 2023-09-08
Upstream opensuse route code

holmanb:
```
Upstream opensuse route code into _write_network_state()

https://build.opensuse.org/package/view_file/Cloud:Tools/cloud-init/cloud-init-write-routes.patch
```

## Additional Context

@rjschwei could you please review? I think that once this lands we can drop the patch in downstream opensuse.

Based on discussion with @rjschwei at cloud-init summit on August 10th, 2023.

### DIFF
--- a/cloudinit/distros/opensuse.py
+++ b/cloudinit/distros/opensuse.py
@@ -10,6 +10,7 @@
 
 import logging
 import os
+from contextlib import suppress
 from typing import Optional
 
 from cloudinit import distros, helpers, subp, util
@@ -402,7 +403,10 @@ class Distro(distros.Distro):
                 for device_name in devices:
                     config_routes = ""
                     device_config = devices[device_name]
-                    try:
+                    # the parser expects another level of nesting
+                    # which should be there in case it's properly
+                    # formatted; if not we may get an exception on items()
+                    with suppress(AttributeError):
                         gateways = [
                             v
                             for k, v in device_config.items()
@@ -421,8 +425,3 @@ class Distro(distros.Distro):
                                 )
                             )
                             util.write_file(route_file, config_routes)
-                    except Exception:
-                        # the parser above expects another level of nesting
-                        # which should be there in case it's properly
-                        # formatted; if not we may get an exception on items()
-                        pass
--- a/cloudinit/distros/opensuse.py
+++ b/cloudinit/distros/opensuse.py
@@ -286,10 +286,14 @@ class Distro(distros.Distro):
         return self._preferred_ntp_clients
 
     def _write_network_state(
-        self, network_state, renderer: Renderer, netconfig: Optional[dict]
+        self,
+        network_state,
+        renderer: Renderer,
+        netconfig: Optional[dict] = None,
     ):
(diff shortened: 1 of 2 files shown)


rjschwei (approved):
LGTM, this was a clever way to get this in, much appreciate the effort and help.

holmanb:
Thanks for the review @rjschwei!

I just force pushed to resolve a conflict with main (due to adjacent import changes in main).

holmanb:
@rjschwei Apologies for requesting the re-review here. We shouldn't ignore all exceptions like the original patch did - do you think that [this change](https://github.com/canonical/cloud-init/pull/4421/commits/c1167306a7026d71e7a5fb5a24d2a444d219d401) will suffice?

If I understand correctly, `.items()` might be called on an object that isn't a `dict` in the case that it isn't nested as deeply. This causes the exception, but in that case we are fine to just skip ahead.

This solution limits the scope of the exceptions we'll capture to the exception we expect to occur: an `AttributeError`.

cloudinit/distros/opensuse.py
```diff
@@ -10,10 +10,14 @@
 
 import logging
 import os
+from contextlib import suppress
+from typing import Optional
 
 from cloudinit import distros, helpers, subp, util
 from cloudinit.distros import rhel_util as rhutil
 from cloudinit.distros.parsers.hostname import HostnameConf
+from cloudinit.net import ipv4_mask_to_net_prefix
```
rjschwei, line 19: There is a bug in the code w.r.t. handling ipv6. I added `net` to line 16

cloudinit/distros/opensuse.py
```diff
@@ -281,3 +285,143 @@ def preferred_ntp_clients(self):
...
+                for route in routes:
+                    dest = route.get("destination") or route.get("network")
+                    if not dest or dest in default_nets:
+                        dest = "default"
+                        if not has_default_route:
+                            has_default_route = True
+                    if dest != "default":
+                        netmask = route.get("netmask")
+                        if netmask:
+                            prefix = ipv4_mask_to_net_prefix(netmask)
```
rjschwei, line 341: ...and then changes the code here to:
```
                            if net.is_ipv4_network(netmask):
                                prefix = net.ipv4_mask_to_net_prefix(netmask)
                            if net.is_ipv6_network(netmask):
                                prefix = net.ipv6_mask_to_net_prefix(netmask)
```
This is currently being tested by the customer that found this issue. This happened because the customer config includes a `netmask` setting for an IPv6 network, which was previously not considered.

rjschwei (commented):
Comments based on recently identified bug.

cloudinit/distros/opensuse.py
holmanb: Hey @rjschwei, did the customer find any issue with this implementation? If not then I'll update this PR to reflect your suggestions.

rjschwei: @holmanb the updated patch is not yet released but has been verified to work.

PR #4424, opened 2023-09-11
Pytestify apt config test modules

blackboxsw:
## Proposed Commit Message
```
tests: convert multiple cc_apt_configure tests to pytest

Adapt test_apt_configure_sources_list_v1.py,
test_apt_configure_sources_list_v3.py and test_apt_source_v3.py
to pytest

One functional change made for test support in cc_apt_configure.
    - add_apt_sources function only prepends target test_apt_source_v3.py path to filename
      if filename does not already start with target.
No production logic invokes add_apt_sources with target= param set.

For pytest conversion, the following procedure was followed:
- Make use of fake_filesystem fixture and avoid subclassing from FilesystemMockingTestCase
- Rector common EXAMPLE_TMPL, reusing it as a base test content/layout representation in
  YAML_TEXT_CUSTOM_SL.
 - use a common setup fixture in a class for typical mocks are needed for tests
 - use mocker instead of unittest.mock where possible
 - drop private _apt_source_list function and complex ExitStack in favor of common setup fixture


Now that we are actually writing tmpl files out and reading them directly instead of mocking out
all isfile, load_file and write_file calls, there are a couple of bugs in the testing of
EXPECTED_CONTENT_MIRROR_CONTENT and EXPECTED_PRIMSEC_CONTENT suites
represented because of the nested ExitStack mocking was mocking the wrong return values.
```

## Checklist:

 - [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)

### DIFF
--- a/tests/unittests/config/test_apt_source_v3.py
+++ b/tests/unittests/config/test_apt_source_v3.py
@@ -738,7 +738,6 @@ class TestAptSourceConfig:
         # create files
         for (opre, _npre, suff) in files:
             fpath = os.path.join(apt_lists_d, opre + suff)
-            print(fpath)
             util.write_file(fpath, content=fpath)
 
         cc_apt_configure.rename_apt_lists(mirrors, target.strpath, arch)
--- a/tests/unittests/config/test_apt_source_v3.py
+++ b/tests/unittests/config/test_apt_source_v3.py
@@ -8,17 +8,14 @@ import glob
 import os
 import pathlib
 import re
-import shutil
 import socket
-import tempfile
-from unittest import TestCase, mock
+from unittest import mock
 from unittest.mock import call
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


tests/unittests/config/test_apt_source_v3.py
```diff
@@ -736,11 +738,11 @@ def test_apt_v3_list_rename_non_slash(self, m_get_dpkg_architecture):
         # create files
         for (opre, _npre, suff) in files:
             fpath = os.path.join(apt_lists_d, opre + suff)
+            print(fpath)
             util.write_file(fpath, content=fpath)
```
TheRealFalcon, line 742: Nit: I'm assuming this is for debugging?

TheRealFalcon (approved):
LGTM!

PR #4426, opened 2023-09-11
sources: do not override datasource detection if None is in list

cjp256:
Users with datasource_list = [Azure, None] started failing to boot
    properly outside of Azure with the changes to override datasource detection.

    If the fallback "None" is included in the datasource_list, do not treat
    the system as configured with a single datasource.

    If users want to force a single datasource regardless of detection,
    they can do so by removing None from the list.

### DIFF
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -352,10 +352,7 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 self,
             )
             return True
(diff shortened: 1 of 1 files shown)


holmanb (approved):
This makes sense. Checking for `[Azure, "None"]` was intended to avoid re-doing the datasource checks that `ds-identify` already does, but if this breaks a migration path that we'd like to support then I don't think this optimization makes sense.

Thanks for fielding this bug and providing a fix.

+1

PR #4429, opened 2023-09-12
Add "phsm" as contributor

phsm:
Add "phsm" as contributor

### DIFF
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -117,6 +117,7 @@ omBratteng
 onitake
 Oursin
 outscale-mdr
+phsm
 phunyguy
 qubidt
 RedKrieg


holmanb:
@phsm What plans do you have to contribute to cloud-init?

phsm:
I would like to make 2 PRs to the CloudStack datasource fixing 2 bugs that I found:
- the host's discovered FQDN  is equal to the hostname. Instead, the domain part of the FQDN should be taken from dhcp leases.
- Searching for the metadata server via dhclient leases on FreeBSD: currently the datasource doesn't inspect FreeBSD's default directory with lease files.

holmanb:
> I would like to make 2 PRs to the CloudStack datasource fixing 2 bugs that I found:
> 
>     * the host's discovered FQDN  is equal to the hostname. Instead, the domain part of the FQDN should be taken from dhcp leases.
> 
>     * Searching for the metadata server via dhclient leases on FreeBSD: currently the datasource doesn't inspect FreeBSD's default directory with lease files.

Sounds good to me. Welcome to cloud-init! I'm looking forward to seeing those PRs :D

holmanb (approved):

PR #4430, opened 2023-09-12
tools/ds-identify: match Azure datasource's ds_detect() behavior

cjp256:
The current identification enables the Azure datasource for all of Hyper-V VMs.  This datasource is strictly intendeded to work on Azure cloud as identified by DMI chassis tag and has largely been limited to this until 23.2 with a guard in the datasource.

Match ds_detect() behavior by removing hypervisor and filesystem label checks.

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1060,18 +1060,8 @@ is_azure_chassis() {
 }
 
 dscheck_Azure() {
-    # http://paste.ubuntu.com/23630873/
-    # $ grep /sr0 /run/blkid/blkid.tab
-    # <device DEVNO="0x0b00" TIME="1481737655.543841"
-    #  UUID="112D211272645f72" LABEL="rd_rdfe_stable.161212-1209"
-    #  TYPE="udf">/dev/sr0</device>
-    #
     is_azure_chassis && return $DS_FOUND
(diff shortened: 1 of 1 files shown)


anhvoms (approved):
LGTM

holmanb (approved):
Looks good, thanks @cjp256!

dermotbradley:
> The current identification enables the Azure datasource for all of Hyper-V VMs. This datasource is strictly intendeded to work on Azure cloud

@cjp256 side question: which DS would you expect people to use for non-Azure Hyper-V VMs then? NoCloud?

cjp256:
> > The current identification enables the Azure datasource for all of Hyper-V VMs. This datasource is strictly intendeded to work on Azure cloud
> 
> @cjp256 side question: which DS would you expect people to use for non-Azure Hyper-V VMs then? NoCloud?

@dermotbradley Yeah, or whatever makes sense for their use case.  The datasource is really the glue to get the configuration data for a given platform (userdata, admin ssh keys, etc.), but there is no general channel in Hyper-V for the host to communicate that sort of information to the VM, at least that I know of.

Azure supports a provisioning media attached to the VM at provisioning as well as Azure Instance Metadata Service to retrieve those bits as configured by the user when they create it.

Issue #4410, opened 2023-09-05
debian/ubuntu: user-data declaring apt keyids breaks on images without gpg


blackboxsw:
# Bug report

If Debian/Ubuntu images do not have gpg package installed, cloud-init will Traceback when optional #cloud-config user-data provides `apt: sources:` configuration.

While cloud-init's `debian/control` declares a `Recommends: gnupg` in debian/control packaging, some images will not have gnupg installed due to space constraints. cloud-init should better handle error and/or warnings when [calling  `gpg --recv-keys`](https://github.com/canonical/cloud-init/blob/main/cloudinit/gpg.py#L81) and or --delete-keys when the gpg command is absent.

Ideal best-practice is to install gpg or other tools if option user-data requires gpg interaction and the image does not contain gpg packages. This install of gpg package will slow boot, but if this is only a `Recommends:` and not a strict `Requires:` it is up to downstream image creates to determine what typical use-cases they want to support in images.

Otherwise we get Tracebacks such as:
```
2023-09-05 19:17:02,104 - subp.py[DEBUG]: Running command ['gpg', '--batch', '--yes', '--delete-keys', 'E4D304DF'] with allowed return codes [0] (shell=False, capture=True)
2023-09-05 19:17:02,105 - gpg.py[WARNING]: Failed delete key "E4D304DF": Unexpected error while running command.
Command: ['gpg', '--batch', '--yes', '--delete-keys', 'E4D304DF']
Exit code: -

```

## Environment details
- Cloud-init version: 23.3
- Operating System Distribution: Ubuntu/debian
- Cloud provider, platform or installer type: all

PR #4110, opened 2023-05-13
launching salt-minion in masterless mode

dhalturin:
## Proposed Commit Message
Launching salt-minion in masterless mode

```
If the minion is launched in masterless mode, then it should not be launched as a daemon.
```

## Additional Context
For more - https://docs.saltproject.io/en/latest/topics/tutorials/quickstart.html

## Test Steps
For testing and reproduce, you can run cloud-init with the following config:
```
#cloud-config
salt_minion:
  conf:
    file_client: local
    fileserver_backend:
     - gitfs
    gitfs_remotes:
     - https://github.com/foo/bar
```
Without a fix, salt-minion will not run correctly in this case.
This fix allows you to run salt-minion both in daemon mode and in masterless mode.

## Checklist:
 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #411, opened 2020-06-02
Ubuntu/devel

blackboxsw:
new-upstream-snapshot for release into groovy

to create it:
```
git checkout upstream/ubuntu/devel -B ubuntu/devel
new-upstream-snapshot 5f7825e2 
sed -i -e "1s/UNRELEASED/groovy/" debian/changelog
git commit -m "releasing cloud-init version 20.2-45-g5f7825e2-0ubuntu1" debian/changelog
```

to validate:
```
check diff of your branch vs mine (should only have a changelog author entry)
build-package
```

PR #4431, opened 2023-09-12
Install gnupg if gpg not found

TheRealFalcon:
Note that this still hasn't been broken in the latest mantic images, so I can't actually confirm that the fix works other than manually uninstalling packages and re-running cloud-init.

## Proposed Commit Message

```
Install gnupg if gpg not found

Ubuntu recommends gnupg in its packaging but does not require it, and
minimal images will no longer contain gnupg. Given that gpg is only
used for apt key handling, rather than having a hard requirement,
this commit installs gnupg if no gpg binary is found.

Fixes GH-4410
```

## Additional Context
https://github.com/canonical/cloud-init/issues/4410

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -8,10 +8,12 @@
 
 """Apt Configure: Configure apt for the user."""
 
+import functools
 import glob
 import os
 import pathlib
 import re
+import shutil
 import signal
 from textwrap import dedent
 
@@ -216,7 +218,7 @@ def apply_apt(cfg, cloud, target):
     LOG.debug("Apt Mirror info: %s", mirrors)
 
     if util.is_false(cfg.get("preserve_sources_list", False)):
-        add_mirror_keys(cfg, target)
+        add_mirror_keys(cfg, cloud, target)
         generate_sources_list(cfg, release, mirrors, cloud)
         rename_apt_lists(mirrors, target, arch)
 
@@ -372,7 +374,7 @@ def rename_apt_lists(new_mirrors, target, arch):
     default_mirrors = get_default_mirrors(arch)
 
     pre = subp.target_path(target, APT_LISTS)
-    for (name, omirror) in default_mirrors.items():
+    for name, omirror in default_mirrors.items():
         nmirror = new_mirrors.get(name)
         if not nmirror:
             continue
@@ -448,11 +450,11 @@ def disable_suites(disabled, src, release):
     return retsrc
 
 
-def add_mirror_keys(cfg, target):
+def add_mirror_keys(cfg, cloud, target):
     """Adds any keys included in the primary/security mirror clauses"""
     for key in ("primary", "security"):
         for mirror in cfg.get(key, []):
-            add_apt_key(mirror, target, file_name=key)
+            add_apt_key(mirror, cloud, target, file_name=key)
 
 
 def generate_sources_list(cfg, release, mirrors, cloud):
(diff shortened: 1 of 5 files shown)


cloudinit/config/cc_apt_configure.py
```diff
@@ -499,12 +501,19 @@ def add_apt_key_raw(key, file_name, hardened=False, target=None):
         raise
 
 
-def add_apt_key(ent, target=None, hardened=False, file_name=None):
+@functools.lru_cache(maxsize=1)
+def _ensure_gpg(cloud):
+    if not shutil.which("gpg"):
```
igalic, line 506: what's the difference between this, and subp.which?

igalic (commented):
❓

cloudinit/config/cc_apt_configure.py
TheRealFalcon: `shutil` is the standard library version. They both do basically the same thing. `subp` version has some slightly different logic that isn't relevant in most cases. I think we can probably move away from the `subp` version at some point.

igalic: I thought the main reason were using our own wrapper is ease of testing

TheRealFalcon: How do see that easing testing?

blackboxsw (approved):
This PR is good as an individual commit, but I think we need a second PR to follow this that also handles software-properties-common which is also a `Recommends:` and also will not be present in ubuntu-minimal images.

Without software-properties-common, we can't run add-apt-repository. So we'll ultimately get tracebacks on missing that dependency as well (which will break ubuntu-minimal image creation scripts).

`FileNotFoundError: [Errno 2] No such file or directory: b'add-apt-repository'`

I could see us adapting the _ensure_gpg to _ensure_dependencies which could:
 -  if not shutil.which("gpg") -> install gnupg
 - if not shutil.which("add-apt-repository"): install software-properties-common
 
It'd be best to do a `single distro.install_packages(missing_deps) to avoid time cost of calling that twice.

We should consider extending our existing integration tests as you said to remove gpg on those images to validate succes in this PR or a follow on for software-properties-common (which represents that add-apt-repository traceback above)

```diff
diff --git a/tests/integration_tests/modules/test_apt.py b/tests/integration_tests/modules/test_apt.py
index 6cf1dc1fb..f921f4e23 100644
--- a/tests/integration_tests/modules/test_apt.py
+++ b/tests/integration_tests/modules/test_apt.py
@@ -11,6 +11,8 @@ from tests.integration_tests.releases import CURRENT_RELEASE, IS_UBUNTU
 
 USER_DATA = """\
 #cloud-config
+bootcmd:
+    - apt-get remove gpg --yes
 apt:
   conf: |
     APT {
```

```
./tools/run-container --package ubuntu/lunar
 CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_23.3-23-g*bddeb_all.deb CLOUD_INIT_KEEP_INSTANCE=1 CLOUD_INIT_OS_IMAGE=lunar tox -e integration-tests ./tests/integration_tests/modules/test_apt.py::TestApt::test_key
```

holmanb (commented):
```
    if not shutil.which("gpg") -> install gnupg
    if not shutil.which("add-apt-repository"): install software-properties-common
```

Confirmed in Debian that these package names and code should also work.

blackboxsw:
Also updated suggestion commit message replaceing GH-411 -> GH-4110

PR #4432, opened 2023-09-13
azure: add an option to allow enabling/disabling secondary ip configuration

KsenijaS:
Currently cloud-init attempts to configure all IP addresses for all interfaces whose configurations are available in IMDS.
To allow users more control over VM network configuration, add an option for cloud-init to enable/disable configuration of secondary IP addresses.
This PR also introduces the datasource option `apply_network_config_for_secondary_ips` for DataSourceAzure, which leverages the new network config above. Default is set to True to maintain compatibility.


cloudinit/sources/DataSourceAzure.py
```diff
@@ -286,6 +286,7 @@ def get_resource_disk_on_freebsd(port_id) -> Optional[str]:
     "data_dir": AGENT_SEED_DIR,
     "disk_aliases": {"ephemeral0": RESOURCE_DISK_PATH},
     "apply_network_config": True,  # Use IMDS published network configuration
+    "apply_ip_network_config": True,  # Set secondary ip addresses
```
cjp256, line 289: `apply_network_config_for_secondary_ips` would be clearer

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1856,6 +1858,7 @@ def load_azure_ds_dir(source_dir):
 @azure_ds_telemetry_reporter
 def generate_network_config_from_instance_network_metadata(
     network_metadata: dict,
+    apply_ip_network_config: bool,
```
cjp256, line 1861: preferably 
```suggestion
    *, apply_ip_network_config: bool,
```
and pass apply_ip_network_config by keyword arg so it's clear what the parameter is

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1894,17 +1897,18 @@ def generate_network_config_from_instance_network_metadata(
...
-                netPrefix = intf[addr_type]["subnet"][0].get(
-                    "prefix", default_prefix
-                )
-                privateIp = addr["privateIpAddress"]
-                if not dev_config.get("addresses"):
-                    dev_config["addresses"] = []
-                dev_config["addresses"].append(
-                    "{ip}/{prefix}".format(ip=privateIp, prefix=netPrefix)
-                )
+            if apply_ip_network_config:
```
cjp256, line 1900: can avoid adding an extra indent by something like:
```
if not apply secondary IPs:
  continue
```

cjp256 (changes requested):

TheRealFalcon:
@KsenijaS and @cjp256 , what's the reasoning for this change? If secondary IP information is coming from the IMDS, doesn't that mean that the NICs have been specifically configured to be that way through the Azure console or CLI? Why should cloud-init ignore that?

cloudinit/sources/DataSourceAzure.py
KsenijaS: Done!

cloudinit/sources/DataSourceAzure.py
KsenijaS: Done!

cloudinit/sources/DataSourceAzure.py
KsenijaS: Done!

KsenijaS:
> @KsenijaS and @cjp256 , what's the reasoning for this change? If secondary IP information is coming from the IMDS, doesn't that mean that the NICs have been specifically configured to be that way through the Azure console or CLI? Why should cloud-init ignore that?

By not assigning the secondary IPs to interfaces, these IPs can otherwise be assigned to other virtual interfaces (ex. containers).

anhvoms:
> @KsenijaS and @cjp256 , what's the reasoning for this change? If secondary IP information is coming from the IMDS, doesn't that mean that the NICs have been specifically configured to be that way through the Azure console or CLI? Why should cloud-init ignore that?

@TheRealFalcon  Elaborating on Ksenija's reply above. There are scenarios (AKS, for example) where the secondary IPs will be configured by another component in the VM (in AKS case, the secondary IPs will be assigned to the container's interfaces)). In those scenarios, cloud-init should not attempt to assign secondary IPs.

TheRealFalcon (approved):
LGTM!

@cjp256 , since you originally reviewed, are you good here too?

doc/rtd/reference/datasources/azure.rst
```diff
@@ -41,6 +41,10 @@ The settings that may be configured are:
   Boolean set to True to use network configuration described by Azure's IMDS
   endpoint instead of fallback network config of DHCP on eth0. Default is
   True.
+* :command:`apply_network_config_for_secondary_ips`
+
+  Boolean set to True to configure primary and secondary IP addresses for each
```
cjp256, line 46:
```suggestion
  Configure secondary IP addresses for each NIC.
  Default is True.
```

cjp256 (changes requested):
LGTM, just one minor doc nit! Thanks!!

doc/rtd/reference/datasources/azure.rst
cjp256: Is there a standard template for documenting these options?

TheRealFalcon: Unfortunately no. We generally try to keep things consistent, but there's no template or specific rules for it. I'll see if that's something we can add.

KsenijaS: done!

PR #4434, opened 2023-09-14
sources/Azure: fix for conflicting reports to platform

cjp256:
When reporting failure the _negotiated flag is not set to True. This may result in reporting success when we do the final check for _negotiated to report ready at the end of the process.

Azure ignores the additional report so there weren't unexpected "successful" deployments but the report is unnecessary and conflicting with the prior log.

- Set or reset _negotiated flag in various _report_ready*() methods.

- Add a test case to verify this scenario.


holmanb (approved):
Looks good, thanks!

PR #4437, opened 2023-09-15
Deb822 default sources list d

blackboxsw:
## Proposed Commit Message

```
    apt_configure: add deb822 support for default sources file

    When features.APT_DEB822_SOURCE_LIST_FILE is set, cloud-init
    will generate an /etc/apt/sources.list.d/<distro>.source file
    in Deb822 format instead of /etc/apt/sources.list.
    
    Introduce a package Recommends dependency on python3-apt.
    If apt_pkg module is available use it to read APT config settings for
    Dir::Etc, Dir::Etc::sourceparts and Dir::Etc::sourcelist.
    
    Add deb822 default templates per distro:
    - templates/sources.list.debian.deb822.tmpl
    - templates/sources.list.ubuntu.deb822.tmpl
    
    When features.APT_DEB822_SOURCE_LIST_FILE is set, cloud-init will
    - read the distro-specific deb822 formatted template file
    - When apt_pkg module is present read the appropriate APT config
      "Dir::Etc::sourceparts" ortherwise use DEFAULT_APT_CFG
    - render /etc/apt/sources.list.d/<distro>.source file from the tmpl
    
    Add function is_deb822_sources_format to check for expected deb822
    keys in rendered APT rendered content before writing deb822 source
    format files. If APT source content is not deb822 format, fallback to
    writing /etc/apt/sources.list.

    The feature flag and format check retains original behavior on stable
    releases where custom cloud-config user-data could contain non-deb822
    APT source content for the keys:
      custom_apt_sources and apt_sources
```

## Additional Context

We would like to target this change for mantic.
When this PR lands, we will also add a `Recommends: python3-apt` to debian/control files.

## Test Steps

```
CLOUD_INIT_KEEP_INSTANCE=1 CLOUD_INIT_OS_IMAGE=mantic CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_23.3-*~bddeb_all.deb tox -e integration-tests tests/integration_tests/modules/test_apt.py
```
## Checklist:


tests/unittests/config/test_apt_configure_sources_list_v1.py
```diff
@@ -14,15 +14,6 @@
 EXAMPLE_TMPL = """\
 ## template:jinja
 ## Note, this file is written by cloud-init on first boot of an instance
-## modifications made here will not survive a re-bundle.
```
blackboxsw, line 17: Just reducing vertical space dedicated to unnecessary extra boilerplate test strings

templates/sources.list.ubuntu.deb822.tmpl
```diff
@@ -0,0 +1,60 @@
...
+##                 may not be under a free licence. Please satisfy yourself as
+##                 to your rights to use the software.
+##                 Also, please note that software in multiverse WILL NOT
+##                 receive any review or updates from the Ubuntu security team.
+##
+## See the sources.list(5) manual page for further settings.
+# Types: deb deb-src
+Types: deb
+URIs: {{mirror}}
+Suites: {{codename}} {{codename}}-updates {{codename}}-backports
```
blackboxsw, line 50: Please pay extra attention to the Suites and Types here and compare to the active default types/suites we 
 have in the non-deb822 format templates in `templates/sources.list.(ubuntu|debian).tmpl` to double check we didn't unintentionally activate or disable some type or suite.

tests/unittests/config/test_apt_source_v3.py
```diff
@@ -1540,3 +1541,57 @@ def test_disable_deb822_suites_disables_proper_suites(
...
+Dir "/";
+Dir::Etc "etc/myapt";
+Dir::Etc::sourcelist "sources.my.list";
+Dir::Etc::sourceparts "sources.my.list.d";
+Dir::Etc::main "apt.conf";
+"""
+
+
+class TestGetAptCfg:
+    @skipIfAptPkg()
```
blackboxsw, line 1557: Because we aren't defining apt_pkg in our requirements.txt or test-requirement.txt file, these tests will always run by default in our development environment. During pkg build and integration tests, python3-apt will almost always be installed due to other build dependencies or base ubuntu images, so these tests will be skipped as they only rely on calling apt-config dump if apt_pkg is not present. So, we get coverage with and without this dependency in various parts of our testing.

blackboxsw:
CC: @enr0n  just FYI as I know you recently added changes to ubuntu-release-ugprader to support upgrade paths and migration to deb822 format. You may have some ideas on shared design we should look at long term within cloud-init to better support U-R-U with shared libraries/functions etc. And please confirm that cloud-init's writing of `/etc/apt/sources.list.d/ubuntu.sources` doesn't collide with any upgrade migration logic in place.  Given that cloud-init could run in systems without python3-apt installed, any shared libs/functions we rely would need to rely on working in the potential absence of aptsources/apt_pkg python modules.

TheRealFalcon (changes requested):
First pass review. Overall looks really good. I left inline comments. I still need to grok the disable code a little deeper and look at tests.

enr0n:
> CC: @enr0n just FYI as I know you recently added changes to ubuntu-release-ugprader to support upgrade paths and migration to deb822 format. You may have some ideas on shared design we should look at long term within cloud-init to better support U-R-U with shared libraries/functions etc. And please confirm that cloud-init's writing of `/etc/apt/sources.list.d/ubuntu.sources` doesn't collide with any upgrade migration logic in place. Given that cloud-init could run in systems without python3-apt installed, any shared libs/functions we rely would need to rely on working in the potential absence of aptsources/apt_pkg python modules.

I have not had time to read the code yet, but FYSA last week we decided we are not going to have deb822 by default for 23.10, so u-r-u will not do the migration on upgrades for this release. I was on travel last week so I haven't updated u-r-u or sent an update on ubuntu-devel yet, but I will today. All of the deb822 logic will still be needed of course, but we will not enable it by default for 23.10.

cloudinit/config/cc_apt_configure.py
```diff
@@ -455,34 +540,127 @@ def add_mirror_keys(cfg, target):
...
+    Only validates that minimal required keys are present in the file, which
+    indicates we are likely deb822 format.
+
+    Doesn't handle if multiple sections all contain deb822 keys.
+
+    Return True if content looks like it is deb822 formatted APT source.
+    """
+    # TODO(At jammy EOL: use aptsources.sourceslist.Deb822SourceEntry.invalid)
+    try:
+        apt_src = safeyaml.load(apt_src_content)
```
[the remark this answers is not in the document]
blackboxsw: Note that specific to apt sources in deb822 are a lot more strict than the general spec, any file that doesn't adhere to either commented line, a single line or whitespace between deb822 source entries and each entry minimally containing Ty=pes, Suites, Components, URIs will generate an error like the following it it is missing any known key
```
E: Malformed entry 2 in sources file /etc/apt/sources.list.d/ubuntu.sources (Suite)
```

Yes supplemental keys can be provided which aren't YAML, so, I agree we should not leverage such a short-cut here.  I've refactored the function to return True if the known deb822 source keys are present, and no known invalid `^deb  or ^- ^deb-src lines exist.

cloudinit/config/cc_apt_configure.py
```diff
@@ -49,6 +51,10 @@
         cloud-init's distro support for instructions on using
         these config options.
 
+        In cloud-init 23.4 or later, cloud-init will generate default
```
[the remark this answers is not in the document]
blackboxsw: dropped the versioning docs. It is confusing to represent something we know we will disable on stable releases to avoid change in behavior.

tests/unittests/config/test_apt_configure_sources_list_v1.py
```diff
@@ -120,12 +148,12 @@ def apt_source_list(self, distro, mirror, tmpdir, mirrorcheck=None):
     def test_apt_v1_source_list_by_distro(self, distro, mirror, tmpdir):
         """Test rendering of a source.list from template for each distro"""
         mycloud = get_cloud(distro)
-        tmpl_file = f"/etc/cloud/templates/sources.list.{distro}.tmpl"
-        util.write_file(tmpl_file, EXAMPLE_TMPL)
+        tmpl_file = f"/etc/cloud/templates/sources.list.{distro}.deb822.tmpl"
```
TheRealFalcon, line 151: The old test is still a valid test and valid behavior, correct? Should we have a version of the test that uses the feature flag and one that doesn't, rather than moving to a version that only does?

The same concern applies to all tests modified/removed in this file and the v3 version.

cloudinit/config/cc_apt_configure.py
```diff
@@ -461,34 +549,135 @@ def add_mirror_keys(cfg, cloud, target):
...
+    """
+    # TODO(At jammy EOL: use aptsources.sourceslist.Deb822SourceEntry.invalid)
+    if re.findall(r"^(deb |deb-src )", apt_src_content, re.M):
+        return False
+    if re.findall(
+        r"^(Types: |Suites: |Components: |URIs: )", apt_src_content, re.M
+    ):
+        return True
+    # Did not match any required deb822 format keys
+    return False
```
TheRealFalcon, line 570: If we actually get here we'll have a sources file that we don't actually know what to do with. Can we log a warning?

TheRealFalcon (commented):
I left some minor inline comments.

templates/sources.list.ubuntu.deb822.tmpl
```diff
@@ -0,0 +1,60 @@
+## template:jinja
```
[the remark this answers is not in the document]
blackboxsw: I don't think it's worth defining a separate fully commented entry because multiple `Types:` can be declared on one line in a given entry.

That said, in ubuntu.deb822.tmpl, I had the commented `# Types: deb deb-src` above our active `Types: deb` in the top entry, but is was missing from the security source entry.

```diff
diff --git a/templates/sources.list.ubuntu.deb822.tmpl b/templates/sources.list.ubuntu.deb822.tmpl
index 7b65d53c2..d75dc088f 100644
--- a/templates/sources.list.ubuntu.deb822.tmpl
+++ b/templates/sources.list.ubuntu.deb822.tmpl
@@ -53,6 +53,7 @@ Signed-By: /usr/share/keyrings/ubuntu-archive-keyring.gpg
 
 ## Ubuntu security updates. Aside from URIs and Suites,
 ## this should mirror your choices in the previous section.
+# Types: deb deb-src
 Types: deb
 URIs: {{security}}
 Suites: {{codename}}-security
```

In sources.list.debian.deb822.tmpl, I left both `Types: deb deb-src` active because that's what debian has done in packaging in the original sources.list template. If we think that's worth changing, we can make debian align w/ ubuntu and avoid updating deb-src repos.

tests/unittests/config/test_apt_configure_sources_list_v1.py
blackboxsw: The deb822 feature is being tested in other unittests and that code path is the same regardless of v1 vs v3 vs deb822 or not. As long as we have a unit test that asserts deb822 vs non-deb822 feature finds the correct deb822 template file we know this will work for all unitttests. Addiing a matrix to each unittest to test v3 vs v1 finding the deb822 template won't produce a signal beyond "deb822 feature flag can't find the right template file".  But, I did add `deb822_feature_aware` unittest names for those tests that are specifically validating deb822.tmpl functionality.

cloudinit/config/cc_apt_configure.py
blackboxsw: Added

TheRealFalcon (approved):
LGTM! I'll leave it to you to merge.

cloudinit/config/cc_apt_configure.py
```diff
@@ -555,20 +555,50 @@ def is_deb822_format(apt_src_content: str) -> bool:
...
 
+APT_CFG_RE = (
+    r"(Dir::Etc|Dir::Etc::sourceparts|Dir::Etc::sourcelist) \"([^\"]+)"
+)
+
 
 def get_apt_cfg() -> Dict[str, str]:
-    """Return a dict of applicable apt configuration or defaults."""
-    return {
-        "sourcelist": apt_pkg.config.find_file(
```
holmanb: The import for apt_pkg still exists at 39c5c06c9e92, which conflicts with the commit message. Please fix this if you want this to be a rebase commit.

Also it feels a bit odd to introduce code (the import for example) in one commit in a series, only to relocate / refactor the same code in a followup commit in the serious. Not a huge issue in this case, since in this case (as you mentioned) it makes the review easier due to the size of this PR, but I think that's something we want to avoid when we can for rebase commits.

cloudinit/config/cc_apt_configure.py
```diff
@@ -575,8 +575,9 @@ def is_deb822_sources_format(apt_src_content: str) -> bool:
 
 
 DEFAULT_APT_CFG = {
-    "sourcelist": "/etc/apt/sources.list",
-    "sourceparts": "/etc/apt/sources.list.d/",
+    "Dir::Etc": "etc/apt",
```
holmanb, line 578: Refactoring code that hasn't landed yet adds churn / noise in the commit log. To avoid this, please squash the changes in this commit into the commit that introduced these lines of code prior to merging.

cloudinit/config/cc_apt_configure.py
```diff
@@ -562,16 +560,18 @@ def is_deb822_format(apt_src_content: str) -> bool:
     Return True if content looks like it is deb822 formatted APT source.
     """
     # TODO(At jammy EOL: use aptsources.sourceslist.Deb822SourceEntry.invalid)
-    try:
-        apt_src = safeyaml.load(apt_src_content)
```
holmanb: This is more refactoring within the series that we don't need in the commit history.

holmanb (changes requested):
@blackboxsw Nice work on this PR. I left a few comments inline, mostly related to how the commits are.

cloudinit/config/cc_apt_configure.py
blackboxsw: +1 on these points here @holman. I had actually squash merged this PR during merge instead of a rebase merge because of the way I had dealt  with review comments and refactoring code after the fact. When I handle some of the review comments, I didn't squash the content into the correct commit in some cases and rather that presenting 10 separate commits for this effort, I thought it was going to be more appropriate to squash merge and represent the set of changes in the single commit message.

cloudinit/config/cc_apt_configure.py
blackboxsw: yes agreed on this point, which had prompted my overall squash merge for this PR given some of the overlap and churn I had introduced when handling review comments and iterations. I left these separate commits in a less than desirable state after handling initial reviews which made separate commits for this effort a bit more work than I wanted to spend on this particular PR.

cloudinit/config/cc_apt_configure.py
blackboxsw: Agreed. If I had stuck with the desire to rebase merge the PR this commit set would have needed tidying up.

PR #4438, opened 2023-09-15
fix: correct vendor data usage context

suryamurugan:
## Proposed Commit Message

```

```

## Checklist:


TheRealFalcon:
@suryamurugan , thanks for contributing to cloud-init, but the change here isn't correct. Because users can ultimately disable vendor data, it shouldn't be used for things that are **required** to bring an instance online. That would mean that users can no longer launch an instance when disabling vendor data. If you'd like to change the wording to reflect that, then we'd still welcome your change.

PR #4441, opened 2023-09-15
apt: install software-properties-common when absent but needed

blackboxsw:
## Proposed Commit Message
```
apt: install software-properties-common when absent but needed

When optional user-data contains APT sources configuration that warrants source repo setup, cloud-init calls add-apt-repository which is packaged in software-properties common.

The software-properties-common package is defined as a Recommends: in debian/control, and some minimal images do not install recommended packages.

When minimal images do not have software-properties-common installed, cloud-init will install this package on first boot only when required by optional apt sources user-data.

The gnupg package is another optional/recommended package that willl be installed is specific apt user-data config requires gpg interaction.

Refactor _ensure_gpg to _ensure_dependencies to only attempt to install missing dependencies once in mininmal image cases where user-data requires both gpg and add-apt-repository commands and the image includes neither gnupg nor software-properties-common.
```

## Additional Context
Ubuntu mantic minimal daily builds do not install Recommends: packages. We need to assert that images without these packages (gnupg, software-properties-common) can 

## Checklist:

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -8,7 +8,6 @@
 
 """Apt Configure: Configure apt for the user."""
 
-import functools
 import glob
 import os
 import pathlib
@@ -36,6 +35,11 @@ CLOUD_INIT_GPG_DIR = "/etc/apt/cloud-init.gpg.d/"
 
 frequency = PER_INSTANCE
 distros = ["ubuntu", "debian"]
+
+PACKAGE_DEPENDENCY_BY_COMMAND = {
+    "add-apt-repository": "software-properties-common",
+    "gpg": "gnupg",
+}
 
 meta: MetaSchema = {
     "id": "cc_apt_configure",
@@ -501,10 +505,19 @@ def add_apt_key_raw(key, file_name, hardened=False, target=None):
         raise
 
 
-@functools.lru_cache(maxsize=1)
-def _ensure_gpg(cloud):
-    if not shutil.which("gpg"):
-        cloud.distro.install_packages(["gnupg"])
+def _ensure_dependencies(apt_sources_dict, aa_repo_match, cloud):
+    """Install missing package dependencies based on apt_sources config."""
+    missing_packages = []
+    required_cmds = set()
+    for ent in apt_sources_dict.values():
+        if set(["key", "keyid"]).intersection(ent):
+            required_cmds.update(["gpg"])
+        if aa_repo_match(ent.get("source", "")):
+            required_cmds.update(["add-apt-repository"])
+    for command in required_cmds:
+        if not shutil.which(command):
+            missing_packages.append(PACKAGE_DEPENDENCY_BY_COMMAND[command])
+    cloud.distro.install_packages(missing_packages)
 
 
 def add_apt_key(ent, cloud, target=None, hardened=False, file_name=None):
@@ -513,7 +526,6 @@ def add_apt_key(ent, cloud, target=None, hardened=False, file_name=None):
     Supports raw keys or keyid's
     The latter will as a first step fetched to get the raw key
     """
-    _ensure_gpg(cloud)
     if "keyid" in ent and "key" not in ent:
         keyserver = DEFAULT_KEYSERVER
         if "keyserver" in ent:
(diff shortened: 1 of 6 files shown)


cloudinit/config/cc_apt_configure.py
```diff
@@ -501,10 +505,19 @@ def add_apt_key_raw(key, file_name, hardened=False, target=None):
...
-def _ensure_gpg(cloud):
-    if not shutil.which("gpg"):
-        cloud.distro.install_packages(["gnupg"])
+def _ensure_dependencies(apt_sources_dict, aa_repo_match, cloud):
+    """Install missing package dependencies based on apt_sources config."""
+    missing_packages = []
+    required_cmds = set()
+    for ent in apt_sources_dict.values():
+        if set(["key", "keyid"]).intersection(ent):
+            required_cmds.update(["gpg"])
```
holmanb, line 514: I think `add()` makes more sense than `update()` here and below, no need to create lists and then union them with sets.

example:
```
>>> required_cmds = set()
>>> required_cmds.add("gpg")
>>> required_cmds.add("add-apt-repository")
>>> required_cmds
{'add-apt-repository', 'gpg'}
```

cloudinit/config/cc_apt_configure.py
```diff
@@ -581,7 +594,7 @@ def add_apt_sources(
             key_file = add_apt_key(ent, cloud, target, hardened=True)
             template_params["KEY_FILE"] = key_file
         else:
-            key_file = add_apt_key(ent, cloud, target)
+            add_apt_key(ent, cloud, target)
```
holmanb, line 597: +1

holmanb (approved):
See nits inline regarding set usage. Otherwise this looks good to me.

cloudinit/config/cc_apt_configure.py
```diff
@@ -513,7 +526,6 @@ def add_apt_key(ent, cloud, target=None, hardened=False, file_name=None):
     Supports raw keys or keyid's
     The latter will as a first step fetched to get the raw key
     """
-    _ensure_gpg(cloud)
```
TheRealFalcon, line 516: This was needed here because `add_mirror_keys` also calls `add_apt_key`. I think we can have it both places if we bring back the LRU cache

cloudinit/config/cc_apt_configure.py
```diff
@@ -501,10 +505,19 @@ def add_apt_key_raw(key, file_name, hardened=False, target=None):
...
-@functools.lru_cache(maxsize=1)
-def _ensure_gpg(cloud):
-    if not shutil.which("gpg"):
-        cloud.distro.install_packages(["gnupg"])
+def _ensure_dependencies(apt_sources_dict, aa_repo_match, cloud):
+    """Install missing package dependencies based on apt_sources config."""
+    missing_packages = []
+    required_cmds = set()
+    for ent in apt_sources_dict.values():
+        if set(["key", "keyid"]).intersection(ent):
```
TheRealFalcon, line 513: Nit:
```suggestion
        if {"key", "keyid"}.intersection(ent):
```
Also, TIL you can intersect a set with a dict.

blackboxsw: less characters for the win, and less conversion of types.

cloudinit/config/cc_apt_configure.py
blackboxsw: @TheRealFalcon shoot I missed the add_mirror_keys call as being outside the scope of `add_apt_sources` the problem I'm trying to solve in this PR is that I don't want us to call _ensure_dependencies twice because distro.install_packages will call the expensive `apt update` prior to apt install gpg or software-properties-common. So I pulled that logic up out of `add_apt_key` and into `add_apt_sources`  because I want to do this just once given a condition check on the user-data provided. It looks like I might want to pull the callsite (and user-data inspection of cfg up into `apply_apt` before `add_mirror_keys`. Then, If both gpg requirements and sw-properties-common are going to be required by the provided cloud-config, let's install them in one call to `distro.install_packages`.

TheRealFalcon: Not sure this matters to the implementation anymore, but `apt update` is protected by a runner containing a semaphore, which (IIRC) means it will only run once per boot.

TheRealFalcon (changes requested):
One more. See https://github.com/canonical/cloud-init/pull/4441/files#r1333307693

TheRealFalcon (approved):
LGTM!

PR #4443, opened 2023-09-18
feat(apport): collect ubuntu-pro logs if keys present in cloud-config (SC-1439)

aciba90:
## Proposed Commit Message

```
feat(apport): collect ubuntu-pro logs if keys present in cloud-config
```

## Additional Context

https://warthogs.atlassian.net/browse/SC-1439

## Checklist:

### DIFF
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -7,7 +7,6 @@
 import json
 import os
 
-from cloudinit import stages
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.cmd.devel.logs import (
     INSTALLER_APPORT_FILES,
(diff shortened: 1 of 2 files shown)


cloudinit/apport.py
```diff
@@ -155,6 +156,25 @@ def attach_installer_files(report, ui=None):
...
+    init = stages.Init()
+    has_up_key = init.cfg and (
+        "ubuntu_advantage" in init.cfg.keys()
+        or "ubuntu-advantage" in init.cfg.keys()
+    )
+    if not has_up_key:
+        return
+
+    realpath = os.path.realpath("/var/log/ubuntu-advantage.log")
+    attach_file_if_exists(report, realpath)
```
blackboxsw, line 170: Any reason why we wouldn't just want to include this log if it exists?
If we only want to conditionally include the ubuntu-pro tag, we could check
if '.var.log.ubuntu-avantage.log' in report.keys():
    report.setdefault("Tags", "")
     if report["Tags"]:
         report["Tags"] += " "
     report["Tags"] += "ubuntu-pro"



Alternatively, if we want to check for ubuntu-advantage or ubuntu_advantage keys in user-data and only include ubuntu-advantage.logs in that case... we probably should read `/run/cloud-init/combined-cloud-config.json`  instead of stages.Init() which has to re-read and merge all system config, user-data and vendor-data.

aciba90: Thanks, @blackboxsw. The presence of the log file is a good enough indicator of 'pro has run'. I have simplified the code.

blackboxsw (approved):
Validated behavior and additional tagging performed on systems with /var/log/ubuntu-advantage.log present.

Issue #4436, opened 2023-09-14
 Datasource vmware vmware-rpctool Permission denied


playerla:
Hi,

Following a security configuration https://github.com/vmware/open-vm-tools/issues/690 (metadata, vendordata or userdata are all readable by any local user with the defautl configuration...) 
I learn that :
> rpctool does not support secure RPCs, and we'd like to deprecate it.
> 'vmtoolsd --cmd' does support secure RPCs, and is the preferred way to execute RPCs.

I'm know stuck with a datasource vmware vmware-rpctool _Permission denied_ :
https://github.com/canonical/cloud-init/blob/97e09dfea2d74ef583bd5650b5c6e40aa19fbb24/cloudinit/sources/DataSourceVMware.py#L92

In what extend rpctool could be replaced with vmtoolsd  ? (optional switch, impact or even deprecation?)

TheRealFalcon:
@playerla , do you have documentation for how `vmtoolsd` is supposed to be used? If it is meant to be a replacement for rpctools, then it makes sense to update in the VMWare datasource.

cc: @sshedi and @akutz

akutz:
This was raised internally as well. The tl;dr:

### Context

* The tool `vmware-rpctool` is failing due to a secure RPC feature implemented in ESXi that currently is only supported by `vmtoolsd`.
* Why did we switch from `vmtoolsd` to `vmware-rpctool` in the first place? For that please see https://github.com/vmware-archive/cloud-init-vmware-guestinfo/pull/23
  * `vmtoolsd` was subject to a performance issue that was patched on the ESXi side an in open-vm-tools
  * It is not possible for the guest to know whether the hypervisor is running with that patch

### Short-term solution

* @ravindravmw recommended that if `vmware-rpctool` fail, we fall back to `vmtoolsd`
* My concern was that returning no data is a perfectly valid scenario, so failing has to be based on a specific condition, not just a non-zero exit code
* We decided to fall back to `vmtoolsd` if `vmware-rpctool` returns a non-zero exit code and writes `Permission denied` to `stderr`. 

### Long-term plan

* In the long-run, we want to move back to `vmtoolsd` anyway as it supports other features new in vSphere 8.0, such as datasets, a more robust data transport model than guestinfo. 
* However, again, Cloud-Init and the guest won't know what the host supports, so we need to develop a migration/fall-back strategy for `vmware-rpctool` --> `vmtoolsd` that:
  * Does not cause performance issues on older hosts
[...]

PR #4444, opened 2023-09-18
vmware: Fall back to vmtoolsd if vmware-rpctool errs

akutz:
Addresses #4436

## Proposed Commit Message

```
vmware: Fallback to vmtoolsd if rpctool errs

This patch udpates the ds-identify script and the
VMware datasource to fall back to using the
vmtoolsd program if vmware-rpctool errors.

Fixes GH-4436
```

## Additional Context

I validated the output and exit code for both commands:

```shell
$ vmtoolsd --cmd "info-get guestinfo.userdata.encoding"; echo $?
gzip+base64
0
```

```shell
$ vmware-rpctool "info-get guestinfo.userdata.encoding"; echo $?
gzip+base64
0
```

```shell
$ vmtoolsd --cmd "info-get guestinfo.userdata.enc"; echo $?
No value found
1
```

```shell
$ vmware-rpctool "info-get guestinfo.userdata.enc"; echo $?
No value found
1
```

## Test Steps

* Locally I verified the OVF datasource with:

    ```shell
    make clean_pyc && PYTHONPATH="$(pwd)" python3 -m pytest -v --log-level=DEBUG tests/unittests/sources/test_ovf.py 
    ```

* Locally I verified the VMware datasource with:

    ```shell
    make clean_pyc && PYTHONPATH="$(pwd)" python3 -m pytest -v --log-level=DEBUG tests/unittests/sources/test_vmware.py 
    ```

* Locally I verified `ds-identify` with:

    ```shell
    make clean_pyc && PYTHONPATH="$(pwd)" python3 -m pytest -v --log-level=DEBUG tests/unittests/test_ds_identify.py
    ```

* I also copied the modified `ds-identify`  and datasources to a VM and ran the following:

    ```shell
      sudo rm -f cloud-init*.log; sudo cloud-init clean; sudo cloud-init init
    ```

    Everything behaved as expected.

* I also replaced `/usr/bin/vmware-rpctool` with:

    ```shell
    #!/bin/sh

    exit 1
    ```

    And ran:

    ```shell
    sudo rm -f cloud-init*.log; sudo cloud-init clean; sudo cloud-init init
    ```

    And `/var/log/cloud-init.log` showed the expected behavior occurred:

    ```log
    2023-09-18 18:29:04,244 - util.py[WARNING]: Failed to get guestinfo value for key metadata: Unexpected error while running command.
    Command: ['/usr/bin/vmware-rpctool', 'info-get guestinfo.metadata']
    Exit code: 1
    Reason: -
    Stdout: 
    Stderr: 
    2023-09-18 18:29:04,244 - util.py[DEBUG]: Failed to get guestinfo value for key metadata: Unexpected error while running command.
    Command: ['/usr/bin/vmware-rpctool', 'info-get guestinfo.metadata']
    Exit code: 1
    Reason: -
    Stdout: 
    Stderr: 
      File "/usr/lib/python3.11/site-packages/cloudinit/sources/DataSourceVMware.py", line 551, in guestinfo_get_value
        (stdout, stderr) = subp(args)
                          ^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/cloudinit/subp.py", line 335, in subp
        raise ProcessExecutionError(
    Command: ['/usr/bin/vmware-rpctool', 'info-get guestinfo.metadata']
    Exit code: 1
    Reason: -
    Stdout: 
    Stderr: 
    2023-09-18 18:29:04,246 - DataSourceVMware.py[DEBUG]: Getting guestinfo value for key metadata with /usr/bin/vmtoolsd
    2023-09-18 18:29:04,246 - subp.py[DEBUG]: Running command ['/usr/bin/vmtoolsd', '--cmd', 'info-get guestinfo.metadata'] with allowed return codes [0] (shell=False, capture=True)
    2023-09-18 18:29:04,295 - DataSourceVMware.py[DEBUG]: Getting guestinfo value for key metadata.encoding with /usr/bin/vmtoolsd
    2023-09-18 18:29:04,295 - subp.py[DEBUG]: Running command ['/usr/bin/vmtoolsd', '--cmd', 'info-get guestinfo.metadata.encoding'] with allowed return codes [0] (shell=False, capture=True)
    2023-09-18 18:29:04,329 - DataSourceVMware.py[DEBUG]: Getting encoded data for key=guestinfo.metadata, enc=gzip+base64
    2023-09-18 18:29:04,330 - DataSourceVMware.py[DEBUG]: Decoding gzip+base64 format guestinfo.metadata
    2023-09-18 18:29:04,330 - DataSourceVMware.py[DEBUG]: Getting guestinfo value for key userdata with /usr/bin/vmtoolsd
    2023-09-18 18:29:04,330 - subp.py[DEBUG]: Running command ['/usr/bin/vmtoolsd', '--cmd', 'info-get guestinfo.userdata'] with allowed return codes [0] (shell=False, capture=True)
    2023-09-18 18:29:04,373 - DataSourceVMware.py[DEBUG]: Getting guestinfo value for key userdata.encoding with /usr/bin/vmtoolsd
    2023-09-18 18:29:04,373 - subp.py[DEBUG]: Running command ['/usr/bin/vmtoolsd', '--cmd', 'info-get guestinfo.userdata.encoding'] with allowed return codes [0] (shell=False, capture=True)
    2023-09-18 18:29:04,411 - DataSourceVMware.py[DEBUG]: Getting encoded data for key=guestinfo.userdata, enc=gzip+base64
    2023-09-18 18:29:04,411 - DataSourceVMware.py[DEBUG]: Decoding gzip+base64 format guestinfo.userdata
    2023-09-18 18:29:04,411 - DataSourceVMware.py[DEBUG]: Getting guestinfo value for key vendordata with /usr/bin/vmtoolsd
    ```

* With `/usr/bin/vmware-rpctool` still replaced, I validated `ds-identify` by running `sudo rm /var/run/cloud-init/.ds-identify.result /var/run/cloud-init/ds-identify.log` and then running `sudo /usr/lib/cloud-init/ds-identify`. The resulting log fail showed the VMware datasource was still detected:

    ```shell
    $ sudo stat /var/run/cloud-init/ds-identify.log 
      File: /var/run/cloud-init/ds-identify.log
      Size: 1360      	Blocks: 8          IO Block: 4096   regular file
    Device: 0,23	Inode: 2318        Links: 1
    Access: (0640/-rw-r-----)  Uid: (    0/    root)   Gid: (    0/    root)
    Access: 2023-09-18 19:54:51.286717602 +0000
    Modify: 2023-09-18 19:54:48.650490201 +0000
    Change: 2023-09-18 19:54:48.650490201 +0000
     Birth: 2023-09-18 19:54:48.498477087 +0000
    ```
    
    ```
    [up 1725229.52s] ds-identify 
    policy loaded: mode=search report=false found=all maybe=all notfound=disabled
    /etc/cloud/cloud.cfg set datasource_list: ['NoCloud', 'ConfigDrive', 'OpenStack', 'VMware', None]
    DMI_PRODUCT_NAME=VMware20,1
    DMI_SYS_VENDOR=VMware, Inc.
    DMI_PRODUCT_SERIAL=VMware-42 1a b5 b3 64 97 90 e6-35 7b d8 97 66 df c3 d0
    DMI_PRODUCT_UUID=b3b51a42-9764-e690-357b-d89766dfc3d0
    PID_1_PRODUCT_NAME=unavailable
    DMI_CHASSIS_ASSET_TAG=No Asset Tag
    DMI_BOARD_NAME=440BX Desktop Reference Platform
    FS_LABELS=
    ISO9660_DEVS=
    KERNEL_CMDLINE=BOOT_IMAGE=/boot/vmlinuz-6.1.10-10.ph5-esx root=PARTUUID=c724cded-cf33-4d1c-b510-59478c322f84 init=/lib/systemd/systemd rcupdate.rcu_expedited=1 rw systemd.show_status=0 quiet noreplace-smp cpu_init_udelay=0 net.ifnames=0 plymouth.enable=0 systemd.unified_cgroup_hierarchy=yes
    VIRT=vmware
    UNAME_KERNEL_NAME=Linux
    UNAME_KERNEL_RELEASE=6.1.10-10.ph5-esx
    UNAME_KERNEL_VERSION=#1-photon SMP Mon Apr 24 22:51:08 UTC 2023
    UNAME_MACHINE=x86_64
    UNAME_NODENAME=photon5-w-tpm
    UNAME_OPERATING_SYSTEM=GNU/Linux
    DSNAME=
    DSLIST=NoCloud ConfigDrive OpenStack VMware None
    MODE=search
    ON_FOUND=all
    ON_MAYBE=all
    ON_NOTFOUND=disabled
    pid=24540 ppid=24539
    is_container=false
    is_ds_enabled(IBMCloud) = false.
    is_ds_enabled(IBMCloud) = false.
    check for 'VMware' returned found
    Found single datasource: VMware
    [up 1725229.68s] returning 0
    ``` 

## Checklist:

### DIFF
--- a/cloudinit/sources/DataSourceOVF.py
+++ b/cloudinit/sources/DataSourceOVF.py
@@ -8,6 +8,11 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
+"""Cloud-Init DataSource for OVF
+
+This module provides a cloud-init datasource for OVF data.
+"""
+
 import base64
 import os
 import re
@@ -20,7 +25,6 @@ LOG = logging.getLogger(__name__)
 
 
 class DataSourceOVF(sources.DataSource):
-
     dsname = "OVF"
 
     def __init__(self, sys_cfg, distro, paths):
@@ -142,7 +146,7 @@ def read_ovf_environment(contents, read_network=False):
     cfg_props = ["password"]
     md_props = ["seedfrom", "local-hostname", "public-keys", "instance-id"]
     network_props = ["network-config"]
-    for (prop, val) in props.items():
+    for prop, val in props.items():
         if prop == "hostname":
             prop = "local-hostname"
         if prop in md_props:
@@ -220,10 +224,9 @@ def maybe_cdrom_device(devname):
 # Transport functions are called with no arguments and return
 # either None (indicating not present) or string content of an ovf-env.xml
 def transport_iso9660(require_iso=True):
-
     # Go through mounts to see if it was already mounted
     mounts = util.mounts()
-    for (dev, info) in mounts.items():
+    for dev, info in mounts.items():
         fstype = info["fstype"]
         if fstype != "iso9660" and require_iso:
             continue
@@ -258,22 +261,85 @@ def transport_iso9660(require_iso=True):
     return None
 
 
+def exec_vmware_rpctool(rpctool, arg):
+    cmd = [rpctool, arg]
+    (stdout, stderr) = subp.subp(cmd)
+    return (cmd, stdout, stderr)
+
+
+def exec_vmtoolsd(rpctool, arg):
+    cmd = [rpctool, "--cmd", arg]
+    (stdout, stderr) = subp.subp(cmd)
+    return (cmd, stdout, stderr)
+
+
 def transport_vmware_guestinfo():
-    rpctool = "vmware-rpctool"
-    not_found = None
-    if not subp.which(rpctool):
-        return not_found
-    cmd = [rpctool, "info-get guestinfo.ovfEnv"]
+    rpctool, rpctool_fn = None, None
+    vmtoolsd = subp.which("vmtoolsd")
+    vmware_rpctool = subp.which("vmware-rpctool")
+
+    # Default to using vmware-rpctool if it is available.
+    if vmware_rpctool:
+        rpctool, rpctool_fn = vmware_rpctool, exec_vmware_rpctool
+        LOG.debug("discovered vmware-rpctool: %s", vmware_rpctool)
+
+    if vmtoolsd:
+        # Default to using vmtoolsd if it is available and vmware-rpctool is
+        # not.
+        if not vmware_rpctool:
+            rpctool, rpctool_fn = vmtoolsd, exec_vmtoolsd
+        LOG.debug("discovered vmtoolsd: %s", vmtoolsd)
+
+    # If neither vmware-rpctool nor vmtoolsd are available, then nothing can
+    # be done.
+    if not rpctool:
+        LOG.debug("no rpctool discovered")
+        return None
+
+    def query_guestinfo(rpctool, rpctool_fn):
+        LOG.info("query guestinfo.ovfEnv with %s", rpctool)
+        try:
+            cmd, stdout, _ = rpctool_fn(rpctool, "info-get guestinfo.ovfEnv")
+            if stdout:
+                return stdout
+            LOG.debug("cmd %s exited 0 with empty stdout", cmd)
+            return None
+        except subp.ProcessExecutionError as error:
+            if error.exit_code != 1:
+                LOG.warning("%s exited with code %d", rpctool, error.exit_code)
+            raise error
+
     try:
-        out, _err = subp.subp(cmd)
-        if out:
-            return out
-        LOG.debug("cmd %s exited 0 with empty stdout: %s", cmd, out)
-    except subp.ProcessExecutionError as e:
-        if e.exit_code != 1:
-            LOG.warning("%s exited with code %d", rpctool, e.exit_code)
-            LOG.debug(e)
-    return not_found
+        # The first attempt to query guestinfo could occur via either
+        # vmware-rpctool *or* vmtoolsd.
+        return query_guestinfo(rpctool, rpctool_fn)
+    except subp.ProcessExecutionError as error:
+        # The second attempt to query guestinfo can only occur with
+        # vmtoolsd.
+
+        # If the first attempt at getting the data was with vmtoolsd, then
+        # no second attempt is made.
+        if vmtoolsd and rpctool == vmtoolsd:
+            # The fallback failed, log the error.
+            util.logexc(
+                LOG, "vmtoolsd failed to get guestinfo.ovfEnv: %s", error
+            )
+            return None
+
+        if not vmtoolsd:
+            LOG.info("vmtoolsd fallback option not present")
+            return None
+
+        try:
+            LOG.info("fallback to vmtoolsd")
+            return query_guestinfo(vmtoolsd, exec_vmtoolsd)
+        except subp.ProcessExecutionError as error:
+            # The fallback failed, log the error.
+            util.logexc(
+                LOG, "vmtoolsd failed to get guestinfo.ovfEnv: %s", error
+            )
+
+    return None
 
 
 def find_child(node, filter_func):
@@ -287,7 +353,6 @@ def find_child(node, filter_func):
 
 
 def get_properties(contents):
-
     dom = minidom.parseString(contents)
     if dom.documentElement.localName != "Environment":
         raise XmlError("No Environment Node")
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -89,7 +89,6 @@ DATA_ACCESS_METHOD_ENVVAR = "envvar"
 DATA_ACCESS_METHOD_GUESTINFO = "guestinfo"
 DATA_ACCESS_METHOD_IMC = "imc"
 
-VMWARE_RPCTOOL = which("vmware-rpctool")
 REDACT = "redact"
 CLEANUP_GUESTINFO = "cleanup-guestinfo"
 VMX_GUESTINFO = "VMX_GUESTINFO"
@@ -146,7 +145,8 @@ class DataSourceVMware(sources.DataSource):
 
         self.cfg = {}
         self.data_access_method = None
-        self.vmware_rpctool = VMWARE_RPCTOOL
+        self.rpctool = None
+        self.rpctool_fn = None
 
         # A list includes all possible data transports, each tuple represents
         # one data transport type. This datasource will try to get data from
@@ -237,7 +237,7 @@ class DataSourceVMware(sources.DataSource):
 
         # Reflect any possible local IPv4 or IPv6 addresses in the guest
         # info.
-        advertise_local_ip_addrs(host_info)
+        advertise_local_ip_addrs(host_info, self.rpctool, self.rpctool_fn)
 
         # Ensure the metadata gets updated with information about the
         # host, including the network interfaces, default IP addresses,
@@ -313,7 +313,9 @@ class DataSourceVMware(sources.DataSource):
             keys_to_redact = self.metadata[CLEANUP_GUESTINFO]
 
         if self.data_access_method == DATA_ACCESS_METHOD_GUESTINFO:
-            guestinfo_redact_keys(keys_to_redact, self.vmware_rpctool)
+            guestinfo_redact_keys(
+                keys_to_redact, self.rpctool, self.rpctool_fn
+            )
 
     def get_envvar_data_fn(self):
         """
@@ -331,13 +333,77 @@ class DataSourceVMware(sources.DataSource):
         """
         check to see if there is data via the guestinfo transport
         """
-        md, ud, vd = None, None, None
-        if self.vmware_rpctool:
-            md = guestinfo("metadata", self.vmware_rpctool)
-            ud = guestinfo("userdata", self.vmware_rpctool)
-            vd = guestinfo("vendordata", self.vmware_rpctool)
 
-        return (md, ud, vd)
+        vmtoolsd = which("vmtoolsd")
+        vmware_rpctool = which("vmware-rpctool")
+
+        # Default to using vmware-rpctool if it is available.
+        if vmware_rpctool:
+            self.rpctool = vmware_rpctool
+            self.rpctool_fn = exec_vmware_rpctool
+            LOG.debug("discovered vmware-rpctool: %s", vmware_rpctool)
+
+        if vmtoolsd:
+            # Default to using vmtoolsd if it is available and vmware-rpctool
+            # is not.
+            if not vmware_rpctool:
+                self.rpctool = vmtoolsd
+                self.rpctool_fn = exec_vmtoolsd
+            LOG.debug("discovered vmtoolsd: %s", vmtoolsd)
+
+        # If neither vmware-rpctool nor vmtoolsd are available, then nothing
+        # can be done.
+        if not self.rpctool:
+            LOG.debug("no rpctool discovered")
+            return (None, None, None)
+
+        def query_guestinfo(rpctool, rpctool_fn):
+            md, ud, vd = None, None, None
+            LOG.info("query guestinfo with %s", rpctool)
+            md = guestinfo("metadata", rpctool, rpctool_fn)
+            ud = guestinfo("userdata", rpctool, rpctool_fn)
+            vd = guestinfo("vendordata", rpctool, rpctool_fn)
+            return md, ud, vd
+
+        try:
+            # The first attempt to query guestinfo could occur via either
+            # vmware-rpctool *or* vmtoolsd.
+            return query_guestinfo(self.rpctool, self.rpctool_fn)
+        except Exception as error:
+            util.logexc(
+                LOG,
+                "Failed to query guestinfo with %s: %s",
+                self.rpctool,
+                error,
+            )
+
+            # The second attempt to query guestinfo can only occur with
+            # vmtoolsd.
+
+            # If the first attempt at getting the data was with vmtoolsd, then
+            # no second attempt is made.
+            if vmtoolsd and self.rpctool == vmtoolsd:
+                return (None, None, None)
+
+            if not vmtoolsd:
+                LOG.info("vmtoolsd fallback option not present")
+                return (None, None, None)
+
+            LOG.info("fallback to vmtoolsd")
+            self.rpctool = vmtoolsd
+            self.rpctool_fn = exec_vmtoolsd
+
+            try:
+                return query_guestinfo(self.rpctool, self.rpctool_fn)
+            except Exception:
+                util.logexc(
+                    LOG,
+                    "Failed to query guestinfo with %s: %s",
+                    self.rpctool,
+                    error,
+                )
+
+                return (None, None, None)
 
     def get_imc_data_fn(self):
         """
@@ -447,25 +513,25 @@ def get_none_if_empty_val(val):
     return val
 
 
-def advertise_local_ip_addrs(host_info):
+def advertise_local_ip_addrs(host_info, rpctool, rpctool_fn):
     """
     advertise_local_ip_addrs gets the local IP address information from
     the provided host_info map and sets the addresses in the guestinfo
     namespace
     """
-    if not host_info:
+    if not host_info or not rpctool or not rpctool_fn:
         return
 
     # Reflect any possible local IPv4 or IPv6 addresses in the guest
     # info.
     local_ipv4 = host_info.get(LOCAL_IPV4)
     if local_ipv4:
-        guestinfo_set_value(LOCAL_IPV4, local_ipv4)
+        guestinfo_set_value(LOCAL_IPV4, local_ipv4, rpctool, rpctool_fn)
         LOG.info("advertised local ipv4 address %s in guestinfo", local_ipv4)
 
     local_ipv6 = host_info.get(LOCAL_IPV6)
     if local_ipv6:
-        guestinfo_set_value(LOCAL_IPV6, local_ipv6)
+        guestinfo_set_value(LOCAL_IPV6, local_ipv6, rpctool, rpctool_fn)
         LOG.info("advertised local ipv6 address %s in guestinfo", local_ipv6)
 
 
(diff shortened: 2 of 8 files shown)


akutz:
@TheRealFalcon I think the integration test is failing for an unrelated reason. I looked into it, and it's a wireguard module test.

akutz:
Okay @TheRealFalcon , it's ready for review. Cc @PengpengSun

akutz:
@TheRealFalcon / @PengpengSun,

I spent an hour refactoring DataSourceVMware a bit to support better unit testing around the fallback case. All tests pass.

cloudinit/sources/DataSourceOVF.py
```diff
@@ -259,21 +269,43 @@ def transport_iso9660(require_iso=True):
 
 
 def transport_vmware_guestinfo():
-    rpctool = "vmware-rpctool"
-    not_found = None
-    if not subp.which(rpctool):
-        return not_found
-    cmd = [rpctool, "info-get guestinfo.ovfEnv"]
+    is_vmtoolsd = False
```
TheRealFalcon, line 272: It's a little hard for me to follow the logic here, and it feels a little more strongly coupled than needed.

What about something like this:
```diff
diff --git a/cloudinit/sources/DataSourceOVF.py b/cloudinit/sources/DataSourceOVF.py
index 7b9bf65af..53a544458 100644
--- a/cloudinit/sources/DataSourceOVF.py
+++ b/cloudinit/sources/DataSourceOVF.py
@@ -269,29 +269,22 @@ def transport_iso9660(require_iso=True):
 
 
 def transport_vmware_guestinfo():
-    is_vmtoolsd = False
-    rpctool = subp.which("vmware-rpctool")
-
-    if not rpctool:
-        rpctool = subp.which("vmtoolsd")
-        is_vmtoolsd = True
-
-    if not rpctool:
-        return None
-
-    try:
-        return transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd)
-    except subp.ProcessExecutionError:
-        rpctool = subp.which("vmtoolsd")
-        is_vmtoolsd = True
-        return transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd)
+    rpctool_path = subp.which("vmware-rpctool")
+    if rpctool_path:
+        result = transport_vmware_guestinfo_rpctool(rpctool_path, "")
+        if result:
+            return result
+    vmtoolsd_path = subp.which("vmtoolsd")
+    if vmtoolsd_path:
+        return transport_vmware_guestinfo_rpctool(vmtoolsd_path, "--cmd")
+    return None
 
 
-def transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd=False):
+def transport_vmware_guestinfo_rpctool(rpctool, cmd_arg=""):
     args = [rpctool]
-    if is_vmtoolsd:
-        args.append("--cmd")
-    args.append("info-get guestinfo.ovfEnv")
+    if cmd_arg:
+        args.append(cmd_arg)
+    args.extend(["info-get", "guestinfo.ovfEnv"])
 
     try:
         out, _err = subp.subp(args)
@@ -302,8 +295,6 @@ def transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd=False):
         if e.exit_code != 1:
             LOG.warning("%s exited with code %d", rpctool, e.exit_code)
             LOG.debug(e)
-        if not is_vmtoolsd:
-            raise e
 
     return None
 
```

cloudinit/sources/DataSourceOVF.py
```diff
@@ -259,21 +269,43 @@ def transport_iso9660(require_iso=True):
...
+        rpctool = subp.which("vmtoolsd")
+        is_vmtoolsd = True
+        return transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd)
+
+
+def transport_vmware_guestinfo_rpctool(rpctool, is_vmtoolsd=False):
+    args = [rpctool]
+    if is_vmtoolsd:
+        args.append("--cmd")
+    args.append("info-get guestinfo.ovfEnv")
```
TheRealFalcon, line 294: rather than a space, these need to be two separate arguments in the list, correct?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -61,6 +61,13 @@
     why netifaces was used in the first place in order to either smooth the
     transition away from netifaces or embrace it further up the cloud-init
     stack.
+
+This module may be unit tested by running the following command from the root
+of the Cloud-Init repository:
+
+    make clean_pyc && PYTHONPATH="$(pwd)" \
```
TheRealFalcon, line 68: Same comment about testing with tox

cloudinit/sources/DataSourceVMware.py
```diff
@@ -507,46 +543,70 @@ def guestinfo_envvar_get_value(key):
...
+            # vmware-rpctool, then raise the exception so the caller can try
+            # again with vmtoolsd.
             util.logexc(
                 LOG,
                 "Failed to get guestinfo value for key %s: %s",
                 key,
                 error,
             )
+            if not is_rpctool_vmtoolsd(rpctool):
+                raise error
```
TheRealFalcon, line 609: There's a few places where this function is called but there's no fallback logic. That would result in an exception where previously it would just return None.

I'm also not sure it makes sense to check every single call. Presumably, if rpctool is going to fail for one call, it would fail for every call, correct? Does a call exist to get the version or something like that? If call succeeds, use rpctool everywhere. If not, use vmtoolsd. If so, we could do that during class initialization to figure out which tool to use. Then for any future calls, we could stick with one tool.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -571,24 +631,31 @@ def guestinfo_set_value(key, value, vmware_rpctool=VMWARE_RPCTOOL):
...
+        # again with vmtoolsd.
         util.logexc(
             LOG,
             "Failed to set guestinfo key=%s to value=%s: %s",
             key,
             value,
             error,
         )
+        if not is_rpctool_vmtoolsd(rpctool):
+            raise error
```
TheRealFalcon, line 658: See my comment about `guestinfo_get_value`

doc/rtd/reference/datasources/ovf.rst
```diff
@@ -6,6 +6,33 @@ OVF
...
+-------------------------
+
+The datasource initially attempts to use the program ``vmware-rpctool`` if it
+is available. However, if the program returns a non-zero exit code, then the
+datasource falls back to using the program ``vmtoolsd`` with the ``--cmd``
+argument.
+
+On some older versions of ESXi and open-vm-tools, the ``vmware-rpctool``
+program is much more performant than ``vmtoolsd``. While this gap was
+closed, it is not reasonable to expect the guest where Cloud-Init is running to
```
TheRealFalcon, line 19: Per our docs standards:
```suggestion
closed, it is not reasonable to expect the guest where cloud-init is running to
```

doc/rtd/reference/datasources/vmware.rst
```diff
@@ -148,6 +148,30 @@ Features
...
+-------------------------
+
+The datasource initially attempts to use the program ``vmware-rpctool`` if it
+is available. However, if the program returns a non-zero exit code, then the
+datasource falls back to using the program ``vmtoolsd`` with the ``--cmd``
+argument.
+
+On some older versions of ESXi and open-vm-tools, the ``vmware-rpctool``
+program is much more performant than ``vmtoolsd``. While this gap was
+closed, it is not reasonable to expect the guest where Cloud-Init is running to
```
TheRealFalcon, line 161: Same as above
```suggestion
closed, it is not reasonable to expect the guest where cloud-init is running to
```

tools/ds-identify
```diff
@@ -972,14 +972,78 @@ vmware_guest_customization() {
...
+
+vmware_guestinfo() {
+    vmware_rpctool_guestinfo "${1}" || vmware_vmtoolsd_guestinfo "${1}"
+}
+
+vmware_guestinfo_err() {
+    vmware_rpctool_guestinfo_err "${1}" || vmware_vmtoolsd_guestinfo_err "${1}"
+}
+
+vmware_guestinfo_ovfenv() {
```
TheRealFalcon, line 1007: I'm only ever seeing the "err" case used here. Can we remove this function entirely and replace it with `vmware_guestinfo_err "ovfEnv"` ?

tools/ds-identify
```diff
@@ -1486,16 +1530,16 @@ dscheck_VMware() {
         return "${DS_NOT_FOUND}"
     fi
 
-    # Do not proceed if the vmware-rpctool command is not present.
-    if ! vmware_has_rpctool; then
+    # Do not proceed if neither the vmware-rpctool or vmtoolsd command exists.
+    if [ ! vmware_has_rpctool ] && [ ! vmware_has_vmtoolsd ]; then
```
TheRealFalcon, line 1534: Nit: brackets aren't needed here

TheRealFalcon (changes requested):
@akutz Thanks for this change! I left some inline comments, but good change overall.

cloudinit/sources/DataSourceVMware.py
akutz: Sure, but I view the logic as low-risk for the way it is organized. It only decides on which tool to use once, but to do what you're suggesting would essentially still require the use of an if/then, pointing to which function to call each time. Either way there's an if/then.

cloudinit/sources/DataSourceVMware.py
akutz: See my response :)

tools/ds-identify
akutz: Yep. I just did it to be consistent with the others.

cloudinit/sources/DataSourceVMware.py
TheRealFalcon: > It only decides on which tool to use once

It's possible I haven't grokked the datasource enough, but I think `guesteinfo_set_value` is a better example. It still has the raise exception in it, but has 4 different calls, 2 in `advertise_local_ip_addrs`, and 2 in `guestinfo_redact_keys`. I don't see exception handling around either of those, so what happens if we raise the exception? Rather than tracing through every possible exception path, I think it's easier to set it once instead of possibly falling back multiple times.

akutz: It won't be raised multiple times. I could remove it from setinfo and probably should. Once `self.rpctool` is set to `vmtoolsd` after the first exception, the only way it's raised again is if it is unrelated to fallback.

akutz: I see what you are saying. The other exception handlers shouldn't have to care about this decision. I guess I was getting a little cute with the way it was handled. Will refactor. Thanks!

cloudinit/sources/DataSourceOVF.py
akutz: Are you asking about `--cmd` and `info-get guestinfo.ovfEnv` or `info-get` and `guestinfo.ovfEnv`? If the latter, `info-get guestinfo.ovfEnv` is one arugment bookended by quotes.

cloudinit/sources/DataSourceOVF.py
akutz: Yeah, and if you feel strongly, I can do something like that. I wanted to minimize the refactor the DS I don't own. I know it has the same issue, hence why I took it on.

akutz: Please note you are incorrectly assuming that `info-get` and the next arg are separate. They aren't. See the examples in the command lines in the description.

doc/rtd/reference/datasources/vmware.rst
akutz: Is that something that can/should be lintable?

cloudinit/sources/DataSourceOVF.py
TheRealFalcon: It was the latter. Sounds good!

cloudinit/sources/DataSourceOVF.py
TheRealFalcon: > Yeah, and if you feel strongly, I can do something like that

If you agree I think it's preferable, but if you have qualms we don't need to go that way.

doc/rtd/reference/datasources/vmware.rst
TheRealFalcon: Good question! We'd need to extend a linter with our own rules which we haven't done before. Have you used anything that makes it easy?

akutz: For markdown, but not for rst. Let me dig into what you're using for rst.

cloudinit/sources/DataSourceOVF.py
PengpengSun: or we combine a cmd directly for either rpctool, and give it to transport_vmware_guestinfo_rpctool(cmd) to execute it, sorry for bad indentation if any.

def transport_vmware_guestinfo():
    rpctool = subp.which("vmware-rpctool")
    if rpctool:
        cmd = [rpctool, "info-get guestinfo.ovfEnv"]
        result = transport_vmware_guestinfo_rpctool(cmd)
        if result:
            return result
    rpctools = subp.which("vmtoolsd")
    if rpctools:
        cmd = [rpctool, "--cmd", "info-get guestinfo.ovfEnv"]
        result = transport_vmware_guestinfo_rpctool(cmd)
        if result:
            return result
    return None

akutz: I ended up resolving it a little differently than both of the above approaches. Interested in your thoughts.

akutz:
Hi @TheRealFalcon,

I am missing something. The tests are passing for me locally, but failing on GH actions.

TheRealFalcon:
@akutz 

From the `tests/unittests/sources/test_vmware.py::TestDataSourceVMwareIMC::test_get_data_cloudinit_metadata_json` failure:
```
E           cloudinit.subp.ProcessExecutionError: Unexpected error while running command.
E           Command: ['/usr/bin/vmware-rpctool', 'info-get guestinfo.metadata']
E           Exit code: 255
E           Reason: -
E           Stdout: 
E           Stderr: Error: /usr/bin/vmware-rpctool must be run inside a virtual machine on a VMware hypervisor product.
```

Are you missing a mock somewhere?

akutz:
> @akutz
> 
> From the `tests/unittests/sources/test_vmware.py::TestDataSourceVMwareIMC::test_get_data_cloudinit_metadata_json` failure:
> 
> ```
> E           cloudinit.subp.ProcessExecutionError: Unexpected error while running command.
> E           Command: ['/usr/bin/vmware-rpctool', 'info-get guestinfo.metadata']
> E           Exit code: 255
> E           Reason: -
> E           Stdout: 
> E           Stderr: Error: /usr/bin/vmware-rpctool must be run inside a virtual machine on a VMware hypervisor product.
> ```
> 
> Are you missing a mock somewhere?

Quite possibly? But why would it work locally?

TheRealFalcon:
> Quite possibly? But why would it work locally?

It looks like it might actually exist on these runners. I don't have it locally and the test passes for me, so I think the missing binary triggering the normal fallback, but the exit code on the runners is causing us to raise early because of the 255 exit code.

akutz:
> > Quite possibly? But why would it work locally?
> 
> It looks like it might actually exist on these runners. I don't have it locally and the test passes for me, so I think the missing binary triggering the normal fallback, but the exit code on the runners is causing us to raise early because of the 255 exit code.

Huh. I don't have it locally either. I tested for it being missing, not existing and returning 255. Let me think about that.

akutz:
> > > Quite possibly? But why would it work locally?
> > 
> > 
> > It looks like it might actually exist on these runners. I don't have it locally and the test passes for me, so I think the missing binary triggering the normal fallback, but the exit code on the runners is causing us to raise early because of the 255 exit code.
> 
> Huh. I don't have it locally either. I tested for it being missing, not existing and returning 255. Let me think about that.

@TheRealFalcon I was able to repro it locally by adding the same command to my `PATH` and have it exit with `255`. My latest change addresses the issue. I did not think there was any issue with raising the exception, but I forgot that this prevents additional sub-platforms from being found.

tools/ds-identify
```diff
@@ -1486,16 +1526,16 @@ dscheck_VMware() {
         return "${DS_NOT_FOUND}"
     fi
 
-    # Do not proceed if the vmware-rpctool command is not present.
-    if ! vmware_has_rpctool; then
+    # Do not proceed if neither the vmware-rpctool or vmtoolsd command exists.
+    if ! vmware_has_rpctool ] && ! vmware_has_vmtoolsd; then
```
TheRealFalcon, line 1530: There's still a close bracket left here (and I'm surprised it doesn't cause any failures)

akutz:
```shell
[0]akutz@akutzKFPN7:cloud-init $ ! false ] && ! true
[1]akutz@akutzKFPN7:cloud-init $ 
```

I am shocked as well (the number at the front of my prompt is the exit code from the previous command).

akutz: Looks like y'all need to implement shellcheck.net for linting. I use it in all my projects for this very reason. Also note it's a warning and not an error. That alone shocks me...

image

TheRealFalcon: Hmmm...I thought we did. I have it in my editor and you'll notice some `# shellcheck disable` annotations thrown around ds-identify. I guess it never made it into CI though.

TheRealFalcon: https://github.com/canonical/cloud-init/issues/4453

cloudinit/sources/DataSourceVMware.py
```diff
@@ -146,7 +145,8 @@ def __init__(self, sys_cfg, distro, paths, ud_proc=None):
 
         self.cfg = {}
         self.data_access_method = None
-        self.vmware_rpctool = VMWARE_RPCTOOL
+        self.rpctool = None
```
TheRealFalcon, line 148: I think this and the line after still have some potential to result in tracebacks. If neither is found or [we have envvars](https://github.com/canonical/cloud-init/blob/fc7b92ba7529cf749722bd5345c9bfdd6d4ca5fd/cloudinit/sources/DataSourceVMware.py#L159) (such that they never get a chance to get set), these will stay None. Then when they're called from places like [here](https://github.com/canonical/cloud-init/blob/fc7b92ba7529cf749722bd5345c9bfdd6d4ca5fd/cloudinit/sources/DataSourceVMware.py#L216) or [here](https://github.com/canonical/cloud-init/blob/fc7b92ba7529cf749722bd5345c9bfdd6d4ca5fd/cloudinit/sources/DataSourceVMware.py#L240), we'll try to call `self.rpctool_fn` with arguments, but it will still be `None`, resulting in a TypeError.

akutz: The first example cannot do that as the detected method is checked first in the redact function. I also just pushed an update that checks if the rpctool and rpctool_fn vars are set before using them in the advertise method.

akutz: To be clear, on the advertise func, I want that to work even if the guestinfo platform wasn't chosen but the capability _exists_. That's why the check I added only validates the rpctool vars.

akutz:
@TheRealFalcon I verified this works on a real system again. Please note the reason I just force pushed one more time is to add a little extra logging. It was not working on the real system, but the reason was not apparent. It ended up being because the current codebase has `b64` in `atomic_helper` not `util` where it used to be. My system had the old location. But that error was being swallowed. Anyway, hopefully this is finally good to go :)

cloudinit/sources/DataSourceVMware.py
akutz: And by the way, the entire purpose of the `self.rpctool_fn` was so that the function to invoke was decided exactly once and then the rest of the datasource never had to have another if statement about the logic.

akutz: @TheRealFalcon I was waiting for you to respond before resolving this.

TheRealFalcon: @akutz , yeah, that works. Thanks

akutz:
Hi @TheRealFalcon,

Please let me know if there is anything else needed to resolve this PR. Thanks!

TheRealFalcon (approved):
Thanks @akutz ! LGTM!

playerla:
Hi,
Have you been able to test with _secure_ RPC  ? That was the reason behind this MR (#4436)

Just got the daily build cloud-init_23.3.daily-202309271631-d9cdc298~ubuntu20.04.1_all.deb. As intented it fallbacks to vmtoolsd. But unexpectedly still get _Permission Denied_

I opened  #4475

PR #4447, opened 2023-09-19
integration test: allow pro service warnings

TheRealFalcon:
## Proposed Commit Message

```
integration test: allow pro service warnings

Even if pro is giving us warnings, that is showing that cloud-init
enabled things correctly.
```

## Additional Context
An integration test is failing due to livepatch warning that a kernel is not supported. While this may be a problem, it should be caught elseware and is not a problem in cloud-init.
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-jammy-ec2/306/testReport/tests.integration_tests.modules.test_ubuntu_advantage/TestUbuntuAdvantagePro/test_custom_services/


holmanb (approved):
Looks good, thanks for the fix!

PR #4450, opened 2023-09-21
net: allow NetworkManager configuration for NetworkManager net renderer

ani-sinha:
This is the equivalent of the change made in the following commit for sysconfig renderer but this time it is for network_manager renderer: 

`67bab5bb804e2346673 ("net: Allow for NetworkManager configuration")`

The reasons for making the change has already been outlined in the above commit. When cloud-init configures DNS and produces `/etc/resolv.conf`, NetworkManager, if enabled, can clobber the above file and could also render it empty. Hence, this change makes cloud-init add a NetworkManager configuration so that it disables management and handling of `/etc/resolv.conf` by NetworkManager.

Some code refactoring has been performed so as to avoid code duplication. Unit tests have been adjusted accordingly.

LP: #1693251


ani-sinha:
Our QE is testing this internally. So please feel free to review but do not push it yet.

ani-sinha:
After looking at the sysconfig and NM renderer code deeper and discussing this internally with QE, we found that sysconfig renderer writes DNS information to /etc/resolv.conf but NM renderer does not. When NM renderer is used, the DNS information is directly written into the NM keyfile and NetworkManager is responsible for managing DNS from that point onwards and writing /etc/resolv.conf. There is no need to keep NM out of the loop. Hence, there is no need to write /etc/NetworkManager/conf.d/99-cloud-init.conf from NM renderer. For sysconfig it is necessary to keep NM from touching /etc/resolv.conf separately and overwriting cloud-init generated file. Thus, this PR can be closed.

PR #4457, opened 2023-09-25
net: remove the word "on instance boot" from cloud-init generated config

ani-sinha:
cloud-init does not necessarily generate its configuration files upon instance boot. If cloud-init service is disabled or if cloud-init network configuration is disabled either using Linux kernel command line `network-config=disabled` or by cloud-init configuration

```
network:
  config: disabled
```
, cloud-init will not generate those config files. Hence, using the word "on instance boot" on the config file header could be misleading on those circumstances and confuse users. Remove it.

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -40,7 +40,7 @@ KNOWN_DISTROS = [
 
 def _make_header(sep="#"):
     lines = [
-        "Created by cloud-init on instance boot automatically, do not edit.",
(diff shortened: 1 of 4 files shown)


ani-sinha:
@TheRealFalcon once you push this, https://github.com/canonical/cloud-init/pull/4450 needs to be rebased and fixed. I can do that. But we need to push this first.

ani-sinha:
@holmanb any issues with this change?

TheRealFalcon (approved):
LGTM!

PR #4456, opened 2023-09-23
delete DSA support

xiaoge1001:
## Proposed Commit Message
OpenSSH 7.0 and greater similarly disable the ssh-dss (DSA) public key algorithm. It too is weak and we recommend against its use. 
refer:https://www.openssh.com/legacy.html

However, cloud-init still generates the DSA key.  It is recommended that cloud-init disable dsa

PR #4458, opened 2023-09-25
do not generate dsa in openeuler

xiaoge1001:
## Proposed Commit Message

OpenSSH 7.0 and greater similarly disable the ssh-dss (DSA) public key algorithm. It too is weak and we against its use in openEuler.

### DIFF
--- a/config/cloud.cfg.tmpl
+++ b/config/cloud.cfg.tmpl
@@ -109,6 +109,11 @@ syslog_fix_perms: ~
 disable_vmware_customization: false
 {% endif -%}
 
+# OpenSSH 7.0 and greater similarly disable the ssh-dss (DSA) public key algorithm. It too is weak and we against its use in openEuler.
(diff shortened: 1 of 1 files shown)


blackboxsw (commented):
Thank you for the PR submittals @xiaoge1001. While this is a good specific change for openeuler dropping support for dsa, I think upstream would probably prefer your more general proposal in #4456. We'll get a review on the branch and also have our security team review the implications as we also need to decide whether this changset is something we want to push into stable releases as well. We also want to think through the implications of any other downstream distributions that may still rely on legacy dsa support to ensure we aren't leaving them out in the cold.   I don't think #4456 should be an issue anymore to drop generally, but we want to do a little confirmation first before landing it.

blackboxsw:
Let's close this PR for the moment as I think we can work a general solution per your other Pull Request when we get review on the approach  there. If we need to, and #4456 is a not possible, then we'll re-open this issue.

PR #4459, opened 2023-09-26
Don't recommend using cloud-init as a library

holmanb:
```
Don't recommend using cloud-init as a library
```

## Additional Context
Whoever put that in the docs didn't know what they were talking about.

### DIFF
--- a/doc/rtd/reference/faq.rst
+++ b/doc/rtd/reference/faq.rst
@@ -81,16 +81,15 @@ IRC (``#ubuntu-server`` on Libera) or Discourse. For Juju support see their
 
 Can I use cloud-init as a library?
 ==================================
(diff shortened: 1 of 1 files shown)


s-makin (approved):
LGTM. My only question is do we need to provide the obvious reasons, or a link to the page that contains the obvious reasons? If no to both, happy to accept as-is.

holmanb:
> LGTM. My only question is do we need to provide the obvious reasons, or a link to the page that contains the obvious reasons? If no to both, happy to accept as-is.

My .02 is that they are already treading in water that we told them not to. If they don't understand why rsyslog getting setup might cause the cloud-init logger to need re-configuring, then that comment serves as a warning for them to do their research.

TheRealFalcon (approved):

Issue #4461, opened 2023-09-27
packages: cloud-config not longer accepts versioned lists as package entries


blackboxsw:
# Bug report

Recent commit 226ba258216466dc0c42d6a9ddb669c788131160 regressed the ability to provide specific versions in generic package install requests with the `packages:` directive in #cloud-config YAML.

## Steps to reproduce the problem
```
cat > pkg.yaml <<EOF
#cloud-config
packages:
  - [hello, 2.10-3]
EOF
lxc launch ubuntu-daily/mantic test-pkg-m
lxc exec test-pkg-m -- cloud-init status --wait --format json
# expect success
lxc exec test-pkg-m -- apt-get remove hello
# Update to daily PPA
lxc exec test-pkg-m -- add-apt-repository ppa:cloud-init-dev/daily -y
lxc exec test-pkg-m -- cloud-init clean --logs --reboot
lxc exec test-pkg-m -- cloud-init status --wait --format json
# Errors seen

  "_schema_version": "1",
  "boot_status_code": "enabled-by-generator",
  "datasource": "lxd",
  "detail": "('package_update_upgrade_install', ValueError(\"Invalid 'packages' yaml specification. Check schema definition.\"))",
  "errors": [
    "('package_update_upgrade_install', ValueError(\"Invalid 'packages' yaml specification. Check schema definition.\"))"
  ],
  "last_update": "Wed, 27 Sep 2023 03:19:47 +0000",
  "schemas": {
    "1": {
      "boot_status_code": "enabled-by-generator",
      "datasource": "lxd",
      "detail": "('package_update_upgrade_install', ValueError(\"Invalid 'packages' yaml specification. Check schema definition.\"))",
      "errors": [
        "('package_update_upgrade_install', ValueError(\"Invalid 'packages' yaml specification. Check schema definition.\"))"
      ],
      "last_update": "Wed, 27 Sep 2023 03:19:47 +0000",
      "status": "error"
    }
  },
  "status": "error"
}
```

The above procedure produces the following traceback on daily images:
```

kages/cloudinit/config/cc_package_update_upgrade_install.py'>) failed
2023-09-27 03:19:47,853 - util.py[DEBUG]: Running module package_update_upgrade_install (<module 'cloudinit.config.cc_package_update_upgrade_install' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_package_update_upgrade_install.py'>) failed
Traceback (most recent call last):
[...]

PR #4462, opened 2023-09-27
WIP: Snap package regression

blackboxsw:
WIP: needs integration test validation/fix.

Three separate commits fixing 3 failure conditions:
1. Fix jenkins failures in puppet integration tests now that distro.install_packages no longer accepts tuples
2. Wire _extract_package_manager to support and Distro subclass with it's own install_packages method in order to extract any generic packages: provided in cloud-config (and error on unsupported package_managers)
3. fix regression in 226ba258 by allowing per-package versions to be supplied in generic packages cloud-config
 ```
packages:
 - [pkg, ver]
```

## Proposed Commit Messages

```
*     packages: fix generic versioned package installs allowed
    
    Fix regression in commit 226ba258.
    
    User-data containing packages: can define specific package versions
    to install by providing a two-item list containing the package name and
    package version such as the following:
    
    packages:
    - [hello, 2.10-3]
    
    Add integration test covering this use case for the hello package on
    various supported Ubuntu releases.
    
    Fixes GH-4461

*     distros: invoke _extract_package_by_manager in all install_packages
    
    Distro subclasses like alpine, arch, bsd, redhat, suse need to
    parse packages: from cloud-config to separate generic package
    install requests for package manager specific install requests.
    
    This allows the packages listed to be aggregated as an installable
    list of generic packages.
    
    Given that only debian and ubuntu currently define
    Distro.package_managers attribute, allow extract_package_by_manager
    to raise an error when packages: cloud-config requests specific
    package managers that aren't supported by that Distro class.
    
    Allow util.expand_package_list to handle the sets of generic_packages
    returned by _extract_package_manager for alpined, arch, bsd, redhat and
    suse.

*  puppet: provide versioned package_install as a tuple in a list
    
    Since distro.install_packages operates on iterables, a versioned
    package install request shoud be provided as a single tuple as a
    list item to avoid iterating on each item in the tuple separately.

```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_puppet.py
+++ b/cloudinit/config/cc_puppet.py
@@ -240,7 +240,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             if package_name is None:  # conf has no package_nam
                 for puppet_name in PUPPET_PACKAGE_NAMES:
                     try:
-                        cloud.distro.install_packages((puppet_name, version))
+                        cloud.distro.install_packages([(puppet_name, version)])
                         package_name = puppet_name
                         break
                     except subp.ProcessExecutionError:
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -183,6 +183,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                             )
             elif isinstance(entry, str):
                 generic_packages.add(entry)
+            elif isinstance(entry, tuple) and len(entry) == 2:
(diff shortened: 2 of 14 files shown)


blackboxsw:
Fixes #4461

TheRealFalcon:
@blackboxsw , thanks for working on this. However, before you go any further, I don't think is the right approach. We can sync up on it later, but I think I have a simpler solution.

blackboxsw:
We'll table this PR and sort a @TheRealFalcon's separate solution for processing the package lists by manager to catch this gap.

PR #4464, opened 2023-09-27
Ubuntu/mantic 23.3.x

holmanb:
Do not merge.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,14 @@
+cloud-init (23.3.1-0ubuntu2) UNRELEASED; urgency=medium
+
+  * New upstream bug fix release based on 23.3.1.
+    + Install missing dependencies in apt module if required for minimal images.
+      - d/p/cpick-0d9f149a-Pytestify-apt-config-test-modules-4424
+      - d/p/cpick-5023e9f9-Refactor-test_apt_source_v1.py-to-use-pytest-4427
+      - d/p/cpick-e9cdd7e3-Install-gnupg-if-gpg-not-found-4431
+      - d/p/cpick-015543d3-apt-install-software-properties-common-when-absent-but
+
+ -- Brett Holman <brett.holman@canonical.com>  Thu, 28 Sep 2023 10:20:51 -0500
+
 cloud-init (23.3.1-0ubuntu1) mantic; urgency=medium
 
   * New upstream bug fix release based on 23.3.1.
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-cloud-init (23.3.1-0ubuntu2) UNRELEASED; urgency=medium
+cloud-init (23.3.1-0ubuntu2) mantic; urgency=medium
 
   * New upstream bug fix release based on 23.3.1.
     + Install missing dependencies in apt module if required for minimal images.
@@ -7,7 +7,7 @@ cloud-init (23.3.1-0ubuntu2) UNRELEASED; urgency=medium
       - d/p/cpick-e9cdd7e3-Install-gnupg-if-gpg-not-found-4431
       - d/p/cpick-015543d3-apt-install-software-properties-common-when-absent-but
 
- -- Brett Holman <brett.holman@canonical.com>  Thu, 28 Sep 2023 10:20:51 -0500
+ -- Brett Holman <brett.holman@canonical.com>  Thu, 28 Sep 2023 12:00:23 -0500
 
 cloud-init (23.3.1-0ubuntu1) mantic; urgency=medium
 
--- a/debian/patches/cpick-015543d3-apt-install-software-properties-common-when-absent-but
+++ b/debian/patches/cpick-015543d3-apt-install-software-properties-common-when-absent-but
@@ -1,0 +1,225 @@
+From 015543d304707907eb09fd286d76184cd21d066f Mon Sep 17 00:00:00 2001
+From: Chad Smith <chad.smith@canonical.com>
+Date: Thu, 21 Sep 2023 15:20:17 -0600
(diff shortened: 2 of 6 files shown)


blackboxsw (changes requested):
~~@holmanb I think we need to first provide a PR against upstream/23.3.x with this set of commits. Then when that approves/merges we'd new_upstream_snapshot down into upstream/ubuntu/mantic-23.3.x from there.~~
I recant that statement. 

This isn't a hot-fix of critical features that need to be in `upstream/23.3.`x. This is  onlysupplemental commits cherry-picked into `ubuntu/devel-23.3.x`. So this approach makes sense. Reviewing now

debian/changelog
```diff
@@ -1,3 +1,11 @@
+cloud-init (23.3.1-0ubuntu2) UNRELEASED; urgency=medium
+
+  * New upstream bug fix release based on 23.3.1.
+  * Install gnupg and software-properties-common
+    in cc_apt_configure if included in base image.
```
blackboxsw, line 5: We should provide the proper indent of scope prefix ` -` for the changes included in the new upstream bug fix. Let's also correct the message as we only install if the pkgs are absent and cloud-config requires those cmds

```suggestion
    + Install gnupg and software-properties-common
      if absent in base image but required by apt cloud-config
```

debian/changelog
```diff
@@ -1,3 +1,11 @@
+cloud-init (23.3.1-0ubuntu2) mantic; urgency=medium
```
blackboxsw, line 1: pkg ver good representing supplemental commits beyond upstream 23.3.1.

blackboxsw, line 1: pkg ver good representing supplemental commits beyond upstream 23.3.1.

blackboxsw (approved):
- APPROVED with minor text change in debian/changelog
- versioning looks good for mantic
- build-package/sbuild good
- content limited to bug fix release and unittests.

blackboxsw (changes requested):
Hrm, since we aren't changing the upstream version 23.3.1 of the package and we aren't providing an official upstream release for this as 23.3.2, we may actually have to represent this cherry picks as quilt patches in debian/patches/series * debian/patches/cpick-* files.

The way we'd do that on this branch is with uss-tableflip `cherry-pick` tool instead of `git cherry-pick`
that was we only upload a debdiff which contains quilt patches based on upstream 23.3.1.orig.tar.gz. We'll probably talk over this tomorrow. I'll drop the procedure in this PR later today.


Procedure to create this cherry-pick based branch since we are performing a debdiff upload and not a full new upstream snapshot for this release:
```
checkout -B upstream/ubuntu/mantic-23.3.x ubuntu/mantic-23.3.x
cherry-pick 0d9f149a3109a36d3ce41e120777ad592640754b -y
cherry-pick 5023e9f9cf01fdd9482e9743e8407dda9438558b -y
cherry-pick e9cdd7e39058f6ef03c38339aa5307335d3573ec -y 
cherry-pick 015543d304707907eb09fd286d76184cd21d066f  -y
dch -e # set you consolidated message
git rebase -i HEAD~9  # fixup/squash all 'update changelog' commits into your single "update changelog"
dch -r -D mantic ''
git commit -am 'releasing cloud-init version 23.3.1-0ubuntu2'
```

The uss-tableflip cherry-pick tool is not recommended for our typical release process because we generally speaking are SRU ing to stable releases via new_upstream_snapshot, or performing new_upstream_snapshots of tip of main into the devel series. 

Given that we are in [BETA FREEZE for mantic](https://discourse.ubuntu.com/t/mantic-minotaur-release-schedule/34989), we want to limit introduction of commits to only bug fixes. In this scenario, it makes sense the quilt patch to provide a simple debdiff against what is already released to mantic because we are not uploading a full new upstream 23.3.2.orig.tar.gz

https://github.com/canonical/uss-tableflip/blob/main/doc/ubuntu_release_process.md#cherry-picked-changes

blackboxsw:
I've pushed https://github.com/blackboxsw/cloud-init/tree/ubuntu/mantic-23.3.x as a comparison point

debian/changelog
```diff
@@ -1,3 +1,13 @@
...
+
+  * Install missing dependencies in apt module if required for minimal images.
+    + d/p/cpick-0d9f149a-Pytestify-apt-config-test-modules-4424
+    + d/p/cpick-5023e9f9-Refactor-test_apt_source_v1.py-to-use-pytest-4427
+    + d/p/cpick-e9cdd7e3-Install-gnupg-if-gpg-not-found-4431
+    + d/p/cpick-015543d3-apt-install-software-properties-common-when-absent-but
```
blackboxsw, line 7: Good representing the individual quilt patches as cloud-init historically needs to represent separate debian directory packaging changes explicitly. 
We also need a way to call out that this is a "New upstream bug fix release based on 23.3.1" 
Are you able to include that as a scoping entry somehow?


```suggestion
  * New upstream bug fix release based on 23.3.1
    + Install missing dependencies in apt module if required for minimal images
      by the following debian/patches:
      - cpick-0d9f149a-Pytestify-apt-config-test-modules-4424
      - cpick-5023e9f9-Refactor-test_apt_source_v1.py-to-use-pytest-4427
      - cpick-e9cdd7e3-Install-gnupg-if-gpg-not-found-4431
      - cpick-015543d3-apt-install-software-properties-common-when-absent-but
```

holmanb: @blackboxsw Good point, done.

blackboxsw (approved):
LGTM! ship it. sponsored upload done ` 23.3.1-0ubuntu2 (Waiting for approval)`

build-package/sbuild deb generated has right supplemental commits and proper version.

PR #4465, opened 2023-09-28
azure: check for primary interface when performing DHCP

cjp256:
For Savable PPS we rely on a connectivity check to IMDS to determine if the NIC is primary.  However, in some cases, connectivity may be delayed and we incorrectly assume the NIC is not primary.

Instead of relying on connectivity, check the DHCP-provided route configuration for the presence of Wireserver and/or IMDS IPs.

- Return bool from _setup_ephemeral_networking() indicating primary NIC or not.

- Remove _check_if_nic_is_primary() used in Savable PPS.

- Add a relevant diagnostic to assist debugging failures when a secondary NIC is chosen.

- Remove some tests that are redundant.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -376,7 +376,6 @@ class DataSourceAzure(sources.DataSource):
         primary = any(
             n in routed_networks
             for n in [
-                "168.63.129.16/32",
                 "169.254.169.254/32",
                 wireserver_route,
             ]
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -25,7 +25,7 @@ from cloudinit.net.dhcp import (
     NoDHCPLeaseInterfaceError,
     NoDHCPLeaseMissingDhclientError,
 )
-from cloudinit.net.ephemeral import EphemeralDHCPv4
+from cloudinit.net.ephemeral import EphemeralDHCPv4, EphemeralIPv4Network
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

anhvoms:
@holmanb can you review and merge? Thanks

holmanb:
> @holmanb can you review and merge? Thanks

Ack, sorry for the delay on this one. EOD now, but will do this tomorrow.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -364,14 +365,33 @@ def _get_subplatform(self):
...
+        if not ephipv4.static_routes:
+            # Primary nics must contain routes.
+            return False
+
+        routed_networks = [r[0] for r in ephipv4.static_routes]
+        wireserver_route = f"{self._wireserver_endpoint}/32"
+        primary = any(
+            n in routed_networks
+            for n in [
+                "168.63.129.16/32",
```
holmanb, line 379: Isn't this the same as the default `wireserver_route`? Why duplicate?

holmanb (approved):
I would also assume this performs a bit better too than getting metadata?

I left one question inline, but then I'll merge.

Looks good to me, thanks!

cloudinit/sources/DataSourceAzure.py
cjp256: You are correct, it should be

holmanb: Nice, thanks!

PR #4466, opened 2023-09-28
Fix regression in package installation

TheRealFalcon:
## Proposed Commit Message

```
Fix regression in package installation

This commit addresses 3 issues introduced by commit 226ba25.
1. We regressed the ability to provide specific
   versions in generic package installs with the `packages:`
   directive in #cloud-config YAML.
2. Package installation failure changed to logging error rather than
   raising Exception.
3. Package lists being passed to package installation became constrained
   to `packages` schema when it was originally more open.

To fix 1, added _validate_entry function to ensure packages specified
at both the top level and under a package manager are parsed the same.

To fix 2, raise Exception rather than log error if package installation
fails. Also, change any intermediary package install failures to INFO
rather than ERROR.

To fix 3, apply typing to the pkglist accepted by `install_packages` to
(mostly) adhere to the `packages` schema defition so that we're using a
consistent interface. Call sites have been updated accordingly.

Fixes GH-4461
```

## Additional Context
Main changes are in `distros/__init__.py`. Pretty much everything else is updating call sites or typing appropriately. I also took some tests from @blackboxsw .

Also ensured `tests/unittests/config/test_cc_package_update_upgrade_install.py` and `tests/integration_tests/modules/test_puppet.py` pass.

### DIFF
--- a/cloudinit/config/cc_puppet.py
+++ b/cloudinit/config/cc_puppet.py
@@ -11,8 +11,10 @@
 import logging
 import os
 import socket
+from contextlib import suppress
 from io import StringIO
 from textwrap import dedent
+from typing import List, Union
 
 import yaml
 
@@ -20,7 +22,7 @@ from cloudinit import helpers, subp, temp_utils, url_helper, util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema, get_meta_doc
-from cloudinit.distros import ALL_DISTROS, Distro
+from cloudinit.distros import ALL_DISTROS, Distro, InstallerError
 from cloudinit.settings import PER_INSTANCE
 
 AIO_INSTALL_URL = "https://raw.githubusercontent.com/puppetlabs/install-puppet/main/install.sh"  # noqa: E501
@@ -237,21 +239,28 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         )
 
         if install_type == "packages":
+            to_install: List[Union[str, List[str]]]
             if package_name is None:  # conf has no package_nam
                 for puppet_name in PUPPET_PACKAGE_NAMES:
-                    try:
-                        cloud.distro.install_packages((puppet_name, version))
+                    with suppress(InstallerError):
+                        to_install = (
+                            [[puppet_name, version]]
+                            if version
+                            else [puppet_name]
+                        )
+                        cloud.distro.install_packages(to_install)
                         package_name = puppet_name
                         break
-                    except subp.ProcessExecutionError:
-                        pass
                 if not package_name:
                     LOG.warning(
                         "No installable puppet package in any of: %s",
                         ", ".join(PUPPET_PACKAGE_NAMES),
                     )
             else:
-                cloud.distro.install_packages((package_name, version))
+                to_install = (
+                    [[package_name, version]] if version else [package_name]
+                )
+                cloud.distro.install_packages(to_install)
 
         elif install_type == "aio":
             install_puppet_aio(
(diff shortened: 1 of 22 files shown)


cloudinit/config/cc_puppet.py
```diff
@@ -237,21 +239,28 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         )
 
         if install_type == "packages":
+            to_install: List[Union[str, List[str]]]
             if package_name is None:  # conf has no package_nam
                 for puppet_name in PUPPET_PACKAGE_NAMES:
-                    try:
-                        cloud.distro.install_packages((puppet_name, version))
+                    with suppress(InstallerError):
```
blackboxsw, line 245: Smart approach w/ the suppress handling as we loop over the install_packages calls.

cloudinit/distros/__init__.py
```diff
@@ -104,6 +105,24 @@
...
+# The Immutable types mentioned there won't work for us because
+# we need to distinguish between a str and a Sequence[str]
+# This also isn't exhaustive. If you have a unique case that adheres to
+# the `packages` schema, you can add it here.
+PackageList = Union[
+    List[str],
+    List[Mapping],
+    List[List[str]],
+    List[Union[str, List[str]]],
+    List[Union[str, List[str], Mapping]],
```
blackboxsw, line 119: good to see tuples excluded from this Union. No longer need to support yet another way to provide package configuration to distro.install_packages

cloudinit/distros/__init__.py
```diff
@@ -156,6 +175,17 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+        elif isinstance(entry, (list, tuple)):
+            if len(entry) == 2:
+                return tuple(entry)
+        raise ValueError(
+            "Invalid 'packages' yaml specification. "
+            "Check schema definition."
+        )
+
     def _extract_package_by_manager(
         self, pkglist: Iterable
```
blackboxsw, line 190: his type can now be set to PackageList right?

cloudinit/distros/__init__.py
```diff
@@ -217,23 +241,23 @@ def install_packages(self, pkglist):
...
-                    pkglist=packages
-                )
+                manager_type.from_config(
+                    self._runner, self._cfg
+                ).install_packages(pkglist=packages)
             )
 
         if uninstalled:
-            LOG.error(error_message, uninstalled)
+            raise InstallerError(error_message, uninstalled)
```
blackboxsw, line 260: This needs to apply the formatting to the string
```suggestion
            raise InstallerError(error_message % uninstalled)
```

tests/unittests/distros/test_init.py
```diff
@@ -306,13 +316,23 @@ def test_non_default_package_manager_fail(
             "cloudinit.distros.package_management.snap.Snap.install_packages",
             return_value=["pkg3"],
         )
-        _get_distro("debian").install_packages(
-            [{"apt": ["pkg1"]}, "pkg2", {"snap": ["pkg3"]}]
-        )
+        with pytest.raises(InstallerError):
```
blackboxsw, line 319: Let's add a match here that would have caught the failure in InstallerError message initialization.
```suggestion
        with pytest.raises(
            InstallerError,
            match="Failed to install the following packages: \['pkg3'\]",
        ):
```

blackboxsw (changes requested):
Thanks @TheRealFalcon, couple of nits and good to go.

TheRealFalcon:
@blackboxsw , I believe I addressed all of your comments

blackboxsw (approved):
LGTM!

PR #4468, opened 2023-09-28
azure/imds: ensure new errors are logged immediately when retrying

cjp256:
When retrying to fetch metadata from IMDS there is a backoff for logging to avoid flooding the logs.  In the current implementation we have to wait until the next backoff log to see the transition.

Always log when the current error does not match the previous one.

### DIFF
--- a/cloudinit/sources/azure/imds.py
+++ b/cloudinit/sources/azure/imds.py
@@ -4,7 +4,7 @@
 
 import logging
 from time import time
-from typing import Dict, Optional
+from typing import Dict, Optional, Type, Union
 
 import requests
 
@@ -45,6 +45,7 @@ class ReadUrlRetryHandler:
         self.retry_deadline = retry_deadline
         self._logging_threshold = 1.0
         self._request_count = 0
+        self._last_error: Union[None, Type, int] = None
 
     def exception_callback(self, req_args, exception) -> bool:
         self._request_count += 1
@@ -79,6 +80,19 @@ class ReadUrlRetryHandler:
             self._logging_threshold *= self.logging_backoff
         else:
             log = False
+
+        # Always log if error does not match previous.
+        if exception.code is not None:
+            if self._last_error != exception.code:
+                log = True
+                self._last_error = exception.code
+        elif (
+            self._last_error is None
+            or not isinstance(self._last_error, type)
+            or not isinstance(exception.cause, self._last_error)
+        ):
+            log = True
+            self._last_error = type(exception.cause)
 
         if log or not retry:
             report_diagnostic_event(
--- a/cloudinit/sources/azure/imds.py
+++ b/cloudinit/sources/azure/imds.py
@@ -83,12 +83,16 @@ class ReadUrlRetryHandler:
 
         # Always log if error does not match previous.
         if exception.code is not None:
+            # This is an HTTP response with failing code, log if different.
             if self._last_error != exception.code:
                 log = True
                 self._last_error = exception.code
         elif (
+            # No previous error to match against.
             self._last_error is None
+            # Previous error is exception code (int).
             or not isinstance(self._last_error, type)
+            # Previous error is different class.
             or not isinstance(exception.cause, self._last_error)
         ):
             log = True
--- a/tests/unittests/sources/azure/test_imds.py
+++ b/tests/unittests/sources/azure/test_imds.py
@@ -481,6 +481,46 @@ class TestFetchMetadataWithApiFallback:
             )
         ]
 
+    def test_logs_unique_errors(
+        self,
+        caplog,
+        mock_requests_session_request,
+        mock_url_helper_time_sleep,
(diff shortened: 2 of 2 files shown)


cloudinit/sources/azure/imds.py
```diff
@@ -80,6 +81,19 @@ def exception_callback(self, req_args, exception) -> bool:
...
             log = False
 
+        # Always log if error does not match previous.
+        if exception.code is not None:
+            if self._last_error != exception.code:
+                log = True
+                self._last_error = exception.code
+        elif (
+            self._last_error is None
+            or not isinstance(self._last_error, type)
```
CalvoM, line 91: Sorry, I am trying to understand this line. Could you please explain what this checks?

cjp256: If no previous error or, or previous error is not an instance of the current one, log it/save it.

cjp256: So if say previous error was a connection timeout and the new error is 404, we'll log the 404.

cjp256: If specifically referring to `or not isinstance(self._last_error, type)`

We're checking that _last_error is a _type_ like saved below.  If it's not a type, it means the previous error was an exception with http code (int).

cjp256: Specifically it handles the case where we transition from http code error (e.g. 404) -> connection timeout or similar exception.

cjp256:
closing temporary to improve doc & test

cjp256:
Added docs to code.  Plan to follow up with a separate PR which migrates all IMDS tests to `responses` which improves the error string checks (there are a couple other IMDS-related PRs open and not worth the conflict).

cloudinit/sources/azure/imds.py
cjp256: Added code docs

tests/unittests/sources/azure/test_imds.py
```diff
@@ -481,6 +481,46 @@ def test_non_json_repsonse(
...
+            fake_http_error_for_code(500),
+            requests.ConnectionError("Fake connection error"),
+            requests.Timeout("Fake connection timeout"),
+        ]
+        attempts = len(fake_errors) + 1
+        mock_requests_session_request.side_effect = fake_errors + [
+            mock.Mock(content=json.dumps(fake_md))
+        ]
+
+        md = imds.fetch_metadata_with_api_fallback(retry_deadline=300)
```
blackboxsw, line 504: This test case doesn't exercise the new code when `log = False` we never fall into a case here where [`self._request_count < self._logging_threshold`](https://github.com/canonical/cloud-init/pull/4468/files#diff-23c0349c65e82f104b4f4df89065e282abd73e7d64a07098383d8b76c7755bc4R82) which means we log is always True as all logs are persisted.


You can see this by removing the changset from imds.py and re-running the test to success or by adding duplicate fake_errors and still seeing an error logged for every duplicate index.


  I don't think `fetch_metadata_with_api_fallback` gives us a chance to provide a ReadUrlRetryHandler with a logging_backoff of the default 1.0.  Wouldn't we need to use `fetch_reprovision_data` function instead as it sets a logging backoff which will give us a chance to set `log = False` in handling.

This test should also attempt to provide duplicated error codes or error types to ensure that you don't get duplicate logs when you are below the next logging_threshold but still have duplicated errors reported.

tests/unittests/sources/azure/test_imds.py
```diff
@@ -481,6 +481,46 @@ def test_non_json_repsonse(
...
+    def test_logs_unique_errors(
+        self,
+        caplog,
+        mock_requests_session_request,
+        mock_url_helper_time_sleep,
+    ):
+        fake_md = {"foo": {"bar": []}}
+        fake_errors = [
+            fake_http_error_for_code(404),
+            fake_http_error_for_code(410),
```
blackboxsw, line 493: Please add a duplicate fake_error here so we can validate logging is initially silenced for duplicates  resulting in only unique logs and not 1:1 for every fake_error

blackboxsw (changes requested):
minor unittests changes on this. thanks for the updated comments in code,

tests/unittests/sources/azure/test_imds.py
cjp256: done!

cjp256:
If the conversion to `responses` looks good I'll finish applying it across the test module in another PR.

cjp256:
Hrm.  Doesn't looks like responses is too thrilled on lowest_supported.  I wonder why...

cjp256:
Hopefully that settles it.

cjp256:
should be ready for review @blackboxsw , thanks!

blackboxsw (approved):
Thanks @cjp256 for the test coverage updates and test refactor.

PR #4469, opened 2023-09-28
Tidy up contributing docs

s-makin:
```
Tidy up contributing docs

In preparation for Hacktoberfest, but also because this was needed,
made a first proper go at reorganising the contributor docs. 

- Organise docs by: Overview - Code - Docs - Community
- Split out some content into separate files to make it more obvious where to find the content
- Simplify new contributor docs
- Add docs_layout to mirror the dir_layout file
- Make docs style guide separate
- Move introduction to explanation and include front page link to help direct people to it
```


doc/rtd/development/dir_layout.rst
```diff
@@ -3,7 +3,7 @@
 Directory layout
 ****************
 
-``Cloud-init``'s directory structure is somewhat different from a regular
+Cloud-init's directory structure is somewhat different from a regular
```
TheRealFalcon, line 6: We should remove the directory listing listed immediately after this line. It is woefully out of date.

doc/rtd/development/style_docs.rst
```diff
@@ -115,12 +107,16 @@ source code.
...
 ------------
 
-There are some common words that should follow specific usage:
+There are some common words that should follow specific usage in text:
 
-- ``cloud-init``: Always hyphenated. Follows sentence case, so only
-  capitalised at the start of a sentence (e.g., ``Cloud-init``).
-- ``metadata``, ``datasource``: One word.
-- ``user data``, ``vendor data``: Two words, not to be combined or hyphenated.
+- **Cloud-init**: Always hyphenated, and follows sentence case, so only
```
TheRealFalcon, line 112: I would leave this lowercase. The point of the bullet is that we generally don't want the name of the project capitalized.

doc/rtd/development/docs_layout.rst
```diff
@@ -0,0 +1,73 @@
...
+            - *links.txt*
+        - rtd_html/
+        - sources/
+
+
+``examples/``
+=============
+
+
+``man/``
```
TheRealFalcon, line 36: Contains the [Linux man pages](https://en.wikipedia.org/wiki/Man_page) for the binaries provided by cloud-init.

doc/rtd/development/docs_layout.rst
```diff
@@ -0,0 +1,73 @@
...
+  define the same URLs on every page and can use a more convenient shorthand
+  when referencing often-used links.
+
+``rtd_html/``
+=============
+
+When the documentation is built locally using ``tox -e doc``, the built pages
+can be found in this folder.
+
+``sources/``
```
TheRealFalcon, line 68: This is...uh...weird. It contains an example project of how to use some particular functionality (OVF) within cloud-init.  I'm guessing the original intention was to keep such demos here, but I didn't know this existed and not really sure it belongs here. We should probably document it as is until we move it though.

TheRealFalcon (changes requested):
I left some inline comments. Overall great content that'll be way more helpful to newcomers.

Is there a reason `development/index.html` no longer renders the CONTRIBUTING doc in Github? I think the separation causes a few issues:
1. If I'm coming into the docs via the Github CONTRIBUTING.[rst|md] (most people will), I saw the same headers/introduction and didn't immediately realize there was more content beneath the Github version. I think it's confusing having to find out where to pick up on the new page.
2. It means the two sources will likely diverge over time, and I'm not sure that's desirable.

Also, I just now realized that the section headings on the left sidebar are pages in and of themselves.
image
E.g., `Contributing overview` is a page unto itself and not just a a heading container for what's below it. That's usually not what I expect from dropdown menus and probably won't be obvious because clicking the arrows doesn't open the page. The headers in the non-development section are essentially a really simple (one or two sentence) overview with links to the rest of the pages, but this PR adds quite a lot of content to the new headers. Would it make more sense to include a separate Overview page at the top of the children, then add a link from the heading to that overview? That way, important information won't be hidden in these section headers.

doc/rtd/development/first_PR.rst
```diff
@@ -0,0 +1,171 @@
...
+
+- Open the branch on GitHub
+
+- You can see a web view of your repository and navigate to the branch at: ::
+
+      https://github.com/GH_USER/cloud-init/tree/my-topic-branch
+
+- Click :guilabel:`Pull Request`.
+
+- Fill out the pull request title, summarizing the change and a longer message
```
TheRealFalcon, line 128: Hrm...I thought I had included a comment here earlier.

Since we have a [PR template](https://github.com/canonical/cloud-init/blob/main/.github/PULL_REQUEST_TEMPLATE.md) that is automatically populated anytime somebody creates  a pull request, I'm not sure we need this template here too. The two have already diverged. We can instead say that when submitting a PR to fill out all pre-populated fields accordingly.

s-makin: Yeah I did wonder about that, but since this looked more like an example I didn't want to just remove it. Happy to take it out though for ongoing maintenance reasons.

TheRealFalcon:
I pushed a commit with my changes (not including the larger heading ones I mentioned).

@blackboxsw wanna review too? Since I already reviewed, we probably don't need a super deep review here.

doc/rtd/development/first_PR.rst
```diff
@@ -0,0 +1,81 @@
+Submit your first pull request
```
TheRealFalcon, line 1: I decided to remove quite a lot from this file and instead point the GitHub quickstart guide. Let me know if you think that was a mistake.

doc/rtd/development/dir_layout.rst
```diff
@@ -3,85 +3,62 @@
...
-            - user-data.txt.i
-      - scripts/
-         - per-boot/
-         - per-instance/
-         - per-once/
-      - seed/
-      - sem/
-
 ``/var/lib/cloud``
+==================
```
blackboxsw, line 7: In dropping the top directory tree, we now have slightly confusing rendering of the subdirectories as headings equal to the font and weight of the parent /var/lib/cloud. Let's make them subheadings so they are deemphasized.

blackboxsw: I've changed this instead to ellipsis to represent that the elements are subdirectories until /var/lib/cloud. The result looks like this:

image

blackboxsw (approved):
LGTM

PR #4470, opened 2023-09-28
azure/imds: retry on 429 errors for reprovisiondata

cjp256:
Allow retries for 429 when fetching reprovisiondata. 429 is used when exceeding rate-limits and is currently retried for fetching instance metadata.

A tight retry-loop was observed on older versions of cloud-init where we had a series of short (2 second) read timeouts.  We expect this had triggered the rate-limiting due to max concurrent connections (from IMDS' perspective). Then _poll_imds() then entered a tight retry loop, spamming the logs until connections were allowed to resume.

The standard sleep between readurl() retries should be an adequate backoff if 429 is observed.

### DIFF
--- a/cloudinit/sources/azure/imds.py
+++ b/cloudinit/sources/azure/imds.py
@@ -185,6 +185,7 @@ def fetch_reprovision_data() -> bytes:
         retry_codes=(
             404,
             410,
+            429,
         ),
         retry_deadline=None,
     )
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

CalvoM (approved):
LGTM

blackboxsw (approved):
Thanks @cjp256 for this update. As long as 1-second retry gaps between 429 retries are far enough apart to avoid triggering additional rate-limiting on Azure during reprovisiondata stage, this looks good to me. If we find we need to space out the retries we should be able to alter `sec_between` param on the `readurl` call in `fetch_reprovision_data`.

PR #4472, opened 2023-09-28
Ubuntu/devel

blackboxsw:
## do not squash merge
Queue Recommends dependency on python3-apt for apt_pkg module since cc_apt_configure uses it for some operations.
Cloud-init falls back to `apt-config dump` command when python module not present.

When this PR is merged, I will reflect it to Focal/Jammy/Lunar branches

## Proposed commit messages
```
update changelog
```
```
    d/contol: add python3-apt to Recommends
    
    cc_apt_configure will use apt_pkg from this deb if present to
    read apt configuration for Dir::Etc::sourceparts/sourcelist values.
    
    When absent, rely on the command apt-config dump.
```

## Checklist:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+cloud-init (23.3-0ubuntu2) UNRELEASED; urgency=medium
+
+  * d/control: add python3-apt as Recommends to read APT config from apt_pkg
+
+ -- Chad Smith <chad.smith@canonical.com>  Thu, 28 Sep 2023 13:57:45 -0600
+
 cloud-init (23.3-0ubuntu1) mantic; urgency=medium
 
   * d/po/templates.pot: refresh with debconf-updatepo
--- a/debian/control
+++ b/debian/control
@@ -43,7 +43,7 @@ Depends: cloud-guest-utils | cloud-utils,
          python3-serial,
          ${misc:Depends},
          ${python3:Depends}
-Recommends: eatmydata, gdisk, gnupg, software-properties-common
+Recommends: eatmydata, gdisk, gnupg, python3-apt, software-properties-common
 Suggests: openssh-server, ssh-import-id
 Description: initialization and customization tool for cloud instances
  Cloud-init is the industry standard multi-distribution method for


holmanb (commented):
@blackboxsw What value is added by this dependency? Looking at the code it looks like both the `apt_pkg` and the `apt-config dump` codepaths do the same thing - is that correct?

holmanb:
> it looks like both the apt_pkg and the apt-config dump codepaths do the same thing - is that correct?

[The PR discussion](https://github.com/canonical/cloud-init/pull/4437), commit message, and code comments don't explain why this was added either. If we can accomplish the same thing without a extra dependency, why wouldn't we just always do that? Maintaining multiple codepaths that do the same thing adds maintenance cost - what do we/users get in exchange for this cost?

TheRealFalcon:
Performance-wise, tradeoff is about 7ms on the critical-chain, though not on the "time to ssh" path.

Adding it as Recommends isn't going to bloat anything. The library already exists in our minimal images.

Is an extra 5 lines of code (plus tests) worth 7ms?

holmanb:
> Performance-wise, tradeoff is about 7ms on the critical-chain, though not on the "time to ssh" path.

Nice. Did I miss this in a comment in the PR?

> Adding it as Recommends isn't going to bloat anything. The library already exists in our minimal images.

Agreed, package bloat was not my concern for a Recommends dependency.

> Is an extra 5 lines of code (plus tests) worth 7ms?

@TheRealFalcon I'm not opposed to keeping it if there is a performance benefit. Let's just please document the tradeoffs being made in this code while the reasons are fresh.

Looking at the fallback code just makes me ask "but why?", because tradeoffs are clearly being made but not explained by the code or commit message. The intent is non-obvious, so it deserves a little context in an easy-to-find place (preferably right in the code as a comment).

I didn't know performance was the reason for this until your comment (and think I read all of the comments in the original PR), and shelling out vs a library that reads files from disk isn't an obvious performance benefit - synchronous reads from disk are required for both of these code paths.

TheRealFalcon (approved):
I'm ok with this as-is, but I agree with Brett that we should document why this is ok.

blackboxsw:
v

> I'm ok with this as-is, but I agree with Brett that we should document why this is ok.

I've added documentation about the performance tradeoff  and Recommends: state of python3-apt vs subp in PR https://github.com/canonical/cloud-init/pull/4484

PR #4478, opened 2023-10-02
Holmanb/machine readable output

holmanb:
```
      cmd: Only write json status files for boot stages
      
      Don't use status_wrapper() for `cloud-init modules --mode init`.
      
      Both status.json and result.json are currently written when `cloud-init
      modules --mode init` is ran. This is unecessary since this is not a boot
      command used by cloud-init, and as a workaround for previous bugs related to
      this behavior this leaves behind unwanted keys in status.json.
      
      This aligns cloud-init code's behavior with documented behavior.
```


TheRealFalcon (commented):
Code LGTM. Can we get a unit test ensuring the correct files are written?

holmanb:
> Code LGTM. Can we get a unit test ensuring the correct files are written?

Thanks for the review @TheRealFalcon. Done.

TheRealFalcon (approved):
LGTM! I'll leave to you to merge

PR #4485, opened 2023-10-04
ds-identify: FreeBSD fixes

igalic:
## Proposed Commit Message

```
ds-identify: Allow disable service and override environment

This PR allows disabling the ds-identify service on FreeBSD, without
having to delete the service file.
It also overrides two environment variables when running the service
to ensure that on FreeBSD, paths are where they are to be expected.

Fixes GH-4481
Co-authored-by: Dave Cottlehuber <dch@skunkwerks.at>
```

## Checklist:

### DIFF
--- a/sysvinit/freebsd/dsidentify.tmpl
+++ b/sysvinit/freebsd/dsidentify.tmpl
@@ -39,5 +39,6 @@ load_rc_config 'cloudinit'
 
 : ${cloudinit_enable="NO"}
 : ${cloudinit_dsidentify_enable="YES"}
+: ${dsidentify_env="PATH_DI_CONFIG={{ prefix }}/etc/cloud/ds-identify.cfg PATH_ETC_CLOUD={{ prefix }}/etc/cloud"}
 
 run_rc_command "$1"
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
LGTM! Does @dch want to review too?

dch:
thanks @igalic for the help on this, and the better solution than my original one.
This works in practice; I am not entirely sure whether all the vars from different
rc files should be mixed or not, it's not something I've seen in other scripts.
Viz,

```
load_rc_config 'cloudinit' <- only loads the cloudinit_* vars

: ${cloudinit_enable="NO"} <- ok
: ${cloudinit_dsidentify_enable="YES"} <- ok too
: ${dsidentify_env="PATH_DI_CONFIG={{ prefix }}/etc/cloud/ds-identify.cfg PATH_ETC_CLOUD={{ prefix }}/etc/cloud"} <- script won't work without it
```

`dsidentify_env` prefix has to match the `${name}` in the script to be loaded, as
we found. I suspect that it's not possible to override `dsidentify_env` this way,
for example. Not that *anybody* should want to do this, ofc.

Anyway, +1 for merging this as-is, the disabling functionality will work, and
the fix for previous issue is resolved.

PR #4498, opened 2023-10-07
Python 3.12 has been released. Require CI tests to pass.

holmanb:
```
Make Python 3.12 CI test non-experimental

Python 3.12 is released upstream. Require it.
```


blackboxsw (approved):
LGTM. Though one thing we may want to consider.

Can we drop "3.6" from the python-version matrix definition because we also have our `lowest-supported` toxenv defined via the include? It seems we are double-dipping on python 3.6 test runners currently.

PR #4499, opened 2023-10-09
Add logging setup fixture

holmanb:
```
tests: Add logging fix

Prior to 6bbbfbbb03, cloud-init's logger was initialized during module
load in cloudinit/log.py and cloudinit/cmd/main.py. Running individual
unittests exposes that many tests assume logging to be configured
automatically as it was before. Configure automatically in conftest.py.
```

Without this fix running various individual test files, such as test_schema.py, causes failures due to unconfigured logging:
```
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_schema_emits_warning_on_missing_jsonschema - AssertionError: assert 'Ignoring schema validation. jsonschema is not present' in ''
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_logs_deprecations[schema0-config0-Deprecated cloud-config provided:\na-b: <desc> Deprecated in version 22.1.-True] - AttributeError: 'Logger' object has no attribute 'deprecated'
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_logs_deprecations[schema1-config1-Deprecated cloud-config provided:\nx: <desc> Deprecated in version 22.1.-True] - AttributeError: 'Logger' object has no attribute 'deprecated'
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_logs_deprecations[schema2-config2-Deprecated cloud-config provided:\nx: <desc> Deprecated in version 22.1. <dep desc>-True] - AttributeError: 'Logger' object has no attribute 'deprecated'
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_logs_deprecations[schema3-config3-Deprecated cloud-config provided:\nx: <desc> Deprecated in version 22.1.-True] - AttributeError: 'Logger' object has no attribute 'deprecated'
FAILED tests/unittests/config/test_schema.py::TestValidateCloudConfigSchema::test_validateconfig_logs_deprecations[schema4-config4-Deprecated cloud-config provided:\nx: <desc> Deprecated in version 22.1.-True] - AttributeError: 'Logger' object has no attribute 'deprecated'
[...]

PR #4503, opened 2023-10-10
Set Debian's default locale to be C.UTF-8

TheRealFalcon:
## Proposed Commit Message

```
Set Debian's default locale to be C.UTF-8

When no locale is found in /etc/default/locale, cloud-init was
defaulting to generating en_US.UTF-8. This can be an expensive operation
and isn't necessary. Instead, change the default to C.UTF-8 which
generally shouldn't require generation.

Also add some logging to make it more obvious where cloud-init is
getting the locale information from.

LP: #2038945
```

## Checklist:

### DIFF
--- a/cloudinit/distros/debian.py
+++ b/cloudinit/distros/debian.py
@@ -54,14 +54,13 @@ class Distro(distros.Distro):
         # calls from repeatly happening (when they
         # should only happen say once per instance...)
         self.osfamily = "debian"
-        self.default_locale = "en_US.UTF-8"
+        self.default_locale = "C.UTF-8"
         self.system_locale = None
         self.apt = Apt.from_config(self._runner, cfg)
         self.package_managers: List[PackageManager] = [self.apt]
 
     def get_locale(self):
         """Return the default locale if set, else use default locale"""
-
         # read system locale value
         if not self.system_locale:
             self.system_locale = read_system_locale()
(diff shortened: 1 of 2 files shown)


cloudinit/distros/debian.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, name, cfg, paths):
         # calls from repeatly happening (when they
         # should only happen say once per instance...)
         self.osfamily = "debian"
-        self.default_locale = "en_US.UTF-8"
+        self.default_locale = "c.UTF-8"
```
mwhudson, line 57: I'm fairly sure it doesn't matter but can you spell this "C.UTF-8" please!

cloudinit/distros/debian.py
```diff
@@ -66,10 +66,21 @@ def get_locale(self):
...
-        )
+        if self.system_locale:
+            LOG.debug(
+                "System locale set to %s via %s",
+                self.system_locale,
+                LOCALE_CONF_FN,
+            )
+            return self.system_locale
+        else:
+            LOG.debug(
```
blackboxsw, line 77: Changeset looks good, but we have no unittest coverage for this path of the conditional

blackboxsw (approved):
Changset looks good to set locale defaults. Let's also get review from Noah on this as debian could prefer images to retain a default of `en_US.UTF-8`  if the image itself didn't provide a default in /etc/default/locale (or by proxy /etc/locale.conf).

cloudinit/distros/debian.py
blackboxsw: tested on images w/out /etc/default/locale. Success regenerating locales in absence of system defaults

TheRealFalcon:
@nmeyerhans Would you mind reviewing this?

cloudinit/distros/debian.py
TheRealFalcon: Looking at this more, I may move the logging around. In the case where no locale has been specified, `get_locale` is going to get called twice so we'll see duplicate logs

TheRealFalcon:
@blackboxsw I moved the log messages so that they don't log multiple times. I also added unit tests. One commit is converting the tests to pytest. The other commit is the actual test changes.

blackboxsw (approved):
Approved w/ unittest coverage and logging messages are only written once

nmeyerhans (approved):
👍 to changing the default locale; it is consistent with what we set in the Debian cloud images, so should be a no-op in most cases.

holmanb:
> Also, tests require a fixed version of #4499 . Without it we get no logs from caplog.

+1 Merged

PR #4506, opened 2023-10-12
subp: add a log when skipping a file for execution for lack of exe permission

ani-sinha:
Sometimes a script may itself have executable permission but the underlying
file system on which the script is located may have been mounted with 'noexec'
mount option. On those cases is_exe() would return false and the file would
be silently skipped. Add a warning log when a file is being skipped so that
it can be noticed and provides useful information while debugging issues.

### DIFF
--- a/cloudinit/subp.py
+++ b/cloudinit/subp.py
@@ -405,6 +405,13 @@ def runparts(dirp, skip_no_exist=True, exe_prefix=None):
             except ProcessExecutionError as e:
                 LOG.debug(e)
                 failed.append(exe_name)
+        else:
+            LOG.warning(
+                "skipping %s as its not executable "
+                "or the underlying file system is mounted without "
+                "executable permissions.",
+                exe_path,
+            )
 
     if failed and attempted:
         raise RuntimeError(


igalic (approved):
sounds reasonable.

holmanb (commented):
@ani-sinha Great idea, I agree this should be logged. In fact, I'm leaning towards this being logged at a `WARNING` level, since I would expect this to be unexpected/undesirable behavior that attention should be drawn towards.

What do you think?

ani-sinha:
> @ani-sinha Great idea, I agree this should be logged. In fact, I'm leaning towards this being logged at a `WARNING` level, since I would expect this to be unexpected/undesirable behavior that attention should be drawn towards.
> 
> What do you think?

Yes I was being conservative. Warning sounds better. In fact I sent this because internally someone reported that their script was not getting executed although it was present but there was nothing in the logs to indicate as to what was happening (the script failed while executing or was not being executed at all). Debugging indicated the noexec mount option issue. Warning would make this more conspicuous to users.

igalic (approved):
👍

holmanb (approved):
Looks good, thanks!

aciba90 (approved):
LGTM

PR #4515, opened 2023-10-16
Ubuntu/mantic 23.3.x

TheRealFalcon:


blackboxsw (approved):
This changeset is good to merge. But, prior to actually uploading/publishing to mantic, we'll need to also pull in Brett's changeset reverting the systemd ordering changes for `Before=systemd-user-sessions.service`  in `cloud-config.service` due to commit b3c9b6a79c85ebc8c87908383c34b0314c2205b6

TheRealFalcon:
Popped off the last commit (as we shouldn't be releasing until Brett's stuff is in) and pushed upstream

PR #4516, opened 2023-10-17
Ubuntu/focal

holmanb:
If/once approved and merged, these changes will be reflected to jammy, lunar, and mantic.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,7 @@
 cloud-init (23.3-0ubuntu0~20.04.1) UNRELEASED; urgency=medium
 
+  * d/p/do-not-block-user-login.patch:
+    - Revert behavior, allow user login after cloud-init stage (LP: #2039505)
   * d/cloud-init.maintscript: Remove the unused hook-network-manager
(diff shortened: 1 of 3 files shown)


blackboxsw (commented):
The quilt patch looks good, but there are a couple things we need to sort.
1. this changelog entry should be collapsed with the UNRELEASED changelog entry below on  ubuntu/focal
2. ubuntu/focal branch is a reflection of what was prerviously SRU'd because we uploaded from ubuntu/focal/23.3.x to represent the hotfix release. So, within ubuntu/focal we also have a supplemental commit adding python3-apt as a Recommends to the package which we'd need to account for in this SRU if we are to try uploading this content. 



I think we may want to work off of the hotfix branch for ubuntu/focal-23.3.x with your quilt patch as well as syncing that changeset back down into ubuntu/focal as well. If you intend on having the patchset here and also in ubuntu/focal-23.3.x branch I think that is fine as well, but I do think the current state of ubuntu/focal is something that is not releasable until our next official SRU of 23.4.

TheRealFalcon (approved):

PR #4519, opened 2023-10-17
fix(cc_apt_configure): avoid unneeded call to apt-install

aciba90:
## Proposed Commit Message

```
If no apt-sources config is given, then no apt dependencies are required to be installed.
This avoids unneeded calls to apt-update and apt-install.

The problem was introduced in 015543d304.
```

## Additional Context

https://github.com/canonical/cloud-init/pull/4441

## Test Steps

```
lxc launch ubuntu-daily:devel mm
lxc shell mm

# without this patch the following happens while cc_apt_configure runs, and no package is installed:
$ grep -E 'apt-update|apt-install' /var/log/cloud-init.log
2023-10-17 14:50:41,775 - util.py[DEBUG]: apt-update [eatmydata apt-get --option=Dpkg::Options::=--force-confold --option=Dpkg::options::=--force-unsafe-io --assume-yes --quiet update] took 8.957 seconds
2023-10-17 14:50:42,142 - util.py[DEBUG]: apt-install [eatmydata apt-get --option=Dpkg::Options::=--force-confold --option=Dpkg::options::=--force-unsafe-io --assume-yes --quiet install] took 0.366 seconds

# with this patch no output expected
```

## Checklist:

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -733,7 +733,8 @@ def _ensure_dependencies(cfg, aa_repo_match, cloud):
     for command in required_cmds:
         if not shutil.which(command):
             missing_packages.append(PACKAGE_DEPENDENCY_BY_COMMAND[command])
-    cloud.distro.install_packages(sorted(missing_packages))
+    if missing_packages:
+        cloud.distro.install_packages(sorted(missing_packages))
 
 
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
Fix is good. I've added supplemental unittests for various configs which can be provided to _ensure_dependencies. When this passes we can merge this and reflect to release branches

PR #4520, opened 2023-10-17
Maas disregard initramfs

holmanb:
```
Make MAAS a local data source

The existing datasource works for ephemeral deployments
(booting over the network). When adapted to run during 
local timeframe, the initramfs brings up the initial
networking.

However, when the system boots from disk initramfs doesn’t bring
up the networking, which means that cloud-init will try to get
the metadata before network is actually up.

The motivation for this change is to support network configuration
for non-Ubuntu operating systems.
```

## Additional Context
https://warthogs.atlassian.net/browse/MAASENG-2140

### DIFF
--- a/cloudinit/sources/DataSourceMAAS.py
+++ b/cloudinit/sources/DataSourceMAAS.py
@@ -12,6 +12,8 @@ import os
 import time
 
 from cloudinit import sources, url_helper, util
+from cloudinit.net.cmdline import KlibcNetworkConfigSource
+from cloudinit.sources import NetworkConfigSource
 
 LOG = logging.getLogger(__name__)
 MD_VERSION = "2012-03-01"
@@ -164,6 +166,24 @@ class DataSourceMAAS(sources.DataSource):
         return self.id_hash == get_id_from_ds_cfg(ncfg)
 
 
+class DataSourceMAASLocal(DataSourceMAAS):
+    network_config_sources = (
+        NetworkConfigSource.CMD_LINE,
+        NetworkConfigSource.SYSTEM_CFG,
+        NetworkConfigSource.DS,
+        NetworkConfigSource.INITRAMFS,
+    )
+
+    def _get_data(self):
+        if not KlibcNetworkConfigSource().is_applicable():
+            # We booted from disk. Initramfs didn't bring up a network, so
+            # nothing to do. Wait until network timeframe to run _get_data()
+            LOG.debug("No initramfs applicable config")
+            return False
+        LOG.debug("Found initramfs applicable config")
+        return super()._get_data()
+
+
 def get_oauth_helper(cfg):
     """Return an oauth helper instance for values in cfg.
 
@@ -300,6 +320,7 @@ class MAASSeedDirMalformed(Exception):
 
 # Used to match classes to dependencies
 datasources = [
+    (DataSourceMAAS, (sources.DEP_FILESYSTEM,)),
     (DataSourceMAAS, (sources.DEP_FILESYSTEM, sources.DEP_NETWORK)),
 ]
 
--- a/tests/unittests/sources/test_common.py
+++ b/tests/unittests/sources/test_common.py
@@ -44,6 +44,7 @@ DEFAULT_LOCAL = [
     Hetzner.DataSourceHetzner,
     IBMCloud.DataSourceIBMCloud,
     LXD.DataSourceLXD,
+    MAAS.DataSourceMAAS,
     NoCloud.DataSourceNoCloud,
     OpenNebula.DataSourceOpenNebula,
     Oracle.DataSourceOracle,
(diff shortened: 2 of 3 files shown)


cloudinit/sources/DataSourceMAAS.py
```diff
@@ -40,6 +43,12 @@ class DataSourceMAAS(sources.DataSource):
     dsname = "MAAS"
     id_hash = None
     _oauth_helper = None
+    network_config_sources: Tuple[NetworkConfigSource, ...] = (
```
TheRealFalcon, line 46: Did this need to be re-ordered in the non-local case? Why does initramfs need to take priority over kernel cmdline?

cloudinit/sources/DataSourceMAAS.py
```diff
@@ -164,6 +173,22 @@ def check_instance_id(self, sys_cfg):
...
+        NetworkConfigSource.SYSTEM_CFG,
+        NetworkConfigSource.DS,
+        NetworkConfigSource.INITRAMFS,
+    )
+
+    def _get_data(self):
+        if not KlibcNetworkConfigSource().is_applicable():
+            LOG.debug("No initramfs applicable config")
+            return False
+        LOG.debug("Found initramfs applicable config")
```
TheRealFalcon, line 188: So if there's no initramfs network configuration, we're not going to try to configure network at all? Why is that?

TheRealFalcon (commented):
I left some inline questions, but they're probably an artifact of me not understanding everything vs anything that needs to change. Can we provide inline comments with these changes so that future us remembers why?

cloudinit/sources/DataSourceMAAS.py
blackboxsw: @TheRealFalcon The return True/False here is just whether the DataSourceMAASLocal is being detected in init-local timeframe, not whether network is  configured by the datasource.


It means that we are going `return False` and avoid detecting DataSourceMAASLocal when there is no applicable initramfs ntework config available in init-local stage. In absence of initramfs network config, cloud-init will fallback to detecting DataSourceMAAS in init-network boot stage and rely on the typical priority of network config sources which still prioritize the typical priority order of network_config_sources.

cloudinit/sources/DataSourceMAAS.py
blackboxsw: I agree here, we don't want to change network config source ordering on the non-local DataSourceMAAS case. We would prefer to avoid that change in behavior and limit the scope of this PR to just what happens in init-local when initramfs network config is present.


From the related MAAS Jira card linked:
` In the case where the system boots from disk, initramfs doesn’t bring up the networking`.  
This means that we should only end up detecting DataSourceMAASLocal during diskless ephemeral boots on MAAS and we don't want to alter any other behavior for this datasources beyond that.

So, I suggest drop this section of the changeset.

blackboxsw (changes requested):

cloudinit/sources/DataSourceMAAS.py
holmanb: Oops, I botched merging in the changes from the patch, sorry for the confusion here.

cloudinit/sources/DataSourceMAAS.py
holmanb: I'll add context in comments inline.

holmanb:
Thanks for the reviews @TheRealFalcon and @blackboxsw. I think that I've addressed all comments on this branch.

cloudinit/sources/DataSourceMAAS.py
```diff
@@ -164,6 +166,24 @@ def check_instance_id(self, sys_cfg):
...
+class DataSourceMAASLocal(DataSourceMAAS):
+    network_config_sources = (
+        NetworkConfigSource.CMD_LINE,
+        NetworkConfigSource.SYSTEM_CFG,
+        NetworkConfigSource.DS,
+        NetworkConfigSource.INITRAMFS,
+    )
+
+    def _get_data(self):
+        if not KlibcNetworkConfigSource().is_applicable():
```
blackboxsw, line 178: While it's not high priority, it would be nice to get some unittest coverage here to assert that _get_data is triggered in DataSourceMAASLocal bcased on KlibcNetworkConfigSource.

blackboxsw (commented):
Minor request for unit test coverage. Ideally we'd start filling in the maas datasource documentation in the future, but that can be something taken to a separate PR. Will approve once we have unittest coverage.

blackboxsw:
@holmanb I just pushed a couple of commits for unittests to the branch, 1 to refactor tests to drop CiTestCase subclassing, and one for` DataSourceMAASLocal._get_data` behavior. If agreeable we can go forward with this branch as additional testing is going to be performed by MAAS team

blackboxsw (approved):

holmanb (commented):
Thanks for the test improvements @blackboxsw. They look good to me.

I just squashed my commits into one. I'll merge this once tests pass.

PR #4521, opened 2023-10-17
Ubuntu/focal 23.3.x

holmanb:
If/once approved and merged, these changes will be reflected to jammy, lunar, and mantic.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,10 +1,11 @@
-cloud-init (23.3.2-0ubuntu0~20.04.1) UNRELEASED; urgency=medium
-
+cloud-init (23.3.2-0ubuntu1~20.04.1) UNRELEASED; urgency=medium
+  * d/p/do-not-block-user-login.patch:
+    - Revert behavior, allow user login after cloud-init stage (LP: #2039505)
   * Upstream snapshot based on 23.3.2. (LP: #2039453).
     List of changes from upstream can be found at
     https://raw.githubusercontent.com/canonical/cloud-init/23.3.2/ChangeLog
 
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,10 +1,11 @@
-cloud-init (23.3.2-0ubuntu0~20.04.1) UNRELEASED; urgency=medium
-
+cloud-init (23.3.2-0ubuntu1~20.04.1) UNRELEASED; urgency=medium
```
TheRealFalcon, line 1: Just FYI, we increment the .1 at the end rather than 0ubuntu0 here. Doesn't need to change here though as this version will be overwritten with 23.3.3 before we upload.

TheRealFalcon (approved):
LGTM! Feel free to push upstream.

debian/changelog
holmanb: > we increment the .1 at the end rather than 0ubuntu0 here

Since this is adding a downstream patch, I think ubuntu1 makes more sense, right? The number after "ubuntu" is typically used to indicate patch revision on top of debian release. Since there is no debian release between us and upstream, (and this is literally adding a downstream patch) I would interpret this to mean patch revision on top of upstream release.

TheRealFalcon: SRU numbers are different. We should only increment the Y in `ubuntuY` during devel.
https://github.com/canonical/ubuntu-maintainers-handbook/blob/main/VersionStrings.md#version-adding-a-change-in-ubuntu-as-a-stable-release-update

holmanb: > SRU numbers are different

Ahhhh, I didn't realize our numbers at the end were an (SRU) exception to the normal package version schema. +1 Thanks, @TheRealFalcon.

PR #4533, opened 2023-10-19
cc_ssh_import_id: Extend support to FreeBSD

igalic:
## Proposed Commit Message

```
cc_ssh_import_id: Extend support to FreeBSD

The module was already enabled in cloud.cfg, but it doesn't officially
state that FreeBSD is supported.
Any platform that has a ssh-import-id package is supported, so add the
dependency and declare FreeBSD supported.

```

## Checklist


igalic:
I should probably add dragonfly to the list, and add a test for this

PR #4543, opened 2023-10-24
Clarify boothook docs

holmanb:
```
Clarify boothook docs
```

## Additional Context
See commit messages for explanations

### DIFF
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -107,7 +107,8 @@ mounted, including ones that have stale (previous instance) references in
 :file:`/etc/fstab`. As such, entries in :file:`/etc/fstab` other than those
 necessary for cloud-init to run should not be done until after this stage.
 
-A part-handler and boothooks will run at this stage.
+A part-handler and :ref:`boothooks<explanation/format:\`\`cloud-boothook\`\`>`
+will run at this stage.
 
 .. _boot-Config:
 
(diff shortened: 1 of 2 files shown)


s-makin (approved):
I think this makes sense. Thanks for taking the lead on this - I agree it's better not to present these terms without them being explained anywhere, and if the user doesn't need to know about boothooks at all then this definition/term should be avoided except in the dev docs.

aciba90 (approved):
LGTM, thanks for cleaning this up!

doc/rtd/explanation/boot.rst
```diff
@@ -107,9 +107,7 @@ mounted, including ones that have stale (previous instance) references in
 :file:`/etc/fstab`. As such, entries in :file:`/etc/fstab` other than those
 necessary for cloud-init to run should not be done until after this stage.
 
-A part-handler will run at this stage, as will boothooks including
-cloud-config ``bootcmd``. The user of this functionality has to be aware
-that the system is in the process of booting when their code runs.
+A part-handler and boothooks will run at this stage.
```
blackboxsw, line 110: Let's cross link using the autosectiondocs.
```suggestion
A part-handler and :ref:`boothooks<explanation/format:\`\`cloud-boothook\`\`>`
+will run at this stage.
```

blackboxsw (changes requested):
MOAR linking FTW. Approved after this changeset.

blackboxsw (approved):
There was a stray `+` as part of my suggested fix. force-pushed that and +1

PR #4548, opened 2023-10-24
Ubuntu/mantic 23.3.x

TheRealFalcon:
`new_upstream_snapshot.py -c 23.3.3 -b 2040291`

I popped off the quilt refresh as it isn't necessary and adds noise to the hotfix

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+cloud-init (23.3.3-0ubuntu0~23.10.1) UNRELEASED; urgency=medium
+
+  * Upstream snapshot based on 23.3.3. (LP: #2040291).
+    List of changes from upstream can be found at
+    https://raw.githubusercontent.com/canonical/cloud-init/23.3.3/ChangeLog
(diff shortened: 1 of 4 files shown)


holmanb (approved):

debian/changelog
```diff
@@ -1,3 +1,11 @@
+cloud-init (23.3.3-0ubuntu0~23.10.1) mantic; urgency=medium
+
+  * Upstream snapshot based on 23.3.3. (LP: #2040291).
```
blackboxsw, line 3: Unrelated, we also need to sync 23.3.x ChangeLog into main as we haven't pulled down are published changes yet into upstream/main

blackboxsw (approved):

debian/changelog
TheRealFalcon: yeah, I synced 23.3.2 when we did that one. I still need 23.3.3, but figured we should get the downstreams done first

PR #4550, opened 2023-10-24
Ubuntu/jammy 23.3.x

TheRealFalcon:
new_upstream_snapshot.py -c 23.3.3 -b 2040291

I popped off the quilt refresh as it isn't necessary and adds noise to the hotfix


blackboxsw (approved):
LGTM

Issue #4525, opened 2023-10-18
ImportError: No distribution found for distro suse


smoshiur1237:
# Bug report
I was trying to install cloud-init from source github repo in SUSE Enterprise 15 SP4. I am getting the ImprotError and also some warnings which eventually failed the process.

## Steps to reproduce the problem
The following command were used to install cloud-init:
```
git clone https://github.com/canonical/cloud-init.git
cd cloud-init
sudo python3 -m pip install -r requirements.txt
sudo python3 setup.py build
sudo python3 setup.py install --init-system systemd
sudo cloud-init init --local   # Failure occurs on this step
sudo cloud-init status
```
We can see the ImportError of suse during `init --local `stage. There was no `suse` distro module in the source code(sles, opensuse and other opensuse, sles option is available) and I think that is the issue of the failure. 

## Environment details
- Cloud-init version: v. 23.3.1
- Operating System Distribution: SUSE Linux Enterprise Server 15 SP4

## cloud-init logs

We are having the following error during installation and setup of cloud-init:
```
Finished processing dependencies for cloud-init==23.3+80.g0cf010e6
 Cloud-init v. 23.3.1 running 'init-local' at Thu, 12 Oct 2023 06:25:28 +0000. Up 601755.92 seconds.
  - util.py[WARNING]: failed stage init-local
 failed run of stage init-local
 ------------------------------------------------------------
 Traceback (most recent call last):
 File "/usr/lib/python3.6/site-packages/cloud_init-23.3+80.g0cf010e6-py3.6.egg/cloudinit/cmd/main.py", line 772, in status_wrapper
 ret = functor(name, args)
 File "/usr/lib/python3.6/site-packages/cloud_init-23.3+80.g0cf010e6-py3.6.egg/cloudinit/cmd/main.py", line 386, in main_init
 init.fetch(existing=existing)
 File "/usr/lib/python3.6/site-packages/cloud_init-23.3+80.g0cf010e6-py3.6.egg/cloudinit/stages.py", line 469, in fetch
 return self._get_data_source(existing=existing)
 File "/usr/lib/python3.6/site-packages/cloud_init-23.3+80.g0cf010e6-py3.6.egg/cloudinit/stages.py", line 362, in _get_data_source
 self.distro,
 File "/usr/lib/python3.6/site-packages/cloud_init-23.3+80.g0cf010e6-py3.6.egg/cloudinit/stages.py", line 155, in distro
 distro_cls = distros.fetch(distro_name)
[...]

PR #4557, opened 2023-10-25
Add suse distro

holmanb:
Proposed commit message:
```
distros: Add suse
```

## Additional Context
Fixes https://github.com/canonical/cloud-init/issues/4525

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -87,6 +87,7 @@ OSFAMILIES = {
         "sle_hpc",
         "sle-micro",
         "sles",
+        "suse",
     ],
     "openeuler": ["openeuler"],
     "OpenCloudOS": ["OpenCloudOS", "TencentOS"],
--- a/cloudinit/distros/suse.py
+++ b/cloudinit/distros/suse.py
@@ -1,0 +1,5 @@
+from cloudinit.distros import opensuse
+
+
+class Distro(opensuse.Distro):
+    pass


holmanb:
Requesting review from @Rozzii, @smoshiur1237, and @rjschwei

Rozzii:
I am not sure about the commit name, I mean I am honored that my "nick" would be in the cloud-init git history but is it okay thing in this project ? I am not familiar with the conventions here.

holmanb:
> I am not sure about the commit name, I mean I am honored that my "handle" would be in the cloud-init git history but is it okay thing in this project ? I am not familiar with the conventions here.

Agreed :) In this project by default we typically squash merge with a commit message that is defined in the Pull request. The message I'm proposing is:
```
distro: Add suse
```

CalvoM (approved):
LGTM

rjschwei:
As stated in the issue I think this is incorrect, as no SUSE distro, no openSUSE or SUSE Linux enterprise server identifies itself as "SUSE". The changes do no harm but are superfluous.

smoshiur1237 (commented):
Thanks for adding the changes. I was able to go forward with my test where distro name is detected as suse for SUSE SP4 Enterprise version.
LGTM

holmanb:
Hey @smoshiur1237, where did you get this image? I'm confused why this is happening if @rjschwei doesn't expect it to. It might "solve your issue", but I'm concerned that we're barking up the wrong tree if the root cause is using a customized change. Do you have a reproducer that I can poke at? Is this a custom image build you are using or something direct from SUSE?

smoshiur1237:
@holmanb The image I am using is named `SLES15-SP4-Minimal-VM.x86_64-OpenStack-Cloud-GM.qcow2` , and it is available in [Suse Enterprise image](https://www.suse.com/download/sles/). After installing cloud-init I was getting the error for suse distro and your changes help to eliminate the error and it is moving on.

Rozzii:
It is not stale, we are waiting for feedback, @holmanb .

holmanb:
> It is not stale, we are waiting for feedback, @holmanb .

I think at this point this just requires a review from another dev (ping @aciba90, @TheRealFalcon @blackboxsw).

@rjschwei doesn't seem to think it is necessary, but apparently this change solves @Rozzii and @smoshiur1237's issue on a suse-provided image, so I'm +1 with including this since it is harmless if unused. If @rjschwei has any thoughts on why that's happening on the image @smoshiur1237 pointed to, I'd be curious, otherwise I'm fine to land this PR and put this issue  behind us.

TheRealFalcon (approved):
LGTM

rjschwei:
OK, and to close the loop on this, and sorry it took me a while to get there. When cloud-init is built in the build service we build it for a specific distribution by including a so called release package in the build system. This release package then assures that the system identifies itself "correctly" for example as SLES. I suspect, but have not checked, that the qcow image being used as the basis to build cloud-init does not have a release package installed, which is why the system would identify itself as "suse".

PR #4558, opened 2023-10-26
azure: report failure to host if ephemeral DHCP was on a secondary NIC

cjp256:
Azure requires the use of the primary interface to query IMDS and report ready to wireserver.  If we determine that the interface is not the primary interface, report the failure to the host.

The exception is when iterating over hotplugged NICs in the Savable PPS scenario.  In this case the secondary may be enumerated first and we would expect a failure.


anhvoms (approved):

anhvoms:
LGTM

blackboxsw (approved):
LGTM!

PR #4559, opened 2023-10-27
fix(cli): fix parsing of argparse subcommands

CalvoM:
## Proposed Commit Message

We are currently not parsing the subcommands correctly. 
This causes the high-level arguments to make parsing of subcommand be erroneous.
Fixes
LP: #2040325

## Test Steps

The steps listed in [this LP bug](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2040325) are sufficient and well explained.

## Checklist

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -863,14 +863,6 @@ def main(sysv_args=None):
         help="Show program's version number and exit.",
     )
     parser.add_argument(
-        "--file",
-        "-f",
-        action="append",
-        dest="files",
-        help="Use additional yaml configuration files.",
-        type=argparse.FileType("rb"),
-    )
-    parser.add_argument(
         "--debug",
         "-d",
         action="store_true",
@@ -903,6 +895,14 @@ def main(sysv_args=None):
         help="Start in local mode (default: %(default)s).",
         default=False,
     )
+    parser_init.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
+    )
     # This is used so that we can know which action is selected +
     # the functor to use to run this subcommand
     parser_init.set_defaults(action=("init", main_init))
@@ -918,6 +918,14 @@ def main(sysv_args=None):
         help="Module configuration name to use (default: %(default)s).",
         default="config",
         choices=("init", "config", "final"),
+    )
+    parser_mod.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
     )
     parser_mod.set_defaults(action=("modules", main_modules))
 
@@ -950,6 +958,14 @@ def main(sysv_args=None):
         nargs="*",
         metavar="argument",
         help="Any additional arguments to pass to this module.",
+    )
+    parser_single.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
     )
     parser_single.set_defaults(action=("single", main_single))
 
@@ -989,7 +1005,10 @@ def main(sysv_args=None):
 
     if sysv_args:
         # Only load subparsers if subcommand is specified to avoid load cost
-        subcommand = sysv_args[0]
+        subcommand = next(
+            (posarg for posarg in sysv_args if not posarg.startswith("-")),
+            None,
+        )
         if subcommand == "analyze":
             from cloudinit.analyze import get_parser as analyze_parser
 
--- a/doc/rtd/reference/cli.rst
+++ b/doc/rtd/reference/cli.rst
@@ -15,14 +15,12 @@ Example output:
 
 .. code-block::
 
-   usage: cloud-init [-h] [--version] [--file FILES] [--debug] [--force]
+   usage: cloud-init [-h] [--version] [--debug] [--force]
                                                                {init,modules,single,query,features,analyze,devel,collect-logs,clean,status,schema} ...
 
     options:
       -h, --help            show this help message and exit
       --version, -v         Show program's version number and exit.
-      --file FILES, -f FILES
-                            Use additional yaml configuration files.
       --debug, -d           Show additional pre-action logging (default: False).
       --force               Force running even if no datasource is found (use at your own risk).
 
@@ -173,6 +171,7 @@ due to semaphores in :file:`/var/lib/cloud/instance/sem/` and
 :file:`/var/lib/cloud/sem`.
 
 * :command:`--local`: Run *init-local* stage instead of *init*.
+* :command:`--file` : Use additional yaml configuration files.
 
 .. _cli_modules:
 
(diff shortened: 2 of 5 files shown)


cloudinit/cmd/main.py
```diff
@@ -987,7 +987,11 @@ def main(sysv_args=None):
 
     if sysv_args:
         # Only load subparsers if subcommand is specified to avoid load cost
-        subcommand = sysv_args[0]
+        subcommand = list(
+            filter(lambda arg: not arg.startswith("-"), sysv_args)
+        )
+        if subcommand:
+            subcommand = subcommand[0]
```
blackboxsw, line 994: Let's just use next here as it'll be faster than the filter + lambda + list cast.
```suggestion
        subcommand = next(
            (posarg for posarg in sysv_args if not posarg.startswith("-")),
            None
        )
```

blackboxsw (changes requested):
Let's use the next iterator as it should execute faster and we don't really need filter/list to generate and cast a full list of params, just the first non-optional/long parameter.


Also please add a unit test to tests/unittests/cmd/test_main.py that covers specifying --debug but no subcommand to exercise this behavior in case we change this in the future.

cloudinit/cmd/main.py
CalvoM: Oh, this is a better approach.

blackboxsw:
Thank you @CalvoM. It just needs unittest coverage and we are good to go.

cloudinit/cmd/main.py
```diff
@@ -987,7 +987,10 @@ def main(sysv_args=None):
 
     if sysv_args:
         # Only load subparsers if subcommand is specified to avoid load cost
-        subcommand = sysv_args[0]
+        subcommand = next(
```
blackboxsw, line 990: One problem with this solution is that the top-level command supports a `--file` parameter like the following `cloud-init --debug --file /my/path.yaml` which will cause us to claim the subcommand  as `/my/path.yaml`.

I think this top-level --file key is really only appropriate for the following subcommands (init, modules and single).

As such, I think it should only be declared on the specific subcommand parser.
Then our top-level --help will be better able to all subcommands and the `next` handling will hold for simple -- optional parameters.

Here's a diff that pushes the applicable --files optional param down into the 3 subcommands which allow it
```diff
diff --git a/cloudinit/cmd/main.py b/cloudinit/cmd/main.py
index d4bf38243..5bce6917b 100755
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -860,14 +860,6 @@ def main(sysv_args=None):
         version="%(prog)s " + (version.version_string()),
         help="Show program's version number and exit.",
     )
-    parser.add_argument(
-        "--file",
-        "-f",
-        action="append",
-        dest="files",
-        help="Use additional yaml configuration files.",
-        type=argparse.FileType("rb"),
-    )
     parser.add_argument(
         "--debug",
         "-d",
@@ -901,6 +893,14 @@ def main(sysv_args=None):
         help="Start in local mode (default: %(default)s).",
         default=False,
     )
+    parser_init.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
+    )
     # This is used so that we can know which action is selected +
     # the functor to use to run this subcommand
     parser_init.set_defaults(action=("init", main_init))
@@ -917,6 +917,14 @@ def main(sysv_args=None):
         default="config",
         choices=("init", "config", "final"),
     )
+    parser_mod.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
+    )
     parser_mod.set_defaults(action=("modules", main_modules))
 
     # This subcommand allows you to run a single module
@@ -949,6 +957,14 @@ def main(sysv_args=None):
         metavar="argument",
         help="Any additional arguments to pass to this module.",
     )
+    parser_single.add_argument(
+        "--file",
+        "-f",
+        action="append",
+        dest="files",
+        help="Use additional yaml configuration files.",
+        type=argparse.FileType("rb"),
+    )
     parser_single.set_defaults(action=("single", main_single))
 
     parser_query = subparsers.add_parser(
```


Longer term than this PR, I think we need to better handle invalid optional params in subcommands.

When subcommands are provided

blackboxsw: filed https://github.com/canonical/cloud-init/issues/4609 to track the above issue with CLI separately

blackboxsw (approved):
Thanks @CalvoM I confirm this addresses the traceback mentioned in the linked bug. We probably have a couple of other work items separately to handle such as #4609 and or potentially supported a `--debug` param for the clean subcommand to print each operation it is performing to educate users about the cleanup being performed, but I think those are lower-weight bugs to address separately.

doc/rtd/reference/cli.rst
```diff
@@ -173,6 +171,7 @@ due to semaphores in :file:`/var/lib/cloud/instance/sem/` and
 :file:`/var/lib/cloud/sem`.
 
 * :command:`--local`: Run *init-local* stage instead of *init*.
+* :command:`--file` : Use additional yaml configuration files.
```
blackboxsw, line 174: Thx for these additions.

embik:
This change was just rolled out to Ubuntu 22.04 on Thursday as far as we can say and it seems to have broken calls to `cloud-init` in the format `cloud-init -f <file> init`. I understand that fixing parsing args was something that needed to be done, but I'm concerned this is not marked as breaking change and just got rolled out into a 2 years old LTS release.

xrstf:
Even worse, there is no syntax that is compatible with both cloud-init versions. `cloud-init init -f foo.cfg` only works in the new version, `clout-init -f foo.cfg init` only works in the old version.

holmanb:
Thanks for reporting! 

> but I'm concerned this is not marked as breaking change and just got rolled out into a 2 years old LTS release.

Oof, not good. I just filed this as a bug in the downstream [bug tracker](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2064300) against the releases which received this change. Feel free to track the issue and add additional details there.

@embik @xrstf Can I ask about your use cloud-init case? Why you are running this cloud-init command manually rather than allowing the init system to run cloud-init on boot?

xrstf:
@holmanb We're using it as part of the bootstrapping for Kubernetes nodes in https://github.com/kubermatic/operating-system-manager. I'm not super familiar with the details, I just had to handle the fallout today 😁

holmanb:
> As @xrstf said, we run cloud-init to bootstrap Kubernetes nodes. Our cloud-init configs grew quite big and we ran into user-data limitations on some cloud providers. So we bootstrap nodes with a lightweight cloud-init config which pulls a larger cloud-init config from a remote location and then executes cloud-init against that downloaded larger config.

Interesting, thanks for sharing @embik. Have you tried using [gzipped configurations](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#gzip-compressed-content)? Alternatively, [#include](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#include-file) configurations seem purpose built for doing what you are attempting to do manually. Is there some reason that one of these configuration formats don't work for you?

embik:
> Interesting, thanks for sharing @embik. Have you tried using [gzipped configurations](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#gzip-compressed-content)? Alternatively, [#include](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#include-file) configurations seem purpose built for doing what you are attempting to do manually. Is there some reason that one of these configuration formats don't work for you?

@holmanb I think our challenge is that we pull the Kubernetes `Secret` object and then need to extract the cloud-init from its data. So pointing directly to the URL would result in a response format that cloud-init doesn't understand. But I will double-check on this.

I believe we weren't aware that this usage of `cloud-init` was discouraged / not supported, so we'll need to investigate solutions, especially if it is going to deprecated.

PR #4560, opened 2023-10-27
azure: workaround to disable reporting IMDS failures on Azure Stack

cjp256:
There are Azure Stack implementations which do not support IMDS.  The only way to detect this is by checking for the presence of the static route to IMDS.  If we don't see the route to IMDS, limit the failure report to the host via KVP.

There will be future work to detect that we are running on Azure Stack and have explicit configuration passed along via ovf-env.xml to toggle IMDS support as the environment dictates.

- refactor _check_if_primary() to track configured routes

- use this to toggle reporting behavior

- add duration check to TestGetMetadataFromImds tests in case we want to toggle duration in the future.

Example diagnostic on Azure public cloud:

```
2023-10-27 12:37:04,634 - azure.py[DEBUG]: Obtained DHCP lease on interface 'eth0' (primary=True driver='hv_netvsc' router='10.0.0.1' routes=[('0.0.0.0/0', '10.0.0.1'), ('168.63.129.16/32', '10.0.0.1'), ('169.254.169.254/32', '10.0.0.1')] lease={'interface': 'eth0', 'fixed-address': '10.0.0.4', 'server-name': 'IAD011091108004SOC', 'subnet-mask': '255.255.255.0', 'dhcp-lease-time': '4294967295', 'routers': '10.0.0.1', 'dhcp-message-type': '5', 'domain-name-servers': '168.63.129.16', 'dhcp-server-identifier': '168.63.129.16', 'dhcp-renewal-time': '4294967295', 'rfc3442-classless-static-routes': '0,10,0,0,1,32,168,63,129,16,10,0,0,1,32,169,254,169,254,10,0,0,1', 'dhcp-rebinding-time': '4294967295', 'unknown-245': 'a8:3f:81:10', 'domain-name': 'uejkdvkrjiqe1jxrijlvafqihe.bx.internal.cloudapp.net', 'renew': '1 2159/12/03 19:05:19', 'rebind': '1 2159/12/03 19:05:19', 'expire': '1 2159/12/03 19:05:19'} imds_routed=True wireserver_routed=True)
```

Example diagnostic on Azure Stack:

```
2023-10-27 12:35:47,363 - azure.py[DEBUG]: Obtained DHCP lease on interface 'eth0' (primary=True driver='hv_netvsc' router='10.126.64.1' routes=[('0.0.0.0/0', '10.126.64.1'), ('168.63.129.16/32', '10.126.64.1')] lease={'interface': 'eth0', 'fixed-address': '10.126.64.35', 'subnet-mask': '255.255.252.0', 'routers': '10.126.64.1', 'dhcp-lease-time': '4294967295', 'dhcp-message-type': '5', 'domain-name-servers': '10.50.10.50,10.50.50.50', 'dhcp-server-identifier': '168.63.129.16', 'interface-mtu': '1500', 'dhcp-renewal-time': '4294967295', 'unknown-245': 'a8:3f:81:10', 'rfc3442-classless-static-routes': '0,10,126,64,1,32,168,63,129,16,10,126,64,1', 'dhcp-rebinding-time': '4294967295', 'domain-name': 'corp.microsoft.com', 'renew': '1 2159/12/03 19:04:02', 'rebind': '1 2159/12/03 19:04:02', 'expire': '1 2159/12/03 19:04:02'} imds_routed=False wireserver_routed=True)
```

We can see that IMDS routing is detected appropriately.

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -17,6 +17,8 @@ from pathlib import Path
 from time import sleep, time
 from typing import Any, Dict, List, Optional
 
+import requests
+
 from cloudinit import net, sources, ssh_util, subp, util
 from cloudinit.event import EventScope, EventType
 from cloudinit.net import device_driver
@@ -334,6 +336,8 @@ class DataSourceAzure(sources.DataSource):
         self._iso_dev = None
         self._network_config = None
         self._ephemeral_dhcp_ctx = None
+        self._route_configured_for_imds = False
+        self._route_configured_for_wireserver = False
         self._wireserver_endpoint = DEFAULT_WIRESERVER_ENDPOINT
         self._reported_ready_marker_file = os.path.join(
             paths.cloud_dir, "data", "reported_ready"
@@ -344,6 +348,8 @@ class DataSourceAzure(sources.DataSource):
 
         self._ephemeral_dhcp_ctx = None
         self._iso_dev = None
+        self._route_configured_for_imds = False
+        self._route_configured_for_wireserver = False
         self._wireserver_endpoint = DEFAULT_WIRESERVER_ENDPOINT
         self._reported_ready_marker_file = os.path.join(
             self.paths.cloud_dir, "data", "reported_ready"
@@ -371,17 +377,21 @@ class DataSourceAzure(sources.DataSource):
             # Primary nics must contain routes.
             return False
 
-        routed_networks = [r[0] for r in ephipv4.static_routes]
-        wireserver_route = f"{self._wireserver_endpoint}/32"
-        primary = any(
-            n in routed_networks
-            for n in [
-                "169.254.169.254/32",
-                wireserver_route,
-            ]
+        routed_networks = [r[0].split("/")[0] for r in ephipv4.static_routes]
+
+        # Expected to be true for all of Azure public cloud and future Azure
+        # Stack versions with IMDS capabilities, but false for existing ones.
+        self._route_configured_for_imds = "169.254.169.254" in routed_networks
+
+        # Expected to be true for Azure public cloud and Azure Stack.
+        self._route_configured_for_wireserver = (
+            self._wireserver_endpoint in routed_networks
         )
 
-        return primary
+        return (
+            self._route_configured_for_imds
+            or self._route_configured_for_wireserver
+        )
 
     @azure_ds_telemetry_reporter
     def _setup_ephemeral_networking(
@@ -503,7 +513,8 @@ class DataSourceAzure(sources.DataSource):
             primary = self._check_if_primary(ephipv4)
             report_diagnostic_event(
                 "Obtained DHCP lease on interface %r "
-                "(primary=%r driver=%r router=%r routes=%r lease=%r)"
+                "(primary=%r driver=%r router=%r routes=%r lease=%r "
+                "imds_routed=%r wireserver_routed=%r)"
                 % (
                     iface,
                     primary,
@@ -511,6 +522,8 @@ class DataSourceAzure(sources.DataSource):
                     ephipv4.router,
                     ephipv4.static_routes,
                     lease,
+                    self._route_configured_for_imds,
+                    self._route_configured_for_wireserver,
                 ),
                 logger_func=LOG.debug,
             )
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM

anhvoms (approved):

cloudinit/sources/DataSourceAzure.py
```diff
@@ -371,17 +377,15 @@ def _check_if_primary(self, ephipv4: EphemeralIPv4Network) -> bool:
...
-            n in routed_networks
-            for n in [
-                "169.254.169.254/32",
-                wireserver_route,
-            ]
+        routed_networks = [r[0].split("/")[0] for r in ephipv4.static_routes]
+        self._route_configured_for_imds = "169.254.169.254" in routed_networks
+        self._route_configured_for_wireserver = (
+            self._wireserver_endpoint in routed_networks
+        )
```
blackboxsw, line 384: Let's add a leading comment for future-us above each attribute here to represent the config we expect from the separate Azure product behaviors.
```suggestion
        # Expected for primary NIC on Azure Public cloud offerings
        self._route_configured_for_imds = "169.254.169.254" in routed_networks
        # Expected support for primary NIC on Azure Stack
        self._route_configured_for_wireserver = (
            self._wireserver_endpoint in routed_networks
        )
```

blackboxsw (changes requested):
Thanks @cjp256, content looks good.

One question: 
  - What problem/symptom is the ultimately avoiding on Azure-side for Azure Stack. Does limiting error reporting to host only via KVM avoid a bunch of additional failure logs in cloud-init.log such as ["Failed to report failure using cached ephemeral dhcp context"](https://github.com/canonical/cloud-init/blob/4c85405809c56cde076c2c9e1e3f085aec601afc/cloudinit/sources/DataSourceAzure.py#L1288-L1291)?

Thanks for clarification, otherwise the approach seems reasonable. Minor ask for comments to clarify expected config for AzureStack vs public cloud

cloudinit/sources/DataSourceAzure.py
cjp256: FWIW _route_configured_for_wireserver = True is expected for all scenarios.  I'm largely tracking it for logging purposes in a manner consistent with imds route.  It would be helpful should we ever see a case where it is reported as false for unknown reasons (perhaps we DHCP'd on secondary interface or somehow DHCP response was incorrect).

I will add documentation to clarify, thanks!

cjp256:
> Thanks @cjp256, content looks good.
> 
> One question:
> 
> * What problem/symptom is the ultimately avoiding on Azure-side for Azure Stack. Does limiting error reporting to host only via KVM avoid a bunch of additional failure logs in cloud-init.log such as ["Failed to report failure using cached ephemeral dhcp context"](https://github.com/canonical/cloud-init/blob/4c85405809c56cde076c2c9e1e3f085aec601afc/cloudinit/sources/DataSourceAzure.py#L1288-L1291)?
> 
> Thanks for clarification, otherwise the approach seems reasonable. Minor ask for comments to clarify expected config for AzureStack vs public cloud

The main thing we're accomplishing here is that we ensure the Azure Stack 

> Thanks @cjp256, content looks good.
> 
> One question:
> 
> * What problem/symptom is the ultimately avoiding on Azure-side for Azure Stack. Does limiting error reporting to host only via KVM avoid a bunch of additional failure logs in cloud-init.log such as ["Failed to report failure using cached ephemeral dhcp context"](https://github.com/canonical/cloud-init/blob/4c85405809c56cde076c2c9e1e3f085aec601afc/cloudinit/sources/DataSourceAzure.py#L1288-L1291)?
> 

Today reporting failures really only count if we report them to the wireserver.  The host report allows us to report the failure using the KVP channel which is not considered for the provisioning protocol today, but helps us to readily identify failures, particularly when wireserver is unavailable (DHCP or wireserver failures, etc.). It still gets reported as a LOG.error().

The plan is to supplement the ovf-env.xml configuration with toggles to indicate whether or not IMDS is available, so we can skip checking for IMDS altogether in these cases.  In the meantime, I'd rather it attempts anyways and emits an error when IMDS is not present [on Azure Stack] rather than risk regression on Azure cloud (even if we expect it to hold true).  Once Azure Stack implements the property and we know for certain we're running in that environment, we can get the best for both worlds and remove the IMDS route check.

cloudinit/sources/DataSourceAzure.py
cjp256: done!

cjp256:
I updated the commit message to provide a little more detail and added the requested comments, should be good to go @blackboxsw, thanks!

blackboxsw (approved):
Thanks for correcting my understanding here.

PR #4561, opened 2023-10-27
Add summit digest report to docs

s-makin:
```
Summit summary added to docs for posterity

Mailing list entry has gone out but no link for it yet
Also moved images out of the repo (including the ones from previous
years) and added to the Ubuntu asset manager for hosting to avoid
bloating the repo as time goes on!
```


blackboxsw (approved):
Force pushed an update to the correct mailinglist URL. Thanks @s-makin

PR #4562, opened 2023-10-27
feat: Add top level properties and additionalProperties: false

TheRealFalcon:
Current set of commits is kind of mess, but given that it has already been partially reviewed, I'm going to leave as-is for further review and then squash into a few more targeted commits.

## Proposed Commit Message

```
feat: Add top level properties and additionalProperties: false

The top-level properties is necessary to restrict additional top-level keys.
```

## Additional Context
We're duplicating keys, but that's better than having to rehaul how we generate our documentation.

`doc/examples/cloud-config.txt` contained keys that don't work in userdata and/or belong in /etc/cloud/cloud.cfg. Because of this, I created a separate `doc/examples/etc_cloud_cfg.txt` with the output from /etc/cloud/cloud.cfg plus these extra keys. This file doesn't get validated by our schema tests.


blackboxsw (commented):
The premise of this branch looks good, and it looks like it also raises a bug in how we were previously validating cloud-int user-data in [early boot](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/main.py#L480-L482) because we were passing in the merged config of both systemcfg and user-data. Now we get to see warnings in cloud-init.log about invalid schema due to system config keys present in /etc/cloud/cloud.cfg.d that are not strictly cloud-config user-data.

`schema.py[WARNING]: Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.`

When adding the following diff to the branch we can see the specific merged system config keys from /tec/cloud/cloud.cfg.d reported:
```diff
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -481,7 +481,7 @@ def main_init(name, args):
         validate_cloudconfig_schema(
             config=init.cfg,
             strict=False,
-            log_details=False,
+            log_details=True,
             log_deprecations=True,
         )
     else:
```
New log 
`/var/log/cloud-init.log-: Additional properties are not allowed ('_log', 'vendor_data2', 'syslog_fix_perms', 'datasource_list', 'def_log_file', 'log_cfgs' were unexpected)`

Short term, now that our schema is strict, it may be best to strictly perform a safeyaml.load of the processed user-data file from init.paths.get_ipath_cur("cloud_config") in cloudinit.cmd.main_init instead of passing in full merged Init.cfg to validate_cloudconfig_schema.

Long term, we may want to provide a separate JSON schema definition for system config keys which would provide schema coverage for these system config keys as they aren't applicable as user-data.

Additionally I think we do need to represent a missing "output" property in our user-data schema as it does provide an option to reset output logs to somewhere other than the image-based default in system config or cloudinit.settings just after user-data is processed during boot.


Sample start on output schema.
```diff
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -3375,6 +3375,61 @@
         }
       }
     },
+    "output_log_operation": {
+      "oneOf": [
+        {
+          "type": "string",
+          "description": "A filepath operation configuration. This is a string containing a filepath and an optional leading operator: '>', '>>' or '|'. Operators '>' and '>>' indicate whether to overwrite or append to the file. The operator '|' pipes content to the command arguments specified."
+        },
+        {
+          "type": "array",
+          "description": "A list specifying filepath operation configuration for stdout and stderror",
+          "items": {
+            "type": [
+              "string"
+            ]
+          },
+          "minItems": 2,
+          "maxItems": 2
+        },
+        {
+          "type": "object",
+          "properties": {
+            "output": {
+              "type": "string",
+              "description": "A filepath operation configuration. This is a string containing a filepath and an optional leading operator: '>', '>>' or '|'. Operators '>' and '>>' indicate whether to overwrite or append to the file. The operator '|' pipes content to the command arguments specified."
+            },
+            "error": {
+              "type": "string",
+              "description": "A filepath operation configuration. This is a string containing a filepath and an optional leading operator: '>', '>>' or '|'. Operators '>' and '>>' indicate whether to overwrite or append to the file. The operator '|' pipes content to the command arguments specified."
+            }
+          }
+        }
+      ]
+    },
+    "output_config": {
+      "type": "object",
+      "properties": {
+        "output": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "all": {
+              "$ref": "#/$defs/output_log_operation"
+            },
+            "init": {
+              "$ref": "#/$defs/output_log_operation"
+            },
+            "config": {
+              "$ref": "#/$defs/output_log_operation"
+            },
+            "final": {
+              "$ref": "#/$defs/output_log_operation"
+            }
+          }
+        }
+      }
+    },
     "reporting_config": {
       "type": "object",
       "properties": {
@@ -3671,6 +3726,9 @@
     },
     {
       "$ref": "#/$defs/reporting_config"
+    },
+    {
+      "$ref": "#/$defs/output_config"
     }
   ],
   "properties": {
@@ -3723,6 +3781,7 @@
     "mounts": {},
     "no_ssh_fingerprints": {},
     "ntp": {},
+    "output": {},
     "package_reboot_if_required": {},
     "package_update": {},
     "package_upgrade": {},
```

cloudinit/config/schema.py
```diff
@@ -668,6 +680,10 @@ def _annotate_content(
         deprecation_footer: List[str] = []
         error_index = 1
         deprecation_index = 1
+        if "generic" in errors_by_line:
```
blackboxsw, line 683: I think this "generic" label is applicable for items which don't have a path associated with them.

I think we can make an improvement here to better annotate the correct line of the top-level key
Most nested addtionalProperties errors will have a non-empty schema_path representing the path of the parent schema node. When we don't have a schema_path and have an additionalProperties validator error, we are likely at root of the document, but we can confirm that by also double-checking schema_error .instance matches our full schema (asserting we are validating top-level schema and not a subschema).

We can perform this proper path fallback earlier in validate_cloudconfig_schema
```diff
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -584,6 +584,17 @@ def validate_cloudconfig_schema(
         validator.iter_errors(config), key=lambda e: e.path
     ):
         path = ".".join([str(p) for p in schema_error.path])
+        if (
+            not path
+            and schema_error.validator == "additionalProperties"
+            and schema_error.schema == schema
+        ):
+            # an issue with invalid top-level property
+            prop_match = re.match(
+                r".*\('(?P<name>.*)' was unexpected\)", schema_error.message
+            )
+            if prop_match:
+                path = prop_match["name"]
         problem = (SchemaProblem(path, schema_error.message),)
         if isinstance(
             schema_error, SchemaDeprecationError
```

If we  we can still properly annotate top-level keys which don't adhere to strict schema by checking whether the json schema.


Testable with any bogus top-level key and an expectation to have an `# E1` annotation comment to the right of the offending invalid key.

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -3668,5 +3672,101 @@
     {
       "$ref": "#/$defs/reporting_config"
     }
-  ]
+  ],
+  "properties": {
```
blackboxsw, line 3676: Now that we have an "empty" top-level "properties" key present `cloudinit.config.schema._get_property_doc` needs to grow smarts to not document completely "empty/undefined properties" for modules which have no specific subschema.

doc/examples/etc_cloud_cfg.txt
```diff
@@ -0,0 +1,184 @@
+# The top level settings are used as module
```
blackboxsw, line 1: Great consolidation of doc content here. It's appropriate and gives context to the intent of system/base configuration.

Since we are filtering this out of cloud-config examples page, is there a good home we can find this example in `doc/rtd/reference/base_config_reference.rst`. Maybe in belongs down under the "Example" section of that page? There is a lot of duplication between this content and the example in the current doc.

blackboxsw (changes requested):
Thanks for progress on this @TheRealFalcon . A couple of functional things we need to sort, but looks promising.

cloudinit/config/schema.py
TheRealFalcon: Whooo, thanks! I didn't think it'd be this easy to add that in.

cloudinit/config/schemas/schema-cloud-config-v1.json
TheRealFalcon: Do you have a problematic example? I'm not sure I understand.

blackboxsw (commented):
@TheRealFalcon I've pushed a number of commits to a branch by the same name for comparison to capture a number of comments I mentioned in this review:
That branch is visible in [this comparison](https://github.com/TheRealFalcon/cloud-init/compare/strict-schema...blackboxsw:cloud-init:strict-schema)

I think we need the following: 
* bug(schema): add output config schema definitions
* bug(schema): only generate property_doc on non-empty properties
* feat(schema): provide annotation path for invalid top-level keys
* bug(cli): main source cloud_config for schema validation

This suggestions are represented as separate commits in 
https://github.com/blackboxsw/cloud-init/tree/strict-schema

I realize that you've pushed some of these changes into your current branch so some of them may now no longer be applicable.

cloudinit/config/schemas/schema-cloud-config-v1.json
blackboxsw: Sorry, I hadn't explained this context. We don't want our schema docs to reprsent this set of empty top-level properties when a given module does not actually have a subschema defined.
The way we avoid producing invalid top-level docs for a module which defined no schema is to avoid generating docs with _get_properties_doc when we see no specialized subschema.

This [property doc "skip" is done in the following commit](https://github.com/canonical/cloud-init/commit/ad5ac9383d1bb321d571060b796e0f5b91142d2e)

doc/rtd/reference/base_config_reference.rst
```diff
@@ -400,6 +407,32 @@ On an Ubuntu system, :file:`/etc/cloud/cloud.cfg` should look similar to:
             security: http://ports.ubuntu.com/ubuntu-ports
       ssh_svcname: ssh
 
+    # configure where output will go
+    output:
+    init: "> /var/log/my-cloud-init.log"
+    config: [ ">> /tmp/foo.out", "> /tmp/foo.err" ]
+    final:
+        output: "| tee /tmp/final.stdout | tee /tmp/bar.stdout"
+        error: "&1"
```
blackboxsw, line 416: Indent issue
```suggestion
      init: "> /var/log/my-cloud-init.log"
      config: [ ">> /tmp/foo.out", "> /tmp/foo.err" ]
      final:
        output: "| tee /tmp/final.stdout | tee /tmp/bar.stdout"
        error: "&1"
```

blackboxsw (commented):
Minor updates on output invalid indentation in schema example  I think this PR minimally needs to avoid documenting empty top-level keys for modules with no subschema as seen  `PYTHONPATH=. python3 -m cloudinit.cmd.main schema --docs cc_rightscale_userdata`

I'd also like to see us [avoid reading merged init.cfg in cloudinit.cmd.main](https://github.com/canonical/cloud-init/commit/3e30c69578e72c1dc2dfd96ae6fbf97df4db5801) because that generates invalid schema warnings in logs for some of the base-config keys.

TheRealFalcon:
@blackboxsw , I have added your commits along with a unit tests for top level properties. I think this is ready for review now.

blackboxsw (approved):
Thanks James for the test coverage here and updated. Validated behavior on invalid user-data generating warnings in logs as well as proper annotations on the cmdline. 
```

root@mmm:~# cloud-init query userdata
#cloud-config
output: 
   all: '| tee -a /var/log/cloud-init-new.log'
bootcmd:
   - echo "YEP"
package_update: false
packages: [sl]
invalid: broken
root@mmm:~# cloud-init status --format=json | jq .recoverable_errors
{
  "WARNING": [
    "Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.",
    "Stdout, stderr changing to (| tee -a /var/log/cloud-init-new.log, | tee -a /var/log/cloud-init-new.log)"
  ]
}
root@mmm:~# cloud-init schema --system --annotate
Found cloud-config data types: user-data, network-config

1. user-data at /var/lib/cloud/instances/ebeadcb1-ba58-4b5e-82b3-c2d8b78355b8/cloud-config.txt:
#cloud-config

# from 1 files
# part-001

---
bootcmd:
- echo "YEP"
invalid: broken		# E1
output:
    all: '| tee -a /var/log/cloud-init-new.log'
package_update: false
packages:
- sl
...

# Errors: -------------
# E1: Additional properties are not allowed ('invalid' was unexpected)



2. network-config at /var/lib/cloud/instances/ebeadcb1-ba58-4b5e-82b3-c2d8b78355b8/network-config.json:
  Valid schema network-config
```

Also validated non-error case, no warnings in logs or on command line.
```

root@mmm:~# 
root@mmm:~# cloud-init query userdata
#cloud-config
output: 
   all: '| tee -a /var/log/cloud-init-new.log'
bootcmd:
   - echo "YEP"
package_update: false
packages: [sl]
root@mmm:~# cloud-init status --format=json | jq .recoverable_errors
{
  "WARNING": [
    "Stdout, stderr changing to (| tee -a /var/log/cloud-init-new.log, | tee -a /var/log/cloud-init-new.log)"
  ]
}
root@mmm:~# cloud-init schema --system --annotate
Found cloud-config data types: user-data, network-config

1. user-data at /var/lib/cloud/instances/edabe2aa-263f-4549-866d-1609270e7e43/cloud-config.txt:
  Valid schema user-data

2. network-config at /var/lib/cloud/instances/edabe2aa-263f-4549-866d-1609270e7e43/network-config.json:
  Valid schema network-config
```

TheRealFalcon:
Most recent force push was re-ordering commits with no functional changes.

PR #4567, opened 2023-10-31
Holmanb/python3.13

holmanb:
## Proposed Commit Message

```
ci: Add Python 3.13
```

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -14,19 +14,28 @@ jobs:
   unittests:
     strategy:
       matrix:
-        python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11", "3.12" ]
-        toxenv: [ py3 ]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
+        toxenv: [py3]
         experimental: [false]
+        check-latest: [false]
+        continue-on-error: [false]
         include:
           - python-version: "3.6"
             toxenv: lowest-supported
+            continue-on-error: false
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
Should we close this PR and re-open when the run can work?

holmanb:
> Should we close this PR and re-open when the run can work?

Or maybe the opposite - we open if failing, so that we can easily check in periodically to see if it is failing because of test failures or still failing because of build failures. WDYT?

TheRealFalcon:
Keeping it with a non-vote to CI is good

TheRealFalcon:
> non-vote

Reading about it...is this an option? https://github.com/actions/runner/issues/2347

holmanb:
Requires https://github.com/canonical/cloud-init/pull/4669

TheRealFalcon:
Will this fail CI if the dev job fails?

holmanb:
> Will this fail CI if the dev job fails?

Yep, over 1000 reactions to [the top comment](https://github.com/actions/runner/issues/2347) and Github still hasn't bothered to implement this.

We can let this grow stale until 3.13 is released, if you want, but we had 3.12 as a job before it was released and didn't have any troubles, so alternatively we could just drop 3.13 if it causes troubles. I got a passing run on 3.13, so I'm happy either way.

TheRealFalcon (approved):
> alternatively we could just drop 3.13 if it causes troubles

Yeah, I think this makes more sense. Better to get the signal if we need it rather than ignore it due to possible breaking changes from Python, which seem pretty unlikely.

PR #4585, opened 2023-11-05
schema: improved CLI error messages on unprocessed raw data files

blackboxsw:
## Proposed Commit Message

```
When cloud-init encounters user-data and vendor-data that does not contain a valid header, cloud-init logs a warning and creates an empty processed user-data or vendor-data file at
/var/lib/cloud/instane/cloud-config.txt or vendor-cloud-config.txt.

In cases where invalid user-data or vendor-data was provided, prefer to process the raw data files to alert the end-user to the specific failure in parsing the raw data provided.

This approach improves error reporting from the CLI when user-data, vendor-data or vendor2-data contain undeclared or invalid headers.

Add a utility function get_processed_or_fallback_path which will return the preferred data path processed or raw fallback file depending on whether the processed file or the raw data file has content.

Two other minor changes:
 - Align exit 1 behavior for cloud-init schema --annotate command with and without --annotate parameter
 - refactor parts of handle_schema_args into simpler helper functions, _assert_exclusive_args and get_config_paths_from_args
```

## Checklist

## Merge type


holmanb (commented):
Does `get_processed_or_fallback_path()` need to go in `helpers.py`? Since code isn't reused in any other modules, or even outside of `get_config_paths_from_args()`, I'd prefer if we don't require importing a single-use function. Maybe a nested function in schema.py makes more sense?
```suggestion
diff --git a/cloudinit/config/schema.py b/cloudinit/config/schema.py
index 9c6cd585a..6ece12135 100644
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -9,6 +9,7 @@ import sys
 import textwrap
 from collections import defaultdict
 from collections.abc import Iterable
+from contextlib import suppress
 from copy import deepcopy
 from functools import partial
 from itertools import chain
@@ -28,7 +29,7 @@ import yaml
 from cloudinit import importer, safeyaml
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.handlers import INCLUSION_TYPES_MAP, type_from_starts_with
-from cloudinit.helpers import get_processed_or_fallback_path
+from cloudinit.helpers import Paths
 from cloudinit.sources import DataSourceNotFoundException
 from cloudinit.util import error, get_modules_from_dir, load_file
 
@@ -1474,6 +1475,29 @@ def _assert_exclusive_args(args):
 
 
 def get_config_paths_from_args(args) -> Tuple[str, Tuple[str, str]]:
+    def get_processed_or_fallback_path(
+        paths: Paths,
+        primary_path_key: str,
+        raw_fallback_path_key: str,
+    ) -> Optional[str]:
+        """Get processed data path when non-empty of fallback to raw data path.
+        - When primary path isn't set, return None as Paths structure is not setup.
+        - When primary path and raw path are both empty, prefer primary path.
+        - When primary path is empty but the raw fallback path is non-empty,
+          this indicates an invalid and ignored raw user-data was provided and
+          cloud-init emitted a warning and did not process unknown raw user-data.
+          In the case of invalid raw user-data header, prefer raw_fallback_path_key
+          so actionalbe sensible warnings can be reported ot the user about
+          the raw unparseable user-data.
+        """
+        primary_datapath = paths.get_ipath(primary_path_key) or ""
+        with suppress(FileNotFoundError):
+            if not os.stat(primary_datapath).st_size:
+                raw_path = paths.get_ipath(raw_fallback_path_key) or ""
+                if os.stat(raw_path).st_size:
+                    return raw_path
+        return primary_datapath or None
+
     try:
         paths = read_cfg_paths(fetch_existing_datasource="trust")
     except DataSourceNotFoundException:
diff --git a/cloudinit/helpers.py b/cloudinit/helpers.py
index f4ec47c32..e00045eff 100644
--- a/cloudinit/helpers.py
+++ b/cloudinit/helpers.py
@@ -14,7 +14,6 @@ import os
 from configparser import NoOptionError, NoSectionError, RawConfigParser
 from io import StringIO
 from time import time
-from typing import Optional
 
 from cloudinit import persistence, type_utils, util
 from cloudinit.settings import CFG_ENV_NAME, PER_ALWAYS, PER_INSTANCE, PER_ONCE
@@ -496,32 +495,3 @@ class DefaultingConfigParser(RawConfigParser):
         if header:
             contents = "\n".join([header, contents, ""])
         return contents
-
-
-def get_processed_or_fallback_path(
-    paths: Paths,
-    primary_path_key: str,
-    raw_fallback_path_key: str,
-) -> Optional[str]:
-    """Get processed data path when non-empty of fallback to raw data path.
-
-    - When primary path isn't set, return None as Paths structure is not setup.
-    - When primary path and raw path are both empty, prefer primary path.
-    - When primary path is empty but the raw fallback path is non-empty,
-      this indicates an invalid and ignored raw user-data was provided and
-      cloud-init emitted a warning and did not process unknown raw user-data.
-      In the case of invalid raw user-data header, prefer raw_fallback_path_key
-      so actionalbe sensible warnings can be reported ot the user about
-      the raw unparseable user-data.
-    """
-    primary_datapath = paths.get_ipath(primary_path_key)
-    if not primary_datapath:
-        return None
-    try:
-        if os.stat(primary_datapath).st_size == 0:
-            raw_path = paths.get_ipath(raw_fallback_path_key)
-            if os.stat(raw_path).st_size:
-                return raw_path
-    except FileNotFoundError:
-        pass
-    return primary_datapath
-- 
2.41.0

```

blackboxsw:
> Does `get_processed_or_fallback_path()` need to go in `helpers.py`? Since code isn't reused in any other modules, or even outside of `get_config_paths_from_args()`, I'd prefer if we don't require importing a single-use function. Maybe a nested function in schema.py makes more sense?

Thanks for the suggestion to move the function into the schema module . Defining `get_processed_or_fallback_path` doesn't have to be in helpers. I thought to put it there as it was Path related functions in general, but this is really only needed for schema module handling of specific config files and fallback. So, it makes sense to move it into the schema module as a single-use function. By nesting it as you suggest we avoid polluting the schema module level namespace with functions we don't expect to be called publicly, so I agree. Thanks for the suppress suggestion to avoid noisy try/except handling. If we encounter any FileNotFoundError on primary_path_key, this represents the datasource not being setup yet and we'll also hit FileNotFoundError for the fallback_path_key, so there is no concern with immediately exiting out of the suppress context and returning an empty string.

cloudinit/config/schema.py
```diff
@@ -1476,39 +1518,39 @@ def handle_schema_args(name, args):
...
-        except FileNotFoundError:
-            # Error handling on absent userdata_file below
-            pass
-
         config_files = (("user-data", userdata_file),)
-        supplemental_config_files = (
-            ("vendor-data", paths.get_ipath("vendor_cloud_config")),
-            ("vendor2-data", paths.get_ipath("vendor2_cloud_config")),
-            ("network-config", paths.get_ipath("network_config")),
+        supplemental_config_files: Tuple[Tuple[str, str], ...] = (
```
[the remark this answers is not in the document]
blackboxsw: Great suggestions here. While it's a bit of a refactor of the logic, now is as good a time as any because a followup PR will make these even harder when we are dealing with network schema version 2. I've added Enums and an InstanceDataType named tuple to better capture the magic strings and clarify specific SchemaTypes associated with user-data, vendor*data and network-config.

cloudinit/config/schema.py
```diff
@@ -1457,10 +1472,37 @@ def handle_schema_args(name, args):
             "Invalid flag combination. Cannot use --annotate with --docs",
             sys_exit=True,
         )
-    full_schema = get_schema(schema_type="cloud-config")
-    if args.docs:
-        print(load_doc(args.docs))
-        return
+
+
+def get_config_paths_from_args(
```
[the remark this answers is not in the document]
blackboxsw: This was only for readability. The complexity and length of `handle_schema_args` was getting unwieldy and incredibly long making it tough to grok all the things the function was doing. I opted to tease out the `get_config_paths_from_args` helper function out of `handle_schema_args` function to perform the initial instance data part detection

cloudinit/config/schema.py
```diff
@@ -1457,10 +1472,37 @@ def handle_schema_args(name, args):
...
+    args,
+) -> Tuple[str, Tuple[Tuple[str, str], ...]]:
+    def get_processed_or_fallback_path(
+        paths: Paths,
+        primary_path_key: str,
+        raw_fallback_path_key: str,
+    ) -> str:
+        """Get processed data path when non-empty of fallback to raw data path.
+        - When primary path isn't set, return empty string as Paths structure
```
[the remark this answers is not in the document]
blackboxsw: You are correct. I've dropped this comment. Prior to cloud-init running, Paths will actually default to an instance-id of `iid-datasource-none` instead of as I supposed with the comment. This leaves our paths.get_ipath("userdata") call with a paths response such as `/var/lib/cloud/instances/iid-datasource-none/user-data.txt.i`. In the case where cloud-init hasn't run yet, (or we just ran cloud-init clean) we'll end up with an error message like:

 `Error: Config file /var/lib/cloud/instances/iid-datasource-none/cloud-config.txt does not exist`

I'll drop this comment as it's incorrect

blackboxsw:
Thanks @holmanb for this I addressed your comments, refactored a bit of the code and amended commit messages per your review. It's ready for review again now.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
+repos:
```
holmanb, line 1: did you mean to include this file? I assume this was an accidental commit

cloudinit/config/schema.py
```diff
@@ -148,6 +143,49 @@ def format(self) -> str:
...
+class InstanceDataType(Enum):
+    """Types of instance data provided to cloud-init"""
+
+    USERDATA = "user-data"
+    NETWORK_CONFIG = "network-config"
+    VENDORDATA = "vendor-data"
+    VENDOR2DATA = "vendor2-data"
+    # METADATA = "metadata"
+
+    def __str__(self):  # pylint: disable=invalid-str-returned
```
holmanb, line 179: I think we want this class to be a [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) here, right? No need to re-invent it.

cloudinit/config/schema.py
```diff
@@ -1519,101 +1603,131 @@ def handle_schema_args(name, args):
...
-        supplemental_config_files = (
-            ("vendor-data", paths.get_ipath("vendor_cloud_config")),
-            ("vendor2-data", paths.get_ipath("vendor2_cloud_config")),
-            ("network-config", paths.get_ipath("network_config")),
         )
-        for cfg_type, cfg_file in supplemental_config_files:
-            if cfg_file and os.path.exists(cfg_file):
-                config_files += ((cfg_type, cfg_file),)
-    if not os.path.exists(config_files[0][1]):
+        supplemental_config_files: List[InstanceDataPart] = [
```
holmanb, line 1636: much more structured - thanks!

holmanb (approved):
Thanks for the refactoring @blackboxsw. I'm guessing that your pre-commit config was accidentally added to this PR?

Otherwise one small nit - the rest looks good to me!

.pre-commit-config.yaml
blackboxsw: whoops dropped.

cloudinit/config/schema.py
blackboxsw: You are right we do want StrEnum, but we can't yet because we still support python 3.8 on Focal and StrEnum wasn't introduced until 3.11. So until the, we "pre-invent" it :)

holmanb:
@blackboxsw you requested a squash merge request, but merged it as a rebase merge. I would have reviewed this differently had I know that you weren't going to squash it. Also, the `.pre-commit-config.yaml` gets added and removed between these two commits. It's not a big deal, and not worth changing at this point, but lets please make sure to mark the merge type accordingly and follow that in the future when merging.

PR #4586, opened 2023-11-06
integration_tests: Support running on non-Ubuntu distros

holmanb:
## Proposed Commit Message

```
integration_tests: Support running on non-Ubuntu distros

Commit 2562f93383 broke cross-platform support. Fix it.

Also remove an unused function.
```

### DIFF
--- a/tests/integration_tests/releases.py
+++ b/tests/integration_tests/releases.py
@@ -5,31 +5,24 @@ from typing import Optional
 from packaging import version
 
 from cloudinit import subp
-from cloudinit.subp import ProcessExecutionError
 from tests.integration_tests import integration_settings
 
 log = logging.getLogger("integration_testing")
 
 
-def get_all_ubuntu_series() -> list:
-    """Use distro-info-data's ubuntu.csv to get a list of Ubuntu series"""
-    out = ""
-    try:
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
The code is fine, but I'm not sure this really buys us anything.

The one call site looks like this:
```python
        if len(parts) == 1:
            image_id = None
            os = "ubuntu"
            series = parts[0]
            version = ubuntu_version_from_series(series)
```

so it is still Ubuntu-specific and will only run if you try to use something like `jammy` as the image specification. We could probably restructure the call site to make that a little more clear, but I don't see an advantage to this change as-is.

holmanb:
>  but I don't see an advantage to this change as-is.

@TheRealFalcon  Integration tests cannot run on a non-ubuntu distro with the `main` branch, but they can on this branch. That seems like an advantage to me ¯\_(ツ)_/¯

holmanb:
> @TheRealFalcon Integration tests cannot run on a non-ubuntu distro with the `main` branch, but they can on this branch. That seems like an advantage to me ¯_(ツ)_/¯

This PR is about the distro of the host, not the distro of the instance being tested. Maybe that is the confusion here?

TheRealFalcon (approved):
> Maybe that is the confusion here?

Yes, I understand now. LGTM!

PR #4587, opened 2023-11-06
Remove dead code

holmanb:
## Additional Context
Remove dead code

## Merge type

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2316,10 +2316,6 @@ def make_header(comment_char="#", base="created"):
     return header
 
 
-def abs_join(base, *paths):
-    return os.path.abspath(os.path.join(base, *paths))
-
-
 # shellify, takes a list of commands
 #  for each entry in the list
 #    if it is an array, shell protect it (with single ticks)
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -826,23 +826,3 @@ def oauth_headers(
     )
     _uri, signed_headers, _body = client.sign(url)
     return signed_headers
-
-
-def retry_on_url_exc(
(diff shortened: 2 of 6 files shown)


cloudinit/util.py
```diff
@@ -2316,10 +2316,6 @@ def make_header(comment_char="#", base="created"):
     return header
 
 
-def abs_join(base, *paths):
```
TheRealFalcon, line 2319: Why are we duplicating this multiple times now?

holmanb: Right, I didn't bother to explain that tradeoff.

My argument here is: "packaging code, and especially packaging code that is only used for building test packages, should not be shipped as part of the cloud-init runtime".

If we agree on that point we have a few options:

1. duplicate (chosen)
2. replace this one function call with the two function calls it simplifies at callsites 
3. create a new module to store shared code

Option 2 involved more typing. Option 3 seemed like overkill for a one line function. Option 1 seemed like a reasonable tradeoff, but I will happily do 2 if preferred. Option 3 sounds like a refactor to make it worthwhile. If there are other sharable/shared functions to consolidate I guess we can do that too.

holmanb: @TheRealFalcon alright, per out of band convo it sounds like 2 is preferred.

TheRealFalcon (commented):
Did `abs_join` get unremoved? I'm seeing it removed here.

holmanb:
> Did `abs_join` get unremoved? I'm seeing it removed here.

Thanks, fixed!

TheRealFalcon (approved):

holmanb:
force pushed the to update the commit wording with PR number

Issue #4044, opened 2023-05-12
Netplan/Systemd/Cloud-init/Dbus Race


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1997124](https://bugs.launchpad.net/cloud-init/+bug/1997124)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = ['netplan', 'systemd (Ubuntu)']
assignee = falcojr
assignee_name = James Falcon
date_closed = None
date_created = 2022-11-18T23:07:12.002484+00:00
date_fix_committed = None
date_fix_released = None
id = 1997124
importance = high
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1997124
milestone = None
owner = holmanb
owner_name = Brett Holman
private = False
status = in_progress
submitter = holmanb
submitter_name = Brett Holman
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Brett Holman(holmanb)** wrote on 2022-11-18T23:07:12.002484+00:00_

Cloud-init is seeing intermittent failures while running `netplan apply`, which appears to be caused by a missing resource at the time of call.

The symptom in cloud-init logs looks like:

Running ['netplan', 'apply'] resulted in stderr output: Failed to connect system bus: No such file or directory

I think that this error[1] is likely caused by cloud-init running netplan apply too early in boot process (before dbus is active).

Today I stumbled upon this error which was hit in MAAS[2]. We have also hit it intermittently during tests (we didn't have a reproducer).

Realizing that this may not be a cloud-init error, but possibly a dependency bug between dbus/systemd we decided to file this bug for broader visibility to other projects.

I will follow up this initial report with some comments from our discussion earlier. 

[1] https://github.com/canonical/netplan/blob/main/src/dbus.c#L801
[2] https://discourse.maas.io/t/latest-ubuntu-20-04-image-causing-netplan-error/5970

ubuntu-server-builder:
_Launchpad user **Brett Holman(holmanb)** wrote on 2022-11-18T23:23:58.106907+00:00_

Some details from a conversation with Chad, James, and Vorlon.

`netplan apply` is executed in cloud-init.service, which runs
`Before=network-online.target` but `After=systemd-networkd-wait-online.service`. 

There is may be a dependency bug between dbus and systemd-networkd because systemd-networkd is a dbus service so when it's "up" it should be accessible over dbus.

[...]

PR #4592, opened 2023-11-06
WIP: Fix netplan systemd dbus race

holmanb:
## Additional Context
This is a rough sketch of how we might fix the persistent dbus/networkd/netplan race that we've been seeing for some time now. The commit messages and comments explain the code changes.

Feel free to bikeshed implementation details.

Fixes https://github.com/canonical/cloud-init/issues/4044

TODO: Clean up logic
TODO: Tests

Review by an SRU reviewer desired after details are hammered out, but before merge.

## Merge type

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -431,8 +431,24 @@ def main_init(name, args):
         # dhcp clients to advertize this hostname to any DDNS services
         # LP: #1746455.
         _maybe_set_hostname(init, stage="local", retry_stage="network")
-    init.apply_network_config(bring_up=bring_up_interfaces)
-
+        init.apply_network_config(bring_up=bring_up_interfaces)
+    else:
+        # Until dbus is available, systemd-networkd cannot reload its
+        # config, and therefore a race has occurred: a new configuration was
+        # written, but systemd-networkd was not notified.
+        # Until dbus is available, we mustn't continue otherwise our config
+        # is lost until the next time systemd-networkd restarts. As a oneshot,
+        # cloud-init.service is started before dbus.socket, but does not block
+        # it. Retry until dbus becomes available.
+        while not init.apply_network_config(bring_up=bring_up_interfaces):
+            LOG.warning(
+                "Dbus is not available to notify systemd-networkd of"
+                "a new configuration from netplan. Retrying."
+            )
+            # Why retry every 0.042? it uses a fraction of a percent of one CPU
+            # This seems like a reasonable tradeoff between "get there fast"
+            # and "leave lots of resources available for others".
+            time.sleep(0.042)
     if mode == sources.DSMODE_LOCAL:
         if init.datasource.dsmode != mode:
             LOG.debug(
--- a/cloudinit/net/activators.py
+++ b/cloudinit/net/activators.py
@@ -24,7 +24,6 @@ def _alter_interface(cmd, device_name, fail_on_stderr=False) -> bool:
         if err and fail_on_stderr and "Failed to connect system bus:" in err:
             LOG.warning("Running %s resulted in stderr output: %s", cmd, err)
             return False
-        LOG.info("Running %s resulted in stderr output: %s", cmd, err)
         return True
     except subp.ProcessExecutionError:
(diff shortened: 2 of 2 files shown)


cloudinit/cmd/main.py
```diff
@@ -431,8 +431,24 @@ def main_init(name, args):
...
+        # cloud-init.service is started before dbus.socket, but does not block
+        # it. Retry until dbus becomes available.
+        while not init.apply_network_config(bring_up=bring_up_interfaces):
+            LOG.warning(
+                "Dbus is not available to notify systemd-networkd of"
+                "a new configuration from netplan. Retrying."
+            )
+            # Why retry every 0.042? it uses a fraction of a percent of one CPU
+            # This seems like a reasonable tradeoff between "get there fast"
+            # and "leave lots of resources available for others".
```
cjp256, line 450: lol this made me laugh, but really I want to know _why 0.042_ :D

holmanb: Heh, I'm a big fan of a certain universal guide ;)

holmanb: Hitchhiker's reference aside, I can run something like:
```python
>>> import time
>>> while 1:
...     time.sleep(N)
```

And it doesn't even register in `htop` above around `N=0.02`. With that in mind, somewhere between `0.02` and `0.1` seemed acceptable for a WIP POC like this - happy to make it whatever seems best :D

cloudinit/net/activators.py
```diff
@@ -17,12 +17,14 @@ class NoActivatorException(Exception):
...
+def _alter_interface(cmd, device_name, fail_on_stderr=False) -> bool:
     LOG.debug("Attempting command %s for device %s", cmd, device_name)
     try:
-        (_out, err) = subp.subp(cmd)
-        if len(err):
+        err = subp.subp(cmd).stderr
+        if err and fail_on_stderr and "Failed to connect system bus:" in err:
             LOG.warning("Running %s resulted in stderr output: %s", cmd, err)
+            return False
+        LOG.info("Running %s resulted in stderr output: %s", cmd, err)
```
TheRealFalcon, line 27: I don't think we want this log when there is no stderr.

cloudinit/net/activators.py
```diff
@@ -17,12 +17,13 @@ class NoActivatorException(Exception):
...
 
 
-def _alter_interface(cmd, device_name) -> bool:
+def _alter_interface(cmd, device_name, fail_on_stderr=False) -> bool:
     LOG.debug("Attempting command %s for device %s", cmd, device_name)
     try:
-        (_out, err) = subp.subp(cmd)
-        if len(err):
+        err = subp.subp(cmd).stderr
+        if err and fail_on_stderr and "Failed to connect system bus:" in err:
```
TheRealFalcon, line 24: I might be a little pedantic here, but don't we want the LOG anytime we have stderr? Something like:
```python
try:
    err = subp.subp(cmd).stderr
    if err:
        LOG.warning("...")
        if fail_on_stderr and "Failed to connect system bus:" in err:
            return False
    ...
```
?

slyon:
This feels like a workaround to me, and I feel we should rather try to fix the original ordering problem.

I do not fully understand the cloud-init ordering stages we're interacting with here. Maybe you can provide some more context?

In the commit message you suggest this ordering:
systemd-netword.service < systemd-networkd-network-wait-online.service < cloud-init.service

But cloud-init.service also needs dbus.socket, therefore implementing some polling to wait for it. Why can't we just define a systemd dependency, so that cloud-init.service runs AFTER systemd-networkd-wait-online.service AND dbus.socket?

Or alternatively have an additional cloud-init stage, that runs BEFORE systemd-networkd.service, only calling `netplan generate` instead of `netplan apply` (so doesn't need DBus and will be applied by systemd-networkd starting afterwards, implicitly). IIUC there is already a previous cloud-init stage where you spin up some ephemeral DHCP connection, using a DHCP client manually, in order to fetch the actual Netplan configuration from an online data-source. So the Netplan YAML config should be available before starting systemd-networkd, right?

PR #4593, opened 2023-11-07
chore: add conventional commits template

TheRealFalcon:
## Proposed Commit Message

```
chore: add conventional commits template
```

We need to actually follow this if merged :smile:


holmanb (approved):
I think the value of classifying commit types in terms of grokking history quickly (and maybe changelog automation or something) outweighs the process cost. +1

Lets leave open for more feedback.

aciba90 (approved):
I totally agree, I think this adds value by itself in the short term and we could used tooling around conventional commits in the mid/long term to extract even more value. Thanks for bringing this dicussion up.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,13 +1,22 @@
...
+
+If the change is related to a particular cloud or particular distro,
+please include the "optional scope" in the summary line. E.g.,
+feat(ec2): Add support for foo to the baz
+
+Types used by this project:
+feat, fix, docs, ci, test, refactor, chore
+-->
 ```
-summary: no more than 70 characters
```
blackboxsw, line 5: I'd like to see us define an acceptable line LENGTH for titile/summary. I think it would be reasonable to set 72 as this width.

blackboxsw (commented):
Please do. I think now that we are allowing/performing more rebase commits instead squash-merged commits at PR-level granularity, we now represent a lot more commits which can be better categorized to assist with release content generation. While this may increase pressure on sensible commit title line-length of 72 in the title of a commit, I hope we can guide contributors to be creative and brief in our commit message titles to avoid exceeding this suggested limit.

.github/PULL_REQUEST_TEMPLATE.md
TheRealFalcon: Addressed in most recent force push

TheRealFalcon:
Given 2 +1s and another +1 to the idea with a change requested that has been addressed, I'm going to merge this. We can always iterate further if needed.

TheRealFalcon:
We had something in the docs originally, but I removed it. It was out of date compared to the actual template we were using, and I didn't see a strong reason to have it duplicated in both places.

What do you think about a more general "we use conventional commits" with a link to the official webpage and then a link back to our repo with the current template?

holmanb:
@TheRealFalcon I already wrote up [this issue](https://github.com/canonical/cloud-init/issues/4646) to track this request before seeing your response, feel free to continue the conversation there.

> I didn't see a strong reason to have it duplicated in both places.

Agreed, duplication feels weird, but when reviewing a PR I'd like to be able to drop a link to a header in our docs on RTD than have to point to the markdown template, which has lots of non-commit-format information to parse.

> What do you think about a more general "we use conventional commits" with a link to the official webpage and then a link back to our repo with the current template?

That works for me. That doesn't cover the line length expectation, which I'd like to see documented too.

PR #4596, opened 2023-11-08
azure: Check for stale pps data from IMDS

KsenijaS:
## preferred commit message
```
azure: Check for stale pps data from IMDS

After reprovisioning, IMDS should return None value for PPS data
type.

Introduce a check to verify it IMDS is returning stale-PPS
data. In case of stale data, report the error to the host and allow
the VM to continue with provisioning. This is part of a larger effort
to improve error handling and error reporting during VM provisioning.
```

After reprovisioning, IMDS should return None value for PPS data type. This PR introduces a check to verify if IMDS is returning stale PPS data. In case of stale data, error will be reported to the host and VM will continue with provisioning.
This is a part of a large larg

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -661,6 +661,15 @@ class DataSourceAzure(sources.DataSource):
             # fetch metadata again as it has changed after reprovisioning
             imds_md = self.get_metadata_from_imds(report_failure=True)
 
+            # validate imds pps metadata
+            imds_ppstype = self._ppstype_from_imds(imds_md)
+            if imds_ppstype != PPSType.NONE.value:
+                self._report_failure(
+                    errors.ReportableErrorImdsInvalidMetadata(
+                        key="extended.compute.ppsType", value=imds_ppstype
+                    )
+                )
+
         # Report errors if IMDS network configuration is missing data.
         self.validate_imds_network_metadata(imds_md=imds_md)
 
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -663,7 +663,7 @@ class DataSourceAzure(sources.DataSource):
 
             # validate imds pps metadata
             imds_ppstype = self._ppstype_from_imds(imds_md)
-            if imds_ppstype != PPSType.NONE.value:
+            if imds_ppstype is not None and imds_ppstype != PPSType.NONE.value:
                 self._report_failure(
                     errors.ReportableErrorImdsInvalidMetadata(
                         key="extended.compute.ppsType", value=imds_ppstype
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -663,7 +663,7 @@ class DataSourceAzure(sources.DataSource):
 
             # validate imds pps metadata
             imds_ppstype = self._ppstype_from_imds(imds_md)
-            if imds_ppstype is not None and imds_ppstype != PPSType.NONE.value:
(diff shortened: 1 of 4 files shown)


cloudinit/sources/DataSourceAzure.py
```diff
@@ -661,6 +661,15 @@ def crawl_metadata(self):
             # fetch metadata again as it has changed after reprovisioning
             imds_md = self.get_metadata_from_imds(report_failure=True)
 
+            # validate imds pps metadata
+            imds_ppstype = self._ppstype_from_imds(imds_md)
+            if imds_ppstype != PPSType.NONE.value:
```
cjp256, line 666: imds_ppstype can be None if we fallback to the old API (and we shouldn't report failure).
```suggestion
            if imds_ppstype is not None and imds_ppstype != PPSType.NONE.value:
```

cjp256 (approved):
LGTM, tests look good, thanks!

cjp256:
@blackboxsw @TheRealFalcon can you tag this for 23.4 too, thanks!

cloudinit/sources/DataSourceAzure.py
```diff
@@ -661,6 +661,15 @@ def crawl_metadata(self):
             # fetch metadata again as it has changed after reprovisioning
             imds_md = self.get_metadata_from_imds(report_failure=True)
 
+            # validate imds pps metadata
+            imds_ppstype = self._ppstype_from_imds(imds_md)
+            if imds_ppstype is not None and imds_ppstype != PPSType.NONE.value:
```
blackboxsw, line 666: easier to read?
```suggestion
            if imds_ppstype not in (None, PPSType.NONE.value):
```

tests/unittests/sources/test_azure.py
```diff
@@ -3759,13 +3770,79 @@ def test_no_pps(self):
...
+
+        # Verify DMI usage.
+        assert self.mock_dmi_read_dmi_data.mock_calls == [
+            mock.call("chassis-asset-tag"),
+            mock.call("system-uuid"),
+            mock.call("system-uuid"),
+        ]
+
+        # Verify reports via KVP.
+        assert len(self.mock_kvp_report_failure_to_host.mock_calls) == 1
```
blackboxsw, line 3827: You can drop this assert line because you test the actual value and length of response below

tests/unittests/sources/test_azure.py
```diff
@@ -3759,13 +3770,79 @@ def test_no_pps(self):
         assert len(self.mock_kvp_report_failure_to_host.mock_calls) == 0
         assert len(self.mock_kvp_report_success_to_host.mock_calls) == 1
 
+    @pytest.mark.parametrize("pps_type", ["Savable", "Running"])
```
blackboxsw, line 3773: Can we get coverage of pps_type = "None" here too to assert it doesn't generate an error in this case? or is this covered already elsewhere?

blackboxsw (changes requested):
Thanks for this submission. Content looks good, minor nit on the conditional. I was wondering if we can get unittest coverage that exercises the PPSType.None value to assert we will not report failures  to KVP in that case. Not critical, but would be nice to have

tests/unittests/sources/test_azure.py
KsenijaS: I think that case is already covered in test_recovery_pps test.

tests/unittests/sources/test_azure.py
KsenijaS: done

cloudinit/sources/DataSourceAzure.py
KsenijaS: done

blackboxsw (approved):
LGTM

PR #4599, opened 2023-11-09
config: Module documentation updates

holmanb:
## Proposed Commit Message
```
config: Module documentation updates

- improve security wording
- standardize default value formatting
```

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -231,15 +231,15 @@
           "type": "boolean"
         },
         "passwd": {
-          "description": "Hash of user password applied when user does not exist. This will NOT be applied if the user already exists. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
+          "description": "Hash of user password applied when user does not exist. This will NOT be applied if the user already exists. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
           "type": "string"
         },
         "hashed_passwd": {
-          "description": "Hash of user password to be applied. This will be applied even if the user is pre-existing. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
+          "description": "Hash of user password to be applied. This will be applied even if the user is preexisting. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
           "type": "string"
         },
         "plain_text_passwd": {
-          "description": "Clear text of user password to be applied. This will be applied even if the user is pre-existing. There are many more secure options than using plain text passwords, such as ``ssh_import_id``. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
+          "description": "Clear text of user password to be applied. This will be applied even if the user is preexisting. There are many more secure options than using plain text passwords, such as ``ssh_import_id``. **Note:** Your password might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
           "type": "string"
         },
         "create_groups": {
@@ -308,7 +308,7 @@
           ]
         },
         "uid": {
-          "description": "The user's ID. Default value [system default].",
+          "description": "The user's ID. Default value [system default]",
           "oneOf": [
             {
               "type": "integer"
@@ -815,7 +815,7 @@
             },
             "conf": {
               "type": "string",
-              "description": "Specify configuration for apt, such as proxy configuration. This configuration is specified as a string. For multiline APT configuration, make sure to follow yaml syntax."
+              "description": "Specify configuration for apt, such as proxy configuration. This configuration is specified as a string. For mult-line APT configuration, make sure to follow YAML syntax."
             },
             "https_proxy": {
               "type": "string",
@@ -1113,7 +1113,7 @@
       "properties": {
         "disable_ec2_metadata": {
           "default": false,
-          "description": "Set true to disable IPv4 routes to EC2 metadata. Default: ``false.``",
+          "description": "Set true to disable IPv4 routes to EC2 metadata. Default: ``false``",
           "type": "boolean"
         }
       }
@@ -1308,7 +1308,7 @@
           "properties": {
             "mode": {
               "default": "auto",
-              "description": "The utility to use for resizing. Default: ``auto``\n\nPossible options:\n\n* ``auto`` - Use any available utility\n\n* ``growpart`` - Use growpart utility\n\n* ``gpart`` - Use BSD gpart utility\n\n* ``off`` - Take no action.",
+              "description": "The utility to use for resizing. Default: ``auto``\n\nPossible options:\n\n* ``auto`` - Use any available utility\n\n* ``growpart`` - Use growpart utility\n\n* ``gpart`` - Use BSD gpart utility\n\n* ``off`` - Take no action",
               "oneOf": [
                 {
                   "enum": [
@@ -2199,7 +2199,7 @@
             },
             "conf": {
               "type": "object",
-              "description": "Every key present in the conf object will be added to puppet.conf. As such, section names should be one of: ``main``, ``server``, ``agent`` or ``user`` and keys should be valid puppet configuration options. The configuration is specified as a dictionary containing high-level ``<section>`` keys and lists of ``<key>=<value>`` pairs within each section. The ``certname`` key supports string substitutions for ``%i`` and ``%f``, corresponding to the instance id and fqdn of the machine respectively.\n\n``ca_cert`` is a special case. It won't be added to puppet.conf. It holds the puppetserver certificate in pem format. It should be a multi-line string (using the | yaml notation for multi-line strings).",
+              "description": "Every key present in the conf object will be added to puppet.conf. As such, section names should be one of: ``main``, ``server``, ``agent`` or ``user`` and keys should be valid puppet configuration options. The configuration is specified as a dictionary containing high-level ``<section>`` keys and lists of ``<key>=<value>`` pairs within each section. The ``certname`` key supports string substitutions for ``%i`` and ``%f``, corresponding to the instance id and fqdn of the machine respectively.\n\n``ca_cert`` is a special case. It won't be added to puppet.conf. It holds the puppetserver certificate in pem format. It should be a multi-line string (using the | YAML notation for multi-line strings).",
               "additionalProperties": false,
               "properties": {
                 "main": {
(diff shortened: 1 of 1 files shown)


cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2636,7 +2636,7 @@
               "description": "Whether to expire all user passwords such that a password will need to be reset on the user's next login. Default: ``true``"
             },
             "users": {
-              "description": "This key represents a list of existing users to set passwords for. Each item under users contains the following required keys: ``name`` and ``password`` or in the case of a randomly generated password, ``name`` and ``type``. The ``type`` key has a default value of ``hash``, and may alternatively be set to ``text`` or ``RANDOM``.",
+              "description": "This key represents a list of existing users to set passwords for. Each item under users contains the following required keys: ``name`` and ``password`` or in the case of a randomly generated password, ``name`` and ``type``. The ``type`` key has a default value of ``hash``, and may alternatively be set to ``text`` or ``RANDOM``. Randomly generated keys may be insecure, use at your own risk.",
```
blackboxsw, line 2639:
```suggestion
              "description": "This key represents a list of existing users to set passwords for. Each item under users contains the following required keys: ``name`` and ``password`` or in the case of a randomly generated password, ``name`` and ``type``. The ``type`` key has a default value of ``hash``, and may alternatively be set to ``text`` or ``RANDOM``. Randomly generated passwords may be insecure, use at your own risk.",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -231,15 +231,15 @@
...
+          "description": "Hash of user password applied when user does not exist. This will NOT be applied if the user already exists. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
           "type": "string"
         },
         "hashed_passwd": {
-          "description": "Hash of user password to be applied. This will be applied even if the user is pre-existing. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** While ``hashed_password`` is better than ``plain_text_passwd``, using ``passwd`` in user-data represents a security risk as user-data could be accessible by third-parties depending on your cloud platform.",
+          "description": "Hash of user password to be applied. This will be applied even if the user is pre-existing. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
```
blackboxsw, line 238:
```suggestion
          "description": "Hash of user password to be applied. This will be applied even if the user is preexisting. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=4096. **Note:** Your password might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -231,15 +231,15 @@
...
           "type": "string"
         },
         "plain_text_passwd": {
-          "description": "Clear text of user password to be applied. This will be applied even if the user is pre-existing. There are many more secure options than using plain text passwords, such as ``ssh_import_id`` or ``hashed_passwd``. Do not use this in production as user-data and your password can be exposed.",
+          "description": "Clear text of user password to be applied. This will be applied even if the user is pre-existing. There are many more secure options than using plain text passwords, such as ``ssh_import_id``. **Note:** Your password might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
```
blackboxsw, line 242:
```suggestion
          "description": "Clear text of user password to be applied. This will be applied even if the user is preexisting. There are many more secure options than using plain text passwords, such as ``ssh_import_id``. **Note:** Your password might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -2833,7 +2833,7 @@
...
+          "description": "A dictionary entries for the public and private host keys of each desired key type. Entries in the ``ssh_keys`` config dict should have keys in the format ``<key type>_private``, ``<key type>_public``, and, optionally, ``<key type>_certificate``, e.g. ``rsa_private: <key>``, ``rsa_public: <key>``, and ``rsa_certificate: <key>``. Not all key types have to be specified, ones left unspecified will not be used. If this config option is used, then separate keys will not be automatically generated. In order to specify multiline private host keys and certificates, use yaml multiline syntax. **Note:** Your ssh keys might possibly be visible to unpriveledged users on your system, depending on your cloud's security model.",
```
blackboxsw, line 2836:
```suggestion
          "description": "A dictionary entries for the public and private host keys of each desired key type. Entries in the ``ssh_keys`` config dict should have keys in the format ``<key type>_private``, ``<key type>_public``, and, optionally, ``<key type>_certificate``, e.g. ``rsa_private: <key>``, ``rsa_public: <key>``, and ``rsa_certificate: <key>``. Not all key types have to be specified, ones left unspecified will not be used. If this config option is used, then separate keys will not be automatically generated. In order to specify multi-line private host keys and certificates, use YAML multi-line syntax. **Note:** Your ssh keys might possibly be visible to unprivileged users on your system, depending on your cloud's security model.",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -308,7 +308,7 @@
           ]
         },
         "uid": {
-          "description": "The user's ID. Default is next available value.",
+          "description": "The user's ID. Default value [system default].",
```
blackboxsw, line 311: We don't want trailing period/stop here as it gets rendered in docs as `uid: (integer/string) The user’s ID. Default value [system default]..`
```suggestion
          "description": "The user's ID. Default value [system default]",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -1113,7 +1113,7 @@
       "properties": {
         "disable_ec2_metadata": {
           "default": false,
-          "description": "Set true to disable IPv4 routes to EC2 metadata. Default: false.",
+          "description": "Set true to disable IPv4 routes to EC2 metadata. Default: ``false.``",
```
blackboxsw, line 1116:
```suggestion
          "description": "Set true to disable IPv4 routes to EC2 metadata. Default: ``false``",
```

cloudinit/config/schemas/schema-cloud-config-v1.json
blackboxsw: Other occasions we could clean up to avoid the duplicated stop ".." 
```diff
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -297,7 +297,7 @@
                 "string",
                 "null"
               ],
-              "description": "Sudo rule to use or false. Absence of a sudo value or ``null`` will result in no sudo rules added for this user."
+              "description": "Sudo rule to use or false. Absence of a sudo value or ``null`` will result in no sudo rules added for this user"
             },
             {
               "type": "boolean",
@@ -308,7 +308,7 @@
           ]
         },
         "uid": {
-          "description": "The user's ID. Default value [system default].",
+          "description": "The user's ID. Default value [system default]",
           "oneOf": [
             {
               "type": "integer"
@@ -1308,7 +1308,7 @@
           "properties": {
             "mode": {
               "default": "auto",
-              "description": "The utility to use for resizing. Default: ``auto``\n\nPossible options:\n\n* ``auto`` - Use any available utility\n\n* ``growpart`` - Use growpart utility\n\n* ``gpart`` - Use BSD gpart utility\n\n* ``off`` - Take no action.",
+              "description": "The utility to use for resizing. Default: ``auto``\n\nPossible options:\n\n* ``auto`` - Use any available utility\n\n* ``growpart`` - Use growpart utility\n\n* ``gpart`` - Use BSD gpart utility\n\n* ``off`` - Take no action",
               "oneOf": [
                 {
                   "enum": [
@@ -2624,7 +2624,7 @@
               "changed_description": "Use of non-boolean values for this field is deprecated."
             }
           ],
-          "description": "Sets whether or not to accept password authentication. ``true`` will enable password auth. ``false`` will disable. Default: leave the value unchanged. In order for this config to be applied, SSH may need to be restarted. On systemd systems, this restart will only happen if the SSH service has already been started. On non-systemd systems, a restart will be attempted regardless of the service state."
+          "description": "Sets whether or not to accept password authentication. ``true`` will enable password auth. ``false`` will disable. Default: leave the value unchanged. In order for this config to be applied, SSH may need to be restarted. On systemd systems, this restart will only happen if the SSH service has already been started. On non-systemd systems, a restart will be attempted regardless of the service state"
         },
         "chpasswd": {
           "type": "object",
@@ -3053,7 +3053,7 @@
       "properties": {
         "manage_etc_hosts": {
           "default": false,
-          "description": "Whether to manage ``/etc/hosts`` on the system. If ``true``, render the hosts file using ``/etc/cloud/templates/hosts.tmpl`` replacing ``$hostname`` and ``$fdqn``. If ``localhost``, append a ``127.0.1.1`` entry that resolves from FQDN and hostname every boot. Default: ``false``.",
+          "description": "Whether to manage ``/etc/hosts`` on the system. If ``true``, render the hosts file using ``/etc/cloud/templates/hosts.tmpl`` replacing ``$hostname`` and ``$fdqn``. If ``localhost``, append a ``127.0.1.1`` entry that resolves from FQDN and hostname every boot. Default: ``false``",
           "oneOf": [
             {
               "enum": [
```

blackboxsw (changes requested):
Minor spelling and avoidance of doubled periods/stops in rendered descriptions otherwise +1

blackboxsw (approved):
@holmanb just force pushed a squash commit with the minor spelling fixes for privileged. and multi-line. Looks good to me now.

setharnold (commented):
I really like the hints about the passwords and keys being visible to untrusted users on the system in some clouds. That's very important to share. It's probably too impolitic to share here which clouds are which, but suggestions on how users can tell would be helpful.

I have a few concerns about the passwd texts:

- the command to generate a hash is far too weak; the default number of rounds is 5000 and that's been known to be too weak for years: https://github.com/besser82/libxcrypt/commit/a80e94cc79b7980bc889f54f66a01a92b6e65298 -- I suggest raising `rounds` to `500000` (half a million), as that takes 0.2 s on my laptop. That's probably still too fast, but is vastly better. (The distribution switched to yescrypt a few years ago, which adds enough memory-hardness that it's harder to brute-force even when it's tuned to run pretty quickly. Once focal has dropped out of scope, consider changing this example to `yescrypt` instead.)
- the password can be used both for logging in and for sudo and so on. So the "There are many more secure options than using plain text passwords" is only partially true. While encouraging people to use ssh keys or certificates is a vastly better way to log in, once they are logged in and need to use sudo, they'll probably need a password. Maybe it bloats this text too much to try to describe the different needs for passwords, but it's probably worth a few small changes:
  -  point out that an exposed plain text password is immediately impactful; an exposed password hash isn't great but if the password hash is strong enough and the password was chosen with a high-entropy process it's not so bad.
  -  point out that ssh keys or certificates are better for logging in, and that if passwords are necessary, supplying a hashed password is the safer choice.

Thanks

holmanb:
Thanks @setharnold for the review! I just pushed updated wording in response to your comments.

TheRealFalcon (approved):

PR #4602, opened 2023-11-10
Proposal: unify service names

holmanb:
## Summary
User, developer, and maintainer confusion regarding the naming of cloud-init services makes reasoning and communicating about cloud-init stages unnecessarily confusing. This is a proposal to change cloud-init's stages to a clearer standard of names.

## Additional Context
There are lots of examples of confusion regarding current names. This change will require a downstream patch in Ubuntu and require distro changes for any distro which carries downstream packaging. I'm happy to help drive those externally too, to get to a cleaner commandline interface so we can spend less time explaining points of confusion in this awful stage naming convention to users and collaborators.

Please do note that the final say of whether this change lands ultimately rests in the hands of those that are much more familiar and comfortable with our current names than those who will benefit the most: our users, first time or infrequent contributors, or external projects trying to interact with cloud-init.

Docs, test, and manpage changes will not be implemented until feedback has been gathered regarding this change. A forum for discussion regarding this change is available [on the page of the public specification](https://discourse.ubuntu.com/t/proposed-specification-cloud-init-service-name-changes/40125) for this change.

## Merge type


TheRealFalcon:
@holmanb , I'm not really sure what to do with this PR. Conceptually, it's a great idea, but it also feels like a monumental change to just merge any time soon. Also, depending on how much of this gets patched out, that will be a fairly significant maintenance burden for existing distros for a fairly long time.

In  your linked spec, you mention:
> If this progresses, I intend to reach out to those communities that I am familiar with to notify them in advance of these required changes

Do you think it's worth reaching out to impacted maintainers first? We have the discourse spec, but I'm not sure very many people have actually seen it. At the very least, I think we need another discussion with the rest of the cloud-init upstream devs to determine if we think it is worth the maintenance from our end.

If that's not happening anytime soon, I'm thinking it might be better to close this PR. The branch still exists and can be viewed/modified as needed. Otherwise this branch will just sit here collecting merge conflicts and stale PR checks.

holmanb:
> If that's not happening anytime soon, I'm thinking it might be better to close this PR.

Fair enough, sounds good

PR #4604, opened 2023-11-13
azure/imds: remove limit for connection errors if route present

cjp256:
Drop the limit for connection errors if a route to IMDS is detected. Since we know we've brought up the primary NIC and connectivity should work, continue to retry until retry_deadline is up.

If there is no route to IMDS detected, such as Azure Stack, keep the current limit of 11.

Fetching reprovision data behavior is unchanged.


anhvoms (approved):
LGTM

blackboxsw (approved):
Ship it.

PR #4606, opened 2023-11-14
feat(subp): Measure subprocess command time

holmanb:
```
feat(subp): Measure subprocess command time
```


holmanb:
Didn't implement this in the same log that subp currently emits since it is useful to see that log before the command is run, for debugging in case cloud-init crashes or hangs.

TheRealFalcon (commented):
This results in a ton of logs like this:
```
2023-12-01 03:32:20,754 - subp.py[DEBUG]: command ['hostname', 'me'] took 0.0s to run
```
and also adds a fair amount of extra logs.

I think we should either change `%.3s` to `%.5s`, or only log if the time elapsed is >= .1

holmanb:
> This results in a ton of logs like this:

Hmm, that serves no one.

> only log if the time elapsed is >= .1

I prefer this way to avoid the extra logs.

PR #4607, opened 2023-11-14
Revert "net: allow dhcp6 configuration from generate_fallback_configu…

TheRealFalcon:
## Proposed Commit Message

```
Revert "net: allow dhcp6 configuration from generate_fallback_configu…
…ration() (#4474)"

This reverts commit 518047aea93d502aefae808921b11d0cfe2e7fd9.

This commit could cause issues in bringing up the network, especially
because in non-netplan NetworkManager environments, may-fail=false .
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/net/__init__.py
+++ b/cloudinit/net/__init__.py
@@ -571,12 +571,7 @@ def generate_fallback_config(config_driver=None):
         match = {
             "macaddress": read_sys_net_safe(target_name, "address").lower()
         }
-    cfg = {
-        "dhcp4": True,
-        "dhcp6": True,
-        "set-name": target_name,
-        "match": match,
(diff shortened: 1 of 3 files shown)


holmanb (approved):
Agreed, this is the right thing for now.

I agree with the original intent of this PR, but a future approach needs to be able to properly handle the following matrix:

1. ipv4-only, ipv6-only, and ipv6+ipv4
2. all currently supported renderers

Since the reverted approach appears to break an ipv4-only system with NetworkManager, and also an ipv6-only system with NetworkManager, reverting makes sense. Hopefully a future implementation will come with tests to validate behavior on at least a couple of different renderer-network configurations.

PR #4614, opened 2023-11-20
test: Add default hello package version

TheRealFalcon:
## Proposed Commit Message
Use message of commit


holmanb (approved):
I had an initial hesitation because this solution to the failing test is not the same as this recommendation to fix the failing test, but since the last couple of changes have had no update to version number, I agree with your change.

PR #4616, opened 2023-11-21
Actually validate or tell user we cannot

holmanb:
## Proposed Commit Message
Use commit message

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -942,7 +942,7 @@ def validate_cloudconfig_file(
             cloudconfig, schema, strict=True, log_deprecations=False
         ):
             print(
-                f"Skipping {schema_type}schema validation."
+                f"Skipping {schema_type} schema validation."
                 " Jsonschema dependency missing."
             )
             return False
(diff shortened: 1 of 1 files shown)


cloudinit/config/schema.py
```diff
@@ -937,9 +938,10 @@ def validate_cloudconfig_file(
             )
             return False
     try:
-        validate_cloudconfig_schema(
+        if not validate_cloudconfig_schema(
             cloudconfig, schema, strict=True, log_deprecations=False
-        )
+        ):
+            print("Cannot validate config, jsonschema dependency missing")
```
blackboxsw, line 944: Let's make this message include `Skipping schema validation` somehow and we should return False just like our network-config skip above as this function returns False when schema validation is not performed.
```suggestion
            print("Cannot validate config, jsonschema dependency missing")
            return False
```

blackboxsw (changes requested):

holmanb:
@blackboxsw Thanks, updated

cloudinit/config/schema.py
```diff
@@ -937,9 +938,14 @@ def validate_cloudconfig_file(
             )
             return False
     try:
-        validate_cloudconfig_schema(
+        if not validate_cloudconfig_schema(
             cloudconfig, schema, strict=True, log_deprecations=False
-        )
+        ):
+            print(
+                "Skipping network-config schema validation."
```
blackboxsw, line 945: Because we don't know if we are skipping network-config or cloud-config
```suggestion
                f"Skipping {schema_type}schema validation."
```

Issue #4613, opened 2023-11-20
Growpart integration test intermittently failing


TheRealFalcon:
I can reproduce this both on 23.3 and on a daily install, so this doesn't represent a regression in 23.4.

# Bug report
[test_growpart](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_growpart.py#L60) will intermittently fail locally and on jenkins.

Relevant logs from /var/log/cloud-init.log:
```
2023-11-20 20:56:58,924 - modules.py[DEBUG]: Running module growpart (<module 'cloudinit.config.cc_growpart' from '/usr/lib/pyt
hon3/dist-packages/cloudinit/config/cc_growpart.py'>) with frequency always
2023-11-20 20:56:58,924 - handlers.py[DEBUG]: start: init-network/config-growpart: running config-growpart with frequency alway
s
2023-11-20 20:56:58,924 - helpers.py[DEBUG]: Running config-growpart using lock (<cloudinit.helpers.DummyLock object at 0x7fa96
e863b50>)
2023-11-20 20:56:58,924 - subp.py[DEBUG]: Running command ['growpart', '--help'] with allowed return codes [0] (shell=False, ca
pture=True)
2023-11-20 20:56:58,927 - util.py[DEBUG]: Reading from /proc/517/mountinfo (quiet=False)
2023-11-20 20:56:58,928 - util.py[DEBUG]: Read 3422 bytes from /proc/517/mountinfo
2023-11-20 20:56:58,928 - util.py[DEBUG]: Reading from /sys/class/block/sda1/partition (quiet=False)
2023-11-20 20:56:58,928 - util.py[DEBUG]: Read 2 bytes from /sys/class/block/sda1/partition
2023-11-20 20:56:58,928 - util.py[DEBUG]: Reading from /sys/devices/pci0000:00/0000:00:01.1/0000:02:00.0/virtio6/host0/target0:
0:0/0:0:0:1/block/sda/dev (quiet=False)
2023-11-20 20:56:58,928 - util.py[DEBUG]: Read 4 bytes from /sys/devices/pci0000:00/0000:00:01.1/0000:02:00.0/virtio6/host0/tar
get0:0:0/0:0:0:1/block/sda/dev
2023-11-20 20:56:58,928 - util.py[DEBUG]: Reading from /proc/517/mountinfo (quiet=False)
[...]

PR #4618, opened 2023-11-21
Fix growpart race

holmanb:
## Proposed Commit Message
Fixes https://github.com/canonical/cloud-init/issues/4613

## Additional Context
See bug

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -19,7 +19,7 @@ from abc import ABC, abstractmethod
 from contextlib import suppress
 from pathlib import Path
 from textwrap import dedent
-from typing import Tuple
+from typing import Optional, Tuple
 
 from cloudinit import subp, temp_utils, util
 from cloudinit.cloud import Cloud
@@ -283,7 +283,7 @@ class ResizeGpart(Resizer):
         return (before, get_size(partdev))
 
 
-def get_size(filename):
+def get_size(filename) -> Optional[int]:
     fd = None
     try:
         fd = os.open(filename, os.O_RDONLY)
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_growpart.py
```diff
@@ -284,11 +284,15 @@ def resize(self, diskdev, partnum, partdev):
 
 
 def get_size(filename):
```
TheRealFalcon, line 286: Let's add an `Optional[int]` return type here since we're changing the return type

TheRealFalcon (commented):
2 minor comments inline.

Can we make sure this is unit tested?

cloudinit/config/cc_growpart.py
```diff
@@ -580,6 +584,15 @@ def resize_devices(resizer, devices):
                         "no change necessary (%s, %s)" % (disk, ptnum),
                     )
                 )
+            elif new is None or old is None:
+                info.append(
+                    (
+                        devent,
+                        RESIZE.CHANGED,
+                        "changed (%s, %s) size, new size is unknown"
+                        % (disk, ptnum, old),
```
TheRealFalcon, line 593: Now there's too many arguments here

holmanb:
Thanks @TheRealFalcon. This one fell off my radar - in part due to test complexity.

`resize_devices()` will require lots of mocks in order to get coverage for the log message logic - enough that I don't think that such a heavily mocked test provides much value. Refactoring that function would likely be required to make it more reasonable to write tests for, but I'm hoping to avoid that for now.

I've pushed a test for `get_size()`, let me know if that suffices for now.

TheRealFalcon:
Yeah, that one's a bit of a doozy. Given that we have some integration test coverage (that triggered this issue), I'm ok skipping that function for now.

TheRealFalcon (approved):

PR #4622, opened 2023-11-23
net/network_manager: do not set "may-fail" to False for both ipv4 and ipv6

ani-sinha:
Please see discussions in https://github.com/canonical/cloud-init/pull/4474 .

The second patch unreverts the revert https://github.com/TheRealFalcon/cloud-init/commit/21e6dfb8298835e93e5a11e6185574e611a1f44b .

The first change tries to the fix the issue due to which the above reversal was made. 

```
   If "may-fail" is set to False in the Network Manager keyfile for both ipv4
    and ipv6, it essentially means both ipv4 and ipv6 network initialization must
    succeed for the overall network configuration to succeed. This means, for
    environments where only ipv4 or ipv6 is available but not both and we need to
    configure both ipv4 and ipv6 options, the overall network configuration will
    fail. This is not what we want. When both ipv4 and ipv6 are configured,
    it is enough for the overall configuration to succeed if any one succeeds.
    Therefore, set "may-fail" to True for both ipv4 and ipv6 if and only if both
    ipv4 and ipv6 are configured in the Network Manager keyfile and "may-fail" is
    set to False for both. If both ipv4 and ipv6 are configured in the keyfile
    and if for any of them "may-fail" is already set to True, then do nothing.
    All other cases remain same as before.
```

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -71,6 +71,57 @@ class NMConnection:
         if not self.config.has_option(section, option):
             self.config[section][option] = value
 
+    def _config_option_is_set(self, section, option):
+        """
+        Checks if a config option is set. Returns True if it is,
+        else returns False.
+        """
+        return self.config.has_section(section) and self.config.has_option(
+            section, option
+        )
+
+    def _get_config_option(self, section, option):
+        """
+        Returns the value of a config option if its set,
+        else returns None.
+        """
+        if self._config_option_is_set(section, option):
+            return self.config[section][option]
+        else:
+            return None
+
+    def _change_set_config_option(self, section, option, value):
+        """
+        Overrides the value of a config option if its already set.
+        Else, if the config option is not set, it does nothing.
+        """
+        if self._config_option_is_set(section, option):
+            self.config[section][option] = value
+
+    def _set_mayfail_true_if_both_false_dhcp(self):
+        """
+        If for both ipv4 and ipv6, 'may-fail' is set to be False,
+        set it to True for both of them.
+        """
+        for family in ["ipv4", "ipv6"]:
+            if self._get_config_option(family, "may-fail") != "false":
+                # if either ipv4 or ipv6 sections are not set/configured,
+                # or if both are configured but for either ipv4 or ipv6,
+                # 'may-fail' is not 'false', do not do anything.
+                return
+            if self._get_config_option(family, "method") not in [
+                "dhcp",
+                "auto",
+            ]:
+                # if both v4 and v6 are not dhcp, do not do anything.
+                return
+
+        # If we landed here, it means both ipv4 and ipv6 are configured
+        # with dhcp/auto and both have 'may-fail' set to 'false'. So set
+        # both to 'true'.
+        for family in ["ipv4", "ipv6"]:
+            self._change_set_config_option(family, "may-fail", "true")
+
     def _set_ip_method(self, family, subnet_type):
         """
         Ensures there's appropriate [ipv4]/[ipv6] for given family
@@ -270,6 +321,14 @@ class NMConnection:
                 self._add_dns_search(family, subnet["dns_search"])
             if family == "ipv4" and "mtu" in subnet:
                 ipv4_mtu = subnet["mtu"]
+
+        # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
+        # at the at the same time. This will make the network configuration
+        # work only when both ipv4 and ipv6 dhcp succeeds. This may not be
+        # what we want. If we have configured both ipv4 and ipv6 dhcp, any one
+        # succeeding should be enough. Therefore, if "may-fail" is set to
+        # False for both ipv4 and ipv6 dhcp, set them both to True.
+        self._set_mayfail_true_if_both_false_dhcp()
 
         if ipv4_mtu is None:
             ipv4_mtu = device_mtu
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -1477,11 +1477,11 @@ NETWORK_CONFIGS = {
 
                 [ipv4]
                 method=auto
-                may-fail=false
+                may-fail=true
 
                 [ipv6]
                 method=auto
-                may-fail=false
+                may-fail=true
 
                 """
             ),
@@ -1650,11 +1650,11 @@ NETWORK_CONFIGS = {
 
                 [ipv6]
                 method=auto
-                may-fail=false
+                may-fail=true
 
                 [ipv4]
                 method=auto
-                may-fail=false
+                may-fail=true
 
                 """
             ),
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -4339,6 +4339,7 @@ class TestGenerateFallbackConfig(CiTestCase):
             "ethernets": {
                 "eth0": {
                     "dhcp4": True,
+                    "dhcp6": True,
                     "set-name": "eth0",
                     "match": {
                         "macaddress": "00:11:22:33:44:55",
@@ -4423,6 +4424,9 @@ iface lo inet loopback
 
 auto eth0
 iface eth0 inet dhcp
+
+# control-alias eth0
+iface eth0 inet6 dhcp
 """
         self.assertEqual(expected.lstrip(), contents.lstrip())
 
@@ -4512,6 +4516,9 @@ iface lo inet loopback
 
 auto eth1
 iface eth1 inet dhcp
+
+# control-alias eth1
+iface eth1 inet6 dhcp
 """
         self.assertEqual(expected.lstrip(), contents.lstrip())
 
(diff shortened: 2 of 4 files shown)


tests/unittests/test_net.py
```diff
@@ -3537,15 +3537,15 @@
 
                 [ipv4]
                 method=manual
-                may-fail=false
+                may-fail=true
```
TheRealFalcon, line 3540: This is just one example, but is this really the change you want? Brett's proposal was to set `may-fail=true` when dhcp4 and dhcp6 are both set. This accomplishes that, but also will set `may-fail=true` anytime any configuration contains both ipv4 and ipv6 sections regardless of dhcp status.

I just want to double check that you're ok with this broader change.

ani-sinha: yes you are right. We need to be as narrow as possible with our fix so that we do not break anything else. I have sent a re-work of the patch. Please let me know how it looks.

holmanb:
> The second patch unreverts

Please note that since we don't want to introduce new commits that put cloud-init in a broken state, so this will be squash merged.

ani-sinha:
> > The second patch unreverts
> 
> Please note that since we don't want to introduce new commits that put cloud-init in a broken state, so this will be squash merged if/when ready.

That is precisely why I arranged the commits in such a way that first the fix is applied and then the revert of the revert of the original commit so that once we re-apply the original commit, at no stage cloud-init is broken.

ani-sinha:
> > first the fix is applied and then the revert of the revert of the original commit so that once we re-apply the original commit, at no stage cloud-init is broken.
> 
> Okay I see now what you are trying to do. However, I still don't think we want git history to look the way that you've proposed this.
> 
> Reasons:
> 
> 1. "Unrevert a reverted commit" -> This describes "what happened", not "what does this code do". This adds a layer of indirection in the commit message that we don't want.
> 2. "Unrevert a reverted commit" -> This claim isn't actually true. [560dede](https://github.com/canonical/cloud-init/commit/560dede84961d2dd79a42f0984ad3bbd1d58facd) and [518047a](https://github.com/canonical/cloud-init/commit/518047aea93d502aefae808921b11d0cfe2e7fd9) are different.
> 
> I think I'm fine with keeping these commits separate if the commit message gets fixed, since "do the right thing for NetworkManager in all cases of dhcp4+dhcp6" is a fundamentally different thing than "change the default configuration to include dhcp6".

Ok I have tried to update the commit message to something that is more appropriate.

cloudinit/net/network_manager.py
```diff
@@ -116,6 +170,14 @@ def _set_ip_method(self, family, subnet_type):
         self.config[family]["method"] = method
         self._set_default(family, "may-fail", "false")
 
+        # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
+        # at the at the same time. This will make the network configuration
+        # work only when both ipv4 and ipv6 dhcp succeeds. This may not be
+        # what we want. If we have configured both ipv4 and ipv6 dhcp, any one
+        # succeeding should be enough. Therefore, if "may-fail" is set to
+        # False for both ipv4 and ipv6 dhcp, set them both to True.
+        self._set_mayfail_true_if_both_false_dhcp()
```
TheRealFalcon, line 179: I don't think this is the right spot for this as we'll run it for every subnet.

I included a patch to this review that moves it to after we configure the subnets.

cloudinit/net/network_manager.py
```diff
@@ -71,6 +71,60 @@ def _set_default(self, section, option, value):
...
+        """
+        if self._config_option_is_set(section, option):
+            self.config[section][option] = value
+
+    def _set_mayfail_true_if_both_false_dhcp(self):
+        """
+        If for both ipv4 and ipv6, 'may-fail' is set to be False,
+        set it to True for both of them.
+        """
+        for family in ["ipv4", "ipv6"]:
```
TheRealFalcon, line 106: Can this be simplified to:
```python
        for family in ["ipv4", "ipv6"]:
            if self._get_config_option(family, "method") not in [
                "dhcp",
                "auto",
            ]:
                # if both v4 and v6 are not dhcp, do not do anything.
                return
```

We don't really care about the `may-fail` value, correct? Rather we care if both methods are using DHCP.

TheRealFalcon (changes requested):
Thanks for the update. This works, but I requested some minor changes inline.

Here's a patch containing what I suggested:
```diff
diff --git a/cloudinit/net/network_manager.py b/cloudinit/net/network_manager.py
index 589ab7368..237b440c8 100644
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -98,24 +98,16 @@ class NMConnection:
         if self._config_option_is_set(section, option):
             self.config[section][option] = value
 
-    def _set_mayfail_true_if_both_false_dhcp(self):
+    def _set_mayfail_true_if_both_dhcp(self):
         """
-        If for both ipv4 and ipv6, 'may-fail' is set to be False,
-        set it to True for both of them.
+        If for both ipv4 and ipv6, method is 'dhcp' or 'auto', set
+        'may-fail' to 'true'.
         """
         for family in ["ipv4", "ipv6"]:
-            if not self._config_option_is_set(family, "may-fail"):
-                # if either ipv4 or ipv6 sections are not set/configured,
-                # do not do anything.
-                return
-            if self._get_config_option(family, "may-fail") == "true":
-                # if for either ipv4 or ipv6, 'may-fail' is 'true', do
-                # not do anything.
-                return
-            if (
-                self._get_config_option(family, "method") != "auto"
-                and self._get_config_option(family, "method") != "dhcp"
-            ):
+            if self._get_config_option(family, "method") not in [
+                "dhcp",
+                "auto",
+            ]:
                 # if both v4 and v6 are not dhcp, do not do anything.
                 return
 
@@ -130,7 +122,6 @@ class NMConnection:
         Ensures there's appropriate [ipv4]/[ipv6] for given family
         appropriate for given configuration type
         """
-
         method_map = {
             "static": "manual",
             "static6": "manual",
@@ -170,14 +161,6 @@ class NMConnection:
         self.config[family]["method"] = method
         self._set_default(family, "may-fail", "false")
 
-        # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
-        # at the at the same time. This will make the network configuration
-        # work only when both ipv4 and ipv6 dhcp succeeds. This may not be
-        # what we want. If we have configured both ipv4 and ipv6 dhcp, any one
-        # succeeding should be enough. Therefore, if "may-fail" is set to
-        # False for both ipv4 and ipv6 dhcp, set them both to True.
-        self._set_mayfail_true_if_both_false_dhcp()
-
     def _add_numbered(self, section, key_prefix, value):
         """
         Adds a numbered property, such as address<n> or route<n>, ensuring
@@ -333,6 +316,14 @@ class NMConnection:
             if family == "ipv4" and "mtu" in subnet:
                 ipv4_mtu = subnet["mtu"]
 
+        # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
+        # at the at the same time. This will make the network configuration
+        # work only when both ipv4 and ipv6 dhcp succeeds. This may not be
+        # what we want. If we have configured both ipv4 and ipv6 dhcp, any one
+        # succeeding should be enough. Therefore, if "may-fail" is set to
+        # False for both ipv4 and ipv6 dhcp, set them both to True.
+        self._set_mayfail_true_if_both_dhcp()
+
         if ipv4_mtu is None:
             ipv4_mtu = device_mtu
         if not ipv4_mtu == device_mtu:

```

Let me know if that works for you.

cloudinit/net/network_manager.py
ani-sinha: > Can this be simplified to:
> 
> ```python
>         for family in ["ipv4", "ipv6"]:
>             if self._get_config_option(family, "method") not in [
>                 "dhcp",
>                 "auto",
>             ]:
>                 # if both v4 and v6 are not dhcp, do not do anything.
>                 return
> ```
> 
> We don't really care about the `may-fail` value, correct? Rather we care if both methods are using DHCP.

We do care about may-fail. If both methods are dhcp and in both cases may-fail is False then we would want to make may-fail as True. So both checks are needed.

cloudinit/net/network_manager.py
ani-sinha: Yes good idea.

cloudinit/net/network_manager.py
ani-sinha: I understand that currently the way the code is written, "may-fail" is always set to False for dhcp but I want to have an explicit check to make it future proof and to show specific intention.

cloudinit/net/network_manager.py
```diff
@@ -71,6 +71,57 @@ def _set_default(self, section, option, value):
...
+            if self._get_config_option(family, "may-fail") != "false":
```
ani-sinha, line 107: I agree that this check is currently not required. But I still want to have this explicitly spelled out.

TheRealFalcon (approved):
Thanks @ani-sinha , this looks good to me.

@holmanb , since you proposed the idea, do you mind reviewing as well?

holmanb (approved):
Looks good, @ani-sinha.

Thanks for your work on this.

PR #4624, opened 2023-11-23
Improve spelling checker

holmanb:
## Additional Context
Our current spellchecker is hacked together in a Makefile.
The spellcheck fixer is rarely used and commits odious shell crimes (I wrote it).

Replace the checker with a sphinx plugin. Do not replace the fixer.
Include an exact match allow-list (built from the output of "misspelled" words).
Include a regex match allow-list.
Fix misspelled words.

The allowlists would be onerous to create from scratch. Thankfully a oneliner on the output of "misspelled words" got us most of the way there:
```
grep 'Spell check' tmp_GB | sed 's/.*Spell check\: //g'| awk -F\: '{print $1}' | tr '[A-Z]' '[a-z]' | sed 's/:$//g' | sort -u | grep -v 'doc/rtd/' > doc/rtd/spelling_word_list.txt
```

A bit of regex made matching the rest easier.

## Test Steps
```
# introduce a spelling error
tox -e doc-spelling
```

## Merge type


doc/rtd/spelling_wordlist.txt
TheRealFalcon, line 1: Can this file be deleted?

doc/rtd/spelling.py
```diff
@@ -0,0 +1,79 @@
...
+                    previous_item = item_a
+            unordered = (
+                f"[..., {previous_item}, {first_missordered}, "
+                f"{next_item}, ...]"
+            )
+            raise Exception(
+                f"Unsorted {self.word_list} detected. "
+                f"Please sort for legibility. Unordered list: {unordered}"
+            )
+        for line in lines:
```
TheRealFalcon, line 50: unless it's the first or last word, we're never going to get here because an uppercase letter will sort differently and get triggered by the unsorted detection above (but print an unhelpful `unordered` list).

Also, if you have the full text, something like `if text != text.lower()` would be a simpler check.

doc/rtd/spelling.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        super().__init__(*args, *kwargs)
+        directory = pathlib.Path(__file__).parent
+        with open(directory.joinpath(self.word_list)) as f:
+            lines = f.read().splitlines()
+            self._validate_lines(lines)
+            self.word_set = set(lines)
+        with open(directory.joinpath(self.regex_list)) as f:
+            self.regex_set = set(f.read().splitlines())
+        print(f"Loaded {self.word_list}: {self.word_set}")
```
TheRealFalcon, line 27: Sets are unorderered, so this output isn't super helpful.

doc/rtd/conf.py
```diff
@@ -36,8 +36,17 @@
...
 ]
 
+
+# Spelling settings for sphinxcontrib.spelling
+# https://docs.ubuntu.com/styleguide/en/
+spelling_warning = True
+
+# Uses case-independent spelling matches from doc/rtd/spelling_word_list.txt
+spelling_filters = ["spelling.WordListFilter"]
+
```
TheRealFalcon, line 49: Should we also include a `spelling_lang='en_GB'`?

doc/rtd/spelling_word_list.txt
```diff
@@ -0,0 +1,294 @@
+akamai
```
TheRealFalcon, line 1: I see a number of US spellings in this list. E.g.,artifact, honored, flavors. Shouldn't we be disallowing these words?

TheRealFalcon (changes requested):
Nice update! I have a few inline comments.

Due to sphinx version, this is python3.8+ . I don't think that's a problem but just wanted to highlight it.

doc/rtd/spelling.py
holmanb: > we're never going to get here because an uppercase letter will sort differently and get triggered by the unsorted detection above (but print an unhelpful unordered list).

any word that begins with a lower case letter breaks this assumption, and tech has more than a couple of those (camelCase, iPod etc)

holmanb:
Rebased on main to fix merge conflict.

doc/rtd/howto/predeploy_testing.rst
```diff
@@ -0,0 +1,141 @@
+.. _predeploy_testing:
```
TheRealFalcon, line 1: Did this file accidentally get added on rebase?

TheRealFalcon:
@holmanb , did you see my more recent comment about the casing?

doc/rtd/spelling.py
holmanb: Ahhh, I see. Yes good point, ordered first would be better.

TheRealFalcon (changes requested):
Change LGTM, but now there's a rebase issue:
```
Warning, treated as error:
cloudinit/config/cc_users_groups.py:docstring of cloudinit.config.cc_users_groups:114: : Spell check: SHA:  (string) Hash of user password applied when user does not exist. This will NOT be applied if the user already exists. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=500000 .
```

`sha` needs to be added to the word list

s-makin:
> Change LGTM, but now there's a rebase issue:
> 
> ```
> Warning, treated as error:
> cloudinit/config/cc_users_groups.py:docstring of cloudinit.config.cc_users_groups:114: : Spell check: SHA:  (string) Hash of user password applied when user does not exist. This will NOT be applied if the user already exists. To generate this hash, run: mkpasswd --method=SHA-512 --rounds=500000 .
> ```
> 
> `sha` needs to be added to the word list

I'd prefer that this piece of code be marked out as code instead, if that's possible. Code is excluded specifically (so it can either be in a code block, or in backticks)

TheRealFalcon (approved):
LGTM!

PR #4629, opened 2023-11-27
SC-1603-jinja-enhancements

a-dubs:
## Proposed Commit Message

```
feat(jinja): better jinja feedback and error catching

When an error is encountered while trying to parse a jinja template, the traceback is automatically parsed to get the specific error and the line number that caused this error, and then this information is turned into a user-friendly single-line error message. 

The cloud-init runtime will log these error messages when jinja syntax errors are encountered in the cloud-config file. For the query, render, and schema commands, the nasty traceback is caught and a single line error message is given to the user instead.

Fixes GH-4360
```

## PR Checklist:

### DIFF
--- a/cloudinit/templater.py
+++ b/cloudinit/templater.py
@@ -19,6 +19,8 @@ import sys
 import traceback
 from typing import Any
 
+from jinja2 import TemplateSyntaxError
+
 from cloudinit import type_utils as tu
 from cloudinit import util
 from cloudinit.atomic_helper import write_file
@@ -45,7 +47,7 @@ BASIC_MATCHER = re.compile(r"\$\{([A-Za-z0-9_.]+)\}|\$([A-Za-z0-9_.]+)")
 MISSING_JINJA_PREFIX = "CI_MISSING_JINJA_VAR/"
 
 
-class CustomParsedJinjaException(Exception):
+class JinjaSyntaxParsingException(Exception):
     pass
 
 
@@ -121,7 +123,7 @@ def detect_template(text):
             )
             return result
 
-        except Exception as JTemplateError:
+        except Exception as jtemplate_error:
             # if the error is a syntax one, we know this next line will fail
             # we want to get the traceback and error message it throws
             try:
@@ -135,38 +137,43 @@ def detect_template(text):
                 jinja_env.parse(content)
                 # if we get here, then the previous line didn't fail,
                 # so we just need to re-raise the original error
-                raise JTemplateError
-            except Exception as useful_jinja_error:
+                raise jtemplate_error
+            except TemplateSyntaxError as template_syntax_error:
                 try:
                     tb = "".join(
-                        traceback.format_tb(useful_jinja_error.__traceback__)
+                        traceback.format_tb(
+                            template_syntax_error.__traceback__
+                        )
                     )
                     line_number_matches = re.findall(
                         r'File "<unknown>", line (\d+)', tb
                     )
                     line_number_of_error = int(line_number_matches[0])
+                    # adjust line number to adjust for the jinja header having been
+                    # removed from the content
+                    if (
+                        len(
+                            re.findall(
+                                "##( )?template:( )?jinja",
+                                content.split("\n")[0].strip().lower(),
+                            )
+                        )
+                        == 0
+                    ):
+                        line_number_of_error += 1
+                    raise JinjaSyntaxParsingException(
+                        "{e} on line {line_no}".format(
+                            e=template_syntax_error,
+                            line_no=line_number_of_error,
+                        )
+                    ) from jtemplate_error
+
                 # if we couldn't parse the traceback for some reason,
                 # just re-raise the original error
                 except Exception:
-                    raise JTemplateError from useful_jinja_error
-                # adjust line number to adjust for the jinja header having been
-                # removed from the content
-                if (
-                    len(
-                        re.findall(
-                            "##( )?template:( )?jinja",
-                            content.split("\n")[0].strip().lower(),
-                        )
-                    )
-                    == 0
-                ):
-                    line_number_of_error += 1
-                raise CustomParsedJinjaException(
-                    "{e} on line {line_no}".format(
-                        e=useful_jinja_error,
-                        line_no=line_number_of_error,
-                    )
-                ) from useful_jinja_error
+                    raise jtemplate_error
+            except Exception:
+                raise jtemplate_error
 
     if text.find("\n") != -1:
         ident, rest = text.split("\n", 1)  # remove the first line
--- a/cloudinit/templater.py
+++ b/cloudinit/templater.py
@@ -119,8 +119,9 @@ def detect_template(text):
             return result
 
         except Exception as JTemplateError:
-            # print(JTemplateError)
-            try:  # we know this next line will fail, but we want to get the traceback and error message it throws
+            # if the error is a syntax one, we know this next line will fail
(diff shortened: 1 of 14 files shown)


tests/unittests/cmd/test_query.py
```diff
@@ -567,3 +567,69 @@ def test_handle_args_list_keys_errors_when_varname_is_not_a_dict(
...
+            user_data="ud",
+            vendor_data="vd",
+            varname=None,
+        )
+        with mock.patch("os.getuid") as m_getuid:
+            m_getuid.return_value = 100
+            assert 1 == query.handle_args("anyname", args)
+        out, err = capsys.readouterr()
+        assert expected_error in caplog.text
+        assert expected_error in err
```
blackboxsw, line 635: @a-dubs I think we determined in our discussion today that this unittest shouldn't be checking stderr setup because the query module no longer invokes `setup_basic_logging` directly to add a stderr stream handler to logging. This is now done up in [cloudinit.cmd.main before calling this submodule gets invoked](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/main.py#L1064)

Since the logging setup is outside of this scope of this unit, I'd suggest we have a supplemental test that asserts that main does the proper setup_basic_logging on all subcommands not in `('modules', 'init')`.  We already have [separate unittest coverage that `setup_basic_logging` properly sets up a stderr handler here](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_log.py#L91). So no need to add extra stderr validation if we minimally assert when `setup_basic_logging` is called below.

```diff
--- a/tests/unittests/test_cli.py
+++ b/tests/unittests/test_cli.py
@@ -2,6 +2,7 @@
 
 import contextlib
 import io
+import logging
 import os
 from collections import namedtuple
 
@@ -172,6 +173,45 @@ class TestCLI:
         for subcommand in expected_subcommands:
             assert subcommand in err
 
+    @pytest.mark.parametrize(
+        "subcommand,log_to_stderr,mocks",
+        (
+            ("init", False, [mock.patch("cloudinit.cmd.main.status_wrapper")]),
+            (
+                "modules",
+                False,
+                [mock.patch("cloudinit.cmd.main.status_wrapper")],
+            ),
+            (
+                "schema",
+                True,
+                [
+                    mock.patch(
+                        "cloudinit.stages.Init._read_cfg", return_value={}
+                    ),
+                    mock.patch("cloudinit.config.schema.handle_schema_args"),
+                ],
+            ),
+        ),
+    )
+    @mock.patch("cloudinit.cmd.main.setup_basic_logging")
+    def test_subcommands_log_to_stderr_via_setup_basic_logging(
+        self, setup_basic_logging, subcommand, log_to_stderr, mocks
+    ):
+        """setup_basic_logging is called for modules to use stderr
+
+        Subcommands with exception of 'init'  and 'modules' use
+        setup_basic_logging to direct logged errors to stderr.
+        """
+        with contextlib.ExitStack() as mockstack:
+            for mymock in mocks:
+                mockstack.enter_context(mymock)
+            self._call_main(["cloud-init", subcommand])
+        if log_to_stderr:
+            setup_basic_logging.assert_called_once_with(logging.WARNING)
+        else:
+            setup_basic_logging.assert_not_called()
+
     @pytest.mark.parametrize("subcommand", ["init", "modules"])
     @mock.patch("cloudinit.cmd.main.status_wrapper")
     def test_modules_subcommand_parser(self, m_status_wrapper, subcommand):
```

a-dubs: Patch has been applied. Thank you!

tests/unittests/config/test_schema.py
```diff
@@ -826,6 +827,41 @@ def test_validateconfig_file_no_cloud_cfg(
...
+
+        _out, err = capsys.readouterr()
+        expected = (
+            "Error:\n"
+            "Failed to render templated user-data. "
+            + JinjaSyntaxParsingException.message_template.format(
+                syntax_error="unexpected '}'",
+                line_no=3,
+                line_content="c:{{ d } }",
+            )
```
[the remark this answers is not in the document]
a-dubs: I added a test to `tests/unittests/test_templating.py` which explicitly tests that the formatting and creation of the
error message string is as expected. This way, we can reuse the naming template throughout various unit tests which will
significantly help reduce tech debt.

cloudinit/templater.py
```diff
@@ -42,6 +46,13 @@
 MISSING_JINJA_PREFIX = "CI_MISSING_JINJA_VAR/"
 
 
+class JinjaSyntaxParsingException(Exception):
+    message_template = (
```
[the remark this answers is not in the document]
a-dubs: Excellent suggestion. Thank you so much for such a well thought out improvement. Let me know if the changes I implemented are inline with what you were hoping for.

a-dubs:
This is ready for a re-review now. 

@blackboxsw I did my best to take your suggested changes and modify them as I saw fit and not just blindly patch them in. Let me know if the changes I made are inline with what you were hoping to see implemented.

cloudinit/templater.py
```diff
@@ -42,6 +44,67 @@
...
+        template = JinjaSyntaxParsingException.message_template
+        if not line_content:
+            template = template.rsplit(":", 1)[
+                0
+            ]  # remove line_content portion
+            return template.format(
+                syntax_error=syntax_error,
+                line_number=line_number,
+            )
+        return template.format(
```
blackboxsw, line 96: Up here we can set line_content to the appropriate suffix if present or empty string if absent. Then we don't need to manipulate a copy ofthe `message_format` attribute.
```suggestion
        if line_content:
            line_conent = f": {line_content}"
        return template.format(
```

cloudinit/templater.py
```diff
@@ -101,19 +164,30 @@ def replacer(match):
 def detect_template(text):
     def jinja_render(content, params):
         # keep_trailing_newline is in jinja2 2.7+, not 2.6
+        content = "##template:jinja\n" + content
```
blackboxsw, line 167: Why are we adding this cloud-init header back into the content passed to JTemplate only to remove it again upon return of JTemplate. jinja2.Template shouldn't care about this cloud-init header comments remotely. Is this to try to establish proper line posisitons??

blackboxsw (changes requested):
A couple of other nits and questions. Thanks @a-dubs . I'll wrap up the review today.

cloudinit/templater.py
a-dubs: Yes it is to establish proper line positions upon failure. But now that I think about it, without even checking if the
header is even there in the first place, this doesn't actually add any flexibility. I'll just revert to how i had it
previously (just always adding 1 to the line number that the original jinja error returns.)

cloudinit/templater.py
```diff
@@ -101,19 +164,30 @@ def replacer(match):
...
+                    extensions=["jinja2.ext.do"],
+                ).render(**params)
+                + add
+            ).replace("##template:jinja\n", "")
+
+        except TemplateSyntaxError as template_syntax_error:
+            # template_syntax_error.lineno += 1
+            raise JinjaSyntaxParsingException(
+                error=template_syntax_error,
+                source=content,
```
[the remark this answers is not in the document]
a-dubs: Good catch!

blackboxsw (approved):
Thanks @a-dubs for the cleanups here. This looks good and shows simplier errors for all of our subcommands when invalid jinja templates are provided.

PR #4631, opened 2023-11-27
schema: non-root fallback to default cfg paths on perm errors

blackboxsw:
## Proposed Commit Message
```
When non-root user calls cloud-init schema to validate cfg files, they will get a permissions error on /var/lib/cloud/instance dir trying to source the cached datasource. Fallback to default read_cfg_paths in this case and log debug level msg about this case.

Since cloud-init schema CLI is used in the context of validating config files and not really searching for cfg path overrides, using default paths to find instance-data.json isn't critical or even a warning condition.

Additionally avoid print() and use LOG.warning when no datasource currently detected as this should indicate that scripts or user may be trying to read instance-data.json which does not exist yet.

Fixes GH-4620
```

## Test Steps

```
lxc launch ubuntu-daily:mantic test-m
lxc exec test-m -- cloud-iniit status --wait
lxc exec tests-m -- add-apt-repository ppa:cloud-init-dev/daily -y
lxc exec tests-m -- apt install cloud-init -y

lxc exec tests-m
su - ubuntu
cloud-init schema -c /usr/share/doc/cloud-init/examples/seed/user-data
...

PermissionError: [Errno 13] Permission denied: '/var/lib/cloud/instance'
```

## Checklist

## Merge type

### DIFF
--- a/tests/unittests/config/test_schema.py
+++ b/tests/unittests/config/test_schema.py
@@ -1,6 +1,5 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
-
 import importlib
 import inspect
 import itertools
@@ -12,6 +11,7 @@ import sys
 import unittest
 from collections import namedtuple
 from copy import deepcopy
+from errno import EACCES
 from pathlib import Path
 from textwrap import dedent
 from types import ModuleType
@@ -39,6 +39,7 @@ from cloudinit.config.schema import (
 from cloudinit.distros import OSFAMILIES
 from cloudinit.safeyaml import load, load_with_marks
 from cloudinit.settings import FREQUENCIES
+from cloudinit.sources import DataSourceNotFoundException
 from cloudinit.util import load_file, write_file
 from tests.hypothesis import given
 from tests.hypothesis_jsonschema import from_schema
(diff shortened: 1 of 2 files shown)


tests/unittests/config/test_schema.py
```diff
@@ -2165,6 +2166,61 @@ class TestHandleSchemaArgs:
...
+        args = self.Args(
+            config_file=str(user_data_fn),
+            schema_type="cloud-config",
+            annotate=False,
+            docs=None,
+            system=None,
+            instance_data=None,
+        )
+        handle_schema_args("unused", args)
+        out, _err = capsys.readouterr()
```
holmanb, line 2218: nit: `capsys.readouterr()` returns a namedtuple with two attributes, `out` and `err`, so this could be prettier / less boilerplate, but functionally equivalent without this line:
```python
        handle_schema_args("unused", args)
        for expected in expected_out:
            assert expected in capsys.readouterr().out
```

holmanb (approved):
One nit, feel free to take it or leave it.

LGTM!

tests/unittests/config/test_schema.py
blackboxsw: yes given that we aren't providing more than one item in the expected_out list, I'll adapt this code to make the direct call to `capsys.readouterr.out` I'll also just drop the parameterization of expected_out too since we are validating the same hard-coded string for both anyway.
thx

PR #4634, opened 2023-11-28
refactor: remove dependency on netifaces

catmsred:
netifaces is no longer being maintained and is only used by the VMWare data source.  As such this commit replaces the calls to netifaces with cloudinit's native netinfo.

## Proposed Commit Message

```
refactor: remove dependency on netifaces

netifaces is no longer being maintained and is only used by the VMWare data source.  As such this commit replaces the calls to netifaces with cloudinit's native netinfo.
```

## Additional Context

This is related to https://github.com/canonical/cloud-init/pull/953.  I have tagged is as WIP at this time because I would like to do more testing on VMWare directly to ensure the updated code works in an environment without mocks.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -774,34 +774,44 @@ def get_default_ip_addrs():
     addresses associated with the device used by the default route for a given
     address.
     """
+
+    # Get ipv4 and ipv6 interfaces associated with default routes
+    ipv4_if = None
+    ipv6_if = None
+    routes = netinfo.route_info()
+    for route in routes["ipv4"]:
+        if route["destination"] == "0.0.0.0":
+            ipv4_if = route["iface"]
+    for route in routes["ipv6"]:
+        if route["destination"] == "::/0":
+            ipv6_if = route["iface"]
+
+    # Get ip address associated with default interface
     ipv4 = None
     ipv6 = None
-    routes = netinfo.route_info()
-    for route in routes:
-        if "G" in route["flags"]:
-            if ipv4 and ipv4 != route["gateway"]:
-                LOG.debug(
-                    "multiple ipv4 gateways: %s, %s", ipv4, route["gateway"]
-                )
-            ipv4 = route["gateway"]
-        elif "UG" in route["flags"]:
-            if ipv6 and ipv6 != route["gateway"]:
-                LOG.debug(
-                    "multiple ipv6 gateways: %s, %s", ipv6, route["gateway"]
-                )
-            ipv6 = route["gateway"]
-
-    # If there is an IPv4 gateway and an IPv6 gateway, return immediately to
-    # avoid extra work
-    if ipv4 and ipv6:
-        return ipv4, ipv4
-
     netdev = netinfo.netdev_info()
+    if ipv4_if in netdev:
+        addrs = netdev[ipv4_if]["ipv4"]
+        if len(addrs) > 1:
+            LOG.debug(
+                "device %s has more than one ipv4 address: %s", ipv4_if, addrs
+            )
+        elif len(addrs) == 1 and "ip" in addrs[0]:
+            ipv4 = addrs[0]["ip"]
+    if ipv6_if in netdev:
+        addrs = netdev[ipv6_if]["ipv6"]
+        if len(addrs) > 1:
+            LOG.debug(
+                "device %s has more than one ipv6 address: %s", ipv6_if, addrs
+            )
+        elif len(addrs) == 1 and "ip" in addrs[0]:
+            ipv6 = addrs[0]["ip"]
+
     # If there is a default IPv4 address but not IPv6, then see if there is a
     # single IPv6 address associated with the same device associated with the
     # default IPv4 address.
     if ipv4 is not None and ipv6 is None:
-        for dev_name in netdev.keys():
+        for dev_name in netdev:
             for addr in netdev[dev_name]["ipv4"]:
                 if addr["ip"] == ipv4 and len(netdev[dev_name]["ipv6"]) == 1:
                     ipv6 = netdev[dev_name]["ipv6"][0]["ip"]
@@ -811,10 +821,10 @@ def get_default_ip_addrs():
     # single IPv4 address associated with the same device associated with the
     # default IPv6 address.
     if ipv4 is None and ipv6 is not None:
-        for dev_name in netdev.keys():
+        for dev_name in netdev:
             for addr in netdev[dev_name]["ipv6"]:
                 if addr["ip"] == ipv6 and len(netdev[dev_name]["ipv4"]) == 1:
-                    ipv4 = netdev[dev_name]["ipv6"][0]["ip"]
+                    ipv4 = netdev[dev_name]["ipv4"][0]["ip"]
                     break
 
     return ipv4, ipv6
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -737,9 +737,11 @@ def get_default_ip_addrs():
     for route in routes["ipv4"]:
         if route["destination"] == "0.0.0.0":
             ipv4_if = route["iface"]
+            break
     for route in routes["ipv6"]:
         if route["destination"] == "::/0":
             ipv6_if = route["iface"]
+            break
 
     # Get ip address associated with default interface
     ipv4 = None
--- a/cloudinit/sources/DataSourceVMware.py
+++ b/cloudinit/sources/DataSourceVMware.py
@@ -16,51 +16,6 @@ multiple transports types, including:
     * EnvVars
     * GuestInfo
     * IMC (Guest Customization)
-
-Netifaces (https://github.com/al45tair/netifaces)
-
-    Please note this module relies on the netifaces project to introspect the
-    runtime, network configuration of the host on which this datasource is
-    running. This is in contrast to the rest of cloud-init which uses the
-    cloudinit/netinfo module.
-
-    The reasons for using netifaces include:
-
-        * Netifaces is built in C and is more portable across multiple systems
-          and more deterministic than shell exec'ing local network commands and
-          parsing their output.
-
-        * Netifaces provides a stable way to determine the view of the host's
-          network after DHCP has brought the network online. Unlike most other
-          datasources, this datasource still provides support for JINJA queries
-          based on networking information even when the network is based on a
-          DHCP lease. While this does not tie this datasource directly to
-          netifaces, it does mean the ability to consistently obtain the
-          correct information is paramount.
-
-        * It is currently possible to execute this datasource on macOS
-          (which many developers use today) to print the output of the
-          get_host_info function. This function calls netifaces to obtain
-          the same runtime network configuration that the datasource would
-          persist to the local system's instance data.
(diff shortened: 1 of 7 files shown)


catmsred:
@akutz @PengpengSun This PR proposes replacing VMWare's dependence on netifaces (which is no longer being actively maintained) with cloud-init's netinfo.  I know there was a previous thread that discussed this and there is an extensive comment at the top of `cloudinit/sources/DataSourceVMware.py` explaining VMWare's preference for netifaces.  There were three main points listed as follows:

- Netifaces is built in C and is more portable across multiple system and more deterministic than shell exec'ing local network commands and parsing their output.  **Is this still true with netifaces no longer being maintained?**
- Netifaces provides a stable way to determine the view of the host's network after DHCP has brought the network online.  **Can you provide more clarification on the specific testing you did that gave you inconsistent data with netinfo?**
- It is currently possible to execute this datasource on macOS (which many developers use today) to print the output of the get_host_info function.  **The commands executed by netdev_info are fully functional on my Mac environment; is this still a concern?**

If you have some time could you validate that this PR works as you expect in a VMWare environment?  I have updated the unit tests but obviously that uses mocks for networking set up and may not hit all the points that concerned you when you originally introduced netifaces into the code base.

Thanks!

holmanb (commented):
Thanks for your effort on this @catmsred!

Dropping this dependency will help reduce cloud image size requirements due to cloud-init and standardize networking code paths across datasources. This gives us one less datasource variance to have to maintain and reason about.

I know that this is still WIP, but there are few more items that we will want to take care of as part of this change. I know that you're probably waiting on feedback from @akutz and probably testing feedback before doing more work on this.

You mentioned removing the comment at the top of `cloudinit/sources/DataSourceVMware.py` already when we chatted earlier today.

Here are some other pointers to required changes in case it helps:

#### main branch updates:

pyproject.toml
tox.ini
requirements.txt
tools/build-on-netbsd
tools/build-on-openbsd

#### ubuntu packaging: update debian/control in the following branches
 - ubuntu/devel
 - ubuntu/mantic
 - ubuntu/lunar
 - ubuntu/jammy
 - ubuntu/focal

I haven't dug into the code yet to review, but a brief pass over it didn't yield any initial concerns.

Thanks again!

holmanb:
@akutz Have you had a chance to look at this? I believe @catmsred is testing this currently and we'll be trying to land it soon.

catmsred:
@akutz @PengpengSun I have been struggling to set up a VMWare testing environment to validate this proposal.  Would either of you have a chance to review and test it?  Thanks!

PengpengSun:
Hi @catmsred , Thanks for your fix, we will review and verify it on VMware platform.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,52 @@ def get_default_ip_addrs():
     addresses associated with the device used by the default route for a given
     address.
     """
-    # TODO(promote and use netifaces in cloudinit.net* modules)
```
PengpengSun, line 779: Firstly, I suggest add a test/few tests for this get_default_ip_addrs() function in TestDataSourceVMware, It seems we only mock the return_value before, since this change focuses on this function, now test is needed.

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,52 @@ def get_default_ip_addrs():
...
     ipv4 = None
     ipv6 = None
+    routes = netinfo.route_info()
+    for route in routes:
+        if "G" in route["flags"]:
+            if ipv4 and ipv4 != route["gateway"]:
+                LOG.debug(
+                    "multiple ipv4 gateways: %s, %s", ipv4, route["gateway"]
+                )
+            ipv4 = route["gateway"]
```
PengpengSun, line 786: I think route["gateway"] should be a gateway address, so here ipv4 should be gw4. Furthermore, can this make sure this is a default gateway?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,52 @@ def get_default_ip_addrs():
...
-    if (
-        netifaces.AF_INET not in default_gw
-        and netifaces.AF_INET6 not in default_gw
-    ):
-        return None, None
-
     ipv4 = None
     ipv6 = None
+    routes = netinfo.route_info()
+    for route in routes:
```
PengpengSun, line 780: I'm not familiar with netinfo.route_info(), but with my local check,  it returns a dict but list, so is this loop working as expected?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,52 @@ def get_default_ip_addrs():
...
+                LOG.debug(
+                    "multiple ipv4 gateways: %s, %s", ipv4, route["gateway"]
+                )
+            ipv4 = route["gateway"]
+        elif "UG" in route["flags"]:
+            if ipv6 and ipv6 != route["gateway"]:
+                LOG.debug(
+                    "multiple ipv6 gateways: %s, %s", ipv6, route["gateway"]
+                )
+            ipv6 = route["gateway"]
```
PengpengSun, line 792: ditto. Is this a gw4 or ipv6?

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,52 @@ def get_default_ip_addrs():
...
-                        "device %s has more than one ipv6 address: %s",
-                        dev6,
-                        af_inet6,
-                    )
-                elif "addr" in af_inet6[0]:
-                    ipv6 = af_inet6[0]["addr"]
+    # If there is an IPv4 gateway and an IPv6 gateway, return immediately to
+    # avoid extra work
+    if ipv4 and ipv6:
+        return ipv4, ipv4
```
PengpengSun, line 797: This function is to get ipv4 and ipv6 addresses of default interface, so return gateway is not expected. And it should be `return ipv4, ipv6`

cloudinit/sources/DataSourceVMware.py
catmsred: Thank you for the feedback!  Do you have any sense of the standard use cases this function gets called in?  I agree that the lack of unit tests makes it harder to validate my changes, but I would like to make sure any unit tests I write capture the original goals of the function rather than my perhaps imperfect understanding of what it should be.

cloudinit/sources/DataSourceVMware.py
catmsred: netinfo.route_info() has _netdev_route_info_iproute and _netdev_route_info_netstat as the underlying functions, both of which should return a list of dicts with keys "destination", "gateway", "metric", etc.  What circumstances are you seeing something different?

cloudinit/sources/DataSourceVMware.py
catmsred: I think I have misunderstood the goal of this function.  Based on your comments, the correct process for this functions is (one flow but actually IPv4/IPv6 for each step):
1. Identify default gateway
2. Extract interface of default gateway
3. Get IP address of interface
4. Return IP address

I will update the code to reflect this flow unless you have further corrections to add.  Thanks again for your input here!

cloudinit/sources/DataSourceVMware.py
PengpengSun: If possible, please mock return_value of netinfo.route_info() and netinfo.netdev_info() to verify below cases:
1. The interface used for the default route has only one ipv4, this function shall return the ipv4 address
2. The interface used for the default route has only one ipv6, this function shall return the ipv6 address
3. The interface used for the default route has both ipv4 and ipv6, this function shall return both ipv4 and ipv6 addresses
4. The interface used for the default route has multiple ipv4 and multiple ipv6,  this function shall return None, None
5. There is no default route, this function shall return None, None

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,62 @@ def get_default_ip_addrs():
...
+    # Get ipv4 and ipv6 interfaces associated with default routes
+    ipv4_if = None
+    ipv6_if = None
+    routes = netinfo.route_info()
+    for route in routes["ipv4"]:
+        if route["destination"] == "0.0.0.0":
+            ipv4_if = route["iface"]
+    for route in routes["ipv6"]:
+        if route["destination"] == "::/0":
+            ipv6 = route["iface"]
```
PengpengSun, line 787: s/ipv6/ipv6_if/

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +774,62 @@ def get_default_ip_addrs():
...
-                    dev6,
-                    af_inet4,
-                )
-            elif "addr" in af_inet4[0]:
-                ipv4 = af_inet4[0]["addr"]
+    if ipv4 is None and ipv6 is not None:
+        for dev_name in netdev:
+            for addr in netdev[dev_name]["ipv6"]:
+                if addr["ip"] == ipv6 and len(netdev[dev_name]["ipv4"]) == 1:
+                    ipv4 = netdev[dev_name]["ipv6"][0]["ip"]
```
PengpengSun, line 827: s/"ipv6"/"ipv4"/

cloudinit/sources/DataSourceVMware.py
PengpengSun: Hi @catmsred 
Overall the latest change looks good to me, only few nits.  
Please bear with me for one more week while I'm out of office for Chinese Spring Festival. In the meanwhile, I will check with @akutz if he want to take a look too.
Again, thanks for working on this.

cloudinit/sources/DataSourceVMware.py
PengpengSun, line 1: From L20 to L63 is the comment about why Netifaces was introduced, now if we are moving away from Netifaces, you can remove this comment, thanks.

akutz:
I have put time on my schedule to review this today and/or tomorrow.

holmanb:
> I have put time on my schedule to review this today and/or tomorrow.

@akutz ping, any updates on this?

catmsred:
@PengpengSun I have requested a re-review when you have the chance!

cloudinit/sources/DataSourceVMware.py
```diff
@@ -776,91 +729,62 @@ def get_default_ip_addrs():
...
+    # Get ipv4 and ipv6 interfaces associated with default routes
+    ipv4_if = None
+    ipv6_if = None
+    routes = netinfo.route_info()
+    for route in routes["ipv4"]:
+        if route["destination"] == "0.0.0.0":
+            ipv4_if = route["iface"]
+    for route in routes["ipv6"]:
+        if route["destination"] == "::/0":
+            ipv6_if = route["iface"]
```
PengpengSun, line 742: ditto.

PengpengSun:
> @PengpengSun I have requested a re-review when you have the chance!

@catmsred Thanks for adding the tests, the change looks good to me.

catmsred:
@PengpengSun Thank you for the feedback -- I have moved this to a non-WIP status but please let me know if there's any additional testing you would like to do before approving!

PengpengSun (approved):
Thanks for fixing this.

holmanb (approved):
Thanks for this work @catmsred!

Looks good to me.

PR #4636, opened 2023-11-28
fix: fix issue with networkd online target

eb3095:
## Proposed Commit Message
fix: fix issue with networkd online target

```
fix: fix issue with networkd online target

This adds systemd-networkd-wait-online.service to the Before's of cloud-init.service to prevent boot hangs and unitfile failures when its in use.
```

## Additional Context
We are currently encountering this on select Ubuntu releases at Vultr.

## Test Steps
We made this change live on a system and in our custom package.

## Checklist
- [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [X] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [X] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

### DIFF
--- a/systemd/cloud-init.service.tmpl
+++ b/systemd/cloud-init.service.tmpl
@@ -8,7 +8,6 @@ Wants=cloud-init-local.service
 Wants=sshd-keygen.service
 Wants=sshd.service
 After=cloud-init-local.service
-After=systemd-networkd-wait-online.service
 {% if variant in ["ubuntu", "unknown", "debian"] %}
 After=networking.service
 {% endif %}
@@ -26,6 +25,7 @@ After=wicked.service
 # would not be guaranteed at this point.
 After=dbus.service
 {% endif %}
+Before=systemd-networkd-wait-online.service
 Before=network-online.target
 Before=sshd-keygen.service
 Before=sshd.service


holmanb:
> We are currently encountering this on select Ubuntu releases at Vultr.

Details, please. This is a pretty invasive change.

holmanb:
> > We are currently encountering this on select Ubuntu releases at Vultr.
> > boot hangs and unitfile failures
> 
> Details, please. This is a pretty invasive change.

You obviously haven't even tested this code, since the rendered service file on suse will also contain:
```
After=systemd-networkd-wait-online.service
```

eb3095:
> Details, please. This is a pretty invasive change.

We are seeing on Ubuntu 22.04, and I think later I'de have to double check, that this particular service is hanging our boots for 2 minutes+ on a fairly vanilla install. With Cloud-Init configuring the network after the service, it would fail, and hang the boot for this length of time.

```
2min 79ms systemd-networkd-wait-online.service
  32.853s apt-daily-upgrade.service
  21.077s fstrim.service
  20.911s cloud-final.service
  11.417s cloud-init-local.service
   4.175s cloud-init.service
```

With Cloud-Init reporting 160s+ execution times. However with the change above, it executed within 30s.

> You obviously haven't even tested this code, since the rendered service file on suse will also contain:
> 
> ```
> After=systemd-networkd-wait-online.service
> ```

We tested it on our builds, but we do not have SUSE and our RHEL didnt hit an issue, just caught the if statement there.

I can fix it. But im curious as to why its explicitly going after the wait target during RHEL distros in light of what we are seeing here. If the network is configuring on cloud-init.service, and not local, is this not going to be an issue there too? I'm assuming because we have it disabled on our RHEL images it doesnt present an issue. However we found attempting to disable networkd or any component seems to get turned back on and still present an issue leaving us with this as a solution.

TheRealFalcon:
The current ordering is intentional. We want to ensure that the network is up and usable before the cloud-init service runs. There are a few reasons for this:
1. If a datasource hasn't been configured yet, we'll need to obtain updated networking configuration and then restart networking. This often requires a network that is up with the fallback configuration.
2. Many of the modules run during this service expect networking to be available.

So I think that changing any service ordering here is a non-starter for us.

Having `systemd-networkd-wait-online.service` take so long is definitely not normal. I would continue investigating the causes of that on your platform, rather than having cloud-init ignore the problem.

> An interesting note I suppose would this started occurring after we switched from DHCP to static on our baremetal servers and when this started occurring. Guessing that the config here is only applied after the fact, and because the system isnt DHCP the wait online is failing.

Have you verified that the configuration that cloud-init writes out in init-local is valid for your new static configuration? Could there be a mismatch between an interface and it's match condition such that when networking is coming up that it is waiting for something that will never appear? Are you sure that the network devices your instance is trying to configure are available when they're being configured? There have been race conditions with cloud-init trying to configure networking before all of the network devices are available.

eb3095:
Maybe a bit of back history here is due. We have had numerous issues with the networking portion of cloud-init and getting it to configure, and not break on us.

* First is the issue with where our configurations are sent from. In this instance, we send them over a network resource from our metadata. 169.254.169.254. On vm's this is from the host and we provide DHCP. However, on baremetal, which are just servers treated as a cloud resource in this aspect, we are providing that via some crazy networking from the switch via BGP.
* This was working fine, but we also have other network products such as virtual private networks, and for this we utilized cloud-init to configure this, as doing so by hand has proven problem-some for most clients we have.
* The next issue was that these were not configuring when network changes came in, so we started adding the hotplug config to the mix.
* The next issue after was the issue of when there is a hiccup or networking was down, where we had to add a config change, I believed we discussed this during a chat on IRC. I'll have to drag out exactly where thats happening and what it was. But the underlying issue was that if the network was not available cloud-init would ignore our data source due to an error and just default to nocloud and start wiping passwords and host keys and annoy our clients. Maybe this is the exception we throw on DHCP failure?
* Then we had an issue where docker network configs were causing the hotplug event to trigger and it was breaking networking on our baremetal machines for clients. For this we decided to switch to static config netcfg definitions.
* Now we have the issue we have here now. It seems like the hotplug config and other configs are causing cloud-init 'init's phase to write the config which is when this comes up leading to an issue where if this phase hasnt run, that unit file always fails.

As before. Im definitely able and willing to supply an unlimited enough resource allocation to a dev account for cannonical for colab and I am willing to rework this for a proper solution. This has just cause us quite some grief.

If theres an enterprise support that can help grease the wheels here I can get that going to!

Thanks for your time. If this isnt the correct fix, I would definitely prefer a correct solution to these issues as we are currently forced to lean on the one displayed here to keep our platform happy. We have as you have seen, had considerable issues getting this to all play nicely in the confines of our particular and apparently unique spec and wide offering!

As for your question, these interfaces should be up just the same as they were before we switched to the static config. And because images are written over the network via a pxe boot to the disk, theres 0 chance the network isnt available, this also seemingly works in distros where this particular service isnt in play and the debug output in our logs is staggeringly limited it seems.

eb3095:
> @eb3095 We're talking internally and will get back to you on a path forward. For now, please understand that the proposed change would break cloud-init so will not land in upstream repository.

Thats perfectly fine. If this isnt the correct approach ide rather get this right once and for all! Ive been vocal about our issues with this one particular aspect so if we can smooth it out here and now and get this corrected, the sooner the better! Ide very much like to let the correct answer to propogate through distros so we can stop providing some wacky custom package here!

If theres some bug we are bumping against i've very much be inclined to get a new proper PR in to address it, or if its something wacky we are doing correct that!

Thanks again for your time. Look forward to the response!

eb3095:
We came to find a few issues that resolved this particular problem. We still have others related to some oddness here, but this particular one seems to of been related to the prefix we use suddenly no longer being accepted.

Issue #4518, opened 2023-10-17
ENI and Network-Manager renderers create wrong IPv6 routes


karibou:
# Bug report
When defining IPv6 route in the v2 network config yaml, the IPv6 route is wrongly defined. The Network-Manager renderer adds it to the [ipv4] section.

## Steps to reproduce the problem
Using the following network-config : 
```
# cat config.yaml
version: 2
ethernets:
  eth0:
    dhcp4: true
    routes:
      - to: 169.254.42.42/32
        via: 62.210.0.1
      - via: fe80::dc00:ff:fe20:186
        to: ::/0
    addresses:
      - 2001:bc8:1210:232:dc00:ff:fe20:185/64
```
 when creating a Network-Manager configuration using : 
```
cloud-init devel net-convert -p config.yaml -k yaml -d . -D centos -O network-manager
```
The [ipv4] section has : 
```
[ipv4]
method=auto
may-fail=false
route1=169.254.42.42/32,62.210.0.1
route2=::/0,fe80::dc00:ff:fe20:186
```
As seen the IPv6 route gets added to the IPv4 section

## Environment details
- Cloud-init version: 23.3.1-0ubuntu1~22.04.1
- Operating System Distribution: Ubuntu Jammy
- Cloud provider, platform or installer type: Scaleway

## cloud-init logs
No logs are necessary, the behavior can be reproduced at will using the cloud-init command

TheRealFalcon:
@karibou , thanks for the bug report. I can confirm the problem in versions of cloud-init supporting network-manager.

karibou:
Hello,
I've been peeking at the code and might have an idea of what is happening.
At line https://github.com/canonical/cloud-init/blob/d0f00bd54649e527d69ad597cbcad6efa8548e58/cloudinit/net/network_manager.py#L256C9-L256C9 we have : 
```
        # Deal with Layer 3 configuration
        for subnet in iface["subnets"]:
            family = "ipv6" if subnet_is_ipv6(subnet) else "ipv4"

            self._set_ip_method(family, subnet["type"])
            if "address" in subnet:
                self._add_address(family, subnet)
            if "gateway" in subnet:
                self.config[family]["gateway"] = subnet["gateway"]
```
With the data given in config.yaml above we get : 
```
iface["subnets"]
[...]

PR #4637, opened 2023-11-28
fix: Correct v2 NetworkManager route rendering

TheRealFalcon:
## Proposed Commit Message

```
fix: Correct v2 NetworkManager route rendering

Because network v2 route defintions can have mixed v4 and v6 routes, we
need to determine the IP family per route rather than per subnet.

Similar, ensure dns-search is rendered correctly.

Fixes GH-4518
```

## Additional Context
https://github.com/canonical/cloud-init/issues/4518


TheRealFalcon:
@karibou , do you want to ensure that this fixes your problem on NM?

@holmanb , since you've looked at this previously, I added you as reviewer.

After this one I have a follow-on PR that adds route MTU support, but the commits seemed disparate enough that they deserved separate PRs.

tests/unittests/test_net.py
```diff
@@ -4154,6 +4154,70 @@
...
+                interface-name=eth0
+
+                [user]
+                org.freedesktop.NetworkManager.origin=cloud-init
+
+                [ethernet]
+
+                [ipv4]
+                method=auto
+                may-fail=false
```
holmanb, line 4177: Oops

tests/unittests/test_net.py
```diff
@@ -4154,6 +4154,70 @@
...
+                route2=169.254.42.43/32,62.210.0.2
+                address1=192.168.1.20/16
+                dns=8.8.8.8;
+                dns-search=lab;home;
+
+                [ipv6]
+                route1=::/0,fe80::dc00:ff:fe20:186
+                route2=fe80::dc00:ff:fe20:188/64,fe80::dc00:ff:fe20:187
+                method=auto
+                may-fail=false
```
holmanb, line 4188: That isn't right.

holmanb (changes requested):
@TheRealFalcon You'll need to rebase on main to pick up 29dd5ace73ad6 and then update the test to pass, currently it doesn't.

TheRealFalcon:
@holmanb updated

cloudinit/net/network_manager.py
```diff
@@ -189,40 +195,38 @@ def _add_address(self, family, subnet):
...
         """
         Extends the ipv[46].dns-search property with a name server.
         """
-
-        self._set_default(family, "dns-search", "")
-        self.config[family]["dns-search"] = (
-            self.config[family]["dns-search"] + ";".join(dns_search) + ";"
-        )
+        for family in ["ipv4", "ipv6"]:
+            if self.config.has_section(family):
```
holmanb, line 223: This means that a configuration that specifies no addresses will now have no search domains rendered, correct?

I'm curious what netplan does in this case, since it seems like an invalid configuration, but I think that this behavior makes sense.

cloudinit/net/network_manager.py
```diff
@@ -158,11 +163,11 @@ def _set_ip_method(self, family, subnet_type):
         if self.config[family]["method"] == "auto" and method == "manual":
             return
 
-        if (
-            subnet_type == "ipv6_dhcpv6-stateful"
-            or subnet_type == "ipv6_dhcpv6-stateless"
-            or subnet_type == "ipv6_slaac"
-        ):
+        if subnet_type in [
```
holmanb, line 166: +1, more succinct

cloudinit/net/network_manager.py
```diff
@@ -334,7 +338,7 @@ def render_interface(self, iface, renderer):
 
         if ipv4_mtu is None:
             ipv4_mtu = device_mtu
-        if not ipv4_mtu == device_mtu:
+        if ipv4_mtu != device_mtu:
```
holmanb, line 341: Your earlier comment:

> After this one I have a follow-on PR that adds route MTU support, but the commits seemed disparate enough that they deserved separate PRs.

Did you mean to make this change?

holmanb (commented):
Thanks @TheRealFalcon, this looks really good for the most part. Great test addition and some nice refactors along the way too.

In the commit message, can you please describe what was previously incorrect about the route rendering?

Also, did you mean to include the mtu change? If so can you please add some context on that one?

cloudinit/net/network_manager.py
TheRealFalcon:
```
root@me:/etc/netplan# cat 50-cloud-init.yaml 
network:
  version: 2
  renderer: NetworkManager
  ethernets:
    eth0:
      nameservers:
        search: [lab, home]
        addresses: [8.8.8.8, FEDC::1]
root@me:/etc/netplan# cat /run/NetworkManager/system-connections/netplan-eth0.nmconnection
[connection]
id=netplan-eth0
type=ethernet
interface-name=eth0

[ethernet]
wake-on-lan=0

[ipv4]
method=link-local
dns=8.8.8.8;

[ipv6]
method=manual
ip6-privacy=0
dns=FEDC::1;
dns-search=lab;home;
```

I have yet to make any conclusions about it :smile:

holmanb: > I have yet to make any conclusions about it 😄

Errr, why isn't netplan adding `dns-search` to the ipv4 entry? Is that a netplan bug?

TheRealFalcon: Yeah, I'm not sure about that either. I'll file a bug and see. 

I did push an update though. This one waits until we've processed all of the subnets, and then adds the dns search entries. That means if we haven't defined ipv6 (or ipv4) at all, then we won't try to add the dns-search entry to that family.

TheRealFalcon: > I'll file a bug

Jk, found this: https://bugs.launchpad.net/netplan/+bug/1786726

TheRealFalcon:
@holmanb I believe this is ready for re-review.

cloudinit/net/network_manager.py
holmanb: Gotcha, that explains the behavior thanks

holmanb (commented):
```yaml
version: 2
ethernets:
  eth0:
    dhcp4: true
    dhcp6: true
    nameservers:
      search: [lab, home]
      addresses: [8.8.8.8, "FEDC::1"]

```

with netplan I see 
```
[connection]
id=netplan-eth0
type=ethernet
interface-name=eth0

[ethernet]
wake-on-lan=0

[ipv4]
method=auto
dns=8.8.8.8;
dns-search=lab;home;

[ipv6]
method=auto
ip6-privacy=0
dns=FEDC::1;
dns-search=lab;home;
```

but with cloud-init we generate an empty `[ethernet]` section (is that even a legal config?) and don't include any of the `dns` or `dns-search` content.

Similar issue when no `dhcp{4,6}: true` is specified, except that the cloud-init renderer drops both ip4 and ipv6 sections

holmanb:
I think we would want to see this test pass.
```diff
diff --git a/tests/unittests/test_net.py b/tests/unittests/test_net.py
index 7ee1cc719..c9714b5c6 100644
--- a/tests/unittests/test_net.py
+++ b/tests/unittests/test_net.py
@@ -4218,6 +4218,95 @@ iface bond0 inet6 static
         """
         ),
     },
+    "v2-dns-no-if-ips": {
+        "expected_network_manager": {
+            "cloud-init-eth0.nmconnection": textwrap.dedent(
+                """\
+                # Generated by cloud-init. Changes will be lost.
+
+                [connection]
+                id=cloud-init eth0
+                uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
+                autoconnect-priority=120
+                type=ethernet
+                interface-name=eth0
+
+                [user]
+                org.freedesktop.NetworkManager.origin=cloud-init
+
+                [ethernet]
+
+                [ipv4]
+                method=auto
+                may-fail=true
+                dns=8.8.8.8;
+                dns-search=lab;home;
+
+                [ipv6]
+                method=auto
+                may-fail=true
+                dns=FEDC::1;
+                dns-search=lab;home;
+
+            """
+            )
+        },
+        "yaml": textwrap.dedent(
+            """\
+            version: 2
+            ethernets:
+              eth0:
+                dhcp4: true
+                dhcp6: true
+                nameservers:
+                  search: [lab, home]
+                  addresses: [8.8.8.8, "FEDC::1"]
+        """
+        ),
+    },
+    "v2-dns-no-dhcp": {
+        "expected_network_manager": {
+            "cloud-init-eth0.nmconnection": textwrap.dedent(
+                """\
+                # Generated by cloud-init. Changes will be lost.
+
+                [connection]
+                id=cloud-init eth0
+                uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
+                autoconnect-priority=120
+                type=ethernet
+                interface-name=eth0
+
+                [user]
+                org.freedesktop.NetworkManager.origin=cloud-init
+
+                [ethernet]
+
+                [ipv4]
+                method=auto
+                may-fail=true
+                dns=8.8.8.8;
+
+                [ipv6]
+                method=auto
+                may-fail=true
+                dns=FEDC::1;
+                dns-search=lab;home;
+
+            """
+            )
+        },
+        "yaml": textwrap.dedent(
+            """\
+            version: 2
+            ethernets:
+              eth0:
+                nameservers:
+                  search: [lab, home]
+                  addresses: [8.8.8.8, "FEDC::1"]
+        """
+        ),
+    },
 }
 
 
@@ -6338,6 +6427,20 @@ class TestNetworkManagerRendering(CiTestCase):
             entry[self.expected_name], self.expected_conf_d, found
         )
 
+    def test_v2_dns_no_ips(self):
+        entry = NETWORK_CONFIGS["v2-dns-no-if-ips"]
+        found = self._render_and_read(network_config=yaml.load(entry["yaml"]))
+        self._compare_files_to_expected(
+            entry[self.expected_name], self.expected_conf_d, found
+        )
+
+    def test_v2_dns_no_dhcp(self):
+        entry = NETWORK_CONFIGS["v2-dns-no-dhcp"]
+        found = self._render_and_read(network_config=yaml.load(entry["yaml"]))
+        self._compare_files_to_expected(
+            entry[self.expected_name], self.expected_conf_d, found
+        )
+
 
 @mock.patch(
     "cloudinit.net.is_openvswitch_internal_interface",
```

TheRealFalcon:
Man...this network state code...
There's a top level dns section that only gets populated in v2, and then a subnet-specific dns section that gets populated in both configs...unless you have no other subnets defined :man_facepalming: 

Thanks for the additional tests. I understand and agree with your first test, but this one:
```
+    "v2-dns-no-dhcp": {
+        "expected_network_manager": {
+            "cloud-init-eth0.nmconnection": textwrap.dedent(
+                """\
+                # Generated by cloud-init. Changes will be lost.
+
+                [connection]
+                id=cloud-init eth0
+                uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
+                autoconnect-priority=120
+                type=ethernet
+                interface-name=eth0
+
+                [user]
+                org.freedesktop.NetworkManager.origin=cloud-init
+
+                [ethernet]
+
+                [ipv4]
+                method=auto
+                may-fail=true
+                dns=8.8.8.8;
+
+                [ipv6]
+                method=auto
+                may-fail=true
+                dns=FEDC::1;
+                dns-search=lab;home;
+
+            """
+            )
+        },
+        "yaml": textwrap.dedent(
+            """\
+            version: 2
+            ethernets:
+              eth0:
+                nameservers:
+                  search: [lab, home]
+                  addresses: [8.8.8.8, "FEDC::1"]
+        """
+        ),
+    },
```

Why `auto` or `may-fail=true`? If there's no dhcp or static addresses, I don't think we should be auto-configuring anything. If we're going to allow a network config with DNS but no DHCP and no static IPs (why would anybody do this??), then I think we need to copy the way netplan does it. That's some additional state and complexity for something that I can't see being a real use case. We could eventually support it, but I don't see it being a high priority or something that needs to be added here. 

I pushed what I think makes the most sense currently. Let me know if you agree.

holmanb:
> If we're going to allow a network config with DNS but no DHCP and no static IPs (why would anybody do this??),

If IP is assigned in initrd or managed by some other service (a routing daemon or other external service), then NetworkManager doesn't need to know about it.

holmanb (approved):
Thanks for this @TheRealFalcon!

This looks like an improvement to me over the current state of code. Other improvements can happen at a later date.

PR #4638, opened 2023-11-28
refactor: Remove metaclass from network_state.py

TheRealFalcon:
## Proposed Commit Message

```
refactor: Remove metaclass from network_state.py

It's magic that makes it hard to follow call hierachies without saving
much code.
```

## Test Steps
Existing unit test coverage should suffice. If this broke anything it should be pretty apparent.

### DIFF
--- a/cloudinit/net/network_state.py
+++ b/cloudinit/net/network_state.py
@@ -123,25 +123,6 @@ def ensure_command_keys(required_keys):
     return wrapper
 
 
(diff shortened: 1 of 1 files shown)


holmanb (approved):
Thanks much for this! Metaclasses are neat and fun to toy around with, but cost us more than they help us since they cause developer confusion and defeat static code analyzers.

PR #4639, opened 2023-11-29
test: fix schema fuzzing test

TheRealFalcon:
## Proposed Commit Message

```
test: fix schema fuzzing test

Since adding the top-level properties key to the schema, the fuzz test
will only generate empty tests. Update the test to remove/ignore the
top-level keys along with ensuring fuzzed additionalProperties are
removed from the test.
```


holmanb:
> Since adding the top-level properties key to the schema, the fuzz test
will only generate empty tests

Curious, do we know why?

TheRealFalcon:
I don't know but if I had to guess the "properties" top-level key (which currently has empty definitions) is chosen over the "allOf" definition

holmanb (approved):

PR #4643, opened 2023-11-30
azure: report failure to eject as error instead of debug

cjp256:
I recently noticed some images do not include `eject` and this failure is lost in the debug messages.  Raising it to error to improve visibility of the error.

Given that the media may contain the plain-text admin password, we strongly prefer to eject it as soon as possible from within the VM.  If the media is not ejected by the client, the host will do so with some delay when cloud-init reports ready/failure to WireServer.


blackboxsw (approved):
LGTM better be noisy about significant impacts to boot config when concerning security.

PR #4644, opened 2023-11-30
Update build-on-openbsd dependencies

CodeBleu:
fix: Update build-on-openbsd dependencies

    * script would fail will needing to do work with python netifaces
      module.
    * linting: SC2806 -  Double quote to prevent globbing and word
      splitting.


igalic (approved):
👍

holmanb (approved):
LGTM

PR #4654, opened 2023-12-02
dmi: support OpenBSD native reading 

igalic:
## Proposed Commit Message

```
fix: add support for OpenBSD's native dmi reading

OpenBSD's sysctl provides the most important DMI data in
its `hw` hierarchy. Switch ds-identify and cloudinit.dmi to using it.

This means we can get rid of the dmidecode dependency, which
requires setting `kern.allowkmem=1` and a reboot before usage.

```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/dmi.py
+++ b/cloudinit/dmi.py
@@ -6,57 +6,74 @@ from collections import namedtuple
 from typing import Optional
 
 from cloudinit import subp
-from cloudinit.util import is_container, is_FreeBSD
+from cloudinit.util import (
+    is_container,
+    is_DragonFlyBSD,
+    is_FreeBSD,
+    is_OpenBSD,
+)
 
 LOG = logging.getLogger(__name__)
 
 # Path for DMI Data
 DMI_SYS_PATH = "/sys/class/dmi/id"
 
-KernelNames = namedtuple("KernelNames", ["linux", "freebsd"])
-KernelNames.__new__.__defaults__ = (None, None)
+KernelNames = namedtuple("KernelNames", ["linux", "freebsd", "openbsd"])
+KernelNames.__new__.__defaults__ = (None, None, None)
 
 # FreeBSD's kenv(1) and Linux /sys/class/dmi/id/* both use different names from
 # dmidecode. The values are the same, and ultimately what we're interested in.
+# Likewise, OpenBSD has the most commonly used things we need in sysctl(8)'s
+# hw hierarchy. Moreover, this means it can be run without kern.allowkmem=1.
 # These tools offer a "cheaper" way to access those values over dmidecode.
 # This is our canonical translation table. If we add more tools on other
 # platforms to find dmidecode's values, their keys need to be put in here.
 DMIDECODE_TO_KERNEL = {
-    "baseboard-asset-tag": KernelNames("board_asset_tag", "smbios.planar.tag"),
+    "baseboard-asset-tag": KernelNames(
+        "board_asset_tag", "smbios.planar.tag", None
+    ),
     "baseboard-manufacturer": KernelNames(
-        "board_vendor", "smbios.planar.maker"
+        "board_vendor", "smbios.planar.maker", None
     ),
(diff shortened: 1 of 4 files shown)


tools/ds-identify
```diff
@@ -534,8 +560,12 @@ read_uname_info() {
     if [ -z "$out" ]; then
         out=$(uname -snrvmo) || {
             ret=$?
-            error "failed reading uname with 'uname -snrvmo'"
-            return $ret
+            error "failed reading uname with 'uname -snrvmo', retrying with -snrvm"
+            out=$(uname -snrvmo) || {
```
CodeBleu, line 564: Adding this here :smile: 
Need to change to `snrvm` instead

holmanb (commented):
@igalic Nice work here, thanks for putting this together. This looks really close, but I have a couple of requests.

I left a couple of stylistic nitpick comments, feel free to take them or leave them. Also a couple of minor requests and questions.

The `-n`,`-r`, `-o` flags aren't accepted in OpenBSD's `uname` I take it? The contents of `DI_UNAME_NODENAME`, `DI_UNAME_KERNEL_RELEASE`, and `DI_UNAME_OPERATING_SYSTEM` were valuable for debugging, but there are other ways to get that data if needed, and their output in the log isn't a strict requirement, so that needs not happen in `ds-identify`. Improved cross-platform compatibility is higher priority than info in the log. 

Commit 69f5c2fc2ccbbcd822b should mention something about DragonFlyBSD.

igalic:
> The `-n`,`-r`, `-o` flags aren't accepted in OpenBSD's `uname` I take it?

everything except -o.
I dropped the others because it made parsing easier.

> The contents of `DI_UNAME_NODENAME`, `DI_UNAME_KERNEL_RELEASE`, and `DI_UNAME_OPERATING_SYSTEM` were valuable for debugging, but there are other ways to get that data if needed, and their output in the log isn't a strict requirement, so that needs not happen in `ds-identify`.

I didn't realise that. i could try harder to parse it.

> Improved cross-platform compatibility is higher priority than info in the log.

👍 

> Commit [69f5c2f](https://github.com/canonical/cloud-init/commit/69f5c2fc2ccbbcd822b56e87e3074872f7738a8c) should mention something about DragonFlyBSD.

👍

cloudinit/dmi.py
```diff
@@ -159,9 +198,12 @@ def read_dmi_data(key: str) -> Optional[str]:
     if is_container():
         return None
 
-    if is_FreeBSD():
+    if is_FreeBSD() or is_DragonFlyBSD():
```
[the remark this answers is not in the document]
igalic: A key benefit to doing that is that we can fall-back to much, much easier.
So we can use the fast method first, and if that fails go to dmidecode (which then may throw an exception because it's not installed, or in OpenBSD's case, because it's not enabled)

cloudinit/dmi.py
```diff
@@ -120,6 +137,28 @@ def _read_kenv(key: str) -> Optional[str]:
...
+    """
+    kmap = DMIDECODE_TO_KERNEL.get(key)
+    if kmap is None or kmap.openbsd is None:
+        return None
+
+    LOG.debug("querying dmi data %s", kmap.openbsd)
+
+    try:
+        cmd = ["sysctl", "-qn", kmap.openbsd]
+        (result, _err) = subp.subp(cmd)
```
holmanb, line 152: style nit: parens aren't required in assignment, named tuple attribute could be referenced also to avoid the unused assignment

```python
        result = subp.subp(cmd).stdout
```

not a required change, just a style suggestion

holmanb (approved):
Thanks @igalic. This looks good to me.

@CodeBleu I see you got in an early review on this one. Does everything look good now?

cloudinit/dmi.py
igalic: @holmanb any idea why unittests are now failing after thischange?

CodeBleu:
I'm good with this change :+1:

PR #4660, opened 2023-12-04
fix(openbsd): services & build tool

CodeBleu:
## Proposed Commit Message
```
fix(openbsd): services & build tool

    * Added net/activator for `ifconfig` that is used with OpenBSD
    * Added sysvinit scripts for OpenBSD
    * Updated build-on-openbsd tool with additional dependencies and
      better logic
```

### DIFF
--- a/cloudinit/net/activators.py
+++ b/cloudinit/net/activators.py
@@ -102,6 +102,33 @@ class IfUpDownActivator(NetworkActivator):
         return _alter_interface(cmd, device_name)
 
 
+class IfConfigActivator(NetworkActivator):
+    @staticmethod
+    def available(target=None) -> bool:
+        """Return true if ifconfig can be used on this system."""
+        expected = "ifconfig"
+        search = ["/sbin"]
+        return subp.which(expected, search=search, target=target)
+
+    @staticmethod
+    def bring_up_interface(device_name: str) -> bool:
+        """Bring up interface using ifconfig <dev> up.
+
+        Return True is successful, otherwise return False
+        """
+        cmd = ["ifconfig", device_name, "up"]
+        return _alter_interface(cmd, device_name)
+
+    @staticmethod
+    def bring_down_interface(device_name: str) -> bool:
+        """Bring up interface using ifconfig <dev> down.
+
+        Return True is successful, otherwise return False
+        """
+        cmd = ["ifconfig", device_name, "down"]
+        return _alter_interface(cmd, device_name)
+
+
 class NetworkManagerActivator(NetworkActivator):
     @staticmethod
     def available(target=None) -> bool:
@@ -220,6 +247,7 @@ DEFAULT_PRIORITY = [
     "netplan",
     "network-manager",
     "networkd",
+    "ifconfig",
 ]
 
 NAME_TO_ACTIVATOR: Dict[str, Type[NetworkActivator]] = {
@@ -227,6 +255,7 @@ NAME_TO_ACTIVATOR: Dict[str, Type[NetworkActivator]] = {
     "netplan": NetplanActivator,
     "network-manager": NetworkManagerActivator,
     "networkd": NetworkdActivator,
+    "ifconfig": IfConfigActivator,
 }
 
 
--- a/sysvinit/openbsd/cloudconfig.tmpl
+++ b/sysvinit/openbsd/cloudconfig.tmpl
@@ -1,0 +1,37 @@
+## template:jinja
+#!/bin/ksh
+
+# PROVIDE: cloudconfig
+# REQUIRE: cloudinit
+# BEFORE: sshd
+
+
+daemon="cloud-init"
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="modules --mode config"
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES" # (undefined or "NO")
+
+rc_start() {
+    if test -e /etc/cloud/cloud-init.disabled; then
+        echo -n "cloud-init is disabled via cloud-init.disabled file." | logger -t ${daemon}
+    else
+        echo -n "Starting..." | logger -t ${daemon}
+        rc_exec "${daemon_execdir}/${daemon} ${daemon_flags}"
+    fi
+}
+
+rc_check() {
+    pgrep -f "${daemon}" >/dev/null
+}
+
+rc_stop() {
+    echo -n "Stopping..." | logger -t ${daemon}
+    pkill -f "${daemon}" >/dev/null
+}
+
+rc_cmd "$1"
--- a/sysvinit/openbsd/cloudfinal.tmpl
+++ b/sysvinit/openbsd/cloudfinal.tmpl
@@ -1,0 +1,35 @@
+## template:jinja
+#!/bin/ksh
+
+# PROVIDE: cloudfinal
+# REQUIRE: LOGIN cloudconfig
+
+daemon="cloud-init"
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="modules --mode final"
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES" # (undefined or "NO")
+
+rc_start() {
+    if test -e /etc/cloud/cloud-init.disabled; then
+        echo -n "cloud-init is disabled via cloud-init.disabled file." | logger -t ${daemon}
+    else
+        echo -n "Starting..." | logger -t ${daemon}
+        rc_exec "${daemon_execdir}/${daemon} ${daemon_flags}"
+    fi
+}
+
+rc_check() {
+    pgrep -f "${daemon}" >/dev/null
+}
+
+rc_stop() {
+    echo -n "Stopping..." | logger -t ${daemon}
+    pkill -f "${daemon}" >/dev/null
+}
+
+rc_cmd "$1"
--- a/sysvinit/openbsd/cloudinit.tmpl
+++ b/sysvinit/openbsd/cloudinit.tmpl
@@ -1,0 +1,35 @@
+## template:jinja
+#!/bin/ksh
+
+# PROVIDE: cloudinit
+# REQUIRE: cloudinitlocal
+
+daemon="cloud-init"
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="init"
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES" # (undefined or "NO")
+
+rc_start() {
+    if test -e /etc/cloud/cloud-init.disabled; then
+        echo -n "cloud-init is disabled via cloud-init.disabled file." | logger -t ${daemon}
+    else
+        echo -n "Starting..." | logger -t ${daemon}
+        rc_exec "${daemon_execdir}/${daemon} ${daemon_flags}"
+    fi
+}
+
+rc_check() {
+    pgrep -f "${daemon}" >/dev/null
+}
+
+rc_stop() {
+    echo -n "Stopping..." | logger -t ${daemon}
+    pkill -f "${daemon}" >/dev/null
+}
+
+rc_cmd "$1"
--- a/sysvinit/openbsd/cloudinitlocal.tmpl
+++ b/sysvinit/openbsd/cloudinitlocal.tmpl
@@ -1,0 +1,38 @@
+## template:jinja
+#!/bin/ksh
+
+# PROVIDE: cloudinitlocal
+# REQUIRE: NETWORKING
+
+# After NETWORKING because we don't want staticroute to wipe
+# the route set by the DHCP client toward the meta-data server.
+
+daemon="cloud-init"
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="init -l"
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES"
+
+rc_start() {
+    if test -e /etc/cloud/cloud-init.disabled; then
+        echo -n "cloud-init is disabled via cloud-init.disabled file." | logger -t ${daemon}
+    else
+        echo -n "Starting..." | logger -t ${daemon}
+        rc_exec "${daemon_execdir}/${daemon} ${daemon_flags}"
+    fi
+}
+
+rc_check() {
+    pgrep -f "${daemon}" >/dev/null
+}
+
+rc_stop() {
+    echo -n "Stopping..." | logger -t ${daemon}
+    pkill -f "${daemon}" >/dev/null
+}
+
+rc_cmd "$1"
--- a/tools/build-on-openbsd
+++ b/tools/build-on-openbsd
@@ -2,6 +2,12 @@
 
 fail() { echo "FAILED:" "$@" 1>&2; exit 1; }
 
+PYTHON=${PYTHON:-python3}
+if ! command -v ${PYTHON} >/dev/null 2>&1; then
+    echo "Please install python first."
+    exit 1
+fi
+
 # Check dependencies:
 depschecked=/tmp/c-i.dependencieschecked
 pkgs="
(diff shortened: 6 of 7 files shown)


cloudinit/net/activators.py
```diff
@@ -216,13 +241,15 @@ def bring_down_interface(device_name: str) -> bool:
 # This section is mostly copied and pasted from renderers.py. An abstract
 # version to encompass both seems overkill at this point
 DEFAULT_PRIORITY = [
+    "ifconfig",
```
igalic, line 244: this should either be last, or alphabetic

cloudinit/net/activators.py
```diff
@@ -216,13 +241,15 @@ def bring_down_interface(device_name: str) -> bool:
...
     "eni",
     "netplan",
     "network-manager",
     "networkd",
 ]
 
 NAME_TO_ACTIVATOR: Dict[str, Type[NetworkActivator]] = {
+    "ifconfig": IfConfigActivator,
```
igalic, line 252: this should either be last, or alphabetic

cloudinit/net/ifconfig.py
```diff
@@ -0,0 +1,22 @@
...
+from cloudinit import subp, util
+from cloudinit.net import ParserError, renderer, subnet_is_ipv6
+from cloudinit.net.network_state import NetworkState
+
+LOG = logging.getLogger(__name__)
+
+def available(target=None):
+    expected = ["ifconfig"]
+    search = ["/sbin"]
+    for p in expected:
```
igalic, line 19: this loop can be eliminated. And honestly, maybe the `which(ifconfig)`, too ;)
We might just check `if is_BSD()` and be done with it

sysvinit/openbsd/cloudconfig.tmpl
```diff
@@ -0,0 +1,56 @@
...
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="modules --mode config"
+#daemon_logger=
+daemon_rtable=0
+daemon_timeout=1
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+#pexp="$(eval echo ${daemon}${daemon_flags:+ ${daemon_flags}})"
```
igalic, line 19: please get rid of everything that's commented out in these files.

sysvinit/openbsd/cloudinitlocal.tmpl
```diff
@@ -0,0 +1,62 @@
+## template:jinja
+#!/bin/ksh
+
+# PROVIDE: cloudinitlocal
+{#
+``cloudinitlocal`` purposefully does not depend on ``dsidentify``.
+That makes it easy for image builders to create images without ``dsidentify``.
```
igalic, line 7: you can delete this comment, and I should delete it from the FreeBSD rc scripts.
I've made it so that dsidentify can be easily disabled

on OpenBSD, that is easier still.

tools/build-on-openbsd
```diff
@@ -16,12 +22,23 @@ pkgs="
    py3-setuptools
    py3-yaml
    sudo--
+   wget
```
igalic, line 25: What's that used for?

igalic (changes requested):

tools/build-on-openbsd
CodeBleu: I don't recall right off, so I will take out and re-run to see if I still get an error in the logs about this.  This might have been an issue when I was doing a build on older version of cloud-init some how and after I cleaned all the cached stuff, this might not be an issue anymore.

cloudinit/net/ifconfig.py
CodeBleu: I'll remove the loop, but leave the `which(ifconfig)` for now

tools/build-on-openbsd
igalic: okay, so that's CloudStack specific.

cloudinit/net/ifconfig.py
```diff
@@ -0,0 +1,21 @@
...
+import copy
+import glob
+import logging
+import os
+import re
+from typing import Optional
+
+from cloudinit import subp, util
+from cloudinit.net import ParserError, renderer, subnet_is_ipv6
+from cloudinit.net.network_state import NetworkState
```
igalic, line 12: The linter should be complaining about lots of unused imports.
Please give it a run and see which ones (most) need to be removed.

sysvinit/openbsd/cloudfinal.tmpl
```diff
@@ -0,0 +1,34 @@
...
+daemon_timeout=1
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES" # (undefined or "NO")
+
+rc_start() {
+    test -e /etc/cloud/cloud-init.disabled \
```
igalic, line 20: please replace this `test && …` with a simple `if` like the FreeBSD rc files have.
an `if` is easier to follow, less easy to mess up.
We'll need to do that in the NetBSD rc files too.

sysvinit/openbsd/cloudinit.tmpl
```diff
@@ -0,0 +1,34 @@
...
+daemon_timeout=1
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES" # (undefined or "NO")
+
+rc_start() {
+    test -e /etc/cloud/cloud-init.disabled \
```
igalic, line 20: same as above

(but isn't it funny what reviewers can suddenly see when the files aren't full of comments ;)

igalic (changes requested):
Great progress!
Looking a lot better.
Now it's time to make linters happy, not just reviewers!

tools/build-on-openbsd
```diff
@@ -2,6 +2,12 @@
...
 
+PYTHON=${PYTHON:-python3}
+if [ ! $(which ${PYTHON}) ]; then
+    echo "Please install python first."
+    exit 1
+fi
+
 # Check dependencies:
 depschecked=/tmp/c-i.dependencieschecked
 pkgs="
```
igalic, line 13: n.b.: following #4654, we can drop the dependency on dmidecode

igalic (commented):
reminder, re #4654

igalic (approved):
👍🏻

tools/build-on-openbsd
```diff
@@ -2,6 +2,12 @@
 
 fail() { echo "FAILED:" "$@" 1>&2; exit 1; }
 
+PYTHON=${PYTHON:-python3}
+if [ ! $(which ${PYTHON}) ]; then
```
holmanb, line 6: Which shell is `/bin/sh` on openbsd? Typically `command -v` is more portable.

sysvinit/openbsd/cloudinitlocal.tmpl
```diff
@@ -0,0 +1,40 @@
...
+daemon_timeout=1
+daemon_user=root
+
+. /etc/rc.d/rc.subr
+
+rc_bg="YES" # (undefined or "YES")
+rc_usercheck="YES"
+
+rc_start() {
+    if test -e /etc/cloud/cloud-init.disabled; then
```
holmanb, line 23: Good that we check this. On other *nixes we also check [kernel commandline](https://cloudinit.readthedocs.io/en/latest/howto/disable_cloud_init.html#method-2-kernel-commandline). Is that doable on OpenBSD? Using a `sysctl kern.something` maybe?

It's nice to allow the user to disable without modifying the root filesystem.

Same comment on the other 3 init scripts.

sysvinit/openbsd/cloudconfig.tmpl
```diff
@@ -0,0 +1,39 @@
+## template:jinja
```
holmanb, line 1: Why is this a template? (same comment on the others)

holmanb (commented):
First pass review, no major objections to this so far @CodeBleu.

I left a few comments and questions inline.

tools/build-on-openbsd
igalic: sh is Bourne Shell.
(the default (root) shell of ksh)

sysvinit/openbsd/cloudinitlocal.tmpl
igalic: see ` daemon_execdir="{{prefix}}/bin" ` for example

igalic, line 23: this file should probably be in `{{prefix}}/etc… `

igalic (changes requested):
one more thing

sysvinit/openbsd/cloudinitlocal.tmpl
CodeBleu: I looked this up initially (kernel options) , but I did not see a way to do this.  When googling it also said to have something like kern.options=cloud-init.disabled would require customizing the kernel.  So, that is why I left the kernel option check out.

sysvinit/openbsd/cloudconfig.tmpl
CodeBleu: Because it is using this in each file:

`daemon_execdir="{{prefix}}/bin"`

sysvinit/openbsd/cloudinitlocal.tmpl
CodeBleu: I did not do this because `etc` is at `/etc` not `/usr/local/etc`

cloudinit/net/ifconfig.py
```diff
@@ -0,0 +1,15 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import logging
+
+from cloudinit import subp
+
+LOG = logging.getLogger(__name__)
+
+
+def available(target=None):
```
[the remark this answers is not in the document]
CodeBleu: Good point. I'll refactor to:

```Python
def available(target=None):
     expected = "ifconfig"
     search = ["/sbin"]
     return subp.which(expected, search=search, target=target)
```

CodeBleu:
@holmanb @igalic  Any objections on just using /etc/rc.local for now?  I think it would be good to have something working for the moment, then we can always come back and make it better.  I'm going to put this PR back in `Draft` until I hear back from you all and make changes needed.

igalic:
given all the feedback we've received from the #OpenBSD folks, this seems like the best option

tools/build-on-openbsd
```diff
@@ -16,12 +22,45 @@ pkgs="
...
+
+cloud-init modules --mode config
+wait
+
+cloud-init modules --mode final
+"
+    if ! test -e $RC_LOCAL; then
+        echo "export PATH=$PATH:/usr/local/sbin:/usr/local/bin" >> $RC_LOCAL
+        echo "$RC_LOCAL_CONTENT" >> $RC_LOCAL
+    elif ! grep -Fq "cloud-init" $RC_LOCAL; then
```
igalic, line 63:
```suggestion
    if ! grep -Fq "PATH" $RC_LOCAL; then;
        echo "export PATH=$PATH:/usr/local/sbin:/usr/local/bin" >> $RC_LOCAL
    fi
    if ! grep -Fq "cloud-init" $RC_LOCAL; then
```

CodeBleu: @igalic @holmanb Drafting this again...uggh.  I thought I had this updated and working and pushed it up, but it's not. @igalic  I had thought I was making it better by just removing `PATH` altogether, and it worked for the most part, but looking at the logs further it complained it couldn't find `wget` again.

CodeBleu: @igalic OK, I have reverted to back to what it was.  Unless there is a strong reason to update with your suggestion, I think I'd just like to keep it the way it was.  I tested that and it seemed to work, and I need to work on some other things for a bit.

tools/build-on-openbsd
```diff
@@ -16,12 +22,45 @@ pkgs="
...
+cloud-init init --local
+wait
+
+cloud-init init
+wait
+
+cloud-init modules --mode config
+wait
+
+cloud-init modules --mode final
```
holmanb, line 58: If the `wait`s are necessary above, why not here?

holmanb (commented):
Great work @CodeBleu. I have another round of questions / requests for you.

sysvinit/openbsd/cloudconfig.tmpl
```diff
@@ -0,0 +1,39 @@
...
+# PROVIDE: cloudconfig
+# REQUIRE: cloudinit
+# BEFORE: sshd
+
+
+daemon="cloud-init"
+daemon_execdir="{{prefix}}/bin"
+daemon_flags="modules --mode config"
+daemon_rtable=0
+daemon_timeout=1
```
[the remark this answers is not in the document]
igalic: from [rc.subr(8)](https://man.openbsd.org/rc.subr.8)

> #### daemon_timeout
>    Maximum time in seconds to wait for the start, stop and reload actions to return. This is only guaranteed with the default rc_start, rc_stop and rc_reload functions.

so, yeah, i think it needs to go.
provided we keep these service files to begin with.

OpenBSD's rc can only really manage services that remain running. It's not very good at one-shot tasks.
We might be able to fake it by having rc_check always return true, but i don't know how well that will actually work. because it doesn't give us a guarantee that a stage is finished

sysvinit/openbsd/cloudconfig.tmpl
```diff
@@ -0,0 +1,39 @@
...
+    fi
+}
+
+rc_check() {
+    pgrep -f "${daemon}" >/dev/null
+}
+
+rc_stop() {
+    echo -n "Stopping..." | logger -t ${daemon}
+    pkill -f "${daemon}" >/dev/null
```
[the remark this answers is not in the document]
igalic: again, if stop is needed for a functioning / fake service, we can fake it

cloudinit/net/activators.py
```diff
@@ -217,13 +243,15 @@ def bring_down_interface(device_name: str) -> bool:
 # version to encompass both seems overkill at this point
 DEFAULT_PRIORITY = [
     "eni",
+    "ifconfig",
```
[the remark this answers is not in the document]
CodeBleu: I had changed this to be alphabetical, but will  change to be at bottom of list.

sysvinit/openbsd/cloudconfig.tmpl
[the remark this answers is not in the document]
CodeBleu: I will test removing this and see how it responds without setting it.  Even though it is set to 1, it returns an (ok) from the start, but because the daemon_bg is set, the process still runs in the background and I'm able to verify this with running a check of the service.  The check returns (ok) while the service is running and as soon as it completes, the check of the service returns (failed).

sysvinit/openbsd/cloudconfig.tmpl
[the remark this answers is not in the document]
CodeBleu: Yes, as I mentioned in IRC, the service files are not being used currently, but were kept in for ability to manually run them if you wanted.

tools/build-on-openbsd
CodeBleu: It doesn't need to wait to execute another command, as that is the last command.  See above comments for more explanation.

cloudinit/net/ifconfig.py
```diff
@@ -0,0 +1,13 @@
+# This file is part of cloud-init. See LICENSE file for license information.
```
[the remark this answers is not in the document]
CodeBleu: Since I'm new to this, I just kept the same boilerplate as I thought it might be needed and would already be there if more things were needed.  This was a quick thing I did, just to get it working.  I can look into reducing the boilerplate though.

tools/build-on-openbsd
```diff
@@ -16,12 +22,45 @@ pkgs="
...
+    python3 setup.py build
+    python3 setup.py install -O1 --distro openbsd --skip-build --init-system sysvinit_openbsd
 
-python3 setup.py build
-python3 setup.py install -O1 --distro openbsd --skip-build
+    echo "Installation completed."
+    RC_LOCAL="/etc/rc.local"
+    RC_LOCAL_CONTENT="
+/usr/lib/cloud-init/ds-identify
+wait
```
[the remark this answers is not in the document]
holmanb: Wait is for background tasks, right? I would understand this if the rc script was itself calling `/etc/rc.d/cloud-init start` (not sure if that's how rc services get invoked, but I think you know what I mean) directly, since that would put the script in the background, but since cloud-init executes syncronously in the foreground I don't think this should be necessary. At least that's what I would expect from bash / posix shell.

sysvinit/openbsd/cloudconfig.tmpl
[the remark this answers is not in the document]
holmanb: +1 I'll resolve this conversation

cloudinit/net/ifconfig.py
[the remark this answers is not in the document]
holmanb: Thanks. If the ifconfig definition were more complicated I'd agree with this choice, but since it's just one short function no need to add the extra file. Just move the contents of the `available()` function here into the `IfConfigActivator.available()` and remove this file and we should be good. 

> Since I'm new to this

No worries :) this is looking like a great early contribution.

sysvinit/openbsd/cloudconfig.tmpl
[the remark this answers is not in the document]
holmanb: > I will test removing this and see how it responds without setting it. Even though it is set to 1, it returns an (ok) from the start, but because the rc_bg is set, the process still runs in the background and I'm able to verify this with running a check of the service. The check returns (ok) while the service is running and as soon as it completes, the check of the service returns (failed).

Right, that's what I'm not so clear on - whether the `daemon_timeout` applies to the daemon in the background (within the call to `rc_exec`) or to the process of putting the daemon in the background (within the call to `rc_start`). If the former, I think this is a problem, but if the latter, then I would think this is fine (so long as cloud-init can keep running in the background after `rc_start` exits).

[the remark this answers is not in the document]
CodeBleu: All tests seemed fine with removing `daemon_timeout` and `daemon_rtable`

holmanb (approved):
Looks good to me, thanks for this contribution @CodeBleu!

PR #4672, opened 2023-12-07
tests/unittests: add a new unit test for network manager net activator

ani-sinha:
Some changes in behavior in network manager net activator was brought in with the commit
d1d5166895da ("net/nm: check for presence of ifcfg files when nm connection files are absent")

This change adds some unit tests that exercizes network manager activator's bring_up_interface() method that tests failure scenarios as well as cases where an ifcfg file is used to bring the interface up.

### DIFF
--- a/tests/unittests/test_net_activators.py
+++ b/tests/unittests/test_net_activators.py
@@ -1,3 +1,4 @@
+import os
 from collections import namedtuple
 from contextlib import ExitStack
 from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


ani-sinha:
@TheRealFalcon sorry I should perhaps have pushed these unit tests along with my previous change. But here it goes now.

TheRealFalcon (approved):
LGTM! Thanks!

PR #4673, opened 2023-12-07
feat(disk_setup) add timeout

### LINKED ISSUE #3386: fs_setup/disk_setup: option to wait for the device to exist before continuing
This bug was originally filed in Launchpad as [LP: #1832645](https://bugs.launchpad.net/cloud-init/+bug/1832645)
<details>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2019-06-12T20:48:53.539989+00:00
date_fix_committed = None
date_fix_released = None
id = 1832645
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1832645
milestone = None
owner = minfrin-y
owner_name = Graham Leggett
private = False
status = in_progress
submitter = minfrin-y
submitter_name = Graham Leggett
tags = []
duplicates = [1907080]
</pre>
</details>

_Launchpad user **Graham Leggett(minfrin-y)** wrote on 2019-06-12T20:48:53.539989+00:00_

When using the AWS::EC2::Volume and AWS::EC2::VolumeAttachment options to add a volume to an AWS::EC2::Instance on AWS EC2, the volume is not immediately available on the instance.

This causes fs_setup and disk_setup to fail.

What would prevent this failure is a "wait" option on both fs_setup and disk_setup, which if true, will cause cloud-init to wait until the device exists (caused by AWS catching up and attaching the device) before continuing.

flokli:
## Proposed Commit Message
```
In a cloud environment, sometimes disks will attach while cloud-init is running and get missed. This adds a configurable timeout to wait for those disks.

Fixes GH-3386
LP: #1832645
```

## Additional Context
This is https://github.com/canonical/cloud-init/pull/710 resurrected.
Fixes #3386.

## Test Steps
Tested with the following cloud-config file on azure:

```yaml
disk_setup:
  /dev/disk/by-lun/10:
    layout: false
    timeout: 60

fs_setup:
 - filesystem: ext4
   partition: auto
   device: /dev/disk/by-lun/10
   label: jenkins

mounts:
 - ["/dev/disk/by-label/jenkins", "/var/lib/jenkins"]
```

I used a `azurerm_virtual_machine_data_disk_attachment` terraform resource, which attaches the disk halfway during bootup, due to https://github.com/hashicorp/terraform-provider-azurerm/issues/6117, but this should also work with setting a reasonably large timeout and manually attaching the disk (at that lun)

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_disk_setup.py
+++ b/cloudinit/config/cc_disk_setup.py
@@ -131,6 +131,13 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
                 LOG.warning("Invalid disk definition for %s", disk)
                 continue
 
+            if definition.get("timeout"):
+                tmout = int(definition.get("timeout", "0"))
+                missing = util.wait_for_files(
+                    [disk], maxwait=tmout, naplen=0.1
+                )
+                if len(missing) > 0:
+                    LOG.warning("Timeout expired waiting for %s", disk)
             try:
                 LOG.debug("Creating new partition table/disk")
                 util.log_time(
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -1261,6 +1261,11 @@
                   "type": "boolean",
                   "default": false,
                   "description": "Controls whether this module tries to be safe about writing partition tables or not. If ``overwrite: false`` is set, the device will be checked for a partition table and for a file system and if either is found, the operation will be skipped. If ``overwrite: true`` is set, no checks will be performed. Using ``overwrite: true`` is **dangerous** and can lead to data loss, so double check that the correct device has been specified if using this option. Default: ``false``"
+                },
+                "timeout": {
+                  "type": "integer",
+                  "default": 0,
+                  "description": "Seconds to wait for a disk to appear. If omitted, no wait is performed. This is useful if a cloud provider has late or inconsistent disk attachments and you need a delay."
                 }
               }
             }
--- a/doc/examples/cloud-config-disk-setup.txt
+++ b/doc/examples/cloud-config-disk-setup.txt
@@ -98,6 +98,7 @@ disk_setup:
 #       table_type: 'mbr'
 #       layout: <LAYOUT|BOOL>
 #       overwrite: <BOOL>
+#       timeout: <INT>
 #
 # Where:
 #   <DEVICE>: The name of the device. 'ephemeralX' and 'swap' are special
@@ -159,6 +160,11 @@ disk_setup:
 #               'true' is cowboy mode. There are no checks and things are
 #                   done blindly. USE with caution, you can do things you
 #                   really, really don't want to do.
+#
+#   timeout=<INT>: Seconds to wait for a disk to appear. If omitted, no wait
+#               is performed. This is useful if a cloud provider has late or
+#               inconsistent disk attachments and you need a delay.
+#
 #
 #
 # fs_setup: Setup the filesystem
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -54,6 +54,7 @@ einsibjarni
 emmanuelthome
 eslerm
 esposem
+flokli
 frantisekz
 GabrielNagy
 garzdin


holmanb:
Ugh. What a mess.

The code might work, but I can't say that I like it.
- timeouts -> unreliable, ugly
- this makes cloud-init block remaining cloud-init runtime (and all of boot) until the disk is ready

Plus, if you're using a systemd-based instance, you should be able to accomplish this already using systemd's builtin mount point generation from fstab. Something like this should do it:

```yaml
mounts:
 - ["/dev/disk/by-label/jenkins", "/var/lib/jenkins", "ext4", "defaults,nofail,x-systemd.makefs"]
```

Since systemd creates mount units from fstab, and mount units get automatically ordered, I think the above should "just work". 

At least, that would work if cloud-init's `cc_mount` module didn't assume that mounts need to be available when `cloud-init.service` is running in order for them to get written into `/etc/fstab`.

Thanks for proposing this @flokli. This might be an interim solution that we can take, but I think a more elegant solution is desirable long term.

flokli:
Using `systemd-fstab-generator` unfortunately won't work at all, no - it's the approach I tried before.

The fstab line creates a `.mount` unit which will depend on a `.device` unit that doesn't exist while systemd plans the transaction, because there's no corresponding `/dev/disk/by-lun/10` while the disk is not attached yet - so all units with `RequiresMountsFor=` - either explicit or implicit via `StateDirectory` will fail due to a missing dependency.

This is not an issue on the second boot, or if I re-attempt to reach mutli-user.target (or just restart the .service unit using the data disk), but the only way to get past this was to do mounting via cloud-init, not systemd.

Let me know if you'd be fine to accept this, happy to then take another pass and fix the linters.

holmanb:
> The fstab line creates a .mount unit which will depend on a .device unit that doesn't exist while systemd plans the transaction, because there's no corresponding /dev/disk/by-lun/10 while the disk is not attached yet - so all units with RequiresMountsFor= - either explicit or implicit via StateDirectory will fail due to a missing dependency.

You're right that depending on only the generator output wouldn't work for first boot. I didn't consider initial transaction failure due to missing device unit.

However, the .device file should later be generated by systemd-udev[1], which will cause subsequent transaction re-calculation to succeed. I think that this is why restarting the .service unit and second boot and re-attempting multi-user all worked. I think that `systemctl daemon-reload` should also do that anytime after the .device file is created.

An alternative that could take advantage of better systemd ordering would be to use `ENV{SYSTEMD_WANTS}+="your-fs.mount"`[2] in a udev rule that matches on the specified block device. This seems like something that `systemd-fstab-generator` could feasibly do: a special mount option with a name like `x-systemd.first-boot` or `x-systemd.new-device` which generates a udev rule for the device which activates the .mount when available - the rule could be disabled / deleted if the .device file already exists.

> This is not an issue on the second boot, or if I re-attempt to reach mutli-user.target (or just restart the .service unit using the data disk), but the only way to get this working straight from the first boot was to do mounting via cloud-init, not systemd.

[1] [systemd.device](https://www.freedesktop.org/software/systemd/man/latest/systemd.device.html)
> systemd will dynamically create device units for all kernel devices that are marked with the "systemd" udev tag (by default all block and network devices, and a few others). 

[2] [systemd-fstab-generator](https://www.freedesktop.org/software/systemd/man/latest/systemd.device.html#SYSTEMD_WANTS=)
> These properties may be used to activate arbitrary units when a specific device becomes available.

holmanb:
Your proposal should be easy to review, so for now if you want to move forward with this approach, please make sure to add the new key to the jsonschema.

flokli:
I added the field to the jsonschema, addressed the linter warning and renamed `need` to `missing` to be more consistent with the other users of this method. PTAL.

holmanb:
@flokli I don't see your username `flokli` connected to any users that have [signed the CLA](https://ubuntu.com/legal/contributors). Could you please do that so that we can accept this contribution to cloud-init?

flokli:
Sorry, please check again, just did it.

holmanb (approved):
The LP reference is superseded by the github bug, so I'm going to drop that from the message.

Thanks for this contribution @flokli, and welcome to cloud-init!

holmanb:
@flokli Sorry about this one more thing. I just noticed that commit 9b2e3dc907dc06d0a2abdaae has @nibalizer as the author, but @flokli's DCO. This project requires CLA, not DCO. Whose code is this, @flokli's or @nibalizer's? If @nibalizer, I need you to sign the DCO please.

Apologies for the hold up on this. I wish this wasn't part of the process but unfortunately it is.

flokli:
Correct, the changes were resurrected from @nibalizer's PR, that's mentioned in the PR description:

> This is https://github.com/canonical/cloud-init/pull/710 resurrected.

nibalizer:
> This project requires CLA, not DCO. Whose code is this, @flokli's or @nibalizer's? If @nibalizer, I need you to sign the DCO please.

I assume you need CLA not DCO.

I wrote this while at IBM. At the time I don't think IBM signed a CCLA with Canonical. I no longer work there. Is it possible we could get this in under "trivial commit" ?

cc @jjasghar @bradtopol

flokli:
I realized you actually can change the timeout of .device by using the [`x-systemd.device-timeout` mount option](https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html#x-systemd.device-timeout=) in your fstab line, or the [`DefaultDeviceTimeout=` `systemd.conf` setting](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#DefaultDeviceTimeoutSec=), so I might not be needing this workaround after all, at least for my specific usecase.

holmanb:
> I realized you actually can change the timeout of .device by using the [`x-systemd.device-timeout` mount option](https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html#x-systemd.device-timeout=) in your fstab line, or the [`DefaultDeviceTimeout=` `systemd.conf` setting](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#DefaultDeviceTimeoutSec=), so I might not be needing this workaround after all, at least for my specific usecase.

Let me know if that works. It looks like the existence of the .device unit [is required](https://github.com/systemd/systemd/blob/main/src/shared/generator.c#L389) for this command (much like `x-systemd.makefs` that I pointed out above), so I'm guessing that this won't help since the transaction will still fail to include the mount unit because the device unit still won't exist on first boot.

holmanb:
> I assume you need CLA not DCO.

Yes, that's what I meant.

> I opened this PR while at IBM. At the time I don't think IBM signed a CCLA with Canonical. I no longer work there.Is it possible we could get this in under "trivial commit" ? Or perhaps all this code is authored by @flokli who can sign the CLA.

I'll check internally, but I'm not sure that there is much that we can do unless we can get a CLA signature from IBM.

> cc @jjasghar @bradtopol

Any insight would be appreciated.

flokli:
> > I realized you actually can change the timeout of .device by using the [`x-systemd.device-timeout` mount option](https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html#x-systemd.device-timeout=) in your fstab line, or the [`DefaultDeviceTimeout=` `systemd.conf` setting](https://www.freedesktop.org/software/systemd/man/latest/systemd-system.conf.html#DefaultDeviceTimeoutSec=), so I might not be needing this workaround after all, at least for my specific usecase.
> 
> @flokli Let me know if that works. It looks like the existence of the .device unit [is required](https://github.com/systemd/systemd/blob/main/src/shared/generator.c#L389) for this command (much like `x-systemd.makefs` that I pointed out above), so I'm guessing that this won't help since the transaction will still fail to include the mount unit because the device unit still won't exist on first boot.

You need to follow one more reference ;-) `is_device_path` just matches the passed string with some known-good path strings: https://github.com/systemd/systemd/blob/1fdab6c3069ab945259d70e22a29e80da8370288/src/basic/path-util.c#L1285

The approach using `x-systemd.device-timeout` worked in principle - but I then ran into the issue that sometimes the `azurerm_linux_virtual_machine` resource /itself/ blocked  returninguntil the machine finished booting up - which it never does if it waits for the block device to be attached, which only happens after the resource returns :see_no_evil: 

In the end I gave up on this journey, and resorted to the `azurerm_virtual_machine` resource, which does support attaching data disks to VMs at creation time.

I assume keeping systemd about the mountpoints in the dark, and all the waiting / filesystem creation / mounting in cloud-init would also work, so I still think this PR is useful to have.

holmanb:
> You need to follow one more reference ;-) is_device_path just matches the passed string with some known-good path strings:

you're right again :P

> The approach using x-systemd.device-timeout worked in principle - but I then ran into the issue that sometimes the azurerm_linux_virtual_machine resource /itself/ blocked returninguntil the machine finished booting up - which it never does if it waits for the block device to be attached, which only happens after the resource returns 🙈

oof

> In the end I gave up on this journey, and resorted to the azurerm_virtual_machine resource, which does support attaching data disks to VMs at creation time.

:+1:, glad you got something figured out for your use case

> I assume keeping systemd about the mountpoints in the dark, and all the waiting / filesystem creation / mounting in cloud-init would also work, so I still think this PR is useful to have.

Agreed

holmanb (changes requested):
Setting this to changes requested for now. Conversations with legal don't appear to be moving quickly, but this is still something we'd like to add.

flokli:
> > This project requires CLA, not DCO. Whose code is this, @flokli's or @nibalizer's? If @nibalizer, I need you to sign the DCO please.
> 
> I assume you need CLA not DCO.
> 
> I opened this PR while at IBM. At the time I don't think IBM signed a CCLA with Canonical. I no longer work there. Is it possible we could get this in under "trivial commit" ? Or perhaps all this code is authored by @flokli who can sign the CLA.
> 
> cc @jjasghar @bradtopol

This is PR is stuck on this. I don't work for IBM, and didn't plan to. If this can't be considered a trivial contribution that doesn't need a CLA, someone at canonical needs to convince IBM to sign this.

Issue #4642, opened 2023-11-29
[wishlist] Resolve pyright errors


holmanb:
## Current use of static type checkers

Cloud-init uses type annotations with Python static analyzers to check for typing errors. This increases code quality, and thereby the confidence that new releases of cloud-init will not experience unexpected runtime Python typing exceptions. Currently cloud-init includes a [mypy ci job](https://github.com/canonical/cloud-init/blob/cd35cadc5cbd952ccd17932688ff9dfef2a160c7/.github/workflows/check_format.yml#L20) to validate new contributions.

## Pyright: a more verbose alternative

[Pyright](https://github.com/microsoft/pyright) provides an alternative static type checker implementation. This checker (or sometimes Microsoft's proprietary version Pylance) is often used by developers as an LSP server (with VSCode, neovim, etc) to assist with development. However, it is much more verbose; Pyright reports many potential issues which mypy does not report. 

## Request: resolve (fix or silence) currently reported errors

This request is to resolve currently reported errors and optionally generate a configuration for Pyright which silences errors that are not helpful. Some of the reported errors may represent potential errors, and many are expected to be code paths that simply do not occur, or false positives.

Resolving this bug is expected to increase the quality of cloud-init's code, increase value of Pyright / Pylance to developers by increasing the signal to noise ratio, and additionally make it possible to run Pyright as a continuous integration test[1].

Slightly over half of Pyright errors occur in test code. Fixing test code is a lower priority than fixing runtime code (code that lives in `cloudinit/`). Many errors may be solved without deep understanding of cloud-init's code base. This is a good first issue that would be an excellent way for a new developer to become familiar with the code base.

## Steps to reproduce the problem
```
$ git clone cloud-init
$ cd cloud-init
$ python3 -m venv venv
$ . venv/bin/activate
$ pip install pyright
$ pyright cloudinit/    # runtime code (fix these first)
$ pyright tests/        # test code
```

[1] continuous integration testing is not an explicit goal of this issue - whether that is desirable is yet to be determined

gajeshbhat:
Hey folks, put up a Pull Request to address some of the errors in this issue. Can you kindly review it ? https://github.com/canonical/cloud-init/pull/6358 CC: @holmanb @aciba90 @TheRealFalcon Please let me know if you have any questions.

gajeshbhat:
I attempted to address this issue with PR #6358, but closed it after feedback. 

[...]

PR #4675, opened 2023-12-08
Fix some issues related to (canonical#4642)

AlexSv04047:
Before changes:
```
pyright cloudinit/ -> 619 errors, 30 warnings
```
After changes:
```
pyright cloudinit/ -> 590 errors, 30 warnings
```


AlexSv04047:
Any feedback about the PR is appreciated

holmanb:
@AlexSv04047 we have a line length requirement of 79 characters which gets automatically checked. Some of the changes here violate that, which makes some of our other linters unhappy (see the failing ruff check). Also it looks like you might be mixing tabs and spaces?

AlexSv04047:
> @AlexSv04047 Thanks for jumping on this! Really appreciate the initiative - help reducing the number of pyright warnings is appreciated.
> 
> We have a line length requirement of 79 characters which gets automatically checked. Some of the changes here violate that, which makes some of our other linters unhappy (see the failing ruff check in Github Actions). Also it looks like you might be mixing tabs and spaces?
> 
> See if you can get Github Actions tests happy then I can take another look.

Got it, I'll try

AlexSv04047:
Okay so I reverted the changes in netinfo.py, now it only fixes 3 errors (which is a little sad 😭) :
```bash
dmi.py:118:46 - error: "cmd" is possibly unbound (reportUnboundVariable)
dmi.py:137:51 - error: "cmd" is possibly unbound (reportUnboundVariable)
gpg.py:44:61 - error: Argument of type "Literal[False]" cannot be assigned to parameter "decode" of type "str" in function "subp"
```
Will be trying to fix more errors daily, I'll probably PR every time I fix around 50/70 of them if that's okay?

AlexSv04047:
> > I'll probably PR every time I fix around 50/70 of them if that's okay?
> 
> Sounds fine by me. Feel free to give a ping when you get there on this one :)

So I'm getting `618 errors, 30 warnings` before changes, `449 errors, 30 warnings` after changes. 😎
Fixed one small error (I think it's an actual error?) and suppressed some false positives.

There should be some actual errors in `netinfo.py`, specifically in the `_netdev_info_ifconfig(ifconfig_data)` function.
I'd like to refactor the function but I would need a sample of `ifconfig_data` and a description of what the function should do?

Also to ignore specific lines from appearing as false positive errors in pyright I tried `# type: ignore`.
It appears there's a check for it from `mypy` in pyproject.toml -> `warn_unused_ignores = "true"`, so I'm guessing I shouldn't use it?

Pyright says {variable} is possibly unbound, during try/except blocks, but try: doesn't create it's own scope.
Do I modify the code to "fix" those errors or do I set up a configuration (without using `# type: ignore`) to silence them?
Sometimes the variable could potentially be unbound in the except block but in this case that won't happen.

AlexSv04047:
@holmanb Not sure if quoted reply automatically pings, so I'll ping just in case 😎

AlexSv04047:
@holmanb 
I'll close this PR and work on issue #4642, if I manage to come up with a nice configuration and fix all the errors I'll send a new PR.
Sorry for the spam / questions, it's my first contribution.

holmanb:
> @holmanb 
> I'll close this PR and work on issue #4642, if I manage to come up with a nice configuration and fix all the errors I'll send a new PR.

@AlexSv04047 I hadn't had time yet to review this, busy with other things but I was planning to get to it. It's up to you whether you want to do this piece-meal or all at once. Though for a change this large as a first-time contribution if you want to do it all at once I would recommend maybe submitting a WIP PR once you have a general approach sketched out so that you don't waste your own time on an approach that isn't the best approach for the project.

I will suggest (and i was going to request this with this PR as you had it) that any suppressions that get added should probably have an explanation for why it is unnecessary and won't catch issues with the code - I'm hesitant to get to "pyright 0 errors" via suppressions, unless those suppressed errors really won't help us.

> Sorry for the spam / questions, it's my first contribution.

Not a problem. And if you have questions that you expect to be easier to find in the docs, that's an area that we're always trying to improve too. New user/contributor feedback is valuable, so don't hesitate to ask if you have a question the docs don't answer well (in an issue on github or #cloud-init on libera).

PR #4677, opened 2023-12-08
fix(runpath): On *BSD, create /run is not ephemeral

igalic:
## Proposed Commit Message

```
fix(runpath): On *BSD, /run is not ephemeral

On *BSD, `/run` is not ephemeral, but the way cloud-init behaves, it
expects it to be. To fix this, we ds-identify's RUN_* initialization after
figuring out which OS we run on.
For the Python we set the config to the expected directories.

Add and fix tests to ensure this behavior is correct and consistent.

Fixes GH-4180
Fixes GH-4231
```

## Checklist

## Merge type

### DIFF
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -18,16 +18,14 @@ from tests.unittests.helpers import (
     populate_dir_with_ts,
 )
 
-UNAME_MYSYS = (
-    "Linux 4.4.0-62-generic #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64"
-)
+UNAME_MYSYS = "Linux #83-Ubuntu SMP Wed Jan 18 14:10:15 UTC 2017 x86_64"
 UNAME_PPC64EL = (
-    "Linux 4.4.0-83-generic #106-Ubuntu SMP "
-    "Mon Jun 26 17:53:54 UTC 2017 "
+    "Linux #106-Ubuntu SMP mon Jun 26 17:53:54 UTC 2017 "
     "ppc64le ppc64le ppc64le"
 )
 UNAME_FREEBSD = (
-    "FreeBSD 12.1-RELEASE-p10 FreeBSD 12.1-RELEASE-p10 GENERIC  amd64"
+    "FreeBSD FreeBSD 14.0-RELEASE-p3 releng/14.0-n265398-20fae1e1699"
+    "GENERIC-MMCCAM amd64"
 )
 UNAME_OPENBSD = "OpenBSD GENERIC.MP#1397 amd64"
 
@@ -340,7 +338,17 @@ class DsIdentifyBase(CiTestCase):
             },
         ]
 
-        written = [d["name"] for d in mocks]
+        uname = "Linux"
+        runpath = "run"
+        written = []
+        for d in mocks:
+            written.append(d["name"])
+            if d["name"] == "uname":
+                uname = d["out"].split(" ")[0]
+        # set runpath so that BSDs use /var/run rather than /run
+        if uname != "Linux":
+            runpath = "var/run"
+
         for data in mocks:
             mocklines.append(write_mock(data))
         for d in default_mocks:
@@ -363,7 +371,7 @@ class DsIdentifyBase(CiTestCase):
             err = e.stderr
 
         cfg = None
-        cfg_out = os.path.join(rootd, "run/cloud-init/cloud.cfg")
+        cfg_out = os.path.join(rootd, runpath, "cloud-init/cloud.cfg")
         if os.path.exists(cfg_out):
             contents = util.load_file(cfg_out)
             try:
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -68,7 +68,6 @@ DI_DISABLED="disabled"
 DI_DEBUG_LEVEL="${DEBUG_LEVEL:-1}"
 
 PATH_ROOT=${PATH_ROOT:-""}
-PATH_RUN=${PATH_RUN:-"${PATH_ROOT}/run"}
 PATH_SYS_CLASS_DMI_ID=${PATH_SYS_CLASS_DMI_ID:-${PATH_ROOT}/sys/class/dmi/id}
 PATH_SYS_HYPERVISOR=${PATH_SYS_HYPERVISOR:-${PATH_ROOT}/sys/hypervisor}
 PATH_SYS_CLASS_BLOCK=${PATH_SYS_CLASS_BLOCK:-${PATH_ROOT}/sys/class/block}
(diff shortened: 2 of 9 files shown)


sysvinit/freebsd/dsidentify.tmpl
```diff
@@ -17,10 +17,19 @@ PATH="{{ prefix }}/sbin:{{ prefix }}/bin:/usr/sbin:/usr/bin:/sbin:/bin"
...
+start_precmd="dsidentify_precmd"
 start_cmd="dsidentify_start"
 stop_cmd=":"
 rcvar="cloudinit_enable"
 
+dsidentify_precmd()
+{
+    if ! test -L /run ; then
+       rm -rf /run
+       ln -s /var/run /run
```
holmanb, line 29: The symlink won't ever be created in the case that /run does not yet exist, since it happens inside of this conditional. Isn't this conditional a workaround for images that have already ran `ds-identify` which created this directory?

In other words, don't we want `/run` to be created as a symlink to `/var/run/` on first boot when no `/run/` exists yet?

I'm thinking that we want something more akin to:
```sh
link_run_to_ephemeral()
{
    if test -L /run; then
       return 0
    elif test -e /run; then
       rm -rf /run
    fi 
    ln -s /var/run /run
}
```

question applies to elsewhere too

igalic: from http://man.openbsd.org/test.1#L (using OpenBSD because it's easiest to link exactly where i want to, they all say the same)

> -L file
>    True if file exists and is a symbolic link. This operator is for compatibility purposes. Do not rely on its existence; use -h instead.

sysvinit/freebsd/dsidentify.tmpl
```diff
@@ -17,10 +17,19 @@ PATH="{{ prefix }}/sbin:{{ prefix }}/bin:/usr/sbin:/usr/bin:/sbin:/bin"
...
 command="{{ prefix }}/lib/cloud-init/ds-identify"
+start_precmd="dsidentify_precmd"
 start_cmd="dsidentify_start"
 stop_cmd=":"
 rcvar="cloudinit_enable"
 
+dsidentify_precmd()
+{
+    if ! test -L /run ; then
+       rm -rf /run
```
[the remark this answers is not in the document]
holmanb: +1 - feel free to iterate here (ping when ready for review) or open a new one. Either works for me.

igalic (commented):
this is some of the best (harmless) shell code I've written yet

tests/unittests/test_ds_identify.py
```diff
@@ -363,8 +366,12 @@ def write_mock(data):
             out = e.stdout
             err = e.stderr
 
+        runpath = "run"
```
holmanb, line 369: nit: Could we set this variable near the other changes and include a comment explaining what it is for?

```diff
diff --git a/tests/unittests/test_ds_identify.py b/tests/unittests/test_ds_identify.py
index 4e8126ba6..1bc9dc42a 100644
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -340,11 +340,16 @@ class DsIdentifyBase(CiTestCase):
         ]
 
         uname = "Linux"
+        runpath = "run"
         written = []
         for d in mocks:
             written.append(d["name"])
             if d["name"] == "uname":
                 uname = d["out"].split(" ")[0]
+        # set runpath so that BSDs use /var/run rather than /run
+        if uname != "Linux":
+            runpath = "var/run"
+
         for data in mocks:
             mocklines.append(write_mock(data))
         for d in default_mocks:
@@ -366,10 +371,6 @@ class DsIdentifyBase(CiTestCase):
             out = e.stdout
             err = e.stderr
 
-        runpath = "run"
-        if uname != "Linux":
-            runpath = "var/run"
-
         cfg = None
         cfg_out = os.path.join(rootd, runpath, "cloud-init/cloud.cfg")
         if os.path.exists(cfg_out):
```

igalic: yeah that reads more sensible

tools/ds-identify
```diff
@@ -1579,6 +1581,10 @@ collect_info() {
 }
 
 print_info() {
+    DI_LOG=stderr
```
igalic, line 1584: my understanding is that this is correct
although it didn't seem to be necessary before

holmanb: Why do we need the `DI_LOG=stderr` and `ensure_sane_path` calls?

I think that I see why `read_uname_info` and `set_run_path` might be required - this function is the other "main" entry point of this function, thus will require the `PATH_RUN*` variables to be properly set, as they were before this function.

tools/ds-identify
```diff
@@ -82,11 +81,14 @@ PATH_PROC_UPTIME=${PATH_PROC_UPTIME:-${PATH_ROOT}/proc/uptime}
 PATH_ETC_CLOUD="${PATH_ETC_CLOUD:-${PATH_ROOT}/etc/cloud}"
 PATH_ETC_CI_CFG="${PATH_ETC_CI_CFG:-${PATH_ETC_CLOUD}/cloud.cfg}"
 PATH_ETC_CI_CFG_D="${PATH_ETC_CI_CFG_D:-${PATH_ETC_CI_CFG}.d}"
-PATH_RUN_CI="${PATH_RUN_CI:-${PATH_RUN}/cloud-init}"
-PATH_RUN_CI_CFG=${PATH_RUN_CI_CFG:-${PATH_RUN_CI}/cloud.cfg}
-PATH_RUN_DI_RESULT=${PATH_RUN_DI_RESULT:-${PATH_RUN_CI}/.ds-identify.result}
 
-DI_LOG="${DI_LOG:-${PATH_RUN_CI}/ds-identify.log}"
+# Declare global here, we'll set it properly in main()
```
holmanb, line 85: Lets add a comment explaining that the default values for these variables are system-specific, therefore must be assigned after collecting some information from the system.

tests/unittests/test_ds_identify.py
```diff
@@ -339,11 +339,16 @@ def write_mock(data):
         ]
```
holmanb, line 339: I'm fine with enabling the config file change in a separate commit if you'd rather, but we shouldn't change `test_ds_identify.py` in adjacent commits. Could you please rebase the changes in this file into the previous commit please? Or if you want this one squashed into the last that would work too - up to you.

Also no need to add me as a co-author.

holmanb (changes requested):
Great work here @igalic, I like the way this one is coming together. I have a couple smallish requests / questions, once resolved I'll merge.

tests/unittests/test_ds_identify.py
igalic: done.

tools/ds-identify
igalic: here's what happens without `DI_LOG=stderr`:

```
tools/ds-identify: 185: 3: Bad file descriptor
tools/ds-identify: 185: 3: Bad file descriptor
```

igalic: but `set_run_path` is not, so i removed that.

holmanb: I meant the other way around:
```diff
diff --git a/tools/ds-identify b/tools/ds-identify
index 6e923a550..154668a00 100755
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -1584,8 +1584,7 @@ collect_info() {
 }
 
 print_info() {
-    DI_LOG=stderr
-    ensure_sane_path
+    set_run_path
     read_uname_info
     collect_info
     _print_info
```

since `set_run_path` sets DI_LOG, we shouldn't unconditionally log to stderr, in case of `DI_MAIN=print_info DI_LOG=/some/file tools/ds-identify` or similar.

The above didn't have the bad descriptor you mentioned.

tools/ds-identify
```diff
@@ -1835,12 +1843,25 @@ read_uptime() {
     return
 }
 
+set_run_path() {
+    if [ "$DI_UNAME_KERNEL_NAME" != "Linux" ]; then
```
igalic, line 1847: make sure this function works on Linux… even if someone forgot to call `read_uname_info()`…

tools/ds-identify
```diff
@@ -154,6 +159,8 @@ debug() {
     shift
     [ "$lvl" -gt "${DI_DEBUG_LEVEL}" ] && return
 
+    [ "$DI_LOG" = "" ] && DI_LOG="stderr"
```
igalic, line 162: make sure the `debug()` function works, even if someone forgot to `set_run_path()`, or `DI_LOG`.

igalic (commented):
i think this is done now…

tools/ds-identify
holmanb: Mentally walking through this:
1. call to `debug` before `set_run_path` causes `DI_LOG` set to `stderr`, `_DI_LOGGED` set to `stderr`, file descriptor 3 redirects to file descriptor 2
2. `set_run_path` redefines `DI_LOG` to `whatever/cloud-init.log`
3. subsequent call to `debug` sees that `"$_DI_LOGGED" != "$DI_LOG"`, re-opens fd 3 pointing to `whatever/cloud-init.log`, reassigns `_DI_LOGGED` to `$DI_LOG`
4. subsequent calls to `debug` use fd 3 which points to the right file

Looks good to me!

holmanb (approved):
Great work @igalic. This looks good to me.

PR #4678, opened 2023-12-08
ci: Run all tests always

holmanb:
## Additional Context
See proposed commit message for details.

## Merge type

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -13,26 +13,23 @@ defaults:
 jobs:
   unittests:
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         toxenv: [py3]
         experimental: [false]
         check-latest: [false]
-        continue-on-error: [false]
         include:
           - python-version: "3.6"
             toxenv: lowest-supported
-            continue-on-error: false
             check-latest: false
             experimental: false
           - python-version: "3.13-dev"
             toxenv: py3
             check-latest: true
             experimental: true
-            continue-on-error: true
     name: unittest / ${{ matrix.toxenv }} / python ${{matrix.python-version}}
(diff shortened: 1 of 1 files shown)


aciba90 (changes requested):
+1 on this change, thanks.

The `experimental` variable is no longer used anywhere after this change, could we please remove it?

```patch
diff --git a/.github/workflows/unit.yml b/.github/workflows/unit.yml
index 8981af56a..b1ae16d25 100644
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -17,17 +17,14 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
         toxenv: [py3]
-        experimental: [false]
         check-latest: [false]
         include:
           - python-version: "3.6"
             toxenv: lowest-supported
             check-latest: false
-            experimental: false
           - python-version: "3.13-dev"
             toxenv: py3
             check-latest: true
-            experimental: true
     name: unittest / ${{ matrix.toxenv }} / python ${{matrix.python-version}}
     runs-on: ubuntu-20.04
     steps:
```

TheRealFalcon:
From your commit message:
```
First time contributors might not know how to run tests locally, and forcing
them to iterate on formatting issues is senseless if the same code is causing
integration failures.
```

That's not what's happening here though. The `continue-on-error` is only going to apply to the rest of the jobs in the workflow. This workflow only runs unit tests. It means e.g., if unit tests on 3.9 fail, then stop running the unit tests for 3.10 and any other versions running. AFAIK, lint checks aren't blocking unit tests or integration tests and vice versa. Given how rare it is that unit tests will fail on different python versions for different reasons, I don't really see value in this change. I think it's more cumbersome seeing 6 different failures that you need to click into only to have the same failure show up in all 6 runs.

holmanb:
@TheRealFalcon hmmm, looks like you're right

Issue #4676, opened 2023-12-08
Failure to launch 22.04 LXD container with cloud-init 23.3.3


skatsaounis:
# Bug report

During MAAS daily system-tests, we launch a 22.04 LXD container where we install MAAS and perform our tests. Starting from midnight yesterday, the container is failing to launch due to an error on this function: https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/status.py#L288-L295

## Environment details
- Cloud-init version: 23.3.3-0ubuntu0~22.04.1
- Operating System Distribution: Ubuntu 22.04
- Cloud provider, platform or installer type: LXD container

## cloud-init logs

```bash
maas_container: ┌ lxc launch -q ubuntu:22.04 -e -c 'user.user-data=#cloud-config\n"package_upgrade": !!bool |-\n  true\n"packages":\n- |-\n  dhcp-probe\n"runcmd":\n- - |-\n    netplan\n  - |-\n    apply\n"snap":\n  "commands":\n  - |-\n    snap set system proxy.http="http://xxx" proxy.https="http://xxx"\n  - |-\n    snap refresh snapd\n  - |-\n    snap install maas-test-db --channel=3.3\n  - |-\n    snap install maas --channel=3.3\n"write_files":\n- "content": |-\n    response_wait_time 10\n  "path": |-\n    /etc/dhcp_probe.cf\n- "content": |\n    network:\n      ethernets:\n        maas-ss-pxe:\n          addresses:\n          - xxx.xxx.xxx.xxx/xx\n      version: 2\n  "path": |-\n    /etc/netplan/99-maas-systemtests.yaml\n' -p maas-system-maas maas-system-maas
maas_container: └ ✔
maas_container: Container maas-system-maas created.
maas_container: Waiting for boot to finish...
maas_container: ┌ timeout 2000 cloud-init status --wait --long
maas_container: |................................................................................................................................................................................................................................................................................................................................................................................................................................…
maas_container: |ETraceback (most recent call last):
maas_container: |E  File "/usr/bin/cloud-init", line 33, in <module>
maas_container: |E    sys.exit(load_entry_point('cloud-init==23.3.3', 'console_scripts', 'cloud-init')())
maas_container: |E  File "/usr/lib/python3/dist-packages/cloudinit/cmd/main.py", line 1080, in main
maas_container: |E    retval = util.log_time(
[...]

PR #4679, opened 2023-12-09
 bug(status): retry systemctl show to avoid dbus disconnect

blackboxsw:
## Proposed Commit Message

```
bug(status): retry systemctl show to avoid dbus disconnect

Fixes GH-4676

## Additional Context
<!-- If relevant -->

## Test Steps
<!-- Please include any steps necessary to verify (and reproduce if
this is a bug fix) this change on a live deployed system,
including any necessary configuration files, user-data,
setup, and teardown. Scripts used may be attached directly to this PR. -->

## Checklist
that apply. -->

## Merge type


holmanb (commented):
This leaves users that want a quick response to `cloud-init status` waiting for dbus. What if we did something different:

In this case, obviously cloud-init isn't in `"done"` or `"degraded done"` state yet, and it can't be in `"not run"` or `"disabled"` either, given the check before the call to `_get_systemd_status()` so it's either `"error"` or `"running"` or `"degraded running"`. Since this function only returns `"running"` or `"error"` (`"degraded"` gets assigned later), and the absence of dbus means that we cannot introspect to discover an `"error"`, I think that it would be reasonable for this function to return `"running"` when this happens, because to the best of our knowledge, it is running.

Also, looking at the call site of `_get_systemd_status()`, I see no reason that this function should be called if we have already seen that status is set to `UXAppStatus.ERROR`. The presence of an error cannot be proven to be "not an error" by systemd introspection. We already saw that there are errors, so in that case this should just fall through.

TheRealFalcon:
> Also, looking at the call site of _get_systemd_status(), I see no reason that this function should be called if we have already seen that status is set to UXAppStatus.ERROR. The presence of an error cannot be proven to be "not an error" by systemd introspection.

error becomes "not an error" when cloud-init is still running. The reason for that code was that `status --wait` was returning with error before cloud-init ever completed. I should've named it better, but `_get_systemd_status()` can only ever return RUNNING, ERROR, or None (ignore my status).

PR #4681, opened 2023-12-11
fix: Handle systemctl commands when dbus not ready

TheRealFalcon:
## Proposed Commit Message
```
fix: Handle systemctl commands when dbus not ready

During `cloud-init status`, we check systemctl to ensure the status
we're reporting is accurate. However, we can get an error from systemctl
if dbus isn't ready yet. This commit will either ignore the error if we
can assume that cloud-init is still running, or retry until we get a
proper response from systemctl.

Fixes GH-4676
```

## Additional Context
https://github.com/canonical/cloud-init/issues/4676

## Merge type

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -337,7 +337,6 @@ def _get_error_or_running_from_systemd_with_retry(
     then we should retry on systemd status so we don't incorrectly report
     error state while cloud-init is still running.
     """
-    last_exception = subp.ProcessExecutionError
     while True:
         try:
             return _get_error_or_running_from_systemd()
@@ -435,7 +434,7 @@ def get_status_details(
         UXAppStatus.DISABLED,
     ):
         systemd_status = _get_error_or_running_from_systemd_with_retry(
-            status, wait
+            status, wait=wait
         )
         if systemd_status:
             status = systemd_status
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -11,6 +11,7 @@ import enum
 import json
 import os
 import sys
+import time
 from copy import deepcopy
 from time import gmtime, sleep, strftime
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
@@ -272,12 +273,15 @@ def get_bootstatus(disable_file, paths) -> Tuple[UXAppBootStatusCode, str]:
     return (bootstatus_code, reason)
 
 
-def _get_systemd_status() -> Optional[UXAppStatus]:
-    """Get status from systemd.
+def _get_error_or_running_from_systemd() -> Optional[UXAppStatus]:
+    """Get if systemd is in error or running state.
 
     Using systemd, we can get more fine-grained status of the
     individual unit. Determine if we're still
-    running or if there's an error we haven't otherwise detected
+    running or if there's an error we haven't otherwise detected.
+
+    If we don't detect error or running, return None as we don't want to
+    report any other particular status based on systemd.
     """
     for service in [
         "cloud-final.service",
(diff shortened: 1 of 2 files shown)


cloudinit/cmd/status.py
```diff
@@ -324,6 +328,35 @@ def _get_systemd_status() -> Optional[UXAppStatus]:
...
+        try:
+            return _get_error_or_running_from_systemd()
+        except subp.ProcessExecutionError:
+            if existing_status in (
+                UXAppStatus.DEGRADED_RUNNING,
+                UXAppStatus.RUNNING,
+            ):
+                return None
+            sleep(0.25)
+    print(
```
TheRealFalcon, line 352: I don't love printing here, but when trying to raise an exception, I didn't really find a better place higher in the call stack to catch and print, so here the print is.

blackboxsw: This works, we'll still be getting an exit 1 in this case indicating a failure anyway. I'd like to improve our message to better indicate the error raised by systemctl so if we have to debug this later we know the failure mode

```diff
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -339,19 +339,21 @@ def _get_error_or_running_from_systemd_with_retry(
     error state while cloud-init is still running.
     """
     start_time = time.time()
+    last_error = None
     while time.time() - start_time < max_wait:
         try:
             return _get_error_or_running_from_systemd()
-        except subp.ProcessExecutionError:
+        except subp.ProcessExecutionError as e:
             if existing_status in (
                 UXAppStatus.DEGRADED_RUNNING,
                 UXAppStatus.RUNNING,
             ):
                 return None
+            last_error = e
             sleep(0.25)
     print(
         f"Failed to get status from systemd after {max_wait} seconds. "
-        "Cloud-init may still be running.",
+        f"Cloud-init may still be running. {last_error}",
         file=sys.stderr,
     )
     return None
```
Otherise +1

blackboxsw (approved):

cloudinit/cmd/status.py
```diff
@@ -324,6 +328,35 @@ def _get_systemd_status() -> Optional[UXAppStatus]:
...
+        f"Failed to get status from systemd after {max_wait} seconds. "
```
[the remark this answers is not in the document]
blackboxsw: This generates the following output:
```
...
Failed to get status from systemd after 5 seconds. Cloud-init may still be running.

status: error
extended_status: error
boot_status_code: disabled-by-generator
last_update: Mon, 11 Dec 2023 21:58:06 +0000
detail:
DataSourceLXD
errors:
	- someerror
recoverable_errors: {}
```

I'd suggest (below) that we may want more of a failure reason from systemctl show instead of just "Failed to get status from systemd"


We probably also may not want to return the semi-stricture content when this error/print is raised?

cloudinit/cmd/status.py
```diff
@@ -324,6 +328,35 @@ def _get_systemd_status() -> Optional[UXAppStatus]:
     return None
 
 
+def _get_error_or_running_from_systemd_with_retry(
+    existing_status, max_wait=5
```
blackboxsw, line 332: In the original bug (on OpenStack) the fix-case the commenter mentioned was a sleep of 30 seconds. Should this max_wait be longer for slower platforms so we don't reproduce this same failure case?

TheRealFalcon:
@holmanb or @blackboxsw , this should be ready for re-review

cloudinit/cmd/status.py
blackboxsw: thanks for the `{last_exception.stderr}` error text

blackboxsw (approved):
Thanks for the rework.

Blocking on `--wait` indefinitely only when cloud-init is in error condition and failures to talk to dbus via `systemctl show` is what we want here as systemctl failures should only exist prior to dbus.service/socket being setup in very early boot. Otherwise calls to `cloud-init status` without the `--wait` param will not block or retry but immediately raise a user-facing error message about inability to contact dbus and provide the known current status values. 


In testing this thoroughly, two unrelated bugs came up that we can address separately
1. `cloud-init status --wait --format=json` should emit progress dots/stops "." to behave like invocations `--wait --long` and `--wait`.
2. When errors are discovered in previous stages, but cloud-init is [still running due to no results.json file created, do not set status to ERROR, but rather DEGRADED_RUNNING](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/status.py#L390-L391)

holmanb:
> 2. When errors are discovered in previous stages, but cloud-init is [still running due to no results.json file created, do not set status to ERROR, but rather DEGRADED_RUNNING](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/status.py#L390-L391)

@blackboxsw I disagree. `DEGRADED_RUNNING` is used to indicate [recoverable error](https://cloudinit.readthedocs.io/en/latest/explanation/failure_states.html#recoverable-failure), errors map to [critical failure](https://cloudinit.readthedocs.io/en/latest/explanation/failure_states.html#critical-failure).

See [my proposal](https://github.com/canonical/cloud-init/issues/4685) for a way to keep this distinction while also better reporting that cloud-init has experienced an error but hasn't yet exited.

PR #4682, opened 2023-12-11
Enable coverage in integration tests

TheRealFalcon:
## Proposed Commit Message
rebase

## Merge type

### DIFF
--- a/tests/integration_tests/assets/enable_coverage.py
+++ b/tests/integration_tests/assets/enable_coverage.py
@@ -1,0 +1,28 @@
+from pathlib import Path
+
+services = [
+    "cloud-init-local.service",
+    "cloud-init.service",
+    "cloud-config.service",
+    "cloud-final.service",
+]
+service_dir = Path("/lib/systemd/system/")
+
+# Check for the existence of the service files
+for service in services:
+    if not (service_dir / service).is_file():
+        print(f"Error: {service} does not exist in {service_dir}")
+        exit(1)
+
+# Prepend the ExecStart= line with 'python3 -m coverage run'
+for service in services:
+    file_path = service_dir / service
+    content = file_path.read_text()
+    content = content.replace(
+        "ExecStart=/usr",
+        (
+            "ExecStart=python3 -m coverage run "
+            "--source=/usr/lib/python3/dist-packages/cloudinit --append /usr"
+        ),
+    )
+    file_path.write_text(content)
--- a/tests/integration_tests/conftest.py
+++ b/tests/integration_tests/conftest.py
@@ -3,11 +3,13 @@ import datetime
 import functools
 import logging
 import os
+import shutil
+import subprocess
 import sys
 from contextlib import contextmanager
 from pathlib import Path
 from tarfile import TarFile
-from typing import Dict, Generator, Iterator, Type
+from typing import Dict, Generator, Iterator, List, Type
 
 import pytest
 from pycloudlib.lxd.instance import LXDInstance
@@ -115,11 +117,18 @@ def setup_image(session_cloud: IntegrationCloud, request):
     So we can launch instances / run tests with the correct image
     """
     source = get_validated_source(session_cloud)
-    if not source.installs_new_version():
+    if not (
+        source.installs_new_version() or integration_settings.INCLUDE_COVERAGE
+    ):
         return
-    log.info("Setting up environment for %s", session_cloud.datasource)
+    log.info("Setting up source image")
     client = session_cloud.launch()
-    client.install_new_cloud_init(source)
+    if source.installs_new_version():
+        log.info("Installing cloud-init from %s", source.name)
+        client.install_new_cloud_init(source)
+    if integration_settings.INCLUDE_COVERAGE:
+        log.info("Installing coverage")
+        client.install_coverage()
     # All done customizing the image, so snapshot it and make it global
     snapshot_id = client.snapshot()
     client.cloud.snapshot_id = snapshot_id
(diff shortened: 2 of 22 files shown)


tests/integration_tests/conftest.py
```diff
@@ -131,28 +143,35 @@ def setup_image(session_cloud: IntegrationCloud, request):
...
-    """
-    if any(
-        [
-            integration_settings.COLLECT_LOGS == "NEVER",
-            integration_settings.COLLECT_LOGS == "ON_ERROR"
-            and not test_failed,
-        ]
-    ):
-        return
+def _collect_logs(instance: IntegrationInstance, log_dir: Path):
```
TheRealFalcon, line 146: To help read the following sections:
Instead of `_collect_logs`, we instead call a new `_collect_artifacts` after every test. This function will check what needs to be setup based on the integration settings, then call `_collect_logs` and/or `_collect_coverage` accordingly.

aciba90 (commented):
Thanks, @TheRealFalcon, great work.

As it is right now:

1. We instrument the cloud-init services with `coverage` in the tested instance.
2. Run pytest from test-runner instance.
3. Pull the coverage metadata files from the tested instance.
4. Generate the coverage reports in the test-runner instance.

The last step fails if cloud-init is not installed in the test-runner instance because the coverage metadata points to `/usr/lib/python3/dist-packages/cloudinit`. See: [log.txt](https://github.com/canonical/cloud-init/files/13687739/log.txt).

I wonder if it would be better to generate the reports in the tested instance and pull them, as coverage is already installed there.

This would remove the need to install cloud-init in developer's machines and would avoid misalignment between cloud-init versions in both machines, which could provoke this same error (I am not sure if the absence or modification of only one file would lead to this error, but it could be).

tests/integration_tests/assets/enable_coverage.py
```diff
@@ -0,0 +1,28 @@
...
+service_dir = Path("/lib/systemd/system/")
+
+# Check for the existence of the service files
+for service in services:
+    if not (service_dir / service).is_file():
+        print(f"Error: {service} does not exist in {service_dir}")
+        exit(1)
+
+# Prepend the ExecStart= line with 'python3 -m coverage run'
+for service in services:
```
holmanb, line 18: I've done similar things (more ad-hoc) for inserting profiler commands into the service files. At some point I'll probably steal this code or make it more generic to do that in automatically in tox (unless someone else wants to do it)

tests/integration_tests/util.py
```diff
@@ -8,12 +8,14 @@
...
 from pathlib import Path
-from typing import Set
+from typing import TYPE_CHECKING, Set
 
 import pytest
 
 from cloudinit.subp import subp
-from tests.integration_tests.instances import IntegrationInstance
+
+if TYPE_CHECKING:
```
holmanb, line 17: Why this? Performance optimization? Can we add a comment or something to the commit message? Also if this isn't required for enabling coverage maybe it should go in its own commit?

holmanb (commented):
Nice, thanks! One inline question / and some random comments. I'll wait for @aciba90's comments to be addressed.


> The last step fails if cloud-init is not installed in the test-runner instance

Is this actually a problem? The package `python3-coverage` gets installed in `tests/integration_tests/instances.py`, so when would this happen?

> I wonder if it would be better to generate the reports in the tested instance and pull them, as coverage is already installed there.
>
> This would remove the need to install cloud-init in developer's machines and would avoid misalignment between cloud-init versions in both machines, which could provoke this same error (I am not sure if the absence or modification of only one file would lead to this error, but it could be).

I think you meant `s/cloud-init/python3-coverage/g` here, and yes - I agree this might forcibly require the host and guest to have the same coverage version for this to work - generation on the host would be better and therefore wouldn't require local install of coverage.

tests/integration_tests/util.py
TheRealFalcon: Avoid circular import. I almost made it a separate commit...so...I guess I will :smile:

holmanb: Could we please include a comment that it's for avoiding circular import?

tests/integration_tests/assets/enable_coverage.py
holmanb: https://github.com/canonical/cloud-init/issues/4729

TheRealFalcon:
> The last step fails if cloud-init is not installed in the test-runner instance because the coverage metadata points to /usr/lib/python3/dist-packages/cloudinit. See: [log.txt](https://github.com/canonical/cloud-init/files/13687739/log.txt).

Thanks! I had the cloud-init package installed locally, so I hadn't noticed this

> This would remove the need to install cloud-init in developer's machines and would avoid misalignment between cloud-init versions in both machines, which could provoke this same error

You've identified two different issues:
1. The need to install the cloud-init package to ensure matching paths 
2. Misalignment between cloud-init versions

For 1, coverage has a configuration option to align a directory on your local machine with paths from coverage files. I pushed a change (see `tox.ini`) that does this.

For 2, that is a valid concern, but it's not one that I'm super concerned about. If there's a mismatch in source, our coverage numbers might be slightly off, but in most cases not enough to make a considerable difference. If we want to avoid this, we can ensure that the code we're testing is the same as the code we're launching from.

> I wonder if it would be better to generate the reports in the tested instance and pull them, as coverage is already installed there.

I'm not sure there's a way as easy way to do this. `coverage` has the option to combine binary results, but I haven't seen an option for combining generated reports. The other option would be to launch one more instance after all tests have run, push all the coverage results there, then have that instance generate them, but that seems like a lot of overhead for something like this. If it's that vs misalignment, I prefer the misalignment.

aciba90:
Thanks for the explanation and addressing 1. I agree, 2 is not worth handling.

I get now the following error:

```
2024-01-16 10:27:25 INFO      integration_testing:conftest.py:363 Generating coverage report                                                                                                                       
/home/aciba/canonical/wt/cloud-init/reviews/.tox/integration-tests/lib/python3.11/site-packages/coverage/data.py:171: CoverageWarning: Couldn't use data file '/tmp/cloud_init_test_logs/240116102518/.coverage.tes
ts-integration_tests-test_paths-TestHonorCloudDir-test_honor_cloud_dir': Looks like a coverage 4.x data file. Are you mixing versions of coverage?
```

Misalignment between inner and outer coverage versions?

Full test session: https://pastebin.canonical.com/p/gShy25PkSq/

TheRealFalcon:
I pushed two new commits to install via pip rather than using python3-coverage.

If the PR is approved, I plan on moving the refactor commit earlier and squashing the fixup commit, but I left them in the current order for easier review.

aciba90 (approved):
LGTM, thanks.

PR #4683, opened 2023-12-12
refactor(IscDhclient): discover DHCP leases at distro-provided location

phsm:
## Proposed Commit Message

```
refactor(IscDhclient): discover DHCP leases at distro-provided location

Use leases directory and lease file provided by Distro object
instead of cycling through all possible locations.
```

## Additional Context

IscDhclient class seems to be only used by CloudStack datasource.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/rhel.py
+++ b/cloudinit/distros/rhel.py
@@ -29,6 +29,12 @@ class Distro(distros.Distro):
     network_script_tpl = "/etc/sysconfig/network-scripts/ifcfg-%s"
     tz_local_fn = "/etc/localtime"
     usr_lib_exec = "/usr/libexec"
+    # RHEL and derivatives use NetworkManager DHCP client by default.
+    # But if NM is configured with using dhclient ("dhcp=dhclient" statement)
+    # then the following location is used:
+    # /var/lib/NetworkManager/dhclient-<uuid>-<network_interface>.lease
+    dhclient_lease_directory = "/var/lib/NetworkManager"
+    dhclient_lease_file_regex = r"dhclient-[\w-]+\.lease"
     renderer_configs = {
         "sysconfig": {
             "control": "etc/sysconfig/network",
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -29,6 +29,9 @@ from cloudinit.net import (
 LOG = logging.getLogger(__name__)
 
 NETWORKD_LEASES_DIR = "/run/systemd/netif/leases"
+DHCLIENT_FALLBACK_LEASE_DIR = "/var/lib/dhclient"
+# Match something.lease or something.leases
+DHCLIENT_FALLBACK_LEASE_REGEX = r".+\.leases?$"
 UDHCPC_SCRIPT = """#!/bin/sh
 log() {
     echo "udhcpc[$PPID]" "$interface: $2"
@@ -467,51 +470,38 @@ class IscDhclient(DhcpClient):
         return static_routes
 
     @staticmethod
-    def get_dhclient_d():
-        # find lease files directory
-        supported_dirs = [
-            "/var/lib/dhclient",
-            "/var/lib/dhcp",
-            "/var/lib/NetworkManager",
-        ]
-        for d in supported_dirs:
-            if os.path.exists(d) and len(os.listdir(d)) > 0:
-                LOG.debug("Using %s lease directory", d)
-                return d
-        return None
-
-    @staticmethod
-    def get_latest_lease(lease_d=None):
-        # find latest lease file
-        if lease_d is None:
-            lease_d = IscDhclient.get_dhclient_d()
-        if not lease_d:
-            return None
-        lease_files = os.listdir(lease_d)
-        latest_mtime = -1
+    def get_latest_lease(lease_dir, lease_file_regex):
         latest_file = None
 
-        # lease files are named inconsistently across distros.
-        # We assume that 'dhclient6' indicates ipv6 and ignore it.
-        # ubuntu:
-        #   dhclient.<iface>.leases, dhclient.leases, dhclient6.leases
-        # centos6:
-        #   dhclient-<iface>.leases, dhclient6.leases
-        # centos7: ('--' is not a typo)
-        #   dhclient--<iface>.lease, dhclient6.leases
-        for fname in lease_files:
-            if fname.startswith("dhclient6"):
-                # avoid files that start with dhclient6 assuming dhcpv6.
-                continue
-            if not (fname.endswith((".lease", ".leases"))):
+        # Try primary dir/regex, then the fallback ones
+        for directory, regex in (
+            (lease_dir, lease_file_regex),
+            (DHCLIENT_FALLBACK_LEASE_DIR, DHCLIENT_FALLBACK_LEASE_REGEX),
+        ):
+            if not directory:
                 continue
 
-            abs_path = os.path.join(lease_d, fname)
-            mtime = os.path.getmtime(abs_path)
-            if mtime > latest_mtime:
-                latest_mtime = mtime
-                latest_file = abs_path
-        return latest_file
+            lease_files = []
+            try:
+                lease_files = os.listdir(directory)
+            except FileNotFoundError:
+                continue
+
+            latest_mtime = -1
+            for fname in lease_files:
+                if not re.search(regex, fname):
+                    continue
+
+                abs_path = os.path.join(directory, fname)
+                mtime = os.path.getmtime(abs_path)
+                if mtime > latest_mtime:
+                    latest_mtime = mtime
+                    latest_file = abs_path
+
+            # Lease file found, skipping falling back
+            if latest_file:
+                return latest_file
+        return None
 
     @staticmethod
     def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
--- a/cloudinit/sources/DataSourceCloudStack.py
+++ b/cloudinit/sources/DataSourceCloudStack.py
@@ -87,10 +87,12 @@ class DataSourceCloudStack(sources.DataSource):
         # Cloudstack has its metadata/userdata URLs located at
         # http://<virtual-router-ip>/latest/
         self.api_ver = "latest"
-        self.vr_addr = get_vr_address()
+
+        self.distro = distro
+        self.vr_addr = get_vr_address(self.distro)
         if not self.vr_addr:
             raise RuntimeError("No virtual router found!")
-        self.metadata_address = "http://%s/" % (self.vr_addr,)
+        self.metadata_address = f"http://{self.vr_addr}/"
         self.cfg = {}
 
     def _get_domainname(self):
@@ -108,7 +110,10 @@ class DataSourceCloudStack(sources.DataSource):
             "Falling back to ISC dhclient"
         )
 
-        lease_file = dhcp.IscDhclient.get_latest_lease()
+        lease_file = dhcp.IscDhclient.get_latest_lease(
+            self.distro.dhclient_lease_directory,
+            self.distro.dhclient_lease_file_regex,
+        )
         if not lease_file:
             LOG.debug("Dhclient lease file wasn't found")
             return None
@@ -259,7 +264,7 @@ def get_default_gateway():
     return None
 
 
-def get_vr_address():
+def get_vr_address(distro):
     # Get the address of the virtual router via dhcp leases
     # If no virtual router is detected, fallback on default gateway.
     # See http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.8/virtual_machines/user-data.html # noqa
@@ -282,7 +287,12 @@ def get_vr_address():
         return latest_address
 
     # Try dhcp lease files next
-    lease_file = dhcp.IscDhclient.get_latest_lease()
+    # get_latest_lease() needs a Distro object to know which directory
+    # stores lease files
+    lease_file = dhcp.IscDhclient.get_latest_lease(
+        distro.dhclient_lease_directory, distro.dhclient_lease_file_regex
+    )
+
     if lease_file:
         latest_address = dhcp.IscDhclient.parse_dhcp_server_from_lease_file(
             lease_file
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -148,6 +148,14 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
     osfamily: str
     dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd, dhcp.Udhcpc]
+    # Directory where the distro stores their DHCP leases.
+    # The children classes should override this with their dhcp leases
+    # directory
+    dhclient_lease_directory: Optional[str] = None
+    # A regex to match DHCP lease file(s)
+    # The children classes should override this with a regex matching
+    # their lease file name format
+    dhclient_lease_file_regex: Optional[str] = None
 
     def __init__(self, name, cfg, paths):
         self._paths = paths
--- a/cloudinit/distros/amazon.py
+++ b/cloudinit/distros/amazon.py
@@ -14,5 +14,11 @@ from cloudinit.distros import rhel
 
 
 class Distro(rhel.Distro):
+    # Amazon Linux 2 stores dhclient leases at following location:
+    # /var/lib/dhclient/dhclient--<iface_name>.leases
+    # Perhaps there could be a UUID in between two "-" in the file name
+    dhclient_lease_directory = "/var/lib/dhcp"
+    dhclient_lease_file_regex = r"dhclient-[\w-]+\.lease"
+
     def update_package_sources(self):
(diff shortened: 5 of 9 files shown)


igalic (approved):
This looks good to me!

This part of the commit message,

> This change adds allows discovering such lease files.

seems like it should be either "adds" or "allows", not both.

phsm:
> This looks good to me!
> 
> This part of the commit message,
> 
> > This change adds allows discovering such lease files.
> 
> seems like it should be either "adds" or "allows", not both.

Thanks, fixed the typo in the commit message.

cloudinit/net/dhcp.py
```diff
@@ -473,6 +473,7 @@ def get_dhclient_d():
             "/var/lib/dhclient",
             "/var/lib/dhcp",
             "/var/lib/NetworkManager",
+            "/var/db",
```
holmanb, line 476: I'm on the fence about whether to make this a (free)bsd check before adding. On one hand, I check a couple of linuxen, and some have the directory, others don't. Why check this directory on those distros? On the other hand, there were few files in that directory and only a single `listdir` called, so it shouldn't be "slow".

This function is one that I've previously considered breaking out into `distros/` anyways, since I think `/var/lib/dhcp/` and `/var/lib/dhclient/` are distro-specific iirc.

Thoughts?

holmanb (commented):
Thanks for this @phsm!

Nice addition and tests, I'm leaning towards making that a (free)bsd check prior to adding to the list.

Justification below.

cloudinit/net/dhcp.py
phsm: I briefly checked the `distros/` directory: seems its purpose is to isolate distro-specific code in there. So this means that the cloud-init has a reliable way to detect which os/distro it is running on? 

If yes, then its indeed better to implement this function per-distro.
Perhaps implement a method in `distros.Distro` that uses a class attribute [such as this](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/debian.py#L34) and override that class attribute per child distro class.

holmanb: > Perhaps implement a method in distros.Distro that uses a class attribute with the dhclient directory [such as this](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/debian.py#L34) and override that class attribute per child distro class.

@phsm yes, this sounds like the right direction. I think we have a reference to the distro instance higher up that we'll just need to pass down the stack as an argument to access it in this function.

phsm: I just made another commit with an example how I see it done:
- main Distro class has two attributes: dhclient_lease_directory, dhclient_lease_file_regex defining the lease directory for lease files in a distro, and a regex helping to find those lease files in there.
- every class of a distro that uses dhclient has those 2 attributes defined.
- Cloudstack's get_vr_address() now detects a distro and supplies dhcp.IscDhclient.get_latest_lease() with the lease_dir an a regex.
- there is no need for get_dhclient_d() anymore
- Although get_latest_lease() function signature is changed, I don't see it being used anywhere except Cloudstack DS so it shouldn't break anything.

The purpose of this commit is to gather your feedback if i'm working in the right direction. 
Also, do I detect the distro correctly here?

cloudinit/sources/DataSourceCloudStack.py
```diff
@@ -282,7 +283,11 @@ def get_vr_address():
         return latest_address
 
     # Try dhcp lease files next...
-    lease_file = dhcp.IscDhclient.get_latest_lease()
+    distro = stages.Init(reporter=True).distro  # detect OS distro
```
holmanb, line 286: Driveby comment:  Why call Init()? `__init__()` already got an instance of distro, lets just save a reference to it in an instance variable and use that rather than calling `Init()`.

phsm: Indeed, I noticed that the Distro object is already passed into `sources.DataSource.__init__()` as an argument. Thus, no need to obtain the Distro object once more.
Switched to using that. 

Does the rest look OK so I can implement it for the other distros?

cloudinit/distros/rhel.py
```diff
@@ -29,6 +29,10 @@ class Distro(distros.Distro):
     network_script_tpl = "/etc/sysconfig/network-scripts/ifcfg-%s"
     tz_local_fn = "/etc/localtime"
     usr_lib_exec = "/usr/libexec"
+    # RHEL and derivatives have the following dhclient lease path:
+    # /var/lib/NetworkManager/dhclient-<uuid>-<network_interface>.lease
+    dhclient_lease_directory = '/var/lib/NetworkManager'
+    dhclient_lease_file_regex = r'dhclient-[\w-]+\.lease'
```
igalic, line 35: where is the equivalent for this in other distros? particularly freebsd

igalic (commented):
❓

cloudinit/distros/rhel.py
phsm: Hi @igalic,
As I mentioned earlier, this commit is intended as a WIP, a draft to show an idea how I intend to implement it. I just picked RHEL distro as an example. 

Once I get Brett's approval on this course, I'll implement it for every distro, FreeBSD included.

cloudinit/net/dhcp.py
phsm: I just updated the PR with the following:
1. Updated the distros files with dhclient lease director/regex for distros that support DHClient. I installed all of them on a virtual machine to verify the paths:

- CentOS: updated centos.py to contain dhclient parameters for Centos 7. I assume no one uses CentOS > 7 on servers because they're EOL.
- RHEL: updated rhel.py with dhclient parameters. To use dhclient in RHEL, `dhcp=dhclient` parameter must be set in NetworkManager config.
- AlmaLinux: identical to RHEL.
- Rocky: identical to RHEL.
- Debian: updated debian.py with dhclient parameters.
- Alpine: doesn't use dhclient, uses udhcpc.
- Amazon Linux: updated amazon.py with dhclient parameters.
- CloudLinux: identical to RHEL.
- EuroLinux: identical to RHEL.
- Fedora: identical to RHEL
- FreeBSD: updated freebsd.py with dhclient parameters.
- Gentoo: doesn't use dhcpcd, uses dhcpcd.
- Mariner: uses systemd-networkd.
- Miracle: identical to RHEL
- Open Cloud OS: I couldn't even download it an ISO image to test. I assume its identical to RHEL.
- OpenEuler: identical to RHEL
- OpenSUSE: doesn't use dhclient, uses wicked.
- SLES: I don't have a copy of SLES, I assume it also uses wicked.
- Virtuozzo: I don't have a copy of it, I assume its identical to RHEL.
- Tencentos: didn't test, assume its identical to RHEL. 

2. Tests: 
- Added tests for finding a lease in different distros that can use dhclient.
- Removed `TestGetLatestLease` testcase from `test_cloudstack.py` in its entirety: Cloudstack tests shouldn't test obtaining DHCP lease, its already covered in `test_dhcp.py`

phsm:
Sorry for commit storm, I was struggling with the linters and the tests.

I also changed the PR description as its not just fixing Dhclient at FreeBSD anymore.

cloudinit/sources/DataSourceCloudStack.py
```diff
@@ -282,7 +289,16 @@ def get_vr_address():
...
     # Try dhcp lease files next
-    lease_file = dhcp.IscDhclient.get_latest_lease()
+    # get_latest_lease() needs a Distro object to know which directory
+    # stores lease files
+    lease_file = None
+    if distro:
+        lease_file = dhcp.IscDhclient.get_latest_lease(
+            distro.dhclient_lease_directory, distro.dhclient_lease_file_regex
+        )
+    else:
```
igalic, line 299: I'm not sure this should happen at all

cloudinit/sources/DataSourceCloudStack.py
```diff
@@ -282,7 +289,16 @@ def get_vr_address():
...
+        LOG.debug("Distro object is not defined, skipping leasefile search")
```
igalic, line 300: and if it did, i don't see how a debug log is useful.

remove the if, and the else, let it crash loudly and clearly

cloudinit/sources/DataSourceCloudStack.py
```diff
@@ -108,7 +110,12 @@ def _get_domainname(self):
             "Falling back to ISC dhclient"
         )
 
-        lease_file = dhcp.IscDhclient.get_latest_lease()
+        lease_file = None
+        if self.distro:
```
igalic, line 114: please get rid of this if, I think it's misleading.
this reads as: this could sensibly happen! rather than: if this happens, nothing here should work

igalic (changes requested):
this looks good, here's some requests for changes

cloudinit/sources/DataSourceCloudStack.py
phsm: Hi @igalic, 

Thank you for your comments. I believe all three your comments can be grouped here together in one notion: "explicit is better than implicit, better return an error than continuing with undefined behavior".

I normally would agree with it but there is one more last resort down the code: look for Cloudstack Metadata server at server's default gateway. So even if something is wrong with the DHCP client code, there's still a chance for cloud-init to get its hands on metadata.

I know its hard to discuss the logical flow without knowing whats going on underneath. That's why I explained how Cloudstack networks work in detail in my previous PR: https://github.com/canonical/cloud-init/pull/4433#discussion_r1366978013

Basically the reason behind not raising an exception here is to allow the code to use the last resort, the default gateway address.

cloudinit/sources/DataSourceCloudStack.py
phsm: This method was added in my previous PR: https://github.com/canonical/cloud-init/pull/4433
You can follow the discussion there (its quite long, though). 

In short: the virtual machine FQDN is assembled by concatenating isntance name + '.' + domain_name_given_by_dhcp.

This method is only used at get_hostname() method that tries to make an FQDN and if it can't obtain DHCP domain name for any reason, it uses the instance name as FQDN (the behavior previous to my PR)

phsm: Oh please disregard my previous comment, I thought you commented similar code in get_vr_address() method.

Well, I of course can let it crash here. But what to do with the get_hostname() method that uses this? The current behavior is to try assembling an FQDN and if that fails, return the instance name as FQDN.

cloudinit/sources/DataSourceCloudStack.py
igalic: I don't doubt that you understand the CloudStack code way better than I do, so let's reformulate my change request into two clear and concise questions:

1. under which circumstances can this `distro` object be `None`?
2. why is `distro=None` a desirable state for a `DataSourceCloudStack` instance to be in?

phsm: 1. I don't have an answer to this question. I found that None is passed into DataSourceCloudStack.__init__() during tests so I thought it is better to include a check for it in case None is somehow passed in.
I would assume that distro=None should normally never happen.

2. It is not a desirable state.

I guess if the crash during distro=None is acceptable, I could assume that distro is always provided, and adjust the tests to be initialized with `distros.ubuntu.Distro`  instead of None.

phsm: Updated the PR: removed the distro=None checks, as suggested.

cloudinit/sources/DataSourceCloudStack.py
phsm: Updated the PR: removed the distro=None checks, as suggested.

cloudinit/sources/DataSourceCloudStack.py
phsm: Updated the PR: removed the distro=None checks, as suggested.

cloudinit/sources/DataSourceCloudStack.py
igalic: if it's only happening during tests we should just fix the tests

igalic (approved):
I like this.
thank you for this monumental effort

holmanb (changes requested):
@phsm nice work here - thanks for putting in this effort.

I left a couple of comments that I think we need to address.

cloudinit/distros/rhel.py
```diff
@@ -29,6 +29,12 @@ class Distro(distros.Distro):
     network_script_tpl = "/etc/sysconfig/network-scripts/ifcfg-%s"
     tz_local_fn = "/etc/localtime"
     usr_lib_exec = "/usr/libexec"
+    # RHEL and derivatives use NetworkManager DHCP client by default.
```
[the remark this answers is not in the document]
phsm: Implemented:
- if the distro-provided dir and regex yield nothing, then try to use the fallback dir and regex. Regex is specifically made to be quite loose (search for any file ending with `.lease` or `.leases`
- Tests for that behavior

cloudinit/distros/centos.py
```diff
@@ -4,4 +4,8 @@
 
 
 class Distro(rhel.Distro):
-    pass
+    # Centos 7 has DHCP lease at the following location:
```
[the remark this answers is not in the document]
phsm: I checked Centos9-stream. As I expected, it is identical to RHEL.
Removed dir and regex definitions under Centos distro, and the test for centos7 case.

cloudinit/net/dhcp.py
```diff
@@ -29,6 +29,8 @@
 LOG = logging.getLogger(__name__)
 
 NETWORKD_LEASES_DIR = "/run/systemd/netif/leases"
+DHCLIENT_FALLBACK_LEASE_DIR = "/var/lib/dhclient"
+DHCLIENT_FALLBACK_LEASE_REGEX = r".+\.leases*$"  # smth.lease or smth.leases
```
holmanb, line 33: I think you meant `.+\.lease*$` to match `.lease` or `.leases`, right?

Also we typically avoid inline comments in this codebase as a style rule, fyi.

holmanb (changes requested):
Nice work - getting close! A couple more changes requested, I'm testing this on Ec2 now.

cloudinit/net/dhcp.py
phsm: well, `s*` means 0 or more occurrences of character `s` ;) So it will match '.lease', '.leases', '.leasessss'.

I moved the inline comment.

igalic: shouldn't it just be `\.leases?$` ?

cloudinit/distros/rhel.py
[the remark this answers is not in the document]
holmanb: @phsm Just looking at this change now, I realize that you made `/var/lib/dhclient/` default for all distros.

From what I can tell, the `dhclient` subdirectory is a feature of RHEL derivatives and Arch. Sorry I missed this before.

The following distros use `/var/lib/dhcp/`: including Alpine, Gentoo (and derivatives I assume), Suse (and derivatives), Debian, etc.

No clue what Photon or Mariner use.

With the change as it stands, `/var/lib/dhclient` will be the only directory checked for distros that do not inherit from FreeBSD / RHEL / Debian, so I think that we need to account for those mentioned above.

@igalic is `/var/lib/dhclient` is a reasonable directory for the non-FreeBSD BSDs?

cloudinit/net/dhcp.py
phsm: Indeed, I totally forgot about the `?` quantifier. Thanks, fixed that.

cloudinit/distros/rhel.py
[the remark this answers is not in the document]
igalic: I'll need to check.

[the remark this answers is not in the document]
igalic: OpenBSD puts it in the same place as FreeBSD:
https://man.openbsd.org/dhclient.8#FILES

NetBSD had switched to `dhcpcd` since 9.0. before that, they put it in the same place, https://man.netbsd.org/NetBSD-8.2/dhclient.8#FILES but with a different name

cloudinit/net/dhcp.py
holmanb: Thanks @phsm for the update. I think that there is a missing detail here.

> doesn't use dhclient

Use of wicked / networkd / networkmanager / etc doesn't necessarily mean that dhclient isn't used. In fact, cloud-init itself uses dhclient during the [local](https://cloudinit.readthedocs.io/en/latest/explanation/boot.html#local) stage. Cloud-init sets up an ephemeral network so that it can reach out to the IMDS to get configuration. This allows cloud-init to write the networking configuration locally *before* the distro's built-in network manager is even started.

You've done some good work on this refactor, but I think that by default all distros should probably be checking for the isc-dhclient path, at least until `udhcpc` is fully featured for alpine, etc (with classless static routes and unknown dhcp option support) and dhcpcd replaces isc-dhclient on other distros.

holmanb (approved):
Great work @phsm, thanks for putting this together.

Looks good to me, so I'll merge.

If I understand correctly, cloudstack might also benefit from parsing the default lease that cloud-init gets during the ephemeral stage. I'll probably include that as part of the work I'm doing with dhcpcd.

PR #4535, opened 2023-10-19
drop systemd order After=snapd.seeded.service from cloud-config.service

blackboxsw:
For review and discussion.  Having a strict systemd ordering `After=snapd.seeded.service` declared for `cloud-config.service` was defined to give cloud-config.service certainty that everything snap has already been set up in case any optional cloud-config modules in cloudinit.config.cc_* needed to rely on snap tooling to be present.

In practice this strict ordering always adds around a 10 second delay while cloud-config.service waits for snapd.seeded.service to complete its attestation, installation and setup yet the snaps seeded are not critical to the default functionality the cloud-init is configuring.

 In practice, there are only a few modules that absolutely require snaps to be seeded because they may be interacting with snapd itself, those modules are:
- cc_lxd
- cc_ubuntu_autoinstall
- cc_snaps
- cc_runcmd (in case option runcmd scripts reference snaps or snap tools)

This PR provides a common `util.wait_for_snap_seeded` which invokes `snap wait system seed.loaded` can be called from any config module that relies on snap setup being complete. This means that the default case for cloud-image installs should no longer have to wait on snaps being seeded to finish initial user configuration and bringing up SSH  and can therefore unblock earlier remote access and login to the system during boot. Each module will attempt to call wait_for_snap_seeded through the semaphore-based cloud.run method to prevent this function from being called more than once by other modules.

## boottime comparisons

TLDR: default launch without runcmd/lxd/autoinstall gets a big reduction in time to SSH/login (~4 seconds) by cloud-config.service not waiting on snap.seeded.service.   Overall boot time via systemd analyze is unchanged as we still have to wait for snap.seeded.service somewhere before mulit-user-.target. When runcmd/lxd/autoinstall needs to snap wait system seed.service, overall boot time is still comparable but time-to-SSH/login still blocks on snaps being seeded.

<details>
   <summary>ppa:cloud-init-dev/daily vs daily+ThisPR with no applicable user-data (default launch case)</summary>

Things to note between run #1 (daily image)  vs run #2 (daily image + this PR)
* slightly faster boot time  time for systemd-analyze  (15.580s vs  14.076s)
* despite slightly longer snapd.seeded.service in run2 around <> ( 3.775s vs 4.277s)
* systemd-user-sessions.service still after cloud-config.service but starts earlier during boot: ( @11.978s vs @5.853s)
* comparable time cost of modules-config boot stage from cloud-init analyze show (00.26000 vs 00.47200)
[...]

PR #4650, opened 2023-12-01
docs: Update documentation on boot hooks SC-1657

CalvoM:
Address the concern in  https://github.com/canonical/cloud-init/issues/4542

## Proposed Commit Message

```
docs: Update documentation on boot hooks

Fixes GH-4542 
```

## Checklist

- [ ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [ ] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [x] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #4687, opened 2023-12-13
cloud-config: after systemd ordering After=snapd.seeded.service

blackboxsw:
## Proposed Commit Message
```
feat(bootspeed): cloud-config.service drop After=snapd.seeded (#4687)

Historically cloud-config.service was declared the boot stage blocked
on After=snapd.seeded.service to allow for any config module to
potentially leverage snap utilities if present.

In practice, only the config modules cc_snap, cc_lxd,
cc_ubuntu_autoinstall and cc_runcmd are the modules that may strictly
rely on snaps being seeded during initial boot. So the systemd
ordering is not necessary in the majority of deployment cases.

Avoid a hard systemd ordering `After=snapd-seeded.service` in
cloud-config.service. Instead, directly call
`snap system wait seed.loaded` if user-data requires that snaps
are present and seeded.

Add a semaphore runner set limited to frequency PER_ONCE
which calls snap wait system seed.loaded one time to block
if any related cloud-config modules are active which may require
snap interaction.

Add this helper to each affected cloud-config module:
  snap, lxd and ubuntu_autoinstall

In runcmd, document the need for call: snap wait system seed.loaded
if the runcmd user-data requires snaps being present.

The runner will prevent cloud-init from making multiple
calls to snap wait even if user-data requires snappy interaction
is applicable for multiple config modules with a dependency on snaps.
```

## Additional Context

boot speed improvements on GCP: https://github.com/canonical/cloud-init/pull/4535#issuecomment-1841008249

By instrumenting a [local test script for testing on QEMU](https://github.com/canonical/server-test-scripts/pull/201), I see negligible performance impacts when launching images with upgraded cloud-init and no user-data with this branch. 

When providing runcmd content in user-data I do see a 3 second degradation to boot time.

<details>
  <summary>Launch empty userdata: boot time to SSH delta: < 1 second diff </summary>

```
python3 qemu_performance.py -c cloud-init-drop-snapd-seeded.deb  -n 4 -c cloud-init-drop-snapd-seeded.deb | tee drop-strict-snap-seeded.log
------------------- Boot speed comparison ---------------------------
Boot image serial: 20231211
Cloud-init from: cloud-init-drop-snapd-seeded.deb
Source image: /srv/pycloudlib/noble/20231211/noble-server-cloudimg-amd64.img
Number of samples: 4
---------------------------------------------------------------------
```

```
--------------------- Performance Deltas Encountered ---------------------
| Delta type and service name        | Original avg/samples | New avg/samples
--------------------------------------------------------------------------
| ***         client_time_to_ssh       | 034.75: ['035.367', '034.398', '034.260', '034.967'] | 035.35: ['035.023', '035.677', '034.719', '035.963']
| ***         time_systemd_userspace   | 020.73: ['020.988', '020.234', '020.967', '020.713'] | 019.82: ['019.558', '019.998', '019.104', '020.612']
| ***DEGRADED cloud-config.service     | 001.15: ['001.078', '001.229', '001.143'] | 002.46: ['002.437', '002.454', '002.468', '002.464']
| ***IMPROVED init-network/config-ssh  | 001.30: ['002.076', '001.080', '000.634', '001.417'] | 001.04: ['001.228', '001.292', '000.931', '000.706']
------------------- Control image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 034.75 | 035.37 | 034.26 | client_time_to_ssh
| 040.44 | 040.69 | 040.04 | client_time_to_cloudinit_done
| 003.90 | 003.93 | 003.88 | time_systemd_kernel
| 020.73 | 020.99 | 020.23 | time_systemd_userspace
| 006.33 | 006.46 | 006.23 | snapd.seeded.service
| 003.33 | 004.05 | 002.66 | cloud-init.service
| 002.69 | 002.70 | 002.68 | snapd.apparmor.service
| 002.10 | 002.12 | 002.08 | cloud-init-local.service
| 001.59 | 001.64 | 001.56 | dev-sda1.device
| 001.27 | 001.41 | 001.16 | pollinate.service
| 001.19 | 001.22 | 001.17 | snapd.service
| 001.36 | 001.86 | 001.17 | systemd-networkd-wait-online.service
| 001.11 | 001.14 | 001.08 | apport.service
| 001.12 | 001.12 | 001.12 | snap.lxd.activate.service
| 001.15 | 001.23 | 001.08 | cloud-config.service
| 001.30 | 002.08 | 000.63 | init-network/config-ssh
| 000.88 | 000.91 | 000.85 | init-network/config-growpart
| 000.53 | 000.53 | 000.52 | init-local/search-NoCloud
| 000.49 | 000.53 | 000.41 | modules-config/config-grub_dpkg
| 000.29 | 000.32 | 000.27 | init-network/config-resizefs
| 000.28 | 000.31 | 000.25 | init-network/config-users_groups
------------------- Updated cloud-init image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 035.35 | 035.96 | 034.72 | client_time_to_ssh
| 039.61 | 040.41 | 039.17 | client_time_to_cloudinit_done
| 003.86 | 003.88 | 003.82 | time_systemd_kernel
| 019.82 | 020.61 | 019.10 | time_systemd_userspace
| 006.57 | 006.84 | 006.30 | snapd.seeded.service
| 003.04 | 003.37 | 002.67 | cloud-init.service
| 002.67 | 002.67 | 002.66 | snapd.apparmor.service
| 002.46 | 002.47 | 002.44 | cloud-config.service
| 002.10 | 002.10 | 002.08 | cloud-init-local.service
| 001.59 | 001.63 | 001.55 | dev-sda1.device
| 001.56 | 001.96 | 001.16 | systemd-networkd-wait-online.service
| 001.38 | 001.40 | 001.37 | snapd.service
| 001.28 | 001.35 | 001.18 | pollinate.service
| 001.19 | 001.23 | 001.14 | apport.service
| 001.13 | 001.18 | 001.07 | snap.lxd.activate.service
| 001.11 | 001.12 | 001.10 | user@1000.service
| 001.04 | 001.29 | 000.71 | init-network/config-ssh
| 000.86 | 000.91 | 000.81 | init-network/config-growpart
| 000.53 | 000.54 | 000.52 | init-local/search-NoCloud
| 000.47 | 000.50 | 000.46 | modules-config/config-grub_dpkg
| 000.27 | 000.29 | 000.24 | init-network/config-users_groups
| 000.30 | 000.31 | 000.28 | init-network/config-resizefs
```
</details>

<details>
  <summary>Launch runcmd in userdata: boot time to SSH delta: ~ 3 second diff </summary>

```
------------------- Boot speed comparison ---------------------------
Boot image serial: 20231211
Cloud-init from: cloud-init-drop-snapd-seeded.deb
Source image: /srv/pycloudlib/noble/20231211/noble-server-cloudimg-amd64.img
Number of samples: 4
---------------------------------------------------------------------


--------------------- Performance Deltas Encountered ---------------------
| Delta type and service name        | Original avg/samples | New avg/samples
--------------------------------------------------------------------------
| ***         client_time_to_ssh       | 034.58: ['033.666', '035.233', '034.777', '034.638'] | 038.20: ['037.362', '039.576', '038.053', '037.821']
| ***         time_systemd_userspace   | 020.42: ['019.875', '021.370', '020.485', '019.947'] | 018.97: ['018.361', '019.130', '018.829', '019.556']
| ***DEGRADED cloud-config.service     | 001.15: ['001.240', '001.000', '001.206'] | 006.99: ['006.719', '007.036', '007.055', '007.153']
| ***DEGRADED modules-config/config-runcmd | 000.00: ['000.001', '000.001', '000.002', '000.001'] | 004.60: ['004.369', '004.594', '004.699', '004.748']
------------------- Control image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 034.58 | 035.23 | 033.67 | client_time_to_ssh
| 040.35 | 041.42 | 039.69 | client_time_to_cloudinit_done
| 003.88 | 003.91 | 003.86 | time_systemd_kernel
| 020.42 | 021.37 | 019.88 | time_systemd_userspace
| 006.27 | 006.40 | 006.17 | snapd.seeded.service
| 002.70 | 002.72 | 002.67 | snapd.apparmor.service
| 002.99 | 003.79 | 002.61 | cloud-init.service
| 002.10 | 002.12 | 002.09 | cloud-init-local.service
| 001.60 | 001.66 | 001.57 | dev-sda1.device
| 001.48 | 001.72 | 001.05 | systemd-networkd-wait-online.service
| 001.15 | 001.24 | 001.00 | cloud-config.service
| 001.25 | 001.36 | 001.17 | pollinate.service
| 001.25 | 001.40 | 001.17 | snapd.service
| 001.17 | 001.29 | 001.11 | apport.service
| 001.13 | 001.18 | 001.09 | snap.lxd.activate.service
| 001.07 | 001.07 | 001.07 | user@1000.service
| 000.85 | 000.88 | 000.80 | init-network/config-growpart
| 001.02 | 001.89 | 000.61 | init-network/config-ssh
| 000.53 | 000.54 | 000.50 | init-local/search-NoCloud
| 000.45 | 000.47 | 000.43 | modules-config/config-grub_dpkg
| 000.28 | 000.29 | 000.25 | init-network/config-users_groups
| 000.26 | 000.29 | 000.21 | init-network/config-resizefs
------------------- Updated cloud-init image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 038.20 | 039.58 | 037.36 | client_time_to_ssh
| 038.84 | 039.97 | 037.93 | client_time_to_cloudinit_done
| 003.87 | 003.90 | 003.84 | time_systemd_kernel
| 018.97 | 019.56 | 018.36 | time_systemd_userspace
| 006.99 | 007.15 | 006.72 | cloud-config.service
| 005.27 | 005.35 | 005.22 | snapd.seeded.service
| 002.79 | 003.43 | 002.43 | cloud-init.service
| 002.71 | 002.75 | 002.68 | snapd.apparmor.service
| 002.12 | 002.14 | 002.11 | cloud-init-local.service
| 001.60 | 001.64 | 001.57 | dev-sda1.device
| 001.34 | 001.39 | 001.29 | snapd.service
| 001.24 | 001.29 | 001.18 | pollinate.service
| 001.17 | 001.20 | 001.13 | apport.service
| 001.63 | 002.02 | 001.06 | systemd-networkd-wait-online.service
| 001.52 | 001.52 | 001.52 | cloud-final.service
| 004.60 | 004.75 | 004.37 | modules-config/config-runcmd
| 000.83 | 001.47 | 000.43 | init-network/config-ssh
| 000.85 | 000.87 | 000.84 | init-network/config-growpart
| 000.53 | 000.54 | 000.53 | init-local/search-NoCloud
| 000.48 | 000.53 | 000.44 | modules-config/config-grub_dpkg
| 000.28 | 000.31 | 000.27 | init-network/config-resizefs
| 000.25 | 000.26 | 000.24 | init-network/config-users_groups
| 000.11 | 000.11 | 000.09 | modules-final/config-keys_to_console

```
</details>

The runcmd launches represent `cloud-config.service` boot costs because we are now explicitly calling `snap wait system seed.loaded` in the runcmd module to block for 4.3 seconds on snap seeding instead of relying on systemd unit ordering After=snapd.seeded.service in `cloud-config.service` config.  This user-facing cost is not percieved as more than a 0.5 second delay in time to SSH because systemd allows `cloud-config.service` to start earlier and run other cloud-config modules prior to hitting the block on runcmd. So more work is done before blocking on snap seeding, which saves us some of that cost of waiting on snap seeding directly in the runcmd module.

This testing does represent that we may want to avoid arbitrarily blocking on `snap seed.loaded` in runcmd as most user-data may not truly depend on seeded snaps being present. We could retain this blocking behavior on stable releases by always calling `snap wait systemd seed.loaded`. For latest devel release, we should document that if your runcmd relies on snaps being present, the runcmd should start with a call to `snap wait systemd seed.loaded`. So that this cost is an opt-in when snaps are necessary to the runcmd operations.

This PR will need a followup PR for downstream `ubuntu/(focal|jammy|lunar|mantic)` series branches to strictly revert this commit because we don't want to risk corner-case exposure and change in behavior of systemd ordering that some customers may rely on.

<details>
<summary> Test with runcmd user-data on noble with no `cloud.run("snap-seeded", util.wait_for_snap_seeded, [], freq=PER_ONCE)`:  time to ssh delta < 0.7 seconds diff   cloud-config.service delta <  1.2 seconds
 </summary>
```
------------------- Boot speed comparison ---------------------------
Boot image serial: 20231211
Cloud-init from: cloud-init-drop-snapd-seeded-from-runcmd.deb
Source image: /srv/pycloudlib/noble/20231211/noble-server-cloudimg-amd64.img
Number of samples: 4
---------------------------------------------------------------------


--------------------- Performance Deltas Encountered ---------------------
| Delta type and service name        | Original avg/samples | New avg/samples
--------------------------------------------------------------------------
| ***         client_time_to_ssh       | 034.12: ['035.189', '032.459', '033.809', '035.009'] | 034.88: ['035.478', '034.135', '035.168', '034.744']
| ***         time_systemd_userspace   | 019.78: ['019.372', '019.216', '019.918', '020.617'] | 019.22: ['019.935', '018.898', '019.235', '018.828']
| ***DEGRADED cloud-config.service     | 001.13: ['001.173', '001.039', '001.168'] | 002.45: ['002.420', '002.460', '002.445', '002.486']
| ***DEGRADED init-network/config-ssh  | 000.56: ['000.623', '000.405', '000.520', '000.678'] | 000.79: ['000.732', '000.563', '000.731', '001.154']
------------------- Control image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 034.12 | 035.19 | 032.46 | client_time_to_ssh
| 040.07 | 040.85 | 039.49 | client_time_to_cloudinit_done
| 003.91 | 003.93 | 003.89 | time_systemd_kernel
| 019.78 | 020.62 | 019.22 | time_systemd_userspace
| 005.92 | 006.44 | 005.48 | snapd.seeded.service
| 002.71 | 002.80 | 002.67 | snapd.apparmor.service
| 002.53 | 002.67 | 002.42 | cloud-init.service
| 002.12 | 002.22 | 002.07 | cloud-init-local.service
| 001.64 | 001.71 | 001.60 | dev-sda1.device
| 001.80 | 002.01 | 001.50 | systemd-networkd-wait-online.service
| 001.22 | 001.33 | 001.13 | pollinate.service
| 001.13 | 001.17 | 001.04 | cloud-config.service
| 001.20 | 001.24 | 001.16 | snapd.service
| 001.12 | 001.17 | 001.06 | apport.service
| 000.85 | 000.87 | 000.82 | init-network/config-growpart
| 000.54 | 000.63 | 000.43 | modules-config/config-grub_dpkg
| 000.56 | 000.68 | 000.41 | init-network/config-ssh
| 000.53 | 000.57 | 000.51 | init-local/search-NoCloud
| 000.27 | 000.28 | 000.26 | init-network/config-users_groups
| 000.26 | 000.29 | 000.22 | init-network/config-resizefs
------------------- Updated cloud-init image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 034.88 | 035.48 | 034.13 | client_time_to_ssh
| 039.15 | 039.94 | 038.65 | client_time_to_cloudinit_done
| 003.87 | 003.88 | 003.86 | time_systemd_kernel
| 019.22 | 019.93 | 018.83 | time_systemd_userspace
| 006.27 | 006.83 | 005.72 | snapd.seeded.service
| 002.68 | 002.69 | 002.66 | snapd.apparmor.service
| 002.78 | 003.14 | 002.55 | cloud-init.service
| 002.45 | 002.49 | 002.42 | cloud-config.service
| 002.10 | 002.10 | 002.10 | cloud-init-local.service
| 001.48 | 001.87 | 001.05 | systemd-networkd-wait-online.service
| 001.61 | 001.67 | 001.56 | dev-sda1.device
| 001.38 | 001.41 | 001.35 | snapd.service
| 001.30 | 001.38 | 001.24 | pollinate.service
| 001.19 | 001.23 | 001.14 | apport.service
| 001.07 | 001.07 | 001.07 | user@1000.service
| 001.03 | 001.03 | 001.03 | snap.lxd.activate.service
| 000.85 | 000.90 | 000.83 | init-network/config-growpart
| 000.79 | 001.15 | 000.56 | init-network/config-ssh
| 000.53 | 000.53 | 000.52 | init-local/search-NoCloud
| 000.47 | 000.48 | 000.45 | modules-config/config-grub_dpkg
| 000.29 | 000.30 | 000.27 | init-network/config-resizefs
| 000.27 | 000.31 | 000.25 | init-network/config-users_groups
```
</details>

## Test Steps

```
./tools/run-container --package ubuntu-daily/noble
# Run simple ubuntu boot speed analysis script
python3 ../server-test-scripts/cloud-init/qemu_performance.py -c cloud-init*deb -n 3
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_lxd.py
+++ b/cloudinit/config/cc_lxd.py
@@ -210,6 +210,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             f" '{type(lxd_cfg).__name__}'"
         )
 
+    util.wait_for_snap_seeded(cloud)
     # Grab the configuration
     init_cfg = lxd_cfg.get("init", {})
     preseed_str = lxd_cfg.get("preseed", "")
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -65,7 +65,7 @@ from cloudinit import (
     url_helper,
     version,
 )
-from cloudinit.settings import CFG_BUILTIN
+from cloudinit.settings import CFG_BUILTIN, PER_ONCE
 
 _DNS_REDIRECT_IP = None
 LOG = logging.getLogger(__name__)
@@ -3092,6 +3092,18 @@ def wait_for_files(flist, maxwait, naplen=0.5, log_pre=""):
         "%sStill missing files after %s seconds: %s", log_pre, maxwait, need
     )
     return need
+
+
+def wait_for_snap_seeded(cloud):
+    """Helper to wait on completion of snap seeding."""
+
+    def callback():
+        if not subp.which("snap"):
+            LOG.debug("Skipping snap wait, no snap command present")
+            return
+        subp.subp(["snap", "wait", "system", "seed.loaded"])
+
+    cloud.run("snap-seeded", callback, [], freq=PER_ONCE)
 
 
 def mount_is_read_write(mount_point):
--- a/cloudinit/cloud.py
+++ b/cloudinit/cloud.py
@@ -57,6 +57,17 @@ class Cloud:
         return copy.deepcopy(self._cfg)
 
     def run(self, name, functor, args, freq=None, clear_on_fail=False):
+        """Run a function gated by a named semaphore for a desired frequency.
+
+        The typical case for this method would be to limit running of the
+        provided func to a single well-defined frequency:
+            PER_INSTANCE, PER_BOOT or PER_ONCE
+
+        The semaphore provides a gate that persists across cloud-init
+        boot stage boundaries so multiple modules can share this state
+        even if they happen to be run in different boot stages or across
+        reboots.
+        """
         return self._runners.run(name, functor, args, freq, clear_on_fail)
 
     def get_template_filename(self, name):
--- a/cloudinit/config/cc_snap.py
+++ b/cloudinit/config/cc_snap.py
@@ -191,7 +191,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             "Skipping module named %s, no 'snap' key in configuration", name
         )
         return
-
+    util.wait_for_snap_seeded(cloud)
     add_assertions(
         cfgin.get("assertions", []),
         os.path.join(cloud.paths.get_ipath_cur(), "snapd.assertions"),
--- a/cloudinit/config/cc_ubuntu_autoinstall.py
+++ b/cloudinit/config/cc_ubuntu_autoinstall.py
@@ -6,6 +6,7 @@ import logging
 import re
 from textwrap import dedent
 
+from cloudinit import util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import (
@@ -83,6 +84,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         )
         return
 
+    util.wait_for_snap_seeded(cloud)
     snap_list, _ = subp(["snap", "list"])
     installer_present = None
     for snap_name in LIVE_INSTALLER_SNAPS:
--- a/systemd/cloud-config.service.tmpl
+++ b/systemd/cloud-config.service.tmpl
@@ -2,7 +2,6 @@
 [Unit]
 Description=Apply the settings specified in cloud-config
 After=network-online.target cloud-config.target
-After=snapd.seeded.service
 Before=systemd-user-sessions.service
 Wants=network-online.target cloud-config.target
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
--- a/tests/integration_tests/modules/test_frequency_override.py
+++ b/tests/integration_tests/modules/test_frequency_override.py
@@ -1,6 +1,7 @@
 import pytest
 
 from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.releases import CURRENT_RELEASE
 
 USER_DATA = """\
 #cloud-config
@@ -17,6 +18,17 @@ def test_frequency_override(client: IntegrationInstance):
         in client.read_from_file("/var/log/cloud-init.log")
     )
     assert client.read_from_file("/var/tmp/hi").strip().count("hi") == 1
+    if CURRENT_RELEASE.os == "ubuntu":
+        if CURRENT_RELEASE.series in ("focal", "jammy", "lunar", "mantic"):
+            # Stable series will block on snapd.seeded.service and create a
+            # semaphore file
+            assert client.execute("test -f /var/lib/cloud/snap-seeded.once").ok
+        else:
+            # Newer series will not block on snapd.seeded.service nor create a
+            # semaphore file
+            assert not client.execute(
+                "test -f /var/lib/cloud/snap-seeded.once"
+            ).ok
 
     # Change frequency of scripts_user to always
     config = client.read_from_file("/etc/cloud/cloud.cfg")
--- a/tests/unittests/config/test_cc_lxd.py
+++ b/tests/unittests/config/test_cc_lxd.py
@@ -11,6 +11,8 @@ from cloudinit.config.schema import (
     get_schema,
     validate_cloudconfig_schema,
 )
+from cloudinit.helpers import Paths
+from cloudinit.util import del_file
 from tests.unittests import helpers as t_help
 from tests.unittests.util import get_cloud
 
@@ -44,7 +46,9 @@ class TestLxd(t_help.CiTestCase):
     )
     def test_lxd_init(self, maybe_clean, which, subp, exists, system_info):
         system_info.return_value = {"uname": [0, 1, "mykernel"]}
-        cc = get_cloud(mocked_distro=True)
+        tmpdir = self.tmp_dir()
+        sem_file = f"{tmpdir}/sem/snap_seeded.once"
+        cc = get_cloud(mocked_distro=True, paths=Paths({"cloud_dir": tmpdir}))
         install = cc.distro.install_packages
 
         for backend, cmd, package in BACKEND_DEF:
@@ -84,21 +88,23 @@ class TestLxd(t_help.CiTestCase):
             if backend == "lvm":
                 self.assertEqual(
                     [
+                        mock.call(sem_file),
                         mock.call(
                             "/lib/modules/mykernel/"
                             "kernel/drivers/md/dm-thin-pool.ko"
-                        )
+                        ),
                     ],
                     exists.call_args_list,
                 )
             else:
-                self.assertEqual([], exists.call_args_list)
+                self.assertEqual([mock.call(sem_file)], exists.call_args_list)
+            del_file(sem_file)
 
     @mock.patch("cloudinit.config.cc_lxd.maybe_cleanup_default")
     @mock.patch("cloudinit.config.cc_lxd.subp")
     @mock.patch("cloudinit.config.cc_lxd.subp.which", return_value=False)
     def test_lxd_install(self, m_which, mock_subp, m_maybe_clean):
-        cc = get_cloud()
+        cc = get_cloud(paths=Paths({"cloud_dir": self.tmp_dir()}))
         cc.distro = mock.MagicMock()
         mock_subp.which.return_value = None
         cc_lxd.handle("cc_lxd", LXD_INIT_CFG, cc, [])
@@ -112,7 +118,7 @@ class TestLxd(t_help.CiTestCase):
     @mock.patch("cloudinit.config.cc_lxd.maybe_cleanup_default")
     @mock.patch("cloudinit.config.cc_lxd.subp")
     def test_no_init_does_nothing(self, mock_subp, m_maybe_clean):
-        cc = get_cloud()
+        cc = get_cloud(paths=Paths({"cloud_dir": self.tmp_dir()}))
         cc.distro = mock.MagicMock()
         cc_lxd.handle("cc_lxd", {"lxd": {}}, cc, [])
         self.assertFalse(cc.distro.install_packages.called)
@@ -122,16 +128,17 @@ class TestLxd(t_help.CiTestCase):
     @mock.patch("cloudinit.config.cc_lxd.maybe_cleanup_default")
     @mock.patch("cloudinit.config.cc_lxd.subp")
     def test_no_lxd_does_nothing(self, mock_subp, m_maybe_clean):
-        cc = get_cloud()
+        cc = get_cloud(paths=Paths({"cloud_dir": self.tmp_dir()}))
         cc.distro = mock.MagicMock()
         cc_lxd.handle("cc_lxd", {"package_update": True}, cc, [])
         self.assertFalse(cc.distro.install_packages.called)
         self.assertFalse(mock_subp.subp.called)
         self.assertFalse(m_maybe_clean.called)
 
+    @mock.patch("cloudinit.config.cc_lxd.util.wait_for_snap_seeded")
     @mock.patch("cloudinit.config.cc_lxd.subp")
-    def test_lxd_preseed(self, mock_subp):
-        cc = get_cloud()
+    def test_lxd_preseed(self, mock_subp, wait_for_snap_seeded):
+        cc = get_cloud(paths=Paths({"cloud_dir": self.tmp_dir()}))
         cc.distro = mock.MagicMock()
         cc_lxd.handle(
             "cc_lxd",
@@ -146,6 +153,7 @@ class TestLxd(t_help.CiTestCase):
             ],
             mock_subp.subp.call_args_list,
         )
+        wait_for_snap_seeded.assert_called_once_with(cc)
 
     def test_lxd_debconf_new_full(self):
         data = {
--- a/tests/unittests/config/test_cc_snap.py
+++ b/tests/unittests/config/test_cc_snap.py
@@ -293,8 +293,11 @@ class TestSnapSchema:
 
 
 class TestHandle:
+    @mock.patch("cloudinit.util.wait_for_snap_seeded")
     @mock.patch("cloudinit.config.cc_snap.subp.subp")
-    def test_handle_adds_assertions(self, m_subp, fake_cloud, tmpdir):
+    def test_handle_adds_assertions(
+        self, m_subp, wait_for_snap_seeded, fake_cloud, tmpdir
+    ):
         """Any configured snap assertions are provided to add_assertions."""
         assert_file = os.path.join(
             fake_cloud.paths.get_ipath_cur(), "snapd.assertions"
@@ -309,3 +312,4 @@ class TestHandle:
         assert util.load_text_file(compare_file) == util.load_text_file(
             assert_file
         )
+        wait_for_snap_seeded.assert_called_once_with(fake_cloud)
--- a/tests/unittests/config/test_cc_ubuntu_autoinstall.py
+++ b/tests/unittests/config/test_cc_ubuntu_autoinstall.py
@@ -11,6 +11,7 @@ from cloudinit.config.schema import (
     get_schema,
     validate_cloudconfig_schema,
 )
+from cloudinit.helpers import Paths
 from tests.unittests.helpers import skipUnlessJsonSchema
 from tests.unittests.util import get_cloud
 
@@ -64,18 +65,20 @@ class TestvalidateConfigSchema:
             cc_ubuntu_autoinstall.validate_config_schema(src_cfg)
 
 
+@mock.patch(MODPATH + "util.wait_for_snap_seeded")
 @mock.patch(MODPATH + "subp")
 class TestHandleAutoinstall:
     """Test cc_ubuntu_autoinstall handling of config."""
 
     @pytest.mark.parametrize(
-        "cfg,snap_list,subp_calls,logs",
+        "cfg,snap_list,subp_calls,logs,snap_wait_called",
         [
             pytest.param(
                 {},
                 SAMPLE_SNAP_LIST_OUTPUT,
                 [],
                 ["Skipping module named name, no 'autoinstall' key"],
+                False,
                 id="skip_no_cfg",
             ),
             pytest.param(
@@ -87,6 +90,7 @@ class TestHandleAutoinstall:
                     " installer snap packages to be present: subiquity,"
                     " ubuntu-desktop-installer"
                 ],
+                True,
                 id="valid_autoinstall_schema_checks_snaps",
             ),
             pytest.param(
@@ -97,6 +101,7 @@ class TestHandleAutoinstall:
                     "Valid autoinstall schema. Config will be processed by"
                     " subiquity"
                 ],
+                True,
                 id="valid_autoinstall_schema_sees_subiquity",
             ),
             pytest.param(
@@ -107,19 +112,33 @@ class TestHandleAutoinstall:
                     "Valid autoinstall schema. Config will be processed by"
                     " ubuntu-desktop-installer"
                 ],
+                True,
                 id="valid_autoinstall_schema_sees_desktop_installer",
             ),
         ],
     )
     def test_handle_autoinstall_cfg(
-        self, subp, cfg, snap_list, subp_calls, logs, caplog
+        self,
+        subp,
+        wait_for_snap_seeded,
+        cfg,
+        snap_list,
+        subp_calls,
+        logs,
+        snap_wait_called,
+        caplog,
+        tmpdir,
     ):
         subp.return_value = snap_list, ""
-        cloud = get_cloud(distro="ubuntu")
+        cloud = get_cloud(distro="ubuntu", paths=Paths({"cloud_dir": tmpdir}))
         cc_ubuntu_autoinstall.handle("name", cfg, cloud, None)
         assert subp_calls == subp.call_args_list
         for log in logs:
             assert log in caplog.text
+        if snap_wait_called:
+            wait_for_snap_seeded.assert_called_once_with(cloud)
+        else:
+            wait_for_snap_seeded.assert_not_called()
 
 
 class TestAutoInstallSchema:
--- a/tests/unittests/util.py
+++ b/tests/unittests/util.py
@@ -33,7 +33,9 @@ def get_cloud(
         myds.metadata.update(metadata)
     if paths:
         paths.datasource = myds
-    return cloud.Cloud(myds, paths, sys_cfg, mydist, None)
+    return cloud.Cloud(
+        myds, paths, sys_cfg, mydist, runners=helpers.Runners(paths)
+    )
 
 
 def abstract_to_concrete(abclass):


cloudinit/util.py
```diff
@@ -3073,6 +3073,14 @@ def wait_for_files(flist, maxwait, naplen=0.5, log_pre=""):
     return need
 
 
+def wait_for_snap_seeded():
+    """Helper to wait on completion of snap seeding."""
+    if not subp.which("snap"):
```
holmanb, line 3078: Will Ubuntu ever not have `snap` installed? All of the callsites are ubuntu-only modules, so I'm curious why we feel the need for this check.

holmanb: I guess it won't hurt anything, I just assume that it is unnecessary

blackboxsw: Good point, while all ubuntu images at the present time all have snaps included or seeded in their images, this may change if the future for specific minimal images. This was just a preliminary pre-flight check to ensure we don't go down a path that would `raise CalledProcessError` on a missing command if we happen to be in an env that doesn't support snaps for some reason.

cloudinit/config/cc_lxd.py
```diff
@@ -210,6 +210,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             f" '{type(lxd_cfg).__name__}'"
         )
 
+    cloud.run("snap-seeded", util.wait_for_snap_seeded, [], freq=PER_ONCE)
```
holmanb, line 213: I don't think that this is in use yet in any of our configuration modules, which I think is intentional, since the modules themselves are run with a configurable frequency. I'm curious about the benefits / costs of using this interface. 

### Specific to this use case

Are we sure that `PER_ONCE` works in all cases? Is forcing snap to un-seed possible? If so, I'd lean towards running `wait_for_snap_seeded()` unconditionally in these three modules.

Is using a fs-based semaphore cheaper than unconditionally using `snap wait system seed.loaded`? The command runs in <0.05 for me.

### More generally about `cloud.run()`

This interface has overlapping behavior and intent with `lru_cache`. `lru_cache` should only persist across stages when pickled, and I see that this interface also has the advantage of flexibility over `lru_cache()` for aggregating multiple callsites together under a single semaphore.

Do you think that we should we use `cloud.run()` over `lru_cache` in other cases as well, to avoid per-stage calls?

In this section of my comment, the first paragraph and answers to questions in the second would be well suited for the docstring of `cloud.run()`. This would make reviewing and maintenance of this code to be easier.

holmanb (commented):
@blackboxsw some questions about `cloud.run()`

blackboxsw:
> @blackboxsw some questions about `cloud.run()`
> 
> Additionally:
> 
> > This PR will need a followup PR for downstream ubuntu/(focal|jammy|lunar|mantic) series branches to also add the following quilt patch to retain current stable behavior for runcmd:
> 
> This is insufficient. We need a full revert of this commit, not just a patched `runcmd`. Cloud-init has other supported interfaces (boot hooks, etc) that will experience a change in behavior as a result of this commit.


Ahh you are right, but not specifically because of `boot-hooks` which are run in the init network stage (before cloud-config.service anyway).

Since we are dropping `After=snapd-seeded.service` from cloud-config.service, we could potentially also affect any of the following supplemental scripts which are run by default in cloud-final.stage.

  - scripts_vendor
  - scripts_per_once
  - scripts_per_boot
  - scripts_per_instance
  - scripts_user

To easily remedy that situation in all cases on stable series, I think it's truly worth us pushing the `After=snapd.seeded.service` into cloud-final.service configuration on stable releases. That'd ensure we don't expose custom scripts which are run late in boot to snap seeding issues if those scripts relied on snaps being present. Are there other areas that you think wouldn't be considered if we retain the wait on snapd seeding in the specific config modules lxd/ubuntu_advantage/snap, add a block in runcmd and also add an `After=snapd.seeded.service` to cloud-final.service?


I've also pushed doc notes to this PR for any of the cc_script* modules to note that upstream versions with scripts relying on snap seeding will have to call `snap wait system seed.loaded`

blackboxsw:
Re-ran performance comparison script showing no degradation in "time to SSH" on instances even with providing the following snap user-data which triggers a snap wait systemd seeded.service from the cc_snap module. note "degradated" state of cloud-config.service (and cc_snap module) because of the wait on snap seeding within the service instead of due to blocking 

user-data:
```
#cloud-config
snap:
  commands:
    00: snap list
```

<details>
 <summary> perf testing of snap userdata: no delta in time to SSH. Blocking cost seen in config-snap module/cloud-config.service due to snap wait call</summary>

```
------------------- Boot speed comparison ---------------------------
Boot image serial: 20231213
Cloud-init from: cloud-init-noble-no-snap-seeded.deb
Source image: pre-existing samples
Number of samples: 4
---------------------------------------------------------------------


--------------------- Performance Deltas Encountered ------------------------
| Control Avg | Upgrade Avg | Avg delta | Delta type and service name
-----------------------------------------------------------------------------
|      034.02 |      033.52 |   -00.50s | ***         client_time_to_ssh      
|      019.74 |      019.34 |   -00.40s | ***         time_systemd_userspace  
|      001.16 |      008.30 |   +07.13s | ***DEGRADED cloud-config.service    
|      000.08 |      005.86 |   +05.78s | ***DEGRADED modules-config/config-snap
|      000.93 |      000.73 |   -00.20s | ***IMPROVED init-network/config-ssh 
|      000.69 |      006.53 |   +05.84s | ***DEGRADED stage/modules-config    
------------------- Control image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 034.02 | 034.59 | 033.36 | client_time_to_ssh
| 039.28 | 040.11 | 038.44 | client_time_to_cloudinit_done
| 003.88 | 003.90 | 003.87 | time_systemd_kernel
| 019.74 | 020.51 | 018.95 | time_systemd_userspace
| 005.88 | 006.25 | 005.61 | snapd.seeded.service
| 002.67 | 002.69 | 002.66 | snapd.apparmor.service
| 002.89 | 004.10 | 002.37 | cloud-init.service
| 002.10 | 002.11 | 002.09 | cloud-init-local.service
| 001.58 | 001.65 | 001.53 | dev-sda1.device
| 001.42 | 001.94 | 001.10 | systemd-networkd-wait-online.service
| 001.19 | 001.26 | 001.12 | pollinate.service
| 001.17 | 001.17 | 001.17 | user@1000.service
| 001.18 | 001.26 | 001.15 | snapd.service
| 001.10 | 001.18 | 001.05 | apport.service
| 001.10 | 001.10 | 001.10 | snap.lxd.activate.service
| 001.16 | 001.27 | 001.05 | cloud-config.service
| 000.87 | 000.91 | 000.83 | init-network/config-growpart
| 000.53 | 000.54 | 000.52 | init-local/search-NoCloud
| 000.48 | 000.63 | 000.40 | modules-config/config-grub_dpkg
| 000.93 | 002.12 | 000.40 | init-network/config-ssh
| 000.27 | 000.27 | 000.26 | init-network/config-resizefs
| 000.25 | 000.26 | 000.24 | init-network/config-users_groups
| 001.19 | 001.19 | 001.18 | stage/init-local
| 002.50 | 003.71 | 001.96 | stage/init-network
| 000.69 | 000.87 | 000.58 | stage/modules-config
| 000.16 | 000.17 | 000.15 | stage/modules-final
------------------- Updated cloud-init image ---------------------------
| Avg    |  Max   |  Min   | Metric Name
------------------------------------------------------------------
| 033.52 | 034.16 | 032.23 | client_time_to_ssh
| 038.93 | 039.50 | 038.71 | client_time_to_cloudinit_done
| 003.86 | 003.90 | 003.82 | time_systemd_kernel
| 019.34 | 020.18 | 018.91 | time_systemd_userspace
| 008.30 | 009.01 | 007.75 | cloud-config.service
| 006.00 | 006.37 | 005.79 | snapd.seeded.service
| 002.65 | 002.66 | 002.64 | snapd.apparmor.service
| 002.66 | 003.01 | 002.42 | cloud-init.service
| 002.09 | 002.10 | 002.08 | cloud-init-local.service
| 001.46 | 001.86 | 001.04 | systemd-networkd-wait-online.service
| 001.60 | 001.65 | 001.57 | dev-sda1.device
| 001.37 | 001.40 | 001.34 | snapd.service
| 001.25 | 001.26 | 001.25 | pollinate.service
| 001.21 | 001.23 | 001.19 | apport.service
| 001.10 | 001.10 | 001.10 | user@1000.service
| 005.86 | 006.47 | 005.40 | modules-config/config-snap
| 000.85 | 000.90 | 000.81 | init-network/config-growpart
| 000.73 | 001.04 | 000.45 | init-network/config-ssh
| 000.52 | 000.53 | 000.51 | init-local/search-NoCloud
| 000.53 | 000.66 | 000.36 | modules-config/config-grub_dpkg
| 000.25 | 000.27 | 000.24 | init-network/config-users_groups
| 000.26 | 000.28 | 000.24 | init-network/config-resizefs
| 001.18 | 001.18 | 001.18 | stage/init-local
| 002.28 | 002.62 | 002.02 | stage/init-network
| 006.53 | 007.28 | 005.90 | stage/modules-config
| 000.16 | 000.25 | 000.13 | stage/modules-final
```
</details>

cloudinit/config/cc_runcmd.py
```diff
@@ -49,6 +49,12 @@
 
     when writing files, do not use /tmp dir as it races with
     systemd-tmpfiles-clean LP: #1707222. Use /run/somedir instead.
+
+.. note::
```
holmanb, line 53: Are we sure that this note belongs here?

Our docs have many little notes like this that might be relevant to their adjacent section, but don't serve the majority of users. This [comment I just made on the boothook docs PR](https://github.com/canonical/cloud-init/pull/4650/files#r1427264695) references another example of is problem in our docs. Both that comment and this one point out what seems to me like a smell, and I think we can improve it. Yes, the information you added might be relevant to a user that is reading this section, but a user looking in that section isn't necessarily using snap commands, so it's not relevant to everyone. This note is relevant only to Ubuntu users, and only the subset which want to run snap commands - definitely not to all users of `cc_runcmd`.

This is an example of a standard documentation practice in this project that I think that we can improve on. Many cloud-init docs were written in such a way that relevant information gets grouped together on adjacent pages, however this dilutes the purpose of the original section and leaves users with information that is irrelevant to them, or leaves them with unanswered questions. Writing documentation in this style means that our documentation looks a lot more like a large collection of "notes" than reference material or a user's manual. I think we should prefer to put information in sections that are relevant to all users of that section, rather than having a long tail of implementation details following each section that only maybe applies. The reason for preferring this is that maybe-relevant-notes after sections makes the experience unwieldy to the user. As a user, am I supposed to read all module notes to know how to use the module? If yes, and if we add a note like this to every affected module every time we change something in upstream cloud-init, then I as a user end up reading lots of irrelevant information for my use-case.

I'd like to make the case that the appropriate spot for this would be in our Changelog, maybe in a section titled "breaking changes", or something like that. If we wrote this module from scratch, I don't think that we would have written the `cc_runcmd` documentation like this with an Ubuntu-specific callout. Maybe if we think this requires a dedicated spot in the docs, it could go in a distro-specific page? A snap-specific page? I think that this would increase the signal to noise ratio in our docs over copy-pasting implementation details to all parts of the docs where the details may be relevant.

Same comment to the other docs changes.

holmanb: @s-makin Thoughts on this?

holmanb (commented):
> Are there other areas that you think wouldn't be considered

I don't know - that's why I'm concerned about behavior change risk on older Ubuntu releases and recommended reverting this change in them rather than crafting a patch to allow reordering. Do we have any other earlier modules which pass user inputs to `subp()`?

What about a user that uses `bootcmd` to write a unit file (and then calls `systemctl daemon-reload`) that calls a snap command before the cloud-final.service? That would certainly cause a behavior change on old releases. Would that change in behavior matter to us? I don't think that there is anything that we can do for such a user if we backport with the approach proposed here.

I think we had agreed on reverting on old Ubuntu series because of the risk of backporting, and I still lean towards that. I don't think that we can do what you suggest safely, so I'm -1 for going that route.

cloudinit/config/cc_runcmd.py
s-makin: As a general principle, I completely agree. I would like us to try and avoid (or at least minimise) the use of notes and asides by finding appropriate "homes" for all these little callouts to live in. I suspect it'll be a bit of a Pandora's box to identify and rehome them all, but probably a worthwhile effort if it helps to streamline the user experience. First step, I think, will be to identify and gather them all so we can gauge the scope of the problem. It might need time on the next roadmap, even if it's a small one.

If there are parts of the docs that *require* distro-specific callouts, it could be worth considering using synced tabs across the docs so that users can select their distro the first time they encounter the tabset, and then only get shown the relevant content by default (instead of ploughing through a list of callouts).

blackboxsw:
> I think we had agreed on reverting on old Ubuntu series because of the risk of backporting, and I still lean towards that. I don't think that we can do what you suggest safely, so I'm -1 for going that route.

Agreed, it is safer to just avoid changing systemd ordering behavior on all stable releases here unless there is an explicit gain to be had as there are other corner cases that may not be covered by this changeset. If we really need to improve boot performance on stable releases, we can entertain this functionality and stress test it well before introducing it into the stable release stream. As it is, we'll make sure we have a clean revert of this functionality in stable releases to avoid incidental change in behavior there.

blackboxsw:
> Additionally:
> 
> > This PR will need a followup PR for downstream ubuntu/(focal|jammy|lunar|mantic) series branches to also add the following quilt patch to retain current stable behavior for runcmd:
> 
> This is insufficient. We need a full revert of this commit, not just a patched `runcmd`. Cloud-init has other supported interfaces (boot hooks, etc) that will experience a change in behavior as a result of this commit.

I've updated the PR description to represent that we will revert this behavior in stable releases to avoid risk of change in behavior due to systemd ordering changes.

cloudinit/config/cc_runcmd.py
blackboxsw: This works for me, I've dropped the snap-related  notes in runcmd and and the scripts in general and represented this `BREAKING CHANGE` in the commit message body as well as in the commit summary with the `!`. I think it's worth us adding process in our release note generation to call-out those breaking changes both in github releases and launchpad releases we represent. For discoverability of BREAKING CHANGES in commits, I suggest we also add the trailing bang `!` to the shortened commit summary for related commits/features that do diverge. If agreed, I'll add a PR to [improve our 
pull request template](https://github.com/canonical/cloud-init/blob/main/.github/PULL_REQUEST_TEMPLATE.md) to better document this commit message practice for approaching BREAKING CHANGES.

cloudinit/config/cc_lxd.py
blackboxsw: lru_cache won't persist even in our pickled objects such as our Datasource. If I add decorate a datasource method as lru_cached and call it per boot stage `.cache_info()` still shows it as empty every single boot stage even with deserialization.
`CacheInfo(hits=0, misses=0, maxsize=128, currsize=0)`.

We'd have to setup a file-based memory backend I think (maybe `joblib.Memory`?) to persist and reference that cache across boot stages to load and reference previously cached method/function calls.

So I think our `cloud.run()` semaphore file is the simplest way we can persist  this cached state across our separate python invocations in each boot stage.

1. PER_ONCE directive will work in all cases because we can't un-seed snaps. seeded snaps could be removed after seeding takes place with external snap commands, but I think worrying about that is off-path/unsupported by cloud-init.
2. the `lru_cache` will only apply once a given function has been memoized. 
3. Is fs-based semaphore cheaper than subp `snap wait system seed.loaded`? Yes much cheaper
Here is a timeit run of 1000 attempts for cloud.run vs subp shelling out to call snap wait system seed.loaded:
```
root@nnn:~# python3 1.py 
use cloud.run semaphore
0.0023272480175364763
invoke subp.subp(snap wait system seed.loaded
0.5457840269955341
```

I'll add a docstr to cloud.run to better explain the use-cases for this method.

blackboxsw:
CC: @enr0n for additional review given that cloud-init is looking to adapt our systemd service ordering in Ubuntu Noble and later it'd be good to get your two cents on this or at least provide some context.  The intent will be to avoid changing cloud-init's systemd unit/service ordering on all stable releases so that we don't change behavior there.

enr0n:
> CC: @enr0n for additional review given that cloud-init is looking to adapt our systemd service ordering in Ubuntu Noble and later it'd be good to get your two cents on this or at least provide some context. The intent will be to avoid changing cloud-init's systemd unit/service ordering on all stable releases so that we don't change behavior there.

This seems fine in principle. But to make sure I understand the boot time impact, is it the case that this causes a `3s` delay in "boot to SSH", but *only* when lxd, snap, or ubuntu_autoinstall are needed?

catmsred (approved):

blackboxsw:
> > CC: @enr0n for additional review given that cloud-init is looking to adapt our systemd service ordering in Ubuntu Noble and later it'd be good to get your two cents on this or at least provide some context. The intent will be to avoid changing cloud-init's systemd unit/service ordering on all stable releases so that we don't change behavior there.
> 
> This seems fine in principle. But to make sure I understand the boot time impact, is it the case that this causes a `3s` delay in "boot to SSH", but _only_ when lxd, snap, or ubuntu_autoinstall are needed?

Thanks @enr0n. The only negative impacts to boot speed at all will be if user-data contains any of the cloud-config keys: `lxd`, `ubuntu_advantage` or `snap` which is a small percentage of use-cases.

Also, it isn't really an 3s overall delay in time to SSH (more like a 0.5s or less delay due to a subprocess call out to snap from python) when those keys are used.

What we may see is that `systemd-analyze blame` shows cloud-config.service as taking a total of 3 seconds longer that it did with the prior/explicit systemd ordering `After=snap.seeded.service` in the cloud-config.service config. But, blame delta is because systemd previously would not start cloud-config.service until snapd.seeded.service was finished. This meant systemd didn't attribute cloud-config.service boot costs at all until snap seeding was done.

With this PR we'll drop that strict/explicit system ordering after snap seeding to favor the general use case for cloud-init which doesn't rely on lxd/snap/ubuntu_advantage user-data configuration. But, if those configs are needed, cloud-config.service waiting on snap seeding will not be represented explicitly from `systemd-anaylze critical-chain cloud-config.service` because it will be handled in the the python config modules for the lxd/snap/UA which will be shelling out to `snap wait system seeded`.

enr0n:
> > > CC: @enr0n for additional review given that cloud-init is looking to adapt our systemd service ordering in Ubuntu Noble and later it'd be good to get your two cents on this or at least provide some context. The intent will be to avoid changing cloud-init's systemd unit/service ordering on all stable releases so that we don't change behavior there.
> > 
> > 
> > This seems fine in principle. But to make sure I understand the boot time impact, is it the case that this causes a `3s` delay in "boot to SSH", but _only_ when lxd, snap, or ubuntu_autoinstall are needed?
> 
> Thanks @enr0n. The only negative impacts to boot speed at all will be if user-data contains any of the cloud-config keys: `lxd`, `ubuntu_advantage` or `snap` which is a small percentage of use-cases.
> 
> Also, it isn't really an 3s overall delay in time to SSH (more like a 0.5s or less delay due to a subprocess call out to snap from python) when those keys are used.
> 
> What we may see is that `systemd-analyze blame` shows cloud-config.service as taking a total of 3 seconds longer that it did with the prior/explicit systemd ordering `After=snap.seeded.service` in the cloud-config.service config. But, blame delta is because systemd previously would not start cloud-config.service until snapd.seeded.service was finished. This meant systemd didn't attribute cloud-config.service boot costs at all until snap seeding was done.
> 
> With this PR we'll drop that strict/explicit system ordering after snap seeding to favor the general use case for cloud-init which doesn't rely on lxd/snap/ubuntu_advantage user-data configuration. But, if those configs are needed, cloud-config.service waiting on snap seeding will not be represented explicitly from `systemd-anaylze critical-chain cloud-config.service` because it will be handled in the the python config modules for the lxd/snap/UA which will be shelling out to `snap wait system seeded`.

Okay, that makes sense now. Thanks for clarifying!

cloudinit/util.py
```diff
@@ -3094,6 +3094,14 @@ def wait_for_files(flist, maxwait, naplen=0.5, log_pre=""):
     return need
 
 
+def wait_for_snap_seeded():
```
holmanb, line 3097: This approach results in the same function call, which uses the same magic string and callback in multiple configuration modules.


Something like this would be cleaner I think:
```python
def wait_for_snap_seeded():
    """Helper to wait on completion of snap seeding."""
    def callback():
        if not subp.which("snap"):
            LOG.debug("Skipping snap wait, no snap command present")
            return
        subp.subp(["snap", "wait", "system", "seed.loaded"])
    cloud.run("snap-seeded", callback, [], freq=PER_ONCE)
```

that would turn our various calls like this:
```python
    cloud.run("snap-seeded", util.wait_for_snap_seeded, [], freq=PER_ONCE)
```

into this:
```python
    util.wait_for_snap_seeded()
```

holmanb (commented):
Thanks @blackboxsw. I have one style preference request, but otherwise I am much more comfortable with the approach outlined than before.

On the Ubuntu side of things do you think we will want feedback from an SRU reviewer, given the systemd changes?

holmanb (changes requested):
see previous comment

cloudinit/util.py
blackboxsw: Thanks @holmanb for the review. What this suggestion doesn't capture though is we would also need to provide the cloud as a param into `wait_for_snap_seeded`. I had wanted to keep dependency util function on Cloud instances to a minimal in `util` just for separation of roles/functions of util module functions and leaving this function as one that could be called outside the scope of an environment that has access to a `Cloud` instance. If we tightly couple the `wait_for_snap_seeded`  function to the presence of the cloud instance, it strictly limits call-sites to having access to the `Cloud` instance.   Maybe that dependency is ok though as we have no other call-sites in mind today.

blackboxsw: @holmanb my defense aside. I can't immediately think of other cases where we may intend as call-sites of `wait_for_snap_seeded` outside a scope where `cloud` is already present. So, let's design for today's call-sites. If we have to revist for other use-cases we'll be better placed/aware of a design that incorporates both needs. And it's trivial to fix that if we ever need to.

holmanb (approved):
@blackboxsw Looks good now, thanks for working on this!

cloudinit/config/cc_runcmd.py
holmanb: > or discoverability of BREAKING CHANGES in commits, I suggest we also add the trailing bang ! to the shortened commit summary for related commits/features that do diverge

@blackboxsw I meant to comment on this before, but I'm not sure that I agree that we want to require this. It's not intuitive what that means, and while it is in the specification I don't see why that would add anything that BREAKING CHANGE doesn't already accomplish (and more intuitively)

PR #4689, opened 2023-12-13
Make logger name more useful

holmanb:
## Additional Context
Using a logger name of `__name__` is useful to identify the source of a log.... Except when that source is named `__init__.py` because we have a few of those that contain lots of code. Use the module's directory name in that case.

## Merge type

### DIFF
--- a/cloudinit/log.py
+++ b/cloudinit/log.py
@@ -166,6 +166,18 @@ def setup_backup_logging():
     logging.Handler.handleError = handleError
 
 
+class CloudInitLogRecord(logging.LogRecord):
+    """reporting the filename as __init__.py isn't very useful in logs
+
+    if the filename is __init__.py, use the parent directory as the filename
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if "__init__.py" == self.filename:
+            self.filename = os.path.basename(os.path.dirname(self.pathname))
+
+
 def configure_root_logger():
     """Customize the root logger for cloud-init"""
 
@@ -179,3 +191,6 @@ def configure_root_logger():
     handler = LogExporter()
     handler.setLevel(logging.WARN)
     logging.getLogger().addHandler(handler)
+
+    # LogRecord allows us to report more useful information than __init__.py
+    logging.setLogRecordFactory(CloudInitLogRecord)


TheRealFalcon (commented):
I think all this did was filter those logs out entirely. E.g., if I search the logs for "Looking for data source" (logged from `sources/__init__.py`), I get nothing.

aciba90:
I think the format is specified here:

https://github.com/canonical/cloud-init/blob/c071f69ac2449ccf95641040976f8bdb9b251c5d/config/cloud.cfg.d/05_logging.cfg#L38

So we would have to change `filename to `module`. More info: https://docs.python.org/3/library/logging.html#logrecord-attributes, in all formatters in that file probably.

aciba90:
Or changing `LOG = logging.getLogger(__name__)` to `LOG = logging.getLogger(__module__)` in `__init__.py` modules to let the formatter calculate the filename from the fully qualified module name.

holmanb:
> So we would have to change filename to module. More info: https://docs.python.org/3/library/logging.html#logrecord-attributes, in all formatters in that file probably.

Hmmmm, so when I try using this, I still see:
```log
2023-12-14 17:55:15,451 - __init__.py[DEBUG]: Looking for data source in: ['NoCloud', 'LXD', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM']
```

> Or changing `LOG = logging.getLogger(__name__)` to `LOG = logging.getLogger(__module__)` in `__init__.py` modules to let the formatter calculate the filename from the fully qualified module name.

It doesn't look like `__module__` is defined.

```
>>> import cloudinit.distros
>>> cloudinit.distros.__module__
AttributeError: module 'cloudinit.distros' has no attribute '__module__'
>>> cloudinit.distros
<module 'cloudinit.distros' from '/home/holmanb/cloud-init/cloudinit/distros/__init__.py'>
>>> dir(cloudinit.distros)
['ALLOW_EC2_MIRRORS_ON_NON_AWS_INSTANCE_TYPES', 'ALL_DISTROS', 'Any', 'Dict', 'Distro', 'LDH_ASCII_CHARS', 'LOG', 'LinuxNetworking', 'List', 'Mapping', 'MutableMapping', 'Networking', 'OSFAMILIES', 'Optional', 'PREFERRED_NTP_CLIENTS', 'PackageInstallerError', 'PackageList', 'PackageManager', 'Renderer', 'Set', 'StringIO', 'Tuple', 'Type', 'Union', '_EC2_AZ_RE', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '_apply_hostname_transformations_to_url', '_get_arch_package_mirror_info', '_get_package_mirror_info', '_sanitize_mirror_url', 'abc', 'activators', 'bsd_utils', 'defaultdict', 'dhcp', 'eni', 'fetch', 'helpers', 'hosts', 'importer', 'iproute2', 'known_package_managers', 'logging', 'net', 'network_state', 'networking', 'os', 'package_management', 'parse_net_config_data', 'parsers', 'persistence', 're', 'renderers', 'set_etc_timezone', 'ssh_util', 'stat', 'string', 'subp', 'temp_utils', 'type_utils', 'urllib', 'uses_systemd', 'util']
```

But I think I can make this work by setting a custom `LogRecord`. I'll update the PR shortly.

aciba90:
Sorry, I meant `__package__` instead of `__module__`.

holmanb:
> Sorry, I meant `__package__` instead of `__module__`.

Also didn't work.

```
2023-12-14 20:24:27,306 - __init__.py[DEBUG]: Looking for data source in: ['NoCloud', 'LXD', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM']
```

But this PR currently does what we want, I think

TheRealFalcon (approved):
LGTM!

Issue #4691, opened 2023-12-13
Duplicated apt sources/list


simondeziel:
# Bug report
For Noble, cloud-init adds `/etc/apt/sources.list.d/ubuntu.sources` but those are are already defined in `/etc/apt/sources.list`. This leads apt to emit many warnings about sources being duplicated.

## Steps to reproduce the problem
```
$ lxc launch ubuntu-daily:24.04 c1
$ lxc exec c1 -- cloud-init status --wait
$ lxc exec c1 -- apt-get update
Hit:1 http://security.ubuntu.com/ubuntu noble-security InRelease
Get:2 http://archive.ubuntu.com/ubuntu noble InRelease [240 kB]
Hit:3 http://archive.ubuntu.com/ubuntu noble-updates InRelease
...
Fetched 27.9 MB in 3s (8032 kB/s)                                   
Reading package lists... Done
W: Target Packages (main/binary-amd64/Packages) is configured multiple times in /etc/apt/sources.list:37 and /etc/apt/sources.list.d/ubuntu.sources:2
W: Target Packages (main/binary-all/Packages) is configured multiple times in /etc/apt/sources.list:37 and /etc/apt/sources.list.d/ubuntu.sources:2
W: Target Translations (main/i18n/Translation-en) is configured multiple times in /etc/apt/sources.list:37 and /etc/apt/sources.list.d/ubuntu.sources:2
...
```

## Environment details
- Cloud-init version: `23.4-0ubuntu1`
- Operating System Distribution: `Ubuntu Noble Numbat (development branch)`
- Cloud provider, platform or installer type: LXD

## cloud-init logs

[cloud-init.tar.gz](https://github.com/canonical/cloud-init/files/13666034/cloud-init.tar.gz)

aciba90:
Thanks, @simondeziel, for reporting this and making cloud-init better.

I can confirm this is reproducible in noble. Those warnings are not hard errors and apt works correctly, but having two sources of config could lead to conflicts / unexpected behavior / hard apt errors in the future. In addition to that, the warnings could be a concern for users, so I think this is a bug that has to be fixed.

One possible simple and safe way to solve this would be, if features.APT_DEB822_SOURCE_LIST_FILE and /etc/apt/source.list is present, then do work with the classic sources lists and emit a warning informing the user about the fact that deb822 style has not been used.

[...]

PR #4699, opened 2023-12-14
fix(apt_configure): disable source.list if rendering deb822

aciba90:
## Proposed Commit Message

```
fix(apt_configure): disable sources.list if rendering deb822
    
If deb822 sources format is used and a preexistent
/etc/apt/sources.list exists, then disable the old sources file.
This prevents user-facing warnings while operating with apt or
apt-get and other possible conflicts.
    
This is safe to do as cloud-init owns the base APT configuration which
tipically/historically involved overwriting /etc/apt/sources.list

Fixes GH-4691
LP: #2045086
```

## Additional Context

I think it is safe to disable `/etc/apt/source.list`, because previously cloud-init override it, before https://github.com/canonical/cloud-init/pull/4437, or now if features.APT_DEB822_SOURCE_LIST_FILE is false or if the provided sources config has the classic style.
https://github.com/canonical/cloud-init/issues/4691

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -604,7 +604,6 @@ def get_apt_cfg() -> Dict[str, str]:
             "Dir::Etc::sourceparts", DEFAULT_APT_CFG["Dir::Etc::sourceparts"]
         )
     except ImportError:
-
         try:
             apt_dump, _ = subp.subp(["apt-config", "dump"])
         except subp.ProcessExecutionError:
@@ -680,7 +679,20 @@ def generate_sources_list(cfg, release, mirrors, cloud):
             )
             aptsrc_file = apt_sources_list
     disabled = disable_suites(cfg.get("disable_suites"), rendered, release)
+    disable_apt_sources_list = False
+    if aptsrc_file == apt_sources_deb822 and os.path.exists(apt_sources_list):
+        disable_apt_sources_list = True
     util.write_file(aptsrc_file, disabled, mode=0o644)
+    if disable_apt_sources_list:
+        LOG.warning(
+            "Disabling %s to favor deb822 source format", apt_sources_list
+        )
+        content = pathlib.Path(apt_sources_list).read_bytes()
+        content = b"# disabled by cloud-init\n" + content
+        util.rename(apt_sources_list, f"{apt_sources_list}.disabled")
+        util.write_file(
+            f"{apt_sources_list}.disabled", content, preserve_mode=True
+        )
 
 
 def add_apt_key_raw(key, file_name, hardened=False):
--- a/tests/integration_tests/modules/test_apt_functionality.py
+++ b/tests/integration_tests/modules/test_apt_functionality.py
@@ -336,6 +336,21 @@ class TestDefaults:
             assert 3 == sources_list.count(sec_deb_line)
             assert 3 == sources_list.count(sec_src_deb_line)
 
+    def test_no_duplicate_apt_sources(self, class_client: IntegrationInstance):
+        r = class_client.execute("apt-get update", use_sudo=True)
+        assert not re.match(
+            r"^W: Target Packages .+ is configured multiple times in", r.stderr
+        )
+
+    def test_disabled_apt_sources(self, class_client: IntegrationInstance):
+        feature_deb822 = is_true(
+            get_feature_flag_value(class_client, "APT_DEB822_SOURCE_LIST_FILE")
+        )
+        if feature_deb822:
+            assert class_client.execute(
+                f"test -f {ORIG_SOURCES_FILE}"
+            ).failed, f"Found unexpected {ORIG_SOURCES_FILE}"
+
 
 DEFAULT_DATA_WITH_URI = _DEFAULT_DATA.format(
     uri='uri: "http://something.random.invalid/ubuntu"'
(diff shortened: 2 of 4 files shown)


blackboxsw:
I'm not certain this is the fix we want. In noble, cloud-init doesn't appear to actually be writing `/etc/apt/sources.list` it seems an artifact of the image build process given  that the timestamp on the file matches /etc/cloud/build.info  not the actual instance launch time/date

aciba90:
> I'm not certain this is the fix we want. In noble, cloud-init doesn't appear to actually be writing `/etc/apt/sources.list` it seems an artifact of the image build process given that the timestamp on the file matches /etc/cloud/build.info not the actual instance launch time/date

That is true and has been like that for previous Ubuntu releases, I believe. Cloud-init did and is overwriting  `/etc/apt/sources.list` if `features.APT_DEB822_SOURCE_LIST_FILE == False`. So the options I see for the case when cloud-init is going to write [1]`ubuntu.sources` but [2]`sources.list` exists are:

1. Write [1] and disable [2]. (This PR).
2. Overwrite [2], instead of [1] and warn about it. (Maintain old behavior).
3. Parse [2], merge with [1] and goto 1. (Complex and error-prone).

cloudinit/config/cc_apt_configure.py
```diff
@@ -680,7 +680,13 @@ def generate_sources_list(cfg, release, mirrors, cloud):
             )
             aptsrc_file = apt_sources_list
     disabled = disable_suites(cfg.get("disable_suites"), rendered, release)
+    disable_apt_sources_list = False
+    if aptsrc_file == apt_sources_deb822 and os.path.exists(apt_sources_list):
+        disable_apt_sources_list = True
     util.write_file(aptsrc_file, disabled, mode=0o644)
+    if disable_apt_sources_list:
+        LOG.debug("Disabling %s", apt_sources_list)
+        util.rename(apt_sources_list, f"{apt_sources_list}.old")
```
blackboxsw, line 689: I think we want to rename this `.disabled` instead of `.old` as a best practice for preserved and ignored apt sources files as we see apt ignores these files by default `
```
$ apt-config dump | grep disable
Dir::Ignore-Files-Silently:: "\.disabled$";
```

I think we may also want to add a header line `# disabled by cloud-init` to this file to leave a breadcrumb for why this rename happened

blackboxsw:
> 
> That is true and has been like that for previous Ubuntu releases, I believe. Cloud-init did and is overwriting `/etc/apt/sources.list` if `features.APT_DEB822_SOURCE_LIST_FILE == False`. So the options I see for the case when cloud-init is going to write [1]`ubuntu.sources` but [2]`sources.list` exists are:
> 
> 1. Write [1] and disable [2]. (This PR).
> 2. Overwrite [2], instead of [1] and warn about it. (Maintain old behavior).
> 3. Parse [2], merge with [1] and goto 1. (Complex and error-prone).

Agreed here, we know that there is work in progress to make sure CPC image builds do not leave this config artifact around on images with cloud-init installed. But, that would only cover the Ubuntu cloudimage case for migration to an APT deb822 compatible system.


Since we know cloud-init needs to support any debian derivateiv, we still need to cover the case where new cloud-init which wants to write deb822 format is running in some debian image which contains an old `/etc/apt/sources.list` unified format.
 In this case, it makes sense to disable the existing `/etc/apt/sources.list`, WARN on this behavior taken by cloud-init and allow cloud-init to emit the new deb822 sources format to `/etc/cloud/cloud.cfg.d` as you have listed.

As you mentioned, we don't want to perform `3.` to parse old unified sources.list due to complexity, but also due to the fact that cloud-init has always blindly overwritten that file anyway, so no initial /etc/apt/sources.list file survives first boot with cloud-init anway, so we don't need to support that case.

We also should rule out `2.` for that same reason, cloud-init should own base APT configuration on a system (which typically involved overwriting /etc/apt/sources.list anyway) so no need to preseve this file on new deb822-compatible installs.

blackboxsw (changes requested):
@aciba90 thank you for clarifying the use-cases we want  to cover here. I think a minor adjustment to the file suffix we use when renaming, and a breadcrumb comment in the file to notify that cloud-init migrated it would be helpful. 
Thanks for covering the more generic cases outside of Ubuntu cloud image builds.

blackboxsw:
General idea looks good, i think we need unittest coverage for this case and a minor addition to the existing integration test tests/integration_tests/modules/test_apt_functionality.py to assert that we don't see warnings about duplicate apt sources being found on any series as well as that we see the `.disabled` file being created.

aciba90:
Thanks, @blackboxsw, for the comments and suggestions! I have addressed them and this PR is ready to be re-reviewed.

tests/integration_tests/modules/test_apt_functionality.py
```diff
@@ -336,6 +336,22 @@ def test_security(self, class_client: IntegrationInstance):
...
+
+    def test_disabled_apt_sources(self, class_client: IntegrationInstance):
+        feature_deb822 = is_true(
+            get_feature_flag_value(class_client, "APT_DEB822_SOURCE_LIST_FILE")
+        )
+        if feature_deb822:
+            assert (
+                class_client.execute(f"ls {ORIG_SOURCES_FILE}").return_code
+                != 0
+            )
```
blackboxsw, line 353: Just a minor improvement to error message traceback.
```suggestion
            assert class_client.execute(
                f"test -f {ORIG_SOURCES_FILE}"
            ).failed, f"Found unexpected {ORIG_SOURCES_FILE}"
```

blackboxsw (approved):
Thanks for the integration test and unittest coverage. +1 with a minor nit in integration test.
I'll apply it and land it. it'd be good to get an upload from main into noble to help out customers affected by this warning.

PR #4702, opened 2023-12-14
test: Fix silent swallowing of unexpected subp error

TheRealFalcon:
## Proposed Commit Message

```
test: Fix silent swallowing of unexpected subp error

We throw an error when supb isn't mocked in a test. However, there
are many broad exception handlers that wind up swallowing the error.
This commit changes it from an AssertionError to a new error that
inherits from BaseException so that it doesn't get silently swallowed.

Also fix any tests that are impacted.
```

## Merge type

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -79,6 +79,14 @@ class _FixtureUtils:
         return result[0]
 
 
+class UnexpectedSubpError(BaseException):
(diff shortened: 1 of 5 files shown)


TheRealFalcon:
cc @cjp256 because a number of Azure tests are affected

aciba90 (approved):
LGTM!

PR #4714, opened 2023-12-21
azure/imds: log http failures as warnings instead of info

cjp256:
This is a more appropriate categorization for these logs and
this log level will ensure they are logged to the console.


anhvoms (approved):
We should consider logging the time/attempts remaining for retrying. Perhaps not in this PR as that will be applicable for both wireserver/imds communication

TheRealFalcon (approved):
LGTM!

PR #4715, opened 2023-12-21
azure: remove cloud-init.log reporting via KVP

cjp256:
With the recent changes in failure reporting and diagnostics, the full log is unnecessary overhead.

Rename push_log_to_kvp() => report_dmesg_to_kvp() now that it only captures the dmesg log.


anhvoms (approved):

holmanb:
@cjp256 Looks like tests are failing.

cjp256:
> @cjp256 Looks like tests are failing.

Thanks! Should be fixed now.

tests/unittests/reporting/test_reporting_hyperv.py
```diff
@@ -234,74 +234,6 @@ def test_report_compressed_event(self):
...
-                2,
-                [
-                    log_content[-azure.MAX_LOG_TO_KVP_LENGTH :].encode(),
-                    extra_content.encode(),
-                ],
-            )
-        finally:
-            instantiated_handler_registry.unregister_item(
-                "telemetry", force=False
-            )
```
blackboxsw, line 303: It may be worth keeping unittest coverage of the dmesg functionality that is still retained.

blackboxsw (changes requested):
I think we need test coverage of the remaining report_dmesg_to_kvp

tests/unittests/reporting/test_reporting_hyperv.py
cjp256: Good point, done!  Moved the call to report_dmesg_to_kvp() up to the Azure data source to decouple it from the walinuxagent shim we'll eventually get rid of.  report_compressed_event() is tested in the hyperv reporting handler.  I didn't want to do it now, but in a future PR will consolidate these kvp bits into the kvp module.

blackboxsw (approved):
LGTM! Thanks @cjp256

Issue #4059, opened 2023-05-12
Add --timeout to cloud-init status --wait


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1999876](https://bugs.launchpad.net/cloud-init/+bug/1999876)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2022-12-16T09:10:36.916235+00:00
date_fix_committed = None
date_fix_released = None
id = 1999876
importance = wishlist
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1999876
milestone = None
owner = aciba
owner_name = Alberto Contreras
private = False
status = triaged
submitter = aciba
submitter_name = Alberto Contreras
tags = ['bitesize']
duplicates = []
</pre>
</details>

_Launchpad user **Alberto Contreras(aciba)** wrote on 2022-12-16T09:10:36.916235+00:00_

Add a supplemental / optional timeout flat to the `cloud-init status --wait` command to not infinitely block.

This could be useful for testing context, as an example.

Context: https://irclogs.ubuntu.com/2022/12/15/%23cloud-init.html

ubuntu-server-builder:
_Launchpad user **Brett Holman(holmanb)** wrote on 2023-01-03T21:09:04.759132+00:00_

Quoting from the irc logs:

> blocking forever is usually a nono, that's my thought

Agreed. However, this is out of our control. We are awaiting on other systemd units to return before exiting.

Maybe we make this wait forever behavior a little more obvious in the help/manpage output?

> this is in a script that runs tests, so specially here I don't want forever blocks

I guess if you don't care what state your machine is in before doing things on the machine, then a timeout might be useful?

If we add this, lets make sure to document well that using `--timeout` means that system configuration will not be completed at exit.

holmanb:
I'm going to close this as not planned. While writing code such as this would be trivial, this problem is also trivially solved by anybody that wants this feature. Coreutils/uutils `timeout` works just fine - or run it in the background with a for-loop. This feature doesn't really add much that a user doesn't already have.

PR #4717, opened 2023-12-23
timeout parameter for cloud-init --wait  #4059

FoRtY-5:
## Proposed Commit Message

```
feature: timeout for cloud-init status --wait command

Added `--timeout` flag for `cloud-init status --wait` command to terminate after 
declared amount of seconds.

Fixes GH-#4059
LP: #1999876
```

## Additional Context

Example usage `cloud-init status --wait --timeout 5`
Time to terminate provided in timeout flag might not be 1:1 with real time in seconds because of the time library behavior (I suppose), the implemented solution meant to be clear and readable and also fulfills requirements assumed based on the context from https://irclogs.ubuntu.com/2022/12/15/%23cloud-init.html
[#4059](https://github.com/canonical/cloud-init/issues/4059)
## Test Steps
```
cd cloud-init/cmd
python main.py status --wait --timeout 10
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -12,7 +12,7 @@ import json
 import os
 import sys
 from copy import deepcopy
-from time import gmtime, sleep, strftime
+from time import gmtime, sleep, strftime, time
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
 
 from cloudinit import safeyaml, subp
@@ -130,6 +130,13 @@ def get_parser(parser=None):
         default=False,
         help="Block waiting on cloud-init to complete",
     )
+    parser.add_argument(
+        "-t",
+        "--timeout",
+        type=int,
+        nargs=1,
+        help="Timeout in seconds for --wait flag - use to avoid forever wait",
+    )
     return parser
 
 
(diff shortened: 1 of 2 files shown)


holmanb:
I'm still not convinced that this is a real value add. The user can already do this themselves with `cloud-init status` in a loop.

If a timeout happens, the user is going to have to do error handling based on the return code anyways (otherwise their code will be making invalid assumptions). Is that really so hard to do for the typical user of this flag?

I'm not going to say we can't do this, but I am hesitant to add this, since the real users of this are automated scripts, which I would assume have the ability to loop since they also need the ability to do conditional execution.

Thoughts?

FoRtY-5:
I think that based on the [conversation](https://irclogs.ubuntu.com/2022/12/15/%23cloud-init.html) it would be handy for testers to use the `cloud-init status --wait` with the `--timeout` flag - without the `--timeout` flag the `--wait` flags blocks the execution of the test until status changes. 
In my opinion it maybe makes sense for testers to have a simple one line command `cloud-init status --wait --timeout SECONDS`, so it makes their job a little bit easier(?)

With that being said I would consider adding this small enhancement to the main branch - personally I don't see a lot of value added to the final product, maybe life of some testers would be a little bit easier after this change.
Consider closing this issue after deciding what to do with this PR. [Issue](https://github.com/canonical/cloud-init/issues/4059)

holmanb:
Thanks for contributing, @FoRtY-5. Please [add yourself to the cla](https://cloudinit.readthedocs.io/en/latest/development/index.html#prerequisites). Also, this adds code but no tests. Please add test coverage in `tests/integration_tests/cmd/test_status.py` and `tests/unittests/cmd/test_status.py`.

> In my opinion it maybe makes sense for testers to have a simple one line command cloud-init status --wait --timeout SECONDS, so it makes their job a little bit easier(?)

Allowing them to add a flag rather than write their own timeout look on `cloud-init status` makes their job easier. It just also adds code that we have to maintain. And if we're going to give users a new feature lets at least document how to use it correctly.

The point that I'm trying to make is that every caller of this flag will absolutely need to know whether this command exited because of timeout or because cloud-init has completed. This call only adds value when the user is able to discern whether cloud-init has timed out or not, and currently in your implementation, there is no way to signal this to the caller without parsing stdout. If status is called with `--timeout 5` and the call took 5 seconds, did the call timeout or did cloud-init actually complete at 5 seconds (which is entirely possible)? Lets give the user of this new feature an example of how to use it correctly.

holmanb (changes requested):

PR #4721, opened 2023-12-30
logrotate config

### LINKED ISSUE #4509: [enhancement]: add logrotate config file
# Enhancement

Please consider adding a logrotate config file to your cloud-init source repository. Looking at bigger
files in /var/log/, cloud-init grows quite big logging files.

Please also see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=926424 which ends
also into a merge request at https://salsa.debian.org/cloud-team/cloud-init/-/merge_requests/6.

best regards,

Florian La Roche

chifac08:
## logrotate config

```
This feature enables log rotation (logrotate.d) for cloud-init logs located in /var/log.

Fixes  GH-#4509 
```

## Additional Context
The creation of a new logfile after rotation has been done is not necessary due to cloud-init takes care of creating the log file during execution. Furthermore, rotation can run periodically as configured by system administrator and will only be rotated if file size is bigger than one megabyte.

## Checklist

## Merge type


dermotbradley:
I'm currently using a similar logrotate config in the cloud-init package that I maintain for Alpine: https://git.alpinelinux.org/aports/tree/community/cloud-init/cloud-init.logrotate

I don't currently have "nocreate" there but will be adding it.

cloudinit/cmd/devel/logs.py
```diff
@@ -242,7 +250,8 @@ def collect_logs(tarfile, include_userdata: bool, verbosity=0):
             verbosity=verbosity,
         )
 
-        for log in CLOUDINIT_LOGS:
+        init.read_cfg()
+        for log in get_config_logfiles(init.cfg):
```
aciba90, line 254: I do not think `get_config_logfiles` is aware of the rotated log files, so this functionality needs to be added there.

tests/unittests/cmd/devel/test_logs.py
```diff
@@ -35,6 +35,7 @@ def test_collect_logs_with_userdata_requires_root_user(
             " Try sudo cloud-init collect-logs\n" == m_stderr.getvalue()
         )
 
+    @pytest.mark.allow_all_subp
```
aciba90, line 38: Allowing subp to run make the unit tests dependent on the underlying system. Instead, I think we should mock the call to `Init`.

aciba90 (changes requested):
Thanks, @chifac08, for the changes!

Apart from the inline comments, there are some things missing:

1. Unit tests verifying that collect-logs grabs rotated logs.
2. cloud-init clean --logs should also be aware of the rotated logs.

2 could probably be split into another issue if you do not have resources to complete it.

cloudinit/cmd/devel/logs.py
chifac08: Added logic to `get_config_logfiles` function to take care of rotated logfiles for cloud-init as well as cloud-init-output log. Rotated logfiles will only be added to list when they exist. This step is necessary to be compliant with diverse unit tests which test or mock this function. Let me know what you think :)

tests/unittests/cmd/devel/test_logs.py
chifac08: I am not quite sure what's the easiest and best way to implement a mock for `Init` but i took a look at unittests implementation for "clean.py" and this looks very promising to me:
`@pytest.fixture(scope="function")
def init_class(clean_paths):
    class FakeInit:
        cfg = {
            "def_log_file": clean_paths.log,
            "output": {"all": f"|tee -a {clean_paths.output_log}"},
        }
        # Ensure cloud_dir has a trailing slash, to match real behaviour
        paths = MyPaths(cloud_dir=f"{clean_paths.cloud_dir}/")

        def __init__(self, ds_deps):
            pass

        def read_cfg(self):
            pass

    return FakeInit`
Am i far off track or is this something we could work with? thank you for your help

aciba90: Here is a patch mocking `Init` and `get_log_files` to test the isolated behavior of `collect-logs`.  The patch also includes a unit test covering the new behavior of `get_log_files`.

```diff
diff --git a/tests/unittests/cmd/devel/test_logs.py b/tests/unittests/cmd/devel/test_logs.py
index b7ea33052..004b4957f 100644
--- a/tests/unittests/cmd/devel/test_logs.py
+++ b/tests/unittests/cmd/devel/test_logs.py
@@ -35,7 +35,6 @@ class TestCollectLogs:
             " Try sudo cloud-init collect-logs\n" == m_stderr.getvalue()
         )
 
-    @pytest.mark.allow_all_subp
     def test_collect_logs_creates_tarfile(self, m_getuid, mocker, tmpdir):
         """collect-logs creates a tarfile with all related cloud-init info."""
         m_getuid.return_value = 100
@@ -57,6 +56,12 @@ class TestCollectLogs:
         )
         output_tarfile = str(tmpdir.join("logs.tgz"))
 
+        mocker.patch(M_PATH + "Init", autospec=True)
+        mocker.patch(
+            M_PATH + "get_config_logfiles",
+            return_value=[log1, log1_rotated, log2, log2_rotated],
+        )
+
         date = datetime.utcnow().date().strftime("%Y-%m-%d")
         date_logdir = "cloud-init-logs-{0}".format(date)
 
@@ -147,7 +152,6 @@ class TestCollectLogs:
         )
         fake_stderr.write.assert_any_call("Wrote %s\n" % output_tarfile)
 
-    @pytest.mark.allow_all_subp
     def test_collect_logs_includes_optional_userdata(
         self, m_getuid, mocker, tmpdir
     ):
@@ -167,6 +171,12 @@ class TestCollectLogs:
         )
         output_tarfile = str(tmpdir.join("logs.tgz"))
 
+        mocker.patch(M_PATH + "Init", autospec=True)
+        mocker.patch(
+            M_PATH + "get_config_logfiles",
+            return_value=[log1, log2],
+        )
+
         date = datetime.utcnow().date().strftime("%Y-%m-%d")
         date_logdir = "cloud-init-logs-{0}".format(date)
 
diff --git a/tests/unittests/test_util.py b/tests/unittests/test_util.py
index 7186780ba..9e7d06378 100644
--- a/tests/unittests/test_util.py
+++ b/tests/unittests/test_util.py
@@ -2344,6 +2344,30 @@ class TestGetConfigLogfiles(helpers.CiTestCase):
             ),
         )
 
+    def test_output_logs_parsed_when_teeing_files_and_rotated(self):
+        """When output configuration is parsed when teeing files and rotated
+        log files are present."""
+        tmpd = self.tmp_dir()
+        log1 = self.tmp_path("my.log", tmpd)
+        log1_rotated = self.tmp_path("my.log.1.gz", tmpd)
+        log2 = self.tmp_path("himom.log", tmpd)
+        log2_rotated = self.tmp_path("himom.log.1.gz", tmpd)
+
+        util.write_file(log1_rotated, "hello")
+        util.write_file(log2_rotated, "hello")
+
+        self.assertEqual(
+            [log2, log2_rotated, log1, log1_rotated],
+            sorted(
+                util.get_config_logfiles(
+                    {
+                        "def_log_file": str(log1),
+                        "output": {"all": f"|tee -a {log2}"},
+                    }
+                )
+            ),
+        )
+
 
 class TestMultiLog(helpers.FilesystemMockingTestCase):
     def _createConsole(self, root):
```

aciba90:
> 2\. cloud-init clean --logs should also be aware of the rotated logs.

Per your current implementation, this is already done, as `clean` uses `get_config_logfiles`.

tests/unittests/cmd/devel/test_logs.py
chifac08: thank you for your help :) . i applied your patch. Furthermore, i enhanced integration tests.

chifac08:
Added "`logrotate`"  package as dependency for various platforms. I am also thinking about adding some sort of documentation about logrotate file but i am uncertain where to place it. Maybe "reference" would be an suitable option. What do you think?

cloudinit/util.py
```diff
@@ -78,6 +78,10 @@
 TRUE_STRINGS = ("true", "1", "on", "yes")
 FALSE_STRINGS = ("off", "0", "no", "false")
 
+# logrotate
```
aciba90, line 81: Being log-rotation implemented as package configuration, I think these hard-coded values are not enough. For example, [alpine configures the number of rotated logs as 6](https://git.alpinelinux.org/aports/tree/community/cloud-init/cloud-init.logrotate). Note that the compression algorithm / extension can also be configured.

To be safer here, I think we should glob and capture any file matching `f"{_log}*"`.

tests/integration_tests/cmd/test_clean.py
aciba90, line 1: `logrotate` is not a required runtime dependency of cloud-init, so instead of requiring it under `packages/pkg-deps.json`, I think we should install it for this particular integration test or simulate manually the rotation.

To install it, add to USER_DATA:

```yaml
packages:
- logrotate
```

aciba90 (changes requested):
Thanks for fixing the unit tests and adding the supplemental integration test.

> I am also thinking about adding some sort of documentation about logrotate file but i am uncertain where to place it. Maybe "reference" would be an suitable option. What do you think?

I think this is going to be tricky, because:

1. The `packages` contain reference packaging implementations, not real ones.
2. Each downstream can choose to add or not log-rotation and which specific config to use.
3. The `docs` in this repo refer to upstream cloud-init.

So I am not sure if it would be a good idea to document this config.

Just to make 1. clear, this PR is adding the functionality regarding log rotation to `cloud-init clean/collect-logs` and a reference config file to `packages`. A subsequent PR to [ubuntu/devel/debian](https://github.com/canonical/cloud-init/tree/ubuntu/devel/debian) enable this in noble.

cloudinit/util.py
chifac08: Agreed and implemented

tests/integration_tests/cmd/test_clean.py
chifac08: I enhanced the integration test by installing `logrotate` package

chifac08:
I was unaware about `packages` folder is just a reference, but know i understand :) thx for explaining this to me. It truly makes no sense to add documentation about `logrotate` to cloud-init upstream but i am asking myself about how you inform package maintainers of other distros about such changes?

cloudinit/util.py
```diff
@@ -1765,7 +1766,16 @@ def get_config_logfiles(cfg):
             logs.append(target)
         elif ["tee", "-a"] == parts[:2]:
             logs.append(parts[2])
-    return list(set(logs))
+
+    # add rotated log files
+    for _log in logs:
```
aciba90, line 1771: Nit: free variables prepended with an underscore mean unused variables in this codebase, a la Rust. `_log` and `_rotated_logfile` do not collide with any Python keyword nor anything else, if that would be the case we could rename them or append an underscore. I wonder if it would be more readable to just rename them as `log` and `rotated_logfile`.

aciba90 (approved):
LGTM, thanks. I will leave it open in case you want to fix the inline nit comment.

cloudinit/util.py
chifac08: renamed vars to be compliant with codebase

Issue #4097, opened 2023-05-12
network-manager renderer not properly configuring /etc/resolv.conf


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #2015494](https://bugs.launchpad.net/cloud-init/+bug/2015494)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2023-04-06T15:54:39.018777+00:00
date_fix_committed = None
date_fix_released = None
id = 2015494
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/2015494
milestone = None
owner = rosario-esposito
owner_name = olaszfiu
private = False
status = triaged
submitter = rosario-esposito
submitter_name = olaszfiu
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **olaszfiu(rosario-esposito)** wrote on 2023-04-06T15:54:39.018777+00:00_

I'm running cloud-init 23.1.1 on a rockylinux 9 virtual machine (running on Proxmox VE) with network-manager renderer.

The cloud-init network-config file is:

version: 1
config:
    - type: physical
      name: eth0
      mac_address: '62:04:02:7b:34:a0'
      subnets:
      - type: static
        address: '172.18.7.1'
        netmask: '255.255.254.0'
        gateway: '172.18.7.254'
    - type: nameserver
      address:
      - '192.84.134.50'
      - '192.84.134.55'
      search:
      - 'na.infn.it'

Inside the VM base image I have, in /etc/cloud/cloud.cfg:

system_info:
   network:
      renderers: ['network-manager']
      activators: ['network-manager']

When I start the VM, everything works fine and the renderer creates the correct NIC config file: /etc/NetworkManager/system-connections/cloud-init-eth0.nmconnection:

# Generated by cloud-init. Changes will be lost.

[connection]
id=cloud-init eth0
uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
type=ethernet

[user]
org.freedesktop.NetworkManager.origin=cloud-init

[ethernet]
mac-address=62:04:02:7B:34:A0

[ipv4]
method=manual
may-fail=false
address1=172.18.7.1/23
gateway=172.18.7.254

The problem is that /etc/resolv.conf is not correctly configured with my DNS servers:

# cat /etc/resolv.conf

# Generated by NetworkManager
search na.infn.it

[...]

PR #4723, opened 2024-01-02
Apply global DNS settings to individual interfaces in network-manager renderer

apollo13:
This basically applies the logic of d29eeccd to network-manager as well.

## Proposed Commit Message

```
net/nm: Apply global DNS settings to interfaces

Sometimes DNS settings in cloud configs are specified globally and
not per interface / subnet. This results in a configuration without
proper nameservers. This was fixed for netplan in d29eeccd and is
now also applied to the network-manager renderer.
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -307,8 +307,8 @@ class NMConnection:
 
         device_mtu = iface["mtu"]
         ipv4_mtu = None
-        found_nameservers = set()
-        found_dns_search = set()
+        found_nameservers = []
+        found_dns_search = []
 
         # Deal with Layer 3 configuration
         for subnet in iface["subnets"]:
@@ -323,28 +323,37 @@ class NMConnection:
                 self._add_route(route)
             # Add subnet-level DNS
             if "dns_nameservers" in subnet:
-                found_nameservers.update(subnet["dns_nameservers"])
+                found_nameservers.extend(subnet["dns_nameservers"])
             if "dns_search" in subnet:
-                found_dns_search.update(subnet["dns_search"])
+                found_dns_search.extend(subnet["dns_search"])
             if family == "ipv4" and "mtu" in subnet:
                 ipv4_mtu = subnet["mtu"]
 
         # Add interface-level DNS
         if "dns" in iface:
-            found_nameservers.update(iface["dns"]["nameservers"])
-            found_dns_search.update(iface["dns"]["search"])
+            found_nameservers += [
+                dns
+                for dns in iface["dns"]["nameservers"]
+                if dns not in found_nameservers
+            ]
+            found_dns_search += [
+                search
+                for search in iface["dns"]["search"]
+                if search not in found_dns_search
+            ]
 
-        # Add global DNS
-        if network_state.dns_nameservers:
-            found_nameservers.update(network_state.dns_nameservers)
-        if network_state.dns_searchdomains:
-            found_dns_search.update(network_state.dns_searchdomains)
+        # We prefer any interface-specific DNS entries, but if we do not
+        # have any, add the global DNS to the connection
+        if not found_nameservers and network_state.dns_nameservers:
+            found_nameservers = network_state.dns_nameservers
+        if not found_dns_search and network_state.dns_searchdomains:
+            found_dns_search = network_state.dns_searchdomains
 
         # Write out all DNS entries to the connection
-        for nameserver in sorted(list(found_nameservers)):
+        for nameserver in found_nameservers:
             self._add_nameserver(nameserver)
         if found_dns_search:
-            self._add_dns_search(sorted(list(found_dns_search)))
+            self._add_dns_search(found_dns_search)
 
         # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
         # at the at the same time. This will make the network configuration
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -246,7 +246,7 @@ class NMConnection:
         """
         return addr.replace("-", ":").upper()
 
-    def render_interface(self, iface, renderer):
+    def render_interface(self, iface, network_state, renderer):
         """
         Integrate information from network state interface information
         into the connection. Most of the work is done here.
@@ -307,6 +307,7 @@ class NMConnection:
 
         device_mtu = iface["mtu"]
         ipv4_mtu = None
+        has_address = False
         found_nameservers = []
         found_dns_search = []
 
@@ -317,6 +318,7 @@ class NMConnection:
             self._set_ip_method(family, subnet["type"])
             if "address" in subnet:
                 self._add_address(family, subnet)
+                has_address = True
             if "gateway" in subnet:
                 self.config[family]["gateway"] = subnet["gateway"]
             for route in subnet["routes"]:
@@ -333,6 +335,14 @@ class NMConnection:
         if "dns" in iface:
             found_nameservers = iface["dns"]["nameservers"]
             found_dns_search = iface["dns"]["search"]
+
+        # Now add in the global DNS information if we have an address but no
+        # DNS information yet
+        if has_address:
+            if not found_nameservers and network_state.dns_nameservers:
+                found_nameservers = network_state.dns_nameservers
+            if not found_dns_search and network_state.dns_searchdomains:
+                found_dns_search = network_state.dns_searchdomains
 
         for nameserver in found_nameservers:
             self._add_nameserver(nameserver)
@@ -453,7 +463,7 @@ class Renderer(renderer.Renderer):
         # Now render the actual interface configuration
         for iface in network_state.iter_interfaces():
             conn = self.connections[iface["name"]]
-            conn.render_interface(iface, self)
+            conn.render_interface(iface, network_state, self)
 
         # And finally write the files
         for con_id, conn in self.connections.items():
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -307,9 +307,8 @@ class NMConnection:
 
         device_mtu = iface["mtu"]
         ipv4_mtu = None
-        has_address = False
-        found_nameservers = []
-        found_dns_search = []
+        found_nameservers = set()
+        found_dns_search = set()
 
         # Deal with Layer 3 configuration
         for subnet in iface["subnets"]:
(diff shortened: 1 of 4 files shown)


TheRealFalcon:
@apollo13 , thanks for the PR and improving cloud-init, but it looks like this same functionality was recently addressed in https://github.com/canonical/cloud-init/pull/4637 . Does that change adequately fix what you're trying to address here? Is there anything else needed by this PR?

apollo13:
I feel a bit stupid now 😔 I will test your changes tomorrow (will need to backport them to 23.1 and rerun against proxmox with it). Will let you know how it works out

apollo13:
@TheRealFalcon I took your changes (21c1d91617454f5e43b20546dfc227efa56a8759) through a quick `devel net-convert` run and supplied the following configuration:
```
version: 1
config:
    - type: physical
      name: eth0
      mac_address: '86:bb:06:ac:d6:b1'
      subnets:
      - type: static
        address: '10.7.203.211'
        netmask: '255.255.224.0'
        gateway: '10.7.192.1'
    - type: nameserver
      address:
      - '10.7.192.2'
      - '10.7.192.3'
      search:
      - 'domain.ext'
```

the rendered network-manager config does not include DNS and looks like this:

```
# Generated by cloud-init. Changes will be lost.

[connection]
id=cloud-init eth0
uuid=1dd9a779-d327-56e1-8454-c65e2556c12c
autoconnect-priority=120
type=ethernet

[user]
org.freedesktop.NetworkManager.origin=cloud-init

[ethernet]
mac-address=86:BB:06:AC:D6:B1

[ipv4]
method=manual
may-fail=false
address1=10.7.203.211/19
gateway=10.7.192.1
```

apollo13:
Running this with my branch "properly" includes:
```
dns=10.7.192.2;10.7.192.3;
dns-search=domain.ext;
```

TheRealFalcon:
@apollo13 , thanks for the feedback. I'll see your piece needs to be applied.

apollo13:
Btw I am having a hard time understanding your changes, eg where is this 'dns' key supposed to come from:
https://github.com/canonical/cloud-init/blob/c2c100e8c9fd8709539b3ab2b0ee34c66ba3f2f7/cloudinit/net/network_manager.py#L314

Do you want to try to address my issue or should I try to rebase my changes onto yours tomorrow -- just don't want us to do double work and I am having a little trouble parsing your last comment as non-native speaker.

TheRealFalcon:
My fault, I mis-typed a word, so my message didn't make much sense.

I'll take a look at your issue and see how it fits into my changes. I'll try to get you a response by the end of the day.

> where is this 'dns' key supposed to come from

I'm primarily thinking about v2-format whereas your example uses v1, but after cloud-init converts the config to the intermediary network-state config, the DNS can be defined both at a top-level config as well as within a subnet config. If you pass `--debug` to a `cloud-init devel net-convert` call, there's a `_network_state["dns"]` but you can also have something like `_network_state["interfaces"][_name]["subnets"]["routes"][0]["dns_nameservers"]`

apollo13:
Ah `"dns"` in iface is coming from here: https://github.com/canonical/cloud-init/blob/21c1d91617454f5e43b20546dfc227efa56a8759/cloudinit/net/network_state.py#L338 -- I forgot that the network_state doesn't match the config exactly.

Btw merging something along the lines of this PR will probably fix #4097 as well.

TheRealFalcon:
After looking closer, yes, our two changes are different. There are actually three different places in the network state that DNS can be found. I fixed/added support for two, but I missed the one you're looking at.

This shows a good example (from the output of `cloud-init devel net-convert ... --debug`:
```
_network_state:
  config:
    ethernets:
      eth99:
        addresses:
        - 192.168.21.3/24
        dhcp4: true
        match:
          macaddress: c0:d6:9f:2c:e8:80
        nameservers:
          addresses:
          - 8.8.8.8
          - 8.8.4.4
          search:
          - barley.maas
          - sach.maas
        routes:
        - metric: 10000
          to: 0.0.0.0/0
          via: 65.61.151.37
        set-name: eth99
    version: 2
  dns:
    nameservers:
    - 8.8.8.8
    - 8.8.4.4
    search:
    - barley.maas
    - sach.maas
  interfaces:
    eth99:
      accept-ra: null
      address: null
      dns:
        nameservers:
        - 8.8.8.8
        - 8.8.4.4
        search:
        - barley.maas
        - sach.maas
      gateway: null
      inet: inet
      mac_address: c0:d6:9f:2c:e8:80
      mode: manual
      mtu: null
      name: eth99
      subnets:
      - routes:
        - gateway: 65.61.151.37
          metric: 10000
          netmask: 0.0.0.0
          network: 0.0.0.0
          prefix: 0
        type: dhcp4
      - address: 192.168.21.3
        dns_nameservers:
        - 8.8.8.8
        - 8.8.4.4
        dns_search:
        - barley.maas
        - sach.maas
        netmask: 255.255.255.0
        prefix: 24
        routes: []
        type: static
      type: physical
      wakeonlan: null
  routes: []
  use_ipv6: false
_version: 2
use_ipv6: false

```

You can see the raw config under the `config` key. You'll notice we have DNS in 3 places:
```
1. top level key
2. interfaces['eth99']['dns']
3. interfaces['eth99']['subnets'][1]['dns_nameservers']  # and dns_search
```

Thats unfortunate, but consolidating those doesn't need to happen here. My changes never addressed the top-level, so yes, your changes are still needed.

I'll provide more detailed review later, but I just wanted to provide this context first.

cloudinit/net/network_manager.py
```diff
@@ -438,7 +461,7 @@ def render_network_state(
         # Now render the actual interface configuration
         for iface in network_state.iter_interfaces():
             conn = self.connections[iface["name"]]
-            conn.render_interface(iface, self)
+            conn.render_interface(iface, self, network_state)
```
TheRealFalcon, line 464: Rather than passing in `network_state` on this line, could we instead loop through the DNS entries and call `conn._add_nameserver()` after this line? We could remove the leading `_` so it's not a private method.

Same for adding the dns search. I think after a rebase that should work without doing the additional address checking.

apollo13: Not sure, it will become ugly either way I fear. Somehow we will need to check if there are static addresses and if yes, if there are no DNS options set already we need to set them. There are probably two options here: 

 * Do this in `render_network_state` which requires basically redoing what you already did (partially) in `render_interface`.
 * Add a method to the connection ala `finalize(network_state)` which then checks the `.config` and if it has addresses but no DNS yet it would apply the top-top-top-level DNS stuff.

No strong feelings and both will probably end up ugly, my gut tells me that a `finalize` option might maybe be nicer (slightly :D)

apollo13:
@TheRealFalcon I have pushed two commits. The first one rewrites your changes somewhat so my changes fit in better (I hope that is okay for you, I honestly think the new code is clearer as well) and the second commit adds my feature.

TheRealFalcon:
@apollo13 , thanks for the updates. A few things:
1. After looking at the different places to specify DNS, it doesn't appear that they should be mutually exclusive. I know my last commit assumed that, but I'm now thinking they should be combined rather than choosing one over the other.
2. As long as the IP family has been defined, it should be safe to add a DNS entry.

What do you think about the following patch?
```
diff --git a/cloudinit/net/network_manager.py b/cloudinit/net/network_manager.py
index 7d7772e11..50f4a2f59 100644
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -307,9 +307,8 @@ class NMConnection:
 
         device_mtu = iface["mtu"]
         ipv4_mtu = None
-        has_address = False
-        found_nameservers = []
-        found_dns_search = []
+        found_nameservers = set()
+        found_dns_search = set()
 
         # Deal with Layer 3 configuration
         for subnet in iface["subnets"]:
@@ -318,36 +317,34 @@ class NMConnection:
             self._set_ip_method(family, subnet["type"])
             if "address" in subnet:
                 self._add_address(family, subnet)
-                has_address = True
             if "gateway" in subnet:
                 self.config[family]["gateway"] = subnet["gateway"]
             for route in subnet["routes"]:
                 self._add_route(route)
+            # Add subnet-level DNS
             if "dns_nameservers" in subnet:
-                found_nameservers.extend(subnet["dns_nameservers"])
+                found_nameservers.update(subnet["dns_nameservers"])
             if "dns_search" in subnet:
-                found_dns_search.extend(subnet["dns_search"])
+                found_dns_search.update(subnet["dns_search"])
             if family == "ipv4" and "mtu" in subnet:
                 ipv4_mtu = subnet["mtu"]
 
-        # Now add our DNS search domains. We add them later because we
-        # only want them if an IP family has already been defined
+        # Add interface-level DNS
         if "dns" in iface:
-            found_nameservers = iface["dns"]["nameservers"]
-            found_dns_search = iface["dns"]["search"]
-
-        # Now add in the global DNS information if we have an address but no
-        # DNS information yet
-        if has_address:
-            if not found_nameservers and network_state.dns_nameservers:
-                found_nameservers = network_state.dns_nameservers
-            if not found_dns_search and network_state.dns_searchdomains:
-                found_dns_search = network_state.dns_searchdomains
-
-        for nameserver in found_nameservers:
+            found_nameservers.update(iface["dns"]["nameservers"])
+            found_dns_search.update(iface["dns"]["search"])
+
+        # Add global DNS
+        if network_state.dns_nameservers:
+            found_nameservers.update(network_state.dns_nameservers)
+        if network_state.dns_searchdomains:
+            found_dns_search.update(network_state.dns_searchdomains)
+
+        # Write out all DNS entries to the connection
+        for nameserver in sorted(list(found_nameservers)):
             self._add_nameserver(nameserver)
         if found_dns_search:
-            self._add_dns_search(found_dns_search)
+            self._add_dns_search(sorted(list(found_dns_search)))
 
         # we do not want to set may-fail to false for both ipv4 and ipv6 dhcp
         # at the at the same time. This will make the network configuration

```

It would require some test updates (I can add those too if you'd like), but I think this ensures all DNS entries are being added.

apollo13:
Fine by me. I am not sure how common such configuration will be in practice, but yeah it would probably make the most sense. If you can update the tests, that would be fine with me. If you don't get to it I'll try myself tomorrow in the office (roughly in 16 hours or so).

TheRealFalcon:
@apollo13 , I took the liberty of pushing the commit along with the tests. Let me know if you have any questions or concerns. There is one xfail I added for a v2 DNS issue I need to follow-up on, but that can come after this PR.

Before we can accept your change, you need to sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html#submitting-your-first-pull-request)

apollo13:
Hups, the CLA entry got lost when rebasing onto your changes. Pushed now… 

Your changes look good. I am still torn whether the most specific DNS entry should win or whether they should get merged, but either will work for my usecase… Something I didn't check though: How are other renderers handling this situation? If nothing else I would opt for consistency with the other renderers.

Since DNS servers are usually queried in parallel (?) nowadays I don't think we will get any ordering problems?

TheRealFalcon:
> Hups, the CLA entry got lost when rebasing onto your changes. Pushed now…

Have you signed the [actual CLA](https://ubuntu.com/legal/contributors/agreement) too? I'm not seeing your information there. If you did, can you provide the date that you signed it?

> I am still torn whether the most specific DNS entry should win or whether they should get merged

Another option is to merge them, but order the most specific first.

> How are other renderers handling this situation?

Very inconsistently. Sysconfig only writes the global entry to resolv.conf and ignores the rest. networkd prefers the interface-specific one if provided but doesn't consider a subnet-specific entry at all. /etc/network/interfaces will use separate entries, putting global dns under a localhost entry. Netplan ignores the dns entries altogether.

Being most consistent with networkd would mean using the more specific DNS entries. That would be the more conservative option as far as functionality change too.

apollo13:
> Have you signed the [actual CLA](https://ubuntu.com/legal/contributors/agreement) too? I'm not seeing your information there. If you did, can you provide the date that you signed it?

Ups, I failed there before. Did so now.

TheRealFalcon:
@apollo13 , any thoughts on the rest of my comment?

apollo13:
@TheRealFalcon I would be fine with using the most specific ones. The at least 2 renderers have the same behavior :D

apollo13:
The new changes look good to me!

tests/unittests/net/test_net_rendering.py
```diff
@@ -88,6 +88,9 @@ def _check_network_manager(network_state: NetworkState, tmp_path: Path):
     "test_name, renderers",
     [("no_matching_mac_v2", Renderer.Netplan | Renderer.NetworkManager)],
 )
+@pytest.mark.xfail(
+    reason="v2 interface-specific DNS errantly gets applied globally"
```
apollo13, line 92: Does this still fail with the new changes?

TheRealFalcon: Yes, because a subnet-specific DNS entry will get added to the global list, so the global entry is getting added to connections that didn't specify any DNS. I'll follow up on this separately.

TheRealFalcon (approved):
@apollo13 , thanks for raising this and working with me to get this complete!

apollo13:
Thank you so much for your help on this. Now I have to figure out how to get that back ported into my distros 😄

PR #4727, opened 2024-01-02
docs: add create_hostname_file to all hostname user-data examples

catmsred:
If a cloud or user sets create_hostname_file to false (as GCP does) the examples in our docs for setting the hostname will not work. This change updates those examples to also set create_hostname_file to true.

## Proposed Commit Message

```
docs: add create_hostname_file to all hostname user-data examples
    
If a cloud or user sets create_hostname_file to false (as GCP does)
the examples in our docs for setting the hostname will not work.
This change updates those examples to also set create_hostname_file
to true.
```

## Additional Context
See https://github.com/canonical/cloud-init/pull/4330 for details on `create_hostname_file` key.

## Test Steps
This is left as a WIP because I want to go through and follow each of the changed examples to ensure they still work with the new key included.

## Checklist

- [ ] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly (N/A)

## Merge type

### DIFF
--- a/cloudinit/config/cc_update_hostname.py
+++ b/cloudinit/config/cc_update_hostname.py
@@ -78,7 +78,7 @@ meta: MetaSchema = {
         dedent(
             """\
         # On a machine without an ``/etc/hostname`` file, don't create it
-        # In most clouds, this will results in a DHCP configured hostname
+        # In most clouds, this will result in a DHCP-configured hostname
         # provided by the cloud
         create_hostname_file: false
         """
--- a/cloudinit/config/cc_update_hostname.py
+++ b/cloudinit/config/cc_update_hostname.py
@@ -65,6 +65,7 @@ meta: MetaSchema = {
         fqdn: external.fqdn.me
         hostname: myhost
         prefer_fqdn_over_hostname: true
+        create_hostname_file: true
         """
         ),
         dedent(
@@ -72,6 +73,14 @@ meta: MetaSchema = {
         # Set hostname to "external" instead of "external.fqdn.me" when
         # cloud metadata provides the ``local-hostname``: "external.fqdn.me".
         prefer_fqdn_over_hostname: false
+        """
+        ),
+        dedent(
+            """\
+        # On a machine without an ``/etc/hostname`` file, don't create it
+        # In most clouds, this will results in a DHCP configured hostname
+        # provided by the cloud
+        create_hostname_file: false
         """
         ),
     ],
--- a/doc/rtd/tutorial/qemu-script.sh
+++ b/doc/rtd/tutorial/qemu-script.sh
@@ -21,8 +21,6 @@ EOF
 
 cat << EOF > meta-data
 instance-id: someid/somehostname
-local-hostname: jammy
-create_hostname_file: true
 EOF
 
 touch vendor-data
--- a/doc/rtd/tutorial/qemu-script.sh
+++ b/doc/rtd/tutorial/qemu-script.sh
@@ -22,6 +22,7 @@ EOF
 cat << EOF > meta-data
 instance-id: someid/somehostname
 local-hostname: jammy
+create_hostname_file: true
 EOF
 
 touch vendor-data
--- a/doc/rtd/tutorial/qemu.rst
+++ b/doc/rtd/tutorial/qemu.rst
@@ -143,8 +143,6 @@ Now let's run the following command, which creates a file named
 
     $ cat << EOF > meta-data
     instance-id: someid/somehostname
-    local-hostname: jammy
-    create_hostname_file: true
 
     EOF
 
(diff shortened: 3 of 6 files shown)


doc/rtd/reference/base_config_reference.rst
```diff
@@ -287,6 +287,9 @@ On an Ubuntu system, :file:`/etc/cloud/cloud.cfg` should look similar to:
     # This will cause the set+update hostname module to not operate (if true)
     preserve_hostname: false
 
+    # This will create the hostname file (e.g. /etc/hostname) if it does not already exist (if true)
+    create_hostname_file: false
```
holmanb, line 291: Why add this?

> On an Ubuntu system, :file:`/etc/cloud/cloud.cfg` should look similar to:

This will only be true on GCE, right?

holmanb (commented):
Thanks @catmsred for this PR. I like the direction of it. I just noticed the WIP tag, but I had already started reviewing so I'll leave what I started below.


Think we want to add this to the [modules examples](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#set-hostname)?

doc/rtd/reference/base_config_reference.rst
catmsred: Good point -- will fix!  And yes, I think it would be good to add to the modules examples for set and update hostname.  Thanks!

doc/rtd/tutorial/qemu-script.sh
```diff
@@ -22,6 +22,7 @@ EOF
 cat << EOF > meta-data
 instance-id: someid/somehostname
 local-hostname: jammy
+create_hostname_file: true
```
TheRealFalcon, line 25: Since this is a tutorial, I think we can remove this line (and probably the line above it) to avoid over-complicating things.

@holmanb , what do you think? Also...where is this file even used/referenced?

doc/rtd/tutorial/qemu.rst
```diff
@@ -144,6 +144,7 @@ Now let's run the following command, which creates a file named
     $ cat << EOF > meta-data
     instance-id: someid/somehostname
     local-hostname: jammy
+    create_hostname_file: true
```
TheRealFalcon, line 147: Same here. This would go into user data, not meta data. Though...I'm not sure it's worth mentioning hostnames at all.

TheRealFalcon (changes requested):

doc/rtd/tutorial/qemu-script.sh
holmanb: > Since this is a tutorial, I think we can remove this line (and probably the line above it) to avoid over-complicating things.

+1 works for me

> Also...where is this file even used/referenced?

This file is referenced from the [Qemu tutorial debugging page](https://cloudinit.readthedocs.io/en/latest/tutorial/qemu-debugging.html).

TheRealFalcon (approved):
LGTM!

@catmsred , I'm assuming this is no longer WIP? If so, it's usually best to remove WIP from the title.

catmsred:
I don't consider it still a WIP but I would like @s-makin to finish her review before it's merged

holmanb (approved):
I know that this touches some docs files, but the changes appear code / config related, so I don't think a review from @s-makin is really required (though it would be welcome). Since there hasn't been any response from @s-makin do you mind if we go ahead with a merge on this one @catmsred?

catmsred:
@holmanb I don't think anyone has validated the NoCloud reference example change starting from the beginning of the example, but if you're okay with merging that I'm fine with it too.

s-makin (commented):
Just a few nits. The tutorial still runs fine, so no problems there, but I did have problems getting tox - doc to run. Some sort of a problem with the pyspelling module?

cloudinit/config/cc_update_hostname.py
```diff
@@ -74,6 +75,14 @@
         prefer_fqdn_over_hostname: false
         """
         ),
+        dedent(
+            """\
+        # On a machine without an ``/etc/hostname`` file, don't create it
+        # In most clouds, this will results in a DHCP configured hostname
+        # provided by the cloud
```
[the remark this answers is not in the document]
catmsred: I didn't make these changes ("this will result in a hostname provided by the cloud" -> "this will result in a hostname being provided by the cloud") because the implication of the second on is, to me, that the key we provide has an effect on the cloud DHCP service, which it does not.  The original statement is grammatically correct in my opinion (provided by the cloud modifies hostname) and is more appropriate to the situation.  Let me know if you feel differently/would like me to reword the comment!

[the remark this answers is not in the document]
s-makin: I'm not sure I agree, but if that's how you interpreted it then chances are other people might also, so happy to leave as is. Thanks for clarifying! Feel free to mark these as resolved.

s-makin (approved):

PR #4730, opened 2024-01-03
feat: datasource check for WSL

CarlosNihelton:
## Proposed Commit Message

```
feat: datasource check for WSL

Adds `dscheck_WSL` function in ds-identify to allow detecting WSL.

Fixes: UDENG-1993
```

## Additional Context

This PR opens up the work for implementing a datasource for WSL.
Subsequent PR's will bring the datasource, its documentation and
the final step should be enabling it in the required datasources lists.

## Checklist

## Merge type

### DIFF
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -30,6 +30,10 @@ UNAME_FREEBSD = (
     "FreeBSD 12.1-RELEASE-p10 FreeBSD 12.1-RELEASE-p10 GENERIC  amd64"
 )
 UNAME_OPENBSD = "OpenBSD GENERIC.MP#1397 amd64"
+UNAME_WSL = (
+    "Linux 5.15.133.1-microsoft-standard-WSL2 #1 SMP Thu Oct 5 21:02:42 "
+    "UTC 2023 x86_64"
+)
 
 BLKID_EFI_ROOT = """
 DEVNAME=/dev/sda1
@@ -243,9 +247,11 @@ MOCK_VIRT_IS_VMWARE = {"name": "detect_virt", "RET": "vmware", "ret": 0}
 # currenty' SmartOS hypervisor "bhyve" is unknown by systemd-detect-virt.
 MOCK_VIRT_IS_VM_OTHER = {"name": "detect_virt", "RET": "vm-other", "ret": 0}
 MOCK_VIRT_IS_XEN = {"name": "detect_virt", "RET": "xen", "ret": 0}
+MOCK_VIRT_IS_WSL = {"name": "detect_virt", "RET": "wsl", "ret": 0}
 MOCK_UNAME_IS_PPC64 = {"name": "uname", "out": UNAME_PPC64EL, "ret": 0}
 MOCK_UNAME_IS_FREEBSD = {"name": "uname", "out": UNAME_FREEBSD, "ret": 0}
 MOCK_UNAME_IS_OPENBSD = {"name": "uname", "out": UNAME_OPENBSD, "ret": 0}
+MOCK_UNAME_IS_WSL = {"name": "uname", "out": UNAME_WSL, "ret": 0}
 
 shell_true = 0
 shell_false = 1
@@ -1141,6 +1147,20 @@ class TestOracle(DsIdentifyBase):
         mycfg = copy.deepcopy(VALID_CFG["Oracle"])
         mycfg["files"][P_CHASSIS_ASSET_TAG] = "Not Oracle"
         self._check_via_dict(mycfg, rc=RC_NOT_FOUND)
+
+
+class TestWSL(DsIdentifyBase):
+    def test_found(self):
+        """Simple positive test of WSL."""
+        self._test_ds_found("WSL-supported")
+
+    def test_not_found(self):
+        """Simple negative test for WSL due other virt."""
+        self._test_ds_not_found("Not-WSL")
+
+    def test_almost_found(self):
+        """Simple negative test by lack of host filesystem mount points."""
+        self._test_ds_not_found("WSL-no-host-mounts")
 
 
 def blkid_out(disks=None):
@@ -2118,6 +2138,44 @@ VALID_CFG = {
         "files": {
             P_PRODUCT_NAME: "Not 3DS Outscale VM\n",
             P_SYS_VENDOR: "3DS Outscale\n",
+        },
+    },
+    "Not-WSL": {
+        "ds": "WSL",
+        "mocks": [
+            MOCK_VIRT_IS_KVM,
+        ],
+    },
+    "WSL-no-host-mounts": {
+        "ds": "WSL",
+        "mocks": [
+            MOCK_VIRT_IS_WSL,
+            MOCK_UNAME_IS_WSL,
+        ],
+        "files": {
+            "proc/mounts": (
+                "/dev/sdd / ext4 rw,errors=remount-ro,data=ordered 0 0\n"
+                "cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec0 0\n"
+                "snapfuse /snap/core22/1033 fuse.snapfuse ro,nodev,user_id=0,"
+                "group_id=0,allow_other 0 0"
+            ),
+        },
+    },
+    "WSL-supported": {
+        "ds": "WSL",
+        "mocks": [
+            MOCK_VIRT_IS_WSL,
+            MOCK_UNAME_IS_WSL,
+        ],
+        "files": {
+            "proc/mounts": (
+                "/dev/sdd / ext4 rw,errors=remount-ro,data=ordered 0 0\n"
+                "cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec0 0\n"
+                "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;\n"
+                "gid=0;symlinkroot=/mnt/...\n"
+                "snapfuse /snap/core22/1033 fuse.snapfuse ro,nodev,user_id=0,"
+                "group_id=0,allow_other 0 0"
+            ),
         },
     },
 }
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -2171,7 +2171,7 @@ VALID_CFG = {
             "proc/mounts": (
                 "/dev/sdd / ext4 rw,errors=remount-ro,data=ordered 0 0\n"
                 "cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec0 0\n"
-                "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;\n"
+                "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;"
                 "gid=0;symlinkroot=/mnt/...\n"
                 "snapfuse /snap/core22/1033 fuse.snapfuse ro,nodev,user_id=0,"
                 "group_id=0,allow_other 0 0"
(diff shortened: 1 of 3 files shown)


tools/ds-identify
```diff
@@ -1565,6 +1565,22 @@ dscheck_VMware() {
...
 
+dscheck_WSL() {
+    if [ ! "${DI_VIRT}" = "wsl" ]; then
+        return "${DS_NOT_FOUND}"
+    fi
+
+    # The datasource needs to find the cloud-config files in the Windows host
+    # filesystem, which is exposed as 9p mount points, one per disk drive (partition).
+    # If none is found, the datasource cannot proceed.
+    if [ "$(grep -c '9p .*drvfs' "${PATH_ROOT}/proc/mounts")" = 0 ]; then
```
holmanb, line 1576: Long options please (`grep --count`), for easier reading.

Also, matching on `'9p .*drvfs'` is still fairly broad - is there anything that we can do to tighten up this match?

tests/unittests/test_ds_identify.py
```diff
@@ -2120,4 +2140,42 @@ def _print_run_output(rc, out, err, cfg, files):
...
+        "ds": "WSL",
+        "mocks": [
+            MOCK_VIRT_IS_WSL,
+            MOCK_UNAME_IS_WSL,
+        ],
+        "files": {
+            "proc/mounts": (
+                "/dev/sdd / ext4 rw,errors=remount-ro,data=ordered 0 0\n"
+                "cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec0 0\n"
+                "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;\n"
```
holmanb, line 2174: Will the line in `/proc/mounts` always match `aname=drvfs`?

holmanb (commented):
Thanks for this contribution @CarlosNihelton! See the changes requested below.

tools/ds-identify
dermotbradley: @holmanb Busybox's grep (as used by Alpine) only handles short options...

dermotbradley:
Um, this PR changes ds-identify to recognise a WSL environment and to then use the WSL datasource...but there is *no* WSL DataSource in cloud-init and this PR does not introduce one...

tools/ds-identify
holmanb: @dermotbradley oh it does?

@CarlosNihelton I retract that part of the request, the other part still stands

tests/unittests/test_ds_identify.py
CarlosNihelton: Indeed it always match. The device name (the first field in `/proc/mounts`) changed in December (it used to be `drvfs` but it's now `LETTER:\\NUMBER`), thus we cannot rely on the device name. Yet across many versions of WSL I could check this part `aname=drvfs` is stable. `path=` is always there too, but I'd be skeptical about how stable that metadata position is to match it via regex in a single match.

I'd say `'^\S* \S* 9p \S*aname=drvfs;.*'` is a strong enough filter. That only matches entry where the third field is '9p' and the fourth contains 'aname=drvfs', where fields are separated by whitespaces.

tools/ds-identify
```diff
@@ -1565,6 +1565,22 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+dscheck_WSL() {
+    if [ ! "${DI_VIRT}" = "wsl" ]; then
+        return "${DS_NOT_FOUND}"
+    fi
+
+    # The datasource needs to find the cloud-config files in the Windows host
+    # filesystem, which is exposed as 9p mount points, one per disk drive (partition).
```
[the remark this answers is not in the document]
CarlosNihelton: The best I could find was a talk about the WSL2 architecture given by Microsoft's principal engineer and the product manager on YouTube:
https://youtu.be/lwhMThePdIo?t=2431&si=JKTHx39TyRgPbzkZ

tools/ds-identify
```diff
@@ -1565,6 +1565,25 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+dscheck_WSL() {
+    if [ ! "${DI_VIRT}" = "wsl" ]; then
```
holmanb, line 1569: WSL only supports Linux, right? Since ds-identify runs on other kernels besides just the linux kernel, lets add an early check to based on information we already know to avoid the grep on non-Linux.

```suggestion
    if [ ! "${DI_VIRT}" = "wsl" -o "$DI_UNAME_KERNEL_NAME" != "Linux" ]; then
```

cc @igalic

tools/ds-identify
```diff
@@ -1565,6 +1565,25 @@ dscheck_VMware() {
...
+        return "${DS_NOT_FOUND}"
+    fi
+
+    # The datasource needs to find the cloud-config files in the Windows host
+    # filesystem, which is exposed as 9p mount points, one per disk drive (partition).
+    # If none is found, the datasource cannot proceed.
+    # See https://youtu.be/lwhMThePdIo?t=2431&si=JKTHx39TyRgPbzkZ and
+    # https://learn.microsoft.com/en-us/windows/wsl/wsl-config#what-is-drvfs
+    # for more information.
+    if [ "$(grep -c '^\S* \S* 9p \S*aname=drvfs;.*' "${PATH_ROOT}/proc/mounts")" = 0 ]; then
```
holmanb, line 1579: I don't think that `\S` is a [Posix regex](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions), which we prefer for portability.

From Gnu grep's `info` page I see:

```info
‘\S’
     Match non-whitespace, it is a synonym for ‘[^[:space:]]’.
```

So I think we want somthing more like:
```suggestion
    if [ "$(grep -c '^[^[:space:]]* [^[:space:]]* 9p [^[:space:]]*aname=drvfs;.*' "${PATH_ROOT}/proc/mo  unts")" = 0 ]; then
```

Which also works in Gnu and busybox (cc: @dermotbradley) grep implementations:
```bash
$ echo "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;" | grep -c '^[^[:space:]]* [^[:space:]]* 9p [^[:space:]]*aname=drvfs;.*'
1
$ echo "C:\\134 /mnt/c 9p rw,dirsync,aname=drvfs;path=C:\\;uid=0;" | busybox grep -c '^[^[:space:]]* [^[:space:]]* 9p [^[:space:]]*aname=drvfs;.*'
1
```

holmanb (changes requested):
@CarlosNihelton Looking better, thanks for the updates. I have a couple more requested changes.

tools/ds-identify
igalic: yeah, if we don't, everything else will also needlessly fail

igalic: the l in wsl stands for Linux, so no other platform should match that, to begin with

blackboxsw (changes requested):
Thanks for this branch @CarlosNihelton and @holmanb for the review here. A note that we should not merge this branch until we also have DataSourceWSL.py defined and merged. Normally this is all done in one giant PR, python datasouce + ds-identify implementation, but I get that separating the PR into two unique PRs may make this review easier to review/land. No need to include DataSourceWSL.py implementation if not desired, we just need to wait to merge this PR until we are certain DataSourceWSL exists.

The reason we can't land a ds-identify only branch before the matching DataSourceWSL.py implementation because if ds-identity does actually detect WSL, it can't actually choose that datasource because it doesn't yet exist in the python module `cloudinit.sources.DatasourceWSL`

If ds-identify does detect the possibility of WSL during systemd generator timeframe, the python code of cloud-init will emit and error like the following:
```
root@nnn:~# cloud-init status --long
status: done
extended_status: degraded done
last_update: Mon, 08 Jan 2024 20:37:18 +0000
detail:
DataSourceLXD
errors: []
recoverable_errors:
ERROR:
	- Could not import DataSourceWSL. Does the DataSource exist and is it importable?
```


Marking 'request changes' just to track on objection to merging until we land the prerequisite PR delivering cloudinit/sources/DataSourceWSL.py

holmanb:
@blackboxsw I agree that typically we wouldn't want to see a new datasource land until it is in a working state. However, this is an exception from the norm for a number of reasons: 1) the upstream devs had an in-person meeting with those implementing the datasource 2) they have a shared their implementation plan and gotten agreement from us 3) we know that engineering time is allocated to this effort. Compare this to PRs where an unknown dev from a cloud provider reaches out for code inclusion without any prior contact or expectations set.

> If ds-identify does detect the possibility of WSL during systemd generator timeframe, the python code of cloud-init will emit and error like the following:

This is only a hypothetical scenario, not a failure path that should be seen on any supported version of cloud-init. A positive identification of WSL must only happen on WSL. If we merge this and someone tries running cloud-init on WSL - it isn't a supported platform of cloud-init (yet) so what is the harm? Additionally, to get to the failure path that you are describing, one would have to either 1) include `WSL` in the `datasource_list` of their image or 2) use the built-in list and detect no other datasource (WSL is last on the built-in list). Both of these are unsupported scenarios - both before and after we merge this PR.

If we merge this and this gets detected on a not-WSL platform, then that is a bug in this implementation of `dscheck_WSL` which would exist whether or not there is a Python datasource implemented. If we merge this before the Python datasource implementation, we could get daily tests running sooner and get an earlier detection of bugs like that.

> if ds-identity does actually detect WSL, it can't actually choose that datasource because it doesn't yet exist in the python module cloudinit.sources.DatasourceWSL

Not wrong, but what is to stop us from implementing this in parts? WSL support would not be "complete", but this wouldn't break any existing datasources.

> And I think we want to wait to land this particular PR until we have an implementation of DataSourceWSL.py merged.

@blackboxsw I'm fine with waiting on this PR if you really want to. In my opinion, however, the primary reason to do this is to keep a standard upstream process for all datasource proposals. The opposing argument is that if we go ahead with a merge we can avoid this PR bitrotting and also get sooner detection of bugs from daily testing. Since the effort to get a `WSL` datasource merged is targeting the 24.04 release of Ubuntu, I think that this is a pretty low risk PR to merge.

holmanb (approved):
Great work on this PR @CarlosNihelton, thanks! This looks ready to me, but you'll need to wait on the concerns of @blackboxsw before this can merge.

blackboxsw:
@holmanb your points are well made. There really isn't a significant concern about a "recoverable error" message in cloud-init on WSL images booting with tip of cloud-init because we, today, have no support whatsoever for WSL images, so it's a minor step toward the end goal and allows the WSL devs to start their local testing during initial datasource development.

And since we have a priority commitment from the devs for WSL datasource support. There is little harm in cloud-init trying to detect WSL environments and emitting a warning/error that WSL support is not yet present.

Let's not let me block this since the work is active and making progress. Thanks for defending good @CarlosNihelton's work :)

tools/ds-identify
```diff
@@ -1565,6 +1565,29 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+dscheck_WSL() {
+    if [ ! "${DI_UNAME_KERNEL_NAME}" = "Linux" ]; then
```
blackboxsw, line 1569: I'm pretty sure all distros we support run ds-identify from POSIX-compliant implementations. But, there is an off chance that a minimal image may have a busybox implementation somewhere that doesn't adhere strictly to POSIX order of operators here.
In this case and below, to be clear in all cases, let's just use the `!=` operator to be safe.
```suggestion
    if [ "${DI_UNAME_KERNEL_NAME}" != "Linux" ]; then
```

tools/ds-identify
```diff
@@ -1565,6 +1565,29 @@ dscheck_VMware() {
...
+        return "${DS_NOT_FOUND}"
+    fi
+
+    if [ ! "${DI_VIRT}" = "wsl" ]; then
```
blackboxsw, line 1573: Safety for potential non-POSIX compliant images
```suggestion
    if [ "${DI_VIRT}" != "wsl" ]; then
```

blackboxsw (changes requested):
Minor suggestion `just in case`™

blackboxsw (approved):
+! LGTM!

PR #4733, opened 2024-01-03
fix(ds-identify): Return code 2 is a valid result, use cached value

holmanb:
```
fix(ds-identify): Return code 2 is a valid result, use cached value
      
Commit de5fc365f1 added a new return code (2) for indicating disabled status.
Without this change, ds-identify will re-run when it shouldn't, and users will
see a bogus log message:

    previous run returned unexpected '$ret'. Re-running.

Fix it.
```


blackboxsw (approved):
LGTM, reasonable update for this. I think we're ok without a unit test here.

PR #4736, opened 2024-01-04
docs: Remove euca reference

TheRealFalcon:
## Proposed Commit Message

```
docs: Remov euca reference and example script

We don't recommend Eucalyptus or include it any of our tutorials or
other examples. Including it without any context or explanation makes
no sense.
```

### DIFF
--- a/doc/rtd/explanation/format.rst
+++ b/doc/rtd/explanation/format.rst
@@ -50,22 +50,6 @@ archive.
 User data scripts can optionally render cloud instance metadata variables using
 jinja templating. See :ref:`instance_metadata` for more information.
 
-Example script
---------------
-
-Create a script file :file:`myscript.sh` that contains the following:
-
-.. code-block::
-
-   #!/bin/sh
-   echo "Hello World.  The time is now $(date -R)!" | tee /root/output.txt
-
-Now run:
-
-.. code-block:: shell-session
-
-   $ euca-run-instances --key mykey --user-data-file myscript.sh ami-a07d95c9
-
 Kernel command line
 ===================
 


doc/rtd/explanation/format.rst
```diff
@@ -50,22 +50,6 @@ archive.
 User data scripts can optionally render cloud instance metadata variables using
 jinja templating. See :ref:`instance_metadata` for more information.
 
-Example script
```
holmanb, line 53: Isn't this still a valid user data script? Why should we remove that part?

TheRealFalcon: Is there value to showing an arbitrary shell script?

holmanb: Not a ton, but I think there is some value. Without an example for each type, our format page is a dense wall of text that is harder to skim quickly.

I'd personally love to see each of our documented "User data format" types show - or link to in the case of gzip (if gzip is even a user data format) - a very simple example of a working user data of that format which could be passed to cloud-init, with an explanation of what it does.

Note: this section is still missing important information (do user-data scripts run on every boot? first boot? etc...)

holmanb: If you still want to remove it, I'm fine with that. It really doesn't add much on its own. That said, I'll probably add an example back in when implementing [this docs change](https://github.com/canonical/cloud-init/issues/4739) at some point in the future.

holmanb (approved):
Up to you. I'm fine with the change as-is, but would like to see eventually an example per format.

TheRealFalcon:
I'll close this PR for now. It was just one of those things I saw when trying to find something else so figured I'd change it right away. Instead I'll add a comment to https://github.com/canonical/cloud-init/issues/4739

holmanb:
> I'll close this PR for now. It was just one of those things I saw when trying to find something else so figured I'd change it right away. Instead I'll add a comment to #4739

Gotcha, sounds good +1

PR #4737, opened 2024-01-04
doc: Document how cloud-init is, not how it was

holmanb:
## Additional Context
What value is this content to a user today?

### DIFF
--- a/doc/rtd/explanation/format.rst
+++ b/doc/rtd/explanation/format.rst
@@ -33,9 +33,8 @@ Begins with: ``#cloud-config`` or ``Content-Type: text/cloud-config`` when
 using a MIME archive.
 
 .. note::
-   New in ``cloud-init`` v. 18.4: Cloud config data can also render cloud
-   instance metadata variables using jinja templating. See
-   :ref:`instance_metadata` for more information.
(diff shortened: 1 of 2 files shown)


doc/rtd/explanation/format.rst
```diff
@@ -33,9 +33,8 @@ Begins with: ``#cloud-config`` or ``Content-Type: text/cloud-config`` when
 using a MIME archive.
 
 .. note::
-   New in ``cloud-init`` v. 18.4: Cloud config data can also render cloud
```
blackboxsw, line 36: It would be nice to have a "New in version XX." in our docs sort of like python project does https://docs.python.org/3/library/collections.html#chainmap-objects which tells users in our latest docs what version they can expect to find support for some significant feature. While this doesn't necessarily apply to this particular doc reference, I think it may be helpful for avoid customers trying a feature they see in docs that doesn't work on a stale version of cloud-init that they have in their images. It would also allow users to avoid back tracking through older docs to see when such a feature was introduced.

blackboxsw: CC @s-makin for your thoughts on how we might think about approaching something like this in cloud-init docs.

blackboxsw (approved):

PR #4740, opened 2024-01-04
feat(apt): remove /etc/apt/sources.list when deb22 preferred

blackboxsw:
## Proposed Commit Message

```
feat(apt): remove /etc/apt/sources.list when deb22 preferred (#....)

Some images may contain /etc/apt/sources.list which cloud-init historically would silently overwrite on first boot. Remove this file rather than preserve a vestigial disabled file as /etc/apt/sources.list.disabled because the disabled file may just confuse users as to the why the file exists and whether they should consider re-enabling it.

If a customer reenables a previously disabled
/etc/apt/sources.list.disabled file and the same APT sources are defined in both /etc/apt/sources.list and deb822 formatted /etc/apt/sources.list.d/ubuntu.sources, apt update and cache commands will emit warnings about duplicate sources being detected which could break tooling which attempts to parse apt command output.

Let's avoid leaving unnecessary artifacts on the sytem given all previous versions of cloud-init would silently overwrite this /etc/apt/sources.list file in on first boot anyway.
```

## Additional Context
Recommendation is that we don't want to leave vestigial apt configuration artifacts around that otherwise cloud-init would have overwritten. Having extra .disabled apt sources file could lead to customers trying to re-enable /etc/apt/sources.list which would collide with the now duplicate deb822 sources entries in /etc/apt/sources.list.d/ubuntu.sources that cloud-init sets up on first boot. The duplicate APT source configs emit warnings for update update apt upgrade and apt-cache commands. 
Better to leave the cloud-init warning in cloud-init.log to report this condition, but still remove the /etc/apt/sources.list file.

## Test Steps
```
CLOUD_INIT_CLOUD_INIT_SOURCE=IN_PLACE CLOUD_INIT_OS_IMAGE=noble tox -e  integration-tests tests/integration_tests/modules/test_apt_functionality.py
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -679,20 +679,12 @@ def generate_sources_list(cfg, release, mirrors, cloud):
             )
             aptsrc_file = apt_sources_list
     disabled = disable_suites(cfg.get("disable_suites"), rendered, release)
-    disable_apt_sources_list = False
-    if aptsrc_file == apt_sources_deb822 and os.path.exists(apt_sources_list):
-        disable_apt_sources_list = True
     util.write_file(aptsrc_file, disabled, mode=0o644)
-    if disable_apt_sources_list:
+    if aptsrc_file == apt_sources_deb822 and os.path.exists(apt_sources_list):
         LOG.warning(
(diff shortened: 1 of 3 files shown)


aciba90 (approved):
LGTM, thanks!

PR #4742, opened 2024-01-05
Describe ds-identify irrespective of distro

holmanb:
## Commit message 
```
doc(ds-identify): Describe ds-identify irrespective of distro
    
- add links to the detect stage where appropriate
- add note about datasource_list requiring a subset of YAML
- rename "Generator" stage to "Detect"
```
## Additional Context

This stage may have originally been specific to systemd-generator, but today it is used by various other init systems for the same purpose. Update the docs to use generic vocabulary, include some links to this section where appropriate, and update language to be direct. Also include the importance of the `datasource_list` key being a subset of YAML.  

## Merge type

### DIFF
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -19,7 +19,8 @@ Detect
 A platform identification tool called ``ds-identify`` runs in the first stage.
 This tool detects which platform the instance is running on. This tool is
 integrated into the init system to disable cloud-init when no platform is
-found, and enable cloud-init when a valid platform is detected.
+found, and enable cloud-init when a valid platform is detected. This stage
+might not be present for every installation of cloud-init.
 
 .. _boot-Local:
 
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -3,23 +3,23 @@
 Boot stages
 ***********
 
-To be able to provide the functionality that it does, ``cloud-init`` must be
-integrated into the boot in a fairly controlled way. There are five
-stages to boot:
+There are five stages to boot:
 
-1. Generator
+1. Detect
 2. Local
 3. Network
 4. Config
 5. Final
 
-.. _boot-Generator:
+.. _boot-Detect:
 
-Generator
-=========
-When booting under ``systemd``, a generator will run that determines if
-cloud-init.target should be included in the boot goals. ``ds-identify``
-runs at this stage.
+Detect
+======
+
+A platform identification tool called ``ds-identify`` runs in the first stage.
+This tool detects which platorm the instance is running on. This tool is
+integrated into the init system to disable cloud-init when no platform is
+found, and enable cloud-init when a valid platform is detected.
 
 .. _boot-Local:
 
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -17,7 +17,7 @@ Detect
 ======
 
 A platform identification tool called ``ds-identify`` runs in the first stage.
-This tool detects which platorm the instance is running on. This tool is
+This tool detects which platform the instance is running on. This tool is
 integrated into the init system to disable cloud-init when no platform is
 found, and enable cloud-init when a valid platform is detected.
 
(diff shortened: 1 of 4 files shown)


doc/rtd/explanation/boot.rst
```diff
@@ -3,23 +3,23 @@
...
-Generator
-=========
-When booting under ``systemd``, a generator will run that determines if
-cloud-init.target should be included in the boot goals. ``ds-identify``
-runs at this stage.
+Detect
+======
+
+A platform identification tool called ``ds-identify`` runs in the first stage.
+This tool detects which platorm the instance is running on. This tool is
```
TheRealFalcon, line 20:
```suggestion
This tool detects which platform the instance is running on. This tool is
```

doc/rtd/explanation/boot.rst
```diff
@@ -3,23 +3,23 @@
...
 
-.. _boot-Generator:
+.. _boot-Detect:
 
-Generator
-=========
-When booting under ``systemd``, a generator will run that determines if
-cloud-init.target should be included in the boot goals. ``ds-identify``
-runs at this stage.
+Detect
```
TheRealFalcon, line 16: Since we've removed the mention of systemd, should we mention that this stage may not be present for every installation of cloud-init?

doc/rtd/reference/base_config_reference.rst
```diff
@@ -232,19 +232,24 @@ Automatically includes ``cloudinit.sources``.
...
+attempts to discover on boot. By default, this is defined in
+:file:`/etc/cloud/cloud.cfg.d`.
+
+There are a few reasons to modify the ``datasource_list``:
+
+1. Override default datasource discovery priority order
+2. Force cloud-init to use a specific datasource: A single entry in
+   the list (or a single entry and ``None``) will override datasource
+   discovery, which will force the specified datasource to run.
+3. Remove known invalid datasources: On distros that do not use ``ds-identify``
```
TheRealFalcon, line 245: Nit: Moving the end clause to the beginning flows better to me. I.e.,  `Remove the known invalid datasources: This might improve boot speed on distros that do not use ...`.

No biggie if you disagree

doc/rtd/reference/datasources/lxd.rst
```diff
@@ -24,8 +24,8 @@ Disabling ``security.devlxd`` over the life of the container will result in
 warnings from ``cloud-init``, and ``cloud-init`` will keep the
 originally-detected LXD datasource.
 
-The LXD datasource is detected as viable by ``ds-identify`` during ``systemd``
-generator time when either ``/dev/lxd/sock`` exists, or
+The LXD datasource is detected as viable by ``ds-identify`` during the
+:ref:`detect stage<boot-Detect>` when either ``/dev/lxd/sock`` exists, or
```
TheRealFalcon, line 28:
```suggestion
:ref:`detect stage<boot-Detect>` when either ``/dev/lxd/sock`` exists or
```
Same reason as above

TheRealFalcon (approved):
Nice improvements! All of my comments are pretty minor so I trust you can make the necessary changes.

doc/rtd/explanation/boot.rst
holmanb: Good question.

I think that the list of supported distros that don't have ds-identify integration is dwindling. At this point I think it is just Alpine, Gentoo, and OpenBSD. From some conversations in IRC, I think Alpine will gain support for it in the future (will require `ds-identify` changes to some flags for busybox tooling compatibility), and I suspect that the OpenBSD work shouldn't be too hard either. I'm happy to do the Gentoo bits if that's the last remaining distro.

I'd like to see this become an expected cloud-init behavior for future distro additions, but I'm fine with documenting how it is for now - up to you.

holmanb: Actually I just added it for now, can remove the statement when it is no longer true.

holmanb:
> Nice improvements! All of my comments are pretty minor so I trust you can make the necessary changes.

Thanks! I made the requested changes.

doc/rtd/explanation/boot.rst
igalic: OpenBSD has basic ds-id support, detecting discs doesn't work yet.

OpenBSD's init also doesn't support running any of our stages under it, because they don't remain running, so it gets really confused.

The strategy right now is to tell people to enable it via init.local(8).
See README in pkg/ dir: https://codeberg.org/meena/openbsd-ports/src/branch/add/cloud-init/sysutils/cloud-init
(technically, this is still init, and your statement(s) below are still correct, it's not the way people are used to from the average daemon)

PR #4744, opened 2024-01-05
Hostname tests

TheRealFalcon:
See individual commits.

## Additional Context
This is due to the new `create_hostname_file` key being necessary on GCE. The layout is a bit verbose, but due to the way we inject userdata with marks, I'm not sure there's a better way to do it.

## Merge type

### DIFF
--- a/tests/integration_tests/modules/test_set_hostname.py
+++ b/tests/integration_tests/modules/test_set_hostname.py
@@ -4,12 +4,12 @@ This module specify two tests: One updates only the hostname and the other
 one updates the hostname and fqdn of the system. For both of these tests
 we will check is the changes requested by the user data are being respected
 after the system is boot.
-
-(This is ported from
-``tests/cloud_tests/testcases/modules/set_hostname.yaml`` and
-``tests/cloud_tests/testcases/modules/set_hostname_fqdn.yaml``.)"""
+"""
 
 import pytest
+
+from tests.integration_tests.integration_settings import PLATFORM
+from tests.integration_tests.releases import CURRENT_RELEASE, NOBLE
 
 USER_DATA_HOSTNAME = """\
 #cloud-config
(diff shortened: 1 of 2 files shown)


tests/integration_tests/modules/test_set_hostname.py
```diff
@@ -30,35 +30,100 @@
...
+USER_DATA_PREFER_FQDN_FILE = """\
+#cloud-config
+prefer_fqdn_over_hostname: {}
+hostname: cloudinit1
+fqdn: cloudinit2.test.io
+create_hostname_file: true
+"""
+
+
+class TestCreateHostnameFile:
```
holmanb, line 104: Do you think we need this test when `False == REQUIRES_FILE_FLAG`? iirc this key is a no-op there, right?

holmanb (approved):
+1, but pinging @catmsred for a review as well since she's more familiar with this change

tests/integration_tests/modules/test_set_hostname.py
TheRealFalcon: It's testing both here. I think it's good to still check that the file is actually created and applied successfully.

...or am I misunderstanding your question?

tests/integration_tests/modules/test_set_hostname.py
```diff
@@ -30,35 +30,100 @@
...
+
 
 @pytest.mark.ci
 class TestHostname:
     @pytest.mark.user_data(USER_DATA_HOSTNAME)
     def test_hostname(self, client):
         hostname_output = client.execute("hostname")
-        assert "cloudinit2" in hostname_output.strip()
+        if REQUIRES_FILE_FLAG:
+            assert "cloudinit2" not in hostname_output.strip()
```
catmsred, line 42: How are the instances created?  I ask because the randomly assigned google DHCP hostname could include cloudinit2 if the instance name includes that.

TheRealFalcon: In integration tests, the instance name will never be that unless you very intentionally instrument it to happen. Instance names should have a 'cloud-init' + timestamp in the name.

catmsred (approved):

tests/integration_tests/modules/test_set_hostname.py
blackboxsw: @catmsred from cloud-init's integration tests, [the launched instance name defined at launch time here](https://github.com/canonical/cloud-init/blob/main/tests/integration_testshe/clouds.py#L333)

Issue #4741, opened 2024-01-05
cloud-init not select `NetworkManager` renderer if `/etc/NetworkManager/NetworkManager.conf` absents


bin456789:
# Bug report

https://download.opensuse.org/tumbleweed/appliances/openSUSE-Tumbleweed-Minimal-VM.x86_64-Cloud.qcow2
This image have `NetworkManager` installed. No `networkd` installed.
But Cloud-init selects `networkd` renderer instead of `NetworkManager`

## Due to
cloud-init not select `NetworkManager` renderer if `/etc/NetworkManager/NetworkManager.conf` absents

## Fix
When select renderers
 - `NetworkManager`: Don't check if `/etc/NetworkManager/NetworkManager.conf` exists 
 - `networkd`: Check if `systemd-networkd` services exists

TheRealFalcon:
Thanks for the bug report!

PR #4745, opened 2024-01-05
fix: relax NetworkManager renderer rules

TheRealFalcon:
## Proposed Commit Message

```
fix: relax NetworkManager renderer rules

Some images don't include /etc/NetworkManager/NetworkManager.conf by
default, so don't require that file to detect NetworkManager.

Fixes GH-4741
```

## Additional Context
See the image in https://github.com/canonical/cloud-init/issues/4741 for an example.

## Checklist

## Merge type


holmanb (approved):
No objections, but are we sure networkmanager even works with no congfiguration? If it does, I'm curious why this was ever there.

TheRealFalcon:
> but are we sure networkmanager even works with no congfiguration?

[Yes](https://www.networkmanager.dev/docs/api/latest/NetworkManager.conf.html), it can live in multiple places or not exist at all.

> why this was ever there

I over-reacted to a previous check that was overly broad, and added this before the systemd service check existed.

PR #4746, opened 2024-01-06
dhcpcd support

holmanb:
## Current State
```
feat(dhcp): Add support for dhcpcd in ephemeral network timeframe

- dhcp: rework client api
- dhcp/isc-dhclient: consolidate duplicate code paths
- distro: make dhcp_client a distro property [1]
- distro: make fallback_interface a distro property [1]
- cloudstack: simplify "domain-name" and "dhcp-server-identifier" codepaths
- scaleway: eliminate unnecessary fallback interface detection
- ec2: eliminate old unnecessary pickle attribute check for "fallback_nic"[2]
- tests/unittest: new dhcpcd tests
- tests/integration: new dhcpcd tests for azure, ec2, gce, and openstack

[1] required because datasource objects can access the Distro attributes but
    not the other way around. Also this is a very distro-specific codepath
    and much simplification to the code will be possible as a result of this.
[2] This "bug" was never a supported use case. Calling cloud-init manually
    after upgrade not via the init system isn't supported, but this was
    "fixed" before the use case was fully understood.
```

### Implementation notes
```
Summary:

Besides implementing the dhcpcd client, several changes were made to the codebase in order
to make using different clients interchangeably possible. The big things that changed are
that the dhcp client instance now exists as a property under the Distro object, and the
Distro object acquired the implementation of the fallback_interface property from DataSource
object. These two things help make the code more re-usable and modular - different linux and
bsd distributions should be able to use the same code base with different clients - and
should dhcpcd ever loose maintainership, we will have an alternative already implemented
(udhcpcd) and alternatives more easily implemented.

dhcp.py:
 - add type annotations, typing fixes, comments
 - unify the dhcp client interface such that all clients have the same ephemeral methods[1]

dhpc.py::Dhcpcd
- implement the client

dhpc.py::IscDhclient
- eliminate duplicate parsing codepaths (between ephemeral and non-ephemeral)
- add get_key_from_latest_lease() to simplify cloudstack's parsing of non-ephemeral keys

distro/__init__.py
- store which dhcp client is instantiated and selected within a Distro attribute
- move dhcp client selection from dhcp.py
- move fallback_interface[2] from source/__init__.py

ephemeral.py
- use already-available distro object to access selected dhcp client

sources/__init__.py
- move fallback interface to distro/__init__.py[2]

DataSourceCloudStack
- use already-available Distro object to access selected dhcp client "domain-name" and
  "dhcp-server-identifier"
- use new dhclient helper (and therefore the unified dhclient parsing codepath)

DataSourceScaleWay:
- eliminate redundant fallback interface detection

DataSourceEc2
- eliminate old unnecessary pickle attribute check for "fallback_nic"[3]


[1] dhclient still has some extra methods that are used for non-ephemeral lease introspection
[2] Required because datasource objects  can access the Distro attributes but not the other
    way around. Also this is a very distro-specific codepath and much simplification to the
    code will be possible as a result of this.
[3] which was never a supported use case. Calling cloud-init manually after upgrade not via
the init system isn't supported, but this was "fixed" before the use case was fully sussed
out. Also a very unlikely upgrade path at this point.
```
### Features
- [x] new integration test [xfail noble uses dhcpcd]
- [x] new unit tests
- [x] timeout
- [x] unit tests
- [x] disable hooks
- [x] openstack support (classless static routes)
- [x] cloudstack support (dhcp-server-identifier, domain-name)

### Remaining Work (separate PR)
- [ ] feature: support azure corner case (unknown options)
- [ ] feature: InfiniBand support (plan to ping @wmousa, who contributed isc-dhclient Infiniband code for review)
- [ ] docs
- [ ] Ubuntu packaging (separate PR)

## Test Steps
Boot an Ubuntu instance
Reorder dhcp client priority in `distros/__init__.py`
Install dhcpcd

```
cloud-init clean --logs
cloud-init init --local
vim /var/log/cloud-init.log
```

## Merge type

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -147,7 +147,6 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     resolve_conf_fn = "/etc/resolv.conf"
 
     osfamily: str
-    dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd, dhcp.Udhcpc]
     # Directory where the distro stores their DHCP leases.
     # The children classes should override this with their dhcp leases
     # directory
@@ -162,14 +161,12 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         self._cfg = cfg
         self.name = name
         self.networking: Networking = self.networking_cls()
-        self.dhcp_client_priority = [
-            dhcp.IscDhclient,
-            dhcp.Dhcpcd,
-            dhcp.Udhcpc,
-        ]
+        self.dhcp_client_priority = dhcp.ALL_DHCP_CLIENTS
         self.net_ops = iproute2.Iproute2
         self._runner = helpers.Runners(paths)
         self.package_managers: List[PackageManager] = []
+        self._dhcp_client = None
+        self._fallback_interface = None
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         """Perform deserialization fixes for Distro."""
@@ -182,6 +179,10 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             # either because it isn't present at all, or because it will be
             # missing expected instance state otherwise.
             self.networking = self.networking_cls()
+        if not hasattr(self, "_dhcp_client"):
+            self._dhcp_client = None
+        if not hasattr(self, "_fallback_interface"):
+            self._fallback_interface = None
 
     def _validate_entry(self, entry):
         if isinstance(entry, str):
@@ -273,6 +274,66 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             "Legacy function '_write_network' was called in distro '%s'.\n"
             "_write_network_config needs implementation.\n" % self.name
         )
+
+    @property
+    def dhcp_client(self) -> dhcp.DhcpClient:
+        """access the distro's preferred dhcp client
+
+        if no client has been selected yet select one - uses
+        self.dhcp_client_priority, which may be overriden in each distro's
+        object to eliminate checking for clients which will not be provided
+        by the distro
+        """
+        if self._dhcp_client:
+            return self._dhcp_client
+
+        # no client has been selected yet, so pick one
+        #
+        # set the default priority list to the distro-defined priority list
+        dhcp_client_priority = self.dhcp_client_priority
+
+        # if the configuration includes a network.dhcp_client_priority list
+        # then attempt to use it
+        config_priority = util.get_cfg_by_path(
+            self._cfg, ("network", "dhcp_client_priority"), []
+        )
+
+        if config_priority:
+            # user or image builder configured a custom dhcp client priority
+            # list
+            found_clients = []
+            LOG.debug(
+                "Using configured dhcp client priority list: %s",
+                config_priority,
+            )
+            for client_configured in config_priority:
+                for client_class in dhcp.ALL_DHCP_CLIENTS:
+                    if client_configured == client_class.client_name:
+                        found_clients.append(client_class)
+                        break
+                else:
+                    LOG.warning(
+                        "Configured dhcp client %s is not supported, skipping",
+                        client_configured,
+                    )
+            # If dhcp_client_priority is defined in the configuration, but none
+            # of the defined clients are supported by cloud-init, then we don't
+            # override the distro default. If at least one client in the
+            # configured list exists, then we use that for our list of clients
+            # to check.
+            if found_clients:
+                dhcp_client_priority = found_clients
+
+        # iterate through our priority list and use the first client that is
+        # installed on the system
+        for client in dhcp_client_priority:
+            try:
+                self._dhcp_client = client()
+                LOG.debug("DHCP client selected: %s", client.client_name)
+                return self._dhcp_client
+            except (dhcp.NoDHCPLeaseMissingDhclientError,):
+                LOG.debug("DHCP client not found: %s", client.client_name)
+        raise dhcp.NoDHCPLeaseMissingDhclientError()
 
     @property
     def network_activator(self) -> Optional[Type[activators.NetworkActivator]]:
@@ -1247,6 +1308,22 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             "-sf",
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
+
+    @property
+    def fallback_interface(self):
+        """Determine the network interface used during local network config."""
+        if self._fallback_interface is None:
+            self._fallback_interface = net.find_fallback_nic()
+            if not self._fallback_interface:
+                LOG.warning(
+                    "Did not find a fallback interface on distro: %s.",
+                    self.name,
+                )
+        return self._fallback_interface
+
+    @fallback_interface.setter
+    def fallback_interface(self, value):
+        self._fallback_interface = value
 
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -17,6 +17,7 @@ import stat
 import string
 import urllib.parse
 from collections import defaultdict
+from contextlib import suppress
 from io import StringIO
 from typing import (
     Any,
@@ -1324,6 +1325,59 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     @fallback_interface.setter
     def fallback_interface(self, value):
         self._fallback_interface = value
+
+    @staticmethod
+    def get_proc_ppid(pid: int) -> Optional[int]:
+        """Return the parent pid of a process by parsing /proc/$pid/stat"""
+        match = Distro._get_proc_stat_by_index(pid, 4)
+        if match is not None:
+            with suppress(ValueError):
+                return int(match)
+            LOG.warning("/proc/%s/stat has an invalid ppid [%s]", pid, match)
+        return None
+
+    @staticmethod
+    def get_proc_pgid(pid: int) -> Optional[int]:
+        """Return the parent pid of a process by parsing /proc/$pid/stat"""
+        match = Distro._get_proc_stat_by_index(pid, 5)
+        if match is not None:
+            with suppress(ValueError):
+                return int(match)
+            LOG.warning("/proc/%s/stat has an invalid pgid [%s]", pid, match)
+        return None
+
+    @staticmethod
+    def _get_proc_stat_by_index(pid: int, field: int) -> Optional[int]:
+        """
+        parse /proc/$pid/stat for a specific field as numbered in man:proc(5)
+
+        param pid: integer to query /proc/$pid/stat for
+        param field: field number within /proc/$pid/stat to return
+        """
+        try:
+            content: str = util.load_file(
+                "/proc/%s/stat" % pid, quiet=True
+            ).strip()  # pyright: ignore
+            match = re.search(
+                r"^(\d+) (\(.+\)) ([RSDZTtWXxKPI]) (\d+) (\d+)", content
+            )
+            if not match:
+                LOG.warning(
+                    "/proc/%s/stat has an invalid contents [%s]", pid, content
+                )
+                return None
+            return int(match.group(field))
+        except IOError as e:
+            LOG.warning("Failed to load /proc/%s/stat. %s", pid, e)
+        except IndexError:
+            LOG.warning(
+                "Unable to match field %s of process pid=%s (%s) (%s)",
+                field,
+                pid,
+                content,  # pyright: ignore
+                match,  # pyright: ignore
+            )
+        return None
 
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -359,12 +359,21 @@ class IscDhclient(DhcpClient):
         ppid = "unknown"
         daemonized = False
         pid_content = None
+        debug_msg = ""
         for _ in range(sleep_cycles):
-            pid_content = util.load_file(pid_file).strip()
             try:
+                pid_content = util.load_file(pid_file).strip()
                 pid = int(pid_content)
+            except FileNotFoundError:
+                debug_msg = (
+                    f"No PID file found at {pid_file}, "
+                    "dhclient is still running"
+                )
             except ValueError:
-                pass
+                debug_msg = (
+                    f"PID file contained [{pid_content}], "
+                    "dhclient is still running"
+                )
             else:
                 ppid = util.get_proc_ppid(pid)
                 if ppid == 1:
@@ -373,6 +382,8 @@ class IscDhclient(DhcpClient):
                     daemonized = True
                     break
             time.sleep(sleep_time)
+        else:
+            LOG.debug(debug_msg)
 
         if not daemonized:
             LOG.error(
@@ -583,15 +594,11 @@ class Dhcpcd(DhcpClient):
         sleep_time = 0.01
         sleep_cycles = int(self.timeout / sleep_time)
 
-        # ISC dhclient needs the interface up to send initial discovery packets
+        # dhcpcd needs the interface up to send initial discovery packets
         # Generally dhclient relies on dhclient-script PREINIT action to bring
         # the link up before attempting discovery. Since we are using
         # -sf /bin/true, we need to do that "link up" ourselves first.
         distro.net_ops.link_up(interface)
-
-        # TODO: disabling hooks means we need to get all of the files in
-        # /lib/dhcpcd/dhcpcd-hooks/ and pass each of those with the --nohook
-        # argument to dhcpcd
         try:
             command = [
                 self.dhcp_client_path,  # pyright: ignore
@@ -612,6 +619,23 @@ class Dhcpcd(DhcpClient):
             # Attempt cleanup and leave breadcrumbs if it fails, but return
             # the lease regardless of failure to clean up dhcpcd.
             if lease:
+                # pid file arrival can take very long, and stdout might contain
+                # the PID, so attempt to use a stdout parsed PID
+                for line in reversed(out.split("\n")):
+                    if "forked to background, child pid " in line:
+                        try:
+                            os.kill(int(line.split()[-1]), signal.SIGKILL)
+                        except ValueError:
+                            LOG.debug(
+                                "Clouldn't kill process: couldn't parse "
+                                "pid from stdout"
+                            )
+                        except ProcessLookupError:
+                            LOG.debug(
+                                "Couldn't kill process, it already exited"
+                            )
+                LOG.debug("PID not in stdout, waiting for PID file")
+
                 # Note: the pid file location depends on the arguments passed
                 # it can be discovered with the -P flag
                 pid_file = subp.subp([*command, "-P"]).stdout
@@ -712,7 +736,7 @@ class Dhcpcd(DhcpClient):
             return self.parse_dhcpcd_lease(
                 subp.subp(
                     [
-                        "dhcpcd",
+                        self.dhcp_client_path,
                         "--dumplease",
                         "--ipv4only",
                         interface,
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -600,6 +600,17 @@ class Dhcpcd(DhcpClient):
         # -sf /bin/true, we need to do that "link up" ourselves first.
         distro.net_ops.link_up(interface)
         try:
+            # Currently dhcpcd doesn't have a workable --oneshot lease parsing
+            # story. All non-daemon lease parsing options on dhcpcd appear
+            # broken:
+            #
+            #   https://github.com/NetworkConfiguration/dhcpcd/issues/285
+            #   https://github.com/NetworkConfiguration/dhcpcd/issues/286
+            #   https://github.com/NetworkConfiguration/dhcpcd/issues/287
+            #
+            # Until fixed, we allow dhcpcd to spawn background processes so
+            # that we can use --dumplease, but when any option above is fixed,
+            # it would be safer to avoid spawning processes using --oneshot
             command = [
                 self.dhcp_client_path,  # pyright: ignore
                 "--ipv4only",  # only attempt configuring ipv4
@@ -619,31 +630,11 @@ class Dhcpcd(DhcpClient):
             # Attempt cleanup and leave breadcrumbs if it fails, but return
             # the lease regardless of failure to clean up dhcpcd.
             if lease:
-                # pid file arrival can take very long, and stdout might contain
-                # the PID, so attempt to use a stdout parsed PID
-                for line in reversed(out.split("\n")):
-                    if "forked to background, child pid " in line:
-                        try:
-                            pid = int(line.split()[-1])
-                            gid = distro.get_proc_pgid(pid)
-                            if gid:
-                                os.kill(gid, signal.SIGKILL)
-                                break
-                        except ValueError:
-                            LOG.debug(
-                                "Clouldn't kill process: couldn't parse "
-                                "pid from stdout"
-                            )
-                        except ProcessLookupError:
-                            LOG.debug(
-                                "Couldn't kill process, it already exited"
-                            )
-                LOG.debug("PID not in stdout, waiting for PID file")
-
                 # Note: the pid file location depends on the arguments passed
                 # it can be discovered with the -P flag
-                pid_file = subp.subp([*command, "-P"]).stdout
+                pid_file = subp.subp([*command, "-P"]).stdout.strip()
                 pid_content = None
+                gid = False
                 debug_msg = ""
                 for _ in range(sleep_cycles):
                     try:
@@ -654,7 +645,15 @@ class Dhcpcd(DhcpClient):
                             LOG.debug(
                                 "killing dhcpcd with pid=%s gid=%s", pid, gid
                             )
-                            os.kill(gid, signal.SIGKILL)
+                            os.killpg(gid, signal.SIGKILL)
+                            break
+                    except ProcessLookupError:
+                        LOG.debug(
+                            "Process group id [%s] has already exited, "
+                            "nothing to kill",
+                            gid,
+                        )
+                        break
                     except FileNotFoundError:
                         debug_msg = (
                             f"No PID file found at {pid_file}, "
@@ -668,7 +667,8 @@ class Dhcpcd(DhcpClient):
                     else:
                         return lease
                     time.sleep(sleep_time)
-                LOG.debug(debug_msg)
+                else:
+                    LOG.debug(debug_msg)
                 return lease
             raise NoDHCPLeaseError("No lease found")
 
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -375,7 +375,7 @@ class IscDhclient(DhcpClient):
                     "dhclient is still running"
                 )
             else:
-                ppid = util.get_proc_ppid(pid)
+                ppid = distro.get_proc_ppid(pid)
                 if ppid == 1:
                     LOG.debug("killing dhclient with pid=%s", pid)
                     os.kill(pid, signal.SIGKILL)
@@ -624,7 +624,11 @@ class Dhcpcd(DhcpClient):
                 for line in reversed(out.split("\n")):
                     if "forked to background, child pid " in line:
                         try:
-                            os.kill(int(line.split()[-1]), signal.SIGKILL)
+                            pid = int(line.split()[-1])
+                            gid = distro.get_proc_pgid(pid)
+                            if gid:
+                                os.kill(gid, signal.SIGKILL)
+                                break
                         except ValueError:
                             LOG.debug(
                                 "Clouldn't kill process: couldn't parse "
@@ -645,8 +649,12 @@ class Dhcpcd(DhcpClient):
                     try:
                         pid_content = util.load_file(pid_file).strip()
                         pid = int(pid_content)
-                        LOG.debug("killing dhcpcd with pid=%s", pid)
-                        os.kill(pid, signal.SIGKILL)
+                        gid = distro.get_proc_pgid(pid)
+                        if gid:
+                            LOG.debug(
+                                "killing dhcpcd with pid=%s gid=%s", pid, gid
+                            )
+                            os.kill(gid, signal.SIGKILL)
                     except FileNotFoundError:
                         debug_msg = (
                             f"No PID file found at {pid_file}, "
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -13,14 +13,12 @@ import signal
 import time
 from contextlib import suppress
 from io import StringIO
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import configobj
 
 from cloudinit import subp, temp_utils, util
 from cloudinit.net import (
-    find_fallback_nic,
-    get_devicelist,
     get_ib_interface_hwaddr,
     get_interface_mac,
     is_ib_interface,
@@ -85,28 +83,6 @@ class NoDHCPLeaseMissingDhclientError(NoDHCPLeaseError):
     """Raised when unable to find dhclient."""
 
 
-class NoDHCPLeaseMissingUdhcpcError(NoDHCPLeaseError):
-    """Raised when unable to find udhcpc client."""
-
-
-def select_dhcp_client(distro):
-    """distros set priority list, select based on this order which to use
-
-    If the priority dhcp client isn't found, fall back to lower in list.
-    """
-    for client in distro.dhcp_client_priority:
-        try:
-            dhcp_client = client()
-            LOG.debug("DHCP client selected: %s", client.client_name)
-            return dhcp_client
-        except (
-            NoDHCPLeaseMissingDhclientError,
-            NoDHCPLeaseMissingUdhcpcError,
-        ):
-            LOG.warning("DHCP client not found: %s", client.client_name)
-    raise NoDHCPLeaseMissingDhclientError()
-
-
 def maybe_perform_dhcp_discovery(distro, nic=None, dhcp_log_func=None):
     """Perform dhcp discovery if nic valid and dhclient command exists.
 
@@ -120,18 +96,8 @@ def maybe_perform_dhcp_discovery(distro, nic=None, dhcp_log_func=None):
         from the dhclient discovery if run, otherwise an empty list is
         returned.
     """
-    if nic is None:
-        nic = find_fallback_nic()
-        if nic is None:
-            LOG.debug("Skip dhcp_discovery: Unable to find fallback nic.")
-            raise NoDHCPLeaseInterfaceError()
-    elif nic not in get_devicelist():
-        LOG.debug(
-            "Skip dhcp_discovery: nic %s not found in get_devicelist.", nic
-        )
-        raise NoDHCPLeaseInterfaceError()
-    client = select_dhcp_client(distro)
-    return client.dhcp_discovery(nic, dhcp_log_func, distro)
+    interface = nic or distro.fallback_interface
+    return distro.dhcp_client.dhcp_discovery(interface, dhcp_log_func, distro)
 
 
 def networkd_parse_lease(content):
@@ -176,6 +142,12 @@ def networkd_get_option_from_leases(keyname, leases_d=None):
 
 class DhcpClient(abc.ABC):
     client_name = ""
+    timeout = 10
+
+    def __init__(self):
+        self.dhcp_client_path = subp.which(self.client_name)
+        if not self.dhcp_client_path:
+            raise NoDHCPLeaseMissingDhclientError()
 
     @classmethod
     def kill_dhcp_client(cls):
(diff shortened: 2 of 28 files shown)


cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
+                    "--waitip=4",  # wait for ipv4 to be configured
+                    "--persistent",  # don't deconfigure when dhcpcd exits
+                    "--noarp",  # don't be slow
+                    interface,
+                ]
+            )
+            return self.parse_dhcp_lease_file(interface)
+        except subp.ProcessExecutionError as error:
+            LOG.debug(
+                "dhclient exited with code: %s stderr: %r stdout: %r",
```
blackboxsw, line 579:
```suggestion
                f"{self.client_name} exited with code: %s stderr: %r stdout: %r",
```

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
 
 
 class Udhcpc(DhcpClient):
     client_name = "udhcpc"
 
-    def __init__(self):
-        self.udhcpc_path = subp.which("udhcpc")
-        if not self.udhcpc_path:
-            LOG.debug("Skip udhcpc configuration: No udhcpc command found.")
-            raise NoDHCPLeaseMissingUdhcpcError()
```
blackboxsw, line 549: We can also drop the exception handling of MissingUdhcpcError  in `select_dhcp_client and drop the definition of this exception now

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
+                        interface,
+                    ],
+                    rcs=[0, 1],
+                ).stdout,
+                interface,
+            )
+
+        except subp.ProcessExecutionError as error:
+            LOG.debug(
+                "dhcpcd exited with code: %s stderr: %r stdout: %r",
```
blackboxsw, line 648:
```suggestion
                f"{self.client_name} exited with code: %s stderr: %r stdout: %r",
```

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
+        pairs a specific lease in order from oldest to newest.
+
+        @raises: InvalidDHCPLeaseFileError on empty or unparseable leasefile
+            content.
+        """
+        try:
+            return cls.parse_dhcpcd_lease(
+                subp.subp(
+                    [
+                        "dhcpcd",
```
blackboxsw, line 637:
```suggestion
                        self.dhcp_client_path,
```

cloudinit/net/dhcp.py
```diff
@@ -199,14 +205,6 @@ def stop_service(cls, dhcp_interface: str, distro):
 class IscDhclient(DhcpClient):
     client_name = "dhclient"
 
-    def __init__(self):
-        self.dhclient_path = subp.which("dhclient")
-        if not self.dhclient_path:
-            LOG.debug(
-                "Skip dhclient configuration: No dhclient command found."
```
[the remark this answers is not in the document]
holmanb: Yep, it was duplicating the same information to the logs so I figured we could drop it here.

blackboxsw (commented):
Looks good @holmanb! This is just a first pass. I am going to manually tinker with dhcpcd now on a system to play around and familiarize myself with the behavior. I agree with static rountes and Azure concerns as areas that need additional review.

tests/unittests/net/test_dhcp.py
```diff
@@ -438,18 +439,14 @@ def test_dhcp_client_failover(self, m_which, m_subp, m_remove, m_fallback):
             subp.ProcessExecutionError(exit_code=-5),
         ]
 
-        m_which.side_effect = [False, True]
```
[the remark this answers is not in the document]
holmanb: I've made another round of changes to this, let me know if your thoughts on  this still stand.

cloudinit/distros/__init__.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, name, cfg, paths):
         self.net_ops = iproute2.Iproute2
         self._runner = helpers.Runners(paths)
         self.package_managers: List[PackageManager] = []
+        self._selected_dhcp_client = None
```
blackboxsw: We may need to be of unpickling across upgrade paths as this class attribute won't exist in the pickled datasource and we are trying to reference it in the `dhcp_client` property below with self._selected_dhcp_client. I think we may need an _unpickle clause for this?

holmanb: > unpickling across upgrade paths

I'm fine with adding an attribute for now, but....

Long term why not just delete pickles on upgrade paths to avoid requiring workarounds like this - we do it across for Python upgrade path already so why not cloud-init? We have some pretty old attributes that we still check for that are very unlikely to ever be used again (see "fallback_nic").

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
     client_name = "dhcpcd"
 
-    def __init__(self):
-        raise NoDHCPLeaseMissingDhclientError("Dhcpcd not yet implemented")
+    def dhcp_discovery(
+        self,
+        interface,
+        dhcp_log_func=None,
+        distro=None,
+    ):
```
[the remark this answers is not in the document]
holmanb: good call - the whole file got a facelift with them :-)

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
+        """Run dhclient on the interface without scripts/filesystem artifacts.
+
+        @param dhclient_cmd_path: Full path to the dhclient used.
```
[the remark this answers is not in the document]
holmanb: nice catch - I updated the rest of the comments too

cloudinit/net/dhcp.py
```diff
@@ -532,22 +530,132 @@ def parse_dhcp_server_from_lease_file(lease_file) -> Optional[str]:
...
+        # -sf /bin/true, we need to do that "link up" ourselves first.
+        distro.net_ops.link_up(interface)
+
+        # TODO: disabling hooks means we need to get all of the files in
+        # /lib/dhcpcd/dhcpcd-hooks/ and pass each of those with the --nohook
+        # argument to dhcpcd
+        try:
+            subp.subp(
+                [
+                    self.client_name,
```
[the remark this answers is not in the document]
holmanb: > Shouldn't we use the full client_path we determined in init?

yes, fixed

> Thanks for the per param comments below!

+1, no problem

cloudinit/net/dhcp.py
holmanb: why do that? this is in dhclient-specific code

holmanb:
Thanks @blackboxsw. A bit of refactoring was required to get nicer "swappable" client implementations working seamlessly. Along the way this involved the distro class getting more code. See the implementation notes above. Also at this point I'm not sure that I have a coherent way to split out into multiple commits so I think that I'll request a squash merge on this one.

Remaining todos that I'd like to include in this PR (pending review on the approach):

- [ ] integration tests:
      - xfail test ubuntu noble uses dhcpcd
      - tests other releases use dhclient
- [ ] the timeout question below

### An open question regarding timeouts:

Currently this code will block forever if dhcpcd never receives a lease, which is not acceptable. However, the timeout flag that dhcpcd supports [is insufficient](https://github.com/NetworkConfiguration/dhcpcd/issues/276). To avoid [manager mode behavior changes](https://github.com/NetworkConfiguration/dhcpcd/discussions/271), I think we still want to use `--oneshot`, but that means we have to craft our own timeout code. Any objections to throwing this call into a separate thread and killing that thread at the timeout (5-10 seconds, whatever timeout we have for dhclient?). It's not an elegant proposal, but I don't have a better one at this time.

cloudinit/distros/__init__.py
blackboxsw: We can't blow away pickls across upgrade path because next boot of cloud-init will find no cached obj.pkl file which triggers cloud-init to rediscover all datasources again and recrawl the entire discovered IMDS which 1, would lead to longer reboot time and 2 expose the VM to not-detecting the datasource that was originally configured in cases where the IMDS is no longer available(sandboxed VMs) or non-responsive resulting in potentially destructive behavior like regenerating SSH host keys.

holmanb: >  1, would lead to longer reboot time 

@blackboxsw This would introduce a longer reboot once per quarter for distros that follow the upstream release cadence, that is only 4 times per year. All of the attribute checks at runtime aren't free either - once the pickle gets blown away the attribute checks wouldn't be required, so this would actually mean faster boots in the future by dropping unnecessary attribute checks when instantiating objects from pickles (which happens 3 times per boot on a fresh boot, 4 times per boot on a new boot). So this would lead to a longer reboot time _4 times a year_, and **shorter** reboot times _N-4 times per year (where N=number of reboots per year)_. The proposal isn't a net loss, it is a tradeoff that would also reduce the risk of introducing bugs due to missing attributes (like this would have been) and also reduce maintenance burden.


> 2 expose the VM to not-detecting the datasource that was originally configured

What new breakage would removing pickles at cloud-init version change introduce which doesn't already exist by removing pickles at Python version change? Surely if this is a supported use case of cloud-init, we must not depend _only_ on the existence of the pickle file but also have some other mechanism to prevent changing discovered datasources whenever a Python major version change happens under cloud-init? If we don't have some other mechanism and we say this is a supported use-case of cloud-init, then I would consider the lack of another mechanism to be a bug in cloud-init (i.e. we could, and probably should, change this regardless of whether we remove old caches on cloud-init upgrade).

Additionally, pickling was only added in release 20.4, ~3 years ago. Before the pickle behavior was added, wouldn't this have been the expected behavior across _every_ reboot (indeed, even every stage of cloud-init)?

TheRealFalcon: > The proposal isn't a net loss, it is a tradeoff that would also reduce the risk of introducing bugs due to missing attributes (like this would have been) and also reduce maintenance burden.

This argument is a bit of a stretch to me. A few attribute checks is going to be incredibly trivial to the overall run time, to the point where I don't think it's worth considering until we've taken much more drastic performance actions.

> What new breakage would removing pickles at cloud-init version change introduce which doesn't already exist by removing pickles at Python version change?

These are two very different upgrade use cases. A new (supported) Python version change is only going to be brought in by a do-release-upgrade or similar for other OSes. That's going to have to be intentionally triggered with ops monitoring and expected changes in behavior. A new version of cloud-init can be installed via unattended upgrade where unexpected changes are much more significant...though I agree that if we have bugs or unwieldly behavior, we should fix it.

Pickling issues have caused some bugs in the past, but our integration tests should handle them now.

That said, I do get that our pickling stuff is a big headache. It also seems less relevant as more clouds have moved towards wanting network data fetched every boot. I'd be open to alternative solutions, but I don't think that removing the pickle on every upgrade is it.

cloudinit/net/dhcp.py
blackboxsw: good pt unnecessary. Was just using my hammer throughout logging comments

cloudinit/net/dhcp.py
```diff
@@ -503,62 +500,224 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+                error.exit_code,
+                error.stderr,
+                error.stdout,
+            )
+            raise NoDHCPLeaseError from error
+
+    @staticmethod
+    def parse_static_routes(routes: str) -> List[Tuple[str, str]]:
+        """
+        classless static routes as returnedd from dhcpcd --dumplease and return
```
blackboxsw, line 666:
```suggestion
        classless static routes as returned from dhcpcd --dumplease and return
```

cloudinit/net/dhcp.py
```diff
@@ -198,82 +170,93 @@ def start_service(cls, dhcp_interface: str, distro):
     def stop_service(cls, dhcp_interface: str, distro):
         distro.manage_service("stop", cls.client_name, rcs=[0, 1])
 
+    @staticmethod
+    def parse_static_routes(routes: str) -> List[Tuple[str, str]]:
+        return []
+
+    def get_newest_lease(self, interface: str) -> Dict[str, Any]:
```
blackboxsw, line 177: All subclasses implement this method any reason to avoid just making it `@abc.abstractmethod` ?

blackboxsw (commented):
Thanks @holmanb for the overview and refactor work here. A number of questions and minor comments inline to digest for this pass and I'll get another pass a bit later today. The refactors look good and I agree with trying to push as much as we can into the distro classes as dhclient selection is appropriate there.

blackboxsw (changes requested):
marking request changes per set of comments above.

cloudinit/distros/__init__.py
```diff
@@ -147,7 +147,6 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     resolve_conf_fn = "/etc/resolv.conf"
 
     osfamily: str
-    dhcp_client_priority = [dhcp.IscDhclient, dhcp.Dhcpcd, dhcp.Udhcpc]
```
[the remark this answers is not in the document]
holmanb: > therwise we see the following across pkg upgrade:

Missing something here, I'm not sure what is supposed to be rendered. 

Also, isn't creating the instance variable in `__init__()` (which we do) sufficient to avoid unpickle issues?

cloudinit/sources/__init__.py
```diff
@@ -610,13 +607,15 @@ def get_vendordata2(self):
     @property
     def fallback_interface(self):
```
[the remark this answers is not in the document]
holmanb: Good point, I think it's doable.

cloudinit/net/dhcp.py
```diff
@@ -503,62 +500,224 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
-                        dhcptok = words[2]
-                        LOG.debug("Found DHCP identifier %s", dhcptok)
-                        latest_address = dhcptok
-        return latest_address
-
-
-class Dhcpcd:
+        lease_file = self.get_newest_lease_file_from_distro(distro)
+        if lease_file:
+            with suppress(FileNotFoundError):
```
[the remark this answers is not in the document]
holmanb: good point

cloudinit/net/dhcp.py
```diff
@@ -503,62 +500,224 @@ def get_latest_lease(lease_dir, lease_file_regex):
                 return latest_file
```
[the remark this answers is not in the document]
holmanb: not sure I follow what you're suggesting. if just this diff is applied, wouldn't this return at the first lease in the directory, which will inevitably have a `mtime > -1.0`?

holmanb:
Thanks for your review @blackboxsw. You made some great points, I think that I've addressed them all. Let me know what you think.

cloudinit/distros/__init__.py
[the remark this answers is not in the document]
blackboxsw: You are correct, I had meant to remove this comment before submitting the review. __init__ will cover this case and perform the necessary setup before we reference self.dhcp_client_priority in our code flow. I had tried reproducing the failure case I expected above and didn't see such failure. Closing this comment as irrelevant.

cloudinit/distros/__init__.py
```diff
@@ -274,6 +279,32 @@ def _write_network(self, settings):
...
+
+        if no client has been selected yet select one - uses
+        self.dhcp_client_priority, which may be overriden in each distro's
+        object to eliminate checking for clients which will not be provided
+        by the distro
+        """
+        if self._dhcp_client:
+            return self._dhcp_client
+        self._select_dhcp_client()
+        return self._dhcp_client
```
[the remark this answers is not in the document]
blackboxsw: Your proposal makes sense to have the private attribute check in only one place to avoid such duplication. And since there are no callers of _select_dhcp_client it makes sense to drop it altogether  thank you for that.

cloudinit/net/dhcp.py
blackboxsw: Since we want this method to be implemented by all subclasses we should drop the `return {}` below correct? Otherwise subclasses won't actually have to implement it.
```diff
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -176,7 +176,7 @@ class DhcpClient(abc.ABC):
 
     @abc.abstractmethod
     def get_newest_lease(self, interface: str) -> Dict[str, Any]:
-        return {}
+        """Get the most recent lease from the ephemeral phase as a dict."""
 
 
 class IscDhclient(DhcpClient):
```

cloudinit/net/dhcp.py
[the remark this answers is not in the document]
blackboxsw: Ahh yes you are right, we want to walk through all leases and get latest lease, where mtime  is later than the previously found "latest_mtime". I forgot that lease organization may have most recent leases appended to the end of the lease file and we don't want the quick exit.
Thanks for challenging this suggestion.

cloudinit/net/dhcp.py
```diff
@@ -503,62 +504,226 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+        distro.net_ops.link_up(interface)
+
+        # TODO: disabling hooks means we need to get all of the files in
+        # /lib/dhcpcd/dhcpcd-hooks/ and pass each of those with the --nohook
+        # argument to dhcpcd
+        try:
+            out, err = subp.subp(
+                [
+                    self.client_name,
+                    "--oneshot",  # get lease then exit
```
blackboxsw, line 561: @holmanb I think we have a problem with --oneshot on dhcpd, it kills all dhcpcd processes upon success... making the next call to dumplease reject due to no dhcpcd services running:

root@ip-192-168-2-172:~# dhcpcd --oneshot --nobackground --ipv4only --waitip=4 --persistent --noarp --script=/bin/true ens5
dhcpcd-10.0.6 starting
DUID 00:01:00:01:2d:3a:fc:cc:02:f2:c2:1c:ca:1f
ens5: IAID c2:1c:ca:1f
ens5: rebinding lease of 192.168.2.172
ens5: leased 192.168.2.172 for 3600 seconds
ens5: adding route to 192.168.0.0/20
ens5: adding default route via 192.168.0.1
exiting due to oneshot
dhcpcd exited
root@ip-192-168-2-172:~# dhcpcd --dumplease ens5
dhcpcd is not running
root@ip-192-168-2-172:~# echo $?
1

cloudinit/distros/__init__.py
```diff
@@ -274,6 +275,66 @@ def _write_network(self, settings):
...
+        if config_priority:
+            # user or image builder configured a custom dhcp client priority
+            # list
+            found_clients = []
+            LOG.debug(
+                "Using configured dhcp client priority list: %s",
+                config_priority,
+            )
+            for client_configured in config_priority:
+                for client_class in dhcp.ALL_DHCP_CLIENTS:
```
blackboxsw, line 310: This should be the distro-supported list of clients right in case subclasses reduce support from ALL_DHCP_CLIENTS.
```suggestion
                for client_class in self.dhcp_client_priority:
```

holmanb: I could go either way on that one. The distro priority list might set a single expected dhcp client, but if the user gets the binary installed and wants to use it, I think that I'd rather just let them use it if they set that in the configuration file.

blackboxsw: +1 I'm good with that justification.

cloudinit/net/dhcp.py
holmanb: I ended up filling some bugs in upstream dhcpcd and a couple of potential fixes in PRs too. For now we'll have to just use the daemon and make sure that we clean up after ourselves by killing it when we've accomplished what we needed. This requires use of the `-P` flag to get the lease file name, since it isn't necessarily consistent and there is no way to set its location.

holmanb:
Thanks for the review @blackboxsw. I've reworked this PR to not use `--oneshot` so that a daemon gets spawned. This allows us to use `--dumplease <ifname>` which will always reach out to the daemon (unlike stdin parsing, which is broken) to dump the lease. After the lease is dumped, we kill the spawned daemon to make sure we don't leak a dhcpcd daemon to the system. It's a little more messy than using `--oneshot`, but that doesn't appear to currently be an option given the bugs in upstream.

I also added some integration tests to check that dhclient is used on < Noble, dhcpcd is used on Noble (currently a failing test), and that when we set the dhcp priority list to prefer dhcpcd, the correct client gets selected and we get a lease (also that clouds that use the EphemeralDHCPv4 helpers work correctly).

I think this is ready for re-review now.

cloudinit/net/dhcp.py
```diff
@@ -503,62 +540,254 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+            error streams.
+        @param distro: a distro object for network interface manipulation
+        @return: dict of lease options representing the most recent dhcp lease
+            parsed from the dhclient.lease file
+        """
+        LOG.debug("Performing a dhcp discovery on %s", interface)
+        sleep_time = 0.01
+        sleep_cycles = int(self.timeout / sleep_time)
+
+        # ISC dhclient needs the interface up to send initial discovery packets
```
blackboxsw, line 586: Did you mean dhcpcd not ISC dhclient in this comment?

cloudinit/net/dhcp.py
```diff
@@ -503,62 +540,254 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+        # Generally dhclient relies on dhclient-script PREINIT action to bring
+        # the link up before attempting discovery. Since we are using
+        # -sf /bin/true, we need to do that "link up" ourselves first.
+        distro.net_ops.link_up(interface)
+
+        # TODO: disabling hooks means we need to get all of the files in
+        # /lib/dhcpcd/dhcpcd-hooks/ and pass each of those with the --nohook
+        # argument to dhcpcd
```
blackboxsw, line 594: Are you planning on taking nohooks frunctionality to a separate PR?

cloudinit/net/dhcp.py
```diff
@@ -503,62 +540,254 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+
+        @param interface: which interface to dump the lease from
+        @raises: InvalidDHCPLeaseFileError on empty or unparseable leasefile
+            content.
+        """
+        try:
+            return self.parse_dhcpcd_lease(
+                subp.subp(
+                    [
+                        "dhcpcd",
```
blackboxsw, line 715: For our --dumplease command we aren't providing the full path to /usr/sbin/dhcpcd as we are for the initial --waitip command. This could lead unexpected issues in the unlikely event that $PATH contains a different locally installed dhcpcd. Shouldn't we be invoking the full path `/usr/sbin/dhcpcd for all commands to avoid potential corner case behavior?

cloudinit/net/dhcp.py
```diff
@@ -503,62 +540,254 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+                for _ in range(sleep_cycles):
+                    try:
+                        pid_content = util.load_file(pid_file).strip()
+                        pid = int(pid_content)
+                        LOG.debug("killing dhcpcd with pid=%s", pid)
+                        os.kill(pid, signal.SIGKILL)
+                    except FileNotFoundError:
+                        debug_msg = (
+                            f"No PID file found at {pid_file}, "
+                            "dhcpcd is still running"
```
blackboxsw, line 629: We are running into this condition on ec2 where we loop 1000 times over 10 seconds before finally giving up and leaving dhcpcd running.

blackboxsw (commented):
Something that needs more digging into is how we are handling the cleanup of dhcpcd. The looping we have isn't obtaining a valid pid file with actionable information within 10 seconds and eventually falls through leaving the dhcpcd running on the system

cloudinit/net/dhcp.py
holmanb: The `/bin/true` handles this behavior, I'll drop this comment.

cloudinit/net/dhcp.py
holmanb: oops, yes thank you

cloudinit/net/dhcp.py
holmanb: yes

cloudinit/net/dhcp.py
holmanb: Hrm, I saw this previously, but that was before I had the `sleep()` call there - I haven't seen this yet since adding it in. Back when I was debugging that I noticed that child PID gets printed to standard out, and had written some code that parsed and killed the child process using that. I just pushed that code, as well as changes to address your previous comments.

holmanb: Since parsing the lease file from stdin is [currently unreliable](https://github.com/NetworkConfiguration/dhcpcd/issues/285), and `dhcpcd --dumplease -4 <ifname>`  expects to be able to talk to a running daemon, and dumping all leases also expects to talk to a running daemon (and [is unreliable](https://github.com/NetworkConfiguration/dhcpcd/issues/287)), I think that we're stuck trying to use a daemon and clean it up for now. Once https://github.com/NetworkConfiguration/dhcpcd/issues/285 is fixed we can change that.

In the meantime, the only alternative that I've thought of involves the `ptrace` syscall. I want to leave the direct use of the syscall out of scope, since that would require architecture-dependent (and probably kernel-dependent) code via ctypes, but perhaps calling `dhcpcd` under `strace --follow-forks --trace=process <dhcpcd call>` or `strace --follow-forks -e trace="clone,fork,vfork"` would be an acceptable stopgap until https://github.com/NetworkConfiguration/dhcpcd/issues/287 is fixed. Unfortunately `strace` isn't going to be available everywhere, so this would have to be optional depending on its availability. It might raise some eyebrows, but it would probably "work".

holmanb: Example of using strace to get PID.
```
sudo strace --daemonize=grandchild --follow-forks --trace=clone,fork,vfork,clone --signal='!all' dhcpcd --ipv4only --waitip=4 --persistent --noarp enp24s0
strace: Process 133407 attached
dhcpcd-9.4.1 starting
clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133412 attached
, child_tidptr=0x7f73cef02a10) = 133412
[pid 133412] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133413 attached
, child_tidptr=0x7f73cef02a10) = 133413
[pid 133412] +++ exited with 0 +++
[pid 133413] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7f73cef02a10) = 133414
strace: Process 133414 attached
[pid 133413] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133415 attached
, child_tidptr=0x7f73cef02a10) = 133415
DUID 00:04:00:00:00:00:00:00:00:00:00:00:30:9c:23:e6:be:f0
strace: Process 133416 attached
[pid 133416] +++ exited with 0 +++
strace: Process 133417 attached
[pid 133417] +++ exited with 0 +++
enp24s0: IAID 23:e6:be:f0
enp24s0: rebinding lease of 192.168.0.104
[pid 133414] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133418 attached
, child_tidptr=0x7f73cef02a10) = 133418
[pid 133414] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133419 attached
, child_tidptr=0x7f73cef02a10) = 133419
enp24s0: leased 192.168.0.104 for 86400 seconds
enp24s0: adding route to 192.168.0.0/24
enp24s0: adding default route via 192.168.0.1
strace: Process 133420 attached
[pid 133420] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133421 attached
, child_tidptr=0x7f3d38fa2a10) = 133421
[pid 133421] +++ exited with 0 +++
[pid 133420] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133422 attached
, child_tidptr=0x7f3d38fa2a10) = 133422
[pid 133422] +++ exited with 0 +++
[pid 133420] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133423 attached
, child_tidptr=0x7f3d38fa2a10) = 133423
[pid 133420] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133424 attached
, child_tidptr=0x7f3d38fa2a10) = 133424
[pid 133423] +++ exited with 0 +++
Dropped protocol specifier '.dhcp' from 'enp24s0.dhcp'. Using 'enp24s0' (ifindex=2).
[pid 133424] +++ exited with 0 +++
[pid 133420] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLDstrace: Process 133427 attached
, child_tidptr=0x7f3d38fa2a10) = 133427
[pid 133427] +++ exited with 0 +++
[pid 133420] +++ exited with 0 +++
forked to background, child pid 133413
[pid 133414] clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD <unfinished ...>
[pid 133407] +++ exited with 0 +++
```

holmanb: This raises a good point however that we should be killing all of the processes, not just one of them. Killing by process group would accomplish that. See the latest commit pushed @blackboxsw.

cloudinit/net/dhcp.py
```diff
@@ -503,62 +551,275 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+                            )
+                        except ProcessLookupError:
+                            LOG.debug(
+                                "Couldn't kill process, it already exited"
+                            )
+                LOG.debug("PID not in stdout, waiting for PID file")
+
+                # Note: the pid file location depends on the arguments passed
+                # it can be discovered with the -P flag
+                pid_file = subp.subp([*command, "-P"]).stdout
```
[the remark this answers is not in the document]
blackboxsw: Also we need to `strip()` this response otherwise we are always hitting issues with `load_file(/run/dhcpcd/*pid\n")` being unable to find the file. This is what resulted in the 1000 retries trying to determine pids on ec2 for me.

cloudinit/net/dhcp.py
blackboxsw: Some points on your ocmments above:
1. > I think that we're stuck trying to use a daemon and clean it up for now."

 Agreed, because of the former bugs, we should spawn and cleanup dhcpcd proceses. Please link in a comment relative to the dhcpcd command the upstream bugs prompting this behavior so we can track the feature/fix in the future and re-evaluate our solution

2. Agreed to stay away from strrace to determine PIDs as this is highly environment dependent and likely fragile. 

3. > In the meantime, the only alternative that I've thought of involves the ptrace syscall. I want to leave the direct use of the syscall out of scope, since that would require architecture-dependent

Why do we need ptrace at all. We can/do obtain the ppid of the dhcpcd process from /run/dhcpcd/*.pid properly which allows us to directly kill that process and any children with a `os.kill(<pid_from_dhcpcd_pid_file>)`


Given the `stdout.strip()` fix to the load_file(pid_file). It looks like the code you have performs the right operation and could kill the dhcpcd parent pid cleanly without all the /proc/<PID>/stat lookup functions.

blackboxsw (changes requested):
I think this approach may not want the gpid lookup right? I think we should be able to kill the parent process id stored in `/run/dhcpcd/enX0-4.pid` without the extra groupid lookup right?

If we do need to worry about process group for some reason , then we need to use `os.killpg(gid...)` instead of  `os.kill(gid...)`


Here's utilzing the os.kill of the pid from /run/dhcpcd/*pid that seems the clean things up on ec2:
```diff
diff --git a/cloudinit/net/dhcp.py b/cloudinit/net/dhcp.py
index c68a5743a..526302cdf 100644
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -642,19 +642,16 @@ class Dhcpcd(DhcpClient):
 
                 # Note: the pid file location depends on the arguments passed
                 # it can be discovered with the -P flag
-                pid_file = subp.subp([*command, "-P"]).stdout
+                pid_file = subp.subp([*command, "-P"]).stdout.strip()
                 pid_content = None
                 debug_msg = ""
                 for _ in range(sleep_cycles):
                     try:
                         pid_content = util.load_file(pid_file).strip()
                         pid = int(pid_content)
-                        gid = distro.get_proc_pgid(pid)
-                        if gid:
-                            LOG.debug(
-                                "killing dhcpcd with pid=%s gid=%s", pid, gid
-                            )
-                            os.kill(gid, signal.SIGKILL)
+                        if pid:
+                            LOG.debug("killing dhcpcd with pid=%s", pid)
+                            os.kill(pid, signal.SIGKILL)
                     except FileNotFoundError:
                         debug_msg = (
                             f"No PID file found at {pid_file}, "
```

blackboxsw:
Also side-note the concerns about looping and blocking for 10 seconds on dhcpcd in early boot are an irrelevant/incorrect assumption on my part. The time cost of 10 seconds wasn't due to dhcpcd not receiving a lease it could parse, or a pid file that was readable, it was only in our incorrect looping trying to retrieve a correctly named /run/dhcpcd/*pid file. In looking over timestamps the /run/dhcpcd/*pid file is created within 5 hundredths of a second of our first call to dhcpcd. Overall, the process of receiving a lease via the dhcpcd request looks to be < 2 seconds on average, and a lot of that time is just waiting on the control plane  in the cloud to respond to that dhcp request.

cloudinit/net/dhcp.py
```diff
@@ -503,62 +551,275 @@ def get_latest_lease(lease_dir, lease_file_regex):
...
+            if dhcp_log_func is not None:
+                dhcp_log_func(out, err)
+            lease = self.get_newest_lease(interface)
+            # Attempt cleanup and leave breadcrumbs if it fails, but return
+            # the lease regardless of failure to clean up dhcpcd.
+            if lease:
+                # pid file arrival can take very long, and stdout might contain
+                # the PID, so attempt to use a stdout parsed PID
+                for line in reversed(out.split("\n")):
+                    if "forked to background, child pid " in line:
```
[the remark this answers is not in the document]
holmanb: > What is the downside of using a more foolproof method to determine the dhcpcd PID by reading the correct /run/dhcpcd/*pid file? 

I revived this code because you reported that it was taking >10s for cloud-init to get its lease - this was an attempt to make timeout faster as a stopgap. Given that you've identified the root cause I agree, we can toss this logic.

cloudinit/net/dhcp.py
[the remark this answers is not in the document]
holmanb: > subp shell out is a bit heavier an operation that just glob("/run/dhcpcd/<interface_name>.pid")

Also I'm not convinced that such a glob would actually be faster - when this command runs, dhcpcd is already in page cache and just requires a fork/exec (of the already-cached binary) then a few instructions. A glob would require `stat`ing the directory, and if the directory isn't already in the dcache then it probably means a round trip to disk and back. I agree that subp is heavy-weight - but I wouldn't assume that it's faster than globbing directories in the case that the binary is already in memory and only a couple of syscalls away from a returned result.

> /run/dhcpcd/<interface_name>.pid

I didn't document it, but I'm pretty sure I saw dhcpcd put the lease in a different directory or filename before. That's why I did this at all, given the manpage documentation of the lease location. I was curious about the existence of `-P` at all - until I saw that happen.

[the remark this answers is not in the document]
holmanb: > Also we need to strip() this response otherwise we are always hitting issues with load_file(/run/dhcpcd/*pid\n") being unable to find the file. This is what resulted in the 1000 retries trying to determine pids on ec2 for me.

Thanks for finding that. I did test that this worked - the original `strip()` must have gotten lost when I rebased for the review.

holmanb:
> Some points on your comments above:
>
>        I think that we're stuck trying to use a daemon and clean it up for now."
>
> Agreed, because of the former bugs, we should spawn and cleanup dhcpcd processes. Please link > in a comment relative to the dhcpcd command the upstream bugs prompting this behavior so we
> can track the feature/fix in the future and re-evaluate our solution

+1 Will do

>    Agreed to stay away from strace to determine PIDs as this is highly environment dependent 
> and likely fragile.

It is actually pretty broadly supported - I bet it's less fragile than you think.

>>        In the meantime, the only alternative that I've thought of involves the ptrace 
>> syscall. I want to leave the direct use of the syscall out of scope, since that would require 
> >architecture-dependent
>
> Why do we need ptrace at all.

I was only proposing that this syscall (used via strace) was likely less fragile and more portable and stable than parsing stdout (which itself was revived in response to the 10s timeout report).

> We can/do obtain the ppid of the dhcpcd process from /run/
> dhcpcd/*.pid properly which allows us to directly kill that process and any children with a 
> os.kill(<pid_from_dhcpcd_pid_file>)

That's not how signal propagation works on Posix-like systems (i.e. linux or unix). To kill the other process we want to send the signal to the process group.

> Given the stdout.strip() fix to the load_file(pid_file). It looks like the code you have 
> performs the right operation

Ack, sorry for the mixup there - when I was prepping this for review I think the `strip()` got lost in the rebase (removal of unwanted code after testing) - I should have tested again before requesting review.

>  and could kill the dhcpcd parent pid cleanly without all the /
> proc//stat lookup functions.

I'm strongly against just hoping that since killing the parent pid seems to work on Ubuntu/ec2 that this is going to work everywhere. Signal propagation is literally what process groups exist for. We should use it as it is designed to be used. Signals don't propagate magically from parent to child. Interactive terminals implement special behavior that I assume you're drawing from (for job control) - but that's not how signals generically work.

I agree that I see the same behavior on Ec2, but we need to be sure that we are taking responsibility for cleaning the processes that we start rather than hoping that since it seems to work on one distro on ec2 it will work everywhere.

If I test killing just the parent PID locally it - as expected - doesn't bring down the child or grandchild processes. I haven't dug into why it behaves that way on ec2, but since I have a counterexample of that behavior not working I don't plan on digging into that further.

> If we do need to worry about process group for some reason , then we need to use os.killpg(gid...) instead of os.kill(gid...)

+1, yes agreed `os.killpg()` is the way to go

holmanb:
@blackboxsw a quick experiment will killing process groups vs just the parent PID.
```python
$ PYTHONPATH=. sudo /usr/bin/python3
>>> import os
>>> from cloudinit.subp import subp
>>> from cloudinit.distros import Distro
>>> print('\n'.join(subp(['pgrep', '-a', 'dhcpcd'], rcs=[0,1]).stdout.split('\n')))

>>> print('\n'.join(subp(['dhcpcd', '--ipv4only', '--waitip=4', '--persistent', '--noarp', 'enp24s0']).stderr.split('\n')))
dhcpcd-9.4.1 starting
DUID 00:04:00:00:00:00:00:00:00:00:00:00:30:9c:23:e6:be:f0
enp24s0: IAID 23:e6:be:f0
enp24s0: rebinding lease of 192.168.0.104
enp24s0: leased 192.168.0.104 for 86400 seconds
enp24s0: adding route to 192.168.0.0/24
enp24s0: adding default route via 192.168.0.1
Dropped protocol specifier '.dhcp' from 'enp24s0.dhcp'. Using 'enp24s0' (ifindex=2).
forked to background, child pid 370956

>>> pid_file = subp(['dhcpcd', '--ipv4only', '--waitip=4', '--persistent', '--noarp', 'enp24s0', '-P']).stdout.strip()
>>> pid = util.load_file(pid_file).strip()
>>> print('\n'.join(subp(['pgrep', '-a', 'dhcpcd'], rcs=[0,1]).stdout.split('\n')))
370956 dhcpcd: enp24s0 [ip4]
370957 dhcpcd: [privileged proxy] enp24s0 [ip4]
370958 dhcpcd: [control proxy] enp24s0 [ip4]
370961 dhcpcd: [BPF ARP] enp24s0 192.168.0.104
370971 dhcpcd: [network proxy] 192.168.0.104

>>> pgid = Distro.get_proc_pgid(pid)
>>> os.killpg(pgid)
>>> os.killpg(pgid, 9)
>>> print('\n'.join(subp(['pgrep', '-a', 'dhcpcd'], rcs=[0,1]).stdout.split('\n')))

>>> print('\n'.join(subp(['dhcpcd', '--ipv4only', '--waitip=4', '--persistent', '--noarp', 'enp24s0']).stderr.split('\n')))
dhcpcd-9.4.1 starting
DUID 00:04:00:00:00:00:00:00:00:00:00:00:30:9c:23:e6:be:f0
enp24s0: IAID 23:e6:be:f0
enp24s0: rebinding lease of 192.168.0.104
enp24s0: leased 192.168.0.104 for 86400 seconds
enp24s0: adding route to 192.168.0.0/24
enp24s0: adding default route via 192.168.0.1
Dropped protocol specifier '.dhcp' from 'enp24s0.dhcp'. Using 'enp24s0' (ifindex=2).
forked to background, child pid 371149

>>> print('\n'.join(subp(['pgrep', '-a', 'dhcpcd'], rcs=[0,1]).stdout.split('\n')))
371149 dhcpcd: enp24s0 [ip4]
371150 dhcpcd: [privileged proxy] enp24s0 [ip4]
371151 dhcpcd: [control proxy] enp24s0 [ip4]
371154 dhcpcd: [BPF ARP] enp24s0 192.168.0.104
371164 dhcpcd: [network proxy] 192.168.0.104

>>> pid_file = subp(['dhcpcd', '--ipv4only', '--waitip=4', '--persistent', '--noarp', 'enp24s0', '-P']).stdout.strip()
>>> pid = util.load_file(pid_file).strip()
>>> os.kill(int(pid), 9)
>>> print('\n'.join(subp(['pgrep', '-a', 'dhcpcd'], rcs=[0,1]).stdout.split('\n')))
371150 dhcpcd: [privileged proxy] enp24s0 [ip4]
371154 dhcpcd: [BPF ARP] enp24s0 192.168.0.104
371164 dhcpcd: [network proxy] 192.168.0.104

```

holmanb:
@blackboxsw If you agree that killing by process group is the right path forward, then I've just pushed a commit that's ready for review. It passes integration tests on AWS, and addresses the remaining comments. I dropped stdout parsing since using the PID file works fine, and it seems that we're on the same page about not needing other alternatives.

If you don't mind giving it another review, I would appreciate it. I've tested the latest commit against Noble/Ec2 and I think it's ready. I'm also ready to queue up PRs for InfiniBand parsing and option parsing support - but I want this to land first before doing that.

blackboxsw (approved):
Thank you for this fix, feature and all the work you put into upsream bug analysis and filing @holmanb. Let's ship it.

> I'm strongly against just hoping that since killing the parent pid seems to work on Ubuntu/ec2 that this is going to work everywhere. Signal propagation is literally what process groups exist for. We should use it as it is designed to be used. Signals don't propagate magically from parent to child. Interactive terminals implement special behavior that I assume you're drawing from (for job control) - but that's not how signals generically work.

Thanks for the discussion, correction and the legwork to establish failure patterns without grp id. I agree with your justification that `os.killpg` is the only correct approach. I'm not certain why I was seeing that the process and all children appeared to be brought down either by just killing the parent PID represented by the `/run/dhcpd/*.pid` file in python via an `os.kill` on Ec2. But, let's do the right thing here as you suggest and not base the solution on an assumption that will not work everywhere.


Some things to address in later PRs: 
- After Infiniband followup  and Azure option 245, we will need a separate PR to prioritize Dchpcd first in ALL_DHCP_CLIENTS in tip of main. (This might have been part of what your checklist mentions as "packaging branches")

PR #4747, opened 2024-01-08
Ubuntu/mantic 23.4.x: Retain exit code in cloud-init status for recoverable errors

aciba90:
## do not squash

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,10 +1,10 @@
-cloud-init (23.4.1-0ubuntu1~23.10.2) UNRELEASED; urgency=medium
+cloud-init (23.4.1-0ubuntu1~23.10.2) mantic; urgency=medium
 
   * d/p/status-retain-recoverable-error-exit-code.patch:
     Retain exit code in cloud-init status for recoverable errors.
     (LP: #2048522).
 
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,3 +1,11 @@
+cloud-init (23.4.1-0ubuntu1~23.10.1ubuntu1) UNRELEASED; urgency=medium
```
TheRealFalcon, line 1: This should be `23.10.2` rather than `1ubuntu1`. File a bug against new_upstream_snapshot.py if we need one.

TheRealFalcon (approved):
Just need to update the version number, but otherwise LGTM.

Lets wait for standup before merging though.

debian/changelog
blackboxsw: Also please bump `UNRELEASED` to `mantic` so we can release this

blackboxsw (commented):
Also right after we upload this let's get PRs against the base branches `upstream/ubuntu/(mantic|lunar|jammy|focal)` the sync both d/changelog and the new quilt patch and series updates into the base branches so we don't miss it on next  SRU

PR #4748, opened 2024-01-08
feat: Add v2 route mtu rendering to NetworkManager

TheRealFalcon:
## Proposed Commit Message

```
feat: Add v2 route mtu rendering to NetworkManager
```

## Merge type

### DIFF
--- a/cloudinit/net/network_manager.py
+++ b/cloudinit/net/network_manager.py
@@ -205,12 +205,13 @@ class NMConnection:
 
         If the line already exists, then append the new key/value pair
         """
-        if not self.config[section].get(f"{route}_options"):
-            self.config[section][f"{route}_options"] = f"{key}={value}"
-        else:
-            self.config[section][
-                f"{route}_options"
-            ] = f'{self.config[section][f"{route}_options"]},{key}={value}'
+        numbered_key = f"{route}_options"
+        route_options = self.config[section].get(numbered_key)
+        self.config[section][numbered_key] = (
+            f"{route_options},{key}={value}"
+            if route_options
+            else f"{key}={value}"
+        )
 
     def _add_address(self, family, subnet):
         """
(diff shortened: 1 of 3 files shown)


cloudinit/net/network_manager.py
```diff
@@ -173,18 +173,44 @@ def _set_ip_method(self, family, subnet_type):
...
+          route: route0
+          key: mtu
+          value: 500
+
+        Create line under [ipv4] section:
+            route0_options=mtu=500
+
+        If the line already exists, then append the new key/value pair
+        """
+        if not self.config[section].get(f"{route}_options"):
```
holmanb, line 208: We re-create the same string and reference the same dict location multiple times here - and also use nested f strings - a bit messy to grok. Assigning a couple of variables would make this a little easier to read.

compare this:
```python
        if not self.config[section].get(f"{route}_options"):
            self.config[section][f"{route}_options"] = f"{key}={value}"
        else:
            self.config[section][
                f"{route}_options"
            ] = f'{self.config[section][f"{route}_options"]},{key}={value}'
```

with this:

```python
        option_number = f"{route}_options"
        route_options = self.config[section].get(option_number)
        if not route_options:
            self.config[section][option_number] = f"{key}={value}"
        else:
            self.config[section][
                option_number
            ] = f'{route_options},{key}={value}'
```

cloudinit/net/network_manager.py
```diff
@@ -173,18 +173,44 @@ def _set_ip_method(self, family, subnet_type):
...
+
+        Example:
+        Given:
+          section: ipv4
+          route: route0
+          key: mtu
+          value: 500
+
+        Create line under [ipv4] section:
+            route0_options=mtu=500
```
holmanb, line 204: I had to dig a little to find docs that reference this syntax, but [I found them](https://networkmanager.pages.freedesktop.org/NetworkManager/NetworkManager/nm-settings-keyfile.html). Do you think we should add a link to the function docstring above?

holmanb (approved):
Thanks @TheRealFalcon, this is a great addition. I have a couple of nitpicks on the code, but feel free to take what you want and ignore the rest.

cloudinit/net/network_manager.py
TheRealFalcon: Most of this file is for generating these keyfiles. I'll add a more general reference, but I think it'd be odd to put it on this function specifically.

PR #4754, opened 2024-01-08
Ubuntu/jammy

aciba90:
## do not squash


blackboxsw (approved):
Sync of jammy-23.4.x to base downstream ubuntu/jammy looks good so we don't miss quilt patch changes on next SRU

PR #4755, opened 2024-01-08
Ubuntu/focal

aciba90:
## do not squash


blackboxsw (approved):
Sync of focal-23.4.x to base downstream ubuntu/focal looks good so we don't miss quilt patch changes on next SRU

PR #4756, opened 2024-01-08
feat: Ensure internal DNS state same for v1 and v2

TheRealFalcon:
When defining interface-level DNS on the network state, v1 uses the "nameservers" key whereas v2 was using an "addresses" key. This commit updates the v2 parsing to set the key as "nameservers".

Also update networkd renderer as this was only renderer using the v2 "addresses" key.

## Proposed Commit Message

```
feat: Ensure internal DNS state same for v1 and v2

When defining interface-level DNS on the network state, v1 uses the
"nameservers" key whereas v2 was using an "addresses" key.
This commit updates the v2 parsing to set the key as "nameservers".

Also update networkd renderer as this was only renderer using the v2
"addresses" key.
```

## Merge type

### DIFF
--- a/cloudinit/net/network_state.py
+++ b/cloudinit/net/network_state.py
@@ -336,7 +336,7 @@ class NetworkStateInterpreter:
             if iface:
                 nameservers, search = dns
                 iface["dns"] = {
-                    "addresses": nameservers,
+                    "nameservers": nameservers,
                     "search": search,
                 }
 
--- a/cloudinit/net/networkd.py
+++ b/cloudinit/net/networkd.py
@@ -221,12 +221,6 @@ class Renderer(renderer.Renderer):
     def parse_dns(self, iface, cfg: CfgParser, ns: NetworkState):
         sec = "Network"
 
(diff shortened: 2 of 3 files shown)


holmanb (approved):
Nice refactor. LGTM, thanks!

CalvoM (approved):

ani-sinha:
@TheRealFalcon It seems the DNS and search information is not getting added to ifcfg files with a config like the following:
```  network:
              config:
                - name: eth0
                  subnets:
                    - address: 10.0.2.2/24
                      gateway: 10.0.2.1
                      type: static
                  type: physical
                - type: nameserver
                  interface: eth0
                  address:
                    - 8.8.8.8
                    - 4.4.4.4
                  search:
                    - example.com
              version: 1
```

Does the sysconfig renderer need to be updated as well?

TheRealFalcon:
@ani-sinha , I saw that you filed https://github.com/canonical/cloud-init/issues/5845 and it looks like the issue is unrelated to this PR, so let's keep the conversation there.

PR #4761, opened 2024-01-10
fix(clean): stop warning when running clean command

### LINKED ISSUE #4760: clean: Noble raises warning about README in /etc/cloud/clean.d
# Bug report

on Noble, tested on lxd containers, when I run:
```
cloud-init clean
```
I get the following warnig:
```
2024-01-10 10:48:52,127 - subp.py[WARNING]: skipping /etc/cloud/clean.d/README as its not executable or the underlying file system is mounted without executable permissions.
```
This is caused by [runparts](https://github.com/canonical/cloud-init/blob/b729a4c4aa32d2bdfdd84fb24b9be5fc9c1eb293/cloudinit/subp.py#L336) running on /etc/cloud/clean.d/ and expecting every file in the directory being executable. However, we have a README file in there.
## Environment details
- Cloud-init version: 24.1~1g1003d438-0ubuntu1
- Operating System Distribution: Noble 
- Cloud provider, platform or installer type: lxc container

CalvoM:
When the clean command is run, runparts is called and README in /etc/cloud/clean.d is run, which leads to a warning.

## Proposed Commit Message

```
fix(clean): stop warning when running clean command

When the clean command is run, runparts is called and README in /etc/cloud/clean.d
 is run, which leads to a warning.

Fixes GH-4760
```
Fixes #4760

## Test Steps

Run the clean command on a noble machine:
```
cloud-init clean
```

## Checklist

## Merge type


TheRealFalcon:
Warning was added in #4506 . LGTM, but should probably get an approval from @blackboxsw since this was somewhat recently added.

holmanb (changes requested):
We will also need to update the following packaging-related files which reference this file & directory.

- packages/suse/cloud-init.spec.in
- packages/redhat/cloud-init.spec.in
- setup.py

I didn't see this file in the ubuntu/devel branch, so I think that `setup.py` installs this for package managers.

setup.py
```diff
@@ -274,7 +274,6 @@ def finalize_options(self):
 
 data_files = [
     (ETC + "/cloud", [render_tmpl("config/cloud.cfg.tmpl", is_yaml=True)]),
-    (ETC + "/cloud/clean.d", glob("config/clean.d/*")),
```
blackboxsw, line 277: The result of no longer delivering this FILE/directory will break external consumers like subiquity which expect the /etc/cloud/clean.d directory to exist. [Subiquity injects supplemental clean.d files](https://github.com/canonical/subiquity/blob/main/subiquity/models/subiquity.py#L453) so when cloud-init is cleaned from live desktop/serve deployed images, the datasource_list is reset to defaults and cloud-init discovery is "re-enabled". We either need to either:
 1. ensure this empty /etc/cloud/clean.d directory still exists
 2. or continue to deliver the /etc/cloud/clean.d/README but provide a param to runparts to be silent on non_exe files, or allow for providing a ignore_files list param which which skip the expected README

blackboxsw (changes requested):
Thanks @CalvoM. I like the doc addition, and we should keep it.

But, this will break subiquity if we don't have an `/etc/cloud/clean.d` directory, and I find that the README context could be helpful for those who stumble upon the subdir as they are poking around cloud-init config files on the system,

Can we find a way to either continue to deliver /etc/cloud/clean.d directory as empty. Or ideally continue to deliver ther README there without raising a warning by extending our sub.runparts to better handle expected non_exec files?

setup.py
TheRealFalcon: Originally this PR just removed the README file. Brett mentioned `setup.py` in his review, but I don't think anything actually needs to be changed, so we should just remove the change on this line.

CalvoM: @blackboxsw I think option 1 is better; any documentation, e.g. README should be on our documentation.

blackboxsw: Works from me since we now have top-level docs on ReadTheDocs for this. thanks

blackboxsw (approved):
LGTM @CalvoM.

-- Correction on versions below. rm_conffile should be referencing the first version that we know drops the README, not the previous version currently released. The diminishing `~` at the end handles matching all version less that the version known to drop the conf file.

Now, to avoid WARNINGS from apt across upgrades we'll need to update the debian/mainstcripts for each series to express that any released package version before our next release has dropped the README file otherwise the package upgrade path will leave the README file in the directory and we'll still get warnings from `cloudinit.runparts`.


To fix the package creation to "delete" config files across upgrade path: add a line to each debian/cloud-init.maintscript file which says when previous package is less than `<next_release_number>~`, ensure to delete /etc/cloud/clean.d/README.

This looks something like the following diff against each downstream release branch:

```diff
--- a/debian/cloud-init.maintscript
+++ b/debian/cloud-init.maintscript
@@ -7,3 +7,4 @@ rm_conffile /etc/init/cloud-init-nonet.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-log-shutdown.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3~4ga31370af-0ubuntu1~
+rm_conffile /etc/cloud/clean.d/README 24.1~4~
```

The above is a diff against https://github.com/canonical/cloud-init/tree/ubuntu/devel
which says if we are upgrading from a version of cloud-init that is less than 24.1~4, then make sure to delete README prior to installing this new version.

The result is seeing a log like the following:
```
Removing obsolete conffile /etc/cloud/clean.d/README ...
```


~~We need to obtain the correct released version number for each series from `rmadison cloud-init` and provide that  the the series-specific `upstream/ubuntu/(focal,jammy,lunar,mantic)` branches~~`

CORRECTION: since dropping the README file will only be introduced to a stable release on our next SRU, we can simplify our rm_conffile version to `24.1-0ubuntu0~XX.YY.1~` where XX.YY is the stable release major.minor indicating our first potential SRU of 24.1 to that stable release.

TheRealFalcon:
@blackboxsw waaaait, why is apt trying to delete `clean.d`? We want the directory to still be there.

CalvoM:
@TheRealFalcon I think he means deleting the README file in the `/etc/cloud/clean.d/` during the upgrade of cloud-init on a machine with README in directory already

blackboxsw:
> @blackboxsw waaaait, why is apt trying to delete `clean.d`? We want the directory to still be there.

Right, @TheRealFalcon @CalvoM sorry that statement I made above is incorrect regarding the warning message. You'd only see that warning message about directory not empty if our deb no longer included  /etc/cloud/cloud.d directory in the binary package and if we also didn't rm_conffile in the maintscript. But, since we've reverted the setup.py changes, and continue to the deliver the empty /etc/cloud/cliean.d directory we won't see that warning. I've updated the comment above with version recommendations based on testing this again. https://github.com/canonical/cloud-init/pull/4761#pullrequestreview-1822306353 and reflected those same sentiments to Calvin's downstream branches.

Issue #4732, opened 2024-01-03
[enhancement]: List `eject` as a suggested/recommended dependency for the deb


Monochromics:
# Enhancement/Bug

The `eject` package is used in the azure datasource: https://github.com/canonical/cloud-init/blob/cb36bf38b823f811a3e938ccffc03d7d13190095/cloudinit/sources/helpers/azure.py#L790

`eject` is not listed as a dependency or a recommendation/suggestion of the deb.  It look's like this is the only datasource that uses this, so listing as a full dep might be a bit heavy handed. This isn't typically and issue, as most images already include eject, however ubuntu minimal images susceptible. 

I'd also argue that the current handling of the missing package is ungraceful (even if mostly benign).

```
2024-01-03 19:58:43,765 - subp.py[DEBUG]: Running command ['eject', '/dev/sr0'] with allowed return codes [0] (shell=False, capture=True)
2024-01-03 19:58:43,767 - azure.py[DEBUG]: Failed ejecting the provisioning iso: Unexpected error while running command.
```

j5awry:
I think we need to gather info about the availability of the command for all supported distros

1. is the command guaranteed to be installed?
2. if so, how is it guaranteed (for instance, in Ubuntu, it'd be in a seed file _or_ as a package dependency)
3. it not, how do we solve this cross-distro?

For more background, we're hitting this in Ubuntu because, previously, `eject` was being supplied as a `recommends` of another package. in Ubuntu Minimal, we no longer install any `recommended` packages. We hit this with other `cloud-init` dependencies while working on the new Minimal seeds, but this is the first time we're seeing a requirement that's in a single cloud. for dependencies used across clouds, Ubuntu has added packages to our Minimal seed as required. But we're hesitant to include a package for a _single_ cloud in a definition of a base image. We're not against adding it directly to our Azure image build _but_, before we take that route, we want to ensure we're not the only ones hitting the issue. 

tl:dr -- if it's just Ubuntu, we can add it directly to our image builds. If it may affect other distros as well, I think we should find a different solution than all of us adding it to our Azure builds.

catmsred:
Since there is no way to restrict dependencies (or recommends/suggests) to a single cloud, would it make sense to include the dependency as a commented out suggestion in the debian packaging?  That would make it clear to anyone experiencing difficulty getting cloud-init to run on Azure what they need to do to fix the issue but also not add bloat to cloud-init installs in non-Azure environment.

j5awry:
[...]

PR #4769, opened 2024-01-11
feat: prefer udev's cdrom_id -e to eject

catmsred:
Azure currently relies on eject for some use cases.  In noble minimal images eject is not installed by default.  This change prefers a udev command that performs the same action before trying eject.

A integration test has been created but is not functional.  Code changes have not been tested.

[1] https://github.com/canonical/cloud-init/issues/4732

## Proposed Commit Message
```
feat: fall back to cdrom_id eject if eject is not available

Azure currently relies on eject for some use cases.  In ubuntu noble
minimal images eject is not installed by default.  This change adds a
fallback of systemd udev's cdrom -e if eject is not found on the distro.

Fixes GH-4732
```

## Test Steps
Completely untested at this time.  @TheRealFalcon offered to work on an integration test for this.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -1316,10 +1316,19 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
     @staticmethod
     def eject_media(device: str) -> None:
-        cmd = ["/lib/udev/cdrom_id", "--eject-media", device]
-        if not uses_systemd():
+        cmd = None
+        if subp.which("eject"):
             cmd = ["eject", device]
+        elif subp.which("/lib/udev/cdrom_id"):
+            cmd = ["/lib/udev/cdrom_id", "--eject-media", device]
+        else:
+            raise subp.ProcessExecutionError(
+                cmd="eject_media_cmd",
+                description="eject command not found",
+                reason="neither eject nor /lib/udev/cdrom_id are found",
+            )
         subp.subp(cmd)
+
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
     """
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -1314,6 +1314,12 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             )
         return None
 
+    @staticmethod
+    def eject_media(device: str) -> None:
+        cmd = ["/lib/udev/cdrom_id", "--eject-media", device]
+        if not uses_systemd():
+            cmd = ["eject", device]
+        subp.subp(cmd)
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
     """
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -719,10 +719,10 @@ class WALinuxAgentShim:
             self.openssl_manager.clean_up()
 
     @azure_ds_telemetry_reporter
-    def eject_iso(self, iso_dev) -> None:
+    def eject_iso(self, iso_dev, distro: distros.Distro) -> None:
+        LOG.debug("Ejecting the provisioning iso")
         try:
-            LOG.debug("Ejecting the provisioning iso")
-            subp.subp(["eject", iso_dev])
+            distro.eject_media(iso_dev)
         except Exception as e:
             report_diagnostic_event(
                 "Failed ejecting the provisioning iso: %s" % e,
@@ -731,7 +731,7 @@ class WALinuxAgentShim:
 
     @azure_ds_telemetry_reporter
     def register_with_azure_and_fetch_data(
-        self, pubkey_info=None, iso_dev=None
+        self, distro: distros.Distro, pubkey_info=None, iso_dev=None
     ) -> Optional[List[str]]:
         """Gets the VM's GoalState from Azure, uses the GoalState information
         to report ready/send the ready signal/provisioning complete signal to
@@ -762,7 +762,7 @@ class WALinuxAgentShim:
         )
 
         if iso_dev is not None:
-            self.eject_iso(iso_dev)
+            self.eject_iso(iso_dev, distro=distro)
 
         health_reporter.send_ready_signal()
         return ssh_keys
@@ -940,13 +940,14 @@ class WALinuxAgentShim:
 @azure_ds_telemetry_reporter
 def get_metadata_from_fabric(
     endpoint: str,
+    distro: distros.Distro,
     pubkey_info: Optional[List[str]] = None,
     iso_dev: Optional[str] = None,
 ):
     shim = WALinuxAgentShim(endpoint=endpoint)
     try:
         return shim.register_with_azure_and_fetch_data(
-            pubkey_info=pubkey_info, iso_dev=iso_dev
+            distro=distro, pubkey_info=pubkey_info, iso_dev=iso_dev
         )
     finally:
         shim.clean_up()
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1366,6 +1366,7 @@ class DataSourceAzure(sources.DataSource):
         try:
             data = get_metadata_from_fabric(
                 endpoint=self._wireserver_endpoint,
+                distro=self.distro,
                 iso_dev=self._iso_dev,
                 pubkey_info=pubkey_info,
             )
--- a/tests/integration_tests/datasources/test_azure.py
+++ b/tests/integration_tests/datasources/test_azure.py
@@ -1,0 +1,51 @@
+import pytest
+from pycloudlib.cloud import ImageType
+
+from tests.integration_tests.clouds import IntegrationCloud
+from tests.integration_tests.conftest import get_validated_source
+from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.integration_settings import PLATFORM
+from tests.integration_tests.releases import CURRENT_RELEASE
+from tests.integration_tests.util import verify_clean_log
+
+
+def _check_for_eject_errors(
+    instance: IntegrationInstance,
+):
+    assert "sr0" not in instance.execute("mount")
+    log = instance.read_from_file("/var/log/cloud-init.log")
+    assert "Failed ejecting the provisioning iso" not in log
+    verify_clean_log(log)
+
+
+@pytest.mark.skipif(PLATFORM != "azure", reason="Test is Azure specific")
+def test_azure_eject(session_cloud: IntegrationCloud):
+    """Integration test for GitHub #4732.
+
+    Azure uses `eject` but that is not always available on minimal images.
+    Ensure udev's eject can be used on systemd-enabled systems.
+    """
+    with session_cloud.launch(
+        launch_kwargs={
+            "image_id": session_cloud.cloud_instance.daily_image(
+                CURRENT_RELEASE.series, image_type=ImageType.MINIMAL
+            )
+        }
+    ) as instance:
+        source = get_validated_source(session_cloud)
+        if source.installs_new_version():
+            instance.install_new_cloud_init(
+                source, take_snapshot=False, clean=True
+            )
+            snapshot_id = instance.snapshot()
+            try:
+                with session_cloud.launch(
+                    launch_kwargs={
+                        "image_id": snapshot_id,
+                    }
+                ) as snapshot_instance:
+                    _check_for_eject_errors(snapshot_instance)
+            finally:
+                session_cloud.cloud_instance.delete_image(snapshot_id)
+        else:
+            _check_for_eject_errors(instance)
--- a/tests/unittests/sources/test_azure.py
+++ b/tests/unittests/sources/test_azure.py
@@ -3754,6 +3754,7 @@ class TestProvisioning:
         assert self.mock_azure_get_metadata_from_fabric.mock_calls == [
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev="/dev/sr0",
                 pubkey_info=None,
             )
@@ -3919,11 +3920,13 @@ class TestProvisioning:
         assert self.mock_azure_get_metadata_from_fabric.mock_calls == [
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev="/dev/sr0",
                 pubkey_info=None,
             ),
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev=None,
                 pubkey_info=None,
             ),
@@ -4039,11 +4042,13 @@ class TestProvisioning:
         assert self.mock_azure_get_metadata_from_fabric.mock_calls == [
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev="/dev/sr0",
                 pubkey_info=None,
             ),
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev=None,
                 pubkey_info=None,
             ),
@@ -4195,11 +4200,13 @@ class TestProvisioning:
         assert self.mock_azure_get_metadata_from_fabric.mock_calls == [
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev="/dev/sr0",
                 pubkey_info=None,
             ),
             mock.call(
                 endpoint="10.11.12.13",
+                distro=self.azure_ds.distro,
                 iso_dev=None,
                 pubkey_info=None,
             ),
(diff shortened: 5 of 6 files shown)


cloudinit/sources/helpers/azure.py
```diff
@@ -788,9 +788,15 @@ def clean_up(self):
 
     @azure_ds_telemetry_reporter
     def eject_iso(self, iso_dev) -> None:
+        eject_cmd = subp.subp(["eject", iso_dev])
+        # Prefer systemd /lib/udev/cdrom_id to avoid eject dependency
+        if self.distro.uses_systemd():
+            eject_cmd = subp.subp(
```
igalic, line 794: this here is distro knowledge, so why not put the eject_cmd into distro?

igalic (commented):
❓

TheRealFalcon:
@catmsred , I added the integration test. Setup is a bit more cumbersome than it should be, so I'd like to update some fixtures to make this kind of use case simpler, but that doesn't need to happen here. Additionally, I moved some things around based on @igalic 's comment.

Here's a patch you can `git apply`:
```diff
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index 79e262356..226280804 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -1240,6 +1240,15 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
 
+    @staticmethod
+    def eject_media(device: str) -> None:
+        cmd = (
+            ["/lib/udev/cdrom_id", "--eject-media", device]
+            if uses_systemd
+            else ["eject", device]
+        )
+        subp.subp(cmd)
+
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
     """
diff --git a/cloudinit/sources/DataSourceAzure.py b/cloudinit/sources/DataSourceAzure.py
index 4c5780f76..d15e84490 100644
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1370,6 +1370,7 @@ class DataSourceAzure(sources.DataSource):
         try:
             data = get_metadata_from_fabric(
                 endpoint=self._wireserver_endpoint,
+                distro=self.distro,
                 iso_dev=self._iso_dev,
                 pubkey_info=pubkey_info,
             )
diff --git a/cloudinit/sources/helpers/azure.py b/cloudinit/sources/helpers/azure.py
index f32940b7e..ea99f9fec 100644
--- a/cloudinit/sources/helpers/azure.py
+++ b/cloudinit/sources/helpers/azure.py
@@ -787,16 +787,10 @@ class WALinuxAgentShim:
             self.openssl_manager.clean_up()
 
     @azure_ds_telemetry_reporter
-    def eject_iso(self, iso_dev) -> None:
-        eject_cmd = subp.subp(["eject", iso_dev])
-        # Prefer systemd /lib/udev/cdrom_id to avoid eject dependency
-        if self.distro.uses_systemd():
-            eject_cmd = subp.subp(
-                ["/lib/udev/cdrom_id", "--eject-media", iso_dev]
-            )
+    def eject_iso(self, iso_dev, distro: distros.Distro) -> None:
+        LOG.debug("Ejecting the provisioning iso")
         try:
-            LOG.debug("Ejecting the provisioning iso")
-            eject_cmd
+            distro.eject_media(iso_dev)
         except Exception as e:
             report_diagnostic_event(
                 "Failed ejecting the provisioning iso: %s" % e,
@@ -805,7 +799,7 @@ class WALinuxAgentShim:
 
     @azure_ds_telemetry_reporter
     def register_with_azure_and_fetch_data(
-        self, pubkey_info=None, iso_dev=None
+        self, distro: distros.Distro, pubkey_info=None, iso_dev=None
     ) -> Optional[List[str]]:
         """Gets the VM's GoalState from Azure, uses the GoalState information
         to report ready/send the ready signal/provisioning complete signal to
@@ -836,7 +830,7 @@ class WALinuxAgentShim:
         )
 
         if iso_dev is not None:
-            self.eject_iso(iso_dev)
+            self.eject_iso(iso_dev, distro=distro)
 
         health_reporter.send_ready_signal()
         return ssh_keys
@@ -1014,13 +1008,14 @@ class WALinuxAgentShim:
 @azure_ds_telemetry_reporter
 def get_metadata_from_fabric(
     endpoint: str,
+    distro: distros.Distro,
     pubkey_info: Optional[List[str]] = None,
     iso_dev: Optional[str] = None,
 ):
     shim = WALinuxAgentShim(endpoint=endpoint)
     try:
         return shim.register_with_azure_and_fetch_data(
-            pubkey_info=pubkey_info, iso_dev=iso_dev
+            distro=distro, pubkey_info=pubkey_info, iso_dev=iso_dev
         )
     finally:
         shim.clean_up()
diff --git a/tests/integration_tests/bugs/test_gh4732.py b/tests/integration_tests/bugs/test_gh4732.py
deleted file mode 100644
index f426ab380..000000000
--- a/tests/integration_tests/bugs/test_gh4732.py
+++ /dev/null
@@ -1,40 +0,0 @@
-""" Integration test for GitHub #4732
-
-Azure uses `eject` but that is not always available on minimal images
-"""
-
-import pytest
-from pycloudlib.cloud import ImageType
-
-from tests.integration_tests.clouds import IntegrationCloud
-from tests.integration_tests.conftest import get_validated_source
-from tests.integration_tests.instances import IntegrationInstance
-from tests.integration_tests.integration_settings import PLATFORM
-from tests.integration_tests.releases import CURRENT_RELEASE
-
-def _check_for_eject_errors(
-    instance: IntegrationInstance,
-):
-    pass    
-
-@pytest.mark.skipif(PLATFORM != "azure", reason="Test is Azure specific")
-def test_azure_eject(session_cloud: IntegrationCloud):
-    source = get_validated_source(session_cloud)
-    if not source.installs_new_version():
-        pytest.skip(
-            "Provide CLOUD_INIT_SOURCE to install expected working cloud-init"
-        )
-    with session_cloud.launch(
-        launch_kwargs={
-            "image_id": session_cloud.cloud_instance.daily_image(
-                CURRENT_RELEASE.series, image_type=ImageType.XXX
-            )
-        }
-    ) as instance:
-        # We can't use setup_image fixture here because we want to avoid
-        # taking a snapshot or cleaning the booted machine after cloud-init
-        # upgrade.
-        instance.install_new_cloud_init(
-            source, take_snapshot=False, clean=False
-        )
-        _check_for_eject_errors(instance)
diff --git a/tests/integration_tests/datasources/test_azure.py b/tests/integration_tests/datasources/test_azure.py
new file mode 100644
index 000000000..896973de6
--- /dev/null
+++ b/tests/integration_tests/datasources/test_azure.py
@@ -0,0 +1,51 @@
+import pytest
+from pycloudlib.cloud import ImageType
+
+from tests.integration_tests.clouds import IntegrationCloud
+from tests.integration_tests.conftest import get_validated_source
+from tests.integration_tests.instances import IntegrationInstance
+from tests.integration_tests.integration_settings import PLATFORM
+from tests.integration_tests.releases import CURRENT_RELEASE
+from tests.integration_tests.util import verify_clean_log
+
+
+def _check_for_eject_errors(
+    instance: IntegrationInstance,
+):
+    assert "sr0" not in instance.execute("mount")
+    log = instance.read_from_file("/var/log/cloud-init.log")
+    assert "Failed ejecting the provisioning iso" not in log
+    verify_clean_log(log)
+
+
+@pytest.mark.skipif(PLATFORM != "azure", reason="Test is Azure specific")
+def test_azure_eject(session_cloud: IntegrationCloud):
+    """Integration test for GitHub #4732.
+
+    Azure uses `eject` but that is not always available on minimal images.
+    Ensure udev's eject can be used on systemd-enabled systems.
+    """
+    with session_cloud.launch(
+        launch_kwargs={
+            "image_id": session_cloud.cloud_instance.daily_image(
+                CURRENT_RELEASE.series, image_type=ImageType.MINIMAL
+            )
+        }
+    ) as instance:
+        source = get_validated_source(session_cloud)
+        if source.installs_new_version():
+            instance.install_new_cloud_init(
+                source, take_snapshot=False, clean=True
+            )
+            snapshot_id = instance.snapshot()
+            try:
+                with session_cloud.launch(
+                    launch_kwargs={
+                        "image_id": snapshot_id,
+                    }
+                ) as snapshot_instance:
+                    _check_for_eject_errors(snapshot_instance)
+            finally:
+                session_cloud.cloud_instance.delete_image(snapshot_id)
+        else:
+            _check_for_eject_errors(instance)
diff --git a/tests/integration_tests/instances.py b/tests/integration_tests/instances.py
index 3fc6558a2..242db9a2c 100644
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -188,6 +188,7 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
+        assert self.execute("apt-get update -q").ok
         assert self.execute("apt-get install -qy python3-passlib").ok
         assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
 
```

TheRealFalcon:
Also, my patch requires https://github.com/canonical/pycloudlib/pull/333

igalic:
i should add a patch to @TheRealFalcon's patch that adds FreeBSD support.

catmsred:
@TheRealFalcon Your changes look good to me -- do you want to push to this PR/start a new one or do you want me to incorporate the patch?

cloudinit/distros/__init__.py
```diff
@@ -1248,6 +1248,13 @@ def build_dhclient_cmd(
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
 
+    @staticmethod
+    def eject_media(device: str) -> None:
+        cmd = ["/lib/udev/cdrom_id", "--eject-media", device]
+        if uses_systemd():
```
TheRealFalcon, line 1254: Wait...did I do this backwards? Don't we want to use `lib/udev/cdrom_id` if we're using systemd?

catmsred: No that was me fighting with with tox (it didn't like your one-line conditional).  Let me fix it now.

cloudinit/distros/__init__.py
```diff
@@ -1248,6 +1248,13 @@ def build_dhclient_cmd(
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
 
+    @staticmethod
+    def eject_media(device: str) -> None:
+        cmd = ["/lib/udev/cdrom_id", "--eject-media", device]
+        if not uses_systemd():
+            cmd = ["eject", device]
+        subp.subp(cmd)
```
holmanb, line 1256: @catmsred @TheRealFalcon Now we're unconditionally using this for all systemd distros. How confident are we that all systemd-based distros ship with this binary? If we don't have high confidence maybe we should fallback to "eject"?

holmanb (changes requested):

cloudinit/distros/__init__.py
holmanb: > It's part of upstream systemd

The presence of a tool in upstream systemd doesn't guarantee its presence on every Linux distro that runs systemd.

In Ubuntu for example, various parts of "systemd" get shipped in separate packages for various reasons. Consider:

```
Package: systemd
Package: systemd-sysv
Package: systemd-container
Package: systemd-journal-remote
Package: systemd-coredump
Package: systemd-timesyncd
Package: systemd-tests
Package: libpam-systemd
Package: libnss-myhostname
Package: libnss-mymachines
Package: libnss-resolve
Package: libnss-systemd
Package: libsystemd0
Package: libsystemd-dev
Package: libsystemd-shared
Package: udev
Package: libudev1
Package: libudev-dev
Package: udev-udeb
Package: libudev1-udeb
Package: systemd-standalone-sysusers
Package: systemd-standalone-tmpfiles
Package: systemd-oomd
Package: systemd-userdbd
Package: systemd-homed
Package: systemd-boot
Package: systemd-boot-efi
Package: systemd-resolved
```

This tool gets shipped as part of the `udev` package in Ubuntu, not the `systemd` package. I very much doubt even ubuntu minimal images are going to drop dependency on udev, but it might be reasonable to not use udev on some Ubuntu-based appliance that uses cloud-init. Honestly I'm less concerned about the Ubuntu story (I expect udev to be in use most places) and more concerned about making assumptions for all the different linux distributions that use cloud-init. Do they all ship this utility as part of some package? It is entirely possible that some distro decided not to include it in their base images because of the pre-existance of `eject`.

I would also expect `util-linux` to be available almost everywhere that the Linux kernel is (which is more ubiquitous than systemd), but in this case it was decided that it wasn't required for minimal ubuntu images. However, I could see the same argument being made but the other way. The `eject` binary is half the size of the `cdrom_id` binary on my system 43240 bytes vs 84696 bytes, respectively, so why not use `eject` in `$some_distro`'s udev rules rather than `cdrom_id`?

I'm not opposed to making the code this way, but if we are going to, I'd say we should probably take a good look at every supported distro's packaging to make sure that this is available [everywhere](https://cloudinit.readthedocs.io/en/latest/reference/distros.html). In my opinion it would be easier (and safer) to just fail over to calling `eject` when `cdrom_id` isn't called - perhaps with an INFO log level message so that future devs can collect data through bug reports about whether this ever gets used.

TheRealFalcon: Good point. We may not even want the systemd check at all. Just check for cdrom_id and if it exists, use it, otherwise use eject (or vice versa).

catmsred: Do we have a strong preference for which (eject or cd_rom) we try first?  My inclination is to try eject first since that is what has been used historically for this function.

TheRealFalcon: > My inclination is to try eject first since that is what has been used historically for this function.

Agreed.

I think I'd prefer we look before we leap when it comes to shelling out though, just because it can be a more expensive operation. Something along the lines of:
```python
if subp.which("eject"):
    call eject
elif subp.which("/lib/udev/cdrom_id"):
    call cdrom_id
else:
    raise a custom exception
```

What do you think?

catmsred: Updated raising a `subp.ProcessExecutionError`; please take a look at whether you think that's a good choice and whether the description/reason are used as they should be.

TheRealFalcon (approved):
LGTM! I think this slightly broadens the scope of `ProcessExecutionError`, but I'm ok with it given that I don't think we really need a separate distinction. We can always add one later if needed.

tests/integration_tests/instances.py
```diff
@@ -216,6 +216,7 @@ def install_deb(self):
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
+        assert self.execute("apt-get update -q").ok
```
holmanb, line 219: Why this change? This is going to add a significant amount of time to each image launch, so I'd like to be sure we actually need this.

FWIW I don't think we need the following python3-passlib line - that was actually required in https://github.com/canonical/cloud-init/commit/f755bad5d3c683a7102acdf053c4121a5d568be8#diff-a525ccb597ad0f02ccca3a3660450a4ab7c846137bc358a2cc902c4f1b81b7c8, but shouldn't be required any longer since [now it is only required when crypt doesn't exist](https://github.com/canonical/cloud-init/commit/5ed70cc09d1ce068cc920d50f987cba2afffe46a) (Python > 3.12) and in that case it should be included in cloud-init's dependency list in distro packaging.

TheRealFalcon: It's just the first launch that creates the custom image, so it shouldn't be much of a problem, but yes, it was because of the passlib dependency.

cloudinit/distros/__init__.py
```diff
@@ -1248,6 +1248,21 @@ def build_dhclient_cmd(
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
 
+    @staticmethod
+    def eject_media(device: str) -> None:
+        cmd = None
+        if subp.which("eject"):
```
igalic, line 1254: i just checked, and this code path will work on OpenBSD, NetBSD, Dragonfly(!), and not on FreeBSD.

so I'll need to submit a patch

tests/integration_tests/instances.py
TheRealFalcon: so then we can remove this line. @catmsred , would you be able to do that?

catmsred: Removed the line and rebased.

cloudinit/distros/__init__.py
igalic: https://github.com/catmsred/cloud-init/pull/1

catmsred: Per @igalic's suggestion should we add camcontrol to the main distro list of ejects to try?  While it's currently only effecting FreeBSD, there may come a time when eject is removed from some of the other distros too.

TheRealFalcon: CAM is a FreeBSD thing, so I don't see a reason to apply it everywhere. I'd rather merge this first and then take the other PR separately.

holmanb (commented):
Thanks @catmsred, this looks good. Once you update the proposed commit message to reflect the current state of this PR we can merge this.

catmsred:
> Thanks @catmsred, this looks good. Once you update the proposed commit message to reflect the current state of this PR we can merge this.

Thanks @holmanb.  Commit message fixed.

holmanb (approved):
Thanks for your effort on this @catmsred!

It looks good to me - I'll merge it.

cjp256:
Much thanks!

PR #4770, opened 2024-01-11
feat(apt): deb822 sources update ubuntu.sources template and replace invalid /etc/apt/sources.list with Ubuntu default

blackboxsw:
## Rebase merge mutiple separate commits
 Updated per changes that just merged into Ubuntu Noble per https://code.launchpad.net/~juliank/livecd-rootfs/+git/livecd-rootfs/+merge/458366. The changset indicates that livecd-rootfs will leave around an "empty" breadcrumb `/etc/apt/sources.list` which will contain comments about the intent of Ubuntu to use deb822 format but no active APT configuration. cloud-init will now leave this breadcrumb in place and not warn in this case by checking md5sum of the file. If it is unaltered, leave it in place.

## Proposed Commit Message(s)
```
--- feat(apt): skip known md5sums /etc/apt/sources.list
    
    Some image creation software will place known stub files in
    /etc/apt/sources.list to inform users in comments of desired
    deb822 format.
    
    When deb822 is preferred and /etc/apt/sources.list exists, do
    not remove that file if it is has a known safe md5sum value.
    
    Add md5um for Ubuntu's default /etc/apt/sources.list which
    has the following content:
     # Ubuntu sources have moved to the /etc/apt/sources.list.d/ubuntu.sources
     # file, which uses the deb822 format. Use deb822-formatted .sources files
     # to manage package sources in the /etc/apt/sources.list.d/ directory.
     # See the sources.list(5) manual page for details.

--- feat(apt): use APT deb822 source format by default
    
    Prefer default APT configuration expressed as deb822 format.
    
    Update default ubuntu.deb822.tmpl documentation and suites/components
    for APT configuration on Ubuntu Noble only which cloud-init renders
    on first boot.
    
    Default deb sources content for Ubuntu is sourced from
    livecd-rootfs:live-build/functions.
```

## Additional Context
A [livecd-rootfs merge request](https://code.launchpad.net/~juliank/livecd-rootfs/+git/livecd-rootfs/+merge/458366) is in flight to update this doc data for other Ubuntu image builds as well and cloud-init should align with common text, suites and components in livecd-rootfs as well. CC: @julian-klode  for review/acceptance on this PR for cloud-init once final livecd-roots PR default text,suites and components is settled.
PR Merged as of Feb 12th

## Test Steps
```
make deb
CLOUD_INIT_KEEP_INSTANCE=1 CLOUD_INIT_OS_IMAGE=noble CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init*bddeb.deb  tox -e integration-tests -- tests/integration-tests/modules/test_apt_functionality.py::TestDefaults
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -184,6 +184,18 @@ PORTS_MIRRORS = {
 PRIMARY_ARCHES = ["amd64", "i386"]
 PORTS_ARCHES = ["s390x", "arm64", "armhf", "powerpc", "ppc64el", "riscv64"]
 
+UBUNTU_DEFAULT_APT_SOURCES_LIST = """\
+# Ubuntu sources have moved to the /etc/apt/sources.list.d/ubuntu.sources
+# file, which uses the deb822 format. Use deb822-formatted .sources files
+# to manage package sources in the /etc/apt/sources.list.d/ directory.
+# See the sources.list(5) manual page for details.
+"""
+
+# List of allowed content in /etc/apt/sources.list when features
+# APT_DEB822_SOURCE_LIST_FILE is set. Otherwise issue warning about
+# invalid non-deb822 configuration.
+DEB822_ALLOWED_APT_SOURCES_LIST = {"ubuntu": UBUNTU_DEFAULT_APT_SOURCES_LIST}
+
 
 def get_default_mirrors(
     arch=None,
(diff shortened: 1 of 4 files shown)


TheRealFalcon (approved):
@blackboxsw , approve from me, but I know you're waiting for a review from Julien.

julian-klode:
I don't think it makes a lot of sense at this point in time, we need to land the livecd-rootfs default first and then that's the point where we should have agreement.

blackboxsw:
> I don't think it makes a lot of sense at this point in time, we need to land the livecd-rootfs default first and then that's the point where we should have agreement.

All good @julian-klode @TheRealFalcon, we'll await closure/consensus on your livecd-rootfs and only merge in cloud-init once we have that content committed to that repository.

blackboxsw:
Setting this to draft state until the livecd-rootfs changeset lands in that project

julian-klode:
How about you just also write the same sources.list with comments in it, then you don't need to track it's md5, just check if you would write the same file and do nothing? Or does it break if you have both? That would be unfortunate

blackboxsw:
`> How about you just also write the same sources.list with comments in it, then you don't need to track it's md5, just check if you would write the same file and do nothing? Or does it break if you have both? That would be unfortunate

Thanks @julian-klode. That'd work on Ubuntu but not Debian, and we would also lose our visibility to potential ignored invalid APT configuration in image creation. If an image creator is trying to configure APT via `/etc/apt/sources.list` in Ubuntu, it'll magically get overwritten by cloud-init on first boot without any warning or notice that config was changed or dropped in favor of /etc/apt/sources.list.d/ubuntu.sources`. Our warning log is attempting to raise the awareness of that condition instead of steamrolling over the top of it.

julian-klode (approved):
lgtm then

blackboxsw:
> lgtm then

@julian-klode thanks, I took your suggestion under consideration and provided an updated commit that actually validates
the existing /etc/apt/sources.list file content. If not an allowed /etc/apt/sources.list content, then cloud-init will write that same file out (after warning). This way the end-state is the same, Ubuntu with an informative, commented /etc/apt/sources.list  and cloud-init alerting that something in the image creation process left around configuration that is disregarded.

blackboxsw:
Updated per your review comments and rekicked jobs.

TheRealFalcon (approved):
LGTM!

PR #4772, opened 2024-01-14
Modify hook-hotplug to be POSIX shell, add OpenRC init script

dermotbradley:
## Proposed Commit Message

```
Change hoot-hotplug from being Bash-specific to POSIX shell.

Add OpenRC init script for starting hotplug "daemon".

Add hotplug "daemon" that corresponds to cloud-init's
cloud-init-hotplugd.service/cloud-init-hotplugd.socket functionality.
```

## Checklist

## Merge type

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -141,6 +141,7 @@ INITSYS_FILES = {
     "sysvinit_openrc": lambda: [
         f for f in glob("sysvinit/gentoo/*") if is_f(f)
     ],
+    "sysvinit_openrc.dep": lambda: ["tools/cloud-init-hotplugd"],
     "systemd": lambda: [
         render_tmpl(f)
         for f in (
@@ -164,6 +165,7 @@ INITSYS_ROOTS = {
     "sysvinit_openbsd": "etc/rc.d",
     "sysvinit_deb": "etc/init.d",
     "sysvinit_openrc": "etc/init.d",
+    "sysvinit_openrc.dep": "usr/lib/cloud-init",
     "systemd": pkg_config_read("systemd", "systemdsystemunitdir"),
     "systemd.generators": pkg_config_read(
         "systemd", "systemdsystemgeneratordir"
--- a/sysvinit/gentoo/cloud-init-hotplug
+++ b/sysvinit/gentoo/cloud-init-hotplug
@@ -1,0 +1,23 @@
+#!/sbin/openrc-run
+
+description="cloud-init hotplug daemon"
+
+command="/usr/lib/cloud-init/cloud-init-hotplugd"
+pidfile="/run/$RC_SVCNAME.pid"
+
+depend() {
+  after cloud-final
+}
+
+start() {
+  if grep -q 'cloud-init=disabled' /proc/cmdline; then
+    ewarn "$RC_SVCNAME is disabled via /proc/cmdline."
+  elif test -e /etc/cloud/cloud-init.disabled; then
+    ewarn "$RC_SVCNAME is disabled via cloud-init.disabled file"
+  else
+    ebegin "$description"
+    start-stop-daemon --start --background --exec $command \
+      --make-pidfile --pidfile $pidfile
+    eend $?
+  fi
+}
--- a/tools/cloud-init-hotplugd
+++ b/tools/cloud-init-hotplugd
@@ -1,0 +1,23 @@
+#!/bin/sh
+# This file is part of cloud-init. See LICENSE file for license information.
+
+# This script is used on non-systemd systems. It is called by the
+# cloud-init-hotplug init.d script.
+#
+# Creates a named pipe and then continually listens to this pipe. The pipe
+# is written to by the hook-hotplug script (which is called by a udev rule
+# upon a network device event). Anything received via the pipe is then
+# passed on via the "cloud-init devel hotplug-hook handle" command.
+
+PIPE="/run/cloud-init/hook-hotplug-cmd"
+
+mkfifo -m700 $PIPE
+
+while true; do
+  # shellcheck disable=SC2162
+  read args < $PIPE
+  # shellcheck disable=SC2086
+  exec /usr/bin/cloud-init devel hotplug-hook $args
+done
+
+exit
--- a/doc/rtd/explanation/events.rst
+++ b/doc/rtd/explanation/events.rst
@@ -72,9 +72,10 @@ metadata, ``cloud-init`` will also bring up/down the newly added interface.
 
 .. warning::
    Due to its use of ``systemd`` sockets, ``hotplug`` functionality is
-   currently incompatible with SELinux. This issue is being `tracked
-   in GitHub #3890`_. Additionally, ``hotplug`` support is considered
-   experimental for non-Debian-based systems.
+   currently incompatible with SELinux on Linux distributions using systemd.
+   This issue is being `tracked in GitHub #3890`_. Additionally, ``hotplug``
+   support is considered experimental for non-Alpine and non-Debian-based
+   systems.
 
 Example
 =======
--- a/tools/hook-hotplug
+++ b/tools/hook-hotplug
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 # This file is part of cloud-init. See LICENSE file for license information.
 
 # This script checks if cloud-init has hotplug hooked and if
@@ -11,12 +11,12 @@ is_finished() {
 if is_finished; then
     # open cloud-init's hotplug-hook fifo rw
     exec 3<>/run/cloud-init/hook-hotplug-cmd
-    env_params=(
-        --subsystem="${SUBSYSTEM}"
-        handle
-        --devpath="${DEVPATH}"
-        --udevaction="${ACTION}"
-    )
+    env_params=" \
+        --subsystem=\"${SUBSYSTEM}\" \
+        handle \
+        --devpath=\"${DEVPATH}\" \
+        --udevaction=\"${ACTION}\" \
+    "
     # write params to cloud-init's hotplug-hook fifo
-    echo "${env_params[@]}" >&3
+    echo "${env_params}" >&3
 fi


sysvinit/gentoo/cloud-init-hotplug
```diff
@@ -0,0 +1,18 @@
+#!/sbin/openrc-run
+
```
blackboxsw, line 2: I don't have much depth in openrc files, so please excuse the questions if they don't make sense.

It seems given other examples of openrc files, that we are missing helpful attributes that allow others to introspect this daemon's purpose, looking over /etc/init.d/sshd I see we can have helpful attributes such as 
```suggestion
description="cloud-init hotplug hook daemon"
```

Without the above diff I can't see much about this daemon
```
~ # rc-service cloud-init-hotplugd describe
 * No description for cloud-init-hotplugd
 * healthcheck: no description
 * unhealthy: no description
 * cgrp_cleanup: Kill all processes in the cgroup
```
I propose we minimally:
- provide descriptions of this service here.
- it would be nice to have an easy comment  header providing context

do we want or need a checkconfig function?

sysvinit/gentoo/cloud-init-hotplug
```diff
@@ -0,0 +1,18 @@
...
+  before cloud-init
+}
+
+start_pre() {
+   if grep -q 'cloud-init=disabled' /proc/cmdline; then
+     ewarn "$RC_SVCNAME is disabled via /proc/cmdline."
+   elif test -e /etc/cloud/cloud-init.disabled; then
+     ewarn "$RC_SVCNAME is disabled via cloud-init.disabled file"
+   fi
+   eend 0
```
blackboxsw, line 17: I don't see us invoking $command anywhere as I've seen in other openrc scripts. Is this done automatically by s6-svcscan for us automatically? I'm also uncertain how we really prevent this service from running if s6-svcscan is automatically calling start on the $command. I'd expect we have some facility to disable trying to start a service in openrc due to unmet dependencies/config conditions.

tools/cloud-init-hotplugd
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
```
blackboxsw, line 2: Let's add a comment or breadcrumb here about how this daemon plays a role in proxying udev events to cloud-init devel hotplug command calls. So others can understand some context of how this daemon is being used if this is the first file they find on a system.

setup.py
```diff
@@ -280,6 +280,7 @@ def finalize_options(self):
     (
         USR_LIB_EXEC + "/cloud-init",
         [
+            "tools/cloud-init-hotplugd",
```
blackboxsw, line 283: I'd like to avoid packaging this file on systems that do have systemd since we are already shipping systemd service/socket files with this behavior. Maybe we can conditionally append this file to data_files if we are on self.init_system != "systemd" in the `finalize_options`?

blackboxsw (changes requested):
Thank you @dermotbradley for fixing this daemon up for non openrc environments. A couple of questions to get my bearing on how we should better provide openrc files and functionality and I'd like exclusion of the hotplugd in packaging for non-systemd targets to avoid confusion or potentiall duplication with existing systemd units/services if someone tries to invoke that hotplugd on a system that is already camping out on a named pipe.

sysvinit/gentoo/cloud-init-hotplug
blackboxsw: Otherwise if s6-svcscan isn't our service supervisor, don't we need a start/stop/restart function in this openrc file that would only attempt to start the service if our warning conditions weren't triggered?

sysvinit/gentoo/cloud-init-hotplug
```diff
@@ -0,0 +1,18 @@
+#!/sbin/openrc-run
+
+command="/usr/lib/cloud-init/cloud-init-hotplugd"
+command_background=yes
+pidfile="/run/$RC_SVCNAME.pid"
+
+depend() {
+  before cloud-init
```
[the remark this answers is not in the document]
dermotbradley: > Why position before cloud-init?

I based this on systemd/cloud-init-hotplugd.socket defining "WantedBy=cloud-init.target" which I took to mean that the hotplug should be run before cloud-init

sysvinit/gentoo/cloud-init-hotplug
dermotbradley: I based this openrc file on the other existing cloud-init openrc files so I configured it accordingly.

Whereas if you look at the hotplug file I've packaged for Alpine for some months it does include a description: https://git.alpinelinux.org/aports/tree/community/cloud-init/cloud-init-hotplugd.initd

So you're flagging a more general issue about openrc files in cloud-init in general.

sysvinit/gentoo/cloud-init-hotplug
dermotbradley: > Is this done automatically by s6-svcscan for us automatically?

I'm not sure why you are focusing on s6-svcscan.

Whilst s6 can be used as a supervisor for OpenRC it is *not* the default on - stop-stop-daemon is. In order to use s6-svcscan (which this script is not doing) you would need to specifically enable it, see: https://github.com/OpenRC/openrc/blob/master/s6-guide.md

sysvinit/gentoo/cloud-init-hotplug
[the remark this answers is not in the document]
blackboxsw: The [cloud-init.target is just a synchronization point in boot that means that all cloud-init related units/services have finished running](https://github.com/canonical/cloud-init/blob/main/systemd/cloud-init.target#L6-L7) It doesn't represent a specific point in time before cloud-init is running, but a point in boot that external services/units can be sure cloud-init's 4 boot stages are all done The `WantedBy=cloud-init.target` in each of our subordinate units/services represented by our separate boot stages basically just registers those stages as blocking the overall status of `cloud-init.target`.  Ultimately, `cloud-init.target` is only completed/unblocked once we ultimately complete running the final boot stage  `cloud-init-final.service` since it's our last ordered systemd unit ordered `After=cloud-config.service. So, I think the equivalent we want here is `after cloud-final` since there is no cloud-init.target represented in gentoo-openrc land

sysvinit/gentoo/cloud-init-hotplug
blackboxsw: Ahh, ok. Looking through those openrc files now, it is a general question/concern about the preexisting gentoo openrc files and doesn't really need to be handled in this PR, but setting a good example for docs in this supplemental addtion would be helpful and we'll reflect it elsewhere as time permits.

sysvinit/gentoo/cloud-init-hotplug
blackboxsw: Yes s6-svscan was just bad google-fu and searching on my part. Thanks for the link.

My question then becomes, this particular openrc script doesn't appear to actually provide any start/stop bit of functionality or invoking any commands. It only tests and emits warning messages about disabled conditions and exits 0. The other gentoo sysvinit openrc scripts all seem to invoke `cloud-init init ...` on the command line when cloud-init is not in a disabled state. Does this current script actually need to invoke `/usr/lib/cloud-init/cloud-init-hotplugd` directly when no disabled/warning conditions exist?

I'm wondering if the openrc script needs to be something like the following so we can be sure that hotplugd service is only started when and if cloud-init is not disabled?

```
start() {
start() {
  if grep -q 'cloud-init=disabled' /proc/cmdline; then
    ewarn "$RC_SVCNAME is disabled via /proc/cmdline."
  elif test -e /etc/cloud/cloud-init.disabled; then
    ewarn "$RC_SVCNAME is disabled via cloud-init.disabled file"
  else
    # only get here if cloud-init is enabled
    start-stop-daemon --start --exec $command --pidfile $pidfile
  fi
  eend 0
}
```

setup.py
dermotbradley: Ok, I've changed it to be conditional on openrc.

tools/cloud-init-hotplugd
dermotbradley: Done

sysvinit/gentoo/cloud-init-hotplug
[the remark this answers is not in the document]
dermotbradley: Ok, I changed this to "after cloud-final".

sysvinit/gentoo/cloud-init-hotplug
```diff
@@ -0,0 +1,18 @@
...
+}
+
+start_pre() {
+   if grep -q 'cloud-init=disabled' /proc/cmdline; then
+     ewarn "$RC_SVCNAME is disabled via /proc/cmdline."
```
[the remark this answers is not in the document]
dermotbradley: Fixed now. I was trying to use start_pre(), rather than writing a start() but it seems these is no way to signal a warning in start_pre() such that the start() function is not executed - only an error can be signally but that then produces a visible error during boot.

So I switched to using a start() function instead.

sysvinit/gentoo/cloud-init-hotplug
dermotbradley: I have tidied up the cloud-init-hotplug openrc script - it now has a description and switch from a start_pre function to a start function.

I'll look at making changes to the other openrc script via a separate PR.

sysvinit/gentoo/cloud-init-hotplug
dermotbradley: > My question then becomes, this particular openrc script doesn't appear to actually provide any start/stop bit of functionality or invoking any commands.

That's the magic of OpenRC - by default you are **not** expected to write start() and stop() functions, OpenRC uses builtin functions which use the "command", "pidfile", and other variables to know what to start and stop. From OpenRC [docs](https://github.com/OpenRC/openrc/blob/master/service-script-guide.md):

```
All scripts must have start/stop/status functions, but defaults are provided and should be used unless you have a very strong reason not to use them.
```

However in this case I've now added a start() function as I couldn't do what I wanted to do via a start_pre() function.

dermotbradley:
I also added a minor change to the events.rst docs.

setup.py
```diff
@@ -141,6 +141,9 @@ def render_tmpl(template, mode=None, is_yaml=False):
     "sysvinit_openrc": lambda: [
         f for f in glob("sysvinit/gentoo/*") if is_f(f)
     ],
+    "sysvinit_openrc.dep": lambda: [
+        f for f in glob("tools/cloud-init-hotplugd") if is_f(f)
+    ],
```
blackboxsw, line 146: No need to glob here since we are just providing one known file with this lambda
```suggestion
    "sysvinit_openrc.dep": lambda: ["tools/cloud-init-holplugd"],
```

blackboxsw (changes requested):
One minor change left and all looks good to me. +1

setup.py
blackboxsw: I've pushed this suggestion to setup.py and will land when CI clears. Thanks again @dermotbradley

blackboxsw (approved):

PR #4775, opened 2024-01-16
fix(debian): Remove/Delete README in /etc/cloud/clean.d/

CalvoM:
## Proposed Commit Message

```
fix(debian): Remove/Delete README in /etc/cloud/clean.d/

Fixes GH-4760
```
Solves #4760 
Works with #4761 

## Checklist

## Merge type

### DIFF
--- a/debian/cloud-init.maintscript
+++ b/debian/cloud-init.maintscript
@@ -7,3 +7,4 @@ rm_conffile /etc/init/cloud-init-nonet.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-log-shutdown.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3~4ga31370af-0ubuntu1~
+rm_conffile /etc/cloud/clean.d/README 24.1~4~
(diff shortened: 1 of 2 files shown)


blackboxsw (changes requested):
Correction on my previous comment: @CalvoM 
1. looks like it will need to change to something less than our next planner release version, which is not yet completely set. But for Noble, that next released version will be 24.1~4<git_hash> or greater (maybe 24.1 itself).... so our version in the maintscript here should be `24.1~4~` since our next release will be at least >= `24.1~4`

2. We need to also have a separate commit that adds a new UNRELEASED debian/changelog that via `dch -i` and mentions:

`* d/cloud-init.maintscript: remove /etc/cloud/clean.d/README`

Same comment on other release branches. Let's just set the version to

debian/cloud-init.maintscript
```diff
@@ -7,3 +7,4 @@ rm_conffile /etc/init/cloud-init-nonet.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-init.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/init/cloud-log-shutdown.conf 0.7.9-243-ge74d775-0ubuntu2~
 rm_conffile /etc/NetworkManager/dispatcher.d/hook-network-manager 23.3~4ga31370af-0ubuntu1~
+rm_conffile /etc/cloud/clean.d/README 24.1~3gb729a4c4-0ubuntu1~
```
blackboxsw, line 10: Turns out we actually need this version to be matching the next expected release version. Since the devel release will adhere to either 24.1~4 if we publish another snapshot prior to upstream release of 24.1, or it'll be `24.1` our best best to match both cases is below

 ```suggestion
rm_conffile /etc/cloud/clean.d/README 24.1~4~
```

CalvoM: Thanks for this. I will update the maintscripts.

blackboxsw (approved):
LGTM and tested on noble thanks much.

PR #4780, opened 2024-01-16
Fix Alpine and Mariner /etc/hosts templates

### LINKED ISSUE #4771: alpine hosts template creates unexpected behavior for hostname -f command
# Bug report
the layout of the current hosts.alpine.tmpl file results in unexpected behavior from the `hostname -f` command which should return the hostname + fqdn but only returns the hostname. 

It looks like this happens because of the changes in https://github.com/canonical/cloud-init/pull/1084.

deploy a new vm based on the alpine 3.19 NoCloud image and provide both hostname and fqdn.
run `hostname` and `hostname -f` in the newly deployed machine, they should not return exactly the same output.
if you switch around hostname and fqdn in /etc/hosts it will work as expected.

## Environment details
- Cloud-init version: 23.4
- Operating System Distribution: alpine 3.19
- Cloud provider, platform or installer type: NoCloud

dermotbradley:
## Proposed Commit Message

```
Fix Alpine and Mariner /etc/hosts templates

Revert the changes I introduced with #1084, these had the side-effect
of causing "hostname" problems. Fixes #4771.

Also remove the "::1 {{fqdn}} {{hostname}}" entry as this this being
results in strange behaviour with both Busybox "hostname -i" (returns
"127.0.0.1", or "127.0.0.1 A.B.C.D" if an additional non-loopback IP
entry in /etc/hosts) and Coreutils "hostname -i" (returns "::1" in
both situations), commands whereas if the "::1" entry is removed then
both Busybox and Coreutils behave the the same fashion returning
"127.0.0.1" or "127.0.0.1 A.B.C.D" as appropriate.

Also correct hosts.mariner.tmpli file it was not valid (every line
was prefixed with "+").

Fixes GH-4771
```

## Checklist

## Merge type


holmanb (approved):
LGTM

PR #4784, opened 2024-01-17
landscape-client: handle already registered client

chifac08:
## landscape-client: handle already registered client

```
This PR makes use of "is-registered" parameter to prevent duplicate entries 
of clients on a Landscape server when cloud-init runs more than once.
```

## Additional Context

The parameter was introduced to landscape-client in version 23.02 (https://github.com/canonical/landscape-client/pull/103) and has been available in "`release (main`)" repository since Lunar. The package is currently in "`proposed`" repository for Focal as well as Jammy and will hopefully be released in near future. So please keep in mind, if this PR should make it's way in the upstream we have to take care about the current release status of all supported Ubuntu releases. Please have a look at [launchpad](https://launchpad.net/landscape-client.)

## Checklist

## Merge type


cloudinit/config/cc_landscape.py
```diff
@@ -151,7 +151,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             ]
         )
     )
-    subp.subp(["landscape-config", "--silent"] + cmd_params)
+    subp.subp(["landscape-config", "--silent", "--is-registered"] + cmd_params)
```
blackboxsw, line 154: I think we instead would want a pre-flight check of "landscape-config",  "--silent", "--is-registered" looking for a success exit 0. If we get exit zero, then emit a warning in logs that landscape-client is already registered. If exit 5, then we fall through to the full landscape-config --silent with the supplemental cmd_params.

blackboxsw (changes requested):
Thank you @chifac08 for continued support and contributions to cloud-init. 
I do think we probably want to pull the landscape-config --is-registered check up before we invoke the full landscape-config --silent command with all the cmd_params. Otherwise, I'm not quite certain how landscape-config is expected to behave when not registered, but when it's also provided with all necessary config options to register.

I think we'll also want to check with @Perfect5th on the estimated SRU verification for the currently queued SRU back to Jammy for landscape-client v. 23 before we land this as we'd have to back out those changes on stable releases for cloud-init which also targets back to Jammy.

cloudinit/config/cc_landscape.py
chifac08: Agreed and implemented. What do you think? 

Unfortunately, integration tests are failing due to a new line character in "expected_calls" array. 
`call(['landscape-config', '--silent', '--is-registered'], rcs=[5]),\n call(['landscape-config', '--silent', '--computer-title', 'My" PC', '--data-path', '/var/lib/landscape/client', '--log-level', 'info', '--ping-url', 'http://landscape.canonical.com/ping', '--url', 'https://landscape.canonical.com/message-system'])`

I can't figure it out why the first "call" tuple of the array will be interpreted with a new line character. Any help is much appreciated. thx

chifac08: Nevermind, figured it out what was causing this issue and fixed it :)

chifac08:
Implemented changes you requested. Furthermore, I added some unit tests. Please review again. thx

cloudinit/config/cc_landscape.py
```diff
@@ -151,8 +151,17 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             ]
         )
     )
-    subp.subp(["landscape-config", "--silent"] + cmd_params)
-    util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
+    try:
+        subp.subp(["landscape-config", "--silent", "--is-registered"], rcs=[5])
```
blackboxsw, line 155: This will not raise a ProcessExecutionError if you pass rcs=[5] on a registered system.
```
# Note I'm registered in this instance and the exit 5
root@nnn:~# landscape-config --silent --is-registered

Registered:    False
Config Path:   /etc/landscape/client.conf
Data Path      /var/lib/landscape/client
root@nnn:~# echo $?
5
# Note that subp with rcs=[5] allows the subprocess to receive the exit code 5 without raising a ProcessExecutionError

root@nnn:~# python3 -c 'from cloudinit.subp import subp; subp(["landscape-config", "--silent", "--is-registered"], rcs=[5])'

# If we drop the rcs=[5], we get the desired ProcessExecutionError

root@nnn:~# python3 -c 'from cloudinit.subp import subp; subp(["landscape-config", "--silent", "--is-registered"])'
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 277, in subp
    raise ProcessExecutionError(
Command: ['landscape-config', '--silent', '--is-registered']
Exit code: 5
Reason: -
Stdout: 
        Registered:    False
        Config Path:   /etc/landscape/client.conf
        Data Path      /var/lib/landscape/client
Stderr: 
```
So,  I think what you want is actually the following and instead of checking for exit_code == 0 below check for exit_code == 5
```suggestion
        subp.subp(["landscape-config", "--silent", "--is-registered"])
```

cloudinit/config/cc_landscape.py
```diff
@@ -151,8 +151,17 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
...
+        subp.subp(["landscape-config", "--silent"] + cmd_params)
+        util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
+    except subp.ProcessExecutionError as e:
+        if e.exit_code == 0:
```
blackboxsw, line 159: Per the above I think we want to actually check for exit_code 5 to indicate already registered
```suggestion
        if e.exit_code == 5:
```

cloudinit/config/cc_landscape.py
chifac08: landscape-client code states that exit code **5** stands for [NOT REGISTERED](https://github.com/canonical/landscape-client/blob/51b8d3243278d1d5405db832c3471d04d784de13/landscape/client/configuration.py#L40) and 0 if client is already registered. Based on this facts,  i think we should continue (on exit code 5) and register the client and log out a warning when exit code is 0. 
I made following tests on Focal with landscape-client 23.02:

```
# Already registered to Landscape
root@test01:/# landscape-config --silent --is-registered

Registered:    True
Config Path:   /etc/landscape/client.conf
Data Path      /var/lib/landscape/client
Account Name:  standalone
root@test01:/# echo $?
0

# Not registered to Landscape
root@test01:/# landscape-config --silent --is-registered

Registered:    False
Config Path:   /etc/landscape/client.conf
Data Path      /var/lib/landscape/client
root@test01:/# echo $?
5
```

I am aware that an exception will be thrown for 
`subp.subp(["landscape-config", "--silent", "--is-registered"], rcs=[5])`. But we do not log the exception to file, which could possibly irritate a user, rather than writing a simple warning message that the client is already registered.


What do you think?

cloudinit/config/cc_landscape.py
chifac08: Please see my explanation above

cloudinit/config/cc_landscape.py
blackboxsw: oops. You are correct @chifac08. I misinterpreted the fact that your rcs=[5] lacked the exit 0, which would end up raising ProcessExecutionError on exit 0 in the case when already registered. Ok this makes perfect sense. And logging a warning is good nowadays too as newer versions of cloud-init will exit(2) when warnings are present, making that conditon more conspicious. I also think we don't want to log full exceptions or tracebacks with warning logs can suffice to explain the situation clearly enough. So, your approach works fine here.

cloudinit/config/cc_landscape.py
blackboxsw: agreed and reject this comment. I misread the rcs=[5] behavior and expectation from landscape-config exit codes.

cloudinit/config/cc_landscape.py
blackboxsw: Ok, resolving this message as your approach will work fine as long as --is-registered parameter is available from lansdcape-config command. (Again we can wait for SRU of landscape-client for this feature on all supported cloud-init Ubuntu series.

cloudinit/config/cc_landscape.py
```diff
@@ -151,8 +151,17 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             ]
         )
     )
-    subp.subp(["landscape-config", "--silent"] + cmd_params)
-    util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
+    try:
+        subp.subp(["landscape-config", "--silent", "--is-registered"], rcs=[5])
+        subp.subp(["landscape-config", "--silent"] + cmd_params)
+        util.write_file(LS_DEFAULT_FILE, "RUN=1\n")
```
blackboxsw, line 157: I actually think writing this file a feature that is now obsolete and unnecessary given that we are running landscape-config directly which will start the landscape-client services up and ensure they are active at boot. It was first introduced in 2012 when this module was writing directly to `/etc/landscape/client.conf` and performing a `service landscape-client restart` after writing configs. I don't think the recent landscape-client on any supported Ubuntu series from Focal onward consumes this RUN=1 setting in /etc/default/landscape-client anymore.  So, let's drop the LS_DEFAULT_FILE variable and this write_file  as well.

blackboxsw: Just validated that all services are appropriately started on Ubuntu focal with landscape-client 18.01-0ubuntu4.3 and no RUN=1 needed in /etc/default/landscape-client. Yes this can be dropped/

tests/unittests/config/test_cc_landscape.py
```diff
@@ -210,6 +224,37 @@ def test_handler_writes_merged_provided_cloudconfig_with_defaults(
         )
         assert expected_calls == m_subp.call_args_list
 
+    @mock.patch(f"{MPATH}.merge_together")
+    def test_handler_client_failed_registering(self, m_subp, m_merge_together):
```
blackboxsw: The decorator ordering is flipped, which results in the unittest errors here.
```suggestion
    def test_handler_client_failed_registering(self, m_merge_together, m_subp):
```

tests/unittests/config/test_cc_landscape.py
```diff
@@ -210,6 +224,37 @@ def test_handler_writes_merged_provided_cloudconfig_with_defaults(
...
+            "Failure registering client:\nUnexpected error while"
+            " running command.\nCommand: -\nExit code: -\nReason: -\n"
+            "Stdout: Could not register client\nStderr: -"
+        )
+        with pytest.raises(RuntimeError, match=match):
+            cc_landscape.handle("notimportant", cfg, mycloud, None)
+
+    @mock.patch(f"{MPATH}.merge_together")
+    def test_handler_client_is_already_registered(
+        self, m_subp, m_merge_together, caplog
```
blackboxsw: ditto per above.

```suggestion
        self, m_merge_together, m_subp, caplog
```

blackboxsw (approved):
Approve the changeset, but I'd prefer to avoid landing this until https://bugs.launchpad.net/landscape-client/+bug/2006402 lands and SRUs to Focal so we don't have to stub out the --is-registered parameter logic.

tests/unittests/config/test_cc_landscape.py
chifac08: thanks for the fix and explanation :)

blackboxsw:
Still waiting on SRU resolution for landscape-client before this lands.

TheRealFalcon:
@blackboxsw , looks like the last comment on that bug is from 2023-10-20 . Do we expect that to be SRUed anytime soon?

blackboxsw:
> @blackboxsw , looks like the last comment on that bug is from 2023-10-20 . Do we expect that to be SRUed anytime soon?

Yep. I had circled back with the team and was assured they had it on their planning and it looks like verification just got unblocked yesterday. We expect this release within a week now that verification is complete for Focal/Jammy. https://bugs.launchpad.net/landscape-client/+bug/2006402/comments/53

blackboxsw:
Checked again on this landscape-client SRU. All tests are verified, SRU review queue is saturated given demand for Noble release reviews. Giving this another week to allow the Noble release week testing to pass this week which should free up SRU vanguards to unblock landscape-client publication.

blackboxsw:
chifac08  looks like landscape-client published yesterday. I just validated your PR works great on both Ubuntu focal and jammy with the newly published landscape-client --is-registered param. Finally landing this without further delay.

PR #4785, opened 2024-01-18
Modify ds-identify for Alpine support and add OpenRC init.d script

dermotbradley:
```
feat(Alpine): Modify ds-identify for Alpine support and add OpenRC init.d script

Add OpenRC init.d script for ds-identify.

Add support to ds-identify for "virt-what" as an alternative to
"systemd-detect-virt" for non-systemd Linux distros.

Change ds-identify to use short option for "modprobe" for Busybox
compatibility.

Change ds-identify to use short options for "grep" as POSIX does not
define long options, they are a GNU-ism.

```

## Checklist

## Merge type

### DIFF
--- a/sysvinit/gentoo/cloud-init-ds-identify
+++ b/sysvinit/gentoo/cloud-init-ds-identify
@@ -1,0 +1,21 @@
+#!/sbin/openrc-run
+
+depend() {
+  after localmount
+  before net
+  before cloud-init-local
+}
+
+start() {
+  ebegin "cloud-init-ds-identify"
+
+  if grep -q 'cloud-init=disabled' /proc/cmdline; then
+    ewarn "$RC_SVCNAME is disabled via /proc/cmdline."
+  elif test -e /etc/cloud/cloud-init.disabled; then
+    ewarn "$RC_SVCNAME is disabled via cloud-init.disabled file"
+  else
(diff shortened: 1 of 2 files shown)


tools/ds-identify
```diff
@@ -747,11 +759,15 @@ probe_floppy() {
     [ -b "$fpath" ] ||
         { STATE_FLOPPY_PROBED=1; return 1; }
```
igalic, line 760: (I didn't realise we had this. I also recently learned we're removing floppy support from FreeBSD in 15 or 16)

sysvinit/gentoo/cloud-init-ds-identify
igalic, line 1: I should adapt this for FreeBSD

igalic (approved):
this looks pretty good at first blush

tools/ds-identify
dermotbradley: That floppy check is only for RHEV (though I though some versions of VMware did/do make use of floppy also).

dermotbradley:
The use of "virt-what" is a bit of a stop-gap, I'm considering writing an alternative (also in Shell) that uses /sys/class/dmi rather than "dmidecode" and that handles containers also (which virt-what doesn't).

holmanb (approved):
This looks good to me, thanks @dermotbradley!

PR #4786, opened 2024-01-18
feat: WSL datasource implementation

CarlosNihelton:
## Proposed Commit Message

```
feat: Implements the WSL datasource

A new datasource is implemented for the WSL platform.
It relies on being able to access the Windows filesystem and execute cmd.exe
to determine where the user data configuration files are.
Reference documentation for that datasource is also added, explaining where
the user should place the config files, the requirements for the datasource to work
and some details that are expected to work differently or not at all.
I'm not sure that list is exhaustive, so we should expect revisions to it in the short future.
For now at least the implementation is focused on locating and ingesting user-data,
without being too smart.

Closes UDENG-1989
```

## Additional Context
Following #4730 , this brings in the Python module with its unit tests implementing the WSL datasource.

Enabling the datasource and integration tests are subject to subsequent PR's, as well as adding a tutorial on how to setup a fresh WSL instance with cloud-init.

This PR brings no inteligence on how to deal with the data supplied, i.e., we ingest whatever user data supplied, whether it may work as intended or not under the WSL platform. While the documentation covers some limitations imposed by the platform, we may soon consider reducing the gaps, such as the discussion we had around setting the hostname. I'd rather have those as separate PR's so we can exhaust the implementation details that would certainly affect not only the datasource.

With those changes plus adding the word "WSL" into a few lists to make cloud-init consider it a viable candidate and exporting an image with the changes already baked in, I could successfully initialize WSL instances with user data stored in my Windows filesystem. The screenshots below depict some scenarios I played with.

image

image

Closes UDENG-1989

## Checklist

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -139,9 +139,7 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
     Return a list of candidate file names that may contain user-data
     in some supported format, ordered by precedence.
     """
-    lsb_rel = util.lsb_release()
-    distribution_id = lsb_rel["id"]
-    release_codename = lsb_rel["codename"]
+    distribution_id, _, release_codename = util.get_linux_distro()
 
     return [
         # WSL instance specific:
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -118,7 +118,7 @@ def win_user_profile_dir() -> PurePath:
             "No output from cmd.exe to show the user profile dir."
         )
 
-    return win_path_2_wsl(home.rstrip())
+    return win_path_2_wsl(home)
 
 
 def machine_id():
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -6,13 +6,15 @@
 """ Datasource to support the Windows Subsystem for Linux platform. """
 
 import logging
+import os
 from pathlib import PurePath
+from typing import List, Optional, cast
 
-from cloudinit import subp
+from cloudinit import sources, subp, util
 
 LOG = logging.getLogger(__name__)
 
-WSLPATH_CMD = "wslpath"
+WSLPATH_CMD = "/usr/bin/wslpath"
 
 
 def wsl_path_2_win(path: str) -> PurePath:
@@ -36,3 +38,86 @@ def instance_name() -> str:
     """
     root_net_path = wsl_path_2_win("/")
     return root_net_path.name
+
+
+def mounted_win_drives() -> List[str]:
+    """
+    Return a list of mount points of the Windows drives inside the current
+    WSL instance, if drives are mounted, or an empty list otherwise
+    """
+    FS_TYPE = "9p"
+    OPTIONS_CONTAIN = "aname=drvfs"
+
+    mounts = util.mounts()
+    mounted = []
+    for mnt in mounts.values():
+        if mnt["fstype"] == FS_TYPE and OPTIONS_CONTAIN in mnt["opts"]:
+            mounted.append(mnt["mountpoint"])
+
+    return mounted
+
+
+def win_path_2_wsl(path: str) -> Optional[PurePath]:
+    """
+    Returns a translation of a Windows path to a Linux path that can be
+    accessed inside the current instance filesystem, or None if failed.
+
+    It requires the Windows drive mounting feature to be enabled and the
+    disk drive must exist for this to succeed.
+
+    Example:
+    # Assuming Windows drives are mounted under /mnt/ and "S:" doesn't exist:
+    p = winpath2wsl("C:\\ProgramData") # p == "/mnt/c/ProgramData/"
+    n = winpath2wsl("S:\\CoolFolder") # n is None (S doesn't exist)
+
+    :param path: string representing a Windows path. The root drive must exist,
+    although the path is not required to.
+    """
+    out, err = subp.subp([WSLPATH_CMD, "-au", path], rcs=[0, 1])
+    if err:
+        LOG.debug(err)
+        return None
+
+    return PurePath(out.rstrip())
+
+
+def cmd_executable() -> Optional[PurePath]:
+    """
+    Returns the Linux path to the Windows host's cmd.exe.
+    """
+
+    mounts = mounted_win_drives()
+    if not mounts:
+        LOG.error("Windows drives are not mounted.")
+        return None
+
+    # cmd.exe path is being stable for decades.
+    candidate = "%s/Windows/System32/cmd.exe"
+    for mnt in mounts:
+        cmd = candidate % mnt
+        if not os.access(cmd, os.X_OK):
+            continue
+
+        LOG.debug("Found cmd.exe at <%s>", cmd)
+        return PurePath(cmd)
+
+    LOG.error("Couldn't find cmd.exe in any mount point.")
+    return None
+
+
+def win_user_profile_dir() -> Optional[PurePath]:
+    """
+    Returns the Windows user profile directory translated as a Linux path
+    accessible inside the current WSL instance.
+    """
+    cmd = cmd_executable()
+    if cmd is None:
+        return None
+
+    home, _ = subp.subp([cmd.as_posix(), "/C", "echo %USERPROFILE%"])
+    home = home.rstrip()
+    if not home:
+        LOG.error("No output from cmd to show the user profile dir.")
+        return None
+
+    return win_path_2_wsl(home.rstrip())
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -132,8 +132,7 @@ def machine_id():
     """
     Returns the local machine ID value from /etc/machine-id.
     """
-    MACHINE_ID_FILE = "/etc/machine-id"
-
+    from cloudinit.settings import MACHINE_ID_FILE
     if util.wait_for_files([MACHINE_ID_FILE], 2.0):
         LOG.debug("%s file not found", MACHINE_ID_FILE)
         return None
@@ -196,8 +195,7 @@ class DataSourceWSL(sources.DataSource):
         # (ref https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity),
         # thus  better prevent it by always relying on case insensitive match.
         existing_files = {
-            ef.name.casefold(): ef.path
-            for ef in os.scandir(seed_dir)
+            ef.name.casefold(): ef.path for ef in os.scandir(seed_dir)
         }
         if not existing_files:
             LOG.warning("%s directory is empty", seed_dir)
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -131,7 +131,7 @@ def machine_id():
         LOG.debug("%s file not found", MACHINE_ID_FILE)
         return None
 
-    return util.load_file(MACHINE_ID_FILE, decode=True)
+    return util.load_binary_file(MACHINE_ID_FILE)
 
 
 def candidate_user_data_file_names(instance_name) -> List[str]:
@@ -209,7 +209,11 @@ class DataSourceWSL(sources.DataSource):
 
         try:
             file = self.find_user_data_file()
-            self.userdata_raw = cast(str, util.load_file(file, decode=True))
+            if file is None:
+                return False
+            self.userdata_raw = cast(
+                str, util.load_binary_file(file.as_posix())
+            )
             return True
         except IOError as err:
             LOG.error("Could not find any user data file: %s", str(err))
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -56,39 +56,34 @@ def mounted_win_drives() -> List[str]:
     return mounted
 
 
-def win_path_2_wsl(path: str) -> Optional[PurePath]:
+def win_path_2_wsl(path: str) -> PurePath:
     """
     Returns a translation of a Windows path to a Linux path that can be
-    accessed inside the current instance filesystem, or None if failed.
+    accessed inside the current instance filesystem.
 
     It requires the Windows drive mounting feature to be enabled and the
-    disk drive must exist for this to succeed.
+    disk drive must be muonted for this to succeed.
 
     Example:
     # Assuming Windows drives are mounted under /mnt/ and "S:" doesn't exist:
     p = winpath2wsl("C:\\ProgramData") # p == "/mnt/c/ProgramData/"
-    n = winpath2wsl("S:\\CoolFolder") # n is None (S doesn't exist)
+    n = winpath2wsl("S:\\CoolFolder") # Exception! S: is not mounted.
 
     :param path: string representing a Windows path. The root drive must exist,
     although the path is not required to.
     """
-    out, err = subp.subp([WSLPATH_CMD, "-au", path], rcs=[0, 1])
-    if err:
-        LOG.debug(err)
-        return None
-
+    out, _ = subp.subp([WSLPATH_CMD, "-au", path])
     return PurePath(out.rstrip())
 
 
-def cmd_executable() -> Optional[PurePath]:
+def cmd_executable() -> PurePath:
     """
     Returns the Linux path to the Windows host's cmd.exe.
     """
 
     mounts = mounted_win_drives()
     if not mounts:
-        LOG.error("Windows drives are not mounted.")
-        return None
+        raise IOError("Windows drives are not mounted.")
 
     # cmd.exe path is being stable for decades.
     candidate = "%s/Windows/System32/cmd.exe"
@@ -100,20 +95,17 @@ def cmd_executable() -> Optional[PurePath]:
         LOG.debug("Found cmd.exe at <%s>", cmd)
         return PurePath(cmd)
 
-    LOG.error(
-        "Couldn't find cmd.exe in any mount point: %s", ", ".join(mounts)
+    raise IOError(
+        "Couldn't find cmd.exe in any mount point: %s" % ", ".join(mounts)
     )
-    return None
 
 
-def win_user_profile_dir() -> Optional[PurePath]:
+def win_user_profile_dir() -> PurePath:
     """
     Returns the Windows user profile directory translated as a Linux path
     accessible inside the current WSL instance.
     """
     cmd = cmd_executable()
-    if cmd is None:
-        return None
 
     # cloud-init runs too early to rely on binfmt to execute Windows binaries.
     # But we know that `/init` is the interpreter, so we can run it directly.
@@ -122,8 +114,9 @@ def win_user_profile_dir() -> Optional[PurePath]:
     home, _ = subp.subp(["/init", cmd.as_posix(), "/C", "echo %USERPROFILE%"])
     home = home.rstrip()
     if not home:
-        LOG.error("No output from cmd to show the user profile dir.")
-        return None
+        raise subp.ProcessExecutionError(
+            "No output from cmd.exe to show the user profile dir."
+        )
 
     return win_path_2_wsl(home.rstrip())
 
@@ -133,6 +126,7 @@ def machine_id():
     Returns the local machine ID value from /etc/machine-id.
     """
     from cloudinit.settings import MACHINE_ID_FILE
+
     if util.wait_for_files([MACHINE_ID_FILE], 2.0):
         LOG.debug("%s file not found", MACHINE_ID_FILE)
         return None
@@ -177,15 +171,9 @@ class DataSourceWSL(sources.DataSource):
         user-data, if any, or None otherwise.
         """
         profile_dir = win_user_profile_dir()
-        if profile_dir is None:
-            LOG.warning(
-                "Cannot proceed without finding the Windows %USERPROFILE% dir."
-            )
-            return None
-
         seed_dir = os.path.join(profile_dir, ".cloud-init")
         if not os.path.isdir(seed_dir):
-            LOG.warning("%s directory doesn't exist.", seed_dir)
+            LOG.error("%s directory doesn't exist.", seed_dir)
             return None
 
         # Notice that by default file name casing is irrelevant here. Windows
@@ -198,7 +186,7 @@ class DataSourceWSL(sources.DataSource):
             ef.name.casefold(): ef.path for ef in os.scandir(seed_dir)
         }
         if not existing_files:
-            LOG.warning("%s directory is empty", seed_dir)
+            LOG.error("%s directory is empty", seed_dir)
             return None
 
         folded_names = [
@@ -209,7 +197,7 @@ class DataSourceWSL(sources.DataSource):
             if filename in existing_files.keys():
                 return PurePath(existing_files[filename])
 
-        LOG.warning(
+        LOG.error(
             "%s doesn't contain any of the expected user-data files", seed_dir
         )
         return None
@@ -227,13 +215,14 @@ class DataSourceWSL(sources.DataSource):
                 self.instance_name, m_id
             )
 
-        file = self.find_user_data_file()
-        if file is None:
-            self.userdata_raw = None
-        else:
+        try:
+            file = self.find_user_data_file()
             self.userdata_raw = cast(str, util.load_file(file, decode=True))
-
-        return True
+            return True
+        except IOError as err:
+            LOG.error("Could not find any user data file: %s", str(err))
+            self.userdata_raw = ""
+            return False
 
 
 # Used to match classes to dependencies
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -235,24 +235,20 @@ class DataSourceWSL(sources.DataSource):
 
     def _get_data(self) -> bool:
         self.vendordata_raw = None
+        self.metadata = dict()
+        seed_dir = cloud_init_data_dir()
         self.instance_name = instance_name()
 
-        self.metadata = dict()
-        m_id = machine_id()
-        if m_id is None:
-            LOG.debug("Instance ID will be the WSL instance name only")
-            self.metadata["instance-id"] = self.instance_name
-        else:
-            self.metadata["instance-id"] = "{}-{}".format(
-                self.instance_name, m_id
-            )
-
         try:
-            file = self.find_user_data_file()
+            self.metadata["instance-id"] = load_metadata_iid(
+                seed_dir, self.instance_name
+            )
+            file = self.find_user_data_file(seed_dir)
             self.userdata_raw = cast(
                 str, util.load_binary_file(file.as_posix())
             )
             return True
+
         except IOError as err:
             LOG.error("Could not find any user data file: %s", str(err))
             self.userdata_raw = ""
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -48,9 +48,8 @@ def mounted_win_drives() -> List[str]:
     FS_TYPE = "9p"
     OPTIONS_CONTAIN = "aname=drvfs"
 
-    mounts = util.mounts()
     mounted = []
-    for mnt in mounts.values():
+    for mnt in util.mounts().values():
         if mnt["fstype"] == FS_TYPE and OPTIONS_CONTAIN in mnt["opts"]:
             mounted.append(mnt["mountpoint"])
 
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -127,19 +127,6 @@ def cloud_init_data_dir() -> PurePath:
     return PurePath(seed_dir)
 
 
-def machine_id():
-    """
-    Returns the local machine ID value from /etc/machine-id.
-    """
-    from cloudinit.settings import MACHINE_ID_FILE
-
-    try:
-        return util.load_binary_file(MACHINE_ID_FILE)
-    except OSError as err:
-        LOG.error(err)
-        return None
-
-
 def candidate_user_data_file_names(instance_name) -> List[str]:
     """
     Return a list of candidate file names that may contain user-data
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -8,7 +8,7 @@
 import logging
 import os
 from pathlib import PurePath
-from typing import List, Optional, cast
+from typing import List, cast
 
 from cloudinit import sources, subp, util
 
@@ -156,7 +156,7 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
 class DataSourceWSL(sources.DataSource):
     dsname = "WSL"
 
-    def find_user_data_file(self) -> Optional[PurePath]:
+    def find_user_data_file(self) -> PurePath:
         """
         Finds the most precendent of the candidate files that may contain
         user-data, if any, or None otherwise.
@@ -164,8 +164,7 @@ class DataSourceWSL(sources.DataSource):
         profile_dir = win_user_profile_dir()
         seed_dir = os.path.join(profile_dir, ".cloud-init")
         if not os.path.isdir(seed_dir):
-            LOG.error("%s directory doesn't exist.", seed_dir)
-            return None
+            raise FileNotFoundError("%s directory doesn't exist." % seed_dir)
 
         # Notice that by default file name casing is irrelevant here. Windows
         # filenames are case insensitive. Even though accessed through Linux,
@@ -177,8 +176,7 @@ class DataSourceWSL(sources.DataSource):
             ef.name.casefold(): ef.path for ef in os.scandir(seed_dir)
         }
         if not existing_files:
-            LOG.error("%s directory is empty", seed_dir)
-            return None
+            raise IOError("%s directory is empty" % seed_dir)
 
         folded_names = [
             f.casefold()
@@ -188,10 +186,9 @@ class DataSourceWSL(sources.DataSource):
             if filename in existing_files.keys():
                 return PurePath(existing_files[filename])
 
-        LOG.error(
-            "%s doesn't contain any of the expected user-data files", seed_dir
+        raise IOError(
+            "%s doesn't contain any of the expected user-data files" % seed_dir
         )
-        return None
 
     def _get_data(self) -> bool:
         self.vendordata_raw = None
@@ -209,8 +206,6 @@ class DataSourceWSL(sources.DataSource):
 
         try:
             file = self.find_user_data_file()
-            if file is None:
-                return False
             self.userdata_raw = cast(
                 str, util.load_binary_file(file.as_posix())
             )
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -5,6 +5,7 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 """ Datasource to support the Windows Subsystem for Linux platform. """
 
+import json
 import logging
 import os
 from pathlib import PurePath
@@ -156,6 +157,35 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
         # generic, valid for all WSL distros and instances.
         "default.user-data",
     ]
+
+
+DEFAULT_INSTANCE_ID = "datasource-wsl"
+
+
+def load_metadata_iid(cloudinitdir: PurePath, instance_name: str) -> str:
+    """
+    Returns the relevant metadata loaded from cloudinit dir based on the
+    instance name
+    """
+    raw = dict()
+    try:
+        raw = json.loads(
+            util.load_binary_file(
+                os.path.join(
+                    cloudinitdir.as_posix(), "%s.meta-data" % instance_name
+                )
+            )
+        )
+
+    except IOError as err:
+        LOG.debug(
+            "Failed to load metadata file from %s for instance %s: %s",
+            cloudinitdir.as_posix(),
+            instance_name,
+            err
+        )
+
+    return raw.get("instance-id", DEFAULT_INSTANCE_ID)
 
 
 class DataSourceWSL(sources.DataSource):
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -5,7 +5,6 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 """ Datasource to support the Windows Subsystem for Linux platform. """
 
-import json
 import logging
 import os
 from pathlib import PurePath
@@ -159,27 +158,17 @@ def load_instance_metadata(cloudinitdir: PurePath, instance_name: str) -> dict:
         cloudinitdir.as_posix(), "%s.meta-data" % instance_name
     )
     try:
-        metadata = json.loads(util.load_binary_file(metadata_path))
-
+        metadata = util.load_yaml(util.load_binary_file(metadata_path))
     except FileNotFoundError:
         LOG.debug(
             "No instance metadata found at %s. Using default instance-id.",
             metadata_path,
         )
-    except json.JSONDecodeError as err:
-        LOG.error(
-            "Unable to setup WSL datasource."
-            " Invalid JSON format found at %s: %s.",
-            metadata_path,
-            err,
-        )
-        raise err
-
-    if "instance-id" not in metadata:
+    if not metadata or "instance-id" not in metadata:
         # Parsed metadata file invalid
         msg = (
             f" Metadata at {metadata_path} does not contain instance-id key."
-            f"Instead received: {metadata}"
+            f" Instead received: {metadata}"
         )
         LOG.error(msg)
         raise ValueError(msg)
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -127,11 +127,11 @@ def machine_id():
     """
     from cloudinit.settings import MACHINE_ID_FILE
 
-    if util.wait_for_files([MACHINE_ID_FILE], 2.0):
-        LOG.debug("%s file not found", MACHINE_ID_FILE)
+    try:
+        return util.load_binary_file(MACHINE_ID_FILE)
+    except OSError as err:
+        LOG.error(err)
         return None
-
-    return util.load_binary_file(MACHINE_ID_FILE)
 
 
 def candidate_user_data_file_names(instance_name) -> List[str]:
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -188,6 +188,9 @@ class DataSourceWSL(sources.DataSource):
             LOG.warning("%s directory doesn't exist.", seed_dir)
             return None
 
+        # Notice that file name casing is irrelevant here. Windows filenames
+        # are case insensitive. Even though accessed through Linux, path
+        # translation just works with whichever casing we try.
         for filename in candidate_user_data_file_names(self.instance_name):
             file = os.path.join(seed_dir, filename)
             if os.path.isfile(file):
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -100,7 +100,9 @@ def cmd_executable() -> Optional[PurePath]:
         LOG.debug("Found cmd.exe at <%s>", cmd)
         return PurePath(cmd)
 
-    LOG.error("Couldn't find cmd.exe in any mount point.")
+    LOG.error(
+        "Couldn't find cmd.exe in any mount point: %s", ", ".join(mounts)
+    )
     return None
 
 
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -146,33 +146,45 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
     ]
 
 
-DEFAULT_INSTANCE_ID = "datasource-wsl"
+DEFAULT_INSTANCE_ID = "iid-datasource-wsl"
 
 
-def load_metadata_iid(cloudinitdir: PurePath, instance_name: str) -> str:
+def load_instance_metadata(cloudinitdir: PurePath, instance_name: str) -> dict:
     """
     Returns the relevant metadata loaded from cloudinit dir based on the
     instance name
     """
-    raw = dict()
+    metadata = {"instance-id": DEFAULT_INSTANCE_ID}
+    metadata_path = os.path.join(
+        cloudinitdir.as_posix(), "%s.meta-data" % instance_name
+    )
     try:
-        raw = json.loads(
-            util.load_binary_file(
-                os.path.join(
-                    cloudinitdir.as_posix(), "%s.meta-data" % instance_name
-                )
-            )
-        )
+        metadata = json.loads(util.load_binary_file(metadata_path))
 
-    except IOError as err:
+    except FileNotFoundError:
         LOG.debug(
-            "Failed to load metadata file from %s for instance %s: %s",
-            cloudinitdir.as_posix(),
-            instance_name,
+            "No instance metadata found at %s. Using default instance-id.",
+            metadata_path,
+        )
+    except json.JSONDecodeError as err:
+        LOG.error(
+            "Unable to setup WSL datasource."
+            " Invalid JSON format found at %s: %s.",
+            metadata_path,
             err,
         )
+        raise err
 
-    return raw.get("instance-id", DEFAULT_INSTANCE_ID)
+    if "instance-id" not in metadata:
+        # Parsed metadata file invalid
+        msg = (
+            f" Metadata at {metadata_path} does not contain instance-id key."
+            f"Instead received: {metadata}"
+        )
+        LOG.error(msg)
+        raise ValueError(msg)
+
+    return metadata
 
 
 class DataSourceWSL(sources.DataSource):
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -187,13 +187,27 @@ class DataSourceWSL(sources.DataSource):
             LOG.warning("%s directory doesn't exist.", seed_dir)
             return None
 
-        # Notice that file name casing is irrelevant here. Windows filenames
-        # are case insensitive. Even though accessed through Linux, path
-        # translation just works with whichever casing we try.
-        for filename in candidate_user_data_file_names(self.instance_name):
-            file = os.path.join(seed_dir, filename)
-            if os.path.isfile(file):
-                return PurePath(file)
+        # Notice that by default file name casing is irrelevant here. Windows
+        # filenames are case insensitive. Even though accessed through Linux,
+        # path translation just works with whichever casing we try.
+        # But users can change that behavior with configuration
+        # (ref https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity),
+        # thus  better prevent it by always relying on case insensitive match.
+        existing_files = {
+            ef.name.casefold(): ef.path
+            for ef in os.scandir(seed_dir)
+        }
+        if not existing_files:
+            LOG.warning("%s directory is empty", seed_dir)
+            return None
+
+        folded_names = [
+            f.casefold()
+            for f in candidate_user_data_file_names(self.instance_name)
+        ]
+        for filename in folded_names:
+            if filename in existing_files.keys():
+                return PurePath(existing_files[filename])
 
         LOG.warning(
             "%s doesn't contain any of the expected user-data files", seed_dir
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -229,7 +229,7 @@ class DataSourceWSL(sources.DataSource):
 
         try:
             metadata = load_instance_metadata(
-                 cloud_init_data_dir(), self.instance_name
+                cloud_init_data_dir(), self.instance_name
             )
             return current == metadata.get("instance-id")
 
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -156,13 +156,6 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
 class DataSourceWSL(sources.DataSource):
     dsname = "WSL"
 
-    def __init__(self, sys_cfg, distro, paths):
-        sources.DataSource.__init__(self, sys_cfg, distro, paths)
-        self._network_config = sources.UNSET
-        self.dsmode = sources.DSMODE_LOCAL
-        self.distro = distro
-        self.instance_name = instance_name()
-
     def find_user_data_file(self) -> Optional[PurePath]:
         """
         Finds the most precendent of the candidate files that may contain
@@ -202,6 +195,7 @@ class DataSourceWSL(sources.DataSource):
 
     def _get_data(self) -> bool:
         self.vendordata_raw = None
+        self.instance_name = instance_name()
 
         self.metadata = dict()
         m_id = machine_id()
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -242,12 +242,11 @@ class DataSourceWSL(sources.DataSource):
 
     def _get_data(self) -> bool:
         self.vendordata_raw = None
-        self.metadata = dict()
         seed_dir = cloud_init_data_dir()
         self.instance_name = instance_name()
 
         try:
-            self.metadata["instance-id"] = load_metadata_iid(
+            self.metadata = load_instance_metadata(
                 seed_dir, self.instance_name
             )
             file = self.find_user_data_file(seed_dir)
@@ -256,9 +255,8 @@ class DataSourceWSL(sources.DataSource):
             )
             return True
 
-        except IOError as err:
-            LOG.error("Could not find any user data file: %s", str(err))
-            self.userdata_raw = ""
+        except (ValueError, IOError) as err:
+            LOG.error("Unable to setup WSL datasource: %s", str(err))
             return False
 
 
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -1,0 +1,38 @@
+# Copyright (C) 2024 Canonical Ltd.
+#
+# Author: Carlos Nihelton <carlos.santanadeoliveira@canonical.com>
+#
+# This file is part of cloud-init. See LICENSE file for license information.
+""" Datasource to support the Windows Subsystem for Linux platform. """
+
+import logging
+from pathlib import PurePath
+
+from cloudinit import subp
+
+LOG = logging.getLogger(__name__)
+
+WSLPATH_CMD = "wslpath"
+
+
+def wsl_path_2_win(path: str) -> PurePath:
+    """
+    Translates a path inside the current WSL instance's filesystem to a
+    Windows accessible path.
+
+    Example:
+    # Running under an instance named "CoolInstance"
+    root = wslpath2win("/") # root == "//wsl.localhost/CoolInstance/"
+
+    :param path: string representing a Linux path, whether existing or not.
+    """
+    out, _ = subp.subp([WSLPATH_CMD, "-am", path])
+    return PurePath(out.rstrip())
+
+
+def instance_name() -> str:
+    """
+    Returns the name of the current WSL instance as seen from outside.
+    """
+    root_net_path = wsl_path_2_win("/")
+    return root_net_path.name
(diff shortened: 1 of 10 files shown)


cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    lsb_rel = util.lsb_release()
+    distribution_id = lsb_rel["id"]
+    release_codename = lsb_rel["codename"]
+
+    return [
+        # WSL instance specific:
+        "%s.user-data" % instance_name,
+        # release codename specific
+        "%s-%s.user-data" % (distribution_id, release_codename),
```
dermotbradley, line 156: Could there be an issue with the "codename" value?

On Alpine Linux (with package lsb-release-minimal installed) "lsb_release -c" returns:

```
Codename:       n/a
```

i.e. there is a "/" in the value returned...

dermotbradley: > The '/' character by itself won't be a problem

I was thinking of whether the "/" might be interpreted as a file path separator.

CarlosNihelton: It will certainly be, but that leads to a file not found, which won't be a catastrophic failure. That composes a candidate filename, not exactly a required file, but rather, one possible file.

The absence of a specific codename would restrict Alpine to either use the instance specific user data file or the distro generic `Allpine-all.user-data` or the most generic `config.user-data`.

If that's an issue, we could use the release numbers instead of the codename, which is more inclusive to the distros, but a little bit more cryptic for the end user, IMHO.

dermotbradley:
@CarlosNihelton

```
At the time of this writing only systemd distros are
supported by the WSL datasource, although there is
nothing hard coded in the implementation code that
requires it, so it's possible that non-systemd
distros find ways to run cloud-init and make it just
work.
```

I maintain Alpine packaging for cloud-init. Any suggestions how a non-systemd-based distro (Alpine uses OpenRC) could achieve this?

I don't have any suitable machines to test this on unfortunately.

CarlosNihelton:
> I maintain Alpine packaging for cloud-init. Any suggestions how a non-systemd-based distro (Alpine uses OpenRC) could achieve this?
> 
> I don't have any suitable machines to test this on unfortunately.

Apparently it's just a matter of commanding WSL itself to run `openRC` via `/etc/wsl.conf` `boot.command` parameter. I'll find some time to test a real Alpine instance to confirm it produces the expected result.

https://wiki.gentoo.org/wiki/Gentoo_in_WSL#Init_systems

cloudinit/sources/DataSourceWSL.py
holmanb: @dermotbradley makes a good point - previously `util.lsb_release()` was only used in `cc_apt_configure`, so we'll want to make sure that this is tested on other distros. 

Additionally, since other distros might not want to add this new dependency (it isn't installed on centos-stream or alpine by default, probably others too), it probably makes more sense to shift this over to the distro python modules over in cloudinit/distro/__init__.py so that distros can customize as they see fit.

holmanb: @CarlosNihelton - This is what I have in mind - it doesn't fix the wsl unittests, but updates the rest of them:
```diff
diff --git a/cloudinit/config/cc_apt_configure.py b/cloudinit/config/cc_apt_configure.py
index 8a01c54c5..465ee8a14 100644
--- a/cloudinit/config/cc_apt_configure.py
+++ b/cloudinit/config/cc_apt_configure.py
@@ -228,7 +228,7 @@ def _should_configure_on_empty_apt():
     return True, "Apt is available."
 
 
-def apply_apt(cfg, cloud):
+def apply_apt(cfg: dict, cloud: Cloud):
     # cfg is the 'apt' top level dictionary already in 'v3' format.
     if not cfg:
         should_config, msg = _should_configure_on_empty_apt()
@@ -238,7 +238,7 @@ def apply_apt(cfg, cloud):
 
     LOG.debug("handling apt config: %s", cfg)
 
-    release = util.lsb_release()["codename"]
+    release = cloud.distro.lsb_release()["codename"]
     arch = util.get_dpkg_architecture()
     mirrors = find_apt_mirror_info(cfg, cloud, arch=arch)
     LOG.debug("Apt Mirror info: %s", mirrors)
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index c857a8737..f16f165dc 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -1248,6 +1248,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             "/bin/true",
         ] + (["-cf", config_file, interface] if config_file else [interface])
 
+    def lsb_release(self) -> dict:
+        return lsb_release()
+
 
 def _apply_hostname_transformations_to_url(url: str, transformations: list):
     """
@@ -1449,3 +1452,32 @@ def uses_systemd():
         return stat.S_ISDIR(res.st_mode)
     except Exception:
         return False
+
+
+def lsb_release() -> dict:
+    fmap = {
+        "Codename": "codename",
+        "Description": "description",
+        "Distributor ID": "id",
+        "Release": "release",
+    }
+
+    data = {}
+    try:
+        out = subp.subp(["lsb_release", "--all"], capture=True)
+        for line in out.stdout.splitlines():
+            fname, _, val = line.partition(":")
+            if fname in fmap:
+                data[fmap[fname]] = val.strip()
+        missing = [k for k in fmap.values() if k not in data]
+        if len(missing):
+            LOG.warning(
+                "Missing fields in lsb_release --all output: %s",
+                ",".join(missing),
+            )
+
+    except subp.ProcessExecutionError as err:
+        LOG.warning("Unable to get lsb_release --all: %s", err)
+        data = dict((v, "UNAVAILABLE") for v in fmap.values())
+
+    return data
diff --git a/cloudinit/sources/DataSourceWSL.py b/cloudinit/sources/DataSourceWSL.py
index 0ed51c3af..100c9e5b4 100644
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -140,27 +140,6 @@ def machine_id():
     return util.load_file(MACHINE_ID_FILE, decode=True)
 
 
-def candidate_user_data_file_names(instance_name) -> List[str]:
-    """
-    Return a list of candidate file names that may contain user-data
-    in some supported format, ordered by precedence.
-    """
-    lsb_rel = util.lsb_release()
-    distribution_id = lsb_rel["id"]
-    release_codename = lsb_rel["codename"]
-
-    return [
-        # WSL instance specific:
-        "%s.user-data" % instance_name,
-        # release codename specific
-        "%s-%s.user-data" % (distribution_id, release_codename),
-        # distribution specific (Alpine, Arch, Fedora, openSUSE, Ubuntu...)
-        "%s-all.user-data" % distribution_id,
-        # generic, valid for all WSL distros and instances.
-        "config.user-data",
-    ]
-
-
 class DataSourceWSL(sources.DataSource):
     dsname = "WSL"
 
@@ -191,7 +170,9 @@ class DataSourceWSL(sources.DataSource):
         # Notice that file name casing is irrelevant here. Windows filenames
         # are case insensitive. Even though accessed through Linux, path
         # translation just works with whichever casing we try.
-        for filename in candidate_user_data_file_names(self.instance_name):
+        for filename in self.candidate_user_data_file_names(
+            self.instance_name
+        ):
             file = os.path.join(seed_dir, filename)
             if os.path.isfile(file):
                 return PurePath(file)
@@ -222,6 +203,26 @@ class DataSourceWSL(sources.DataSource):
 
         return True
 
+    def candidate_user_data_file_names(self, instance_name) -> List[str]:
+        """
+        Return a list of candidate file names that may contain user-data
+        in some supported format, ordered by precedence.
+        """
+        lsb_rel = self.distro.lsb_release()
+        distribution_id = lsb_rel["id"]
+        release_codename = lsb_rel["codename"]
+
+        return [
+            # WSL instance specific:
+            "%s.user-data" % instance_name,
+            # release codename specific
+            "%s-%s.user-data" % (distribution_id, release_codename),
+            # distribution specific (Alpine, Arch, Fedora, openSUSE, Ubuntu...)
+            "%s-all.user-data" % distribution_id,
+            # generic, valid for all WSL distros and instances.
+            "config.user-data",
+        ]
+
 
 # Used to match classes to dependencies
 datasources = [
diff --git a/cloudinit/util.py b/cloudinit/util.py
index 6401a196f..2838e4a29 100644
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -94,36 +94,6 @@ def get_dpkg_architecture():
     return out.stdout.strip()
 
 
-@lru_cache()
-def lsb_release():
-    fmap = {
-        "Codename": "codename",
-        "Description": "description",
-        "Distributor ID": "id",
-        "Release": "release",
-    }
-
-    data = {}
-    try:
-        out = subp.subp(["lsb_release", "--all"], capture=True)
-        for line in out.stdout.splitlines():
-            fname, _, val = line.partition(":")
-            if fname in fmap:
-                data[fmap[fname]] = val.strip()
-        missing = [k for k in fmap.values() if k not in data]
-        if len(missing):
-            LOG.warning(
-                "Missing fields in lsb_release --all output: %s",
-                ",".join(missing),
-            )
-
-    except subp.ProcessExecutionError as err:
-        LOG.warning("Unable to get lsb_release --all: %s", err)
-        data = dict((v, "UNAVAILABLE") for v in fmap.values())
-
-    return data
-
-
 def decode_binary(blob, encoding="utf-8"):
     # Converts a binary type into a text type using given encoding.
     if isinstance(blob, str):
diff --git a/tests/unittests/config/test_apt_configure_sources_list_v1.py b/tests/unittests/config/test_apt_configure_sources_list_v1.py
index 8772625f2..2a98cae69 100644
--- a/tests/unittests/config/test_apt_configure_sources_list_v1.py
+++ b/tests/unittests/config/test_apt_configure_sources_list_v1.py
@@ -89,7 +89,7 @@ class TestAptSourceConfigSourceList:
             subp, "subp", return_value=("PPID   PID", "")
         )
         mocker.patch("cloudinit.config.cc_apt_configure._ensure_dependencies")
-        lsb = mocker.patch("cloudinit.util.lsb_release")
+        lsb = mocker.patch("cloudinit.distros.lsb_release")
         lsb.return_value = {"codename": "fakerelease"}
         m_arch = mocker.patch("cloudinit.util.get_dpkg_architecture")
         m_arch.return_value = "amd64"
diff --git a/tests/unittests/config/test_apt_configure_sources_list_v3.py b/tests/unittests/config/test_apt_configure_sources_list_v3.py
index 260c8a647..65f6b6321 100644
--- a/tests/unittests/config/test_apt_configure_sources_list_v3.py
+++ b/tests/unittests/config/test_apt_configure_sources_list_v3.py
@@ -160,7 +160,7 @@ class TestAptSourceConfigSourceList:
             "subp",
             return_value=("PPID   PID", ""),
         )
-        lsb = mocker.patch("cloudinit.util.lsb_release")
+        lsb = mocker.patch("cloudinit.distros.lsb_release")
         lsb.return_value = {"codename": "fakerel"}
         m_arch = mocker.patch("cloudinit.util.get_dpkg_architecture")
         m_arch.return_value = "amd64"
diff --git a/tests/unittests/config/test_apt_source_v1.py b/tests/unittests/config/test_apt_source_v1.py
index cb0503ace..a1f858c20 100644
--- a/tests/unittests/config/test_apt_source_v1.py
+++ b/tests/unittests/config/test_apt_source_v1.py
@@ -70,7 +70,7 @@ class TestAptSourceConfig:
     @pytest.fixture(autouse=True)
     def common_mocks(self, mocker):
         mocker.patch(
-            "cloudinit.util.lsb_release",
+            "cloudinit.distros.lsb_release",
             return_value={"codename": self.release},
         )
         mocker.patch(
diff --git a/tests/unittests/config/test_apt_source_v3.py b/tests/unittests/config/test_apt_source_v3.py
index e94bc8b8e..eb7c6d0a4 100644
--- a/tests/unittests/config/test_apt_source_v3.py
+++ b/tests/unittests/config/test_apt_source_v3.py
@@ -53,7 +53,7 @@ class TestAptSourceConfig:
     @pytest.fixture(autouse=True)
     def setup(self, mocker, tmpdir):
         mocker.patch(
-            f"{M_PATH}util.lsb_release",
+            "cloudinit.distros.lsb_release",
             return_value=MOCK_LSB_RELEASE_DATA.copy(),
         )
         mocker.patch(f"{M_PATH}_ensure_dependencies")
```

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def find_user_data_file(self) -> Optional[PurePath]:
+        """
+        Finds the most precendent of the candidate files that may contain
+        user-data, if any, or None otherwise.
+        """
+        profile_dir = win_user_profile_dir()
+        if profile_dir is None:
+            LOG.warning(
+                "Cannot proceed without finding the Windows %USERPROFILE% dir."
```
blackboxsw, line 182: If we cannot proceed without this dir, shouldn't this be an error instead of warning?

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+            )
+            return None
+
+        seed_dir = os.path.join(profile_dir, ".cloud-init")
+        if not os.path.isdir(seed_dir):
+            LOG.warning("%s directory doesn't exist.", seed_dir)
```
blackboxsw, line 188: Same, should this be error instead of warning? Or are there conditions that you want to support where the profile_dir does exist, but the seed_dir doesn't exist because there is no applicable user-data for some reason?

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+        # are case insensitive. Even though accessed through Linux, path
+        # translation just works with whichever casing we try.
+        for filename in candidate_user_data_file_names(self.instance_name):
+            file = os.path.join(seed_dir, filename)
+            if os.path.isfile(file):
+                return PurePath(file)
+
+        LOG.warning(
+            "%s doesn't contain any of the expected user-data files", seed_dir
+        )
```
blackboxsw, line 201: Since we require that at least one user-data file exists which meets the launch requirements for this instance, this should be an error, and the error message should probably state that a user-data file was that isn't present is something like
`%USERPROFILE%\.cloud-init\config.user-data`.  

Most clouds support an empty user-data file and cloud-init handles that fine because user-data content is optional, but the structure of valid metadata service configuration requires that we at least 

If possible, ./tools/ds-identify should probably also validate that a top-level `%USERPROFILE%\.cloud-init\config.user-data`  exists on the system if we can check that early enough in boot. Without it, we can't officially detect WSL datasource with the python get_data call.

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,173 @@
...
+   22.04 Jammy Jellyfish image, if a more specific configuration file does not
+   match.
+
+3. ``%USERPROFILE%\.cloud-init\<ID>-all.user-data`` for the distro specific
+   configuration, matched by the distro ID entry in ``/etc/os-release``,
+   regardless of the release version. Example: ``debian-all.user-data`` will
+   affect any instance created from any Debian GNU/Linux image, regardless of
+   which release, if a more specific configuration file does not match.
+
+4. ``%USERPROFILE%\.cloud-init\config.user-data`` for the configuration
```
blackboxsw, line 81: When WSL interoperability and automount is enabled, will the WSL agent provide this provide this empty clonfig.user-data file by default to all USERPROFILE? Or, does the user always need to set this default up so prior to using cloud-init?

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: What if we just replace `util.lsb_release()` by `util.get_linux_distro()` (https://github.com/canonical/cloud-init/blob/main/cloudinit/util.py#L582) ?
That function is more distro-agnostic, doesn't imply dependencies and provides all the information this datasource needs.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+
+    Example:
+    # Assuming Windows drives are mounted under /mnt/ and "S:" doesn't exist:
+    p = winpath2wsl("C:\\ProgramData") # p == "/mnt/c/ProgramData/"
+    n = winpath2wsl("S:\\CoolFolder") # n is None (S doesn't exist)
+
+    :param path: string representing a Windows path. The root drive must exist,
+    although the path is not required to.
+    """
+    out, err = subp.subp([WSLPATH_CMD, "-au", path], rcs=[0, 1])
```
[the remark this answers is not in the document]
CarlosNihelton: That binary only fails if the Windows disk drive is not mounted (which is also the case if it doesn't exist as well).

```sh
❯ wslpath -au 'c:\Users'
/mnt/c/Users
❯ echo $?
0
❯ wslpath -au 'K:\Users'
wslpath: K:\Users
❯ echo $?
1
❯ wslpath -au 'C:\ThisDoesntExist'
/mnt/c/ThisDoesntExist
❯ echo $?
0
❯ wslpath -au 'K:\ThisDoesntExist'
wslpath: K:\ThisDoesntExist
❯ echo $?
1
❯ umount /mnt/c
❯ wslpath -au 'C:\ThisDoesntExist'
wslpath: C:\ThisDoesntExist
❯ echo $?
1
```

Currently we only call it to translate the contents of the %USERPROFILE% Windows user environment variable to a Linux path. If correctly set this should never fail, but there is a remote possibility of the user misconfiguring that environment variable to a non-existing disk, which would lead to the error we are foreseeing here.

But, thinking better, being the above mentioned a remote possibility, I think it's fine to consider that exceptional, thus remove the `rcs` parameter and deal with exceptions in the data source. That also prevents backtraces in the future if that tool adds more failure modes.

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: Indeed.

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: The other specification we talked about may add more places to look for user-data, but you're right. Let's not anticipate the future here and consider this an error.

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: Instead of relying on an always present user data file, we could just assign the empty string to `self.userdata_raw` if a file is not found. Wouldn't that have the same effect?

Since the datasource only craws user data, if there is none, there is nothing for this datasource to do. Ideally we should really do nothing, not even DataSourceNone should play any role under such situation. Is there a good way to create that behavior?

cloudinit/sources/DataSourceWSL.py
holmanb: That should work.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    cmd = cmd_executable()
+    if cmd is None:
+        return None
+
+    # cloud-init runs too early to rely on binfmt to execute Windows binaries.
+    # But we know that `/init` is the interpreter, so we can run it directly.
+    # See /proc/sys/fs/binfmt_misc/WSLInterop[-late]
+    # inside any WSL instance for more details.
+    home, _ = subp.subp(["/init", cmd.as_posix(), "/C", "echo %USERPROFILE%"])
```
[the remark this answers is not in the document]
CarlosNihelton: Exception being handled in the upper level: datasource's `_get_data()` method.

CarlosNihelton:
> > I maintain Alpine packaging for cloud-init. Any suggestions how a non-systemd-based distro (Alpine uses OpenRC) could achieve this?
> > I don't have any suitable machines to test this on unfortunately.
> 
> Apparently it's just a matter of commanding WSL itself to run `openRC` via `/etc/wsl.conf` `boot.command` parameter. I'll find some time to test a real Alpine instance to confirm it produces the expected result.
> 
> https://wiki.gentoo.org/wiki/Gentoo_in_WSL#Init_systems

@dermotbradley indeed adding the boot.command to start OpenRC did work once I added cloud-init to the default run-level.
Some issues happened, though:
- The datasource checking assumed the datasource to match must have DEP_FILESYSTEM and DEP_NETWORK. I don't know if that's because Alpine doesn't rely on ds-identify. For Ubuntu I got only DEP_FILESYSTEM, which is what I'd expect.
- Forcing the datasource to include DEP_NETWORK to make it be chosen under Alpine reveals that running openRC breaks some WSL features, such as the `wslpath` binary now missing.
- SSH host keygen fails.
- Attempting to run it manually with `cloud-init init` breaks very early on missing `serial` Python module.

My understanding is that there is some distro work to be done before being able to integrate cloud-init running as a service on Alpine on the WSL platform, but that should be doable without much changes in the datasource itself.

dermotbradley:
> Some issues happened, though:
> 
> * The datasource checking assumed the datasource to match must have DEP_FILESYSTEM and DEP_NETWORK. I don't know if that's because Alpine doesn't rely on ds-identify.


Alpine doesn't currently rely/use ds-identify. I had a PR #4785 merged recently to add Alpine support to ds-identify but it requires further work (Alpine doesn't have systemd-detect-virt and virt-what is only a partial solution - it doesn't detect WSL - so I'm working on an alternative).

> * Forcing the datasource to include DEP_NETWORK to make it be chosen under Alpine reveals that running openRC breaks some WSL features, such as the `wslpath` binary now missing.

Doesn't that require 9p to be setup?

> * SSH host keygen fails.

Fails how exactly? Have you installed the Alpine openssh-keygen package? It is not a dep for cloud-init package as some people may want to use cloud-init without SSH (The package's [README](https://git.alpinelinux.org/aports/tree/community/cloud-init/README.Alpine#n130) file talks about this), i.e. via serial/hypervisor console only.

> * Attempting to run it manually with `cloud-init init` breaks very early on missing `serial` Python module.

I'm assuming you're running Alpine with the Alpine cloud-init package and then you've replaced that with cloud-init git master code...and so some of the packaged Alpine config may have been lost - the Alpine cloud-init package provides a [modified](https://git.alpinelinux.org/aports/tree/community/cloud-init/04-set-default-datasource-list.patch) /etc/config/cloud.cfg.tmpl that disables 2 DataSources that rely on py3-serial, this modification has likely been lost when you put cloud-init git main files in place.

> My understanding is that there is some distro work to be done before being able to integrate cloud-init running as a service on Alpine on the WSL platform, but that should be doable without much changes in the datasource itself.

Well basically either ds-identify is enabled and it is able to identify WSL (requires my detection code to be written, packaged for Alpine, and listed as a package dependency), or else for WSL to be present in the active DataSources list that the Alpine package defaults to.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class DataSourceWSL(sources.DataSource):
+    dsname = "WSL"
+
+    def __init__(self, sys_cfg, distro, paths):
+        sources.DataSource.__init__(self, sys_cfg, distro, paths)
+        self._network_config = sources.UNSET
+        self.dsmode = sources.DSMODE_LOCAL
+        self.distro = distro
```
blackboxsw, line 163: I don't think we need/want any of these attributes set in __init__for a couple reasons:
1. _network_config is only referenced in datasources that try to cache the read network config and override the DataSource.network_config property to return the cached value instead of recalculating it
2. dsmode is unused in this DataSource and was used only in other datasources to print specific information about what bode the datasource was detected in. Since this datasource is only detected in local stage per the `datasources: ....WSL, sources.DEP_FILESYSTEM` registration at the bottom, we don't really care about tracking and reporting the dsmode in any string representation of this class.
3. self.distro is set in the superclass `__init__` anyway
4. We probably don't want self.instance_id setup in __init__ because next reboot of this instance will leverage the cached datasource in /var/lib/cloud/instance/obj.pkl and will not re-run __init__ when deserializing the pickled datasource cache. So, if the instance name changed across reboot, the _get_data calls will be using the old instance name instead if the instance name has the ability to change across reboots.


I think we should drop the __init_ from DataSourceWSL and also push self.instance_name calculdation down into `_get_data()` which will be called every boot.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+        LOG.error(
+            "%s doesn't contain any of the expected user-data files", seed_dir
+        )
+        return None
+
+    def _get_data(self) -> bool:
+        self.vendordata_raw = None
+
+        self.metadata = dict()
+        m_id = machine_id()
```
blackboxsw, line 207: So, if someone removes or clears /etc/machine-id  or sets it to `uninitialized` triggering a recreation of /etc/machine-id on next boot, cloud-init will re-run and re-apply all user-data discovered on this machine, regenerating SSH host keys etc.   I think this is reasonable, but I wanted to call it out in case we have concerns about a /etc/machine-id manipulation forcing cloud-init to re-run on next boot. @holmanb @TheRealFalcon any concerns about this?

doc/rtd/reference/datasources/wsl.rst
blackboxsw: We may want to just rename this ``%USERPROFILE%\.cloud-init\all.user-data``  to also align with the ``<DISTRO_ID>-all.user-data`` encoded filename above or maybe ``%USERPROFILE%\.cloud-init\default.user-data``.  It seems strange to pivot from ``<DISTRO_ID>-all.user-data`` to ``config.user-data`` as the default.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+        if m_id is None:
+            LOG.debug("Instance ID will be the WSL instance name only")
+            self.metadata["instance-id"] = self.instance_name
+        else:
+            self.metadata["instance-id"] = "{}-{}".format(
+                self.instance_name, m_id
+            )
+
+        try:
+            file = self.find_user_data_file()
```
blackboxsw, line 217: [Bump on my other comment in docs](https://github.com/canonical/cloud-init/pull/4786/files#r1462495371). Since it is a hard requirement that the WSL datasource has to minimally find at least one user-data file in order to use WSL images, I think the datasource documentation needs to express a default `config.user-data`  or `default.user-data` or `all.user-data` file needs to be present in order for cloud-init to be enabled on boot.

blackboxsw (changes requested):
Thanks @CarlosNihelton for the patience here. A few more nits and a bikeshed question about whether we should name the top-level default cloud-config file `default.user-data` or `all.user-data` instead of `config.user-data`. A couple of requests for docs and dropping the DataSourceWSL.__init__ method as it seems unnecessary.

cloudinit/sources/DataSourceWSL.py
blackboxsw: > Instead of relying on an always present user data file, we could just assign the empty string to self.userdata_raw if a file is not found. Wouldn't that have the same effect?

Sorry I missed this response. I think it is a good constraint/restriction to place on the WSL datasource to only activate/discover if it sees some applicable user-data file. This would allow customizatation of some environments that may the ability to deactivate cloud-init on a WSL image stream which happens to have cloud-init installed (and configured to potentially use the WSL datasource). This would then behave like our nocloud datasource, if some expected NoCloud user-data file doesn't exist in /var/lib/cloud/seed/nocloud/user-data, then cloud-init won't try to discover that datasource and configure the machine. There may be use-cases we haven't thought of where external to the WSL image we want a mechanism to leave cloud-init inert, and the absence of any applicable user-data file would be the easiest knob to expose to the host windows system/environment.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+    return win_path_2_wsl(home.rstrip())
+
+
+def machine_id():
+    """
+    Returns the local machine ID value from /etc/machine-id.
+    """
+    from cloudinit.settings import MACHINE_ID_FILE
+
+    if util.wait_for_files([MACHINE_ID_FILE], 2.0):
```
holmanb, line 130: We can't just wait for the existence of this file - it might say "uninitialized" and not yet have a unique id. That means that a subsequent boot would unexpectedly re-run cloud-init.

cloudinit/sources/DataSourceWSL.py
holmanb: > if someone removes or clears /etc/machine-id or sets it to uninitialized triggering a recreation of /etc/machine-id on next boot, cloud-init will re-run and re-apply all user-data discovered on this machine, regenerating SSH host keys etc. 

Machine-id changing is not a random event. It is an intentional act by an image builder or admin to say "treat this like a new instance boot - do the first boot stuffs". I think that this is exactly what we want cloud-init to do when machine-id changes - regardless of datasource.

From [man machine-id](https://www.freedesktop.org/software/systemd/man/latest/machine-id.html#)

> For operating system images which are created once and used on multiple machines, for example for containers or in the cloud, /etc/machine-id should be either missing or an empty file in the generic file system image (the difference between the two options is described under "First Boot Semantics" below). An ID will be generated during boot and saved to this file if possible

Getting instance id from the network is [known to be problematic and flaky](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/1885527) (when were networks ever reliable?). A local file would be more reliable than instance-id.

I'm not saying we should replace instance-id with machine-id (at least right now), but I am saying that I think that if machine-id has changed, that cloud-init should probably treat it as first boot - regardless of whether or not the cloud thinks the instance id has changed.

cloudinit/sources/DataSourceWSL.py
holmanb: To work reliably, we should probably order cloud-init after [first-boot-complete.target](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#first-boot-complete.target).

holmanb (changes requested):

cloudinit/sources/DataSourceWSL.py
holmanb: This also raises a point about init systems - `/etc/machine-id` isn't going to be created by Alpine, for example, since it doesn't use systemd. This should really be implemented in the base distro class in `cloudinit/distros/__init__.py` so that other distros can override behavior in a way that makes sense to them.

dermotbradley: @holmanb

> This also raises a point about init systems - /etc/machine-id isn't going to be created by Alpine, for example, since it doesn't use systemd.

It will be created on Alpine **if** the machine-id init.d service is enabled and runs before the DataSource runs: https://git.alpinelinux.org/aports/tree/main/openrc/machine-id.initd

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: Let's use the version numbers instead. The Alpine example from above would look too cryptic IMHO: `Alpine-Alpine_Linux_v3.19.user-data` (Rocky seems OK: `Rocky-Blue_Onyx.user-data`).

doc/rtd/reference/datasources/wsl.rst
CarlosNihelton: `%USERPROFILE%\.cloud-init\default.user-data` looks better. I'm going with that suggestion.

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: Must that ordering be specific to WSL? I mean, for other use cases under systemd, would it be fine to have that ordering applied?

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: Gotcha! Unfortunately we'll need to keep some logic duplicated in between ds-identify and the datasource until all distros support the first, but that's fine.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,234 @@
...
+        LOG.debug("%s file not found", MACHINE_ID_FILE)
+        return None
+
+    return util.load_file(MACHINE_ID_FILE, decode=True)
```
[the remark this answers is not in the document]
CarlosNihelton: I rebased over the latest main due that. New commits start at edde2df14157f9a7d72ff5a686fc7aedf7a90a45.

cloudinit/sources/DataSourceWSL.py
blackboxsw: I'd be hesitant adding general systemd ordering changes in order to support this work without us providing a more thorough specification for review to get input on the impacts on other distros/clouds when /etc/machine-id changes  as well as impacts to boot speed and boot ordering cycles on any systemd environment we currently support. Maybe it makes sense to drop this reading of /etc/machine-id as instance-id from WSL, since WSL already has a unique identifier per user which should be the instance name.

The assertions I'd like confirmation on:
1.  A single user cannot have two WSL instances with the same instance name
2. If a WSL instance or snapshot or image gets its name changed, we want cloud-init to re-run
3. The only way to trigger cloud-init to re-run and apply user-data again would be to rename the instance and reboot

If the above are true, and we don't expect a use case where WSL instances to would need trigger cloud-init to re-run on the same instance name, I don't think we need to build in a dependency on /etc/machine-id for this datasource

blackboxsw:
> I'm not saying we should replace cloud-init's concept of an instance-id with machine-id (at least right now), but I am saying that I think that if machine-id has changed, that cloud-init should probably treat it as first boot regardless of whether or not the cloud thinks the instance id has changed. I think that we should listen to the user if they want to do first boot things, and since we integrate with the init system and the init system has a way of communicating to services whether or not to do first boot things (which is based on this file), we should respect the user interface and do what the user expects.

I agree that long-term vision makes sense. But, I don't think it's warranted for this WSL datasource as it feels generally like WSL instances use-case we are trying to solve here feels more like a one-time launch with static user-data. If there are config changes that need applying via cloud-init my understanding is that those WSL instance would be re-create that WSL instance from scratch instead of relying on a changed /etc/machine-id.  

@CarlosNihelton I had a thought/concern though, if our persisted instance-id is just the instance_name "dog", and that instance is snapshotted and launched by another user who also calls the instance "dog", cloud-init may not have enough information to know that it's a brand new instance in a different user's account.

cloudinit/sources/DataSourceWSL.py
CarlosNihelton: > 1. A single user cannot have two WSL instances with the same instance name

That's true.
image

> 2. If a WSL instance or snapshot or image gets its name changed, we want cloud-init to re-run

There is no convenient interface at the moment to rename an instance (either from inside or outside), but that's possible through the Windows registry:

image

Having the datasource recomputing and comparing the id with the cached value can ensure we react to the instance name changing.

> 3. The only way to trigger cloud-init to re-run and apply user-data again would be to rename the instance and reboot

The discussion around `/etc/machine-id` shows that it's quite brittle to rely on it, since it's not universaly available for all Linux distros and the timing around its generation might also be an issue.

So, @blackboxsw and myself brainstormed about this and come to a conclusion that the machine ID must come from the outside, being dependent on:
1. **The Windows host**. An image exported from a computer and imported in another computer would detect it's a new fresh instance and trigger cloud-init to re-run.
2. **The Windows user**. An image exported to a location in disk where other users can access and create new instances imported from that image wouldn't fool cloud-init, i.e. it would detect those new instances as new and re-run.
3. **Some internal knowledge WSL has about the instance**. A user could export a snapshot of a distro instance, unregister it - delete the instance - and reimport the instance under the same name. None of the pieces of information above would be enough to detect that it's a new instance. But WSL internally stores UUID's in the Windows registry. Even in the case mentioned above the new instance receives a new UUID. We cannot ensure how unique is that UUID in a broader scope, but that value combined with the other pieces of information already mentioned allows us to compute a quite unique ID for the instance.

So, besides implementing a method `check_instance_id()` for this datasource, I need to compute the that ID differently, by launching some nasty Windows commands via `subp` to acquire the following information:

1. **The Windows host**: `powershell.exe -Command '$(Get-WmiObject -Class "Win32_ComputerSystemProduct").UUID'` => UUID tied to the Windows hardware
2. **The Windows user**: `powershell.exe -Command 'whoami.exe'` => we're not gonna store it, but hash it soon
3. **Some internal knowledge WSL has about the instance**: `powershell.xe -Command "reg.exe query 'HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Lxss\'" => We can find the UUID for the current instance looking inside this registry location.

Since the user can disable access to those binaries anytime after the first boot, if the datasource fails to compute the current instance ID due lack of access, we'd assume nothing has changed.

I think that summarizes the brainstorming we did.

CarlosNihelton:
> @CarlosNihelton I had a thought/concern though, if our persisted instance-id is just the instance_name "dog", and that instance is snapshotted and launched by another user who also calls the instance "dog", cloud-init may not have enough information to know that it's a brand new instance in a different user's account.

I think this comment summarizes it:
https://github.com/canonical/cloud-init/pull/4786#discussion_r1476692586

cloudinit/sources/DataSourceWSL.py
holmanb: @CarlosNihelton Great investigation - thanks for looking into the options.

After our discussion today about using the instance name and instance-id provided by the metadata file, @blackboxsw and @TheRealFalcon  and I talked a little more. Since it seems rather arbitrary that instance name change would retrigger cloud-init, I think we're leaning towards using a known default instance id ([as in the case of nocloud](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceNoCloud.py#L53)) unless the user provides one in meta-data.

This solution should be very simple, and hopefully more robust than calling into powershell and more obvious to users of cloud-init than using instance name as a trigger for whether cloud-init should run.

blackboxsw: Thanks @CarlosNihelton for the continuation of this discussion. You referenced in another discussion that if cloud-init relies on powershell functionality such as Get_WmiObject, whoami or reg.exe, a user could disable access to any of those commands on the host, When that config visibility is disabled, it would leave cloud-init with the inability to determine instance-id on those machines. Since that configuration change is security-related and may become policy for an organization, we determined it is best to avoid such commands as they may be ineffective in some configurations leaving cloud-init in a state where cannot determine instance-id across a reboot boundary.

As a result, we decided the WSL datasource can instead rely on setting a default `instance-id: datasource-wsl` in metadata, but allow for reading an optional `<instance-name>`.meta-data file on the Windows host can be provided alongside the applicable `.user-data`. If `<instance-name>.meta-data` exists, it could contain a JSON dicttionary optionally providing a some value for instance id such as:  `'{"instance-id": "x-y-z"}'`. As with other cloud platforms, that optional meta-data file can change the value of instance-id to trigger cloud-init to re-run across reboot. If users are looking to share snapshots or relaunch a snapshot where cloud-init is re-triggered, they would either run `sudo cloud-init clean --logs`  on the instance before snapshot/export, or create the appropriate <new-instance-name>.meta-data file containing  `'{"instance-id": "some-new-instance-id"}'`.

blackboxsw (changes requested):
Thank you @CarlosNihelton.  A number of inline comments and some refactor suggestions.

Minimally, I think we need to:
1.  ensure the entire instance metadata file, when present, is assigned to self.metadata on this instance as the DataSource parent has helper methods that react to instance-id, and other optional keys that child classes don't have to be aware of explicitly.  
2. I also think we need more strict format/error messaging around the metadata file because this is potentially crafted by a person and typically error-prone.

holmanb:
@s-makin could you please look at `doc/rtd/reference/datasources/wsl.rst` in this PR?

blackboxsw:
FWIW @holmanb @CarlosNihelton here's the consolidated functional diff I was thinking about for this datasource capturing any inline review comments I had above
```diff
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -146,33 +146,48 @@ def candidate_user_data_file_names(instance_name) -> List[str]:
     ]
 
 
-DEFAULT_INSTANCE_ID = "datasource-wsl"
+DEFAULT_INSTANCE_ID = "iid-datasource-wsl"
 
 
-def load_metadata_iid(cloudinitdir: PurePath, instance_name: str) -> str:
+def load_instance_metadata(cloudinitdir: PurePath, instance_name: str) -> dict:
     """
     Returns the relevant metadata loaded from cloudinit dir based on the
     instance name
     """
-    raw = dict()
+    metadata = {"instance-id": DEFAULT_INSTANCE_ID}
+    metadata_path = os.path.join(
+        cloudinitdir.as_posix(), "%s.meta-data" % instance_name
+    )
     try:
-        raw = json.loads(
+        metadata = json.loads(
             util.load_binary_file(
                 os.path.join(
                     cloudinitdir.as_posix(), "%s.meta-data" % instance_name
                 )
             )
         )
-
-    except IOError as err:
+    except FileNotFoundError as err:
         LOG.debug(
-            "Failed to load metadata file from %s for instance %s: %s",
-            cloudinitdir.as_posix(),
-            instance_name,
+            "No instance metadata found at %s. Using default instance-id",
+            metadata_path,
+        )
+    except json.JSONDecodeError as err:
+        LOG.error(
+            "Unable to setup WSL datasource. Invalid JSON format found at %s: %s.",
+            metadata_path,
             err,
         )
+        raise err
 
-    return raw.get("instance-id", DEFAULT_INSTANCE_ID)
+    if "instance-id" not in metadata:
+        # Parsed metadata file invalid
+        msg = (
+            f" Metadata at {metadata_path} does not contain instance-id key."
+            f"Instead received: {metadata}"
+        )
+        LOG.error(msg)
+        raise ValueError(msg)
+    return metadata
 
 
 class DataSourceWSL(sources.DataSource):
@@ -208,19 +223,6 @@ class DataSourceWSL(sources.DataSource):
             "%s doesn't contain any of the expected user-data files" % seed_dir
         )
 
-    def check_instance_id(self, sys_cfg):
-        current = self.get_instance_id()
-        if not current:
-            return None
-
-        try:
-            return current == load_metadata_iid(
-                cloud_init_data_dir(), self.instance_name
-            )
-        except IOError as err:
-            LOG.error("Could not load updated instance ID: %s", err)
-            return None
-
     def _get_data(self) -> bool:
         self.vendordata_raw = None
         self.metadata = dict()
@@ -228,7 +230,7 @@ class DataSourceWSL(sources.DataSource):
         self.instance_name = instance_name()
 
         try:
-            self.metadata["instance-id"] = load_metadata_iid(
+            self.metadata = load_instance_metadata(
                 seed_dir, self.instance_name
             )
             file = self.find_user_data_file(seed_dir)
@@ -237,9 +239,8 @@ class DataSourceWSL(sources.DataSource):
             )
             return True
 
-        except IOError as err:
-            LOG.error("Could not find any user data file: %s", str(err))
-            self.userdata_raw = ""
+        except (ValueError, IOError) as err:
+            LOG.error("Unable to setup WSL datasource: %s", str(err))
             return False
 
```

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    Returns the local machine ID value from /etc/machine-id.
+    """
+    from cloudinit.settings import MACHINE_ID_FILE
+
+    try:
+        return util.load_binary_file(MACHINE_ID_FILE)
+    except OSError as err:
+        LOG.error(err)
+        return None
```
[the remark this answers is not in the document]
CarlosNihelton: This part was already removed, but there were leftovers from promoting the MACHINE_ID_FILE constant which I just reset back to the original state.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,254 @@
...
+        ]
+        for filename in folded_names:
+            if filename in existing_files.keys():
+                return PurePath(existing_files[filename])
+
+        raise IOError(
+            "%s doesn't contain any of the expected user-data files" % seed_dir
+        )
+
+    def check_instance_id(self, sys_cfg):
```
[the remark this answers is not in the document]
CarlosNihelton: Ok! That approach makes more sense to me based on what I saw in other datasources.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,254 @@
...
+    def _get_data(self) -> bool:
+        self.vendordata_raw = None
+        self.metadata = dict()
+        seed_dir = cloud_init_data_dir()
+        self.instance_name = instance_name()
+
+        try:
+            self.metadata["instance-id"] = load_metadata_iid(
+                seed_dir, self.instance_name
+            )
```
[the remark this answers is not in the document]
CarlosNihelton: I was hesitant to allow all sorts of metadata relying on the base class behavior and figuring out some of that is not applicable under the WSL environment, so instead of blocklisting the unsuitable bits I'd though we should just allow-list the parts we know for sure to work. But I think you convinced me the proper way to deal with this situation  is by figuring out the usage in the wild and adapting the unwanted behavior as we find them.

CarlosNihelton:
> FWIW @holmanb @CarlosNihelton here's the consolidated functional diff I was thinking about for this datasource capturing any inline review comments I had above
> 

Thanks for summarizing it here. It became much easier to track the train-of-thought than by following the inline comments.
I think I addressed it all, except for the usage of  `seed_dir` in the exception handler of the `check_instance_id` method that would be unbound in there (setting it should happen inside the try-block because the function `cloudinit_data_dir()` could raise exceptions). The potential exceptions raised by the lower levels report the affected paths if they were to blame.

cloudinit/sources/DataSourceWSL.py
```diff
@@ -0,0 +1,271 @@
...
+    """
+    Returns the relevant metadata loaded from cloudinit dir based on the
+    instance name
+    """
+    metadata = {"instance-id": DEFAULT_INSTANCE_ID}
+    metadata_path = os.path.join(
+        cloudinitdir.as_posix(), "%s.meta-data" % instance_name
+    )
+    try:
+        metadata = json.loads(util.load_binary_file(metadata_path))
```
blackboxsw: In further discussion with the team, we came to the conclusion that we don't want to force JSON on this file format, but instead rely on YAML format like DataSourceNoCloud.
YAML also has the benefit  of not requiring as many characters to express the simple dictionary values that is probably less prone to errors as well:
`instance-id: some-uuid`

@CarlosNihelton sorry for thrashing on this. 
Can we instead use `util.load_yaml?


Also let's get some unittest coverage on load_instance_metadata. We'll need to update docs for the meta-data file to simplify the value to `instance-id: some-uuid` from the JSON equivalent.

I'll push this suggested diff for review on this branch to short-circuit the work on your end if you agree.

tests/unittests/sources/test_wsl.py
```diff
@@ -0,0 +1,321 @@
...
+    "hugetblfs": {
+        "fstype": "hugetblfs",
+        "mountpoint": "/dev/hugepages",
+        "opts": "rw,relatime...",
+    },
+}
+SAMPLE_LINUX_DISTRO = ("ubuntu", "24.04", "noble")
+
+
+class TestWSLHelperFunctions(CiTestCase):
```
blackboxsw: As a project, we are trying to move away from CiTestCase wherever possible to make use of pytest constructs, fixtures and features. As such I've adapted your unittests here to move away from the unittest-based self.assert* and tmp directory setup. I have also added unittest coverage for the read_instance_metadata function.

blackboxsw (commented):
@CarlosNihelton looks really good thanks. Only concerns I think are requiring meta-data file to be YAML instead of json for easy syntax and alignment with other datasources and deriving from CiTestCases. I've taken the liberty to push the latest review commit to this branch to ease review. Please take a look at let us know what you think.

Thanks!

tools/ds-identify
```diff
@@ -1601,12 +1623,39 @@ dscheck_WSL() {
...
+    if [ -z "$cloudinitdir" ]; then
+        debug 1 "%USERPROFILE/.cloud-init/ directory not found"
+        return "${DS_NOT_FOUND}"
+    fi
+
+    # and the applicable userdata file. Notice the ordering in the for-loop
+    # must match our expected precedence, so the file we find is what the
+    # datasource must process.
+    instance_name="$(WSL_instance_name)"
+    distro_id=$(grep "^ID=" "${PATH_ROOT}/etc/os-release" | cut -d'=' -f2 )
```
holmanb, line 1645: Typically we try to avoid parsing using external tools like grep / cut in favor of using built-in operations.

holmanb (changes requested):

tests/unittests/sources/test_wsl.py
```diff
@@ -132,15 +129,87 @@ def test_candidate_files(self, m_gld):
...
+                ["Failed loading yaml blob. Invalid format at line 1"],
+                "",
+                id="error_on_md_invalid_yaml",
+            ),
+        ),
+    )
+    def test_load_instance_metadata(
+        self, md_content, raises, errors, warnings, md_expected, tmpdir, caplog
+    ):
+        """meta-data file is optional. Errors are raised on ivalid content."""
```
CarlosNihelton, line 181:
```suggestion
        """meta-data file is optional. Errors are raised on invalid content."""
```

CarlosNihelton (commented):
@blackboxsw your changes make sense to me, specially replacing JSON with YAML. Thanks for the work around CiTestCase, I was not aware of such movement.

I just spotted one minor typo which I'll address at next.

tools/ds-identify
blackboxsw: Maybe we'd want to `source /etc/os-release` and use `{$ID}` and `${VERSION_ID}`

holmanb: >  some versions of redhat which will only contain /etc/redhat-release

@blackboxsw Which versions do you mean? I haven't seen a RHEL or derivative without it before - it is required by systemd last time I checked.

> Maybe we'd want to source /etc/os-release and use {$ID} and ${VERSION_ID}

From `man 5 os-release`:

       The format of os-release is a newline-separated list of environment-like shell-compatible
       variable assignments. It is possible to source the configuration from Bourne shell scripts,
       however, beyond mere variable assignments, no shell features are supported (this means
       variable expansion is explicitly not supported), allowing applications to read the file
       without implementing a shell compatible execution engine.

So that seems viable

blackboxsw: @holmanb good points. Yes that was just my recall that this was the case on some images. But specifically we added support for CentOS6 and Rawhide at the time. Neither image stream had an /etc/os-release file.   Given we don't support CentOS6 anymore because of old python and rawhide now carries /etc/os-release this is a non-issue. I agree `. /etc/os-release` instead of `source /etc/os-release` good catch on POSIX there. @CarlosNihelton I think this is approach otherwise holds water and should support any Linux distros we currently support.

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+Since WSL instances are scoped by Windows user, having the user data files
+inside the ``%USERPROFILE%`` directory (typically ``C:\Users\<USERNAME>``)
+ensures that WSL instances initialization won't be subject to naming conflicts
+if the Windows host is shared by multiple users.
+
+
+Vendor and metadata
+===================
+
+The current implementation doesn't allow supplying vendor data and metadata.
```
blackboxsw, line 100:
```suggestion
The current implementation doesn't allow supplying vendor data.
```

blackboxsw (approved):
Thanks @CarlosNihelton. I walked through the recent commits and changeset looks good to me an I think all comments are now addressed. Minor doc nit a +1 from me.

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
+.. _datasource_wsl:
+
+WSL
+***
+
+The Windows Subsystem for Linux (WSL) somewhat resembles a container
+hypervisor, a Windows user may have as many Linux distro instances as they
```
s-makin, line 7:
```suggestion
hypervisor. A Windows user may have as many Linux distro instances as they
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+
+3. **The init system must be aware of cloud-init**. WSL has opt-in support for
+   systemd, thus for distros that rely on it, such as Ubuntu, cloud-init will
+   run automatically if systemd is enabled via the ``/etc/wsl.conf``. The
+   Ubuntu applications distributed via Microsoft Store enable systemd in the
+   first boot, so no action is required if the user sets up a new instance by
+   using them. Users of other distros may find it surprising that cloud-init
+   doesn't run automatically by default. At the time of this writing, only
+   systemd distros are supported by the WSL datasource, although there is
+   nothing hard coded in the implementation code that requires it, so
```
s-makin, line 31:
```suggestion
   nothing hard-coded in the implementation code that requires it, so
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+   non-systemd distros may find ways to run cloud-init and make it just work.
+
+Notice that requirements 1 and 2 are met by default, i.e. WSL grants those
+features enabled, being the user allowed to opt-out if so desired.
```
s-makin, line 35: I found this line to be a little confusingly worded - I don't know exactly what was intended by "being the user allowed to opt-out" so I don't have an alternative suggestion.

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+For more information about how to configure WSL,
+`check the official documentation <https://learn.microsoft.com/windows/wsl/wsl-config#configuration-settings-for-wslconf>`_.
+
+User data Configuration
```
s-makin, line 39:
```suggestion
User data configuration
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+default. If none are found, cloud-init remains disabled.
+
+.. note::
+   Some users may have configured case sensitivity for file names on Windows.
+   Note that user data files will still be matched case-insensitively. If there
+   are both `InstanceName.user-data` and `instancename.user-data`, which one
+   will be chosen is arbitrary and should not be relied on. Thus it's
+   recommended to avoid that scenario to prevent confusion.
+
+Since WSL instances are scoped by Windows user, having the user data files
```
s-makin, line 91:
```suggestion
Since WSL instances are scoped by the Windows user, having the user data files
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+inside the ``%USERPROFILE%`` directory (typically ``C:\Users\<USERNAME>``)
+ensures that WSL instances initialization won't be subject to naming conflicts
```
s-makin, line 93:
```suggestion
ensures that WSL instance initialization won't be subject to naming conflicts
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+include file.
+
+* Network configuration is not supported.
+
+  WSL has full control of the instances' networking features and configuration.
+  A limited set of options for networking is exposed to the user via
+  ``/etc/wsl.conf``. Those options don't fit well with the networking model
+  cloud-init expects or understands.
+
+* Set Hostname.
```
s-makin, line 156:
```suggestion
* Set hostname.
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+        default=j
+
+* Disk setup, Growpart, Mounts and Resizefs.
+
+  The root filesystem must have the layout expected by WSL. Other mount points
+  may work, depending on how the hardware devices are exposed by the Windows
+  host, and fstab processing during boot is subject to configuration via
+  ``/etc/wsl.conf``, so users should expect limited functionality.
+
+* Grub Dpkg.
```
s-makin, line 195:
```suggestion
* GRUB dpkg.
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+
+  WSL controls the boot process, meaning that attempts to install and configure
+  Grub as any other bootloader won't be effective.
```
s-makin, line 198:
```suggestion
  GRUB as any other bootloader won't be effective.
```

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -0,0 +1,205 @@
...
+
+* Resolv Conf and Update Etc Hosts.
```
s-makin, line 200:
```suggestion
* Resolv conf and update etc/ hosts.
```

s-makin (commented):
Generally LGTM, just a few suggestions/nits :)

doc/rtd/reference/datasources/wsl.rst
s-makin: Thanks for the change, I think it's clearer now :)

s-makin (approved):

tests/unittests/test_ds_identify.py
```diff
@@ -113,6 +114,22 @@
 MOCK_UNAME_IS_FREEBSD = {"name": "uname", "out": UNAME_FREEBSD, "ret": 0}
 MOCK_UNAME_IS_OPENBSD = {"name": "uname", "out": UNAME_OPENBSD, "ret": 0}
 MOCK_UNAME_IS_WSL = {"name": "uname", "out": UNAME_WSL, "ret": 0}
+MOCK_WSL_INSTANCE_DATA = {
+    "name": "Noble-MLKit",
+    "distro": "ubuntu",
+    "version": "24.04",
+    "os_release": (
```
holmanb, line 121: style: We typically typically prefer `textwrap.dedent()` for more readable file strings, but I wont force you to change that - it works.

tools/ds-identify
```diff
@@ -1586,6 +1586,28 @@ dscheck_VMware() {
...
 }
 
+WSL_cloudinit_dir_in() {
+    _RET=""
+    for m in "$@"; do
+        cmdexe="$m/Windows/System23/cmd.exe"
+        if command -v "$cmdexe" > /dev/null 2>&1; then
+            cloudinitdir=$(/init "$cmdexe" /c echo %USERPROFILE% 2>/dev/null)
+            if [ -n "$cloudinitdir" ]; then
+                val=$(wslpath -au "$cloudinitdir") && _RET="$val"
```
holmanb, line 1596: Is the wslpath binary from wsl? Can we describe its origin in a comment please?

tools/ds-identify
```diff
@@ -1586,6 +1586,28 @@ dscheck_VMware() {
...
+            fi
+        fi
+    done
+
+    return 1
+}
+
+WSL_instance_name() {
+    instance_name=$(wslpath -am /)
+    val=$(basename "$instance_name")
```
holmanb, line 1607: We should be able to avoid this external dependency and shelling out by making use of the following builtin:
```

$ echo "$fname"
/usr/bin/sort
$ echo "$(basename $fname)"
sort
$ echo "${fname##*/}"
sort
```

tools/ds-identify
```diff
@@ -1586,6 +1586,28 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+WSL_cloudinit_dir_in() {
+    _RET=""
+    for m in "$@"; do
+        cmdexe="$m/Windows/System23/cmd.exe"
+        if command -v "$cmdexe" > /dev/null 2>&1; then
+            cloudinitdir=$(/init "$cmdexe" /c echo %USERPROFILE% 2>/dev/null)
```
holmanb, line 1594: I can guess what this does based on the context, but a comment describing what it does would be very helpful.

tools/ds-identify
```diff
@@ -1586,6 +1586,28 @@ dscheck_VMware() {
     return "${DS_NOT_FOUND}"
 }
 
+WSL_cloudinit_dir_in() {
+    _RET=""
+    for m in "$@"; do
+        cmdexe="$m/Windows/System23/cmd.exe"
```
holmanb, line 1592: we should probably declare the cmdexe variable as local to `WSL_cloudinit_dir_in` using the `local` keyword at the top of this function.

tools/ds-identify
holmanb, line 1594: cloudinitdir should be a local variable

tools/ds-identify
holmanb, line 1596: val should be a local variable

tools/ds-identify
```diff
@@ -1601,12 +1623,40 @@ dscheck_WSL() {
     # See https://youtu.be/lwhMThePdIo?t=2431&si=JKTHx39TyRgPbzkZ and
     # https://learn.microsoft.com/en-us/windows/wsl/wsl-config#what-is-drvfs
     # for more information.
-    if [ "$(grep -c '^[^[:space:]]* [^[:space:]]* 9p [^[:space:]]*aname=drvfs;.*' "${PATH_ROOT}/proc/mounts")" = 0 ]; then
+    mountpoints=$(grep '^[^[:space:]]* [^[:space:]]* 9p [^[:space:]]*aname=drvfs;.*' "${PATH_ROOT}/proc/mounts" | cut -f2 -d' ')
```
holmanb, line 1626: Same comments about `local` variables throughout this function: `mountpoints`, `cloudinitdir`, `instance_name`, and `candidate` should all be declared local at the start of this function.

holmanb (changes requested):
Thanks for the updates @CarlosNihelton. I have a few more changes requested for ds-identify. Once those land I'll be happy to land this.

holmanb (approved):
Thanks for all of your effort. I'm satisfied with how this turned out. Great work @CarlosNihelton.

Issue #4794, opened 2024-01-23
cloud-init-23.4 cannot read "- Azure" datasource_list format


zhaohuijuan:
# Bug report
In Image Builder images, the datasource_list in /etc/cloud/cloud.cfg.d/91-azure_datasource.cfg is:

datasource_list:
 \- Azure

which cannot be parsed by cloud-init.

2024-01-18 10:27:12,433 - stages.py[DEBUG]: Using distro class <class 'cloudinit.distros.rhel.Distro'>
2024-01-18 10:27:12,441 - _{_}init{_}_.py[DEBUG]: Looking for data source in: [\\{'datasource_list': None}, 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM', 'NETWORK']
2024-01-18 10:27:12,441 - util.py[WARNING]: failed stage init
2024-01-18 10:27:12,441 - util.py[DEBUG]: failed stage init
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/cloudinit/cmd/main.py", line 781, in status_wrapper
    ret = functor(name, args)
  File "/usr/lib/python3.6/site-packages/cloudinit/cmd/main.py", line 394, in main_init
    init.fetch(existing=existing)
  File "/usr/lib/python3.6/site-packages/cloudinit/stages.py", line 493, in fetch
    return self._get_data_source(existing=existing)
  File "/usr/lib/python3.6/site-packages/cloudinit/stages.py", line 367, in _get_data_source
    self.reporter,
  File "/usr/lib/python3.6/site-packages/cloudinit/sources/_{_}init{_}_.py", line 1001, in find_source
    ds_list = list_sources(cfg_list, ds_deps, pkg_list)
  File "/usr/lib/python3.6/site-packages/cloudinit/sources/_{_}init{_}_.py", line 1047, in list_sources
    ds_name = importer.match_case_insensitive_module_name(ds)
  File "/usr/lib/python3.6/site-packages/cloudinit/importer.py", line 40, in match_case_insensitive_module_name
    if "nocloud-net" == mod_name.lower():
AttributeError: 'dict' object has no attribute 'lower'

## Steps to reproduce the problem
1. Create a VM on Azure, change the /etc/cloud/cloud.cfg.d/91-azure_datasource.cfg to:

datasource:
  Azure:
    apply_network_config: false
datasource_list:
 \- Azure
2. Clean cloud-init and reboot VM

 $sudo cloud-init clean
[...]

Issue #4796, opened 2024-01-23
ds-identify yaml parsing is broken


holmanb:
# Bug report
Any call to `check_config()` in `ds-identify` is subtly broken, and could result in false positives as a result of this key in `/etc/cloud/cloud.cfg` or `/etc/cloud/cloud.cfg/.d*`. The most impactful use of this false positive can result in the false positive detection of MAAS as the datasource.

For example, the following configuration:
```yaml
datasource:
   MAAS:
     metadata_urls: [ 'blah.com' ]
```

Will result in MAAS being detected, even if it is not the correct platform.

```bash
# cat /run/cloud-init/cloud.cfg
datasource_list: [ MAAS, LXD, None ]
```

This may impact users that excpect to use a single configuration for multiple datasources, by causing cloud-init to attempt detection of maas at runtime:
```python
2024-01-23 19:59:14,496 - __init__.py[DEBUG]: Looking for data source in: ['MAAS', 'LXD', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM']
2024-01-23 19:59:14,499 - __init__.py[DEBUG]: Searching for local data source in: ['DataSourceMAAS', 'DataSourceLXD']
2024-01-23 19:59:14,499 - handlers.py[DEBUG]: start: init-local/search-MAAS: searching for local data from DataSourceMAAS
2024-01-23 19:59:14,499 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceMAAS.DataSourceMAAS'>
2024-01-23 19:59:14,500 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: boot-new-instance
2024-01-23 19:59:14,500 - __init__.py[DEBUG]: Detected platform: DataSourceMAAS [None]. Checking for active instance data
2024-01-23 19:59:14,500 - __init__.py[DEBUG]: Datasource DataSourceMAAS [None] not updated for events: boot-new-instance
2024-01-23 19:59:14,500 - handlers.py[DEBUG]: finish: init-local/search-MAAS: SUCCESS: no local data found from DataSourceMAAS
2024-01-23 19:59:14,500 - handlers.py[DEBUG]: start: init-local/search-LXD: searching for local data from DataSourceLXD
```

holmanb:
[...]

PR #4797, opened 2024-01-23
Revert ds identify optimization

holmanb:
## Additional Context
Reported in https://github.com/canonical/cloud-init/issues/4794, when an invalid configuration includes an invalid (empty) `datasource_list`, ds-identify will identify the wrong datasource.

While the datasource_list is invalid, it does represent a regression because previous ds-identify behavior was to silently ignore the invalid configuration and now it needlessly prevents cloud-init from running rather than identifying the correct datasource like it used to.

This issue was introduced in https://github.com/canonical/cloud-init/pull/4327.

## Test Steps
```
$ tox -e py3 tests/unittests/test_ds_identify.py
```

## Merge type

### DIFF
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -6,6 +6,8 @@ from collections import namedtuple
 from textwrap import dedent
 from uuid import uuid4
 
+import pytest
+
 from cloudinit import atomic_helper, safeyaml, subp, util
 from cloudinit.sources import DataSourceIBMCloud as ds_ibm
 from cloudinit.sources import DataSourceOracle as ds_oracle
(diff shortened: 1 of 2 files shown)


TheRealFalcon:
Not against a revert, but would it be better to add a warning/error upstream, and then patch this behavior downstream?

holmanb:
> Not against a revert, but would it be better to add a warning/error upstream, and then patch this behavior downstream?

We could do that, but by doing so we are choosing to break users when we don't have to in a common scenario of mistaken configuration (i.e.: "I thought cloud.cfg was a yaml file, what gives?").

I'm in agreement for complaining louder when user configurations are wrong, but I don't want to do that without actually making it reasonable for users to fix their issues. Most users aren't going to know to go digging `ds-identify.log`, they'll just see the following traceback in cloud-init.log, and assume that cloud-init is at fault. I suppose that we could make this more user friendly by checking for the invalid content in `/run/cloud/cloud.cfg` -> `datasource_list: [ datasource_list:, None ]` in Python and create a more user-friendly warning message in `cloud-init.log`.

However, if we only warn about this issue, that still leaves the part of this that amplifies https://github.com/canonical/cloud-init/issues/4796 unfixed (which I haven't mentioned directly yet). Notice the (non-xfail) MAAS tests. Do we really want cloud-init to run `DataSourceMAAS._get_data()` for every system that has the word `MAAS` appear in a `cloud.cfg` comment? That's just broken and very surprising for a user.


```
2024-01-18 10:27:12,441 - {}init{}.py[DEBUG]: Looking for data source in: [\{'datasource_list': None}, 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM', 'NETWORK']
2024-01-18 10:27:12,441 - util.py[WARNING]: failed stage init
2024-01-18 10:27:12,441 - util.py[DEBUG]: failed stage init
File "/usr/lib/python3.6/site-packages/cloudinit/cmd/main.py", line 781, in status_wrapper
ret = functor(name, args)
File "/usr/lib/python3.6/site-packages/cloudinit/cmd/main.py", line 394, in main_init
init.fetch(existing=existing)
File "/usr/lib/python3.6/site-packages/cloudinit/stages.py", line 493, in fetch
return self.get_data_source(existing=existing)
File "/usr/lib/python3.6/site-packages/cloudinit/stages.py", line 367, in get_data_source
self.reporter,
File "/usr/lib/python3.6/site-packages/cloudinit/sources/{}init{}.py", line 1001, in find_source
ds_list = list_sources(cfg_list, ds_deps, pkg_list)
File "/usr/lib/python3.6/site-packages/cloudinit/sources/{}init{}.py", line 1047, in list_sources
ds_name = importer.match_case_insensitive_module_name(ds)
File "/usr/lib/python3.6/site-packages/cloudinit/importer.py", line 40, in match_case_insensitive_module_name
if "nocloud-net" == mod_name.lower():
AttributeError: 'dict' object has no attribute 'lower'
```

users aren't going to see it. This is an issue in ds-identify, and most users aren't going to go digging through ds-identify logs. The "invalid" configuration is valid yaml, so unless we implement the same broken parser in Python we can't put it in cloud-init.log.

tests/unittests/test_ds_identify.py
```diff
@@ -460,6 +319,37 @@ def test_wb_print_variables(self):
...
             self.assertIn("{0}=".format(var), err)
 
+    @pytest.mark.xfail(reason="GH-4796")
+    def test_maas_not_detected_1(self):
+        """Don't incorrectly identify maas"""
+        config = "LXD-kvm-not-MAAS-1"
+        self._test_ds_found(config)
+
+    def test_maas_not_detected_2(self):
+        """Don't incorrectly identify maas
```
TheRealFalcon, line 329: Think it's worth putting an issue/commit reference in all of these tests (not just the xfail)? They just feel a bit random without the underlying context and if approaching otherwise I would think "why are we only testing maas here and not every other single datasource?"

tests/unittests/test_ds_identify.py
```diff
@@ -460,6 +319,37 @@ def test_wb_print_variables(self):
...
+    def test_maas_not_detected_3(self):
+        """Don't incorrectly identify maas
+
+        The bug reported in 4794 combined with the previously existing bug
+        reported in 4796 made for very loose MAAS false-positives.
+        """
+        config = "LXD-kvm-not-MAAS-3"
+        self._test_ds_found(config)
+
+    # fmt: off
```
TheRealFalcon, line 346: Why's this needed?

TheRealFalcon (approved):
Yeah, you raise good points. I was initially thinking we could have ds-identify fail more loudly if it can't parse something correctly, but that requires more smarts than ds-identify currently has (hence this issue).

tests/unittests/test_ds_identify.py
holmanb: black disagrees about how to format the docstring. If you turn that off, black will require `"""` at the end of the single-line string, which makes it longer than 79 characters long.

tests/unittests/test_ds_identify.py
holmanb: > Think it's worth putting an issue/commit reference in all of these tests

I think you missed the rest of the docstring?

>         The bug reported in 4794 combined with the previously existing bug
>         reported in 4796 made for very loose MAAS false-positives.

Or maybe I misunderstand what you are asking for.

holmanb:
@smoser fyi ^^

tests/unittests/test_ds_identify.py
holmanb: I've added more context to the tests.

PR #4805, opened 2024-01-24
Ubuntu/focal 23.4.x

aciba90:
## do not squash

- [x] build-package -- -S -d
- [x] sbuild --dist=focal --arch=amd64 --arch-all ../out/cloud-init_23.4.2-0ubuntu0~20.04.1.dsc
- [x] No new lintian errors / warnings

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,11 +1,11 @@
-cloud-init (23.4.2-0ubuntu0~20.04.1) UNRELEASED; urgency=medium
+cloud-init (23.4.2-0ubuntu0~20.04.1) focal; urgency=medium
 
   * Upstream snapshot based on 23.4.2. (LP: #2045582).
     List of changes from upstream can be found at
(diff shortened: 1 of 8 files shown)


debian/changelog
```diff
@@ -1,3 +1,12 @@
+cloud-init (23.4.2-0ubuntu0~20.04.1) focal; urgency=medium
+
+  * Upstream snapshot based on 23.4.2. (LP: #2045582).
```
TheRealFalcon, line 3: I think we can remove this bug reference and replace it with the one in the bottom bullet

blackboxsw (approved):
uploaded to -proposed. Will let you merge this.

debian/changelog
aciba90: That make sense for hotfix releases that only contain one bug fix (which would be most of the cases). I will point directly to bug being hotfixed for future releases, thanks!

PR #4810, opened 2024-01-26
refactor: Remove unused networking code

TheRealFalcon:
## Proposed Commit Message

```
refactor: Remove unused networking code

distros/__init__.py employs two ways of configuring network.
There was an old way in which the distro directly wrote out its
networking config, and a new (as of 7 years ago) way which involves
selecting a network renderer and rendering network using that.

We have kept the old way around to support Arch and Gentoo, whose code
never switched to using the new renderers, and write netctl and netifrc
respectively. This code is no longer needed.

On Arch, the only networking system included in the base image is
systemd-networkd, and the only supported init system is systemd. The
networkd renderer automatically gets selected in these images, so the
netctl code is no longer needed.

On Gentoo, the systemd images act the same as Arch. Due to a bug in
cloud-init, the openrc images never fall back to the netifrc code,
and instead throw an exception when they cannot find a suitable
renderer. This behavior has been in place for several years, which
means that the netifrc code has gone unused for several years.

If there's value in either of these forms of writing network, we can
always bring them back as separate renderers, but removing them
simplifies the distro code immensely.
```

## Additional Context
The gentoo bug mentioned in the commit message:
`apply_network_config` calls `self._get_renderer()` and falls back to the "old" code if `NotImplementedError` is raised from it. `_get_renderer()` then calls `renderers.select()`, and if no renderer can be found, `RendererNotFoundError` is raised. Since there's no handler for `RendererNotFoundError`, this doesn't get caught by the exception handler, and the "old" code is never invoked. The Arch code solved this by overriding `_get_renderer()` to catch `RendererNotFoundError` and raise `NotImplementedError` from it, but the Gentoo code has no such override.

I have observed this on a current Gentoo openrc LXD image:
```
2024-01-25 22:29:09,829 - stages.py[ERROR]: Unable to render networking. Network config is likely broken: No available network renderers found. Searched through list: ['eni', 'sysconfig', 'netplan', 'network-manager', 'freebsd', 'netbsd', 'openbsd', 'networkd']
2024-01-25 22:29:09,829 - main.py[DEBUG]: [local] Exiting without datasource
2024-01-25 22:29:09,830 - atomic_helper.py[DEBUG]: Atomically writing to file /var/lib/cloud/data/status.json (via temporary file /var/lib/cloud/data/tmpmvz2nxeh) - w: [644] 475 bytes/chars
2024-01-25 22:29:09,830 - util.py[DEBUG]: Reading from /proc/uptime (quiet=False)
2024-01-25 22:29:09,831 - util.py[DEBUG]: Read 14 bytes from /proc/uptime
2024-01-25 22:29:09,831 - util.py[DEBUG]: cloud-init mode 'init' took 0.116 seconds (0.11)
2024-01-25 22:29:09,831 - handlers.py[DEBUG]: finish: init-local: FAIL: searching for local datasources
```

## Test Steps
Unit tests pass. I have not tested these on a live system as I'm unsure how to build arch or gentoo images.

## Merge type


cloudinit/distros/arch.py
```diff
@@ -62,56 +58,6 @@ def install_packages(self, pkglist: PackageList):
         self.update_package_sources()
         self.package_command("", pkgs=pkglist)
 
-    def _get_renderer(self) -> Renderer:
```
holmanb, line 65: The base class still has this method, do you think we still need it?

holmanb (approved):
One nit, then LGTM.

If anyone wants to bring back support for Gentoo networking, we can always include it in upstream, but since it's been broken for so long in upstream I don't see any harm in removing it.

cloudinit/distros/arch.py
TheRealFalcon: Yes, this is still used by `apply_network_config`

PR #4813, opened 2024-01-26
refactor: Remove cc_rightscale_userdata

TheRealFalcon:
## Proposed Commit Message

```
refactor: Remove cc_rightscale_userdata

Rightscale no longer exists. Furthermore, even if the company that
bought it still uses it the same, the 2nd line of the handler does a
check that can never be True and then exits. Based on this, this
module no longer provides any value.
```

## Additional Context
The beginning of the handler:
```python
def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
    get_userdata_raw = getattr(cloud, "get_userdata_raw", None)
    if not get_userdata_raw or not callable(get_userdata_raw):
        LOG.debug("Failed to get raw userdata in module %s", name)
        return
```
`get_userdata_raw` is not a member of `Cloud`, and it hasn't been since the `0.6.*` days, back when the class was called `CloudInit`.

## Merge type


cloudinit/config/cc_rightscale_userdata.py
```diff
@@ -1,130 +0,0 @@
...
-from cloudinit.distros import ALL_DISTROS
-from cloudinit.settings import PER_INSTANCE
-
-MY_NAME = "cc_rightscale_userdata"
-MY_HOOKNAME = "CLOUD_INIT_REMOTE_HOOK"
-
-"""Rightscale Userdata: Support rightscale configuration hooks"""
-
-MODULE_DESCRIPTION = """\
-This module adds support for RightScale configuration hooks to cloud-init.
```
holmanb, line 27: This sounds a lot like an attempt to create a datasource inside of a configuration module. I'm curious why it was never done that way in the first place.

holmanb (commented):
I see a few other references to rightscale in docs:
```
doc/rtd/reference/base_config_reference.rst
351:    - rightscale_userdata

doc/rtd/explanation/instancedata.rst
465:     "rightscale_userdata",

doc/rtd/reference/performance_analysis.rst
88:         00.00100s (modules-final/config-rightscale_userdata)

doc/rtd/spelling_word_list.txt
205:rightscale

doc/rtd/explanation/analyze.rst
81:        00.00100s (modules-final/config-rightscale_userdata)
```

Do you think we should remove any of those?

TheRealFalcon:
Removed additional references. If docs check pass, it should be ready for re-review.

holmanb (approved):
LGTM!

PR #4817, opened 2024-01-29
test: Cleanup unwanted logger setup calls

holmanb:
## Proposed Commit Message
```
test: Cleanup unwanted logger setup calls

This was causing duplicate logs in test output.   
Also use modern class self() calls (less error-prone, easier to audit).
```

## Additional Context
While digging through test failures I noted that extra logging handlers were getting added in test code causing duplicate logs. Since the logging cleanup recently, a single root logger is used in unittests. Avoid polluting the logger with unwanted handlers - manual logger setup shouldn't be required in tests anymore. Also the CiTestCase class was overwriting the log level incorrectly.   

## Merge type

### DIFF
--- a/tests/unittests/config/test_cc_ca_certs.py
+++ b/tests/unittests/config/test_cc_ca_certs.py
@@ -8,9 +8,7 @@ from unittest import mock
 
 import pytest
 
-from cloudinit import distros, helpers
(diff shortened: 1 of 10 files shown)


TheRealFalcon (approved):
LGTM

PR #4823, opened 2024-01-30
Split `load_file` based on return type

TheRealFalcon:
## Proposed Commit Message

```
Use individual commits
```

## Additional Context
I did see a few places we could further refactor things, but I didn't want to complicate this PR any more than necessary. Other than the function definitions of `load_text_file` and `load_binary_file`, all other changes should be replacing the `load_file` as necessary. There is also one additional mypy related commit that surfaced from these changes.

## Merge type

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -49,6 +49,8 @@ from typing import (
     Optional,
     Sequence,
     TypeVar,
+    Union,
+    cast,
 )
 from urllib import parse
 
@@ -124,18 +126,14 @@ def lsb_release():
     return data
 
 
-def decode_binary(blob, encoding="utf-8") -> str:
+def decode_binary(blob: Union[str, bytes], encoding="utf-8") -> str:
     # Converts a binary type into a text type using given encoding.
-    if isinstance(blob, str):
-        return blob
-    return blob.decode(encoding)
+    return blob if isinstance(blob, str) else blob.decode(encoding=encoding)
 
 
-def encode_text(text, encoding="utf-8") -> bytes:
+def encode_text(text: Union[str, bytes], encoding="utf-8") -> bytes:
     # Converts a text string into a binary type using given encoding.
-    if isinstance(text, bytes):
-        return text
-    return text.encode(encoding)
+    return text if isinstance(text, bytes) else text.encode(encoding=encoding)
 
 
 def maybe_b64decode(data: bytes) -> bytes:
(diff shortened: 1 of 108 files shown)


cloudinit/util.py
```diff
@@ -1595,25 +1597,31 @@ def load_file(fname, read_cb=None, quiet=False, decode=True):
...
+    return contents
+
+
+def load_text_file(
+    fname: Union[str, os.PathLike],
+    *,
+    read_cb: Optional[Callable[[int], None]] = None,
+    quiet: bool = False,
+) -> str:
+    return decode_binary(load_binary_file(fname, read_cb=read_cb, quiet=quiet))
```
holmanb, line 1609: This means that every text file that we load is first read in binary mode then we manually decode it to text after it is in our buffer. Is there a reason we don't just `open(<file>, "r")` which will do that for us?

TheRealFalcon: I don't see a reason no...but I didn't want to further complicate this PR by changing any behavior. This is what was done before, so I'd rather not change more behavior on an enormous PR.

holmanb: > but I didn't want to further complicate this PR by changing any behavior

Okay, that's fair.

aciba90:
If the value of spliting load_files into to is only type checking, we could solve this by doing:

```
if typing.TYPE_CHECKING:
    @typing.overload
    def load_file(
        fname: Union[str, os.PathLike],
        *,
        read_cb: Optional[Callable[[int], None]] = None,
        quiet: bool = False,
        decode: typing.Literal[True] = True
    ) -> str:
        ...


    @typing.overload
    def load_file(
       fname: Union[str, os.PathLike],
       *,
       read_cb: Optional[Callable[[int], None]] = None,
       quiet: bool = False,
       decode: typing.Literal[False] = False
    ) -> bytes:
        ...


def load_file(
    fname: Union[str, os.PathLike],
    *,
    read_cb: Optional[Callable[[int], None]] = None,
    quiet: bool = False,
    decode: bool = False,
) -> Union[str, bytes]:
```

holmanb (approved):
Logically this doesn't do much, but it helps with typing static analysis tools and simplifies reasoning about what functions. Whether `encode` is turning binary into utf-8 text or the inverse of  that is something that may not be obvious to readers of the code - human grokking is easier with this change.

I have one nit, but feel free to take it or leave it.

Also I raised it in a comment (and I'm fine with this being a separate change), but probably worth looking into someday in the future is whether reading `open()` in text mode would be more performant than manually encoding - I would expect that it is written in C and therefore is probably faster than post-processing in Python, but I may be wrong - that's just a hunch.

TheRealFalcon:
I'm going to leave it for now to not add any additional complexity, but I think there's further improvements we could make here.

PR #4825, opened 2024-01-30
fix(source): fix argument boundaries when parsing cmdline

gilbsgilbs:
## Proposed Commit Message

```
fix(source): fix argument boundaries when parsing cmdline

When the command-line includes an argument that contains `ds=`, the rest
of the argument is always interpreted as a datasource, regardless of
wether the argument is `ds` or just the suffix of another argument.
While this doesn't cause issues in a majority of cases, it starts
conflicting with other arguments that also include `ds=` but do not
refer to a cloud-init datasource. For instance, setting
`bonding.max_bonds=0` in the command-line (which is a real-world
use-case) conflicts with the current implementation.

To fix this issue, this commit ensures that `ds`, `ci.ds` and
`ci.datasource` arguments are either at the very beginning of the
command-line, or just before a whitespace, so that we don't match
another argument.
```

## Test Steps

Add `bonding.max_bonds=0` to the command-line, and then `ds=nocloud-net;s=https://some`. Result: the parsed datasource is `0` instead of `nocloud-net;s=https://some`.

## Checklist

- [ ] (irrelevant as it is a bugfix) I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

### DIFF
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -1185,9 +1185,9 @@ def parse_cmdline() -> str:
     Passing by command line overrides runtime datasource detection
     """
     cmdline = util.get_cmdline()
-    ds_parse_0 = re.search(r"ds=([^\s;]+)", cmdline)
-    ds_parse_1 = re.search(r"ci\.ds=([^\s;]+)", cmdline)
-    ds_parse_2 = re.search(r"ci\.datasource=([^\s;]+)", cmdline)
+    ds_parse_0 = re.search(r"(?:^|\s)ds=([^\s;]+)", cmdline)
+    ds_parse_1 = re.search(r"(?:^|\s)ci\.ds=([^\s;]+)", cmdline)
+    ds_parse_2 = re.search(r"(?:^|\s)ci\.datasource=([^\s;]+)", cmdline)
     ds = ds_parse_0 or ds_parse_1 or ds_parse_2
     deprecated = ds_parse_1 or ds_parse_2
     if deprecated:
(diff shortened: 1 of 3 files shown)


holmanb:
Hey @gilbsgilbs, thanks for proposing this contribution. Can you please [sign the CLA](https://ubuntu.com/legal/contributors) so that we may accept this contribution?

gilbsgilbs:
I'd like to, but it's asking me for a launchpad id that I can't manage to find it on my launchpad account.

cloudinit/sources/__init__.py
```diff
@@ -1185,9 +1185,9 @@ def parse_cmdline() -> str:
     Passing by command line overrides runtime datasource detection
     """
     cmdline = util.get_cmdline()
-    ds_parse_0 = re.search(r"ds=([^\s;]+)", cmdline)
-    ds_parse_1 = re.search(r"ci\.ds=([^\s;]+)", cmdline)
-    ds_parse_2 = re.search(r"ci\.datasource=([^\s;]+)", cmdline)
+    ds_parse_0 = re.search(r"(?:^|\s)ds=([^\s;]+)", cmdline)
```
holmanb, line 1188: The `?:` is meaningless in these regexes, did you intend something else?

holmanb (changes requested):
@gilbsgilbs Great, thanks for signing! Can you please include in this PR your github username in [this file](https://github.com/canonical/cloud-init/blob/main/tools/.github-cla-signers)?

I also have one question about the regex.

cloudinit/sources/__init__.py
gilbsgilbs: It makes the group non-capturing. The other group in these regexes is referred to as `.group(1)`, so if you capture the first group, you have to shift the group index. As the first group is only used to match what we want, I think making it non-capturing is "more correct" than shifting the group at the end of this function.

gilbsgilbs:
> @gilbsgilbs Great, thanks for signing! Can you please include in this PR your github username in [this file](https://github.com/canonical/cloud-init/blob/main/tools/.github-cla-signers)?

Done.

holmanb (approved):
Thanks for the bug fix and welcome to cloud-init @gilbsgilbs!

PR #4829, opened 2024-01-30
fix: Ensure NetworkManager renderer works without gateway

TheRealFalcon:
## Proposed Commit Message

```
fix: Ensure NetworkManager renderer works without gateway

Gateway isn't required when specifying routes. Currently,
NetworkManager requires one. Make it optional.
```

## Merge type


holmanb (approved):
Think we should add a test for this behavior?

TheRealFalcon:
Added a test for the behavior. Without this change, the test fails with:
```
    def _add_route(self, route):
        """Adds a ipv[46].route<n> property."""
        # Because network v2 route definitions can have mixed v4 and v6
        # routes, determine the family per route based on the gateway
>       family = "ipv6" if is_ipv6_network(route["gateway"]) else "ipv4"
E       KeyError: 'gateway'

cloudinit/net/network_manager.py:228: KeyError
```

PR #4834, opened 2024-02-01
Add ftp support to NoCloud

holmanb:
## Proposed Commit Message
```
feat: Add support for FTP and FTP over TLS

Use Python's ftplib to connect to and read instance configurations from a
remove server. Like the current http support, nocloud's configuration URI
is defined via kernel command line or dmi product serial.

Notes on FTP and FTP over TLS
=============================

FTP is insecure, use at your own risk.

The FTP over TLS implementation follows library best practices[1][2].
From the ssl documentation regarding create_default_context():

> It will load the system’s trusted CA certificates, enable certificate
> validation and hostname checking, and try to choose reasonably secure
> protocol and cipher settings.

Test Changes
============

tests/i/util.py
---------------

- new helper function verify_clean_boot() to verify cloud-init state
- acquire override_kernel_cmdline() for reuse in other tests
- acquire restart_cloud_init() for reuse in other tests

tests/i/datasources/test_nocloud.py
-----------------------------------

- new test class for FTP/FTPS uses pyftpdlib for server, mkcert for test certs
- 2 test additions for FTP
- 2 test additions for FTPS

Implements via ftplib
=====================

RFC 959 - File Transfer Protocol
RFC 2640 - FTP Internationalization (UTF-8)
RFC 4217 - FTP over TLS

[1] https://docs.python.org/3/library/ftplib.html#ftp-tls-objects
[2] https://docs.python.org/3/library/ssl.html#ssl-security
```

## Test Steps
```
PLATFORM="lxd_vm" CLOUD_INIT_SOURCE="./cloud-init_all.deb" python -m
pytest -vv --log-cli-level=INFO --durations 10 tests/integration_tests/datasources/test_nocloud.py::TestFTP
```

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -369,7 +369,7 @@ class DataSourceNoCloudNet(DataSourceNoCloud):
             "http://",
             "https://",
             "ftp://",
-            "stfp://",
+            "ftps://",
         )
 
     def ds_detect(self):
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -288,7 +288,9 @@ def load_cmdline_data(fill, cmdline=None):
 
         seedfrom = fill.get("seedfrom")
         if seedfrom:
-            if seedfrom.startswith(("http://", "https://")):
+            if seedfrom.startswith(
+                ("http://", "https://", "ftp://", "ftps://")
+            ):
                 fill["dsmode"] = sources.DSMODE_NETWORK
             elif seedfrom.startswith(("file://", "/")):
                 fill["dsmode"] = sources.DSMODE_LOCAL
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -365,7 +365,12 @@ def _merge_new_seed(cur, seeded):
 class DataSourceNoCloudNet(DataSourceNoCloud):
     def __init__(self, sys_cfg, distro, paths):
         DataSourceNoCloud.__init__(self, sys_cfg, distro, paths)
-        self.supported_seed_starts = ("http://", "https://")
+        self.supported_seed_starts = (
+            "http://",
+            "https://",
+            "ftp://",
+            "stfp://",
+        )
 
     def ds_detect(self):
         """Check dmi and kernel commandline for dsname
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -421,3 +421,14 @@ datasources = [
 # Return a list of data sources that match this set of dependencies
 def get_datasource_list(depends):
     return sources.list_from_depends(depends, datasources)
+
+
+if __name__ == "__main__":
+    from sys import argv
+
+    logging.basicConfig(level=logging.DEBUG)
+    seedfrom = argv[1]
+    md_seed, ud, vd = util.read_seeded(seedfrom)
+    print(f"seeded: {md_seed}")
+    print(f"ud: {ud}")
+    print(f"vd: {vd}")
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -123,21 +123,6 @@ def read_ftps(url: str, timeout: float = 5.0, **kwargs: dict) -> "FtpResponse":
                     port=port,
                     timeout=timeout or 5.0,  # uses float internally
                 )
-                LOG.debug("Attempting to login with user [%s]", user)
-                ftp_tls.login(
-                    user=user,
-                    passwd=url_parts.password or "",
-                )
-                LOG.debug("Creating a secure connection")
-                ftp_tls.prot_p()
-                LOG.debug("Reading file: %s", url_parts.path)
-                ftp_tls.retrbinary(
-                    f"RETR {url_parts.path}", callback=buffer.write
-                )
-                response = FtpResponse(url_parts.path, contents=buffer)
-                LOG.debug("Closing connection")
-                ftp_tls.close()
-                return response
             except ftplib.all_errors as e:
                 code = ftp_get_return_code_from_exception(e)
                 raise UrlError(
@@ -149,6 +134,21 @@ def read_ftps(url: str, timeout: float = 5.0, **kwargs: dict) -> "FtpResponse":
                     headers=None,
                     url=url,
                 ) from e
+            LOG.debug("Attempting to login with user [%s]", user)
+            ftp_tls.login(
+                user=user,
+                passwd=url_parts.password or "",
+            )
+            LOG.debug("Creating a secure connection")
+            ftp_tls.prot_p()
+            LOG.debug("Reading file: %s", url_parts.path)
+            ftp_tls.retrbinary(
+                f"RETR {url_parts.path}", callback=buffer.write
+            )
+            response = FtpResponse(url_parts.path, contents=buffer)
+            LOG.debug("Closing connection")
+            ftp_tls.close()
+            return response
         else:
             try:
                 ftp = ftplib.FTP()
@@ -159,17 +159,6 @@ def read_ftps(url: str, timeout: float = 5.0, **kwargs: dict) -> "FtpResponse":
                     port=port,
                     timeout=timeout or 5.0,  # uses float internally
                 )
-                LOG.debug("Attempting to login with user [%s]", user)
-                ftp.login(
-                    user=user,
-                    passwd=url_parts.password or "",
-                )
-                LOG.debug("Reading file: %s", url_parts.path)
-                ftp.retrbinary(f"RETR {url_parts.path}", callback=buffer.write)
-                response = FtpResponse(url_parts.path, contents=buffer)
-                LOG.debug("Closing connection")
-                ftp.close()
-                return response
             except ftplib.all_errors as e:
                 code=ftp_get_return_code_from_exception(e),
                 raise UrlError(
@@ -181,6 +170,17 @@ def read_ftps(url: str, timeout: float = 5.0, **kwargs: dict) -> "FtpResponse":
                     headers=None,
                     url=url,
                 ) from e
+            LOG.debug("Attempting to login with user [%s]", user)
+            ftp.login(
+                user=user,
+                passwd=url_parts.password or "",
+            )
+            LOG.debug("Reading file: %s", url_parts.path)
+            ftp.retrbinary(f"RETR {url_parts.path}", callback=buffer.write)
+            response = FtpResponse(url_parts.path, contents=buffer)
+            LOG.debug("Closing connection")
+            ftp.close()
+            return response
 
 
 def _read_file(path: str, **kwargs) -> "FileResponse":
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -9,6 +9,8 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 import copy
+import ftplib
+import io
 import json
 import logging
 import os
@@ -16,17 +18,17 @@ import threading
 import time
 from concurrent.futures import ThreadPoolExecutor, TimeoutError, as_completed
 from email.utils import parsedate
-from errno import ENOENT
 from functools import partial
 from http.client import NOT_FOUND
 from itertools import count
+from ssl import create_default_context
 from typing import Any, Callable, Iterator, List, Optional, Tuple, Union
-from urllib.parse import quote, urlparse, urlunparse
+from urllib.parse import quote, urlparse, urlsplit, urlunparse
 
 import requests
 from requests import exceptions
 
-from cloudinit import version
+from cloudinit import util, version
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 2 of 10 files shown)


cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -365,7 +365,12 @@ def _merge_new_seed(cur, seeded):
 class DataSourceNoCloudNet(DataSourceNoCloud):
     def __init__(self, sys_cfg, distro, paths):
         DataSourceNoCloud.__init__(self, sys_cfg, distro, paths)
-        self.supported_seed_starts = ("http://", "https://")
+        self.supported_seed_starts = (
+            "http://",
+            "https://",
+            "ftp://",
+            "stfp://",
```
dermotbradley, line 372: This is a typo, I assume "sftp" was meant, however it really should be "ftps" as "sftp" typically refers to FTP-over-SSH.

Also regarding "ftp" support - unencrypted FTP isn't really considered "safe" these days AFAIK, e.g. most if not all browsers have removed support for it, should "plain" FTP really be added as a supported protocol?

holmanb: > This is a typo, I assume "sftp" was meant, however it really should be "ftps" as "sftp" typically refers to FTP-over-SSH.

Nice catch, thanks. ftps was intended here.

> e.g. most if not all browsers have removed support for it, should "plain" FTP really be added as a supported protocol?

Hmm, good point. The thought was to allow administrators on local trusted networks to use cloud-init with less hassle, but you make a good point about other tooling removing support for ftp - it does look odd to add in 2024. The idea here (docs to come) is that when using "ftps://", only encrypted ftp would work - with unencrypted only allowed if the user uses "ftp://" for their scheme. But you might be right, that we should just add ftps and not bother with unencrypted ftp.

cloudinit/util.py
```diff
@@ -2968,8 +2968,6 @@ def is_x86(uname_arch=None):
 
 
 def message_from_string(string):
-    if sys.version_info[:2] < (2, 7):
```
TheRealFalcon, line 2971: Bwahaha!

cloudinit/sources/DataSourceNoCloud.py
TheRealFalcon: I'm not sure I agree about vanilla ftp. Isn't it still commonly used over internal networks? In browsers, support doesn't really buy much functionality other than not having to use another client, so it doesn't justify the maintenance burden. In cloud-init, it could potentially be a lot more useful, and the maintenance is few lines of fallback code.

However, I do suspect that this could be a feature looking for a user, so it's not a strongly held opinion.

TheRealFalcon:
@holmanb this still WIP?

holmanb:
> @holmanb this still WIP?

I'm working on an integration test for it currently, but I think it is ready for review.

TheRealFalcon (commented):
Besides the test issues, I wanted to provide my other comments here now

cloudinit/url_helper.py
```diff
@@ -59,7 +61,137 @@ def combine_single(url, add_on):
...
+                    ),
+                    code=code,
+                    headers=None,
+                    url=url,
+                ) from e
+            LOG.info(
+                "Connecting to ftp server over tls "
+                "failed for url %s [%s]", url, code
+            )
+        try:
```
TheRealFalcon, line 143: I'm wondering if there's a possible security concern here. If somebody has an ftps server, but then specifies "ftp://", will it still connect? If so, they may think they have a secure connection, but a MITM could host a different ftp server and trigger this fallback code.

cloudinit/url_helper.py
```diff
@@ -59,7 +61,108 @@ def combine_single(url, add_on):
...
+def read_ftps(url: str, timeout: Optional[float] = None) -> "FtpResponse":
+    """connect to URL using ftp over TLS and read a file
+
+    when using strict mode (ftps://), raise exception in event of failure
+    when not using strict mode (ftp://), fall back to using unencrypted ftp
+
+    returns: UrlResponse
+    """
+
+    def get_return_code_from_exception(exc):
```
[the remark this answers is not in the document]
holmanb: Those are all fair reasons to dislike avoid them.

I personally like them for the encapsulation that they provide. They make it easier to understand the intended context of the nested function and groups together related code.

cloudinit/url_helper.py
```diff
@@ -59,7 +61,108 @@ def combine_single(url, add_on):
...
+                timeout=timeout or 0.0,  # Python docs are wrong about types
+                context=create_default_context(),
+            )
+            ftp_tls.login()
+            ftp_tls.prot_p()
+            ftp_tls.retrbinary(f"RETR {url_parts.path}", callback=buffer.write)
+            return FtpResponse(url_parts.path, contents=buffer)
+        except ftplib.all_errors as e:
+            if "ftps" == url_parts.scheme:
+                raise UrlError(
```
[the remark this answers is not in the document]
holmanb: > Can just the connect line be in the try block while letting all other exceptions escape?

I'm hesitant to do that in the first try/except block, but I'm fine with doing it in the second.

If `prot_p()` throws an exception and we're using `ftp://` (which permits unencrypted ftp) I think we should try to recover so we can attempt unencrypted.

cloudinit/url_helper.py
holmanb: > I'm wondering if there's a possible security concern here. If somebody has an ftps server, but then specifies "ftp://", will it still connect?

By default, yes.

I believe this can be prevented by the user by configuring the base image with CA certs which don't trust the attacker's certs.    

> If so, they may think they have a secure connection, but a MITM could host a different ftp server and trigger this fallback code.

This is possible regardless of the fallback code (via "ftps://"). If an attacker can MITM the instance (and the instance's CA chain trusts the attacker's TLS certs), then the instance will have a TLS-encrypted connection to the MITM server which could itself host an ftp server.

I don't believe that the security concern mentioned here is any different than connecting via HTTPS, which we already support in this codepath. Both can be avoided by only initializing over trusted networks or (I think) using a custom CA.

Signed user-data is something we could implement someday to help these scenarios.

This is all probably better answered by @setharnold.

cloudinit/url_helper.py
```diff
@@ -59,7 +61,108 @@ def combine_single(url, add_on):
...
+                user=url_parts.username or "",
+                passwd=url_parts.password or "",
+                timeout=timeout or 0.0,  # Python docs are wrong about types
+                context=create_default_context(),
+            )
+            ftp_tls.login()
+            ftp_tls.prot_p()
+            ftp_tls.retrbinary(f"RETR {url_parts.path}", callback=buffer.write)
+            return FtpResponse(url_parts.path, contents=buffer)
+        except ftplib.all_errors as e:
```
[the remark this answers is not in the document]
holmanb: Heh, there's a few things about this standard library that I would have done differently.

cloudinit/url_helper.py
TheRealFalcon: > I believe this can be prevented by the user by configuring the base image with CA certs which don't trust the attacker's certs.
If an attacker can MITM the instance (and the instance's CA chain trusts the attacker's TLS certs), then the instance will have a TLS-encrypted connection to the MITM server which could itself host an ftp server.

But the attacker wouldn't need certs at all. If I put up a plain unencrypted, no-cert ftp server between cloud-init and the fully secure target ftps server, but specify `ftp://`, our code will see the MITM, attempt to connect securely, fail on the (normally used) ftps connection, log a message, then fallback to an unsecured ftp connect to the MITM. Right?

The obvious answer is "don't connect with ftp:// when you mean ftps://", but I'm just concerned that this is a potential security footgun.

Regardless, I agree it's best to check with security.

holmanb: > But the attacker wouldn't need certs at all.

Correct, that is how unencrypted ftp would work regardless of this fallback path.

The user explicitly asked to support an insecure unencrypted ftp by setting "ftp://", so attempting to use TLS first isn't introducing any vulnerabilities - it simply attempts a more secure connection first in case the server supports TLS. The "fallback" is to do exactly what the user asked.

> The obvious answer is "don't connect with ftp:// when you mean ftps://", but I'm just concerned that this is a potential security footgun.

The alternative is "cloud-init shouldn't support unencrypted ftp", and to quote your earlier comment on this PR:

> I'm not sure I agree about vanilla ftp. Isn't it still commonly used over internal networks? In browsers, support doesn't really buy much functionality other than not having to use another client, so it doesn't justify the maintenance burden. In cloud-init, it could potentially be a lot more useful, and the maintenance is few lines of fallback code.
>
>  it could potentially be a lot more useful, and the maintenance is few lines of fallback code.

Do you think we should support unencrypted ftp or don't you?

I'm in favor of it, since users may want it for use on a secure private management network. I think that providing the option, documenting the implications, and then allowing the user to define their own security posture makes sense. I also think that it doesn't do any harm to try ftps first, just in case it is supported by the server.

TheRealFalcon: > Do you think we should support unencrypted ftp or don't you?

You're framing this as if I just decided I want two totally opposite things. As I reviewed the code more carefully, some security implications stood out to me. I want to make sure this won't be a problem. Yes, I still want ftp...but I don't want it if it exposes a security vulnerability. I'm trying to figure that out and I'd rather not sweep those concerns under the rug. False dichotomies won't help the process.

> The alternative is "cloud-init shouldn't support unencrypted ftp"

Is that the only alternative? What if we required `ftps://` to connect to a secure server? That way there's no pretense that they'll be getting a secure connection when they won't be.

holmanb: > You're framing this as if I just decided I want two totally opposite things. As I reviewed the code more carefully, some security implications stood out to me. I want to make sure this won't be a problem.

My apologies, I'm not trying to shut down the conversation. It's good to talk through and make sure we understand the implementation details and users' understanding of the tool.

Ftp is not a secure transfer protocol and if we include it will have to be documented as such, regardless of whether we quietly attempt TLS for a more secure connection.

>> The alternative is "cloud-init shouldn't support unencrypted ftp"
>
> Is that the only alternative? What if we required ftps:// to connect to a secure server? That way there's no pretense that they'll be getting a secure connection when they won't be.

I think we were thinking about this in different ways. In your last comment I thought you were concerned that using TLS even when the user didn't request it would somehow make it easier for an attacker to intercept communication than normal unencrypted ftp would. Requiring `ftps://` to use TLS wouldn't make things any technically more secure, so I guess this is still about user perception of this protocol? I haven't put together docs yet, but if I think I would say something like:

| scheme | title | description |
| ------ | ----- | ----------- |
| ftps:// | FTP over TLS | acquire configurations from a server with TLS support |
| ftp:// | unencrypted FTP | not safe for use over untrusted networks, use at your own risk |

TheRealFalcon: > Requiring ftps:// to use TLS wouldn't make things any technically more secure, so I guess this is still about user perception of this protocol?

I think so? At the end of the day, I think you doc change would be good enough, but I think there's still a potential footgun there that's worth addressing.

I think it's a footgun because we're upgrading the connection to TLS, but in reality, it can only provide encryption, not identity. Somebody can look at the docs once, see we provide fpt(s) support, standup an ftps server, connect using "ftp://" and think they're secure because they see a TLS connection. Our fallback behavior prevents identity (because it can be MITMed), so "upgrading" to TLS without the possibility of real identity gives a false sense of security. At the end of the day we can say "you didn't RTFM", but if we could prevent that by only attempting a TLS connection when we see "ftps", then I think we should. I say that not really knowing the library though, so I don't actually know how hard that would be.

holmanb: fair enough, I've made the ftp:// scheme only support ftp, and the ftps:// scheme only support ftp over tls

cloudinit/url_helper.py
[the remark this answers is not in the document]
holmanb: @TheRealFalcon I think I've done what you asked, but I'm not certain I fully understood. Let me know.

holmanb:
@TheRealFalcon I think I've addressed your comments. I've added integration tests and updated the commit message.

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +95,237 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
...
+                assert client.execute("apt install -yq golang").ok
+
+                # build mkcert from source
+                #
+                # we could check out a tag, but the project hasn't
+                # been updated in 2 years
+                #
+                # instructions from https://github.com/FiloSottile/mkcert
+                assert client.execute(
+                    "git clone https://github.com/FiloSottile/mkcert && "
```
setharnold, line 211: This changes where this can be run, and how safe it is to run this test. Could we package a fixed cert.pem and key.pem (set to expire in a hundred years or so?) to skip this download, compile, and execute?

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +95,237 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
...
+                "mkcert -install -cert-file /cert.pem -key-file /key.pem "
+                "localhost 127.0.0.1 0.0.0.0 ::1"
+            ).ok
+
+        client.write_to_file(
+            "/lib/systemd/system/local-ftp.service",
+            dedent(
+                """\
+                [Unit]
+                Description=run a local ftp server against
```
setharnold, line 233: I think this was accidentally truncated. Consider:

```
Description=TESTING USE ONLY ftp server
```

or something else that would stand out pretty loudly if someone accidentally runs these tests on their development workstation.

tests/integration_tests/util.py
```diff
@@ -42,6 +45,130 @@ def verify_ordered_items_in_text(to_verify: list, text: str):
...
+    require_warnings = require_warnings or []
+    status = json.loads(c.execute("cloud-init status --format=json"))
+
+    unexpected_errors = set()
+    unexpected_warnings = set()
+
+    required_warnings_found = set()
+    required_errors_found = set()
+
+    for current_error in status["errors"]:
```
setharnold, line 104: These loops are more complex than I'd like. They might work perfectly well, but it isn't obvious to me that they do. I would find set operations easier, eg:

```
missing_errors = require_errors.subtract(status["errors"])
surplus_errors = status["errors"].subtract(require_errors).subtract(ignore_errors)
```

(or whatever would actually run.)

setharnold (commented):
I added some nitpicks, feel free to disregard if you disagree strongly enough. Thanks

tests/integration_tests/datasources/test_nocloud.py
holmanb: > Could we package a fixed cert.pem and key.pem (set to expire in a hundred years or so?) to skip this download, compile, and execute?

Given that this will be dropped as soon as we are no longer releasing to focal, I'm not sure that it is worth it. Also, I could use help with cert generation for that. 

The challenging bit wasn't actually getting a cert.pem and key.pem, it was building a root cert and getting it into the CA store so that the client would trust it - self signed certs won't work for that. I'm not very well versed in the openssl commandline, but if someone is able to generate a test root CA, cert, and key I'd be happy to use that rather than using mkcert.

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +95,237 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
...
+        self, client: IntegrationInstance
+    ):
+        """check that ftp:// succeeds to unencrypted ftp server
+
+        this mode allows administrators to choose unencrypted ftp,
+        at their own risk
+        """
+        cmdline = "ds=nocloud;seedfrom=ftp://0.0.0.0:2121"
+        self._boot_with_cmdline(cmdline, client)
+        verify_clean_boot(client, ignore_warnings=self.expected_warnings)
```
TheRealFalcon, line 291: Can we also verify here (and the other tests) that the userdata worked correctly? I.e., verify the hostname or add a bootmcd to the userdata to touch a file and show the file was actually touched.

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +95,237 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
     client.restart()
     assert client.execute("cloud-init status").ok
     assert "seeded_vendordata_test_file" in client.execute("ls /var/tmp")
+
+
+@pytest.mark.skipif(PLATFORM != "lxd_vm", reason="Modifies grub config")
+@pytest.mark.lxd_use_exec
+class TestFTP:
```
TheRealFalcon, line 102: Nice work getting all of this setup!

TheRealFalcon (changes requested):
Generally LGTM...just mostly test requests inline.

tests/integration_tests/datasources/test_nocloud.py
holmanb: ◉‿◉

tests/integration_tests/datasources/test_nocloud.py
setharnold: "no longer releasing to focal" may be a very long way away.

The security team has some rough-and-ready python for testing OpenSSL that might be suitable for stealing:

https://git.launchpad.net/qa-regression-testing/tree/scripts/test-openssl.py#n238

tests/integration_tests/util.py
```diff
@@ -222,3 +349,63 @@ def get_feature_flag_value(client: "IntegrationInstance", key):
     if "NameError" in value:
         raise NameError(f"name '{key}' is not defined")
     return value
+
+
+def restart_cloud_init(c: "IntegrationInstance"):
```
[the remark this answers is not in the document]
holmanb: > Why is this needed?

If I remember correctly, this was created because `.restart()` didn't work on lxd vms. This was originally added back in a60c0845806baff72c7, and this PR only refactors it to be reusable (i.e. by two different tests that both use it).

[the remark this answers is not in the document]
holmanb: Although, now that I say that I can't reproduce the issue that I recall seeing before. I'll drop it.

tests/integration_tests/util.py
holmanb: I agree that this would be easier logic to follow, however set operations require exact matches rather than allowing substring matches (what the code currently does), or regex matches (what we probably want in the future). I haven't yet thought of an elegant way to accomplish this using sets.

cloudinit/url_helper.py
```diff
@@ -68,22 +209,16 @@ def read_file_or_url(url, **kwargs) -> Union["FileResponse", "UrlResponse"]:
...
-        try:
-            with open(file_path, "rb") as fp:
-                contents = fp.read()
-        except IOError as e:
-            code = e.errno
-            if e.errno == ENOENT:
-                code = NOT_FOUND
-            raise UrlError(cause=e, code=code, headers=None, url=url) from e
-        return FileResponse(file_path, contents=contents)
+    parsed = urlparse(url)
```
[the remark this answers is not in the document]
holmanb: Addressed with exception handling

tests/integration_tests/datasources/test_nocloud.py
holmanb: done

tests/integration_tests/util.py
```diff
@@ -42,6 +45,130 @@ def verify_ordered_items_in_text(to_verify: list, text: str):
...
+    # if only one error put the header and the error message on a single line
+    if 1 == len(items):
+        return f"\n{header}: {items.pop(0)}"
+
+    # otherwise make a list after header
+    else:
+        return f"\n{header}:\n\t- " + "\n\t- ".join(items)
+
+
+def verify_clean_boot(
```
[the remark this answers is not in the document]
holmanb: > At the very least I think the new function should call verify_clean_log for now.

I think that at the moment, verify_clean_log just won't work unless you pass matching expected failures. This is the result of the "Found unexpected errors/warnings" code paths - it just won't work unless you've passed everything that you expect to see in the `ignored_warnings` and `ignored_errors` keys. I don't think that calling `verify_clean_log()` inside of `verify_clean_boot()` will fix that - in fact we can't because we _require_ failures in some of our tests, but I'm happy to call that manually as well in the integration tests that we expect a clean log in.

One idea that I envision for this helper is a matrix of platforms and series that allows us to specify exactly when an expected warning message that will be seen. This would be more precise than the current "ignore this pattern everywhere" paradigm that `verify_clean_log()` uses. That said, I'm not sure that I want to include that refactor in this PR, which has already grown sizeable.

holmanb:
@TheRealFalcon I believe that I've addressed all of your comments.

I have one more review item left to address, which is the request to include a manually generated cert rather than installing mkcert from source for focal. I won't be able to work on that request until Friday at the earliest, so if you'd like to re-review the non-test code, I don't have any more planned changes there.

tests/integration_tests/util.py
```diff
@@ -42,6 +45,130 @@ def verify_ordered_items_in_text(to_verify: list, text: str):
...
+    c: "IntegrationInstance",
```
TheRealFalcon, line 65: `c` variable still exists.

tests/integration_tests/util.py
```diff
@@ -219,3 +346,41 @@ def get_feature_flag_value(client: "IntegrationInstance", key):
     if "NameError" in value:
         raise NameError(f"name '{key}' is not defined")
     return value
+
+
+def override_kernel_cmdline(ds_str: str, c: "IntegrationInstance"):
```
TheRealFalcon, line 351: `c` variable still exists.

TheRealFalcon (approved):
Nit: still some `c` variables in function signatures, but otherwise LGTM

holmanb:
rebased and force pushed - addressed remaining comment from @TheRealFalcon and merging

PR #4835, opened 2024-02-01
Add option 245 support to dhcpcd

holmanb:
## Commit message

```
feat(dhcpcd): Read dhcp option 245 for azure wireserver

Move lease parsing logic into the dhcp class.
Add tests for this logic.
```

## Additional Context

Reading the binary lease ourselves is currently required for dhcpcd due to https://github.com/NetworkConfiguration/dhcpcd/issues/282.

I have tested both dhcpcd and dhclient on Azure with this change. Both appear to have worked as expected.

## Test Steps
```
tox -e py3
```

## Merge type


cjp256:
nice! is this going to be the expected default in 24.04?

holmanb:
> nice! is this going to be the expected default in 24.04?

@cjp256 Yep! That's the plan. This and the InfiniBand PR are the two remaining code changes expected for dhcpcd in cloud-init.

cloudinit/net/dhcp.py
```diff
@@ -679,6 +703,38 @@ def dhcp_discovery(
...
 
+    @staticmethod
+    def parse_unknown_options_from_packet(
+        data: bytes, number: int
+    ) -> Optional[bytes]:
+        """get a specific option from a binary lease file
+
+        @param data: Binary lease data
+        @param number: Option number to return
+        @return: the option (bytes) or None
```
blackboxsw, line 714: Let's add a comment linking to your upstream dhcpcd issue here that this function is only needed temporarily until upstream dhcpcd supports option 245 https://github.com/NetworkConfiguration/dhcpcd/issues/282

cloudinit/net/dhcp.py
```diff
@@ -679,6 +703,38 @@ def dhcp_discovery(
...
+        """
+        # comes right after the magic cookie
```
blackboxsw, line 716: Can we reference the rfc that defined magic cookie at index 240?

Cross referencing other rfcs I see good mention of magic cookie as the first "marker" in the vend field but nothing specifically defining that index/position of the field.
https://www.ietf.org/rfc/rfc1497.txt and https://www.ietf.org/rfc/rfc2132.txt make reference to it, but not it's index.

I know this implementation is correct as I've poked around and tested everything from option 1,3,6 and 51 and seeing the expected netmask GW etc.

When looking through the following https://www.ietf.org/rfc/rfc951.txt I can see the bootstrap protocol which looks to define the `vend` field which looks to be nearly the same byte index 240 if I add up all the preceding fields.

blackboxsw: Maybe we just reference the struct here?https://github.com/NetworkConfiguration/dhcpcd/blob/master/src/dhcp.h#L167

cloudinit/sources/helpers/azure.py
```diff
@@ -223,20 +221,6 @@ def cd(newdir):
         os.chdir(prevdir)
 
 
-def get_ip_from_lease_value(fallback_lease_value):
```
blackboxsw, line 226: nvm. posted the comment without seeing the migration to `IscDhclient.get_ip_from_lease_value`. +1

cloudinit/net/dhcp.py
```diff
@@ -679,6 +703,38 @@ def dhcp_discovery(
             )
             raise NoDHCPLeaseError from error
 
+    @staticmethod
+    def parse_unknown_options_from_packet(
+        data: bytes, number: int
```
blackboxsw, line 708: nit: lets make this param name a bit more specific?
```suggestion
        data: bytes, dhcp_option_number: int
```

blackboxsw (approved):
+1 pending your review of Minor nits below and decision on integration test either way.

1. on explanatory comments or references explaining the index 240
2.  nit on param name in `parse_unknown_options_from_packet`. 
3.   Since we already have a `verify_clean_log` in our integration tests, this will have caught issues on the Azure platform https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/net/test_dhcp.py#L49... Given azure has a specific log of the parsed/processed lease. Do we want to add a check that we have a log on Azure containing the processed unknown-245 option?

The Azure logs will have the following format:
```
2024-02-06 17:45:47,275 - azure.py[DEBUG]: Obtained DHCP lease on interface 'eth0' (primary=True driver='hv_netvsc' router='10.0.0.1' routes=[('0.0.0.0/0', '10.0.0.1'), ('168.63.129.16/32', '10.0.0.1'), ('169.254.169.254/32', '10.0.0.1')] lease={'broadcast-address': '10.0.0.255', 'dhcp-lease-time': '4294967295', 'dhcp-message-type': '5', 'dhcp-rebinding-time': '4294967295', 'dhcp-renewal-time': '4294967295', 'dhcp-server-identifier': '168.63.129.16', 'domain-name': 'ba4rf02gucbeboddf4le52hkgf.gx.internal.cloudapp.net', 'domain-name-servers': '168.63.129.16', 'network-number': '10.0.0.0', 'routers': '10.0.0.1', 'server-name': 'DSM101050420009', 'subnet-cidr': '24', 'subnet-mask': '255.255.255.0', 'interface': 'eth0', 'fixed-address': '10.0.0.4', 'static_routes': '0.0.0.0/0 10.0.0.1 168.63.129.16/32 10.0.0.1 169.254.169.254/32 10.0.0.1', 'unknown-245': '168.63.129.16'} imds_routed=True wireserver_routed=True)
```
Approve pending supplemental comments and your decision on whether the integration tests needs to grow option 245 awareness.

cloudinit/sources/helpers/azure.py
holmanb: I'm happy to consolidate codepaths if that works. I just haven't tested yet and was just relocating the code so that arbitrary clients would work. 

I don't fully grok the parsing done in `get_ip_from_lease_value()` so I didn't want to mess with that for now, but I agree that it is probably a future improvement we could do.

cloudinit/net/dhcp.py
holmanb: > I can see the bootstrap protocol which looks to define the vend field which looks to be nearly the same byte index 240 if I add up all the preceding fields.

Yep, the vend field starts on 236, so if you add the 4 bytes that the magic cookie takes up that's where we get index 240. I've updated the comments to make that easier to discover for future readers of this code.

holmanb:
@blackboxsw thanks for the review, great feedback.

I think I've addressed all of your concerns in the update I just pushed.

blackboxsw (approved):

PR #4838, opened 2024-02-01
distro: add FreeBSD code path for eject

igalic:
## Proposed Commit Message

```
distro: add eject FreeBSD code path

OpenBSD, and NetBSD both have an eject(1), so they should be covered in
the default code path.

FreeBSD and Dragonfly however, do not have eject in base. Here, eject is
an (unmaintained) port. In base, we do however, have camcontrol(8) and
cdcontrol(1), both of which have an eject subcommand.

Let's use camcontrol(8) here.

```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -226,3 +226,7 @@ class Distro(cloudinit.distros.bsd.BSD):
         return [path, "-l", lease_file, "-p", pid_file] + (
             ["-c", config_file, interface] if config_file else [interface]
         )
+
+    @staticmethod
+    def eject_media(device: str) -> None:
+        subp.subp(["camcontrol", "eject", device])


holmanb (commented):
Since the Dragonflybsd class inherits from the Freebsd class this actually overrides dragonfly's eject_media() function as well, which I don't think we want.

I think we also need the following:
```diff
diff --git a/cloudinit/distros/dragonflybsd.py b/cloudinit/distros/dragonflybsd.py
index 5032ff5a7..9a7a53eee 100644
--- a/cloudinit/distros/dragonflybsd.py
+++ b/cloudinit/distros/dragonflybsd.py
@@ -2,8 +2,13 @@
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
-import cloudinit.distros.freebsd
+from cloudinit.distros import Distro as base
+from cloudinit.distros import freebsd
 
 
-class Distro(cloudinit.distros.freebsd.Distro):
+class Distro(freebsd.Distro):
     home_dir = "/home"
+
+    @staticmethod
+    def eject_media(device: str) -> None:
+        base.eject_media(device)
```

igalic:
yeah, we could do that, but dragonfly has [camcontrol(8)](https://man.dragonflybsd.org/?command=camcontrol&section=8).

cam is very old, as you can see from the history section.

more importantly: i have misread https://man.dragonflybsd.org/?command=eject&section=ANY this is just the same port (unmaintained) eject port: https://man.freebsd.org/cgi/man.cgi?eject

so, that commit message needs adapting.
I'll also add a test case. tomorrow.

holmanb:
> yeah, we could do that, but dragonfly has [camcontrol(8)](https://man.dragonflybsd.org/?command=camcontrol&section=8).
> 
> cam is very old, as you can see from the history section.
> 
> more importantly: i have misread https://man.dragonflybsd.org/?command=eject&section=ANY this is just the same (unmaintained) eject port: https://man.freebsd.org/cgi/man.cgi?eject
> 
> so, that commit message needs adapting. I'll also add a test case. tomorrow.

That sounds good, thanks @igalic!

igalic:
n.b.: Tomorrow has turned into much later, as I'm at a conference right now.

holmanb (approved):
LGTM

PR #4839, opened 2024-02-01
tests(dhcp): Add udhcpc client to test matrix

holmanb:
## Proposed Commit Message
```
tests(dhcp): Add udhcpc client to the integration test matrix

Udhcpc support is incomplete, however it does function.
Add it to the test matrix for improved test coverage.
```

### DIFF
--- a/tests/integration_tests/net/test_dhcp.py
+++ b/tests/integration_tests/net/test_dhcp.py
@@ -47,19 +47,19 @@ class TestDHCP:
         reason="earlier Ubuntu releases have a package named dhcpcd5",
     )
     @pytest.mark.parametrize(
-        "client, package",
+        "dhcp_client, package",
         [
             ("dhcpcd", "dhcpcd-base"),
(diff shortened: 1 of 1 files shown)


tests/integration_tests/net/test_dhcp.py
```diff
@@ -46,12 +46,20 @@ def test_noble_and_newer_uses_dhcpcd_by_default(self, client):
         CURRENT_RELEASE < NOBLE,
         reason="earlier Ubuntu releases have a package named dhcpcd5",
     )
-    def test_noble_and_newer_force_dhcp(self, client):
+    @pytest.mark.parametrize(
+        "client, package",
```
TheRealFalcon, line 50: Won't this override the client fixture? I think we need a different name than `client`.

TheRealFalcon (changes requested):
One change, but otherwise good

TheRealFalcon (approved):

PR #4840, opened 2024-02-02
Cleanup test_net.py

TheRealFalcon:
## Proposed Commit Message

```
test: cleanup test_net.py

* Move ~3500 lines of network config into its own file
* Replace unittest idioms with pytest idioms
* Remove CiTestCase and FilesystemMockingTestCase
* Parametrize the obvious tests
* Remove commented tests and debug print statements
```

## Additional Context
There's still more that could be done here, but I want to stop here after most of the "easy" changes.

Commits are technically independently applicable, but I think I'd rather squash merge the whole thing as my current commits are fairly noisy. A few of the commits will have some extra context in the commit message though.

## Merge type


blackboxsw (approved):
Much cleaner to look at and reason about. Thx for catching and documenting the failed/misspelled test-cases.

blackboxsw:
+1 on the squash here. it was helpful for the review to have separate commts, but it's not really a functional set of differences that I'd see future-me wanting to distinguish.

PR #4841, opened 2024-02-02
fix: Add types to network v1 schema

TheRealFalcon:
## Proposed Commit Message

```
fix: Add types to network v1 schema

Even though it has conflicted with our documentation, we have allowed
nameserver address to a be a string, mtu to be empty, and nameserver
search to be missing. Since we have allowed these, expand our schema
and documentation accordingly.

Fixes GH-4710
```

## Merge type

### DIFF
--- a/doc/rtd/reference/network-config-format-v1.rst
+++ b/doc/rtd/reference/network-config-format-v1.rst
@@ -251,9 +251,10 @@ Nameserver
 Users can specify a ``nameserver`` type. Nameserver dictionaries include
 the following keys:
 
-- ``address``: List of IPv4 or IPv6 address of nameservers.
-- ``search``: List of hostnames to include in the :file:`resolv.conf` search
-  path.
+- ``address``: List of IPv4 or IPv6 address of nameservers or string with
+  single nameserver.
+- ``search``: Optional. List of hostnames to include in the :file:`resolv.conf`
+  search path.
 - ``interface``: Optional. Ties the nameserver definition to the specified
   interface. The value specified here must match the ``name`` of an interface
   defined in this config. If unspecified, this nameserver will be considered
--- a/doc/rtd/reference/network-config-format-v1.rst
+++ b/doc/rtd/reference/network-config-format-v1.rst
@@ -251,8 +251,7 @@ Nameserver
 Users can specify a ``nameserver`` type. Nameserver dictionaries include
 the following keys:
 
-- ``address``: List of IPv4 or IPv6 address of nameservers or string with
-  single nameserver.
+- ``address``: List of IPv4 or IPv6 address of nameservers.
 - ``search``: Optional. List of hostnames to include in the :file:`resolv.conf`
   search path.
 - ``interface``: Optional. Ties the nameserver definition to the specified
--- a/cloudinit/config/schemas/schema-network-config-v1.json
+++ b/cloudinit/config/schemas/schema-network-config-v1.json
@@ -25,7 +25,10 @@
           "description": "The lowercase MAC address of the physical device."
         },
         "mtu": {
-          "type": "integer",
+          "type": [
+            "integer",
+            "null"
+          ],
           "description": "The MTU size in bytes. The ``mtu`` key represents a device's Maximum Transmission Unit, which is the largest size packet or frame, specified in octets (eight-bit bytes), that can be sent in a packet- or frame-based network. Specifying ``mtu`` is optional. Values too small or too large for a device may be ignored by that device."
         },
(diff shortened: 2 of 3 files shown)


doc/rtd/reference/network-config-format-v1.rst
```diff
@@ -251,9 +251,10 @@ Nameserver
 Users can specify a ``nameserver`` type. Nameserver dictionaries include
 the following keys:
 
-- ``address``: List of IPv4 or IPv6 address of nameservers.
-- ``search``: List of hostnames to include in the :file:`resolv.conf` search
-  path.
+- ``address``: List of IPv4 or IPv6 address of nameservers or string with
+  single nameserver.
```
holmanb, line 255: nit: documenting every special case for multiple types may be more complete, but the relative benefit to users for supporting multiple types seems pretty low relative to the benefit since it often leads to unclarity - a user might interpret "list or string of IPv4 address of nameservers" to mean the following could be valid:

`"1.1.1.1 8.8.8.8 4.4.4.4"`

I agree that when our schema is wrong, we should fix it to be correct, however I lean towards a leaner "one way to do it" preference for future features visible in the UI - and documentation that reflects that mentality as well.

holmanb (approved):
This sort of thing is inconsistently handled throughout cloud-init.

This fix looks looks good, so I'm +1 to the change, though in general I think I'd prefer follow the following guidelines in the future:

#### new keys:

1. always use plural key names in the case that "one or more items" is allowed (`"addresses"`, not "address")
2. allow one input type per key

the reason for this preference is that the overhead of extra schema complexity, confusion caused by multiple ways for users to do the same thing, resulting code complexity, and documentation required for such a "feature" to be of any use is a really high cost for the benefit it gives the user (two key presses: `[` and `]`).

#### existing keys:

1. Document the one generic case that always works (i.e.: list of items where one or more item is possible) rather than special-casing special key names for "single" and "multiple" keys. The intent is to prefer reducing user-facing complexity rather than to prefer giving (and documenting) multiple configuration formats for users to accomplish the same task.

@TheRealFalcon thoughts?

I'm fine with merging as is - I don't want to block this PR, but if you're in agreement generally with the principle I can toss a PR to document this preference going forward?

TheRealFalcon:
@holmanb , in general I agree (and I intentionally didn't document mtu), but I went back and forth on address for a few reasons:

1. The key is called "address" but takes a list. That on its own is confusing and might make more sense having a string option.
2. This is documentation for us as much as anybody else. The schema was generated from the docs. If our schema becomes our new source of source that's fine, but that is a bit of a shift.
3. We have tests for this behavior.

But long term your reasoning makes more sense, so I'll update the docs for this line accordingly.

holmanb (approved):

ani-sinha:
@TheRealFalcon  Is it possible to get a back-port or a fix for this for 23.4. We are struggling to backport it downstream on c9s.

ani-sinha:
> @TheRealFalcon Is it possible to get a back-port or a fix for this for 23.4. We are struggling to backport it downstream on c9s.

I had to do this in order to make unit tests work for 23.4. Hope this is ok.
```
diff --git a/tests/unittests/config/test_schema.py b/tests/unittests/config/test_schema.py
index 63247bd1..52667332 100644
--- a/tests/unittests/config/test_schema.py
+++ b/tests/unittests/config/test_schema.py
@@ -2059,7 +2059,6 @@ class TestNetworkSchema:
                     }
                 },
                 does_not_raise(),
-                "",
                 id="GH-4710_mtu_none_and_str_address",
             ),
         ),
```

TheRealFalcon:
> Is it possible to get a back-port or a fix for this for 23.4

Are you asking for a 23.4.4 or something else?

24.1 is planning to be released upstream next week, so at this point I don't think it makes sense to do another upstream hotfix. Does that work for you?

ani-sinha:
> > Is it possible to get a back-port or a fix for this for 23.4
> 
> Are you asking for a 23.4.4 or something else?
> 
> 24.1 is planning to be released upstream next week, so at this point I don't think it makes sense to do another upstream hotfix. Does that work for you?

On the downstream side we are stuck with 23.4 for now. So was wondering if there was a way to get the patch for 23.4. In any case, can you please comment on my change above? Other than that change, the patch applies cleanly.

TheRealFalcon:
> In any case, can you please comment on my change above?

Yes, your change makes sense. After 23.4 an option was added to check messages in log files, so that's why the parametrized test has an extra argument. It can be removed when cherry-picking the fix.

PR #4842, opened 2024-02-02
fix: Handle systemctl when dbus not ready

TheRealFalcon:
## Proposed Commit Message

```
fix: Handle systemctl when dbus not ready

See commit d29b744, but this commit generalizes the solution for all
calls in status.py
```

## Additional Context
#4681 #4676 LP: #2046483

## Checklist

## Merge type

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -93,6 +93,36 @@ boot_status_code: {boot_code}
 {description}"""
 
 
+def query_systemctl(
+    systemctl_args: List[str],
+    *,
+    wait: bool,
+    existing_status: Optional[UXAppStatus] = None,
+) -> str:
+    """Query systemd with retries and return output."""
+    while True:
+        try:
+            return subp.subp(["systemctl", *systemctl_args]).stdout.strip()
+        except subp.ProcessExecutionError as e:
+            if existing_status and existing_status in (
+                UXAppStatus.DEGRADED_RUNNING,
+                UXAppStatus.RUNNING,
+            ):
+                return ""
+            last_exception = e
+            if wait:
+                sleep(0.25)
+            else:
+                break
+    print(
+        "Failed to get status from systemd. "
+        "Cloud-init status may be inaccurate. ",
+        f"Error from systemctl: {last_exception.stderr}",
+        file=sys.stderr,
+    )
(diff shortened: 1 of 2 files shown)


cloudinit/cmd/status.py
```diff
@@ -93,6 +93,36 @@ class StatusDetails(NamedTuple):
 {description}"""
 
 
+def query_systemctl(
```
blackboxsw, line 96: This is a good utility function to generalize interaction with systemctl. While we see other call-sites throughout cloud-init to systemctl in cloudini/net/network_manager, cloudinit/net/activators.py, cloudinit/sources/helpers/azure.py and cloudinit/config/cc_mounts.py, these entry points are much later in boot than a potential early-boot invocation of cloud-init status from external consumers. So I don't necessarily think we "need" this logic elsewhere for this PR. It may be worth generalizing it in util or somewhere if we find another call-site that warrants early boot interaction and exposure to the socket not yet being responsive.

blackboxsw: The current implemntation is very status specific as well. So I don't think it's worth generalizing further at this time.

tests/unittests/cmd/test_status.py
```diff
@@ -1005,13 +999,70 @@ def test_retry_no_wait(self, mocker, capsys):
...
+
+    def test_query_systemctl_wait_with_exception_status(self, mocker):
+        m_sleep = mocker.patch(f"{M_PATH}sleep")
+        m_subp = mocker.patch(
+            f"{M_PATH}subp.subp",
+            side_effect=[
+                subp.ProcessExecutionError("Message recipient disconnected"),
+                subp.ProcessExecutionError("Message recipient disconnected"),
+                subp.ProcessExecutionError("Message recipient disconnected"),
+            ],
```
blackboxsw, line 1056: Can drop the other side_effects
```suggestion
            side_effect=subp.ProcessExecutionError("Message recipient disconnected"),
```

blackboxsw (approved):
+1 and minor unittest fix. Both otherwise looks good to me.

blackboxsw:
@TheRealFalcon please add the footer LP: #2046483 to the commit message when squashed/rebase merged.

PR #4843, opened 2024-02-02
fix(schema cli): avoid netplan validation on net-config version 1

blackboxsw:
## Proposed Commit Message take from rebase
```
fix(schema cli): avoid netplan validation on net-config version 1

Commit e168b4a1 moved network-config version 2 checking out of netplan_validate_network_schema and up into the callsite in netplan_validate_network_schema.

An oversight didn't apply the same version 2 check to the second callsite in validate_cloudconfig_file.

Add unittest coverage to supplement the integration test which already caught this failure in tests.integration_tests.cmd.test_schema.
```

## Additional Context

Failed jenkins job https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-noble-ec2/lastSuccessfulBuild/testReport/junit/tests.integration_tests.cmd.test_schema/TestSchemaDeprecations/test_network_config_schema_validation/
## Test Steps
```
# test main to see errors vs this branch for the fix
CLOUD_INIT_OS_IMAGE=noble CLOUD_INIT_CLOUD_INIT_SOURCE=IN_PLACE CLOUD_INIT_PLATFORM=lxd_container  tox -e integration-tests  -- tests/integration_tests/cmd/test_schema.py::TestSchemaDeprecations::test_network_config_schema_validation
```
## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -1119,11 +1119,13 @@ def validate_cloudconfig_file(
         if not cloudconfig.get("network", cloudconfig):
             print("Skipping network-config schema validation on empty config.")
             return False
-        if netplan_validate_network_schema(
-            network_config=cloudconfig, strict=True, annotate=annotate
-        ):
-            return True  # schema validation performed by netplan
-        if network_schema_version(cloudconfig) != 1:
+        network_version = network_schema_version(cloudconfig)
+        if network_version == 2:
+            if netplan_validate_network_schema(
+                network_config=cloudconfig, strict=True, annotate=annotate
+            ):
+                return True  # schema validation performed by netplan
+        if network_version != 1:
             # Validation requires JSON schema definition in
             # cloudinit/config/schemas/schema-network-config-v1.json
             print(
--- a/tests/unittests/config/test_schema.py
+++ b/tests/unittests/config/test_schema.py
@@ -1892,6 +1892,14 @@ class TestMain:
                 "Skipping network-config schema validation. No network schema"
                 " for version: 2",
             ),
+            (
+                "network-config",
+                (
+                    b"network:\n version: 1\n config:\n  - type: physical\n"
+                    b"    name: eth0\n    subnets:\n      - type: dhcp\n"
+                ),
+                "Valid schema",
+            ),
         ),
     )
     def test_main_validates_config_file(


cloudinit/config/schema.py
```diff
@@ -1119,11 +1119,13 @@ def validate_cloudconfig_file(
...
-        ):
-            return True  # schema validation performed by netplan
-        if network_schema_version(cloudconfig) != 1:
+        network_version = network_schema_version(cloudconfig)
+        if network_version == 2:
+            if netplan_validate_network_schema(
+                network_config=cloudconfig, strict=True, annotate=annotate
+            ):
+                return True  # schema validation performed by netplan
+        if network_version != 1:
```
TheRealFalcon, line 1128: Nit: Do we need this `if` statement? Everything under it will be under the `if network_version == 2` block, which I think would be adequate?

TheRealFalcon (approved):
Just one question inline, but I don't think it's major enough warranting a re-push.

cloudinit/config/schema.py
blackboxsw: On systems without netplan but with network-config version 2, the `netplan_validate_network_schema` will have returned False on line 1123. In those cases, we want to fail with an informative message like `No network schema for version 2`. Instead of a json schema validation error that isn't as informative `Invalid network schema: 'version' not in [1]`. Also, in the case we are version 1 here, we want to fall through to the standard `validate_cloudconfig_schema` below to actually validate that network-config v1 file instead of claiming to skip it and early return

TheRealFalcon: Yeah, I think I understand that. I'm just saying we have this now:
```python
        if network_version == 2:
            if netplan_validate_network_schema(
                network_config=cloudconfig, strict=True, annotate=annotate
            ):
                return True  # schema validation performed by netplan
        if network_version != 1:
            # Validation requires JSON schema definition in
            # cloudinit/config/schemas/schema-network-config-v1.json
            print(
                "Skipping network-config schema validation."
                " No network schema for version:"
                f" {network_schema_version(cloudconfig)}"
            )
            return False
```

Isn't `if network_version == 2` effectively the same as `if network_version != 1`? Couldn't it be one block?
```python
        if network_version == 2:
            if netplan_validate_network_schema(
                network_config=cloudconfig, strict=True, annotate=annotate
            ):
                return True  # schema validation performed by netplan
            # Validation requires JSON schema definition in
            # cloudinit/config/schemas/schema-network-config-v1.json
            print(
                "Skipping network-config schema validation."
                " No network schema for version:"
                f" {network_schema_version(cloudconfig)}"
            )
            return False
```

PR #4848, opened 2024-02-02
Ubuntu/mantic 23.4.x

TheRealFalcon:
DO NOT SQUASH


blackboxsw (approved):
build-package/sbuild and d/changelog version are good here for this sync. Already uploaded to unapproved mantic upload queue and pinged SRU release managers.

PR #4851, opened 2024-02-05
Noble hotplug udev

aciba90:
## Do not squash

Rename /etc/udev/rules.d/10-cloud-init-hook-hotplug.rules to /etc/udev/rules.d/99-cloud-init-hook-hotplug.rules after https://github.com/canonical/cloud-init/pull/4799


blackboxsw (approved):
Applied same suggested diff to pre-flight check presence of hotplug file before moving it to avoid errors in postinst

PR #4853, opened 2024-02-05
tests: add status --wait blocking test from early boot

blackboxsw:
## rebase merge commit

```
tests: add status --wait blocking test from early boot

In bugs such as LP: #2046483, cloud-init status --wait produces a traceback early when dbus or other resources are not yet available resulting in an early unblock from status --wait.

Add an integration test inserting a systemd unit before cloud-init-local.service that will block until cloud-init is done.

Assert that the timestamp of marker file created when this test unit unblocks, occurs after cloud-init.target was reached.
```

## Additional Context

Supplemental test coverage to make sure future changes to cloud-init boot stages don't break cloud-init --wait expected behavior.

## Test Steps
```
CLOUD_INIT_KEEP_INSTANCE=1 CLOUD_INIT_OS_IMAGE=noble CLOUD_INIT_CLOUD_INIT_SOURCE=IN_PLACE tox -e integration-tests -- tests/integration_tests/cmd/test_status.py::test_status_block_through_all_boot_status
```

## Checklist

## Merge type

### DIFF
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -4,6 +4,7 @@ import json
 import pytest
 
 from tests.integration_tests.clouds import IntegrationCloud
+from tests.integration_tests.decorators import retry
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.integration_settings import PLATFORM
 from tests.integration_tests.releases import CURRENT_RELEASE, IS_UBUNTU, JAMMY
@@ -17,6 +18,12 @@ def _remove_nocloud_dir_and_reboot(client: IntegrationInstance):
     old_boot_id = client.instance.get_boot_id()
     client.execute("cloud-init clean --logs --reboot")
     client.instance._wait_for_execute(old_boot_id=old_boot_id)
+
+
+@retry(tries=5, delay=1)
+def retry_read_from_file(client: IntegrationInstance, path: str):
+    """Retry read_from_file expecting it shortly"""
+    return client.read_from_file(path)
 
 
 @pytest.mark.skipif(not IS_UBUNTU, reason="Only ever tested on Ubuntu")
@@ -82,3 +89,73 @@ def test_status_json_errors(client):
     assert "Invalid cloud-config provided" in json.loads(status_json)[
         "recoverable_errors"
     ].get("WARNING").pop(0)
+
+
+EARLY_BOOT_WAIT_USER_DATA = """\
+#cloud-config
+runcmd: [systemctl enable before-cloud-init-local.service]
+write_files:
+- path: /waitoncloudinit.sh
+  permissions: '0755'
+  content: |
+    #!/bin/sh
+    if [ -f /var/lib/cloud/data/status.json ]; then
+        MARKER_FILE="/$1.start-hasstatusjson"
+    else
+        MARKER_FILE="/$1.start-nostatusjson"
+    fi
+    cloud-init status --wait --long > $1
+    date +%s.%N > $MARKER_FILE
+- path: /lib/systemd/system/before-cloud-init-local.service
+  permissions: '0644'
+  content: |
+    [Unit]
+    Description=BEFORE cloud-init local
+    DefaultDependencies=no
+    Before=cloud-init-local.service
+    Before=shutdown.target
+    Before=sysinit.target
+    Conflicts=shutdown.target
+    RequiresMountsFor=/var/lib/cloud
+
+    [Service]
+    Type=simple
+    ExecStart=/waitoncloudinit.sh /before-local
+    RemainAfterExit=yes
+    TimeoutSec=0
+
+    # Output needs to appear in instance console output
+    StandardOutput=journal+console
+
+    [Install]
+    WantedBy=cloud-init.target
+"""  # noqa: E501
+
+
+@pytest.mark.user_data(EARLY_BOOT_WAIT_USER_DATA)
+@pytest.mark.lxd_use_exec
+@pytest.mark.skipif(
+    PLATFORM not in ("lxd_container", "lxd_vm"),
+    reason="Requires use of lxd exec",
+)
+def test_status_block_through_all_boot_status(client):
+    """Assert early boot cloud-init status --wait does not exit early."""
+    client.execute("cloud-init clean --logs --reboot")
+    wait_for_cloud_init(client).stdout.strip()
+    client.execute("cloud-init status --wait")
+
+    # Assert that before-cloud-init-local.service started before
+    # cloud-init-local.service could create status.json
+    client.execute("test -f /before-local.start-hasstatusjson").failed
+
+    early_unit_timestamp = retry_read_from_file(
+        client, "/before-local.start-nostatusjson"
+    )
+    # Assert the file created at the end of
+    # before-cloud-init-local.service is newer than the last log entry in
+    # /var/log/cloud-init.log
+    events = json.loads(client.execute("cloud-init analyze dump").stdout)
+    final_cloud_init_event = events[-1]["timestamp"]
+    assert final_cloud_init_event < float(
+        early_unit_timestamp
+    ), "Systemd unit didn't block on cloud-init status --wait"
--- a/tests/integration_tests/decorators.py
+++ b/tests/integration_tests/decorators.py
@@ -20,7 +20,7 @@ def retry(*, tries: int = 30, delay: int = 1):
             last_error = None
             for _ in range(tries):
                 try:
-                    func(*args, **kwargs)
+                    retval = func(*args, **kwargs)
                     break
                 except Exception as e:
                     last_error = e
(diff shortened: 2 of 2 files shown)


tests/integration_tests/cmd/test_status.py
```diff
@@ -82,3 +82,57 @@ def test_status_json_errors(client):
...
+    wait_for_cloud_init(client).stdout.strip()
+    status_json = client.execute("cloud-init status --wait --format=json")
+
+    # Assert that before-cloud-init-local.service started before
+    # cloud-init-local.service could create status.json
+    assert client.execute("test -f /before-local.start-nostatusjson")
+    # Assert the file created at the end of
+    # before-cloud-init-local.service is newer than the last log entry in
+    # /var/log/cloud-init.log
+    assert client.execute("test /before-local -nt /var/log/cloud-init.log").ok
```
holmanb, line 137: Depending on timestamps could be unreliable since reads may update timestamps for any filesystems mounted without `noatime` or `A` attribute set, since any external tool that reads `/var/log/cloud-init.log` would update the timestamp of the file. Maybe the file could contain a time for a more robust test?

tests/integration_tests/cmd/test_status.py
```diff
@@ -82,3 +82,57 @@ def test_status_json_errors(client):
...
+    print(status_json)
```
holmanb, line 138: We shouldn't need this print statement.

tests/integration_tests/cmd/test_status.py
```diff
@@ -82,3 +82,57 @@ def test_status_json_errors(client):
...
+    # Output needs to appear in instance console output
+    StandardOutput=journal+console
+
+    [Install]
+    WantedBy=cloud-init.target
+"""  # noqa: E501
+
+
+@pytest.mark.user_data(EARLY_BOOT_WAIT_USER_DATA)
+@pytest.mark.lxd_use_exec
```
TheRealFalcon, line 124: We should add a skip if not LXD here. Any other platform we need to wait for ssh access to get in and run the command.

tests/integration_tests/cmd/test_status.py
```diff
@@ -82,3 +82,57 @@ def test_status_json_errors(client):
...
+def test_status_block_through_all_boot_status(client):
```
TheRealFalcon, line 125: Using the [lxd_setup mark](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/modules/test_disk_setup.py#L56) would allow us to test this first boot and avoid the need for a reboot. It would require a few `lxc push`s rather than userdata. Otherwise, I could see failures due to following statements getting in before the reboot happens.

TheRealFalcon (changes requested):
Thanks for this! I left a few inline comments.

tests/integration_tests/cmd/test_status.py
blackboxsw: Good suggestion, I've adapted the unit to echo an ending timestamp and validate that the time is later than the timestamp of the last log exposed by `cloud-init analyze dump`.

tests/integration_tests/cmd/test_status.py
blackboxsw: added.

tests/integration_tests/cmd/test_status.py
blackboxsw: @TheRealFalcon, thanks. Sorry to take so long on this part. I did instrument a whole separate test with lxd_setup_mark but found that it is challening to inject a new systemd unit and enable it via symlinks prior to first boot without also triggering a `systemctl daemon-reload` which we can't do until the system is up and running.

I toyed around with something like the following patch and kept hitting issues with the before-cloud-init-local.service never being enabled
```diff
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -1,5 +1,10 @@
 """Tests for `cloud-init status`"""
 import json
+import os
+from cloudinit.subp import subp
+from cloudinit.util import write_file
+from pycloudlib.lxd.instance import LXDInstance
+from tempfile import TemporaryDirectory
 
 import pytest
 
@@ -91,6 +96,104 @@ def test_status_json_errors(client):
     ].get("WARNING").pop(0)
 
 
+
+
+EARLY_BOOT_GENERATOR = """\
+#!/bin/sh
+# Provide a symlink that enables before-cloud-init-local.service
+UNIT_NAME="before-cloud-init-local.service"
+LINK_PATH="/etc/systemd/system/multi-user.target.wants/$UNIT_NAME"
+UNIT_PATH="/lib/systemd/system/$UNIT_NAME"
+[ -d "${LINK_PATH%/*}" ] || mkdir -p "${LINK_PATH%/*}"
+cd /etc/systemd/system/multi-user.target.wants
+ln -snf "${UNIT_PATH}" "${LINK_PATH}"
+exit 0
+"""
+
+EARLY_BOOT_SYSTEMD_UNIT = """\
+[Unit]
+Description=BEFORE cloud-init local
+DefaultDependencies=no
+Before=cloud-init-local.service
+Before=shutdown.target
+Before=sysinit.target
+Conflicts=shutdown.target
+RequiresMountsFor=/var/lib/cloud
+
+[Service]
+Type=simple
+ExecStart=/waitoncloudinit.sh /before-local
+RemainAfterExit=yes
+TimeoutSec=0
+
+# Output needs to appear in instance console output
+StandardOutput=journal+console
+
+[Install]
+WantedBy=cloud-init.target
+"""
+
+EARLY_BOOT_SCRIPT = """\
+#!/bin/sh
+if [ -f /var/lib/cloud/data/status.json ]; then
+    MARKER_FILE="/$1.start-hasstatusjson"
+else
+    MARKER_FILE="/$1.start-nostatusjson"
+fi
+cloud-init status --wait --long > $1
+date +%s.%N > $MARKER_FILE
+"""
+
+
+
+def setup_early_systemd_unit(instance: LXDInstance):
+    """Add systemd unit Before=cloud-init-local.service."""
+    with TemporaryDirectory() as tmp_dir:
+        generator_file = os.path.join(tmp_dir, "before-cloud-init-generator")
+        write_file(generator_file, EARLY_BOOT_GENERATOR, mode=0o755)
+        subp(
+            "lxc file push {} {}/lib/systemd/system-generators/".format(
+                generator_file, instance.name
+            ).split()
+        )
+        unit_file = os.path.join(tmp_dir, "before-cloud-init-local.service")
+        write_file(unit_file, EARLY_BOOT_SYSTEMD_UNIT)
+        subp(
+            "lxc file push {} {}/lib/systemd/system/".format(
+                unit_file, instance.name
+            ).split()
+        )
+        unit_script = os.path.join(tmp_dir, "waitoncloudinit.sh")
+        write_file(unit_script, EARLY_BOOT_SCRIPT, mode=0o755)
+        subp(f"lxc file push {unit_script} {instance.name}/".split())
+
+
+@pytest.mark.lxd_setup.with_args(setup_early_systemd_unit)
+@pytest.mark.skipif(
+    PLATFORM not in ("lxd_container", "lxd_vm"),
+    reason="Test requires lxc setup before launch",
+)
+def test_try_in_lxd_init(client: IntegrationInstance):
+    # Assert that before-cloud-init-local.service started before
+    # cloud-init-local.service could create status.json
+    client.execute("cloud-init status --wait")
+    client.execute("test -f /before-local.start-hasstatusjson").failed
+    assert "enabled" in client.execute("systemctl show -p UnitFileState before-cloud-init-local.service").stdout
+    early_unit_timestamp = retry_read_from_file(
+        client, "/before-local.start-nostatusjson"
+    )
+    # Assert the file created at the end of
+    # before-cloud-init-local.service is newer than the last log entry in
+    # /var/log/cloud-init.log
+    events = json.loads(client.execute("cloud-init analyze dump").stdout)
+    final_cloud_init_event = events[-1]["timestamp"]
+    assert final_cloud_init_event < float(
+        early_unit_timestamp
+    ), "Systemd unit didn't block on cloud-init status --wait"
+
+
+
+
 EARLY_BOOT_WAIT_USER_DATA = """\
 #cloud-config
 runcmd: [systemctl enable before-cloud-init-local.service]

```

TheRealFalcon: Ah sorry, I forgot that systemd requires a symlink and not just a file in the right location.

It can be hackily done like this:
```
# In an empty directory...
cow:a $ touch before-cloud-init-local.service  # Use the actual file here                  
cow:a $ mkdir cloud-init.target.wants                      
cow:a $ ln -s ../before-cloud-init-local.service cloud-init.target.wants/before-cloud-init-local.service
cow:a $ lxc init ubuntu:jammy test
Creating test
cow:a $ lxc file push -r * test/etc/systemd/system
cow:a $ lxc start test 
```

blackboxsw: Thanks for the suggestion there, TIL about pushing pre-created symlinks into the unstarted instance. I also limited this test to lxd_container platform as we can't lxc file push into an lxd_vm

TheRealFalcon (approved):
LGTM (when you fix the linters)!

PR #4855, opened 2024-02-05
add expected dhcp6 key

holmanb:
## Proposed Commit Message
```
test: Add expected dhcp6 key
```

## Additional Context
@TheRealFalcon think it is safe to ignore this one? Also this one strikes me as one to maybe run as part of ci, but not sure we should really add more to ci runtime.

https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-gce/384/testReport/junit/tests.integration_tests/test_upgrade/test_clean_boot_of_upgraded_package/

## Test Steps

Not yet tested

## Merge type

### DIFF
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -3,6 +3,7 @@ import logging
 import os
 
 import pytest
+import yaml
 
 from tests.integration_tests.clouds import IntegrationCloud
 from tests.integration_tests.conftest import get_validated_source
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
Heh...so this got added then reverted and instead of adding it again I thought we could wait till 23.4 released...but that's proving to have been a bad idea. So yes, we should add it here.

I don't think this solution is complete though. Most cloud's won't give you a fallback config, so this shouldn't actually be true on most clouds.
https://github.com/canonical/cloud-init/pull/4612/files

As far as moving it to CI, I think that's probably a good idea, though I think we could probably make this one CI, and the one further down in the file not CI...or maybe just remove that one entirely.

holmanb:
> I don't think this solution is complete though. Most cloud's won't give you a fallback config, so this shouldn't actually be true on most clouds.
> https://github.com/canonical/cloud-init/pull/4612/files

Good point, thanks. I just pushed with that fix.

> As far as moving it to CI, I think that's probably a good idea, though I think we could probably make this one CI, and the one further down in the file not CI...or maybe just remove that one entirely.

I just realized that the test I was proposing moving to CI was Bionic-only, and the other one is release agnostic, so nevermind on that proposal.

TheRealFalcon (approved):

PR #4859, opened 2024-02-07
handle error when log file is empty

AliyevH:
## Proposed Commit Message

```
feat: handle error when log file is empty

Fixes GH-4686
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/analyze/show.py
+++ b/cloudinit/analyze/show.py
@@ -371,7 +371,7 @@ def load_events_infile(infile):
     :return: json version of logfile, raw file
     """
     data = infile.read()
-    if not data:
+    if not data.strip():
         sys.stderr.write("Empty file %s\n" % infile.name)
         sys.exit(1)
     try:
--- a/cloudinit/analyze/show.py
+++ b/cloudinit/analyze/show.py
@@ -7,6 +7,7 @@
 import datetime
 import json
 import os
+import sys
 import time
 
 from cloudinit import subp, util
@@ -370,6 +371,9 @@ def load_events_infile(infile):
     :return: json version of logfile, raw file
     """
     data = infile.read()
+    if not data:
+        sys.stderr.write("Empty file %s\n" % infile.name)
+        sys.exit(1)
     try:
         return json.loads(data), data
     except ValueError:
--- a/tests/unittests/analyze/test_show.py
+++ b/tests/unittests/analyze/test_show.py
@@ -1,5 +1,7 @@
-import pytest
 from collections import namedtuple
+
+import pytest
+
 from cloudinit.analyze import analyze_show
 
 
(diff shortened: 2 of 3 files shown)


AliyevH:
@a-dubs Already signed CLA several times. Don't know why it does not work :)
Added username and proposed commit message.
Thanks!

a-dubs:
### Adding this to make linking back to issue easy
Fixes: 4686

TheRealFalcon (commented):
@AliyevH , thanks for the contribution. While this does solve the problem, adding a `sys.exit()` into what is essentially a helper function several calls deep isn't ideal. It doesn't allow for any recovery for any other caller that might call the function.

Instead, can we raise a custom exception here (something like `EmptyFileException`), and then catch that exception at the bottom of `cloudinit/analyze/__init__.py` and print the proper text there?

AliyevH:
@TheRealFalcon Thanks for review!
1. sys exit is called in `load_events_infile` function.
2. It has been called several times only in `show.py`
3. If file is empty i think it does not make sense to continue.
4. `cloudinit/analyze/__init__.py` here we can not catch exception, because this file is not called directly. 
`if __name__ == "__main__":`
(sorry, maybe i am wrong. Trying to understand architecture. )

cloudinit/analyze/show.py
```diff
@@ -370,6 +371,9 @@ def load_events_infile(infile):
     :return: json version of logfile, raw file
     """
     data = infile.read()
+    if not data:
```
TheRealFalcon, line 374: Can we add a `.strip()` here?

TheRealFalcon (changes requested):
> cloudinit/analyze/__init__.py here we can not catch exception, because this file is not called directly.

Sorry about that. You're right. It's actually called from `main.py`. 

I still don't love having a `sys.exit(1)` here, but doing it a different way would require some more substantial refactors which isn't fair to ask for in this PR. It is ok as-as.

I left one inline comment and can we also get a test? Since we don't already have a unit test file for `cloud-init analyze show`, you can create one at `tests/unittests/analyze/test_show.py`. It should look something like

```python

import pytest

from cloudinit.analyze import analyze_show


@pytest.fixture
def mock_io(tmp_path):
    """Mock args for configure_io function"""
    infile = tmp_path / "infile"
    outfile = tmp_path / "outfile"
    return namedtuple("MockIO", ["infile", "outfile"])(infile, outfile)


class TestAnalyzeShow:
    """Test analyze_show (and/or helpers) in cloudinit/analyze/__init__.py"""

    def test_empty_logfile(self, mock_io, capsys):
        """Test analyze_show with an empty logfile"""
        mock_io.infile.write_text("")
        with pytest.raises(SystemExit):
            analyze_show("dontcare", mock_io)
        assert capsys.readouterr().err == f"Empty file {mock_io.infile}\n"

```

cloudinit/analyze/show.py
AliyevH: Sure. I will add strip test.

TheRealFalcon (approved):
LGTM! Thanks for the contribution.

AliyevH:
Thank you for guiding me !

PR #4864, opened 2024-02-08
Refactor status.py

TheRealFalcon:
## Proposed Commit Message

```
refactor: Refactor status.py

The one major functional change in this commit is around how we detect
running vs error states. Status reporting has a fundamental problem in
that we can't accurately tell if cloud-init is done because cloud-init
is actually several processes. There isn't always a way to tell whether
a service isn't running because it simply hasn't started yet vs the
service being blocked/crashed and will never start/finish.

In the past, if any of the cloud-init services reported an error, we
would assume that cloud-init as a whole has crashed and report that
cloud-init is "done", but with error. This commit flips that logic to
assume that cloud-init is always running unless we see indication that
cloud-init has completely finished. This means that
`cloud-init status --wait` may run forever if cloud-init has crashed or
is blocked on another service. This is preferable to returning early
and potentially allowing provisioning scripts that wait for cloud-init
to continue provisioning. On systemd-enabled systems, there is extra
logic to inspect the state of the services, so this should rarely be a
problem in practice.

Additionally, this commit includes the following refactoring:
- Split UXAppStatus into RunningStatus and ConditionStatus so they can
  be tracked independently
- Simplify the tabular printing
- On error print extended_status as "error - done" or "error - running"
- Add several helper functions in `get_status_details` to simplify
  logic
- Rename `_get_error_or_running_from_systemd` to `systemd_failed`
  and only return if error is detected
- Change "is running" logic to be determined solely by the existence
  of the status.json and results.json files.
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -13,7 +13,7 @@ import os
 import sys
 from copy import deepcopy
 from time import gmtime, sleep, strftime
-from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 
 from cloudinit import safeyaml, subp
 from cloudinit.cmd.devel import read_cfg_paths
@@ -24,35 +24,27 @@ from cloudinit.util import get_cmdline, load_json, load_text_file
 CLOUDINIT_DISABLED_FILE = "/etc/cloud/cloud-init.disabled"
 
 
-# customer visible status messages
 @enum.unique
-class UXAppStatus(enum.Enum):
+class RunningStatus(enum.Enum):
     """Enum representing user-visible cloud-init application status."""
 
-    NOT_RUN = "not run"
+    NOT_STARTED = "not started"
     RUNNING = "running"
     DONE = "done"
-    ERROR = "error"
-    DEGRADED_DONE = "degraded done"
-    DEGRADED_RUNNING = "degraded running"
     DISABLED = "disabled"
 
 
-# Extend states when degraded
-UXAppStatusDegradedMap = {
-    UXAppStatus.RUNNING: UXAppStatus.DEGRADED_RUNNING,
-    UXAppStatus.DONE: UXAppStatus.DEGRADED_DONE,
-}
+@enum.unique
+class ConditionStatus(enum.Enum):
+    """Enum representing user-visible cloud-init condition status."""
 
-# Map extended states back to simplified states
-UXAppStatusDegradedMapCompat = {
-    UXAppStatus.DEGRADED_RUNNING: UXAppStatus.RUNNING,
-    UXAppStatus.DEGRADED_DONE: UXAppStatus.DONE,
-}
+    ERROR = "error"  # cloud-init exited abnormally
+    DEGRADED = "degraded"  # we have warnings
+    PEACHY = "healthy"  # internal names can be fun, right?
 
 
 @enum.unique
-class UXAppBootStatusCode(enum.Enum):
+class EnabledStatus(enum.Enum):
     """Enum representing user-visible cloud-init boot status codes."""
 
     DISABLED_BY_GENERATOR = "disabled-by-generator"
@@ -67,17 +59,18 @@ class UXAppBootStatusCode(enum.Enum):
 
 DISABLED_BOOT_CODES = frozenset(
     [
-        UXAppBootStatusCode.DISABLED_BY_GENERATOR,
-        UXAppBootStatusCode.DISABLED_BY_KERNEL_CMDLINE,
-        UXAppBootStatusCode.DISABLED_BY_MARKER_FILE,
-        UXAppBootStatusCode.DISABLED_BY_ENV_VARIABLE,
+        EnabledStatus.DISABLED_BY_GENERATOR,
+        EnabledStatus.DISABLED_BY_KERNEL_CMDLINE,
+        EnabledStatus.DISABLED_BY_MARKER_FILE,
+        EnabledStatus.DISABLED_BY_ENV_VARIABLE,
     ]
 )
 
 
 class StatusDetails(NamedTuple):
-    status: UXAppStatus
-    boot_status_code: UXAppBootStatusCode
+    running_status: RunningStatus
+    condition_status: ConditionStatus
+    boot_status_code: EnabledStatus
     description: str
     errors: List[str]
     recoverable_errors: Dict[str, List[str]]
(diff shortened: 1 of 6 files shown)


cloudinit/cmd/status.py
```diff
@@ -425,36 +441,70 @@ def get_status_details(
...
+        running_status == RunningStatus.RUNNING
+        and uses_systemd()
+        and systemd_failed(wait=wait)
+    ):
+        running_status = RunningStatus.DONE
+        condition_status = ConditionStatus.ERROR
+        description = "Failed due to systemd unit failure"
+        errors.append(
+            "Failed due to sysetmd unit failure. Ensure all cloud-init "
+            "services are enabled, and check 'systemctl' or 'journalctl' "
```
holmanb, line 498: Which service to check is unknown to the user by this message, but we could easily return the service name that failed from `systemd_failed()` for a more informative error message.

I'm not requesting this change now, but it's an idea for a future improvement we could make.

cloudinit/cmd/status.py
```diff
@@ -425,36 +441,70 @@ def get_status_details(
...
+    elif recoverable_errors:
+        condition_status = ConditionStatus.DEGRADED
+
+    running_status = get_running_status(
+        status_file, result_file, boot_status_code, latest_event
+    )
+
+    if (
+        running_status == RunningStatus.RUNNING
+        and uses_systemd()
```
holmanb, line 490: We might eventually want to abstract `systemd_failed()` into the `distros/` so that other init systems can be better extended to check status. This isn't something we need to do today, just an idea for the future.

cloudinit/cmd/status.py
```diff
@@ -353,69 +346,92 @@ def _get_error_or_running_from_systemd(
...
+    return os.path.exists(status_file) and not os.path.exists(result_file)
 
-    @param paths: An initialized cloudinit.helpers.paths object.
 
-    Values are obtained from parsing paths.run_dir/status.json.
+def get_running_status(
+    status_file, result_file, boot_status_code, latest_event
+) -> RunningStatus:
+    """Return the running status of cloud-init."""
+    if is_running(status_file, result_file):
```
holmanb, line 371: this logic normally should work, but I think it would be slightly safer to do `if boot_status_code in DISABLED_BOOT_CODES` before `is_running(status_file, result_file)`. This is because I think that `boot_status_code` is a higher confidence signal than the file existence. 

A system that experiences a cloud-init crash that leaves behind a status file but not a result file will return `RunningStatus.RUNNING` (non-systemd). While this makes sense, if cloud-init is subsequently disabled we will still report `RunningStatus.RUNNING`.

cloudinit/cmd/status.py
```diff
@@ -250,72 +236,80 @@ def handle_status_args(name, args) -> int:
...
-    """Get if systemd is in error or running state.
-
-    Using systemd, we can get more fine-grained status of the
-    individual unit. Determine if we're still
-    running or if there's an error we haven't otherwise detected.
-
-    If we don't detect error or running, return None as we don't want to
-    report any other particular status based on systemd.
-    """
+def systemd_failed(wait: bool) -> bool:
```
holmanb, line 311: much more intuitive name and signature to understand +1

cloudinit/cmd/status.py
```diff
@@ -425,36 +441,70 @@ def get_status_details(
...
+    if (
+        running_status == RunningStatus.RUNNING
+        and uses_systemd()
+        and systemd_failed(wait=wait)
+    ):
+        running_status = RunningStatus.DONE
+        condition_status = ConditionStatus.ERROR
+        description = "Failed due to systemd unit failure"
+        errors.append(
+            "Failed due to sysetmd unit failure. Ensure all cloud-init "
```
holmanb, line 497: spelling

cloudinit/cmd/status.py
```diff
@@ -89,26 +82,21 @@ class StatusDetails(NamedTuple):
...
     *,
     wait: bool,
-    existing_status: Optional[UXAppStatus] = None,
 ) -> str:
     """Query systemd with retries and return output."""
     while True:
         try:
             return subp.subp(["systemctl", *systemctl_args]).stdout.strip()
         except subp.ProcessExecutionError as e:
-            if existing_status and existing_status in (
```
holmanb, line 107: After this change, the message to stderr below might get printed once from both `get_status_details()::get_bootstatus()` and again in `get_status_details()::systemd_failed()` when `cloud-init status` is called early in boot. Is this intentional? +1 either way, but it just seems unnecessary from a user's perspective.

cloudinit/cmd/status.py
```diff
@@ -24,35 +24,27 @@
...
+    """Enum representing user-visible cloud-init condition status."""
 
-# Map extended states back to simplified states
-UXAppStatusDegradedMapCompat = {
-    UXAppStatus.DEGRADED_RUNNING: UXAppStatus.RUNNING,
-    UXAppStatus.DEGRADED_DONE: UXAppStatus.DONE,
-}
+    ERROR = "error"  # cloud-init exited abnormally
+    DEGRADED = "degraded"  # we have warnings
+    PEACHY = "healthy"  # internal names can be fun, right?
```
holmanb, line 43: hehe, +1

cloudinit/cmd/status.py
```diff
@@ -163,82 +151,80 @@ def get_parser(parser=None):
...
+    running: RunningStatus, condition: ConditionStatus
+) -> Tuple[str, str]:
+    """Translate running and condition status to human readable strings.
+
+    Returns (status, extended_status).
+    Much of this is for backwards compatibility
+    """
+    # If we're done and have errors, we're in an error state
+    if condition == ConditionStatus.ERROR:
+        return ("error", f"{condition.value} - {running.value}")
```
holmanb, line 164: style question: Was it intentional to use parenthesis on this line and the next return but not the final one? I personally prefer never using them when returning tuples, but I'm curious if there is some stylistic reason for doing it this way?

holmanb (changes requested):
Thanks @TheRealFalcon, huge maintainability improvements here. Thanks for doing this. I have a couple of minor change requests inline - mostly just comments and questions.

We have a couple of strings that still need to be updated in the docs, comments, and tests:
```bash
$ rg "degraded running"
cloudinit/cmd/status.py
165:    # Handle the "degraded done" and "degraded running" states

doc/rtd/howto/status.rst
58:    "degraded running"

$ rg "degraded done"
doc/rtd/howto/status.rst
43:      "extended_status": "degraded done",
57:    "degraded done"

doc/rtd/explanation/exported_errors.rst
27:      "extended_status": "degraded done",
83:      "extended_status": "degraded done",

cloudinit/cmd/status.py
165:    # Handle the "degraded done" and "degraded running" states

tests/unittests/cmd/test_status.py
671:                    "extended_status": "degraded done",
```

cloudinit/cmd/status.py
TheRealFalcon: Nope, just lack of conscious thought. I was also notorious at a previous company for mixing double quotes and single quotes in the same file/function, but thankfully black has made that a non-issue for me. :smile: 

I removed the parens from these.

cloudinit/cmd/status.py
TheRealFalcon: I refactored it some to account for this. Let me know if that works for you

TheRealFalcon:
Also force pushed a rebase

holmanb (approved):
LGTM, thanks @TheRealFalcon!

PR #4873, opened 2024-02-13
Don't warn if cloud-init is disabled

holmanb:
## Proposed Commit Message
Use individual commit messages

## Additional Context
Also relocate apport functionality to main branch for better testing and less code duplication.

Fixes https://github.com/canonical/cloud-init/issues/4875, which is [this bug](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2025376) which was never fixed in `ubuntu/focal` so it came back again in 23.3.

If we had included the apport code in the main branch rather than in packaging branches, our unit tests would have caught the invalid typing.

## Merge type

### DIFF
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -14,6 +14,7 @@ from cloudinit.cmd.devel.logs import (
     INSTALLER_APPORT_FILES,
     INSTALLER_APPORT_SENSITIVE_FILES,
 )
+from cloudinit.cmd.status import is_cloud_init_enabled
 
 try:
     from apport.hookutils import (
@@ -297,6 +298,8 @@ def general_add_info(report, _) -> None:
     /run/cloud/instance-data.json that will be helpful for debugging.
     """
     try:
+        if not is_cloud_init_enabled():
+            return
         with open("/run/cloud-init/instance-data.json", "r") as fopen:
             instance_data = json.load(fopen)
     except FileNotFoundError:
--- a/cloudinit/apport.py
+++ b/cloudinit/apport.py
@@ -7,6 +7,7 @@
 import json
 import logging
 import os
+from typing import Dict
 
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.cmd.devel.logs import (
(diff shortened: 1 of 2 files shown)


cloudinit/apport.py
```diff
@@ -226,3 +240,102 @@ def add_info(report, ui):
...
+    """Entry point for Apport.
+
+    This hook runs for every apport report
+
+    Add a subset of non-sensitive cloud-init data from
+    /run/cloud/instance-data.json that will be helpful for debugging.
+    """
+    try:
+        if not is_cloud_init_enabled():
+            return
```
blackboxsw, line 309: Confirmed that this is the only diff of substance between the downstream packaged debian/apport-general-hooks.py and our upstream version, I approve the "new/migrated" function additions.

cloudinit/apport.py
```diff
@@ -71,6 +72,13 @@
 CLOUDINIT_OUTPUT_LOG = "/var/log/cloud-init-output.log"
 
 
+def is_cloud_init_enabled() -> bool:
+    return (
+        get_status_details(read_cfg_paths()).boot_status_code
+        not in DISABLED_BOOT_CODES
+    )
```
blackboxsw, line 79: Given that this is more status-related functionality than apport-related It is worth just defining this helper function in cloudinit.status?

blackboxsw (approved):
Minor request to mograte is_cloud_init_enabled into cloudinit.status module so we aren't importing multiple "implementation details" over into apport. But +1 either way if you disagree with that suggestion.

cloudinit/apport.py
holmanb: That works for me!

holmanb: Aside: It does feel funny to have code under `cloudinit/cmd/` that ends up getting imported by external tools.

I would almost lean towards a `cloudinit/lib/` to have documented acceptable interfaces for external users with a very small subset of functions that we guarantee to have a stable interface. But that's very much a "someday" idea - for now I think that your suggested approach is best. Thanks for the review @blackboxsw!

PR #4874, opened 2024-02-13
apport: Don't collect logs if cloud-init is disabled

holmanb:
## Proposed Commit Message
See individual commits

## Additional Context
Requires https://github.com/canonical/cloud-init/pull/4873 and updates to the other branches {focal,jammy,mantic}

## Merge type

- [ ] Squash merge using "Proposed Commit Message"
- [x] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #4878, opened 2024-02-13
Ubuntu/jammy

holmanb:
See https://github.com/canonical/cloud-init/pull/4874

### DIFF
--- a/debian/apport-general-hook.py
+++ b/debian/apport-general-hook.py
@@ -1,61 +1,6 @@
 """General Apport hook for all reports that are using cloud-init."""
 
-import json
-import logging
-
-
-def _get_azure_data(ds_data) -> dict[str, str]:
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
LGTM

blackboxsw:
I'm inclined  to merge this branch for 3 reasons:
  - the "Integration test" workflow shouldn't have run on this branch anyway as it should be skipped by the following workflow config because we do want to disregard the package build costs for interrim branch releases
```
on:
  pull_request:
    branches-ignore:
      - 'ubuntu/**'
```
- We can see in this separate PR #4883 against ubuntu/jammy that this "integration test" action was skipped properly by CI
- There is an additional fix in upstream integration tests that should specifically moderates how features are being checked when the package is being built by `packages/bddeb` as we do in the CI runner for branches targeting merge into main.

PR #4881, opened 2024-02-14
Ubuntu/devel

holmanb:
## Proposed Commit Message
use individual commits

## Additional Context

Since https://github.com/canonical/cloud-init/commit/551f560d1a2decdb58f3a8352c2e5cbe6f56c5d7 landed, snapd will no longer be ordered before ssh. The performance regression [introduced by making ssh order after snapd](https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2039505) should no longer be an issue (we should verify this).

That means two things:

1) we should reapply [the fix](https://bugs.launchpad.net/ubuntu/+source/livecd-rootfs/+bug/2013403) for some systems that currently get a login prompt before users are created
2) 551f560d1a2de needs to be patched out of non-devel releases

This PR handles the first

## Merge type


TheRealFalcon (approved):
LGTM!

PR #4895, opened 2024-02-15
postinst: pre-flight check hotplug file exists before forcibly moving it

blackboxsw:
An oversight in postinst to pre-flight check of hotplug file exists before trying to rename it to avoid errors in postinst when hotplug file doesn't exist


TheRealFalcon (approved):
Changelog already present and unreleased, so LGTM!

PR #4897, opened 2024-02-16
Tiobe cs

TheRealFalcon:
## Proposed Commit Message

```
Rebase...see individual commits
```

## Additional Context
I'm ignoring tests because we often set instance attributes in fixtures, and I don't think we should be avoiding that.

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -112,6 +112,7 @@ class SocketHTTPConnection(HTTPConnection):
     def __init__(self, socket_path):
         super().__init__("localhost")
         self.socket_path = socket_path
+        self.sock = None
 
     def connect(self):
         self.sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -166,7 +166,7 @@ class DataSourceLXD(sources.DataSource):
     dsname = "LXD"
 
     _network_config: Union[Dict, str] = sources.UNSET
-    _crawled_metadata: Union[Dict, str] = sources.UNSET
+    _crawled_metadata: Optional[Union[Dict, str]] = sources.UNSET
 
     sensitive_metadata_keys: Tuple[
         str, ...
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -17,7 +17,7 @@ import pickle
 import re
 from collections import namedtuple
 from enum import Enum, unique
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from cloudinit import atomic_helper, dmi, importer, net, type_utils
 from cloudinit import user_data as ud
@@ -195,6 +195,8 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
     #  - seed-dir (<dirname>)
     _subplatform = None
 
+    _crawled_metadata: Optional[Union[Dict, str]] = None
+
     # The network configuration sources that should be considered for this data
     # source.  (The first source in this list that provides network
     # configuration will be used without considering any that follow.)  This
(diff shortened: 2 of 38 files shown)


cloudinit/sources/__init__.py
```diff
@@ -349,6 +354,18 @@ def _unpickle(self, ci_pkl_version: int) -> None:
                 raise DatasourceUnpickleUserDataError() from e
         if not hasattr(self, "extra_hotplug_udev_rules"):
             self.extra_hotplug_udev_rules = None
+        if not hasattr(self, "metadata_address"):
+            self.metadata_address = None
```
blackboxsw, line 358: In this function we have so many of the boilerplate setup of instance attributes to the same default values. Would it make sense to loop over expected_attrs and set the expect defaults to avoid the if not hasattr duplication?
```diff
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -331,12 +331,26 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
     def _unpickle(self, ci_pkl_version: int) -> None:
         """Perform deserialization fixes for Paths."""
-        if not hasattr(self, "vendordata2"):
-            self.vendordata2 = None
-        if not hasattr(self, "vendordata2_raw"):
-            self.vendordata2_raw = None
-        if not hasattr(self, "skip_hotplug_detect"):
-            self.skip_hotplug_detect = False
+        expected_attrs = [
+            "_crawled_metadata",
+            "_platform_type",
+            "_subplatform",
+            "ec2_metadata",
+            "extra_hotplug_udev_rules",
+            "metadata_address",
+            "network_json",
+            "skip_hotplug_detect",
+            "vendordata2",
+            "vendordata2_raw",
+        ]
+        value_defaults = {
+            "skip_hotplug_detect": False,
+            "ec2_metadata": UNSET,
+            "network_json": UNSET,
+        }
+        for attr in expected_attrs:
+            if not hasattr(self, attr):
+                setattr(self, attr, value_defaults.get(attr))
         if hasattr(self, "userdata") and self.userdata is not None:
             # If userdata stores MIME data, on < python3.6 it will be
             # missing the 'policy' attribute that exists on >=python3.6.
@@ -352,20 +366,6 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
                     e,
                 )
                 raise DatasourceUnpickleUserDataError() from e
-        if not hasattr(self, "extra_hotplug_udev_rules"):
-            self.extra_hotplug_udev_rules = None
-        if not hasattr(self, "metadata_address"):
-            self.metadata_address = None
-        if not hasattr(self, "_platform_type"):
-            self._platform_type = None
-        if not hasattr(self, "_subplatform"):
-            self._subplatform = None
-        if not hasattr(self, "_crawled_metadata"):
-            self._crawled_metadata = None
-        if not hasattr(self, "network_json"):
-            self.network_json = UNSET
-        if not hasattr(self, "ec2_metadata"):
-            self.ec2_metadata = UNSET
 
     def __str__(self):
         return type_utils.obj_name(self)
```

cloudinit/sources/__init__.py
```diff
@@ -349,6 +354,18 @@ def _unpickle(self, ci_pkl_version: int) -> None:
...
+        if not hasattr(self, "_platform_type"):
```
blackboxsw, line 359: I think we can drop the hasattr check on _platform_type in DataSourceNoCloud.py and Ec2 too

cloudinit/sources/DataSourceLXD.py
```diff
@@ -166,7 +167,7 @@ class DataSourceLXD(sources.DataSource):
     dsname = "LXD"
 
     _network_config: Union[Dict, str] = sources.UNSET
-    _crawled_metadata: Union[Dict, str] = sources.UNSET
+    _crawled_metadata: Optional[Union[Dict, str]] = sources.UNSET
```
blackboxsw, line 170: Hrm, LXD seems to be an outlier here with _crawled_metadata being set to sources.UNSET instead of None which the parent class attribute is initialized to. Should this actually by None instead of _UNSET to avoid potential issues with treating _UNSET as valid value != None? I see persist_instance_data treats != None as a viable value, even though we don't call the method until after the end of a successful get_data call.

blackboxsw: A change to LXD._crawled_metadata default value would involve a change to DataSourceLXD.network_config as well.

TheRealFalcon: Agreed. I wasn't sure if there was a good reason for having `UNSET` rather than `None`, but I'd rather not make that change here as it's unrelated to the changeset of this PR. If you know of no good reason, I can do it in a follow-up PR.

TheRealFalcon:
@blackboxsw I believe I addressed all of your comments.

TheRealFalcon:
a rebase also required a change to DataSourceWSL, so I added that as a fixup commit

blackboxsw (approved):

PR #4905, opened 2024-02-18
Bring back ds-identify fix

holmanb:
## Additional Context
Originally proposed by @smoser, then [reverted](https://github.com/canonical/cloud-init/pull/4327) due to [a reported bug](https://github.com/canonical/cloud-init/issues/4794) that it introduced. This brings back the original proposal by Scott plus the fixes required to resolve the issues it introduced. Also makes some minor shellcheck and test changes.

## Merge type

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -815,21 +815,24 @@ check_config() {
     if [ "$1" = "$files" -a ! -f "$1" ]; then
         return 1
     fi
-    local fname="" line="" ret="" found=0 found_fn=""
-    # shellcheck disable=2094
-    for fname in "$@"; do
-        [ -f "$fname" ] || continue
-        while read line; do
-            line=${line%%#*}
-            case "$line" in
-                $key:\ *|"${key}":)
-                    ret=${line#*:};
-                    ret=${ret# };
-                    found=$((found+1))
-                    found_fn="$fname";;
-            esac
-        done <"$fname"
+    local line="" ret="" found=0 found_fn="" oifs="$IFS" out=""
+    out=$(grep "$key\"\?:" "$@" 2>/dev/null)
+    IFS=${CR}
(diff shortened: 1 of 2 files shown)


tools/ds-identify
```diff
@@ -630,14 +629,14 @@ read_datasource_list() {
             dslist=${_RET}
             ;;
     esac
-    if [ -z "$dslist" ] && check_config datasource_list; then
+    if [ -z "$dslist" ] && check_config "$key" && get_value "$key" "$_RET"; then
```
TheRealFalcon, line 632: If I understand this correctly, it checks that _something_ exists after the `datasource_list:`, correct? Even if it's just a space? Wouldn't `datasource_list: ` pass the check here? Can we instead check that there's at least an opening bracket somewhere on the line or some kind of non-whitespace?

TheRealFalcon (changes requested):
One request inline. Also, can we get a test that some multiline yaml will be invalid? I think the one from the reported issue would work:
```yaml
datasource_list:
 - Azure
```

tools/ds-identify
holmanb: > If I understand this correctly, it checks that something exists after the datasource_list:, correct? Even if it's just a space? Wouldn't datasource_list: pass the check here?

Correct

> Can we instead check that there's at least an opening bracket somewhere on the line

This would break the `disable_vmware_customization` callsite, which expects a boolean not a list.

> or some kind of non-whitespace?

Yeah we can make this more strict.

TheRealFalcon (approved):
LGTM! Really nice cleanup!

holmanb:
> LGTM! Really nice cleanup!

Thanks!

I just force pushed to merge a couple of related commits and update some commit wording. Will merge once tests pass.

Issue #4907, opened 2024-02-19
Azure: ubuntu minimal images getting warnings due to missing gdisk (/usr/sbin/sgdisk)


blackboxsw:
# Bug report
Ubuntu Minimal Azure images do not contain gdisk package (and resulting sgdisk), resulting in failure to partition /dev/sda.
This results in a warning in cloud-init logs `Failed to partitioning operation` and the inability to partition /dev/sda.

## Environment details
- Cloud-init version: 24.1~
- Operating System Distribution: Ubuntu Noble minimal
- Cloud provider, platform or installer type: Azure

## cloud-init logs
```
2024-02-19 03:16:07,680 - cc_disk_setup.py[DEBUG]: Checking if device layout matches
2024-02-19 03:16:07,680 - subp.py[DEBUG]: Running command [None, '-p', '/dev/sda'] with allowed return codes [0] (shell=False, capture=True)
2024-02-19 03:16:07,680 - util.py[DEBUG]: Creating partition on /dev/disk/cloud/azure_resource took 0.022 seconds
2024-02-19 03:16:07,680 - util.py[WARNING]: Failed partitioning operation
Error running partition command on /dev/sda
'NoneType' object has no attribute 'encode'
2024-02-19 03:16:07,684 - util.py[DEBUG]: Failed partitioning operation
Error running partition command on /dev/sda
'NoneType' object has no attribute 'encode'
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_disk_setup.py", line 491, in check_partition_gpt_layout
    out, _err = subp.subp(prt_cmd, update_env=LANG_C_ENV)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 244, in subp
    bytes_args = [
                 ^
  File "/usr/lib/python3/dist-packages/cloudinit/subp.py", line 245, in <listcomp>
    x if isinstance(x, bytes) else x.encode("utf-8") for x in args
                                   ^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'encode'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/cloudinit/config/cc_disk_setup.py", line 136, in handle
    util.log_time(
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 2838, in log_time
[...]

PR #4909, opened 2024-02-19
tests: use apt install instead of dpkg -i to install pkg deps

blackboxsw:
## Proposed Commit Message
```
tests: use apt instead of dpkg -i to install pkg deps

On Ubuntu Noble dhcpcd-base is a hard package Depends and
also not included by default in ubuntu-minimal images until
cloud-init the new cloud-init images contain latest cloud-init.

When provided with a DEB_PATH source, use apt instead of dpkg -i
to any missing install package dependencies.

Fixes: GH-4908
```

## Additional Context
Failed jenkins job
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-noble-azure/51/console

There is a separate bug with Ubuntu Minimal  images (missing sgdisk https://github.com/canonical/cloud-init/issues/4907) which leaves a WARNING in logs that will not be addressed by this particular PR.

## Checklist

## Merge type

### DIFF
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -216,7 +216,13 @@ class IntegrationInstance:
             local_path=integration_settings.CLOUD_INIT_SOURCE,
             remote_path=remote_path,
         )
-        assert self.execute("dpkg -i {path}".format(path=remote_path)).ok
+        # Update APT cache so all package data is recent to avoid inability
+        # to install missing dependency errors due to stale cache.
+        self.execute("apt update")
+        # Use apt install instead of dpkg -i to pull in any changed pkg deps
+        assert self.execute(
+            "apt install {path} --yes".format(path=remote_path)
+        ).ok
 
     @retry(tries=30, delay=1)
     def upgrade_cloud_init(self):


tests/integration_tests/datasources/test_azure.py
```diff
@@ -34,6 +37,15 @@ def test_azure_eject(session_cloud: IntegrationCloud):
     ) as instance:
         source = get_validated_source(session_cloud)
         if source.installs_new_version():
+            if (
```
TheRealFalcon, line 40: Why only in this test? Wouldn't we need to do an update on any test being launched on noble that installs something new? I would think this should go into `install_new_cloud_init`or `install_deb` directly.

tests/integration_tests/datasources/test_azure.py
```diff
@@ -34,6 +37,15 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
+                source == CloudInitSource.DEB_PACKAGE
+                and CURRENT_RELEASE >= NOBLE
+            ):
+                # Minimal Noble images do not contain dhcpcd-base and
+                # Noble and newer have an explicit Depends.
+                # Update APT cache so dependencies are
+                #
```
TheRealFalcon, line 47: Sentence didn't get finished here.

TheRealFalcon (changes requested):

tests/integration_tests/datasources/test_azure.py
blackboxsw: That's a better place, I had thought about it but considered that this should only affect MINIMAL images in general or days in which we've changed strict packaging dependencies in an additive fashion and haven't released those pkg changes to cloud-init yet. I had also thought to avoid the cost of the apt update... but it's only a cost paid at image setup time so that's ok. I'll move it.

holmanb (approved):
Hmmm, this would resolve itself automatically once using an image that contains the up-to-date version of cloud-init, but better to silence this failure now and forever so that future packaging dependency changes don't require debugging. Thanks for this. 

Alternatively including isc-dhcp-client as an alternate to the dhcpcd-base requirement in debian/control would work (and would technically be correct), but I didn't bother specifying that because of the plans to drop isc-dhcp-client from the archive - it didn't seem necessary to keep it as an alternate dependency. 

+1 to this approach, apt will be more robust against future packaging dependency changes in cloud-init

TheRealFalcon (approved):
LGTM!

PR #4914, opened 2024-02-19
test: fix tmpdir in test_cc_apk_configure

TheRealFalcon:
## Proposed Commit Message

```
test: fix tmpdir in test_cc_apk_configure

cc_apk_configure uses temp_utils.py, which has special logic to return
if the user is root that can't be retargeted using our current fixtures.
Fix it by setting that tmpdir in the test setup.
```

## Additional Context
Running tests in `tests/unittests/config/test_cc_apk_configure.py` as root will demonstrate the failure

## Merge type

### DIFF
--- a/tests/unittests/config/test_cc_apk_configure.py
+++ b/tests/unittests/config/test_cc_apk_configure.py
@@ -10,7 +10,7 @@ import textwrap
 
 import pytest
 
-from cloudinit import cloud, helpers, util
+from cloudinit import cloud, helpers, temp_utils, util
 from cloudinit.config import cc_apk_configure
 from cloudinit.config.schema import (
     SchemaValidationError,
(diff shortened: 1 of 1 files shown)


TheRealFalcon:
cc @ani-sinha . For 23.4, it probably makes more sense to pin pytest to the version we were using when we released cloud-init 23.4 as you are doing, but this should fix the root cause of the problem you asked about in IRC.

tests/unittests/config/test_cc_apk_configure.py
```diff
@@ -60,6 +60,7 @@ def setUp(self):
         self.name = "apk_configure"
         self.cloud = cloud.Cloud(None, self.paths, None, None, None)
         self.args = []
+        temp_utils._TMPDIR = self.new_root
```
holmanb, line 63: This sets a global for that module. Shouldn't we clean this up after tests run?

holmanb (commented):
Nice fix, but don't we want to do cleanup of this as well?

tests/unittests/config/test_cc_apk_configure.py
TheRealFalcon: Should be fixed now

holmanb (approved):
Looks good, thanks!

Issue #4765, opened 2024-01-10
[bug] udhcp support is incomplete


holmanb:
## Steps to reproduce the problem

PR #4190 introduced udhcpc client support. As it is, udhcpc may work on some platforms, but is missing support for unknown dhcp options. This is used by microsoft azure to update the wireserver IP address via [option 245](https://github.com/canonical/cloud-init/blob/b729a4c4aa32d2bdfdd84fb24b9be5fc9c1eb293/cloudinit/sources/DataSourceAzure.py#L504). I believe that this is still a requirement for microsoft azure, so this should be implemented for udhcpc as well for correct behavior on microsoft Azure.

holmanb:
> I believe that this is still a requirement

@cjp256 can you please confirm?

holmanb:
@jfroche In #4190 you said you would follow up to complete udhcpc support. Are you still planning to do that?

PR #4924, opened 2024-02-20
test: xfail udhcpc on azure

holmanb:
## Proposed Commit Message
```
test: xfail udhcpc on azure
```

## Additional info
see https://github.com/canonical/cloud-init/issues/4765

## Merge type


TheRealFalcon (approved):
I'm slightly more wary of xfail's in integration tests given the run time, but as long as we don't accumulate a ton of them I think it's fine.

holmanb:
> I'm slightly more wary of xfail's in integration tests given the run time, but as long as we don't accumulate a ton of them I think it's fine.

Agreed. I'd actually love to get #4765 fixed in the next cycle if possible, which is why I left it for now. We can always rip it out if we decide that's not a priority.

PR #4925, opened 2024-02-20
test: fix test_combined_cloud_config_json

TheRealFalcon:
## Proposed Commit Message

```
test: fix test_combined_cloud_config_json

The old version cannot pass all possible scenarios with code residing
on the test runner. The only way to ensure that features advertised
in cloud-config.json match is to check the features file on the SUT.
```

## Additional Context
See https://github.com/canonical/cloud-init/pull/4868 .
There is no way for us to know beforehand which features will be enabled for a given test. For example, for any current given series, what's in PROPOSED (i.e., 23.4) is going to vary from what's in the daily PPA.

## Merge type

### DIFF
--- a/tests/integration_tests/modules/test_combined.py
+++ b/tests/integration_tests/modules/test_combined.py
@@ -15,12 +15,11 @@ from pathlib import Path
 import pytest
 
 import cloudinit.config
-from cloudinit.features import get_features
 from cloudinit.util import is_true
 from tests.integration_tests.decorators import retry
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.integration_settings import PLATFORM
-from tests.integration_tests.releases import CURRENT_RELEASE, IS_UBUNTU, NOBLE
+from tests.integration_tests.releases import CURRENT_RELEASE, IS_UBUNTU
 from tests.integration_tests.util import (
     get_feature_flag_value,
     get_inactive_modules,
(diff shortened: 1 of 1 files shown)


tests/integration_tests/modules/test_combined.py
```diff
@@ -323,15 +322,12 @@ def test_combined_cloud_config_json(
             "/run/cloud-init/combined-cloud-config.json"
         )
         data = json.loads(combined_json)
-        expected_features = get_features()
-        is_devel_pkg = (
-            "bddeb" in client.execute("cloud-init -v").split()[-1]
-            or client.settings.CLOUD_INIT_SOURCE == "IN_PLACE"
+        expected_features = json.loads(
```
blackboxsw, line 325: What we lose with this is visibility to whether our daily images actually did correctly disable NETPLAN_CONFIG_ROOT_ROOT_ONLY. We are now just comparing that our python features.get_features() module equals /run/cloud-init/combined-cloud-config.json regardless of whether that content has the correct published expected settings and/or is well-formed. I'm ok with not checking the actual values. But, we risk accidentally forgetting to disable some feature in the future when tip of main enables something that we want to disable on stable downstreams.

blackboxsw (commented):
Per your commit description, I agree it's challenging to validate all use-cases. That's probably why our IN_PLACE vs deb_path vs PPA daily integration test runs are falling over frequently on this test. The test now becomes just validation that the SUT `features.get_featres()` == `/run/cloud-init/combined-instance-data.json["features"]`Is there any other we can can think to easily assert that stable releases have effectively turned off a feature we expect to be False?

tests/integration_tests/modules/test_combined.py
blackboxsw: well we do assert well-formed because you use json.loads. But, we aren't asserting anywhere the expected packaged values of our features in stable releases AFAICT.

TheRealFalcon: > The test now becomes just validation that the SUT features.get_featres() == /run/cloud-init/combined-instance-data.json["features"]

Isn't that the point of the test? We want to ensure that `cloud-init query` matches what's actually set for the features. At least that seemed to be the point when you added it. The only way to really test that the values are what we think they should be is to keep a static map of cloudinit version ranges for each series and each expected value. The problem with that is we have the same problem as before...we have to remember to update the map like we have to remember to create the quilt patches. So I don't see a ton of value to that kind of test either.

blackboxsw: Yes. that was the original point of the initial test when I added it- assert the whatever features were configured were exposed in /run/cloud-init/combined-instance-data.json. 

I just thought when Alberto added supplemental checks it would be nice to have that level of validation that we did package what we thought we packaged. But, that's really something that we should be checking already during both our downstream SRU reviews and validation and our quilt patch reviews. So, maybe this is not really something we need in ongoing integration tests due to the fact that we are duplicating tracking of what is already tracked in debian/patches anyway.

+1 on leaving this test flexible and limiting it to ensure the original intent that features are surfaced in the /run file. If we come up with an easy to maintain strict validation of known feature settings on each stable release we can deal with that in a separate test. I agree that it's the same problem. if we are forgetting to disable a feature via a quilt patch, we are also going to forget to update the test map.

PR #4926, opened 2024-02-20
d/cloud-init.templates: enable WSL datasource in packaging.

blackboxsw:
## Proposed Commit Message
```
d/cloud-init.templates: enable WSL datasource in packaging.

Also update po translation files
```

## Additional Context
Now that WSL datasource has landed in tip of main, enable it in packaging by adding it to cloud-init.templates

## Test Steps
```
dhc -r -D noble ''
git commit -am 'test full package build'
build-package
sbuild --dist=noble --arch=amd64 --arch-all ../out/cloud-init_*.dsc
lxc launch ubuntu-daily:noble test-n
lxc file push cloud-init*_all.deb test-n/
lxc exec test-n -- apt -i /cloud-init*deb
lxc exec test-n -- dpkg-reconfigure cloud-init # confirm you can select  WSL from datasource search options
```

## Checklist

## Merge type


TheRealFalcon (approved):
Not for this PR but...

Is there a reason we keep the translation stuff around even though we have no translations? It's not required for a debian package, right? If not, it's just overhead with no current benefit. Can we remove it and then add if/when we ever need it?

blackboxsw:
> Not for this PR but...
> 
> Is there a reason we keep the translation stuff around even though we have no translations? It's not required for a debian package, right? If not, it's just overhead with no current benefit. Can we remove it and then add if/when we ever need it?

We should double check that there isn't a fall-back/default option for translatable strings/message ids that is used when language packs are installed and a package (e.g.cloud-init) doesn't define specific localization/translations of such messages if they happen to match.

Since we really only have rudimentary 'internationalization groundwork' coverage of the dpkg-reconfigure prompts and since cloud-init doesn't use getttext throughout our codebase for messaging, the impact of dropping the initial po files should not be much even if there was a fallback global translation mechanism that catches any cloud-init strings. Let's circle back on that issue with other package maintainers to determine best course of action here.

Issue #4927, opened 2024-02-20
networkd renderer does not respect accept-ra


philsphicas:
# Bug report

The networkd renderer only includes `IPv6AcceptRA` in the generated files if DHCP is enabled. It should be possible to toggle this setting whenever `accept-ra` is set, for example, to reject RAs when static IPv4 addressing is used.

## Steps to reproduce the problem

This network-config:

```yaml
network:
    config:
    -   accept-ra: false
        mac_address: 00:11:22:33:44:55
        name: eth0
        subnets:
        -   address: 10.0.0.10/24
            type: static
        type: physical
    version: 1
```

Should render as:

```ini
[Address]
Address=10.0.0.10/24

[Match]
MACAddress=00:11:22:33:44:55
Name=eth0

[Network]
DHCP=no
IPv6AcceptRA=False
```

TheRealFalcon:
Thanks for the bug report @philsphicas . Looks like [networkd.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/networkd.py#L195) needs an update.

philsphicas:
> Thanks for the bug report @philsphicas . Looks like [networkd.py](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/networkd.py#L195) needs an update.

Hi @TheRealFalcon I noticed that the linked PR got marked as stale ... is there anything that needs updating?

TheRealFalcon:
@philsphicas , no sorry about the wait on that. I'll try to get it reviewed this week.

PR #4928, opened 2024-02-20
networkd: Always respect accept-ra if set

philsphicas:
Support the explicit rejection of router advertisements in cases where
the network may be configured to send them, but they are not desireable,
for example, in a case where static IPv4 addressing is used.

Before this change, IPv6AcceptRA is only included in the rendered
networkd files if DHCP is used.

Now, if a boolean value for accept-ra is present, it will be included in
the rendered output, regardless of the configured addressing.

Fixes: #4927

### DIFF
--- a/cloudinit/net/networkd.py
+++ b/cloudinit/net/networkd.py
@@ -192,9 +192,7 @@ class Renderer(renderer.Renderer):
 
         cfg.update_section(sec, "DHCP", dhcp)
 
-        if dhcp in ["ipv6", "yes"] and isinstance(
-            iface.get("accept-ra", ""), bool
-        ):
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
@philsphicas , thanks for this PR. The change looks good to me.

I noticed you added yourself to the CLA signers file, but have you signed the [actual CLA](https://ubuntu.com/legal/contributors/agreement) too? I'm not seeing your information there. If you did, can you provide the date that you signed it?

philsphicas:
Sorry, I thought I had done this a long time ago, but apparently not. Signed and submitted just now.

TheRealFalcon (approved):
Thanks for the contribution!

PR #4931, opened 2024-02-21
feat(azurelinux): Add new distro 'azurelinux' for Microsoft Azure Linux.

ddstreetmicrosoft:
## Proposed Commit Message
<!-- Include a proposed commit message because PRs are squash merged
by default.

See https://www.conventionalcommits.org/en/v1.0.0/#specification
for our commit message convention.

If the change is related to a particular cloud or particular distro,
please include the "optional scope" in the summary line. E.g.,
feat(ec2): Add support for foo to the baz

Types used by this project:
feat, fix, docs, ci, test, refactor, chore
-->
```
This adds support for the new (renamed) distro 'Azure Linux'.                                                                                                                                 
                                                                                                                                                                                              
Azure Linux is the renamed next version of the existing 'Mariner'                                                                                                                             
distro. It is similar to the 'Mariner' distro but changes the base                                                                                                                            
inherited Distro() class.                                                                                                                                                                     

```

## Additional Context
<!-- If relevant -->

## Checklist
that apply. -->

## Merge type

### DIFF
--- a/cloudinit/config/cc_ntp.py
+++ b/cloudinit/config/cc_ntp.py
@@ -25,6 +25,7 @@ NR_POOL_SERVERS = 4
 distros = [
     "almalinux",
     "alpine",
+    "azurelinux",
     "centos",
     "cloudlinux",
     "cos",
@@ -107,6 +108,15 @@ DISTRO_CLIENT_CONFIG = {
             "confpath": "/etc/ntp.conf",
             "packages": [],
             "service_name": "ntpd",
+        },
+    },
+    "azurelinux": {
+        "chrony": {
+            "service_name": "chronyd",
+        },
+        "systemd-timesyncd": {
+            "check_exe": "/usr/lib/systemd/systemd-timesyncd",
+            "confpath": "/etc/systemd/timesyncd.conf",
         },
     },
     "centos": {
--- a/cloudinit/distros/azurelinux.py
+++ b/cloudinit/distros/azurelinux.py
@@ -1,0 +1,72 @@
+# Copyright (C) 2024 Microsoft Corporation
+#
+# Author: Dan Streetman <ddstreet@microsoft.com>
+#
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import logging
+
+from cloudinit import subp, util
+from cloudinit.distros import rhel
+from cloudinit.net.netplan import CLOUDINIT_NETPLAN_FILE
+
+LOG = logging.getLogger(__name__)
+
+NETWORK_FILE_HEADER = """\
+# This file is generated from information provided by the datasource.  Changes
+# to it will not persist across an instance reboot. To disable cloud-init's
+# network configuration capabilities, write a file
+# /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with the following:
+# network: {config: disabled}
+"""
+
+
+class Distro(rhel.Distro):
+    def __init__(self, name, cfg, paths):
+        super().__init__(name, cfg, paths)
+        self.osfamily = "azurelinux"
+
+        self.network_conf_dir = "/etc/systemd/network/"
+        self.systemd_locale_conf_fn = "/etc/locale.conf"
+        self.resolve_conf_fn = "/etc/systemd/resolved.conf"
+        self.init_cmd = ["systemctl"]
+
+        self.network_conf_fn = {"netplan": CLOUDINIT_NETPLAN_FILE}
+        self.renderer_configs = {
+            "networkd": {
+                "resolv_conf_fn": self.resolve_conf_fn,
+                "network_conf_dir": self.network_conf_dir,
+            },
+            "netplan": {
+                "netplan_path": self.network_conf_fn["netplan"],
+                "netplan_header": NETWORK_FILE_HEADER,
+                "postcmds": "True",
+            },
+        }
+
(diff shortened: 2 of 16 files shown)


ddstreetmicrosoft:
CC @rmhsawyer @christopherco

cloudinit/config/cc_ntp.py
```diff
@@ -25,6 +25,7 @@
 distros = [
     "almalinux",
     "alpine",
+    "azure",
```
christopherco, line 28: azurelinux

cloudinit/distros/azurelinux.py
```diff
@@ -0,0 +1,81 @@
...
+            cmd = ["tdnf"]
+        else:
+            LOG.debug("Using YUM for package management")
+            # the '-t' argument makes yum tolerant of errors on the command
+            # line with regard to packages.
+            #
+            # For example: if you request to install foo, bar and baz and baz
+            # is installed; yum won't error out complaining that baz is already
+            # installed.
+            cmd = ["yum", "-t"]
```
christopherco, line 65: Consider deprecating yum. Default to `tdnf`

cloudinit/config/cc_ntp.py
ddstreetmicrosoft: thanks! fixed

cloudinit/distros/azurelinux.py
ddstreetmicrosoft: `yum` is only being used as a final fallback here, after attempting `dnf` and `tdnf` first...we could switch the final fallback over to `tdnf` instead (and drop falling back to `yum` completely) if that's what you mean?

christopherco: Yes, that is what I meant. I was thinking we should switch the final fallback to `tdnf` and drop `yum`, since `yum` for us is really a symlink to `tdnf`, and we ought to remove this in our upcoming major release - https://github.com/microsoft/CBL-Mariner/blob/3.0-dev/SPECS/tdnf/tdnf.spec#L147

christopherco (approved):
lgtm

ddstreet:
rebased

ddstreet:
updated `config/cloud.cfg.tmpl` in some locations I missed before

TheRealFalcon:
@ddstreetmicrosoft Thanks for this PR. Does this mean that the Mariner datasource is no longer relevant? Should we remove it at some point?

If you check the CI runs, you'll see there's a few lint failures that need to be fixed before we can merge this.

Additionally, before we can accept your change, you need to sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html#submitting-your-first-pull-request) . It looks like you may have already signed the CLA under a different github name, so if you're using two (`ddstreet` and `ddstreetmicrosoft`), please add them both the signers file.

cjp256:
LGTM.  Looks like `black` needs to be run against some of the files.

ddstreet:
> @ddstreetmicrosoft Thanks for this PR. Does this mean that the Mariner datasource is no longer relevant? Should we remove it at some point?

> LGTM. Looks like black needs to be run against some of the files. Is there need to keep Mariner after this change?

While the `Mariner` naming is currently still a supported release, that name will retire at the end of its support lifetime; I could update this PR to completely switch over (i.e. remove 'Mariner' while adding 'Azure Linux') if you think that's appropriate? Since the `Mariner` release is in a stable period (and shouldn't be updating its cloud-init) there shouldn't be any reason to need to keep it in the upstream code here.

Alternately, I could open a separate PR to remove 'Mariner', either soon or closer to when it goes EOL (which will be late this year at the earliest, more likely sometime next year).

> 
> If you check the CI runs, you'll see there's a few lint failures that need to be fixed before we can merge this.

ack, thanks, i ran `tox -e do_format` and added the change, and a full `tox` run succeeded, so should be ok now i hope.

> 
> Additionally, before we can accept your change, you need to sign [the CLA](https://ubuntu.com/legal/contributors/agreement) and add your Github username (alphabetically) to [the cla signers file](https://github.com/canonical/cloud-init/blob/master/tools/.github-cla-signers) . The full details are described in the last bullet point [of the documentation](https://cloudinit.readthedocs.io/en/latest/development/contributing.html#submitting-your-first-pull-request) . It looks like you may have already signed the CLA under a different github name, so if you're using two (`ddstreet` and `ddstreetmicrosoft`), please add them both the signers file.

i signed the CLA and included both my github usernames (`ddstreet` and `ddstreetmicrosoft`) in the signers file, in this pr.

TheRealFalcon:
> Alternately, I could open a separate PR to remove 'Mariner', either soon or closer to when it goes EOL (which will be late this year at the earliest, more likely sometime next year).

Yes, a separate PR sounds good. If there is truly no need for the upstream code anymore, then soon is fine, but there is also no harm in keeping it here for the time being.

TheRealFalcon:
The 3.13 test failure is unrelated and being addressed in https://github.com/canonical/cloud-init/pull/5052

TheRealFalcon (approved):
LGTM! Thanks @ddstreet

PR #4939, opened 2024-02-22
fix(cmd/main.py): use correct return value

sshedi:
## Proposed Commit Message
```
fix(cmd/main.py): use correct return value
```

## Additional Context
```
$ cloud-init -d modules -m init
Above command returns non zero exit status
```

## Test Steps
```
root@ph4dev [ ~/cloud-init ]# cloud-init -d modules -m init
2024-02-22 19:10:07,659 - handlers.py[DEBUG]: start: modules-init: running modules for init
2024-02-22 19:10:07,659 - util.py[DEBUG]: Reading from /proc/uptime (quiet=False)
...
Cloud-init v. 23.4 running 'modules:init' at Thu, 22 Feb 2024 19:10:07 +0000. Up 26599.38 seconds.
[]

root@ph4dev [ ~/cloud-init ]# echo $?
1
```

## Checklist

- [] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly

## Merge type

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -126,12 +126,12 @@ def run_module_section(mods: Modules, action_name, section):
         )
         sys.stderr.write("%s\n" % (msg))
         LOG.debug(msg)
-        return []
+        return 1
     else:
         LOG.debug(
             "Ran %s modules with %s failures", len(which_ran), len(failures)
         )
-        return failures
+        return failures if failures else 0
 
 
 def apply_reporting_cfg(cfg):
@@ -786,7 +786,10 @@ def status_wrapper(name, args, data_d=None, link_d=None):
         else:
             errors = ret
 
-        v1[mode]["errors"] = [str(e) for e in errors]
+        if isinstance(errors, list):
+            v1[mode]["errors"] = [str(e) for e in errors]
+        else:
+            v1[mode]["errors"] = []
 
     except Exception as e:
         util.logexc(LOG, "failed stage %s", mode)
--- a/tests/unittests/cmd/test_main.py
+++ b/tests/unittests/cmd/test_main.py
@@ -76,7 +76,7 @@ class TestMain(FilesystemMockingTestCase):
             "init",
             cmdargs,
         )
-        self.assertEqual([], item2)
+        self.assertEqual(0, item2)
         # Instancify is called
         instance_id_path = "var/lib/cloud/data/instance-id"
         self.assertEqual(
@@ -146,7 +146,7 @@ class TestMain(FilesystemMockingTestCase):
             "init",
             cmdargs,
         )
-        self.assertEqual([], item2)
+        self.assertEqual(0, item2)
         # Instancify is called
         instance_id_path = "var/lib/cloud/data/instance-id"
         self.assertEqual(


sshedi:
I'm not sure how to add a test for this. Any inputs on the same would be helpful. Thanks.

cloudinit/cmd/main.py
```diff
@@ -786,7 +786,9 @@ def status_wrapper(name, args, data_d=None, link_d=None):
         else:
             errors = ret
 
-        v1[mode]["errors"] = [str(e) for e in errors]
+        v1[mode]["errors"] = []
+        for e in errors:
+            v1[mode]["errors"].append(str(e))
```
a-dubs, line 791: What is the purpose of this change? The new code is functionally identical to the code it replaces.

a-dubs:
Also, could you please modify your proposed commit message section so that your commit message appears as a code block like the following:
```
fix(cmd/main.py): use correct return value
```

And could you move the 
```
$ cloud-init -d modules -m init
Above command returns non zero exit status
```
information in the PR description to the `Additional Context` section and add some more context for why this change is needed and what the change is seeking to accomplish?

Thanks!

cloudinit/cmd/main.py
sshedi: `errors` can be an integer value here, not always a list.

sshedi:
> @sshedi Do you use this for something? This doesn't normally run during boot, so I'm curious what you use it for.

In one of our CI/CD pipelines, I tried adding:
```
echo -e "\nRun cloud-init with all configured modules"
for m in init config final; do
  cmd="cloud-init --file user-data -d modules --mode $m"
  if $cmd; then
    echo -e "\nPASS: $cmd"
  else
    echo -e "\nFAIL: $cmd" 1>&2
  fi
done
```
And caught this issue. Nothing major, but I think the return value should be appropriate.

sshedi:
> Also, could you please modify your proposed commit message section so that your commit message appears as a code block like the following:
> 
> ```
> fix(cmd/main.py): use correct return value
> ```
> 
> And could you move the
> 
> ```
> $ cloud-init -d modules -m init
> Above command returns non zero exit status
> ```
> 
> information in the PR description to the `Additional Context` section and add some more context for why this change is needed and what the change is seeking to accomplish?
> 
> Thanks!

Thanks for the input and suggestions. Previously while raising PRs, this section was simple. I feel it's a bit complex now.
I understand the intent behind this but writing good commit message is already a hard task and filling this template is a bit confusing.

I will try.

cloudinit/cmd/main.py
sshedi: Indeed there was in issue in the code change, fixed it. Thanks.

aciba90:
Thanks, @sshedi, for raising this issue and proposing a fix. I am closing this one in favor of #5017, which includes a correct fix and integration tests. I hope that's okay from your side.

sshedi:
> Thanks, @sshedi, for raising this issue and proposing a fix. I am closing this one in favor of #5017, which includes a correct fix and integration tests. I hope that's okay from your side.

No problem. Thanks for fixing it.

PR #4942, opened 2024-02-22
test: limit temp dhcp6 changes to < NOBLE

TheRealFalcon:
## Proposed Commit Message

```
test: limit temp dhcp6 changes to < NOBLE

#4474 landed in Noble, so this block isn't needed.
```

## Additional Context
https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-noble-gce/49/testReport/junit/tests.integration_tests/test_upgrade/test_clean_boot_of_upgraded_package/

## Merge type

### DIFF
--- a/tests/integration_tests/test_upgrade.py
+++ b/tests/integration_tests/test_upgrade.py
@@ -8,7 +8,12 @@ import yaml
 from tests.integration_tests.clouds import IntegrationCloud
 from tests.integration_tests.conftest import get_validated_source
 from tests.integration_tests.integration_settings import PLATFORM
-from tests.integration_tests.releases import CURRENT_RELEASE, FOCAL, IS_UBUNTU
+from tests.integration_tests.releases import (
+    CURRENT_RELEASE,
+    FOCAL,
+    IS_UBUNTU,
+    NOBLE,
+)
 from tests.integration_tests.util import verify_clean_log
 
 LOG = logging.getLogger("integration_testing.test_upgrade")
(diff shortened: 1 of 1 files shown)


a-dubs:
Nice catch! Looks good. My eyes don't work too well tho...

blackboxsw:
@TheRealFalcon @holmanb don't we need to patch out this dhcp6 in network config from generate_fallback_config to avoid change in behavior on stable releases?

TheRealFalcon:
Is there reason to think this would break anything? I didn't think we needed to patch this one.

blackboxsw:
> Is there reason to think this would break anything? I didn't think we needed to patch this one.

As long as our rendered network config definition is not blocking/waiting indefinitely on marking the network interface "up" on having a viable dhcp6 route/response from the DHCP server, then it shouldn't impact anything such as boot time while we await for network-online.target.

TheRealFalcon:
>  not blocking/waiting indefinitely on marking the network interface "up"

Yeah, that's a valid concern. I initially reverted this one because it would cause exactly that on NetworkManager, but we were able to get that fixed in a follow-up to https://github.com/canonical/cloud-init/pull/4474

blackboxsw:
Thanks for this. I found the supplemental commit that fixed this issue for NetworkManager rendered config 29dd5ace73ad60c7452c39b840045fb47fe0711f.  And also confirmed that on systemd-networkd environments rendered from netplan config, we don't see issues with timing in environments with no dhcp6 provided from the dhcp service with /run/systemd/network/*.network   config setting `[Network]\nDHCP=yes` which is generated due to dhcp6 and dhcp4 vs just `DHCP=ipv4` which is rendered when we only specific `dhcp4: true` config in netplan.

I approve this work and it looks like our common net config rendered cases won't introduce blocking in environments where the DHCP service is not providing IPV6 config attributes.

blackboxsw (approved):

Issue #4945, opened 2024-02-23
nocloud-net not working in cloud-init >=23.4


rvandegrift:
# Bug report
I noticed that the nocloud-net datasource isn't working in 23.4.3.  If I set a seed via dmi, my http server never sees requests from cloud-init.  The same setup works fine with 23.3.1.  nocloud works fine in 23.4.3.

`git bisect` reports a commit first released in 23.4:
```
76b7b381497af9e4a6197606fd4d5fd5269b180b is the first bad commit
commit 76b7b381497af9e4a6197606fd4d5fd5269b180b
Author: Chris Patterson <cpatterson@microsoft.com>
Date:   Tue Sep 12 10:51:14 2023 -0400

    sources: do not override datasource detection if None is in list (#4426)

    Users with datasource_list = [Azure, None] started failing to boot
    properly outside of Azure with the changes to override datasource detection.

    If the fallback "None" is included in the datasource_list, do not treat
    the system as configured with a single datasource.

    If users want to force a single datasource regardless of detection,
    they can do so by removing None from the list.

    Signed-off-by: Chris Patterson <cpatterson@microsoft.com>

 cloudinit/sources/__init__.py | 5 +----
 1 file changed, 1 insertion(+), 4 deletions(-)
```

I wasn't able to test `main` with this reverted due to some test failures.  But the logs reveal that `override_ds_detect` is returning `False`.  In the working 23.3.1 case, the logs have this:
```
$ grep single cloud-init.log
2024-02-19 21:11:03,359 - __init__.py[DEBUG]: Machine is configured to run on single datasource DataSourceNoCloud [seed=None][dsmode=net].
2024-02-19 21:11:05,392 - __init__.py[DEBUG]: Machine is configured to run on single datasource DataSourceNoCloudNet [seed=None][dsmode=net].
```
But it's gone in 23.4.3.

## Steps to reproduce the problem
1. Start a webserver to serve a seed, nothing specific is required
2. Boot a qemu vm pointing to that server: `-smbios type=1,serial=ds='nocloud;s=http://10.0.2.2:8000/'` (`nocloud` vs `nocloud-net` makes no difference)
3. no requests are seen on the webserver, cloud-init uses `DataSourceNone`

## Environment details
- Cloud-init version: 23.4.3, current main (`89ebcc394`)
- Operating System Distribution: debian sid
- Cloud provider, platform or installer type: qemu

## cloud-init logs
logs from current main are attached.

[...]

PR #4949, opened 2024-02-23
fix serial datasource definition for nocloud

holmanb:
## Proposed Commit Message
```
fix(nocloud): NoCloud serial datasource definition is broken

Fix it.
Deprecate the "nocloud-net" name.
```

## Additional Context
https://github.com/canonical/cloud-init/issues/4945
https://github.com/canonical/cloud-init/issues/4948

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -11,6 +11,7 @@
 import errno
 import logging
 import os
+from functools import partial
 
 from cloudinit import dmi, sources, util
 from cloudinit.net import eni
@@ -368,12 +369,40 @@ class DataSourceNoCloudNet(DataSourceNoCloud):
         self.supported_seed_starts = ("http://", "https://")
 
     def ds_detect(self):
-        """NoCloud requires "nocloud-net" as the way to specify
-        seeding from an http(s) address. This diverges from all other
-        datasources in that it does a kernel commandline match on something
-        other than the datasource dsname for only DEP_NETWORK.
+        """Check dmi and kernel commandline for dsname
+
+        NoCloud historically used "nocloud-net" as its dsname
+        for network timeframe (DEP_NETWORK), which supports http(s) urls.
+        For backwards compatiblity, check for that dsname.
         """
-        return "nocloud-net" == sources.parse_cmdline()
+        log_deprecated = partial(
+            util.deprecate,
+            deprecated="The 'nocloud-net' datasource name",
+            deprecated_version="24.1",
+            extra_message=(
+                "Use 'nocloud' instead, which uses the seedfrom protocol"
+                "scheme (http// or file://) to decide how to run."
+            ),
+        )
+
+        if "nocloud-net" == sources.parse_cmdline():
+            log_deprecated()
+            return True
+
+        serial = sources.parse_cmdline_or_dmi(
+            dmi.read_dmi_data("system-serial-number") or ""
+        ).lower()
+
+        if serial in (self.dsname.lower(), "nocloud-net"):
+            LOG.debug(
+                "Machine is configured by dmi serial number to run on "
+                "single datasource %s.",
+                self,
+            )
+            if serial == "nocloud-net":
+                log_deprecated()
+            return True
+        return False
 
 
 # Used to match classes to dependencies
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -1195,10 +1195,13 @@ def parse_cmdline() -> str:
     """Check if command line argument for this datasource was passed
     Passing by command line overrides runtime datasource detection
     """
-    cmdline = util.get_cmdline()
-    ds_parse_0 = re.search(r"(?:^|\s)ds=([^\s;]+)", cmdline)
-    ds_parse_1 = re.search(r"(?:^|\s)ci\.ds=([^\s;]+)", cmdline)
-    ds_parse_2 = re.search(r"(?:^|\s)ci\.datasource=([^\s;]+)", cmdline)
+    return parse_cmdline_or_dmi(util.get_cmdline())
+
+
+def parse_cmdline_or_dmi(input: str) -> str:
+    ds_parse_0 = re.search(r"(?:^|\s)ds=([^\s;]+)", input)
+    ds_parse_1 = re.search(r"(?:^|\s)ci\.ds=([^\s;]+)", input)
+    ds_parse_2 = re.search(r"(?:^|\s)ci\.datasource=([^\s;]+)", input)
     ds = ds_parse_0 or ds_parse_1 or ds_parse_2
     deprecated = ds_parse_1 or ds_parse_2
     if deprecated:
--- a/tests/integration_tests/datasources/test_nocloud.py
+++ b/tests/integration_tests/datasources/test_nocloud.py
@@ -1,4 +1,6 @@
 """NoCloud datasource integration tests."""
+from textwrap import dedent
+
 import pytest
 from pycloudlib.lxd.instance import LXDInstance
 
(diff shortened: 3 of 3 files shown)


cloudinit/sources/__init__.py
```diff
@@ -366,7 +366,8 @@ def override_ds_detect(self) -> bool:
         does not run, _something_ needs to detect the kernel command line
         definition.
         """
-        if self.dsname.lower() == parse_cmdline().lower():
+        dsname = self.dsname.lower()
```
TheRealFalcon, line 369: Nit: This change isn't needed anymore (but also doesn't hurt anything)

TheRealFalcon (commented):
Changes LGTM. Can we get tests?

holmanb:
> Changes LGTM. Can we get tests?

Good call.

FYI - The qemu tutorial is currently broken by this regression and could be used to manually test with a modified image (mount it locally, chroot into it, install a test cloud-init deb).

I assume that using qemu via pycloudlib would be best so we can set the bios serial, but I haven't tried it out yet, and I don't think that we have any Jenkins runners set up to use it yet, right?

@TheRealFalcon Is that what you were thinking of or did you have something else in mind?

TheRealFalcon:
I was really only thinking about unit tests.

I was also thinking we would need QEMU, but it looks like LXD has recently added support for something like this:
```
$ lxc launch ubuntu:jammy jammy-test --vm -c raw.qemu="-smbios type=1,serial=HITHERE"
Creating jammy-test
Starting jammy-test
$ lxc exec jammy-test -- cat /sys/class/dmi/id/product_serial              
HITHERE
```

holmanb:
> I was really only thinking about unit tests.

Gotcha

> I was also thinking we would need QEMU, but it looks like LXD has recently added support for something like this:

Oh sweet.

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -362,18 +363,50 @@ def _merge_new_seed(cur, seeded):
     return ret
 
 
+def _parse_dmi() -> str:
+    serial = dmi.read_dmi_data("product_serial")
```
rvandegrift, line 367: s/product_serial/system-serial-number/ I think, otherwise this always fails

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -362,18 +363,50 @@ def _merge_new_seed(cur, seeded):
...
+                "scheme (http// or file://) to decide how to run."
+            ),
+        )
+
+        if "nocloud-net" == sources.parse_cmdline():
+            log_deprecated()
+            return True
+
+        serial = _parse_dmi().lower()
+        if serial in (self.dsname, "nocloud-net"):
```
rvandegrift, line 400: I think s/self.dsname/self.dsname.lower()/ is needed to fix the `ds=nocloud` case

rvandegrift:
@holmanb with the above changes, this is working!

holmanb:
> @holmanb with the above changes, this is working!

Thanks for the review! I just pushed the requested changes.

TheRealFalcon:
@holmanb 

I think something like this could work as a basic integration test. It should be hard to modify for http roots. Also, I did attempt to do the file pushing in the setup code, but LXD VMs need to be started before you can do that.
```python
SMBIOS_USERDATA = """\
#cloud-config
runcmd:
  - touch /var/tmp/smbios_test_file
"""
SMBIOS_SEED_DIR = "/smbios_seed"


def setup_nocloud_serial(instance: LXDInstance):
    subp(
        [
            "lxc",
            "config",
            "set",
            instance.name,
            "raw.qemu=-smbios "
            f"type=1,serial=ds=nocloud;s=file://{SMBIOS_SEED_DIR};h=myhost",
        ]
    )


@pytest.mark.lxd_setup.with_args(setup_nocloud_serial)
@pytest.mark.lxd_use_exec
@pytest.mark.skipif(
    PLATFORM != "lxd_vm",
    reason="Requires NoCloud with raw QEMU serial setup",
)
def test_smbios_seed(client: IntegrationInstance):
    client.execute(f"mkdir {SMBIOS_SEED_DIR}")
    client.write_to_file(f"{SMBIOS_SEED_DIR}/user-data", SMBIOS_USERDATA)
    client.write_to_file(f"{SMBIOS_SEED_DIR }/smbios_seed/meta-data", "")
    client.execute("cloud-init clean --logs")
    client.restart()
    assert client.execute("test -f /var/tmp/smbios_test_file").ok

```

cloudinit/sources/DataSourceNoCloud.py
holmanb: @rvandegrift It looks like applying this actually broke the test case, what were you seeing that led to this suggestion?

TheRealFalcon: @holmanb , I think `system-serial-number` is "more right" here. We use the map @rvandegrift mentioned to read the DMI value from `/sys`. If that fails, we fallback to reading using `dmidecode`. It looks like specifying `product_serial` as a key will trigger the fallback code, but we should be using `/sys` if it's available.

> broke the test case

Which test case? Mine will be broken unless you apply my other suggested fix :grimacing:

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +88,38 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
...
+@pytest.mark.lxd_setup.with_args(setup_nocloud_serial)
+@pytest.mark.lxd_use_exec
+@pytest.mark.skipif(
+    PLATFORM != "lxd_vm",
+    reason="Requires NoCloud with raw QEMU serial setup",
+)
+def test_smbios_seed(client: IntegrationInstance):
+    client.execute(f"mkdir {SMBIOS_SEED_DIR}")
+    client.write_to_file(f"{SMBIOS_SEED_DIR}/user-data", SMBIOS_USERDATA)
+    client.write_to_file(f"{SMBIOS_SEED_DIR }/smbios_seed/meta-data", "")
```
blackboxsw, line 122: This integration test fails due to the path typo duplicating SMBIOS_SEED_DIR/smbois_seed

```suggestion
    client.write_to_file(f"{SMBIOS_SEED_DIR}/meta-data", "")
```

blackboxsw (changes requested):
Minor test fix then approve

TheRealFalcon (dismissed):

cloudinit/sources/DataSourceNoCloud.py
```diff
@@ -362,18 +363,50 @@ def _merge_new_seed(cur, seeded):
...
+            util.deprecate,
+            deprecated="The 'nocloud-net' datasource name",
+            deprecated_version="24.1",
+            extra_message=(
+                "Use 'nocloud' instead, which uses the seedfrom protocol"
+                "scheme (http// or file://) to decide how to run."
+            ),
+        )
+
+        if "nocloud-net" == sources.parse_cmdline():
```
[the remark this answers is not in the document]
holmanb: > I think that we also need to change this line to if 'nocloud' in sources.parse_cmdline(), or something similar.

This is [done already in the base class](https://github.com/canonical/cloud-init/blob/beddd97c146e0bdba57b1fadc39e7b707351cef8/cloudinit/sources/__init__.py#L369).

> Otherwise, since nocloud-net is no longer wanted (or documented), specifying nocloud on the kernel CLI with an http(s) server will not be recognized.

We would have [several failing tests](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/test_kernel_commandline_match.py) if that was broken.

TheRealFalcon:
There's an extra `.swp` file that got added, but otherwise LGTM! I can reproduce nocloud-net failure on main, but pass on this branch.

blackboxsw (changes requested):
Since we are spending an integration test on NoCloudNet let's also double-dip and get a check on the deprecation logic too.

Also we need to pass the value we pull out of system-serial-number `/sys/class/dmi/id/product_serial` through our `new parse_cmdline_or_dmi` otherwise ds_detect is trying to find a full `ds=nocloud-net;s=http://blah` in (`nocloud`, `nocloud-net`)  which never matches.

```diff
--- a/cloudinit/sources/DataSourceNoCloud.py
+++ b/cloudinit/sources/DataSourceNoCloud.py
@@ -389,7 +389,9 @@ class DataSourceNoCloudNet(DataSourceNoCloud):
             log_deprecated()
             return True
 
-        serial = (dmi.read_dmi_data("product_serial") or "").lower()
+        serial = sources.parse_cmdline_or_dmi(
+            dmi.read_dmi_data("system-serial-number") or ""
+        ).lower()
         if serial in (self.dsname.lower(), "nocloud-net"):
             LOG.debug(
                 "Machine is configured by dmi serial number to run on "
diff --git a/tests/integration_tests/datasources/.test_nocloud.py.swp b/tests/integration_tests/datasources/.test_nocloud.py.swp
deleted file mode 100644
index 97bda3f6b..000000000
Binary files a/tests/integration_tests/datasources/.test_nocloud.py.swp and /dev/null differ
diff --git a/tests/integration_tests/datasources/test_nocloud.py b/tests/integration_tests/datasources/test_nocloud.py
index 2e5ea458f..9ab7ecd5b 100644
--- a/tests/integration_tests/datasources/test_nocloud.py
+++ b/tests/integration_tests/datasources/test_nocloud.py
@@ -121,7 +121,7 @@ def setup_nocloud_network_serial(instance: LXDInstance):
             "set",
             instance.name,
             "raw.qemu=-smbios "
-            "type=1,serial=ds=nocloud;s=http://0.0.0.0/;h=myhost",
+            "type=1,serial=ds=nocloud-net;s=http://0.0.0.0/;h=myhost",
         ]
     )
 
@@ -185,4 +185,7 @@ class TestSmbios:
         client.write_to_file(f"{SMBIOS_SEED_DIR}/vendor-data", "")
         assert client.execute("cloud-init clean --logs").ok
         client.restart()
+        assert "'nocloud-net' datasource name is deprecated" in client.execute(
+            "cloud-init status --format json"
+        )
         assert client.execute("test -f /var/tmp/smbios_test_file").ok
```

blackboxsw (approved):
LGTM! tests pass locally here too.

tests/integration_tests/datasources/test_nocloud.py
```diff
@@ -88,3 +90,99 @@ def test_nocloud_seedfrom_vendordata(client: IntegrationInstance):
...
+                """
+            ),
+        )
+        assert client.execute(
+            "chmod 644 /lib/systemd/system/local-server.service"
+        ).ok
+        assert client.execute("systemctl enable local-server.service").ok
+        client.write_to_file(
+            "/etc/cloud/cloud.cfg.d/91_do_not_use_lxd.cfg",
+            "datasource_list: [ NoCloud, None ]\n",
```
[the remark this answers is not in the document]
TheRealFalcon: If we do, we won't exercise the `ds_detect` logic

PR #4981, opened 2024-02-29
feat(net): provide network config to netplan.State for render

blackboxsw:
## Proposed Commit Message

```
feat(net): provide network config to netplan.State for render (#4981)

Rely on netplan API where present to write network config to
50-cloud-init.yaml. (SC-1402)

Allow cloud-init to be decoupled from security policies of netplan
libraries when rendering potentially senstive network configuratiion.

Newer netplan versions may filter sensitive network configuration parts
into separate files or subdirectories. Allow cloud-init to rely on
netplan's treatment of sensitive keys by using netplan.state.State to
write_yaml_file with cloud-init's requested base filename:
50-cloud-init.yaml.

If netplan security policy changes in the future, those changes will
be adopted under the hood of netplan.State.
```

## Additional Context
Some concerns with this approach: 
1.   Given that netplan API doesn't yet surface a public method, we are using _write_yaml_file under expectation that this interface does't change (needs netplan team sign-off for awareness) CC: @slyon 
2. The current `State._write_yaml_file` doesn't accept a header, strips any inline comments when processing network config yaml. This means that 50-cloud-init.yaml, as rendered by netplan's State._write_yaml will no long contain the helpful comment/header:
```
# network: {config: disabled}
````

If we want to take this approach to use netplan API to avoid policy changes in netplan, we probably should look to a feature request for `netplan.State._write_yaml_file` to allow a header or footer parameter to allow extending the config content that is written by netplan.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -4,7 +4,6 @@ import copy
 import io
 import logging
 import os
-import shutil
 import textwrap
 from tempfile import SpooledTemporaryFile
 from typing import Optional, cast
@@ -19,7 +18,6 @@ from cloudinit.net import (
     subnet_is_ipv6,
 )
 from cloudinit.net.network_state import NET_CONFIG_TO_V2, NetworkState
-from cloudinit.temp_utils import mkdtemp
 
 CLOUDINIT_NETPLAN_FILE = "/etc/netplan/50-cloud-init.yaml"
 
@@ -250,10 +248,9 @@ def netplan_api_write_yaml_file(net_config_content: str) -> bool:
             CLOUDINIT_NETPLAN_FILE,
         )
         return False
-    cfg_dir = mkdtemp()
     try:
-        with SpooledTemporaryFile(dir=cfg_dir, mode="w+b") as f:
-            f.write(util.encode_text(net_config_content))
+        with SpooledTemporaryFile(mode="w") as f:
+            f.write(net_config_content)
             f.flush()
             f.seek(0, io.SEEK_SET)
             parser = Parser()
@@ -276,34 +273,29 @@ def netplan_api_write_yaml_file(net_config_content: str) -> bool:
             e,
         )
         return False
-    if os.path.exists(cfg_dir):
-        shutil.rmtree(cfg_dir)
     LOG.debug("Rendered netplan config using netplan python API")
     return True
 
 
-def fallback_write_netplan_yaml(filename: str, content: str) -> bool:
-    """Write netplan config to filename reporting when content is unchanged.
+def has_netplan_config_changed(cfg_file: str, content: str) -> bool:
+    """Return True when new netplan config has changed vs previous."""
+    if not os.path.exists(cfg_file):
+        # This is our first write of netplan's cfg_file, representing change.
+        return True
+    # Check prev cfg vs current cfg. Ignore comments
+    prior_cfg = util.load_yaml(util.load_text_file(cfg_file))
+    return prior_cfg != util.load_yaml(content)
 
-    Return bool: True if new content is the same as original content.
-    """
-    # Determine if existing config files have the same content
-    same_netplan_config = False
-    if os.path.exists(filename):
-        hashed_content = util.hash_buffer(io.BytesIO(content.encode()))
-        with open(filename, "rb") as f:
-            hashed_original_content = util.hash_buffer(f)
-        if hashed_content == hashed_original_content:
-            same_netplan_config = True
 
+def fallback_write_netplan_yaml(cfg_file: str, content: str):
+    """Write netplan config to cfg_file because python API was unavailable."""
     mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
-    if not same_netplan_config and os.path.exists(filename):
-        current_mode = util.get_permissions(filename)
+    if os.path.exists(cfg_file):
+        current_mode = util.get_permissions(cfg_file)
         if current_mode & mode == current_mode:
             # preserve mode if existing perms are more strict
             mode = current_mode
-    util.write_file(filename, content, mode=mode)
-    return same_netplan_config
+    util.write_file(cfg_file, content, mode=mode)
 
 
 class Renderer(renderer.Renderer):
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cloudinit/net/netplan.py [hunk 1]
```diff
@@ -235,6 +237,61 @@ def _clean_default(target=None):
...
+            parser = Parser()
+            parser.load_yaml(f)
+            state_output_file = State()
+            state_output_file.import_parser_results(parser)
+
+            # Write our desired basename 50-cloud-init.yaml, allow netplan to
+            # determine default root-dir /etc/netplan and/or specialized
+            # filenames or read permissions based on whether this config
+            # contains secrets.
+            state_output_file._write_yaml_file(
```
blackboxsw, line 278: @slyon points here for awareness:
 1. cloud-init is calling this private method from netplan in absence of a public netplan API intended to let netplan choose ultimately what file to write and in which directories. Can we consider this a `safe` method to call that won't change until cloud-init is able to call a public method in the future?
 2. If netplan grows the ability in the future, do you expect the functionality to separate sensitive config form world-readable config will live under this method call?
 3.

cloudinit/net/netplan.py
```diff
@@ -235,6 +237,61 @@ def _clean_default(target=None):
...
+    except ImportError:
+        LOG.debug(
+            "No netplan python module. Fallback to write %s",
+            CLOUDINIT_NETPLAN_FILE,
+        )
+        return False
+    cfg_dir = mkdtemp()
+    try:
+        with open(os.path.join(cfg_dir, "ci-net-cfg.yaml"), "wb+") as f:
+            f.write(util.encode_text(net_config_content))
```
slyon, line 266: note (non-blocking): You could consider doing this all in-memory, using `io.StringIO` and/or `os.memfd_create`

cloudinit/net/netplan.py
slyon: 1. Yes
2. Yes

There is an idea of introducing a public `.write_yaml()` method. But we have to work out some kinks first. `_write_yaml_file()` is good for now, as its underlying CAPI is a public symbol `netplan_state_write_yaml_file()`. And it's also what is being used in the Netplan CLI.

slyon (approved):
lgtm

cloudinit/net/netplan.py
blackboxsw: @slyon excellent and thank you. I was wondering what you though about a feature request for an optional `header` param to _write_yaml_file in the future that would allow cloud-init to inject custom comments into the file being written?

cloudinit/net/netplan.py
TheRealFalcon: Or `tempfile.SpooledTemporaryFile`

cloudinit/net/netplan.py
slyon: I was already pondering about ways to keep the comments, but that's not currently possible with the underlying `libyaml`. I like the idea of a `header` parameter.

TheRealFalcon: Another thing that might be helpful is returning the path(s) of any files written so that at the very least we could add our own header after the fact.

cloudinit/net/netplan.py
```diff
@@ -287,22 +344,24 @@ def render_network_state(
...
-
-        mode = 0o600 if features.NETPLAN_CONFIG_ROOT_READ_ONLY else 0o644
-        if not same_content and os.path.exists(fpnplan):
-            current_mode = util.get_permissions(fpnplan)
-            if current_mode & mode == current_mode:
-                # preserve mode if existing perms are more strict than default
-                mode = current_mode
-        util.write_file(fpnplan, content, mode=mode)
+        if not netplan_api_write_yaml_file(content):
+            # No netplan api, fallback to write config  directly.
```
TheRealFalcon, line 349: Nit: As we're adding some extra complexity here, I think the code under this `if block` would be better as its own function. `fallback_write_yaml_file` or something like that.

TheRealFalcon (commented):
Generally LGTM. I do agree with slyon that we should use something like StringIO or perhaps a SpooledTemporaryFile rather than the current tmpdir approach.

Also, will netplan currently write a separate private config for the sensitive parts? If so, I think it's worth a separate integration test ensuring the pieces get written correctly.

slyon:
> Also, will netplan currently write a separate private config for the sensitive parts? If so, I think it's worth a separate integration test ensuring the pieces get written correctly.

Netplan does not currently do that. It's just a card in our backlog so far, see [FR-3285](https://warthogs.atlassian.net/browse/FR-3285) for more details.

cloudinit/net/netplan.py
```diff
@@ -235,6 +237,61 @@ def _clean_default(target=None):
...
+        from netplan.state import State
+    except ImportError:
+        LOG.debug(
+            "No netplan python module. Fallback to write %s",
+            CLOUDINIT_NETPLAN_FILE,
+        )
+        return False
+    cfg_dir = mkdtemp()
+    try:
+        with open(os.path.join(cfg_dir, "ci-net-cfg.yaml"), "wb+") as f:
```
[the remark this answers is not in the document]
blackboxsw: This is resolved/dropped as I'm now using SpooledTemporaryFile which defaults to "w+b" and ensuring we are encoding as bytes

TheRealFalcon:
I think we can simplify a bit more, can't we?
```diff
diff --git a/cloudinit/net/netplan.py b/cloudinit/net/netplan.py
index fe97fdcc9..46cbe3a9d 100644
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -250,11 +250,9 @@ def netplan_api_write_yaml_file(net_config_content: str) -> bool:
             CLOUDINIT_NETPLAN_FILE,
         )
         return False
-    cfg_dir = mkdtemp()
     try:
-        with SpooledTemporaryFile(dir=cfg_dir, mode="w+b") as f:
-            f.write(util.encode_text(net_config_content))
-            f.flush()
+        with SpooledTemporaryFile(mode="w") as f:
+            f.write(net_config_content)
             f.seek(0, io.SEEK_SET)
             parser = Parser()
             parser.load_yaml(f)
@@ -276,8 +274,6 @@ def netplan_api_write_yaml_file(net_config_content: str) -> bool:
             e,
         )
         return False
-    if os.path.exists(cfg_dir):
-        shutil.rmtree(cfg_dir)
     LOG.debug("Rendered netplan config using netplan python API")
     return True
 
```

blackboxsw:
@TheRealFalcon correct, we can simplify this. I also adapted our netplan config delta check to allow it to be used for both netplan API and fallback method to ensure we don't run `netplan generate` unnecessarily when netplan config hasn't changed.
`CLOUD_INIT_CLOUD_INIT_SOURCE=IN_PLACE CLOUD_INIT_OS_IMAGE=jammy .tox/integration-tests/bin/pytest --  tests/integration_tests/test_networking.py::TestNetplanGenerateBehaviorOnReboot::test_skip`   for testing.

TheRealFalcon (commented):
+1 to the primary changes, but you have test failures now.

TheRealFalcon (approved):
LGTM!

Test nit: We don't really have any unit test coverage of the new functionality. But also:
1. We have integration test coverage.
2. The whole thing is essentially a wrapper to an outside call, so there's not really much value to a unit test.
3. It's all wrapped in exceptions to fallback on failure.

Given these 3 things, I'm ok with it.

Issue #4688, opened 2023-12-13
[enhancement]: Relocate json schemas somewhere more sensible


holmanb:
# Enhancement

All of our jsonschemas are hidden down under `cloudinit/config/schemas/`, which doesn't make sense:

a) these files contain data not Python code
b) some of these files are not specific to configuration modules
c) the `__init__.py` was never required

These file should exist in a directory under the top directory, probably named something like `schema` (alternatively `jsonschema` or maybe something more generic like `data`).

Expected changes:
- cloudinit/
- test/unittest/
- `setup.py`
- possibly distro packaging too.

antwigambrah:
Hi @holmanb  , can this task be assigned to me?

aciba90:
Hello, @antwigambrah, thanks for improving cloud-init. We do not typically assign issues to people, but please, feel free to put together a PR. If you have any questions regarding the implementation, please ask them here or in [the IRC channel on libera #cloud-init](https://docs.cloud-init.io/en/latest/index.html#project-and-community).

Contributing guidelines: https://docs.cloud-init.io/en/latest/development/index.html

antwigambrah:
> Hello, @antwigambrah, thanks for improving cloud-init. We do not typically assign issues to people, but please, feel free to put together a PR. If you have any questions regarding the implementation, please ask them here or in [the IRC channel on libera #cloud-init](https://docs.cloud-init.io/en/latest/index.html#project-and-community).
> 
> Contributing guidelines: https://docs.cloud-init.io/en/latest/development/index.html

Just seeing this . Thanks @holmanb @aciba90 Added the pr

afbjorklund:
It would be better to deploy these files somewhere under cloud-init.io domain, instead of directly going to github...

https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json

holmanb:
> It would be better to deploy these files somewhere under cloud-init.io domain, instead of directly going to github...
>
>https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json

We publish them already to the jsonschema store, if you want a non-github location to pull from.

afbjorklund:
The current location works OK with the tools, just thinking that normally you would download files from a web server?

The https://www.schemastore.org/json/ seems to be linking to the same address, but I  just embedded a copy for now.

PR #4991, opened 2024-03-01
refactor : Relocate json schemas

antwigambrah:
## Proposed Commit Message

```
refactor: relocate json schemas 

Fixes GH-4688 
```

## Additional Context

fixes https://github.com/canonical/cloud-init/issues/4688

## Checklist

## Merge type


catmsred:
Thank you for helping improve cloud-init!

A general question -- why did you decide on the schemas for the top level directory name as opposed to data as suugested in the bug report?

I think you may also need to update the schemas themselves, e.g. `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/versions.schema.cloud-config.json"` to `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/schemas/versions.schema.cloud-config.json"` although I haven't checked.

Can you provide some context on how you've tested your changes?

cloudinit/config/schema.py
```diff
@@ -1429,7 +1430,7 @@ def load_doc(requested_modules: list) -> str:
 
 
 def get_schema_dir() -> str:
-    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "schemas")
+    return os.path.join(Path(__file__).parent.parent.parent, "schemas")
```
TheRealFalcon, line 1433: This unfortunately won't work. Only the `cloudinit` directory gets deployed as a unit, not the parent `cloud-init` directory. In order to accommodate this, we'll need to update `setup.py` to write the schema files to correct location on installation.

I would think somewhere like `/usr/lib/cloud-init/schemas/` would be most appropriate, but I don't think we have a definite directory for something like this yes.

@holmanb , thoughts?

TheRealFalcon (changes requested):

antwigambrah:
> Thank you for helping improve cloud-init!
> 
> A general question -- why did you decide on the schemas for the top level directory name as opposed to data as suugested in the bug report?
> 
> I think you may also need to update the schemas themselves, e.g. `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/versions.schema.cloud-config.json"` to `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/schemas/versions.schema.cloud-config.json"` although I haven't checked.
> 
> Can you provide some context on how you've tested your changes?

For testing, I didn't add any additional test code but made sure   test_shema.py. Any additional kind of test you would like me run is welcome . Thanks for letting me know of the other location, will update it accordingly.

cloudinit/config/schema.py
holmanb: > This unfortunately won't work. Only the cloudinit directory gets deployed as a unit, not the parent cloud-init directory. In order to accommodate this, we'll need to update setup.py to write the schema files to correct location on installation.

+1, agreed

> I would think somewhere like /usr/lib/cloud-init/schemas/ would be most appropriate, but I don't think we have a definite directory for something like this yes.

Agreed, although /usr/lib is for [libraries, shared objects, and internal executables](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html). Maybe a cloud-init directory under [/usr/share](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) would be more appropriate?

blackboxsw: +1 on `/usr/share/cloud-init/schemas` as that aligns well with FHS for architecture independent data. We have some good examples of json schema files from other packages:
- vscode: /usr/share/code/resources
- /usr/share/lintian/data
- /usr/share/gnome-shell/extensions


Also note when we move forward on this approach we'll may need a one-liner followup PR against [schemastore updating the github URLs for the base schema](https://github.com/SchemaStore/schemastore/pull/2252/files#diff-1b8f038f5afb1158263a1fc83b9c0ca5a7438cceb90adad99d87168f70edc815R785) so that cloud-init schema can be auto-ingested by tools that automatically validate our user-data schema

antwigambrah: I can see in setup.py there is an existing /share/doc/cloud-init folder in which the schemas folder can also be created . Should it be maintained I should stick with moving the files to /usr/share/cloud-init/schemas .

antwigambrah:
> Thank you for helping improve cloud-init!
> 
> A general question -- why did you decide on the schemas for the top level directory name as opposed to data as suugested in the bug report?
> 
> I think you may also need to update the schemas themselves, e.g. `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/versions.schema.cloud-config.json"` to `"$id": "https://raw.githubusercontent.com/canonical/cloud-init/main/schemas/versions.schema.cloud-config.json"` although I haven't checked.
> 
> Can you provide some context on how you've tested your changes?

I have update the code based on suggestions made by @catmsred  @TheRealFalcon  and @blackboxsw  . I have added  test cases for the new file location and the ability to load from the location .

cloudinit/config/schema.py
holmanb: > I can see in setup.py there is an existing /usr/share/doc/cloud-init folder in which the schemas folder can also be created . Should it be maintained I should stick with moving the files to /usr/share/cloud-init/schemas .

Yes, I think that we agree that it should go there.

holmanb (changes requested):
@antwigambrah sorry for the delayed response. Yes, lets put it in `/usr/shar/cloud-init/schemas/`. That aligns with FHS expectations.

antwigambrah:
> @antwigambrah sorry for the delayed response. Yes, lets put it in `/usr/shar/cloud-init/schemas/`. That aligns with FHS expectations.

updated schema locations

antwigambrah:
Hi @TheRealFalcon  any update on this pr

cloudinit/config/schema.py
```diff
@@ -1553,8 +1553,12 @@ def load_doc(requested_modules: list) -> str:
     return docs
 
 
-def get_schema_dir() -> str:
-    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "schemas")
+def get_schema_dir(schema_dir=None) -> str:
+
+     if schema_dir:
+        return os.path.join(schema_dir, "schemas")
```
holmanb, line 1559: This path never actually gets called does it?

cloudinit/config/schema.py
```diff
@@ -42,7 +42,7 @@
     load_text_file,
     write_file,
 )
-
```
holmanb, line 45: unnecessary whitespace change

tests/unittests/config/test_schema.py
```diff
@@ -128,6 +130,7 @@ def test_versioned_cloud_config_schema_is_valid_json(
         self, schema, error_msg
     ):
         schema_dir = get_schema_dir()
+
```
holmanb, line 133: unnecessary whitespace change

holmanb (changes requested):
@antwigambrah Thanks for following up on this PR, and great work so far! I see that you've signed the CLA, but we need to add your github username to https://github.com/canonical/cloud-init/blob/main/tools/.github-cla-signers. Can you please include that in this PR?

This is generally looking good. I left a couple of nitpicks inline.

The issue that have with the current state of this PR is that, while unittests pass currently, I don't see a way for cloud-init to find the correct location on the filesystem to load these files when packaged for distros. I'm not sure about the best way to resolve this, but I suspect that it may involve a default path in `settings.py` so that distros can patch it if needed, however I'm not certain how best to make sure that when we want cloud-init to load the files from the relative codepath it will work while still working when installed. I'll try to come up with something, but if you have any ideas that would be appreciated!

PR #4992, opened 2024-03-01
Switch to pyproject from distutils in debian packaging post-Noble

holmanb:
## Additional Context
The commit message explains it:
```
d/control: Add build-depends on pybuild-plugin-pyproject

This switches the pybuild backend plugin from distutils to pyproject.
Distutils, the most commonly used build system plugin, is deprecated[1]
and removed from the standard library in upstream Python starting in
3.12. This plugin still works on Noble (which provides Python 3.12) due
to the python3-distutils package, which is based on distutils from Python
3.11. 

The pyproject plugin is expected to replace distutils, due to its
support of PEP-517. Upstream cloud-init currently supports PEP-517[2].

To enable building with the pyproject build system,
pybuild-plugin-pyproject must be a build dependency, as well as any
build tools specified in pyproject.toml[3]. Cloud-init already build
depends on python3-setuptools, so only pybuild-plugin-pyproject needs
to be added.

The pybuild-plugin-pyproject package is in universe, however it is
acceptable because this is a build-only dependency[4].

[1] https://docs.python.org/3.10/whatsnew/3.10.html#distutils-deprecated
[2] https://github.com/canonical/cloud-init/commit/8836c7824533025768ec82c350a965b76a744b1e
[3] https://manpages.ubuntu.com/manpages/noble/man1/pybuild.1.html
[4] https://github.com/canonical/ubuntu-mir (see [Dependencies] section)
```

Note: I added the branch `ubuntu/noble` to the repository, which is a copy of `ubuntu/devel`. This change is only intended to take effect in 24.10 and later. Noble and earlier releases will continue to use the distutils pybuild plugin.

## Test Steps
```bash
build-package -- -S -d
sbuild --dist=noble --arch=amd64 ../out/cloud-init_24.1-0ubuntu2.dsc
```

## Merge type


TheRealFalcon (approved):
Thanks for this. I have been frustrated that distutils is still the default even though it's been deprecated forever and being removed. Also, thanks for the references. I didn't know that a build-time dependency could rely on a universe package.

PR #4994, opened 2024-03-02
bug(wait_for_url): when exceptions occur url is unset, use url_exc

blackboxsw:
When processing multiple urls in wait_for_url and a timeout error occurs, the local variable url is None yet url_exc.url is set.

Fallback to url_exc.url with url is unet to provide sensible warning messages:
 Calling '<full_failed_url>' failed [3/120s]: request error

instead of:
 Calling 'None' failed [3/120s]

LP: #2055077

* two more commits to fix other unit tests (reduce test time for test_ec2, fix openstack test to also mock requests against IPv6 route

## Additional Context
Confusing logs generated from Ec2 datasource detection on any system where cloud-init is disabled where 
Failure to wait_for_urls finds no resolvable URLs.

```
 2024-02-26 19:29:49,617 WARNING cloudinit.sources.DataSourceEc2:583 Calling 'None' failed [119/120s]: request error [HTTPConnectionPool(host='169.254.169.254', port=80): Max retries exceeded with url: /2009-04-04/meta-data/instance-id (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x74a6d282ad70>, 'Connection to 169.254.169.254 timed out. (connect timeout=18.0)'))]
```
Fallback to url_exc.url which is set with the appropriate failed url to improve logged warnings.

## Test Steps
```
lxc launch ubuntu-daily:noble nn
cat > 91-ds-only-ec2.cfg <<EOF
datasource_list: [ Ec2, None ]
EOF
lxc file push 91-ds-only-ec2.cfg  nn/etc/cloud/cloud.cfg.d/
lxc exec nn -- cloud-init clean --logs
lxc exec nn -- rm -rf /run/cloud-init
lxc exec nn -- cloud-init init
```
## Checklist

## Merge type

### DIFF
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -575,7 +575,7 @@ def wait_for_url(
         time_taken = int(time.time() - start_time)
         max_wait_str = "%ss" % max_wait if max_wait else "unlimited"
         status_msg = "Calling '%s' failed [%s/%s]: %s" % (
-            url,
+            url or getattr(url_exc, "url", "url ? None"),
             time_taken,
             max_wait_str,
             reason,
--- a/tests/unittests/sources/test_ec2.py
+++ b/tests/unittests/sources/test_ec2.py
@@ -695,29 +695,13 @@ class TestEc2(test_helpers.ResponsesTestCase):
             m_readurl.side_effect = (
                 conn_error,
                 conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
-                conn_error,
                 mock_success,
             )
             with mock.patch("cloudinit.url_helper.time.sleep"):
                 self.assertTrue(ds.wait_for_metadata_service())
 
         # Just one /latest/api/token request
-        self.assertEqual(19, len(m_readurl.call_args_list))
+        self.assertEqual(3, len(m_readurl.call_args_list))
         for readurl_call in m_readurl.call_args_list:
             self.assertIn("latest/api/token", readurl_call[0][0])
 
--- a/tests/unittests/sources/test_openstack.py
+++ b/tests/unittests/sources/test_openstack.py
@@ -136,7 +136,7 @@ def _register_uris(version, ec2_files, ec2_meta, os_files, *, responses_mock):
 
     responses_mock.add_callback(
         responses.GET,
-        re.compile(r"http://169.254.169.254/.*"),
+        re.compile(r"http://(169.254.169.254|\[fe80::a9fe:a9fe\])/.*"),
         callback=get_request_callback,
     )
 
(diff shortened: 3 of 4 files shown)


holmanb:
> on any system where cloud-init is disabled where
Failure to wait_for_urls finds no resolvable URLs.

This isn't a supported path, but +1 for better debugging messaging regardless. 

Also unit tests fail currently.

blackboxsw:
> > on any system where cloud-init is disabled where
> > Failure to wait_for_urls finds no resolvable URLs.
> 
> This isn't a supported path, but +1 for better debugging messaging regardless.
> 
> Also unit tests fail currently.

Agreed and cloud-init stagins.Init() shouldn't ever be called when ds-identify has disabled cloud-init. But, this failure path still exists if ds-identify isn't involved in the boot process. Our `cloud-init init` will use our python runtime via `DataSourceEc2.get_data` to try discovering ec2 endpoints on a system and will fail with this log warning 'None' an any environment without ds-identify if the configured `datasource_list:` contains `Ec2`. I'll sort the unittest issue there. also looks like we have some mock/sleep leaks missing a mocked sleep in DataSourceScaleway tests as those tests shouldn't take 2 seconds to run. May fix that too.

holmanb (approved):
+1, looks good

> because systemd-based, and BSD environments now, should not invoke cloud-init init because ds-identify is run. 

also [openrc](https://github.com/canonical/cloud-init/blob/main/sysvinit/openrc/cloud-init-ds-identify) I think that's all of the supported ones, am I missing any? 

> But, this failure path still exists if ds-identify isn't involved in the early boot process. In those cases, `cloud-init init` will use our python runtime via `DataSourceEc2.get_data` to still try discovering ec2 endpoints on a system and will fail with this log warning 'None' an any environment if the configured `datasource_list:` contains `Ec2`.

+1, agreed we should fix this. Now that ds-identify support has gotten more standardized, I just think that this is getting further and further from the supported path. 

Aside: This is another example of cloud-init's lack of a consumable programmatic api leaving consumers of cloud-init doing awkward things and (both us and them) learning by trial and error/traceback. Maybe we should start to think about what a well-defined library to meet the common needs of cloud-init consumers might look like?

cloudinit/url_helper.py
```diff
@@ -575,7 +575,7 @@ def read_url_handle_exceptions(
         time_taken = int(time.time() - start_time)
         max_wait_str = "%ss" % max_wait if max_wait else "unlimited"
         status_msg = "Calling '%s' failed [%s/%s]: %s" % (
-            url,
+            url or getattr(url_exc, "url", None),
```
TheRealFalcon, line 578: Instead of `None`, can we use something like `url`? `None` in a log message will make me smell a logic error.

TheRealFalcon:
> I just think that this is getting further and further from the supported path

To be fair, we'll hit this if we request a url that times out. I just ran into this when trying to make some Oracle changes and independently fixed it not knowing that this PR was up.

blackboxsw:
> Maybe we should start to think about what a well-defined library to meet the common needs of cloud-init consumers might look like?

I think trying to drive parts of cloud-init from well-defined API entrypoints would be worthwhile for performing certain levels of tasks.
The typical entry-points our known consumers: blocking on cloud-init completion, subiuquity/pro-client are really looking at introspection of cloud-init metadata/user-data after a run to understand what platform I'm on and whether they need to perform certain operations actions based on presence of certain user-data. I don't know if we have a strong motivation at the moment to surface callable cloud-init boot stages cleanly from an API, but when we look at boot performance optimization, that may come into play.

PR #4997, opened 2024-03-04
Fall back to cached local ds if no valid ds found

### LINKED ISSUE #3402: OVF datasource should check if instant id is still on VMware Platform
This bug was originally filed in Launchpad as [LP: #1835205](https://bugs.launchpad.net/cloud-init/+bug/1835205)
<details>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2019-07-03T11:24:36.299696+00:00
date_fix_committed = None
date_fix_released = None
id = 1835205
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1835205
milestone = None
owner = pengpengs
owner_name = Pengpeng Sun
private = False
status = triaged
submitter = pengpengs
submitter_name = Pengpeng Sun
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Pengpeng Sun(pengpengs)** wrote on 2019-07-03T11:24:36.299696+00:00_

Currently DatasourceOVF does not check the instance id to determine if current instance is a new instance or not, so that every booting, cloud-init will go through entire datasource list.

This leads an issue:
When a VM's network is customized to static IP by cloud-init DatasourceOVF and then the VM is rebooted, the VM's network will be changed to DHCP after rebooting since this time no customization config file will be found. cloud-init uses DatasourceNone which change network to "default" configuration.

The expected behaviors are:
1. cloud-init checks the instance id to know if this instance is "iid-vmware-xxxxxx" which means it's a VMware VM. If yes, it should always use datasourceOVF.
2. When there has customization config file, datasourceOVF parse the configures and enable nics as usual.
3. When there is no customization config file, datasourceOVF wait for it until timeout and then cloud-init does NOT apply other datasources.

PengpengSun:
## Proposed Commit Message

```
Fall back to cached local ds if no valid ds found

Rebooting an instance which has finished VMware guest
customization with DataSourceVMware will load
DataSourceNone due to metadata is NOT available.

This is mostly a re-post of PR#229, few differences are:
1. Let ds decide if fallback is allowed, not always fall back
   to previous cached LOCAL ds.
2. No comparing instance-id of cached ds with previous instance-id
   due to I think they are always identical.

Fixes GH-3402 (GitHub Issue number. Remove line if irrelevant)
LP: #1835205 (Launchpad bug number. Remove line if irrelevant)
```

## Additional Context
Fixes #3402 

## Checklist

## Merge type


cloudinit/sources/__init__.py
```diff
@@ -925,6 +925,10 @@ def check_instance_id(self, sys_cfg):
         # quickly (local check only) if self.instance_id is still
         return False
 
+    def check_fallback(self):
```
holmanb, line 928: Could we make this method name more descriptive please? Also we should add a docstring that will communicate to other datasource authors what the purpose of this method is. There are many fallbacks in cloud-init, so this isn't very meaningful on its own.

holmanb (changes requested):
When this happens, does ds-identify still correctly identify the platform?

cloudinit/sources/__init__.py
PengpengSun: Thanks for the review.
Yes, I've renamed this method(please suggest a better name if needed), and added a docstring.

PengpengSun:
> When this happens, does ds-identify still correctly identify the platform?

Not sure if I get your question correctly, with my understanding, ds-identify identifies the platform before init --local, so this shall not have impact on ds identify process. The purpose of this change is to add an opportunity for the cached ds to be restored after no data can be gotten from ds in the ds list.

cloudinit/stages.py
```diff
@@ -359,20 +359,36 @@ def _get_data_source(self, existing) -> sources.DataSource:
...
-                self.distro,
-                self.paths,
-                copy.deepcopy(self.ds_deps),
-                cfg_list,
-                pkg_list,
-                self.reporter,
-            )
-            LOG.info("Loaded datasource %s - %s", dsname, ds)
+            try:
+                (cfg_list, pkg_list) = self._get_datasources()
```
holmanb, line 363: nit: parens in assignment not needed

cloudinit/stages.py
```diff
@@ -359,20 +359,36 @@ def _get_data_source(self, existing) -> sources.DataSource:
...
+                # Deep copy so that user-data handlers can not modify
+                # (which will affect user-data handlers down the line...)
+                (ds, dsname) = sources.find_source(
```
holmanb, line 366: nit: parens in assignment are not needed

cloudinit/stages.py
PengpengSun: Parens removed, thanks.

cloudinit/stages.py
PengpengSun: Ditto.

PengpengSun:
Hi @holmanb , I'm removing WIP from title, please kindly review this again and let me know if any updates needed. Thanks!

cloudinit/stages.py
```diff
@@ -359,20 +359,36 @@ def _get_data_source(self, existing) -> sources.DataSource:
...
+                )
+                util.del_file(self.paths.instance_link)
+                LOG.info("Loaded datasource %s - %s", dsname, ds)
+            except sources.DataSourceNotFoundException as e:
+                if existing != "check":
+                    raise e
+                ds = self._restore_from_cache()
+                if (
+                    ds
+                    and hasattr(ds, "check_if_fallback_is_allowed")
```
holmanb, line 383: This is undesirable for two reasons:

- old pickles will never acquire this attribute
- encapsulation: we shift pickle handling outside of the class definition which should handle pickle issues directly

How about the following modification?

```diff
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index 2c4d54871..e225e4ee8 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -185,6 +185,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             self._dhcp_client = None
         if not hasattr(self, "_fallback_interface"):
             self._fallback_interface = None
+        if not hasattr(self, "check_if_fallback_is_allowed"):
+            self.check_if_fallback_is_allowed = lambda: False
 
     def _validate_entry(self, entry):
         if isinstance(entry, str):
diff --git a/cloudinit/stages.py b/cloudinit/stages.py
index 13bc26d26..c228805d1 100644
--- a/cloudinit/stages.py
+++ b/cloudinit/stages.py
@@ -378,11 +378,7 @@ class Init:
                 if existing != "check":
                     raise e
                 ds = self._restore_from_cache()
-                if (
-                    ds
-                    and hasattr(ds, "check_if_fallback_is_allowed")
-                    and ds.check_if_fallback_is_allowed()
-                ):
+                if ds and ds.check_if_fallback_is_allowed():
                     LOG.info(
                         "Restored fallback datasource from checked cache: %s",
                         ds,
```

PengpengSun: Hi @holmanb,
Thanks for the review! I've updated the code accordingly(I think it'd be sources/__init__.py), PTAL.

PengpengSun: tox is not happy with the last commit, need to fix it.

PengpengSun: Hi @holmanb , I've updated code again which fixed the test error (linkcheck error seems not related), PTAL, thanks!

ani-sinha:
@holmanb is it possible to backport this patch to 24.1 train maybe in 24.1.5?

PR #4999, opened 2024-03-04
Switch back to distutils due to bug in dh-python

holmanb:
## Additional Context
Switch back to distutils due to bug in dh-python

I filed [a bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1065441).


blackboxsw (approved):
LGTM. Let's back this out until we have a path forward here w/ pybuild-plugin-pyproject and our configuration.

PR #5009, opened 2024-03-05
feat(url_helper): add default timeout to readurl() to prevent hanging

a-dubs:
## Proposed Commit Message

```
feat(url_helper): add default timeout to readurl() to prevent hanging

If there is an issue with networking, we don't want any calls to hang 
forever. This is never a desireable outcome, because this prevents 
cloud-init from finishing and thus makes debugging much more difficult.
```

## Additional Context
This issue was discovered when attempting to contact the IMDS from an
oracle VM that had failed setting up its networking.  

## Checklist

- [X] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [X] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [X] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

### DIFF
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -193,7 +193,7 @@ def _get_ssl_args(url, ssl_details):
 def readurl(
     url,
     data=None,
-    timeout=None,
+    timeout=30,
     retries=0,
     sec_between=1,
     headers=None,
@@ -215,7 +215,8 @@ def readurl(
     :param data: Optional form data to post the URL. Will set request_method
         to 'POST' if present.
     :param timeout: Timeout in seconds to wait for a response. May be a tuple
-        if specifying (connection timeout, read timeout).
+        if specifying (connection timeout, read timeout). Default is 30 so that
+        we don't hang forever.
     :param retries: Number of times to retry on exception if exception_cb is
         None or exception_cb returns True for the exception caught. Default is
         to fail with 0 retries on exception.
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -193,7 +193,7 @@ def _get_ssl_args(url, ssl_details):
 def readurl(
     url,
     data=None,
-    timeout=30,
+    timeout=60,
     retries=0,
     sec_between=1,
     headers=None,
@@ -215,7 +215,7 @@ def readurl(
     :param data: Optional form data to post the URL. Will set request_method
         to 'POST' if present.
     :param timeout: Timeout in seconds to wait for a response. May be a tuple
-        if specifying (connection timeout, read timeout). Default is 30 so that
+        if specifying (connection timeout, read timeout). Default is 60 so that
         we don't hang forever.
     :param retries: Number of times to retry on exception if exception_cb is
         None or exception_cb returns True for the exception caught. Default is


TheRealFalcon (commented):
LGTM with an inline comment.

Let's get a +1 from one of the other committers that hasn't been staring at this code all week :zany_face:

cloudinit/url_helper.py
```diff
@@ -193,7 +193,7 @@ def _get_ssl_args(url, ssl_details):
 def readurl(
     url,
     data=None,
-    timeout=None,
+    timeout=30,
```
TheRealFalcon, line 196: I found today that the [EC2 datasource](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceEc2.py#L90) sets a default timeout to 50, so I think we should bump this up to 60 as a global default.

a-dubs: 50s? Are their server rooms full of carrier pigeons??? 
Just kidding. I agree 60s is good!

holmanb (changes requested):
> we don't want any calls to hang forever. This is never a desireable outcome

Yeah this would probably have made debugging that issue easier. I agree that this proposal is probably fine for most cloud environments, but I'm a little worried about regression potential of this kind of change. How confident are we that "never" really means "never"?

I'm imagining a hypothetical workload which might kick off a cloud-init instance to start, simultaneously fire off building an imds from scratch (or something that might take really long to finish), and let the blocked instance wait "indefinitely" (30 minutes, 2 hours, idk) until the server is available and serving before the instance would continue.

Are we sure that this can't happen in any cloud-init codepaths? I see that NoCloud uses a timeout of 5s in `util.read_seeded()`, so that shouldn't be a culprit, but are there any other code paths that might do something like this?

Some scientific computing clusters take a really long time to boot (think minutes to hours), and I know that cloud-init is now used in many of those environments.

I'd like to make sure that we're really confident that we **never** want to wait forever before we change the default to a timeout. 
I'm -1 to this changeset for now, but happy to be convinced otherwise - I'm probably being too cautious.

a-dubs:
@holmanb Yeah good point. I should really _never_ use "never" like that. 

But yeah I'm honestly okay with keeping it cloud-specific for now. For oracle, I'm quite certain a small timeout is reasonable so we'll just put the timeout there - I believe @TheRealFalcon is working on a PR that will add short timeout (5s) to the oracle datasource's calls to the IMDS via read_url().

Also, I am not sure I understand how this timeout would matter in terms of boot speed of the instance it's running on though? Since this is called against the datasource, it doesn't matter how long the instance would take to boot because once it boots the datasource should respond in a normally timely manner, right? Or I am misunderstanding? Thanks!

holmanb:
> Also, I am not sure I understand how this timeout would matter in terms of boot speed of the instance it's running on though? Since this is called against the datasource, it doesn't matter how long the instance would take to boot because once it boots the datasource should respond in a normally timely manner, right? 

I wouldn't expect the timeout to negatively affect boot speed.

> Or I am misunderstanding? Thanks!

I wasn't trying to raise concerns about boot speed. My concerns were actually about the reliability (and changing behavior) for systems that boot slowly and therefore depend on cloud-init not having a timeout.


Question:

@a-dubs When systems don't use ds-identify, in theory they have to iterate through datasources until they find one that responds. Wouldn't oracle's "no timeout" have broken this design? Or am I missing something? In theory this doesn't really matter too much anymore, since most (all?) supported init systems now have support for ds-identify, but I'm curious if this would have broken that model.

a-dubs:
> Question:
@a-dubs When systems don't use ds-identify, in theory they have to iterate through datasources until they find one that responds. Wouldn't oracle's "no timeout" have broken this design? Or am I missing something? In theory this doesn't really matter too much anymore, since most (all?) supported init systems now have support for ds-identify, but I'm curious if this would have broken that model.

That's a good question. For Oracle specifically, the  readurl() call against the imds that has no timeout currently, is done once Oracle is determined to be the datasource. Let's sync up on this @holmanb because it sounds like we both need to just clarify a few things. I know I'd benefit from it at least.

TheRealFalcon:
Good points @holmanb . +1 to closing this PR.

holmanb:
>> Wouldn't oracle's "no timeout" have broken this design? Or am I missing something?
>
> DataSourceOracle has a custom ds_detect() which eventually checks the [chassis-asset-tag](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceOracle.py#L341) . If this returns False, _get_data() isn't called.

I should have checked the code before asking. That explains it.

TheRealFalcon:
Closing for the reasons mentioned by holmanb . If there's something left to be done here, feel free to re-open.

PR #5011, opened 2024-03-05
release: sync changelog and version

holmanb:
## Proposed Commit Message
```
release: sync changelog and version
```

## Additional Context
The release should have been to main originally, but it's too late to put the tagged commit on main, so just sync the changelog.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,255 @@
+24.1
+ - fix: Don't warn on vendor directory (#4986)
+ - apt: kill spawned keyboxd after gpg cmd interaction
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
Shoot...I think `git describe` is going to continue to show as `23.4-<number>` rather than `24.1`.

I know the Ubuntu downstream used to depend on it for generating version numbers, so I'm curious if any other downstreams do. If so...I'm not sure there's an easy way to fix it.

holmanb:
> I'm not sure there's an easy way to fix it.

same, unless we want to rewrite the history of main

Issue #4973, opened 2024-02-28
test_net test fail in the calico network env


xiaoge1001:
# Bug report
RuntimeError: duplicate mac found! both 'calixxx' and 'caliyyy' have mac 'ee:ee:ee:ee:ee:ee'.

## Steps to reproduce the problem

pytest tests/unittests/test_net.py
 
## Environment details
- Cloud-init version:23.4.1
- Operating System Distribution:openeuler

## cloud-init logs

```
2024-02-27 10:05:36                 msg = "duplicate mac found! both '%s' and '%s' have mac '%s'." % (
2024-02-27 10:05:36                     name,
2024-02-27 10:05:36                     ret[mac],
2024-02-27 10:05:36                     mac,
2024-02-27 10:05:36                 )
2024-02-27 10:05:36 >               raise RuntimeError(msg)
2024-02-27 10:05:36 E               RuntimeError: duplicate mac found! both 'caliaadab05bca1' and 'cali7a099f8b737' have mac 'ee:ee:ee:ee:ee:ee'.
2024-02-27 10:05:36 
2024-02-27 10:05:36 cloudinit/net/__init__.py:987: RuntimeError
```

xiaoge1001:
I plan to use the following fix (verified to work), but don't know if it's correct.
fix patch: https://gitee.com/tong_1001/cloud-init/blob/master/Ignoring-duplicate-macs-for-calico-network.patch

catmsred:
Thank you for reporting this bug!  23.4.1 has been superceded by 23.4.4 -- are you able to reproduce this issue in 23.4.4 as well?

TheRealFalcon:
> pytest tests/unittests/test_net.py

`test_net.py` contains 223 tests. Can you provide the exact test or tests that failed?

xiaoge1001:
> > pytest tests/unittests/test_net.py
> 
> `test_net.py` contains 223 tests. Can you provide the exact test or tests that failed?

cloud-init-23.4.1:
https://gitee.com/src-openeuler/cloud-init/issues/I93ZJT?from=project-issue    （See the attachment.）

xiaoge1001:
I will use the latest version for the past two days to test it as well.

blackboxsw:
@xiaoge1001 I expect https://github.com/canonical/cloud-init/pull/5012 may solve the build issues you are seeing. If you get a chance to test #5012 in your environment it will help give us a signal as to whether this resolves the issue you filed.

[...]

PR #5012, opened 2024-03-05
bug(tests): mock reads of host's /sys/class/net via get_sys_class_path

blackboxsw:
Create pytest fixture `disable_sysfs_net` to mock cloudinit.net.get_sys_class_path and 
avoid reading /sys/class/net on the host system. This fixture is only applied sparingly to modules that interact with `cloudinit.net.get_sys_class_path` because the cost globally setting  `autouse=True` on this fixture adds about 10 seconds to the test runtime. In modules where we expect most tests to be consuming `get_sys_class_path` we autouse at the module level, otherwise leverage the `disable_sysfs_net` fixture in individual test methods where needed.

This is an answer to GH #4973 where unitests are being run in non-production environments which exhibit duplicate MAC addresses causing many tests to fail.

The plan is it squash merge these separate commits after review.

## Proposed Commit Message
```
bug(tests): mock reads of host's /sys/class/net via get_sys_class_path

Avoid leaking reads to the underlying host's /sys/class/net files.
Some test environments contain virtual network hardware and
configuration such as Calico network devices which provide
duplicated MAC addresses for each device in /sys/class/net. This
results in errors from unittests calling cloudinit.net.get_interfaces.

Provide a disable_sysfs_net` fixture and use it in net-related
modules or functions to avoid unrelated test failures due to
environmental differences.

Intentionally avoid a global `autouse=True` for this fixtures as adds
around 10 seconds to test runs if used for every single unittest.

This should also resolve test platform failures seen in openeuler.

Fixes GH-4973
```

## Test Steps
```
# ensure all tests except for 2 are mocked and prevent access to host's /sys/class/net/
sed -i 's/return SYS_CLASS_NET/raise AssertionError("should not get here")/' cloudinit/net/__init__.py
# expect 2 failures
tox -e py3
```

## Checklist

## Merge type

### DIFF
--- a/tests/unittests/conftest.py
+++ b/tests/unittests/conftest.py
@@ -63,6 +63,26 @@ def fake_filesystem(mocker, tmpdir):
     yield str(tmpdir)
 
 
+@pytest.fixture(scope="session", autouse=True)
+def disable_sysfs_net(request, tmpdir_factory):
+    """Avoid tests which read the undertying host's /syc/class/net.
+
+    To allow unobscured reads of /sys/class/net on the host we can
+    parametrize the fixture with:
+
+    @pytest.mark.parametrize("disable_sysfs_net", [False], indirect=True)
+    """
+    if hasattr(request, "param") and getattr(request, "param") is False:
+        # Test disabled this fixture, perform no mocks.
+        yield
+        return
+    mock_sysfs = f"{tmpdir_factory.mktemp('sysfs')}/"
+    with mock.patch(
+        "cloudinit.net.get_sys_class_path", return_value=mock_sysfs
+    ):
+        yield mock_sysfs
+
+
 @pytest.fixture(autouse=True)
 def disable_dns_lookup(request):
     if "allow_dns_lookup" in request.keywords:
--- a/tests/unittests/net/test_dhcp.py
+++ b/tests/unittests/net/test_dhcp.py
@@ -374,15 +374,13 @@ class TestDHCPParseStaticRoutes(CiTestCase):
         )
 
 
-class TestDHCPDiscoveryClean(CiTestCase):
-    with_logs = True
-
+class TestDHCPDiscoveryClean:
     @mock.patch("cloudinit.distros.net.find_fallback_nic", return_value="eth9")
     @mock.patch("cloudinit.net.dhcp.os.remove")
     @mock.patch("cloudinit.net.dhcp.subp.subp")
     @mock.patch("cloudinit.net.dhcp.subp.which")
     def test_dhcpcd_exits_with_error(
-        self, m_which, m_subp, m_remove, m_fallback
+        self, m_which, m_subp, m_remove, m_fallback, caplog
     ):
         """Log and do nothing when nic is absent and no fallback is found."""
         m_subp.side_effect = [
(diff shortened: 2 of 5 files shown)


TheRealFalcon (commented):
> the cost globally setting autouse=True on this fixture adds about 10 seconds to the test runtime

A session-scoped fixture only gets applied once and didn't affect runtime at all in my tests. Since we never write anything to `/sys/class/net`, we shouldn't need to worry about each test getting a unique path, correct? Do you foresee any problems with using the same temporary path to be shared across all of the tests? If not, I think this PR could be simplified to something like this:
```diff
diff --git a/tests/unittests/conftest.py b/tests/unittests/conftest.py
index c99f5d860..73b99d758 100644
--- a/tests/unittests/conftest.py
+++ b/tests/unittests/conftest.py
@@ -52,6 +52,16 @@ FS_FUNCS = {
 }
 
 
+@pytest.fixture(scope="session", autouse=True)
+def disable_sysfs_net(tmpdir_factory):
+    """Disables sysfs net directory"""
+    mock_sysfs = f"{tmpdir_factory.mktemp('sysfs')}/"
+    with mock.patch(
+        "cloudinit.net.get_sys_class_path", return_value=mock_sysfs
+    ):
+        yield mock_sysfs
+
+
 @pytest.fixture
 def fake_filesystem(mocker, tmpdir):
     """Mocks fs functions to operate under `tmpdir`"""
diff --git a/tests/unittests/net/test_init.py b/tests/unittests/net/test_init.py
index 60a44186e..c4be1609e 100644
--- a/tests/unittests/net/test_init.py
+++ b/tests/unittests/net/test_init.py
@@ -26,13 +26,13 @@ class TestSysDevPath(CiTestCase):
         """sys_dev_path returns a path under SYS_CLASS_NET for a device."""
         dev = "something"
         path = "attribute"
-        expected = net.SYS_CLASS_NET + dev + "/" + path
+        expected = net.get_sys_class_path() + dev + "/" + path
         self.assertEqual(expected, net.sys_dev_path(dev, path))
 
     def test_sys_dev_path_without_path(self):
         """When path param isn't provided it defaults to empty string."""
         dev = "something"
-        expected = net.SYS_CLASS_NET + dev + "/"
+        expected = net.get_sys_class_path() + dev + "/"
         self.assertEqual(expected, net.sys_dev_path(dev))
 
```

tests/unittests/conftest.py
```diff
@@ -63,6 +63,19 @@ def fake_filesystem(mocker, tmpdir):
     yield str(tmpdir)
 
 
+@pytest.fixture(scope="session", autouse=True)
+def disable_sysfs_net(request, tmpdir_factory):
+    if hasattr(request, "param") and getattr(request, "param") is False:
```
TheRealFalcon, line 68: Nit: The `hasattr` call isn't needed, right? `getattr` will return `None` if the argument doesn't exist?

TheRealFalcon (approved):
I left some inline nits, but no code changes needed.

Also...Wooooooooooah! I knew about indirect parametrization, but I didn't know you could parametrize a session-scoped fixture. If we refactor `disable_subp_usage` and `disable_dns_lookup` to work the same, we could potentially save a lot of test setup time.

tests/unittests/conftest.py
blackboxsw: True story.

blackboxsw:
> I left some inline nits, but no code changes needed.
> 
> Also...Wooooooooooah! I knew about indirect parametrization, but I didn't know you could parametrize a session-scoped fixture. If we refactor `disable_subp_usage` and `disable_dns_lookup` to work the same, we could potentially save a lot of test setup time.

Yep, I was having fun looking over [pytest docs on this this morning](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest.Metafunc.parametrize).  And agreed I thought the same for disable_subp/dns and dropping some of those `allow` markers from tox.ini

holmanb:
>>    Also...Wooooooooooah! I knew about indirect parametrization, but I didn't know you could parametrize a session-scoped fixture. If we refactor disable_subp_usage and disable_dns_lookup to work the same, we could potentially save a lot of test setup time.
>
> Yep, I was having fun looking over [pytest docs on this this morning](https://docs.pytest.org/en/7.3.x/reference/reference.html#pytest.Metafunc.parametrize). And agreed I thought the same for disable_subp/dns and dropping some of those allow markers from tox.ini

And additionally we can add some other common-sense mocks that would otherwise be "too expensive to consider".

Nice find!

PR #5016, opened 2024-03-06
fix: use correct disabled path in ds-identify

TheRealFalcon:
## Proposed Commit Message

```
fix: use correct disabled path in ds-identify
```

## Merge type


holmanb (approved):
Geez, I thought we had a 24.1 regression.

I'm glad it's just the logged message.

PR #5021, opened 2024-03-07
bug(vmware): initialize new DataSourceVMware attributes at unpickle

blackboxsw:
## Proposed Commit Message
```
bug(vmware): initialize new DataSourceVMware attributes at unpickle

When a datasource has already been cached from a previous boot, the the datasource is deserialized from the cached object instead of running __init__ method. Across a cloud-init package upgrade if new instance attributes have been initialized in __init__, we need to also initialize them in the class  _unpickle method to avoid AttributeErrors for any instance methods referencing new attributes.

Fix AttributeError on rpctool and rpctool_fn

LP: #2056439
```

## Checklist

## Merge type


tests/unittests/test_upgrade.py
```diff
@@ -28,6 +30,112 @@
...
+            "dsmode",
+            "metadata_address",
+            "retries",
+            "timeout",
+            "userdata_address",
+            "wait_retry",
+        },
+        "IBMCloud": {"source", "_network_config", "network_json", "platform"},
+        "RbxCloud": {"seed"},
+        # TODO(Fix Scaleway, max_wait, userdata_url, metadata_urls, has_ipv4)
```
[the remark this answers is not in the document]
TheRealFalcon: Adding them won't hurt anything, right? I think we might as well go ahead and do it. We'd be sad if we had to add them later.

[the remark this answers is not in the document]
blackboxsw: +1

TheRealFalcon (commented):
Good way to approach testing this! Generally LGTM, but there's some failing CI jobs.

tests/unittests/test_upgrade.py
```diff
@@ -39,7 +144,67 @@ def previous_obj_pkl(self, request):
...
+            mocker.patch("cloudinit.util.is_container", return_value=False)
+            mocker.patch("cloudinit.dmi.read_dmi_data", return_value="")
+            ds = ds_class(sys_cfg={}, distro=MockDistro(), paths=paths)
+            init_attrs = (
+                set(ds.__dict__)
+                - class_attrs
+                - common_instance_attrs
+                - self.ds_expected_unpickle_attrs.get(ds_class.dsname, set())
+            )
+            # if ds.dsname == "VMware":
```
holmanb, line 195: we can remove this line

tests/unittests/test_upgrade.py
```diff
@@ -39,7 +144,67 @@ def previous_obj_pkl(self, request):
...
+    )
+    def test_all_ds_init_vs_unpickle_attributes(self, mocker, paths, tmpdir):
+        """Unpickle resets any instance attributes created in __init__
+
+        This test asserts that deserialization of a datasource cache
+        does proper initialization of any 'new' instance attributes
+        created as a side-effect of the __init__ method.
+
+        Without proper _unpickle coverage for newly introduced attributes,
+        the new deserialized instance will hit AttributeErrors at runtime.
```
holmanb, line 162: This is a great!

nit: This explains what the test validates, but not how. Usually that's not a problem, but since this is innovative and not typical of cloud-init test code, a sentence describing how it asserts the proper initialization would go along ways. A comment or two inline adding context wouldn't hurt either.

holmanb (commented):
Nice tests and fixes @blackboxsw!

I left a couple of nits inline, feel free to take them or leave them.

TheRealFalcon:
Heads up, my merge caused a conflict here.

tests/unittests/test_upgrade.py
```diff
@@ -39,7 +156,88 @@ def previous_obj_pkl(self, request):
...
+        """
+        # Load all cloud-init DataSource classes
+        for ds_class in sources.list_sources(
+            settings.CFG_BUILTIN["datasource_list"],
+            [sources.DEP_FILESYSTEM],
```
holmanb, line 179: are only DEP_FILESYSTEM datasources unpickled?

holmanb (approved):
With the following diff on this branch:
```
index bd654cc0e..9e95b6a4f 100644
@@ -339,6 +339,7 @@ class DataSourceAzure(sources.DataSource):
         self._reported_ready_marker_file = os.path.join(
             paths.cloud_dir, "data", "reported_ready"
         )
+        self.new_attr = True
 
@@ -354,7 +355,7 @@ class DataSourceAzure(sources.DataSource):
 
     def __str__(self):
         root = sources.DataSource.__str__(self)
-        return "%s [seed=%s]" % (root, self.seed)
+        return "%s [seed=%s] attr: %s" % (root, self.seed, self.new_attr)
 
     def _get_subplatform(self):
         """Return the subplatform metadata source details."""
diff --git a/cloudinit/sources/DataSourceNone.py b/cloudinit/sources/DataSourceNone.py
index 40e89ae8a..dbcc9935e 100644
--- a/cloudinit/sources/DataSourceNone.py
+++ b/cloudinit/sources/DataSourceNone.py
@@ -15,8 +15,11 @@ class DataSourceNone(sources.DataSource):
         sources.DataSource.__init__(self, sys_cfg, distro, paths, ud_proc)
         self.metadata = {}
         self.userdata_raw = ""
+        self.new_attr = True
 
     def _get_data(self):
+        if self.new_attr:
+            pass
         # If the datasource config has any provided 'fallback'
         # userdata or metadata, use it...
         if "userdata_raw" in self.ds_cfg:
```

I would expect to see two new failing tests, right?

holmanb (changes requested):
oops, that was supposed to be just a comment, so dismissing the approval with this "request changes"

blackboxsw:
> oops, that was supposed to be just a comment, so dismissing the approval with this "request changes"

Great review, I had overlooked network-stage classes as you suggested. I've also parametrized the test now and added a parent class attribute check  if we are a datasource subclass with the same dsname to better source common parent attributes as well.

holmanb (approved):
I like this a lot. Thanks @blackboxsw!

Issue #5019, opened 2024-03-07
ubuntu_advantage not valid as module key in base config after rename to ubuntu_pro


aciba90:
# Bug report

After [renaming cc_ubuntu_advantage to cc_ubuntu_pro](https://github.com/canonical/cloud-init/pull/4865), `ubuntu_advantage` is not [a valid module key in the base config](https://docs.cloud-init.io/en/latest/reference/base_config_reference.html#module-keys).

Upgrade-path: while `/etc/cloud/cloud.cfg` gets regenerated at upgrade time, if admins or image creators drop a config file in `/etc/cloud/cloud.cfg.d/` containing a reference to `ubuntu_advantage` in any of `cloud_{init,config,final}_modules` keys, then cloud-init is not able to find and execute `cc_ubuntu_pro`, if re-running for any reason.

Other situation where this could happen is if other downstreams than Ubuntu have references to `ubuntu_advantage` under `/etc/cloud/cloud.cfg.d/`, this arguably not a supported thing.

Expected behavior: `ubuntu_advantage` working properly as module key in base config.

## Steps to reproduce the problem

```
cat >/tmp/ud <<EOF
> #cloud-config
> ubuntu_advantage:
>   enable: []
> EOF

lxc launch ubuntu-daily:j ua -c user.user-data="$(cat /tmp/ud)"
lxc shell ua

root@ua:~# cloud-init --version
/usr/bin/cloud-init 23.4.4-0ubuntu0~22.04.1

root@ua:~# cat >/etc/cloud/cloud.cfg.d/50-other-config.cfg << EOF
# Custom config modules
cloud_config_modules:                                                                                    
  - ssh_import_id                                                                                        
  - keyboard                                  
  - locale                                                                                               
  - set_passwords                                                                                        
  - grub_dpkg     
  - apt_configure
  - ubuntu_advantage
EOF

add-apt-repository ppa:cloud-init-dev/daily
apt update && apt upgrade -y

root@ua:~# cloud-init --version
/usr/bin/cloud-init 23.4.daily-202403062033-67685d65f~ubuntu22.04.1

cloud-init clean --logs --reboot

root@ua:~# cloud-init status --long
status: done
extended_status: degraded done
boot_status_code: enabled-by-generator
last_update: Thu, 07 Mar 2024 09:43:51 +0000
detail: DataSourceLXD
errors: []
recoverable_errors:
WARNING:
[...]

PR #5030, opened 2024-03-08
pro: honor but warn on custom ubuntu_advantage in /etc/cloud/cloud.cfg

blackboxsw:
Add RENAMED_MODULES to allow support for custom /etc/cloud/cloud.cfg.d/* which may still contain the old ubuntu_advantage module name instead of ubuntu_pro.

While the typical package upgrade path would replace ubuntu_avantage with ubuntu_pro in /etc/cloud/cloud.cfg. It's possible a customer has a customized /etc/cloud/cloud.cfg that they do not wish to update with the version delivered by the cloud-init upstream package.

In cases where ubuntu_advantage is listed in system config, allow cloud-init to WARN about the rename, but still load and run ubuntu_pro.

Fixes GH-5019

## Proposed Commit Message

```
<type>(optional scope): <summary>  # no more than 72 characters

made if the summary line is insufficient.  This should be wrapped at
72 characters.

```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/modules.py
+++ b/cloudinit/config/modules.py
@@ -35,6 +35,10 @@ REMOVED_MODULES = [
     "cc_rightscale_userdata",  # Removed in 24.1
 ]
 
+RENAMED_MODULES = {
+    "cc_ubuntu_advantage": "cc_ubuntu_pro",  # Renamed 24.1
+}
+
 
 class ModuleDetails(NamedTuple):
     module: ModuleType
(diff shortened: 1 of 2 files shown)


cloudinit/config/modules.py
```diff
@@ -198,6 +202,14 @@ def _fixup_modules(self, raw_mods) -> List[ModuleDetails]:
                 # Misconfigured in /etc/cloud/cloud.cfg. Reset so cc_* module
                 # default meta attribute "frequency" value is used.
                 freq = None
+            if mod_name in RENAMED_MODULES:
+                LOG.warning(
```
holmanb, line 206: think this should be a `LOG.deprecated()`?

blackboxsw: Good thought. Done

holmanb (approved):
One nit, otherwise LGTM. Thanks!

tests/integration_tests/modules/test_ubuntu_pro.py
```diff
@@ -140,10 +140,23 @@ def get_services_status(client: IntegrationInstance) -> dict:
 )
 class TestUbuntuAdvantage:
     @pytest.mark.user_data(ATTACH_FALLBACK.format(token=CLOUD_INIT_UA_TOKEN))
-    def test_valid_token(self, client: IntegrationInstance):
+    def test_valid_token(self, client: IntegrationInstance, tmpdir):
```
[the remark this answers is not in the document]
blackboxsw: thanks. This  was cruft from an earlier integration test approach on this PR where I created the local cloud.cfg.d  file and tried pushing it to target before going with the inline sed approach. dropping it.

Issue #4987, opened 2024-03-01
[enhancement]: don't wait for instance deletion in integration tests


holmanb:
# Enhancement

Our integration tests synchronously wait for [instance destruction](https://github.com/canonical/cloud-init/blob/12126f57412f9b3916fbc83ec73289e1157bab34/tests/integration_tests/instances.py#L278) on every cloud.

On a [recent integration test run of GCE](https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-noble-gce/52/consoleText), I see 90 warnings of instance termination timeouts:
```
grep "WARNING   pycloudlib.instance:instance.py:178 Instance not terminated after 30 seconds."  text  | wc -l
90
```

Not waiting for instance deletion would cut the total integration test runtime on GCE by 45 minutes, (likely quite a bit more) per test run. Since the whole test run takes approximately 4 hours, this would cut the runtime by 25-50%.

LXD may require timeouts still, to avoid system saturation, but the clouds likely don't.

I propose that we either don't wait for instance deletion in integration tests, or we wait synchronously in a background thread, on a per-cloud basis (for LXD).

PR #5037, opened 2024-03-09
tests: Don't wait for GCE instance teardown

holmanb:
## Proposed Commit Message
```
tests: Don't wait for GCE instance teardown

Fixes GH-4987
```

## Additional Context
It feels like an abstraction would be better here for cloud-specific instance behavior customizations, but I think that would be a bigger refactor.

Fixes https://github.com/canonical/cloud-init/issues/4987


blackboxsw (approved):
I'm +1 on this improvement. We need to make sure we are on monitoring the gce cleanup jobs to ensure they are lot leaving around active VMs. So let's watch jenkins next week once this lands

PR #5041, opened 2024-03-11
packages/debian: remove dependency on isc-dhcp-client

cjp256:
With support for new clients, remove hard dependency on dhclient.


TheRealFalcon (approved):
LGTM!

PR #1479, opened 2022-05-26
schema: write_files defaults, versions $ref full URL and add vscode

blackboxsw:
## Proposed Commit Message
```
schema: write_files defaults, versions $ref full URL and add vscode

* Set write_files.permissions and content defaults the correct "string"
   data type. This aids in vscode auto-completion which injects default
   values for defined cloud-config keys.

* Make the $ref URI a full URL to our schema-cloud-config-v1.json
   JSON schema Draft-04 does not support relative $ref URLs which
   prevents validators like https://www.jsonschemavalidator.net/
   from parsing our versions schema with the error:
      Error when resolving schema reference './schema-cloud...json'.

* Add vscode exstentions for validatating #cloud-config YAML
    Developers using the cloud-init repo are likely to also
    want to create #cloud-config YAML files for development. Provide
    the suggested YAML extension in .vscode/extensions.json.
    
    This YAML extension will prompt vscode developers with the following
    message upon opening the cloud-init repo:
    
      'YAML' extension is recommended for this repository. Do you want
      to install?
    
    This RedHat YAML extension sources public schemas from
    https://schemastore.org and will treat the following file formats as
    as #cloud-config files:

      cloud-init.yaml
      cloudinit.yaml
      *.cloud-init.yaml
      *.cloudinit.yaml
  
    Any filenames matching those formats will get automatic validation
    against cloud-config schema and auto-complete known #cloud-config
    keys.
```

## Test Steps
1. [install vscode](https://code.visualstudio.com/download)
2. Start vscode in the cloud-init directory with this PR
```bash
cd cloud-init
code
```
3. Open the new cloud-init project in vscode
4. See and click "Blue Install button in lower right corner for the recommended YAML extension
5. Use the file navigator to open the cloud-init directory 
6. Create a new file named cloud-config.yaml    
7. Notice that a header link to "schema-cloud-config-v1.json" is rendered at the top you the cloud-config.yaml file
8. start typing any characters and see drop down menus of auto-completion vscode intellisense suggestions for known cloud-config schema.

## Checklist:

 - [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/topics/contributing.html)
 - [ ] I have updated or added any unit tests accordingly
 - [ ] I have updated or added any documentation accordingly

PR #5049, opened 2024-03-12
docs: Explain exit code 2

holmanb:
## Proposed Commit Message
```
doc: Explain breaking change in status code
```

## Merge type

### DIFF
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -55,7 +55,7 @@ JSON schema
 ^^^^^^^^^^^
 
 Cloud-init has defined a JSON schema which fully documents the user-data
-cloud-config. This jsonschema may be used in several different ways:
+cloud-config. This JSON schema may be used in several different ways:
 
 Text editor integration
 """""""""""""""""""""""
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -80,10 +80,10 @@ configuration.
 Return codes
 ^^^^^^^^^^^^
 
-Cloud-init historically used two return codes from the :code:`cloud-init status`
-subcommand: 0 to indicate success and 1 to indicate failure. These return codes
-lacked nuance. Return code 0 (success) included the in-between when something
-went wrong, but cloud-init was able to finish.
+Cloud-init historically used two return codes from the
+:code:`cloud-init status` subcommand: 0 to indicate success and 1 to indicate
+failure. These return codes lacked nuance. Return code 0 (success) included
+the in-between when something went wrong, but cloud-init was able to finish.
 
 Many users of cloud-init run :code:`cloud-init status --wait` and expect that
 when complete, cloud-init has finished. Since cloud-init is not guaranteed to
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -61,7 +61,7 @@ Text editor integration
 """""""""""""""""""""""
 
 Thanks to `yaml-language-server`_, cloud-init's jsonschema may be
-used for yaml syntax checking, warnings when invalid keys are used, and
+used for YAML syntax checking, warnings when invalid keys are used, and
 autocompletion. Several different text editors are capable of this.
 See this `blog post on configuring this for neovim`_, or for VScode one
 can install the `extension`_ and then a file named ``cloud-config.yaml``
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -64,7 +64,7 @@ Thanks to `yaml-language-server`_, cloud-init's jsonschema may be
 used for yaml syntax checking, warnings when invalid keys are used, and
 autocompletion. Several different text editors are capable of this.
 See this `blog post on configuring this for neovim`_, or for VScode one
-may just install the `extension`_ and then a file named cloud-config.yaml
+can install the `extension`_ and then a file named ``cloud-config.yaml``
 will automatically use cloud-init's JSON schema.
 
 
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -60,7 +60,7 @@ cloud-config. This jsonschema may be used in several different ways:
 Text editor integration
 """""""""""""""""""""""
 
-Thanks to `yaml-language-server`_, cloud-init's jsonschema may be
+Thanks to `yaml-language-server`_, cloud-init's JSON schema may be
 used for YAML syntax checking, warnings when invalid keys are used, and
 autocompletion. Several different text editors are capable of this.
 See this `blog post on configuring this for neovim`_, or for VScode one
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -36,9 +36,12 @@ Pain points
 3) Cloud-init's reported status was only capable of telling the user
    whether cloud-init crashed. Cloud-init would report a status of
    "done" in the following cases:
+
    * a user's configuration was invalid
-   * if the operating system or cloud environment experienced some error that prevented cloud-init from configuring the instance
-   * if cloud-init internally experienced an error - all of these previously reported a status of "done".
+   * if the operating system or cloud environment experienced some error that
+     prevented cloud-init from configuring the instance
+   * if cloud-init internally experienced an error - all of these previously
+     reported a status of "done".
 
 .. _return_codes_improvements:
 
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -65,7 +65,7 @@ used for yaml syntax checking, warnings when invalid keys are used, and
 autocompletion. Several different text editors are capable of this.
 See this `blog post on configuring this for neovim`_, or for VScode one
 may just install the `extension`_ and then a file named cloud-config.yaml
-will automatically use cloud-init's jsonschema.
+will automatically use cloud-init's JSON schema.
 
 
 Cloud-init schema subcommand
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -80,7 +80,7 @@ configuration.
 Return codes
 ^^^^^^^^^^^^
 
-Cloud-init historically used two return codes from the `cloud-init status`
+Cloud-init historically used two return codes from the :code:`cloud-init status`
 subcommand: 0 to indicate success and 1 to indicate failure. These return codes
 lacked nuance. Return code 0 (success) included the in-between when something
 went wrong, but cloud-init was able to finish.
--- a/doc/rtd/explanation/return_codes.rst
+++ b/doc/rtd/explanation/return_codes.rst
@@ -54,7 +54,7 @@ points described above.
 JSON schema
 ^^^^^^^^^^^
 
-Cloud-init has defined a jsonschema which fully documents the user-data
+Cloud-init has defined a JSON schema which fully documents the user-data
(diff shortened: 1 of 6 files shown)


doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+
+1) Invalid configurations were historically "ignored".
+
+2) Log verbosity is unfriendly to end users that may not know
+  what to look for. Verbose logs means users often ignore real
+  errors.
+
+3) Cloud-init's reported status was only capable of telling the user
+  whether cloud-init crashed. If a user's configuration was invalid,
+  or if the operating system or cloud environment experienced some
```
TheRealFalcon, line 43: This sentence feels a bit long. Maybe convert to bullets or somehow separate into two sentences.

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Cloud-init has defined a jsonschema which fully documents the user-data
+cloud-config. This jsonschema may be used in several different ways:
+
+Text editor integration
+"""""""""""""""""""""""
+
+Thanks to `yamlls<yaml-language-server>`, cloud-init's jsonschema may be
+used for yaml syntax checking, warnings when invalid keys are used, and
+even autocompletion. Several different text editors are capable of this.
```
TheRealFalcon, line 67:
```suggestion
autocompletion. Several different text editors are capable of this.
```

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+See this `blog post on configuring this for neovim`_, or 
+
+# TODO someone who knows about vscode add some details
```
TheRealFalcon, line 70: Install this extension: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml . Create a file called `cloud-config.yaml` and it should work automatically.

Also see test steps here: https://github.com/canonical/cloud-init/pull/1479

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Cloud-init schema subcommand
+""""""""""""""""""""""""""""
+
+The cloud-init package includes a cloud-init subcommand, `cloud-init schema`
```
TheRealFalcon, line 75: Should link to https://cloudinit.readthedocs.io/en/latest/howto/debug_user_data.html

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+which uses the schema to validate either the configuration passed to the
+instance that you are running the command on, or to validate an arbitrary
+text file containing a configuration.
+
+Nuanced Status Reporting
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In release 23.4, cloud-init introduced extended status reporting to users.
```
TheRealFalcon, line 83: Should link to https://cloudinit.readthedocs.io/en/latest/howto/status.html#cloud-init-s-extended-status

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+This makes it easier for users to identify problems that didn't result in a
+crash.
+
+Extended status and recoverable errors
```
TheRealFalcon, line 87: I think this section can be merged with the one above it

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+Cloud-init historically used two return codes from the `cloud-init status`
+subcommand: 0 to indicate success and 1 to indicate failure. These return codes
+lacked nuance. Return code 0 (success) included the in-between when something
+went wrong, but cloud-init was able to finish.
+
+Many users of cloud-init run :code:`cloud-init status --wait` and expect that
+when complete, cloud-init has finished. Since cloud-init is not guaranteed to
+succeed, users should also be check the return code of this command.
+
+A backwards-incompatible change was recently introduced: errors that did not
```
TheRealFalcon, line 109: Since we could neglect to update this for a while, let's not mention recently and instead just say something like "As of 23.4...".

Also, I don't remember the exact page, but I remember you expressing frustration about an older page talking about this history of how things were years ago instead of describing how it works. I could see that quickly being relevant here and I'm not sure the two paragraphs above this one are necessary.

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+crash cloud-init now have an exit code of 2. After this change, exit 1 still
+means that cloud-init crashed, but exit code 0 now more correctly means that
+cloud-init succeeded. Anyone that previously checked for exit code 0 should
+probably update their assumptions in one of the following two ways:
+
+Users that wish to take advantage of cloud-init's new error reporting
+capabilities should check for exit code of 2 from :code:`cloud-init status`.
+An example of this:
+
+.. code-block:: python
```
TheRealFalcon, line 119: Shell might be a better example here just as that's the general the lowest common denominator for scripting.

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    if 1 == completed.return_code:
+        # cloud-init completely failed
+        logger.error("cloud-init crashed, all bets are off!")
+        exit(1)
+
+    # cloud-init might have failed, but this code ignores that possibility
+    # in preference of backwards compatibility
+
+Some distros (such as RHEL and Ubuntu) are likely to patch out this
```
TheRealFalcon, line 163: Not sure I'd call out specific distros here.

Also, this paragraph feels a little weird to call out here. We don't generally point out every possible breaking change that has been patched out in various downstreams. Not sure we need anything, but if so I'd reduce it to a note with something along the lines of "Distributions have likely removed this behavior for stable releases".

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+Why did `cloud-init status` start returning exit code 2?
+========================================================
+
+Cloud-init introduced a new error code in 23.4.
+This page describes the purpose of this change and gives
+some context for why this change was made.
+
+.. _return_codes_history:
+
+History
```
TheRealFalcon, line 12:
```suggestion
Background
```
Only because history is boring :smile:

doc/rtd/howto/return_codes.rst
```diff
@@ -0,0 +1,172 @@
...
+-------
+
+In the early days of cloud-init, the purpose of cloud-init
+was simple: provision ssh keys and a user and exit. Cloud-init
+only supported a couple of clouds, and things were simple.
+Since cloud-init provides access to cloud instances, the
```
TheRealFalcon, line 18: I don't think we need the sentences before this. I don't care about the early days of cloud-init, more just the reasoning behind why things work the way they do.

TheRealFalcon (changes requested):
I left some comments inline. Docs don't build on this branch and after fixing one issue and running into another, I'll leave it to you to fix that. 

More generally, the page feels very disjoint. Much of the history and description of various things we've added to report warnings aren't super relevant to the question being asked in the heading. I think it is useful information though, and I'm wondering if this would make more sense as a "Dealing with problems" explanation. The `why exit 2` can just be a subset of a larger explanation.

We should probably also update https://cloudinit.readthedocs.io/en/latest/reference/cli.html#status

doc/rtd/howto/return_codes.rst
holmanb: I was originally going to use shell, but it doesn't lend well to making use of the machine-readable output available in cloud-init, which real languages would want to make use of, which is why I went with Python.

doc/rtd/howto/return_codes.rst
holmanb: Thanks!

holmanb:
@TheRealFalcon Ready for re-review, I think I've addressed everything except the preference to shell. Shell isn't great for handling machine-readable output without jq, and I think we're in agreement that avoiding that if possible. Python is pretty readable for most people, but I can add another language example if that works (Go / Rust / ???). If you feel strongly about using shell I'll do that, but my goal with those examples was to demonstrate the proper way of handling `exit 2`, which in my opinion is logging the errors raised from cloud-init at a minimum.

holmanb:
> @TheRealFalcon Ready for re-review, I think I've addressed everything except the preference to shell. Shell isn't great for handling machine-readable output without jq, and I think we're in agreement that avoiding that if possible. Python is pretty readable for most people, but I can add another language example if that works (Go / Rust / ???). If you feel strongly about using shell I'll do that, but my goal with those examples was to demonstrate the proper way of handling `exit 2`, which in my opinion is logging the errors raised from cloud-init at a minimum.

I suppose I could use an example that just prints _all_ of the output from `cloud-init status --long` - thoughts?

TheRealFalcon (approved):
LGTM! Should we get a review from @s-makin before merging?

doc/rtd/explanation/return_codes.rst
```diff
@@ -0,0 +1,150 @@
...
+^^^^^^^^^^
+
+Cloud-init has defined a jsonschema which fully documents the user-data
+cloud-config. This jsonschema may be used in several different ways:
+
+Text editor integration
+"""""""""""""""""""""""
+
+Thanks to `yaml-language-server`_, cloud-init's jsonschema may be
+used for yaml syntax checking, warnings when invalid keys are used, and
```
s-makin, line 64:
```suggestion
used for YAML syntax checking, warnings when invalid keys are used, and
```

doc/rtd/explanation/return_codes.rst
```diff
@@ -0,0 +1,150 @@
...
+autocompletion. Several different text editors are capable of this.
+See this `blog post on configuring this for neovim`_, or for VScode one
+may just install the `extension`_ and then a file named cloud-config.yaml
```
s-makin, line 67:
```suggestion
can install the `extension`_ and then a file named ``cloud-config.yaml``
```

doc/rtd/explanation/return_codes.rst
```diff
@@ -0,0 +1,150 @@
...
+will automatically use cloud-init's jsonschema.
```
s-makin, line 68:
```suggestion
will automatically use cloud-init's JSON schema.
```

doc/rtd/explanation/return_codes.rst
```diff
@@ -0,0 +1,150 @@
...
+The cloud-init package includes a cloud-init subcommand,
+:ref:`cloud-init schema<check_user_data_cloud_config>` which uses the schema
+to validate either the configuration passed to the instance that you are
+running the command on, or to validate an arbitrary text file containing a
+configuration.
+
+Return codes
+^^^^^^^^^^^^
+
+Cloud-init historically used two return codes from the `cloud-init status`
```
s-makin, line 83:
```suggestion
Cloud-init historically used two return codes from the :code:`cloud-init status`
```

doc/rtd/reference/cli.rst
```diff
@@ -404,12 +404,14 @@ module default frequency of ``instance``:
 :command:`status`
 =================
 
-Report whether ``cloud-init`` is running, done, disabled or errored. Exits
-non-zero if an error is detected in ``cloud-init``.
+Report `cloud-init's`` current status.
```
s-makin, line 407:
```suggestion
Report cloud-init's current status.
```

s-makin:
> You made some good points about the Python. Cloud-init is in python, so the python syntax shouldn't cause anybody reading the docs any problems.

My $0.02 is basically this - it's a Python project, having the Python code in the docs should be totally fine and expected imo. I don't see anything that's in there with zero context so given that all the code is explained, I'd be happy to see it as-is.

s-makin (approved):
Thanks @holmanb :) lgtm!

holmanb:
Thanks for the reviews @TheRealFalcon @s-makin!

PR #5055, opened 2024-03-13
Ubuntu/focal 24.1.x

holmanb:
sru 24.1.1

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,268 @@
+24.1.1
+ - fix: Include DataSourceCloudStack attribute in unpickle test (#5039)
+ - bug(vmware): initialize new DataSourceVMware attributes at unpickle (#5021)
+ - fix(apt): Don't warn on apt 822 source format (#5028)
+ - fix: Add "broadcast" to network v1 schema (#5034)
+ - pro: honor but warn on custom ubuntu_advantage in /etc/cloud/cloud.cfg (#5030)
+ - net/dhcp: handle timeouts for dhcpcd (#5022)
+ - fix: Make wait_for_url respect explicit arguments
+ - bug(wait_for_url): when exceptions occur url is unset, use url_exc
+ - test: Fix scaleway retry assumptions
+ - fix: Make DataSourceOracle more resilient to early network issues (#5025)
+ - tests: Fix wsl test (#5008)
+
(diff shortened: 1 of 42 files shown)


TheRealFalcon:
I see two additional patch refreshes that were also present in `upstream/ubuntu/focal`. Were they removed intentionally? The bottom one was also refreshed in this snapshot.
(partial diff but the rest isn't relevant):
```diff
diff --git a/debian/changelog b/debian/changelog
index 32b6de474..2b50972b3 100644
--- a/debian/changelog
+++ b/debian/changelog
@@ -19,36 +19,16 @@ cloud-init (24.1.1-0ubuntu1~20.04.1) focal; urgency=medium
     - d/p/retain-netplan-world-readable.patch
     - d/p/status-do-not-remove-duplicated-data.patch
     - d/p/status-retain-recoverable-error-exit-code.patch
+    - d/p/keep-dhclient-as-priority-client.patch
+    - d/p/revert-551f560d-cloud-config-after-snap-seeding.patch
```

holmanb:
> Were they removed intentionally? The bottom one was also refreshed in this snapshot.

Those patches were added in this release, so yes I deleted them. They shouldn't be in the `upstream/ubuntu/focal` changelog either, fwiw, but I hadn't noticed them there.

TheRealFalcon:
Yay unused translations needing updates...
```
W: cloud-init source: newer-debconf-templates
N: 
N:    debconf-updatepo has not been run since the last change to your debconf
N:    templates.
N:    
N:    You should run debconf-updatepo whenever debconf templates files are
N:    changed so that translators can be warned that their files are outdated.
N:    
N:    This can be ensured by running debconf-updatepo in the 'clean' target of
N:    debian/rules. PO files will then always be up-to-date when building the
N:    source package.
N:    
N:    Severity: warning
N:    
N:    Check: debian/po-debconf
```

If you run `debconf-updatepo` from the cloud-init directory, it should add the necessary change.

Also, since we're stuck having to make this change, can we add a little detail to this line of the changelog?
```
  * d/p/keep-dhclient-as-priority-client.patch
```
It's calling out the file but we usually have a short description for new quilt patches.

TheRealFalcon (approved):
LGTM!

PR #5058, opened 2024-03-14
 feat(hotlug): trigger hotplug after cloud-init.service

aciba90:
## Proposed Commit Message

See individual commits.

## Additional Context

https://warthogs.atlassian.net/browse/SC-1716

## Test Steps

```sh
#!/bin/bash

set -eux -o pipefail

make deb

export CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init_all.deb
export CLOUD_INIT_PLATFORM=ec2
tox -e integration-tests -- tests/integration_tests/modules/test_hotplug.py::test_nics_before_config_trigger_hotplug
```

## Checklist

## Merge type

### DIFF
--- a/systemd/cloud-init-hotplugd.service
+++ b/systemd/cloud-init-hotplugd.service
@@ -12,12 +12,13 @@
 [Unit]
 Description=Cloud-init: Hotplug Hook
 After=cloud-init-hotplugd.socket
+After=cloud-init.target
 Requires=cloud-init-hotplugd.socket
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
 
 [Service]
-Type=simple
+Type=oneshot
 ExecStart=/bin/bash -c 'read args <&3; echo "args=$args"; \
                         exec /usr/bin/cloud-init devel hotplug-hook $args; \
                         exit 0'
--- a/systemd/cloud-init-hotplugd.service
+++ b/systemd/cloud-init-hotplugd.service
@@ -16,6 +16,7 @@ After=cloud-init.target
 Requires=cloud-init-hotplugd.socket
 ConditionPathExists=!/etc/cloud/cloud-init.disabled
 ConditionKernelCommandLine=!cloud-init=disabled
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
 [Service]
 Type=oneshot
--- a/tests/integration_tests/modules/test_hotplug.py
+++ b/tests/integration_tests/modules/test_hotplug.py
@@ -1,6 +1,7 @@
 import time
 from collections import namedtuple
 
+import paramiko
 import pytest
 import yaml
 
@@ -13,7 +14,11 @@ from tests.integration_tests.releases import (
     FOCAL,
     UBUNTU_STABLE,
 )
-from tests.integration_tests.util import verify_clean_log
+from tests.integration_tests.util import (
+    push_and_enable_systemd_unit,
+    verify_clean_log,
+    wait_for_cloud_init,
+)
 
 USER_DATA = """\
 #cloud-config
@@ -374,3 +379,156 @@ def test_no_hotplug_triggered_by_docker(client: IntegrationInstance):
     assert "enabled" == client.execute(
         "cloud-init devel hotplug-hook -s net query"
     )
+
+
+def wait_for_cmd(
+    client: IntegrationInstance, cmd: str, return_code: int
+) -> None:
+    for _ in range(60):
+        try:
+            res = client.execute(cmd)
+        except paramiko.ssh_exception.SSHException:
+            pass
+        else:
+            if res.return_code == return_code:
+                return
+        time.sleep(1)
+    assert False, f"`{cmd}` never exited with {return_code}"
+
+
+def assert_systemctl_status_code(
+    client: IntegrationInstance, service: str, return_code: int
+):
+    result = client.execute(f"systemctl status {service}")
+    assert result.return_code == return_code, (
+        f"status of {service} expected to be {return_code} but was"
+        f" {result.return_code}\nstdout: {result.stdout}\n"
+        f"stderr {result.stderr}"
+    )
+
+
+BLOCK_CLOUD_CONFIG = """\
+[Unit]
+Description=Block cloud-config.service
+After=cloud-config.target
+Before=cloud-config.service
+
+DefaultDependencies=no
+Before=shutdown.target
+Conflicts=shutdown.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sleep 360
+TimeoutSec=0
+
+# Output needs to appear in instance console output
+StandardOutput=journal+console
+
+[Install]
+WantedBy=cloud-config.service
+"""  # noqa: E501
+
+
+BLOCK_CLOUD_FINAL = """\
+[Unit]
+Description=Block cloud-final.service
+After=cloud-config.target
+Before=cloud-final.service
+
+DefaultDependencies=no
+Before=shutdown.target
+Conflicts=shutdown.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/sleep 360
+TimeoutSec=0
+
+# Output needs to appear in instance console output
+StandardOutput=journal+console
+
+[Install]
+WantedBy=cloud-final.service
+"""  # noqa: E501
+
+
+def _customize_environment(client: IntegrationInstance):
+    push_and_enable_systemd_unit(
+        client, "block-cloud-config.service", BLOCK_CLOUD_CONFIG
+    )
+    push_and_enable_systemd_unit(
+        client, "block-cloud-final.service", BLOCK_CLOUD_FINAL
+    )
+
+    # Disable pam_nologin for 1000(ubuntu) user to allow ssh access early
+    # during boot. Without this we get:
+    #
+    # System is booting up. Unprivileged users are not permitted to log in yet.
+    # Please come back later. For technical details, see pam_nologin(8).
+    #
+    # sshd[xxx]: fatal: Access denied for user ubuntu by PAM account
+    # configuration [preauth]
+    #
+    # See: pam(7), pam_nologin(8), pam_succeed_id(8)
+    contents = client.read_from_file("/etc/pam.d/sshd")
+    contents = (
+        "account [success=1 default=ignore] pam_succeed_if.so quiet uid eq"
+        " 1000\n\n" + contents
+    )
+    client.write_to_file("/etc/pam.d/sshd", contents)
+
+    client.instance.shutdown(wait=True)
+    client.instance.start(wait=False)
+
+
+@pytest.mark.skipif(
+    PLATFORM != "ec2",
+    reason="test is ec2 specific but should work on other platforms with the"
+    " ability to add_network_interface",
+)
+@pytest.mark.user_data(USER_DATA)
+def test_nics_before_config_trigger_hotplug(client: IntegrationInstance):
+    """
+    Test that NICs added/removed after the Network boot stage but before
+    the rest boot stages do trigger cloud-init-hotplugd.
+
+    Note: Do not test first boot, as cc_install_hotplug runs at
+    config-final.service time.
+    """
+    _customize_environment(client)
+
+    # wait until we are between cloud-config.target done and
+    # cloud-config.service
+    wait_for_cmd(client, "systemctl status cloud-config.target", 0)
+    wait_for_cmd(client, "systemctl status block-cloud-config.service", 3)
+
+    # socket active but service not
+    assert_systemctl_status_code(client, "cloud-init-hotplugd.socket", 0)
+    assert_systemctl_status_code(client, "cloud-init-hotplugd.service", 3)
+
+    assert_systemctl_status_code(client, "cloud-config.service", 3)
+    assert_systemctl_status_code(client, "cloud-final.service", 3)
+
+    added_ip_0 = client.instance.add_network_interface()
+
+    assert_systemctl_status_code(client, "cloud-config.service", 3)
+    assert_systemctl_status_code(client, "cloud-final.service", 3)
+
+    # unblock cloud-config.service
+    assert client.execute("systemctl stop block-cloud-config.service").ok
+    wait_for_cmd(client, "systemctl status cloud-config.service", 0)
+    wait_for_cmd(client, "systemctl status block-cloud-final.service", 3)
+    assert_systemctl_status_code(client, "cloud-final.service", 3)
+
+    # hotplug didn't run before cloud-final.service
+    _wait_till_hotplug_complete(client, expected_runs=0)
+
+    # unblock cloud-final.service
+    assert client.execute("systemctl stop block-cloud-final.service").ok
+
+    wait_for_cloud_init(client)
+    _wait_till_hotplug_complete(client, expected_runs=1)
+
+    client.instance.remove_network_interface(added_ip_0)
+    _wait_till_hotplug_complete(client, expected_runs=2)
--- a/tools/hook-hotplug
+++ b/tools/hook-hotplug
@@ -2,21 +2,26 @@
 # This file is part of cloud-init. See LICENSE file for license information.
 
 # This script checks if cloud-init has hotplug hooked and if
-# cloud-init has finished; if so invoke cloud-init hotplug-hook
+# cloud-init is ready; if so invoke cloud-init hotplug-hook
 
-is_finished() {
-    [ -e /run/cloud-init/result.json ]
+fifo=/run/cloud-init/hook-hotplug-cmd
+
+should_run() {
+    if [ -d /run/systemd ]; then
+        # check that the socket is ready
+        [ -p $fifo ]
+    else
+        # on non-systemd, check cloud-init fully finished.
+        [ -e /run/cloud-init/result.json ]
+    fi
 }
 
-if is_finished; then
-    # open cloud-init's hotplug-hook fifo rw
-    exec 3<>/run/cloud-init/hook-hotplug-cmd
-    env_params=" \
-        --subsystem=${SUBSYSTEM} \
-        handle \
-        --devpath=${DEVPATH} \
-        --udevaction=${ACTION} \
-    "
-    # write params to cloud-init's hotplug-hook fifo
-    echo "${env_params}" >&3
+if ! should_run; then
+    exit 0
 fi
+
+# open cloud-init's hotplug-hook fifo rw
+exec 3<>$fifo
+env_params=" --subsystem=${SUBSYSTEM} handle --devpath=${DEVPATH} --udevaction=${ACTION}"
+# write params to cloud-init's hotplug-hook fifo
+echo "${env_params}" >&3
--- a/systemd/cloud-init-hotplugd.socket
+++ b/systemd/cloud-init-hotplugd.socket
@@ -6,6 +6,9 @@
 [Unit]
 Description=cloud-init hotplug hook socket
 After=cloud-config.target
+ConditionPathExists=!/etc/cloud/cloud-init.disabled
+ConditionKernelCommandLine=!cloud-init=disabled
+ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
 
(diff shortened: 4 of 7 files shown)


tests/integration_tests/modules/test_hotplug.py
```diff
@@ -394,3 +395,145 @@ def test_no_hotplug_triggered_by_docker(client: IntegrationInstance):
...
+    # wait until we are between cloud-config.target done and
+    # cloud-config.service
+    wait_for_cmd(client, "systemctl status cloud-config.target", 0)
+    wait_for_cmd(client, "systemctl status block-cloud-config.service", 3)
+
+    assert_systemctl_status_code(client, "cloud-config.service", 3)
+    assert_systemctl_status_code(client, "cloud-final.service", 3)
+
+    added_ip_0 = client.instance.add_network_interface()
+    _wait_till_hotplug_complete(client, expected_runs=1)
```
holmanb, line 521: Before we `_wait_till_...`, I think that we should assert that we are still between `cloud-config.target` done and `cloud-config.service`? If not, then this test didn't do what we thought it did, right?

tools/hook-hotplug
```diff
@@ -2,21 +2,31 @@
...
-    # write params to cloud-init's hotplug-hook fifo
-    echo "${env_params}" >&3
+if ! should_run; then
+    exit 0
 fi
+
+# open cloud-init's hotplug-hook fifo rw
+exec 3<>$fifo
+env_params=" \
+--subsystem=${SUBSYSTEM} \
```
holmanb, line 26: nit: Could we indent this and the following lines evaluate to a single string? It looks like you did this to avoid the extra indentation between the args (which I appreciate).  I think we can have the best of both worlds by quoting the strings on each line so that they get concatenated together before assignment.

tests/integration_tests/modules/test_hotplug.py
```diff
@@ -394,3 +395,145 @@ def test_no_hotplug_triggered_by_docker(client: IntegrationInstance):
...
+    client.instance.start(wait=False)
+
+
+@pytest.mark.skipif(
+    PLATFORM != "ec2",
+    reason="test is ec2 specific but should work on other platforms with the"
+    " ability to add_network_interface",
+)
+@pytest.mark.user_data(USER_DATA)
+def test_nics_before_config_trigger_hotplug(client: IntegrationInstance):
```
holmanb, line 502: Nice work getting this test set up!

holmanb (approved):
@aciba90 Great job on this, and nice test! This looks good to me (besides the comments inline)

I think it'd be good to get a review from another teammate and also probably an SRU reviewer because of the system change, so lets not merge this yet.

holmanb:
@TheRealFalcon since you've worked on hotplug before could you please give this another review before we put this in front of SRU reviewer's eyes?

tools/hook-hotplug
aciba90: Do you mean something like:

```sh
env_params=""\
   " --subsystem=${SUBSYSTEM}"\
   " handle"\
   " --devpath=${DEVPATH}"\
   " --udevaction=${ACTION}"
```

This would be nice, but it is not POSIX compatible, it fails in dash and probably on other shells than bash.

I would just KISS and define this var in single line like the following, yes it has 90 characters, but it is more portable and readable IMO.

```sh
env_params=" --subsystem=${SUBSYSTEM} handle --devpath=${DEVPATH} --udevaction=${ACTION}"
```

Or we could use the short version of some parameter if 90 is not acceptable:

```sh
env_params=" -s=${SUBSYSTEM} handle --devpath=${DEVPATH} --udevaction=${ACTION}"
```

tests/integration_tests/modules/test_hotplug.py
aciba90: What I am currently assert is that `cloud-config.target` is finished in line [514: wait_for_cmd(client, "systemctl status cloud-config.target", 0)](https://github.com/canonical/cloud-init/pull/5058/files/6ee132506d6b966e80ee2e2c54639d3bbe6c55af#diff-fe6975e54f120fbe110935bb6e5cd8ba9b5dfe6613eae48787bcbc079c582299R514) and that `cloud-config.service` has not yet started in line  [517: assert_systemctl_status_code(client, "cloud-config.service", 3)](https://github.com/canonical/cloud-init/pull/5058/files/6ee132506d6b966e80ee2e2c54639d3bbe6c55af#diff-fe6975e54f120fbe110935bb6e5cd8ba9b5dfe6613eae48787bcbc079c582299R517). I think that is exactly what you mean, or am I missing something?

tests/integration_tests/modules/test_hotplug.py
```diff
@@ -394,3 +395,145 @@ def test_no_hotplug_triggered_by_docker(client: IntegrationInstance):
...
+# Output needs to appear in instance console output
+StandardOutput=journal+console
+
+[Install]
+WantedBy=cloud-final.service
+"""  # noqa: E501
+
+
+def _customize_environment(client: IntegrationInstance):
+    def enable_block_service(service_name: str, content: str):
```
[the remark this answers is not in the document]
aciba90: That's a good idea. I did a quick search and didn't find tests pausing cloud-init services. Which ones are you referring to? I'd be happy to refactor / offer this testing ability in a generic manner, but, could we split it as a different issue? And I am also okay with you or anyone else doing it.

tests/integration_tests/modules/test_hotplug.py
```diff
@@ -394,3 +395,145 @@ def test_no_hotplug_triggered_by_docker(client: IntegrationInstance):
...
+        client.write_to_file(service_filename, content)
+        client.execute(f"chmod 0644 {service_filename}", use_sudo=True)
+        client.execute(
+            f"systemctl enable {service_name}.service", use_sudo=True
+        )
+
+    enable_block_service("block-cloud-config", BLOCK_CLOUD_CONFIG)
+    enable_block_service("block-cloud-final", BLOCK_CLOUD_FINAL)
+
+    # Allows ssh access for 1000 user before boot has finished
```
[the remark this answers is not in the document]
aciba90: Sure, let me find a pointer.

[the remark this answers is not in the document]
aciba90: done

tools/hook-hotplug
holmanb: > This would be nice, but it is not POSIX compatible, it fails in dash and probably on other shells than bash.

Oops, sorry for the bad suggestion.

> I would just KISS and define this var in single line like the following, yes it has 90 characters, but it is more portable and readable IMO.

+1, I'm fine with that

tests/integration_tests/modules/test_hotplug.py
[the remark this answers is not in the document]
holmanb: >> There are some other tests that also do effectively this - this looks reusable. No need to make a change now, but I might refactor this at some point if you don't.
>
> That's a good idea. I did a quick search and didn't find tests pausing cloud-init services.

I was referring to enabling services, but it looks like both of the tests that I was thinking of [[1](https://github.com/canonical/cloud-init/blob/3ad7f287664e41bf6b2a668df78176e62924fc6f/tests/integration_tests/modules/test_ansible.py#L85)][[2](https://github.com/canonical/cloud-init/blob/3ad7f287664e41bf6b2a668df78176e62924fc6f/tests/integration_tests/cmd/test_status.py#L109)] actually do this in userdata. That said, I don't think there's a real advantage to doing it in user-data. It's probably harder to understand and modify. I'm just noting that "install & enable new service and reboot" is a common paradigm in integration tests, so we could create a helper around to simplify things.

> could we split it as a different issue?

Of course - no need to do that in this PR.

tests/integration_tests/modules/test_hotplug.py
holmanb: Those assertions tell us before we add the interface that cloud-config.target is finished and that cloud-config.service has not yet started, but I think that a race is possible. If adding the network interface has some pre-flight check that takes a long time (or later adds something like this on accident), wouldn't it be possible that `add_network_interface()` doesn't actually add the interface until after cloud-config.service has started?

holmanb: With a 6 minute sleep this is all unlikely to happen, but I think that an assertion after adding the nic that we're still in the state we thought we were in would assert that the test is testing what we think it is.

TheRealFalcon (commented):
Shouldn't we also include a `After=cloud-final.service` in `cloud-init-hotplugd.service`?

With this PR as it is, we will capture all hotplug events that have happened after networking has come up and respond to them immediately. Instead, I thought we wanted to capture those events, wait for the regular cloud-init services to finish, and then respond to the hotplug events. This is due to (theoretical) races with reading and writing the pickle with multiple cloud-init services running simultaneously. 

Maybe this isn't a real problem, in which case this PR would be fine.

tests/integration_tests/modules/test_hotplug.py
aciba90: Gotcha, updated.

tools/hook-hotplug
aciba90: Updated to single line.

tests/integration_tests/modules/test_hotplug.py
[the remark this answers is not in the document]
aciba90: Refactored in d935c37954399ec2bf6989d8c18d3f4ab2777ad, thanks.

systemd/cloud-init-hotplugd.service
```diff
@@ -12,6 +12,7 @@
 [Unit]
 Description=cloud-init hotplug hook daemon
 After=cloud-init-hotplugd.socket
+After=cloud-init.target
```
TheRealFalcon, line 15: This should be `cloud-final.servce`, correct?

aciba90: I think it is equivalent, as `cloud-init.target` is a sync point after `cloud-config.service` and `cloud-final.service`, but more resilient for future changes.

But I am good either way.

TheRealFalcon (approved):
LGTM!

aciba90:
Since this is touching systemd units, could you, @basak, please take a look?

aciba90:
`main` rebased to fix merge conflicts.

TheRealFalcon:
@aciba90 , I think we'll need to actively pursue a reviewer if we want them to take a look at this. Your ping was from before 24.04 release and I'm guessing it has been long lost and forgotten in an inbox.

aciba90:
> @aciba90 , I think we'll need to actively pursue a reviewer if we want them to take a look at this. Your ping was from before 24.04 release and I'm guessing it has been long lost and forgotten in an inbox.

I did so a couple of times in Jira and MM! Re-pinging @basak.

basak:
Ack. I was going to try and look at this with Brett this week.

holmanb:
@aciba90 A couple of concerns:

This is going to enable running a cloud-init service while others are currently running. This carries risk of things that weren't previously possible.

I have some questions:

- What would be the correct behavior if two interface uevents happen simultaneously, or one shortly after the other?

- Is pickling used by the hotplug event? If so, we need to be careful that the primary service cannot be writing to the pickle while the hotplug command is reading and vice versa. This could crash cloud-init.

- Can a socket unit and its paired service unit have different ordering? Since hotplugd.socket depends on cloudinit.service but hotplugd.service depends on cloud-final.service, I haven't found any documentation about what should happen in this configuration.

aciba90:
Thanks for the reviews and questions. I have reworked the PR a bit.

I have modified 62ace19d6796ba4b8bee7db8c583b257b0b409ef and added 57c022d652abf53854546af5abcfffdef7fa3c6f. Please, read the commit messages for full context.

tldr: As architected right now, the `hotplugd.socket` will be active after `cloud-init.service` has finished, enqueueing udev hotplug events, `hotplugd.service` will process the events, one at the time, after `cloud-final.service` has finished. Thus, no races accessing `obj.pkl` or processing udev events will happen.

With this, all the questions are solved, please let me know if you, @holmanb and @basak, have more concerns.

holmanb (commented):
> and added https://github.com/canonical/cloud-init/commit/57c022d652abf53854546af5abcfffdef7fa3c6f

+1 no objections this matches the other services

> tldr: As architected right now, the hotplugd.socket will be active after cloud-init.service has finished, enqueueing udev hotplug events, hotplugd.service will process the events, one at the time, after cloud-final.service has finished. Thus, no races accessing obj.pkl or processing udev events will happen.

If multiple events are queued, then a read from the pipe will yield:

```
--subsystem=${SUBSYSTEM1} handle --devpath=${DEVPATH1} --udevaction=${ACTION1} --subsystem=${SUBSYSTEM2} handle --devpath=${DEVPATH2} --udevaction=${ACTION2} 
```

Right? It doesn't look like the argparser code will handle this correctly, will it?

As an aside, (and I know that you didn't design it this way) there are several other things that I really don't like about this hotplug code:

- This is the only remaining runtime `bash` dependency that cloud-init has, and it is unjustified - all it does is just read from a file descriptor and run another command.
- The hotplug call chain is unnecessarily convoluted. In `hook-hotplug` we create some cli arguments and then write them to a FIFO. In `cloud-init-hotplugd.service` we have the bash one-liner which concatenates the arguments and calls cloud-init with the subcommands. We could just as easily have passed `SUBSYSTEM`, `DEVPATH` and `ACTION` to the FIFO and in `cloud-init-hotplugd.service` just call `cloud-init devel hotplug-hook` and done the necessary processing in Python. All we would need to do is an `os.read(3, ...)` to read from the fifo that systemd sets up for the socket activated service and then do all of the required processing internal to python code.
- calling something "hotplugd" when it isn't actually a daemon is ugly and confusing. Socket activation is event driven and doesn't involve any daemons - several different people* have been confused by this misnomer. Using socket activation is a significantly better use of resources than a daemon-based implementation would have been.

dermotbradley:
> * This is the only remaining runtime `bash` dependency that cloud-init has, and it is unjustified - all it does is just read from a file descriptor and run another command.

Are you referring to tools/hook-hotplug?

I changed this to be POSIX shell rather than Bash-specific in #4772.

holmanb:
> > * This is the only remaining runtime `bash` dependency that cloud-init has, and it is unjustified - all it does is just read from a file descriptor and run another command.
> 
> @holmanb Are you referring to tools/hook-hotplug?
> 
> I changed this to be POSIX shell rather than Bash-specific in #4772.

No I was talking about this one:

https://github.com/canonical/cloud-init/blob/00ab32835d25527ff8986fc3cc0ae1c649aa2202/systemd/cloud-init-hotplugd.service#L21

aciba90:
> > and added [57c022d](https://github.com/canonical/cloud-init/commit/57c022d652abf53854546af5abcfffdef7fa3c6f)
> 
> +1 no objections this matches the other services
> 
> > tldr: As architected right now, the hotplugd.socket will be active after cloud-init.service has finished, enqueueing udev hotplug events, hotplugd.service will process the events, one at the time, after cloud-final.service has finished. Thus, no races accessing obj.pkl or processing udev events will happen.
> 
> If multiple events are queued, then a read from the pipe will yield:
> 
> ```
> --subsystem=${SUBSYSTEM1} handle --devpath=${DEVPATH1} --udevaction=${ACTION1} --subsystem=${SUBSYSTEM2} handle --devpath=${DEVPATH2} --udevaction=${ACTION2} 
> ```
> 
> Right? It doesn't look like the argparser code will handle this correctly, will it?

No, that is not right. `hook-hotplug` echoes one line per udev hotplug event to the socket:

```
# contents of /run/cloud-init/hook-hotplug-cmd:
--subsystem=${SUBSYSTEM1} handle --devpath=${DEVPATH1} --udevaction=${ACTION1}
--subsystem=${SUBSYSTEM2} handle --devpath=${DEVPATH2} --udevaction=${ACTION2}
```

Then, systemd will activate a single instance of `cloud-init-hotplugd.service`. 
`cloud-init-hotplugd.service` will use the bash read command which reads only one line, process it and finish.
Then, as the socket has content, systemd  will activate another instance of the service.
And so on.
 
> As an aside, (and I know that you didn't design it this way) there are several other things that I really don't like about this hotplug code:
> 
> * This is the only remaining runtime `bash` dependency that cloud-init has, and it is unjustified - all it does is just read from a file descriptor and run another command.
> 
> * The hotplug call chain is unnecessarily convoluted. In `hook-hotplug` we create some cli arguments and then write them to a FIFO. In `cloud-init-hotplugd.service` we have the bash one-liner which concatenates the arguments and calls cloud-init with the subcommands. We could just as easily have passed `SUBSYSTEM`, `DEVPATH` and `ACTION` to the FIFO and in `cloud-init-hotplugd.service` just call `cloud-init devel hotplug-hook` and done the necessary processing in Python. All we would need to do is an `os.read(3, ...)` to read from the fifo that systemd sets up for the socket activated service and then do all of the required processing internal to python code.

As said above, the use of `read` does ensure only one event is process at the same time. Not sure how `read` behaves in other shell implementations or if it is POSIX complaint, but in any case, I am pretty sure we can implement a *read only one line* POSIX compliant command. But I would vote split that work out of this PR, to limit the scope and because this PR has been open since mid-March.

> * calling something "hotplugd" when it isn't actually a daemon is ugly and confusing. Socket activation is event driven and doesn't involve any daemons - several different people* have been confused by this misnomer. Using socket activation is a significantly better use of resources than a daemon-based implementation would have been.

Daemons can also be activated via sockets: sshd, libvirtd, lvm2-lvmpolld, snapd, uuidd, virtlockd, virtlockd, virtlogd. 

From a user POV, the {udev rule, socket, service} combo acts as a daemon, but yes, it is weird that the *daemon* service is not a long-running process.

I see the value in the renaming, to not break users, we could rename the service and socket and maintain a systemd unit alias pointing to the older names, as a separated ticket too.

Let me know what you think about it, please.

aciba90:
And [57c022d](https://github.com/canonical/cloud-init/commit/57c022d652abf53854546af5abcfffdef7fa3c6f) needs to be patch out for older Ubuntu downstream releases, as those key are not included in older systemd versions.

basak:
> tldr: As architected right now, the hotplugd.socket will be active after cloud-init.service has finished, enqueueing udev hotplug events, hotplugd.service will process the events, one at the time, after cloud-final.service has finished. Thus, no races accessing obj.pkl or processing udev events will happen.

What if a hotplug event arrives after cloud-init.service has read the instance metadata, but before it completes and hotplugd.service has started? Wouldn't processing for that event get missed?

basak:
> and hotplugd.service has started

nor hotplugd.socket, for that matter.

aciba90:
@basak:

> What if a hotplug event arrives after cloud-init.service has read the instance metadata, but before it completes and hotplugd.service has started? Wouldn't processing for that event get missed? nor hotplugd.socket, for that matter.

In that case, those events are intentionally ignored because the instance could be configured to update the network on [BOOT_LEGACY](https://docs.cloud-init.io/en/latest/explanation/events.html#id1) events, i.e. on every boot: once during `cloud-init-local.service` and another time during `cloud-init.service`, and then if BOOT_LEGACY and HOTPLUG are enabled and if HOTPLUG would run after `cloud-init-local.service`, interfaces could be configured two times.

This is documented in the 62ace19d6796ba4b8bee7db8c583b257b0b409ef's message:

> Note 2: The reason to not extend the scope of this to after init-local is
completed is that during init, network config can be applied if
BOOT_LEGACY is configured as an update event and could race with hotplug.

As this PR is right now, users that want full coverage can enable BOOT_LEGACY and HOTPLUG network update events.

Even if the instance is configured with BOOT_LEGACY and HOTPLUG there is still a small gap when using datasources that determine what NICs to configure via IMDS. Between cloud-init fetches the IMDS at `cloud-init.service` timeframe and `cloud-init.service` finishes where hotplug events will be missed.

I think this gap is okay versus configuring two times the same NIC, and we can think about how to solve that gap in the future if it's seen as a problem.

holmanb (approved):
> daemon service is not a long-running process.
>
> I see the value in the renaming, to not break users, we could rename the service and socket and maintain a systemd unit alias pointing to the older names, as a separated ticket too.
>
> Let me know what you think about it, please.

- Yes this seems reasonable. Either an alias on old releasrs or maybe we just patch out the renames there. Either way works for me (in a separate PR).

Thanks @aciba90 for your efforts on this! I think that you have addressed my concerns, let's land it.

PR #5061, opened 2024-03-14
distro(freebsd): add_user: respect homedir

igalic:
## Proposed Commit Message

```
distro(freebsd): add_user: respect homedir

reportedly, passing `homedir` to a user struct doesn't actually change
the homedir to the desired value. This is because we hard-code the path.
Respect the value we are passed, and only make up a default when we
don't get a `homedir`.

Simplify style to remove cerimonious verbosity around strings, and amend
tests to verify what we did.


```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -121,14 +121,10 @@ class Distro(cloudinit.distros.bsd.BSD):
             pw_useradd_cmd.append("-d/nonexistent")
             log_pw_useradd_cmd.append("-d/nonexistent")
         else:
-            pw_useradd_cmd.append(
-                "-d{home_dir}/{name}".format(home_dir=self.home_dir, name=name)
-            )
+            homedir = kwargs.get("homedir", f"{self.home_dir}/{name}")
+            pw_useradd_cmd.append("-d" + homedir)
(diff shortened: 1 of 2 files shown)


cloudinit/distros/freebsd.py
```diff
@@ -121,13 +121,14 @@ def add_user(self, name, **kwargs):
...
             log_pw_useradd_cmd.append("-d/nonexistent")
         else:
-            pw_useradd_cmd.append(
-                "-d{home_dir}/{name}".format(home_dir=self.home_dir, name=name)
+            homedir = "{home_dir}/{name}".format(
+                home_dir=self.home_dir, name=name
             )
+            if "homedir" in kwargs:
+                homedir = kwargs["homedir"]
+            pw_useradd_cmd.append("-d{homedir}".format(homedir=homedir))
```
holmanb, line 129: Nit: This is a little verbose, I think the lines above could be simplified to the following:
```
homedir = kwargs.get("homedir", f"{self.home_dir}/{name}")
pw_useradd_cmd.append("-d" + homedir)
```

igalic: done!

holmanb (approved):
Thanks @igalic!

PR #5065, opened 2024-03-18
test: mocking updates

TheRealFalcon:
## Proposed Commit Message

```
See individual commits
```

## Merge type

### DIFF
--- a/tests/unittests/conftest.py
+++ b/tests/unittests/conftest.py
@@ -88,18 +88,8 @@ def fake_filesystem(mocker, tmpdir):
 
 
 @pytest.fixture(scope="session", autouse=True)
-def disable_sysfs_net(request, tmpdir_factory):
-    """Avoid tests which read the undertying host's /syc/class/net.
-
-    To allow unobscured reads of /sys/class/net on the host we can
-    parametrize the fixture with:
-
-    @pytest.mark.parametrize("disable_sysfs_net", [False], indirect=True)
-    """
-    if hasattr(request, "param") and getattr(request, "param") is False:
-        # Test disabled this fixture, perform no mocks.
-        yield
-        return
+def disable_sysfs_net(tmpdir_factory):
+    """Avoid tests which read the underlying host's /syc/class/net."""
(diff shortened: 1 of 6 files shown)


tests/unittests/conftest.py
```diff
@@ -88,18 +88,8 @@ def fake_filesystem(mocker, tmpdir):
...
-    parametrize the fixture with:
-
-    @pytest.mark.parametrize("disable_sysfs_net", [False], indirect=True)
-    """
-    if hasattr(request, "param") and getattr(request, "param") is False:
-        # Test disabled this fixture, perform no mocks.
-        yield
-        return
+def disable_sysfs_net(tmpdir_factory):
+    """Avoid tests which read the undertying host's /syc/class/net."""
```
holmanb, line 92: sp: underlying

holmanb (changes requested):
with this change, shouldn't `get_sys_class_path()` return an empty directory? I'm currently toying around with a sample test and this PR and I see that `get_devicelist()` still returns my underlying interfaces, which seems incorrect.

```python
from cloudinit.net import get_devicelist

def test_sysfs():
    assert not get_devicelist()
```

result:

```python
================================================== FAILURES ===================================================
_________________________________________________ test_sysfs __________________________________________________

    def test_sysfs():
>       assert not get_devicelist()
E       AssertionError: assert not ['wlp1s0', 'virbr0', 'lo', 'lxdbr0']
E        +  where ['wlp1s0', 'virbr0', 'lo', 'lxdbr0'] = get_devicelist()
```

TheRealFalcon:
How are you running it? I created a test file at `tests/unittests/test_example.py` containing the code you pasted, ran it with `pytest tests/unittests/test_example.py` and it passes for me.

holmanb:
> How are you running it? I created a test file at tests/unittests/test_example.py containing the code you pasted, ran it with pytest tests/unittests/test_example.py and it passes for me.

My bad you're right this works as expected - I was running one directory higher since I (incorrectly) thought that this was in the top-level conftest.py.

holmanb (approved):
LGTM

PR #5067, opened 2024-03-18
Growpart refactor distro specific code

holmanb:
## Context
Refactor growpart code into distro definitions.

I wrote this a while ago but forgot to submit it.

## Merge type

### DIFF
--- a/cloudinit/distros/bsd.py
+++ b/cloudinit/distros/bsd.py
@@ -149,3 +149,7 @@ class BSD(distros.Distro):
         """
         ppid, _ = subp.subp(["ps", "-oppid=", "-p", str(pid)])
         return int(ppid.strip())
+
+    @staticmethod
+    def get_mapped_device(blockdev: str) -> Optional[str]:
+        return None
--- a/cloudinit/distros/bsd.py
+++ b/cloudinit/distros/bsd.py
@@ -1,5 +1,6 @@
 import logging
 import platform
+import re
 from typing import List, Optional
 
 import cloudinit.net.netops.bsd_netops as bsd_netops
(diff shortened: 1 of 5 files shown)


holmanb:
requesting review from @dermotbradley and @igalic

cloudinit/distros/bsd.py
```diff
@@ -149,3 +150,23 @@ def get_proc_ppid(pid):
...
+
+    @staticmethod
+    def get_mapped_device(blockdev: str) -> Optional[str]:
+        return None
+
+    @staticmethod
+    def device_part_info(devpath: str) -> tuple:
+        # FreeBSD doesn't know of sysfs so just get everything we need from
+        # the device, like /dev/vtbd0p2.
+        part = util.find_freebsd_part(devpath)
```
igalic, line 162: this function is FreeBSD (and Dragonfly) specific.
This is the BSD base-class, and should work on all OSes;
I'm not sure we have such a thing, but might be able to cobble it together.

igalic (changes requested):
This will only work on FreeBSD

cloudinit/distros/bsd.py
holmanb: +1 If you don't mind helping that would be appreciated. I believe that this code works as it did before, since this is what would get called if `util.is_BSD()` returned true.

igalic: in trying to patch this up, I realied that right now, we don't actually support growpart/growfs on NetBSD or OpenBSD, so, yeah, right now, that code will work just as before. OpenBSD at least doesn't support online resizing, so it's no big deal there.

I think we can go forth as is, and then when NetBSD people finally complain we can start fixing it.
meanwhile, I'll build an image to test growpart on FreeBSD

cloudinit/distros/bsd.py
```diff
@@ -149,3 +150,23 @@ def get_proc_ppid(pid):
...
+        if part:
+            fpart = f"/dev/{part}"
+            # Handle both GPT partitions and MBR slices with partitions
+            m = re.search(
+                r"^(?P<dev>/dev/.+)[sp](?P<part_slice>\d+[a-z]*)$", fpart
+            )
+            if m:
+                return m["dev"], m["part_slice"]
+
+        return distros.Distro.device_part_info(devpath)
```
TheRealFalcon, line 172: I'm guessing that the lack of `super()` here is intentional due to it being a static method?

TheRealFalcon (approved):
LGTM! Nice refactor

holmanb:
Force pushed to add PR# to commit message

PR #5073, opened 2024-03-19
fix: Ensure network config in DataSourceOracle can be unpickled

TheRealFalcon:
## Proposed Commit Message

```
fix: Ensure network config in DataSourceOracle can be unpickled

`_network_config` is only explicitly set on the instance if
`_is_iscsi_root()` is True. This means that when `_is_iscsi_root()` is
False, we're modifying the class variable directly. When the instance
gets pickled, the class variable is not included so such changes do
not get persisted. This commit fixes this.
```

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceOracle.py
+++ b/cloudinit/sources/DataSourceOracle.py
@@ -123,7 +123,6 @@ class DataSourceOracle(sources.DataSource):
         sources.NetworkConfigSource.INITRAMFS,
     )
 
-    _network_config: dict = {"config": [], "version": 1}
     perform_dhcp_setup = True
 
     # Careful...these can be overridden in __init__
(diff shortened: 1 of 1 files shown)


a-dubs:
Changes look good! Are there any unit tests or anything to test pickling to prevent future regressions?

TheRealFalcon:
I was wondering about unit tests. Without the `if not hasattr(self, "_network_config"):` section, the other changes will fail the `test_upgrade.py` unit test. This issue was found due to failing tests (though somewhat reliant on ordering), so I think that's probably sufficient here. I think that anything else would likely be overly contrived, but I'm also open to ideas.

blackboxsw (approved):
+1 This LGTM. Better to have this mutable attribute specifically on the instance. And I'm good with tests/unittests/test_upgrade.py effectively catching your move from class attribute -> instance attribute and the assertion that the hasattr covereage is required to initialize the default value of _network_config across deserialization.

PR #5074, opened 2024-03-19
Cleanup test side effects

TheRealFalcon:
## Proposed Commit Message

```
See individual commits
```

## Additional Context
Issues were found using a combination of
https://pypi.org/project/pytest-randomly/
and
https://pypi.org/project/pytest-bisect-tests/

## Merge type

### DIFF
--- a/cloudinit/sources/helpers/openstack.py
+++ b/cloudinit/sources/helpers/openstack.py
@@ -763,7 +763,7 @@ def convert_net_json(network_json=None, known_macs=None):
                 cfg["type"] = "infiniband"
 
     for service in services:
-        cfg = service
+        cfg = copy.deepcopy(service)
         cfg.update({"type": "nameserver"})
         config.append(cfg)
 
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -35,14 +35,7 @@ from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.config import cc_set_hostname
 from cloudinit.config.modules import Modules
 from cloudinit.config.schema import validate_cloudconfig_schema
-from cloudinit.log import (
-    LogExporter,
-    setup_basic_logging,
-    setup_logging,
-    reset_logging,
-    configure_root_logger,
-    DEPRECATED,
-)
+from cloudinit import log
 from cloudinit.reporting import events
 from cloudinit.safeyaml import load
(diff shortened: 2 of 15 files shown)


aciba90 (approved):
LGTM. Thanks for this effort!

Do you think it is worth creating a scheduled GH action exercising [pytest-randomly](https://pypi.org/project/pytest-randomly/) and [pytest-bisect-tests](https://pypi.org/project/pytest-bisect-tests/)?

a-dubs:
> Do you think it is worth creating a scheduled GH action exercising [pytest-randomly](https://pypi.org/project/pytest-randomly/) and [pytest-bisect-tests](https://pypi.org/project/pytest-bisect-tests/)?

I agree with @aciba90. Definitely does not need to be a PR action, but would be good to catch unit tests issues sooner.

cloudinit/sources/helpers/openstack.py
```diff
@@ -763,7 +763,7 @@ def convert_net_json(network_json=None, known_macs=None):
                 cfg["type"] = "infiniband"
 
     for service in services:
-        cfg = service
+        cfg = copy.deepcopy(service)
```
a-dubs, line 766: what necessitated this change? just curious because I don't see how this related to leaked/failed mocks?

TheRealFalcon: With the copy issues, it's not necessarily about mocks and instead about causing side effects that effect other tests.

TheRealFalcon:
> Do you think it is worth creating a scheduled GH action exercising [pytest-randomly](https://pypi.org/project/pytest-randomly/) and [pytest-bisect-tests](https://pypi.org/project/pytest-bisect-tests/)?

Yes, I'd like to have a daily that does that along with running unit tests on another platform to further ensure things are mocked enough.

cloudinit/sources/helpers/openstack.py
a-dubs: thank you! that makes more sense now.

holmanb:
> > Do you think it is worth creating a scheduled GH action exercising [pytest-randomly](https://pypi.org/project/pytest-randomly/) and [pytest-bisect-tests](https://pypi.org/project/pytest-bisect-tests/)?
> 
> Yes, I'd like to have a daily that does that along with running unit tests on another platform to further ensure things are mocked enough.

@TheRealFalcon when you get a chance, I've [proposed an Alpine ci PR for this](https://github.com/canonical/cloud-init/pull/5121).

PR #5081, opened 2024-03-20
Ubuntu/noble: Remove netifaces

TheRealFalcon:
Remove netifaces build and runtime dependency


blackboxsw (commented):
Let's avoid merging this change until we've uploaded and gotten 24.1.2 accepted for SRU.

TheRealFalcon:
That's fine, but I don't actually think this branch will get touched as we have an ubuntu/noble-24.1.x branch.

blackboxsw:
> That's fine, but I don't actually think this branch will get touched as we have an ubuntu/noble-24.1.x branch.

You are correct. I had forgotten we'd already created 24.1.x branch last week for the hotfix uploads. All good here.

blackboxsw (approved):

Issue #5075, opened 2024-03-19
rename attempt of already renamed device


rjschwei:
# Bug report
When cloud-init is instructed to rename an interface it should check if the name is already being used or not before trying to attempt to rename a network device.

We also have to decide what the behavior should be when the name is detected to exist.

## Steps to reproduce the problem
There is an inherent race condition between cloud-init and udev that occurs when the system is being rebooted before cloud-init is completely finished. In a case where a network interface gets assigned a name by the user through the network configuration cloud-init will write this information into the udev rules file (persistent-net-cloud-init.rules) generated during the init-local phase. If after cloud-init local is complete the system gets rebooted and the same network config is provided by the user cloud-init will start again, as usual, but on this boot the udev rules file also exists and as such udev will run and process the rules in the previously generated persistent-net-cloud-init.rules file. While the file itself will get overwritten during the new run udev will have already read the old file and process it. At this point it is just a matter of time that udev will create the renamed device setup and by the time cloud-init comes around the rename will have already happened and cloud-init will generate an error:

"""
Command: ['ip', 'link', 'set', 'enp55s1f0', 'name', 'ccd_ban']
Exit code: 2
Reason: -
Stdout: 
Stderr: RTNETLINK answers: File exists
"""

As a result the interface remains down.

## Environment details
- Cloud-init version: 23.3
- Operating System Distribution: SLES
- Cloud provider, platform or installer type:

## cloud-init logs
cloud-init starts at 2024-02-08 05:54:17,618 then the network devices are enumerated starting at 2024-02-08 05:54:17,790 and ending at 2024-02-08 05:54:17,829

Then network devices are enumerated again starting at 2024-02-08 05:54:17,838 and ending at 2024-02-08 05:54:17,843. This is pretty much at the end of the clod-init-local phase. After all this we write out  /etc/udev/rules.d/85-persistent-net-cloud-init.rules

The udev rules are written at

2024-02-08 05:54:17,846

Then at 2024-02-08 05:55:00,230 cloud-init starts it's "init" phase with net device configuration starting again at 2024-02-08 05:55:00,276 and that ends at 2024-02-08 05:55:00,292

Then at 2024-02-08 05:55:03,848 the final phase of cloud-init starts and at 2024-02-08 05:56:08,100 cloud-init gets killed. -> Reboot is triggered

[...]

PR #5086, opened 2024-03-21
Avoid network configuration if we already generated a udev file

rjschwei:
## Proposed Commit Message

In a case where the initialization of an instance gets interrupted by a reboot it is possible that we have already generated the udev rules for the network devices. When the system comes up again we are in a race with udev to apply any renaming that may need to be done. If udev gets there first the device will be in a down state after cloud-init is finished. Set the semaphore to indicate the network is already configured if we find the generated udev rules file.

<type>(optional scope): <summary>  Skip network config if rules are present

Fixes: #5075

## Additional Context
This is a proposal to address the issue, not yet confirmed to work. Therefore I have not yet added a test.

## Checklist

- [ x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)

## Merge type

### DIFF
--- a/cloudinit/net/sysconfig.py
+++ b/cloudinit/net/sysconfig.py
@@ -8,7 +8,7 @@ import os
 import re
 from typing import Mapping, Optional
 
-from cloudinit import subp, util
+from cloudinit import helpers, subp, util
 from cloudinit.distros.parsers import networkmanager_conf, resolv_conf
 from cloudinit.net import (
     IPV6_DYNAMIC_TYPES,
@@ -18,6 +18,7 @@ from cloudinit.net import (
     subnet_is_ipv6,
 )
 from cloudinit.net.network_state import NetworkState
+from cloudinit.settings import PER_ONCE
 
 LOG = logging.getLogger(__name__)
 KNOWN_DISTROS = [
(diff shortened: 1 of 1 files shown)


cloudinit/net/sysconfig.py
```diff
@@ -1021,6 +1022,16 @@ def render_network_state(
         if self.netrules_path:
             netrules_content = self._render_persistent_net(network_state)
             netrules_path = subp.target_path(target, self.netrules_path)
+            if os.path.exists(netrules_path):
+                # There should be a more elegant way to get the path
```
holmanb, line 1026: The `Paths` class has `.get_runpath("sem")`, but I don't think we have access to the `Paths` object in the renderers currently.

To use this we'll need to pass the distros' `self._paths` value to the renderer instantiation (`render_cls()`) in `_get_renderer()` under `cloudinit/distros/__init__.py`.

That will require a new arg in the renderer classes, but I imagine that other renderers will probably need this as well someday, so I think that this would be reasonable to do.

rjschwei:
Closing this. With the new message in place it shows that this is actually  too late already. But the good news is that the code does the correct thing already, i.e. we recognize that there should not be another round of network configuration. That said we still have the problem in #5075

Issue #3781, opened 2023-05-12
sysconfig: NM_CONTROLLED=False should not be set on RHEL8


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1894837](https://bugs.launchpad.net/cloud-init/+bug/1894837)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2020-09-08T12:21:36.599977+00:00
date_fix_committed = None
date_fix_released = None
id = 1894837
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1894837
milestone = None
owner = amansi26
owner_name = Aman Kumar Sinha
private = False
status = triaged
submitter = amansi26
submitter_name = Aman Kumar Sinha
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Aman Kumar Sinha(amansi26)** wrote on 2020-09-08T12:21:36.599977+00:00_

Environment Details: 
Management Control Plane : OpenStack (Ussuri Release)
cloud-init version : 19.1 (community)
Data Source : Config Drive
OS/platform of deployed VM : RHEL 8.2

I am using cloud-init v19.1 where the control plane (OpenStack nova service) passes information (data source) via configdrive during VM deployment.

On a RHEL8.2 VM deployed from the above environment, it is observed that IPv4 interfaces do not come up. This behavior is observed only when NM_CONTROLLED is set to no in the interface files. This value is set from cloud-init src code at the below lines :

https://github.com/canonical/cloud-init/blob/stable-19.4/cloudinit/net/sysconfig.py#L275

we are setting NM_CONTROLLED = no using the code
    iface_defaults = tuple([
        ('ONBOOT', True),
        ('USERCTL', False),
        ('NM_CONTROLLED', False),
        ('BOOTPROTO', 'none'),
        ('STARTMODE', 'auto'),
    ])

under the file [1] . Due to which the Network Manager is not able to handle the interfaces

[1] cloudinit/net/sysconfig.py

When the above pieces of code is updated to set True to NM_CONTROLLED, the IPV4 interfaces come up fine.

Making ('NM_CONTROLLED', True) fixes the issue.

ubuntu-server-builder:
_Launchpad user **Chad Smith(chad.smith)** wrote on 2020-09-08T16:42:45.304727+00:00_

[...]

PR #5089, opened 2024-03-22
fix(rhel): Fix network ordering in upstream cloud-init

holmanb:
## Proposed Commit Message
```
fix(rhel): Fix network ordering in upstream cloud-init

1. NM_CONTROLLED=true should be set to allow cloud-init to wait until network
   devices are online.
2. RHEL derivatives should be using the NetworkManager renderer, so set that
   in the default config.

Fixes GH-3781
```

## Additional Context

Issues were discovered with testing upstream-built RPMs. This should fix any remaining rhel-derivative sysconfig users (are there any?) as well as correctly prioritize NetworkManager when available.

Fixes https://github.com/canonical/cloud-init/issues/3781

## Merge type


holmanb:
Requesting review from @ani-sinha. I assume that this will conflict with some RHEL patches, but I think this should align upstream and downstream by both 1) selecting NetworkManager over sysconfig in the default config and 2) fix sysconfig, if it is in fact still used by any RHEL derivatives.

blackboxsw:
While both changes seem scoped at just Redhat deriviates. Some of these changes could be "risky" to existing downstreams in that it represents a change in behavior which render our cloud.cfg.tmpl directly during package build vs a runtime changeset in the configs we write. Please separate this PR into two different PRs for discussions of each changeset so we can evaluate those separate changes more specifically to make sure we aren't missing something. One PR for dropping the `NM_CONTROLLED=False` logic and one PR for the cloud.cfg.tmpl.

ani-sinha (approved):

blackboxsw (approved):
Thank you for confiming @ani-sinha this as desired downstream for Redhat derivatives. We've agreed we can get behind this in upstream provided that these commits also declare the  "BREAKING CHANGE" footer and maybe even include a `!` in the commit summary to highlight the change in behavior for any redhat derivative downstream using `config/cloud.cfg.tmpl` during their package build process

Issue #5091, opened 2024-03-22
Openstack Datasource not detected in Oracle cloud


palash-delphix:
# Bug report

It looks like recent changes from this PR https://github.com/canonical/cloud-init/pull/4426 have caused an issue in cloud-init where the datasource is not detected causing cloud-init configuration to be ignored.
Here's what we see in the logs:
```
2024-03-19 22:51:34,786 - __init__.py[DEBUG]: Looking for data source in: ['OpenStack', 'None'], via packages ['', 'cloudinit.sources'] that matches dependencies ['FILESYSTEM']
2024-03-19 22:51:34,802 - __init__.py[DEBUG]: Searching for local data source in: ['DataSourceOpenStackLocal']
2024-03-19 22:51:34,802 - handlers.py[DEBUG]: start: init-local/search-OpenStackLocal: searching for local data from DataSourceOpenStackLocal
2024-03-19 22:51:34,802 - __init__.py[DEBUG]: Seeing if we can get any data from <class 'cloudinit.sources.DataSourceOpenStack.DataSourceOpenStackLocal'>
2024-03-19 22:51:34,802 - __init__.py[DEBUG]: Update datasource metadata and network config due to events: boot-new-instance
2024-03-19 22:51:34,803 - dmi.py[DEBUG]: querying dmi data /sys/class/dmi/id/product_name
2024-03-19 22:51:34,803 - dmi.py[DEBUG]: querying dmi data /sys/class/dmi/id/chassis_asset_tag
2024-03-19 22:51:34,803 - util.py[DEBUG]: Reading from /proc/1/environ (quiet=False)
2024-03-19 22:51:34,803 - util.py[DEBUG]: Read 358 bytes from /proc/1/environ
2024-03-19 22:51:34,803 - __init__.py[DEBUG]: Datasource type DataSourceOpenStackLocal [net,ver=None] is not detected.
2024-03-19 22:51:34,803 - __init__.py[DEBUG]: Datasource DataSourceOpenStackLocal [net,ver=None] not updated for events: boot-new-instance
2024-03-19 22:51:34,803 - handlers.py[DEBUG]: finish: init-local/search-OpenStackLocal: SUCCESS: no local data found from DataSourceOpenStackLocal
```
It looks like the `ds-identify` script creates a list with "None":
```
$ cat /run/cloud-init/cloud.cfg
datasource_list: [ OpenStack, None ]
```
In our cloud-init configuration, the last file to be read lexicographically contains:
```
[...]

PR #5098, opened 2024-03-25
fix: Always use single datasource if specified

TheRealFalcon:
## Proposed Commit Message

```
fix: Always use single datasource if specified

ds-identify would automatically append "None" to the datasource_list
if a single entry was provided in /etc/cloud/cloud.cfg[.d].
This wasn't a problem in the past as the python code would detect
a single datasource along with None as an indication to automatically
use that datasource. Since the python code no longer does that,
we should ensure that one specified datasource results in one specified
datasource after ds-identify has run.

Fixes GH-5091
```

## Additional Context
Prior to this commit, in `/etc/cloud/cloud.cfg[.d]`, an entry of `datasource_list: [OpenStack]` results in `/run/cloud-init/cloud.cfg` containing `datasource_list: [ OpenStack, None ]` (OpenStack is just an example here...it's true for any datasource).

## Checklist

## Merge type

### DIFF
--- a/tools/ds-identify
+++ b/tools/ds-identify
@@ -2018,7 +2018,11 @@ _main() {
     # if there is only a single entry in $DI_DSLIST
     if [ $# -eq 1 ] || [ $# -eq 2 -a "$2" = "None" ] ; then
         debug 1 "single entry in datasource_list ($DI_DSLIST) use that."
-        found "$@"
+        if [ $# -eq 1 ]; then
+            write_result "datasource_list: [ $1 ]"
+        else
+            found "$@"
+        fi
         return
     fi
 
--- a/tests/unittests/test_ds_identify.py
+++ b/tests/unittests/test_ds_identify.py
@@ -680,7 +680,7 @@ class TestDsIdentify(DsIdentifyBase):
         mydata = copy.deepcopy(VALID_CFG["Ec2-hvm"])
         cfgpath = "etc/cloud/cloud.cfg.d/myds.cfg"
         mydata["files"][cfgpath] = 'datasource_list: ["NoCloud"]\n'
-        self._check_via_dict(mydata, rc=RC_FOUND, dslist=["NoCloud", DS_NONE])
+        self._check_via_dict(mydata, rc=RC_FOUND, dslist=["NoCloud"])
 
     def test_configured_list_with_none(self):
         """When datasource_list already contains None, None is not added.


a-dubs:
looks good! unit tests just need updated!

tools/ds-identify
```diff
@@ -2018,7 +2018,11 @@ _main() {
     # if there is only a single entry in $DI_DSLIST
     if [ $# -eq 1 ] || [ $# -eq 2 -a "$2" = "None" ] ; then
         debug 1 "single entry in datasource_list ($DI_DSLIST) use that."
-        found "$@"
+        if [ $# -eq 1 ]; then
+            write_result "datasource_list: [ $1 ]"
```
blackboxsw, line 2022: Walking through this, the only way we can get `$# -eq 1` is if one of the following was configs was present:
1. kernel cmdline with `(ds|ci.ds|ci.datasource)=<some_ds_name>`
2. kernel cmdline with `cc:datasource_list: [ <some_ds_name> ] end_cc`    # MAAS does this
3. `/etc/cloud/ds-identify.cfg` with `datasource: <some_ds_name>`     # we should probably deprecate this as i'm unsure where that's used anymore
4. `/etc/cloud/cloud.cfg.d/*` containing `datasource_list [ <some_ds_name> ]`

In those cases, the environment configuration is intentionally limiting the datasource list selection to a single datasource. 

If ds-identify avoids appending fallback `None` to `datasource_list` in `/run/cloud-init/cloud.cfg`ds-identify`,  the system will no longer fallback to DataSourceNone behavior if the primary single datsource is undiscoverable. This will leave us with a slew of WARNINGS which show up in `cloud-init status --format=yaml and /var/log/cloud-init-output.log`
```
    - No instance datasource found! Likely bad things to come!
    - Can not apply stage config, no datasource found! Likely bad things to come!
    - Can not apply stage final, no datasource found! Likely bad things to come!
```

This ultimately leaves us with a system that hasn't run any config modules at all which could make the system error conditions even less discoverable because cloud-init will not have generated SSH host keys or imported SSH pub keys to the instance.   I'm not sure we want to go down this specific path, though it could be argued that a filesystem config or kernel command line that wishes to have a sensible fallback option should either provide `cc:datasource_list: [ DS_NAME, None ] ` or provide that fallback in `/etc/cloud/cloud.cfg.d/*.cfg` as `datasource_list: [ DS_NAME, None]`

blackboxsw: > hasn't run any config modules at all which could make the system error conditions even less discoverable

I'm actually not sure I really believe this statement I made as the DataSourceNone on something like an Azure/Oracle/GCe/Ec2 cloud provides "more" visibility for triage. DataSourceNone won't be reading any cloud platform specific instance metadata configuration anyway which would have provided sensible config for which project/user SSH keys to import anyway. Also, most likely it wouldn't have been provided enough network configuration information to bring up the network anyway unless there was already functional static config in the image (which is possible). So, getting into that broken instance would still be problematic anyway and probably require virtual serial console to access the instance for triage.

TheRealFalcon: I know you've walked back most of your objection, but I just wanna make sure we're on the same page about what's changing.
> If ds-identify avoids appending fallback None to datasource_list in /run/cloud-init/cloud.cfgds-identify,  the system will no longer fallback to DataSourceNone behavior if the primary single datsource is undiscoverable. This will leave us with a slew of WARNINGs which show up in cloud-init status --format=yaml and /var/log/cloud-init-output.log`

Because of #2060 , I don't think this has been true since 23.2. In that PR we skipped datasource detection on the python side if we saw a single datasource or `[datasource, None]`, so DataSourceNone would never actually get used. Am I understanding this correctly? My goal with this commit is to take us back to what things were post 23.2 if a single datasource is specified.

blackboxsw: > Because of #2060 , I don't think this has been true since 23.2. In that PR we skipped datasource detection on the python side if we saw a single datasource or [datasource, None], so DataSourceNone would never actually get used. Am I understanding this correctly? My goal with this commit is to take us back to what things were post 23.2 if a single datasource is specified.

You are right, I think it was 23.1 that we introduced the `override_ds_detect`. Ok, so this behavior observing the single configured datasource has really been active for a while. The `None` fallback has been ignored as a fallback option in any such case where the single datasource went undiscoverable due to `ds_detect` returning False. It just hadn't hit any significant cases until this OpenStack on Oracle config issue #5091.

That said I think this current  #5091 issue could also be resolved without any changes to cloud-init or ds-identify given a simple config file change if they have reasons to retain OpenStack datasource on Oracle platform.

From the looks of the [DataSourceOpenStack.ds_detect](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceOpenStack.py#L246). If their platform needs to use OpenStack datasource over Oracle DataSource. It looks like the following config could be provided in `/etc/cloud/cloud.cfg.d/99-delphix-datasource.cfg`

```
datasource_list: [ OpenStack, Oracle ]
```

This will tell the unique DataSourceOpenStack.ds_detect logic to allow OpenStack datasource detection on any system which exposes chassis asset tag=`OracleCloud.com`.

blackboxsw: That said, I think your suggestion James is the correct long-term solution. If ds-identify only sees one datasource defined in datasource_list, then it really should only persist that single datasource to `/run/cloud-init/cloud.cfg`. It would honor the config provided.

blackboxsw:
The additional context we have here is PR https://github.com/canonical/cloud-init/pull/4426/files which represented a breaking change for some Azure environments which had desired to make use of the fallback DataSourceNone when Azure datasource IMDS was unavailable.

In this case, the base configuration on the system was `datasource_list: [ Azure, None ]` and cloud-init version 23.2 (pre- PR #4426), would defer to only select Azure datasource and [avoid running DataSourceAzure.ds_detect](https://github.com/canonical/cloud-init/blob/23.2/cloudinit/sources/DataSourceAzure.py#L704). In this case, the desire was to actually fallback to DataSourceNone. @TheRealFalcon's PR here will still preserve the base config setting `datasource_list: [ Azure, None ]` which should meet the needs fixed in #4426, but it will also allow any image to customize and limit `datasource_list: [ Azure ]` and avoid falling back to DataSourceNone if Azure datasource is not discovered or errors during `get_data`

blackboxsw (approved):
I think we've enumerate the potential areas where this could cause problems and the behavior that different warnings that would happen for any image that sets a single `datasource_list: [ MyDS ]` . 

- For images which don't specify any custom `datasource_list` in `/etc/cloud/cloud.cfg*` this change doesn't affect them as datasource_list defaults are used and ds-identify will limit viable datasources early in boot
- for images with custom `datasource_list`
  - Those environments with just a single `datasource_list: [ MyDs ]` will continue to only try to detect MYDs and emit warnings about unable to detect datasource in those cases resulting in not trying to run any cloud-config cc_* modules
  - Those environments that want to have a fallback to DataSourceNone configuration if MyDs happens to fail would have to append `None` to the `datasource_list: [ MyDs, None ]`  that is already being customized in their images anyway.


Changeset looks good to me. Let's mark this commit as a breaking change in the commit message and note the workaround for environments which expect a fallback option of None.

PR #5128, opened 2024-04-01
net/dhcp: raise InvalidDHCPLeaseFileError on error parsing dhcpcd lease

cjp256:
Seeing a fairly large number of lease parsing failures on Azure similar to:
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py", line 851, in _get_data
    crawled_data = util.log_time(
                   ^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/util.py", line 2828, in log_time
    ret = func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/helpers/azure.py", line 45, in impl
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py", line 660, in crawl_metadata
    self._wait_for_pps_savable_reuse()
  File "/usr/lib/python3/dist-packages/cloudinit/sources/helpers/azure.py", line 45, in impl
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py", line 1236, in _wait_for_pps_savable_reuse
    self._wait_for_hot_attached_primary_nic(nl_sock)
  File "/usr/lib/python3/dist-packages/cloudinit/sources/helpers/azure.py", line 45, in impl
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py", line 1142, in _wait_for_hot_attached_primary_nic
    primary_nic_found = self._setup_ephemeral_networking(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/helpers/azure.py", line 45, in impl
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/cloudinit/sources/DataSourceAzure.py", line 440, in _setup_ephemeral_networking
    lease = self._ephemeral_dhcp_ctx.obtain_lease()
[...]

PR #5129, opened 2024-04-01
fix(dhcpcd): Make lease parsing more robust

holmanb:
## Proposed Commit Message
```
fix(dhcpcd): Make lease parsing more robust
```

## Additional Context
Followup to https://github.com/canonical/cloud-init/pull/5128 (rebase before merge)

## Merge type

### DIFF
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -796,8 +796,15 @@ class Dhcpcd(DhcpClient):
                     if "=" in a
                 ]
             )
+            if not lease:
+                msg = (
+                    "No valid DHCP lease configuration "
+                    "found in dhcpcd lease: %r"
+                )
+                LOG.error(msg, lease_dump)
+                raise InvalidDHCPLeaseFileError(msg % lease_dump)
         except ValueError as error:
-            LOG.error("Error parsing dhcpcd lease: %r", error)
+            LOG.error("Error parsing dhcpcd lease: %r", lease_dump)
             raise InvalidDHCPLeaseFileError from error
 
         # this is expected by cloud-init's code
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -791,8 +791,9 @@ class Dhcpcd(DhcpClient):
         try:
             lease = dict(
                 [
-                    a.split("=")
+                    a.split("=", maxsplit=1)
                     for a in lease_dump.strip().replace("'", "").split("\n")
+                    if "=" in a
                 ]
             )
         except ValueError as error:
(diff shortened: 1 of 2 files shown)


cloudinit/net/dhcp.py
```diff
@@ -791,8 +791,9 @@ def parse_dhcpcd_lease(lease_dump: str, interface: str) -> Dict:
         try:
             lease = dict(
                 [
-                    a.split("=")
+                    a.split("=", maxsplit=1)
                     for a in lease_dump.strip().replace("'", "").split("\n")
+                    if "=" in a
                 ]
             )
         except ValueError as error:
```
blackboxsw, line 799: Can we also check that we've actually parsed some non-empty valid lease config after this try/except block?
```
    if not lease:
            msg = "No valid DHCP lease configuration found in dhcpcd lease"
            LOG.error(msg)
            raise NoDHCPLeaseError(msg)
```

If we proceed beyond this with no valid data, we'll only be injecting a worthless `interface: ens5` in our dhcp lease, which does not provide cloud-init with actionable config anyway.

blackboxsw: Can we also add the `lease_dump` to the ValueError `LOG.error` message? The reason being that there is not enough information provided in the error level message beyond the actually dict comprehension traceback to tell us what content we failed to parse.  Even though the source dump_lease content is logged as a `debug` level before, some log aggregators (and cloud-init status --format=json) will ignore debug level messages forcing a triager to jump into the full cloud-init log to better understand the dhcp lease content or absence of content which was unable to be parsed.

tests/unittests/net/test_dhcp.py
```diff
@@ -1199,12 +1199,63 @@ def test_parse_lease_dump(self):
...
         assert "192.168.0.1" == parsed_lease["routers"]
 
+    @pytest.mark.parametrize(
+        "lease, parsed",
+        (
+            pytest.param(
+                """
+                fail
+                """,
+                {},
```
blackboxsw, line 1209: I don't think we want this to succeed if the lease doesn't provide any readable lease content. I think it's more appropriate for us to raise an InvalidDHCPLease error or NoDHCPLeaseError in this case to avoid trying to proceed on completely empty/invalid config.

blackboxsw (changes requested):
Thanks for the improvement to avoid breakage. Can we also take this a step further to do the following:
- log in the ValueError handling the source dump_lease content for reference in the error-level logs
- raise an error when the processed dhcpcd.lease file has no viable parsed  config options

holmanb:
Thanks for the review @blackboxsw. I think that I've addressed all of your comments.

Requesting re-review.

blackboxsw (approved):
LGTM!

PR #5131, opened 2024-04-02
test: fix upgrade dhcp6 on ec2

aciba90:
## Proposed Commit Message

```
After [0, 1], dhcp6 is going to be always false after upgrading cloud-init. Correct this in the integration test.

Refs:
[0] https://github.com/canonical/cloud-init/issues/3980
[1] https://bugs.launchpad.net/cloud-init/+bug/1976526
```

## Additional Context

https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-focal-ec2/411/testReport/junit/tests.integration_tests/test_upgrade/test_clean_boot_of_upgraded_package/

## Test Steps

```
export CLOUD_INIT_PLATFORM=ec2
export CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:cloud-init-dev/daily

tox -e integration-tests -- -vvv tests/integration_tests/test_upgrade.py::test_clean_boot_of_upgraded_package
```

## Checklist

- [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [x] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [x] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #5104, opened 2024-03-26
fix(ec2): Do not enable dhcp6 on EC2

major:
## Proposed Commit Message
```
fix(ec2): Do not enable dhcp6 on EC2

When cloud-init finds any ipv6 information in the instance metadata, it
automatically enables dhcp6 for the network interface. However, this
brings up the instance with a broken IPv6 configuration because SLAAC
should be used for almost all situations on EC2.

Red Hat BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2092459
Fedora Pagure: https://pagure.io/cloud-sig/issue/382
Upstream: https://bugs.launchpad.net/cloud-init/+bug/1976526

Fixes: canonical/cloud-init#3980

Signed-off-by: Major Hayden <major@redhat.com>
```

## Additional Context

* Red Hat BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2092459
* Fedora Pagure: https://pagure.io/cloud-sig/issue/382

## Test Steps

1. Deploy an instance on AWS with a VPC attached that has IPv6 subnets
2. The instance should come online with an IPv6 address configured

## Checklist
- [x] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [x] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [x] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #5132, opened 2024-04-02
focal: Add dhcp6 patch for ec2

holmanb:
## Context
Followup for reverting https://github.com/canonical/cloud-init/pull/5131 and https://github.com/canonical/cloud-init/pull/5104 on old series to avoid breaking change.

## Merge type


blackboxsw (approved):
Changeset is good, d/changelog doesn't have repeated refresh entries, content looks good. Only diff between focal and jammy/mantic is that the new_upstream_snapshot to focal was pulled one commit earlier than jammy/mantic. This isn't a big deal at all because we aren't releasing this and we'll be performing a new_upstream_snapshot again before we release/SRU anyway.

PR #5134, opened 2024-04-02
mantic: Add dhcp6 patch for ec2

holmanb:
## Context
Followup for reverting https://github.com/canonical/cloud-init/pull/5131 and https://github.com/canonical/cloud-init/pull/5104 on old series to avoid breaking change.

## Merge type


debian/patches/revert-dhcp6-on-ec2.patch
```diff
@@ -0,0 +1,134 @@
+--- a/cloudinit/sources/DataSourceEc2.py
```
blackboxsw, line 1: Ideally we want a quilt dep3 header associated with this patch which can be added while editing this quilt patch via `quilt header dep3 -e` before the `quilt pop -a`

Here's suggestion for this patch if you like:
```suggestion

@@ -0,0 +1,143 @@
Description: Revert net config changes for ec2 which dropped dhcp6: True
 To avoid change in behavior on stable releases, retain original net config
 behavior on ec2 which set dhcp6: True and dhcp6-overrides in network config.
 While this is desired upstream behavior on new cloud-init releases, we prefer
 to retain previous behavior to avoid breaking custom environments in Ec2 which
 may have configured their own dhcp service providing IPv6 configuration.
Author: Brett Holman <brett.holman@canonical.com>
Origin: other
Last-Update: 2024-04-02
--- a/cloudinit/sources/DataSourceEc2.py
```

blackboxsw (approved):
LGTM. minor request fora dep3 header in the quilt patch with metadata of your choosing

PR #5126, opened 2024-04-01
azure: parse ProvisionGuestProxyAgent as bool

KsenijaS:
Parse ProvisionGuestProxyAgent  property as bool instead of string.

PR #5138, opened 2024-04-03
azure: check azure-proxy-agent status

KsenijaS:
## Proposed Commit Message

```
feat(azure): check azure-proxy-agent status

Azure Guest Proxy Agent is a new feature in Azure that offers a key exchange protocol
to secure communication between guest and host.

This PR checks the status of Azure Guest Proxy Agent and in case of an error it fails provisioning.

```

## Additional Context

This PR depends on my previous PR https://github.com/canonical/cloud-init/pull/5126
## Checklist

## Merge type


cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,35 @@ def _is_ephemeral_networking_up(self) -> bool:
             or self._ephemeral_dhcp_ctx.lease is None
         )
 
+    def _check_azure_proxy_agent_status(self) -> None:
+        try:
+            cmd = [
+                "azure-proxy-agent",
```
holmanb, line 560: What is this and why do you propose that cloud-init calls it? 

Please add a docstring to this function explaining the purpose of this.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,35 @@ def _is_ephemeral_networking_up(self) -> bool:
...
+                "--status",
+                "--wait",
+                "120",
+            ]
+            (out, err) = subp.subp(cmd, capture=True)
```
holmanb, line 565: parens aren't needed, nor is capture=True

tests/unittests/sources/test_azure.py
```diff
@@ -3799,6 +3802,171 @@ def test_no_pps(self):
         # Verify dmesg reported via KVP.
         assert len(self.mock_report_dmesg_to_kvp.mock_calls) == 1
 
+    def test_no_pps_gpa(self):
```
holmanb, line 3805: Please add a docstring on this test and the next one which describes what is being tested.

holmanb (changes requested):
Can you please add some context to this change?

A quick grep of the cloud-init codebase, the walinuxagent codebase and the first page of google doesn't show any hits for "azure-proxy-agent".

Which package is it installed by? Will this package need to be a new cloud-init requirement? Is it already installed in all of the distro images on Azure?

holmanb:
> > Can you please add some context to this change?
> > A quick grep of the cloud-init codebase, the walinuxagent codebase and the first page of google doesn't show any hits for "azure-proxy-agent".
> > Which package is it installed by? Will this package need to be a new cloud-init requirement? Is it already installed in all of the distro images on Azure?
> 
> This PR depends on my previous one: #5126

That doesn't help me understand what this is for and what it does.

KsenijaS:
> > > Can you please add some context to this change?
> > > A quick grep of the cloud-init codebase, the walinuxagent codebase and the first page of google doesn't show any hits for "azure-proxy-agent".
> > > Which package is it installed by? Will this package need to be a new cloud-init requirement? Is it already installed in all of the distro images on Azure?
> > 
> > 
> > This PR depends on my previous one: #5126
> 
> That doesn't help me understand what this is for and what it does.

Azure GuestProxyAgent is a new feature in Azure that offers key exchange protocol to secure communication between guest and host. It should come pre-installed on Azure endorsed images.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,38 @@ def _is_ephemeral_networking_up(self) -> bool:
...
+                "--wait",
+                "120",
+            ]
+            out, err = subp.subp(cmd)
+            report_diagnostic_event(
+                "Running azure-proxy-agent",
+                logger_func=LOG.info,
+            )
+            if err:
+                report_diagnostic_event(
```
cjp256, line 571: consolidate this with the above diagnostic

tests/unittests/sources/test_azure.py
```diff
@@ -4487,6 +4654,37 @@ def test_imds_failure_results_in_provisioning_failure(self):
         assert len(self.mock_kvp_report_success_to_host.mock_calls) == 0
 
 
+class TestCheckAzureProxyAgent:
+    @pytest.fixture(autouse=True)
+    def proxy_setup(self, azure_ds, mock_subp_subp):
+        self.azure_ds = azure_ds
```
cjp256, line 4660: since all the tests use caplog too, perhaps include it here too?

tests/unittests/sources/test_azure.py
```diff
@@ -4487,6 +4654,37 @@ def test_imds_failure_results_in_provisioning_failure(self):
...
+        self.mock_subp_subp = mock_subp_subp
+
```
cjp256, line 4662: consider adding a mock for the failure report and verify the reports are as expected

tests/unittests/sources/test_azure.py
cjp256: be sure to link to https://github.com/azure/guestproxyagent

cloudinit/sources/DataSourceAzure.py
KsenijaS: done

tests/unittests/sources/test_azure.py
KsenijaS: done

cloudinit/sources/DataSourceAzure.py
KsenijaS: done

tests/unittests/sources/test_azure.py
KsenijaS: done

tests/unittests/sources/test_azure.py
KsenijaS: done

cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,35 @@ def _is_ephemeral_networking_up(self) -> bool:
             or self._ephemeral_dhcp_ctx.lease is None
         )
 
+    def _check_azure_proxy_agent_status(self) -> None:
+        """check if azure-proxy-agent is fully running and can procede to
```
cjp256, line 558: nit: would encourage following https://peps.python.org/pep-0257/ and https://peps.python.org/pep-0008/#documentation-strings

and do a short one-liner followed by a longer paragraph.

"""Check if azure-proxy-agent is ready for communication to Wireserver/IMDS.

If ProvisionGuestProxyAgent is true, query azure-proxy-agent status, waiting up to 120 seconds for the proxy to negotiate with Wireserver and configure an eBPF proxy.  Once azure-proxy-agent is ready, it will exit with code 0 and cloud-init can then expect to be able to communicate with these services.

Fail deployment if azure-proxy-agent is not found or otherwise returns an error.

For more information, check out:
https://github.com/azure/guestproxyagent
"""
```

File: cloudinit/sources/DataSourceAzure.py [hunk 8]
```diff
@@ -554,6 +554,35 @@ def _is_ephemeral_networking_up(self) -> bool:
...
+                "--wait",
+                "120",
+            ]
+            out, err = subp.subp(cmd)
+            report_diagnostic_event(
+                "Running azure-proxy-agent %s resulted"
+                "in stderr output: %s with stdout: %s" % (cmd, err, out),
+                logger_func=LOG.debug,
+            )
+
```
cjp256, line 575: nit: remove the extra lines around except blocks for consistency with rest of module

tests/unittests/sources/test_azure.py
```diff
@@ -3729,13 +3748,101 @@ def provisioning_setup(
...
+        assert self.wrapped_util_write_file.mock_calls == []
+        assert self.patched_reported_ready_marker_path.exists() is False
+
+        # Verify reports via KVP.
+        assert len(self.mock_kvp_report_failure_to_host.mock_calls) == 0
+        assert len(self.mock_azure_report_failure_to_fabric.mock_calls) == 0
+        assert len(self.mock_kvp_report_success_to_host.mock_calls) == 1
+
+        # Verify dmesg reported via KVP.
+        assert len(self.mock_report_dmesg_to_kvp.mock_calls) == 1
```
cjp256, line 3823: would be good to add assert for self.mock_subp_subp.mock_calls comparable to test_no_pps_gpa

tests/unittests/sources/test_azure.py
```diff
@@ -4487,6 +4679,65 @@ def test_imds_failure_results_in_provisioning_failure(self):
...
+            subp.ProcessExecutionError(
+                cmd=["failed", "azure-proxy-agent"],
+                stdout="test_stdout",
+                stderr="test_stderr",
+                exit_code=4,
+            ),
+        ]
+        self.azure_ds._check_azure_proxy_agent_status()
+        assert "azure-proxy-agent status failure" in self.caplog.text
+        exception = subp.ProcessExecutionError(
```
cjp256, line 4727: can this exception be reused for the side_effect?

holmanb:
> It should come pre-installed on Azure endorsed images.

Which images have this dependency installed?

> This PR checks if agent works correctly, and if not it fails provisioning.

Do I understand correctly that after this change, cloud-init will be expected to fail if this dependency is not installed or does not work correctly?

cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,35 @@ def _is_ephemeral_networking_up(self) -> bool:
...
+        """check if azure-proxy-agent is fully running and can procede to
+        communicate with IMDS and Wire Server. If not, fail provisioning.
+        Azure Guest Proxy Agent source code:
+        https://github.com/azure/guestproxyagent"""
+        try:
+            cmd = [
+                "azure-proxy-agent",
+                "--status",
+                "--wait",
+                "120",
```
holmanb, line 567: The project doesn't have much for documentation, so without reading the code I can only assume that this waits for the service to start and then reports a status. Is that correct?

holmanb: This is written in a way that waits for an external daemon that isn't explicitly or implicitly [ordered to start prior to cloud-init](https://github.com/Azure/GuestProxyAgent/blob/66d12429911002eb2bae710cd85f5c66e202d253/proxy_agent_setup/src/linux/GuestProxyAgent.service#L5). This means that, as written cloud-init may spend time waiting for the init system to initiate this daemon - making cloud-init slower for no reason.

Cloud-init already gets blamed for being slow (when measured via `systemd-analyze {blame,critical-chain}`) despite a significant amount of time being spent waiting for external commands and daemons to start. Since the init system is capable of providing ordering, please use those primitives rather than forcing cloud-init to needlessly wait on systemd to start a service and for that service to become ready.

cjp256: The plan is to start proxy agent prior to cloud-init.  The challenge is that it needs networking to communicate with Wireserver.  For cloud-init, the challenge is that access to IMDS/Wireserver may be restricted until proxy-agent performs its function.

The options considered:

1. We can order it prior to cloud-init local service and have it bring-up networking to do its check and tear it back down.  Cloud-init local won't start until agent service is complete.

- Time won't count against cloud-init

- Will likely increase boot times overall as it duplicates efforts made by cloud-init

2. We can order it prior to, or in parallel to, cloud-init local and allow cloud-init to bring up ephemeral networking then ask it for status.

- This allows the service to come up and use this as a signal that networking is ready

- Doesn't introduce ephemeral networking into proxy agent

3. Cloud-init does nothing at all.  If the proxy agent is required for connectivity to IMDS, existing retry logic for fetching instance metadata is relied on (up to 5 minutes) with the hopes that the agent is running and functional.  In the broken cases, we'll spin for 5 minutes and then fail provisioning.

- Worst case failure handling, but no changes required for cloud-init.

cjp256:
> > Which images have this dependency installed?
> 
> Ubuntu does not install this at this time. It was just brought to the Canonical Public Cloud's team attention a few weeks ago, as a replacement for logic current in `walinuxagent`

No-one has packaged it yet as it's under development and still in private preview phase.  There is no equivalent logic in walinuxagent, Perhaps you are mistaking it for azure-init?

> 
> @cjp256 Last I had heard, this package was not ready to be installed or used. It also isn't packaged yet for Ubuntu, so i think we have a bit of an ordering issue. Could you come to the next Canonical / Microsoft LSG sync and fill us in on current status?

Absolutely.

Keep in mind this is a feature that's disabled by default and would only be enabled in a limited manner for the short-term.  The expectation is that anyone using this feature would be required to use a custom image with this package installed.  This PR is primary intended to improve the failure handling when the feature is enabled and either (a) proxy agent is not installed which will result in failure or (b) proxy agent is malfunctioning for whatever reason.

cloudinit/sources/DataSourceAzure.py
holmanb: Thanks for the response @cjp256. 

>  For cloud-init, the challenge is that access to IMDS/Wireserver may be restricted until proxy-agent performs its function.

This is a pretty big change in requirements for cloud-init to adapt to. That explains why this was proposed, thanks!

holmanb: I still wonder if there are other options that might meet both needs. I'm also hesitant due to potential future ordering issues. The nice thing about keeping ordering in the init system is that systemd transactions prevent deadlocks - a service gets booted from the transaction when an ordering cycle occurs (proactive) rather than trying  (reactively) to understand why cloud-init blocks forever. I'm less concerned about short term ordering and more concerned about what will happen in the future. New projects have a habit of growing scope after adoption. 

For example, cloud-init could grow more systemd services to allow more nuanced ordering. If we were to break cloud-init into more stages, we could keep ordering in the init system while preventing time from counting against cloud-init and yet not increasing boot time.

Hypothetical Example:

1. cloud-init and proxy agent start in parallel
2. cloud-init brings up ephemeral network, then publishes completion through the init system (ex: `cloud-ephemeral.service`)
3. a proxyagent service is triggered which executes the wait command (ex: `azure-proxy-agent-wait-online.service`)
4. completion of the proxy wait service triggers a cloud-init service to continue with network (ex: `cloud-network.service`) - in the case that no `azure-proxy-agent-wait-online.service` is in the boot order, no blocking would occur 
5. when complete would trigger the end of the existing `cloud-init-local.service`

The obvious drawback to this proposal is added complexity.

Obviously Python module load time is slow, which would make the above seem unappealing for performance, however I have a proof of concept that I may try to add to cloud-init in the next cycle which improves the performance cloud-init (and the entire boot) runtime by avoiding module reload while retaining existing systemd services[1].

[1] It works by changing cloud-init to start once and communicates start/stop events with the current cloud-init systemd services over unix sockets. Exposing further synchronization points using the groundwork that this will lay would be straightforward.

holmanb:
@KsenijaS @cjp256 is there any plan for GuestProxyAgent to support the BSDs? I only see support advertised for Linux 5.15+ and Windows. 

cc @igalic

igalic:
WALinuxAgent's FreeBSD support us rather spotty, and i haven't had time to dedicate to fixing it yet. It would be nice if we caught this one earlier

KsenijaS:
Can you please reopen this PR? This is not a default feature and it gracefully handle the case when proxy agent is not installed.
Is there something else we need to do for this PR to be merged in the next release cycle?

blackboxsw:
Adding this to milestone 24.3 to ensure we track and land this by next release. We will be uploading this to Ubuntu 24.10 (Oracular) devel images prior to our 24.3 upstream release date to give time to validate behavior

cjp256:
> Adding this to milestone 24.3 to ensure we track and land this by next release. We will be uploading this to Ubuntu 24.10 (Oracular) devel images prior to our 24.3 upstream release date to give time to validate behavior

Thanks @blackboxsw !

holmanb:
@KsenijaS @cjp256 What is the expected timeline of the proxy agent's availability in distro repositories?

>  This is not a default feature and it gracefully handle the case when proxy agent is not installed.

Not default because currently azure-proxy-agent is not enabled in the ovf-env.xml file, correct?

Currently if the proxy agent is not installed and the ovf-env.xml file enables azure-proxy-agent, a `self._report_failure(reportable_error)` is called, which would cause provisioning to fail, right?

Under what circumstances does azure plan to enable this feature? Can this feature be different per ami? by distro type?

cjp256:
> @KsenijaS @cjp256 What is the expected timeline of the proxy agent's availability in distro repositories?
> 

There is work being done right now to get it ready for submission to Debian, Fedora, etc.  It is currently available in Azure Linux.

There are preview versions currently available in packages.microsoft.com for Ubuntu, for ex:
https://packages.microsoft.com/repos/microsoft-ubuntu-focal-prod/pool/main/a/azure-proxy-agent/
https://packages.microsoft.com/repos/microsoft-ubuntu-jammy-prod/pool/main/a/azure-proxy-agent/

> > This is not a default feature and it gracefully handle the case when proxy agent is not installed.
> 
> Not default because currently azure-proxy-agent is not enabled in the ovf-env.xml file, correct?
> 

Correct.  If a customer turns on this feature while creating the VM, the flag will be set to true to indicate this is enabled.

> Currently if the proxy agent is not installed and the ovf-env.xml file enables azure-proxy-agent, a `self._report_failure(reportable_error)` is called, which would cause provisioning to fail, right?
> 

Correct.  At that point the customer's request to create the VM should return with the reported failure.

> Under what circumstances does azure plan to enable this feature? Can this feature be different per ami? by distro type?

It will be configurable for all instances.  There is currently no plan to gate this feature to particular set of images or vm sizes.  The expectation is the customer can customize an image to include the necessary dependencies.

This PR is mostly about improving the failure experience (i.e. azure-proxy-agent not found) vs some more obscure failure (e.g. 401 http code when connecting to IMDS, etc.).  It also prevents a race condition by giving proxy agent a chance to finish configuration before proceeding with calls to imds/wireserver.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -633,6 +670,10 @@ def crawl_metadata(self):
 
         imds_md = {}
         if self._is_ephemeral_networking_up():
+            # check if azure-proxy-agent is enabled in the ovf-env.xml file.
```
blackboxsw, line 673: Let's comment that this is disabled by default, and customer opt-in please.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -554,6 +554,43 @@ def _is_ephemeral_networking_up(self) -> bool:
...
+                logger_func=LOG.debug,
+            )
+        except FileNotFoundError:
+            error = errors.ReportableErrorProxyAgentNotFound()
+            self._report_failure(error)
+        except subp.ProcessExecutionError as error:
+            reportable_error = errors.ReportableErrorProxyAgentStatusFailure(
+                error
+            )
+            self._report_failure(reportable_error)
```
blackboxsw, line 592: What I don't particularly like about this logic, is that there appears to be nothing reported currently to me on stdout or stderr when this command exits 1 on unsupported environments (lxd containers using the preliminary jammy deb package linked)  . Can we add a sample out/err string that we expect to the existing unit tests so we can see how this failure expectation looks due to lack of response from the wireserver?

cloudinit/sources/azure/errors.py
```diff
@@ -195,3 +195,17 @@ def __init__(self, exception: Exception) -> None:
 
         self.supporting_data["exception"] = repr(exception)
         self.supporting_data["traceback_base64"] = trace_base64
+
+
+class ReportableErrorProxyAgentNotFound(ReportableError):
+    def __init__(self) -> None:
+        super().__init__("azure-proxy-agent not found")
```
blackboxsw, line 202: Given that no official debian/ubuntu or fedora images yet have a published version of this package, let's provide a bit more context in the error message that tells admin's what triggered this attempt, and 


```suggestion
        super().__init__(
            "Unable to activate ProvisionGuestProxyAgent."
            " azure-proxy-agent package not found"
        )
```

blackboxsw (commented):
Just a couple of minor nits about @holmanb's review.

blackboxsw:
Targetting to milestone 24.2 if we are able to get any outstanding review comments done By Wed June 25th

cloudinit/sources/DataSourceAzure.py
KsenijaS: Hey, I wonder if I understand correctly. In addition to the remote logging, which could fail on unsupported environments, and the cloud init logs which would work even on unsupported environments, you'd like us to log in stderr?

I'm also not sure about the testing part. Do you want this logging to happen only for the sake of the unit test? If so, I could mock/assert the cloud init logging logic, although we have unit tests for it.

Thank you for your time!

cloudinit/sources/DataSourceAzure.py
KsenijaS: done

cloudinit/sources/azure/errors.py
KsenijaS: done

blackboxsw (approved):
I'm good with this approach as preliminary support for a tool that has yet to be released in Linux images. It is opt-in only and provides fallbacks in this event that the necessary command is unavailable in the environment and logs at debug severity level in the event of failures in using the proxy agent.

Minor ask on the example unittest stdout/stderr for a known or likely failure condition, but otherwise +1.

tests/unittests/sources/test_azure.py
```diff
@@ -3829,11 +3929,96 @@ def test_no_pps(self):
...
 
+    def test_no_pps_gpa_fail(self):
+        """test full provisioning scope when azure-proxy-agent is enabled and
+        throwing an exception during provisioning."""
+        self.mock_subp_subp.side_effect = [
+            subp.ProcessExecutionError(
+                cmd=["failed", "azure-proxy-agent"],
+                stdout="test_stdout",
+                stderr="test_stderr",
+                exit_code=4,
```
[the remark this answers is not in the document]
cjp256: The --status error outputs _do still_ need some work as you indicated.  You are not alone in wishing for some better outputs, but they are going to give it more attention in the near future and we'll update this accordingly.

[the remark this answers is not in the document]
cjp256: fwiw I also get no output and code=1 if I check status as non-root.

blackboxsw:
> Thanks @blackboxsw ! Do you know if the proxy agent will be available for 24.10 too or are you just referring to cloud-init?

For clarity, this old comment was only referring to cloud-init published to 24.10. I believe there may be separate efforts to get the proxy agent into a releasable state in various downstream images in Ubuntu, but I have no knowledge of whether those efforts are in earnest for 24.10.

tests/unittests/sources/test_azure.py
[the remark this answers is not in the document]
blackboxsw: > but they are going to give it more attention in the near future and we'll update this accordingly.

I am ok with cloud-init adding this opt-in feature for a pre-release visibility to better enable testing and ongoing development of this feature on Azure. Given that the tooling features and functionality are not yet final, we can come back to this work in the future to provide more extensive unit testing or integration testing once we have images which contain the proxy agent.

cloudinit/sources/DataSourceAzure.py
blackboxsw: I'm ok with this approach and feel our concerns with cloud-init being blamed for being slow due to any agent-status blocking would be easy to determine from cloud-init logs as our subp call adds a suffix "took <X.Yseconds> to run" for any command that takes > 0.1 seconds. So, if there is a blocking command that holds us up on Azure, we can point the finger if need be from logs with our current subp logging without need for a secondary systemd service in this case.

tests/unittests/sources/test_azure.py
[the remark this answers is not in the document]
blackboxsw: If this does show to a be a frequent issue, we can invest more in a more comprehensive approach to systemd ordering of this proxy-agent service. I think folks had already settled on this approach anyway, but let's see where this development leads for this proxy agent before adding a bit more complexity.

PR #5139, opened 2024-04-03
Release bug-fix only cherry-picks from 24.1.4 upstream release

blackboxsw:
Sync applicable bug fixes from 24.1.4 to ubuntu/noble-24.1.x for release.

We are in beta freeze for Ubuntu noble, and we want to limit changes published to noble to Ubuntu bug-fixes only.
So this PR is a cherry-picked set of changes from 24.1.4 with the exclusion of the commit a6e09d9bfd3ef0770bc4d9452d978ba93dbe1245 which is  run_dir support for BSD and not applicable to Ubuntu images.

## process to create this branch:
```
git fetch upstream
git checkout upstream/ubuntu/noble-24.1.x -B ubuntu/noble-24.1.x

#  Pull in 7 of the 8 commits in 24.1.4 into noble-24.1.x branch as cpick files
for hash in accdfe6a8b2ffff82731c8bae2d672ff7c3f0974 144782a838123099334fe44c07566b07503fd67b f74b61eae24706dd313efc06792198a718f06672 8f806e20bda1bdf0e00ff180183990470e241837 372e80f8e6e01441e69d277782e2367678994abb 77771023ab5bda43a25cdc3dd4eb69a79169b3cb f6ac6ee8a4321ba959ae8f3ae9f1b9ac3b9fafd0; do cherry-pick $hash; done

# Manually update d/changelog to the right package version `24.1.3-0ubuntu2` and refactor the d/changelog entry to represent 
# functional bug fixes vs test fixes
git rebase -i HEAD~10 # collapse all  `update changelog` commits
dch -r -D noble ''
 git commit -am 'releasing cloud-init version 24.1.3-0ubuntu2'
```
## Proposed Commit Message [REBASE merge]
See individ commits.
## Merge type

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,26 @@
+cloud-init (24.1.3-0ubuntu2) UNRELEASED; urgency=medium
+
+  * Upstream bug fix release based on 24.1.4.
+    + functional fixes in debian/patches:
+      - cpick-f74b61ea-ephemeral-dhcpcd-Set-dhcpcd-interface-down
+        When dhcpcd client sets up a device to discover IMDS, ensure
+        link is set back down to avoid interface rename failures (GH-5019)
+      - cpick-372e80f8-net-dhcp-bump-dhcpcd-timeout-to-300s-5127
+        Extend default timeout of dhcpcd command to 300 seconds
+      - cpick-77771023-net-dhcp-raise-InvalidDHCPLeaseFileError-on-error
+        Raise any dhcpcd lease parse failures as InvalidDHCPLeaseFileError
+        (GH-5128)
+      - cpick-f6ac6ee8-fix-dhcpcd-Make-lease-parsing-more-robust-5129
+        Avoid tracebacks on empty invalid lease content from dhcpcd
+        and raise InvalidDHCPLeaseFileError with full lease parsed when errors
+        occur. (GH-5128)
+    + test fixes in debian/patches:
+      - d/p/cpick-accdfe6a-refactor-Import-log-module-rather-than-functions-5074
+        Avoid mock failures in unit tests by module-level import of log
+      - cpick-144782a8-test-Remove-side-effects-from-tests-5074
+
+ -- Chad Smith <chad.smith@canonical.com>  Wed, 03 Apr 2024 12:09:38 -0600
+
 cloud-init (24.1.3-0ubuntu1) noble; urgency=medium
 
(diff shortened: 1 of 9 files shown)


holmanb:
Why use cpicks rather than cherry-picking them directly onto this branch? Are you just trying to avoid uploading a new orig.tar.gz? or is there something else?

blackboxsw:
> Why use cpicks rather than cherry-picking them directly onto this branch? Are you just trying to avoid uploading a new orig.tar.gz? or is there something else?

@holmanb, exactly. Since we are not generating an official new cloud-init-24.1.4.orig.tar.gz due to our avoidance of the run_dir(BSD) changes, our package delta needs to be represented in quilt patches in debian/patches just as we generally do for other stable downstream releases which are derived from typical upstream releases, but have ubuntu series-specific cherry-picks of modifications to revert upstream behavior.  I was trying to avoid uploading that extra new clud-init-24.1.3.2.orig.tar.gz as it's not an official upstream release. We are just patching the previous released 23.1.3.orig.tar.gz and instead providing a simple packaging delta with quilt cpicks against the already released 24.1.3.orig.tar.gz.

debian/changelog
```diff
@@ -1,3 +1,26 @@
+cloud-init (24.1.3-0ubuntu2) noble; urgency=medium
+
+  * Upstream bug fix release based on 24.1.4.
```
holmanb, line 3: Think this will raise eyebrows given the upstream version is based on 24.1.3?

holmanb (approved):
One nit, but LGTM either way!

debian/changelog
blackboxsw: I think we are good here as we are saying it's `Upstream bug fix release` not `Upstream snapshot release`. The difference being that we are only pulling in bug fixes, not any features from 24.1.4

blackboxsw:
Validated integration test run on noble experienced only the expected current failures:
 - modules/test_apt_functionality.py # issues finding gpg keys in integration tests due to unfixed/published gpg 2.4.4 which relies on keyboxd and doesn't list-keys from keyrings to stdout. This should be resolved once gpg package publishes to noble daily images
 - modules/test_package_update_upgrade_install  SIGTERM  due to needrestart interrupting cloud-final.service preventing the service from completing all it's APT operations. This is tracked by https://bugs.launchpad.net/ubuntu/+source/needrestart/+bug/2059337

PR #5140, opened 2024-04-03
fix: Logging sensitive data

holmanb:
```
fix: Don't log sensitive data

Since /var/log/cloud-init.log is a priviledged file, this
does not expose a secure system (no CVE). However, we don't
want to log this information so that users can file reports
without having to manually redact logs.

Standardize log messages so that redacted and non-redacted logs match.
```


blackboxsw (approved):
LGTM, thanks for the cleanup.

PR #5142, opened 2024-04-03
chore(safeyaml): Remove legacy Python 2 unicode helper

holmanb:
## Proposed Commit Message
```
chore(safeyaml): Remove legacy Python 2 unicode helper

PyYAML has built-in unicode support in Python3+.

The original code[1] was added as a helper to add
support for unicode to `yaml.safe_load()`. We
don't need this anymore, and can jettison it and
prefer `yaml.safe_load()`.

[1] a7a9de1a079
```

Note that the unittests that were added with this helper back in the python2 days still pass.

Rational: custom code that offers nothing over library code is an unnecessary complexity, remove it.

### DIFF
--- a/cloudinit/cmd/devel/net_convert.py
+++ b/cloudinit/cmd/devel/net_convert.py
@@ -115,7 +115,7 @@ def handle_args(name, args):
     if args.kind == "eni":
         pre_ns = eni.convert_eni_data(net_data)
     elif args.kind == "yaml":
-        pre_ns = safeyaml.load(net_data)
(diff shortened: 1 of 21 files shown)


CalvoM:
Tested with tests and they all passed. I will be testing the package built.

holmanb:
> Tested with tests and they all passed. I will be testing the package built.

Thanks for the review @CalvoM!

TheRealFalcon (approved):
LGTM!

Also...jettison is a great verb :smile:

PR #5144, opened 2024-04-04
fix(url_helper): fix TCP connection leak on readurl() retries

cjp256:
In scenarios where a lot of retries are expected, Ubuntu 24.04 fails regularly with "Too many open files".  The `ulimit -n` shows the same number of allowed open files in Ubuntu 20.04 (1024), but the connections must be holding out longer on 24.04 for some reason. As the retry count gets close to 1024 in readurl(), the open file limit is hit and exceptions sprout up in a number of places.

The leak appears to occur because the same session is reused and all prior failures are recorded in excps[] which in turn appears to keep the connection open.

- create a new Session() for each connection attempt if session was not passed in explicitly

- raise exceptions immediately when required rather than saving them to do outside of the exception handler

### DIFF
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -242,6 +242,7 @@ def readurl(
     :param stream: if False, the response content will be immediately
     downloaded.
     """
+    create_sessions = session is None
     url = _cleanurl(url)
     req_args = {
         "url": url,
@@ -281,11 +282,9 @@ def readurl(
     if sec_between is None:
         sec_between = -1
 
-    excps = []
     # Handle retrying ourselves since the built-in support
     # doesn't handle sleeping between tries...
-    # Infinitely retry if infinite is True
-    for i in count() if infinite else range(manual_tries):
+    for i in count():
         req_args["headers"] = headers_cb(url)
         filtered_req_args = {}
         for (k, v) in req_args.items():
@@ -300,7 +299,6 @@ def readurl(
             else:
                 filtered_req_args[k] = v
         try:
-
             if log_req_resp:
                 LOG.debug(
                     "[%s/%s] open '%s' with %s configuration",
@@ -310,7 +308,7 @@ def readurl(
                     filtered_req_args,
                 )
 
-            if session is None:
+            if create_sessions:
                 session = requests.Session()
 
             with session as sess:
@@ -329,6 +327,10 @@ def readurl(
             # subclass for responses, so add our own backward-compat
             # attrs
             return UrlResponse(r)
+        except exceptions.SSLError as e:
+            # ssl exceptions are not going to get fixed by waiting a
+            # few seconds
+            raise UrlError(e, url=url) from e
         except exceptions.RequestException as e:
             if (
                 isinstance(e, (exceptions.HTTPError))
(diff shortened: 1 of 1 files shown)


cjp256:
Ubuntu 24.04 reproducer:
```
$ python3 -c "from cloudinit.url_helper import readurl; readurl('http://191.235.228.36/fail', infinite=True)" & sleep 30 && sudo lsof -n  | grep 191.235.228.36 | wc -l
[1] 1590
20
```

Versus Ubuntu 20.04:
```
$ python3 -c "from cloudinit.url_helper import readurl; readurl('http://191.235.228.36/fail', infinite=True)" & sleep 30 && sudo lsof -n  | grep 191.235.228.36 | wc -l
[1] 13516
0
```

cloudinit/url_helper.py
```diff
@@ -310,7 +308,7 @@ def _cb(url):
                     filtered_req_args,
                 )
 
-            if session is None:
+            if create_sessions:
```
cjp256, line 311: I actually need to come back to this and re-use the session, but kill the subsequent context manager (re-use) which seems to break on 24.04

cjp256:
This test quickly shows the difference in behavior between 20.04 and 24.04 without cloud-init in the picture:

This script run fine on Ubuntu 20.04 but not 24.04.

```
import requests
import subprocess
import time

WITH_CONTEXT_MANAGER = True
SAVE_RESPONSES = True

def print_connections(i):
    lines = subprocess.run(["sudo", "lsof", "-n"], capture_output=True, text=True).stdout.splitlines()
    conns = [c for c in lines if "191.235.228.36" in c]
    print(f"{i}:")
    for conn in conns:
      print(" -", conns)


session = requests.Session()
req_args = {'url': 'http://191.235.228.36/fail', 'stream': False, 'allow_redirects': True, 'method': 'GET', 'headers': {'User-Agent': 'Cloud-Init/24.1-0ubuntu1'}}
responses = []

for i in range(1, 10000):
    if WITH_CONTEXT_MANAGER:
        with session as sess:
            r = sess.request(**req_args)
    else:
        r = session.request(**req_args)

    if SAVE_RESPONSES:
        responses.append(r)
    print_connections(i)
    time.sleep(0.1)
```

cloudinit/url_helper.py
cjp256: done

anhvoms (approved):
LGTM

holmanb:
Huge thanks for this work @cjp256. This looks good to me.

I think that there is more work to do here, including reusing TCP connections, but for a bugfix PR this is perfect - thanks for the cleanup too.

Using the reproducer and I can see that this reproduces on Noble and Mantic, and does not on Focal.

On Mantic, I see that this reproduces even with cloud-init 23.3.1.

On Noble hosts, this does not reproduce in a focal lxc container, but does with mantic and noble. I think that this tells us that the change in behavior is somewhere in the software stack - Python or requests probably.

I'd like to get test coverage for this into cloud-init too, but I'm not having success with a pure-python solution, so I may end up creating an integration test that does something like one of your reproducers.

holmanb (approved):

PR #5153, opened 2024-04-05
Another fix for noble

holmanb:
## Additional Context
This fixes an occasionally seen issue on long-booting instances that take many retries - it has been seen since mantic was released and appears to be due to a change in Python or requests.

## Test Steps
```
python3 -c "from cloudinit.url_helper import readurl; readurl('http://191.235.228.36/fail', infinite=True)" & sleep 30 && sudo lsof -n  | grep 191.235.228.36 | wc -l
```
This shows higher numbers on reproducible systems and 0 or 1 on non-reproducing systems. 

Note: the kernel doesn't immediately close sockets when the process that opens the socket is killed or exits. This makes testing for this behavior difficult.

## Merge type

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+cloud-init (24.1.3-0ubuntu3) UNRELEASED; urgency=medium
+
+  * cherry-pick 516fad6d: fix(url_helper): fix TCP connection leak on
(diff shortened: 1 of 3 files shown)


TheRealFalcon (approved):
LGTM!

PR #5147, opened 2024-04-04
docs: Add breaking changes section to reference docs

catmsred:
## Proposed Commit Message
```
docs: Add breaking changes section to reference
```

## Additional Context
In order to have snapd seeding changes to reference from the noble release docs

## Checklist

- [ ] My code follows the process laid out in [the documentation](https://cloudinit.readthedocs.io/en/latest/development/index.html)
- [ ] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly
- [x] I have updated or added any [documentation](https://cloudinit.readthedocs.io/en/latest/development/contribute_docs.html) accordingly

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #5165, opened 2024-04-10
docs: Add DataSourceNone documentation

TheRealFalcon:
## Proposed Commit Message

```
docs: Add DataSourceNone documentation
```

## Merge type

### DIFF
--- a/doc/rtd/reference/datasources/nocloud.rst
+++ b/doc/rtd/reference/datasources/nocloud.rst
@@ -15,8 +15,8 @@ Configuration Methods:
 .. warning::
     User data placed under ``/etc/cloud/`` will **not** be recognized as a
     source of configuration data by the NoCloud datasource. While it may
-    be acted upon by cloud-init, this use case may be better served
-    by :ref:`DataSourceNone<datasource_none_example>`.
+    be acted upon by cloud-init, using
+    :ref:`DataSourceNone<datasource_none_example>` should be preferred.
 
 Method 1: Local filesystem, labeled filesystem
 ----------------------------------------------
--- a/doc/rtd/reference/datasources/nocloud.rst
+++ b/doc/rtd/reference/datasources/nocloud.rst
@@ -12,6 +12,12 @@ provide configurations.
 Configuration Methods:
 ======================
 
+.. warning::
+    User data placed under ``/etc/cloud/`` will **not** be recognized as a
+    source of configuration data by the NoCloud datasource. While it may
+    be acted upon by cloud-init, this use case may be better served
+    by :ref:`DataSourceNone<datasource_none_example>`.
+
 Method 1: Local filesystem, labeled filesystem
 ----------------------------------------------
 
--- a/doc/rtd/reference/datasources/none.rst
+++ b/doc/rtd/reference/datasources/none.rst
@@ -1,0 +1,51 @@
+.. _datasource_none:
+
+None
+****
+
+The data source ``None`` may be used when no other viable datasource is
+present on disk. This has three primary use cases:
+
+1. Testing cloud-init in an environment where no datasource is
+   available.
+2. As a fallback for when a datasource is otherwise intermittently
+   unavailable.
+3. Providing user data to cloud-init from on-disk configuration when
+   no other datasource is present.
+
+When the datasource is ``None``, cloud-init is unable to obtain or
+render networking configuration. Additionally, when cloud-init
+completes, a warning is logged that DataSourceNone is being used.
+
+Configuration
+=============
+
+User data and meta data may be passed to cloud-init via system
+configuration in :file:`/etc/cloud/cloud.cfg` or
+:file:`/etc/cloud/cloud.cfg.d/`.
+
+``userdata_raw``
+----------------
+
+A **string** containing the user data (including header) to be used by
+cloud-init.
+
+``metadata``
+------------
+The metadata to be used by cloud-init.
+
+.. _datasource_none_example:
+
+Example configuration
+---------------------
+
+.. code-block:: yaml
+
+  datasource:
+    None:
+      metadata:
+        local-hostname: "myhost.internal"
+      userdata_raw: |
+        #cloud-config
+        runcmd:
+        - echo 'mydata' > /var/tmp/mydata.txt
--- a/doc/rtd/reference/datasources/none.rst
+++ b/doc/rtd/reference/datasources/none.rst
@@ -4,14 +4,12 @@ None
 ****
 
 The data source ``None`` may be used when no other viable datasource is
-present on disk. This has three primary use cases:
+present on disk. This has two primary use cases:
 
-1. Testing cloud-init in an environment where no datasource is
-   available.
+1. Providing user data to cloud-init from on-disk configuration when
+   no other datasource is present.
 2. As a fallback for when a datasource is otherwise intermittently
    unavailable.
-3. Providing user data to cloud-init from on-disk configuration when
-   no other datasource is present.
 
 When the datasource is ``None``, cloud-init is unable to obtain or
 render networking configuration. Additionally, when cloud-init
@@ -22,7 +20,7 @@ Configuration
 
 User data and meta data may be passed to cloud-init via system
 configuration in :file:`/etc/cloud/cloud.cfg` or
-:file:`/etc/cloud/cloud.cfg.d/`.
+:file:`/etc/cloud/cloud.cfg.d/*.cfg`.
 
 ``userdata_raw``
 ----------------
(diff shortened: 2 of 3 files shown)


doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -12,6 +12,12 @@ provide configurations.
 Configuration Methods:
 ======================
 
+.. warning::
```
TheRealFalcon, line 15: I don't love putting this here, but I do think we've had enough confusion about this that it should be called out on this page. I'm open to suggestions for a better location.

doc/rtd/reference/datasources/none.rst
```diff
@@ -0,0 +1,51 @@
...
+1. Testing cloud-init in an environment where no datasource is
+   available.
+2. As a fallback for when a datasource is otherwise intermittently
+   unavailable.
+3. Providing user data to cloud-init from on-disk configuration when
+   no other datasource is present.
+
+When the datasource is ``None``, cloud-init is unable to obtain or
+render networking configuration. Additionally, when cloud-init
+completes, a warning is logged that DataSourceNone is being used.
```
holmanb, line 18: > a warning is logged that DataSourceNone is being used.

Is it logged if DatasourceNone is being used, or is it logged if this datasource is used as a fallback? I don't think that this is behavior that we actually want if only [None] is specified by the user, right?

doc/rtd/reference/datasources/none.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Configuration
+=============
+
+User data and meta data may be passed to cloud-init via system
+configuration in :file:`/etc/cloud/cloud.cfg` or
+:file:`/etc/cloud/cloud.cfg.d/`.
```
holmanb, line 25: The files in `cloud.cfg.d/` require a specific file ending that is not a standard *.yaml/*.yml, right? I know ds-identify code does require that, at least.

Do you think that is worth implying with `/etc/cloud/cloud.cfg.d/*.cfg`?

holmanb (changes requested):
Thanks @TheRealFalcon, this is looking great!

I have a couple of nits, and a couple of questions since some of the described behavior seems unintuitive and/or undesireable.

doc/rtd/reference/datasources/none.rst
TheRealFalcon: It happens regardless.

https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_final_message.py#L114

I can create an issue to remove it if None has userdata/metadata or is the only entry in the datasource list.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -12,6 +12,12 @@ provide configurations.
...
+    User data placed under ``/etc/cloud/`` will **not** be recognized as a
+    source of configuration data by the NoCloud datasource. While it may
+    be acted upon by cloud-init, this use case may be better served
```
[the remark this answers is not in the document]
TheRealFalcon: That technically isn't true as it's used as base config and not considered associated with any datasource (unless specifically placed under the datasource definition). I updated the wording though. Let me know if that works for you.

TheRealFalcon:
@holmanb I believe I addressed your comments

doc/rtd/reference/datasources/nocloud.rst
a-dubs: I'm not even 100% clear on what this warning block means? I think this is a fine location for it though. And a warning is definitely worthwhile to provide, especially if its a known confusion point!

catmsred:
Can we also make sure this is reference/linked in the Breaking changes page created https://github.com/canonical/cloud-init/pull/5147 since we've had a lot of people breaking in different ways on this?

TheRealFalcon:
@catmsred , nothing in this PR is new or is referencing behavior that has changed recently. If you're referring to the `datasource_list` changes, I'll file that separately.

TheRealFalcon:
See https://github.com/canonical/cloud-init/pull/5171

doc/rtd/reference/datasources/none.rst
```diff
@@ -0,0 +1,51 @@
...
+****
+
+The data source ``None`` may be used when no other viable datasource is
+present on disk. This has three primary use cases:
+
+1. Testing cloud-init in an environment where no datasource is
+   available.
+2. As a fallback for when a datasource is otherwise intermittently
+   unavailable.
+3. Providing user data to cloud-init from on-disk configuration when
```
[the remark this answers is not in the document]
holmanb: no objections to leaving it as is, that makes sense to me

doc/rtd/reference/datasources/none.rst
holmanb: sounds good to me!

holmanb (approved):
Thanks @TheRealFalcon, this looks good to me!

doc/rtd/reference/datasources/none.rst
TheRealFalcon: https://github.com/canonical/cloud-init/issues/5192

doc/rtd/reference/datasources/nocloud.rst
ani-sinha: @holmanb could you please point me the exact change in 24.1? Some internal user did get caught by surprise by this change you talk about.

zhaohuijuan: @holmanb I have a further question for this change, could you please help to clarify this?
Just as @ani-sinha mentioned, we have some internal user did get caught by surprise by this change you talked about. 

It reports "DataSourceNoCloud.py[DEBUG]: Seed from https://..../userdata/ not supported by DataSourceNocloud" in cloud-init.log with below config in cloud-init-23.4(But no such issue before 23.1):

# cat << EOM > /etc/cloud/cloud.cfg.d/10_datasource.cfg
datasource_list: [NoCloud]
datasource:
  NoCloud:
    seedfrom: https://satellite.example.com/userdata/
EOM


Could I know how to adapt it after the change?
Thanks!

holmanb: @zhaohuijuan @ani-sinha sometimes comments like get missed when commenting on old merged PRs - github notifications are the only way that these messages get notified and since there is no queue for reviewing old comments comments like this are more likely to get lost. IRC would be more likely to get a response.

To answer your question, I don't think that this log message indicates an actual issue:
```

Comment (zhaohuijuan) [reply doc/rtd/reference/datasources/nocloud.rst:15]: Thanks @holmanb for the reminder and quick response. Seems I can not login the IRC #cloud-init ... So asking here.

It indeed broke the cloud-init implementation, not sure if  this log related. 

Per the latest cloud-init doc[1]: User data placed under /etc/cloud/ will not be recognized as a source of configuration data by the NoCloud datasource.
[1] https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html

And the doc provides 4 configuration methods(e.g. command line), should we config NoCloud and seedfrom with one of the 4 methods?

But there is another example for NoCloud config[2], is it conflict with the above doc[1]?
[2] https://cloudinit.readthedocs.io/en/latest/reference/examples.html#configure-data-sources

Comment (holmanb) [reply doc/rtd/reference/datasources/nocloud.rst:15]: > Thanks @holmanb for the reminder and quick response. Seems I can not login the IRC #cloud-init ... So asking here.
> 
> It indeed broke the cloud-init implementation, not sure if  this log related. 

I don't think that this log is related, but it sounds like there may be an issue that should be reported. There have been lots of different changes in this part of the code over the last two years. Without logs of the working and non-working instances this may be difficult to understand.

Can you please file a bug with the relevant logs?

> Per the latest cloud-init doc[1]: User data placed under /etc/cloud/ will not be recognized as a source of configuration data by the NoCloud datasource.
> [1] https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html
>
> And the doc provides 4 configuration methods(e.g. command line), should we config NoCloud and seedfrom with one of the 4 methods?
> 
> But there is another example for NoCloud config[2], is it conflict with the above doc[1]?
> [2] https://cloudinit.readthedocs.io/en/latest/reference/examples.html#configure-data-sources

This sounds like possibly confusion between user-data and system configuration, the doc referenced is not describing user-data I think.

Comment (holmanb) [reply doc/rtd/reference/datasources/nocloud.rst:15]: Ahh, you're referring to this:
```
    # these are optional, but allow you to basically provide a datasource

    # right here

    user-data: |

      # This is the user-data verbatim

    meta-data:

      instance-id: i-87018aed

      local-hostname: myhost.internal
```

Yes, we should probably update the doc wording. You can use this subkey, which is technically defined within cloud.cfg files.

zhaohuijuan: Thanks @holmanb 
I think you already involved in the issue: https://github.com/canonical/cloud-init/issues/5288, thanks for your clarification

PR #5166, opened 2024-04-11
Enable interactively running cloud-init local --local

holmanb:
## Proposed Commit Message
```
fix(dhcp): Enable interactively running cloud-init init --local (#5166)

Check network interface state before and after receiving dhcp lease. Return
link and address to pre-existing state prior to running dhcp client.

While this isn't a supported cloud-init use case, it does make some
testing easier.
      
- add function typing and a docstring to a netinfo function
- remove unnecessary connectivity url check from EphemeralIPv4Network
```

## Test Steps
Build a deb from this branch
On both noble and mantic, do the following:
- `cloud-init clean --logs`
- `cloud-init init --local`
- still connected? cool

My logs: [interactive-init-local-support.tar.gz](https://github.com/canonical/cloud-init/files/14952135/interactive-init-local-support.tar.gz)

## Merge type

### DIFF
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -35,7 +35,6 @@ class EphemeralIPv4Network:
         broadcast,
         interface_addrs_before_dhcp: dict,
         router=None,
-        connectivity_url_data: Optional[Dict[str, Any]] = None,
         static_routes=None,
     ):
         """Setup context manager and validate call signature.
@@ -46,8 +45,6 @@ class EphemeralIPv4Network:
             prefix.
         @param broadcast: Broadcast address for the IPv4 network.
         @param router: Optionally the default gateway IP.
-        @param connectivity_url_data: Optionally, a URL to verify if a usable
-           connection already exists.
         @param static_routes: Optionally a list of static routes from DHCP
         """
         if not all([interface, ip, prefix_or_mask, broadcast]):
@@ -64,7 +61,6 @@ class EphemeralIPv4Network:
                 "netmask: {0}".format(e)
             ) from e
 
-        self.connectivity_url_data = connectivity_url_data
         self.interface = interface
         self.ip = ip
         self.broadcast = broadcast
@@ -77,9 +73,6 @@ class EphemeralIPv4Network:
         self.interface_addrs_before_dhcp = interface_addrs_before_dhcp.get(
             self.interface, {}
         )
-        self.interface_addrs_after_dhcp = netinfo.netdev_info().get(
-            self.interface, {}
-        )
 
     def __enter__(self):
         """Set up ephemeral network if interface is not connected.
@@ -88,20 +81,6 @@ class EphemeralIPv4Network:
         addresses, routes, etc
         """
 
-        if self.connectivity_url_data:
-            if net.has_url_connectivity(self.connectivity_url_data):
-                LOG.debug(
-                    "Skip ephemeral network setup, instance has connectivity"
-                    " to %s",
-                    self.connectivity_url_data["url"],
-                )
-                return
-            else:
-                LOG.debug(
-                    "Instance does not have connectivity to %s. Bringing up "
-                    "ephemeral network now.",
-                    self.connectivity_url_data["url"],
-                )
         try:
             try:
                 self._bringup_device()
@@ -152,11 +131,13 @@ class EphemeralIPv4Network:
             self.cidr,
             self.broadcast,
         )
-        has_link = self.interface_addrs_after_dhcp.get("up")
+        interface_addrs_after_dhcp = netinfo.netdev_info().get(
+            self.interface, {}
+        )
+        has_link = interface_addrs_after_dhcp.get("up")
         had_link = self.interface_addrs_before_dhcp.get("up")
         has_ip = self.ip in [
-            ip.get("ip")
-            for ip in self.interface_addrs_after_dhcp.get("ipv4", {})
+            ip.get("ip") for ip in interface_addrs_after_dhcp.get("ipv4", {})
         ]
         had_ip = self.ip in [
             ip.get("ip")
@@ -380,8 +361,6 @@ class EphemeralDHCPv4:
             ] = self.distro.dhcp_client.parse_static_routes(
                 kwargs["static_routes"]
             )
-        if self.connectivity_url_data:
-            kwargs["connectivity_url_data"] = self.connectivity_url_data
         ephipv4 = EphemeralIPv4Network(
             self.distro,
             interface_addrs_before_dhcp=self.interface_addrs_before_dhcp,
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -8,11 +8,8 @@ from functools import partial
 from typing import Any, Callable, Dict, List, Optional
 
 import cloudinit.net as net
-from cloudinit.net.dhcp import (
-    Dhcpcd,
-    NoDHCPLeaseError,
-    maybe_perform_dhcp_discovery,
-)
+import cloudinit.netinfo as netinfo
+from cloudinit.net.dhcp import NoDHCPLeaseError, maybe_perform_dhcp_discovery
 from cloudinit.subp import ProcessExecutionError
 
 LOG = logging.getLogger(__name__)
@@ -36,6 +33,7 @@ class EphemeralIPv4Network:
         ip,
         prefix_or_mask,
         broadcast,
+        interface_addrs_before_dhcp: dict,
         router=None,
         connectivity_url_data: Optional[Dict[str, Any]] = None,
         static_routes=None,
@@ -76,9 +74,20 @@ class EphemeralIPv4Network:
         self.cleanup_cmds: List[Callable] = []
         self.distro = distro
         self.cidr = f"{self.ip}/{self.prefix}"
+        self.interface_addrs_before_dhcp = interface_addrs_before_dhcp.get(
+            self.interface, {}
+        )
+        self.interface_addrs_after_dhcp = netinfo.netdev_info().get(
+            self.interface, {}
+        )
 
     def __enter__(self):
-        """Perform ephemeral network setup if interface is not connected."""
+        """Set up ephemeral network if interface is not connected.
+
+        This context manager handles the lifecycle of the network interface,
+        addresses, routes, etc
+        """
+
         if self.connectivity_url_data:
             if net.has_url_connectivity(self.connectivity_url_data):
                 LOG.debug(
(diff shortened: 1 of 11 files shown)


cloudinit/net/ephemeral.py
```diff
@@ -76,9 +74,20 @@ def __init__(
         self.cleanup_cmds: List[Callable] = []
         self.distro = distro
         self.cidr = f"{self.ip}/{self.prefix}"
+        self.interface_addrs_before_dhcp = interface_addrs_before_dhcp.get(
+            self.interface, {}
+        )
+        self.interface_addrs_after_dhcp = netinfo.netdev_info().get(
```
TheRealFalcon, line 80: Nit: Is there a reason to make this an instance variable rather than putting it in `_bringup_device()`?

TheRealFalcon (commented):
One nit inline, but the approach LGTM!

blackboxsw (commented):
Thanks @holmanb this does look good and will avoid teardown on the CLI when networking is already up. 

While not strictly related to this PR, our rename problem on Hetzner is not resolved on Noble per commit f74b61eae24706dd313efc06792198a718f06672 because Hetzner also provides a connectivity_url which is checked before EphemeralDHCPv4 is called.

During init-local timeframe:
1. EphemeralDHCPv4 is called passing a hetzner connectivity_url
2. when EphemeralDHCPv4.__enter__ is called connectivity_url is inaccessible because interface has no link yet, 
3. EphemeralDHCPv4.obtain_lease calls maybe_perform_dhcp_discovery which brings link_up and calls dhcpcd client
4. dhcpcd sets assigns ipv4 address/nemask etc to the interface
5. EphemeralIPv4Network.__enter__ context is called where connectivity_url is accessible so no setup,address/route assignments or teardown operations are performed
6. Rename fails because link and address from dhcpcd in EphemeralDHCPv4 is still left in 'up' state.

What I think needs to happen in EphemeralDHCPv4 is that it should perform link_down if it brought the link up due to calling `maybe_perrom_dhcp_discovery`.

This patch would be the following: (including James' comment about relocating the instance attributes out of __init__)
```diff
diff --git a/cloudinit/net/ephemeral.py b/cloudinit/net/ephemeral.py
index 02827e0bd..b0beb9dca 100644
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -74,12 +74,7 @@ class EphemeralIPv4Network:
         self.cleanup_cmds: List[Callable] = []
         self.distro = distro
         self.cidr = f"{self.ip}/{self.prefix}"
-        self.interface_addrs_before_dhcp = interface_addrs_before_dhcp.get(
-            self.interface, {}
-        )
-        self.interface_addrs_after_dhcp = netinfo.netdev_info().get(
-            self.interface, {}
-        )
+        self.interface_addrs_before_dhcp = interface_addrs_before_dhcp
 
     def __enter__(self):
         """Set up ephemeral network if interface is not connected.
@@ -152,11 +147,14 @@ class EphemeralIPv4Network:
             self.cidr,
             self.broadcast,
         )
-        has_link = self.interface_addrs_after_dhcp.get("up")
+        interface_addrs_after_dhcp = netinfo.netdev_info().get(
+            self.interface, {}
+        )
+        has_link = interface_addrs_after_dhcp.get("up")
         had_link = self.interface_addrs_before_dhcp.get("up")
         has_ip = self.ip in [
             ip.get("ip")
-            for ip in self.interface_addrs_after_dhcp.get("ipv4", {})
+            for ip in interface_addrs_after_dhcp.get("ipv4", {})
         ]
         had_ip = self.ip in [
             ip.get("ip")
@@ -307,9 +305,9 @@ class EphemeralDHCPv4:
         self._ephipv4 = None
         self.lease = None
         self.dhcp_log_func = dhcp_log_func
+        self.cleanup_cmds: List[Callable] = []
         self.connectivity_url_data = connectivity_url_data
         self.distro = distro
-        self.interface_addrs_before_dhcp = netinfo.netdev_info()
 
     def __enter__(self):
         """Setup sandboxed dhcp context, unless connectivity_url can already be
@@ -333,6 +331,12 @@ class EphemeralDHCPv4:
         self.lease = None
         if self._ephipv4:
             self._ephipv4.__exit__(None, None, None)
+            if not self._ephipv4.cleanup_cmds:
+                # No setup or teardown was performed by self._ephipv4.
+                # We must cleanup any of our own setup due to dhcp_discovery
+                for cmd in self.cleanup_cmds:
+                    cmd()
+
 
     def obtain_lease(self):
         """Perform dhcp discovery in a sandboxed environment if possible.
@@ -345,9 +349,18 @@ class EphemeralDHCPv4:
         """
         if self.lease:
             return self.lease
+        interface_addrs_before_dhcp = netinfo.netdev_info().get(self.iface, {})
         self.lease = maybe_perform_dhcp_discovery(
             self.distro, self.iface, self.dhcp_log_func
         )
+        # link_up is performed by dhcp_discovery. Queue link_down operation
+        self.cleanup_cmds.append(
+            partial(
+                self.distro.net_ops.link_down,
+                self.iface,
+                family="inet",
+            )
+        )
         if not self.lease:
             raise NoDHCPLeaseError()
         LOG.debug(
@@ -384,7 +397,7 @@ class EphemeralDHCPv4:
             kwargs["connectivity_url_data"] = self.connectivity_url_data
         ephipv4 = EphemeralIPv4Network(
             self.distro,
-            interface_addrs_before_dhcp=self.interface_addrs_before_dhcp,
+            interface_addrs_before_dhcp=interface_addrs_before_dhcp,
             **kwargs,
         )
         ephipv4.__enter__()

```

blackboxsw:
Additionally, GCE and Ec2 datasources do not provide connectivity_url to the context manager, which means that we'll be skipping the connectivity_url tests. In cases where we don't have a connectivity_url param, should EphemeralDHCPv4 instead be checking for link up and non-zero length of IPs in order to skip EphemeralDHCP setup?
This would make the diff the following:
```diff
TBD
```

blackboxsw:
Observed failure logs for dhcpcd with the current PR as-is on Hetzner when removing /etc/netplan/50-cloud-init.yaml and running cloud-init clean --logs --reboot.
[cloud-init.log.hetzner.dhcpcd.connectivity.renamebreak.txt](https://github.com/canonical/cloud-init/files/14974115/cloud-init.log.hetzner.dhcpcd.connectivity.renamebreak.txt)


Note as well that dhclient with the same version does not break because dhclient doesn't apply the  ipv4 address to the interface, leaving it unable to connect to the connectivity_url, which means that it performs and EphemeralIPNetwork setup and teardown, leaving the interface down so renames can actually happen.
dhclient success log 
[cloud-init.log.hetzner.dhclient.rename.noconnectivity.txt](https://github.com/canonical/cloud-init/files/14974123/cloud-init.log.hetzner.dhclient.rename.noconnectivity.txt)

holmanb:
> While not strictly related to this PR, our rename problem on Hetzner is not resolved on Noble per commit f74b61e because Hetzner also provides a connectivity_url which is checked before EphemeralDHCPv4 is called.

@blackboxsw are you saying that f74b61e didn't resolve the rename issue or that this PR would reintroduce the rename issue?

holmanb:
@blackboxsw thanks for the proposals and testing. See my feedback below.

> What I think needs to happen in EphemeralDHCPv4 is that it should perform link_down if it brought the link up due to calling maybe_perrom_dhcp_discovery.

The point of the url check in this context manager is to check if we can avoid doing unnecessary work. 

While the patch you provided may accomplish the stated goal for hetzner, I wonder if the proposed patch adds unnecessary complexity. Wouldn't it be simpler to reorder the url check prior to obtaining the lease?

> In cases where we don't have a connectivity_url param, should EphemeralDHCPv4 instead be checking for link up and non-zero length of IPs in order to skip EphemeralDHCP setup?

I don't think that we can safely assume that a nonzero IP means that no work is required. This might work in some cases but I can imagine others where it wouldn't work.

holmanb:
> One nit inline, but the approach LGTM!

holmanb:
> One nit inline, but the approach LGTM!

Thanks for the review @TheRealFalcon

blackboxsw:
> Wouldn't it be simpler to reorder the url check prior to obtaining the lease?

   Yes, and I think we already have that [early check in EphemeralDHCPv4.__enter__](https://github.com/canonical/cloud-init/blob/33c7fc03307c808b17d1a571102205e0935153d1/cloudinit/net/ephemeral.py#L317-L325)


> @blackboxsw are you saying that [f74b61e](https://github.com/canonical/cloud-init/commit/f74b61eae24706dd313efc06792198a718f06672) didn't resolve the rename issue or that this PR would reintroduce the rename issue?

I was saying this rename issue isn't actually resolved on Hetzner with this new PR due to that connectivity_url having enough configuration "up" after the initial sandboxed dhcpcd cmd to obtain a lease which avoids EphemeralIPv4Network altogether due to the [second connectivity_url check in EphemeralIPv4Network.__enter__](https://github.com/canonical/cloud-init/blob/33c7fc03307c808b17d1a571102205e0935153d1/cloudinit/net/ephemeral.py#L91-L014).

You can see this behavior logged in the [attached dhcpcd...renamebeak.txt](https://github.com/canonical/cloud-init/files/14974115/cloud-init.log.hetzner.dhcpcd.connectivity.renamebreak.txt):
```
2024-04-13 22:20:28,133 - ephemeral.py[DEBUG]: Skip ephemeral network setup, instance has connectivity to http://169.254.169.254/hetzner/v1/metadata/instance-id
```

It turns out that we can avoid the ugly of EphemeralDHCPv4 being made aware of whether self._ephipv4.cleanup_cmds was run if we just avoid [the second connectivity_url check in EphemeralIPv4Network.__enter__](https://github.com/canonical/cloud-init/blob/33c7fc03307c808b17d1a571102205e0935153d1/cloudinit/net/ephemeral.py#L91-L014)

If we drop the connectivity URL, per the below diff, we then allow EphemeralIPv4Network._bringup_device to do any approporiate setup and or teardown necessary based on the interface current state (and it's had_link/had_ip check)

This will keep things simpler as you suggested (and just tested without any of the cleanup_cmds changes on Hetzner successfully.

```diff
--- a/cloudinit/net/ephemeral.py
+++ b/cloudinit/net/ephemeral.py
@@ -88,20 +88,6 @@ class EphemeralIPv4Network:
         addresses, routes, etc
         """
 
-        if self.connectivity_url_data:
-            if net.has_url_connectivity(self.connectivity_url_data):
-                LOG.debug(
-                    "Skip ephemeral network setup, instance has connectivity"
-                    " to %s",
-                    self.connectivity_url_data["url"],
-                )
-                return
-            else:
-                LOG.debug(
-                    "Instance does not have connectivity to %s. Bringing up "
-                    "ephemeral network now.",
-                    self.connectivity_url_data["url"],
-                )
         try:
             try:
                 self._bringup_device()
```

holmanb:
Thanks for the reviews and discussion @TheRealFalcon and @blackboxsw!

I just pushed a change dropping the extra connectivity url, I think this is ready for re-review.

blackboxsw (approved):
The updated approach makes sense and correctly limits our teardown to only cases where EphemeralIPv4Network performed the link/address setup in the first place. While we still have the case where we are spending some time during `cloud-init clean; cloud-init init -local` re-running the dhcp client unnecessarily due to maybe_perform_dhcp_discovering in EphemeralDHCPv4.obtain_lease. cloud-init won't attempt to perform any additional setup or teardown of the primary interface due to link_up and has_ip checks in EphemeralIPv4Network. Works on both Hetzner and Ec2 platforms without issue.

PR #5168, opened 2024-04-11
docs: Add deprecated system_info to schema

TheRealFalcon:
## Proposed Commit Message

```
docs: Add deprecated system_info to schema

In some cases, `system_info` can be passed via user data or vendor data
to override the system_info in /etc/cloud/cloud.cfg . While this
technically can work, this is a use case we no longer support and should
indicate that it is deprecated.

Also remove/update examples.
```

## Merge type

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -577,6 +577,12 @@
         },
         "merge_type": {
           "$ref": "#/$defs/merge_definition"
+        },
+        "system_info": {
+          "type": "object",
+          "description": "System and/or distro specific settings. This is not intended to be overridden by user data or vendor data.",
+          "deprecated": true,
+          "deprecated_version": "24.2"
         }
       }
     },
@@ -3876,6 +3882,7 @@
     "ssh_pwauth": {},
     "ssh_quiet_keygen": {},
     "swap": {},
+    "system_info": {},
     "timezone": {},
     "ubuntu_advantage": {},
     "ubuntu_pro": {},
--- a/tests/data/user_data.1.txt
+++ b/tests/data/user_data.1.txt
@@ -3,13 +3,3 @@ write_files:
 -   content: blah
     path: /etc/blah.ini
     permissions: 493
-
-system_info:
- package_mirrors:
-   - arches: [i386, amd64, blah]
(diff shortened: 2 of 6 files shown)


cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -577,6 +577,12 @@
         },
         "merge_type": {
           "$ref": "#/$defs/merge_definition"
+        },
+        "system_info": {
+          "type": "object",
+          "description": "System and/or distro specific settings. This is not intended to be overridden by user data or vendor data.",
+          "deprecated": true,
+          "deprecated_version": "24.2"
```
a-dubs, line 585: Is the purpose of adding this to the schema, now that it's deprecated, to hopefully give user's warnings if they have the json schema loaded and attempt to use this "unsupported" config?

tests/data/user_data.1.txt
```diff
@@ -3,13 +3,3 @@ write_files:
...
-
-system_info:
- package_mirrors:
-   - arches: [i386, amd64, blah]
-     failsafe:
-      primary: http://my.archive.mydomain.com/ubuntu
-      security: http://my.security.mydomain.com/ubuntu
-     search:
-      primary: []
-      security: []
```
a-dubs, line 15: so this had to be removed because this is testing the unsupported behavior of setting `system_info` via user data? thus, it is now just baked into the cfg in the test below?

cloudinit/config/schemas/schema-cloud-config-v1.json
holmanb: +1 yes, basically. This should help avoid user confusion if they know the config works.

Ideally we would point them to the preferred option in a deprecation message, though this one is a little tricky since it's not really a 1:1 mapping - the users key is the preferred replacement for default_user, but other than that I'm not sure we really have a good story.

holmanb (commented):
Thanks @TheRealFalcon! I see a reference in unittests/sources/test_vultr.py, could we remove that one too with this?

TheRealFalcon:
>  I see a reference in unittests/sources/test_vultr.py, could we remove that one too with this?

Done!

holmanb (approved):
LGTM, thanks @TheRealFalcon!

PR #5169, opened 2024-04-11
growpart: Fix behaviour for ZFS datasets

igalic:
## Proposed Commit Message

```
growpart: Fix behaviour for ZFS datasets

on FreeBSD, ZFS datasets aren't exposed directly as disks devices, so
stats will fail. Currently, this means that growpart will be skipped for
FreeBSD ZFS volumes, even tho we are perfectly capable of  resizing ZFS.

- Return fs from devent2dev(), which is now inappropriately named
- Add get_zfs_size() helper, and extend get_size() to accept an `fs`
  parameter.
- Add _call_resizer() helper, to avoid an indent on `if fs == zfs`

Sponsored by:	The FreeBSD Foundation
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -161,7 +161,7 @@ class Resizer(ABC):
         ...
 
     @abstractmethod
-    def resize(self, diskdev, partnum, partdev):
+    def resize(self, diskdev, partnum, partdev, fs):
         ...
 
 
@@ -178,8 +178,8 @@ class ResizeGrowPart(Resizer):
             pass
         return False
 
-    def resize(self, diskdev, partnum, partdev):
-        before = get_size(partdev)
+    def resize(self, diskdev, partnum, partdev, fs):
+        before = get_size(partdev, fs)
 
         # growpart uses tmp dir to store intermediate states
         # and may conflict with systemd-tmpfiles-clean
@@ -211,7 +211,7 @@ class ResizeGrowPart(Resizer):
                 util.logexc(LOG, "Failed: growpart %s %s", diskdev, partnum)
                 raise ResizeFailedException(e) from e
 
-        return (before, get_size(partdev))
+        return (before, get_size(partdev, fs))
 
 
 class ResizeGrowFS(Resizer):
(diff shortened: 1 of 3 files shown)


cloudinit/config/cc_growpart.py
```diff
@@ -270,36 +270,46 @@ def resize(self, diskdev, partnum, partdev):
...
+    try:
+        size, _ = subp.subp(["zpool", 'get', '-Hpovalue', 'size', zpool])
+    except subp.ProcessExecutionError as e:
+        LOG.debug("Failed: zpool get size %s", zpool)
+        return None
+    return int(size.strip())
+
 def devent2dev(devent):
-    if devent.startswith("/dev/"):
-        return devent
```
igalic, line 297: instead of removing this short-cut, maybe i should `return devent, None` here.

cloudinit/config/cc_growpart.py
```diff
@@ -270,36 +270,46 @@ def resize(self, diskdev, partnum, partdev):
...
+def get_zfs_size(dataset) -> Optional[int]:
+    zpool = dataset.split('/')[0]
+    try:
+        size, _ = subp.subp(["zpool", 'get', '-Hpovalue', 'size', zpool])
+    except subp.ProcessExecutionError as e:
+        LOG.debug("Failed: zpool get size %s", zpool)
+        return None
+    return int(size.strip())
+
 def devent2dev(devent):
```
igalic, line 307: might wanna rename this function now that it returns more, maybe `deventinfo`?

cloudinit/config/cc_growpart.py
```diff
@@ -433,84 +446,85 @@ def resize_devices(resizer, devices, distro: Distro):
...
 
-        try:
-            statret = os.stat(blockdev)
-        except OSError as e:
-            info.append(
-                (
-                    devent,
-                    RESIZE.SKIPPED,
-                    "stat of '%s' failed: %s" % (blockdev, e),
+        if fs != "zfs":
```
igalic, line 449: I do *not* like this.
There has to be a better way, but I'm not yet sure what it is.

igalic (commented):
This is the kind of code that a `goto` would make more readable.
Unfortunately, Python has no `goto`, so I'll have to think a little harder about how to do this.

cloudinit/config/cc_growpart.py
igalic: added a helper function to remove this indent.

cloudinit/config/cc_growpart.py
igalic: done.

igalic:
tested on aws: Still works!

holmanb:
> tested on aws: Still works!

Nice, thanks for this. Busy at a conference now but I'll try to review (and land it if ready) in the next few days.

cloudinit/config/cc_growpart.py
holmanb: > might wanna rename this function now that it returns more, maybe deventinfo?

Agreed, how about a snake case `device_entry_info()` or `dev_entry_info()` or something?

igalic: sounds good, I'll change that tomorrow or the day after.

holmanb (approved):
Sorry for the long delay on this one @igalic. This looks good to me, thanks for putting in the work on it!

PR #5176, opened 2024-04-15
feat(alpine): add support for Busybox adduser/addgroup

dermotbradley:
```
feat(alpine): add support for Busybox adduser/addgroup

By default Alpine Linux provides Busybox utilities such as adduser
and addgroup for managing users and groups. Optionally the Alpine
"shadow" package provides the traditional useradd/groupadd utilities.

Add fallback support for the Busybox user/group management utilities
for Alpine Linux.
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/alpine.py
+++ b/cloudinit/distros/alpine.py
@@ -8,8 +8,10 @@
 
 import logging
 import os
+import re
 import stat
-from typing import Optional
+from datetime import datetime
+from typing import Any, Dict, Optional
 
 from cloudinit import distros, helpers, subp, util
 from cloudinit.distros.parsers.hostname import HostnameConf
@@ -32,6 +34,7 @@ class Distro(distros.Distro):
     keymap_path = "/usr/share/bkeymaps/"
     locale_conf_fn = "/etc/profile.d/50-cloud-init-locale.sh"
     network_conf_fn = "/etc/network/interfaces"
+    shadow_fn = "/etc/shadow"
     renderer_configs = {
         "eni": {"eni_path": network_conf_fn, "eni_header": NETWORK_FILE_HEADER}
     }
@@ -201,6 +204,339 @@ class Distro(distros.Distro):
             self._preferred_ntp_clients = ["chrony", "ntp"]
 
         return self._preferred_ntp_clients
+
+    def add_user(self, name, **kwargs):
+        """
+        Add a user to the system using standard tools
+
+        On Alpine this may use either 'useradd' or 'adduser' depending
+        on whether the 'shadow' package is installed.
+        """
+        if util.is_user(name):
+            LOG.info("User %s already exists, skipping.", name)
+            return
+
+        if "selinux_user" in kwargs:
+            LOG.warning("Ignoring selinux_user parameter for Alpine Linux")
+            del kwargs["selinux_user"]
+
+        # If 'useradd' is available then use the generic
+        # add_user function from __init__.py instead.
+        if subp.which("useradd"):
+            return super().add_user(name, **kwargs)
+
+        create_groups = kwargs.pop("create_groups", True)
+
+        adduser_cmd = ["adduser", "-D"]
+
+        # Since we are creating users, we want to carefully validate
+        # the inputs. If something goes wrong, we can end up with a
+        # system that nobody can login to.
+        adduser_opts = {
+            "gecos": "-g",
+            "homedir": "-h",
+            "primary_group": "-G",
+            "shell": "-s",
+            "uid": "-u",
+        }
+
+        adduser_flags = {"system": "-S"}
+
+        # support kwargs having groups=[list] or groups="g1,g2"
+        groups = kwargs.get("groups")
+        if groups:
+            if isinstance(groups, str):
+                groups = groups.split(",")
+            elif isinstance(groups, dict):
+                util.deprecate(
+                    deprecated=f"The user {name} has a 'groups' config value "
+                    "of type dict",
+                    deprecated_version="22.3",
+                    extra_message="Use a comma-delimited string or "
+                    "array instead: group1,group2.",
+                )
+
+            # remove any white spaces in group names, most likely
+            # that came in as a string like: groups: group1, group2
+            groups = [g.strip() for g in groups]
+
+            # kwargs.items loop below wants a comma delimited string
+            # that can go right through to the command.
+            kwargs["groups"] = ",".join(groups)
+
+            if kwargs.get("primary_group"):
+                groups.append(kwargs["primary_group"])
+
+        if create_groups and groups:
+            for group in groups:
+                if not util.is_group(group):
+                    self.create_group(group)
+                    LOG.debug("created group '%s' for user '%s'", group, name)
+        if "uid" in kwargs:
+            kwargs["uid"] = str(kwargs["uid"])
+
+        unsupported_busybox_values: Dict[str, Any] = {
+            "groups": [],
+            "expiredate": None,
+            "inactive": None,
+            "passwd": None,
+        }
+
+        # Check the values and create the command
+        for key, val in sorted(kwargs.items()):
+            if key in adduser_opts and val and isinstance(val, str):
+                adduser_cmd.extend([adduser_opts[key], val])
+            elif (
+                key in unsupported_busybox_values
+                and val
+                and isinstance(val, str)
+            ):
+                # Busybox's 'adduser' does not support specifying these
+                # options so store them for use via alternative means.
+                if key == "groups":
+                    unsupported_busybox_values[key] = val.split(",")
+                else:
+                    unsupported_busybox_values[key] = val
+            elif key in adduser_flags and val:
+                adduser_cmd.append(adduser_flags[key])
+
+        # Don't create the home directory if directed so
+        # or if the user is a system user
+        if kwargs.get("no_create_home") or kwargs.get("system"):
+            adduser_cmd.append("-H")
+
+        # Busybox's 'adduser' puts username at end of command
+        adduser_cmd.append(name)
+
+        # Run the command
+        LOG.debug("Adding user %s", name)
+        try:
+            subp.subp(adduser_cmd)
+        except subp.ProcessExecutionError as e:
+            LOG.warning("Failed to create user %s", name)
+            raise e
+
+        # Process remaining options that Busybox's 'adduser' does not support
+
+        # Separately add user to each additional group as Busybox's
+        # 'adduser' does not support specifying additional groups.
+        for addn_group in unsupported_busybox_values[
+            "groups"
+        ]:  # pylint: disable=E1133
+            LOG.debug("Adding user to group %s", addn_group)
+            try:
+                subp.subp(["addgroup", name, addn_group])
+            except subp.ProcessExecutionError as e:
+                util.logexc(
+                    LOG, "Failed to add user %s to group %s", name, addn_group
+                )
+                raise e
+
+        if unsupported_busybox_values["passwd"]:
+            # Separately set password as Busybox's 'adduser' does
+            # not support passing password as CLI option.
+            super().set_passwd(
+                name, unsupported_busybox_values["passwd"], hashed=True
+            )
+
+        # Busybox's 'adduser' is hardcoded to always set the following field
+        # values (numbered from "0") in /etc/shadow unlike 'useradd':
+        #
+        # Field                          Value set
+        #
+        #   3    minimum password age    0 (no min age)
+        #   4    maximum password age    99999 (days)
+        #   5    warning period          7 (warn days before max age)
+        #
+        # so modify these fields to be empty.
+        #
+        # Also set expiredate (field '7') and/or inactive (field '6')
+        # values directly in /etc/shadow file as Busybox's 'adduser'
+        # does not support passing these as CLI options.
+
+        expiredate = unsupported_busybox_values["expiredate"]
+        inactive = unsupported_busybox_values["inactive"]
+
+        shadow_contents = None
+        shadow_file = self.shadow_fn
+        try:
+            shadow_contents = util.load_text_file(shadow_file)
+        except FileNotFoundError as e:
+            LOG.warning("Failed to read %s file, file not found", shadow_file)
+            raise e
+
+        # Find the line in /etc/shadow for the user
+        original_line = None
+        for line in shadow_contents.splitlines():
+            new_line_parts = line.split(":")
+            if new_line_parts[0] == name:
+                original_line = line
+                break
+
+        if original_line:
+            # Modify field(s) in copy of user's shadow file entry
+            update_type = ""
+
+            # Minimum password age
+            new_line_parts[3] = ""
+            # Maximum password age
+            new_line_parts[4] = ""
+            # Password warning period
+            new_line_parts[5] = ""
+            update_type = "password aging"
+
+            if expiredate is not None:
+                # Convert date into number of days since 1st Jan 1970
+                days = (
+                    datetime.fromisoformat(expiredate)
+                    - datetime.fromisoformat("1970-01-01")
+                ).days
+                new_line_parts[7] = str(days)
+                if update_type != "":
+                    update_type = update_type + " & "
+                update_type = update_type + "acct expiration date"
+            if inactive is not None:
+                new_line_parts[6] = inactive
+                if update_type != "":
+                    update_type = update_type + " & "
+                update_type = update_type + "inactivity period"
+
+            # Replace existing line for user with modified line
+            shadow_contents = shadow_contents.replace(
+                original_line, ":".join(new_line_parts)
+            )
+            LOG.debug("Updating %s for user %s", update_type, name)
+            try:
+                util.write_file(
+                    shadow_file, shadow_contents, omode="w", preserve_mode=True
+                )
+            except IOError as e:
+                util.logexc(LOG, "Failed to update %s file", shadow_file)
+                raise e
+        else:
+            util.logexc(
+                LOG, "Failed to update %s for user %s", shadow_file, name
+            )
+
+    def lock_passwd(self, name):
+        """
+        Lock the password of a user, i.e., disable password logins
+        """
+
+        # Check whether Shadow's or Busybox's version of 'passwd'.
+        # If Shadow's 'passwd' is available then use the generic
+        # lock_passwd function from __init__.py instead.
+        if not os.path.islink(
+            "/usr/bin/passwd"
+        ) or "bbsuid" not in os.readlink("/usr/bin/passwd"):
+            return super().lock_passwd(name)
+
+        cmd = ["passwd", "-l", name]
+        # Busybox's 'passwd', unlike Shadow's 'passwd', errors
+        # if password is already locked:
+        #
+        #   "passwd: password for user2 is already locked"
+        #
+        # with exit code 1
+        try:
+            (_out, err) = subp.subp(cmd, rcs=[0, 1])
+            if re.search(r"is already locked", err):
+                return True
+        except subp.ProcessExecutionError as e:
+            util.logexc(LOG, "Failed to disable password for user %s", name)
+            raise e
+
+    def expire_passwd(self, user):
+        # Check whether Shadow's or Busybox's version of 'passwd'.
+        # If Shadow's 'passwd' is available then use the generic
+        # expire_passwd function from __init__.py instead.
+        if not os.path.islink(
+            "/usr/bin/passwd"
+        ) or "bbsuid" not in os.readlink("/usr/bin/passwd"):
+            return super().expire_passwd(user)
+
+        # Busybox's 'passwd' does not provide an expire option
+        # so have to manipulate the shadow file directly.
+        shadow_contents = None
+        shadow_file = self.shadow_fn
+        try:
+            shadow_contents = util.load_text_file(shadow_file)
+        except FileNotFoundError as e:
+            LOG.warning("Failed to read %s file, file not found", shadow_file)
+            raise e
+
+        # Find the line in /etc/shadow for the user
+        original_line = None
+        for line in shadow_contents.splitlines():
+            new_line_parts = line.split(":")
+            if new_line_parts[0] == user:
+                LOG.debug("Found /etc/shadow line matching user %s", user)
+                original_line = line
+                break
+
+        if original_line:
+            # Replace existing line for user with modified line
+            #
+            # Field '2' (numbered from '0') in /etc/shadow
+            # is the "date of last password change".
+            if new_line_parts[2] != "0":
+                # Busybox's 'adduser' always expires password so only
+                # need to expire it now if this is not a new user.
+                new_line_parts[2] = "0"
+                shadow_contents = shadow_contents.replace(
+                    original_line, ":".join(new_line_parts), 1
+                )
+
+                LOG.debug("Expiring password for user %s", user)
+                try:
+                    util.write_file(
+                        shadow_file,
+                        shadow_contents,
+                        omode="w",
+                        preserve_mode=True,
+                    )
+                except IOError as e:
+                    util.logexc(LOG, "Failed to update %s file", shadow_file)
+                    raise e
+            else:
+                LOG.debug("Password for user %s is already expired", user)
+        else:
+            util.logexc(LOG, "Failed to set 'expire' for %s", user)
+
+    def create_group(self, name, members=None):
+        # If 'groupadd' is available then use the generic
+        # create_group function from __init__.py instead.
+        if subp.which("groupadd"):
+            return super().create_group(name, members)
+
+        group_add_cmd = ["addgroup", name]
+        if not members:
+            members = []
+
+        # Check if group exists, and then add if it doesn't
+        if util.is_group(name):
+            LOG.warning("Skipping creation of existing group '%s'", name)
+        else:
+            try:
+                subp.subp(group_add_cmd)
+                LOG.info("Created new group %s", name)
+            except subp.ProcessExecutionError:
+                util.logexc(LOG, "Failed to create group %s", name)
+
+        # Add members to the group, if so defined
+        if len(members) > 0:
+            for member in members:
+                if not util.is_user(member):
+                    LOG.warning(
+                        "Unable to add group member '%s' to group '%s'"
+                        "; user does not exist.",
+                        member,
+                        name,
+                    )
+                    continue
+
+                subp.subp(["addgroup", member, name])
+                LOG.info("Added user '%s' to group '%s'", member, name)
 
     def shutdown_command(self, mode="poweroff", delay="now", message=None):
         # called from cc_power_state_change.load_power_state
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -914,8 +914,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
         if hashed:
             # Need to use the short option name '-e' instead of '--encrypted'
-            # (which would be more descriptive) since SLES 11 doesn't know
-            # about long names.
+            # (which would be more descriptive) since Busybox and SLES 11
+            # chpasswd don't know about long names.
             cmd.append("-e")
 
         try:
--- a/tests/unittests/distros/test_alpine.py
+++ b/tests/unittests/distros/test_alpine.py
@@ -1,0 +1,78 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+
+from unittest import mock
+
+import pytest
+
+from cloudinit import distros, util
+from tests.unittests.helpers import TestCase
+
+
+class TestAlpineBusyboxUserGroup:
+    @mock.patch("cloudinit.distros.alpine.subp.subp")
+    @mock.patch("cloudinit.distros.subp.which", return_value=False)
+    def test_busybox_add_group(self, m_which, m_subp):
+        distro = distros.fetch("alpine")("alpine", {}, None)
+
+        group = "mygroup"
+
+        distro.create_group(group)
+
+        m_subp.assert_called_with(["addgroup", group])
+
+    @pytest.mark.usefixtures("fake_filesystem")
+    @mock.patch("cloudinit.distros.alpine.subp.subp")
+    @mock.patch("cloudinit.distros.subp.which", return_value=False)
+    def test_busybox_add_user(self, m_which, m_subp, tmpdir):
+        distro = distros.fetch("alpine")("alpine", {}, None)
+
+        shadow_file = tmpdir.join("/etc/shadow")
+        shadow_file.dirpath().mkdir()
+
+        user = "me2"
+
+        # Need to place entry for user in /etc/shadow as
+        # "adduser" is stubbed and so will not create it.
+        root_entry = "root::19848:0:::::"
+        shadow_file.write(
+            root_entry + "\n" + user + ":!:19848:0:99999:7:::" + "\n"
+        )
+
+        distro.shadow_fn = shadow_file
+
+        distro.add_user(user, lock_passwd=True)
+
+        m_subp.assert_called_with(["adduser", "-D", user])
+
+        contents = util.load_text_file(shadow_file)
+        expected = root_entry + "\n" + user + ":!:19848::::::" + "\n"
+
+        assert contents == expected
+
+
+class TestAlpineShadowUserGroup(TestCase):
+    distro = distros.fetch("alpine")("alpine", {}, None)
+
+    @mock.patch("cloudinit.distros.alpine.subp.subp")
+    @mock.patch(
+        "cloudinit.distros.subp.which", return_value=("/usr/sbin/groupadd")
+    )
+    def test_shadow_add_group(self, m_which, m_subp):
+        group = "mygroup"
+
+        self.distro.create_group(group)
+
+        m_subp.assert_called_with(["groupadd", group])
+
+    @mock.patch("cloudinit.distros.alpine.subp.subp")
+    @mock.patch(
+        "cloudinit.distros.subp.which", return_value=("/usr/sbin/useradd")
+    )
+    def test_shadow_add_user(self, m_which, m_subp):
+        user = "me2"
+
+        self.distro.add_user(user)
+
+        m_subp.assert_called_with(
+            ["useradd", user, "-m"], logstring=["useradd", user, "-m"]
+        )


cloudinit/distros/alpine.py
```diff
@@ -202,6 +204,423 @@ def preferred_ntp_clients(self):
...
+            LOG.info("User %s already exists, skipping.", name)
+            return
+
+        if "create_groups" in kwargs:
+            create_groups = kwargs.pop("create_groups")
+        else:
+            create_groups = True
+
+        # Decide whether 'useradd' or 'adduser' will be used
+        if subp.which("useradd"):
```
TheRealFalcon, line 224: If we're using `useradd`, does the default implementation work? Instead of separate paths for both, at the top of this method could we put something like:
```python
if subp.which("useradd"):
    return super().add_user(name, **kwargs)
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +204,423 @@ def preferred_ntp_clients(self):
...
+            redact_opts = ["passwd"]
+
+        # support kwargs having groups=[list] or groups="g1,g2"
+        groups = kwargs.get("groups")
+        if groups:
+            if isinstance(groups, str):
+                groups = groups.split(",")
+
+            if isinstance(groups, dict):
+                LOG.warning(
```
TheRealFalcon, line 284: Why not use `util.deprecate()` here?

cloudinit/distros/alpine.py
```diff
@@ -202,6 +204,423 @@ def preferred_ntp_clients(self):
...
+
+        if create_groups and groups:
+            for group in groups:
+                if not util.is_group(group):
+                    self.create_group(group)
+                    LOG.debug("created group '%s' for user '%s'", group, name)
+        if "uid" in kwargs.keys():
+            kwargs["uid"] = str(kwargs["uid"])
+
+        addn_groups = None
```
TheRealFalcon, line 312: Nit: I think this section could be simplified. Something like:
```python
unsupported_busybox_values = {
    "addn_groups": None,
    "expirydate": None,
    "inactive": None,
    "passwd": None
}
```
Then later something along the lines of:
```python
for key, val in sorted(kwargs.items()):
    if key in unsupported_busybox_values:
        unsupported_busybox_values[key] = val
```

TheRealFalcon (commented):
@dermotbradley , thanks for this! It'll be good to have extra busybox support.

First pass, I mentioned this inline, but I'm wondering if we can re-use the `adduser` implementation that already exists rather than re-adding most of the existing code in the Alpine context. If there's a reason we can't call `super()` directly, is there a way we could modify the existing code so that it could be used in both places?

cloudinit/distros/alpine.py
dermotbradley: Hmm. I hadn't considered that.

This Alpine-specific function doesn't define "--selinux-user" in the list of valid useradd options for example as Alpine doesn't support Selinux.

Also a lot of the code in the Alpine-specific add_user function would still need to remain for "adduser" support so I'm not convinced of the benefits of splitting "useradd" and "adduser" between 2 separate functions.

cloudinit/distros/alpine.py
dermotbradley: I've had this code sitting around locally from before util.deprecate() was added in Feb 2013. I thought it was about time to upstream this :-)

I'll change it to use that now.

cloudinit/distros/alpine.py
TheRealFalcon: When I walked through it, almost all of the `useradd` code was essentially just duplicated into the alpine code behind if statements. For `--selinux-user`, we can pop that off before calling it. You sure something like this wouldn't work?
```diff
diff --git a/cloudinit/distros/alpine.py b/cloudinit/distros/alpine.py
index 1b8d52c27..9d88a4b30 100644
--- a/cloudinit/distros/alpine.py
+++ b/cloudinit/distros/alpine.py
@@ -11,7 +11,7 @@ import os
 import re
 import stat
 from datetime import datetime
-from typing import Optional
+from typing import Any, Dict, Optional
 
 from cloudinit import distros, helpers, subp, util
 from cloudinit.distros.parsers.hostname import HostnameConf
@@ -215,64 +215,36 @@ class Distro(distros.Distro):
             LOG.info("User %s already exists, skipping.", name)
             return
 
+        if "selinux_user" in kwargs:
+            LOG.warning("Ignoring selinux_user parameter for Alpine Linux")
+            del kwargs["selinux_user"]
+
+        if subp.which("useradd"):
+            return super().add_user(name, **kwargs)
+
         if "create_groups" in kwargs:
             create_groups = kwargs.pop("create_groups")
         else:
             create_groups = True
 
-        # Decide whether 'useradd' or 'adduser' will be used
-        if subp.which("useradd"):
-            # Shadow useradd
-            use_useradd = True
-        else:
-            # Busybox adduser
-            use_useradd = False
-
         # Since we are creating users, we want to carefully validate the
         # inputs. If something goes wrong, we can end up with a system
         # that nobody can login to.
-        if use_useradd:
-            # Shadow package is installed, use 'useradd'
-            add_cmd = ["useradd", name]
-            log_add_cmd = ["useradd", name]
-
-            add_opts = {
-                "expiredate": "--expiredate",
-                "gecos": "--comment",
-                "groups": "--groups",
-                "homedir": "--home",
-                "inactive": "--inactive",
-                "passwd": "--password",
-                "primary_group": "--gid",
-                "shell": "--shell",
-                "uid": "--uid",
-            }
-
-            add_flags = {
-                "no_log_init": "--no-log-init",
-                "no_user_group": "--no-user-group",
-                "system": "--system",
-            }
-        else:
-            # Shadow package is not installed, use Busybox 'adduser'
-            add_cmd = ["adduser", "-D"]
-
-            add_opts = {
-                "expiredate": "",
-                "gecos": "-g",
-                "groups": "",
-                "homedir": "-h",
-                "inactive": "",
-                "passwd": "",
-                "primary_group": "-G",
-                "shell": "-s",
-                "uid": "-u",
-            }
-
-            add_flags = {"system": "-S"}
-
-        if use_useradd:
-            redact_opts = ["passwd"]
+        add_cmd = ["adduser", "-D"]
+
+        add_opts = {
+            "expiredate": "",
+            "gecos": "-g",
+            "groups": "",
+            "homedir": "-h",
+            "inactive": "",
+            "passwd": "",
+            "primary_group": "-G",
+            "shell": "-s",
+            "uid": "-u",
+        }
+
+        add_flags = {"system": "-S"}
 
         # support kwargs having groups=[list] or groups="g1,g2"
         groups = kwargs.get("groups")
@@ -309,148 +281,114 @@ class Distro(distros.Distro):
         if "uid" in kwargs.keys():
             kwargs["uid"] = str(kwargs["uid"])
 
-        addn_groups = None
-        expirydate = None
-        inactive = None
-        passwd = None
+        unsupported_busybox_values: Dict[str, Any] = {
+            "groups": None,
+            "expirydate": None,
+            "inactive": None,
+            "passwd": None,
+        }
 
         # Check the values and create the command
         for key, val in sorted(kwargs.items()):
-
             if key in add_opts and val and isinstance(val, str):
-                if not use_useradd and (
-                    key == "expiredate"
-                    or key == "groups"
-                    or key == "inactive"
-                    or key == "passwd"
-                ):
+                if key in unsupported_busybox_values:
                     # Busybox 'adduser' does not support specifying these
                     # options so store them for use via alternative means.
-                    if key == "expiredate":
-                        expirydate = val
-                    elif key == "groups":
-                        addn_groups = val.split(",")
-                    elif key == "inactive":
-                        inactive = val
-                    elif key == "passwd":
-                        passwd = val
+                    if key == "groups":
+                        unsupported_busybox_values[key] = val.split(",")
+                    else:
+                        unsupported_busybox_values[key] = val
                 else:
                     add_cmd.extend([add_opts[key], val])
 
-                if use_useradd:
-                    # Redact certain fields from the logs
-                    if key in redact_opts:
-                        log_add_cmd.extend([add_opts[key], "REDACTED"])
-                    else:
-                        log_add_cmd.extend([add_opts[key], val])
-
             elif key in add_flags and val:
                 add_cmd.append(add_flags[key])
-                if use_useradd:
-                    log_add_cmd.append(add_flags[key])
 
         # Don't create the home directory if directed so or if the user is a
         # system user
         if kwargs.get("no_create_home") or kwargs.get("system"):
-            if use_useradd:
-                # Shadow useradd
-                add_cmd.append("-M")
-                log_add_cmd.append("-M")
-            else:
-                # Busybox adduser
-                add_cmd.append("-H")
-        else:
-            if use_useradd:
-                # Shadow useradd
-                add_cmd.append("-m")
-                log_add_cmd.append("-m")
+            # Busybox adduser
+            add_cmd.append("-H")
 
-        if not use_useradd:
-            # Busybox 'adduser' puts username at end of command
-            add_cmd.append(name)
+        # Busybox 'adduser' puts username at end of command
+        add_cmd.append(name)
 
         # Run the command
-        LOG.debug("Adding user %s", name)
-        if use_useradd:
-            try:
-                subp.subp(add_cmd, logstring=log_add_cmd)
-            except Exception as e:
-                util.logexc(LOG, "Failed to create user %s", name)
-                raise e
-        else:
-            try:
-                subp.subp(add_cmd)
-            except Exception as e:
-                util.logexc(LOG, "Failed to create user %s", name)
-                raise e
+        try:
+            subp.subp(add_cmd)
+        except Exception as e:
+            util.logexc(LOG, "Failed to create user %s", name)
+            raise e
 
-        if not use_useradd:
-            if addn_groups is not None:
-                # Separately add user to each additional group as Busybox
-                #  'adduser' does not support specifying additional groups.
-                for addn_group in addn_groups:
-                    LOG.debug("Adding user to group %s", addn_group)
-                    try:
-                        subp.subp(["addgroup", name, addn_group])
-                    except Exception as e:
-                        util.logexc(
-                            LOG,
-                            "Failed to add user %s to group %s",
-                            name,
-                            addn_group,
-                        )
-                        raise e
-
-            if passwd is not None:
-                # Separately set password as Busybox 'adduser' does not
-                # support passing password as CLI option.
-                self.set_passwd(name, passwd, hashed=True)
-
-            if expirydate is not None or inactive is not None:
-                # Separately set expirydate and/or inactive values directly
-                # in /etc/shadow file as Busybox 'adduser' does not support
-                # passing these as CLI options.
-                shadow_contents = None
-                with open("/etc/shadow", "r") as file:
-                    shadow_contents = file.read()
-
-                # Find the line for the user
-                original_line = None
-                for line in shadow_contents.splitlines():
-                    current_user = line.split(":")[0]
-                    if current_user == name:
-                        original_line = line
-
-                # Modify field(s) in copy of user's shadow file entry
-                new_list = original_line.split(":")
-                update_type = ""
-                if expirydate is not None:
-                    # Convert date into number of days since 1st Jan 1970
-                    days = (
-                        datetime.fromisoformat(expirydate)
-                        - datetime.fromisoformat("1970-01-01")
-                    ).days
-                    new_list[7] = str(days)
-                    update_type = "acct expiration date"
-                if inactive is not None:
-                    new_list[6] = inactive
-                    if update_type != "":
-                        update_type = update_type + " & "
-                    update_type = update_type + "inactivity period"
-
-                new_line = ":".join(new_list)
+        # Process remaining options that Busybox 'adduser' does not support
+        addn_groups = unsupported_busybox_values["groups"]
+        if addn_groups is not None:
+            # Separately add user to each additional group as Busybox
+            #  'adduser' does not support specifying additional groups.
+            for addn_group in addn_groups:
+                LOG.debug("Adding user to group %s", addn_group)
+                try:
+                    subp.subp(["addgroup", name, addn_group])
+                except Exception as e:
+                    util.logexc(
+                        LOG,
+                        "Failed to add user %s to group %s",
+                        name,
+                        addn_group,
+                    )
+                    raise e
+
+        passwd = unsupported_busybox_values["passwd"]
+        if passwd is not None:
+            # Separately set password as Busybox 'adduser' does not
+            # support passing password as CLI option.
+            self.set_passwd(name, passwd, hashed=True)
+
+        expirydate = unsupported_busybox_values["expirydate"]
+        inactive = unsupported_busybox_values["inactive"]
+        if expirydate is not None or inactive is not None:
+            # Separately set expirydate and/or inactive values directly
+            # in /etc/shadow file as Busybox 'adduser' does not support
+            # passing these as CLI options.
+            shadow_contents = None
+            with open("/etc/shadow", "r") as file:
+                shadow_contents = file.read()
 
-                # Replace existing line for user with modified line
-                shadow_contents = shadow_contents.replace(
-                    original_line, new_line
-                )
-                LOG.debug("Updating %s for user %s", update_type, name)
-                util.write_file(
-                    "/etc/shadow",
-                    shadow_contents,
-                    omode="w",
-                    preserve_mode=True,
-                )
+            # Find the line for the user
+            original_line = None
+            for line in shadow_contents.splitlines():
+                current_user = line.split(":")[0]
+                if current_user == name:
+                    original_line = line
+
+            # Modify field(s) in copy of user's shadow file entry
+            new_list = original_line.split(":")
+            update_type = ""
+            if expirydate is not None:
+                # Convert date into number of days since 1st Jan 1970
+                days = (
+                    datetime.fromisoformat(expirydate)
+                    - datetime.fromisoformat("1970-01-01")
+                ).days
+                new_list[7] = str(days)
+                update_type = "acct expiration date"
+            if inactive is not None:
+                new_list[6] = inactive
+                if update_type != "":
+                    update_type = update_type + " & "
+                update_type = update_type + "inactivity period"
+
+            new_line = ":".join(new_list)
+
+            # Replace existing line for user with modified line
+            shadow_contents = shadow_contents.replace(original_line, new_line)
+            LOG.debug("Updating %s for user %s", update_type, name)
+            util.write_file(
+                "/etc/shadow",
+                shadow_contents,
+                omode="w",
+                preserve_mode=True,
+            )
 
     def lock_passwd(self, name):
         """
```

It's a lot simpler to me and worked for me with a quick test.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +204,423 @@ def preferred_ntp_clients(self):
...
+                # Separately set password as Busybox 'adduser' does not
+                # support passing password as CLI option.
+                self.set_passwd(name, passwd, hashed=True)
+
+            if expirydate is not None or inactive is not None:
+                # Separately set expirydate and/or inactive values directly
+                # in /etc/shadow file as Busybox 'adduser' does not support
+                # passing these as CLI options.
+                shadow_contents = None
+                with open("/etc/shadow", "r") as file:
```
blackboxsw, line 414: Do we want to use `util.load_text_file` here instead of `open` to ensure we log what cloud-init is reading? I't probably not critical as we are logging our write to /etc/shadow below which is indication enough in logs that cloud-init is meddling in /etc/shadow manipulation.

dermotbradley: I had already (locally) rewritten the code to use util.load_text_file.

I've just pushed my latest changes.

cloudinit/distros/alpine.py
dermotbradley: Done.

cloudinit/distros/alpine.py
dermotbradley: Done.

cloudinit/distros/alpine.py
dermotbradley: I've rewritten the code to make use of the default functions where possible.

dermotbradley:
> @dermotbradley , thanks for this! It'll be good to have extra busybox support.
> 
> First pass, I mentioned this inline, but I'm wondering if we can re-use the `adduser` implementation that already exists rather than re-adding most of the existing code in the Alpine context. If there's a reason we can't call `super()` directly, is there a way we could modify the existing code so that it could be used in both places?

Done.

dermotbradley:
Am unsure how to resolve the pylint failure:

> cloudinit/distros/alpine.py:330: [E1133(not-an-iterable), Distro.add_user] Non-iterable value addn_groups is used in an iterating context

I added a type annotation to addn_groups but that had no effect.

dermotbradley:
> Am unsure how to resolve the pylint failure:
> 
> > cloudinit/distros/alpine.py:330: [E1133(not-an-iterable), Distro.add_user] Non-iterable value addn_groups is used in an iterating context
> 
> I added a type annotation to addn_groups but that had no effect.

I ended up adding a pylint ignore - it seems there are [known issues](https://github.com/pylint-dev/pylint/issues/3105) with this warning

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+                # Not Shadow's or Busybox's 'passwd', assume
+                # it functions the same as Shadow's 'passwd'
+                use_shadow = True
+        else:
+            # Shadow's 'passwd --help' does not output anything to stderr
+            use_shadow = True
+
+        # If Shadow's 'passwd' is available then use the generic
+        # expire_passwd function from __init__.py instead.
+        if use_shadow:
```
TheRealFalcon, line 488: I think we can replace this and the entire if block above it with:
```python
if not err.startswith("BusyBox")
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+            return super().expire_passwd(user)
+
+        # Busybox's 'passwd' does not provide an expire option
+        # so have to manipulate the shadow file directly.
+        shadow_contents = None
+        shadow_file = self.shadow_fn
+        try:
+            shadow_contents = util.load_text_file(shadow_file)
+        except FileNotFoundError as e:
+            util.logexc(
```
TheRealFalcon, line 498: I think a `LOG.warning(...)` would suffice here rather than logging the traceback.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+            members = []
+
+        # Check if group exists, and then add if it doesn't
+        if util.is_group(name):
+            LOG.warning("Skipping creation of existing group '%s'", name)
+        else:
+            try:
+                subp.subp(group_add_cmd)
+                LOG.info("Created new group %s", name)
+            except Exception:
```
TheRealFalcon, line 558:
```suggestion
            except subp.ProcessExecutionException:
```

TheRealFalcon (changes requested):
I left some more inline comments. They're hopefully simpler asks this time around.

One more general is that this Alpine `add_user` is becoming quite long. I won't block the PR on it since the parent code it's modeled after is also somewhat long and complicated, but I do think there are some sections like group handling or shadow handling that could easily be moved into helper functions.

That would also make the code easier to test. For how much was added here, it would be nice if we could get some additional test coverage. There are quite a few lines/functions with no coverage at all. Since us upstream devs don't have the Alpine expertise, I think it'd really benefit you long term so that we don't accidentally break things if we ever need to modify things here.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+
+        addn_groups = unsupported_busybox_values["groups"]
+        if addn_groups is not None:
+            # Separately add user to each additional group as Busybox's
+            # 'adduser' does not support specifying additional groups.
+            for addn_group in addn_groups:  # pylint: disable=E1133
+                LOG.debug("Adding user to group %s", addn_group)
+                try:
+                    subp.subp(["addgroup", name, addn_group])
+                except Exception as e:
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+
+        # Busybox's 'adduser' puts username at end of command
+        adduser_cmd.append(name)
+
+        # Run the command
+        LOG.debug("Adding user %s", name)
+        try:
+            subp.subp(adduser_cmd)
+        except Exception as e:
+            util.logexc(LOG, "Failed to create user %s", name)
```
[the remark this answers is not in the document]
dermotbradley: Done. This code was copied from __init__.py's add_user function so the same is still performed there.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+
+        expiredate = unsupported_busybox_values["expiredate"]
+        inactive = unsupported_busybox_values["inactive"]
+
+        shadow_contents = None
+        shadow_file = self.shadow_fn
+        try:
+            shadow_contents = util.load_text_file(shadow_file)
+        except FileNotFoundError as e:
+            util.logexc(
```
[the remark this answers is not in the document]
dermotbradley: Done.

cloudinit/distros/alpine.py
dermotbradley: Done.

cloudinit/distros/alpine.py
dermotbradley: Done.

cloudinit/distros/alpine.py
dermotbradley: Done.

dermotbradley: Actually pylint doesn't like this change:

```
cloudinit/distros/alpine.py:528: [E1101(no-member), Distro.create_group] Module 'cloudinit.subp' has no 'ProcessExecutionException' member
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+            primary_group = kwargs.get("primary_group")
+            if primary_group:
+                groups.append(primary_group)
+
+        if create_groups and groups:
+            for group in groups:
+                if not util.is_group(group):
+                    self.create_group(group)
+                    LOG.debug("created group '%s' for user '%s'", group, name)
+        if "uid" in kwargs.keys():
```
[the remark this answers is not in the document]
dermotbradley: Done.

cloudinit/distros/alpine.py
TheRealFalcon: Bah, sorry. I tried (and failed) from memory. It should be ProcessExecutionError

dermotbradley:
> One more general is that this Alpine `add_user` is becoming quite long. I won't block the PR on it since the parent code it's modeled after is also somewhat long and complicated, but I do think there are some sections like group handling or shadow handling that could easily be moved into helper functions.

Ok, I'll look into this as a future improvement.

> For how much was added here, it would be nice if we could get some additional test coverage. There are quite a few lines/functions with no coverage at all. Since us upstream devs don't have the Alpine expertise, I think it'd really benefit you long term so that we don't accidentally break things if we ever need to modify things here.

Agreed, I do intend to extend the testing for this and yes this would become easier once things are split up more.

cloudinit/distros/alpine.py
dermotbradley: Fixed.

TheRealFalcon (approved):
LGTM! Thanks for the changes here @dermotbradley 

I think @blackboxsw was also reviewing, so I'll wait to see if there's anything else from him.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+                try:
+                    subp.subp(["addgroup", name, addn_group])
+                except subp.ProcessExecutionError as e:
+                    util.logexc(
+                        LOG,
+                        "Failed to add user %s to group %s",
+                        name,
+                        addn_group,
+                    )
+                    raise e
```
blackboxsw, line 341: Here we can drop the if addn_groups and  left-shift this conditional scope if we initialize unsupported_busybox_groups["goups"] to an empty list instead of None above.
```suggestion
        # Separately add user to each additional group as Busybox's
        # 'adduser' does not support specifying additional groups.
        for addn_group in unsupported_busybox_values[
            "groups"
        ]:# pylint: disable=E1133
            LOG.debug("Adding user to group %s", addn_group)
            try:
                subp.subp(["addgroup", name, addn_group])
            except subp.ProcessExecutionError as e:
                    util.logexc(
                        LOG,
                        "Failed to add user %s to group %s",
                        name,
                        addn_group,
                    )
                    raise e
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+        """
+        Lock the password of a user, i.e., disable password logins
+        """
+
+        # Check whether Shadow's or Busybox's version of 'passwd'.
+        # If Shadow's 'passwd' is available then use the generic
+        # lock_passwd function from __init__.py instead.
+        (_out, err) = subp.subp(["passwd", "--help"])
+        if not err.startswith("BusyBox"):
+            return super().lock_passwd(name)
```
blackboxsw, line 438: Our check should probably avoid string matching on the stderr output of the command if possible.
What do you think of using os.readlink instead?
```suggestion
        if os.path.islink("/usr/bin/passwd") and "busybox" not in os.readlink(
            "/usr/bin/passwd"
        ):
            return super().lock_passwd(name)
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+
+        # Find the line in /etc/shadow for the user
+        original_line = None
+        for line in shadow_contents.splitlines():
+            current_user = line.split(":")[0]
+            if current_user == user:
+                LOG.debug("Found /etc/shadow line matching user %s", user)
+                original_line = line
+
+        if original_line is not None:
```
blackboxsw, line 481:
```suggestion
        if original_line:
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+            # Replace existing line for user with modified line
+            new_list = original_line.split(":")
```
blackboxsw, line 483: just use the new_line_parts var split from above

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+            adduser_cmd.append("-H")
+
+        # Busybox's 'adduser' puts username at end of command
+        adduser_cmd.append(name)
+
+        # Run the command
+        LOG.debug("Adding user %s", name)
+        try:
+            subp.subp(adduser_cmd)
+        except Exception as e:
```
blackboxsw, line 320: I think we expect this to be subp.ProcessExecutionError. We like to avoid generic Exception handling when we know what exceptions to deal with.
```suggestion
        except subp.ProcessExecutionError as e:
```

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+        # if password is already locked:
+        #
+        #   "passwd: password for user2 is already locked"
+        #
+        # with exit code 1
+        try:
+            (_out, err) = subp.subp(cmd, rcs=[0, 1])
+            if re.search(r"is already locked", err):
+                return True
+        except Exception as e:
```
blackboxsw, line 451:
```suggestion
        except subp.ProcessExecutionError as e:
```

blackboxsw (commented):
Thanks so much for your patience here @dermotbradley.
- let's avoid calling `set_passwd` when password value is an empty string (drop the `is not None`)
- 
I've added a number of minor inline comments for minor performance issues:
- early exit from for loops
- avoid double split operations

It may be better to also look at the symlink of /usr/bin/passwd rather than string parsing the --help output when deciding to use busybox logic for password handling.

The full diff of suggestions to take as at your discretion:
```diff
diff --git a/cloudinit/distros/alpine.py b/cloudinit/distros/alpine.py
index 1a0fa488f..8dc4314f5 100644
--- a/cloudinit/distros/alpine.py
+++ b/cloudinit/distros/alpine.py
@@ -225,10 +225,7 @@ class Distro(distros.Distro):
         if subp.which("useradd"):
             return super().add_user(name, **kwargs)
 
-        if "create_groups" in kwargs:
-            create_groups = kwargs.pop("create_groups")
-        else:
-            create_groups = True
+        create_groups = kwargs.pop("create_groups", True)
 
         adduser_cmd = ["adduser", "-D"]
 
@@ -250,8 +247,7 @@ class Distro(distros.Distro):
         if groups:
             if isinstance(groups, str):
                 groups = groups.split(",")
-
-            if isinstance(groups, dict):
+            elif isinstance(groups, dict):
                 util.deprecate(
                     deprecated=f"The user {name} has a 'groups' config value "
                     "of type dict",
@@ -264,13 +260,12 @@ class Distro(distros.Distro):
             # that came in as a string like: groups: group1, group2
             groups = [g.strip() for g in groups]
 
-            # kwargs.items loop below wants a comma delimeted string
+            # kwargs.items loop below wants a comma delimited string
             # that can go right through to the command.
             kwargs["groups"] = ",".join(groups)
 
-            primary_group = kwargs.get("primary_group")
-            if primary_group:
-                groups.append(primary_group)
+            if kwargs.get("primary_group"):
+                groups.append(kwargs["primary_group"])
 
         if create_groups and groups:
             for group in groups:
@@ -281,7 +276,7 @@ class Distro(distros.Distro):
             kwargs["uid"] = str(kwargs["uid"])
 
         unsupported_busybox_values: Dict[str, Any] = {
-            "groups": None,
+            "groups": [],
             "expiredate": None,
             "inactive": None,
             "passwd": None,
@@ -317,34 +312,35 @@ class Distro(distros.Distro):
         LOG.debug("Adding user %s", name)
         try:
             subp.subp(adduser_cmd)
-        except Exception as e:
+        except subp.CalledProcessError as e:
             LOG.warning("Failed to create user %s", name)
             raise e
 
         # Process remaining options that Busybox's 'adduser' does not support
 
-        addn_groups = unsupported_busybox_values["groups"]
-        if addn_groups is not None:
-            # Separately add user to each additional group as Busybox's
-            # 'adduser' does not support specifying additional groups.
-            for addn_group in addn_groups:  # pylint: disable=E1133
-                LOG.debug("Adding user to group %s", addn_group)
-                try:
-                    subp.subp(["addgroup", name, addn_group])
-                except subp.ProcessExecutionError as e:
-                    util.logexc(
-                        LOG,
-                        "Failed to add user %s to group %s",
-                        name,
-                        addn_group,
-                    )
-                    raise e
+        # Separately add user to each additional group as Busybox's
+        # 'adduser' does not support specifying additional groups.
+        for addn_group in unsupported_busybox_values[
+            "groups"
+        ]:  # pylint: disable=E1133
+            LOG.debug("Adding user to group %s", addn_group)
+            try:
+                subp.subp(["addgroup", name, addn_group])
+            except subp.ProcessExecutionError as e:
+                util.logexc(
+                    LOG,
+                    "Failed to add user %s to group %s",
+                    name,
+                    addn_group,
+                )
+                raise e
 
-        passwd = unsupported_busybox_values["passwd"]
-        if passwd is not None:
+        if unsupported_busybox_values["passwd"]:
             # Separately set password as Busybox's 'adduser' does
             # not support passing password as CLI option.
-            super().set_passwd(name, passwd, hashed=True)
+            super().set_passwd(
+                name, unsupported_busybox_values["passwd"], hashed=True
+            )
 
         # Busybox's 'adduser' is hardcoded to always set the following field
         # values (numbered from "0") in /etc/shadow unlike 'useradd':
@@ -375,21 +371,21 @@ class Distro(distros.Distro):
         # Find the line in /etc/shadow for the user
         original_line = None
         for line in shadow_contents.splitlines():
-            current_user = line.split(":")[0]
-            if current_user == name:
+            new_line_parts = line.split(":")
+            if new_line_parts[0] == name:
                 original_line = line
+                break
 
-        if original_line is not None:
+        if original_line:
             # Modify field(s) in copy of user's shadow file entry
-            new_list = original_line.split(":")
             update_type = ""
 
             # Minimum password age
-            new_list[3] = ""
+            new_line_parts[3] = ""
             # Maximum password age
-            new_list[4] = ""
+            new_line_parts[4] = ""
             # Password warning period
-            new_list[5] = ""
+            new_line_parts[5] = ""
             update_type = "password aging"
 
             if expiredate is not None:
@@ -398,20 +394,20 @@ class Distro(distros.Distro):
                     datetime.fromisoformat(expiredate)
                     - datetime.fromisoformat("1970-01-01")
                 ).days
-                new_list[7] = str(days)
+                new_line_parts[7] = str(days)
                 if update_type != "":
                     update_type = update_type + " & "
                 update_type = update_type + "acct expiration date"
             if inactive is not None:
-                new_list[6] = inactive
+                new_line_parts[6] = inactive
                 if update_type != "":
                     update_type = update_type + " & "
                 update_type = update_type + "inactivity period"
 
-            new_line = ":".join(new_list)
-
             # Replace existing line for user with modified line
-            shadow_contents = shadow_contents.replace(original_line, new_line)
+            shadow_contents = shadow_contents.replace(
+                original_line, ":".join(new_line_parts)
+            )
             LOG.debug("Updating %s for user %s", update_type, name)
             try:
                 util.write_file(
@@ -433,8 +429,9 @@ class Distro(distros.Distro):
         # Check whether Shadow's or Busybox's version of 'passwd'.
         # If Shadow's 'passwd' is available then use the generic
         # lock_passwd function from __init__.py instead.
-        (_out, err) = subp.subp(["passwd", "--help"])
-        if not err.startswith("BusyBox"):
+        if os.path.islink("/usr/bin/passwd") and "busybox" not in os.readlink(
+            "/usr/bin/passwd"
+        ):
             return super().lock_passwd(name)
 
         cmd = ["passwd", "-l", name]
@@ -448,7 +445,7 @@ class Distro(distros.Distro):
             (_out, err) = subp.subp(cmd, rcs=[0, 1])
             if re.search(r"is already locked", err):
                 return True
-        except Exception as e:
+        except subp.CalledProcessError as e:
             util.logexc(LOG, "Failed to disable password for user %s", name)
             raise e
 
@@ -456,8 +453,9 @@ class Distro(distros.Distro):
         # Check whether Shadow's or Busybox's version of 'passwd'.
         # If Shadow's 'passwd' is available then use the generic
         # expire_passwd function from __init__.py instead.
-        (_out, err) = subp.subp(["passwd", "--help"])
-        if not err.startswith("BusyBox"):
+        if os.path.islink("/usr/bin/passwd") and "busybox" not in os.readlink(
+            "/usr/bin/passwd"
+        ):
             return super().expire_passwd(user)
 
         # Busybox's 'passwd' does not provide an expire option
@@ -473,23 +471,22 @@ class Distro(distros.Distro):
         # Find the line in /etc/shadow for the user
         original_line = None
         for line in shadow_contents.splitlines():
-            current_user = line.split(":")[0]
-            if current_user == user:
+            new_line_parts = line.split(":")
+            if new_line_parts[0] == user:
                 LOG.debug("Found /etc/shadow line matching user %s", user)
                 original_line = line
+                break
 
-        if original_line is not None:
+        if original_line:
             # Replace existing line for user with modified line
-            new_list = original_line.split(":")
             # Field '2' (numbered from '0') in /etc/shadow
             # is the "date of last password change".
-            if new_list[2] != "0":
+            if new_line_parts[2] != "0":
                 # Busybox's 'adduser' always expires password so only
                 # need to expire it now if this is not a new user.
-                new_list[2] = "0"
-                new_line = ":".join(new_list)
+                new_line_parts[2] = "0"
                 shadow_contents = shadow_contents.replace(
-                    original_line, new_line, 1
+                    original_line, ":".join(new_line_parts), 1
                 )
 
                 LOG.debug("Expiring password for user %s", user)
```

blackboxsw (approved):
Pending at least avoidance of an attempting to set_passwd on an empty string passwd value on line 344 of alpine.py. it'd be nice to check `os.readlink` on `/usr/bin/passwd` instead of stderr from --help. But not necessarily blocking.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+
+        # If 'useradd' is available then use the generic
+        # add_user function from __init__.py instead.
+        if subp.which("useradd"):
+            return super().add_user(name, **kwargs)
+
+        if "create_groups" in kwargs:
+            create_groups = kwargs.pop("create_groups")
+        else:
+            create_groups = True
```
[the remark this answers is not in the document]
dermotbradley: Done.

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+
+        adduser_flags = {"system": "-S"}
+
+        # support kwargs having groups=[list] or groups="g1,g2"
+        groups = kwargs.get("groups")
+        if groups:
+            if isinstance(groups, str):
+                groups = groups.split(",")
+
+            if isinstance(groups, dict):
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,374 @@ def preferred_ntp_clients(self):
...
+            # that came in as a string like: groups: group1, group2
+            groups = [g.strip() for g in groups]
+
+            # kwargs.items loop below wants a comma delimeted string
+            # that can go right through to the command.
+            kwargs["groups"] = ",".join(groups)
+
+            primary_group = kwargs.get("primary_group")
+            if primary_group:
+                groups.append(primary_group)
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+                    util.logexc(
+                        LOG,
+                        "Failed to add user %s to group %s",
+                        name,
+                        addn_group,
+                    )
+                    raise e
+
+        passwd = unsupported_busybox_values["passwd"]
+        if passwd is not None:
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+        except FileNotFoundError as e:
+            LOG.warning("Failed to read %s file, file not found", shadow_file)
+            raise e
+
+        # Find the line in /etc/shadow for the user
+        original_line = None
+        for line in shadow_contents.splitlines():
+            current_user = line.split(":")[0]
+            if current_user == name:
+                original_line = line
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+
+        if original_line is not None:
+            # Modify field(s) in copy of user's shadow file entry
+            new_list = original_line.split(":")
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+        except Exception as e:
+            util.logexc(LOG, "Failed to disable password for user %s", name)
+            raise e
+
+    def expire_passwd(self, user):
+        # Check whether Shadow's or Busybox's version of 'passwd'.
+        # If Shadow's 'passwd' is available then use the generic
+        # expire_passwd function from __init__.py instead.
+        (_out, err) = subp.subp(["passwd", "--help"])
+        if not err.startswith("BusyBox"):
```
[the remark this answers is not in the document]
dermotbradley: Done.

I had to change the logic to check if /usr/bin/passwd **not** being a softlink **or** the softlink not pointing to "**bbsuid**" (Alpine uses "bbsuid" binary rather than "busybox" for commands that require suid)

cloudinit/distros/alpine.py
dermotbradley: Done.

I had to change the logic to check if /usr/bin/passwd **not** being a softlink **or** the softlink not pointing to "**bbsuid**" (Alpine uses "bbsuid" binary rather than "busybox" for commands that require suid)

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+            shadow_contents = util.load_text_file(shadow_file)
+        except FileNotFoundError as e:
+            LOG.warning("Failed to read %s file, file not found", shadow_file)
+            raise e
+
+        # Find the line in /etc/shadow for the user
+        original_line = None
+        for line in shadow_contents.splitlines():
+            current_user = line.split(":")[0]
+            if current_user == user:
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
dermotbradley: Done

cloudinit/distros/alpine.py
```diff
@@ -202,6 +205,344 @@ def preferred_ntp_clients(self):
...
+        if create_groups and groups:
+            for group in groups:
+                if not util.is_group(group):
+                    self.create_group(group)
+                    LOG.debug("created group '%s' for user '%s'", group, name)
+        if "uid" in kwargs:
+            kwargs["uid"] = str(kwargs["uid"])
+
+        unsupported_busybox_values: Dict[str, Any] = {
+            "groups": None,
```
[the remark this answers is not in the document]
dermotbradley: Done

cloudinit/distros/alpine.py
dermotbradley: Done

cloudinit/distros/alpine.py
dermotbradley: Done

cloudinit/distros/alpine.py
dermotbradley: Done.

This was lifted from add_user in __init__.py so, like some of @TheRealFalcon previous comments, the same improvements could be made there also

cloudinit/distros/alpine.py
dermotbradley: Done.

Similar issue exists in lock_passwd in __init__.py so, like some of @TheRealFalcon previous comments, the same improvements could be made there also

dermotbradley:
Ready to merge now I guess...

blackboxsw:
Thank you for this correction @dermotbradley. I was basking my suggestion
off of adduser behavior as seen in incus/lxd alpine cloud images and
assumed it'd be comparable for /usr/bin/passwd

~ # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.19.1
PRETTY_NAME="Alpine Linux v3.19"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
~ # ls -l /usr/sbin/adduser
lrwxrwxrwx    1 root     root            12 Jan 26 17:53 /usr/sbin/adduser
-> /bin/busybox

PR #5180, opened 2024-04-17
fix(azure): Avoid non-primary nics from having routes to DNS CPC-4224

CalvoM:
## Proposed Commit Message

```
fix(azure): Avoid non-primary nics from having routes to DNS

From jammy, we have routes to dhcp by all NICS,since  RoutesToDns=true 
is the default in networkd.
```

## Additional Context

From jammy, RouteToDns=true is the default for systemd-networkd.
This causes slow resolution of urls and failures in journalctl.
We need to limit the route to DHCP to the primary NIC.

## Test Steps

```
CLOUD_INIT_PLATFORM=azure CLOUD_INIT_OS_IMAGE=jammy CLOUD_INIT_CLOUD_INIT_SOURCE=cloud-init-bddeb.deb tox -e integration-tests -- tests/integration_tests/datasources/test_azure.py::test_azure_multi_nic_setup
```

## Checklist

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -1482,18 +1482,23 @@ class DataSourceAzure(sources.DataSource):
     def _generate_network_config(self):
         """Generate network configuration according to configuration."""
         # Use IMDS network metadata, if configured.
+        fxn = add_dhcp_config_for_primary_and_secondary_nics_to_network_config
         if (
             self._metadata_imds
             and self._metadata_imds != sources.UNSET
             and self.ds_cfg.get("apply_network_config")
         ):
             try:
-                return generate_network_config_from_instance_network_metadata(
-                    self._metadata_imds["network"],
-                    apply_network_config_for_secondary_ips=self.ds_cfg.get(
-                        "apply_network_config_for_secondary_ips"
-                    ),
+                netcfg = (
+                    generate_network_config_from_instance_network_metadata(
+                        self._metadata_imds["network"],
+                        apply_network_config_for_secondary_ips=self.ds_cfg.get(
+                            "apply_network_config_for_secondary_ips"
+                        ),
+                    )
                 )
+                net_cfg = fxn(netcfg)
+                return net_cfg
             except Exception as e:
                 LOG.error(
                     "Failed generating network config "
@@ -1943,6 +1948,30 @@ def load_azure_ds_dir(source_dir):
 
     md, ud, cfg = read_azure_ovf(contents)
     return (md, ud, cfg, {"ovf-env.xml": contents})
+
+
+def add_dhcp_config_for_primary_and_secondary_nics_to_network_config(
+    netconfig: dict,
+) -> dict:
+    """
+    Configure the ephemeral and hotplug details.
+    parameters:
+        netconfig: dict, Network config details with other ethernet devices.
+
+    returns the updated network configuration
+    """
+    netconfig["ethernets"]["primary"] = {
+        "dhcp4": True,
+        "match": {"driver": "hv_netvsc", "name": "eth0"},
+    }
+    netconfig["ethernets"]["secondary"] = {
+        "dhcp4": True,
+        "dhcp4-overrides": {"use-dns": False},
+        "optional": True,
+        "match": {"driver": "hv_netvsc", "name": "!eth0"},
+    }
+
+    return netconfig
 
 
 @azure_ds_telemetry_reporter
--- a/tests/integration_tests/datasources/test_azure.py
+++ b/tests/integration_tests/datasources/test_azure.py
@@ -1,11 +1,15 @@
+import datetime
+
 import pytest
+import yaml
+from pycloudlib.azure.util import AzureCreateParams, AzureParams
 from pycloudlib.cloud import ImageType
 
 from tests.integration_tests.clouds import IntegrationCloud
 from tests.integration_tests.conftest import get_validated_source
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.integration_settings import PLATFORM
-from tests.integration_tests.releases import CURRENT_RELEASE
+from tests.integration_tests.releases import BIONIC, CURRENT_RELEASE
 
 
 def _check_for_eject_errors(
@@ -45,3 +49,42 @@ def test_azure_eject(session_cloud: IntegrationCloud):
                 session_cloud.cloud_instance.delete_image(snapshot_id)
         else:
             _check_for_eject_errors(instance)
+
+
+@pytest.mark.skipif(PLATFORM != "azure", reason="Test is Azure specific")
+@pytest.mark.skipif(
+    CURRENT_RELEASE < BIONIC, reason="Easier to test on Bionic+"
+)
+def test_azure_multi_nic_setup(
+    setup_image, session_cloud: IntegrationCloud
+) -> None:
+    """Integration test for https://warthogs.atlassian.net/browse/CPC-3999.
+
+    Azure should have the primary NIC only route to DNS.
+    Ensure other NICs do not have route to DNS.
+    """
+    us = datetime.datetime.now().strftime("%f")
+    rg_params = AzureParams(f"ci-test-multi-nic-setup-{us}", None)
+    nic_one = AzureCreateParams(f"ci-nic1-test-{us}", rg_params.name, None)
+    nic_two = AzureCreateParams(f"ci-nic2-test-{us}", rg_params.name, None)
+    with session_cloud.launch(
+        launch_kwargs={
+            "resource_group_params": rg_params,
+            "network_interfaces_params": [nic_one, nic_two],
+        }
+    ) as snapshot_instance:
+        _check_for_eject_errors(snapshot_instance)
+        if CURRENT_RELEASE.series == "bionic":
+            ret = snapshot_instance.execute("systemd-resolve --status")
+            assert ret.ok, ret.stderr
+            assert ret.stdout.count("Current Scopes: DNS") == 1
+        else:
+            ret = snapshot_instance.execute("resolvectl dns")
+            assert ret.ok, ret.stderr
+            routes = yaml.safe_load(ret.stdout)
+            routes_devices = list(routes.keys())
+            eth1_dev = [dev for dev in routes_devices if "(eth1)" in dev][0]
+            assert routes[eth1_dev] is None
+            # check the instance can resolve something
+            res = snapshot_instance.execute("resolvectl query google.com")
+            assert res.ok, res.stderr
--- a/tests/unittests/sources/test_azure.py
+++ b/tests/unittests/sources/test_azure.py
@@ -976,7 +976,17 @@ class TestNetworkConfig:
                     "dhcp6": False,
                     "match": {"macaddress": "00:0d:3a:04:75:98"},
                     "set-name": "eth0",
-                }
+                },
+                "primary": {
+                    "dhcp4": True,
+                    "match": {"driver": "hv_netvsc", "name": "eth0"},
+                },
+                "secondary": {
+                    "dhcp4": True,
+                    "dhcp4-overrides": {"use-dns": False},
+                    "optional": True,
+                    "match": {"driver": "hv_netvsc", "name": "!eth0"},
+                },
             },
             "version": 2,
         }
@@ -1557,7 +1567,17 @@ scbus-1 on xpt0 bus 0
                     "dhcp6": False,
                     "dhcp4": True,
                     "dhcp4-overrides": {"route-metric": 100},
-                }
+                },
+                "primary": {
+                    "dhcp4": True,
+                    "match": {"driver": "hv_netvsc", "name": "eth0"},
+                },
+                "secondary": {
+                    "dhcp4": True,
+                    "dhcp4-overrides": {"use-dns": False},
+                    "optional": True,
+                    "match": {"driver": "hv_netvsc", "name": "!eth0"},
+                },
             },
             "version": 2,
         }
@@ -1595,6 +1615,16 @@ scbus-1 on xpt0 bus 0
                     "dhcp4": True,
                     "dhcp4-overrides": {"route-metric": 300},
                 },
+                "primary": {
+                    "dhcp4": True,
+                    "match": {"driver": "hv_netvsc", "name": "eth0"},
+                },
+                "secondary": {
+                    "dhcp4": True,
+                    "dhcp4-overrides": {"use-dns": False},
+                    "optional": True,
+                    "match": {"driver": "hv_netvsc", "name": "!eth0"},
+                },
             },
             "version": 2,
         }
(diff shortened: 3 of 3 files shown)


aciba90:
Hey, @cjp256, could you also take a look at this one? Thanks.

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1945,6 +1948,29 @@ def load_azure_ds_dir(source_dir):
...
+    returns the updated network configuration
+    """
+    netconfig["ethernets"]["ephemeral"] = {
+        "dhcp4": True,
+        "dhcp4-overrides": {"use-dns": False},
+        "optional": True,
+        "match": {"driver": "hv_netvsc", "name": "!eth0"},
+    }
+
+    netconfig["ethernets"]["hotpluggedeth0"] = {
```
danpdraper, line 1966: How about naming the ethernets in the configuration `primary` and `secondary` instead of `hotpluggedeth0` and `ephemeral`? I believe that the former better expresses the relationship between the two. That said, I am open to other suggestions.

tests/integration_tests/datasources/test_azure.py
```diff
@@ -45,3 +48,43 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
         else:
             _check_for_eject_errors(instance)
+
+
+@pytest.mark.skipif(PLATFORM != "azure", reason="Test is Azure specific")
+def test_azure_multi_nic_setup(session_cloud: IntegrationCloud) -> None:
+    """Integration test for https://warthogs.atlassian.net/browse/CPC-3999
+
+    Azure should have the primary NIC only route to DNS.
+    Ensure other NICs do not have route to DNS.
```
danpdraper, line 58: I do not believe this test exercises the behaviour of concern. As described in https://warthogs.atlassian.net/browse/CPC-3999 , there is no route to the DHCP/DNS server through the secondary NIC on Mantic hosts (per `ip route`), but the DNS resolver has still associated the DNS server with the secondary NIC and thus transmits DNS queries through that NIC. As such, I believe that this test should instead verify that there is no longer a DNS server associated with the secondary NIC in the output of `resolvectl dns`.

danpdraper (changes requested):

danpdraper:
@CalvoM Please update the commit message. As described in https://warthogs.atlassian.net/browse/CPC-3999 , the association of the DNS server with the secondary NIC is a problem as of at least Focal. I would also suggest that you launch Bionic and Xenial VMs to see how far back the fix needs to be ported.

cjp256:
I think I've only noticed RoutesToDns=True default in 22.04+.  Will test this out.

Overall, I think maybe if this is an Azure-specific change, would it work to just conditionally add the use-dns flag here for all but the first nic?
https://github.com/CalvoM/cloud-init/blob/e41c2bca2535216336d09ba263e4130cbd13878a/cloudinit/sources/DataSourceAzure.py#L1993

tests/integration_tests/datasources/test_azure.py
```diff
@@ -45,3 +48,43 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
+                CURRENT_RELEASE.series, image_type=ImageType.GENERIC
+            ),
+        }
+    ) as instance:
+        source = get_validated_source(session_cloud)
+        if source.installs_new_version():
+            instance.install_new_cloud_init(source, clean=True)
+            snapshot_id = instance.snapshot()
+            try:
+                with session_cloud.launch(
```
aciba90, line 76: Instead of manually creating a base image, install new cloud-init, create a snapshot and launch the final test image, one can use the pytest fixture [setup_image](https://github.com/canonical/cloud-init/blob/8fe1494493f2cf2f6a04bba633be17937b93af50/tests/integration_tests/test_networking.py#L260). I guess this is a copy-paste from the other test in this file. This could be simplified using the aforementioned setup_image fixture.

aciba90 (changes requested):
Functionally, the approach looks good to me, thanks.

[SF: 00380708](https://bootstack.canonical.com/cases/00380708) says that the networkd change of defauts starts in jammy and onwards, do we need to patch this new config out for focal, or is this a noop change for focal and back? @CalvoM, @danpdraper.

tests/integration_tests/datasources/test_azure.py
CalvoM: I agree. Let me see how parse `resolvectl dns`. Thanks for this.

danpdraper:
> I think I've only noticed RoutesToDns=True default in 22.04+. Will test this out.
> 
> Overall, I think maybe if this is an Azure-specific change, would it work to just conditionally add the use-dns flag here for all but the first nic (that is, idx != 0)? https://github.com/CalvoM/cloud-init/blob/e41c2bca2535216336d09ba263e4130cbd13878a/cloudinit/sources/DataSourceAzure.py#L1993
> 
> Thanks for finding a solution!

@cjp256 Our investigation revealed that the association of the DNS server with the second NIC by `systemd-resolved` pre-dates Jammy (try running `resolvectl dns` on a Focal host). That said, DNS resolution latency only appears to be affected on Jammy+ hosts; the `cloud-init` logs on my Focal host reveal that the first instance of resolution of `azure.archive.ubuntu.com` was a sub-second operation. The `systemd-resolved.service(8)` man pages in both Focal and Jammy contain language to the following effect:

> The lookup requests that systemd-resolved.service receives are routed to the available DNS servers, LLMNR, and MulticastDNS interfaces according to the following rules:
> ...
> * Queries for multi-label names are routed via unicast DNS on local interfaces that have a DNS server configured, plus the globally configured DNS servers if there are any. Which interfaces are used is determined by the routing logic based on search and route-only domains, described below.
>
> ...
>
> If lookups are routed to multiple interfaces, the first successful response is returned...
> ...
> The following query routing logic applies for unicast DNS lookups initiated by systemd-resolved.service:
> ...
> * If a query does not match any configured routing domain (either per-link or global), it is sent to all DNS servers that are configured on links with the DefaultRoute= option set, as well as the globally configured DNS server.

Assuming the foregoing is consistent with the implementation, DNS requests should be transmitted through both NICs on both Focal and Jammy hosts (`DefaultRoute=Yes` is set for both NICs on both hosts). This suggests that resolution latency should be affected on both hosts. That said, the statement "[i]f lookups are routed to multiple interfaces, the first successful response is returned" is ambiguous. Does it mean that `systemd-resolved` will forward the first successful response to the client as soon as it arrives, or wait for all responses to come back (or the requests to time-out) before returning the first successful response? I wonder if the former is true in Focal, and the latter is true in Jammy+?

In any case, @CalvoM 's solution will remove the secondary NIC from the DNS resolution path and thus will address any associated resolution latency.

danpdraper:
> Functionally, the approach looks good to me, thanks.
> 
> [SF: 00380708](https://bootstack.canonical.com/cases/00380708) says that the networkd change of defauts starts in jammy and onwards, do we need to patch this new config out for focal, or is this a noop change for focal and back? @CalvoM, @danpdraper.

Good question. Even though resolution latency does not appear to be affected on hosts running pre-Jammy releases, we know that DNS resolution through the second NIC is not supported, so a suggestion in the `systemd-resolved` configuration that resolution through that NIC is supported would be a misrepresentation. As such, I believe we should back-port the change to Focal, as well as Bionic and Xenial if necessary (@CalvoM please confirm whether Bionic and Xenial are affected).

tests/integration_tests/datasources/test_azure.py
CalvoM: Bionic does not have `resolvectl`, so that will be interesting....will figure out how to update the test.

tests/integration_tests/datasources/test_azure.py
CalvoM: Updated.

tests/integration_tests/datasources/test_azure.py
```diff
@@ -45,3 +49,39 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
+    nic_one = AzureCreateParams(f"ci-nic1-test-{us}", rg_params.name, None)
+    nic_two = AzureCreateParams(f"ci-nic2-test-{us}", rg_params.name, None)
+    with session_cloud.launch(
+        launch_kwargs={
+            "resource_group_params": rg_params,
+            "network_interfaces_params": [nic_one, nic_two],
+        }
+    ) as snapshot_instance:
+        _check_for_eject_errors(snapshot_instance)
+        if CURRENT_RELEASE.series == "bionic":
```
danpdraper, line 77: What about Xenial?

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1945,6 +1948,28 @@ def load_azure_ds_dir(source_dir):
     return (md, ud, cfg, {"ovf-env.xml": contents})
 
 
+def configure_ephemeral_and_hotplug_device(netconfig: dict) -> dict:
```
danpdraper, line 1951: We should probably rename this function to something like `add_dhcp_config_for_primary_and_secondary_nics_to_network_config`. The doc string should be updated accordingly.

danpdraper (changes requested):

tests/integration_tests/datasources/test_azure.py
CalvoM: I am still trying to figure out how to reliably test on xenial.

CalvoM:
We had a discussion with @danpdraper and we see Xenial is not affected by this issue, so we will not backport it to xenial.

tests/integration_tests/datasources/test_azure.py
```diff
@@ -45,3 +49,39 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
+            ret = snapshot_instance.execute("systemd-resolve --status")
+            assert ret.ok
+            assert ret.stdout.count("Current Scopes: DNS") == 1
+        else:
+            ret = snapshot_instance.execute("resolvectl dns")
+            assert ret.ok
+            routes = yaml.safe_load(ret.stdout)
+            routes_devices = list(routes.keys())
+            eth1_dev = [dev for dev in routes_devices if "(eth1)" in dev][0]
+            assert routes[eth1_dev] is None
```
aciba90, line 87: Could we check that the instance can resolve some url?

```suggestion
            assert routes[eth1_dev] is None

            # check the instance can resolve something
            res = snapshot_instance.execute("resolvectl dns google.com")
            asssert res.ok, res.stderr
```

cloudinit/sources/DataSourceAzure.py
```diff
@@ -1945,6 +1950,30 @@ def load_azure_ds_dir(source_dir):
     return (md, ud, cfg, {"ovf-env.xml": contents})
 
 
+def add_dhcp_config_for_primary_and_secondary_nics_to_network_config(
```
aciba90, line 1953: Why is this config only applied to NICs with `hv_netvsc` as driver?

aciba90 (changes requested):
Thanks, @CalvoM, for fixing this. I have added some inline comments/requests.

tests/integration_tests/datasources/test_azure.py
```diff
@@ -45,3 +49,39 @@ def test_azure_eject(session_cloud: IntegrationCloud):
...
+    ) as snapshot_instance:
+        _check_for_eject_errors(snapshot_instance)
+        if CURRENT_RELEASE.series == "bionic":
+            ret = snapshot_instance.execute("systemd-resolve --status")
+            assert ret.ok
+            assert ret.stdout.count("Current Scopes: DNS") == 1
+        else:
+            ret = snapshot_instance.execute("resolvectl dns")
+            assert ret.ok
+            routes = yaml.safe_load(ret.stdout)
```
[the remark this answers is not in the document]
CalvoM: yeah I would try coming up with a logic to parse the output. I think I saw the output was something similar to 
```
Global:
Link 4 (enP13718s1):
Link 3 (eth1): 168.63.129.16
Link 2 (eth0): 168.63.129.16
```
Thus I thought yaml parsing will be easier.

cloudinit/sources/DataSourceAzure.py
CalvoM:
```
When a virtual machine (VM) is created in Azure, a synthetic network interface is created for each virtual NIC in its configuration. The synthetic interface is a VMbus device and uses the netvsc driver. Network packets that use this synthetic interface flow through the virtual switch in the Azure host and onto the datacenter's physical network.
```
```
The synthetic interface always has a name in the form eth\<n\>. Depending on the Linux distribution, the VF interface might have a name in the form eth\<n\>. Or it might have a different name in the form of enP\<n\> because of a udev rule that does renaming
```
Source: https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-how-it-works

tests/integration_tests/datasources/test_azure.py
CalvoM: I assume you meant `resolvectl query google.com`

cloudinit/sources/DataSourceAzure.py
aciba90: I understand that:

1. Synthetic interfaces are identifiable by the `hv_netvsc` driver.
2. Synthetic interfaces are the only affected interface kinds affected by the dns problem?

Could please add a breadcrumb pointing to this, as a comment and/or in the commit message?

aciba90: > The synthetic interface always has a name in the form eth\<n\>. 
Depending on the Linux distribution, the VF interface might have a name in the form eth\<n\>. 
Or it might have a different name in the form of enP\<n\> because of a udev rule that does renaming

Your proposed solution only matches against `eth*` names, should we also match against `enP\<n\>` ones? Or is there a reason I do not see to restrict this to `eth*`?

aciba90:
@cjp256:

> Overall, I think maybe if this is an Azure-specific change, would it work to just conditionally add the use-dns flag here for all but the first nic (that is, idx != 0)? https://github.com/CalvoM/cloud-init/blob/e41c2bca2535216336d09ba263e4130cbd13878a/cloudinit/sources/DataSourceAzure.py#L1993

That would simplify the fix, but per:

> The nics returned by the network call are not guaranteed to be in order.

in https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#response-1

It looks like cloud-init cannot rely on positions in the IMDS response to determine which NIC is the primary one. Or is that ensured?

cjp256:
> @cjp256:
> 
> > Overall, I think maybe if this is an Azure-specific change, would it work to just conditionally add the use-dns flag here for all but the first nic (that is, idx != 0)? https://github.com/CalvoM/cloud-init/blob/e41c2bca2535216336d09ba263e4130cbd13878a/cloudinit/sources/DataSourceAzure.py#L1993
> 
> That would simplify the fix, but per:
> 
> > The nics returned by the network call are not guaranteed to be in order.
> 
> in https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#response-1
> 
> It looks like cloud-init cannot rely on positions in the IMDS response to determine which NIC is the primary one. Or is that ensured?

The primary nic will always be first in the response.  The ordering of secondary NICs are not guaranteed today (but someday...).  That why I think all we need to do is simplify the change to just add the `use-dns` flag for idx=0:
https://github.com/CalvoM/cloud-init/blob/e41c2bca2535216336d09ba263e4130cbd13878a/cloudinit/sources/DataSourceAzure.py#L1993

aciba90:
Closing this as https://github.com/canonical/cloud-init/pull/5314 superseded.

PR #5184, opened 2024-04-17
test: bump pycloudlib to use latest version

TheRealFalcon:
## Proposed Commit Message

```
test: bump pycloudlib to use latest version

This is to ensure we can get the latest Azure noble images.
```

## Additional Context
I ran the CI subset of tests against Azure on Noble using this version and they passed.

## Merge type

### DIFF
--- a/integration-requirements.txt
+++ b/integration-requirements.txt
@@ -1,7 +1,7 @@
 # PyPI requirements for cloud-init integration testing
 # https://cloudinit.readthedocs.io/en/latest/topics/integration_tests.html
 #
-pycloudlib>=5.10.0,<1!6
+pycloudlib>=5.10.0,<1!7
 
 # Avoid breaking change in `testpaths` treatment forced
 # test/unittests/conftest.py to be loaded by our integration-tests tox env


blackboxsw (approved):
LGTM. Ran through a couple of tests against Azure ensuring we didn't see any immediate incompatibility across 5.x -> 6.x transition.

blackboxsw (changes requested):
[UPDATE] The below issue surfaced two issues: 1. pycloudlib also had a stale Azure minimal image offering which is fixed by https://github.com/canonical/pycloudlib/pull/361  and 2. the specific azure tests/integration_tests/datasources/test_azure.py::test_azure_eject failed because it was using ImageType.MINIMAL which has no add-apt-repository command. This second part is fixed by https://github.com/canonical/cloud-init/pull/5185

Rejecting my former approval after running through integration tests. This image stream actually seems to be minimize images as evidenced by presence of `ubuntu-cloud-minimal` deb package and /etc/motd announcing
```
This system has been minimized by removing packages and content that are
not required on a system that users do not log into.
```
We may need to find the correct image stream to source in pycloudlib for non-minimized images before moving to pycloudlib v. 6.5.0 in cloud-init integration tests.

blackboxsw:
Given the update to pyclloudlib to fix minimal image stream for Azure noble and a cloud-init integration test fix to ensure software-properties-common is installed when on minimal images, this PR can land.

blackboxsw (approved):
Approved.

Once: https://github.com/canonical/cloud-init/pull/5185 and https://github.com/canonical/pycloudlib/pull/361 land integration tests should be happy.

PR #5186, opened 2024-04-18
fix: Warn when signal is handled

holmanb:
```
fix: Warn when a signal is handled

Cloud-init is expected to run to completion. It does not expect to be
stopped or restarted. Therefore, cloud-init should loudly warn when this
happens so that the user is aware that the unexpected has happened. 

Problems:

- When systemd stops cloud-init, no warning is logged.
- When systemd restarts cloud-init, status info and return code from the
  current stage is lost.

The net effect of these problems is that when cloud-init is restarted
no warning is logged, and cloud-init status indicates no problem.

Communicate unexpected state by:

- Increase signal handling log level to ERROR, add log to stderr.
- Write status.json before exiting when sys.exit() is called.
- Warn when status.json contains existing data when the stage
  starts[1].
- Append existing errors and recoverable status.json errors rather
  than overwriting[2].

This should make cloud-init properly warn in status output and logs that
something is amiss.

[1] This should never happen and indicates that something restarted the
    service. When cloud-init doesn't have the opportunity to handle a
    signal (SIGKILL), other mitigations will fail.
[2] Pre-existing warnings were previously silently lost from
    `cloud-init status`'s output.

Fixes GH-5190
```

Fixes https://github.com/canonical/cloud-init/issues/5190

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -745,12 +745,7 @@ def status_wrapper(name, args):
     else:
         raise ValueError("unknown name: %s" % name)
 
-    if mode not in (
-        "init",
-        "init-local",
-        "modules-config",
-        "modules-final",
-    ):
+    if mode not in STAGE_NAME:
         raise ValueError(
             "Invalid cloud init mode specified '{0}'".format(mode)
         )
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -815,6 +815,12 @@ def status_wrapper(name, args):
         v1[mode]["errors"].append(str(e))
         raise
     except SystemExit as e:
+        # All calls to sys.exit() will resume running here. Before exiting,
+        # cloud-init will:
+        # 1) Write status.json (and result.json if in Final stage).
+        # 2) Write the final log message containing module run time.
+        # 3) Flush any queued reporting event handlers.
+        #
         # silence a pylint false positive
         # https://github.com/pylint-dev/pylint/issues/9556
         if e.code:  # pylint: disable=using-constant-test
@@ -823,7 +829,6 @@ def status_wrapper(name, args):
             LOG.exception("failed stage %s", mode)
             print_exc("failed run of stage %s" % mode)
             v1[mode]["errors"].append(f"sys.exit({str(e.code)}) called")
-        raise
     finally:
         # It is desireable to write to status.json before exiting, even when
         # sys.exit() is called.
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -788,7 +788,7 @@ def status_wrapper(name, args):
     if v1[mode]["start"] and not v1[mode]["finished"]:
         # This stage was restarted, which isn't expected.
         LOG.warning(
-            "Unexpected start time found for %s. " "Was this stage restarted?",
+            "Unexpected start time found for %s. Was this stage restarted?",
             STAGE_NAME[mode],
         )
 
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -737,18 +737,31 @@ def status_wrapper(name, args):
     else:
         raise ValueError("unknown name: %s" % name)
 
-    modes = (
+    if mode not in (
         "init",
         "init-local",
         "modules-config",
         "modules-final",
-    )
-    if mode not in modes:
+    ):
         raise ValueError(
             "Invalid cloud init mode specified '{0}'".format(mode)
         )
 
-    status = None
+    nullstatus = {
+        "errors": [],
+        "recoverable_errors": [],
+        "start": None,
+        "finished": None,
+    }
+    status = {
+        "v1": {
+            "datasource": None,
+            "init": nullstatus.copy(),
+            "init-local": nullstatus.copy(),
+            "modules-config": nullstatus.copy(),
+            "modules-final": nullstatus.copy(),
+        }
+    }
     if mode == "init-local":
         for f in (status_link, result_link, status_path, result_path):
             util.del_file(f)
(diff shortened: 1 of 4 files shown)


CalvoM:
Is there a way of reproducing the issue, so that testing the difference becomes easy?

cloudinit/cmd/main.py
```diff
@@ -764,6 +765,24 @@ def status_wrapper(name, args):
 
     v1 = status["v1"]
     v1["stage"] = mode
+    start_time = v1[mode]["start"]
```
TheRealFalcon, line 768: I think we should drop most of this section. It gives us a bunch of recoverable errors, and it winds up giving a messy status (that's also shown twice?)
```
root@me:~# cloud-init status --long
status: error
extended_status: error - done
last_update: Thu, 25 Apr 2024 20:00:25 +0000
detail: DataSourceLXD
errors:
	- sys.exit(1) called
recoverable_errors:
WARNING:
	- [modules-final] Unexpected value found in status.json. Was this stage restarted?
	- unexpected start_time: 1714075207.5654504
	- unexpected errors: ['sys.exit(1) called']
	- unexpected recoverable_errors: {'ERROR': ['Cloud-init 24.1.3 received SIGTERM, exiting...\n  Filename: /usr/lib/python3.12/subprocess.py\n  Function: _try_wait\n  Line number: 2011\n    Filename: /usr/lib/python3.12/subprocess.py\n    Function: _wait\n    Line number: 2053\n      Filename: /usr/lib/python3.12/subprocess.py\n      Function: wait\n      Line number: 1264', 'failed stage modules-final']}
	- [modules-final] Unexpected value found in status.json. Was this stage restarted?
	- unexpected start_time: 1714075207.5654504
	- unexpected errors: ['sys.exit(1) called']
	- unexpected recoverable_errors: {'ERROR': ['Cloud-init 24.1.3 received SIGTERM, exiting...\n  Filename: /usr/lib/python3.12/subprocess.py\n  Function: _try_wait\n  Line number: 2011\n    Filename: /usr/lib/python3.12/subprocess.py\n    Function: _wait\n    Line number: 2053\n      Filename: /usr/lib/python3.12/subprocess.py\n      Function: wait\n      Line number: 1264', 'failed stage modules-final']}
```

Way further down there's a section like this:
```python
        # merge new recoverable errors into exising recoverable error list
        old_recoverable_errors = v1[mode]["recoverable_errors"]
        new_recoverable_errors = next(
            filter(
                lambda h: isinstance(h, log.LogExporter), root_logger.handlers
            )
        ).export_logs()
        for key in new_recoverable_errors.keys():
            if key in old_recoverable_errors:
                old_recoverable_errors[key].extend(new_recoverable_errors[key])
            else:
                old_recoverable_errors[key] = new_recoverable_errors[key]
```

Could we turn this section into it's own function, and then rather than
```python
    v1[mode]["recoverable_errors"] = next(
        filter(lambda h: isinstance(h, log.LogExporter), root_logger.handlers)
    ).export_logs()
```

we call the new function to add the new recoverable errors to the old ones? The first run of cloud-final will log that it got a sigterm. I would think that should give enough signal as to what happened rather than introducing new warnings that are less precise.

TheRealFalcon (changes requested):
One major question inline

cloudinit/cmd/main.py
holmanb: Oops, those duplicates definitely aren't desired.

> The first run of cloud-final will log that it got a sigterm

That covers systemd's default behavior, yes, but what about if a SIGSTOP or SIGKILL happens?

If cloud-init receives a SIGSTOP or SIGKILL, our logging and `cloud-init status` tooling will not know that this happened. This section was a proposed solution to that problem.

TheRealFalcon: That makes sense. So here, can we check that start time exists, but finished time does not (indicating a run that started but never finished), and then combine the errors/recoverable errors to the output of the next run?

holmanb:
@TheRealFalcon I just pushed an update. Is that along the lines of what you had in mind?

Unittests are failing, but if this is what you were thinking I'll fix up the tests prior to merge.

cloudinit/cmd/main.py
```diff
@@ -727,18 +727,31 @@ def status_wrapper(name, args):
...
-    if mode not in modes:
+    ):
         raise ValueError(
             "Invalid cloud init mode specified '{0}'".format(mode)
         )
 
-    status = None
+    nullstatus = {
+        "errors": [],
+        "recoverable_errors": [],
```
TheRealFalcon, line 742:
```suggestion
        "recoverable_errors": {},
```

TheRealFalcon (commented):
Yes, I think the general idea makes sense.

If `cloud-final.service` is restarted:
```
root@me:~# systemctl restart cloud-final.service
Job for cloud-final.service failed because the control process exited with error code.
See "systemctl status cloud-final.service" and "journalctl -xeu cloud-final.service" for details.
root@me:~# cloud-init status --long
status: error
extended_status: error - done
last_update: Thu, 02 May 2024 04:37:23 +0000
detail: DataSourceLXD
errors:
	- sys.exit(1) called
recoverable_errors:
ERROR:
	- Cloud-init 24.1.3 received SIGTERM, exiting...   Filename: /usr/lib/python3.12/subprocess.py   Function: _try_wait   Line number: 2011     Filename: /usr/lib/python3.12/subprocess.py     Function: _wait     Line number: 2053       Filename: /usr/lib/python3.12/subprocess.py       Function: wait       Line number: 1264
	- failed stage modules-final
```
vs being killed:
```
root@me:~# ps aux | grep cloud-init
root         623  0.0  0.0  50444 39840 ?        Ss   04:38   0:00 /usr/bin/python3 /usr/bin/cloud-init modules --mode=final
root         627  0.0  0.0   2800  1600 ?        S    04:38   0:00 /bin/sh -c tee -a /var/log/cloud-init-output.log
root         628  0.0  0.0   6120  1600 ?        S    04:38   0:00 tee -a /var/log/cloud-init-output.log
root         645  0.0  0.0   7080  1760 pts/1    S+   04:38   0:00 grep --color=auto cloud-init
root@me:~# kill -9 623
root@me:~# systemctl restart cloud-final.service
root@me:~# cloud-init status --long
status: done
extended_status: degraded done
last_update: Thu, 02 May 2024 04:38:43 +0000
detail: DataSourceLXD
errors: []
recoverable_errors:
WARNING:
	- Unexpected start time for stage modules-final. Was this stage restarted?
```

tests/unittests/test_cli.py
```diff
@@ -414,3 +412,137 @@ def test_features_hook_subcommand(self, m_features):
...
+                FakeArgs(
+                    (
+                        None,
+                        lambda *_: sys.exit(1),
+                    ),
+                    False,
+                    "bogusmode",
+                ),
+            )
+            assert 1 == e.code
```
TheRealFalcon, line 472: We should assert here that the error message wasn't logged.

TheRealFalcon (changes requested):
A few more minor issues inline

cloudinit/cmd/main.py
```diff
@@ -795,27 +812,56 @@ def status_wrapper(name, args):
...
+    except SystemExit as e:
+        # silence a pylint false positive
+        # https://github.com/pylint-dev/pylint/issues/9556
+        if e.code:  # pylint: disable=using-constant-test
+            # Only log errors when sys.exit() is called with a non-zero
+            # exit code
+            LOG.exception("failed stage %s", mode)
+            print_exc("failed run of stage %s" % mode)
+            v1[mode]["errors"].append(f"sys.exit({str(e.code)}) called")
+        raise
```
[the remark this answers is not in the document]
holmanb: Good question - and I'm not sure myself.

Raising will cause cloud-init to exit more quickly, but there's not really much else that cloud-init does after this point in the code.

If we don't raise, I only see three more things that cloud-init will do before exiting:

1. write result.json (if in Final stage)
2. log completion of the stage and the time it took to complete("cloud-init mode %s")
3. flush reporting events

I lean slightly towards cloud-init doing the same behavior for both zero and non-zero exit modes. Prior to this change, none of these three were done when `sys.exit()` was called. Is there any harm in trying to write out these logs if they do happen? I assume that any service that systemd tries to restart with SIGTINT will just receive a bigger hammer after a timeout period to ensure that they actually restart, but I'm not certain of that.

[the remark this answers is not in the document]
holmanb: I'm leaning towards not re-raising from the `SystemExit` handler at all. What do you think @TheRealFalcon?

holmanb:
@TheRealFalcon I've addressed your remaining comments. Please let me know what you think of my comments regarding re-raising.

TheRealFalcon (approved):
Ship it!

TheRealFalcon (approved):
Ship it again!

Issue #4911, opened 2024-02-19
Cloud config schema errors: config.0.subnets.1: Additional properties are not allowed ('ipv6' was unexpected)


TomEros:
Hello !

# Bug report

Spawning a new Ubuntu 24.04 with Qemu on Openstack and get some warning in the cloud-init log 

`2024-02-16 16:29:56,175 - schema.py[WARNING]: Invalid network-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.`

```
# cloud-init schema --system
Found cloud-config data types: user-data, network-config

1. user-data at /var/lib/cloud/instances/0175e4fe-952d-4ee7-adc5-7982d184d0cf/cloud-config.txt:
Empty 'cloud-config' found at /var/lib/cloud/instances/0175e4fe-952d-4ee7-adc5-7982d184d0cf/cloud-config.txt. Nothing to validate.

2. network-config at /var/lib/cloud/instances/0175e4fe-952d-4ee7-adc5-7982d184d0cf/network-config.json:
  Invalid network-config /var/lib/cloud/instances/0175e4fe-952d-4ee7-adc5-7982d184d0cf/network-config.json
  Error: Cloud config schema errors: config.0.subnets.1: Additional properties are not allowed ('ipv6' was unexpected)

Error: Invalid schema: network-config
```

cat /var/lib/cloud/instances/0175e4fe-952d-4ee7-adc5-7982d184d0cf/network-config.json

```json
{
 "config": [
  {
   "accept-ra": false,
   "mac_address": "fa:16:3e:7f:1a:62",
   "mtu": 1500,
   "name": "ens3",
   "subnets": [
    {
     "type": "dhcp4"
    },
    {
     "address": "x:x:x:x::x",
     "ipv6": true,
     "netmask": "ffff:ffff:ffff:ff00::",
     "routes": [
      {
       "gateway": "x:x:x:x::1",
       "netmask": "::",
       "network": "::"
      }
     ],
     "type": "static6"
    }
   ],
   "type": "physical"
  },
  {
   "address": "1.1.1.1",
   "type": "nameserver"
  }
 ],
 "version": 1
}
```

I thought this was fixed by https://github.com/canonical/cloud-init/issues/4632

## Steps to reproduce the problem
you can spawn a vm on openstack with the image of Ubuntu 24.04
you need to have ipv4 and ipv6

## Environment details
- Cloud-init version: 24.1~5g1f6eddd5-0ubuntu1
[...]

PR #5191, opened 2024-04-18
fix: Add subnet ipv4/ipv6  to network schema

TheRealFalcon:
## Proposed Commit Message

```
fix: Add subnet ipv4/ipv6  to network schema

These are used by our openstack network_data.json parsing code and
get used by the sysconfig renderer.

Fixes GH-4911
```

## Additional Context
https://github.com/canonical/cloud-init/issues/4911
and

https://github.com/canonical/cloud-init/blob/70581bb0833f0ce77a2ca0ae1ee203ecc2a7bf62/cloudinit/net/sysconfig.py#L623-L636

## Merge type

### DIFF
--- a/cloudinit/config/schemas/schema-network-config-v1.json
+++ b/cloudinit/config/schemas/schema-network-config-v1.json
@@ -532,6 +532,14 @@
           "items": {
             "$ref": "#/$defs/anyOf_type_route"
           }
+        },
+        "ipv4": {
+          "type": "boolean",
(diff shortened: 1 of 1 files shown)


holmanb (approved):
I'm almost tempted to ask if we can reorder that config change to after validation rather than supporting this as a user-facing config option for one (not used anymore?) renderer, but I haven't looked to see how difficult that would be.

Either way, looks good to me!

PR #5193, opened 2024-04-19
doc: Describe ftp

holmanb:
## Proposed Commit Message
```
doc(nocloud): Describe ftp and ftp over tls implementation

Standardize and improve nocloud documentation.
```

PR #5204, opened 2024-04-23
doc(nocloud): Document network-config file

holmanb:
## Additional Context
Based on https://github.com/canonical/cloud-init/pull/5193#pullrequestreview-2013137727


doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -106,15 +132,19 @@ port to use (default is ``21``).
 Path Resource
 -------------
 
-The path pointed to by the URI will contain the following files:
+The path pointed to by the URI must contain the following files:
```
blackboxsw, line 135:
```suggestion
The path pointed to by the URI must contain the following files or it will be ignored:
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -21,8 +21,8 @@ Configuration Methods:
 Method 1: Labeled filesystem
 ----------------------------
 
-A labeled `vfat`_ or `iso9660` filesystem containing user data and metadata
-files may be used. The filesystem volume must be labelled ``CIDATA``.
+A labeled `vfat`_ or `iso9660` filesystem may be used. The filesystem volume
```
blackboxsw, line 24: Note that the even disks labeled `CIDATA`  which don't have both `meta-data` and `user-data` files will be ignored. I think that's what the original text was trying to allude to. But, it fell short of that goal. Maybe in the Path resource section below we can mention `The path pointed to by the URI must contain the following files or it will be ignored`

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -65,19 +78,32 @@ tell cloud-init where on the filesystem to look.
...
+The only required key is:
 
-* ``h`` or ``local-hostname``
-* ``i`` or ``instance-id``
 * ``s`` or ``seedfrom``
 
-A valid ``seedfrom`` value consists of a URI which must contain a
-trailing ``/``.
+A valid ``seedfrom`` value consists of a URI which must contain a trailing
+``/``. The short key ``s`` is only supported by kernel commandline or
```
blackboxsw, line 99: The short keys/aliases s h and i are all intended only for use on the kernel command-line is it worth separating this "The short keys ``s`` , ``h`` and ``i`` are only supported when reading values from kernel commandline or SMBIOS. When using a base configuration ``*.cfg`` files, the long key name is required.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -65,19 +78,32 @@ tell cloud-init where on the filesystem to look.
...
+SMBIOS, when using a ``*.cfg`` file, ``seedfrom`` is required.
```
blackboxsw, line 100:
```suggestion
SMBIOS. When using a ``*.cfg`` file, ``seedfrom`` is required.
```

blackboxsw (commented):
LGTM, couple of nits.

doc/rtd/reference/datasources/nocloud.rst
holmanb: it would be awkward to describe the optional keys under the section describing the required key, but I can move this out into a note

holmanb:
@blackboxsw I think that I've addressed your comments.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -65,22 +78,44 @@ tell cloud-init where on the filesystem to look.
...
+  datasource:
+    NoCloud:
+      seedfrom: file://10.42.42.42/cloud-init/configs/
+
 
 Permitted keys
 ==============
 
-The permitted keys are:
+Currenty three keys (and their aliases) are permitted for configuring
```
blackboxsw, line 94: minor typo
```suggestion
Currently three keys (and their aliases) are permitted for configuring
```

blackboxsw (approved):
LGTM, minor typo and a question about command line vs commandline.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -65,22 +78,44 @@ tell cloud-init where on the filesystem to look.
...
+be removed in the future.
+
+* ``local-hostname`` alias: ``h`` (:ref:`cloud-config<mod-set_hostname>`
+  preferred)
+* ``instance-id`` alias: ``i``  (set instance id  in :file:`meta-data` instead)
+
+.. note::
+
+   The aliases ``s`` , ``h`` and ``i`` are only supported by kernel
+   commandline or SMBIOS. When configured in a ``*.cfg`` file, the long key
```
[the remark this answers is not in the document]
holmanb: will update the 15 incorrect instances of commandline

holmanb:
@blackboxsw I've updated the docs, docstrings comments in cloud-init, and user-facing text with the correct spelling. Some file names and function names remain incorrect, but fixing those would be more invasive so for now they stay.

doc/rtd/explanation/index.rst
```diff
@@ -19,7 +19,7 @@ knowledge and become better at using and configuring ``cloud-init``.
    vendordata.rst
    security.rst
    analyze.rst
-   kernel-cmdline.rst
```
blackboxsw, line 22: Let's avoid doc file renaming throughout unless we have a really compelling reason This causes the generated docs generated to be new html pages and routes and we'd also need to setup redirect rules in RTD for each of these changes in the event of external links in blog posts etc.

blackboxsw (commented):
Changes look good thanks, let's just avoid any top-level docs file renames to avoid thrashing with potential external links to cloud-init docs.

blackboxsw (changes requested):
Let's avoid rst doc file renames, otherwise +1.

blackboxsw (approved):
LGTM. I've reverted your last commit based on the fact that I disagree with my former review comment. We should move toward the best represented document names/slugs/urls we can and sort historic links/aliases and redirects appropriately in RTD configuration. Eventually be may run out of redirects on the project for renames as I think we may be capped at 100 redirects on the proejct, but it makes little sense to avoid common sense URL/slug/page renames which improve discoverability an use common terminology throughout cloud-init.

PR #5208, opened 2024-04-24
Do not hardcode reboot command

holmanb:
## Proposed Commit Message
```
fix(cmd): Do not hardcode reboot command

Attempting a `cloud-init clean --reboot` fails on alpine because this
command is hard-coded. This code already has an Init object available,
which has a Distro attribute. Stamp out the duplicate hard-coded
implementation and re-use distro reboot code to acquire cross-distro
compatiblity.

Error:
Could not reboot this system using "['shutdown', '-r', 'now']": Unexpected error while running comma  nd.
Command: ['shutdown', '-r', 'now']
Exit code: -
Reason: [Errno 2] No such file or directory: b'shutdown'
Stdout: -
Stderr: -
```

# Context
This change is mostly test updates - the code change is trivial: just reuse the existing reboot code which already has multi-distro support.

### DIFF
--- a/tests/unittests/test_traceback.py
+++ b/tests/unittests/test_traceback.py
@@ -1,0 +1,9 @@
+import pytest
+
+
+class TestLogExc:
+    def test_logexc(self, caplog):
+        with pytest.raises(Exception):
+            _ = 1 / 0
+
+        assert caplog.record_tuples == []
--- a/tests/unittests/test_traceback.py
+++ b/tests/unittests/test_traceback.py
@@ -1,9 +1,0 @@
-import pytest
-
-
-class TestLogExc:
-    def test_logexc(self, caplog):
-        with pytest.raises(Exception):
-            _ = 1 / 0
-
-        assert caplog.record_tuples == []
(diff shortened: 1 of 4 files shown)


dermotbradley:
I guess I've never tried a `cloud-init clean --reboot` before.

holmanb:
> I guess I've never tried a `cloud-init clean --reboot` before.

It's honestly not much of a value add, so I'm not terribly surprised. I like it because it shaves off a command when iterating for development and it gets used a fair amount in integration testing so I figure it's worth supporting for all distros.

tests/unittests/test_traceback.py
```diff
@@ -0,0 +1,9 @@
+import pytest
```
TheRealFalcon, line 1: What's the point of this file?

TheRealFalcon (approved):
One question, but otherwise LGTM!

tests/unittests/test_traceback.py
holmanb: thanks, I'll drop this

PR #5209, opened 2024-04-24
feat: Warn user of unexpected run mode

holmanb:
## Proposed Commit Message
```
feat: Warn user of unexpected run mode
  
On systemd, services are started by PID 1. When this doesn't happen, cloud-init
is in an unknown run state and should warn the user.

- Reorder pid log to be able to reuse Distro information.
- Add docstring deprecating util.is_Linux().
```

## Merge type

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -79,17 +79,29 @@ def print_exc(msg=""):
     sys.stderr.write("\n")
 
 
-def log_ppid():
-    if util.is_Linux():
+def log_ppid(distro, bootstage_name):
+    if distro.is_linux:
         ppid = os.getppid()
-        LOG.info("PID [%s] started cloud-init.", ppid)
+        log = LOG.info
+        extra_message = ""
+        if 1 != ppid and distro.uses_systemd():
+            log = LOG.warning
+            extra_message = (
+                " Unsupported configuration: boot stage called "
+                "outside of systemd"
+            )
+        log(
+            "PID [%s] started cloud-init '%s'.%s",
+            ppid,
+            bootstage_name,
+            extra_message,
+        )
 
 
 def welcome(action, msg=None):
     if not msg:
         msg = welcome_format(action)
     util.multi_log("%s\n" % (msg), console=False, stderr=True, log=LOG)
-    log_ppid()
     return msg
 
 
@@ -333,10 +345,8 @@ def main_init(name, args):
     #    objects config as it may be different from init object
     # 10. Run the modules for the 'init' stage
     # 11. Done!
-    if not args.local:
-        w_msg = welcome_format(name)
-    else:
-        w_msg = welcome_format("%s-local" % (name))
+    bootstage_name = "init-local" if args.local else "init"
+    w_msg = welcome_format(bootstage_name)
     init = stages.Init(ds_deps=deps, reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
@@ -364,6 +374,7 @@ def main_init(name, args):
     # config applied.  We send the welcome message now, as stderr/out have
     # been redirected and log now configured.
     welcome(name, msg=w_msg)
+    log_ppid(init.distro, bootstage_name)
 
     # re-play early log messages before logging was setup
     for lvl, msg in early_logs:
@@ -591,7 +602,8 @@ def main_modules(action_name, args):
     #    the modules objects configuration
     # 5. Run the modules for the given stage name
     # 6. Done!
-    w_msg = welcome_format("%s:%s" % (action_name, name))
+    bootstage_name = "%s:%s" % (action_name, name)
+    w_msg = welcome_format(bootstage_name)
     init = stages.Init(ds_deps=[], reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
@@ -628,6 +640,7 @@ def main_modules(action_name, args):
 
     # now that logging is setup and stdout redirected, send welcome
     welcome(name, msg=w_msg)
+    log_ppid(init.distro, bootstage_name)
 
     if name == "init":
         util.deprecate(
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -79,23 +79,30 @@ def print_exc(msg=""):
     sys.stderr.write("\n")
 
 
+DEPRECATE_BOOT_STAGE_MESSAGE = (
+    "Triggering cloud-init boot stages outside of intial system boot is not a"
+    " fully supported operation which can lead to incomplete or incorrect"
+    " configuration. As such, cloud-init is deprecating this feature in the"
+    " future. If you currently use cloud-init in this way,"
+    " please file an issue describing in detail your use case so that"
+    " cloud-init can better support your needs:"
+    " https://github.com/canonical/cloud-init/issues/new"
+)
+
+
 def log_ppid(distro, bootstage_name):
     if distro.is_linux:
         ppid = os.getppid()
-        log = LOG.info
-        extra_message = ""
         if 1 != ppid and distro.uses_systemd():
-            log = LOG.warning
-            extra_message = (
-                " Unsupported configuration: boot stage called "
-                "outside of systemd"
+            util.deprecate(
+                deprecated=(
+                    "Unsupported configuration: boot stage called "
+                    f"by PID [{ppid}] outside of systemd"
+                ),
+                deprecated_version="24.3",
+                extra_message=DEPRECATE_BOOT_STAGE_MESSAGE,
             )
-        log(
-            "PID [%s] started cloud-init '%s'.%s",
-            ppid,
-            bootstage_name,
-            extra_message,
-        )
+    LOG.info("PID [%s] started cloud-init '%s'.", ppid, bootstage_name)
 
 
 def welcome(action, msg=None):
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -495,6 +495,12 @@ def multi_log(
 
 @lru_cache()
 def is_Linux():
+    """deprecated: prefer Distro object's `is_linux` property
+
+    Multiple sources of truth is bad, and already know whether we are
+    working with Linux from the Distro class. Using Distro offers greater code
+    reusablity, cleaner code, and easier maintenance.
+    """
     return "Linux" in platform.system()
 
 
--- a/tests/integration_tests/util.py
+++ b/tests/integration_tests/util.py
@@ -66,6 +66,7 @@ def verify_clean_boot(
     instance: "IntegrationInstance",
     ignore_warnings: Optional[Union[List[str], bool]] = None,
     ignore_errors: Optional[Union[List[str], bool]] = None,
+    ignore_tracebacks: Optional[Union[List[str], bool]] = None,
     require_warnings: Optional[list] = None,
     require_errors: Optional[list] = None,
 ):
@@ -94,14 +95,17 @@ def verify_clean_boot(
 
     def append_or_create_list(
         maybe_list: Optional[Union[List[str], bool]], value: str
-    ) -> List[str]:
+    ) -> Optional[Union[List[str], bool]]:
         """handle multiple types"""
         if isinstance(maybe_list, list):
             maybe_list.append(value)
-        elif maybe_list is None or isinstance(maybe_list, bool):
+        elif maybe_list is True:
+            return True  # Ignoring all texts, so no need to append.
+        elif maybe_list in (None, False):
             maybe_list = [value]
         return maybe_list
 
+    traceback_texts = []
     # Define exceptions by matrix of platform and Ubuntu release
     if "azure" == PLATFORM:
         # Consistently on all Azure launches:
@@ -122,11 +126,16 @@ def verify_clean_boot(
         ignore_errors = append_or_create_list(
             ignore_warnings, "Stderr: RTNETLINK answers: File exists"
         )
+        traceback_texts.append("Stderr: RTNETLINK answers: File exists")
         # LP: #1833446
         ignore_warnings = append_or_create_list(
             ignore_warnings,
             "UrlError: 404 Client Error: Not Found for url: "
             "http://169.254.169.254/latest/meta-data/",
+        )
+        traceback_texts.append(
+            "UrlError: 404 Client Error: Not Found for url: "
+            "http://169.254.169.254/latest/meta-data/"
         )
         # Oracle has a file in /etc/cloud/cloud.cfg.d that contains
         # users:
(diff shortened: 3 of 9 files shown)


cloudinit/cmd/main.py
```diff
@@ -324,6 +328,7 @@ def main_init(name, args):
     init = stages.Init(ds_deps=deps, reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
+    log_ppid(init.distro)
```
blackboxsw, line 331: This cannot happen until after `log.setup_logging` has been called.

cloudinit/cmd/main.py
```diff
@@ -580,6 +585,7 @@ def main_modules(action_name, args):
     init = stages.Init(ds_deps=[], reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
+    log_ppid(init.distro)
```
blackboxsw, line 588: Need to move this down a bit too after `log.setup_logging` otherwise no logs are written to /var/log/cloud-init.log. To retain original behavior, we may want to log this just after the `welcome(... w_msg)` to log it during the boot stage block.

cloudinit/cmd/main.py
```diff
@@ -72,17 +72,21 @@ def print_exc(msg=""):
...
-    if util.is_Linux():
+def log_ppid(distro):
+    if distro.is_linux:
         ppid = os.getppid()
-        LOG.info("PID [%s] started cloud-init.", ppid)
+        log = LOG.info
+        extra_message = ""
+        if 1 != ppid and distro.uses_systemd():
+            log = LOG.warning
+            extra_message = ("Not a supported configuration.",)
```
blackboxsw, line 82: Let's add details in the extra message to tell use what's unsupported about it.
`Unsupported configuration: boot stage called outside of systemd`

cloudinit/util.py
```diff
@@ -490,6 +490,12 @@ def multi_log(
 
 @lru_cache()
 def is_Linux():
+    """deprecated: prefer Distro object's `is_linux` property
```
blackboxsw, line 493: We could just drop this function altogether if we can figure a better approach to sort `cloud-init  analyze dump` as the only other use-case right?

cloudinit/cmd/main.py
```diff
@@ -72,17 +72,21 @@ def print_exc(msg=""):
...
+        log("PID [%s] started cloud-init. %s", ppid, extra_message)
```
blackboxsw, line 83: Are we intentionally not logging the PID when `not distro.is_linux`?

blackboxsw (changes requested):
Just checking some boxes:
- ✔ Alpine (openrc-based) will invoke cloud-init boot stages from non-1 PPID, but we are ignoring non-systemd cases
- ✔non-linux and therefore non-systemd is ignored

Here's a suggested diff to ensure the logs are actually written, and the log message has a bit more data about the launched boot stage. It doesn't resolve either the lack of logging PID on non-linux/non-systemd or the dropping of `util.is_Linux`.
```diff
diff --git a/cloudinit/cmd/main.py b/cloudinit/cmd/main.py
index ad9910e04..aff4d380f 100644
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -72,15 +72,20 @@ def print_exc(msg=""):
     sys.stderr.write("\n")
 
 
-def log_ppid(distro):
+def log_ppid(distro, bootstage_name):
     if distro.is_linux:
         ppid = os.getppid()
         log = LOG.info
         extra_message = ""
         if 1 != ppid and distro.uses_systemd():
             log = LOG.warning
-            extra_message = ("Not a supported configuration.",)
-        log("PID [%s] started cloud-init. %s", ppid, extra_message)
+            extra_message = " Unsupported configuration: boot stage called outside of systemd"
+        log(
+            "PID [%s] started cloud-init '%s'.%s",
+            ppid,
+            bootstage_name,
+            extra_message,
+        )
 
 
 def welcome(action, msg=None):
@@ -321,14 +326,11 @@ def main_init(name, args):
     #    objects config as it may be different from init object
     # 10. Run the modules for the 'init' stage
     # 11. Done!
-    if not args.local:
-        w_msg = welcome_format(name)
-    else:
-        w_msg = welcome_format("%s-local" % (name))
+    bootstage_name = "%s-local" % (name) if args.local else name
+    w_msg = welcome_format(bootstage_name)
     init = stages.Init(ds_deps=deps, reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
-    log_ppid(init.distro)
     # Stage 2
     outfmt = None
     errfmt = None
@@ -354,6 +356,7 @@ def main_init(name, args):
     # config applied.  We send the welcome message now, as stderr/out have
     # been redirected and log now configured.
     welcome(name, msg=w_msg)
+    log_ppid(init.distro, bootstage_name)
 
     # re-play early log messages before logging was setup
     for lvl, msg in early_logs:
@@ -581,11 +584,11 @@ def main_modules(action_name, args):
     #    the modules objects configuration
     # 5. Run the modules for the given stage name
     # 6. Done!
-    w_msg = welcome_format("%s:%s" % (action_name, name))
+    bootstage_name = "%s:%s" % (action_name, name)
+    w_msg = welcome_format(bootstage_name)
     init = stages.Init(ds_deps=[], reporter=args.reporter)
     # Stage 1
     init.read_cfg(extract_fns(args))
-    log_ppid(init.distro)
     # Stage 2
     try:
         init.fetch(existing="trust")
@@ -620,6 +623,7 @@ def main_modules(action_name, args):
 
     # now that logging is setup and stdout redirected, send welcome
     welcome(name, msg=w_msg)
+    log_ppid(init.distro, bootstage_name)
 
     if name == "init":
         util.deprecate(
```

cloudinit/cmd/main.py
holmanb: That's what the code did before, but I suppose we should be able to log PPID on non-linux too. I'm curious whether the PPID=1 should hold true on any of the BSDs, and it wouldn't hurt to log it.

cloudinit/util.py
holmanb: yep!

blackboxsw (changes requested):
One thing I think we need to sort here is integration tests which call cloud-init --local over SSH as they'll now exit non-zero due to the warning.

Tests affected are at least the following which both check `response.ok` from a cloud-init init --local call.
- tests/integration_tests/datasources/test_ec2_ipv6.py
- tests/integration_tests/test_upgrade.py

We might be able to adapt those tests to generally disregard exit 2 or maybe specifically call verify_clean_boot with a non-empty ignore_warnings item to ignore this expected warning log.

TheRealFalcon:
Because we patched out the `exit 2` behavior on old releases, there's no backwards incompatibility here, correct? Just an extra warning in the logs?

blackboxsw:
> Because we patched out the `exit 2` behavior on old releases, there's no backwards incompatibility here, correct? Just an extra warning in the logs?
> 
> We would see an `exit 2` when we SRU to Noble though...

yep and yep. I think it's worth a followup quilt patch in ubuntu/noble:debian/patches/retain-ppid-unsupported-config-as-info-level.patch to ensure we [don't set warning level on the logs in noble](https://github.com/canonical/cloud-init/pull/5209/files#diff-5978682d3313395486da4847761e13e175e6303251fe7ddf81552db99779760eR81).

holmanb:
> yep and yep. I think it's worth a followup quilt patch in ubuntu/noble:debian/patches/retain-ppid-unsupported-config-as-info-level.patch to ensure we [don't set warning level on the logs in noble](https://github.com/canonical/cloud-init/pull/5209/files#diff-5978682d3313395486da4847761e13e175e6303251fe7ddf81552db99779760eR81).

we could always just make this a deprecation log and deal with this in the same way that we choose to deal with other new deprecations on older series - a patch per log like this seems really expensive to maintain in the long run

blackboxsw (commented):
So on stable releaeses, we'll patch the warning to a deprecation. Only thing left to land this PR is integration test fixes per [this comment](https://github.com/canonical/cloud-init/pull/5209#pullrequestreview-2029527908)

blackboxsw:
I think this is just waiting on two integration test fixes. But, I'd like this in 24.3 as we are introducing a "breaking change" as far as warnings being emitted where there were not warnings before-hand and I'd like us to avoid trying to quickly stuff quilt patches into /noble for this across the SRU boundary for 24.2

holmanb:
> I think this is just waiting on two integration test fixes. But, I'd like this in 24.3 as we are introducing a "breaking change" as far as warnings being emitted where there were not warnings before-hand and I'd like us to avoid trying to quickly stuff quilt patches into /noble for this across the SRU boundary for 24.2

Resolving this involves either:

1) ignore this warning always
2) do this somehow without using `verify_clean_log()`

I don't think that we want 1) (it involves muddying `verify_clean_log()` with even more single-use ignores aka tech debt), but 2) requires implementing what was described [here](https://github.com/canonical/cloud-init/pull/4834#discussion_r1529551537):

> One idea that I envision for this helper is a matrix of platforms and series that allows us to specify exactly when an expected warning message that will be seen. This would be more precise than the current "ignore this pattern everywhere" paradigm that verify_clean_log() uses. That said, I'm not sure that I want to include that refactor in this PR, which has already grown sizeable.

In order to make `verify_clean_log()` truly be a general-purpose drop-in replacement for `verify_clean_boot()`, I think we need to grow the ability to track tracebacks in `cloud-init status --format json` (which would be useful for users anyways). Unfortunately, that is not something that will happen quickly, but also I don't think that this is a strong requirement today for implementing this test, since ignoring tracebacks is an oracle-only requirement.

I'll share what I've been working on shortly.

holmanb:
@blackboxsw Last commit contains what I had in mind.

With this commit I can successfully run `tests/integration_tests/test_upgrade.py::test_clean_boot_of_upgraded_package` and get an instance with this log:
```
# grep WARN /var/log/cloud-init.log 
2024-06-04 01:07:26,672 - main.py[WARNING]: PID [2082] started cloud-init 'init-local'. Unsupported configuration: boot stage called outside of systemd
2024-06-04 01:07:27,299 - main.py[WARNING]: PID [2082] started cloud-init 'init'. Unsupported configuration: boot stage called outside of systemd
2024-06-04 01:07:27,956 - main.py[WARNING]: PID [2082] started cloud-init 'modules:config'. Unsupported configuration: boot stage called outside of systemd
2024-06-04 01:07:28,548 - main.py[WARNING]: PID [2082] started cloud-init 'modules:final'. Unsupported configuration: boot stage called outside of systemd
```

And it passes.

Thoughts?

blackboxsw:
Thanks @holmanb . I've added two supplemental commits to your branch for reviiew
- move from LOG.warning to util.deprecate() so we can better prompt any consumers of this feature that it will go away. It also allows us to reduce this deprectaion from WARNING level to INFO level on stable releases to avoid non-zero exit codes in the future once the feature spec'd at WIP brach #5411 lands
- fixup verify_clean_boot to support `ignore_(warnings|errrors|tracebacks)=True` params just like verify_clean_log supports. Allowing individual tracebacks to be ignored or all tracebacks if `ignore_tracebacks=True`.

cloudinit/cmd/main.py
```diff
@@ -72,23 +72,32 @@ def print_exc(msg=""):
...
 
 
+DEPRECATE_BOOT_STAGE_MESSAGE = (
+    "Triggering cloud-init boot stages outside of intial system boot is not a"
+    " fully supported operation which can lead to incomplete of incorrect"
+    " configuration due to either race-conditions with other active services"
+    " or incomplete triggering of all cloud-init boot stages. As such,"
+    " cloud-init is deprecating this undocumented feature in the future. If"
+    " there are use-cases in an environment which requires calling cloud-init"
+    " boot stages sometime after boot, please file an issue to allow"
```
holmanb, line 82: Lets tell the users what we are looking for, exactly, from this request to file an issue.


Something like this maybe?

"If you currently use cloud-init in this way, please file an issue describing in detail your use case so that cloud-init can better support your needs."

tests/integration_tests/util.py
```diff
@@ -146,13 +155,27 @@ def append_or_create_list(
...
+    if ignore_tracebacks:
+        for expected_traceback in ignore_tracebacks:
+            expected_traceback_count += int(
+                instance.execute(
+                    f"grep --count '{expected_traceback}'"
+                    " /var/log/cloud-init.log"
+                ).stdout.strip()
+            )
+    assert expected_traceback_count == traceback_count, (
+        f"{traceback_count - expected_traceback_count} unexpected traceback(s)"
```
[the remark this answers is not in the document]
blackboxsw: I agree with this too, I did also look at processing/extracting tracebacks from the cloud-init.log and thought it a bit more complex than trying to piggy back on this PR. But, agreed a richer UX would be helpful in the event of failures and manually jumping into the logged artifacts to get this data after the fact is painful.

cloudinit/cmd/main.py
```diff
@@ -72,23 +72,32 @@ def print_exc(msg=""):
     sys.stderr.write("\n")
 
 
+DEPRECATE_BOOT_STAGE_MESSAGE = (
+    "Triggering cloud-init boot stages outside of intial system boot is not a"
+    " fully supported operation which can lead to incomplete of incorrect"
+    " configuration due to either race-conditions with other active services"
```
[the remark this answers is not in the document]
blackboxsw: +1

cloudinit/cmd/main.py
```diff
@@ -72,23 +72,32 @@ def print_exc(msg=""):
...
+    " or incomplete triggering of all cloud-init boot stages. As such,"
+    " cloud-init is deprecating this undocumented feature in the future. If"
```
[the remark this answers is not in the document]
blackboxsw: That probably makes sense in this PR too.

tests/integration_tests/util.py
```diff
@@ -146,13 +155,27 @@ def append_or_create_list(
         require_warnings=require_warnings,
         require_errors=require_errors,
     )
-    # assert no Tracebacks
-    assert (
-        "0"
-        == instance.execute(
+    if ignore_tracebacks is True:
```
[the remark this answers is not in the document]
blackboxsw: @holmanb yes will do.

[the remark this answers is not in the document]
blackboxsw: done

blackboxsw:
@holmanb I think I resolved your last pass of review comments. Please take a look and make sure I didn't miss anything.

blackboxsw:
@holmanb I think we want this #5209 in milestone 24.2 only if #5491 merged because #5491will give us an easy downstream/stable release mechanism to avoid warnings on "new" deprecation messages introduced across SRU boundary. If we think 5491 will be reviewable for milestone 24.2 today/tomorrow, I'll spend some time on that PR now and could provide some feedback here to get 5491 across the line.

blackboxsw:
I think now that milestone 24.2 has been released. We can merge this pending your approval @holmanb

blackboxsw (approved):

holmanb:
> I think now that milestone 24.2 has been released. We can merge this pending your approval @holmanb

Thanks @blackboxsw taking one last look at this now, I think that I'd like to see two minor changes before merge:

1. Use the new util deprecation helper rather than ubuntu series to conditionally check for depreciation. 

2. Squash git history. I think that the modifications to verify_clean_boot probably deserve to be a separate (initial) commit, and the rest can probably be squashed into a single coauthored commit.

WDYT?

blackboxsw:
> > I think now that milestone 24.2 has been released. We can merge this pending your approval @holmanb
> 
> Thanks @blackboxsw taking one last look at this now, I think that I'd like to see two minor changes before merge:
> 
> 1. Use the new util deprecation helper rather than ubuntu series to conditionally check for depreciation.
> 2. Squash git history. I think that the modifications to verify_clean_boot probably deserve to be a separate (initial) commit, and the rest can probably be squashed into a single coauthored commit.
> 
> WDYT?
@holmanb +1 on using deprecation helper  in integration tests  and agreed on squash merge intent.

cloudinit/cmd/main.py
```diff
@@ -79,23 +79,30 @@ def print_exc(msg=""):
...
-            log = LOG.warning
-            extra_message = (
-                " Unsupported configuration: boot stage called "
-                "outside of systemd"
+            util.deprecate(
+                deprecated=(
+                    "Unsupported configuration: boot stage called "
+                    f"by PID [{ppid}] outside of systemd"
+                ),
+                deprecated_version="24.2",
```
blackboxsw, line 102: This should be 24.3 now

Issue #5158, opened 2024-04-08
tests/unittests/analyze/test_dump.py timestamp format errors on Alpine Linux


ncopa:
# Bug report

```
FAILED tests/unittests/analyze/test_dump.py::TestParseTimestamp::test_parse_unexpected_timestamp_format_with_date_command - AssertionError: assert 1723137300.0 == 1723130100.0 
FAILED tests/unittests/analyze/test_dump.py::TestParseCILogLine::test_parse_logline_returns_event_for_amazon_linux_2_line - AssertionError: assert {'description...oudinit', ...} == {'description...oudinit', ...}  
```

## Steps to reproduce the problem

Run pytest (8) on alpine linux with python 3.12.

## Environment details
- Cloud-init version: 24.1.3
- Operating System Distribution: Alpine Linux (edge / 3.20_alpha)
- Cloud provider, platform or installer type: n/a

## test logs

```
============================================= FAILURES =============================================                                            
___________ TestParseTimestamp.test_parse_unexpected_timestamp_format_with_date_command ____________                                            
                                                                                                                                                
self = <test_dump.TestParseTimestamp object at 0x7f3925a7d2e0>                                                                                  
                                                                                                                                                
    @skipIf(not which("date"), "'date' command not available.")                                                                                 
    @skipIf(                                                                                                                                    
        not is_Linux() and not which("gdate"),                                                                                                  
        "'GNU date' command not available.",                                                                                                    
    )                                                                                                                                           
    @pytest.mark.allow_subp_for("date", "gdate")                                                                                                
    def test_parse_unexpected_timestamp_format_with_date_command(self):                                                                         
        """Dump sends unexpected timestamp formats to date for processing."""                                                                   
        new_fmt = "%H:%M %m/%d %Y"                                                                                                              
        new_stamp = "17:15 08/08"                                                                                                               
        # convert stamp ourselves by adding the missing year value                                                                              
        year = datetime.now().year                                                                                                              
        dt = datetime.strptime(new_stamp + " " + str(year), new_fmt)                                                                            
                                                                                                                                                
>       assert float(dt.strftime("%s.%f")) == parse_timestamp(new_stamp)                                                                        
E       AssertionError: assert 1723137300.0 == 1723130100.0                                                                                     
E        +  where 1723137300.0 = float('1723137300.000000')                                                                                     
E        +    where '1723137300.000000' = <built-in method strftime of datetime.datetime object at 0x7f3922cce2b0>('%s.%f')                     
[...]

PR #5179, opened 2024-04-16
test(alpine): skip timezone-tests for non-glibc

catmsred:
Python 3.12 datetime.now() behaves differently on glibc and non- glibc regarding whether it incorporates a timezone or not.  These tests fail on non-glibc systems as a result.

## Proposed Commit Message

```
test(alpine): skip timezone-tests for non-glibc

Python 3.12 datetime.now() behaves differently on glibc and non-
glibc regarding whether it incorporates a timezone or not.  These
tests fail on non-glibc systems as a result

Fixes GH-5158
```

## Test Steps

See comments

## Checklist

- [x] I have updated or added any [unit tests](https://cloudinit.readthedocs.io/en/latest/development/testing.html) accordingly

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

PR #5214, opened 2024-04-25
fix: Ensure dump timestamps parsed as UTC

TheRealFalcon:
## Proposed Commit Message

```
fix: Ensure dump timestamps parsed as UTC

When parsing timestamps in `cloud-init analyze dump`, most log parsing
used UTC, but if shelling out to `date` was required, it used the local
timezone instead. This is wrong because cloud-init logs are in UTC.
This commit ensures that all timestamps are UTC-based.

Also update test code to explicitly convert generated times to UTC
before checking against what is being tested.

Fixes GH-5158
```

## Additional Context
https://github.com/canonical/cloud-init/issues/5158
and
https://github.com/canonical/cloud-init/pull/5179

## Test Steps
Besides CI, I ran through the reproducer in #5158 as well

## Merge type


holmanb (approved):
> but if shelling out to `date` was required, it used the local timezone instead

Nice catch.

Good work, this looks good to me!

Also tested on alpine and it works.

holmanb:
Fixes https://github.com/canonical/cloud-init/issues/5158

Issue #5192, opened 2024-04-19
Remove DataSourceNone warning in cc_final_message 


TheRealFalcon:
# Bug report
In the past, `DataSourceNone` was used purely as a fallback for when a primary datasource failed. However, `DataSourceNone` can be a datasource in an of itself and preferrable to `NoCloud` if all configuration is baked into the image. Because of this, it no longer makes sense to issue a warning in cc_final_message if None was expected. If there is defined userdata/metadata under the datasource definition in /etc/cloud or `DataSourceNone` is the only entry in the `datasource_list`, we should not warn that `DataSourceNone` was used.

Meulengracht:
We use data source [None] when testing Ubuntu Core, and I believe Certification team is too. The warning it produces to use this causes snapd to handle `cloud-init status` as an error, even when it actually produces the output `status: done`.

It seems like this is a new thing with the Noble build, on Jammy this seems a non issue

Issue #4739, opened 2024-01-04
[docs]: Overhaul user data formats page


holmanb:
Cloud-init supports multiple formats, which are described on the [user data formats](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#user-data-formats) page.

This page has a few improvements that should be made:

1. Each format deserves a very brief example. This will make our format page more readable and allows skimming and reader intuition to fill in details faster than reading a wall of text.
2. This page contains an unrelated [kernel command line](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#kernel-command-line) section. This section does not describe a format, and should be relocated or (more likely) deleted.
3. This page contains an unrelated [how to disable user data](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#disabling-user-data) section. This section does not describe a format, and should be relocated.
4. This page contains a [how to write your own part-handler](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#part-handler) section. Future plans for this section have already been described in [another issue](https://github.com/canonical/cloud-init/issues/4649). tl;dr - this belongs elsewhere
5. Jinja templated user data is not documented on this page (it is best described on the [instance metadata](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#using-instance-data) page)
6. Currently these formats are not categorized, but could be categorized to make them easier to skim / grok by the reader (see the next section).

### Recommended format categorization

These formats fall into two categories, which I will call **base formats** and **meta-formats** for now.

**Base formats** are formats that allow the user to configure the instance via cloud-init. **Meta formats** are formats that modify or aggregate **base formats**.

**Base formats** include the following: [cloud config data](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#cloud-config-data), [user data scripts](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#user-data-script), [cloud-boothooks](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#cloud-boothook)

**Meta formats** include the following: [include files](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#include-file), [gzip compressed content](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#gzip-compressed-content), [mime archives](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#mime-multi-part-archive), jinja templates (need to be documented)

TheRealFalcon:
[...]

PR #5215, opened 2024-04-25
Fix snapd growpart and final message warnings

blackboxsw:
## Proposed Commit Message(s)
*   fix(final_message): do not warn on datasourcenone when single ds
    
    Avoid warning level log "Used fallback datasource" when base
    configuration defines `datasource_list: [ None ]`.
    
    Using DataSourceNone in this case is desired behavior.
    
    Also drop unused is_disconnected property from DataSource
    which was only used by DataSourceNone in cc_final_message.
    
    Instead compare based on dsname property.
    
    Fixes: GH-5192

* fix(growpart): correct growpart log message to include value of mode

## Additional Context
Affects snapd UC24 which sets `datasource_list: [ None ]' in some configs

## Checklist

## Merge type

### DIFF
--- a/cloudinit/config/cc_final_message.py
+++ b/cloudinit/config/cc_final_message.py
@@ -110,5 +110,6 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     except Exception:
         util.logexc(LOG, "Failed to write boot finished file %s", boot_fin_fn)
 
-    if cloud.datasource.is_disconnected:
-        LOG.warning("Used fallback datasource")
+    if cloud.datasource.dsname == "None":
+        if cloud.datasource.sys_cfg.get("datasource_list") != ["None"]:
+            LOG.warning("Used fallback datasource")
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -600,7 +600,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     if util.is_false(mode):
         if mode != "off":
             util.deprecate(
-                deprecated="Growpart's 'mode' key with value '{mode}'",
+                deprecated=f"Growpart's 'mode' key with value '{mode}'",
                 deprecated_version="22.2",
                 extra_message="Use 'off' instead.",
             )
--- a/doc/rtd/explanation/configuration.rst
+++ b/doc/rtd/explanation/configuration.rst
@@ -11,12 +11,16 @@ higher priority source overwrites the lower priority source.
 Base configuration
 ==================
 
+The base configuration format uses `YAML version 1.1`_, but may be
+declared as jinja templates which cloud-init will render at runtime with
+:ref:`instance data <instancedata-Using>` variables.
+
 From lowest priority to highest, configuration sources are:
 
-- **Hardcoded config**: Config_ that lives within the source of ``cloud-init``
+- **Hardcoded config** Config_ that lives within the source of ``cloud-init``
   and cannot be changed.
 - **Configuration directory**: Anything defined in :file:`/etc/cloud/cloud.cfg`
-  and :file:`/etc/cloud/cloud.cfg.d`.
+  and :file:`/etc/cloud/cloud.cfg.d/*.cfg`.
 - **Runtime config**: Anything defined in :file:`/run/cloud-init/cloud.cfg`.
 - **Kernel command line**: On the kernel command line, anything found between
   ``cc:`` and ``end_cc`` will be interpreted as cloud-config user data.
@@ -25,6 +29,11 @@ These four sources make up the base configuration. The contents of this
 configuration are defined in the
 :ref:`base configuration reference page<base_config_reference>`.
 
+.. note::
+   Base configuration may contain
+   :ref:`cloud-config<explanation/format:Cloud config data>` which may be
+   overridden by vendor data and user data.
+
 Vendor and user data
 ====================
 
@@ -32,11 +41,6 @@ Added to the base configuration are :ref:`vendor data<vendordata>` and
 :ref:`user data<user_data_formats>` which are both provided by the datasource.
 
 These get fetched from the datasource and are defined at instance launch.
-
-.. note::
-   While much of what is defined in the base configuration can be overridden by
-   vendor data and user data, base configuration sources do not conform to
-   :ref:`#cloud-config<user_data_formats>`.
 
 Network configuration
 =====================
@@ -81,3 +85,4 @@ images.
 
 .. _Config: https://github.com/canonical/cloud-init/blob/b861ea8a5e1fd0eb33096f60f54eeff42d80d3bd/cloudinit/settings.py#L22
 .. _cloud.cfg template: https://github.com/canonical/cloud-init/blob/main/config/cloud.cfg.tmpl
+.. _YAML version 1.1: https://yaml.org/spec/1.1/current.html
--- a/doc/rtd/explanation/format.rst
+++ b/doc/rtd/explanation/format.rst
@@ -3,28 +3,26 @@
 User data formats
 *****************
 
-User data that will be acted upon by ``cloud-init`` must be in one of the
-following types.
+User data is opaque configuration data provided by a platform to an instance at
+launch configure the instance. User data can be one of the following types.
 
 .. _user_data_formats-cloud_config:
 
 Cloud config data
 =================
 
-Cloud-config is the simplest way to accomplish some things via user data.
-Using cloud-config syntax, the user can specify certain things in a
-human-friendly format.
+Cloud-config is the preferred user data format. The cloud config format is a
+declarative syntax which uses `YAML version 1.1`_ with keys which describe
+desired instance state. Cloud-config can be used to define how an instance
+should be configured in a human-friendly format.
 
-These things include:
+These things may include:
 
-- ``apt upgrade`` should be run on first boot
-- a different ``apt`` mirror should be used
-- additional ``apt`` sources should be added
-- certain SSH keys should be imported
+- performing package upgrades on first boot
+- configuration of different package mirrors or sources
+- initial user or group setup
+- importing certain SSH keys or host keys
 - *and many more...*
-
-.. note::
-   This file must be valid YAML syntax.
 
 See the :ref:`yaml_examples` section for a commented set of examples of
 supported cloud config formats.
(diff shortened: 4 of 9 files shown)


blackboxsw:
cc: @Meulengracht  for snap-related concerns.

Meulengracht:
Thank you to everyone involved here for getting this ball rolling and fixed so quickly!

cloudinit/config/cc_growpart.py
```diff
@@ -570,7 +570,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     if util.is_false(mode):
         if mode != "off":
             util.deprecate(
-                deprecated="Growpart's 'mode' key with value '{mode}'",
```
holmanb, line 573: I'm think maybe we should drop this deprecation altogether (here and in the jsonschema), since it is going to be a common tripping point.

```
python3 -c 'import yaml; print(yaml.safe_load("{mode: off}"))'
{'mode': False}
```

yet [the specification only considers `true` and `false`](https://yaml.org/spec/1.2.2/#10212-boolean) to be booleans

and our schema validation isn't any help:
```
growpart.mode:  Changed in version 22.3. Specifying a boolean ``false`` value for ``mode`` is deprecated. Use ``off`` instead.
```

blackboxsw: I feel like holding our more strict schema deprecation message is still desired here because we don't want to interpret YAML-interpreted boolean values long-term, and we'd prefer only discrete strings for this value type.  We deprecated this back in 22.2/22.3 (2 years ago), so this deprecation shouldn't come as a surprise. The issue is that consumers are now becoming more aware of those deprecation warnings that because of non-zero exit codes from cloud-init status. Given that python3-yaml happens to interpret the value `off` as boolean False shouldn't be a concern as long as we correctly document the desired value `'off'` that we know will be loaded as the right data type.

holmanb: > yet [the specification only considers true and false](https://yaml.org/spec/1.2.2/#10212-boolean) to be booleans

oh I see why PyYAML is doing this - it only supports the 1.1 spec, never updated to 1.2

blackboxsw: @holmanb interesting note on pyyaml 1.2.2 spec only stating true/false ... the pyyaml 1.1 spec though establishes other values are honored [1](https://yaml.org/spec/1.1/current.html#id864510) and [2](https://yaml.org/type/bool.html)  and I think it may be retained for backward-compat.

doc/rtd/explanation/format.rst
```diff
@@ -24,7 +24,7 @@ These things include:
 - *and many more...*
 
 .. note::
-   This file must be valid YAML syntax.
+   This file must be valid `YAML 1.1`_ syntax.
```
holmanb, line 27: I know you didn't write this line, but while we're touching this line can we please update it to something that makes a little more sense?

The file isn't a syntax. YAML isn't a syntax either. YAML is a standardized language for defining data objects, and it has a syntax for describing these data objects as a stream of characters. Cloud-config is a format which uses the YAML syntax described in specification version 1.1 and a set of keys defined [in the jsonschema](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schemas/schema-cloud-config-v1.json).

> This file ...

Additionally, this configuration is often times not even a file. Users may copy-paste a cloud-config into a web interface, for example, and ultimately on the instance they usually do not write a configuration file, so the mention of it being a file may be misleading to some.

Perhaps this paragraph:

> Cloud-config is the simplest way to accomplish some things via user data.
> Using cloud-config syntax, the user can specify certain things in a
> human-friendly format.

Could be updated to say something like:

> Cloud-config is the preferred user data format. The cloud config format is a 
> declarative syntax which uses YAML version 1.1 with [keys](https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schemas/schema-cloud-config-v1.json) that describe instance
> state. Cloud-config can be used to define how an instance should be configured in
> a human-friendly format.

And then remove the `This file must be ...` note?

doc/rtd/explanation/format.rst
```diff
@@ -34,7 +34,7 @@ using a MIME archive.
 
 .. note::
    Cloud config data can also render cloud instance metadata variables using
-   jinja templating. See :ref:`instance_metadata` for more information.
+   ref:`jinja templates <instancedata-Using>`.
```
holmanb, line 37: I still think that these notes under each section [aren't the best way to present meta-formats for user-data](https://github.com/canonical/cloud-init/issues/4739 ). However, until that issue gets implemented incremental improvements are welcome, +1.

doc/rtd/explanation/instancedata.rst
```diff
@@ -71,6 +71,7 @@ Using ``instance-data``
 * Command line interface via :command:`cloud-init query` or
   :command:`cloud-init devel render`.
 
+
```
holmanb, line 74: I assume this was unintentional?

holmanb (commented):
Hey @blackboxsw, thanks for this PR. I just added some wording nits to hopefully make our docs a bit clearer.

doc/rtd/explanation/configuration.rst
```diff
@@ -79,3 +83,4 @@ images.
 
 .. _Config: https://github.com/canonical/cloud-init/blob/b861ea8a5e1fd0eb33096f60f54eeff42d80d3bd/cloudinit/settings.py#L22
 .. _cloud.cfg template: https://github.com/canonical/cloud-init/blob/main/config/cloud.cfg.tmpl
+.. _YAML 1.1: https://yaml.org/spec/1.1/current.html
```
s-makin, line 86:
```suggestion
.. _YAML version 1.1: https://yaml.org/spec/1.1/current.html
```

s-makin: The mismatch is (I think) what's throwing the spellcheck error. Even though this isn't an error in the spelling.

doc/rtd/explanation/format.rst
```diff
@@ -221,5 +221,6 @@ appliances. Setting ``allow_userdata: false`` in the configuration will disable
 ``cloud-init`` from processing user data.
 
 .. _make-mime: https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/devel/make_mime.py
+.. _YAML 1.1: https://yaml.org/spec/1.1/current.html
```
s-makin, line 224:
```suggestion
.. _YAML version 1.1: https://yaml.org/spec/1.1/current.html
```

s-makin (commented):
I don't have much to add beyond a couple of nits

doc/rtd/explanation/configuration.rst
```diff
@@ -32,9 +37,9 @@ Added to the base configuration are :ref:`vendor data<vendordata>` and
 These get fetched from the datasource and are defined at instance launch.
 
 .. note::
-   While much of what is defined in the base configuration can be overridden by
-   vendor data and user data, base configuration sources do not conform to
-   :ref:`#cloud-config<user_data_formats>`.
+   While config keys defined in the base configuration can be overridden by
```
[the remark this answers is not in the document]
blackboxsw: @holmanb This "base configuration" is referring to any config defined in /etc/cloud/cloud.cfg.d, not specifically the deprecated `system_info` key in user-data. But, it does feel a bit confusing to dump that in the user-data/vendor-data section. I've moved it up into Base configuration instead

cloudinit/config/cc_final_message.py
```diff
@@ -110,5 +110,10 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
...
 
-    if cloud.datasource.is_disconnected:
-        LOG.warning("Used fallback datasource")
+    if cloud.datasource.dsname == "None":
+        if cloud.datasource.sys_cfg.get("datasource_list") == ["None"]:
+            # No warn on expected behavior when datasource_list: [ None ]
+            log_func = LOG.debug
+        else:
+            log_func = LOG.warning
+        log_func("Used fallback datasource")
```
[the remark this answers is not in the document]
holmanb: @blackboxsw I'm guessing this was forgotten?

doc/rtd/explanation/format.rst
```diff
@@ -3,29 +3,29 @@
 User data formats
 *****************
 
-User data that will be acted upon by ``cloud-init`` must be in one of the
-following types.
+User data is opaque configuration data provided by a platform to an instance at
+launch to perform supplemental configuration operations on that instance at
+first boot. User data that will be acted upon by ``cloud-init`` must be in one
+of the following types.
```
holmanb, line 9: This feels more verbose than it needs to be. I think that we could probably strike some words to be more succinct without loosing the important bits:

```suggestion
User data is configuration data provided by a platform to an instance at
launch to configure the instance. User data can be one of the following types.
```

doc/rtd/explanation/format.rst
```diff
@@ -3,29 +3,29 @@
...
+These things may include:
 
-- ``apt upgrade`` should be run on first boot
-- a different ``apt`` mirror should be used
-- additional ``apt`` sources should be added
-- certain SSH keys should be imported
+- performing package upgrades on first boot
+- configuration different package mirrors or sources
+- initial user or group setup
+- importing certain SSH keys or host keys to configure
```
holmanb, line 26:
```suggestion
- importing certain SSH keys or host keys
```

holmanb (commented):
> I've pulled the majority into the first paragraph of Cloud-config section as well as rewriting a couple of the apt-specific examples to more package format packaging format agnostic terms.

Great idea. This is looking much better! I left a couple more questions and suggestions inline.

holmanb (changes requested):
I have no more updates to make, I'm just setting this to "request changes" to make it more obvious that this is waiting on @blackboxsw right now.

holmanb:
> I have no more updates to make, I'm just setting this to "request changes" to make it more obvious that this is waiting on @blackboxsw right now.

@blackboxsw since this will make DataSourceNone users no longer have a warning message we might want to squeeze this one into 24.2. I'm adding it to the milestone so that it doesn't slip by without a discussion (we can remove it if we decide it's not a priority).

doc/rtd/explanation/format.rst
blackboxsw: Who'd have though less words, more clear meaning :). Thanks for these suggestions.

blackboxsw (commented):
Thanks for the more specific text suggestions @holmanb . I've added them all.

holmanb (commented):
Thanks @blackboxsw! Let's land it.

holmanb (approved):

PR #5217, opened 2024-04-25
fix(pdb): Enable running cloud-init under pdb

holmanb:
## Proposed Commit Message

```
fix(pdb): Enable running cloud-init under pdb

Cloud-init closes stdin on startup, which breaks interactively running cloud-init under pdb.

Leave stdin open if it is connected to a tty, which fixes pdb use.


File "/usr/lib/python3.12/bdb.py", line 90, in trace_dispatch
  return self.dispatch_line(frame)
         ^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3.12/bdb.py", line 115, in dispatch_line
    if self.quitting: raise BdbQuit
                      ^^^^^^^^^^^^^
bdb.BdbQuit
------------------------------------------------------------
The program exited via sys.exit(). Exit status: 1
```

## Merge type

### DIFF
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -328,9 +328,12 @@ def main_init(name, args):
     outfmt = None
     errfmt = None
     try:
-        early_logs.append((logging.DEBUG, "Closing stdin."))
-        util.close_stdin()
-        (outfmt, errfmt) = util.fixup_output(init.cfg, name)
+        if not os.isatty(sys.stdin.fileno()):
+            early_logs.append((logging.DEBUG, "Closing stdin."))
+            util.close_stdin()
+        else:
+            LOG.warning("Not closing stdin, stdin is a tty.")
+        outfmt, errfmt = util.fixup_output(init.cfg, name)
     except Exception:
         msg = "Failed to setup output redirection!"
         util.logexc(LOG, msg)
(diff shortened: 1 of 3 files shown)


CalvoM:
Thanks for this. I ran into this issue when trying to debug some issues.

CalvoM (approved):
LGTM.

cloudinit/cmd/main.py
```diff
@@ -328,9 +328,12 @@ def main_init(name, args):
...
     errfmt = None
     try:
-        early_logs.append((logging.DEBUG, "Closing stdin."))
-        util.close_stdin()
-        (outfmt, errfmt) = util.fixup_output(init.cfg, name)
+        if not os.isatty(sys.stdin.fileno()):
+            early_logs.append((logging.DEBUG, "Closing stdin."))
+            util.close_stdin()
+        else:
+            LOG.warning("Not closing stdin, stdin is a tty.")
```
TheRealFalcon, line 335: Is there a reason for this to be a warning? I think it makes more sense as a debug.

cloudinit/cmd/main.py
```diff
@@ -328,9 +328,12 @@ def main_init(name, args):
     outfmt = None
     errfmt = None
     try:
-        early_logs.append((logging.DEBUG, "Closing stdin."))
-        util.close_stdin()
-        (outfmt, errfmt) = util.fixup_output(init.cfg, name)
+        if not os.isatty(sys.stdin.fileno()):
```
TheRealFalcon, line 331: Nit: These 5 lines are duplicated a few other times. Any reason we can't move them all into `util.close_stdin()` so each call site is just the single call to the function?

TheRealFalcon (changes requested):
I think we should change the log level. The rest of the comments are nits.

Nit: Think there's any need to keep the `_CLOUD_INIT_SAVE_STDIN` env var? If its only use was for debugging, I don't see a reason to keep it around.

cloudinit/cmd/main.py
holmanb: I also bumped this function over to `cloudinit/cmd/main.py` since it isn't used anywhere else

holmanb:
@TheRealFalcon @CalvoM Thanks for the reviews! I think I've addressed all of the comments, requesting re-review.

TheRealFalcon (approved):
LGTM!

PR #5218, opened 2024-04-25
azure: add request identifier to IMDS requests

KsenijaS:
Set x-ms-client-request-id uuid per request and log it. 
This way we can easily find the request in the IMDS logs based on the uuid number.


tests/unittests/sources/azure/test_imds.py
```diff
@@ -168,7 +167,7 @@ def test_basic(
             mock.call(
                 self.default_url,
                 timeout=self.timeout,
-                headers=self.headers,
+                headers_cb=mock.ANY,
```
cjp256, line 170: we don't want to lose coverage of validating the header's expected contents, is there a good way to restore that even if separately?

tests/unittests/sources/azure/test_imds.py
```diff
@@ -129,6 +129,18 @@ def regex_for_http_error(error):
...
+class TestHeaders:
+    default_url = (
+        "http://169.254.169.254/metadata/instance?"
+        "api-version=2021-08-01&extended=true"
+    )
+
+    def test_headers_cb(self):
+        headers = imds.headers_cb(self.default_url)
+        assert list(headers.keys()) == ["Metadata", "x-ms-client-request-id"]
+        assert headers.get("Metadata") == "true"
```
cjp256, line 141: just add check to validate uuid

TheRealFalcon (approved):
LGTM!

PR #5219, opened 2024-04-26
chore: Remove broken maas code

holmanb:
## Proposed Commit Message
```
chore: Fix broken maas code
```

## Additional Context
Better late than never, eh?

### DIFF
--- a/cloudinit/sources/DataSourceMAAS.py
+++ b/cloudinit/sources/DataSourceMAAS.py
@@ -43,6 +43,10 @@ class DataSourceMAAS(sources.DataSource):
     id_hash = None
     _oauth_helper = None
 
+    # Setup read_url parameters per get_url_params.
+    url_max_wait = 120
+    url_timeout = 50
+
     def __init__(self, sys_cfg, distro, paths):
         sources.DataSource.__init__(self, sys_cfg, distro, paths)
         self.base_url = None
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
Think we should try to still respect a timeout if it's given?

blackboxsw (changes requested):
We should probably just use the general get_url_params right?
```diff
diff --git a/cloudinit/sources/DataSourceMAAS.py b/cloudinit/sources/DataSourceMAAS.py
index a8c043082..a74d91a92 100644
--- a/cloudinit/sources/DataSourceMAAS.py
+++ b/cloudinit/sources/DataSourceMAAS.py
@@ -43,6 +43,10 @@ class DataSourceMAAS(sources.DataSource):
     id_hash = None
     _oauth_helper = None
 
+    # Setup read_url parameters per get_url_params.
+    url_max_wait = 120
+    url_timeout = 50
+
     def __init__(self, sys_cfg, distro, paths):
         sources.DataSource.__init__(self, sys_cfg, distro, paths)
         self.base_url = None
@@ -116,25 +120,18 @@ class DataSourceMAAS(sources.DataSource):
         return "seed-dir (%s)" % self.base_url
 
     def wait_for_metadata_service(self, url):
-        mcfg = self.ds_cfg
-        max_wait = 120
-        try:
-            max_wait = int(mcfg.get("max_wait", max_wait))
-        except Exception:
-            util.logexc(LOG, "Failed to get max wait. using %s", max_wait)
-
-        if max_wait == 0:
+        url_params = self.get_url_params()
+        if url_params.max_wait_seconds == 0:
             return False
 
-        timeout = 50
-
         starttime = time.time()
-        if url.endswith("/"):
-            url = url[:-1]
+        url = url.rstrip("/")
         check_url = "%s/%s/meta-data/instance-id" % (url, MD_VERSION)
         urls = [check_url]
         url, _response = self.oauth_helper.wait_for_url(
-            urls=urls, max_wait=max_wait, timeout=timeout
+            urls=urls,
+            max_wait=url_params.max_wait_seconds,
+            timeout=url_params.timeout_seconds,
         )
 
         if url:
```

holmanb:
I didn't originally propose supporting timeout because it would raise a lot of questions. i.e. does maas (or some user) set this to some value?

If we're all comfortable with just fixing it rather than deleting it, I can do that.

@TheRealFalcon @blackboxsw thoughts?

blackboxsw:
> I didn't originally propose supporting timeout because it would raise a lot of questions. i.e. does maas (or some user) set this to some value?
> 
> If we're all comfortable with just fixing it rather than deleting it, I can do that.
> 
> @TheRealFalcon @blackboxsw thoughts?

@holmanb +1 to fixing this and making it configurable. Given that it's been ignored for ever, this is a feature/bugfix that improves configurability. I'd agree with that direction and maybe even adding a minimal bit to the doc/rtd/reference/datasources/maas.rst to describe the command DS config options supported.

blackboxsw (approved):
LGTM!

Issue #5143, opened 2024-04-04
Option package_upgrade runs `snap refresh` even if snapd is not installed in the image


loxK:
# Bug report
I use images.linuxcontainers.org cloud images. If using the option `package_upgrade`, cloud_init fails because snap isn't installed in these images.

## Steps to reproduce the problem
**test LXD/Incus profile :**

```yaml
config:
  user.vendor-data: |
    #cloud-config
    package_upgrade: true
description: ""
devices: {}
name: test
```

**Script that creates the container :** 
```bash
#!/bin/bash

CONTAINER_NAME="myname"

echo 'Creating container'
incus init images:ubuntu/22.04/cloud $CONTAINER_NAME -p default -p test
echo 'Container created'

echo 'Starting it'
incus start $CONTAINER_NAME

echo 'Waiting for cloud-init to finish up ...'

incus exec $CONTAINER_NAME -- cloud-init status --wait  > /dev/null 2>&1

[ $? -ne 0 ] && echo Cloud-init failed && incus exec $CONTAINER_NAME -- cloud-init status --long && exit 1
echo -n 'Cloud-init finished at '
echo -n `date -R`
```

**The error :** 

```bash
~ ./test.sh
Creating container
Création de myname
Container created
Starting it
Waiting for cloud-init to finish up ...
Cloud-init failed
status: error
extended_status: error
boot_status_code: enabled-by-generator
last_update: Thu, 04 Apr 2024 00:44:15 +0000
detail:
DataSourceNoCloud [seed=/var/lib/cloud/seed/nocloud-net][dsmode=net]
errors:
	- ('package_update_upgrade_install', ProcessExecutionError("Unexpected error while running command.\nCommand: ['snap', 'refresh']\nExit code: -\nReason: [Errno 2] No such file or directory: b'snap'\nStdout: -\nStderr: -"))
recoverable_errors:
WARNING:
	- Package upgrade failed
	- 1 failed with exceptions, re-raising the last one
	- Running module package_update_upgrade_install (<module 'cloudinit.config.cc_package_update_upgrade_install' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_package_update_upgrade_install.py'>) failed
Cloud-init finished at Thu, 04 Apr 2024 11:44:16 +1100
```

**Snap  isn't installed :** 
```bash
root@myname:~ dpkg -l | grep snap
root@myname:~ which snap
root@myname:~ snap
-bash: snap: command not found
```

## Environment details
[...]

PR #5224, opened 2024-04-29
bug(package_update): avoid snap refresh in images without snap command

blackboxsw:
## Proposed Commit Message
When package_update or package_upgrade: true are provided in cloud-config do not call snap refresh on systems that do not have the snap command.

This was intended as fixed in cbe5f3a119 but the fix only avoided snap refresh during package install, not the generic package update operation.

Fixes: GH-5143
## Additional Context

Testable with either of the following procedures:
```
$ cat > pkg_update.yaml <<EOF
#cloud-config
package_update: true
EOF


lxc launch ubuntu-daily/jammy no-snap -c cloud-init.user-data="$(pkg_update.yaml)"
sleep 3
lxc exec no-snap -- apt remove --purge snapd
lxc exec no-snap -- cloud-init clean --logs --reboot 
sleep 3
lxc exec no-snap -- cloud-init status --wait --format=yaml
... ## expect errors


# Or use incus ubuntu cloud images which don't contain snapd package by default.
incus launch images:ubuntu/22.04/cloud  no-snap -c cloud-init.user-data="$(pkg_update.yaml)"
sleep 3
incus exec no-snap -- cloud-init status --wait --format=yaml
.... # expect errors 
```

## Checklist

## Merge type

### DIFF
--- a/tests/unittests/distros/test_init.py
+++ b/tests/unittests/distros/test_init.py
@@ -33,6 +33,8 @@ for separator in [":", ".", "/", "#", "?", "@", "[", "]"]:
     # unable to parse URLs ("[", "]").
     INVALID_URL_CHARS.remove(separator)
 
+M_PATH = "cloudinit.distros.package_management."
+
 
 class TestGetPackageMirrorInfo:
     """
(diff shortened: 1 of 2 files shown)


tests/unittests/distros/test_init.py
```diff
@@ -253,31 +255,130 @@ def test_valid_substitution(
...
+            ),
+        ],
+    )
+    def test_log_errors_with_updating_package_source(
+        self, apt_error, snap_error, expected_logs, mocker, caplog
+    ):
+        """Log error raised from any package_manager.update_package_sources."""
+        mocker.patch(M_PATH + "apt.Apt.available", return_value=True)
+        mocker.patch(M_PATH + "snap.Snap.available", return_value=True)
+        if apt_error:
```
TheRealFalcon, line 288: Nit: Is there a functional difference between `return_value=None` and `side_effect=None`? I think we can remove the if/else's in this function and just include what under the `if` conditions.

TheRealFalcon (approved):
Need to fix the lint but otherwise LGTM! A few nits inline but nothing blocking.

tests/unittests/distros/test_init.py
```diff
@@ -253,31 +255,130 @@ def test_valid_substitution(
...
+        self, apt_available, snap_available, expected_logs, mocker, caplog
+    ):
+        """Avoid update_package_sources on unavailable package managers"""
+
+        mocker.patch(M_PATH + "apt.Apt.available", return_value=apt_available)
+        mocker.patch(
+            M_PATH + "snap.Snap.available",
+            return_value=snap_available,
+        )
+        with mock.patch(
```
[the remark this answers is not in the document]
blackboxsw: right, none, I had started writing the test w/ mock.patch and switched to mocker after I already had the context managers in place. Using mocker throughout.

tests/unittests/distros/test_init.py
blackboxsw: fixed. good thought

PR #5229, opened 2024-04-29
Ubuntu/noble 24.1.x

blackboxsw:
Cherry pick a6f7577d to fix  SRU regression https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2064132

### procedure
```
git fetch upstream
git checkout upstream/ubuntu/noble-24.1.x -B ubuntu/noble-24.1.x
cherry-pick a6f7577d
# Fixup debian/changelog text s/bug/fix/
sed -i '`s/UNRELEASED/noble/'` debian/changelog
git commit -am 'releasing cloud-init version 24.1.3-0ubuntu4'
```

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-cloud-init (24.1.3-0ubuntu3.1) UNRELEASED; urgency=medium
+cloud-init (24.1.3-0ubuntu3.1) noble; urgency=medium
 
   * cherry-pick a6f7577d: bug(package_update): avoid snap refresh in
     images without (LP: #2064132)
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+cloud-init (24.1.3-0ubuntu3.1) UNRELEASED; urgency=medium
+
+  * cherry-pick a6f7577d: bug(package_update): avoid snap refresh in
+    images without (LP: #2064132)
+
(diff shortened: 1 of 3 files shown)


debian/changelog
```diff
@@ -1,3 +1,10 @@
+cloud-init (24.1.3-0ubuntu4) noble; urgency=medium
+
+  * cherry-pick a6f7577d: fix(package_update): avoid snap refresh in
```
TheRealFalcon, line 3: `bug` is what committed, so we should keep this as the same title as in the original commit.

debian/changelog
```diff
@@ -1,3 +1,10 @@
+cloud-init (24.1.3-0ubuntu4) noble; urgency=medium
```
TheRealFalcon, line 1: Since this is our first SRU to noble, I think this need to be changed to `24.1.3-0ubuntu3~24.04.1`.

TheRealFalcon (commented):
Two inline comments changelog related. I'm gonna double check in the ubuntu handbook for version numbering.

debian/changelog
blackboxsw: Since we're in the stange space where we are in 0-day SRU territory and devel release hasn't opened yet. Let's go for ubuntu3.1.  If we get suggestions otherwise on upload we'll adjust for the ` ~24.04.1` suffix.

TheRealFalcon (approved):
LGTM!

Issue #5105, opened 2024-03-27
cloud init is not able to identify data source 


ShivChittora:
Using the AWS console, I'm launching an instance using a custom AMI. A few XML files need to be updated as part of the system initialization. Although I entered the user data, it doesn't appear to be working.

Upon checking ds-idenify.log and cloud-init-generator.log, I can see that cloud-init is unable to obtain the data source.

**cat ds-identify.log**
```
[up 1.17s] ds-identify
policy loaded: mode=search report=false found=all maybe=all notfound=disabled
no datasource_list found, using default: MAAS ConfigDrive NoCloud AltCloud Azure Bigstep CloudSigma CloudStack DigitalOcean AliYun Ec2 GCE OpenNebula OpenStack OVF SmartOS Scaleway Hetzner IBMCloud Oracle Exoscale RbxCloud
DMI_PRODUCT_NAME=t3.small
DMI_SYS_VENDOR=Amazon EC2
DMI_PRODUCT_SERIAL=ec2972df-81d1-c72e-f54d-0da7741efa46
DMI_PRODUCT_UUID=DF7229EC-D181-2EC7-F54D-0DA7741EFA46
PID_1_PRODUCT_NAME=unavailable
DMI_CHASSIS_ASSET_TAG=Amazon EC2
FS_LABELS=any:scandir,any:scandir,any:scandir,any:scandir,var,config,deferred,logs,wk,boot,opt,root,notused,rescue
ISO9660_DEVS=
KERNEL_CMDLINE=BOOT_IMAGE=/vmlinuz-x.x.xx-x.xx..0.x86_64 ro root=LABEL=root quiet nmi_watchdog=2 fsck.repair=yes fips=1 elevator=noop rootdelay=300 console=ttyS0,115200n8 aws nvme_core.io_timeout=255 rootfstype=ext4
VIRT=kvm
UNAME_KERNEL_NAME=Linux
UNAME_KERNEL_RELEASE=x.x.xx-x.xx..0.x86_64
UNAME_KERNEL_VERSION=#1 SMP Thu Mar 21 13:43:44 UTC 2024
UNAME_MACHINE=x86_64
UNAME_NODENAME=mlos-tiny
UNAME_OPERATING_SYSTEM=GNU/Linux
DSNAME=
[...]

PR #5236, opened 2024-04-29
fix(ec2): Fix broken uuid match with other-endianness

holmanb:
```
fix(ec2): Fix broken uuid match with other-endianness

EC2 documents that the system-uuid may be reported in different endianness[1].

A user has reported a case where cloud-init is broken due to inability to
detect the system platform. Fix it.

Behavior change:

Cloud-init was previously making the assumption that uuid and serial would match
on ec2. This assumption was:

1) not documented as a valid way to identify ec2[1]

2) proven invalid on ec2 by the DMI_PRODUCT_SERIAL and DMI_PRODUCT_UUID reported in #5105

3) used in the logic which warns about not running on the "real" ec2

Preserving this warning logic exactly as it was presents several challenges:

a) Risk of regression outside of our control: Since this logic relied upon undocumented
   behavior, AWS could change this at any point, which would break all cloud-init
   instances.

b) Risk of incorrect implementation: What format is the uuid and product serial actually
   in? We don't know. It's easy and safe to just swap the byteorder of the first segment
   of the uuid because this is documented, but matching the whole uuid is problematic
   because UUID formats may be presented as mixed encoding (partially little endian and
   partially big endian). To implement this behavior while fixing this bug we would have
   to make even more assumptions than before. I propose we stop assuming and if a cloud
   happens to implement the same as EC2 (minus the serial/product match), then we just
   don't emit that warning. It's simpler, it's safer, and I really don't think that it is
   a huge change. This is a "change in behavior", but the change is that the code more
   correctly identifies EC2 and would no longer emit a warning on valid ec2 instances, so
   I don't think that this would require omitting this change from SRU.

c) Implementing whatever assumptions we make in b) would require implementing a
   byteswapping algorithm in POSIX shell, which is possible but best to avoid this if
   possible.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html

Fixes GH-5105
```

## Additional Context
Cloud-init cannot identify that it is running on ec2 when the uuid is reported in swapped endianness, which is a [valid documented way for ec2 to report](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html).

https://github.com/canonical/cloud-init/issues/5105

## Test Steps
See new unit tests.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -12,6 +12,8 @@ import copy
 import logging
 import os
 import time
+import uuid
+from contextlib import suppress
 from typing import Dict, List
 
 from cloudinit import dmi, net, sources
@@ -794,11 +796,17 @@ def identify_aliyun(data):
 
 def identify_aws(data):
     # data is a dictionary returned by _collect_platform_data.
-    if data["uuid"].startswith("ec2") and (
-        data["uuid_source"] == "hypervisor" or data["uuid"] == data["serial"]
-    ):
-        return CloudNames.AWS
-
+    uuid_str = data["uuid"]
+    with suppress(ValueError):
+        if uuid_str.startswith("ec2"):
+            # example same-endian uuid:
+            # EC2E1916-9099-7CAF-FD21-012345ABCDEF
+            return CloudNames.AWS
+        elif uuid.UUID(uuid_str).bytes_le.hex().startswith("ec2"):
+            # check for other endianness
+            # example other-endian uuid:
+            # 45E12AEC-DCD1-B213-94ED-012345ABCDEF
+            return CloudNames.AWS
     return None
 
 
@@ -853,7 +861,6 @@ def _collect_platform_data():
 
     Keys in the dictionary are as follows:
        uuid: system-uuid from dmi or /sys/hypervisor
-       uuid_source: 'hypervisor' (/sys/hypervisor/uuid) or 'dmi'
        serial: dmi 'system-serial-number' (/sys/.../product_serial)
        asset_tag: 'dmidecode -s chassis-asset-tag'
        vendor: dmi 'system-manufacturer' (/sys/.../sys_vendor)
@@ -862,37 +869,23 @@ def _collect_platform_data():
     On Ec2 instances experimentation is that product_serial is upper case,
     and product_uuid is lower case.  This returns lower case values for both.
     """
-    data = {}
-    try:
+    uuid = None
+    with suppress(OSError):
         uuid = util.load_text_file("/sys/hypervisor/uuid").strip()
-        data["uuid_source"] = "hypervisor"
-    except Exception:
-        uuid = dmi.read_dmi_data("system-uuid")
-        data["uuid_source"] = "dmi"
 
-    if uuid is None:
-        uuid = ""
-    data["uuid"] = uuid.lower()
+    uuid = uuid or dmi.read_dmi_data("system-uuid") or ""
+    serial = dmi.read_dmi_data("system-serial-number") or ""
+    asset_tag = dmi.read_dmi_data("chassis-asset-tag") or ""
+    vendor = dmi.read_dmi_data("system-manufacturer") or ""
+    product_name = dmi.read_dmi_data("system-product-name") or ""
 
-    serial = dmi.read_dmi_data("system-serial-number")
-    if serial is None:
-        serial = ""
-
-    data["serial"] = serial.lower()
-
-    asset_tag = dmi.read_dmi_data("chassis-asset-tag")
-    if asset_tag is None:
-        asset_tag = ""
-
-    data["asset_tag"] = asset_tag.lower()
-
-    vendor = dmi.read_dmi_data("system-manufacturer")
-    data["vendor"] = (vendor if vendor else "").lower()
-
-    product_name = dmi.read_dmi_data("system-product-name")
-    data["product_name"] = (product_name if product_name else "").lower()
-
-    return data
+    return {
+        "uuid": uuid.lower(),
+        "serial": serial.lower(),
+        "asset_tag": asset_tag.lower(),
+        "vendor": vendor.lower(),
+        "product_name": product_name.lower(),
+    }
 
 
 def _build_nic_order(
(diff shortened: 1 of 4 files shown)


tests/unittests/sources/test_ec2.py
```diff
@@ -855,7 +854,7 @@ def test_unknown_platform_with_strict_true(self, mocker, tmpdir):
         """Unknown platform data with strict_id true should return False."""
         uuid = "ab439480-72bf-11d3-91fc-b8aded755F9a"
         ds = self._setup_ds(
-            platform_data={"uuid": uuid, "uuid_source": "dmi", "serial": ""},
+            platform_data={"uuid": uuid, "serial": ""},
```
aciba90, line 857: Why do we remove the `uuid_source` key?

aciba90 (commented):
LGTM, thanks for finding and fixing this issue. I left some non-0blocking questions, comments, suggestions.

cloudinit/sources/DataSourceEc2.py
```diff
@@ -862,37 +869,23 @@ def _collect_platform_data():
     On Ec2 instances experimentation is that product_serial is upper case,
     and product_uuid is lower case.  This returns lower case values for both.
     """
-    data = {}
-    try:
+    uuid = None
+    with suppress(OSError):
```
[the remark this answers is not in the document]
holmanb: > Did we check that no other exceptions than OSError cannot happen while executing util.load_text_file? If unsure, we could do the following to maintain old behavior:

Good question. I think that this is pretty safe. I suppose this could possibly throw `UnicodeDecodeError` if a malformed UUID exists which cannot be converted to utf-8, so I'll add that exception handler, but I think that all other cases should be handled.

Let's take a look at the relevant functions:

```python3
def pipe_in_out(in_fh, out_fh, chunk_size=1024, chunk_cb=None):
    bytes_piped = 0
    while True:
        data = in_fh.read(chunk_size)
        if len(data) == 0:
            break
        else:
            out_fh.write(data)
            bytes_piped += len(data)
            if chunk_cb:
                chunk_cb(bytes_piped)
    out_fh.flush()
    return bytes_piped

def decode_binary(blob: Union[str, bytes], encoding="utf-8") -> str:
    # Converts a binary type into a text type using given encoding.
    return blob if isinstance(blob, str) else blob.decode(encoding=encoding)

def load_binary_file(
    fname: Union[str, os.PathLike],
    *,
    read_cb: Optional[Callable[[int], None]] = None,
    quiet: bool = False,
) -> bytes:
    LOG.debug("Reading from %s (quiet=%s)", fname, quiet)
    with io.BytesIO() as ofh:
        try:
            with open(fname, "rb") as ifh:
                pipe_in_out(ifh, ofh, chunk_cb=read_cb)
        except FileNotFoundError:
            if not quiet:
                raise
        contents = ofh.getvalue()
    LOG.debug("Read %s bytes from %s", len(contents), fname)
    return contents


def load_text_file(
    fname: Union[str, os.PathLike],
    *,      
    read_cb: Optional[Callable[[int], None]] = None,
    quiet: bool = False,
) -> str:   
    return decode_binary(load_binary_file(fname, read_cb=read_cb, quiet=quiet))
```

`io.BytesIO()` [is documented as raising only exceptions which inherit from OSError](https://docs.python.org/3/library/io.html#high-level-module-interface) (with the exception of `ValueError` when used incorrectly, which we don't)

`OSError` [is really broad](https://docs.python.org/3/library/exceptions.html#exception-hierarchy) and would get thrown by open/read/close operations when file doesn't exist or reads are blocking or permissions error or anything like that. 

Since `load_binary_file()` returns `bytes`, `decode_binary()` should always call `.decode()` on the input (the only source of the aforementioned `UnicodeDecodeError`).

I don't see any other sources of exceptions except for programming errors which I don't think we want to silence (see [this PR for rational](https://github.com/canonical/cloud-init/pull/5202)).

tests/unittests/sources/test_ec2.py
holmanb: Good question. I should document this better in the commit message.

Cloud-init was previously making the assumption that uuid and serial would match on ec2. This assumption was:

1) [not documented as a valid way to identify ec2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html)

2) proven invalid on ec2 by the `DMI_PRODUCT_SERIAL` and `DMI_PRODUCT_UUID` reported in https://github.com/canonical/cloud-init/issues/5105

3) used in the logic which warns about not running on the "real" ec2

Preserving this warning logic exactly as it was presents several challenges:

a) Risk of regression outside of our control: Since this logic relied upon undocumented behavior, AWS could change this at any point, which would break all cloud-init instances.

b) Risk of incorrect implementation: What format is the uuid and product serial actually in? We don't know. It's easy and safe to just swap the byteorder of the first segment of the uuid because this is documented, but matching the whole uuid is problematic because UUID formats may be presented as mixed encoding (partially little endian and partially big endian). To implement this behavior while fixing this bug we would have to make even more assumptions than before. I propose we stop assuming and if a cloud happens to implement the same as EC2 (minus the serial/product match), then we just don't emit that warning. It's simpler, it's safer, and I really don't think that it is a huge change. This is a "change in behavior", but the change is that the code more correctly identifies EC2 and would no longer emit a warning on valid ec2 instances, so I don't think that this would require omitting this change from SRU.

c) Implementing whatever assumptions we make in b) would require a manual byteswapping algorithm in posix shell, which is more shell poo that I'd rather not write & maintain if I can convince folks that this preserving this strict ec2 warning _exactly_ as it was isn't worth it.

holmanb: > Why do we remove the uuid_source key?

To answer this question: because after the changes which resulted from the above comment, that key was no longer used anywhere.

holmanb:
> LGTM, thanks for finding and fixing this issue. I left some non-0blocking questions, comments, suggestions.

Thanks for the review @aciba90. I've updated the commit message to better justify and explain the change. I also pushed another commit with the suggested changes.

cloudinit/sources/DataSourceEc2.py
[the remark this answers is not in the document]
aciba90: Thanks for checking and adding the exception!

aciba90 (approved):
LGTM, thanks for the detailed explanations and fixing this!

PR #5238, opened 2024-04-30
Suppressed errors in pyright

### LINKED ISSUE #4642: [wishlist] Resolve pyright errors

aadityasinha-dotcom:
## Proposed Commit Message
Resolves #4642

## Checklist

## Merge type

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,3 +59,15 @@ ignore = [
 ]
 [tool.ruff.per-file-ignores]
 "cloudinit/cmd/main.py" = ["E402"]
+
+[tool.pyright]
+# 1. Suppress "Method {m} overrides class {c} in an incompatible manner"
+reportIncompatibleMethodOverride = false
+# 2. Suppress "{Method} is not a known member of None"
+reportOptionalMemberAccess = false
+# 3. Exclude specified files & folders from the check (overwrites include)
+exclude = ["cloudinit/apport.py"]
+# 4. Include only specified files & folders to the check
+include = []
+# 5. Suppress missing module errors
+reportOptionalSubscript = false


pyproject.toml
```diff
@@ -59,3 +59,15 @@ ignore = [
 ]
 [tool.ruff.per-file-ignores]
 "cloudinit/cmd/main.py" = ["E402"]
+
+[tool.pyright]
```
aadityasinha-dotcom, line 63: Need some review on that :relaxed:

TheRealFalcon:
@aadityasinha-dotcom , thanks for the contribution. Can you explain your changes on this PR? I don't understand why the configuration you're submitted would be a considered an improvement over the defaults.

aadityasinha-dotcom:
Hi sorry I should mention the changes that were reflected in the description.
Before the configuration `pyright cloudinit/` was giving `556 errors` now it is giving `368 errors`. I tried to supress the warnings in the pyright.

TheRealFalcon:
@aadityasinha-dotcom , Thanks, but the intention of #4642 was to fix the actual errors over time, rather than just mask them. What is the reasoning for suppressing the issues that this PR is suppressing? Do you believe them to be invalid in some way?

aadityasinha-dotcom:
@TheRealFalcon yes I think you are correct. We should fix the actual errors. There was also a point "`optionally generate a configuration for Pyright which silences errors that are not helpful`". I will look into the errors again and will try to find the way to reslove them.

TheRealFalcon:
@aadityasinha-dotcom , thanks. We aren't expecting all errors to be resolved in one go. That would be an enormous effort. The spirit of the issue was more to highlight that those errors exist, and we're always happy to reduce the number where we can. We're also ok silencing errors if the errors don't make sense or are too strict. We'd just need some justification as to why they don't make sense.

Should we close this PR as-is, or are there still changes here you want to make?

aadityasinha-dotcom:
@TheRealFalcon yeah sure I will close this PR and will try to come up with the solution :+1:

PR #5242, opened 2024-04-30
ci: Print isort diff

holmanb:
## Proposed Commit Message
```
ci: Print isort diff

When it passes locally, it would be good to know why ci fails.
```

## Merge type


blackboxsw (approved):
ship it.

PR #5244, opened 2024-05-01
Multiple fixes related to system time

### LINKED ISSUE #2423: wait_for_url doesn't account for system clock being changed
This bug was originally filed in Launchpad as [LP: #1273255](https://bugs.launchpad.net/cloud-init/+bug/1273255)
<details>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2014-01-27T14:55:58.869097+00:00
date_fix_committed = None
date_fix_released = None
id = 1273255
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1273255
milestone = None
owner = smoser
owner_name = Scott Moser
private = False
status = confirmed
submitter = smoser
submitter_name = Scott Moser
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Scott Moser(smoser)** wrote on 2014-01-27T14:55:58.869097+00:00_

wait_for_url takes a 'max_wait' input, and then does:
 start = time.time()
 ...
  now = time.time()

The problem is that when this runs early in boot, ntp (or anything else really) might have set the clock backwards.

I'm looking at a console log that shows:
2014-01-27 14:46:24,743 - url_helper.py[WARNING]: Calling 'http://169.254.169.2
4/2009-04-04/meta-data/instance-id' failed [-16620/120s]: request error [(<urll
compat_monitor0 console  

Ie, the clock got set back 17000 seconds or something.

Asking in # python, I was told that in python3.3 I could use 'time.monotonic'.

In python 2.X, it seems that reading /proc/cpuinfo might be my only option.

### LINKED ISSUE #3149: timeout can be wrong if system time change
This bug was originally filed in Launchpad as [LP: #1762392](https://bugs.launchpad.net/cloud-init/+bug/1762392)
<details>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2018-04-09T11:40:05.587731+00:00
date_fix_committed = None
date_fix_released = None
id = 1762392
importance = low
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1762392
milestone = None
owner = thomas-faivre
owner_name = Thomas Faivre
private = False
status = triaged
submitter = thomas-faivre
submitter_name = Thomas Faivre
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Thomas Faivre(thomas-faivre)** wrote on 2018-04-09T11:40:05.587731+00:00_

Seen on Xenial with a bad time on my hypervisor:

Apr  9 08:57:03 localhost cloud-init[1242]: 2018-04-09 08:57:03,749 - url_helper.py[WARNING]: Calling 'http://10.81.0.218/latest/meta-data/instance-id' failed [0/5s]: bad status code [404]
Apr  9 08:53:59 localhost ntpdate[1245]: step time server 10.81.0.218 offset -184.864291 sec
Apr  9 08:53:59 localhost systemd[1]: Time has been changed
Apr  9 08:53:59 localhost cloud-init[1242]: 2018-04-09 08:53:59,889 - url_helper.py[WARNING]: Calling 'http://10.81.0.218/latest/meta-data/instance-id' failed [-183/5s]: bad status code [404]
Apr  9 08:54:00 localhost cloud-init[1242]: 2018-04-09 08:54:00,893 - url_helper.py[WARNING]: Calling 'http://10.81.0.218/latest/meta-data/instance-id' failed [-182/5s]: bad status code [404]

I could just set the right time on my hypervisor but I believe there should be some mechanism in cloud-init to prevent this from happening.
In my case, it just increase the timeout by 3 minutes but with a positive offset, it could wrongfully skip the timeout.

A fix could be to prevent ntpdate to run until cloud-init is done but I am not sure if this is possible.

holmanb:
## Additional Context
This PR contains two commits related to system time.

The first fixes container to be more accurate with `cloud-init analyze` commands.
The second hardens cloud-init against clock changes, which are possible (and likely)
during early boot.

Fixes #2423
Fixes #3149

## Test Steps
See commit messages for test steps.

## Merge type

### DIFF
--- a/cloudinit/analyze/show.py
+++ b/cloudinit/analyze/show.py
@@ -6,7 +6,6 @@
 
 import datetime
 import json
-import os
 import sys
 import time
 
--- a/cloudinit/analyze/show.py
+++ b/cloudinit/analyze/show.py
@@ -243,19 +243,6 @@ def gather_timestamps_using_systemd():
         # lxc based containers do not set their monotonic zero point to be when
         # the container starts, instead keep using host boot as zero point
         if util.is_container():
-            # clock.monotonic also uses host boot as zero point
-            base_time = float(time.time()) - float(time.monotonic())
-            # TODO: lxcfs automatically truncates /proc/uptime to seconds
-            # in containers when https://github.com/lxc/lxcfs/issues/292
-            # is fixed, util.uptime() should be used instead of stat on
-            try:
-                file_stat = os.stat("/proc/1/cmdline")
-                kernel_start = file_stat.st_atime
-            except OSError as err:
-                raise RuntimeError(
-                    "Could not determine container boot "
-                    "time from /proc/1/cmdline. ({})".format(err)
-                ) from err
             status = CONTAINER_CODE
         kernel_end = base_time + delta_k_end
         cloudinit_sysd = base_time + delta_ci_s
--- a/cloudinit/cmd/main.py
+++ b/cloudinit/cmd/main.py
@@ -16,7 +16,6 @@ import argparse
 import json
 import os
 import sys
-import time
 import traceback
 import logging
 import yaml
@@ -764,7 +763,8 @@ def status_wrapper(name, args):
 
     v1 = status["v1"]
     v1["stage"] = mode
-    v1[mode]["start"] = time.time()
+    uptime = util.uptime()
+    v1[mode]["start"] = float(uptime)
     v1[mode]["recoverable_errors"] = next(
         filter(lambda h: isinstance(h, log.LogExporter), root_logger.handlers)
     ).export_logs()
@@ -791,7 +791,7 @@ def status_wrapper(name, args):
         print_exc("failed run of stage %s" % mode)
         v1[mode]["errors"] = [str(e)]
 
-    v1[mode]["finished"] = time.time()
+    v1[mode]["finished"] = float(util.uptime())
     v1["stage"] = None
 
     # Write status.json after running init / module code
--- a/cloudinit/config/cc_power_state_change.py
+++ b/cloudinit/config/cc_power_state_change.py
@@ -223,7 +223,7 @@ def run_after_pid_gone(pid, pidcmdline, timeout, condition, func, args):
     # is no longer alive.  After it is gone, or timeout has passed
     # execute func(args)
     msg = None
-    end_time = time.time() + timeout
+    end_time = time.monotonic() + timeout
 
     def fatal(msg):
         LOG.warning(msg)
@@ -232,7 +232,7 @@ def run_after_pid_gone(pid, pidcmdline, timeout, condition, func, args):
     known_errnos = (errno.ENOENT, errno.ESRCH)
 
     while True:
-        if time.time() > end_time:
+        if time.monotonic() > end_time:
             msg = "timeout reached before %s ended" % pid
             break
 
--- a/cloudinit/sources/DataSourceAzure.py
+++ b/cloudinit/sources/DataSourceAzure.py
@@ -14,7 +14,7 @@ import socket
 import xml.etree.ElementTree as ET  # nosec B405
 from enum import Enum
 from pathlib import Path
-from time import sleep, time
+from time import monotonic, sleep, time
 from typing import Any, Dict, List, Optional
 
 import requests
@@ -428,7 +428,7 @@ class DataSourceAzure(sources.DataSource):
         )
 
         lease = None
-        start_time = time()
+        start_time = monotonic()
         deadline = start_time + timeout_minutes * 60
         with events.ReportEventStack(
             name="obtain-dhcp-lease",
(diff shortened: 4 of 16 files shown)


cloudinit/analyze/show.py
```diff
@@ -243,19 +242,6 @@ def gather_timestamps_using_systemd():
         # lxc based containers do not set their monotonic zero point to be when
         # the container starts, instead keep using host boot as zero point
         if util.is_container():
```
igalic, line 244: I don't know many people using cloud-init in jails, but i need to test if this code works in FreeBSD

igalic (commented):
note to self

cloudinit/analyze/show.py
holmanb: @igalic +1 good idea, upstream cloud-init gets lots of usage and testing in lxc, but it would be good to know how this code performs in jails. Just running the following in a jail would probably tell us all that we need to know:
```python3
#!/usr/bin/python3
import os, time
from cloudinit.util import uptime
print(f"{uptime(),os.stat('/proc/1/cmdline').st_atime, time.monotonic()}")
```

igalic: well, there is no '/proc/1` in a FreeBSD jail, and if there was, there'd be no `/proc/1/cmdline`

FreeBSD doesn't provide PID name spaces (yet, like Solaris zones), and the `/proc` is very stripped down for security reasons

holmanb: > well, there is no '/proc/1`in a FreeBSD jail, and if there was, there'd be no`/proc/1/cmdline`
> 
> FreeBSD doesn't provide PID name spaces (yet, like Solaris zones), and the `/proc` is very stripped down for security reasons

I see - the others would still be useful to know

TheRealFalcon: `uptime()` takes a different path on BSD, so no `/proc/1` needed.

cloudinit/cmd/main.py
```diff
@@ -764,7 +763,8 @@ def status_wrapper(name, args):
 
     v1 = status["v1"]
     v1["stage"] = mode
-    v1[mode]["start"] = time.time()
+    uptime = util.uptime()
+    v1[mode]["start"] = float(uptime)
```
TheRealFalcon, line 767: I had assumed that this field was used to compare against actual timestamps to know when things happened rather than just tracking elapsed time. Maybe that's a bad assumption, though.

Thoughts?

TheRealFalcon (commented):
Generally LGTM. One question inline.

cloudinit/cmd/main.py
holmanb: > I had assumed that this field (along with finished) was used to compare against actual timestamps to know when things happened rather than just tracking elapsed time. Maybe that's a bad assumption, though.

The spot that I see it used is in `status.py` which uses the relative time between the stages, which is why `time.monotonic()` won't work.

I didn't find timestamps compared anywhere, and if the code was doing that with `time.time()`, such an implementation would also be susceptible to time change. With uptime in this field it can actually be used to manually correlate with events reported in dmesg / systemd logs / etc, so I think the new code is more correct than it was before

TheRealFalcon (approved):
LGTM!

cloudinit/analyze/show.py
igalic: for completeness:

```python
>>> print(f"{uptime(), time.monotonic()}")
('4598464.319990158', 4598464.320018191)
>>> os.system("uptime")
 9:20PM  up 53 days,  5:24, 0 users, load averages: 0.16, 0.23, 0.20
```

by default, FreeBSD jails don't have `/proc` mounted; uptime is a syscall which returns the kernel's uptime.

```
root@ports:~ # ps awuxdj
USER   PID %CPU %MEM   VSZ  RSS TT  STAT STARTED    TIME COMMAND           PPID  PGID   SID JOBC
root 21923  0.0  0.0 12768 2016  -  SsJ  Mon16   0:00.52 /usr/sbin/syslog     1 21923 21923    0
root 31780  0.0  0.1 14000 3324  5  SJ   21:10   0:00.06 tcsh             31429 31780 31429    1
root 12419  0.0  0.1 13600 2580  5  R+J  21:27   0:00.00 ps awuxdj        31780 12419 31429    1
root 50023  0.0  0.1 16560 2108  3  I+J  Mon16   0:00.03 tcsh             49556 50023 49556    1
root@ports:~ # 
```

the oldest process' PPID here is the host system's PID 1

Anyway, I'll look into this as soon as people start seriously using cloud-init in jails, or if I find a good way of testing it under jails

holmanb: @igalic thanks! it looks like `uptime()` should work for jails too

igalic: uptime works, but it's the host's uptime, so often can be a lie. then again, I don't know how important this is to be accurate

holmanb: +1 Sounds good. This isn't a very high priority without a bunch of users. This change is only relevant for `analyze` subcommands - users that are manually digging into performance.

PR #5247, opened 2024-05-01
test: Add v2 test coverage to test_net.py

TheRealFalcon:
## Proposed Commit Message

```
test: Add v2 test coverage to test_net.py

Ensure there is a corresponding v2 test for every v1 test in
`network_configs.py` where possible/reasonable.

Many tests are marked with `xfail` as they will not pass without
additional changes to the network rendering code.
```

## Additional Context
I left the individual commits as they may (or may not) be more useful in reviewing. This PR should be squashed though.

## Merge type


tests/unittests/net/network_configs.py
```diff
@@ -2265,41 +3495,41 @@
     bond-downdelay 10
     bond-fail-over-mac active
     bond-master bond0
+    bond_miimon 100
```
holmanb, line 3498: is this change intentional? (repeat the same question for the next several changes)

tests/unittests/net/network_configs.py
```diff
@@ -2586,22 +3729,22 @@
                 method=manual
                 may-fail=false
                 address1=192.168.0.2/24
-                gateway=192.168.0.1
-                route1=10.1.3.0/24,192.168.0.3
+                route1=0.0.0.0/0,192.168.0.1
```
holmanb, line 3732: is this change intentional? (repeat for the next several changes)

tests/unittests/test_net.py
```diff
@@ -1905,18 +1923,28 @@ def test_all_config(self, caplog):
...
         )
         self._compare_files_to_expected(entry[self.expected_name], found)
         self._assert_headers(found)
         expected_msg = (
             "Network config: ignoring iface0 device-level mtu:8999"
             " because ipv4 subnet-level mtu:9000 provided."
         )
-        assert expected_msg in caplog.text
+        if yaml_file == "yaml_v1":
+            assert expected_msg in caplog.text
```
holmanb, line 1947: No test for `yaml_v2`?

tests/unittests/test_net.py
```diff
@@ -2765,42 +2810,77 @@ def test_v4_and_v6_static_config(self, caplog):
...
+                        "DNS not rendered when DHCP is enabled."
+                    ),
+                ),
+            ),
         ],
     )
-    def test_config(self, expected_name):
+    def test_config(self, expected_name, yaml_name):
         entry = NETWORK_CONFIGS[expected_name]
-        yaml_name = "yaml" if "yaml" in entry else "yaml_v2"
```
holmanb, line 2796: +1 for cleanup

holmanb (commented):
I'm just getting back to this after a long while, sorry for the delay.

Overall this looks good, thanks for the PR! I just have a couple of questions inline.

holmanb (changes requested):
> is this change intentional

I assume not, so setting this to "Request changes".

tests/unittests/net/network_configs.py
TheRealFalcon: Yes. Github unfortunately renders this very poorly. You're looking at the v2 addition, and when using v2 things get rendered alphabetically. This is a change between v1 and v2 rendering, but it results in no functional difference so it is fine.

This may better demonstrate the difference here. Notice the left has expanded the expected v1 whereas the right has expanded the expected v2:
image

tests/unittests/net/network_configs.py
TheRealFalcon: Yes, like the miimon v2 renders slightly differently and github is doing a poor job rendering it. v2 renders a 0.0.0.0 route rather than gateway, but functionally it is the same.

image

tests/unittests/test_net.py
TheRealFalcon: We get no warning in v2 because we wind up testing slightly different things.  v2 has no concept of interface level and subnet level MTU, so that's not something we can test for in v2.

yaml v1:
```
            version: 1
            config:
              - type: 'physical'
                name: 'iface0'
                mtu: 8999
                subnets:
                  - type: static
                    address: 192.168.14.2/24
                    mtu: 9000
                  - type: static
                    address: 2001:1::1/64
                    mtu: 1500
```

yaml v2:
```
            version: 2
            ethernets:
                iface0:
                    addresses:
                    - 192.168.14.2/24
                    - 2001:1::1/64
                    mtu: 9000
```

TheRealFalcon:
@holmanb , I believe I answered all of your questions. Ready for re-review.

holmanb (approved):
Thanks for responding to my questions @TheRealFalcon. Looks good to me!

PR #5250, opened 2024-05-01
chore: Update systemd description

holmanb:
## Proposed Commit Message
```
chore: Update systemd description

The current descriptions do not follow systemd guidance and are misleading since some services have multiple roles.

systemd.unit(5) on Description:

    A short human readable title of the unit. This may be used by systemd
    (and other UIs) as a user-visible label for the unit, so this string
    should identify the unit rather than describe it, despite the name.

Yet these service files attempt to describe the unit rather than identify it.

Before:
May 01 09:49:49.238629 tiny systemd[1]: Starting cloud-config.service - Apply the settings specified in cloud-config...


After:
May 01 09:49:49.238629 tiny systemd[1]: Starting cloud-config.service - Cloud-init: Config Stage...
```

## Additional Context
There's also the fact that the service files allude to the hotplug command being a daemon when it actually isn't, but that's for a different PR. 

## Merge type

### DIFF
--- a/systemd/cloud-init-local.service.tmpl
+++ b/systemd/cloud-init-local.service.tmpl
@@ -1,6 +1,7 @@
 ## template:jinja
 [Unit]
-Description=Initial cloud-init job (pre-networking)
+# https://cloudinit.readthedocs.io/en/latest/explanation/boot.html
+Description=Cloud-init: Local Stage
 {% if variant in ["ubuntu", "unknown", "debian", "rhel" ] %}
 DefaultDependencies=no
 {% endif %}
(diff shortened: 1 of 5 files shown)


systemd/cloud-init-local.service.tmpl
```diff
@@ -1,6 +1,7 @@
 ## template:jinja
 [Unit]
-Description=Initial cloud-init job (pre-networking)
+# https://cloudinit.readthedocs.io/en/latest/explanation/boot.html
+Description=Cloud-init: Local Stage
```
blackboxsw, line 4: I really want to keep `(pre-networking)` somewhere in there as Local turns into something generic that is not understood unless someone actually opens the uint file, sees the commented link and reads about the different stages. pre-network also implies a more clear ordering that is easier to decipher.
WDYT?
```suggestion
Description=Cloud-init: Local Stage (pre-network)
```

blackboxsw (approved):
Other than a nit for a slightly more details suffix for init-local +1

PR #5252, opened 2024-05-02
fix(crypt): Add salt to passlib calls

holmanb:
## Proposed Commit Message
```
fix(crypt): Add salt to passlib calls

- Refactor multiple crypt callsites to a single Python module.
- Add tests to verify standardized behavior between passlib and crypt.

[1] https://github.com/canonical/cloud-init/issues/4791
```

## Additional Context
It would be trivial to add openssl support as well, but this isn't a dependency in minimal images currently so probably not worth it.

In the standard library `hashlib.sha512()` also exists, but this isn't in the right format and I couldn't find documentation to make this compatible with `crypt`. 

## Merge type

### DIFF
--- a/cloudinit/crypt.py
+++ b/cloudinit/crypt.py
@@ -1,0 +1,36 @@
+# This file is part of cloud-init. See LICENSE file for license information.
+from contextlib import suppress
+
+
+def _not_available(secret: str, salt: str):
+    """Raise when called so that importing this module doesn't throw
+    ImportError when ds_detect() returns false. In this case, crypt
+    and passlib are not needed.
+    """
+    raise ImportError("crypt/passlib not found, missing dependency")
+
+
+encrypt_pass = _not_available
+_passlib_crypt = _not_available
+_deprecated_crypt = _not_available
+
+
+# lowest priority: external python dependency with an uncertain future
+with suppress(ImportError, AttributeError):
+    import passlib.hash
+
+    def _passlib_crypt(secret: str, salt: str):
+        """crypt.crypt() interface based on passlib"""
+        return passlib.hash.sha512_crypt.hash(secret, salt=salt, rounds=5000)
+
+    encrypt_pass = _passlib_crypt
+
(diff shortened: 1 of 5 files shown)


TheRealFalcon (commented):
Nice refactor, but is there a reason to provide our own salt? Passlib will automatically provide a salt by default.

cloudinit/crypt.py
```diff
@@ -0,0 +1,36 @@
...
+_deprecated_crypt = _not_available
+
+
+# lowest priority: external python dependency with an uncertain future
+with suppress(ImportError, AttributeError):
+    import passlib.hash
+
+    def _passlib_crypt(secret: str, salt: str):
+        """crypt.crypt() interface based on passlib"""
+        return passlib.hash.sha512_crypt.hash(secret, salt=salt, rounds=5000)
```
TheRealFalcon, line 24: Doesn't BSD require `bcrypt` rather than `sha512_crypt`?

Also, this method of calling `hash` is [deprecated](https://passlib.readthedocs.io/en/stable/lib/passlib.ifc.html#passlib.ifc.PasswordHash.hash). I think it should instead look something like
```python
passlib.hash.sha512_crypt.using(salt=salt, rounds=5000).hash(secret)
```

holmanb:
> Nice refactor, but is there a reason to provide our own salt? Passlib will automatically provide a salt by default.

I guess not really then - the idea of making sure that the behavior "matches" was appealing from a testing perspective, but so long as passlib actually has a future, then I guess library defaults might be better.

Aside: before I pushed this PR I actually had an openssl implementation that also matched the passlib and crypt output, but I deleted it because I didn't want cloud-init to pull it into minimal images.

holmanb:
> Doesn't BSD require bcrypt rather than sha512_crypt?

Ugh, you're right. I'm going to close this PR since there isn't much benefit to it as is, and there really isn't much code to share between the two call sites.

PR #5262, opened 2024-05-05
chore: Add docstrings and types to Version class

holmanb:
## Additional Context
chore: Add docstrings and types to Version class

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3113,13 +3113,63 @@ def error(msg, rc=1, fmt="Error:\n{}", sys_exit=False):
 
 @total_ordering
 class Version(namedtuple("Version", ["major", "minor", "patch", "rev"])):
-    def __new__(cls, major=-1, minor=-1, patch=-1, rev=-1):
+    """A class for comparing versions.
+
+    Implemented as a named tuple with all ordering methods.
+
+    Args:
+
+        major: the most significant number in a version
+        minor: next greatest significant number after major
+        patch: next greatest significant number after minor
+        rev: the least significant number in a version
+
+    Raises:
+
+        TypeError: If invalid arguments are given.
+        ValueError: If invalid arguments are given.
+
+    Examples:
+
+        >> Version(2, 9) == Version.from_str("2.9")
+        True
+        >> Version(2, 9, 1) > Version.from_str("2.9.1")
+        False
+        >> Version(3, 10) > Version.from_str("3.9.9.9")
+        True
+        >> Version(3, 7) >= Version.from_str("3.7")
+        True
+
+    Note:
+
+        Comparisons between X.Y.N and X.Y always treats the more
+        specific number as larger.
+    """
+
+    def __new__(
+        cls, major: int = -1, minor: int = -1, patch: int = -1, rev: int = -1
+    ) -> "Version":
         """Default of -1 allows us to tiebreak in favor of the most specific
         number"""
         return super(Version, cls).__new__(cls, major, minor, patch, rev)
 
     @classmethod
-    def from_str(cls, version: str):
+    def from_str(cls, version: str) -> "Version":
+        """Create a Version object from a string.
+
+        Args:
+
+            version: A period-delimited version string, max 4 segments.
+
+        Return:
+
+            A version object.
+
+        Exception:
+
+            TypeError: Raised if invalid arguments are given.
+            ValueError: Raised if invalid arguments are given.
+        """
         return cls(*(list(map(int, version.split(".")))))
 
     def __gt__(self, other):
@@ -3144,7 +3194,10 @@ class Version(namedtuple("Version", ["major", "minor", "patch", "rev"])):
     def __str__(self):
         return ".".join(self)
 
-    def _compare_version(self, other) -> int:
+    def __hash__(self):
+        return hash(str(self))
+
+    def _compare_version(self, other: "Version") -> int:
         """
         return values:
             1: self > v2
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):
I like that you are following google's python style guide: https://google.github.io/styleguide/pyguide.html

Let's avoid extra empty newlines in the docstrings as we are dedicating a lot of vertical real-estate to these docstrings already.

TheRealFalcon:
I have thoughts...let's not merge this quite yet

TheRealFalcon:
nit: If we change `>>` to `>>>`, these would be valid doctest examples. We don't use it, but it'd be an easy way to gain test coverage for utility functions in the future and ensure our docstring examples are kept up to date.

More broadly, if we're wanting to standardize on a format, I'd prefer we use the [Sphinx](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) format. Reasons:
1. Since we're already using Sphinx, if we ever decide to autogenerate docs docstrings, it should be a fairly easy transition.
2. Since the format is very similar to the epytext strings that are scattered through the codebase, it'd be a much simpler find/replace for those than switching them to Google format would be.
3. Minor point, but Google convention technically breaks pep-8/pep-257 in things like recommending non-imperative mood. If we use a linter and say "use google style", our currently existing docstrings that use imperative mood may not match.

Since [existing tooling](https://pypi.org/project/pydoclint/) generally supports both formats, I don't think there's any loss from not using Google format.

holmanb:
> nit: If we change `>>` to `>>>`, these would be valid doctest examples. We don't use it, but it'd be an easy way to gain test coverage for utility functions in the future and ensure our docstring examples are kept up to date.

No objections from me for cheap test coverage :D

> More broadly, if we're wanting to standardize on a format, I'd prefer we use the [Sphinx](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) format. Reasons:
> 
>     1. Since we're already using Sphinx, if we ever decide to autogenerate docs docstrings, it should be a fairly easy transition.
> 
>     2. Since the format is very similar to the epytext strings that are scattered through the codebase, it'd be a much simpler find/replace for those than switching them to Google format would be.
> 
>     3. Minor point, but Google convention technically breaks pep-8/pep-257 in things like recommending non-imperative mood. If we use a linter and say "use google style", our currently existing docstrings that use imperative mood may not match.
> 
> 
> Since [existing tooling](https://pypi.org/project/pydoclint/) generally supports both formats, I don't think there's any loss from not using Google format.

Good points.

The only reservations that I have are that sphinx has a little more boilerplate and a little more work for my eyes to parse than google-styled docstrings. I'm sure I'll get used to sphinx docstrings without issue. I'll update this PR unless @blackboxsw or @aciba90 has any strong objections.

TheRealFalcon:
> boilerplate

I don't think we need to include every field every time. E.g., there's a [stackoverflow](https://stackoverflow.com/a/24385103/37205) answer that has this:
```python
"""
This is a reST style.

:param param1: this is a first param
:param param2: this is a second param
:returns: this is a description of what is returned
:raises keyError: raises an exception
"""
```
Which is a lot less boilerplate. Since we (ideally) have types specific in the function definition, we shouldn't also need them in the docstring.

aciba90:
Thanks for the discussion.

I do not care much about which docstring format we use, but more about picking up one and be consistent with it.

If we choose a new one, we could use https://github.com/dadadel/pyment to convert existent docstrings if we want.

To lint / enforce docstrings to follow a particular style, we could use https://docs.astral.sh/ruff/settings/#lint_pydocstyle_convention, since https://github.com/PyCQA/pydocstyle has been deprecated in favor of ruff.

holmanb:
> If we choose a new one, we could use https://github.com/dadadel/pyment to convert existent docstrings if we want.

I can see this being useful to get started with some modules that are more important. These are the ones that I would start with

```
stages.py/main.py -> because it defines core logic
util.py -> because is so heavily used across the codebase
distros/__init__.py -> because it defines an interface which gets implemented by all distros
sources/__init__.py -> because it defines an interface which gets implemented by all platforms
```

> To lint / enforce docstrings to follow a particular style, we could use https://docs.astral.sh/ruff/settings/#lint_pydocstyle_convention, since https://github.com/PyCQA/pydocstyle has been deprecated in favor of ruff.

+1

holmanb:
>> To lint / enforce docstrings to follow a particular style, we could use https://docs.astral.sh/ruff/settings/#lint_pydocstyle_convention, since https://github.com/PyCQA/pydocstyle has been deprecated in favor of ruff.
>
> +1

@aciba90 fyi https://github.com/canonical/cloud-init/pull/5277

aciba90:
> > > To lint / enforce docstrings to follow a particular style, we could use https://docs.astral.sh/ruff/settings/#lint_pydocstyle_convention, since https://github.com/PyCQA/pydocstyle has been deprecated in favor of ruff.
> > 
> > 
> > +1
> 
> @aciba90 fyi #5277

Thanks for pinging!

@holmanb I went into a rabbit hole: https://github.com/canonical/pycloudlib/pull/374

holmanb:
@TheRealFalcon @aciba90 @blackboxsw I just updated with a sphinx docstring fyi

TheRealFalcon (approved):
This version LGTM, but lets give @blackboxsw and @aciba90 time for any additional bikeshedding.

holmanb:
@aciba90 I misread your earlier comment, my apologies. I thought that this was suggesting to add docstrings by default to all functions, but now that I've re-read I understand what you said.

> If we choose a new one, we could use https://github.com/dadadel/pyment to convert existent docstrings if we want.

Yes 100% agreed. I think that we should do a bulk update of all comments in the project (or at least in `cloudinit/`) to match this format. Piecemeal updates will take forever. I just filed https://github.com/canonical/cloud-init/issues/5346 to track this effort.

PR #5299, opened 2024-05-17
Add support for Aeza.net hosting provider

cofob:
## Proposed Commit Message

```
Add support for Aeza.net hosting provider

Added a DataSource for the Aeza.net hosting provider, enabling
metadata and user data retrieval and instance provisioning via
cloud-init.
```

## Additional Context

Test urls:
- <http://77.221.156.49/v1/cloudinit/3c0fff62-b1db-4165-a2b2-f819c5ddf83f/meta-data>
- <http://77.221.156.49/v1/cloudinit/3c0fff62-b1db-4165-a2b2-f819c5ddf83f/user-data>
- <http://77.221.156.49/v1/cloudinit/3c0fff62-b1db-4165-a2b2-f819c5ddf83f/vendor-data>

## Test Steps

At the moment cloud-init support in our service is in draft state. We have tested the functionality of these changes.

## Checklist

## Merge type


cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,81 @@
...
+    "userdata_url": BASE_URL_V1 + "/{id}/user-data",
+    "vendordata_url": BASE_URL_V1 + "/{id}/vendor-data",
+}
+
+class DataSourceAeza(sources.DataSource):
+
+    dsname = "Aeza"
+
+    def __init__(self, sys_cfg, distro, paths, ud_proc=None):
+        sources.DataSource.__init__(self, sys_cfg, distro, paths, ud_proc)
```
blackboxsw, line 26: let's use super() here
```suggestion
        super().__init__(self, sys_cfg, distro, paths, ud_proc)
```

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,81 @@
...
+        self.vendordata_raw = None
+        self.userdata_raw = ""
+
+        self.ds_cfg = util.mergemanydict(
+            [
+                util.get_cfg_by_path(sys_cfg, ["datasource", "Aeza"], {}),
+                BUILTIN_DS_CONFIG,
+            ]
+        )
+
```
blackboxsw, line 37: super().__init__ already sets these up no need to redefine here.
```suggestion
```

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,81 @@
...
+        self.timeout = self.ds_cfg["timeout"]
+        self.wait_retry = self.ds_cfg["wait_retry"]
+        self.retries = self.ds_cfg["retries"]
```
blackboxsw, line 40: Generally you should avoid this setting in the instance as you can call self.get_url_params() which will use the standard config option names: timeout, max_wait, retries and sec_between_retries. Then you don't have to define datasource-specific config keys.

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.metadata_address = self.ds_cfg["metadata_url"]
+        self.userdata_address = self.ds_cfg["userdata_url"]
+        self.vendordata_address = self.ds_cfg["vendordata_url"]
+
+    def _get_data(self):
```
blackboxsw, line 46: we need a ds_detect method on this class which returns True/False that cloud-init python code queries first to see if this platform is even detectable as Azea before trying to read metadata. We don't want to waste time trying to reach out to a remote IP on systems if they are unlikely to be Aeza as it costs boot time for every environment which is not aeza. Some images do not call ds-identify early in boot, and those systems have to walk through all python datasources to run _get_data which ends up taking 10+ minutes in some cases due to network retries and timeouts if datasources don't have an early exit via ds_detect() which is called before _get_data

tools/ds-identify
```diff
@@ -1558,6 +1558,11 @@ dscheck_NWCS() {
     return ${DS_NOT_FOUND}
 }
 
+dscheck_Aeza() {
+    dmi_sys_vendor_is Aeza && return ${DS_FOUND}
```
blackboxsw, line 1562: Given that we check sys vendor id for Aeza as a match for the datasource, in ds-identify, this same detection logic should be performed in a `DataSourceAeza.ds_detect` method in python which returns a boolean True if vendor = dmi.read_dmi_data("system-manufacturer") == "Aeza"

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,81 @@
...
+
+        self.metadata = md
+        self.userdata_raw = ud
+        self.vendordata_raw = vd
+
+        return True
+
+
+datasources = [
+    (DataSourceAeza, (sources.DEP_NETWORK, sources.DEP_FILESYSTEM)),
```
blackboxsw, line 76: Does Aeza provide viable dhcp services to all instances when started? If so, cloud-init does have machanisms to invoke a dhclient in a sandbox environment earlier in boot (only needing sources.DEP_FILESYSTEM) which allows cloud-init to just get enough network up to talk to the IMDS IP address to get full system and network config. See Azure/ec2.vultr/hetzner/GCE uses of EphemeralDHCP for examples

cloudinit/sources/helpers/aeza.py
```diff
@@ -0,0 +1,87 @@
...
+# This file is part of cloud-init. See LICENSE file for license information.
+
+import requests
+import time
+from requests.exceptions import HTTPError, RequestException
+
+from cloudinit import util, dmi
+
+
+def read_url(url, timeout=None, sec_between=1, retries=0):
```
blackboxsw, line 14: Let's not reinvent read_url helpers if we can avoid it as it makes it harder to maintain utility code that evolves over time. Is there something that [cloudinit.url_helper.readurl](https://github.com/canonical/cloud-init/blob/main/cloudinit/url_helper.py#L353) doesn't do for you?

cloudinit/sources/helpers/aeza.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def format_url(url, system_uuid=None):
+    if system_uuid is None:
+        system_uuid = read_system_uuid()
+    return url.format(id=system_uuid)
+
+
+def read_system_uuid():
+    return dmi.read_dmi_data("system-uuid")
```
blackboxsw, line 87: Small functions like this are probably unnecessary and can just be inlined in DataSourceAeza.py module for clarity. It just adds a layer of obscurity where not a lot of complexity is needed.

cloudinit/sources/helpers/aeza.py
```diff
@@ -0,0 +1,87 @@
...
+        response = read_url(
+            url, timeout=timeout, sec_between=sec_between, retries=retries
+        )
+        if not response.ok:
+            raise RuntimeError("unable to read data at %s" % url)
+        return response.content
+    except HTTPError as e:
+        if e.response.status_code == 404:
+            return None
+        raise e
```
blackboxsw, line 77: These feel like they could be functions defined in DataSourceAeza.py if needed to avoid sprawl of separate modules in sources/helpers if we can avoid it.

blackboxsw (changes requested):
Thank you much @cofob for thinking of adding cloud-init support and driving this pull request to make cloud-init better on your platform. I've added some intiial review comments, suggestions and questions in this first pass. I'd like to see if we can include this support in our next [24.3 release milestone](https://github.com/canonical/cloud-init/milestone/2) and we'll work with you to get this across the line.

cloudinit/sources/DataSourceAeza.py
blackboxsw: I'm guessing that given that the IP address is a publicly routable IP address. that it's unlikely that the platform would support some private network bringup with access to the primary URL. But I thought I'd ask.

blackboxsw:
Also a general security question here. Given that user-data and vendor-data likely contains potentially sensitive information and Aeza.net's instance metadata endpoint is providing content over http that appears to be accessible to anyone with the URL http://77.221.156.49/v1/cloudinit/3c0fff62-b1db-4165-a2b2-f819c5ddf83f/meta-data and UUIDs can be guessable, what sort of authentication layer do you have planned to avoid potential exposure of sensitive information to unwanted parties?

cloudinit/sources/DataSourceAeza.py
cofob: Fixed

cloudinit/sources/DataSourceAeza.py
cofob: Fixed in 4350dc221feb40ba2e277706e890126f92cee282
`self.ds_cfg` merge had to remain, otherwise the tests didn't work.

cloudinit/sources/DataSourceAeza.py
cofob: Fixed in 4350dc221feb40ba2e277706e890126f92cee282

cloudinit/sources/DataSourceAeza.py
cofob: Fixed in 4350dc221feb40ba2e277706e890126f92cee282

tools/ds-identify
cofob: Fixed in 4350dc221feb40ba2e277706e890126f92cee282

cloudinit/sources/DataSourceAeza.py
cofob: This IP is publicly available, Aeza has no DHCP and the virtual machine is created with a pre-configured network. cloud-init should not perform any network configuration.

cloudinit/sources/helpers/aeza.py
cofob: Removed `read_system_uuid` in 4350dc221feb40ba2e277706e890126f92cee282
I don't see need in inlining `format_url` function

cloudinit/sources/helpers/aeza.py
cofob: Fixed in 4350dc221feb40ba2e277706e890126f92cee282

cloudinit/sources/helpers/aeza.py
cofob: Also it doesn't support `max_wait` parameter.

cofob:
@blackboxsw can you please take a look?

cofob:
@TheRealFalcon can you please take a look?

cloudinit/sources/DataSourceAeza.py
blackboxsw: Thank you. My mistake on inclusion there. I overlooked the BUILTIN_DS_CONFIG you are merging as well.

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,172 @@
...
+        self.vendordata_address = self.ds_cfg["vendordata_url"]
+
+    @staticmethod
+    def ds_detect():
+        return dmi.read_dmi_data("system-manufacturer") == "Aeza"
+
+    def _get_data(self):
+        system_uuid = dmi.read_dmi_data("system-uuid")
+
+        md = read_metadata(
```
blackboxsw, line 137: The way read_metadata is written, 404s in any of these route returns a None value.

I don't think we generally want to allow for absent or optional metadata or user-data endpoints to have a correctly discovered datasource. 

For cloud-init to be certain it has contacted a viable instance metadata service that matches the discovered datasource type and not some bogus service that happened to be stood up at this public IP address, each datasource generally requires  minimally meta-data and user-data files or components to exist (see DataSourceNoCloud as an example).
 
The meta-data should really be a yaml dict that minimally contains an unique `instance-id` which cloud-init caches and acts as an indicator to cloud-init that it has applied the desired configuration from the Aeza.net, if that instance-id changes, cloud-init then knows desired configuration has changed for this instance, and it needs to read and apply new configuration for this node. This includes re-creating SSH host keys, setting up default users or whatever other config operations are declared from meta-data/user-data files. 

Other keys in meta-data that are optional: 
- `local-hostname`: the cloud platform's desired hostname for this node
- `cloud-name`: if desired (it will fallback to the `DataSourceAeza.dsname.lower()`)
- `public-keys`: List of public keys to import for the default user
- `region`: string representing the physical region
-  `availability-zone`:  string representing separate AZ if applicable 


Again best practice is to provide minimally `instance-id` that give the platform a knob to trigger a re-run of cloud-init across system boot if necessary due to mandated config/provisioning changes. The rest are optional.

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,172 @@
...
+            format_url(self.metadata_address, system_uuid),
+            timeout_seconds=self.timeout_seconds,
+            max_wait_seconds=self.max_wait_seconds,
+            sec_between_retries=self.sec_between_retries,
+            retries=self.retries,
+        )
+        ud = read_data(
```
blackboxsw, line 144: As mentioned above, we ideally want to see a user-data component from an IMDS endpoint, even if it is an empty string "" or and empty cloud-config:
```
#cloud-config
{}
```

If we don't see both non-empty md and ud values, it's probably something we want to log and `return False`, "Aeza.net doesn't have applicable meta-data and/or user-data for machine {id}". 


This would allow cloud-init to attempt to discover other lower priority datasources  or would raise an error if only one `datasource_list: [ aeza ]` is defined and the IMDS didn't have applicable config for this system-uuid.

doc/rtd/reference/datasources/aeza.rst
```diff
@@ -0,0 +1,36 @@
...
+Configuration
+=============
+
+Aeza's datasource can be configured as follows: ::
+
+  datasource:
+    Aeza:
+      metadata_url: "http://77.221.156.49/v1/cloudinit/{id}/meta-data"
+      userdata_url: "http://77.221.156.49/v1/cloudinit/{id}/user-data"
+      vendordata_url: "http://77.221.156.49/v1/cloudinit/{id}/vendor-data"
```
blackboxsw, line 19: One design question from a configuration standpoint: do we really want to have 3 separate configuration keys for meta-data/vendor-data and user-data? There are all routes under `http://77.221.156.49/v1/cloudinit/{id}/`. Do you expect some deployments would require very unique separate URLs that don't share the same prefix?

If not may we just use `metadata_url: "http://77.221.156.49/v1/cloudinit/{id}/"`?

tools/ds-identify
blackboxsw: Thank you, looks good!

doc/rtd/reference/datasources/aeza.rst
```diff
@@ -0,0 +1,36 @@
...
+
+* ``metadata_url``
+
+  Specifies the URL to retrieve the VPS meta-data. (optional)
```
blackboxsw, line 23: If the intent of `(optional)` is to say you don't have to provide this datasource config key, let's also provide the `Default: http://77.221.156.49/v1/cloudinit/{id}/`

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,172 @@
...
+
+        self.ds_cfg = util.mergemanydict([self.ds_cfg, BUILTIN_DS_CONFIG])
+
+        url_params = self.get_url_params()
+        self.timeout_seconds = url_params.timeout_seconds
+        self.max_wait_seconds = url_params.max_wait_seconds
+        self.retries = url_params.num_retries
+        self.sec_between_retries = url_params.sec_between_retries
+
+        self.metadata_address = self.ds_cfg["metadata_url"]
```
blackboxsw, line 126: This is good to set in __init__ and on the instance itself because [get_sub_platform](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L679) references this instance atttribute for user/script-facing instance data queries like `cloud-init query v1` or `cloud-id --long`.

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,172 @@
...
+        self.userdata_address = self.ds_cfg["userdata_url"]
+        self.vendordata_address = self.ds_cfg["vendordata_url"]
```
blackboxsw, line 128: I don't see why we need this instance attrs assigned in __init__. they are only used in _get_data below.
```suggestion
```

cloudinit/sources/DataSourceAeza.py
blackboxsw: Looks good thank you.__

cloudinit/sources/DataSourceAeza.py
blackboxsw: Ok thank you for the context.

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,172 @@
...
+            retries=self.retries,
+        )
+        ud = read_data(
+            format_url(self.userdata_address, system_uuid),
+            timeout_seconds=self.timeout_seconds,
+            max_wait_seconds=self.max_wait_seconds,
+            sec_between_retries=self.sec_between_retries,
+            retries=self.retries,
+        )
+        vd = read_data(
```
blackboxsw, line 151: This behavior closely aligns with our existing DataSourceNoCloud which reads mandatory metadata and user-data files and optional vendor-data and network-config files. Do you think instead of developing a specialized read_data/read_url set of functions, you could use the existing [util.read_seeded function which will walk throgh md, ud and vd for you with a timeout and retries while coping with missing vendor-data](https://github.com/canonical/cloud-init/blob/main/cloudinit/util.py#L1064)?

blackboxsw (changes requested):
Thank you for the pings on this Pull request and continued work here. PTO got in the way her . I think there are still a couple of open discussion points we should understand (mandatory metadata/user-data files) possibly the use of util.read_seeded and moving some unnecessary side-effects out of `__init__`

blackboxsw:
Hi @cofob, I don't see that your github username has signed the cloud-init's [Contributor License Agreement](https://cloudinit.readthedocs.io/en/latest/development/first_PR.html#sign-the-cla). So that I can enable CI on this PR and move toward merging this PR, would you mind reviewing and signing our CLA so that we may include this work in upstream cloud-init?

cloudinit/sources/DataSourceAeza.py
TheRealFalcon: I don't think the lack of user data is really a problem here. It gets set in the parent class to `None` and if not present cloud-init carries on without it.

cloudinit/sources/DataSourceAeza.py
TheRealFalcon: > The meta-data should really be a yaml dict that minimally contains an unique instance-id which cloud-init caches

I think we might be being a tad too prescriptive here. At the end of the day, the metadata is useful so that the cloud can get the desired functionality from cloud-init. The form it takes doesn't really matter as long as the cloud-init datasource can scrape it and set the internal structures correctly. Cloud-init can even work just fine without any metadata, but I agree that in order to get cloud-init to do the right thing on snapshots and new instances, a cloud should really be giving us an instance-id...but at the end of the day I wouldn't block on it.

cofob:
> Hi @cofob, I don't see that your github username has signed the cloud-init's [Contributor License Agreement](https://cloudinit.readthedocs.io/en/latest/development/first_PR.html#sign-the-cla). So that I can enable CI on this PR and move toward merging this PR, would you mind reviewing and signing our CLA so that we may include this work in upstream cloud-init?

I signed the CLA

cloudinit/sources/DataSourceAeza.py
cofob: Fixed in 4e8b32dfd7ce0751ca0781a6894600e2e70f1a2a

cloudinit/sources/DataSourceAeza.py
cofob: Fixed in 4e8b32dfd7ce0751ca0781a6894600e2e70f1a2a

doc/rtd/reference/datasources/aeza.rst
cofob: Fixed in 4e8b32dfd7ce0751ca0781a6894600e2e70f1a2a

doc/rtd/reference/datasources/aeza.rst
cofob: Fixed in 4e8b32dfd7ce0751ca0781a6894600e2e70f1a2a

cofob:
@blackboxsw I have made the changes according to your comments.

doc/rtd/reference/datasources/aeza.rst
```diff
@@ -0,0 +1,27 @@
...
+Aeza's datasource can be configured as follows: ::
+
+  datasource:
+    Aeza:
+      metadata_url: "http://77.221.156.49/v1/cloudinit/{id}/"
+
+* ``metadata_url``
+
+  Specifies the URL to retrieve the VPS meta-data. (optional)
+  Default: ``http://http://77.221.156.49/v1/cloudinit/{id}/``
```
marat2509, line 22:
```suggestion
  Default: ``http://77.221.156.49/v1/cloudinit/{id}/``
```

cofob:
@blackboxsw can you review my changes please?

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+            + "%s"
+        )
+
+    @staticmethod
+    def ds_detect():
+        return dmi.read_dmi_data("system-manufacturer") == "Aeza"
+
+    def _get_data(self):
+        md, ud, vd = util.read_seeded(
```
blackboxsw, line 46: A commit commit landed in tip of main that is incompatible with this changeset as read_seeded now returns a 4-tuple from util.read_seeded which looks for a supplemental/optional network-config file.We need to git rebase on this branch to refresh to latest commit on main and possibly ignore that network-config 4-tuple return value. I haven't rebased your PR against upstream/main and pushed my suggested changes as a commit because we may actually want to think about providing a supplemental parameter to allow read_seeded to ignore GETs of network-config since network-config isdisregared by this platform and we don't want to waste URL retries on an unexpected/irrelevant file.

blackboxsw (changes requested):
This update looks better thank you @cofob. I have a number of comments inline, one of which would be that we need to rebase this branch against tip of main and adapt the util.read_seeded to either expect an ignored _network_config return value as the 4th item of the 4-tuple,  or add a new param to util.read_seeded to `ignore_files=['network-config']` to avoid that round trip on your platform.

Below is a diff of all other inline changes/suggestions I've made in review
```diff
commit 1e5417d14fc44e868f5e87e01e707976b2c321ac (HEAD -> aeza)
Author: Chad Smith <chad.smith@canonical.com>
Date:   Thu Aug 15 17:45:27 2024 -0600

    review: move side-effects out of __init__ and into _get_data
    
    - factor side-effects out of __init__ and into _get_data
    - raise source.InvalidMetaDataError on unexpected format
    - update docs to mention other datasource config options retries/timeout
    - add unittests for ds_detect
    - use pytest instead of CiTestCase

diff --git a/cloudinit/sources/DataSourceAeza.py b/cloudinit/sources/DataSourceAeza.py
index 62f757095..fd43be43b 100644
--- a/cloudinit/sources/DataSourceAeza.py
+++ b/cloudinit/sources/DataSourceAeza.py
@@ -6,7 +6,7 @@
 
 import logging
 
-from cloudinit import sources, dmi, util
+from cloudinit import dmi, sources, util
 
 LOG = logging.getLogger(__name__)
 
@@ -24,45 +24,35 @@ class DataSourceAeza(sources.DataSource):
 
         self.ds_cfg = util.mergemanydict([self.ds_cfg, BUILTIN_DS_CONFIG])
 
-        url_params = self.get_url_params()
-        self.timeout_seconds = url_params.timeout_seconds
-        self.max_wait_seconds = url_params.max_wait_seconds
-        self.retries = url_params.num_retries
-        self.sec_between_retries = url_params.sec_between_retries
+    @staticmethod
+    def ds_detect():
+        return dmi.read_dmi_data("system-manufacturer") == "Aeza"
 
+    def _get_data(self):
         system_uuid = dmi.read_dmi_data("system-uuid")
-        self.metadata_address = (
+        metadata_address = (
             self.ds_cfg["metadata_url"].format(
                 id=system_uuid,
             )
             + "%s"
         )
-
-    @staticmethod
-    def ds_detect():
-        return dmi.read_dmi_data("system-manufacturer") == "Aeza"
-
-    def _get_data(self):
+        url_params = self.get_url_params()
         md, ud, vd = util.read_seeded(
-            self.metadata_address,
-            timeout=self.timeout_seconds,
-            retries=self.retries,
+            metadata_address,
+            timeout=url_params.timeout_seconds,
+            retries=url_params.num_retries,
         )
 
         if md is None:
-            LOG.warn(
-                "Failed to read metadata from %s",
-                self.metadata_address,
+            raise sources.InvalidMetaDataException(
+                f"Failed to read metadata from {metadata_address}",
             )
-            return False
         if not isinstance(md.get("instance-id"), str):
-            LOG.warn(
-                "Metadata does not contain instance-id",
+            raise sources.InvalidMetaDataException(
+                f"Metadata does not contain instance-id: {md}"
             )
-            return False
         if not isinstance(ud, bytes):
-            LOG.warn("Userdata is not bytes")
-            return False
+            raise sources.InvalidMetaDataException("Userdata is not bytes")
 
         self.metadata, self.userdata_raw, self.vendordata_raw = md, ud, vd
 
diff --git a/doc/rtd/reference/datasources/aeza.rst b/doc/rtd/reference/datasources/aeza.rst
index 767215594..430cf3fd2 100644
--- a/doc/rtd/reference/datasources/aeza.rst
+++ b/doc/rtd/reference/datasources/aeza.rst
@@ -21,6 +21,21 @@ Aeza's datasource can be configured as follows: ::
   Specifies the URL to retrieve the VPS meta-data. (optional)
   Default: ``http://77.221.156.49/v1/cloudinit/{id}/``
 
+* ``timeout``
+
+  The timeout value provided to ``urlopen`` for each individual http request.
+  This is used both when selecting a ``metadata_url`` and when crawling the
+  metadata service.
+
+  Default: 10
+
+* ``retries``
+
+  The number of retries that should be attempted for an http request. This
+  value is used only after ``metadata_url`` is selected.
+
+  Default: 5
+
 .. note::
    ``{id}`` in URLs is system-uuid DMI value.
 
diff --git a/tests/unittests/sources/test_aeza.py b/tests/unittests/sources/test_aeza.py
index 4f9e0363e..47d05675f 100644
--- a/tests/unittests/sources/test_aeza.py
+++ b/tests/unittests/sources/test_aeza.py
@@ -3,10 +3,14 @@
 # Author: Egor Ternovoy <cofob@riseup.net>
 #
 # This file is part of cloud-init. See LICENSE file for license information.
+import re
+
+import pytest
 
 from cloudinit import helpers, settings, util
-from cloudinit.sources import DataSourceAeza
-from tests.unittests.helpers import CiTestCase, mock
+from cloudinit.sources import DataSourceAeza as aeza
+from cloudinit.sources import InvalidMetaDataException
+from tests.unittests.helpers import mock
 
 METADATA = util.load_yaml(
     """---
@@ -25,42 +29,117 @@ runcmd:
 """
 
 
-class TestDataSourceAeza(CiTestCase):
-    """
-    Test reading the meta-data
-    """
+M_PATH = "cloudinit.sources.DataSourceAeza."
 
-    def setUp(self):
-        super(TestDataSourceAeza, self).setUp()
-        self.tmp = self.tmp_dir()
 
-    def get_ds(self):
-        distro = mock.MagicMock()
-        distro.get_tmp_exec_path = self.tmp_dir
-        ds = DataSourceAeza.DataSourceAeza(
-            settings.CFG_BUILTIN,
-            distro,
-            helpers.Paths({"run_dir": self.tmp}),
-        )
-        return ds
+class TestDataSourceAeza:
+    """Test Aeza.net reading instance-data"""
+
+    @pytest.mark.parametrize(
+        "system_manufacturer,expected",
+        (
+            pytest.param("Aeza", True, id="dmi_platform_match_aeza"),
+            pytest.param("aeza", False, id="dmi_platform_match_case_sensitve"),
+            pytest.param("Aezanope", False, id="dmi_platform_strict_match"),
+        ),
+    )
+    @mock.patch(f"{M_PATH}dmi.read_dmi_data")
+    def test_ds_detect(self, m_read_dmi_data, system_manufacturer, expected):
+        """Only strict case-senstiive match on DMI system-manfacturer Aeza"""
+        m_read_dmi_data.return_value = system_manufacturer
+        assert expected is aeza.DataSourceAeza.ds_detect()
 
+    @pytest.mark.parametrize(
+        "sys_cfg,expected_calls",
+        (
+            pytest.param(
+                {},
+                [
+                    mock.call(
+                        "http://77.221.156.49/v1/cloudinit/1dd9a779-uuid/%s",
+                        timeout=10,
+                        retries=5,
+                    )
+                ],
+                id="default_sysconfig_ds_url_retry_and_timeout",
+            ),
+            pytest.param(
+                {
+                    "datasource": {
+                        "Aeza": {
+                            "timeout": 7,
+                            "retries": 8,
+                            "metadata_url": "https://somethingelse/",
+                        }
+                    }
+                },
+                [
+                    mock.call(
+                        "https://somethingelse/%s",
+                        timeout=7,
+                        retries=8,
+                    )
+                ],
+                id="custom_sysconfig_ds_url_retry_and_timeout_overrides",
+            ),
+        ),
+    )
     @mock.patch("cloudinit.util.read_seeded")
-    @mock.patch("cloudinit.sources.DataSourceAeza.DataSourceAeza.ds_detect")
+    @mock.patch(f"{M_PATH}dmi.read_dmi_data")
     def test_read_data(
         self,
-        m_ds_detect,
+        m_read_dmi_data,
         m_read_seeded,
+        sys_cfg,
+        expected_calls,
+        paths,
+        tmpdir,
     ):
-        m_ds_detect.return_value = True
+        m_read_dmi_data.return_value = "1dd9a779-uuid"
         m_read_seeded.return_value = (METADATA, USERDATA, VENDORDATA)
 
-        with self.allow_subp(True):
-            ds = self.get_ds()
-        ret = ds.get_data()
-        self.assertTrue(ret)
+        ds = aeza.DataSourceAeza(
+            sys_cfg=sys_cfg, distro=mock.Mock(), paths=paths
+        )
+        with mock.patch.object(ds, "ds_detect", return_value=True):
+            assert True is ds.get_data()
+
+        assert m_read_seeded.call_args_list == expected_calls
+        assert ds.get_public_ssh_keys() == METADATA.get("public-keys")
+        assert isinstance(ds.get_public_ssh_keys(), list)
+        assert ds.userdata_raw == USERDATA
+        assert ds.vendordata_raw == VENDORDATA
+
+    @pytest.mark.parametrize(
+        "metadata,userdata,error_msg",
+        (
+            ({}, USERDATA, "Metadata does not contain instance-id: {}"),
+            (
+                None,
+                USERDATA,
+                "Failed to read metadata from http://77.221.156.49/v1/cloudinit/1dd9a779-uuid/%s",
+            ),
+            ({"instance-id": "yep"}, "non-bytes", "Userdata is not bytes"),
+        ),
+    )
+    @mock.patch("cloudinit.util.read_seeded")
+    @mock.patch(f"{M_PATH}dmi.read_dmi_data")
+    def test_not_detected_on_invalid_instance_data(
+        self,
+        m_read_dmi_data,
+        m_read_seeded,
+        metadata,
+        userdata,
+        error_msg,
+        paths,
+    ):
+        """Assert get_data returns False for unexpected format conditions."""
+        m_read_dmi_data.return_value = "1dd9a779-uuid"
+        m_read_seeded.return_value = (metadata, userdata, VENDORDATA)
 
-        self.assertTrue(m_read_seeded.called)
-        self.assertEqual(ds.get_public_ssh_keys(), METADATA.get("public-keys"))
-        self.assertIsInstance(ds.get_public_ssh_keys(), list)
-        self.assertEqual(ds.get_userdata_raw(), USERDATA)
-        self.assertEqual(ds.get_vendordata_raw(), VENDORDATA)
+        ds = aeza.DataSourceAeza(sys_cfg={}, distro=mock.Mock(), paths=paths)
+        with pytest.raises(
+            InvalidMetaDataException, match=re.escape(error_msg)
+        ):
+            with mock.patch.object(ds, "ds_detect", return_value=True):
+                ds.get_data()
diff --git a/tests/unittests/sources/test_common.py b/tests/unittests/sources/test_common.py
index 6a7b2d527..1e06b0efc 100644
--- a/tests/unittests/sources/test_common.py
+++ b/tests/unittests/sources/test_common.py
@@ -4,6 +4,7 @@ from unittest.mock import patch
 
 from cloudinit import importer, settings, sources, type_utils
 from cloudinit.sources import DataSource
+from cloudinit.sources import DataSourceAeza as Aeza
 from cloudinit.sources import DataSourceAkamai as Akamai
 from cloudinit.sources import DataSourceAliYun as AliYun
 from cloudinit.sources import DataSourceAltCloud as AltCloud
@@ -34,7 +35,6 @@ from cloudinit.sources import DataSourceUpCloud as UpCloud
 from cloudinit.sources import DataSourceVMware as VMware
 from cloudinit.sources import DataSourceVultr as Vultr
 from cloudinit.sources import DataSourceWSL as WSL
-from cloudinit.sources import DataSourceAeza as Aeza
 from tests.unittests import helpers as test_helpers
 
 DEFAULT_LOCAL = [
```

I can push this commit to your PR if you wish, and rebase against main if you like to ensure we minimally get the updated util.read_seeded changes, but I wanted to leave the decision to you as to what you would prefer to do? Do you want your platform to ignore network-config by default? If so, I can provide a separate PR if you like against tip of main to allow avoiding network-config requests from read_seeded that we could just rebase into this branch after review or we could incorporate that changeset into this branch.

What would you prefer?

blackboxsw:
@cofob if the proposed changes work for you, I think the final thing we'd need (once we agree on the approach for how to address `util.read_seeded`) would be to have you test this version of cloud-init in your environment and paste the following:
- captured output of `sudo DI_LOG=stderr /usr/lib/cloud-init/ds-identify --force` on a working system to ensure ds-identify is working well.
- Captured /var/log/cloud-init.log after running `sudo cloud-init init --local ` and `sudo cloud-init init`
- Captured output of `sudo cloud-init query --all`

blackboxsw:
Ping @cofob. I think this PR needs a `git rebase upstream/main` to ensure it has visibility to the incompatible changes in util.read_seeded which surfaces a 4-tuple return value.  Let us know how we should proceed with the previous change suggestions.

cloudinit/util.py
```diff
@@ -1069,42 +1072,48 @@ def read_seeded(base="", ext="", timeout=5, retries=10):
...
+        except url_helper.UrlError as e:
+            LOG.debug("No network config provided: %s", e)
+        else:
+            if network_resp.ok():
+                network = load_yaml(network_resp.contents)
+
     md = None
-    if md_resp.ok():
-        md = load_yaml(md_resp.contents, default={})
+    if "meta-data" in ignore_files:
```
blackboxsw, line 1089: This should be
```suggestion
    if "meta-data" not in ignore_files:
```

cloudinit/util.py
```diff
@@ -1069,42 +1072,48 @@ def read_seeded(base="", ext="", timeout=5, retries=10):
...
+        if md_resp.ok():
+            md = load_yaml(md_resp.contents, default={})
 
-    ud_resp = url_helper.read_file_or_url(
-        ud_url, timeout=timeout, retries=retries
-    )
     ud = None
-    if ud_resp.ok():
-        ud = ud_resp.contents
+    if "user-data" in ignore_files:
```
blackboxsw, line 1097: This should be:
```suggestion
    if "user-data" not in ignore_files:
```

cloudinit/util.py
```diff
@@ -1069,42 +1072,48 @@ def read_seeded(base="", ext="", timeout=5, retries=10):
...
-    try:
-        vd_resp = url_helper.read_file_or_url(
-            vd_url, timeout=timeout, retries=retries
-        )
-    except url_helper.UrlError as e:
-        LOG.debug("Error in vendor-data response: %s", e)
-    else:
-        if vd_resp.ok():
-            vd = vd_resp.contents
+    if "vendor-data" in ignore_files:
```
blackboxsw, line 1105: ditto
```suggestion
    if "vendor-data" not in ignore_files:
```

cloudinit/sources/DataSourceAeza.py
```diff
@@ -0,0 +1,69 @@
...
+        if md is None:
+            raise sources.InvalidMetaDataException(
+                f"Failed to read metadata from {metadata_address}",
+            )
+        if not isinstance(md.get("instance-id"), str):
+            raise sources.InvalidMetaDataException(
+                f"Metadata does not contain instance-id: {md}"
+            )
+        if not isinstance(ud, bytes):
+            raise sources.InvalidMetaDataException("Userdata is not bytes")
```
blackboxsw, line 56: I'm not sure we really expect bytes here from util.read_seeded for ud. [self.userdata_raw certainly expects str](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/__init__.py#L322). I think we can drop this check here as this is also not needed for [DataSourceNoCloudNet when it reaches out to a remote URL for user-data files](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceNoCloud.py#L193-L200). 
```suggestion
```

blackboxsw (commented):
Looks great!
Minor diff remaining to fix mypy, doc-check and ruff and dropped that extra unittest.

And lastly, can we get [output of a successful run of the suggested commands?](https://github.com/canonical/cloud-init/pull/5299#issuecomment-2292588823)

```diff
diff --git a/cloudinit/sources/DataSourceAeza.py b/cloudinit/sources/DataSourceAeza.py
index d55dc16e0..40917efb3 100644
--- a/cloudinit/sources/DataSourceAeza.py
+++ b/cloudinit/sources/DataSourceAeza.py
@@ -52,9 +52,6 @@ class DataSourceAeza(sources.DataSource):
             raise sources.InvalidMetaDataException(
                 f"Metadata does not contain instance-id: {md}"
             )
-        if not isinstance(ud, bytes):
-            raise sources.InvalidMetaDataException("Userdata is not bytes")
-
         self.metadata, self.userdata_raw, self.vendordata_raw = md, ud, vd
 
         return True
diff --git a/doc/rtd/reference/datasources/aeza.rst b/doc/rtd/reference/datasources/aeza.rst
index 430cf3fd2..3b5b452a7 100644
--- a/doc/rtd/reference/datasources/aeza.rst
+++ b/doc/rtd/reference/datasources/aeza.rst
@@ -37,6 +37,6 @@ Aeza's datasource can be configured as follows: ::
   Default: 5
 
 .. note::
-   ``{id}`` in URLs is system-uuid DMI value.
+   ``{id}`` in URLs is ``system-uuid`` DMI value.
 
 .. _Aeza: https://wiki.aeza.net/cloud-init
diff --git a/tests/unittests/sources/test_aeza.py b/tests/unittests/sources/test_aeza.py
index a07cf43c3..0374e168a 100644
--- a/tests/unittests/sources/test_aeza.py
+++ b/tests/unittests/sources/test_aeza.py
@@ -7,7 +7,7 @@ import re
 
 import pytest
 
-from cloudinit import helpers, settings, util
+from cloudinit import util
 from cloudinit.sources import DataSourceAeza as aeza
 from cloudinit.sources import InvalidMetaDataException
 from tests.unittests.helpers import mock
@@ -121,7 +121,6 @@ class TestDataSourceAeza:
                 USERDATA,
                 "Failed to read metadata from http://77.221.156.49/v1/cloudinit/1dd9a779-uuid/%s",
             ),
-            ({"instance-id": "yep"}, "non-bytes", "Userdata is not bytes"),
         ),
     )
     @mock.patch("cloudinit.util.read_seeded")
```

Issue #5305, opened 2024-05-20
[enhancement]: Deal with empty ssh key files


ani-sinha:
# Enhancement
Start an instance running centos on aws t2.large system, after reboot system via sysrq 'b', the system is not accessible via ssh

```
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: sshd.service: Service RestartSec=42s expired, scheduling restart.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: sshd.service: Scheduled restart job, restart counter is at 14.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: Stopped OpenSSH server daemon.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: Stopped target sshd-keygen.target.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: Stopping sshd-keygen.target.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: Reached target sshd-keygen.target.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: Starting OpenSSH server daemon...
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key "/etc/ssh/ssh_host_rsa_key": invalid format
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key: /etc/ssh/ssh_host_rsa_key
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key "/etc/ssh/ssh_host_ecdsa_key": invalid format
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key: /etc/ssh/ssh_host_ecdsa_key
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key "/etc/ssh/ssh_host_ed25519_key": invalid format
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: Unable to load host key: /etc/ssh/ssh_host_ed25519_key
Feb 2 09:22:47 ip-10-22-1-137 sshd[1454]: sshd: no hostkeys available – exiting.
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: sshd.service: Main process exited, code=exited, status=1/FAILURE
Feb 2 09:22:47 ip-10-22-1-137 systemd[1]: sshd.service: Failed with result 'exit-code'.
[...]

PR #5306, opened 2024-05-20
Handle zero byte ssh key files

ani-sinha:
## Proposed Commit Message
Sometimes, due to file system corruption from a linux kernel crash or for some other reasons, the system may be left with zero byte ssh keyfiles. These zero byte files are of no use and generate errors like the following upon next reboot:
sshd[1454]: Unable to load host key: /etc/ssh/ssh_host_rsa_key sshd[1454]: Unable to load host key "/etc/ssh/ssh_host_ecdsa_key": invalid format

Therefore, we should check for the presence of zero-byte files and if found delete and regenerate the key files in order to recover from the error.

Fixes: GH-5305

## Merge type

### DIFF
--- a/cloudinit/config/cc_ssh.py
+++ b/cloudinit/config/cc_ssh.py
@@ -264,9 +264,20 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
 
         for keytype in key_names:
             keyfile = KEY_FILE_TPL % (keytype)
-            if os.path.exists(keyfile):
+            if os.path.exists(keyfile) and os.path.getsize(keyfile) != 0:
+                # no need to generate keys if the keyfile exists and
+                # is not an empty file.
                 continue
             util.ensure_dir(os.path.dirname(keyfile))
+            if os.path.exists(keyfile) and os.path.getsize(keyfile) == 0:
+                # remove empty ssh key files.
+                try:
+                    util.del_file(keyfile)
+                except Exception:
+                    util.logexc(
+                        LOG, "Failed deleting empty key file %s", keyfile
+                    )
+
             cmd = ["ssh-keygen", "-t", keytype, "-N", "", "-f", keyfile]
 
             # TODO(harlowja): Is this guard needed?
--- a/tests/unittests/config/test_cc_ssh.py
+++ b/tests/unittests/config/test_cc_ssh.py
@@ -103,9 +103,17 @@ class TestHandleSsh:
     @mock.patch(MODPATH + "glob.glob")
     @mock.patch(MODPATH + "ug_util.normalize_users_groups")
     @mock.patch(MODPATH + "os.path.exists")
+    @mock.patch(MODPATH + "os.path.getsize")
     @mock.patch(MODPATH + "util.fips_enabled")
     def test_handle_no_cfg(
-        self, m_fips, m_path_exists, m_nug, m_glob, m_setup_keys, fips_enabled
+        self,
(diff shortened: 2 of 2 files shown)


TheRealFalcon:
@ani-sinha , thanks for the PR, but I don't think that this will work. The ssh module runs once per instance, meaning after first boot it won't run again unless you've launched a new instance (or if your filesystem corruption also didn't write the semaphore file).

More generally, I'm not sure cloud-init should be concerned with trying to recover from filesystem corruption. If corruption broke your instance, then you have a broken instance. Why would we want to paper over that? In the case of SSH host keys, if one set of keys were generated (but not written to disk yet), then a new set of keys next boot would incorrectly indicate that there's a security issue. We don't want to signal that either.

From the [bug](https://github.com/canonical/cloud-init/issues/5305) you filed:
```
after reboot system via sysrq 'b', the system is not accessible via ssh
```

`sysrq b` is a bit extreme. Can you instead find a way to reboot that ensures that your filesystems get unmounted cleanly first? This sounds like a problem with the test harness, not cloud-init.

ani-sinha:
> More generally, I'm not sure cloud-init should be concerned with trying to recover from filesystem corruption. If corruption broke your instance, then you have a broken instance. Why would we want to paper over that? In the case of SSH host keys, if one set of keys were generated (but not written to disk yet), then a new set of keys next boot would incorrectly indicate that there's a security issue. We don't want to signal that either.

I hear you and I agree with your points. However, I was wondering what would be the difference between no certs and zero-byte certs? Could we treat them in the same category and generate new certs in those cases? 
Another thing I was thinking is, maybe cloud-init could be made more resilient in addressing this case. 

>sysrq b is a bit extreme. Can you instead find a way to reboot that ensures that your filesystems get unmounted cleanly first? This sounds like a problem with the test harness, not cloud-init.

That is just a test case scenario. Actual scenario would be a kernel crash for example due to a bug or for some other reason. When that happens, buffer cache is not sync'd and we end up with zero byte files.
The question then is, is there anything cloud-init can do to recover from this situation? Can we give users some option to regenerate the certs and get going?

TheRealFalcon:
> However, I was wondering what would be the difference between no certs and zero-byte certs? Could we treat them in the same category and generate new certs in those cases?

We still have the problem that the SSH module doesn't run on 2nd boot. Without some new machinery like a new module, we have no way of even checking these files on the next boot.

> Another thing I was thinking is, maybe cloud-init could be made more resilient in addressing this case.
Actual scenario would be a kernel crash for example due to a bug or for some other reason. When that happens, buffer cache is not sync'd and we end up with zero byte files.

If your kernel is crashing within 30 seconds of first boot, I'm not sure why regenerating SSH host keys would be a primary concern. This still feels like a very contrived scenario to me and quite literally any file written by cloud-init (or any other process) could have the same problem. Why are we focusing on host keys? If your concern is troubleshooting or debugging, the serial console should still be usable.

Is there something broader that you're trying to accomplish here or a use case that I'm not understanding?

ani-sinha:
> If your kernel is crashing within 30 seconds of first boot, I'm not sure why regenerating SSH host keys would be a primary concern

Yes I think that if the kernel crashed leaving a bunch of corrupted files around, there are lot of other things that may not function properly, not just ssh or cloud-init. It would be a system wide issue and not specific to cloud-init. Making cloud-init resilient may not make the system usable depending on the extent of the corruption. So  I will close this PR since it seems to me that this problem is not worth fixing in cloud-init.

Issue #5300, opened 2024-05-17
[DataSourceLXD] Wrongly decodes response containing special char


phvalguima:
# Bug report

Running:
Ubuntu 22.04
LXD 
LXC created with 22.04 image

Juju is passing snap-assertions as part of the cloud-init user-data whenever using snap-store-proxy. However, some of these assertions contain a ` display-name` which points to the name a given user registered in the Charmhub / Snap store upstream.

In my case, I registered: "Guimarães", which contains an "ã". That gets wrongly decoded by DataSourceLXD class as it leaves to `requests` to decide the encoding, and concludes on `ISO-8859-1` instead of  `UTF-8` using [chardet](https://github.com/psf/requests/blob/9a40d1277807f0a4f26c9a37eea8ec90faa8aadc/src/requests/compat.py#L30), called by [apparent_encoding](https://github.com/psf/requests/blob/9a40d1277807f0a4f26c9a37eea8ec90faa8aadc/src/requests/models.py#L790-L797).

Using `UTF-8` as encoding resolves the problem.

```
(Pdb) l
377  	            cfg_key = config_route.rpartition("/")[-1]
378  	            # Leave raw data values/format unchanged to represent it in
379  	            # instance-data.json for cloud-init query or jinja template
380  	            # use.
381  ->	            config["config"][cfg_key] = config_route_response.text
382  	            # Promote common CONFIG_KEY_ALIASES to top-level keys.
383  	            if cfg_key in CONFIG_KEY_ALIASES:
384  	                # Due to sort of config_routes, promote cloud-init.*
385  	                # aliases before user.*. This allows user.* keys to act as
386  	                # fallback config on old LXD, with new cloud-init images.

(Pdb) p config_route_response.encoding
None
(Pdb) p config_route_response.apparent_encoding
'ISO-8859-1'
(Pdb) bt
  /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceLXD.py(480)<module>()
-> atomic_helper.json_dumps(read_metadata(metadata_keys=MetaDataKeys.ALL))
  /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceLXD.py(454)read_metadata()
-> return _MetaDataReader(api_version=api_version)(
  /usr/lib/python3/dist-packages/cloudinit/sources/DataSourceLXD.py(410)__call__()
-> md.update(self._process_config(session))
[...]

PR #5309, opened 2024-05-20
fix(lxd): Properly handle unicode from LXD socket

TheRealFalcon:
## Proposed Commit Message

```
fix(lxd): Properly handle unicode from LXD socket

Fixes GH-5300
```

## Additional Context
See #5300

## Merge type

### DIFF
--- a/tests/integration_tests/modules/test_combined.py
+++ b/tests/integration_tests/modules/test_combined.py
@@ -511,6 +511,7 @@ class TestCombined:
         client = class_client
         assert "💩" == client.read_from_file("/var/tmp/unicode_data")
 
+
 @pytest.mark.user_data(USER_DATA)
 class TestCombinedNoCI:
     @retry(tries=30, delay=1)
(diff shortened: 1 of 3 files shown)


blackboxsw (commented):
functionally looks good. 
I think we are missing correct mocks here in tests/unittests/sources/test_lxd.py::TestReadMetadata::test_read_metadata_handles_unexpected_content_or_http_status as we are failing on our mock here in py 3.11 due to our mocking of response.text instead of response.content.  May need to update that unittest to avoid the 8 failures there.

TheRealFalcon:
Doh, I missed the unit test failures. Thanks! Should be updated now.

tests/integration_tests/modules/test_combined.py
```diff
@@ -506,6 +507,10 @@ def test_instance_cloud_id_across_reboot(
         assert client.execute(f"test -f /run/cloud-init/{cloud_file}").ok
         assert client.execute("test -f /run/cloud-init/cloud-id").ok
 
+    def test_unicode(self, class_client: IntegrationInstance):
+        client = class_client
+        assert "💩" == client.read_from_file("/var/tmp/unicode_data")
```
blackboxsw, line 512: And, who says we don't know how to have fun in this project?

blackboxsw (approved):
Bring it home!

PR #5321, opened 2024-05-23
fix: Check renderer for netplan-specific code

TheRealFalcon:
Draft PR as this has not been tested in a real environment yet.

## Proposed Commit Message

```
fix: Check renderer for netplan-specific code

DataSourceEc2 has netplan-specific code that was gated by the
selected activator. However, in practice, the activators are rarely
used. Somebody may have selected a non-netplan renderer without
selecting the proper activator with it. Since this involves rendering
configuration, we should gate on the renderer instead.

Fixes GH-5318
```

## Merge type

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -349,7 +349,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         except activators.NoActivatorException:
             return None
 
-    def _get_renderer(self) -> Renderer:
+    @property
+    def network_renderer(self) -> Renderer:
         priority = util.get_cfg_by_path(
             self._cfg, ("network", "renderers"), None
         )
@@ -442,7 +443,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
         Returns True if any devices failed to come up, otherwise False.
         """
-        renderer = self._get_renderer()
+        renderer = self.network_renderer
         network_state = parse_net_config_data(netconfig, renderer=renderer)
         self._write_network_state(network_state, renderer)
 
--- a/cloudinit/sources/DataSourceEc2.py
+++ b/cloudinit/sources/DataSourceEc2.py
@@ -21,7 +21,7 @@ from cloudinit import url_helper as uhelp
 from cloudinit import util, warnings
 from cloudinit.distros import Distro
 from cloudinit.event import EventScope, EventType
-from cloudinit.net import activators
+from cloudinit.net import netplan
 from cloudinit.net.dhcp import NoDHCPLeaseError
 from cloudinit.net.ephemeral import EphemeralIPNetwork
 from cloudinit.sources import NicOrder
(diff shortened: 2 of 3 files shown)


blackboxsw (commented):
This is the correct approach given that our . I wonder if it's worth use validating if in network that the selected `convert_ec2_metadata_network_config` is actually trying to generate the network config we are attempting to render, not the network config we are attempting to "activate" anyway so cloud-init should align based on renderer instead of activator anyway.

I wonder if it's worth a separate issue to track and warn if the selected network `activator not in (distro.network_renderer, None)`. But, I don't think that is warranted in this fix.

TheRealFalcon:
>  I wonder if it's worth use validating if in network that the selected convert_ec2_metadata_network_config is actually trying to generate the network config we are attempting to render, not the network config we are attempting to "activate" anyway so cloud-init should align based on renderer instead of activator anyway.

I'm not sure what you're wanting to validate here.

> warn if the selected network activator not in (distro.network_renderer, None). But, I don't think that is warranted in this fix.

This isn't super straightforward as it's not always a 1:1 mapping. I.e., sysconfig and network-manager renderer can both use the network-manager activator. netplan renderer could techincally use the networkd and network-manager activators (but I hope nobody is doing this), ifconfig or ifupdown activators could both be used with eni renderer. I think it's good to keep them decoupled.

TheRealFalcon:
I'm marking this ready for review. The original tests used activators as an assumption, so they've been updated to use renderer instead.

I was able to reproduce the original issue manually, then verified that the fix here prevents the traceback and renders the correct ENI code.

blackboxsw:
> > I wonder if it's worth us validating if the network convert_ec2_metadata_network_config is actually trying to generate the network config we are attempting to render, not the network config we are attempting to "activate" anyway so cloud-init should align based on renderer instead of activator anyway.
> 
> I'm not sure what you're wanting to validate here.
> 

Sorry that was a totally unintelligible comment.  I was curious whether we wanted to provide other unittest conditions where we change the mocked distro.renderer to something other than netplan.Render and assert the dict values returned in those non-netplan environments eni-prioritized environments.


> > warn if the selected network activator not in (distro.network_renderer, None). But, I don't think that is warranted in this fix.
> 
> This isn't super straightforward as it's not always a 1:1 mapping. I.e., sysconfig and network-manager renderer can both use the network-manager activator. netplan renderer could techincally use the networkd and network-manager activators (but I hope nobody is doing this), ifconfig or ifupdown activators could both be used with eni renderer. I think it's good to keep them decoupled.

makes sense, nothing for this PR in either case. It would be good to expand our unittest coverage at some point to eni-based renderers for ec2, but not necessary for this PR. 

I also validated failure mode on multi-nic scenarios with only renderers: [ eni, netplan] and ifupdown installed and validated your propsed fix resolves that config inconsistency and avoids the Traceback during network render stage.

blackboxsw (approved):
Works for overridden network:renderers. Validated prior failure on multi-nic instance and the rendered /etc/network/interfaces.d/50-cloud-init.cfg after your proposed PR. The result is renders the correct net config on multi-nic environment and avoids the traceback seen in 24.1. We should spend some time extending our ENI support validation to ensure we don't regress this in the future, for non-netplan environments, but as-is this resolves the issue at hand.

PR #5322, opened 2024-05-23
docs: Renderers/Activators have downstream overrides

TheRealFalcon:
## Proposed Commit Message

```
docs: Renderers/Activators have downstream overrides
```

## Merge type

### DIFF
--- a/doc/rtd/reference/network-config.rst
+++ b/doc/rtd/reference/network-config.rst
@@ -223,6 +223,11 @@ for BSD flavors at the moment.
 Network output policy
 =====================
 
+.. note::
+
+   These are **upstream** defaults and are known to be overridden by
+   downstream distributions.
+
 The default policy for selecting a network ``renderer`` (in order of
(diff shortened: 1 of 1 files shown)


blackboxsw (approved):

doc/rtd/reference/network-config.rst
```diff
@@ -223,6 +223,11 @@ for BSD flavors at the moment.
 Network output policy
 =====================
 
+.. note::
+
+   These are **upstream** defaults and are known to be overridden by
```
blackboxsw, line 228: Do we want to cross link to [cloudinit/net/renderers.py:2](https://github.com/canonical/cloud-init/blob/main/cloudinit/net/renderers.py#L29) here? It may diverge over time, so I'm on the fence about that

TheRealFalcon: I don't think so...especially since it's configurable in cloud.cfg

Issue #3950, opened 2023-05-12
cloud init failed to set up link for lo


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1961355](https://bugs.launchpad.net/cloud-init/+bug/1961355)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2022-02-18T03:57:46.793614+00:00
date_fix_committed = None
date_fix_released = None
id = 1961355
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1961355
milestone = None
owner = zhangjl02
owner_name = Jinsheng Zhang
private = False
status = triaged
submitter = zhangjl02
submitter_name = Jinsheng Zhang
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Jinsheng Zhang(zhangjl02)** wrote on 2022-02-18T03:57:46.793614+00:00_

A error message was found in '/var/lib/cloud/data/result.json' when booting a Ubuntu20.04 server.
Using cloud-init 18.5
ERROR INFO:
-------------ERROR INFO--------
cat /var/lib/cloud/data/result.json
{
 "v1": {
  "datasource": "DataSourceConfigDrive [net,ver=2][source=/dev/vdb]",
  "errors": [
   "Unexpected error while running command.\nCommand: ['udevadm', 'test-builtin', 'net_setup_link', '/sys/class/net/lo']\nExit code: 1\nReason: -\nStdout: \nStderr: Load module index\n        Network interface NamePolicy= disabled on kernel command line, ignoring.\n        Skipping empty file: /etc/systemd/network/99-default.link\n        Parsed configuration file /usr/lib/systemd/network/73-usb-net-by-mac.link\n        Parsed configuration file /run/systemd/network/10-netplan-eth0.link\n        Created link configuration context.\n        Builtin command 'net_setup_link' fails: No such file or directory\n        Unload module index\n        Unloaded link configuration context."
  ]
 }
}

cloudinit.util.ProcessExecutionError: Unexpected error while running command.
Command: ['udevadm', 'test-builtin', 'net_setup_link', '/sys/class/net/lo']
Exit code: 1
Reason: -
Stdout: 
Stderr: Load module index
        Network interface NamePolicy= disabled on kernel command line, ignoring.
        Skipping empty file: /etc/systemd/network/99-default.link
[...]

PR #5339, opened 2024-05-28
fix(netplan): Fix predictable interface rename issue

holmanb:
## Proposed Commit Message
```
fix(netplan): Fix predictable interface rename issue

When predictable naming is disabled, the following command may exit with
a non-zero exit code.

    udevadm test-builtin net_setup_link

This code only ran to check for udev rename races, which cannot happen
when systemd renaming is disabled. Skip when disabled.

Fixes GH-3950
```

## Additional Context
https://github.com/canonical/cloud-init/issues/3950


TheRealFalcon (commented):
Code LGTM, but you need to fix a broken test. Likely a new mock is needed.

PR #5350, opened 2024-06-01
Various mypy fixes

holmanb:
## Additional Context
See individual commits.

This work enables using more of mypy's static analysis features.

This PR does various cleanups including:

- remove unnecessary type ignore comments
- add missing type annotations
- remove invalid type annotations
- minor refactors to eliminate the need for type annotations
- eliminate unused code
- auto-format network jsonschema
- tox: fix minor issues
- tox: add stub libraries

Note to reviewers:

Most of the changes here are no-ops. That said, please pay special attention to the commit titled `refactor(typing): Remove unused code paths`. Invalid type annotations can lead mypy to warn incorrectly that a code path is unused (this is how I found many of the type errors in this commit series). I do believe that I was thorough in verifying this commit, but this one carries the most risk of the series.

### DIFF
--- a/cloudinit/config/cc_power_state_change.py
+++ b/cloudinit/config/cc_power_state_change.py
@@ -78,8 +78,8 @@ def check_condition(cond):
 
 def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
     try:
-        (args, timeout, condition) = load_power_state(cfg, cloud.distro)
-        if args is None:
+        (arg_list, timeout, condition) = load_power_state(cfg, cloud.distro)
+        if arg_list is None:
             LOG.debug("no power_state provided. doing nothing")
             return
     except Exception as e:
@@ -99,7 +99,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
 
     devnull_fp = open(os.devnull, "w")
 
-    LOG.debug("After pid %s ends, will execute: %s", mypid, " ".join(args))
+    LOG.debug("After pid %s ends, will execute: %s", mypid, " ".join(arg_list))
 
     util.fork_cb(
         run_after_pid_gone,
@@ -108,7 +108,7 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
         timeout,
         condition,
         execmd,
-        [args, devnull_fp],
+        [arg_list, devnull_fp],
     )
 
 
--- a/cloudinit/config/schemas/schema-network-config-v2.json
+++ b/cloudinit/config/schemas/schema-network-config-v2.json
@@ -10,43 +10,43 @@
       ]
     },
     "dhcp-overrides": {
-        "type": "object",
-        "description": "DHCP behaviour overrides. Overrides will only have an effect if the corresponding DHCP type is enabled.",
-        "additionalProperties": false,
-        "properties": {
-          "hostname": {
-            "type": "string",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          },
-          "route-metric": {
-            "type": "integer",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          },
-          "send-hostname": {
-            "type": "boolean",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          },
-          "use-dns": {
-            "type": "boolean"
-          },
-          "use-domains": {
-            "type": "string"
-          },
-          "use-hostname": {
-            "type": "boolean"
-          },
-          "use-mtu": {
-            "type": "boolean",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          },
-          "use-ntp": {
-            "type": "boolean"
-          },
-          "use-routes": {
-            "type": "boolean",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          }
+      "type": "object",
+      "description": "DHCP behaviour overrides. Overrides will only have an effect if the corresponding DHCP type is enabled.",
+      "additionalProperties": false,
+      "properties": {
+        "hostname": {
+          "type": "string",
+          "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
+        },
+        "route-metric": {
+          "type": "integer",
+          "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
+        },
+        "send-hostname": {
+          "type": "boolean",
+          "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
+        },
+        "use-dns": {
+          "type": "boolean"
+        },
+        "use-domains": {
+          "type": "string"
+        },
+        "use-hostname": {
+          "type": "boolean"
+        },
+        "use-mtu": {
+          "type": "boolean",
+          "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
+        },
+        "use-ntp": {
+          "type": "boolean"
+        },
+        "use-routes": {
+          "type": "boolean",
+          "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
         }
+      }
     },
     "gateway": {
       "type": "string",
@@ -56,17 +56,33 @@
       "type": "object",
       "properties": {
         "renderer": {
-            "$ref": "#/$defs/renderer"
+          "$ref": "#/$defs/renderer"
         },
         "dhcp4": {
-          "type": ["boolean", "string"],
+          "type": [
+            "boolean",
+            "string"
+          ],
           "description": "Enable DHCP for IPv4. Off by default.",
-          "enum": ["yes", "no", true, false]
+          "enum": [
+            "yes",
+            "no",
+            true,
+            false
+          ]
         },
         "dhcp6": {
-          "type": ["boolean", "string"],
+          "type": [
+            "boolean",
+            "string"
+          ],
           "description": "Enable DHCP for IPv6. Off by default.",
-          "enum": ["yes", "no", true, false]
+          "enum": [
+            "yes",
+            "no",
+            true,
+            false
+          ]
         },
         "dhcp4-overrides": {
           "$ref": "#/$defs/dhcp-overrides"
@@ -89,7 +105,7 @@
         },
         "mtu": {
           "type": "integer",
-          "description": "The MTU key represents a device’s Maximum Transmission Unit, the largest size packet or frame, specified in octets (eight-bit bytes), that can be sent in a packet- or frame-based network. Specifying mtu is optional."
+          "description": "The MTU key represents a device\u2019s Maximum Transmission Unit, the largest size packet or frame, specified in octets (eight-bit bytes), that can be sent in a packet- or frame-based network. Specifying mtu is optional."
         },
         "nameservers": {
           "type": "object",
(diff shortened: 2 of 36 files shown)


TheRealFalcon (commented):
I don't have time to do a full review today, but I realized I have a stale "Pending" comment that is still relevant.

cloudinit/config/cc_power_state_change.py
```diff
@@ -124,7 +124,7 @@ def check_condition(cond):
         return False
 
 
-def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
+def handle(name: str, cfg: Config, cloud: Cloud, _) -> None:
```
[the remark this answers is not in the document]
holmanb: > This is part of every single handler...

This is the only one that has an argument name which later has a different type. If you'd rather we didn't fix it this way, then I'll change the shadowed variable name.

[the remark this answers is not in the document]
TheRealFalcon: Ah, I see now. I thought you were doing something else. Yeah, I'm ok changing this one here. I think I'd still prefer the name change though.

cloudinit/net/netplan.py
```diff
@@ -482,9 +479,18 @@ def _render_content(self, network_state: NetworkState) -> str:
...
             elif if_type == "bridge":
                 # required_keys = ['name', 'bridge_ports']
+                #
+                # Rather than raise an exception on `sorted(None)`, log a
+                # warning and skip this interface when invalid configuration is
+                # received.
                 bridge_ports = ifcfg.get("bridge_ports")
-                # mypy wrong error. `copy(None)` is supported:
-                ports = sorted(copy.copy(bridge_ports))  # type: ignore
+                if bridge_ports is None:
```
TheRealFalcon, line 487: Is there a reason we're downgrading to a warning here? I don't think we should be changing unrelated functionality in this kind of PR.

cloudinit/util.py
```diff
@@ -3231,8 +3229,8 @@ def deprecate(
 
     Note: uses keyword-only arguments to improve legibility
     """
-    if not hasattr(deprecate, "_log"):
-        deprecate._log = set()  # type: ignore
+    if not deprecate.__dict__.get("log"):
```
TheRealFalcon, line 3232: [modifying __dict__ directly is not recommended](https://docs.python.org/3/library/stdtypes.html#modules). Why not `getattr` and `setattr` here, further down in this file, and in conftest.py?

cloudinit/config/schemas/schema-network-config-v2.json
```diff
@@ -10,43 +10,43 @@
...
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          },
-          "use-ntp": {
-            "type": "boolean"
-          },
-          "use-routes": {
-            "type": "boolean",
-            "description": "Unsupported  for dhcp6-overrides when used with the networkd renderer."
-          }
+      "type": "object",
```
TheRealFalcon, line 13: Don't need to change in this PR, but separate functionality should really go in separate PRs. Multiple commits make sense in a PR when the functionality in a later commit requires the change from the earlier commit, but lets try to keep PRs are small and focused as possible.

TheRealFalcon (changes requested):
Comments inline

cloudinit/net/netplan.py
holmanb: This was incorrectly silencing a mypy error which is valid.

```python
>>> sorted(copy.copy(None))
TypeError: 'NoneType' object is not iterable
```

holmanb: The current code's comment is incorrect, and I suspect we could handle this situation more gracefully.

The current code prevents iteration from continuing. The intent here was to allow iteration to continue while also logging a warning, which I think is a more graceful way to handle invalid network state than throwing an exception which prevents the code from iterating through other interfaces.

setup_utils.py
```diff
@@ -35,15 +35,6 @@ def pkg_config_read(library: str, var: str) -> str:
     return path
 
 
-def in_virtualenv() -> bool:
```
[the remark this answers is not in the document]
holmanb: Let me know what needs to be added to the commit message:
```
  commit 5b0d2fa1bf714d7a952a27a59f1ce070159ff352
  Author: Brett Holman <brett.holman@canonical.com>
  Date:   Fri May 31 12:56:50 2024 -0600
  
      ci(mypy): Set default follow_imports value (#5350)
      
      Remove code for detecting a virtualenv. This has been broken since ~3.7.
  
   pyproject.toml |  1 -
   setup.py       | 17 +++++++----------
   setup_utils.py |  9 ---------
   3 files changed, 7 insertions(+), 20 deletions(-)
```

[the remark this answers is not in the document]
holmanb: Mypy correctly points out that this isn't used anywhere on 3.12. Checking for the presence of a virtualenv hasn't worked since Python 3.7, and since nobody has bothered to make it work on >3.7, the assumption is that nobody cares about it working on 3.7 either. This appears to be a relic of the past.

Other factors: 

- We planned to drop support for 3.7 a year ago. We held out for a few stragglers that still wanted support, but I think it is time for 3.7 to go.
- Lots of disruption is coming in the packaging space in the upcoming cycle for PEP517. This is just packaging code. If someone still really wants this, they can maintain it themselves.

Aside: Earlier this year when I was trying to work on PEP517 support in my spare time, this code was actually a stumbling block due to the added complexity. Culling unnecessary junk prior to a big refactor or change in functionality helps a lot in the long run to understand what we actually need.

cloudinit/config/schemas/schema-network-config-v2.json
holmanb: +1, my intention with this was to make this part of the mypy + jsonschema type generation stuff, for which this makes more sense, but without I guess it feels like a bolt-on PR without that work.

cloudinit/url_helper.py
```diff
@@ -278,8 +278,6 @@ def __init__(self, response: requests.Response):
 
     @property
     def contents(self) -> bytes:
-        if self._response.content is None:
```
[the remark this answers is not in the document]
holmanb: I'll need to dig in further about this one

setup_utils.py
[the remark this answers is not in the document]
TheRealFalcon: > Mypy correctly points out that this isn't used anywhere on 3.12

Didn't realize that. I don't disagree with anything you said. I knew this needed to go; I just didn't see the relation to this PR.

> 3.7

Let's start with 3.6 :wink: 

According to our original proposed calendar, we should still be supporting 3.7 until the end of the month.

cloudinit/net/netplan.py
TheRealFalcon: +1. At first I thought we would want to throw an exception if we have issues here as to not render invalid config, but warning makes sense.

cloudinit/url_helper.py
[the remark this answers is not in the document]
holmanb: @TheRealFalcon Thanks for finding that. It turns out that this was due to a bug in typeshed. I just [submitted a PR to fix it](https://github.com/python/typeshed/pull/12180). I'll change this PR to silence this warning until a fix is released.

holmanb:
@TheRealFalcon I think I've addressed all comments. If tests pass, I think this is ready for review.

cloudinit/util.py
```diff
@@ -3231,8 +3229,8 @@ def deprecate(
 
     Note: uses keyword-only arguments to improve legibility
     """
-    if not hasattr(deprecate, "_log"):
-        deprecate._log = set()  # type: ignore
+    if not hasattr(deprecate, "log"):
```
TheRealFalcon, line 3232: Looking at this again, I don't think we want this (or the test updates based on it).

See the first comment on https://stackoverflow.com/a/16955098/37205

.git-blame-ignore-revs
```diff
@@ -2,3 +2,7 @@
 
 # Apply black and isort formatting
 bae9b11da9ed7dd0b16fe5adeaf4774b7cc628cf
+
+# Apply json formatting
+04c82eafd8fe34c5f02ffecfdb5dba888f3baa3f
```
TheRealFalcon, line 7: These changed again. Think it might make sense to update these after the fact?

tox.ini
```diff
@@ -78,7 +82,9 @@ deps =
     types-requests=={[format_deps]types-requests}
     types-setuptools=={[format_deps]types-setuptools}
     typing-extensions=={[format_deps]typing-extensions}
-commands = {envpython} -m mypy cloudinit/ tests/ tools/
+# TODO: add --check-untyped-defs to allow checking untyped function bodies
```
TheRealFalcon, line 85: This is done as of https://github.com/canonical/cloud-init/pull/5409 , correct?

TheRealFalcon (changes requested):
I left a few more inline. ~~The logging one is the only major one~~ (deleted the comment).

tox.ini
holmanb: Yes!

holmanb:
Looks like [the recent addition](355560312461caab01e914bb2424c6cd8268c153) to mypy features expanded the scope slightly, so I had to update some test code to make it pass. I think I just need to fix the merge conflict before this is ready to merge @TheRealFalcon.

tests/integration_tests/instances.py
```diff
@@ -60,7 +59,7 @@ class IntegrationInstance:
     def __init__(
         self,
         cloud: "IntegrationCloud",
-        instance: BaseInstance,
+        instance,
```
TheRealFalcon, line 62: This was fairly useful when it comes to editor autocomplete. Why does mypy not like it?

tests/integration_tests/instances.py
```diff
@@ -283,20 +282,6 @@ def _apt_update(self):
                     f"Failed to update packages: {response.stderr}"
                 )
 
-    def ip(self) -> str:
```
TheRealFalcon, line 286: This is used in `__exit__` (though the implementation is buggy). Why is it being removed?

TheRealFalcon (commented):
2 questions inline

tests/integration_tests/instances.py
holmanb: The whole purpose of that was to to make it easier for someone to ssh into an instance on a public cloud after a test, however since only lxd has `execute_via_ssh`, its purpose is defeated. I've been meaning to clean this up for a while, but hadn't gotten to it yet. Mypy is complaining about the types in this code which is why I finally made the change.

TheRealFalcon: So can we fix it instead of dropping it? It's valid if we don't want to fix it here, but we can use the `# typing: ignore` directives in that case.

holmanb: Sure, that's fair. Looking at this again I think the fix might be easier than I previously thought.

holmanb:
@TheRealFalcon Just force pushed to fix the merge conflict, and while I was at it I squashed the fixup commits and updated commit messages with PR number. Assuming tests pass this one is ready for re-review.

tests/unittests/conftest.py
```diff
@@ -152,7 +152,7 @@ def clear_deprecation_log():
     # Since deprecations are de-duped, the existance (or non-existance) of
     # a deprecation warning in a previous test can cause the next test to
     # fail.
-    util.deprecate._log = set()
+    setattr(util.deprecate, "log", set())
```
TheRealFalcon, line 155: Are you sure this doesn't change any functionality? If so, +1, but I just want to make sure.

TheRealFalcon (approved):
(deleted original comment...the `_log` thing keeps confusing me for a second)

LGTM!

PR #5355, opened 2024-06-02
fix: cc_user_groups incorrectly assumes "useradd" never locks password field

dermotbradley:
## Proposed Commit Message

```
fix: cc_user_groups incorrectly assumes "useradd" never locks password

Currently cc_user_groups assumes that "useradd" never locks the password
field of newly created users. This is an incorrect assumption.

From the useradd manpage:

'-p, --password PASSWORD
The encrypted password, as returned by crypt(3). The default is to
disable the password.'

That is, if cloud-init runs 'useradd' but does not pass it the "-p"
option (with an encrypted password) then the new user's password field
will be locked by "useradd".

cloud-init only passes the "-p" option when calling "useradd" when
user-data specifies the "passwd" option for a new user. For user-data
that specifies either the "hashed_passwd" or "plain_text_passwd"
options instead then cloud-init calls "useradd" without the "-p" option
and so the password field of such a user will be locked by "useradd".

For user-data that specifies "hash_passwd" for a new user then "useradd"
is called with no "-p" option, so causing "useradd" to lock the
password field, however then cloud-init calls "chpasswd -e" to set the
encrypted password which also results in the password field being
unlocked.

For user-data that specifies either "plain_text_passwd" for a new user
then "useradd" is called with no "-p" option, so causing "useradd" to
lock the password. cloud-init then calls "chpasswd" to set the password
which also results in the password field being unlocked.

For user-data that specifies no password at all for a new user then
"useradd" is called with no "-p" option, so causing "useradd" to lock
the password. The password field is left locked.

In all the above scenarios "passwd -l" may be called later by
cloud-init to enforce "lock_passwd: true").

Conversely where "lock_passwd: false" applies the above "usermod"
situation (for "hash_passwd", "plain_text_passwd" or no
password) means that newly created users may have password
fields locked when they should be unlocked.

For Alpine, "adduser" does not support any form of password being
passed and it always locks the password field (the same point
applies about password field being unlocked when/if "chpasswd" is
called). Therefore in some situations (i.e. no password specified
in user-data) the password needs to be unlocked if
"lock_passwd: false".

This PR changes add_user (in both __init__.py and alpine.py) to
explicitly call either lock_passwd or unlock_passwd at all times to
achieve the desired final result.
```

## Test Steps

User-data:

```
users:
  - name: hash
    hash_passwd: $6$nonsense
    locked_passd: false
  - name: password
    passwd: $6$nonsense
    locked_passd: false
  - name: plain
    plain_text_passwd: test
    locked_passd: false
  - name: none
    locked_passd: false
```

Then check /etc/shadow to see if all the above users have locked or unlocked password fields.

## Checklist

## Merge type

### DIFF
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -136,6 +136,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     doas_fn = "/etc/doas.conf"
     ci_sudoers_fn = "/etc/sudoers.d/90-cloud-init-users"
     hostname_conf_fn = "/etc/hostname"
+    shadow_fn = "/etc/shadow"
     tz_zone_dir = "/usr/share/zoneinfo"
     default_owner = "root:root"
     init_cmd = ["service"]  # systemctl, service etc
@@ -655,19 +656,21 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     def get_default_user(self):
         return self.get_option("default_user")
 
-    def add_user(self, name, **kwargs):
+    def add_user(self, name, **kwargs) -> bool:
         """
         Add a user to the system using standard GNU tools
 
         This should be overridden on distros where useradd is not desirable or
         not available.
+
+        Returns False if user already exists, otherwise True.
         """
         # XXX need to make add_user idempotent somehow as we
         # still want to add groups or modify SSH keys on pre-existing
         # users in the image.
         if util.is_user(name):
             LOG.info("User %s already exists, skipping.", name)
-            return
+            return False
 
         if "create_groups" in kwargs:
             create_groups = kwargs.pop("create_groups")
@@ -771,6 +774,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             util.logexc(LOG, "Failed to create user %s", name)
             raise e
 
+        # Indicate that a new user was created
+        return True
+
     def add_snap_user(self, name, **kwargs):
         """
         Add a snappy user to the system using snappy tools
@@ -798,6 +804,62 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
         return username
 
+    def _check_if_password_field_matches(
+        self, username, pattern1, pattern2, pattern3=None, check_file=None
+    ) -> bool:
+        """
+        Check whether ``username`` user has a hashed password matching
+        either pattern.
+
+        FreeBSD, NetBSD, and OpenBSD use 3 patterns, others only use
+        2 patterns.
+
+        Returns either 'True' to indicate a match, otherwise 'False'.
+        """
+
+        if not check_file:
+            check_file = self.shadow_fn
+
+        cmd = [
+            "grep",
+            "-q",
+            "-e",
+            "^%s%s" % (username, pattern1),
+            "-e",
+            "^%s%s" % (username, pattern2),
+        ]
+        if pattern3 is not None:
+            cmd.extend(["-e", "^%s%s" % (username, pattern3)])
+        cmd.append(check_file)
+        try:
+            subp.subp(cmd)
+        except subp.ProcessExecutionError as e:
+            if e.exit_code == 1:
+                # Exit code 1 means 'grep' didn't find empty password
+                return True
+            else:
+                util.logexc(
+                    LOG,
+                    "Failed to check the status of password for user %s",
+                    username,
+                )
+                raise e
+        return False
+
+    def _check_if_existing_password(self, username, shadow_file=None) -> bool:
+        """
+        Check whether ``username`` user has an existing password (regardless
+        of whether locked or not).
+
+        Returns either 'True' to indicate a password present, or 'False'
+        for no password set.
+        """
+
+        status = not self._check_if_password_field_matches(
+            username, "::", ":!:", check_file=shadow_file
+        )
+        return status
+
     def create_user(self, name, **kwargs):
         """
         Creates or partially updates the ``name`` user in the system.
@@ -824,20 +886,103 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             return self.add_snap_user(name, **kwargs)
 
         # Add the user
-        self.add_user(name, **kwargs)
+        pre_existing_user = not self.add_user(name, **kwargs)
 
-        # Set password if plain-text password provided and non-empty
-        if "plain_text_passwd" in kwargs and kwargs["plain_text_passwd"]:
-            self.set_passwd(name, kwargs["plain_text_passwd"])
+        has_existing_password = False
+        ud_blank_password_specified = False
+        ud_password_specified = False
+        password_key = None
 
-        # Set password if hashed password is provided and non-empty
-        if "hashed_passwd" in kwargs and kwargs["hashed_passwd"]:
-            self.set_passwd(name, kwargs["hashed_passwd"], hashed=True)
+        if "plain_text_passwd" in kwargs:
+            ud_password_specified = True
+            password_key = "plain_text_passwd"
+            if kwargs["plain_text_passwd"]:
+                # Set password if plain-text password provided and non-empty
+                self.set_passwd(name, kwargs["plain_text_passwd"])
+            else:
+                ud_blank_password_specified = True
 
-        # Default locking down the account.  'lock_passwd' defaults to True.
-        # lock account unless lock_password is False.
+        if "hashed_passwd" in kwargs:
+            ud_password_specified = True
+            password_key = "hashed_passwd"
+            if kwargs["hashed_passwd"]:
+                # Set password if hashed password is provided and non-empty
+                self.set_passwd(name, kwargs["hashed_passwd"], hashed=True)
+            else:
+                ud_blank_password_specified = True
+
+        if pre_existing_user:
+            if not ud_password_specified:
+                if "passwd" in kwargs:
+                    password_key = "passwd"
+                    # Only "plain_text_passwd" and "hashed_passwd"
+                    # are valid for an existing user.
+                    LOG.warning(
+                        "'passwd' in user-data is ignored for existing "
+                        "user %s",
+                        name,
+                    )
+
+                # As no password specified for the existing user in user-data
+                # then check if the existing user's hashed password value is
+                # blank (whether locked or not).
+                if util.system_is_snappy():
+                    has_existing_password = self._check_if_existing_password(
+                        name, "/var/lib/extrausers/shadow"
+                    )
+                    if not has_existing_password:
+                        # Check /etc/shadow also
+                        has_existing_password = (
+                            self._check_if_existing_password(name)
+                        )
+                else:
+                    has_existing_password = self._check_if_existing_password(
+                        name
+                    )
+        else:
+            if "passwd" in kwargs:
+                ud_password_specified = True
+                password_key = "passwd"
+                if not kwargs["passwd"]:
+                    ud_blank_password_specified = True
+
+        # Default locking down the account. 'lock_passwd' defaults to True.
+        # Lock account unless lock_password is False in which case unlock
+        # account as long as a password (blank or otherwise) was specified.
         if kwargs.get("lock_passwd", True):
             self.lock_passwd(name)
+        elif has_existing_password or ud_password_specified:
+            # 'lock_passwd: False' and either existing account already with
+            # non-blank password or else existing/new account with password
+            # explicitly set in user-data.
+            if ud_blank_password_specified:
+                LOG.debug(
+                    "Allowing unlocking empty password for %s based on empty"
+                    " '%s' in user-data",
+                    name,
+                    password_key,
+                )
+
+            # Unlock the existing/new account
+            self.unlock_passwd(name)
+        elif pre_existing_user:
+            # Pre-existing user with no existing password and none
+            # explicitly set in user-data.
+            LOG.warning(
+                "Not unlocking blank password for existing user %s."
+                " 'lock_passwd: false' present in user-data but no existing"
+                " password set and no 'plain_text_passwd'/'hashed_passwd'"
+                " provided in user-data",
+                name,
+            )
+        else:
+            # No password (whether blank or otherwise) explicitly set
+            LOG.warning(
+                "Not unlocking password for user %s. 'lock_passwd: false'"
+                " present in user-data but no 'passwd'/'plain_text_passwd'/"
+                "'hashed_passwd' provided in user-data",
+                name,
+            )
 
         # Configure doas access
         if "doas" in kwargs:
@@ -914,6 +1059,50 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             util.logexc(LOG, "Failed to disable password for user %s", name)
             raise e
 
+    def unlock_passwd(self, name: str):
+        """
+        Unlock the password of a user, i.e., enable password logins
+        """
+        # passwd must use short '-u' due to SLES11 lacking long form '--unlock'
+        unlock_tools = (["passwd", "-u", name], ["usermod", "--unlock", name])
+        try:
+            cmd = next(tool for tool in unlock_tools if subp.which(tool[0]))
+        except StopIteration as e:
+            raise RuntimeError(
+                "Unable to unlock user account '%s'. No tools available. "
+                "  Tried: %s." % (name, [c[0] for c in unlock_tools])
+            ) from e
+        try:
+            _, err = subp.subp(cmd, rcs=[0, 3])
+        except Exception as e:
+            util.logexc(LOG, "Failed to enable password for user %s", name)
+            raise e
+        if err:
+            # if "passwd" or "usermod" are unable to unlock an account with
+            # an empty password then they display a message on stdout. In
+            # that case then instead set a blank password.
+            passwd_set_tools = (
+                ["passwd", "-d", name],
+                ["usermod", "--password", "''", name],
+            )
+            try:
+                cmd = next(
+                    tool for tool in passwd_set_tools if subp.which(tool[0])
+                )
+            except StopIteration as e:
+                raise RuntimeError(
+                    "Unable to set blank password for user account '%s'. "
+                    "No tools available. "
+                    "  Tried: %s." % (name, [c[0] for c in unlock_tools])
+                ) from e
+            try:
+                subp.subp(cmd)
+            except Exception as e:
+                util.logexc(
+                    LOG, "Failed to set blank password for user %s", name
+                )
+                raise e
+
     def expire_passwd(self, user):
         try:
             subp.subp(["passwd", "--expire", user])
@@ -948,6 +1137,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             )
             + "\n"
         )
+        # Need to use the short option name '-e' instead of '--encrypted'
+        # (which would be more descriptive) since Busybox and SLES 11
+        # chpasswd don't know about long names.
         cmd = ["chpasswd"] + (["-e"] if hashed else [])
         subp.subp(cmd, data=payload)
 
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -907,8 +907,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 # As no password specified for the existing user in user-data
                 # then check if the existing user's hashed password value is
                 # empty (whether locked or not).
-                has_existing_password = not self._shadow_file_has_empty_user_password(
-                    name
+                has_existing_password = not (
+                    self._shadow_file_has_empty_user_password(name)
                 )
         else:
             if "passwd" in kwargs:
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -137,6 +137,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     ci_sudoers_fn = "/etc/sudoers.d/90-cloud-init-users"
     hostname_conf_fn = "/etc/hostname"
     shadow_fn = "/etc/shadow"
+    shadow_extrausers_fn = "/var/lib/extrausers/shadow"
+    # /etc/shadow match patterns indicating empty passwords
+    shadow_empty_locked_passwd_patterns = ["^{username}::", "^{username}:!:"]
     tz_zone_dir = "/usr/share/zoneinfo"
     default_owner = "root:root"
     init_cmd = ["service"]  # systemctl, service etc
@@ -804,61 +807,39 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
         return username
 
-    def _check_if_password_field_matches(
-        self, username, pattern1, pattern2, pattern3=None, check_file=None
-    ) -> bool:
+    def _shadow_file_has_empty_user_password(self, username) -> bool:
         """
-        Check whether ``username`` user has a hashed password matching
-        either pattern.
+        Check whether username exists in shadow files with empty password.
 
-        FreeBSD, NetBSD, and OpenBSD use 3 patterns, others only use
-        2 patterns.
-
-        Returns either 'True' to indicate a match, otherwise 'False'.
+        Support reading /var/lib/extrausers/shadow on snappy systems.
         """
-
-        if not check_file:
-            check_file = self.shadow_fn
-
-        cmd = [
-            "grep",
-            "-q",
-            "-e",
-            "^%s%s" % (username, pattern1),
-            "-e",
-            "^%s%s" % (username, pattern2),
-        ]
-        if pattern3 is not None:
-            cmd.extend(["-e", "^%s%s" % (username, pattern3)])
-        cmd.append(check_file)
-        try:
-            subp.subp(cmd)
-        except subp.ProcessExecutionError as e:
-            if e.exit_code == 1:
-                # Exit code 1 means 'grep' didn't find empty password
-                return True
-            else:
-                util.logexc(
-                    LOG,
-                    "Failed to check the status of password for user %s",
-                    username,
-                )
-                raise e
-        return False
-
-    def _check_if_existing_password(self, username, shadow_file=None) -> bool:
-        """
-        Check whether ``username`` user has an existing password (regardless
-        of whether locked or not).
-
-        Returns either 'True' to indicate a password present, or 'False'
-        for no password set.
-        """
-
-        status = not self._check_if_password_field_matches(
-            username, "::", ":!:", check_file=shadow_file
+        if util.system_is_snappy():
+            shadow_files = [self.shadow_extrausers_fn, self.shadow_fn]
+        else:
+            shadow_files = [self.shadow_fn]
+        shadow_empty_passwd_re = "|".join(
+            [
+                pattern.format(username=username)
+                for pattern in self.shadow_empty_locked_passwd_patterns
+            ]
         )
-        return status
+        for shadow_file in shadow_files:
+            if not os.path.exists(shadow_file):
+                continue
+            shadow_content = util.load_text_file(shadow_file)
+            if not re.findall(rf"^{username}:", shadow_content, re.MULTILINE):
+                LOG.debug("User %s not found in %s", username, shadow_file)
+                continue
+            LOG.debug(
+                "User %s found in %s. Checking for empty password",
+                username,
+                shadow_file,
+            )
+            if re.findall(
+                shadow_empty_passwd_re, shadow_content, re.MULTILINE
+            ):
+                return True
+        return False
 
     def create_user(self, name, **kwargs):
         """
@@ -925,20 +906,10 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
 
                 # As no password specified for the existing user in user-data
                 # then check if the existing user's hashed password value is
-                # blank (whether locked or not).
-                if util.system_is_snappy():
-                    has_existing_password = self._check_if_existing_password(
-                        name, "/var/lib/extrausers/shadow"
-                    )
-                    if not has_existing_password:
-                        # Check /etc/shadow also
-                        has_existing_password = (
-                            self._check_if_existing_password(name)
-                        )
-                else:
-                    has_existing_password = self._check_if_existing_password(
-                        name
-                    )
+                # empty (whether locked or not).
+                has_existing_password = (
+                    not self._shadow_file_has_empty_user_password(name)
+                )
         else:
             if "passwd" in kwargs:
                 ud_password_specified = True
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -907,8 +907,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                 # As no password specified for the existing user in user-data
                 # then check if the existing user's hashed password value is
                 # empty (whether locked or not).
-                has_existing_password = (
-                    not self._shadow_file_has_empty_user_password(name)
+                has_existing_password = not self._shadow_file_has_empty_user_password(
+                    name
                 )
         else:
             if "passwd" in kwargs:
--- a/cloudinit/distros/alpine.py
+++ b/cloudinit/distros/alpine.py
@@ -205,16 +205,18 @@ class Distro(distros.Distro):
 
         return self._preferred_ntp_clients
 
-    def add_user(self, name, **kwargs):
+    def add_user(self, name, **kwargs) -> bool:
         """
         Add a user to the system using standard tools
 
         On Alpine this may use either 'useradd' or 'adduser' depending
         on whether the 'shadow' package is installed.
+
+        Returns False if user already exists, otherwise True.
         """
         if util.is_user(name):
             LOG.info("User %s already exists, skipping.", name)
-            return
+            return False
 
         if "selinux_user" in kwargs:
             LOG.warning("Ignoring selinux_user parameter for Alpine Linux")
@@ -418,6 +420,9 @@ class Distro(distros.Distro):
                 LOG, "Failed to update %s for user %s", shadow_file, name
             )
 
+        # Indicate that a new user was created
+        return True
+
     def lock_passwd(self, name):
         """
         Lock the password of a user, i.e., disable password logins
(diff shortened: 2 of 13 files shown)


dermotbradley:
@holmanb I'm guessing your "exception" changes (#5202) suggestion would also apply here to __init__.py

dermotbradley:
@blackboxsw @holmanb 

I'd like to get this into 24.2 if possible.

dermotbradley:
Hmm, I just noticed that freebsd.py does not have an unlock_passwd function, whereas netbsd.py and opensdb.py do.

The BSDs are a little strange in that there is no distinction between password locking and account locking/disabling, unlike Linux.

@igalic any thoughts? Would "pw usermod <name> -w none" be safe/sensible for a FreeBSD unlock_passwd function in response to "lock_passwd: false" user-data?

cloudinit/distros/__init__.py
```diff
@@ -908,6 +910,50 @@ def lock_passwd(self, name):
...
+            util.logexc(LOG, "Failed to enable password for user %s", name)
+            raise e
+        if err:
+            # if "passwd" or "usermod" are unable to unlock an account with
+            # an empty password then they display a message on stdout. In
+            # that case then instead set a blank password.
+            passwd_set_tools = (
+                ["passwd", "-d", name],
+                ["usermod", "--password", "''", name],
+            )
```
blackboxsw, line 938: I'm very hesitant to call this behavior as fallback if `lock_passwd: False` as it is a poor security policy decision which is why the passwd tooling exits non-zero (3) to warn about this and be noisy.  I'm not thrilled about cloud-init trying to support this lax security policy, but I can think of a couple of cases where it may make sense to have a completely open set of instances on a private network to simplify tooling and automation.

I think minimally cloud-init should:
- log a warning about this behavior which will allow cloud-init status to exit non-zero(2) telling the admin that they have an "poor security practice open system user account with an empty password defined" if someone configures their user with lock_passwd false and doesn't provide a valid non-empty passwd/plain_text_passwd

Ideally, if we think a person really wants to force an empty password and completely expose that instance to simple remote console attacks without getting a warning,  then we should probably avoid calling this `passwd -d` unless either an empty  `plain_text_passwd: ""` or `passwd: ""`  was explicitly provided. Otherwise, we risk opening up password-less access for folks who didn't realize the impact of providing ` lock_passwd: false`  without some form of explicit passwd defined.

I'd be a little more convinced to support this password-less account if it requires explicit additional configuration (such as providing an empty string to `*passwd` cloud-config user keys in user-data to define "no I really do want empty passwords for this use, I know the risks"


So ideal way I'd like to allow folks to express this would be:
```
#cloud-config 
users:
- default
- name: openuser
  gecos: Open User with no password
  primary_group: openuser
  lock_passwd: false
  passwd: ""  
  # or
  plain_text_passwd: ""
```

blackboxsw (changes requested):
@dermotbradley thank  you for putting together this PR and continued refinement of what user setup means on cloud-init.  I feel like this represents a behavior change broadly that now allows for setup of password-less user accounts which I didn't think cloud-init allowed previously because of security concerns.

I think we may want to avoid 24.2 for this particular behavior change because I think it warrants an extra set of eyes from our security team to think through the implications of a configured vm w/  user accounts that can be allowed to have empty passwords.

I believe if we can come to a strict requirement in cloud-config configuration that establishes a pattern where the admin can be explicit that they do want a password-less user account, I'd feel better about this approach instead of just keying on `lock_passwd: false`.

blackboxsw:
Given that we want security review on this behavior let's push this to 24.3 milestone to make sure we can come up with an acceptable/explicit solution that allows password-less account and test this broadly

cloudinit/distros/__init__.py
```diff
@@ -908,6 +910,50 @@ def lock_passwd(self, name):
...
+        unlock_tools = (["passwd", "-u", name], ["usermod", "--unlock", name])
+        try:
+            cmd = next(tool for tool in unlock_tools if subp.which(tool[0]))
+        except StopIteration as e:
+            raise RuntimeError(
+                "Unable to unlock user account '%s'. No tools available. "
+                "  Tried: %s." % (name, [c[0] for c in unlock_tools])
+            ) from e
+        try:
+            _, err = subp.subp(cmd, rcs=[0, 3])
```
blackboxsw, line 927: Because we are falling back on a non-empty stderr due to exit 3 from passwd -u <user>, we are taking automated steps in cloud-init to essentially ignore the warning message about risky password policies that passwd is trying to show us with that exit 3
```
passwd: unlocking the password would result in a passwordless account.
You should set a password with usermod -p to unlock the password of this account.
```

Note that usermod -U <user> would also advise us against this too with a similar message:
```
usermod: unlocking the user's password would result in a passwordless account.
You should set a password with usermod -p to unlock this user's password
```

I really think we should make the admin be intentional in their cloud-config somehow if they really need as passwordless user.  As it is currently if a user is defined with cloud-config: `lock_passwd: false` and they mistakenly omitted a passwd/hashed_passwd/plain_text_passwd related key, cloud-init would happily setup that user with an empty password without any sort of notification about this potentially risky configuration. As mentioned in the review, if we can find a more clear expression of "I really want an empty password"  in cloud-config, such as providing a specific empty string (not None/null) for one of those password keys, I think that would make me more comfortable with the admin's intent.

cloudinit/distros/__init__.py
```diff
@@ -908,6 +910,50 @@ def lock_passwd(self, name):
...
+                    tool for tool in passwd_set_tools if subp.which(tool[0])
+                )
+            except StopIteration as e:
+                raise RuntimeError(
+                    "Unable to set blank password for user account '%s'. "
+                    "No tools available. "
+                    "  Tried: %s." % (name, [c[0] for c in unlock_tools])
+                ) from e
+            try:
+                _, err = subp.subp(cmd, rcs=[0, 3])
```
blackboxsw, line 950: Additionally I think we don't actually want to succeed on exit 3 here do we? That means empty password setup didn't work and we noop'd.
```suggestion
                _, err = subp.subp(cmd, rcs=[0, 3])
```
```suggestion
                _, err = subp.subp(cmd)
```

blackboxsw (changes requested):
Additionally I think we need:
- the second subp which call which attempts to delete the user password should not pass on exit 3. 
- we should get some unittest coverage of the fallback to delete user passwords when empty string passwd, plain_text_passwd or hashed_passwd.

Here's a combined diff of my suggestions including some additional tests:
```diff
diff --git a/cloudinit/distros/__init__.py b/cloudinit/distros/__init__.py
index 60466eba5..f26420afd 100644
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -833,7 +833,17 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         if kwargs.get("lock_passwd", True):
             self.lock_passwd(name)
         else:
-            self.unlock_passwd(name)
+            allow_empty_passwd = False
+            for passwd_key in ("passwd", "hashed_passwd", "plain_text_passwd"):
+                if "" == kwargs.get(passwd_key):
+                    allow_empty_passwd = True
+                    LOG.debug(
+                        "Allowing empty password for %s based on empty %s in"
+                        " cloud-config",
+                        name,
+                        passwd_key,
+                    )
+            self.unlock_passwd(name, allow_empty_passwd)
 
         # Configure doas access
         if "doas" in kwargs:
@@ -910,7 +920,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             util.logexc(LOG, "Failed to disable password for user %s", name)
             raise e
 
-    def unlock_passwd(self, name):
+    def unlock_passwd(self, name: str, allow_empty_passwd: bool = False):
         """
         Unlock the password of a user, i.e., enable password logins
         """
@@ -929,6 +939,14 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
             util.logexc(LOG, "Failed to enable password for user %s", name)
             raise e
         if err:
+            if not allow_empty_passwd:
+                LOG.warning(
+                    "Leaving password locked of user %s. lock_passwd: false in"
+                    " user-data but no passwd, plain_text_passwd or"
+                    " hashed_passwd provided",
+                    name,
+                )
+                return
             # if "passwd" or "usermod" are unable to unlock an account with
             # an empty password then they display a message on stdout. In
             # that case then instead set a blank password.
@@ -947,7 +965,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
                     "  Tried: %s." % (name, [c[0] for c in unlock_tools])
                 ) from e
             try:
-                _, err = subp.subp(cmd, rcs=[0, 3])
+                subp.subp(cmd)
             except Exception as e:
                 util.logexc(
                     LOG, "Failed to set blank password for user %s", name
diff --git a/tests/unittests/distros/test_create_users.py b/tests/unittests/distros/test_create_users.py
index 3e13316c9..459820433 100644
--- a/tests/unittests/distros/test_create_users.py
+++ b/tests/unittests/distros/test_create_users.py
@@ -36,10 +36,11 @@ class TestCreateUser:
         )
 
     @pytest.mark.parametrize(
-        "create_kwargs,expected",
+        "create_kwargs,subp_side_effect, expected",
         [
             pytest.param(
                 {},
+                None,
                 [
                     _useradd2call([USER, "-m"]),
                     mock.call(["passwd", "-l", USER]),
@@ -48,6 +49,7 @@ class TestCreateUser:
             ),
             pytest.param(
                 {"no_create_home": True},
+                None,
                 [
                     _useradd2call([USER, "-M"]),
                     mock.call(["passwd", "-l", USER]),
@@ -56,6 +58,7 @@ class TestCreateUser:
             ),
             pytest.param(
                 {"system": True},
+                None,
                 [
                     _useradd2call([USER, "--system", "-M"]),
                     mock.call(["passwd", "-l", USER]),
@@ -64,6 +67,7 @@ class TestCreateUser:
             ),
             pytest.param(
                 {"create_no_home": False},
+                None,
                 [
                     _useradd2call([USER, "-m"]),
                     mock.call(["passwd", "-l", USER]),
@@ -72,14 +76,82 @@ class TestCreateUser:
             ),
             pytest.param(
                 {"lock_passwd": False},
+                None,
                 [
                     _useradd2call([USER, "-m"]),
                     mock.call(["passwd", "-u", USER], rcs=[0, 3]),
                 ],
                 id="unlocked",
             ),
+            pytest.param(
+                {"lock_passwd": False},
+                [
+                    ("", ""),
+                    (
+                        "",
+                        "usermod: unlocking the user's password would result in a passwordless account.",
+                    ),
+                ],
+                [
+                    _useradd2call([USER, "-m"]),
+                    mock.call(["passwd", "-u", USER], rcs=[0, 3]),
+                ],
+                id="unlocked_without_empty_string_passwd_does_not_unlock",
+            ),
+            pytest.param(
+                {"lock_passwd": False, "passwd": ""},
+                [
+                    ("", ""),
+                    (
+                        "",
+                        "usermod: unlocking the user's password would result in a passwordless account.",
+                    ),
+                    ("", ""),
+                ],
+                [
+                    _useradd2call([USER, "-m"]),
+                    mock.call(["passwd", "-u", USER], rcs=[0, 3]),
+                    mock.call(["passwd", "-d", USER]),
+                ],
+                id="unlocked_with_empty_string_passwd_does_unlock",
+            ),
+            pytest.param(
+                {"lock_passwd": False, "plain_text_passwd": ""},
+                [
+                    ("", ""),
+                    (
+                        "",
+                        "usermod: unlocking the user's password would result in a passwordless account.",
+                    ),
+                    ("", ""),
+                ],
+                [
+                    _useradd2call([USER, "-m"]),
+                    mock.call(["passwd", "-u", USER], rcs=[0, 3]),
+                    mock.call(["passwd", "-d", USER]),
+                ],
+                id="unlocked_with_empty_string_plain_text_passwd_does_unlock",
+            ),
+            pytest.param(
+                {"lock_passwd": False, "hashed_passwd": ""},
+                [
+                    ("", ""),
+                    (
+                        "",
+                        "usermod: unlocking the user's password would result in a passwordless account.",
+                    ),
+                    ("", ""),
+                ],
+                [
+                    _useradd2call([USER, "-m"]),
+                    mock.call(["passwd", "-u", USER], rcs=[0, 3]),
+                    mock.call(["passwd", "-d", USER]),
+                ],
+                id="unlocked_with_empty_string_hashed_passwd_does_unlock",
+            ),
             pytest.param(
                 {"passwd": "passfoo"},
+                None,
                 [
                     _useradd2call([USER, "--password", "passfoo", "-m"]),
                     mock.call(["passwd", "-l", USER]),
@@ -88,8 +160,13 @@ class TestCreateUser:
             ),
         ],
     )
-    def test_create_options(self, m_subp, dist, create_kwargs, expected):
-        m_subp.return_value = ("", "")
+    def test_create_options(
+        self, m_subp, dist, create_kwargs, subp_side_effect, expected
+    ):
+        if subp_side_effect:
+            m_subp.side_effect = subp_side_effect
+        else:
+            m_subp.return_value = ("", "")
         dist.create_user(name=USER, **create_kwargs)
         assert m_subp.call_args_list == expected
 
```

dermotbradley:
> I feel like this represents a behavior change broadly that now allows for setup of password-less user accounts which I didn't think cloud-init allowed previously because of security concerns.

Well cloud-init has always allowed the setup of password-less user accounts when neither "passwd", "hashed_passwd", nor "plain_text_passwd" are specified for a user.

I assume you meant 'password-less and not-locked user accounts'. cloud-init has always only enforced the locking of passwords whenever "lock_passwd: true" (whether as defined by default in /etc/cloud/cloud.cfg or explicitly specified in user-data).

If "lock_passwd: true" is not specified, indeed if "lock_passwd" is not specified at all, then cloud-init makes no attempt to lock passwords.

> I think we may want to avoid 24.2 for this particular behavior change because I think it warrants an extra set of eyes from our security team to think through the implications of a configured vm w/ user accounts that can be allowed to have empty passwords.

Ok.

> I believe if we can come to a strict requirement in cloud-config configuration that establishes a pattern where the admin can be explicit that they do want a password-less user account, I'd feel better about this approach instead of just keying on `lock_passwd: false`.

There are 2 distinct issues here: the issue of empty/blank passwords, and the issue of password-accessible accounts (i.e. via SSH) whether they have passwords or not.

One suggestion would be to deprecate "lock_passwd" and as a replacement to define "unlock_passwd" instead (which is "false" by default) and perhaps to also define "unlock_when_passwd_blank" (again "false" by default) as an additional level of safety.

dermotbradley:
@blackboxsw

I looked at your comments out of order. I'll look into the combination of ```lock_passwd: false``` and ```passwd: ""``` as you suggested.

dermotbradley:
@blackboxsw 

How would you suggest treating "lock_passwd" for **existing** users? i.e. this config:

```
users:
  - name: existinguser
    lock_passwd: false
```

Should the code grep /etc/shadow to check whether the existing user's password is blank and only unlock when it is not blank?

blackboxsw:
> @blackboxsw
> 
> How would you suggest treating "lock_passwd" for **existing** users? i.e. this config:
> 
> ```
> users:
>   - name: existinguser
>     lock_passwd: false
> ```
> 
> Should the code grep /etc/shadow to check whether the existing user's password is blank and only unlock when it is not blank?

@dermotbradley That adds a bit of complexity certainly. But, I'm understanding that the use-case here would be to support custom images w/ precreated user account which allows image creators to define an opaque password that we don't want to represent in sensitive user-data. I think we actually may have to treat that case as I believe [Azure has cases where they may set lock_passwd = False during preprovisioning and not provide a hashed_password when the password was redacted due to lack of complexity](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceAzure.py#L1854-L1856).  In those cases where user-exists from an earlier provisioning stage, but we don't represent that in current boot's user-data, we'd probably want to allow that account to login, but only if it has some representation of complexity (like non-empty).

CC: @cjp256 FYI here as this PR in progress touches represents a potential change in behavior as it will explicitly perform an unlock of a account's password via `passwd -u <default_username>` when `lock_passwd: false` is provided for the user in configuration. This differs from previous behavior on `lock_passwd: false` which only would only avoid calling `passwd -l <username>` in [distros.lock_passwd](https://github.com/canonical/cloud-init/blob/main/cloudinit/distros/__init__.py#L833-L834)

dermotbradley:
> @dermotbradley That adds a bit of complexity certainly. But, I'm understanding that the use-case here would be to support custom images w/ precreated user account which allows image creators to define an opaque password that we don't want to represent in sensitive user-data.

@blackboxsw I already went ahead and implemented /etc/shadow password value checking for pre-existing user accounts.

I have reworked my changes based on your feedback and manually tested it locally. I haven't yet pushed my changes as I'm trying to determine which tests to modify or add for this.

I could go ahead now and push my code changes (without testcase changes) anyway if you want to have a look at it...

blackboxsw:
@dermotbradley please feel free to push and I'll understand more may come or you may squash merge testing as added. I could assist if needed on test generation front for this.

cloudinit/distros/__init__.py
dermotbradley: I have addressed this now.

cloudinit/distros/__init__.py
dermotbradley: Addressed.

cloudinit/distros/__init__.py
dermotbradley: Addressed.

dermotbradley:
> @dermotbradley please feel free to push and I'll understand more may come or you may squash merge testing as added. I could assist if needed on test generation front for this.

Ok, I've pushed my current version.

Still to be addressed:

- testcases

- FreeBSD/NetBSD/OpenBSD changes

dermotbradley:
@blackboxsw could you have a look at this?

cloudinit/distros/__init__.py
```diff
@@ -818,20 +823,108 @@ def create_user(self, name, **kwargs):
...
+                # check if the existing user's hashed password value in
+                # /etc/shadow is blank (whether locked or not).
+                cmd = [
+                    "grep",
+                    "-q",
+                    "-e",
+                    "^%s::" % name,
+                    "-e",
+                    "^%s:!:" % name,
+                    "/etc/shadow",
```
blackboxsw, line 872: There are some enviroments (util.is_snappy()) where /etc/shadow is read-only and any added users by cloud-init or other are forced to use `--extrausers` flag to store those users in /var/lib/extrausers/shadow. We may want to grep that file too if util.is_snappy()

blackboxsw: Hardcoding a file path like `/etc/shadow` here is probably better places as an attr on the class or instance such as `shadow_fns: [...]` which gets  is queried based on the distro you are running in. In most cases I expect this to just be a one-item list, but in snappy case it'd be two files to process/parse/grep

cloudinit/distros/__init__.py
```diff
@@ -818,20 +823,108 @@ def create_user(self, name, **kwargs):
...
+                        # Exit code 1 means 'grep' didn't find empty password
+                        has_password = True
+                    else:
+                        util.logexc(
+                            LOG,
+                            "Failed to Check if user %s already has "
+                            "a password",
+                            name,
+                        )
+                        raise e
```
blackboxsw, line 887: Let's make a encapsulate this logic in an instance method `is_user_password_empty_or_locked()` to make it easier to digest. Then that method can grep against `shadow_files = [f for f in [self.shadow_fn, self.extrausers_shadow_fn] if os.path.exists(f)]`

cloudinit/distros/__init__.py
```diff
@@ -818,20 +823,108 @@ def create_user(self, name, **kwargs):
...
+            # 'lock_passwd: False' and password explicitly set
+            if blank_password and not pre_existing_user:
+                LOG.debug(
+                    "Allowing unlocking empty password for %s based on empty"
+                    " '%s' in user-data",
+                    name,
+                    password_key,
+                )
+
+            self.unlock_passwd(name)
```
blackboxsw, line 910: This unlock is still being  called even if `blank_password and pre_existing_user.` I think we still don't want to unlock the password in this case.

I'm working on a couple of unittests for these paths so we can be more certain we aren't adding exposure to unlocking insecure accounts by default.

blackboxsw (changes requested):
I think we may need to look at:
1.  when we are calling self.unlock_passwd still to avoid unlocks when and 
2. minor refactor suggestions to create a method for the grep of shadow files and hang self.shadow_fn  or shadow_extrausers_fn on the Distro instance only trying to grep those files if they exist.
3. Below is a start for unittest coverage of some of this logic, what I think needs coverage is the shadow file parsing and determination of whether an empty or locked passwd exists for a pre_existing user.
4.  It's probably worth extending one of our existing integration tests in tests/integration_tests/modules/test_users_groups.py or tests/integration_tests/modules/test_set_password.py to provide hashed_passwd and/or plain_test_password to assert this behavior


```diff
diff --git a/tests/unittests/distros/test_create_users.py b/tests/unittests/distros/test_create_users.py
index 039723aaa..6823d4327 100644
--- a/tests/unittests/distros/test_create_users.py
+++ b/tests/unittests/distros/test_create_users.py
@@ -16,6 +16,21 @@ def common_mocks(mocker):
     mocker.patch("cloudinit.distros.util.system_is_snappy", return_value=False)
 
 
+def _existing_shadow_grep(name: str):
+    """Return a mock of grep of /etc/shadow call based on username."""
+    return mock.call(
+        ["grep", "-q", "-e", "^foo_user::", "-e", f"^{name}:!:", "/etc/shadow"]
+    )
+
+
+def _chpasswdmock(name: str, password: str, hashed: bool = False):
+    """Return a mock of chpasswd call based on args"""
+    cmd = ["chpasswd", "-e"] if hashed else ["chpasswd"]
+    return mock.call(
+        cmd, data=f"{name}:{password}", logstring=f"chpasswd for {name}"
+    )
+
+
 def _useradd2call(args: List[str]):
     # return a mock call for the useradd command in args
     # with expected 'logstring'.
@@ -76,17 +91,97 @@ class TestCreateUser:
                 id="unlocked",
             ),
             pytest.param(
-                {"passwd": "passfoo"},
+                {"passwd": "$6$rounds=..."},
+                [
+                    _useradd2call([USER, "--password", "$6$rounds=...", "-m"]),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                id="set_implicit_encrypted_password",
+            ),
+            pytest.param(
+                {"passwd": ""},
+                [
+                    _useradd2call([USER, "-m"]),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                id="set_empty_passwd_new_user",
+            ),
+            pytest.param(
+                {"plain_text_passwd": "clearfoo"},
+                [
+                    _useradd2call([USER, "-m"]),
+                    _chpasswdmock(USER, "clearfoo"),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                id="set_plain_text_password",
+            ),
+            pytest.param(
+                {"hashed_passwd": "$6$rounds=..."},
+                [
+                    _useradd2call([USER, "-m"]),
+                    _chpasswdmock(USER, "$6$rounds=...", hashed=True),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                id="set_explicitly_hashed_password",
+            ),
+        ],
+    )
+    @mock.patch("cloudinit.distros.util.is_user", return_value=False)
+    def test_create_options(
+        self, _is_user, m_subp, dist, create_kwargs, expected
+    ):
+        dist.create_user(name=USER, **create_kwargs)
+        assert m_subp.call_args_list == expected
+
+    @pytest.mark.parametrize(
+        "create_kwargs,expected,expected_logs",
+        [
+            pytest.param(
+                {"passwd": "$6$rounds=..."},
                 [
-                    _useradd2call([USER, "--password", "passfoo", "-m"]),
+                    _existing_shadow_grep(USER),
                     mock.call(["passwd", "-l", USER]),
                 ],
-                id="set_password",
+                [
+                    "'passwd' in user-data is ignored for existing user foo_user"
+                ],
+                id="skip_passwd_set_on_existing_user",
+            ),
+            pytest.param(
+                {"plain_text_passwd": "clearfoo"},
+                [
+                    _chpasswdmock(USER, "clearfoo"),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                [],
+                id="set_plain_text_password_on_existing_user",
+            ),
+            pytest.param(
+                {"hashed_passwd": "$6$rounds=..."},
+                [
+                    _chpasswdmock(USER, "$6$rounds=...", hashed=True),
+                    mock.call(["passwd", "-l", USER]),
+                ],
+                [],
+                id="set_explicitly_hashed_password",
             ),
         ],
     )
-    def test_create_options(self, m_subp, dist, create_kwargs, expected):
+    @mock.patch("cloudinit.distros.util.is_user", return_value=True)
+    def test_create_passwd_existing_user(
+        self,
+        m_is_user,
+        m_subp,
+        dist,
+        create_kwargs,
+        expected,
+        expected_logs,
+        caplog,
+    ):
+        """When user exists, don't unlock on empty or locked passwords."""
         dist.create_user(name=USER, **create_kwargs)
+        for log in expected_logs:
+            assert log in caplog.text
         assert m_subp.call_args_list == expected
 
     @mock.patch("cloudinit.distros.util.is_group")
```

cloudinit/distros/__init__.py
cjp256: May be worth considering using `passwd --status <user>`

$ sudo passwd --status <user> | awk '{print $2}'
P  => password
NP => no password
L => locked

blackboxsw: Excellent suggestion @cjp256! that does take into account both /etc/shadow and /var/lib/extrausers sources. It'd be better to rely on tooling than a grep as well.

blackboxsw: The question remains, is `passwd --status` supported broadly in most distros.

dermotbradley: > May be worth considering using `passwd --status <user>`

Nope, that cannot distinguish between a password field value of "!" and "!$6$..." (in both cases it indicates "L") which is **exactly** what the grep is intended to do.

dermotbradley: > There are some enviroments (util.system_is_snappy()) where /etc/shadow is read-only and any added users by cloud-init or other are forced to use `--extrausers` flag to store those users in /var/lib/extrausers/shadow. We may want to grep that file too if util.system_is_snappy()

Ah ok, I know nothing about snappy.

So with snappy should the /var/lib/extrausers/shadow file be grepped first and then /etc/shadow only grepped if the user has no entry at all in the /var/lib/extrausers/shadow file?

cloudinit/distros/__init__.py
dermotbradley: @blackboxsw

Yeah I'm going to move this logic into a function, especially as I just realised that it doesn't cater for the BSDs - FreeBSD uses /etc/master.passwd instead of /etc/shadow, I need to work out what the other BSDs use.

cloudinit/distros/__init__.py
dermotbradley: I'm renaming some of the variables to make it clear what their purpose is, so "blank_password" becomes "ud_blank_password_specified" to indicate this is where user-data contains a passwd/plain_text_passwd/hashed_passwd value of "".

cloudinit/distros/__init__.py
blackboxsw: @dermotbradley yes. if /var/lib/extrausers/shadow exists we can grep that first and fallback to /etc/shadow if the user has no entry at all in /var/lib/extrausers/shadow +1

cloudinit/distros/__init__.py
blackboxsw: +1

cloudinit/distros/__init__.py
blackboxsw: @dermotbradley thanks on this. If we get a chance to get this updated with any changes/refactors please ping. I'd like to get this into 24.3 which is scheduled for the week of Aug 19th.

dermotbradley: @blackboxsw Ok, I've pushed further changes.

Still needed are tests for snappy and for the various BSDs - neither of those are currently tested at all in test_create_users.py.

I'd also like to get this in 24.3 if possible.

cloudinit/distros/__init__.py
dermotbradley: @blackboxsw

logic moved into function as discussed. Also now caters for NetBSD/FreeBSD/OpenBSD behaviour correctly (I did some, non cloud-init, testing with VMs of those).

cloudinit/distros/__init__.py
dermotbradley: @blackboxsw done.

dermotbradley:
Ok, added some DragonflyBSD/FreeBSD/OpenBSD tests.

cloudinit/distros/__init__.py
```diff
@@ -792,6 +798,62 @@ def add_snap_user(self, name, **kwargs):
...
+
+        Returns either 'True' to indicate a password present, or 'False'
+        for no password set.
+        """
+
+        status = not self._check_if_password_field_matches(
+            username, "::", ":!:", check_file=shadow_file
+        )
+        return status
+
```
blackboxsw, line 856: Thank you @dermotbradley. In your latest squash commit you added these two methods but overrode _check_if_existing_pasword on all BSDs.

Instead of hard-coding minor differences this method in each BSD subclass, can we hang a class attribute `shadow_empty_locked_passwd_patterns` on the distro class that we override when needed?

It'd look like this:
```diff
--- a/cloudinit/distros/__init__.py
+++ b/cloudinit/distros/__init__.py
@@ -134,6 +134,8 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
     ci_sudoers_fn = "/etc/sudoers.d/90-cloud-init-users"
     hostname_conf_fn = "/etc/hostname"
     shadow_fn = "/etc/shadow"
+    # /etc/shadow match patterns indicating empty passwords
+    shadow_empty_locked_passwd_patterns = ["^{username}::", "^{username}:!:"]
     tz_zone_dir = "/usr/share/zoneinfo"
     default_owner = "root:root"
     init_cmd = ["service"]  # systemctl, service etc
@@ -799,7 +801,7 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         return username
 
     def _check_if_password_field_matches(
-        self, username, pattern1, pattern2, pattern3=None, check_file=None
+        self, username, patterns=[], check_file=None
     ) -> bool:
         """
         Check whether ``username`` user has a hashed password matching
@@ -817,13 +819,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         cmd = [
             "grep",
             "-q",
-            "-e",
-            "^%s%s" % (username, pattern1),
-            "-e",
-            "^%s%s" % (username, pattern2),
         ]
-        if pattern3 is not None:
-            cmd.extend(["-e", "^%s%s" % (username, pattern3)])
+        for pattern in patterns:
+            cmd.extend(["-e", pattern.format(username=username)])
         cmd.append(check_file)
         try:
             subp.subp(cmd)
@@ -850,7 +848,9 @@ class Distro(persistence.CloudInitPickleMixin, metaclass=abc.ABCMeta):
         """
 
         status = not self._check_if_password_field_matches(
-            username, "::", ":!:", check_file=shadow_file
+            username,
+            self.shadow_empty_locked_passwd_patterns,
+            check_file=shadow_file,
         )
         return status
 
diff --git a/cloudinit/distros/freebsd.py b/cloudinit/distros/freebsd.py
index ccd961159..7a91dc75b 100644
--- a/cloudinit/distros/freebsd.py
+++ b/cloudinit/distros/freebsd.py
@@ -41,6 +41,16 @@ class Distro(cloudinit.distros.bsd.BSD):
     dhclient_lease_directory = "/var/db"
     dhclient_lease_file_regex = r"dhclient.leases.\w+"
 
+    # /etc/shadow match patterns indicating empty passwords
+    # For FreeBSD (from https://man.freebsd.org/cgi/man.cgi?passwd(5)) a
+    # password field of "" indicates no password, and a password
+    # field value of either "*" or "*LOCKED*" indicate differing forms of
+    # "locked" but with no password defined.
+    shadow_empty_locked_passwd_patterns = [
+        "^{username}::", "^{username}:*:", "^{username}:*LOCKED*:"
+    ]
+
+
     @classmethod
     def reload_init(cls, rcs=None):
         """
@@ -148,25 +158,6 @@ class Distro(cloudinit.distros.bsd.BSD):
         # Indicate that a new user was created
         return True
 
-    def _check_if_existing_password(self, username, shadow_file=None) -> bool:
-        """
-        Check whether ``username`` user has an existing password (regardless
-        of whether locked or not).
-
-        For FreeBSD (from https://man.freebsd.org/cgi/man.cgi?passwd(5)) a
-        password field of "" indicates no password, and a password
-        field value of either "*" or "*LOCKED*" indicate differing forms of
-        "locked" but with no password defined.
-
-        Returns either 'True' to indicate a password present, or 'False'
-        for no password set.
-        """
-
-        status = not self._check_if_password_field_matches(
-            username, "::", ":*:", ":*LOCKED*:", check_file=shadow_file
-        )
-        return status
-
     def expire_passwd(self, user):
         try:
             subp.subp(["pw", "usermod", user, "-p", "01-Jan-1970"])
diff --git a/cloudinit/distros/netbsd.py b/cloudinit/distros/netbsd.py
index b9e038e89..2f687eddb 100644
--- a/cloudinit/distros/netbsd.py
+++ b/cloudinit/distros/netbsd.py
@@ -49,6 +49,18 @@ class NetBSD(cloudinit.distros.bsd.BSD):
     ci_sudoers_fn = "/usr/pkg/etc/sudoers.d/90-cloud-init-users"
     group_add_cmd_prefix = ["groupadd"]
 
+    # For NetBSD (from https://man.netbsd.org/passwd.5) a password field
+    # value of either "" or "*************" (13 "*") indicates no password,
+    # a password field prefixed with "*LOCKED*" indicates a locked
+    # password, and a password field of "*LOCKED*" followed by 13 "*"
+    # indicates a locked and blank password.
+    shadow_empty_locked_passwd_patterns = [
+        "^{username}::",
+        "^{username}:*************:",
+        "^{username}:*LOCKED**************:",
+    ]
+
+
     def __init__(self, name, cfg, paths):
         super().__init__(name, cfg, paths)
         if os.path.exists("/usr/pkg/bin/pkgin"):
@@ -120,30 +132,6 @@ class NetBSD(cloudinit.distros.bsd.BSD):
         # Indicate that a new user was created
         return True
 
-    def _check_if_existing_password(self, username, shadow_file=None) -> bool:
-        """
-        Check whether ``username`` user has an existing password (regardless
-        of whether locked or not).
-
-        For NetBSD (from https://man.netbsd.org/passwd.5) a password field
-        value of either "" or "*************" (13 "*") indicates no password,
-        a password field prefixed with "*LOCKED*" indicates a locked
-        password, and a password field of "*LOCKED*" followed by 13 "*"
-        indicates a locked and blank password.
-
-        Returns either 'True' to indicate a password present, or 'False'
-        for no password set.
-        """
-
-        status = not self._check_if_password_field_matches(
-            username,
-            "::",
-            ":*************:",
-            ":*LOCKED**************:",
-            check_file=shadow_file,
-        )
-        return status
-
     def set_passwd(self, user, passwd, hashed=False):
         if hashed:
             hashed_pw = passwd
diff --git a/cloudinit/distros/openbsd.py b/cloudinit/distros/openbsd.py
index 418e99bc4..c7d40d6c7 100644
--- a/cloudinit/distros/openbsd.py
+++ b/cloudinit/distros/openbsd.py
@@ -14,6 +14,14 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
     hostname_conf_fn = "/etc/myname"
     init_cmd = ["rcctl"]
 
+    # For OpenBSD (from https://man.openbsd.org/passwd.5) a password field
+    # of "" indicates no password, and password field values of either
+    # "*" or "*************" (13 "*") indicate differing forms of "locked"
+    # but with no password defined.
+    shadow_empty_locked_passwd_patterns = [
+        "^{username}::", "^{username}:*:", "^{username}:*************:"
+    ]
+
     def _read_hostname(self, filename, default=None):
         return util.load_text_file(self.hostname_conf_fn)
 
@@ -45,25 +53,6 @@ class Distro(cloudinit.distros.netbsd.NetBSD):
         cmd = list(init_cmd) + list(cmds[action])
         return subp.subp(cmd, capture=True, rcs=rcs)
 
-    def _check_if_existing_password(self, username, shadow_file=None) -> bool:
-        """
-        Check whether ``username`` user has an existing password (regardless
-        of whether locked or not).
-
-        For OpenBSD (from https://man.openbsd.org/passwd.5) a password field
-        of "" indicates no password, and password field values of either
-        "*" or "*************" (13 "*") indicate differing forms of "locked"
-        but with no password defined.
-
-        Returns either 'True' to indicate a password present, or 'False'
-        for no password set.
-        """
-
-        status = not self._check_if_password_field_matches(
-            username, "::", ":*:", ":*************:", check_file=shadow_file
-        )
-        return status
-
     def lock_passwd(self, name):
         try:
             subp.subp(["usermod", "-p", "*", name])
```

blackboxsw (commented):
@dermotbradley unittest coverage looks good here and the functional changes to support password unlock of empty passwords cover clear use-cases. I think what we may need additionally is an integration test extension to cover pre-existing user with empty or locked password in either tests/integration_tests/modules/test_set_password.py or tests/integration_tests/modules/test_users_groups.py to assert warning condition and not-unlocked user when password is empty.

dermotbradley:
Yupe, I think we're good to go...

PR #5363, opened 2024-06-04
Increase docs font weight

s-makin:
## Proposed Commit Message
```
docs: Adjust CSS to increase font weight across the docs.

A common problem users encounter is that the font weight in the
documentation is too thin/hard to read. 

Fixes GH-4146
```


blackboxsw (approved):
LGTM

PR #5370, opened 2024-06-04
fix: Stop attempting to resize ZFS in cc_growpart on Linux

TheRealFalcon:
## Proposed Commit Message

```
fix: Stop attempting to resize ZFS in cc_growpart on Linux

In e520c944, growing ZFS was added to cc_growpart to support BSD
systems. However, this left Linux in a state where it will fail to
resize and log a warning anytime it encounters ZFS. This commit
ensures that the cc_growpart resize is only attempted on BSD (or any
other system that uses `growfs`). This leaves cc_resizefs to grow the
filesystem on Linux.

This is admittedly a hack solution and long term, we should ensure
that ZFS gets grown in one module, and find a better long-term
abstraction for volume managers.


```

## Merge type

### DIFF
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -503,7 +503,7 @@ def resize_devices(resizer: Resizer, devices, distro: Distro):
         # Ideally we would grow the FS for both OSes in the same module.
         if fs == "zfs" and isinstance(resizer, ResizeGrowFS):
             info += _call_resizer(resizer, devent, disk, ptnum, blockdev, fs)
-            return info
+            continue
 
         try:
             statret = os.stat(blockdev)
--- a/cloudinit/config/cc_growpart.py
+++ b/cloudinit/config/cc_growpart.py
@@ -118,36 +118,6 @@ class RESIZE:
 LOG = logging.getLogger(__name__)
 
 
-def resizer_factory(mode: str, distro: Distro, devices: list):
-    resize_class = None
-    if mode == "auto":
-        for _name, resizer in RESIZERS:
-            cur = resizer(distro)
-            if cur.available(devices=devices):
-                resize_class = cur
-                break
-
-        if not resize_class:
-            raise ValueError("No resizers available")
-
-    else:
-        mmap = {}
-        for k, v in RESIZERS:
-            mmap[k] = v
-
-        if mode not in mmap:
-            raise TypeError("unknown resize mode %s" % mode)
-
-        mclass = mmap[mode](distro)
-        if mclass.available(devices=devices):
-            resize_class = mclass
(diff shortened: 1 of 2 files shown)


cloudinit/config/cc_growpart.py
```diff
@@ -280,6 +250,36 @@ def resize(self, diskdev, partnum, partdev, fs):
         return (before, get_size(partdev, fs))
 
 
+def resizer_factory(mode: str, distro: Distro, devices: list) -> Resizer:
```
TheRealFalcon, line 253: This entire function was moved as-is so to be able to reference the `Resizer` return type

holmanb (commented):
I haven't tested this, but the approach looks good for now.

I agree that it's not ideal to have resize happening in different places on different OSes. I do think that growpart is a slightly more natural fit ZFS resizing, but we can bikeshed that another day.

cloudinit/config/cc_growpart.py
```diff
@@ -496,7 +496,12 @@ def resize_devices(resizer, devices, distro: Distro):
...
         LOG.debug("growpart found fs=%s", fs)
-        if fs == "zfs":
+        # TODO: This seems to be the wrong place for this. On Linux, we the
+        # `os.stat(blockdev)` call below will fail on a ZFS filesystem.
+        # We then delay resizing the FS until calling cc_resizefs. Yet
+        # the code here is to accommodate the FreeBSD `growfs` service.
+        # Ideally we would grow the FS for both OSes in the same module.
+        if fs == "zfs" and isinstance(resizer, ResizeGrowFS):
             info += _call_resizer(resizer, devent, disk, ptnum, blockdev, fs)
             return info
```
blackboxsw, line 506: This should be a continue, not a return in case there are other devices listed in `growpart: devices: [..]`
```suggestion
            continue
```

blackboxsw (changes requested):
This looks good for the time being. It retains BSD support for growfs which is what was intended and doesn't degrade the current skip behavior on all non-BSD/non-growfs environments.

We need to change that early return to a continue. But beyond that +1.


Also we need to fix the resizer class `myresizer` in the following test to subclass from  ResizeGrowFS 
 to avoid failing the unit test now. tests/unittests/config/test_cc_growpart.py::TestResizeZFS::test_zroot[zroot/ROOT/nochange-NOCHANGE]

TheRealFalcon:
Addressed comment from @blackboxsw , updated tests, and manually verified that ZFS systems (both from containers and a mounted disk in a VM) result in no log warnings. Ready for review.

blackboxsw (approved):
Content and behavior look good for this PR. Thanks @TheRealFalcon

PR #5391, opened 2024-06-07
perf(netplan): Improve network v1 -> network v2 performance

holmanb:
## Commit message
```
feat(netplan): Improve network v1 -> network v2 performance

When converting network v1 to network v2, cloud-init might need to
know whether the backend network system supports ipv6 mtu. Previously
cloud-init was always calling `netplan info` when converting from v1
to v2 to know whether this feature is supported. With this change,
cloud-init only calls `netplan info` if it is required to render the
configuration (i.e. if MTU needs to be configured for ipv6).

This code change works by changing a property which always gets
referenced into a function which only gets referenced when it is
needed.

This removes ~0.2s from network render time on all Ubuntu platforms
that require converting from network v1 to network v2.

On a local LXD VM running Ubuntu Noble (24.04), this removes 0.2 seconds
from time to ssh - and cuts cloud-init's reported time in Network stage
down from 0.501 to 0.301.
```

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -6,7 +6,7 @@ import logging
 import os
 import textwrap
 from tempfile import SpooledTemporaryFile
-from typing import Optional, cast
+from typing import Callable, List, Optional, cast
 
 from cloudinit import features, safeyaml, subp, util
 from cloudinit.net import (
@@ -48,7 +48,7 @@ def _get_params_dict_by_match(config, match):
     )
 
 
-def _extract_addresses(config: dict, entry: dict, ifname, features=None):
+def _extract_addresses(config: dict, entry: dict, ifname, features: Callable):
     """This method parse a cloudinit.net.network_state dictionary (config) and
        maps netstate keys/values into a dictionary (entry) to represent
        netplan yaml. (config v1 -> netplan)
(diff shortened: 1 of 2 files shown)


aciba90 (approved):
LGTM, thanks!

cloudinit/net/netplan.py
```diff
@@ -48,7 +48,7 @@ def _get_params_dict_by_match(config, match):
     )
 
 
-def _extract_addresses(config: dict, entry: dict, ifname, features=None):
+def _extract_addresses(config: dict, entry: dict, ifname, features: Callable):
```
TheRealFalcon, line 51: Rather than taking a callable here, whose typing will be fairly opaque, can we instead perform the call on line 476 so that this takes a `List[str]` instead?

TheRealFalcon (commented):
One comment inline. Additionally, on the commit message, "perf" isn't one of the project's conventional commit options.

```
Types used by this project:
feat, fix, docs, ci, test, refactor, chore
```

We can add it, but lets add it to the PR template too then

cloudinit/net/netplan.py
holmanb: This would make us incur the cost of calling `netplan apply` in many cases that we don't need it. I would rather only call it when needed, given how costly this is.

holmanb:
> One comment inline. Additionally, on the commit message, "perf" isn't one of the project's conventional commit options.
> 
> ```
> Types used by this project:
> feat, fix, docs, ci, test, refactor, chore
> ```
> 
> We can add it, but lets add it to the PR template too then

Updated

TheRealFalcon (approved):
LGTM!

PR #5394, opened 2024-06-07
test: Add ds-identify integration test coverage

holmanb:
## Proposed Commit Message
```
test: Add ds-identify integration test coverage

LXD instances don't have a defined datasource_list and therefore
implicitly exercise ds-identify. Add coverage for all other platforms.
```

## Additional Context
Currently LXD integration tests exercise ds-identify, but I don't know of any others that do - just some [unit tests](https://github.com/canonical/cloud-init/blob/main/tests/unittests/test_ds_identify.py).

### DIFF
--- a/tests/integration_tests/test_ds_identify.py
+++ b/tests/integration_tests/test_ds_identify.py
@@ -1,7 +1,5 @@
 """test that ds-identify works as expected"""
 
-import json
-
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.integration_settings import OS_IMAGE, PLATFORM
 from tests.integration_tests.util import verify_clean_log, wait_for_cloud_init
--- a/tests/integration_tests/test_ds_identify.py
+++ b/tests/integration_tests/test_ds_identify.py
@@ -11,6 +11,8 @@ MAP_PLATFORM_TO_DATASOURCE = {
     "lxd_container": "lxd",
     "lxd_vm": "lxd",
     "qemu": "nocloud",
+    "ec2": "aws",
+    "oci": "oracle",
 }
 
 
@@ -28,9 +30,9 @@ def test_ds_identify(client: IntegrationInstance):
     verify_clean_log(client.execute("cat /var/log/cloud-init.log"))
     assert client.execute("cloud-init status --wait")
 
-    datasource = MAP_PLATFORM_TO_DATASOURCE.get(PLATFORM)
+    datasource = MAP_PLATFORM_TO_DATASOURCE.get(PLATFORM, PLATFORM)
     if "lxd" == datasource and "focal" == OS_IMAGE:
         datasource = "nocloud"
-    assert datasource == json.loads(
-        client.execute("cloud-init status --format json").stdout
-    ).get("datasource")
+    cloud_id = client.execute("cloud-id")
+    assert cloud_id.ok
+    assert datasource == cloud_id.stdout.rstrip()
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
We should have an assertion that the found datasource is the expected datasource.

TheRealFalcon (commented):
This won't work on all currently supported platforms (e.g., ec2). Was it your goal for this to run everywhere? If so, lets add the rest. If not, we should skip this test on those platforms.

tests/integration_tests/test_ds_identify.py
```diff
@@ -0,0 +1,38 @@
...
+from tests.integration_tests.integration_settings import OS_IMAGE, PLATFORM
+from tests.integration_tests.util import verify_clean_log, wait_for_cloud_init
+
+DATASOURCE_LIST_FILE = "/etc/cloud/cloud.cfg.d/90_dpkg.cfg"
+MAP_PLATFORM_TO_DATASOURCE = {
+    "lxd_container": "lxd",
+    "lxd_vm": "lxd",
+    "qemu": "nocloud",
+    "ec2": "aws",
+    "oci": "oracle",
```
a-dubs, line 15: I tested that this works for oracle, qemu, as well as ibm
```suggestion
    "oci": "oracle",
    "ibm": "nocloud",
```

tests/integration_tests/test_ds_identify.py
```diff
@@ -0,0 +1,38 @@
...
+    assert client.execute("cloud-init clean --logs")
+    client.restart()
+    wait_for_cloud_init(client)
+    verify_clean_log(client.execute("cat /var/log/cloud-init.log"))
+    assert client.execute("cloud-init status --wait")
+
+    datasource = MAP_PLATFORM_TO_DATASOURCE.get(PLATFORM, PLATFORM)
+    if "lxd" == datasource and "focal" == OS_IMAGE:
+        datasource = "nocloud"
+    cloud_id = client.execute("cloud-id")
```
a-dubs, line 36: is this any functionally different than:
```
json.loads(
        client.execute("cloud-init status --format json").stdout
    ).get("datasource")
```

or is it the same under the hood and just a much cleaner way of doing the same thing?

tests/integration_tests/test_ds_identify.py
holmanb: @a-dubs nice thanks!

I tested on ec2, gce, lxd, and have some educated guesses based on the code that it should work on azure and openstack (but can't test openstack currently).

tests/integration_tests/test_ds_identify.py
a-dubs: Now that i see that you added EC2 back in as "aws" I assume the cloud-id _is_ functionally different? It prevents the issue of the various EC2 datasource names?

holmanb: Yeah, the do pretty different things - "datasource" reports the name of the class that was used without the prefix (on ec2, DataSourceEc2Local gets shortened to "ec2local"). `cloud-id` is more user-facing but can come directly from the cloud's IMDS (and it falls back [to `dsname`](https://docs.cloud-init.io/en/latest/reference/datasource_dsname_map.html)).

TheRealFalcon (approved):
Well this was unexpected...
`FAILED tests/integration_tests/test_ds_identify.py::test_ds_identify - AssertionError: assert 'ibm' == 'nocloud'`

Looks like this is probably a real failure though, so PR LGTM.

PR #5403, opened 2024-06-13
[docs] Add explanation about cloud-config

s-makin:
## Proposed Commit Message

```
docs: Add explanation of cloud-config

- The cloud-config user data file creates a lot of confusion
- Want to clarify what it is, how it's used and how to structure
```

## Merge type

### DIFF
--- a/doc/rtd/explanation/about-cloud-config.rst
+++ b/doc/rtd/explanation/about-cloud-config.rst
@@ -66,18 +66,6 @@ written in any order.
    The :ref:`Users and Groups <mod_cc_users_groups>` module is a special case
    where ordering matters if you want to add users to a group list.
 
-Cloud-config for the live installer
------------------------------------
-
-For the special case where your cloud-config file is will be consumed by the
-Ubuntu installer, you will need to include the the ``autoinstall:``
-top level key. The presence of this key will instruct cloud-init not to process
-the user-data itself, but instead to pass it directly to the installer for
-processing.
-
-For more detailed instructions for this case, refer to the installer
-documentation on using `cloud-init with the autoinstaller`_.
-
 Checking your cloud-config file
 ===============================
 
@@ -161,6 +149,18 @@ examples that can be copy/pasted and adapted to your needs.
      message: Rebooting after initial setup
      timeout: 30
      condition: True
+
+Ubuntu installer cloud-config
+-----------------------------
+
+For the special case where your cloud-config file is will be consumed by the
+Ubuntu live installer, you will need to include the ``autoinstall:`` top level
+key. The presence of this key will instruct cloud-init not to process
+the user-data itself, but instead to pass it directly to the installer for
+processing.
+
+For more detailed instructions for this case, refer to the installer
+documentation on using `cloud-init with the autoinstaller`_.
 
 .. LINKS
 .. _cloud-init with the autoinstaller: https://canonical-subiquity.readthedocs-hosted.com/en/latest/tutorial/providing-autoinstall.html#autoinstall-by-way-of-cloud-config
--- a/doc/rtd/explanation/about-cloud-config.rst
+++ b/doc/rtd/explanation/about-cloud-config.rst
@@ -56,6 +56,11 @@ are the various configuration options for the module shown as key-value pairs:
 A full list of modules can be found :ref:`on our modules page <modules>`. This
 list also shows the valid schema keys for every module, and YAML examples.
 
+.. note::
+   When a module declares no `Activate only on keys:` section, that module will
+   attempt to run regardless of specific schema keys.
+
+
 Module ordering
 ---------------
 
@@ -89,17 +94,6 @@ examples that can be copy/pasted and adapted to your needs.
    hostname: example-host
    fqdn: example-host.example.com
 
-   # Configure storage
-   storage:
-     files:
-       - path: /etc/example_file.txt
-         content: |
-           Some text to be stored in the file
-       - path: /etc/example_script.txt
-         content: |
-           #!/bin/bash
-           echo "Some text to be run in the script"
-
    # User setup configuration
    users:
      - name: exampleuser
@@ -111,9 +105,11 @@ examples that can be copy/pasted and adapted to your needs.
        ssh_authorized_keys:
          - ssh-rsa AAAAB3...restofpublickey user@host
 
-   # Change passwords using chpasswd
+   # Change passwords for exampleuser using chpasswd
    chpasswd:
-     exampleuser: terriblepassword12345
+     expire: false
+     users:
+     - {name: exampleuser, password: terriblepassword12345, type: text}
 
    # Package management
    package_update: true
--- a/doc/rtd/explanation/about-cloud-config.rst
+++ b/doc/rtd/explanation/about-cloud-config.rst
@@ -1,0 +1,85 @@
+.. _about-cloud-config:
+
+About the cloud-config file
+***************************
+
+The ``#cloud-config`` file is a type of user data that cloud-init can consume
+to automatically set up various aspects of the system.
+
+It overrides all other types of user data, so if there is a conflict between
+the config you specify in other :ref:`user data formats <user_data_formats>`,
+the contents of the ``#cloud-config`` file will take precendence.
+
+Note that networks are not configurable via the ``#cloud-config`` file because
+:ref:`network configuration <network_config>` occurs before user data is
+consumed and applied.
+
+How do I create a cloud-config file?
+====================================
+
+The cloud-config file can be written in any valid YAML but the first line
+**must start** with ``#cloud-config``. This line identifies the file to
+cloud-init and ensures that it will be processed as intended.
+
+After the first line, every aspect of the system's configuration is controlled
+through specific cloud-init **modules**. Each module included in the
+``cloud-config`` file can be thought of as a section.
+
+Let us consider the example of the "keyboard" module. The module has a
+corresponding top-level key (``keyboard:``, in this case), and beneath that
+the various configuration parameters are defined:
+
+.. code-block:: yaml
+
+   keyboard:
+     layout: us
+     model: pc105
+     variant: nodeadkeys
+     options: compose:rwin
+
+A full list of modules can be found :ref:`on our modules page <modules>`. This
+list also shows the valid schema keys for every module, and YAML examples.
+
+Module ordering
+---------------
+
+The order of the different module "sections" is mostly unimportant and modules
+can be shown in any order -- except where there are dependencies. For example,
+if you want to create users and also put them in a specific group, then the
+``groups`` section must go before ``users`` so that the groups are created
+first.
+
+Cloud-config for the live installer
+-----------------------------------
+
+For the special case where your cloud-config file is will be consumed by the
+Ubuntu installer, you will need to include the the ``autoinstall:``
+top level key. The presence of this key will instruct cloud-init not to process
+the user-data itself, but instead to pass it directly to the installer for
+processing.
+
+For more detailed instructions for this case, refer to the installer
+documentation on using `cloud-init with the autoinstaller`_.
+
+Checking your cloud-config file
+===============================
+
+Once you have constructed your cloud-config file, you can check it against
+the :ref:`cloud-config validation tool <check_user_data_cloud_config>`. This
+tool tests the YAML in your file against the cloud-init schema and will warn
+you of any errors.
+
+Example cloud-config file
+=========================
+
+The following code is an example of a working user data cloud-config file.
+The :ref:`cloud-config example library <yaml_examples>` contains further
+examples that can be copy/pasted and adapted to your needs.
+
+.. code-block:: yaml
+
+   #cloud-config
+   <put example cloud-config here>
+
+.. LINKS
+.. _cloud-init with the autoinstaller: https://canonical-subiquity.readthedocs-hosted.com/en/latest/tutorial/providing-autoinstall.html#autoinstall-by-way-of-cloud-config
--- a/doc/rtd/explanation/about-cloud-config.rst
+++ b/doc/rtd/explanation/about-cloud-config.rst
@@ -4,33 +4,49 @@ About the cloud-config file
 ***************************
 
 The ``#cloud-config`` file is a type of user data that cloud-init can consume
-to automatically set up various aspects of the system.
-
-It overrides all other types of user data, so if there is a conflict between
-the config you specify in other :ref:`user data formats <user_data_formats>`,
-the contents of the ``#cloud-config`` file will take precendence.
+to automatically set up various aspects of the system. It is the preferred way
+to pass this type of configuration to cloud-init, and is commonly referred to
+as "cloud config". 
 
 Note that networks are not configurable via the ``#cloud-config`` file because
-:ref:`network configuration <network_config>` occurs before user data is
-consumed and applied.
+:ref:`network configuration <network_config>` comes from the cloud.
 
 How do I create a cloud-config file?
 ====================================
 
-The cloud-config file can be written in any valid YAML but the first line
-**must start** with ``#cloud-config``. This line identifies the file to
-cloud-init and ensures that it will be processed as intended.
+The cloud-config file uses `YAML version 1.1`_. The file is composed of a
+**header** and one or more **modules**.
 
-After the first line, every aspect of the system's configuration is controlled
-through specific cloud-init **modules**. Each module included in the
-``cloud-config`` file can be thought of as a section.
+* **The header**:
+  The first line **must start** with ``#cloud-config``. This line identifies
+  the file to cloud-init and ensures that it will be processed as intended.
 
-Let us consider the example of the "keyboard" module. The module has a
-corresponding top-level key (``keyboard:``, in this case), and beneath that
-the various configuration parameters are defined:
+* **The modules**:
+  After the header, every aspect of the system's configuration is controlled
+  through specific cloud-init modules.
+
+Each module is declared through the use of its **top-level key**, and the
+configuration options are set using key-value pairs (or list keys) according
+to the config schema. It follows this general format:
 
 .. code-block:: yaml
 
+   #cloud-config
+   top-level-key:   
+     config-key-1: config-value-1
+     config-key-2: config-value-2
+     list-key:
+     - list-value-1
+       additional-list-value-1
+     - list-value-2
+
+Let us consider a real example using the :ref:`Keyboard <mod_cc_keyboard>`
+module. The top-level key for this module is ``keyboard:``, and beneath that
+are the various configuration options for the module shown as key-value pairs:
+
+.. code-block:: yaml
+
+   #cloud-config
    keyboard:
      layout: us
      model: pc105
@@ -43,11 +59,12 @@ list also shows the valid schema keys for every module, and YAML examples.
 Module ordering
 ---------------
 
-The order of the different module "sections" is mostly unimportant and modules
-can be shown in any order -- except where there are dependencies. For example,
-if you want to create users and also put them in a specific group, then the
-``groups`` section must go before ``users`` so that the groups are created
-first.
+The order of the ``cloud-config`` keys is unimportant and modules can be
+written in any order.
+
+.. note::
+   The :ref:`Users and Groups <mod_cc_users_groups>` module is a special case
+   where ordering matters if you want to add users to a group list.
 
 Cloud-config for the live installer
 -----------------------------------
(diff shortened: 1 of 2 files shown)


doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
+.. _about-cloud-config:
+
+About the cloud-config file
```
s-makin, line 3: Q: Is the title helpful/descriptive enough?

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+=========================
+
+The following code is an example of a working user data cloud-config file.
+The :ref:`cloud-config example library <yaml_examples>` contains further
+examples that can be copy/pasted and adapted to your needs.
+
+.. code-block:: yaml
+
+   #cloud-config
+   <put example cloud-config here>
```
s-makin, line 82: Q: What modules should I include to make it a useful example? Does it matter for an example (i.e., can I use "any", or should it be an actual working file?)

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+can be shown in any order -- except where there are dependencies. For example,
+if you want to create users and also put them in a specific group, then the
+``groups`` section must go before ``users`` so that the groups are created
+first.
+
+Cloud-config for the live installer
+-----------------------------------
+
+For the special case where your cloud-config file is will be consumed by the
+Ubuntu installer, you will need to include the the ``autoinstall:``
```
blackboxsw, line 56:
```suggestion
Ubuntu installer, you will need to include the ``autoinstall:``
```

doc/rtd/explanation/about-cloud-config.rst
blackboxsw: I think it'd be helpful to represent an example with runcmd: write_files and ssh_authorized_keys:  chpasswd as they seem to be the most-searched keys in documentation. We can provide a working example that we can document here to detail the operations performed and expected config as a result.

doc/rtd/explanation/about-cloud-config.rst
blackboxsw: Looks good enough as this is a general topic that covers how #cloud-config files are interpreted.

blackboxsw (commented):
First pass on this looks pretty good @s-makin. I think we may have a bit of back and forth on the module ordering and inter-dependency discussion, because the user/groups ordering is sort of a special case where ordering actually matters. Every other module is strictly ordered in cloud.cfg to make sure it deals with most cross-module dependencies that may exist to ensure all dependencies are satisfied before starting a new config module. The only unique cases we may have are within a single module, such as `user_groups` (and I'm sure there may be other minor corner cases in other modules) where some hard-coded operations may require that something else is already present.    Other cross-module dependencies we may look to call out could be things like:
  - bootcmd can't expect to reference utilities installed by `packages:`  because `packages` module is run later that bootcmd.   Or
  -  default `write_files` ordering can't write to a directory that happens to be installed and created packages installed by 'packages:` , so `write_files:` grew a `deferred: true` option which allows the file creation to happen during the cloud.cfg `write_files_deferred` module position during `cloud_final_modules`. 


I'll follow up tomorrow with a good cloud-config  example to include

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
+.. _about-cloud-config:
+
+About the cloud-config file
+***************************
+
+The ``#cloud-config`` file is a type of user data that cloud-init can consume
+to automatically set up various aspects of the system.
+
+It overrides all other types of user data, so if there is a conflict between
```
TheRealFalcon, line 9: I don't understand what this means. User data takes the form of a single file/blob so I don't understand how we could have conflicts.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+the config you specify in other :ref:`user data formats <user_data_formats>`,
+the contents of the ``#cloud-config`` file will take precendence.
+
+Note that networks are not configurable via the ``#cloud-config`` file because
+:ref:`network configuration <network_config>` occurs before user data is
```
TheRealFalcon, line 14: This isn't really true. The main reason is that network configuration comes from the cloud and applying a different configuration could make the instance unreachable.

In most cases we retrieve the network config at the same time as the cloud-config and we *could* extract network-config at that time if we wanted to. We've floating the idea of supporting it a number of times.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+list also shows the valid schema keys for every module, and YAML examples.
+
+Module ordering
+---------------
+
+The order of the different module "sections" is mostly unimportant and modules
+can be shown in any order -- except where there are dependencies. For example,
+if you want to create users and also put them in a specific group, then the
+``groups`` section must go before ``users`` so that the groups are created
+first.
```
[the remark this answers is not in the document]
TheRealFalcon: To demonstrate the issue:
This config:
```
#cloud-config
users:
  - name: chad
  - name: sally
    groups: sallygroup
groups:
 - sallygroup
 - mygroup: [chad, sally]
```
will be translated exactly the same as this config:
```
#cloud-config
groups:
 - sallygroup
 - mygroup: [chad, sally]
users:
  - name: chad
  - name: sally
    groups: sallygroup
```

Using the `users` key I can create the `chad` and `sally` users and even add a user to a particular group. However, regardless of how I order the cloud-config,
```
groups:
 - mygroup: [chad, sally]
```
will fail regardless of the yaml ordering if the `chad` and `sally` users didn't already exist on the system, because the code that creates the groups runs before the code that creates the users.

This is a quirk of the Users/Groups module that I don't think should be mentioned in this part of the documentation.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Cloud-config for the live installer
```
TheRealFalcon, line 52: I would replace `live` with `Ubuntu`.

I don't love that we're so prominently calling out subiquity in upstream documentation that should be distro independent. Tbh, I think buried in https://cloudinit.readthedocs.io/en/latest/reference/faq.html#autoinstall-preruncmd-postruncmd is the right place for it.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+through specific cloud-init **modules**. Each module included in the
+``cloud-config`` file can be thought of as a section.
+
+Let us consider the example of the "keyboard" module. The module has a
+corresponding top-level key (``keyboard:``, in this case), and beneath that
+the various configuration parameters are defined:
+
+.. code-block:: yaml
+
+   keyboard:
```
TheRealFalcon, line 34: Lets add the `#cloud-config` header here.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+     variant: nodeadkeys
+     options: compose:rwin
+
+A full list of modules can be found :ref:`on our modules page <modules>`. This
+list also shows the valid schema keys for every module, and YAML examples.
+
+Module ordering
+---------------
+
+The order of the different module "sections" is mostly unimportant and modules
```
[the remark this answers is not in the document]
TheRealFalcon: Chad makes a good point. I think the issue here is that there's really two different concepts that need to be explained independently. The first is the `#cloud-config`, which is the declarative yaml used to describe how to setup your system, but the second is how that config gets translated, which is a bit more "messy" in the these various modules run a specified order and are not idempotent. `/etc/cloud/cloud.cfg` is a good way to see the module ordering, but I'm hesitant to call it out because we do not support reordering/removing modules, and showing the config will show users the place to reorder/remove modules.

TheRealFalcon (changes requested):
Comments inline

doc/rtd/explanation/about-cloud-config.rst
s-makin: Good catch!

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+How do I create a cloud-config file?
+====================================
+
+The cloud-config file can be written in any valid YAML but the first line
+**must start** with ``#cloud-config``. This line identifies the file to
+cloud-init and ensures that it will be processed as intended.
+
+After the first line, every aspect of the system's configuration is controlled
+through specific cloud-init **modules**. Each module included in the
+``cloud-config`` file can be thought of as a section.
```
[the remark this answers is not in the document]
s-makin: I think one of the reasons this is hard to document is that we use a lot of terminology quite loosely/interchangeably as a shorthand without it being rigorously defined. What's the difference between a top level key and a top level config module?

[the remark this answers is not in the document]
s-makin: Nevermind, I think I answered my own question :upside_down_face:

doc/rtd/explanation/about-cloud-config.rst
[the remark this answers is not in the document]
s-makin: So, if the modules are run in a set order that's handled behind the scenes, why do we not just run `groups` before `users` so that we don't have this special case?

doc/rtd/explanation/about-cloud-config.rst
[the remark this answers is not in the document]
s-makin: Understood. Since this is the only case where ordering matters, I've taken out the discussion entirely and instead put a note to call it out specifically with a link to the module. Thanks for the clarification! (I will probably put this into the explanation stuff about creating users and groups, because I think it will help illustrate the quirk nicely - but I agree, this is not the place for it)

doc/rtd/explanation/about-cloud-config.rst
s-makin: Good catch!

doc/rtd/explanation/about-cloud-config.rst
s-makin: Perhaps, but I'm trying to slim down the FAQ so I'd rather not put it there. Since it does represent a frequently asked question (or at least, something that does cause a lot of problems/consternation) I'm actually OK with putting it in the docs - although I'm also happy to have it at the end of the file so that it's less prominent, if you think that would be better?

doc/rtd/explanation/about-cloud-config.rst
s-makin: I've added an example using those modules (plus some extra ones). I'm happy for us to add some more explanation to the inline comments but I don't know that we necessarily want to add a huge amount of explanation here, I'd rather save that for the example library. What I think we should have here is an illustrative example that shows how a reasonably complete cloud-config file *could* look. Let me know if you think this example suits that purpose (or any other tweaks you'd like me to make to it)

doc/rtd/explanation/about-cloud-config.rst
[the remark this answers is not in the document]
s-makin: I think it's right not to go into that much detail on a page that's meant to be a friendly intro to the file and how it should be constructed. We may want to have some sort of an explanation of the translation, but it seems to me like an implementation detail that belongs more rightly in the dev docs. Wdyt?

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,85 @@
...
+     options: compose:rwin
+
+A full list of modules can be found :ref:`on our modules page <modules>`. This
+list also shows the valid schema keys for every module, and YAML examples.
+
+Module ordering
+---------------
+
+The order of the different module "sections" is mostly unimportant and modules
+can be shown in any order -- except where there are dependencies. For example,
```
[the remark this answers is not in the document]
s-makin: Nice suggestion, but I've sidestepped it by calling out the users/groups as a special case in a note. Let me know if that's better :)

doc/rtd/explanation/about-cloud-config.rst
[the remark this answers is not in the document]
blackboxsw: no prob. to clarify, a config module in cloud-init can take action based on the presence of `0 to n` top-level cloud-config keys. Each module defines the cloud-config keys that will activate this module as documented in the module reference page in the `Activate only on keys: ansible` metadata associated with that module. Some examples:
- The [Bootcmd module is only run if the top-level key `bootcmd` is present.](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#bootcmd)
- The  [Disk setup module is run if either ``disk_setup`` or ``fs_setup`` are present](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#disk-setup)
- The [Locale config module does not define a limited activate on schema keys, so it will be run whether or not a `locale` key is present in cloud-config or not](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#locale)

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,167 @@
...
+list also shows the valid schema keys for every module, and YAML examples.
+
+Module ordering
+---------------
+
+The order of the ``cloud-config`` keys is unimportant and modules can be
+written in any order.
+
+.. note::
+   The :ref:`Users and Groups <mod_cc_users_groups>` module is a special case
```
holmanb, line 66: Do we need this note?

There are other examples where ordering matters too - any place in the schema where a list is used rather than a k/v pair is likely to have an ordering effect.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,167 @@
...
+     timeout: 30
+     condition: True
+
+Ubuntu installer cloud-config
+-----------------------------
+
+For the special case where your cloud-config file will be consumed by the
+Ubuntu live installer, you will need to include the ``autoinstall:`` top level
+key. The presence of this key instructs cloud-init not to process
+the user-data itself, but instead to pass it directly to the installer for
```
holmanb, line 159: > The presence of this key instructs cloud-init not to process
the user-data itself, but instead to pass it directly to the installer for
processing.

That's not true, is it? The presence of this key doesn't prevent cloud-init from processing the user-data.

Cloud-init just ignores keys nested under the `autoinstall` key.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,167 @@
...
+   final_message: "The system is up, after $UPTIME seconds"
+
+   # Reboot the instance after configuration
+   power_state:
+     mode: reboot
+     message: Rebooting after initial setup
+     timeout: 30
+     condition: True
+
+Ubuntu installer cloud-config
```
holmanb, line 153: I don't think that this section makes sense in cloud-init's documentation about cloud-config. Cloud-init doesn't do anything special when autoinstall is included, it is just ignored (and subiquity does whatever it does with it later).

holmanb (changes requested):

blackboxsw (commented):
Had some inline comments and suggestions about user-data that needs a fix here. Pushed them as a separate commit for review.

doc/rtd/explanation/about-cloud-config.rst
blackboxsw: A counterpoint here is that we are getting lots of questions in IRC in cloud-init about this and a lot of keyword searches `past 30 days: autoinstall (1 result) 136 searches`.   I'd rather have a link handy in our docs that says, not our deal so that hopefully it is discoverable before people ask in #cloud-init channel in IRC.

doc/rtd/explanation/about-cloud-config.rst
blackboxsw: good pt. cloud-init just ignores it, it doesn't take any action to "pass it to the installer"

doc/rtd/explanation/about-cloud-config.rst
holmanb: > a lot of keyword searches past 30 days: autoinstall (1 result) 136 searches. I'd rather have a link handy in our docs that says, not our deal so that hopefully it is discoverable before people ask in #cloud-init channel in IRC.

Why is is this page the right place for it? There are many things that one could say about cloud-config, but there is value in keeping it short and on-topic without cluttering it with random details. Also, go search autoinstall in the docs. [I get exactly one hit](https://docs.cloud-init.io/en/latest/reference/faq.html#autoinstall-preruncmd-postruncmd). We don't need to say "not our deal" in two places, so I'd rather not add another unless we're planning on removing the existing one - and if we're going to relocate it lets put it where it belongs rather than from one section to the next. Your points make it sound like it is a frequently asked question, which is exactly the purpose of the page that that section is currently on.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,163 @@
+.. _about-cloud-config:
+
+About the cloud-config file
+***************************
+
+The ``#cloud-config`` file is a type of user data that cloud-init can consume
+to automatically set up various aspects of the system. It is the preferred way
```
TheRealFalcon, line 7: Preferred by who? I don't think the answer is "users" as I remember a previous maintainer telling me that the vast majority of user data is just simple bash scripts. We could poll the maintainers, but I don't have any particular preference for cloud-config.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,163 @@
...
+
+* **The header**:
+  The first line **must start** with ``#cloud-config``. This line identifies
+  the file to cloud-init and ensures that it will be processed as intended.
+
+* **The modules**:
+  After the header, every aspect of the system's configuration is controlled
+  through specific cloud-init modules.
+
+Each module is declared through the use of its **top-level key**, and the
```
TheRealFalcon, line 28: It would be nice if we were this consistent, but not all modules have a single top-level key. E.g., the [set passwords](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#set-passwords) module reacts to the `password`, `chpasswd`, and `ssh_pwauth` keys.

Additionally, there are multiple modules that may react to the same key. E.g., `write_files` key is used by both the `write_files` and `write_files_deferred` module and `hostname` and `fqdn` keys are used by both the `set_hostname` and `update_hostname` modules.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,163 @@
...
+
+.. note::
+   When a module declares no `Activate only on keys:` section, that module will
+   attempt to run regardless of specific schema keys.
+
+
+Module ordering
+---------------
+
+The order of the ``cloud-config`` keys is unimportant and modules can be
```
TheRealFalcon, line 67: Similar to what I commented above, keys don't correspond 1:1 with modules, so I don't think it makes sense to talk about the cloud-config as if we're writing or specifying modules.

doc/rtd/explanation/about-cloud-config.rst
```diff
@@ -0,0 +1,167 @@
...
+       - path: /etc/example_script.txt
+         content: |
+           #!/bin/bash
+           echo "Some text to be run in the script"
+
+   # User setup configuration
+   users:
+     - name: exampleuser
+       gecos: Example User
+       sudo: ['ALL=(ALL) NOPASSWD:ALL']
```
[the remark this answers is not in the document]
TheRealFalcon: @blackboxsw , we originally missed this case. This should be valid as of https://github.com/canonical/cloud-init/pull/5418 .

doc/rtd/explanation/about-cloud-config.rst
TheRealFalcon: I agree that this doesn't belong here. I made a similar comment on an earlier version of this PR and would prefer to keep this info buried in the FAQ, but FAQs don't fit nicely into diataxis and are seen by some as an indication of bad documentation.

Could we keep it in the FAQ for now and perhaps grow a "Projects that use cloud-init" or similar page that point to other well known (and not just Canonical) projects that use cloud-init along with important things to be aware of?

TheRealFalcon (commented):
A few more comments inline

doc/rtd/explanation/about-cloud-config.rst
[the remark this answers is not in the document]
blackboxsw: Yep correct. backing down here.

PR #5414, opened 2024-06-19
Add trace logging and refactor logs.py

TheRealFalcon:
## Proposed Commit Message

```
See individual commits
```

## Additional Context
I have functional changes coming, but since the non-functional stuff grew a bit bigger than I was expecting, I decided to separate it as its own PR for easier review.

I'm also open to squashing a few of these if that makes more sense, though the tracing should probably remain separate.

## Merge type

### DIFF
--- a/cloudinit/cmd/devel/logs.py
+++ b/cloudinit/cmd/devel/logs.py
@@ -7,14 +7,17 @@
 """Define 'collect-logs' utility and handler to include in cloud-init cmd."""
 
 import argparse
+import logging
 import os
+import pathlib
 import shutil
 import subprocess
 import sys
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import NamedTuple, Optional
+from typing import NamedTuple, Optional, cast
 
+from cloudinit import log
 from cloudinit.cmd.devel import read_cfg_paths
 from cloudinit.stages import Init
 from cloudinit.subp import ProcessExecutionError, subp
@@ -26,6 +29,8 @@ from cloudinit.util import (
     get_config_logfiles,
     write_file,
 )
+
+LOG = cast(log.CustomLoggerType, logging.getLogger(__name__))
 
 
 class LogPaths(NamedTuple):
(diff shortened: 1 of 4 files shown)


a-dubs:
So with these changes, anything logged at TRACE level will only be logged to `/var/log/cloud-init.log` but only if log level is set to TRACE, which by default it is not? How can the log level be set to trace? Via user-data?

a-dubs:
Also, what purpose does deprecated logging serve? I see that currently it has not been called in any source code, only in a unit test. So is this for logging messages that tells the user that something they did is deprecated? Or is it some weird thing where we use this to "deprecate" a log call that was previously output as debug or info or whatever?

a-dubs:
Also good work on adding type hints to the functions you modified :D

holmanb (changes requested):
Thanks for this @TheRealFalcon. For the most part it looks good to me, just a couple of questions and nits inline.

cloudinit/cmd/devel/logs.py
```diff
@@ -8,14 +8,13 @@
 
 import argparse
 import logging
-import os
 import pathlib
+import os
```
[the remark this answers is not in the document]
TheRealFalcon: What part? The import logging? It's used further below now.

cloudinit/cmd/devel/logs.py
```diff
@@ -8,8 +8,8 @@
 
 import argparse
 import logging
-import pathlib
 import os
+import pathlib
```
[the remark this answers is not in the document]
TheRealFalcon: It got introduced a commit too early, but this gets used a number of times in this module now.

cloudinit/cmd/devel/logs.py
```diff
@@ -157,35 +156,42 @@ def _get_copytree_ignore_files(paths: LogPaths):
...
+    msg: str,
+) -> Optional[str]:
     """Helper which runs a command and writes output or error to filename."""
-    ensure_dir(os.path.dirname(filename))
+    file_path.parent.mkdir(parents=True, exist_ok=True)
     try:
         output = subp(cmd).stdout
-    except ProcessExecutionError as e:
-        write_file(filename, str(e))
+    except (ProcessExecutionError, subprocess.CalledProcessError) as e:
```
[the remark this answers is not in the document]
TheRealFalcon: Whoops...thanks, at one point I was going to join the `_write_command_output_to_file` and `_stream_command_output_to_file` functions, but they were still different enough that they still made sense separate.

cloudinit/cmd/devel/logs.py
```diff
@@ -152,111 +157,103 @@ def _get_copytree_ignore_files(paths: LogPaths):
...
 
-def _write_command_output_to_file(cmd, filename, msg, verbosity):
+def _write_command_output_to_file(cmd, filename, msg):
     """Helper which runs a command and writes output or error to filename."""
     ensure_dir(os.path.dirname(filename))
     try:
         output = subp(cmd).stdout
     except ProcessExecutionError as e:
         write_file(filename, str(e))
-        _debug("collecting %s failed.\n" % msg, 1, verbosity)
```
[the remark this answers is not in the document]
TheRealFalcon: No, see the `_setup_logger` function near the bottom of this file. It should function exactly the same as it did before.

cloudinit/cmd/devel/logs.py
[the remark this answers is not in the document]
holmanb: I commented on the commit which relocated the import from one line to another

cloudinit/cmd/devel/logs.py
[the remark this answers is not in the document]
holmanb: +1 I see now, thanks. It might be nice to be able to re-use this functionality in other parts of the commandline code.

TheRealFalcon:
@holmanb  I know there's still a merge conflict, but I believe I addressed all your issues.

cloudinit/log.py
```diff
@@ -45,14 +62,20 @@ def flush_loggers(root):
...
 
-    def deprecated(self, message, *args, **kwargs):
-        if self.isEnabledFor(lvl):
-            self._log(lvl, message, args, **kwargs)
+        return log_at_level
 
-    logging.Logger.deprecated = deprecated
+    logging.addLevelName(DEPRECATED, "DEPRECATED")
+    logging.addLevelName(TRACE, "TRACE")
+    logging.Logger.deprecated = new_logger(DEPRECATED)  # type: ignore
```
holmanb, line 77: why not `setattr` instead of `#type: ignore`?

TheRealFalcon: Done

TheRealFalcon:
Rebased to fix merge conflict.

cloudinit/cmd/devel/logs.py
```diff
@@ -312,25 +302,37 @@ def collect_logs(
...
+    log.reset_logging()
+    if verbosity == 0:
+        level = logging.INFO
+    elif verbosity == 1:
+        level = logging.DEBUG
+    else:
+        level = log.TRACE
+    LOG.setLevel(level)
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("%(message)s"))
```
holmanb, line 333: I prefer this a lot over `sys.stderr.write()` all over the place. Are you thinking about replacing manual stderr writes with a StreamHandler in other parts of cloud-init?

cloudinit/cmd/devel/logs.py
```diff
@@ -27,6 +29,8 @@
     write_file,
 )
 
+LOG = cast(log.CustomLoggerType, logging.getLogger(__name__))
```
holmanb, line 32: I don't love using `cast()` - since it can hide real problems, but it seems that mypy doesn't have a better way to introspect custom log levels in the logger singleton from `getLogger()`, so I'm fine with this. Maybe we can find a better way someday.

cloudinit/log.py
```diff
@@ -23,6 +23,23 @@
...
+class CustomLoggerType(logging.Logger):
+    """A hack to get mypy to stop complaining about custom logging methods.
+
+    When using deprecated or trace logging, rather than:
+        LOG = logging.getLogger(__name__)
+    Instead do:
+        LOG = cast(CustomLoggerType, logging.getLogger(__name__))
+    """
+
+    def trace(self, *args, **kwargs):
```
holmanb, line 38: Think we should document our custom loggers in the dev docs?

I think that describing when one should choose debug vs choose trace would help guide the project to use these in standard ways.

holmanb (approved):
I have a couple of followup questions inline, but the code looks good to merge as is.

cloudinit/cmd/devel/logs.py
TheRealFalcon: I wasn't thinking of that. I honestly don't know when or why we log to stderr in other places so I'd have to grok that first to know if something like this is a suitable alternative.

cloudinit/cmd/devel/logs.py
TheRealFalcon: Agreed. I spent a while looking for a better alternative, but I think this is the best we have atm.

cloudinit/log.py
TheRealFalcon: Good idea. Since we have a docs sprint coming up, I filed https://github.com/canonical/cloud-init/issues/5525 and hopefully we can address it then.

cloudinit/cmd/devel/logs.py
holmanb: > I wasn't thinking of that, but it sounds like a good idea. I honestly don't know when or why we log to stderr in other places so I'd have to grok that first to know if something like this is a suitable alternative.


iirc the places that I've seen them besides in under `cmd/` usually had comments about it existing as a fallback to ensure errors get to a serial console in case the log file fails to get written - the reasonable thing to do in error paths - just not using the best interfaces imo

PR #5415, opened 2024-06-19
fix(pytest): Fix broken pytest gdb flag

holmanb:
## Proposed Commit Message
```
fix(pytest): Fix broken pytest gdb flag

When pyest is called with --pdb, some tests fail. The test failure
appears to be due to unittest mocking subp prior to pytest fixtures
being created, which causes autospec to fail on subp, since subp is
already mocked. Fix it by converting these tests to use pytest.
```

## Test Steps

The following:
```
.tox/py3/bin/python -m pytest -vvvv --showlocals --pdb
```
now passes

### DIFF
--- a/tests/unittests/config/test_cc_disk_setup.py
+++ b/tests/unittests/config/test_cc_disk_setup.py
@@ -1,4 +1,5 @@
 # This file is part of cloud-init. See LICENSE file for license information.
+# pylint: disable=attribute-defined-outside-init
 
 import random
 import tempfile
--- a/tests/unittests/config/test_cc_disk_setup.py
+++ b/tests/unittests/config/test_cc_disk_setup.py
@@ -14,15 +14,13 @@ from cloudinit.config.schema import (
 from tests.unittests.helpers import (
     CiTestCase,
     ExitStack,
-    TestCase,
     mock,
     skipUnlessJsonSchema,
 )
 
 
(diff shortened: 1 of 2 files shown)


a-dubs:
Since you've already touched a bunch of functions with this PR, would it be a good idea to add docstrings and type hinting to at least some of the functions while you're at it? Or does that just add unnecessary overhead and make landing these changes harder than it needs to be?

TheRealFalcon (approved):
LGTM!

> The test failure appears to be due to unittest mocking subp prior to pytest fixtures
being created, which causes autospec to fail on subp, since subp is
already mocked. 

I ran into some weirdness that I couldn't figure out when trying to optimize how we apply subp mocking. I'm hoping this was the cause of it.

holmanb:
> Since you've already touched a bunch of functions with this PR, would it be a good idea to add docstrings and type hinting to at least some of the functions while you're at it? Or does that just add unnecessary overhead and make landing these changes harder than it needs to be?

@a-dubs good sentiment, and usually I would try to do that but since these are just test functions for now I'm going to be lazy and merge it since I'm trying to wrap up some other things for now. Thanks for pointing this out!

> I ran into some weirdness that I couldn't figure out when trying to optimize how we apply subp mocking. I'm hoping this was the cause of it.

+1 agreed

TheRealFalcon (approved):

Issue #5399, opened 2024-06-11
[enhancement]: Schema validator run from `main_init` can produce confusing error messages


ani-sinha:
# Enhancement
We have a customer who has raised some serious concerns over confusing warning messages produced by the schema validator that is run from `main_init` in stage 7 in the following hunk of code :

```
  # Validate user-data adheres to schema definition                                                                                                                                             
    cloud_cfg_path = init.paths.get_ipath_cur("cloud_config")
    if os.path.exists(cloud_cfg_path) and os.stat(cloud_cfg_path).st_size != 0:
        validate_cloudconfig_schema(
            config=yaml.safe_load(util.load_text_file(cloud_cfg_path)),
            strict=False,
            log_details=False,
            log_deprecations=True,
        )
    else:
        LOG.debug("Skipping user-data validation. No user-data found.")
```
For example, with the following configuration:

```
#cloud-config

users:
  - name: osadmin
    lock_passwd: false
    sudo: ["ALL=(ALL) NOPASSWD:ALL"]
    ssh-authorized-keys:
      - ssh-rsa "AAAAB3NzaC1yc2EAAAADAQABAAABAQCc/K1T62elOkJKc94sWovn06RE7D274Fxx5n+eT/oClAwk7632awMtWjO+lIRpzOlEcYCR6wexlmL9mZNEFJoROMMRyCxRilNAiDKOq0X5j1HXY2ky9KjtLxo2b8yXFRBcYKK4W8aHU94NkNek81jdi9lC+L6jUrZ25d7xylqnEbd4TgrBphiowdh5B3RY+j6ePthLWWhHopRcVDgq91yYacelCCIfSDaGBya9iXPZSwoKtPin4n5PdSGmSOA3fIkvV5JjLS1QbuY5tzhXKuH9MdAqrvutF1bprs/GDf0IY8DO7DDoo8rXjhhgGtCdg1i5UitZIAM3hn/Vk5F//5eJ JOOS8312@EB-OR6120158"
 ```

raises the schema errors on the lines:
(1) `sudo: ["ALL=(ALL) NOPASSWD:ALL"]` and 
(2) `ssh-authorized-keys:`

and this will generate a warning message while starting cloud-init :

```
[   13.764433] cloud-init[973]: 2024-05-13 08:36:30,490 - schema.py[WARNING]: Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.
```

The schema validator complains because the schema definition says
```
[...]

PR #5417, opened 2024-06-19
fix(jsonschema): Add missing sudo definition

holmanb:
```
fix(jsonschema): Add missing sudo definition

This configuration:

users:
  - name: osadmin
    lock_passwd: false
    sudo: ["ALL=(ALL) NOPASSWD:ALL"]

Is valid syntax but is missing from the jsonschema definition.

Fixes GH-5399
```

https://github.com/canonical/cloud-init/issues/5399

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -405,6 +405,15 @@
               "description": "Sudo rule to use or false. Absence of a sudo value or ``null`` will result in no sudo rules added for this user."
             },
             {
+              "type": "array",
+              "items": {
+                "type": [
+                  "string",
+                  "null"
+                ]
+              }
+            },
+            {
               "type": "boolean",
               "changed": true,
               "changed_version": "22.2",


TheRealFalcon (approved):
LGTM!

ani-sinha:
In that case, I would recommend that we also update the doc to specify valid values of `sudo` property here: https://cloudinit.readthedocs.io/en/latest/reference/modules.html#users-and-groups or if that is not quite acceptable, add an explicit example of this format here: https://cloudinit.readthedocs.io/en/latest/reference/examples.html#yaml-examples and update the following lines
```
#   sudo: Defaults to none. Accepts a sudo rule string, a list of sudo rule
#         strings or False to explicitly deny sudo usage. Examples:
#
#         Allow a user unrestricted sudo access.
#             sudo:  ALL=(ALL) NOPASSWD:ALL
#
#         Adding multiple sudo rule strings.
#             sudo:
#               - ALL=(ALL) NOPASSWD:/bin/mysql
#               - ALL=(ALL) ALL
#
#         Prevent sudo access for a user.
#             sudo: False
```

ani-sinha:
> In that case, I would recommend that we also update the doc to specify valid values of `sudo` property here: https://cloudinit.readthedocs.io/en/latest/reference/modules.html#users-and-groups or if that is not quite acceptable, add an explicit example of this format here: https://cloudinit.readthedocs.io/en/latest/reference/examples.html#yaml-examples and update the following lines
> 
> ```
> #   sudo: Defaults to none. Accepts a sudo rule string, a list of sudo rule
> #         strings or False to explicitly deny sudo usage. Examples:
> #
> #         Allow a user unrestricted sudo access.
> #             sudo:  ALL=(ALL) NOPASSWD:ALL
> #
> #         Adding multiple sudo rule strings.
> #             sudo:
> #               - ALL=(ALL) NOPASSWD:/bin/mysql
> #               - ALL=(ALL) ALL
> #
> #         Prevent sudo access for a user.
> #             sudo: False
> ```

I have tried to update the doc https://github.com/canonical/cloud-init/pull/5418 . The doc update applies on top of this fix otherwise with just the doc update, unit tests will fail.

ani-sinha (approved):

holmanb:
Thanks @ani-sinha! I'll close this PR in favor of #5418

PR #5418, opened 2024-06-20
doc: update examples to reflect alternative ways to provide `sudo` option

ani-sinha:
For creating users and groups, it is possible to pass a `sudo` option to the
 config file that accepts a sudo rule. The option can be a sudo rule string,
 a list of sudo rule strings or `False` to explicitly deny sudo usage. Update
 examples to show how a list of strings can be used with `sudo` option.

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -405,6 +405,15 @@
               "description": "Sudo rule to use or false. Absence of a sudo value or ``null`` will result in no sudo rules added for this user."
             },
             {
+              "type": "array",
+              "items": {
+                "type": [
+                  "string",
+                  "null"
+                ]
+              }
+            },
+            {
               "type": "boolean",
               "changed": true,
               "changed_version": "22.2",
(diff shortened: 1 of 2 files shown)


holmanb (approved):
Looks great, thanks!

ani-sinha:
@holmanb there is also a broad issue. The schema validator complains that `ssh-authorized-keys` (and likely other parameters with a `-` instead of `_` are invalid. Yet, the `ug_util.py` has this logic
```
    # Ensure user options are in the right python friendly format                                                                                                                                 
    if users:
        c_users = {}
        for uname, uconfig in users.items():
            c_uconfig = {}
            for k, v in uconfig.items():
                k = k.replace("-", "_").strip()
                if k:
                    c_uconfig[k] = v
            c_users[uname] = c_uconfig
        users = c_users
```
Thus `-` work. So what to do about this? The schema validator can ignore differences between names with`-` and names with `_` with a warning to place it to canonical form.

holmanb:
@ani-sinha I nearly missed this comment. Please open a new issue referencing the closed PR or issue in the future so that things like this don't get missed.

> The schema validator complains that ssh-authorized-keys (and likely other parameters with a - instead of _ ) are invalid

Since `ssh-authorized-keys` [is deprecated](https://github.com/canonical/cloud-init/blob/8c396a4bbc34db1133d6aa52404cffef4a7db605/cloudinit/config/schemas/schema-cloud-config-v1.json#L364), I think that it is doing the right thing here (I'd have to see an exact config in order to know, maybe I missed something). However, you're right that this code means that we're missing a couple of keys that technically work but aren't represented in the schema. The ones that I think need a schema entry are: `plain_text_passwd`, `create_groups`, `selinux_user`, `primary_group`, `hashed_passwd`, `no_log_init`, `no_create_home`, `no_user_group`.

PR #5420, opened 2024-06-20
fix(azure): fix discrepancy for monotonic() vs time()

cjp256:
Transient failures in fetching metadata from IMDS would not retry. get_metadata_from_imds() is now using time.monotonic() for measuring time instead of time.time() causing the deadline check to fail on first check (monotonic() values will always be less than time()).

Update azure.imds.ReadUrlRetryHandler to use time.monotonic() instead of time.time().

### DIFF
--- a/cloudinit/sources/azure/imds.py
+++ b/cloudinit/sources/azure/imds.py
@@ -4,7 +4,7 @@
 
 import logging
 import uuid
-from time import time
+from time import monotonic
 from typing import Dict, Optional, Type, Union
 
 import requests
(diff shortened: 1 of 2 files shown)


cjp256:
PR currently under test.

a-dubs:
Launchpad is angy. Seems like the integration test failed on importing a user from launchpad.

a-dubs (approved):
Changes look good!

holmanb (approved):
@cjp256 Thanks for catching this!

It looks I broke this in https://github.com/canonical/cloud-init/commit/9b1b2944018dc5a49d42535d784621e2b7e5c667. I just went back through and double checked the other uses of `time.time()`, I think that the remaining ones (after a60a705aa09d8ff03f3cf6ad59b733f3b81d7c9b) are warranted.

> PR currently under test.

Want me to hold off on merging it? This looks good to me.

cjp256:
Thanks @holmanb, testing is done on our end and things look great in advance for the tag.

PR #5423, opened 2024-06-20
fix: Use monotonic time

holmanb:
```
fix: Use monotonic time

Prevent time change races.
```


a-dubs:
Why is this change only needed in these two package related source files?

holmanb:
> Why is this change only needed in these two package related source files?

After https://github.com/canonical/cloud-init/pull/5420 lands, I *think* these are the only remaining spots where `time()` was used to diff against `time()` in cloud-init. There are a few spots in cloud-init where `time()` compares against system uptime, which could also be problematic, but those are harder to solve.

a-dubs (changes requested):
Any chance you want to add typehints and a docstring to `_fire_reboot`? :P

TheRealFalcon (approved):
LGTM!

PR #5428, opened 2024-06-21
test: bump pycloudlib to get azure oracular images

blackboxsw:


TheRealFalcon (approved):
I think we can remove the `>=6.7.0` as it's probably supposed to be `>=1!6.7.0`, but this works too.

PR #5431, opened 2024-06-23
Ubuntu/mantic: new_upstream_snapshot, drop quilt patch to fix daily build recipe

blackboxsw:
# New upstream snapshot to fix daily builds

# context
upstream commit 4981ea9a5ba3b5b267f1c4862501634adfc7f55eq introduced drift in system ordering that caused ubuntu downstream quilt patches to fail to apply.
build recipe failure: https://launchpadlibrarian.net/736166492/buildlog.txt.gz

# procedure to create this branch:
```
# Prepare for snapshot
git refresh upstream
git checkout upstream/ubuntu/mantic -B ubuntu/mantic

# First drop the now unnecessary do-not-block-user-login.patch because upstream aligns with this systemd ordering now
git rm debian/patches/do-not-block-user-login.patch
vi debian/patches/series # redact the do-not-block-user-login.patch line from series
git commit -am 'drop: d/p/do-not-block-user-login: upstream has cloud-init.service Before=systemd-user-sessions'

new_upstream_snapshot.py   # expect quilt failure
quilt push
quilt push -f
vi systemd/cloud0conif.service.tmpl   (add snapd.seeded.service to both After and Wants lines)
quilt refresh
quilt pop -a
git commit -am 'refresh patches'
new_upstream_snapshot.py -p quilt
# Then manually edit d/changelog to include dropping d/p/do-not-block-user-login.patch
vi debian/changelog
```

[blackboxsw/ubuntu/focal](https://github.com/blackboxsw/cloud-init/tree/ubuntu/focal) and [blackboxsw/ubuntu/noble](https://github.com/blackboxsw/cloud-init/tree/ubuntu/noble) branches follow this same procedure


debian/patches/revert-551f560d-cloud-config-after-snap-seeding.patch
```diff
@@ -88,14 +88,17 @@ Last-Update: 2024-02-14
...
  # https://cloudinit.readthedocs.io/en/latest/explanation/boot.html
  Description=Cloud-init: Config Stage
- After=network-online.target cloud-config.target
-+After=snapd.seeded.service
- Before=systemd-user-sessions.service
- Wants=network-online.target cloud-config.target
+-After=network-online.target cloud-config.target
+-Wants=network-online.target cloud-config.target
++After=network-online.target cloud-config.target snapd.seeded.service
++Wants=network-online.target cloud-config.target snapd.seeded.service
```
TheRealFalcon, line 98: Adding `snapd.seeded.service` here has no effect, right? It doesn't hurt anything, but I also don't see why it should be added.

TheRealFalcon (approved):
One nit inline but otherwise LGTM!

debian/patches/revert-551f560d-cloud-config-after-snap-seeding.patch
blackboxsw: Agreed on both counts. I'll back that out on Monday having just missed this comment.

blackboxsw: Fixed. focal/jammy/mantic. We don't want that Wants= in there.

Issue #3189, opened 2023-05-11
bionic: DNS search domains lost from cloud-init to netplan


ubuntu-server-builder:
This bug was originally filed in Launchpad as [LP: #1774540](https://bugs.launchpad.net/cloud-init/+bug/1774540)
<details>
<summary>Launchpad details</summary>
<pre>
affected_projects = []
assignee = None
assignee_name = None
date_closed = None
date_created = 2018-06-01T00:30:23.851349+00:00
date_fix_committed = None
date_fix_released = None
id = 1774540
importance = medium
is_complete = False
lp_url = https://bugs.launchpad.net/cloud-init/+bug/1774540
milestone = None
owner = fo0bar
owner_name = Ryan Finnie
private = False
status = confirmed
submitter = fo0bar
submitter_name = Ryan Finnie
tags = []
duplicates = []
</pre>
</details>

_Launchpad user **Ryan Finnie(fo0bar)** wrote on 2018-06-01T00:30:23.851349+00:00_

On a deployed node with a MAAS datasource, /etc/cloud/cloud.cfg.d/50-curtin-networking.cfg contains:

network:
  config:
  - id: enp1s0
    mac_address: 52:54:00:9a:b4:64
    mtu: 1500
    name: enp1s0
    subnets:
    - address: 10.48.7.48/21
      dns_nameservers:
      - 10.48.0.5
      gateway: 10.48.0.1
      type: static
    type: physical
  - address:
    - 10.48.0.5
    search:
    - bos01.canonistack.internal
    type: nameserver
  version: 1

But the generated /etc/netplan/50-cloud-init.yaml contains no search domains under nameservers for the interface:
# This file is generated from information provided by
# the datasource.  Changes to it will not persist across an instance.
# To disable cloud-init's network configuration capabilities, write a file
# /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with the following:
# network: {config: disabled}
network:
    version: 2
    ethernets:
        enp1s0:
            addresses:
            - 10.48.7.48/21
            gateway4: 10.48.0.1
            match:
                macaddress: 52:54:00:9a:b4:64
            mtu: 1500
            nameservers:
                addresses:
                - 10.48.0.5
            set-name: enp1s0

Related bugs:
 * bug 1750884: /etc/resolv.conf not configured correctly in Bionic, leads to no DNS resolution

ubuntu-server-builder:
[...]

PR #5438, opened 2024-06-24
fix netplan dns search

holmanb:
## Proposed Commit Message
```
fix(netplan): dns search
```

## Additional Context
Originally proposed in https://github.com/canonical/cloud-init/issues/3189

## Merge type

### DIFF
--- a/cloudinit/net/netplan.py
+++ b/cloudinit/net/netplan.py
@@ -534,9 +534,14 @@ class Renderer(renderer.Renderer):
             nscfg = {"addresses": nameservers, "search": searchdomains}
             for section in [ethernets, wifis, bonds, bridges, vlans]:
                 for _name, cfg in section.items():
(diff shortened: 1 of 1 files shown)


TheRealFalcon (commented):
Thanks for adding this, but I still see a few issues:
* If DNS is specified, it will only be applied to interfaces that don't already have DNS specified. I think that if the global option is specified, it should be applied everywhere.
* It only applies to static addresses. Shouldn't it also apply when using DHCP?
* Less important, but it also ignores the `interface` specifier. If I add
```
    - type: nameserver
      interface: interface0
      address:
        - 1.1.1.1
      search:
        - aaaa
```
to my v1 config, it's not getting added to my interface0

holmanb:
Thanks for the review @TheRealFalcon, and thanks for catching those issues. I'm focused on other things at the moment, so I'm going to close this and pick it up again later.

Issue #5373, opened 2024-06-05
ec2, hotplug: race between udev add event and IMDS data not being available


aciba90:
# Bug report

Cloud-init hits what @nmeyerhans mentioned in https://github.com/canonical/cloud-init/pull/4799#discussion_r1472287942: 

> One thing to be aware of when accessing IMDS in response to udev add events is that IMDS propagates asynchronously, and the data corresponding with a newly attached network interface may not be complete at the time that the handler is running.

Sometimes, the IMDS does not have the full metadata about a hotplugged NIC when triggered by a hotplug udev event.

We can see the difference in instance network metadata after a reboot:

```diff
--- tmp/pre/cloud-init-logs-2024-06-05/run/cloud-init/instance-data.json        2024-06-05 11:46:07.000000000 +0200
+++ tmp/post/cloud-init-logs-2024-06-05/run/cloud-init/instance-data.json       2024-06-05 12:49:30.000000000 +0200
@@ -91,7 +91,7 @@
      "info": {
       "AccountId": "937157663530",
       "Code": "Success",
-      "LastUpdated": "2024-06-05T09:44:21Z"
+      "LastUpdated": "2024-06-05T10:33:52Z"
      }
     }
    },
@@ -141,6 +141,7 @@
       "06:dc:23:8e:6c:bc": {
        "device-number": "1",
        "interface-id": "eni-08f52ea023734d621",
+       "ipv6s": "2a05:d011:311:a00:c4f1:eb1b:92d2:200e",
        "local-hostname": "ip-192-168-5-5.eu-south-2.compute.internal",
        "local-ipv4s": "192.168.5.5",
        "mac": "06:dc:23:8e:6c:bc",
@@ -155,8 +156,11 @@
        ],
        "subnet-id": "subnet-0877465f33a6d8dde",
        "subnet-ipv4-cidr-block": "192.168.0.0/20",
+       "subnet-ipv6-cidr-blocks": "2a05:d011:311:a00:0:0:0:0/64",
        "vpc-id": "vpc-031a740d79767510f",
-       "vpc-ipv4-cidr-block": "192.168.0.0/20"
+       "vpc-ipv4-cidr-block": "192.168.0.0/20",
+       "vpc-ipv4-cidr-blocks": "192.168.0.0/20",
+       "vpc-ipv6-cidr-blocks": "2a05:d011:311:a00:0:0:0:0/56"
       }
      }
     }
```

## Steps to reproduce the problem

```
[...]

PR #5451, opened 2024-06-26
test(hotplug): disable ipv6 checks due to race

aciba90:
## Proposed Commit Message

```
test(hotplug): disable ipv6 checks due to race

Those checks are going to fail due to and until #5373 is fixed. Disable
them to get a better feedback from integration tests.
```

## Additional Context

#5373

## Merge type


blackboxsw:
Thank you @aciba90. I we want to keep this test active and know that it is flaky based on race-condition symptoms out of cloud-init's control. We'll monitor the frequency of this race hitting our integration-tests and how costly it is for us to categorize this as a platform problem rather than a cloud-init problem.

There could be some work on cloud-init's side to validate 'racy' config conditions from ec2 IMDS, such as presence of `ipv4-cidr-block` but absent `ipv4-cidr-blocks` config keys or partial ipv6 config keys in IMDS data. But, looking for these arbitrary corner-cases don't really solve the fact that udev rules may run sooner on an instance than IMDS data is actually "updated and completely represented"

PR #5411, opened 2024-06-18
deprecation version boundary

holmanb:
## Additional Context

Add deprecation version knob for stable release distros. This allows distros to set an "initial release" version for the cloud-init release on a stable distro, and any newly added deprecation in future releases should log at an INFO level. 

with the following cloud-config:
```yaml
#cloud-config
grub-dpkg:
  enabled:
    false
users:
  - name: osadmin
    lock-passwd: false
    sudo: ["ALL=(ALL) NOPASSWD:ALL"]
    ssh-authorized-keys:
      - ssh-rsa "AAAAB3NzaC1yc2EAAAADAQABAAABAQCc/K1T62elOkJKc94sWovn06RE7D274Fxx5n+eT/oClAwk7632awMtWjO+lIRpzOlEcYCR6wexlmL9mZNEFJoROMMRyCxRilNAiDKOq0X5j1HXY2ky9KjtLxo2b8yXFRBcYKK4W8aHU94NkNek81jdi9lC+L6jUrZ25d7xylqnEbd4TgrBphiowdh5B3RY+j6ePthLWWhHopRcVDgq91yYacelCCIfSDaGBya9iXPZSwoKtPin4n5PdSGmSOA3fIkvV5JjLS1QbuY5tzhXKuH9MdAqrvutF1bprs/GDf0IY8DO7DDoo8rXjhhgGtCdg1i5UitZIAM3hn/Vk5F//5eJ JOOS8312@EB-OR6120158"

```
this PR yields the following schema annotate output:
```yaml
# cloud-init schema --system --annotate
Found cloud-config data types: user-data, network-config

1. user-data at /var/lib/cloud/instances/cb88d1af-f827-4381-bc89-4fad787bcaaf/cloud-config.txt:
#cloud-config

# from 1 files
# part-001

---
grub-dpkg:		# D1
    enabled: false
users:
-   lock-passwd: false		# D2
    name: osadmin
    ssh-authorized-keys:		# D3
[...]

PR #5456, opened 2024-06-26
Schema support deprecated users groups hyphenated keys

blackboxsw:
## Proposed Commit Message
```
    fix(schema): permit deprecated hyphenated keys under users key
    
    Both hyphenated and underscore delimited key names are permitted
    by cloudinit/distros/ug_util.py#L114 due to magic replacement
    of key names.
    
    Since this is still valid json schema, add the necessary hyphenated
    aliases for all users/groups keys. Because the goal in the future is
    to only support one config key for a given configuraion option, add
    deprecated keys to thos schema definitions.
    
    Also drop the description key from the deprecates lock-passwd schema
    key.
    
    Any deprecated schema key which provides a suggested replacement should
    not provide duplicated key descriptions as the preferred replacement
    will provided the necessary context.
    
    Fixes GH-5454
```

## Additional Context
Expect a supplemental PR which allows downstream stable images which include earlier versions of cloud-init  to pin the cloud-init version which determines whether new schema deprecations introduced will be logged at deprecation level or only at info level. https://github.com/canonical/cloud-init/pull/5411.

This PR will allow downstreams to avoid seeing non-zero exit codes from `cloud-init status` due to new deprecations which can be introduced by upgrading the cloud-init package to a newer version of cloud-init on otherwise stable images.

## Test Steps
```
tox -e py3 -- tests/unittests/config/test_cc_users_groups.py
```

## Merge type

### DIFF
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -267,9 +267,9 @@
               "patternProperties": {
                 "^.+$": {
                   "label": "<group_name>",
-                  "description": "When providing an object for users.groups the ``<group_name>`` keys are the groups to add this user to",
                   "deprecated": true,
                   "deprecated_version": "23.1",
+                  "deprecated_description": "The use of ``object`` type is deprecated. Use ``string`` or ``array`` of ``string`` instead.",
                   "type": [
                     "null"
                   ],
@@ -292,9 +292,7 @@
           "type": "string"
         },
         "lock-passwd": {
-          "default": true,
           "type": "boolean",
-          "description": "Default: ``true``",
           "deprecated": true,
           "deprecated_version": "22.3",
           "deprecated_description": "Use ``lock_passwd`` instead."
@@ -304,15 +302,33 @@
           "description": "Disable password login. Default: ``true``",
           "type": "boolean"
         },
+        "no-create-home": {
+          "type": "boolean",
+          "deprecated": true,
+          "deprecated_version": "24.2",
+          "deprecated_description": "Use ``no_create_home`` instead."
+        },
         "no_create_home": {
           "default": false,
           "description": "Do not create home directory. Default: ``false``",
           "type": "boolean"
         },
+        "no-log-init": {
+          "type": "boolean",
+          "deprecated": true,
+          "deprecated_version": "24.2",
+          "deprecated_description": "Use ``no_log_init`` instead."
+        },
         "no_log_init": {
           "default": false,
           "description": "Do not initialize lastlog and faillog for user. Default: ``false``",
           "type": "boolean"
+        },
+        "no-user-group": {
+          "type": "boolean",
+          "deprecated": true,
+          "deprecated_version": "24.2",
+          "deprecated_description": "Use ``no_user_group`` instead."
         },
         "no_user_group": {
           "default": false,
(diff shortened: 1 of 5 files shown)


blackboxsw:
cc: @ani-sinha

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -292,28 +292,69 @@
           "type": "string"
         },
         "lock-passwd": {
-          "default": true,
-          "type": "boolean",
-          "description": "Default: ``true``",
-          "deprecated": true,
-          "deprecated_version": "22.3",
-          "deprecated_description": "Use ``lock_passwd`` instead."
+          "allOf": [
```
holmanb, line 295: Why shift the definition under an `allOf`?

blackboxsw: @holmanb Thanks for questioning this. I originally thought it was to better represent ordering of documentation content on the CLI that ordered the placement of "default" values before the `Deprecated in version 24.2.` documentation. But, I just dumped full schema docs `cloud-init schema --docs all` with allOf schema declaration vs just combining both deprecated* keys and the property constraints and there is no significant difference between either cli-based documentation dumps for with and without the allOf. So, I've avoided this allOf declaration as it equates to the much simpler schema definitions which have the property type constraints and deprecation constraints at the same depth in the schema.
```bash
$ diff cli-docs-with-all-of.out cli-docs-without-all-of.out
546c546
<       **Activate only on keys:** ``ubuntu_pro``, ``ubuntu_advantage``, ``ubuntu-advantage``
---
>       **Activate only on keys:** ``ubuntu_pro``, ``ubuntu-advantage``, ``ubuntu_advantage``
```

blackboxsw: I've pushed this simpler, flat schema defs for this changeset now to add the coverage needed for the remaining hyphenated key support for no-create-home, plain-text-passwd, hashed-passwd, no-user-group, selinux-user and ssh-redirect-user.

ani-sinha (approved):

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -362,20 +409,24 @@
           "minItems": 1
         },
         "ssh-authorized-keys": {
-          "allOf": [
```
holmanb, line 365: +1

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -292,9 +292,7 @@
           "type": "string"
         },
         "lock-passwd": {
-          "default": true,
           "type": "boolean",
-          "description": "Default: ``true``",
```
holmanb, line 297: +1 probably best not to describe a deprecated key.

blackboxsw: Yes I felt in the rendered docs it just gave wasted real-estate for duplicated description text to the large wall of text for documented keys. I've done the same dropping any documented "default" values for the deprecated keys as long as they have a suggested replacement which is documented.

blackboxsw:
> Thanks for these additions @blackboxsw, and also the cleanups!
> 
> There are a few other cleanup that we should probably do:
> 
> ca-certs -> unnecessary allOf nesting system_info - missing deprecated_description, unnecessary description grub-dpkg - unnecessary description apt_update - unnecessary description apt_upgrade - unnecessary description apt_reboot_if_required - unnecessary description chpasswd.list - unnecessary description groups - object type has unnecessary description lock-passwd - unneccessary description
> 
> we can do these in a separate pr

Good thought. Let's standardize while we are here. I just squash merged those changes you suggested into this commit.

Here's the supplemental diff I've squashed in. It includes also dropping "default" schema documentation on deprecated keys apt_upgrade, apt_update, no-log-init and remove-defaults

holmanb:
> > Thanks for these additions @blackboxsw, and also the cleanups!
> > There are a few other cleanup that we should probably do:
> > ca-certs -> unnecessary allOf nesting system_info - missing deprecated_description, unnecessary description grub-dpkg - unnecessary description apt_update - unnecessary description apt_upgrade - unnecessary description apt_reboot_if_required - unnecessary description chpasswd.list - unnecessary description groups - object type has unnecessary description lock-passwd - unneccessary description
> > we can do these in a separate pr
> 
> Good thought. Let's standardize while we are here. I just squash merged those changes you suggested into this commit.
> 
> Here's the supplemental diff I've squashed in. It includes also dropping "default" schema documentation on deprecated keys apt_upgrade, apt_update, no-log-init and remove-defaults

@blackboxsw I think we're still missing these two:

chpasswd.list - unnecessary description
groups - object type has unnecessary description

blackboxsw:
> 
> chpasswd.list - unnecessary description groups - object type has unnecessary description

Got those too

holmanb (approved):
@blackboxsw LGTM once tests pass.

PR #5460, opened 2024-06-28
fix(wsl): Special handling Landscape client config tags

CarlosNihelton:
## Proposed Commit Message

```
fix(wsl): Special handling Landscape client config tags

UP4W business logic is so that its data overrides user at a key (module) 
level.
That means the entire Landscape config is overriden if both agent data
and user data contains config for that module.
Yet, for better usability, computer tags must be assignable per instance.
That's not possible with agent.yaml, because it's meant to be global.
Its config data affects all Ubuntu WSL instances.

Thus this aims to make a special case for landscape.client.tags,
if present in user provided data (either Landscape or local user -
  whatever is picked up before merging with agent.yaml)
its value overwrites any tags set by agent.yaml.

Only landscape.client.tags are treated specially.
The pre-existing merge rules still apply for any other value present in
both agent.yaml and user provided data.

Fixes UDENG-2464
```

## Test Steps

Fixes UDENG-2464
## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -328,16 +328,12 @@ class DataSourceWSL(sources.DataSource):
         # provides them instead.
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
+        user_tags: str = ""
         overridden_keys: typing.List[str] = []
-        # We've already checked, this is just to please mypy.
-        assert user_data is not None
-        user_tags = (
-            user_data.get("landscape", {}).get("client", {}).get("tags")
-        )
         if user_data:
             merged = user_data
             user_tags = (
-                merged.get("landscape", {}).get("client", {}).get("tags")
+                merged.get("landscape", {}).get("client", {}).get("tags", "")
             )
         if agent_data:
             if user_data:
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -353,14 +353,13 @@ class DataSourceWSL(sources.DataSource):
                         ", ".join(overridden_keys)
                     )
                 )
-            if user_tags:
+            if user_tags and merged.get("landscape", {}).get("client"):
                 LOG.debug(
-                    " Landscape client conf updated with user-data"
+                    "Landscape client conf updated with user-data"
                     " landscape.client.tags: %s",
                     user_tags,
                 )
-                if merged.get("landscape", {}).get("client"):
-                    merged["landscape"]["client"]["tags"] = user_tags
+                merged["landscape"]["client"]["tags"] = user_tags
 
         self.userdata_raw = "#cloud-config\n%s" % yaml.dump(merged)
         return True
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -329,8 +329,12 @@ class DataSourceWSL(sources.DataSource):
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
         overridden_keys: typing.List[str] = []
+        user_tags = None
         if user_data:
             merged = user_data
+            user_tags = (
+                merged.get("landscape", {}).get("client", {}).get("tags")
+            )
         if agent_data:
             if user_data:
                 LOG.debug("Merging both user_data and agent.yaml configs.")
@@ -344,6 +348,15 @@ class DataSourceWSL(sources.DataSource):
                         " agent.yaml overrides config keys: "
                         ", ".join(overridden_keys)
                     )
+                )
+            if user_tags:
+                LOG.debug(
+                    " Landscape client conf updated with user-data tags: %s",
+                    user_tags,
+                )
+                # Does nothing if there is no "landscape"."client" key.
+                merged.get("landscape", {}).get("client", {}).update(
+                    {"tags": user_tags}
                 )
 
         self.userdata_raw = "#cloud-config\n%s" % yaml.dump(merged)
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -329,7 +329,9 @@ class DataSourceWSL(sources.DataSource):
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
         overridden_keys: typing.List[str] = []
-        user_tags = None
+        user_tags = (
+            user_data.get("landscape", {}).get("client", {}).get("tags")
+        )
         if user_data:
             merged = user_data
             user_tags = (
@@ -351,13 +353,12 @@ class DataSourceWSL(sources.DataSource):
                 )
             if user_tags:
                 LOG.debug(
-                    " Landscape client conf updated with user-data tags: %s",
+                    " Landscape client conf updated with user-data"
+                    " landscape.client.tags: %s",
                     user_tags,
                 )
-                # Does nothing if there is no "landscape"."client" key.
-                merged.get("landscape", {}).get("client", {}).update(
-                    {"tags": user_tags}
-                )
+                if merged.get("landscape", {}).get("client"):
+                    merged["landscape"]["client"]["tags"] = user_tags
 
         self.userdata_raw = "#cloud-config\n%s" % yaml.dump(merged)
         return True
--- a/cloudinit/sources/DataSourceWSL.py
+++ b/cloudinit/sources/DataSourceWSL.py
@@ -329,6 +329,8 @@ class DataSourceWSL(sources.DataSource):
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
         overridden_keys: typing.List[str] = []
+        # We've already checked, this is just to please mypy.
+        assert user_data is not None
         user_tags = (
             user_data.get("landscape", {}).get("client", {}).get("tags")
         )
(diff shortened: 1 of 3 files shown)


tests/unittests/sources/test_wsl.py
```diff
@@ -480,3 +482,103 @@ def test_data_precedence(self, m_get_linux_dist, tmpdir, paths):
         assert (
             "landscapetest" not in userdata and "agenttest" in userdata
         ), "Landscape account name should have been overriden by agent data"
+        assert (
+            "tag_aiml" in userdata and "tag_dev" in userdata
+        ), "User-data should override agent data's Landscape computer tags"
+
+        # Set up some Landscape provided user data without tags
```
blackboxsw, line 489: This unittest is getting super long,(228 lines) can we break this down into smaller named tests with a a short one-linter docstr that explains the intent of the scoped test please?

cloudinit/sources/DataSourceWSL.py
```diff
@@ -345,6 +349,15 @@ def _get_data(self) -> bool:
                         ", ".join(overridden_keys)
                     )
                 )
+            if user_tags:
+                LOG.debug(
+                    " Landscape client conf updated with user-data tags: %s",
+                    user_tags,
+                )
```
blackboxsw, line 356:
```suggestion
                    " Landscape client conf updated with user-data"
                    " landscape.client.tags: %s",
                    user_tags,
                )
```

cloudinit/sources/DataSourceWSL.py
```diff
@@ -345,6 +349,15 @@ def _get_data(self) -> bool:
...
+                # Does nothing if there is no "landscape"."client" key.
+                merged.get("landscape", {}).get("client", {}).update(
+                    {"tags": user_tags}
+                )
```
blackboxsw, line 360: While this is neat, the fact that we need to document it with a comment means it lacks readability.
Can we make this easier to interpret with the following?

```suggestion
                if merged.get("landscape", {}).get("client"):
                    merged["landscape"]["client"]["tags"] = user_tags
```

cloudinit/sources/DataSourceWSL.py
```diff
@@ -329,8 +329,12 @@ def _get_data(self) -> bool:
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
         overridden_keys: typing.List[str] = []
+        user_tags = None
```
blackboxsw, line 332: We can define this inline and drop the  set of this below. It'll end up None if no landscape.client.tags are defined. May have to check line length on this suggestion.
```suggestion
        user_tags = user_data.get("landscape", {}).get("client", {}).get("tags")
```

cloudinit/sources/DataSourceWSL.py
```diff
@@ -329,8 +329,12 @@ def _get_data(self) -> bool:
...
         if user_data:
             merged = user_data
+            user_tags = (
+                merged.get("landscape", {}).get("client", {}).get("tags")
```
blackboxsw, line 336: Our agent data overriding user-data policy in WSL leads us down this path where we may start defining multiple exceptions to the "rule". If we end up coming up with a second case were other user-data keys are preferred over agent data keys, we may want to define a separate  userdata_override_paths dict or list or a separate function that extracts a dict of userdata_overrides from the original data so we have one place that defines these expected userdata paths.

blackboxsw (changes requested):

doc/rtd/reference/datasources/wsl.rst
```diff
@@ -66,7 +66,9 @@ following paths:
    the Ubuntu Pro for WSL agent. If this file is present, its modules will be
    merged with (1), overriding any conflicting modules. If (1) is not provided,
    then this file will be merged with any valid user-provided configuration
-   instead.
+   instead. Exception is made for Landscape client config computer tags. If
+   user provided data contains a value for that subkey it will be used instead
```
blackboxsw, line 70:
```suggestion
   user provided data contains a value for ``landscape.client.tags`` it will be used instead
```

tests/unittests/sources/test_wsl.py
CarlosNihelton: Indeed, I'm splitting that into smaller test cases.

tests/unittests/sources/test_wsl.py
```diff
@@ -436,17 +447,48 @@ def test_data_precedence(self, m_get_linux_dist, tmpdir, paths):
...
+
+        m_get_linux_dist.return_value = SAMPLE_LINUX_DISTRO
+
+        user_file = tmpdir.join(".cloud-init", "ubuntu-24.04.user-data")
+        user_file.dirpath().mkdir()
+        user_file.write(
+            """#cloud-config
+landscape:
+  client:
+    account_name: user
```
blackboxsw, line 466: This is invalid JSON schema  for landscape as it also requires a computer_title. maybe good to have that represented.

cloudinit/sources/DataSourceWSL.py
blackboxsw: Given that every computer needs a different landscape.client.computer_title would we expect to override this with user-data too?

tests/unittests/sources/test_wsl.py
blackboxsw: I'm looking for separate simple `def test_`  cases defined to better describe the individual tests. Is that work still uncommitted?

blackboxsw (commented):
Beyond the outstanding question on whether we need user-data overrides for landscape.client.computer_name as well I ran this test on Azure Windows Server and can verify correct behavior in various permutations of agent.yaml w/ and w/out landscape.client config and user-data w/ and w/out landscape config tags. 


Just minor nits on when to log that merge data line and breaking up the last unit test into two separate unit tests

tests/unittests/sources/test_wsl.py
CarlosNihelton: I'm pushing updates with more realistic Landscape configs. Yet, we don't need to represent the computer title in the UP4W-related cases, as that piece of data is handled internally.

tests/unittests/sources/test_wsl.py
CarlosNihelton: Well, we should be in a better shape now.

cloudinit/sources/DataSourceWSL.py
blackboxsw: Great thanks!

cloudinit/sources/DataSourceWSL.py
```diff
@@ -329,8 +329,16 @@ def _get_data(self) -> bool:
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
         overridden_keys: typing.List[str] = []
+        # We've already checked, this is just to please mypy.
+        assert user_data is not None
+        user_tags = (
+            user_data.get("landscape", {}).get("client", {}).get("tags")
+        )
```
blackboxsw: heh. I saw this and originally agreed, but this is a real issue mypy has for use. at this point agent_data OR user_data could be non-None, so there is still a possibility that user_data could be None here, which would raise an AssertionError and break cloud-init. Since we are actually setting this 3 lines below after our if user_data check, let's just initialize user_tags to None here and set it if present a couple lines below.

blackboxsw: I'll apply this diff and approve after CI.
```
    mypy: init user_tags to empty string, only set it when present in userdata

diff --git a/cloudinit/sources/DataSourceWSL.py b/cloudinit/sources/DataSourceWSL.py
index bd71cd431..7a75ff4e6 100644
@@ -328,16 +328,12 @@ class DataSourceWSL(sources.DataSource):
         # provides them instead.
         # That's the reason for not using util.mergemanydict().
         merged: dict = {}
+        user_tags: str = ""
         overridden_keys: typing.List[str] = []
-        # We've already checked, this is just to please mypy.
-        assert user_data is not None
-        user_tags = (
-            user_data.get("landscape", {}).get("client", {}).get("tags")
-        )
         if user_data:
             merged = user_data
             user_tags = (
-                merged.get("landscape", {}).get("client", {}).get("tags")
+                merged.get("landscape", {}).get("client", {}).get("tags", "")
             )
         if agent_data:
             if user_data:
```

tests/unittests/sources/test_wsl.py
blackboxsw: thank you!

tests/unittests/sources/test_wsl.py
```diff
@@ -469,14 +560,175 @@ def test_data_precedence(self, m_get_linux_dist, tmpdir, paths):
...
             " nor ignored"
         )
-        assert "landscape" in userdata
         assert (
             "landscapetest" not in userdata and "agenttest" in userdata
         ), "Landscape account name should have been overriden by agent data"
+        # Make sure we have tags from Landscape data, not agent's
+        assert (
+            "tag_aiml" in userdata and "tag_dev" in userdata
+        ), "User-data should override agent data's Landscape computer tags"
```
blackboxsw: Let's strictly assert we don't see `wsl` as well from agent.yaml. Also note landscape client config schema disallows spaces between tags. so our example should be `tags: x,y`
```suggestion
        assert (
            "tags: tag_aiml,tag_dev" in userdata
        ), "User-data should override agent data's Landscape computer tags"
        assert "wsl" not in userdata
```

blackboxsw (approved):
minor test nits and mypy fix being pushed Will land it after CI passes.
Thanks @CarlosNihelton!

CarlosNihelton:
> minor test nits and mypy fix being pushed Will land it after CI passes.
> Thanks @CarlosNihelton!

Thanks for the collaboration as usual!

PR #5464, opened 2024-06-28
fix(schema): Don't report changed keys as deprecated

holmanb:
## Proposed Commit Message
```
fix(schema): Don't report changed keys as deprecated
```

## Additional Context
also fixes issue where unittests run with version boundary set

### DIFF
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -362,7 +362,7 @@ def _validator(
 ):
     """Jsonschema validator for `deprecated` items.
 
-    It raises a instance of `error_type` if deprecated that must be handled,
+    It yields an instance of `error_type` if deprecated that must be handled,
     otherwise the instance is consider faulty.
     """
     if deprecated:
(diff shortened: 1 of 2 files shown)


TheRealFalcon (approved):
LGTM!

holmanb:
updated the commit messages

cloudinit/config/schema.py
```diff
@@ -797,8 +797,9 @@ def validate_cloudconfig_schema(
         ):  # pylint: disable=W1116
             if (
                 "devel" != features.DEPRECATION_INFO_BOUNDARY
-                and Version.from_str(schema_error.version)
-                > Version.from_str(features.DEPRECATION_INFO_BOUNDARY)
+                and (schema_error.version == "devel"
+                or Version.from_str(schema_error.version)
+                > Version.from_str(features.DEPRECATION_INFO_BOUNDARY))
```
blackboxsw, line 802: this won't work for our default "devel" case as Version.from_str("devel") will traceback
```
ValueError: invalid lteral for int() with base 10: 'devel'
```

blackboxsw: Resolving this comment as we only see schema_error.version == "devel" when no deprecated_version is set in the schema definition, which happens to be when we have the changed_version changed_message keys... Which I realize is what @holmanb was proposing to fix here.

I also can confirm this fix doesn't report deprecation messages for "changed" keys

PR #5466, opened 2024-06-28
test: Fix deprecation test failures

TheRealFalcon:
## Proposed Commit Message

```
test: Fix deprecation test failures

8906e17e is causing some test failures on older releases.
Fix them
```

## Merge type

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3210,6 +3210,18 @@ class Version(namedtuple("Version", ["major", "minor", "patch", "rev"])):
         return -1
 
 
+def should_log_deprecation(version: str) -> bool:
+    """Determine if a deprecation message should be logged.
+
+    :param version: The version in which the thing was deprecated.
+
+    :return: True if the message should be logged, else False.
+    """
+    return features.DEPRECATION_INFO_BOUNDARY == "devel" or Version.from_str(
+        version
+    ) <= Version.from_str(features.DEPRECATION_INFO_BOUNDARY)
+
+
 def deprecate(
     *,
     deprecated: str,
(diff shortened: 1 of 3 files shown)


cloudinit/util.py
```diff
@@ -3210,6 +3210,18 @@ def _compare_version(self, other: "Version") -> int:
         return -1
 
 
+def should_log_deprecation(version: str) -> bool:
```
holmanb, line 3213: this could be used too (simplify this logic too](https://github.com/canonical/cloud-init/blob/5ce2ee36e243b4cc178fef28bc0da06212ac7607/cloudinit/config/schema.py#L800)

if we're making a helper, could we re-use it there too?

TheRealFalcon: I'm not sure I understand what that particular `if` is trying to convey. It has more stuff than `should_log_deprecations()`, but I'm not sure why.

blackboxsw: @TheRealFalcon I think it just means we want to invert the conditional to append to deprecations if should_log_deprecation else to info_deprecations:
```diff
--- a/cloudinit/config/schema.py
+++ b/cloudinit/config/schema.py
@@ -43,6 +43,7 @@ from cloudinit.util import (
     get_modules_from_dir,
     load_text_file,
     load_yaml,
+    should_log_deprecation,
     write_file,
 )
 
@@ -795,16 +796,12 @@ def validate_cloudconfig_schema(
         if isinstance(
             schema_error, SchemaDeprecationError
         ):  # pylint: disable=W1116
-            if "devel" != features.DEPRECATION_INFO_BOUNDARY and (
-                schema_error.version == "devel"
-                or Version.from_str(schema_error.version)
-                > Version.from_str(features.DEPRECATION_INFO_BOUNDARY)
-            ):
+            if should_log_deprecation(schema_error.version):
+                deprecations.append(SchemaProblem(path, schema_error.message))
+            else:
                 info_deprecations.append(
                     SchemaProblem(path, schema_error.message)
                 )
-            else:
-                deprecations.append(SchemaProblem(path, schema_error.message))
         else:
             errors.append(SchemaProblem(path, schema_error.message))
 

```

blackboxsw: With the above diff, downstream unittests pass on jammy with all quilt patches applied (which pins deprecation_info_boundary)

blackboxsw (approved):
bring it home

PR #5469, opened 2024-07-01
tests: integration tests aware of features.DEPRECATION_INFO_BOUNDARY

blackboxsw:
## Proposed Commit Message
```
tests: integration tests aware of features.DEPRECATION_INFO_BOUNDARY
```

## Additional Context

Failed jenkins jobs represented deprecation message errors in integration tests https://jenkins.canonical.com/server-team/view/cloud-init/job/cloud-init-integration-jammy-lxd_container/454/ .

## Test Steps
```
CLOUD_INIT_OS_IMAGE=jammy CLOUD_INIT_PLATFORM=lxd_container CLOUD_INIT_CLOUD_INIT_SOURCE=ppa:cloud-init-dev/daily tox -e integration-tests -- tests/integration_tests/cmd/test_schema.py::TestSchemaDeprecations::test_clean_log tests/integration_tests/cmd/test_schema.py::TestSchemaDeprecations::test_schema_deprecations tests/integration_tests/cmd/test_status.py::test_status_json_errors tests/integration_tests/modules/test_combined.py::TestCombined::test_deprecated_message
```

## Merge type

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -3210,16 +3210,17 @@ class Version(namedtuple("Version", ["major", "minor", "patch", "rev"])):
         return -1
 
 
-def should_log_deprecation(version: str) -> bool:
+def should_log_deprecation(version: str, boundary_version: str) -> bool:
     """Determine if a deprecation message should be logged.
 
     :param version: The version in which the thing was deprecated.
+    :param boundary_version: The version at which deprecation level is logged.
 
     :return: True if the message should be logged, else False.
     """
-    return features.DEPRECATION_INFO_BOUNDARY == "devel" or Version.from_str(
+    return boundary_version == "devel" or Version.from_str(
         version
-    ) <= Version.from_str(features.DEPRECATION_INFO_BOUNDARY)
+    ) <= Version.from_str(boundary_version)
 
 
 def deprecate(
@@ -3263,7 +3264,9 @@ def deprecate(
         f"{deprecated_version} and scheduled to be removed in "
         f"{version_removed}. {message}"
     ).rstrip()
-    if not should_log_deprecation(deprecated_version):
+    if not should_log_deprecation(
+        deprecated_version, features.DEPRECATION_INFO_BOUNDARY
+    ):
         level = logging.INFO
     elif hasattr(LOG, "deprecated"):
         level = log.DEPRECATED
--- a/tests/integration_tests/modules/test_combined.py
+++ b/tests/integration_tests/modules/test_combined.py
@@ -15,7 +15,7 @@ from pathlib import Path
 import pytest
 
 import cloudinit.config
-from cloudinit.util import is_true
+from cloudinit.util import is_true, should_log_deprecation
 from tests.integration_tests.decorators import retry
 from tests.integration_tests.instances import IntegrationInstance
 from tests.integration_tests.integration_settings import PLATFORM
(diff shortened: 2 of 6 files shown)


tests/integration_tests/modules/test_combined.py
```diff
@@ -131,8 +131,20 @@ def test_deprecated_message(self, class_client: IntegrationInstance):
         """Check that deprecated key produces a log warning"""
         client = class_client
         log = client.read_from_file("/var/log/cloud-init.log")
-        assert "Deprecated cloud-config provided" in log
-        assert "The value of 'false' in user craig's 'sudo' config is " in log
+        version_boundary = get_feature_flag_value(
+            class_client, "DEPRECATION_INFO_BOUNDARY"
+        )
+        # the deprecation_version is 22.2 in schema for apt_* keys in
+        # user-data. Pass 22.2 in cagainst the client's version_boundary.
```
aciba90, line 138:
```suggestion
        # user-data. Pass 22.2 in against the client's version_boundary.
```

cloudinit/util.py
```diff
@@ -3210,16 +3210,20 @@ def _compare_version(self, other: "Version") -> int:
         return -1
 
 
-def should_log_deprecation(version: str) -> bool:
+def should_log_deprecation(
+    version: str, boundary_version: Optional[str] = None
+) -> bool:
     """Determine if a deprecation message should be logged.
 
     :param version: The version in which the thing was deprecated.
```
holmanb, line 3218: Please document parameter

TheRealFalcon (approved):

holmanb (changes requested):
At a minimum, lets document the new parameter.

However, adding parameters to cloud-init runtime code that only exist for testing purposes seems like a smell to me. This codebase has extensive history of adding test testing parameters, but this seems like exactly the purpose of a mock.

Why not just use a mock? Something like this perhaps?

```python
with mock.patch("cloudinit.features.DEPRECATION_INFO_BOUNDARY", "some_version"):
    should_log = should_log_deprecation("22.2")
```

TheRealFalcon:
> However, adding parameters to cloud-init runtime code that only exist for testing purposes seems like a smell to me. 

Why? It seems like a pretty common practice and (outside of this code base) is usually preferred to mocking.

> this seems like exactly the purpose of a mock.

I'd really like to not use mocks in integration tests unless absolutely necessary. I prefer the proposed solution here.

holmanb:
> > However, adding parameters to cloud-init runtime code that only exist for testing purposes seems like a smell to me.
> 
> Why? 

Additional conditional branches through the codebase which don't actually get used by production code:

1. reduce code coverage
2. increase cyclomatic complexity 
3. slow down runtime code
4. add boilerplate

> It seems like a pretty common practice and (outside of this code base) is usually preferred to mocking.

oh? 

> > this seems like exactly the purpose of a mock.
> 
> I'd really like to not use mocks in integration tests unless absolutely necessary.

Sure, that's fine.

An alternative would just to be to make this non-optional. We could just convert the other (2) callsites to use this new parameter, which has the added benefit of avoiding reliance on side effects.

TheRealFalcon (approved):

blackboxsw:
Good points.

I also tend to want to avoid static mocking in integration tests here because I want these tests to assert specifically that our downstream packaging actually modified an unmocked value to a valid value for features.DEPRECATION_INFO_BOUNDARY. If we statically mocked this value, then the test we have isn't exposed to whatever our downstream quilt packages have done. So, we risk not seeing the actual values from jammy/focal/mantic and we don't catch a problem that could have been introduced in our real downstream patches


> > > However, adding parameters to cloud-init runtime code that only exist for testing purposes seems like a smell to me.
> > 
> > 
> > Why?
> 
> Additional conditional branches through the codebase which don't actually get used by production code:
> 
> 1. reduce code coverage
> 2. increase cyclomatic complexity
> 3. slow down runtime code
> 4. add boilerplate
> 
> > It seems like a pretty common practice and (outside of this code base) is usually preferred to mocking.
> 
> oh?
> 
> > > this seems like exactly the purpose of a mock.
> > 
> > 
> > I'd really like to not use mocks in integration tests unless absolutely necessary.
> 
> Sure, that's fine.
> 
> An alternative would just to be to make this non-optional. We could just convert the other (2) callsites to use this new parameter, which has the added benefit of avoiding reliance on side effects.

Those are good points, and I think ultimately I'd like to settle on your final suggestion which will make both params mandatory so we don't have complexity in the function, and can use it appropriately in integration test where we need to source the feature value from a remote system instead of the 'local' features value from the current repo.

holmanb (approved):
LGTM once CI passes. I just kicked CI since it looked like there was a spurious failure.

PR #5476, opened 2024-07-01
Focal test updates

TheRealFalcon:
## Proposed Commit Message

```
test: Update integration tests to pass on focal

- Update deprecation test to include focal.
- Update test using `cloud-id` appropriately
```

## Merge type

### DIFF
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -3,14 +3,12 @@ import json
 
 import pytest
 
-from cloudinit.util import should_log_deprecation
 from tests.integration_tests.clouds import IntegrationCloud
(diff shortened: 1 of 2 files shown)


blackboxsw (approved):
Ship it. Agreed it makes sense to avoid the per-series deprecation-boundary based conditionals given that our lowest supported series focal is below the 22.1 deprecation_boundary of ca-certs. ssh-authorized-keys as a deprecation_version of 18.3 which is below all our stable series that get SRUs, so no need to complicate the integration test with per-series conditionals if we can select a deprecated key that happens to be derpecated everywhere. We have other integration tests that cover cases where some stable series log at INFO level and some at DEPRECATED level so we don't need that complexity/duplication of testing here as well.

PR #5478, opened 2024-07-02
test: Fix no default user in test_status.py

TheRealFalcon:
## Proposed Commit Message

```
test: Fix no default user in test_status.py
```

## Merge type

### DIFF
--- a/tests/integration_tests/cmd/test_status.py
+++ b/tests/integration_tests/cmd/test_status.py
@@ -66,6 +66,7 @@ USER_DATA = """\
 users:
   - name: something
     ssh-authorized-keys: ["something"]
+  - default
 ca-certs:
   invalid_key: true
 """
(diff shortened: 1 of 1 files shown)


tests/integration_tests/cmd/test_status.py
```diff
@@ -66,6 +66,7 @@ def test_wait_when_no_datasource(session_cloud: IntegrationCloud, setup_image):
 users:
   - name: something
     ssh-authorized-keys: ["something"]
+  - default
 ca-certs:
```
blackboxsw, line 70: I think we also need to change ca-certs to ca_certs to avoid getting integration test failures in the following test:
```
FAILED tests/integration_tests/cmd/test_status.py::test_status_json_errors - assert 'Deprecated cloud-config provided: users.0.ssh-authorized-keys' in 'Deprecated cloud-config provided: ca-certs:  Deprecated in version 22.3. Use ``ca_certs`` instead., use...
```

blackboxsw (changes requested):
Can we move away from deprecated key ca-certs too?

blackboxsw (approved):
Merge away

PR #5479, opened 2024-07-02
test: Add jsonschema guard in test_cc_ubuntu_pro.py

TheRealFalcon:
jsonschema is still a soft dependency, so don't fail it not present.

## Proposed Commit Message

```
test: Add jsonschema guard in test_cc_ubuntu_pro.py

jsonschema is still a soft dependency, so don't fail it not present.
```

## Test Steps
Ran tests with jsonschema both installed and uninstalled.

## Merge type


blackboxsw (approved):
LGTM tested in virtenv w/out jsonschema.

PR #5487, opened 2024-07-03
chore(postinst): Delete unnecessary postinst code

holmanb:
## Proposed Commit Message
```
chore(postinst): Delete unnecessary postinst code
```

## Additional Context

1) Delete version gated code
- most recent version gate was 22.1-14-g2e17a0d6-0ubuntu1~22.04.5
- If someone somehow upgraded a Jammy instance to noble without getting the newer Jammy cloud-init package available in jammy-updates, the fix in the postinst would still run since this code exists in all releases of noble.

2) Delete old preseed code. Debian doesn't use our postinst, and we don't use preseed, so just drop it -> we can probably do this with the maas preseed code too but that requires more investigation

## Merge type

### DIFF
--- a/debian/cloud-init.postinst
+++ b/debian/cloud-init.postinst
@@ -196,6 +196,8 @@ EOF
 
    fix_lp1889555 "$2"
 
+   rename_hook_hotplug_udev_rule
+
 fi
 
 #DEBHELPER#
--- a/debian/cloud-init.postinst
+++ b/debian/cloud-init.postinst
@@ -191,11 +191,10 @@ datasource_list: [ $values ]
 EOF
    fi
 
-   # if there are maas settings pre-seeded apply them
-   handle_preseed_maas
-
    # fix issue where cloud-init misidentifies the location of grub
    fix_1336855
+
+   fix_lp1889555 "$2"
 
 fi
 
--- a/debian/cloud-init.postinst
+++ b/debian/cloud-init.postinst
@@ -51,64 +51,6 @@ with open(fname, "w") as fp:
     fp.write(yaml.dump(cfg))' "$@"
 }
 
-handle_preseed_maas() {
-   local cfg_file="/etc/cloud/cloud.cfg.d/90_dpkg_maas.cfg"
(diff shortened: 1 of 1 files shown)


debian/cloud-init.postinst
```diff
@@ -109,22 +109,6 @@ for k in sys.argv[2:]:
    db_unregister "cloud-init/maas-metadata-credentials" || :
 }
 
-handle_preseed_local_cloud_config() {
```
TheRealFalcon, line 112: The call site needs to be removed too

debian/cloud-init.postinst
```diff
@@ -390,23 +216,6 @@ EOF
...
-
-   remove_deprecated_cloud_config_modules "$2"
-
-   fix_azure_upgrade_1611074 "$2"
-
-   cleanup_ureadahead "$2"
-   fix_lp1889555 "$2"
-   change_cloud_init_output_log_permissions "$2"
-
-   rename_hook_hotplug_udev_rule
```
TheRealFalcon, line 404: Did you mean to include this one? The definition hasn't been deleted and this is a relatively recent addition.

TheRealFalcon (changes requested):
I left some inline comments.

If we're confident there's no way we can install with preseed anymore, than I'm not sure why we can't also delete the MAAS preseed code.

debian/cloud-init.postinst
holmanb: No I didn't - fixed.

holmanb:
> If we're confident there's no way we can install with preseed anymore, than I'm not sure why we can't also delete the MAAS preseed code.

+1 agreed

@TheRealFalcon I think that I've addressed all of your comments.

TheRealFalcon (changes requested):
Two more inline.

debian/cloud-init.postinst
TheRealFalcon: I'm not seeing it.

TheRealFalcon:
@holmanb , just because it is showing up under the resolved comments, I wanted to make sure you can see my latest comment: https://github.com/canonical/cloud-init/pull/5487#discussion_r1686522844

PR #5488, opened 2024-07-04
fix(test): Fix ip printer for non-lxd

holmanb:
## Proposed Commit Message

```
fix(test): Fix ip printer for non-lxd
```

## Additional Context

https://github.com/canonical/cloud-init/commit/343d2c72f32db8fbe47eb8afa98c15987e9ffba6 gave us a  log with the IP address of an instance, if `KEEP_INSTANCE = True`, which allows on to ssh into an instance immediately after a test finishes, useful for debugging and developing tests.

https://github.com/canonical/cloud-init/commit/937e84c4def1e8c88e27d7e270529403b9a63dcc worked around [a pycloudlib bug](https://github.com/canonical/pycloudlib/issues/316) in this logic which caused some lxd instances to hang/fail, but broke the IP printer logic for all non-lxd platforms.

This brings back the IP printer logic. I tested this on ec2 and lxd.

## Test Steps
Run integration test with `KEEP_INSTANCE=True`, check for `Keeping Instance, public ip: x.x.x.x` in the "live log teardown" pytest section.

### DIFF
--- a/tests/integration_tests/instances.py
+++ b/tests/integration_tests/instances.py
@@ -294,6 +294,8 @@ class IntegrationInstance:
                 and self.instance.execute_via_ssh
             ):
                 self._ip = self.instance.ip
+            elif not isinstance(self.instance, LXDInstance):
+                self._ip = self.instance.ip
         except NotImplementedError:
             self._ip = "Unknown"
         return self._ip


TheRealFalcon (approved):
Nit: This could be simplified to:
```python
            if (
                not isinstance(self.instance, LXDInstance)
                or self.instance.execute_via_ssh
            ):
                self._ip = self.instance.ip
```

but +1 either way.

Issue #5445, opened 2024-06-25
[enhancement]: fix typing of untyped-defs


aciba90:
# Enhancement

In https://github.com/canonical/cloud-init/pull/5409 we enabled the `check_untyped_defs` mypy lint to not include more type issues and disabled modules not passing the check. Fix them all!

## Disabled modules

- [ ] cloudinit.analyze
- [ ] cloudinit.analyze.dump
- [ ] cloudinit.analyze.show
- [ ] cloudinit.cmd.devel.hotplug_hook
- [ ] cloudinit.cmd.devel.make_mime
- [ ] cloudinit.cmd.devel.net_convert
- [ ] cloudinit.cmd.main
- [ ] cloudinit.config.cc_apt_configure
- [ ] cloudinit.config.cc_ca_certs
- [ ] cloudinit.config.cc_growpart
- [ ] cloudinit.config.cc_ntp
- [ ] cloudinit.config.cc_power_state_change
- [ ] cloudinit.config.cc_rsyslog
- [ ] cloudinit.config.cc_ubuntu_pro
- [ ] cloudinit.config.modules
- [ ] cloudinit.config.schema
- [ ] cloudinit.distros
- [ ] cloudinit.distros.alpine
- [ ] cloudinit.distros.azurelinux
- [ ] cloudinit.distros.bsd
- [ ] cloudinit.distros.opensuse
- [ ] cloudinit.distros.parsers.hostname
- [ ] cloudinit.distros.parsers.hosts
- [ ] cloudinit.distros.parsers.resolv_conf
- [ ] cloudinit.distros.ubuntu
- [ ] cloudinit.distros.ug_util
- [ ] cloudinit.helpers
- [ ] cloudinit.log
- [ ] cloudinit.mergers
- [ ] cloudinit.net
- [ ] cloudinit.net.cmdline
- [ ] cloudinit.net.dhcp
- [ ] cloudinit.net.eni
- [ ] cloudinit.net.ephemeral
- [ ] cloudinit.net.freebsd
- [ ] cloudinit.net.netbsd
- [ ] cloudinit.net.netplan
- [ ] cloudinit.net.network_manager
- [ ] cloudinit.net.network_state
- [ ] cloudinit.net.networkd
- [ ] cloudinit.net.sysconfig
- [ ] cloudinit.netinfo
- [ ] cloudinit.reporting.handlers
- [ ] cloudinit.sources.DataSourceAzure
[...]

PR #5490, opened 2024-07-05
typing: fix check_untyped_defs in cloudinit.util

aciba90:
GH-5445

## Proposed Commit Message

```
typing: fix check_untyped_defs in cloudinit.util

GH-5445
```

## Merge type

### DIFF
--- a/cloudinit/util.py
+++ b/cloudinit/util.py
@@ -2454,14 +2454,16 @@ def is_lxd():
     return os.path.exists("/dev/lxd/sock")
 
 
-def get_proc_env(pid, encoding="utf-8", errors="replace"):
+def get_proc_env(
+    pid, encoding: str = "utf-8", errors: str = "replace"
+) -> Dict[str, str]:
     """
     Return the environment in a dict that a given process id was started with.
 
-    @param encoding: if true, then decoding will be done with
-                     .decode(encoding, errors) and text will be returned.
-                     if false then binary will be returned.
-    @param errors:   only used if encoding is true."""
+    @param encoding: decoding will be done with .decode(encoding, errors) and
+    text will be returned.
+    @param errors: passed through .decode(encoding, errors).
+    """
     fn = os.path.join("/proc", str(pid), "environ")
 
     contents: Union[str, bytes]
@@ -2471,20 +2473,13 @@ def get_proc_env(pid, encoding="utf-8", errors="replace"):
         return {}
 
     env = {}
-    null: Union[str, bytes]
-    equal: Union[str, bytes]
-    null, equal = (b"\x00", b"=")
-    if encoding:
-        null, equal = ("\x00", "=")
-        contents = contents.decode(encoding, errors)
+    null, equal = ("\x00", "=")
+    contents = contents.decode(encoding, errors)
 
-    # mypy doesn't know that the types of null, equal and contents are the same
-    # depending on the previous if branch, see:
-    # https://github.com/python/mypy/issues/6233
-    for tok in contents.split(null):  # type: ignore
+    for tok in contents.split(null):
         if not tok:
             continue
-        (name, val) = tok.split(equal, 1)  # type: ignore
+        (name, val) = tok.split(equal, 1)
         if name:
             env[name] = val
     return env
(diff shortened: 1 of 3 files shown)


cloudinit/util.py
```diff
@@ -835,7 +839,9 @@ def set_subprocess_umask_and_gid():
                 stdin=subprocess.PIPE,
                 preexec_fn=set_subprocess_umask_and_gid,
             )
-            new_fp = proc.stdin
+            # As stdin is PIPE, then proc.stdin is IO[bytes]
+            # https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stdin
+            new_fp = cast(IO[Any], proc.stdin)
```
TheRealFalcon, line 844: It looks like it's possible for `proc.stdin` to be `None`. I think a better way to solve this would be to declare `new_fp: Optional[IO]` somewhere earlier in the function, and then do a None check after setting `new_fp`.

Same comment for the next one.

cloudinit/util.py
```diff
@@ -1739,7 +1747,8 @@ def get_output_cfg(cfg, mode):
     for i in range(len(ret)):
         if not ret[i]:
             continue
-        val = ret[i].lstrip()
+        # str because the previous conditional continue
+        val = cast(str, ret[i]).lstrip()
```
TheRealFalcon, line 1751: What about changing the for loop to be something like
`if index, item in enumerate(ret):`

Mypy will understand those semantics better and not complain after seeing a `if not item`. It's also a little easier to read (imo), and then the only place the index needs to be used in the last line of the loop with `ret[index] = val`.

cloudinit/util.py
```diff
@@ -1928,10 +1937,11 @@ def mounts():
                     (dev, mp, fstype, opts, _freq, _passno) = mpline.split()
                 else:
                     m = re.search(mountre, mpline)
-                    dev = m.group(1)
-                    mp = m.group(2)
-                    fstype = m.group(3)
-                    opts = m.group(4)
+                    # safe to type-ignore because of the try-except wrapping
```
TheRealFalcon, line 1940: nit: This might be outside the context of this PR and doesn't need to change here, but I think a nicer refactor would be to remove the try/except entirely and add a None check after the `re.search()` call.

TheRealFalcon (changes requested):
I left some minor comments inline.

cloudinit/util.py
aciba90: As explained in the comment, `Popen.stdin` is `None` only if the `stdin` argument is not `PIPE`, see [1]. Thus, `None` cannot be unless we change the code in the previous lines.

I think a cast is more correct in this case, as this is mypy not being intelligent enough to detect this case, or the Python upstream typing not expressing enough.

And even if we insist in checking for `None`, what would we do in that case, raise a new exception? The code is going to crash anyway when trying to operate with `new_fp`.

[1] https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stdin

aciba90: In addition to the previous arguments, a `cast` is a type check thing and checking for `None` would incur in a runtime penalty.

cloudinit/util.py
```diff
@@ -2455,21 +2465,27 @@ def get_proc_env(pid, encoding="utf-8", errors="replace"):
     @param errors:   only used if encoding is true."""
     fn = os.path.join("/proc", str(pid), "environ")
 
+    contents: Union[str, bytes]
```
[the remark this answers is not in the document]
aciba90: Good point, I have simplified it.

cloudinit/util.py
aciba90: Good point, with this, it is much more clear.

cloudinit/util.py
aciba90: I have refactored it, dealing with more conditions that the try/except was able to catch.

cloudinit/util.py
TheRealFalcon: Good points!

TheRealFalcon (approved):
LGTM!

PR #5491, opened 2024-07-05
fix(cloudstack): Fetch metadata every boot

TheRealFalcon:
In draft state as I'd like feedback from the bug submitters as to if this fixes their issue.

## Proposed Commit Message

```
fix(cloudstack): Fetch metadata every boot

Fixes GH-5486
```

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceCloudStack.py
+++ b/cloudinit/sources/DataSourceCloudStack.py
@@ -22,6 +22,7 @@ from struct import pack
 from cloudinit import sources, subp
 from cloudinit import url_helper as uhelp
 from cloudinit import util
+from cloudinit.event import EventScope, EventType
 from cloudinit.net import dhcp
 from cloudinit.sources.helpers import ec2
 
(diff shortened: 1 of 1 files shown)


benj-n:
I assume that, as Cloudstack Datasource is not running during the init --local stage, the `EventType.BOOT_LEGACY` should be also included in this change.
Otherwise:

```
2024-07-08 18:53:47,433 - stages.py[DEBUG]: Allowed events: {<EventScope.NETWORK: 'network'>: {<EventType.BOOT_NEW_INSTANCE: 'boot-new-instance'>, <EventType.BOOT: 'boot'>}}
2024-07-08 18:53:47,433 - stages.py[DEBUG]: Event Denied: scopes=['network'] EventType=boot-legacy
2024-07-08 18:53:47,433 - stages.py[DEBUG]: No network config applied. Neither a new instance nor datasource network update allowed
```

More important is the following:
As far as I understand, as per [1], the metadata will be refreshed if and only if cloud-init is configured to manage the network config.
In the case I'm facing now, we rely on other methods to configure the network so we do have the configuration option:
```
network:
  config: disabled
```

The current code creates a dependency between cloud-init managing the network and being able to refresh the metadata at will. It doesn't look like it can be "fixed" in the DataSource and that could be considered a breaking change ... 


[1] [stages.py#L1042](https://github.com/canonical/cloud-init/blob/0af459eac9c670f2a7215ecfe96f4a4b6444eaba/cloudinit/stages.py#L1042)

PR #5495, opened 2024-07-08
Set MTU for bond parent interface

jcmoore3:
If an MTU has been provided for the bond interface, ensure it is properly configured. Support for jumbo frames requires that the underlying physical interfaces and the parent bond interface all have the larger MTU configured, not just the physical interfaces. 

## Proposed Commit Message

```
fix: Set MTU for bond parent interface

If an MTU has been provided for the bond interface, ensure it is
properly configured. Support for jumbo frames requires that the
underlying physical interfaces and the parent bond interface all
have the larger MTU configured, not just the physical interfaces.
```

## Merge type


TheRealFalcon (approved):
LGTM!

PR #5505, opened 2024-07-10
feat: Support URI sources in `write_files` module

### LINKED ISSUE #5500: [enhancement]: Support downloading file contents with `write_files` module
# Enhancement

## Proposal

We propose the addition of an extra key to the `write_files` module (something along the lines of `source`?) which would accept a URI from which to retrieve file contents. This would conflict with the existing `content` key, i.e. only one or the other would be allowed at a time.

## Reasoning

Currently, it is only possible to specify file contents inline, via the `write_files` module's `content` key. This poses a number of difficulties, notably for cases such as:
- Files produced via inter-team collaboration
  - Combining multiple cloud-configs is impractical, since either a) later `write_file` invocations with the same target would overwrite previous ones, or b) if the `append` option is set, the configuration ceases to become idempotent.
  - Some use-cases may require one team to remove or edit content created by another team, which is impossible via simple appending to the target file.
- Files which must change frequently, especially in an automated workflow
  - While automation tools could, in some cases, touch the cloud-init config directly, it is far simpler for them to expose their outputs for download by cloud-init clients.
- Large binary files
  - Encoding into base64, as is required when including binary files in a cloud-init configuration, introduces a 33% space overhead. Removing this could be beneficial for network performance.
  - From an aesthetic perspective, throwing a block of base64 into an otherwise human-readable YAML document is awkward. Including a download URL for the relevant content should be significantly cleaner.

These challenges could be (and, in some cases, likely are) accomplished via preprocessing by the cloud-init server. However, this would require the use of a template-like language embedded into the cloud-init config, which is likely to be both work-intensive and unnecessarily painful.

We envision a use case such as the following:

> In an HPC cluster, nodes are configured via cloud-init. Configurations are managed by multiple teams, who may need to touch the same sets of files (for example, an MPI config), and therefore cannot be created using multiple `write_files` invocations, as later ones would simply overwrite their earlier counterparts.
> The desired configurations could be managed by the teams in a collaborative Git repository, accessible from the cluster. During configuration, cloud-init could retrieve the desired configuration file(s) and write them to the desired location.

LRitzdorf:
## Proposed Commit Message
```
feat(write_files): support URI content sources

This change adds an optional `source` key to the `write_files` module,
allowing users to specify a URI from which to load file contents. This
facilitates more flexible multi-part configurations, as file contents
can be managed via external sources such as independent Git
repositories.

Fixes GH-5500
```

## Additional Context
Resolves #5500

## Test Steps

Unit tests, included, cover:
- `file://` and HTTP URI functionality
- Fallback behavior (i.e. use inline `content` if the provided URI fails to return usable data)
- Updated schema validation

## Merge type

### DIFF
--- a/cloudinit/config/cc_write_files.py
+++ b/cloudinit/config/cc_write_files.py
@@ -90,13 +90,17 @@ def write_files(name, files, owner: str, ssl_details: Optional[dict] = None):
         path = os.path.abspath(path)
         # Read content from provided URL, if any, or decode from inline
         contents = read_url_or_decode(
-            f_info.get("source", None), ssl_details,
-            f_info.get("content", None), f_info.get("encoding", None)
+            f_info.get("source", None),
+            ssl_details,
+            f_info.get("content", None),
+            f_info.get("encoding", None),
         )
         if contents == None:
             LOG.warning(
                 "No content could be loaded for entry %s in module %s;"
-                " skipping", i + 1, name
+                " skipping",
+                i + 1,
+                name,
             )
             continue
         # Only create the file if content exists. This will not happen, for
@@ -149,8 +153,10 @@ def read_url_or_decode(source, ssl_details, content, encoding):
             ).contents
         except Exception:
             util.logexc(
-                LOG, 'Failed to retrieve contents from source "%s";'
-                ' falling back to data from "contents" key', url
+                LOG,
+                'Failed to retrieve contents from source "%s"; falling back to'
+                ' data from "contents" key',
+                url,
             )
             use_url = False
     # If inline content is provided, and URL is not provided or is
--- a/cloudinit/config/cc_write_files.py
+++ b/cloudinit/config/cc_write_files.py
@@ -14,7 +14,6 @@ from cloudinit import url_helper, util
 from cloudinit.cloud import Cloud
 from cloudinit.config import Config
 from cloudinit.config.schema import MetaSchema
-from cloudinit.helpers import Paths
 from cloudinit.settings import PER_INSTANCE
 
 DEFAULT_PERMS = 0o644
@@ -45,7 +44,8 @@ def handle(name: str, cfg: Config, cloud: Cloud, args: list) -> None:
             name,
         )
         return
-    write_files(name, filtered_files, cloud.distro.default_owner, cloud.paths)
+    ssl_details = util.fetch_ssl_details(cloud.paths)
+    write_files(name, filtered_files, cloud.distro.default_owner, ssl_details)
 
 
 def canonicalize_extraction(encoding_type):
@@ -73,7 +73,7 @@ def canonicalize_extraction(encoding_type):
     return [TEXT_PLAIN_ENC]
 
 
-def write_files(name, files, owner: str, paths: Paths):
+def write_files(name, files, owner: str, ssl_details: dict|None = None):
     if not files:
         return
 
@@ -98,7 +98,7 @@ def write_files(name, files, owner: str, paths: Paths):
                     url,
                     retries=3,
                     sec_between=3,
-                    ssl_details=util.fetch_ssl_details(paths),
+                    ssl_details=ssl_details,
                 )
             except Exception:
                 util.logexc(
--- a/cloudinit/config/cc_write_files.py
+++ b/cloudinit/config/cc_write_files.py
@@ -87,31 +87,11 @@ def write_files(name, files, owner: str, ssl_details: dict|None = None):
             )
             continue
         path = os.path.abspath(path)
-        contents = ""
-        # Fetch file content from source URL, if provided
-        url = f_info.get("source", "")
-        use_url = bool(url)
-        if use_url:
-            try:
-                # NOTE: These retry parameters are arbitrarily chosen defaults.
-                # They have no significance, and may be changed if appropriate
-                contents = url_helper.read_file_or_url(
-                    url,
-                    retries=3,
-                    sec_between=3,
-                    ssl_details=ssl_details,
-                ).contents
-            except Exception:
-                util.logexc(
-                    LOG, "Failed to retrieve contents from source \"%s\";"
-                    " falling back to data from \"contents\" key", url
-                )
-                use_url = False
-        if not use_url:
-            # NOTE: This is not simply an "else"! Notice that `use_url` can
-            # change in the previous "if" block
-            extractions = canonicalize_extraction(f_info.get("encoding"))
-            contents = extract_contents(f_info.get("content", ""), extractions)
+        # Read content from provided URL, if any, or decode from inline
+        contents = read_url_or_decode(
+            f_info.get("source", None), ssl_details,
+            f_info.get("content", None), f_info.get("encoding", None)
+        )
         (u, g) = util.extract_usergroup(f_info.get("owner", owner))
         perms = decode_perms(f_info.get("permissions"), DEFAULT_PERMS)
         omode = "ab" if util.get_cfg_option_bool(f_info, "append") else "wb"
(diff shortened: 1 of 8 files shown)


LRitzdorf:
As I continue to work on these changes, I have a few questions for anyone with real cloud-init experience:
- Do we want to support templating of the `source` URI?
  The `phone_home` module [seems to do this](https://github.com/canonical/cloud-init/blob/18d76ac60d96186a6f89fd8b0b3ace4c70bbd174/cloudinit/config/cc_phone_home.py#L126-L130), but... only with the instance ID?
- Which, if any, of the other [kwargs supported by `readurl`](https://github.com/canonical/cloud-init/blob/18d76ac60d96186a6f89fd8b0b3ace4c70bbd174/cloudinit/url_helper.py#L354-L405) should be exposed as module parameters?
  Retry behavior and headers, in particular, stick out to me as settings that the user might want to control. On the other hand, how many parameters is too many?

LRitzdorf:
I have one new test currently, which accesses content from a `file://` URI. There should probably also be one for a "real" HTTP(S) URI as well, but I'm not sure about setting that up — would we also need to launch a lightweight HTTP server? Is there already some kind of support for this in the testing framework?

cloudinit/config/cc_write_files.py
```diff
@@ -86,8 +87,30 @@ def write_files(name, files, owner: str):
             )
             continue
         path = os.path.abspath(path)
-        extractions = canonicalize_extraction(f_info.get("encoding"))
-        contents = extract_contents(f_info.get("content", ""), extractions)
+        contents = ""
+        # Fetch file content from source URL, if provided
+        url = f_info.get("source", "")
+        use_url = bool(url)
+        if use_url:
```
TheRealFalcon, line 94: I think this new behavior would fit nicely into its own function

TheRealFalcon (changes requested):
Thanks for this contribution! Overall, things look good here. I left a few inline comments as well as answers to your questions here.

> Do we want to support templating of the source URI?

cloud-config supports [jinja templating](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html#example-cloud-config-with-instance-data), so module specific templating shouldn't be necessary (if I understood the question correctly). The cc_phone_home snippet you found is very old and likely predates the ability to use jinja.

> Which, if any, of the other kwargs supported by readurl should be exposed as module parameters?

That's a great question with no great answer. I prefer to be opinionated as to remove the number of options that need to be specified. The retry defaults you provided seem reasonable, though I'd throw a comment in specifying that they're completely arbitrary and just seemed like good defaults. It makes sense to include headers for any kind of token or basic auth, but I don't think we need more options than that. To accomodate this, I think it'd be better to make `source` a dictionary.  It can then have subkeys like `uri` which would be required and `headers` that would be optional. If in the future we decide we need more than that, we can then add them to the dictionary.

> I have one (now two) new tests currently, which access content from a file:// URI. There should probably also be one for a "real" HTTP(S) URI as well, but I'm not sure about setting that up — would we also need to launch a lightweight HTTP server? Is there already some kind of support for this in the testing framework?

For unit tests, we use [responses](https://github.com/canonical/cloud-init/blob/e0e6a427fdc6826bf7b11d52157a1c5f9b3dde4d/tests/unittests/test_url_helper.py#L86-L95) to mock the server responses. For this effort I think that should be sufficient. Can we also get a test for the fallback behavior?

Let me know if you have any extra questions!

LRitzdorf:
Thanks for the feedback — that's very useful, especially since I'm new to cloud-init. Especially the "`source` as a dictionary" bit; that'll be much cleaner. Will have updates soon!

(Also, side note: there's already a test for the fallback behavior; that's the second one I alluded to in my previous edit. It currently uses a file URI, but that could change if desired.)

TheRealFalcon:
> (Also, side note: there's already a test for the fallback behavior; that's the second one I alluded to in my previous edit. It currently uses a file URI, but that could change if desired.)

Ah, sorry, I missed that. That works!

LRitzdorf:
All right, comments addressed — and a fair bit of formatting fixed; tests under older Pythons should actually pass now!

FWIW, I did try changing the fallback test to use an HTTP URI, but that takes several seconds to fail (due to going through the OS's networking stack), so I've left it with a file URI. Not worth slowing down the tests for a tiny bit of added realism.

cloudinit/config/cc_write_files.py
```diff
@@ -86,8 +88,23 @@ def write_files(name, files, owner: str):
...
-        contents = extract_contents(f_info.get("content", ""), extractions)
+        # Read content from provided URL, if any, or decode from inline
+        contents = read_url_or_decode(
+            f_info.get("source", None),
+            ssl_details,
+            f_info.get("content", None),
+            f_info.get("encoding", None),
+        )
+        if contents == None:
+            LOG.warning(
```
TheRealFalcon, line 99: Sorry, my last comment wasn't clear, but I don't think this is what we want either.

If we have a file source, we're expecting to pull in some specific data, and if that fails, we shouldn't be leaving empty files around.

However,  `content` can be empty and if so, we should expect to write an empty file. E.g.,
```
#cloud-config
write_files:
  - path: /tmp/thing
```

With that, we'd expect `/tmp/thing` to be written and empty. As implemented here, this no longer works. So to summarize:
* If source and content, attempt source but fallback to content
* If source and no content, attempt source but do not create file if we fail to retrieve it
* If content and no source, write the content
* If no source and no content, write an empty file

doc/module-docs/cc_write_files/example6.yaml
```diff
@@ -0,0 +1,5 @@
+#cloud-config
+write_files:
+- source: https://gitlab.example.com/some_ci_job/artifacts/hello
```
TheRealFalcon, line 3: This example will need to be updated. Probably best to include both `uri` and `headers`.

cloudinit/config/cc_write_files.py
```diff
@@ -118,6 +135,40 @@ def decode_perms(perm, default):
...
+def read_url_or_decode(source, ssl_details, content, encoding):
+    contents = None
+    # Fetch file content from source URL, if provided
+    url = None if source == None else source.get("uri", None)
+    use_url = bool(url)
+    if use_url:
+        try:
+            # NOTE: These retry parameters are arbitrarily chosen defaults.
+            # They have no significance, and may be changed if appropriate
+            contents = url_helper.read_file_or_url(
```
TheRealFalcon, line 147: Nit: `content` and `contents` being used in various places in the same function can be a little messy. Maybe make this `uri_contents` or something similar?

TheRealFalcon (changes requested):
Thanks for the updates! I left a few more minor inline comments, but I think it's really close now. I'm also approving the CI runs, so check for any failures there too.

doc/module-docs/cc_write_files/example6.yaml
LRitzdorf: Ah, I fixed the schema but forgot the example. 🤦

cloudinit/config/cc_write_files.py
LRitzdorf: Ah, I somehow thought I'd handled that. Fixed now, and there's a test for this behavior as well.

LRitzdorf:
Great, thanks! I've been running tests manually on my system as well, but it looks like I missed `ruff`. That's fixed now, along with your other comments!

cloudinit/config/schemas/schema-cloud-config-v1.json
```diff
@@ -3386,6 +3386,26 @@
...
+                "type": "object",
+                "description": "Optional specification for content loading from an arbitrary URI",
+                "additionalProperties": false,
+                "properties": {
+                  "uri": {
+                    "type": "string",
+                    "format": "uri",
+                    "description": "URI from which to load file content. If loading fails repeatedly, ``content`` is used instead."
+                  },
+                  "headers": {
```
holmanb, line 3399: Cloud-init passes the value from headers to requests. Requests expects that the value should be string, bytestring, or unicode[1]. It would be ideal if we could validate that the inputs to cloud-init are valid types for requests.

If a user wants to pass an integer as a value and accidentally uses an unquoted integer (which in YAML will not be of type string), this change will benefit the user by validating the input and logging a warning about invalid configuration.

As it is, the header type isn't validated.

[1] https://requests.readthedocs.io/en/latest/user/quickstart/

holmanb (changes requested):
@LRitzdorf Thanks for this! I see a small change that we should make to the schema before merging. See my comments below.

cloudinit/config/schemas/schema-cloud-config-v1.json
LRitzdorf: Aha, good point. Should we maybe just string-ify whatever we're passed? That might be a more forgiving option if the user forgets their quotes, but I could see it causing weirdness if they forget quotes and also supply, say, a value prefixed with `0o` for some reason (which would be decoded from octal, then string-ified out in decimal). It's a case of strictness versus weirdness, I suppose.

LRitzdorf: Ah, okay, forcing string keys is much easier than I realized, and is probably the safest option anyway. Commit incoming...

holmanb (approved):
This looks fine to me. Thanks @LRitzdorf, and welcome to cloud-init!

TheRealFalcon (approved):
LGTM! Thanks for the contribution @LRitzdorf !

PR #5507, opened 2024-07-11
ubuntu/jammy: update changelog from SRU comments

TheRealFalcon:
DO NOT SQUASH


blackboxsw (approved):
Changeset looks good, we'll have to retag this downstream for the CL difference but otherwise good. 
Looks like all the dropped changelog entries were already released in 24.1.1 so no surprise there on dropping them from this upload as they are irrelevant. 

We'll sort how I copy-pasta'd the already released content (and a devel changelog entry separately)
+1h

PR #5509, opened 2024-07-12
refactor and mypy up ibm datasource

a-dubs:
## Proposed Commit Message

```
refactor: refactor and fix check_untyped_defs in DataSourceIBMCloud.py
    
GH-5445
```

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceIBMCloud.py
+++ b/cloudinit/sources/DataSourceIBMCloud.py
@@ -96,6 +96,7 @@ import base64
 import json
 import logging
 import os
+from typing import Any, Callable, Dict, Optional, Tuple
 
 from cloudinit import atomic_helper, sources, subp, util
 from cloudinit.sources.helpers import openstack
@@ -176,7 +177,7 @@ class DataSourceIBMCloud(sources.DataSource):
             # environment handles networking configuration. Not cloud-init.
             return {"config": "disabled", "version": 1}
         if self._network_config is None:
-            if self.network_json is not None:
+            if self.network_json not in (sources.UNSET, None):
                 LOG.debug("network config provided via network_json")
                 self._network_config = openstack.convert_net_json(
                     self.network_json, known_macs=None
@@ -186,7 +187,12 @@ class DataSourceIBMCloud(sources.DataSource):
         return self._network_config
 
 
-def _read_system_uuid():
+def _read_system_uuid() -> Optional[str]:
+    """
+    Read the system uuid.
+
+    @return: the system uuid or None if not available.
+    """
     uuid_path = "/sys/hypervisor/uuid"
     if not os.path.isfile(uuid_path):
         return None
@@ -194,6 +200,9 @@ def _read_system_uuid():
 
 
 def _is_xen():
+    """
+    Return boolean indicating if this is a xen hypervisor.
+    """
     return os.path.exists("/proc/xen")
 
 
@@ -201,7 +210,7 @@ def _is_ibm_provisioning(
     prov_cfg="/root/provisioningConfiguration.cfg",
     inst_log="/root/swinstall.log",
     boot_ref="/proc/1/environ",
-):
+) -> bool:
     """Return boolean indicating if this boot is ibm provisioning boot."""
     if os.path.exists(prov_cfg):
         msg = "config '%s' exists." % prov_cfg
@@ -229,7 +238,7 @@ def _is_ibm_provisioning(
     return result
 
 
-def get_ibm_platform():
+def get_ibm_platform() -> Tuple[Optional[str], Optional[str]]:
     """Return a tuple (Platform, path)
 
     If this is Not IBM cloud, then the return value is (None, None).
@@ -242,7 +251,7 @@ def get_ibm_platform():
         return not_found
 
     # fslabels contains only the first entry with a given label.
-    fslabels = {}
+    fslabels: Dict[str, Dict] = {}
     try:
         devs = util.blkid()
     except subp.ProcessExecutionError as e:
@@ -289,7 +298,7 @@ def get_ibm_platform():
     return not_found
 
 
-def read_md():
+def read_md() -> Optional[Dict[str, Any]]:
     """Read data from IBM Cloud.
 
     @return: None if not running on IBM Cloud.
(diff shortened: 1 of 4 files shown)


TheRealFalcon (commented):
Nice updates!

I think we can simplify the metadata transformation a bit more. Would something like this work?
```diff
diff --git a/cloudinit/sources/DataSourceIBMCloud.py b/cloudinit/sources/DataSourceIBMCloud.py
index c86d68516..05ca8f056 100644
--- a/cloudinit/sources/DataSourceIBMCloud.py
+++ b/cloudinit/sources/DataSourceIBMCloud.py
@@ -341,7 +341,7 @@ def metadata_from_dir(source_dir: str) -> Dict[str, Any]:
     keys.
 
     This function has a lot in common with ConfigDriveReader.read_v2 but
-    there are a number of inconsistencies, such as key renames and only 
+    there are a number of inconsistencies, such as key renames and only
     presenting a 'latest' version, which make it an unlikely candidate to share
     code.
 
@@ -356,11 +356,11 @@ def metadata_from_dir(source_dir: str) -> Dict[str, Any]:
         return json.loads(blob.decode("utf-8"))
 
     def load_file(
-        path: str, translator: Optional[Callable[[bytes], Any]] = None
+        path: str, translator: Callable[[bytes], Any]
     ) -> Any:
         try:
             raw = util.load_binary_file(path)
-            return translator(raw) if translator else raw
+            return translator(raw)
         except IOError as e:
             LOG.debug("Failed reading path '%s': %s", path, e)
             return None
@@ -370,32 +370,27 @@ def metadata_from_dir(source_dir: str) -> Dict[str, Any]:
     files = [
         # tuples of (results_name, path, translator)
         ("metadata_raw", opath("meta_data.json"), load_json_bytes),
-        ("userdata", opath("user_data"), None),
+        ("userdata", opath("user_data"), lambda x: x),
         ("vendordata", opath("vendor_data.json"), load_json_bytes),
         ("networkdata", opath("network_data.json"), load_json_bytes),
     ]
 
-    raw_results: Dict[str, Optional[bytes]] = {}
-    translated_results: Dict[str, Any] = {}
+    results: Dict[str, Any] = {}
 
     for name, path, transl in files:
         fpath = os.path.join(source_dir, path)
-        raw_results[name] = load_file(fpath)
-        if raw_results[name] is not None and transl is not None:
-            translated_results[name] = load_file(fpath, transl)
-        else:
-            translated_results[name] = raw_results[name]
+        results[name] = load_file(fpath, transl)
 
-    if raw_results["metadata_raw"] is None:
+    if results["metadata_raw"] is None:
         raise sources.BrokenMetadata(
             "%s missing required file 'meta_data.json'",
             source_dir,
         )
 
-    translated_results["metadata"] = {}
+    results["metadata"] = {}
 
-    md_raw = translated_results["metadata_raw"]
-    md = translated_results["metadata"]
+    md_raw = results["metadata_raw"]
+    md = results["metadata"]
 
     if "random_seed" in md_raw:
         try:
@@ -416,7 +411,7 @@ def metadata_from_dir(source_dir: str) -> Dict[str, Any]:
         if old_key in md_raw:
             md[new_key] = md_raw[old_key]
 
-    return translated_results
+    return results
 
 
 # Used to match classes to dependencies
```

a-dubs:
@TheRealFalcon Thank you for the patch! I really appreciate that since i was definitely still not nearly satisfied with how messy that was even after the refactor.

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -186,22 +187,30 @@ def network_config(self):
         return self._network_config
 
 
-def _read_system_uuid():
+def _read_system_uuid() -> Optional[str]:
+    """
+    Read the system uuid.
+
+    @return: the system uuid or None if not available.
```
TheRealFalcon, line 194: We decided to standarized on the [Sphix format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format), which should use `:return:`

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -325,81 +334,82 @@ def read_md():
...
-        results[name] = data
+    for name, path, transl in files:
+        fpath = os.path.join(source_dir, path)
+        results[name] = load_file(fpath, transl)
 
-    if results.get("metadata_raw") is None:
+    if results["metadata_raw"] is None:
         raise sources.BrokenMetadata(
-            "%s missing required file 'meta_data.json'" % source_dir
+            "%s missing required file 'meta_data.json'",
```
TheRealFalcon, line 386: The comma syntax only works for logging because the logging module has some smarts to not interpolate the strings if it's not gonna output the log anywhere.

`%` is correct here, but since we're modifying it anyway, we could update it to be an f-string instead.

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -325,81 +334,82 @@ def read_md():
...
     md_raw = results["metadata_raw"]
     md = results["metadata"]
+
     if "random_seed" in md_raw:
         try:
             md["random_seed"] = base64.b64decode(md_raw["random_seed"])
         except (ValueError, TypeError) as e:
             raise sources.BrokenMetadata(
-                "Badly formatted metadata random_seed entry: %s" % e
+                "Badly formatted metadata random_seed entry: %s",
```
TheRealFalcon, line 400: Same comment about the comma

TheRealFalcon (changes requested):
I have a few more minor comments inline.

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -325,27 +334,39 @@ def read_md():
...
     @return: Dict containing translated metadata, userdata, vendordata,
         networkdata as present.
     """
 
-    def opath(fname):
+    def opath(fname: str) -> str:
         return os.path.join("openstack", "latest", fname)
 
-    def load_json_bytes(blob):
+    def load_json_bytes(blob: bytes) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
a-dubs: sorry I'm somewhat unclear what you're referring to. are you suggesting the following?
```suggestion
    def load_json_bytes(blob: bytes) -> Dict[str, Union[List,str]]:
```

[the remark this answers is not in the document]
TheRealFalcon: The return type is specified as `Dict[...]`, but `json.loads` can return `List` and `str` too. Changing it here would likely will make the code that calls this more annoying, and I think it's ok to assume we're going to get a dict. We may just want to add a comment that this may not be technically correct.

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -325,81 +336,89 @@ def read_md():
...
+        """
+        Load JSON from a byte string.
+
+        This technically could return a list or a str, but we are only
+        assuming a dict here.
+
+        :param blob: The byte string to load JSON from.
+        :return: The loaded JSON object.
+        """
```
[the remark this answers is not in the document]
TheRealFalcon: Works for me!

TheRealFalcon:
@a-dubs , the changes here look good to me, but you still have some linter failures to fix first

tests/unittests/sources/test_ibmcloud.py
```diff
@@ -272,6 +272,8 @@ def test_template_live(self, m_platform, m_sysuuid):
         )
 
         ret = ibm.read_md()
+        if ret is None:  # this is needed for mypy - ensures ret is not None
+            self.fail("read_md returned None unexpectedly")
```
a-dubs, line 276: @TheRealFalcon is there a better way I should be doing this? I think this is pretty okay but just curious. thanks!

TheRealFalcon: I'm tempted to disable mypy for tests, but for mypy complaining this looks good!

a-dubs: Well now that you mention it, mypy is supposed to be disabled for tests. All I did in this PR was remove the line disabling mypy checks for the ibm datasource file, not the unit test file for it.

cloudinit/sources/DataSourceIBMCloud.py
```diff
@@ -308,7 +308,7 @@ def read_md() -> Optional[Dict[str, Any]]:
              and optional fields: userdata, vendordata, networkdata.
     Also includes the system uuid from /sys/hypervisor/uuid."""
     platform, path = get_ibm_platform()
-    if platform is None:
+    if platform is None or path is None:
```
TheRealFalcon, line 311: Is this actually true? Based on a quick look at `get_ibm_platform()`, it looks like if platform is in provisioning state then `path` can be None. Also, are we sure there aren't times when we'd want to hit the RuntimeError later if for some other reason `path` winds up being None for some other reason (if that's possible)?

TheRealFalcon:
LGTM except there's a lint check failing.

TheRealFalcon (approved):
Thanks!

PR #5510, opened 2024-07-12
Allow ability to specify fqdn via metadata

jcmoore3:
Similar to hostname, also allow fqdn to be specified via metadata

## Proposed Commit Message

```
fix: Allow ability to specify fqdn via metadata

Similar to hostname, also allow fqdn to be specified via metadata.
In non-DHCP use cases where static IP information is set via network_config and
no search domains are available network_config or DHCP, there is no way to set
the FQDN of the host, except for specifying hostname itself as a FQDN. In many
cases, it is desired to keep hostname as a non-FQDN while also allowing for a
FQDN. Prior to this change, fqdn could be specified via cloud.cfg but this was
not dynamically configurable via meta_data.
```

## Additional Context
In `get_hostname_fqdn()`, it seems that static config is first checked for fqdn and if it is not present, then the cloud information should be queried:
(https://github.com/canonical/cloud-init/blob/e0e6a427fdc6826bf7b11d52157a1c5f9b3dde4d/cloudinit/util.py#L1263-L1279)

This change extends this functionality and will use the fqdn data from the cloud metadata (if provided) before finally falling through to the hostname from the cloud metadata.

## Merge type

### DIFF
--- a/cloudinit/sources/__init__.py
+++ b/cloudinit/sources/__init__.py
@@ -859,6 +859,7 @@ class DataSource(CloudInitPickleMixin, metaclass=abc.ABCMeta):
             # if there is an ipv4 address in 'local-hostname', then
             # make up a hostname (LP: #475354) in format ip-xx.xx.xx.xx
             lhost = self.metadata["local-hostname"]
+            metadata_fqdn = self.metadata.get("fqdn", None)
             if net.is_ipv4_address(lhost):
                 toks = []
                 if resolve_ip:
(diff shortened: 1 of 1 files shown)


holmanb:
Cloud-init already supports hostname and fqdn in vendor-data, which is provided by the cloud. A vendor-data could provide the following to configure hostname and fqdn:

```yaml
#cloud-config
hostname: myhost
fqdn: myhost.example.com
```

Is there some reason that using this feature that is already supported won't work for you?

jcmoore3:
Thanks for your reply.

In my experience (at least with OpenStack), the `vendor_data.json` file is static and global to the entire cloud (region), ie it does not contain instance specific information.  The `vendor_data2.json` file can contain dynamic information, if certain additional external hooks are setup within the infrastructure.

In my case, I'm looking for a way to configure the FQDN via an instance specific data source, such as `meta_data.json`.

holmanb:
> least with OpenStack), the `vendor_data.json` file is static and global to the entire cloud (region), ie it does not contain instance specific information. The `vendor_data2.json` file can contain dynamic information, if certain additional external hooks are setup wit

I see, thanks for elaborating. I'm hesitant to set the precedent that meta-data should provide instance-specific vendor-data, and sounds like openstack already provides this capability.

PR #5515, opened 2024-07-16
Improve nocloud detection / improve logging

holmanb:
See individual commits, relates to https://github.com/canonical/cloud-init/issues/5389

PR #5521, opened 2024-07-17
Update NoCloud docs

holmanb:
## Proposed Commit Message
```
doc(NoCloud): Categorize the different configuration types

Formally document providing runtime configuration in system configuration.
Introduce names to identify previously unnamed NoCloud concepts.
Add more structure - discrete sections for:
- runtime configuration types
- discovery configuration
- configuration sources
```

## Additional context

Related to https://github.com/canonical/cloud-init/pull/5515

I'm not opposed to changing the new names if better ones are proposed.

I think that it is important to name and group different classes of configuration to create a more structured model for users (and developers) to think about cloud-init.

### DIFF
--- a/doc/rtd/reference/datasources/nocloud.rst
+++ b/doc/rtd/reference/datasources/nocloud.rst
@@ -22,14 +22,14 @@ meta-configuration can be delivered to cloud-init in different ways, but is
 different from the configurations that cloud-init uses to configure the
 instance at runtime.
 
-user-data
+user data
 ---------
 
-User-data is a :ref:`configuration format<user_data_formats>` that allows a
+User data is a :ref:`configuration format<user_data_formats>` that allows a
 user to configure an instance.
 
-meta data
----------
+metadata
+--------
 
 The ``meta-data`` file is a YAML-formatted file.
 
@@ -37,7 +37,7 @@ vendor data
 -----------
 
 Vendor data may be used to provide default cloud-specific configurations which
-may be overriden by user-data. This may be useful, for example, to configure an
+may be overriden by user data. This may be useful, for example, to configure an
 instance with a cloud provider's repository mirror for faster package
 installation.
 
@@ -49,7 +49,7 @@ cloud-specific network configurations, or a reasonable default is set by
 cloud-init (typically cloud-init brings up an interface using DHCP).
 
 Since NoCloud is a generic datasource, network configuration may be set the
-same way as user-data, meta data, vendor data.
+same way as user data, metadata, vendor data.
 
 See the :ref:`network configuration<network_config>` documentation for
 information on network configuration formats.
@@ -122,7 +122,7 @@ files which are stored in :file:`/etc/cloud/cloud.cfg.d`.
 Configuration sources
 =====================
 
-User-data, meta data, network config, and vendor data may be sourced from one
+User-data, metadata, network config, and vendor data may be sourced from one
 of several possible locations, either locally or remotely.
 
 Source 1: Local filesystem
--- a/doc/rtd/reference/datasources/nocloud.rst
+++ b/doc/rtd/reference/datasources/nocloud.rst
@@ -4,97 +4,237 @@ NoCloud
 *******
 
 The data source ``NoCloud`` is a flexible datasource that can be used in
-multiple different ways. With NoCloud, one can provide configurations to
-the instance without running a network service (or even without having a
-network at all). Alternatively, one can use HTTP/HTTPS or FTP/FTPS to provide
-a configuration.
+multiple different ways.
 
-Configuration Methods:
+With NoCloud, one can provide configuration to the instance locally (without
+network access) or alternatively NoCloud can fetch the configuration from a
+remote server.
+
+Much of the following documentation describes how to tell cloud-init where
+to get its configuration.
+
+Runtime configurations
 ======================
 
-.. warning::
-    User data placed under ``/etc/cloud/`` will **not** be recognized as a
-    source of configuration data by the NoCloud datasource. While it may
-    be acted upon by cloud-init, using
-    :ref:`DataSourceNone<datasource_none_example>` should be preferred.
+Cloud-init discovers four types of configuration at runtime. The source of
+these configuration types is configurable with a meta-configuration. This
+meta-configuration can be delivered to cloud-init in different ways, but is
+different from the configurations that cloud-init uses to configure the
+instance at runtime.
 
-Method 1: Labeled filesystem
+user-data
+---------
+
+User-data is a :ref:`configuration format<user_data_formats>` which allows a
+user to configure an instance.
+
+meta-data
+---------
+
+The ``meta-data`` file is a YAML-formatted file. This is a required
+configuration type.
+
+vendor-data (optional)
+----------------------
+
+Vendor-data may be used to provide default cloud-specific configurations which
+may be overriden by user-data. This may be useful, for example, configure an
+instance with a cloud provider's repository mirror for faster package
+installation.
+
+network-config (optional)
+-------------------------
+
+Network configuration typically comes from the cloud provider to set
+cloud-specific network configurations or a reasonable default is set by
+cloud-init (typically cloud-init brings up an interface using DHCP).
+
+Since NoCloud is a generic datasource, network configuration may be set the
+same way as user-data, meta-data, vendor-data.
+
+See the :ref:`network configuration<network_config>` documentation for
+information on network configuration formats.
+
+Meta-configuration
+==================
+
+The purpose of the meta-configuration is to tell cloud-init where it can find
+the runtime configurations described above.
+
+There are two methods for cloud-init to receive a meta-configuration.
+
+Method 1: Line configuration
 ----------------------------
 
-A labeled `vfat`_ or `iso9660` filesystem may be used. The filesystem volume
-must be labelled ``CIDATA``.
+The "line configuration" is a single string of text which is passed to an
+instance at boot time via either the kernel command line or in the serial
+number exposed via DMI (sometimes called SMBIOS).
 
+Example: ::
 
-Method 2: Custom webserver
+  ds=nocloud;s=https://10.42.42.42/configs/
+
+In the above line configuration, ``ds=nocloud`` tells cloud-init to use the
+NoCloud datasource, and ``s=https://10.42.42.42/configs/`` tells cloud-init to
+fetch configurations using ``https`` from the URI
+``https://10.42.42.42/configs/``.
+
+We will describe the possible values in a line configuration in the following
+sections.
+
+.. note::
+
+   If using kernel command line arguments with GRUB, note that an
+   unescaped semicolon is intepreted as the end of a statement.
+   See: `GRUB quoting`_
+
+Method 2: System configuration
+------------------------------
+
+System configurations are YAML-formatted files and have names that end in
+``.cfg``. These are located under :file:`/etc/cloud/cloud.cfg.d/`.
+
+Example:
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      seedfrom: https://10.42.42.42/configs/
+
+The above system configuration tells cloud-init that it is using NoCloud and
+that it can find configurations at ``https://10.42.42.42/configs/``.
+
+The scope of this section is limited to its use for selecting the source of
+its configuration, however it is worth mentioning that the system configuration
+provides more than just the meta-configuration.
+
+In addition to defining where cloud-init can find runtime configurations, the
+system configuration also controls many of cloud-init's default behaviors.
+Most users shouldn't need to modify these defaults, however it is worth noting
+that downstream distributions often use them to set reasonable default
+behaviors for cloud-init. This includes things such as which distro to behave
+as and which networking backend to use.
+
+The default values in :file:`/etc/cloud/cloud.cfg` may be overriden by drop-in
+files which are stored in :file:`/etc/cloud/cloud.cfg.d`.
+
+Configuration sources
+=====================
+
+User-data, meta-data, network-config (optional), and vendor-data
+(optional) may be sourced from one of several possible locations, either
+locally or remotely.
+
+Source 1: Local filesystem
 --------------------------
 
-Configuration files can be provided to cloud-init over HTTP(s). To tell
-cloud-init the URI to use, arguments must be passed to the instance via the
-kernel command line or SMBIOS serial number. This argument might look like: ::
+System configuration may provide cloud-init runtime configuration directly
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      meta-data: |
+        instance-id: l-eadfbe
+      user-data: |
+        #cloud-config
+        runcmd: [ echo "it worked!" > /tmp/example.txt ]
+
+Local filesystem: custom location
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Cloud-init makes it possible to find system configuration in a custom
+filesystem path for those that require more flexibility. This may be
+done with a line configuration: ::
+
+
+  ds=nocloud;s=file://path/to/directory/
+
+Or a system configuration:
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      seedfrom: file://path/to/directory
+
+Source 2: Drive with labeled filesystem
+---------------------------------------
+
+A labeled `vfat`_ or `iso9660` filesystem may be used. The filesystem volume
+must be labelled ``CIDATA``. The configuration files must be in the root
+directory of the filesystem.
+
+Source 3: Custom webserver
+--------------------------
+
+Configuration files can be provided to cloud-init over HTTP(S) using a
+line configuration: ::
 
   ds=nocloud;s=https://10.42.42.42/cloud-init/configs/
 
-.. note::
-   If using kernel command line arguments with GRUB, note that an
-   unescaped semicolon is intepreted as the end of a statement.
-   Consider using single-quotes to avoid this pitfall. See: `GRUB quoting`_
-   ds=nocloud;s=http://10.42.42.42/cloud-init/configs/
+or using system configuration:
 
-Alternatively, this URI may be defined in a configuration in a file
-:file:`/etc/cloud/cloud.cfg.d/*.cfg` like this: ::
+.. code-block:: yaml
 
   datasource:
     NoCloud:
       seedfrom: https://10.42.42.42/cloud-init/configs/
 
-Method 3: FTP Server
+Source 4: FTP Server
 --------------------
 
 Configuration files can be provided to cloud-init over unsecured FTP
-or alternatively with FTP over TLS. To tell cloud-init the URL to use,
-arguments must be passed to the instance via the kernel command line or SMBIOS
-serial number. This argument might look like: ::
+or alternatively with FTP over TLS using a line configuration ::
 
   ds=nocloud;s=ftps://10.42.42.42/cloud-init/configs/
 
-Alternatively, this URI may be defined in a configuration in a file
-:file:`/etc/cloud/cloud.cfg.d/*.cfg` like this: ::
+or using system configuration
+
+.. code-block:: yaml
 
   datasource:
     NoCloud:
       seedfrom: ftps://10.42.42.42/cloud-init/configs/
 
-Method 4: Local filesystem
---------------------------
+Source files
+------------
 
-Configuration files can be provided on the local filesystem at specific
-filesystem paths using kernel command line arguments or SMBIOS serial number to
-tell cloud-init where on the filesystem to look.
+The base path pointed to by the URI in the above sources can provide content
+using the following final path components:
 
-.. note::
-   Unless arbitrary filesystem paths are required, one might prefer to use
-   :ref:`DataSourceNone<datasource_none_example>`, since it does not require
-   modifying the kernel command line or SMBIOS.
+* ``user-data`` (required)
+* ``meta-data`` (required)
+* ``vendor-data`` (optional)
+* ``network-config`` (optional)
 
-This argument might look like: ::
+For example, if the ``seedfrom`` value of ``seedfrom`` is
+``https://10.42.42.42/``, then the following files will be fetched from the
+webserver at first boot:
 
-  ds=nocloud;s=file://path/to/directory/
+.. code-block:: sh
 
-Alternatively, this URI may be defined in a configuration in a file
-:file:`/etc/cloud/cloud.cfg.d/*.cfg` like this: ::
+    https://10.42.42.42/user-data
+    https://10.42.42.42/vendor-data
+    https://10.42.42.42/meta-data
+    https://10.42.42.42/network-config
 
-  datasource:
-    NoCloud:
-      seedfrom: file://10.42.42.42/cloud-init/configs/
+If the required files don't exist, this datasource will be skipped.
 
 
-Permitted keys
-==============
+Line configuration in detail
+============================
 
-Currently three keys (and their aliases) are permitted for configuring
-cloud-init.
+The line configuration has several options.
 
-The only required key is:
+Permitted keys (DMI and kernel command line)
+--------------------------------------------
+
+Currently three keys (and their aliases) are permitted in cloud-init's kernel
+command line and DMI (sometimes called SMBIOS) serial number.
+
+There is only one required key in a line configuration:
 
 * ``seedfrom`` alias: ``s``
 
@@ -108,11 +248,6 @@ be removed in the future.
   preferred)
 * ``instance-id`` alias: ``i``  (set instance id  in :file:`meta-data` instead)
 
-.. note::
-
-   The aliases ``s`` , ``h`` and ``i`` are only supported by kernel
-   command line or SMBIOS. When configured in a ``*.cfg`` file, the long key
-   name is required.
 
 Seedfrom: HTTP and HTTPS
 ------------------------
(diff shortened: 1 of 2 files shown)


doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
-.. warning::
-    User data placed under ``/etc/cloud/`` will **not** be recognized as a
-    source of configuration data by the NoCloud datasource. While it may
-    be acted upon by cloud-init, using
-    :ref:`DataSourceNone<datasource_none_example>` should be preferred.
+Cloud-init discovers four types of configuration at runtime. The source of
+these configuration types is configurable with a meta-configuration. This
+meta-configuration can be delivered to cloud-init in different ways, but is
+different from the configurations that cloud-init uses to configure the
+instance at runtime.
```
s-makin, line 23: I found this quite confusing - even after reading it a few times, I'm still a bit lost. I *think* what it's saying here is:
"There are four configuration types that cloud-init uses to configure the instance at runtime. The source of each type can be set using a meta-configuration, which can be delivered to cloud-init in two ways."
(let me know if I've understood that wrongly)

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+
+user-data
+---------
+
+User-data is a :ref:`configuration format<user_data_formats>` which allows a
```
s-makin, line 28:
```suggestion
User-data is a :ref:`configuration format<user_data_formats>` that allows a
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+---------
+
+The ``meta-data`` file is a YAML-formatted file. This is a required
+configuration type.
+
+vendor-data (optional)
+----------------------
+
+Vendor-data may be used to provide default cloud-specific configurations which
+may be overriden by user-data. This may be useful, for example, configure an
```
s-makin, line 41:
```suggestion
may be overriden by user-data. This may be useful, for example, to configure an
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+instance with a cloud provider's repository mirror for faster package
+installation.
+
+network-config (optional)
+-------------------------
+
+Network configuration typically comes from the cloud provider to set
+cloud-specific network configurations or a reasonable default is set by
```
s-makin, line 49:
```suggestion
cloud-specific network configurations, or a reasonable default is set by
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+System configurations are YAML-formatted files and have names that end in
+``.cfg``. These are located under :file:`/etc/cloud/cloud.cfg.d/`.
+
+Example:
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      seedfrom: https://10.42.42.42/configs/
```
s-makin, line 103: I think this section needs an additional space indentation to make the code block render correctly

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      meta-data: |
+        instance-id: l-eadfbe
+      user-data: |
+        #cloud-config
+        runcmd: [ echo "it worked!" > /tmp/example.txt ]
```
s-makin, line 142: This code block may also need another space in indentation

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
 
-Alternatively, this URI may be defined in a configuration in a file
-:file:`/etc/cloud/cloud.cfg.d/*.cfg` like this: ::
+Or a system configuration:
+
+.. code-block:: yaml
+
+  datasource:
+    NoCloud:
+      seedfrom: file://path/to/directory
```
s-makin, line 160: Same here

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -138,26 +273,41 @@ Where ``scheme`` can be ``ftp`` or ``ftps``, ``userinfo`` will be
...
+advantages and disadvantages to each option, neither is clearly a better
+choice - so it is left to the user to decide.
+
+Line configuration
+------------------
 
-The path pointed to by the URI can contain the following
-files:
+Advantages
+^^^^^^^^^^
```
s-makin, line 288: I would probably not use sub-subheaders here - there aren't enough bullet points to make it a true section, so I'd probably turn the Advantages/Disadvantages in each case into regular text, but in bold.

s-makin (commented):
Sorry, I thought I was adding my comments in a single review, turns out not >< anyway, added a few comments etc

doc/rtd/reference/datasources/nocloud.rst
aciba90: I agree. We are introducing yet another new term `meta-configuration`, which in https://docs.cloud-init.io/en/latest/explanation/configuration.html#base-configuration is called base-configuration. For the sake of simplicity, could we standardize on only one name?

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
-    source of configuration data by the NoCloud datasource. While it may
-    be acted upon by cloud-init, using
-    :ref:`DataSourceNone<datasource_none_example>` should be preferred.
+Cloud-init discovers four types of configuration at runtime. The source of
+these configuration types is configurable with a meta-configuration. This
+meta-configuration can be delivered to cloud-init in different ways, but is
+different from the configurations that cloud-init uses to configure the
+instance at runtime.
+
+user-data
```
aciba90, line 25: For symmetry with vendor-data:

```suggestion
user-data (optional)
--------------------
```

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -138,26 +273,41 @@ Where ``scheme`` can be ``ftp`` or ``ftps``, ``userinfo`` will be
...
+Advantages
+^^^^^^^^^^
 
-If the seedfrom URI doesn't contain the required files, this datasource
-will be skipped.
+* simple: requires only modifying a file
 
-The ``user-data`` file uses :ref:`user data format<user_data_formats>`. The
-``meta-data`` file is a YAML-formatted file.
+Disadvantates
```
aciba90, line 307:
```suggestion
Disadvantages
```

doc/rtd/reference/datasources/nocloud.rst
holmanb: > "There are four configuration types that cloud-init uses to configure the instance at runtime. The source of each type can be set using a meta-configuration, which can be delivered to cloud-init in two ways."
(let me know if I've understood that wrongly)

+1  yes that is the correct understanding - I'm happy to use the wording you've suggested if that is preferred.

> I agree. We are introducing yet another new term meta-configuration, which in https://docs.cloud-init.io/en/latest/explanation/configuration.html#base-configuration is called base-configuration.

I wouldn't call these two the same thing. Base configuration maps to what I call "system configuration" in this page. This new term was introduced to describe something that both the "line configuration" and "base configuration" are capable of providing - that is, the source of each type.

> For the sake of simplicity, could we standardize on only one name?

I'm happy to cross-link and standardize (either base configuration or system configuration works for me) with the page you linked.

@s-makin @aciba90 thoughts?

doc/rtd/reference/datasources/nocloud.rst
holmanb: I agree that this symmetry would be nice, however it [that's not how it works](https://github.com/canonical/cloud-init/blob/550c685c98551f65c30832b186fe091721b48477/cloudinit/sources/DataSourceNoCloud.py#L92). This implementation choice was a little odd from a UI perspective, since user-data can actually be empty.

Maybe it would be best to standardize this interface and allow all of them to be optional?

doc/rtd/reference/datasources/nocloud.rst
TheRealFalcon: [user data](https://cloudinit.readthedocs.io/en/latest/explanation/format.html), [vendor data](https://cloudinit.readthedocs.io/en/latest/explanation/vendordata.html), [metadata](https://cloudinit.readthedocs.io/en/latest/explanation/instancedata.html), and [network config](https://cloudinit.readthedocs.io/en/latest/reference/network-config.html) already have their own pages, so I'm not sure we need to re-explain them. Can we instead just link to them and explain where they need to go?

doc/rtd/reference/datasources/nocloud.rst
TheRealFalcon: meta data and user data are sometimes optional depending on the source. It's maybe not worth documenting separately, but e.g., when using the kernel cmdline, I don't think you need either.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+Cloud-init discovers four types of configuration at runtime. The source of
+these configuration types is configurable with a meta-configuration. This
+meta-configuration can be delivered to cloud-init in different ways, but is
+different from the configurations that cloud-init uses to configure the
+instance at runtime.
+
+user-data
+---------
+
+User-data is a :ref:`configuration format<user_data_formats>` that allows a
```
TheRealFalcon, line 28: For all the sections here: https://cloudinit.readthedocs.io/en/latest/development/style_docs.html#common-words

doc/rtd/reference/datasources/nocloud.rst
holmanb: > meta data and user data are sometimes optional depending on the source.

Thanks for the clarification @TheRealFalcon, and my apologies @aciba90 you were right too - we were describing the behavior of different sources. I guess I missed that they behave differently for different sources. That's.... not at all intuitive. It looks like wasn't always that way either - in https://github.com/canonical/cloud-init/commit/c1253945761c33bfa89289a63cdb8799fc18d019 user-data and meta-data changed from being optional to being required for labeled filesystems and for the local filesystem. Unfortunately the commit message doesn't even mention this behavior change so we have no justification for why these two sources behave differently.

> when using the kernel cmdline, I don't think you need either.

Confirmed. On further examination of the code it appears that all four types are optional for the following sources: kernel command line, DMI, system configuration at the top level[1], system configuration nested under the NoCloud key[2], HTTP, FTP.

Unless we think that there is a good reason to require I'm tempted to standardize this behavior so that we can have a more intuitive and user-friendly user-interface. Currently a local datasource that doesn't provide a meta-data file or a user-data file is already on a failure path, so just silently removing this superficial requirement would probably be safe to include in SRU. While this still would technically be required for users of nested system configuration in order for ds-identify to identify the system (in the case that no override is provided `ds=nocloud` isn't in the line configuration and `datasource_list` has >1 datasources included), at least cloud-init's Python code wouldn't be imposing a false requirement.

@aciba90 @TheRealFalcon Thoughts?

[1] note, however, that ds-identify nocloud detection is completely broken and always has been, however these configurations will still by applied by cloud-init if None is included in the datasource list (it usually is unless only a single datasource is specified)
[2] however ds-identify nocloud detection fails unless user-data and meta-data are both provided

doc/rtd/reference/datasources/nocloud.rst
s-makin: I think linking to them makes sense, but the links should come with context, otherwise readers might not know which type they're interested in. I'd be happy for us to leave the subsections as they are to provide this info at the point where users might be looking for it, but with the links added to show where to find more detailed info.

Re: standardizing on terms, I'm always in favour of simplifying and using the minimum amount of jargon possible. I don't have strong opinions on which term is better, so happy to leave that to your discretion, but I agree with Alberto that we should pick one and use that everywhere. When you say "line configuration", I don't know if I know what that means and haven't seen that term used elsewhere. Do you mean passing config via the command line? If so, I'm not sure that "line configuration" describes it in a satisfactory way because it feels like it could mean a few different things.

holmanb: > When you say "line configuration", I don't know if I know what that means and haven't seen that term used elsewhere.

It is a new term. Please let me know if there is something missing from the current written definition.

In cloud-init we have a way of handling a configuration that gets passed either on the kernel command line or SMBIOS. Both sources of line configuration use the same format, but nobody ever bothered to give this configuration format a name, so any time we want to reference it, we have to call it "the configuration format that gets accepted by BIOS/DMI and kernel command line". Using this mouthful of a phrase over and over again is tiring to parse, and if a user is unfamiliar with one of the two modes of providing line configuration then it becomes even more labour intensive to grok the passage that the phrase gets used in.

> Do you mean passing config via the command line?

That is one of the ways that a line configuration can be passed, yes, but it is broader than that.

> If so, I'm not sure that "line configuration" describes it in a satisfactory way because it feels like it could mean a few different things.

The intent is to have a term that could mean multiple things because a line configuration is a configuration format that can also be provided as a serial number through smbios/DMI.

doc/rtd/reference/datasources/nocloud.rst
aciba90: > I'm tempted to standardize this behavior so that we can have a more intuitive and user-friendly user-interface.

Just to confirm, you want to try to standardize on making user-data and meta-data optional for every possible ways of configuring NoCloud, right? If that's the case, I think it is worth the implementation and SRU effort.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+
+   datasource:
+     NoCloud:
+       seedfrom: file://path/to/directory
+
+Source 2: Drive with labeled filesystem
+---------------------------------------
+
+A labeled `vfat`_ or `iso9660` filesystem may be used. The filesystem volume
+must be labelled ``CIDATA``. The configuration files must be in the root
```
aciba90, line 166: What are the expected / allowed names of those files? Are they the same as in `reference/datasources/nocloud.html#source-files`?

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
-Configuration files can be provided to cloud-init over HTTP(s). To tell
-cloud-init the URI to use, arguments must be passed to the instance via the
-kernel command line or SMBIOS serial number. This argument might look like: ::
+In the above line configuration, ``ds=nocloud`` tells cloud-init to use the
+NoCloud datasource, and ``s=https://10.42.42.42/configs/`` tells cloud-init to
+fetch configurations using ``https`` from the URI
+``https://10.42.42.42/configs/``.
 
-  ds=nocloud;s=https://10.42.42.42/cloud-init/configs/
+We will describe the possible values in a line configuration in the following
```
aciba90, line 82: Could we explicitly link to reference/datasources/nocloud.html#line-configuration-in-detail ? In that way, we are directly pointing to the user where those documentation bits are and it is hopefully easier for them.

aciba90 (changes requested):
Many thanks for overhauling this page. I really like the separation of configuration sources and meta configurations.

I left some extra inline comment / requests.

In addition to that, regarding user-data, meta-data, vendor-data, network-config. We sometimes refer to them with the `-` in between the both words e.g. `user-data`, other times all together e.g., `userdata`, and other times with a space in between e.g. `user data`. I know this happens all over the place and it is not exclusive of this document page but, do you think it is worth trying to adhere to only one of those naming styles? I am thinking about users trying to look for those terms.

TheRealFalcon (commented):
I left some comments inline. Additionally,

We lost the warning at the top of the page.
> User data placed under /etc/cloud/ will not be recognized as a source of configuration data by the NoCloud datasource. While it may be acted upon by cloud-init, using [DataSourceNone](https://cloudinit.readthedocs.io/en/latest/reference/datasources/none.html#datasource-none-example) should be preferred.

Was that intentional?

Also, not something you added, but since you're overhauling the page, can we get a `Network Configuration` section at the bottom? [Example meta-data](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html#example-meta-data) flows right into talking about network config without a new header anywhere. It makes it hard to link to that part of the page.

doc/rtd/reference/datasources/nocloud.rst
holmanb: I've filed https://github.com/canonical/cloud-init/issues/5553 to track this.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,237 @@ NoCloud
...
+cloud-init (typically cloud-init brings up an interface using DHCP).
 
-Method 1: Labeled filesystem
+Since NoCloud is a generic datasource, network configuration may be set the
+same way as user-data, meta data, vendor data.
+
+See the :ref:`network configuration<network_config>` documentation for
+information on network configuration formats.
+
+Meta-configuration
```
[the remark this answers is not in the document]
holmanb: > What do you think about calling this "Discovery configuration"? Meta seems a bit generic to me and might cause confusion when referring to meta data.

Nice suggestion, thanks!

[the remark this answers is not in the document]
holmanb: > Long-term, I don't think it makes sense for this section to live here. Neither form of configuration listed here is specific to NoCloud. It also feels a little redundant when reading the Configuration Sources section after this one, but for now I'm not sure we have a better place for it.

Agreed - and for that matter I think that the [Example: Creating a disk](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html#example-meta-data) section probably deserves its own "How to" page that can be linked from this page.

holmanb:
> Many thanks for overhauling this page. I really like the separation of configuration sources and meta configurations.

Thanks!

> I left some extra inline comment / requests.

I think that I've addressed them all.
 
> In addition to that, regarding user-data, meta-data, vendor-data, network-config. We sometimes refer to them with the `-` in between the both words e.g. `user-data`, other times all together e.g., `userdata`, and other times with a space in between e.g. `user data`. I know this happens all over the place and it is not exclusive of this document page but, do you think it is worth trying to adhere to only one of those naming styles? I am thinking about users trying to look for those terms.

+1 I've updated these terms to follow [the style guide](https://cloudinit.readthedocs.io/en/latest/development/style_docs.html#common-words), but I do have misgivings about the guide as it is. I'll raise those separately.

holmanb:
> I left some comments inline. Additionally,

I think that I've addressed them all

> We lost the warning at the top of the page.
> 
> > User data placed under /etc/cloud/ will not be recognized as a source of configuration data by the NoCloud datasource. While it may be acted upon by cloud-init, using [DataSourceNone](https://cloudinit.readthedocs.io/en/latest/reference/datasources/none.html#datasource-none-example) should be preferred.
> 
> Was that intentional?

Yes, I forgot to mention that.

I don't think that the warning serves much purpose after this change, since the docs now describe how to include user-data directly in the base configuration for NoCloud - so what reason is there to prefer DataSourceNone? If the user reads this page, they will know that placing user data under /etc/cloud isn't how NoCloud expects to receive its configuration.


> Also, not something you added, but since you're overhauling the page, can we get a `Network Configuration` section at the bottom? [Example meta-data](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html#example-meta-data) flows right into talking about network config without a new header anywhere. It makes it hard to link to that part of the page.

Let me know if the change I've made makes sense - I just renamed to match the example meta-data format.

holmanb:
Thanks for the reviews @TheRealFalcon @aciba90! I think I've addressed all comments. Requesting re-review.

TheRealFalcon:
> If the user reads this page, they will know that placing user data under /etc/cloud isn't how NoCloud expects to receive its configuration.

Yeah, this is much better. I'd still prefer putting some kind of warning or note when talking about system configuration to look to the `Source 1` example of doing things rather than trying to put cloud-config directly into files in under /etc/cloud. I'm guessing that people will probably still think "I have something called cloud-config, so it should go in the cloud.cfg", but if you disagree I think it's ok as-is.

> Let me know if the change I've made makes sense - I just renamed to match the example meta-data format.

Sorry, I meant before [line 439](https://github.com/canonical/cloud-init/pull/5521/files#diff-b13894a8a0e9a392b92a800d61c85ff4f8a9128ae0ca7875b5c69dde0f41c4b9R439). The page switches from talking about meta data to talking about network config, but never introduces a new section about it.

doc/rtd/reference/datasources/nocloud.rst
```diff
@@ -4,97 +4,239 @@ NoCloud
...
+Runtime configurations
 ======================
 
-.. warning::
-    User data placed under ``/etc/cloud/`` will **not** be recognized as a
-    source of configuration data by the NoCloud datasource. While it may
-    be acted upon by cloud-init, using
-    :ref:`DataSourceNone<datasource_none_example>` should be preferred.
+Cloud-init discovers four types of configuration at runtime. The source of
+these configuration types is configurable with a meta-configuration. This
```
TheRealFalcon, line 20: Can we s/meta-configuration/discovery configuration/g on this page, or did you want to keep the concept of a meta-configuration?

TheRealFalcon:
This isn't your change, so not necessary to change here, but the aliases emphasis strength makes it hard to understand:
image
I initially read it as `seedfrom alias: s` (without the emphasis) which is kind of hard to parse. Also, what does `cloud-config preferred` mean?

holmanb:
> Also, what does cloud-config preferred mean?

Providing multiple ways of accomplishing the same thing is not ideal. Since we're discouraging use of this, we should provide an alternative. That's what this was trying to accomplish.

holmanb:
> Sorry, I meant before [line 439](https://github.com/canonical/cloud-init/pull/5521/files#diff-b13894a8a0e9a392b92a800d61c85ff4f8a9128ae0ca7875b5c69dde0f41c4b9R439). The page switches from talking about meta data to talking about network config, but never introduces a new section about it.

Gotcha, I've added that. Also I removed the bits about /etc/network/interfaces files because they are not relevant to NoCloud and debian is using netplan in cloud images.

aciba90 (approved):
LGTM, many thanks!

TheRealFalcon:
> Also I removed the bits about /etc/network/interfaces files because they are not relevant to NoCloud and debian is using netplan in cloud images.

It is specific to NoCloud though...
https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/DataSourceNoCloud.py#L224

holmanb:
>> and debian is using netplan in cloud images.
>
> I'm not sure why Debian is relevant here.

Debian is relevant because it is [one of the three distros that were called out as eni users](https://docs.cloud-init.io/en/latest/reference/network-config.html#eni) (Alpine, Debian, and Ubuntu). Since Debian no longer uses eni on cloud images, that leaves only Alpine users that might be using this format on their stock cloud images. Since a shrinking number of distributions are using this as their default distro format, this isn't as relevant as it used to be.

> It is specific to NoCloud though...

What I'm proposing is a step in the direction of deprecating ENI as an input configuration format. This format has been called "legacy" for years and is only supported by NoCloud and ConfigDrive - although [the docs (erroneously) claim NoCloud and OpenStack](https://docs.cloud-init.io/en/latest/reference/network-config-format-eni.html#network-config-eni).

Do we want to support 3 user-facing input formats indefinitely? I would prefer if we only supported 2 of these formats (or 1 really). I think energy would be better spent on cloud-init's networking stack if we focused users (and developer effort) onto fewer formats. Un-documenting this legacy format is a step in that direction, and if we have agreement on this direction I'm happy to remove the other reference to the eni format from the docs as well (in a followup PR).

blackboxsw:
> > > and debian is using netplan in cloud images.
> > 
> > 
> > I'm not sure why Debian is relevant here.
> 
> Debian is relevant because it is [one of the three distros that were called out as eni users](https://docs.cloud-init.io/en/latest/reference/network-config.html#eni) (Alpine, Debian, and Ubuntu). Since Debian no longer uses eni on cloud images, that leaves only Alpine users that might be using this format on their stock cloud images. Since a shrinking number of distributions are using this as their default distro format, this isn't as relevant as it used to be.
> 
> > It is specific to NoCloud though...
> 
> What I'm proposing is a step in the direction of deprecating ENI as an input configuration format. This format has been called "legacy" for years and is only supported by NoCloud and ConfigDrive - although [the docs (erroneously) claim NoCloud and OpenStack](https://docs.cloud-init.io/en/latest/reference/network-config-format-eni.html#network-config-eni).
> 
> Do we want to support 3 user-facing input formats indefinitely? I would prefer if we only supported 2 of these formats (or 1 really). I think energy would be better spent on cloud-init's networking stack if we focused users (and developer effort) onto fewer formats. Un-documenting this legacy format is a step in that direction, and if we have agreement on this direction I'm happy to remove the other reference to the eni format from the docs as well (in a followup PR).

I think we agree with the sentiment that we want to support less formats and spend energy maintaining and growing network v1 and v2.  We also discussed that the best path forward would be to allow our code to represent a given feature as deprecated when it encounters that format, allow the deprecation message to suggest the desirable alternative format and provide a reasonable deprecation window before removing the supporting code. Omitting the documentation of deprecated features when we have active `deprecation()` messaging in our code makes sense.

PR #5527, opened 2024-07-19
Ubuntu/devel

aciba90:
new upstream snapshot release

- [x] build-package
- [x] sbuild


aciba90:
I get this lintian error:

```
E: cloud-init: alien-tag package-supports-alternative-init-but-no-init.d-script [usr/share/lintian/overrides/cloud-init:7]
```

We can probably drop: https://github.com/canonical/cloud-init/blob/1852ff08069fdaf4d729bd5365e31c4257e7723e/debian/cloud-init.lintian-overrides#L7

It sounds like that lintian check got removed.

debian/changelog
```diff
@@ -1,3 +1,10 @@
+cloud-init (24.3~6g6e4153b3-0ubuntu1) oracular; urgency=medium
```
blackboxsw, line 1: In devel series we run with the development version format  `<UNRELEASED_MAJOR>.<UNRELEASED_MINOR>~<NUM_OF_DEVEL_PUBLISHES>g<UPSTREAM_GIT_HASH>` to ensure that devel series package version always remains greater than any 24.2-XX release that we push to stable series as part of an SRU. This greater sort order allows the package to remain a viable "upgrade" for images performing dist-upgrade from a stable series to oracular. The tilde `~` makes this package sort lower than the official `24.3-<something>` that we will eventually release to oracular when upstream cuts the official 24.3 release and we package it as `24.3-0ubuntu1`

Since this is the first upload post 24.2 upstream release we have content greater than 24.2, so we need to manually bump this version to 24.3~ based... Since it's the first upload as 24.3~X the X should be 1 not 6.
```suggestion
cloud-init (24.3~1g6e4153b3-0ubuntu1) oracular; urgency=medium
```

blackboxsw (approved):
- Jenkins integration runners healthy on major platforms
- build-package and sbuild success, no new lintian errors/warnings
- same snapshot locally vs proposed branch.

PR #5536, opened 2024-07-23
feat: Eliminate redundant configuration reads

holmanb:
## Proposed Commit Message

```
feat: Eliminate redundant configuration reads

When instance id hasn't changed and datasource hasn't changed, don't
forcibly reload configuration.
```

## Additional Context
On a test system, this avoids reloading the configuration on the second `instancify()`

See the new messages in logs: 
[no-reload.log](https://github.com/user-attachments/files/16351318/no-reload.txt)

### DIFF
--- a/tests/integration_tests/integration_settings.py
+++ b/tests/integration_tests/integration_settings.py
@@ -37,7 +37,7 @@ INSTANCE_TYPE: Optional[str] = None
 # <image_id>[::<os>::<release>::<version>].  If given, os and release should
 # describe the image specified by image_id.  (Ubuntu releases are converted
 # to this format internally; in this case, to "None::ubuntu::focal::20.04".)
-OS_IMAGE = "jammy"
+OS_IMAGE = "focal"
 
 # Populate if you want to use a pre-launched instance instead of
 # creating a new one. The exact contents will be platform dependent
@@ -69,7 +69,7 @@ EXISTING_INSTANCE_ID: Optional[str] = None
 #   Install from a PPA. It MUST start with 'ppa:'
 # <file path>
 #   A path to a valid package to be uploaded and installed
-CLOUD_INIT_SOURCE = "IN_PLACE"
+CLOUD_INIT_SOURCE = "NONE"
 
 # Before an instance is torn down, we run `cloud-init collect-logs`
 # and transfer them locally. These settings specify when to collect these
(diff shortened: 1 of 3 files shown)


TheRealFalcon:
What happens on a new instance id? The old instance directory and link won't go away, so won't we miss removing and re-linking the new instance directory?

holmanb:
> What happens on a new instance id? The old instance directory and link won't go away, so won't we miss removing and re-linking the new instance directory?

The remove and link code will still run if the directory that the symlink already resolves to is different from the one that we would otherwise create.

TheRealFalcon (commented):
The changes look good, but given it's age and how core it is to how cloud-init functions, I really don't think I'm comfortable changing anything here unless we have tests around ways instance-ids may change. Specifically:
* booting with the first instance-id
* booting with a new instance-id from previous boot
* booting with the same instance when there's no way to verify the instance-id on disk
* booting with a new instance id when there's no way to verify the instance-id on disk

holmanb:
> The changes look good, but given it's age and how core it is to how cloud-init functions, I really don't think I'm comfortable changing anything here unless we have tests around ways instance-ids may change. Specifically:
> 
>     * booting with the first instance-id

see the latest test addition

>     * booting with a new instance-id from previous boot

see the latest integration test addition

>     * booting with the same instance when there's no way to verify the instance-id on disk
>     * booting with a new instance id when there's no way to verify the instance-id on disk

there is a test already in tests/integration_tests/datasources/test_caching.py that covers this case

holmanb:
Rebased to pick up the labeler fix.

tests/integration_tests/integration_settings.py
```diff
@@ -37,7 +37,7 @@
 # <image_id>[::<os>::<release>::<version>].  If given, os and release should
 # describe the image specified by image_id.  (Ubuntu releases are converted
 # to this format internally; in this case, to "None::ubuntu::focal::20.04".)
-OS_IMAGE = "focal"
+OS_IMAGE = "jammy"
```
TheRealFalcon, line 40: Were these changes intentional?

TheRealFalcon (approved):
LGTM, thanks for the tests!

You made some changes to integration_settings.py that I'm assuming weren't intentional, but once those are removed, +1.

tests/integration_tests/integration_settings.py
holmanb: nope, thanks!

holmanb:
> LGTM, thanks for the tests!
> 
> You made some changes to integration_settings.py that I'm assuming weren't intentional, but once those are removed, +1.

PR #5539, opened 2024-07-24
doc: add diagram with boot stages

aciba90:
## Proposed Commit Message

```
doc: add diagram with boot stages

SC-1835
Fixes GH-5334
```

## Test Steps

```
tox -e doc && (cd doc/rtd_html/ && python3 -m http.server)
open http://0.0.0.0:8000/explanation/boot.html
```

## Merge type

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -9,4 +9,5 @@ sphinx-design
 sphinx-copybutton
 sphinx-notfound-page
 sphinxcontrib.datatemplates
+sphinxcontrib-mermaid
 sphinxcontrib-spelling
--- a/doc/rtd/conf.py
+++ b/doc/rtd/conf.py
@@ -44,6 +44,7 @@ extensions = [
     "sphinx.ext.autosectionlabel",
     "sphinx.ext.viewcode",
     "sphinxcontrib.datatemplates",
+    "sphinxcontrib.mermaid",
     "sphinxcontrib.spelling",
 ]
 
--- a/doc/rtd/explanation/boot.rst
+++ b/doc/rtd/explanation/boot.rst
@@ -11,6 +11,43 @@ There are five stages to boot:
 4. Config
 5. Final
 
+Visual representation:
+
+.. mermaid::
+
+  graph LR
+
+    D[Detect] ---> L
+
+    L --> NU([Network up])
+    L & NU --> N
+    subgraph L[Local]
+        direction TB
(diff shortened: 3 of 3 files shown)


aciba90:
Diff:

image

s-makin (commented):
Looks really good!

I only have two real suggestions for changes. 
1) Can we make the diagram vertical rather than horizontal? Where it gets sized to fit on the screen width it makes the text quite small. I think a vertical approach would avoid that problem and also make it easier for people to view on smaller screens.
2) Can we make the "Detect" box yellow as well? I think since we have the list of stages at the top, and the other four are in the larger yellow boxes, it would make it nicely correspond to the list if all the main stages looked the same as each other in the diagram

a-dubs:
> Can we make the "Detect" box yellow as well? I think since we have the list of stages at the top, and the other four are in the larger yellow boxes, it would make it nicely correspond to the list if all the main stages looked the same as each other in the diagram

@aciba90 @s-makin Is "detect" an actual stage of cloud-init? If not, shouldn't it be left as is so that only official cloud-init stages are given yellow boxes?

holmanb:
> > Can we make the "Detect" box yellow as well? I think since we have the list of stages at the top, and the other four are in the larger yellow boxes, it would make it nicely correspond to the list if all the main stages looked the same as each other in the diagram
> 
> @aciba90 @s-makin Is "detect" an actual stage of cloud-init? If not, shouldn't it be left as is so that only official cloud-init stages are given yellow boxes?

A small legend which describes the differences would give the boxes and color more meaning to users.

It might make sense to break out [the boot page](https://docs.cloud-init.io/en/latest/explanation/boot.html) into a "logical" main page with "implementation" sub-pages so that we can talk about the systemd implementation (and other implementations) in more detail without making the logical boot stages page too busy.

a-dubs:
+ 1 to @holmanb's idea about breaking out into more details sub pages. But just to be clear, that work should be captured in an issue and tackled separately, right?

s-makin (approved):
LGTM :)

blackboxsw:
Taking [this comment](https://github.com/canonical/cloud-init/pull/5539#issuecomment-2248642808) as a separate issue to follow up on separately

blackboxsw (approved):
diagram looks good. I changed diagram alignment to be top to bottom TB and provided supplemental links to various boot stage sections in this page so others can dig into details easily by clicking elements in the diagram.

Issue #5467, opened 2024-06-29
[docs]: revise headings levels


noREAVER:
# Documentation request

Shouldn't "Deprecation schedule and versions" be a Note instead and not a top heading?
Or at least the same heading level with the rest of the module names?

---
*Reported from: https://docs.cloud-init.io/en/latest/reference/modules.html

PR #5550, opened 2024-07-25
doc(modules): add section to wrap modules' doc

aciba90:
Fixes GH-5467

## Proposed Commit Message

```
doc(modules): add section to wrap modules' doc

Fixes GH-5467
```

## Additional Context

image

## Merge type


s-makin (approved):
I usually don't like headings and subheadings with no content between, but I think in this case it makes sense as a fence. Thanks!

blackboxsw (approved):
LGTM

Issue #5549, opened 2024-07-25
#cloud-config-jsonp is almost completely useless


TheRealFalcon:
`#cloud-config-jsonp` is a way to specify [JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902) operations to
apply to cloud-init's existing configuration. The problem is, in its current form, it is almost completely useless.

I think that the idea is to have a convenient way (even though we already have merge configuration?) to merge user data with previous config. The problem is due to the ordering of how we process cloud-config, it will often attempt to be applied before it ever has a chance to patch anything.

[Our docs](https://cloudinit.readthedocs.io/en/latest/explanation/vendordata.html#overview) call out using it as a way for user data to override vendor data, but this does not work, because the jsonp part gets processed before vendor data.

The only way this can work as expected is if you have a MIME multipart user data in which the jsonp attachment is ordered after an earlier attachment that you want to patch. This seems like an extreme corner case because:
1. Merge strategy is available
2. If you control your user data parts, why would structure them such that overlap in a way that needs to be merged this way?
3. If you don't control your user data parts, how can you ensure that they get ordered correctly?

Given that feature also pulls in the `jsonpatch` dependency, I think we should remove this feature soon in a upcoming release.

holmanb:
> [Our docs](https://cloudinit.readthedocs.io/en/23.4/explanation/vendordata.html) call out using it as a way for user data to override vendor data, but this does not work, because the jsonp part gets processed before vendor data.

Thanks for dropping this incorrect documentation in #5551.

> Given that feature also pulls in the jsonpatch dependency, I think we should remove this feature soon in a upcoming release.

It sounds like it is error prone/broken and doesn't add value over other types. Agreed - lets ditch it.

PR #5546, opened 2024-07-25
doc: Update docs on boothooks

aciba90:
## Proposed Commit Message

```
doc: Update docs on boothooks

Improve explaination on #cloud-boothook for end-users.

Reorder and improve doc/rtd/explanation/format.rst.

SC-1657
Fixes GH-4542

Co-authored-by: Calvin Mwadime <calvin.mwadime@canonical.com>
```

## Additional Context

https://github.com/canonical/cloud-init/pull/5541#pullrequestreview-2196889423

## Merge type

- [x] Squash merge using "Proposed Commit Message"
- [ ] Rebase and merge unique commits. Requires commit messages per-commit each referencing the pull request number (#<PR_NUM>)

Issue #4649, opened 2023-11-30
[docs]: Document drop-in 3rd party modules


holmanb:
# Documentation request
Cloud-init provides code for drop-in 3rd party modules. These can be [custom mergers](https://cloudinit.readthedocs.io/en/latest/reference/merging.html#customisation), [custom handlers](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#part-handler), custom datasources, custom configuration modules, or [custom cleaners](https://github.com/canonical/cloud-init/pull/4761/files).

These are categorically different than typical user configuration. Users are expected to run commands and write configurations. This customization to cloud-init expect the user to write code, which lands them in a grey zone between user and developer - call them "super user" for now.

Currently only custom mergers and custom handlers are documented, and the documentation for those is intermingled with "normal user" documentation. This "super user" documentation is likely confusing alongside normal configuration documentation, and should perhaps be relocated somewhere different for clarity (and linked from "normal user" documentation).

I imagine a document structure like this:
```
doc/rtd/reference/custom_modules/index.rst
doc/rtd/reference/custom_modules/custom_datasource.rst
doc/rtd/reference/custom_modules/custom_configuration_module.rst
doc/rtd/reference/custom_modules/custom_mergers.rst
doc/rtd/reference/custom_modules/custom_formats.rst
doc/rtd/reference/custom_modules/custom_cleaners.rst
```

The index.rst page should discuss the merits, drawbacks, and strategies for using these interfaces. Each sub-page should describe the interface and provide a brief example.

blackboxsw:
big +1 on separating those docs for "power users" with custom downstream image needs and behaviors. These scopres are atypical for most cloud-init users.

aciba90:
Distros can also be dropped-in, but there is no documentation on how to create a new distro. Thus, I would propose to create a separate issue/s for documenting how to create a new distro and how to drop in a new distro.

PR #5551, opened 2024-07-25
docs: Overhaul user data formats documentation

TheRealFalcon:
## Proposed Commit Message

```
docs: Overhaul user data formats documentation

Fixes GH-4739
```

## Additional Context
I kept (and updated) the part handler piece as I think it should stay here until https://github.com/canonical/cloud-init/issues/4649 is implemented. 

I started documenting cloud-config-jsonp, but decided against it and filed https://github.com/canonical/cloud-init/issues/5549 instead. I don't think we should be using it or recommending it to anybody. I also updated the vendor data page accordingly to remove the suggestion there.

There's unfortunately some significant overlap with https://github.com/canonical/cloud-init/pull/5546 . I stole some of the language there because a full-on cherry-pick would have been too hard to merge. If the commit there can be simplified, I'm happy to cherry-pick it and rebase mine on top to give Alberto some credit for his additions.

I'm not married to the format I chose, so if a reviewer has any better ideas, feel free to suggest.

## Merge type


doc/rtd/explanation/format.rst
```diff
@@ -3,18 +3,65 @@
 User data formats
 *****************
 
-User data is opaque configuration data provided by a platform to an instance at
-launch configure the instance. User data can be one of the following types.
+User data is configuration data provided by a user of a cloud platform to an
+instance at launch. User data can be passed to cloud-init in any of many
+formats documented here. Each format will document a required header or
+content-type that must be present in the user data to be recognized as that
+format.
```
s-makin, line 10: "Each format will document a required header or content-type that must be present in the user data to be recognized as that format."
I found this a bit confusing - particularly "each format will document". Is the format doing the documenting? Or is it that each format requires a specific header or content type?

doc/rtd/explanation/format.rst
```diff
@@ -3,18 +3,65 @@
...
 
-Cloud-config is the preferred user data format. The cloud config format is a
-declarative syntax which uses `YAML version 1.1`_ with keys which describe
-desired instance state. Cloud-config can be used to define how an instance
-should be configured in a human-friendly format.
+| **Header:** #cloud-config
+| **Content-Type:** text/cloud-config
+
+Example
+-------
```
s-makin, line 50: I don't know that we need this or the explanation header to be subheaders. For consistency with the header and content type, would it make sense to set these to **Example** and **Explanation**? The sections don't feel long enough to justify having subsections in them.

s-makin (commented):
Overall I really love this change. I think it feels more standardised and definitely a lot more straightforward, and the consistent format for each section in the format page makes it easier to parse the information. Just have one small suggestion, but otherwise lgtm :)

doc/rtd/explanation/format.rst
TheRealFalcon: I think I prefer subheadings rather than bold, but I pushed a bold commit for comparison.

doc/rtd/explanation/format.rst
TheRealFalcon: I updated that sentence. Let me know if that's any better.

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
+        echo "this is from a boothook." > /var/tmp/boothook.txt
+    - type: "text/cloud-config"
+    content: |
+        bootcmd:
+        - echo "this is from a cloud-config." > /var/tmp/bootcmd.txt
+
+Explanation
+-----------
+
+A cloud-config-archive is a way to specify more than one type of data
```
holmanb, line 242: Are there any benefits over a MIME multi-part archive? Why would anyone want to use this over MIME?

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
-================
 
-This content is an :file:`include` file.
+Cloud config archive
+====================
+
+| **Header:** #cloud-config-archive
+| **Content-Type:** text/cloud-config-archive
+
+Example
```
holmanb: Nice example!

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
+-------
+
+.. code-block:: shell
+
+    #cloud-config-archive
+    - type: "text/cloud-boothook"
+    content: |
```
holmanb, line 232: should this be indented further?

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
+        #!/bin/sh
+        echo "this is from a boothook." > /var/tmp/boothook.txt
+    - type: "text/cloud-config"
+    content: |
```
holmanb, line 236: should this be indented further?

doc/examples/part-handler.txt
```diff
@@ -1,22 +1,58 @@
...
-    #           no filename is given attribute is present
-    # payload: the content of the part (empty for begin or end)
+    """Return a list of mime-types that are handled by this module."""
+    return ["text/x-my-path"]
+
+
+def handle_part(data: Cloud, ctype: str, filename: str, payload: Any):
+    """Handle a part with the given mime-type.
+
+    This function will get called mulltiple times. The first time is
```
holmanb, line 33: sp: multiple

doc/examples/part-handler.txt
```diff
@@ -1,22 +1,58 @@
 #part-handler
 
+"""This is a trivial example part-handler that creates a file with the path
```
holmanb, line 3: Much needed context throughout this file, thanks for adding it!

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
 
-The file contains a list of URLs, one per line. Each of the URLs will be read
-and their content will be passed through this same set of rules, i.e., the
-content read from the URL can be gzipped, MIME multi-part, or plain text. If
-an error occurs reading a file the remaining files will not be read.
+Required fields:
 
-Begins with: ``#include`` or ``Content-Type: text/x-include-url``  when using
-a MIME archive.
+* ``type``: The content type of the MIME part
```
holmanb, line 250: This format doesn't use a MIME part, does it? Maybe something like this instead?

> type: The Content-Type identifier for the type of user data in content
> content: The user data configuration

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
-``Cloud-init`` will then call the ``handle_part`` function once before it
-handles any parts, once per part received, and once after all parts have been
-handled. The ``'__begin__'`` and ``'__end__'`` sentinels allow the part
-handler to do initialisation or teardown before or after receiving any parts.
+An include file contains a list of URLs, one per line. Each of the URLs will
+be read and their content can be any kind of user data format, both base
+config and meta config. If an error occurs reading a file the remaining files
+will not be read.
+
+Gzip compressed content
```
holmanb, line 318: Is this format supported on all platforms? I seem to recall there being some platform-specific oddities about this but don't remember specifics.

holmanb (changes requested):
I really like this work! The organization and standardization is much improved. It is much easier to read and adds lots of context.

I left some suggestions and questions inline.

doc/rtd/explanation/format.rst
```diff
@@ -141,85 +216,150 @@ Create user data containing 3 shell scripts:
 
     $ cloud-init devel make-mime -a always.sh:x-shellscript-per-boot -a instance.sh:x-shellscript-per-instance -a once.sh:x-shellscript-per-once
 
-``include`` file
-================
 
-This content is an :file:`include` file.
+Cloud config archive
```
holmanb, line 220: Aside: is cloud config archive nesting supported? I've not checked the code, but I'm curious whether I could stuff have:

- cloud config archive -> cloud config archive -> cloud config archive
- cloud config archive -> mime -> cloud config archive
- mime -> cloud config archive -> mime

and where/when things might start to break with nesting

doc/rtd/explanation/format.rst
```diff
@@ -3,18 +3,65 @@
...
+User data is configuration data provided by a user of a cloud platform to an
+instance at launch. User data can be passed to cloud-init in any of many
+formats documented here. Each format will document a required header or
+content-type that must be present in the user data to be recognized as that
+format.
+
+Configuration types
+===================
+
+User data can be categorized into **base config**
```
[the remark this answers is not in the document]
TheRealFalcon: What do you think about not trying to define them at all. This section could be updated to say something like

User data formats can be split into those that  directly configure the instance and those that serve as a container, template, or means to obtain or modify another configuration.

Formats that directly configure the instance:
...

Formats that deal with other user data formats:
...

doc/rtd/explanation/format.rst
```diff
@@ -24,64 +71,101 @@ These things may include:
...
+* It is run on every boot
+* The environment variable ``INSTANCE_ID`` is set to the current instance ID
+  for use within the script.
 
 MIME multi-part archive
 =======================
 
-This list of rules is applied to each part of this multi-part file.
+| **Header:** Content-Type: multipart/mixed;
+| **Content-Type:** multipart/mixed
```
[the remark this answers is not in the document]
TheRealFalcon: Heh. The header is the content-type declaration...so...it's a bit weird.

doc/rtd/explanation/format.rst
TheRealFalcon: I'll leave that as an exercise for you to figure out :joy:

doc/rtd/explanation/format.rst
```diff
@@ -3,18 +3,65 @@
...
 
 Cloud config data
 =================
 
-Cloud-config is the preferred user data format. The cloud config format is a
-declarative syntax which uses `YAML version 1.1`_ with keys which describe
-desired instance state. Cloud-config can be used to define how an instance
-should be configured in a human-friendly format.
+| **Header:** #cloud-config
+| **Content-Type:** text/cloud-config
```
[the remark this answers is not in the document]
TheRealFalcon: Yeah...I was having a hard time figuring out how to fit these in. I think these are really both pieces of reference information, but I didn't think it made sense to have reference pages that only list header and content-types for user data formats.

The thing is these content types also matter for the cloud config archive and part handlers.

I get that these formats (along with the MIME archive) are going to much less used and basically "expert mode" for anybody using them...so...maybe I can try to cook up a reference page with enough info.

doc/rtd/explanation/format.rst
TheRealFalcon: Why would anyone NOT want to use this over MIME? Have you tried building one yourself? Do you like calling helper commands to build your user data? :wink: 

I can add a "for those that prefer to use YAML" to the end of this section. Are you wanting more of an explanation than that in the docs?

doc/rtd/explanation/format.rst
TheRealFalcon: It translates fairly directly to a mime part that gets generated from the config. I can pull the MIME details out of this section, but IMO it'll make some of the other fields harder to understand.

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
+cloud-config and user data scripts. Any
+:ref:`instance-data variables<instance_metadata-keys>` may be used
+as jinja template variables. Any jinja templated configuration must contain
+the original header along with the new jinja header above it.
 
-Begins with: ``#cloud-boothook`` or ``Content-Type: text/cloud-boothook`` when
-using a MIME archive.
+.. note::
+    Use of Jinja templates is ONLY supported for cloud-config and user data
+    scripts. Jinja templates are not supported for cloud-boothooks or
```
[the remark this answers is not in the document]
TheRealFalcon: Yep, good idea. https://github.com/canonical/cloud-init/issues/5554

doc/rtd/explanation/format.rst
TheRealFalcon: As far as I know it should. If your cloud mangles your user data (like LXD does), it won't work, but that's the only limitation I'm aware of.

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
-handler to do initialisation or teardown before or after receiving any parts.
+An include file contains a list of URLs, one per line. Each of the URLs will
+be read and their content can be any kind of user data format, both base
+config and meta config. If an error occurs reading a file the remaining files
+will not be read.
+
+Gzip compressed content
+=======================
+
+| **Header** n/a
```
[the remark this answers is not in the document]
TheRealFalcon: Yeah, this was never meant to call out generic file headers. It's for cloud-init required headers which is why it's n/a for gzip.

doc/rtd/explanation/format.rst
```diff
@@ -141,66 +225,128 @@ Create user data containing 3 shell scripts:
...
+| **Content-Type** n/a
```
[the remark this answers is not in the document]
TheRealFalcon: Yes, but given that it's a binary file, it can't be specified anywhere.

doc/rtd/explanation/format.rst
s-makin: It's a lot clearer to me :) thanks!

doc/rtd/explanation/format.rst
s-makin: It's nothing I'd be willing to hold up the doc over. If you're happier with the subheadings, we'll leave them as they are.

doc/rtd/explanation/format.rst
holmanb: Yeah, I get it :)

>  It can be seen as an alternative to building a MIME multi-part archive manually.

We present two alternatives, without any recommended / preferred choice or cost / benefit analysis to the end user. If they've tried MIME before, then they'll know, but if not then they have to learn the hard way. Maybe we could save them the trouble?

doc/rtd/explanation/format.rst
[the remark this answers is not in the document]
holmanb: I can see it being useful to have a name for these than "formats that directly configure the instance", but I'm fine with what you've suggested if that's what you would prefer.

s-makin (approved):
+1 from me :)

doc/rtd/explanation/format.rst
```diff
@@ -24,64 +69,113 @@ These things may include:
...
 
-Content found to be gzip compressed will be uncompressed.
-The uncompressed data will then be used as if it were not compressed.
-This is typically useful because user data is limited to ~16384 [#]_ bytes.
+A cloud boothook is similar to a :ref:`user data script<user_data_script>`
+in that it is a shell script run on boot. The boothook is different in that:
+
+* It is run very early in boot, even before the ``cc_bootcmd`` module
+* It is run on every boot
+* The environment variable ``INSTANCE_ID`` is set to the current instance ID
```
aciba90, line 141: `INSTANCE_ID` is also provided for [cc_bootcmd](https://github.com/canonical/cloud-init/blob/779dd6b009f8a568fe74e98df9803ad590b83044/doc/module-docs/cc_bootcmd/data.yaml#L5), could you please remove this item?

aciba90 (commented):
Many thanks for this effort. It looks great!

TheRealFalcon:
Not all comments have been addressed, but I wanted to push what I had so far.

doc/rtd/explanation/format.rst
```diff
@@ -24,64 +71,101 @@ These things may include:
...
+    #!/bin/sh
+    echo "Hello World" > /var/tmp/output.txt
 
-Create a script file :file:`myscript.sh` that contains the following:
+Explanation
+-----------
 
-.. code-block::
+A user data script is a single shell script to be executed once per instance.
+User data scripts are run relatively late in the boot process, after most
```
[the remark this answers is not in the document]
TheRealFalcon: I do think that our stages go long enough that it makes sense to place these relative to a specific module. I don't think it's confusing info if the user doesn't need it. Let me know if you think it makes sense.

doc/rtd/explanation/format.rst
```diff
@@ -24,64 +71,101 @@ These things may include:
...
+Explanation
+-----------
 
-Content found to be gzip compressed will be uncompressed.
-The uncompressed data will then be used as if it were not compressed.
-This is typically useful because user data is limited to ~16384 [#]_ bytes.
+A cloud boothook is similar to a :ref:`user data script<user_data_script>`
+in that it is a shell script run on boot. The boothook is different in that:
+
+* It is run very early in boot, even before the ``cc_bootcmd`` module
```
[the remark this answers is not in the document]
TheRealFalcon: I do think that our stages go long enough that it makes sense to place these relative to a specific module. I don't think it's confusing info if the user doesn't need it. Let me know if you think it makes sense.

doc/rtd/explanation/format.rst
```diff
@@ -24,64 +71,121 @@ These things may include:
...
-User data scripts can optionally render cloud instance metadata variables using
-:ref:`jinja templates <instancedata-Using>`.
+Explanation
+-----------
 
-Example script
---------------
+A user data script is a single shell script to be executed once per instance.
+User data scripts are run relatively late in the boot process, during the
+'modules:final' stage as part of the "cc_scripts_user" module.
```
holmanb, line 102: I know that `modules:final" text that is commonly seen in the log, but I think it would be easier to standardize on the more generic "final" stage, since the "modules" entrypoint is really just an implementation detail that the user shouldn't need to know about.

Also, could we link to the respective stages on the boot order page both here and in the cloud boothook section?

TheRealFalcon:
I believe the only fix left is to move the part handler stuff. Since that requires a rebase and merge, I figured it'd be best to push first.

TheRealFalcon:
Most recent commit adds the part handler split. Tbh, I don't love how it's split, but I do think it makes sense to keep information in both places. What do people think of having the example in both locations?

I believe with this commit I have addressed all of the comments.

doc/rtd/spelling_word_list.txt
```diff
@@ -211,6 +212,7 @@ scaleway
 seedurl
 serverurl
 setup-keymap
+shellscript
```
TheRealFalcon, line 215: I don't love this as in most places this would be a legitimate misspelling, but I couldn't find any way to ignore a spelling for a single line.

holmanb: Would not be legitimate? 

I think Chad submitted a PR in the last day or two that ignored spelling for a single line.

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
-Begins with: ``#!`` or ``Content-Type: text/x-shellscript`` when using a MIME
-archive.
+Explanation
+-----------
 
-User data scripts can optionally render cloud instance metadata variables using
-:ref:`jinja templates <instancedata-Using>`.
+A user data script is a single shell script to be executed once per instance.
+User data scripts are run relatively late in the boot process, during the
+cloud-init's final stage as part of the "cc_scripts_user" module. When run,
```
holmanb, line 87: I would suggest a link to the cc_scripts_user module, however it's really more of an implementation detail and that module's documentation isn't very informative.

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
 
-Begins with: ``#!`` or ``Content-Type: text/x-shellscript`` when using a MIME
-archive.
+Explanation
+-----------
 
-User data scripts can optionally render cloud instance metadata variables using
-:ref:`jinja templates <instancedata-Using>`.
+A user data script is a single shell script to be executed once per instance.
+User data scripts are run relatively late in the boot process, during the
```
holmanb, line 86: "during the cloud-init's final stage" sounds funny

doc/rtd/explanation/format.rst
```diff
@@ -3,18 +3,53 @@
...
-should be configured in a human-friendly format.
+Example
+-------
+
+.. code-block:: yaml
+
+    #cloud-config
+    password: password
+    chpasswd:
+    expire: False
```
holmanb, line 45: indent

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
+module to determine the run frequency.
 
-Begins with: ``#cloud-config`` or ``Content-Type: text/cloud-config`` when
-using a MIME archive.
-
-.. note::
-   Cloud config data can also render cloud instance metadata variables using
-   :ref:`jinja templates <instancedata-Using>`.
+See the :ref:`yaml_examples` section for a set of commented examples of
+supported cloud config formats.
```
holmanb, line 67: > set of commented examples of supported cloud config formats.

I know that you didn't write this in this PR, but I don't think that it makes sense to call these "cloud config formats". 

Also, this seems like an appropriate place to link to the modules reference page as well. Could we do that while making this change?

How about something like this:
```rst
For more information, see the cloud config 
:ref:`example configurations page <link_to_examples_page>`
 or the cloud config :ref:`modules reference page<link_to_modules_page>`.
```

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
+    echo "Hello World" > /var/tmp/output.txt
 
-Begins with: ``#!`` or ``Content-Type: text/x-shellscript`` when using a MIME
-archive.
+Explanation
+-----------
 
-User data scripts can optionally render cloud instance metadata variables using
-:ref:`jinja templates <instancedata-Using>`.
+A user data script is a single shell script to be executed once per instance.
```
holmanb, line 85: It doesn't have to be a shell script does it? Wouldn't anything that can run with a shebang work (lua, perl, make, tcl, guile, etc)? Maybe just drop the word "shell"?

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
+   sudo echo $INSTANCE_ID > $PERSIST_ID
 
-   $ euca-run-instances --key mykey --user-data-file myscript.sh ami-a07d95c9
+Explanation
+-----------
 
-Kernel command line
-===================
+A cloud boothook is similar to a :ref:`user data script<user_data_script>`
+in that it is a shell script run on boot. When run,
```
holmanb, line 128: same comment as above about "shell"

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
+* It is run very early in boot, during the 'init' stage, before any
+  cloud-init modules are run.
+* It is run on every boot
 
-Content found to be gzip compressed will be uncompressed.
-The uncompressed data will then be used as if it were not compressed.
-This is typically useful because user data is limited to ~16384 [#]_ bytes.
+.. _user_data_formats-mime_archive:
 
 MIME multi-part archive
```
holmanb, line 140: How would you feel about ordering the "formats that deal with other user data formats" in order of increasing complexity in the list above and also in the sections starting here?

If that seems reasonable, I would suggest:

Include file
Jinja template
cloud config archive
Mime multi-part archive
Part handler

doc/rtd/explanation/format.rst
holmanb, line 87: Could we link to the final stage on the boot stages page from here? That would be useful for anybody who isn't familiar with cloud-init's ordering.

doc/rtd/explanation/format.rst
```diff
@@ -24,93 +59,132 @@ These things may include:
...
 
-When using the NoCloud datasource, users can pass user data via the kernel
-command line parameters. See the :ref:`NoCloud datasource<datasource_nocloud>`
-and :ref:`explanation/kernel-command-line:Kernel command line` documentation
-for more details.
+The boothook is different in that:
 
-Gzip compressed content
-=======================
+* It is run very early in boot, during the 'init' stage, before any
```
holmanb, line 134: Could we link to the init stage on the boot stages page from here? That would be useful for anybody who isn't familiar with cloud-init's ordering.

holmanb (commented):
@TheRealFalcon I think this is really close. I left a couple more comments.

doc/rtd/spelling_word_list.txt
TheRealFalcon: > Not sure that I understand, did you mean to say "... would not be a legitimate mispelling"?

No. I think that most of the time we want "shell script" and so I think adding this has the potential to silence legitimate misspellings.

doc/rtd/explanation/format.rst
TheRealFalcon: > Wouldn't anything that can run with a shebang work 

Isn't that by definition a shell script then? Shell doesn't just mean `sh` or one of its descendants. Most of your examples there are listed as scripting languages [here](https://en.wikipedia.org/wiki/Scripting_language#Notable_languages)

TheRealFalcon: (ignore my other comment here if you saw it)

doc/rtd/explanation/format.rst
TheRealFalcon: Yes, but I want to put the mime one first as cloud config archive is framed as an alternative to the mime type.

holmanb (approved):

PR #5556, opened 2024-07-26
doc: improve integrations testing configuration instructions

a-dubs:
## Proposed Commit Message
```
docs: improve integrations testing configuration instructions
```

## Merge type

### DIFF
--- a/doc/rtd/development/integration_tests.rst
+++ b/doc/rtd/development/integration_tests.rst
@@ -62,7 +62,7 @@ All possible configuration values are defined in
 `tests/integration_tests/integration_settings.py`_. Look in this file for
 the full list of variables that are available and for context on what each
 variable does and what the default values are.
-by supplying values in :file:`tests/integration_tests/user_settings.py` or by
+Defaults can be overriden by supplying values in :file:`tests/integration_tests/user_settings.py` or by
 providing an environment variable of the same name prepended with
 ``CLOUD_INIT_``. For example, to set the ``PLATFORM`` setting:
 
@@ -132,8 +132,6 @@ By default, the logs are collected to the ``/tmp/cloud_init_test_logs``
 directory. To change the directory, set the ``LOCAL_LOG_PATH`` variable to
 the desired path.
 
-TODO: can the collect log path be set to a relative path?
-
 .. tab-set::
 
     .. tab-item:: Inline environment variable
@@ -177,18 +175,6 @@ Addtionally, for profiling the integration tests, set the ``INCLUDE_PROFILE``
 variable to ``True``. This will generate a profile report for the integration
 test run, and the report will be stored in the directory specified by
 ``LOCAL_LOG_PATH``.
-# We default our coverage to False because it involves modifying the
-# cloud-init systemd services, which is too intrusive of a change to
-# enable by default. If changed to true, the test directory corresponding
-# to the test run under LOCAL_LOG_PATH defined above will contain an
-# `html` directory with the coverage report.
-INCLUDE_COVERAGE = False
-
-# We default our profile to False because it involves modifying the
-# cloud-init systemd services, which is too intrusive of a change to
-# enable by default. If changed to true, the test directory corresponding
-# to the test run under LOCAL_LOG_PATH defined above will contain a report
-INCLUDE_PROFILE = False
 
 .. tab-set::
 
@@ -273,10 +259,10 @@ tests against the image in question. If it's a RHEL8 image, then we
 would expect Ubuntu-specific tests to fail (and vice versa).
 
 To address this, a full image specification can be given. This is of
-the form: ``<image_id>[::<os>[::<release>]]`` where ``image_id`` is a
+the form: ``<image_id>[::<os>::<release>::<version>]`` where ``image_id`` is a
 cloud's image ID, ``os`` is the OS name, and ``release`` is the OS
 release name. So, for example, Ubuntu 18.04 (Bionic Beaver) on LXD is
-``ubuntu:bionic::ubuntu::bionic`` or RHEL8 on Amazon is
+``ubuntu:jammy::ubuntu::jammy::22.04`` or RHEL8 on Amazon is
 ``ami-justanexample::rhel::8``. When a full specification is given,
 only tests which are intended for use on that OS and release will be
 executed.
@@ -364,7 +350,7 @@ The ``client`` fixture should be used for most test cases. It ensures:
 ``module_client`` and ``class_client`` fixtures also exist for the
 purpose of running multiple tests against a single launched instance.
 They provide the exact same functionality as ``client``, but are
-scoped to the module or class respectively.
+scoped to the module or class respectively.ci
 
 ``session_cloud``
 -----------------
--- a/doc/rtd/development/integration_tests.rst
+++ b/doc/rtd/development/integration_tests.rst
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
 
 .. code-block:: bash
 
-    $ tox -e integration-tests
+    tox -e integration-tests
 
-``pytest`` arguments may also be passed. For example:
+``pytest`` arguments may also be passed.
+
+Or, specify a path to run all tests inside the given file or directory:
 
 .. code-block:: bash
 
-    $ tox -e integration-tests tests/integration_tests/modules/test_combined.py
+    tox -e integration-tests tests/integration_tests/modules/test_combined.py
+
+
+.. code-block:: bash
+
+    tox -e integration-tests tests/integration_tests/modules
+
+Or, run a specific test:
+
+.. code-block:: bash
+
+    tox -e integration-tests tests/integration_tests/modules/test_combined.py::test_bootcmd
+
+TODO: Maybe make these 3 commands in one tab group??
+
 
 Configuration
 =============
 
 All possible configuration values are defined in
-`tests/integration_tests/integration_settings.py`_. Defaults can be overridden
+`tests/integration_tests/integration_settings.py`_. Look in this file for
+the full list of variables that are available and for context on what each
+variable does and what the default values are.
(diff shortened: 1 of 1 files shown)


doc/rtd/development/integration_tests.rst
```diff
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
 
 .. code-block:: bash
 
-    $ tox -e integration-tests
+    tox -e integration-tests
 
-``pytest`` arguments may also be passed. For example:
+``pytest`` arguments may also be passed.
+
+Or, specify a path to run all tests inside the given file or directory:
```
TheRealFalcon, line 38: I'm not sure we need 3 separate examples here showing how to run a test. One of the benefits of using pytest and tox is that things "just work" like they do for any other pytest or tox project. I think one example here would suffice with perhaps a link to https://docs.pytest.org/en/7.1.x/how-to/usage.html

doc/rtd/development/integration_tests.rst
```diff
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
...
+
 
 Configuration
 =============
 
 All possible configuration values are defined in
-`tests/integration_tests/integration_settings.py`_. Defaults can be overridden
+`tests/integration_tests/integration_settings.py`_. Look in this file for
+the full list of variables that are available and for context on what each
+variable does and what the default values are.
```
TheRealFalcon, line 64: You chopped off the `Defaults can be overridden` text so now the following sentence is a fragment.

doc/rtd/development/integration_tests.rst
```diff
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
...
+
+By default, logs are collected only when a test fails, by running ``cloud-init
+collect-logs`` on the instance. To collect logs after every test run, set the
+``COLLECT_LOGS`` variable to ``ALWAYS``.
+
+By default, the logs are collected to the ``/tmp/cloud_init_test_logs``
+directory. To change the directory, set the ``LOCAL_LOG_PATH`` variable to
+the desired path.
+
+TODO: can the collect log path be set to a relative path?
```
TheRealFalcon, line 135: Yes

doc/rtd/development/integration_tests.rst
```diff
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
...
+        .. code-block:: python
+
+            INCLUDE_COVERAGE = True
+
+
+Addtionally, for profiling the integration tests, set the ``INCLUDE_PROFILE``
+variable to ``True``. This will generate a profile report for the integration
+test run, and the report will be stored in the directory specified by
+``LOCAL_LOG_PATH``.
+# We default our coverage to False because it involves modifying the
```
TheRealFalcon, line 180: You left some copypasta here

doc/rtd/development/integration_tests.rst
```diff
@@ -31,26 +31,179 @@ integration tests. To run all integration tests, you would run:
...
 providing an environment variable of the same name prepended with
 ``CLOUD_INIT_``. For example, to set the ``PLATFORM`` setting:
 
 .. code-block:: bash
 
-    CLOUD_INIT_PLATFORM='ec2' pytest tests/integration_tests/
+    CLOUD_INIT_PLATFORM='ec2' tox -e integration_tests -- tests/integration_tests/
+
+
+Common integration test run configurations
```
TheRealFalcon, line 74: Nice explanations! For the examples, by the end they're starting to feel redundant.  I'm wondering if it'd be better to have a single example at the top and then another example at the bottom showing multiple options being set at once. Either way, I think it's ok as-is though, so if you disagree, no changes needed.

doc/rtd/development/integration_tests.rst
```diff
@@ -65,6 +218,39 @@ For a minimal setup using LXD, write the following to
...
 
+
+For more information on configuring pycloudlib, see the
+`pycloudlib configuration documentation`_.
+
+To specify a specific cloud to test against, first, ensure that your pycloudlib
+configuration is set up correctly. Then, modify the ``PLATFORM`` variable to be
+on of:
+
+- ``azure``: Microsoft Azure
```
TheRealFalcon, line 229: This feels like one of those lists we'll forget to update as time goes on. [integration_settings.py](https://github.com/canonical/cloud-init/blob/779dd6b009f8a568fe74e98df9803ad590b83044/tests/integration_tests/integration_settings.py#L18) already enumerates the list of valid clouds. Any way we can just point there instead?

TheRealFalcon (changes requested):
Some great updates here, thanks! I left some comments inline. Additionally, there are a few things you didn't write, but when reading the page, I noticed should be updated.

`Platform-specific
marks can be used to limit tests to particular platforms.` This is no longer true. Can we remove it? Related, would you be able to add a section about skipping tests by platform or OS? We used to use marks for this, but that got really complicated, so now we use `@pytest.mark.skipif(...)` with things imported from [releases.py](https://github.com/canonical/cloud-init/blob/main/tests/integration_tests/releases.py).

E.g.,
```
from tests.integration_tests.releases import CURRENT_RELEASE, NOBLE
...
@pytest.mark.skipif(CURRENT_RELEASE < NOBLE, reason="blah blah blah")
...
```

Also, in the image selection section, `<image_id>[::<os>[::<release>]]` should now be `<image_id>[::<os>::<release>::<version>]`.

doc/rtd/development/integration_tests.rst
a-dubs: I was wondering about that too, but was thinking that this is probably one of the areas that will change the slowest, right? Updating docs as code changes is just kind of the pain that comes with documentation, and the code in this case will seldom ever change, so I think worrying about that is futile, no?

doc/rtd/development/integration_tests.rst
a-dubs: good catch XD

doc/rtd/development/integration_tests.rst
a-dubs: Hmm... I might be swayed by this, but still think it is worth directly calling out, because I think specifying a path is one of the most common use cases for integration tests. Rarely will a new contributor be running the entire integrations sweet due to the insanely long run time? 

Let me see how this looks in tabular form so it takes up less space and maybe this will feel less like a waste of space.

a-dubs: So worst comes to worst, I agree, 3 different examples is not necessary and just one command showing how to specify a path would suffice.

a-dubs:
> `Platform-specific marks can be used to limit tests to particular platforms.` This is no longer true. 

Would you be able to elaborate what you mean by this?

Isn't the following an example of this in action?
https://github.com/canonical/cloud-init/blob/779dd6b009f8a568fe74e98df9803ad590b83044/tests/integration_tests/datasources/test_ec2_ipv6.py#L22C1-L22C70

Once I have some clarity on this, I agree that it would be valuable to expand further on how to develop new integration tests / modify existing ones. (this is actually my next documentation task)

doc/rtd/development/integration_tests.rst
a-dubs: I'd be happy to move this to a different file or to the bottom of the document or something. I agree that it starts feeling very repetitive but I do really see value in showing all the examples. So any thoughts on breaking this out into a separate page or something?

TheRealFalcon: Nah, if you think it's ok, then just leave it. You have a fresher perspective on this, so I trust you here more than me.

doc/rtd/development/integration_tests.rst
TheRealFalcon: I guess...to me it's already documented in the integration_settings.py file, so it feels a bit redundant here. I'll leave it up to you.

TheRealFalcon:
> @TheRealFalcon Would you be able to elaborate what you mean by this?

See [this example](https://github.com/canonical/cloud-init/blob/da4d72d72139fe561d50d49e6b06179b98c0ac2f/tests/integration_tests/bugs/test_gh868.py#L17-L23) from 22.3. We now just use `skipif` instead of individual marks.

TheRealFalcon (commented):
Two more minor things I saw on another read through, and then we should be good!

doc/rtd/development/integration_tests.rst
```diff
@@ -27,30 +27,168 @@ Test execution
...
+            KEEP_INSTANCE = True
+
+
+Use in-place cloud-init source code
+-------------------------------------
+
+The simplest way to test an integraton test using your current cloud-init
+changes is to set the ``CLOUD_INIT_SOURCE`` to ``IN_PLACE``. This works ONLY
+on LXD containers. This will mount the source code as-is directly into
+the container to override the pre-existing cloudinit module. This
```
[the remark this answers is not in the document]
a-dubs: good catch!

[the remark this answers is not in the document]
a-dubs: good catch!

a-dubs:
@TheRealFalcon made all changes as requested. hopefully this is good to go!

doc/rtd/development/integration_tests.rst
```diff
@@ -65,6 +204,39 @@ For a minimal setup using LXD, write the following to
...
+
+For more information on configuring pycloudlib, see the
+`pycloudlib configuration documentation`_.
+
+To specify a specific cloud to test against, first, ensure that your pycloudlib
+configuration is set up correctly. Then, modify the ``PLATFORM`` variable to be
+on of:
+
+- ``azure``: Microsoft Azure
+- ``ec2``: Amazon EC2t
```
TheRealFalcon, line 216: Whoops. After this one we should be really actually for sure good :wink:

a-dubs: haha great catch. thanks @TheRealFalcon !

TheRealFalcon (approved):

PR #5561, opened 2024-07-26
chore(eni): Deprecate ENI as an input configuration format

holmanb:
## Proposed Commit Message

```
chore(eni): Deprecate ENI as an input configuration format
```

## Additional Context

The documentation around ENI had numerous factual errors: cloud-init doesn't support ENI configuration from OpenStack or OpenNebula datasources. Only NoCloud and ConfigDrive actually support this format. Both provide distro-agnostic alternatives.

I don't think that we want to support 3 user-facing input formats indefinitely. Energy would be better spent on cloud-init's networking stack if we focused users (and developer effort) onto fewer formats.

Deprecate and un-document.

## Merge type


TheRealFalcon (approved):
LGTM!

holmanb:
Pushed to fix ruff failure

PR #5563, opened 2024-07-29
dhcp: add option to use NetworkManager for DHCP discovery

lkundrak:
Many distros nowadays, especially that the ISC DHCP client got abandoned usptream, ship NetworkManager as their only DHCP client. Allow using it for the init stage networking.

- [x] I have signed the CLA: https://ubuntu.com/legal/contributors
- [x] I have added my Github username to ``tools/.github-cla-signers``
- [x] I have included a comprehensive commit message using the guide below
- [x] I have added unit tests to cover the new behavior under ``tests/unittests/``
  - Test files should map to source files i.e. a source file ``cloudinit/example.py`` should be tested by ``tests/unittests/test_example.py``
  - Run unit tests with ``tox -e py3``
- [x] I have kept the change small, avoiding unnecessary whitespace or non-functional changes.
- [x] I have added a reference to issues that this PR relates to in the PR message (Refs GH-1234, Fixes GH-1234)
- [ ] I have updated the documentation with the changed behavior.
  - If the change doesn't change the user interface and is trivial, this step may be skipped.
  - Cloud-config documentation is generated from the jsonschema.
  - Generate docs with ``tox -e docs``.

### DIFF
--- a/cloudinit/net/dhcp.py
+++ b/cloudinit/net/dhcp.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2017 Canonical Ltd.
 #
 # Author: Chad Smith <chad.smith@canonical.com>
+# Author: Lubomir Rintel <lkundrak@v3.sk>
 #
 # This file is part of cloud-init. See LICENSE file for license information.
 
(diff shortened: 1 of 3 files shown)


holmanb (commented):
Thanks for this proposal @lkundrak! I haven't tested it yet, but I just left a couple of comments inline for you and other community members. Changing the whole Local stage to be after NetworkManager would break some expected behaviors, but perhaps those things don't need to stay as they are in the long term.

PR #5569, opened 2024-07-29
chore(debian): Remove vestigial postinst and preinst code

holmanb:
## Proposed Commit Message

```
chore(debian): Remove vestigial postinst and preinst code

Both were version gated and do not run.
```

## Merge type


blackboxsw (approved):

packages/debian/cloud-init.postinst
```diff
@@ -1,16 +0,0 @@
-#!/bin/sh
-cleanup_lp1552999() {
-    local oldver="$1" last_bad_ver="0.7.7~bzr1178"
```
blackboxsw, line 3: Given the release version gate is >8 years ago, this makes sense to drop. There are many interim releases/upgrade that should have taken care of this upgrade path.

PR #5573, opened 2024-07-31
Add new helper, refactor utility functions into separate module

holmanb:
## Additional Context

This PR changes no functionally in cloud-init. It adds a small helper (4 LOC, excluding docstring). The intent for this helper was mentioned [here](https://github.com/canonical/cloud-init/pull/5202#issuecomment-2259369639).

- commit 1) -> factor out some lifecycle management code into a shared module - at 3.3k LOC and growing util.py should be split into smaller chunks 
- commit 2) -> Add a helper function that will help us get https://github.com/canonical/cloud-init/pull/5202 landed. This PR adds lots of warnings which adds downstream risk for stable releases. This will assist us in introducing warnings while reducing the risk to downstream releases.

## Merge type

### DIFF
--- a/cloudinit/lifecycle.py
+++ b/cloudinit/lifecycle.py
@@ -121,8 +121,13 @@ def should_log_deprecation(version: str, boundary_version: str) -> bool:
     ) <= Version.from_str(boundary_version)
 
 
-def log_if_acceptable(
-    *, logger: logging.Logger, version: str, level: int, msg: str, args
+def log_with_downgradable_level(
+    *,
+    logger: logging.Logger,
+    version: str,
+    requested_level: int,
+    msg: str,
+    args,
 ):
     """Log a message at the requested level, if that is acceptable.
 
@@ -134,12 +139,12 @@ def log_if_acceptable(
     :param version: Version string of the version that this log was introduced
     :param level: Preferred level at which this message should be logged
     :param msg: Message, as passed to the logger.
-    :param args: Message parameters
+    :param args: Message formatting args, ass passed to the logger
 
     :return: True if the message should be logged, else False.
     """
     if should_log_deprecation(version, features.DEPRECATION_INFO_BOUNDARY):
-        logger.log(level, msg, args)
+        logger.log(requested_level, msg, args)
     else:
         logger.debug(msg, args)
 
@@ -155,21 +160,21 @@ def deprecate(
     """Mark a "thing" as deprecated. Deduplicated deprecations are
     logged.
 
-    @param deprecated: Noun to be deprecated. Write this as the start
+    :param deprecated: Noun to be deprecated. Write this as the start
         of a sentence, with no period. Version and extra message will
         be appended.
-    @param deprecated_version: The version in which the thing was
+    :param deprecated_version: The version in which the thing was
         deprecated
-    @param extra_message: A remedy for the user's problem. A good
+    :param extra_message: A remedy for the user's problem. A good
         message will be actionable and specific (i.e., don't use a
         generic "Use updated key." if the user used a deprecated key).
         End the string with a period.
-    @param schedule: Manually set the deprecation schedule. Defaults to
+    :param schedule: Manually set the deprecation schedule. Defaults to
         5 years. Leave a comment explaining your reason for deviation if
         setting this value.
-    @param skip_log: Return log text rather than logging it. Useful for
+    :param skip_log: Return log text rather than logging it. Useful for
         running prior to logging setup.
-    @return: NamedTuple containing log level and log message
+    :return: NamedTuple containing log level and log message
         DeprecationLog(level: int, message: str)
 
     Note: uses keyword-only arguments to improve legibility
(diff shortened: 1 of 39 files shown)


cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+
+    :param logger: Logger object to log with
+    :param version: Version string of the version that this log was introduced
+    :param level: Preferred level at which this message should be logged
+    :param msg: Message, as passed to the logger.
+    :param args: Message parameters
+
+    :return: True if the message should be logged, else False.
+    """
+    if should_log_deprecation(version, features.DEPRECATION_INFO_BOUNDARY):
```
TheRealFalcon, line 141: If I were to call `log_if_acceptable()` at level INFO, then it would downgrade it to DEBUG? I'm guessing there's no good reason to do that, so I don't think this needs to change, but I just want to make sure we don't think we need to do any `level` checking here first.

cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+    :param boundary_version: The version at which deprecation level is logged.
+
+    :return: True if the message should be logged, else False.
+    """
+    return boundary_version == "devel" or Version.from_str(
+        version
+    ) <= Version.from_str(boundary_version)
+
+
+def log_if_acceptable(
```
TheRealFalcon, line 124: Nit: Can we come up with a more descriptive name? Are `log_at_variable_level()` or `log_with_degradable_level()` any better?

cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+    *, logger: logging.Logger, version: str, level: int, msg: str, args
```
TheRealFalcon, line 125: Nit: I think `requested_level` would be descriptive here.

TheRealFalcon (approved):
Nice refactor! I left a few nits, and a "make sure this is fine" inline, but nothing blocking.

cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+
+    If the log level is too high due to the version boundary, log at DEBUG
+    level. Useful to add new warnings to previously unguarded code without
+    disrupting stable downstreams.
+
+    :param logger: Logger object to log with
+    :param version: Version string of the version that this log was introduced
+    :param level: Preferred level at which this message should be logged
+    :param msg: Message, as passed to the logger.
+    :param args: Message parameters
```
a-dubs, line 137: i feel like this is somewhat better for clarity but maybe others feel it is unnecessary 
```suggestion
    :param args: Message formatting args, as passed to the logger
```

cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+    @param extra_message: A remedy for the user's problem. A good
+        message will be actionable and specific (i.e., don't use a
+        generic "Use updated key." if the user used a deprecated key).
+        End the string with a period.
+    @param schedule: Manually set the deprecation schedule. Defaults to
+        5 years. Leave a comment explaining your reason for deviation if
+        setting this value.
+    @param skip_log: Return log text rather than logging it. Useful for
+        running prior to logging setup.
+    @return: NamedTuple containing log level and log message
```
a-dubs, line 172: Shouldn't all of these `@param`'s and this `@return` be `:param ...:` and `:return:` ?

cloudinit/lifecycle.py
```diff
@@ -0,0 +1,237 @@
...
+    """Mark a "thing" as deprecated. Deduplicated deprecations are
+    logged.
+
+    @param deprecated_version: The version in which the thing was
+        deprecated
+    @param extra_message: A remedy for the user's problem. A good
+        message will be actionable and specific (i.e., don't use a
+        generic "Use updated key." if the user used a deprecated key).
+        End the string with a period.
+    @param schedule: Manually set the deprecation schedule. Defaults to
```
a-dubs, line 215: same comment as above for `@param` vs `:param ...:`

a-dubs (changes requested):
few small changes requested but looks good otherwise! nice refactor, sir holman.

cloudinit/lifecycle.py
holmanb: > If I were to call log_if_acceptable() at level INFO, then it would downgrade it to DEBUG?

Correct

> I'm guessing there's no good reason to do that, so I don't think this needs to change, but I just want to make sure we don't think we need to do any level checking here first.

Thanks for raising this point. It wouldn't make sense to use it in that way, but this is just a simple helper intended to make this logic less error-prone and callsites more readable. I'm not too worried about misuse because the docstring explains its intent, and this should be used on a case-by-case basis, and also the consequences of misuse are pretty minor. I agree with leaving it as is without level checking.

holmanb:
@TheRealFalcon @a-dubs Thanks for the reviews!

a-dubs:
Anything for you, @holmanb

Issue #5577, opened 2024-08-01
Cloud-init fails on AWS if IMDSv2 returns a 503 error.


taylorpn:
# Bug report
Cloud-init does not retry IMDSv2 token calls when 503 is returned. According to the public documentation, a 503 error code should be retried and not result in a failure in and of itself.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

## Steps to reproduce the problem
The problem cannot be reproduced on demand.

## Environment details
- Cloud-init version: 19.3 (All versions suspected.)
- Operating System Distribution: Amazon Linux 2023 (But could happen on any cloud-init-enabled OS running on AWS.)
- Cloud provider, platform or installer type: Amazon Web Services

## cloud-init logs
```
2024-07-24 08:13:06,797 - DataSourceEc2.py[DEBUG]: Fetching Ec2 IMDSv2 API Token
2024-07-24 08:13:06,797 - url_helper.py[DEBUG]: [0/1] open 'http://169.254.169.254/latest/api/token' with {'url': 'http://169.254.169.254/latest/api/token', 'allow_redirects': True, 'method': 'PUT', 'timeout': 50.0, 'headers': {'User-Agent': 'Cloud-Init/22.2-0ubuntu1~20.04.3', 'X-aws-ec2-metadata-token-ttl-seconds': 'REDACTED'}} configuration
2024-07-24 08:13:06,801 - url_helper.py[DEBUG]: Read from http://169.254.169.254/latest/api/token (503, 359b) after 1 attempts
2024-07-24 08:13:06,801 - url_helper.py[WARNING]: Calling 'http://169.254.169.254/latest/api/token' failed [0/120s]: bad status code [503]
2024-07-24 08:13:06,801 - DataSourceEc2.py[WARNING]: Fatal error while requesting Ec2 IMDSv2 API tokens
2024-07-24 08:13:06,801 - DataSourceEc2.py[WARNING]: IMDS's HTTP endpoint is probably disabled
```
```
[   15.417209] cloud-init[10775]: Feb 12 11:15:27 cloud-init[10775]: url_helper.py[WARNING]: Calling 'http://169.254.169.254/latest/api/token' failed [0/120s]: bad status code [503]
[   15.417267] cloud-init[10775]: Feb 12 11:15:27 cloud-init[10775]: DataSourceEc2.py[WARNING]: Fatal error while requesting Ec2 IMDSv2 API tokens
[...]

PR #5578, opened 2024-08-02
feat(url_helper): Retry on 503 error

holmanb:
If the server is busy, no need to fail.
Add type hints to adjascent code paths.

Fixes GH-5577

## Proposed Commit Message

```
feat(url_helper): Retry on 503 error

If the server is busy, no need to fail.
Add type hints to adjascent code paths.

Fixes GH-5577
```

## Additional Context
Fixes https://github.com/canonical/cloud-init/issues/5577

## Merge type

### DIFF
--- a/cloudinit/url_helper.py
+++ b/cloudinit/url_helper.py
@@ -698,7 +698,7 @@ def wait_for_url(
             time.monotonic() - start_time + sleep_time > max_wait
         )
 
-    def handle_url_response(response, url):
+    def handle_url_response(response, url) -> Tuple[Optional[Exception], str]:
         """Map requests response code/contents to internal "UrlError" type"""
         if not response.contents:
             reason = "empty response [%s]" % (response.code)
@@ -752,7 +752,7 @@ def wait_for_url(
             # in the future, for example this is what the MAAS datasource
             # does.
             exc_cb(msg=status_msg, exception=url_exc)
-        return (url, url_exc)
+        return url, url_exc
 
     def read_url_cb(url: str, timeout: int) -> UrlResponse:
         return readurl(
@@ -835,14 +835,20 @@ def wait_for_url(
         url = do_read_url(start_time, timeout, exception_cb, status_cb)
         if url:
             address, response = url
-            if isinstance(response, UrlError) and 503 == response.code:
-                # server isn't available, so retry
-                LOG.warning(
-                    "IMDS returned 503 error code. Retrying in %s",
-                    current_sleep_time,
-                )
-            elif isinstance(response, UrlResponse):
+            if isinstance(response, UrlResponse):
                 return (address, response.contents)
+            elif isinstance(response, UrlError):
+                if 503 == response.code:
+                    # server isn't available, so retry
+                    LOG.warning(
+                        "IMDS returned 503 error code. Retrying in %s",
+                        current_sleep_time,
+                    )
+                raise response
+            elif isinstance(response, Exception):
+                raise response
+            else:
+                raise TypeError("Unexpected return type: %s", repr(response))
 
         if timeup(max_wait, start_time, current_sleep_time):
             break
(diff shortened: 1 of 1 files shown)


tekenstam (approved):

cloudinit/url_helper.py
```diff
@@ -827,7 +835,14 @@ def read_url_parallel(start_time, timeout, exc_cb, log_cb):
...
         if url:
             address, response = url
-            return (address, response.contents)
+            if isinstance(response, UrlError) and 503 == response.code:
+                # server isn't available, so retry
+                LOG.warning(
+                    "IMDS returned 503 error code. Retrying in %s",
+                    current_sleep_time,
+                )
+            elif isinstance(response, UrlResponse):
```
TheRealFalcon, line 844: So it looks like the old code was making some bad assumptions...particularly that `response` is a `UrlResponse` even though it could also be `UrlError` if an error occurred. It also looks like the callers of this code aren't really doing anything to handle exceptions. Previously if we did get a UrlError, the old `response.contents` would throw an error due to `contents` not being a member of UrlError. It was raising the wrong error obviously, but it still wound up having the same effect as the requests' `raise_for_status()`.

With this though, if we get an error that isn't a 503, we'll retry until we timeout, even if it's a non-recoverable error. I think we should instead raise the UrlError if it's found and not a 503.

TheRealFalcon (changes requested):
See the inline comment

aciba90 (commented):
Thanks, @holmanb, for fixing this. 

[503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) responses are recommended to contain a [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header containing an estimated time for the recovery of the service. 

Do we want to be good citizens and build logic to respect that recommended wait-time if present and not contribute to the server overload?

cloudinit/url_helper.py
holmanb: Good point, thanks for catching that @TheRealFalcon!

cloudinit/url_helper.py
```diff
@@ -827,7 +835,20 @@ def read_url_parallel(start_time, timeout, exc_cb, log_cb):
...
+            if isinstance(response, UrlResponse):
+                return (address, response.contents)
+            elif isinstance(response, UrlError):
+                if 503 == response.code:
+                    # server isn't available, so retry
+                    LOG.warning(
+                        "IMDS returned 503 error code. Retrying in %s",
+                        current_sleep_time,
+                    )
+                raise response
```
TheRealFalcon, line 847: We need an `else` here, right?

TheRealFalcon:
Is it also possible to get a unit test or two showing contents return correctly on 200, retry happens on 503, and exception is raised on anything else? I'm guessing one or two of those already exist, but we should test the 503 case.

cloudinit/url_helper.py
holmanb: > It looks like the old code was making some bad assumptions...particularly that response is a UrlResponse even though it could also be UrlError if an error occurred. 
> ...
> Previously if we did get a UrlError, the old response.contents would throw an error due to contents not being a member of UrlError. It was raising the wrong error obviously, but it still wound up having the same effect as the requests' raise_for_status().

This isn't correct. Previously, only UrlResponse was returned. See the change on line 755.

> It also looks like the callers of this code aren't really doing anything to handle exceptions.

Exceptions are optionally handled by the function passed in via `exception_cb`. Note the line in the original bug report:
```
2024-07-24 08:13:06,801 - DataSourceEc2.py[WARNING]: Fatal error while requesting Ec2 IMDSv2 API tokens
```

Looking into this a bit further, I think that we actually need to make a change to the function that caused this log to allow 503 exceptions to retry.

Issue #5304, opened 2024-05-18
Cloud-init status always reporting running


Overmorrow1501:
After boot the newly created VM and every task cloud-init should perform is done, cloud-init status still reports status "Running"

## cloud-init configuration
/etc/cloud/cloud.cfg:
```yaml
cloud_init_modules:
  - growpart
  - resizefs

network: {config: disabled}
datasource_list: [ 'None' ]

growpart:
  mode: auto
  devices: ["/"]

resize_rootfs: true

system_info:
  distro: ubuntu
  paths:
    cloud_dir: /var/lib/cloud/
    templates_dir: /etc/cloud/templates/
```
###Steps to reproduce the problem

```bash
$ cloud-init clean --logs --seed --machine-id --reboot

$ cloud-init status --long
status: running
extended_status: running
boot_status_code: enabled-by-generator
last_update: Sat, 18 May 2024 14:21:26 +0000
detail:
Cloud-init enabled by systemd cloud-init-generator
errors: []
recoverable_errors: {}

$ cloud-init status --format json
{
  "_schema_version": "1",
  "boot_status_code": "enabled-by-generator",
  "datasource": null,
  "detail": "Cloud-init enabled by systemd cloud-init-generator",
  "errors": [],
  "extended_status": "running",
  "init": {
    "errors": [],
    "finished": null,
    "start": null
  },
  "init-local": {
    "errors": [],
    "finished": null,
    "start": null
  },
  "last_update": "Sat, 18 May 2024 14:21:26 +0000",
  "modules-config": {
    "errors": [],
    "finished": 1716042085.1915476,
    "recoverable_errors": {},
    "start": 1716042085.1726842
  },
  "modules-final": {
    "errors": [],
    "finished": 1716042086.0076993,
    "recoverable_errors": {},
    "start": 1716042085.9895425
  },
  "recoverable_errors": {},
  "schemas": {
    "1": {
      "boot_status_code": "enabled-by-generator",
      "datasource": null,
      "detail": "Cloud-init enabled by systemd cloud-init-generator",
      "errors": [],
      "extended_status": "running",
      "init": {
        "errors": [],
        "finished": null,
        "start": null
      },
      "init-local": {
        "errors": [],
        "finished": null,
        "start": null
      },
      "last_update": "Sat, 18 May 2024 14:21:26 +0000",
      "modules-config": {
        "errors": [],
        "finished": 1716042085.1915476,
        "recoverable_errors": {},
[...]

PR #5579, opened 2024-08-02
feat: Return error in `cloud-init status` if stage not run

TheRealFalcon:
In draft state as I'm looking for feedback and still needs test updates.

## Proposed Commit Message

```
feat: Return error in `cloud-init status` if stage not run

Fixes GH-5304
```

## Additional Context
The original issue in `GH-5304` was not reproducible on current versions of cloud-init. Since that version of cloud-init, we've updated our status checking to only be running if [`status.json` exists but `results.json` doesn't exist](https://github.com/canonical/cloud-init/blob/d15a7704b6e1317b1fa186667085372c69e6f586/cloudinit/cmd/status.py#L374), and checking systemd status can only change cloud-init status [from RUNNING to ERROR](https://github.com/canonical/cloud-init/blob/d15a7704b6e1317b1fa186667085372c69e6f586/cloudinit/cmd/status.py#L501) rather than the previous code that [could change status back to RUNNING](https://github.com/canonical/cloud-init/blob/781f77c2f673189bba81df2d0d9acd231972bc78/cloudinit/cmd/status.py#L362).

So instead I addressed this comment:
`Long term cloud-init status should return a fatal error code and indicate this issue`

Thinking about it more though, I'm not actually sure we want this behavior, and if we do it probably shouldn't be applied to existing releases. While the supported cloud-init path is to run all 4 services, we have seen numerous use cases where people do things we wouldn't endorse. Amazon Linux completely removing cloud-init-local.service comes to mind. I'm not proposing we officially support those use cases, but I'm also not sure it makes sense to declare there was an error when the lack of a service running may have been intended behavior and technically nothing "bad" happened.

This PR makes it error to have not run one of the stages. After masking cloud-init-local.service and cloud-init.service, I get these results:
```
root@me:~# cloud-init status --long
status: error
extended_status: error - done
last_update: Thu, 01 Jan 1970 00:00:02 +0000
detail: Cloud-init enabled by systemd cloud-init-generator
errors:
	- Failed to start stage: init
	- Failed to start stage: init-local
recoverable_errors:
WARNING:
	- Failed to write boot finished file /var/lib/cloud/instance/boot-finished
	- Used fallback datasource
```

```
root@me:~# cloud-init status --format json
{
  "boot_status_code": "enabled-by-generator",
  "datasource": null,
  "detail": "Cloud-init enabled by systemd cloud-init-generator",
  "errors": [
    "Failed to start stage: init",
    "Failed to start stage: init-local"
  ],
  "extended_status": "error - done",
  "init": {
    "errors": [],
    "finished": null,
    "recoverable_errors": {},
    "start": null
  },
  "init-local": {
    "errors": [],
    "finished": null,
    "recoverable_errors": {},
    "start": null
  },
  "last_update": "Thu, 01 Jan 1970 00:00:02 +0000",
  "modules-config": {
    "errors": [],
    "finished": 2.23,
    "recoverable_errors": {},
    "start": 2.12
  },
  "modules-final": {
    "errors": [],
    "finished": 2.5,
    "recoverable_errors": {
      "WARNING": [
        "Failed to write boot finished file /var/lib/cloud/instance/boot-finished",
        "Used fallback datasource"
      ]
    },
    "start": 2.44
  },
  "recoverable_errors": {
    "WARNING": [
      "Failed to write boot finished file /var/lib/cloud/instance/boot-finished",
      "Used fallback datasource"
    ]
  },
  "stage": null,
  "status": "error"
}
```
(the timestamps are related to my environment and not an issue)

## Merge type

### DIFF
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -435,12 +435,12 @@ def get_not_started_errors(status_v1) -> List[str]:
     """Return a list of errors from status_v1 that are not started."""
     return [
         f"Failed to start stage: {stage_name}"
-        for stage_name, stage_info in status_v1.items()
-        if stage_name in STAGE_NAME.keys() and stage_info.get("start") is None
+        for stage_name in STAGE_NAME
+        if status_v1.get(stage_name, {}).get("start") is None
     ]
 
 
-def get_errors(status_v1) -> Tuple[List, Dict]:
+def get_errors(status_v1, running_status: RunningStatus) -> Tuple[List, Dict]:
     """Return a list of errors and recoverable_errors from status_v1."""
     errors = []
     recoverable_errors = {}
@@ -461,7 +461,8 @@ def get_errors(status_v1) -> Tuple[List, Dict]:
                     recoverable_errors[err_type].extend(
                         current_recoverable_errors[err_type]
                     )
-    errors.extend(get_not_started_errors(status_v1)) 
+    if running_status == RunningStatus.DONE:
+        errors.extend(get_not_started_errors(status_v1))
     return errors, recoverable_errors
 
 
@@ -496,15 +497,15 @@ def get_status_details(
         else ""
     )
 
-    errors, recoverable_errors = get_errors(status_v1)
+    running_status = get_running_status(
+        status_file, result_file, boot_status_code, latest_event
+    )
+
+    errors, recoverable_errors = get_errors(status_v1, running_status)
     if errors:
         condition_status = ConditionStatus.ERROR
     elif recoverable_errors:
         condition_status = ConditionStatus.DEGRADED
-
-    running_status = get_running_status(
-        status_file, result_file, boot_status_code, latest_event
-    )
 
     if (
         running_status == RunningStatus.RUNNING
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -17,6 +17,7 @@ from typing import Any, Dict, List, NamedTuple, Optional, Tuple
 
 from cloudinit import safeyaml, subp
 from cloudinit.cmd.devel import read_cfg_paths
+from cloudinit.cmd.main import STAGE_NAME
 from cloudinit.distros import uses_systemd
 from cloudinit.helpers import Paths
 from cloudinit.util import get_cmdline, load_json, load_text_file
@@ -430,6 +431,15 @@ def get_latest_event(status_v1):
     return latest_event
 
 
+def get_not_started_errors(status_v1) -> List[str]:
+    """Return a list of errors from status_v1 that are not started."""
+    return [
+        f"Failed to start stage: {stage_name}"
+        for stage_name, stage_info in status_v1.items()
+        if stage_name in STAGE_NAME.keys() and stage_info.get("start") is None
+    ]
+
+
 def get_errors(status_v1) -> Tuple[List, Dict]:
     """Return a list of errors and recoverable_errors from status_v1."""
     errors = []
@@ -451,6 +461,7 @@ def get_errors(status_v1) -> Tuple[List, Dict]:
                     recoverable_errors[err_type].extend(
                         current_recoverable_errors[err_type]
                     )
+    errors.extend(get_not_started_errors(status_v1)) 
     return errors, recoverable_errors
 
 
--- a/tests/unittests/cmd/test_status.py
+++ b/tests/unittests/cmd/test_status.py
@@ -355,10 +355,12 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
                         "blah": {"finished": 123.456},
                         "init": {
                             "errors": [],
-                            "start": 124.567,
-                            "finished": 125.678,
+                            "start": 12.3,
+                            "finished": 12.4,
                         },
-                        "init-local": {"start": 123.45, "finished": 123.46},
+                        "init-local": {"start": 12.1, "finished": 12.2},
+                        "modules-config": {"start": 12.5, "finished": 12.6},
+                        "modules-final": {"start": 12.7, "finished": 12.8},
                     }
                 },
                 None,
@@ -379,8 +381,10 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
                             "[seed=/var/.../seed/nocloud-net]"
                             "[dsmode=net]"
                         ),
-                        "init": {"start": 124.567, "finished": 125.678},
-                        "init-local": {"start": 123.45, "finished": 123.46},
+                        "init": {"start": 12.3, "finished": 12.4},
+                        "init-local": {"start": 12.1, "finished": 12.2},
+                        "modules-config": {"start": 12.5, "finished": 12.6},
+                        "modules-final": {"start": 12.7, "finished": 12.8},
                     }
                 },
                 None,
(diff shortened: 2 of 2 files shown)


cloudinit/cmd/status.py
```diff
@@ -430,6 +431,15 @@ def get_latest_event(status_v1):
     return latest_event
 
 
+def get_not_started_errors(status_v1) -> List[str]:
+    """Return a list of errors from status_v1 that are not started."""
+    return [
+        f"Failed to start stage: {stage_name}"
+        for stage_name, stage_info in status_v1.items()
+        if stage_name in STAGE_NAME.keys() and stage_info.get("start") is None
+    ]
```
blackboxsw, line 440: Couldn't this logic instead be iterating by known STAGE_NAMES instead of looping over status_v1.items and filtering out?
Also this has the affect of reporting errors while cloud-init is still running because early calls to status may not have crossed the threshold where all cloud-init stages have yet run. Call to status before cloud-init-local even starts would report. Do we want to also check is_running() too before trying to return these errors?   

Also another concern, is if cloud-final.service is ever not part of the systemd (or other) boot targets, we would never create /run/cloud-init/result.json file which would leave us in a state where we probably want to peek at whether the system's desired run level was reached and the init system is 'up/active' too. 

```suggestion
    return [
        f"Failed to start stage: {stage_name}"
        for stage_name in STAGE_NAME.keys()
        if status_v1.get(stage_name, {}).get("start") is None
    ]
```

blackboxsw (commented):
Potential alternative that at least takes into account DONE status. We also don't want to error if RunningStatus.DISABLED as nothing should run at that point
```diff
--- a/cloudinit/cmd/status.py
+++ b/cloudinit/cmd/status.py
@@ -435,12 +435,12 @@ def get_not_started_errors(status_v1) -> List[str]:
     """Return a list of errors from status_v1 that are not started."""
     return [
         f"Failed to start stage: {stage_name}"
-        for stage_name, stage_info in status_v1.items()
-        if stage_name in STAGE_NAME.keys() and stage_info.get("start") is None
+        for stage_name in STAGE_NAME
+        if status_v1.get(stage_name, {}).get("start") is None
     ]
 
 
-def get_errors(status_v1) -> Tuple[List, Dict]:
+def get_errors(status_v1, running_status: RunningStatus) -> Tuple[List, Dict]:
     """Return a list of errors and recoverable_errors from status_v1."""
     errors = []
     recoverable_errors = {}
@@ -461,7 +461,8 @@ def get_errors(status_v1) -> Tuple[List, Dict]:
                     recoverable_errors[err_type].extend(
                         current_recoverable_errors[err_type]
                     )
-    errors.extend(get_not_started_errors(status_v1)) 
+    if running_status == RunningStatus.DONE:
+        errors.extend(get_not_started_errors(status_v1))
     return errors, recoverable_errors
 
 
@@ -496,16 +497,16 @@ def get_status_details(
         else ""
     )
 
-    errors, recoverable_errors = get_errors(status_v1)
+    running_status = get_running_status(
+        status_file, result_file, boot_status_code, latest_event
+    )
+
+    errors, recoverable_errors = get_errors(status_v1, running_status)
     if errors:
         condition_status = ConditionStatus.ERROR
     elif recoverable_errors:
         condition_status = ConditionStatus.DEGRADED
 
-    running_status = get_running_status(
-        status_file, result_file, boot_status_code, latest_event
-    )
-
     if (
         running_status == RunningStatus.RUNNING
         and uses_systemd()
```

holmanb:
> Thinking about it more though, I'm not actually sure we want this behavior, and if we do it probably shouldn't be applied to existing releases. While the supported cloud-init path is to run all 4 services, we have seen numerous use cases where people do things we wouldn't endorse. Amazon Linux completely removing cloud-init-local.service comes to mind.

> I'm not proposing we officially support those use cases, but I'm also not sure it makes sense to declare that there was an error when the lack of a service running may have been intended behavior and technically nothing "bad" happened.

Downstreams can do whatever they want, but if they choose to do really weird things like skip entire stages that doesn't mean that it's upstream's responsibility to maintain their code. The issue that you have addressed here should be fixed in upstream. This doesn't prevent their downstream from working; they will just have to patch this behavior to work for their modified downstream version. Downstreams doing weird things doesn't make it upstream's maintenance problem.

If there were a way to fix this that was the same amount of work that didn't require breaking a downstream, I would probably advocate for that (taking into account good design, maintainability, etc). However it isn't feasible to only allow those changes which do not break any downstreams - mutually exclusive changes could exist which would paralyze an upstream that operates this way. I think that we should do what is right by fixing this issue, and if we're feeling generous then a ping in their downstream bug tracker about this change would be a friendly thing to do.

TheRealFalcon:
@holmanb , I said `I'm not proposing we officially support those use cases`, and then it kind of felt like your response was about how we shouldn't support those use cases. I'm not disagreeing on that point. I'm more wondering if this really is a problem or error case in the first place.

holmanb:
> @holmanb , I said `I'm not proposing we officially support those use cases`, and then it kind of felt like your response was about how we shouldn't support those use cases. I'm not disagreeing on that point. I'm more wondering if this really is a problem or error case in the first place.

I guess I don't understand what you're trying to say. Under which officially supported use case would this not be a problem or error condition?

cloudinit/cmd/status.py
TheRealFalcon: Thanks for the comment and patch. Also, my apologies, I meant to mark this PR as draft as I knew it wasn't quite ready for review yet. I've applied your patch and added an additional unit test.

> Another concern, is if cloud-final.service is ever not part of the systemd (or other) boot targets, we would never create /run/cloud-init/result.json file which would leave us in a state where we probably want to peek at whether the system's desired run level was reached and the init system is 'up/active' too.

I'm open to seeing an implementation of this, but my previous attempt at doing this wasn't it. :wink:

Since that case is outside of what we support, I see it as a future nice to have rather than a need to have.

TheRealFalcon:
> I guess I don't understand what you're trying to say. Under which officially supported use case would this not be a problem or error condition?

Just to record what me and @holmanb discussed side-channel, my main concern was the space between "supported use case" and "break at will". I didn't really see the need for adding this while knowing that it could cause some behavior change for some unsupported use cases. He made a good point that dependency ordering can create a system where cloud-init is unintentionally removed from boot, and it's right for cloud-init to report this.

TheRealFalcon:
If https://github.com/canonical/cloud-init/pull/5586 lands, it will eliminate the mypy issue.

cloudinit/cmd/status.py
blackboxsw: Yep, that concern is just noting a gap in behavior if cloud-final is kicked out of systemd boot goals due to ordering cycle, but, in those events `systemctl status list-units --failed` or `systemctl status cloud-init-main` or `cloud-final` will all show the error condition as well.

blackboxsw:
> If there were a way to fix this that was the same amount of work that didn't require breaking a downstream, I would probably advocate for that (taking into account good design, maintainability, etc). However it isn't feasible to only allow those changes which do not break any downstreams - mutually exclusive changes could exist which would paralyze an upstream that operates this way. I think that we should do what is right by fixing this issue, and if we're feeling generous then a ping in their downstream bug tracker about this change would be a friendly thing to do.

Even if we made this a "simpler" breaking change such as a config flag for "required_boot_stages" or some such flag, downstreams which don't want to run init-local will still have to carry a patch that modifies that config switch. In this case, any downstream that wants to avoid calling our intended boot stages can just remove that stage key from [cmd.main.STAGE_NAME](https://github.com/canonical/cloud-init/blob/main/cloudinit/cmd/main.py#L67)

blackboxsw:
Now that singleprocess work landed in 143bc9e40f7f33695216db60f73e10a900d1d1cd, cloud-init will not be able to actually get to the point where result.json  is written if we are missing a boot stage because with cloud-init-main will block on a sync point for each expected stage. As such, I'm not sure we need this anymore for typical use cases.

PR #5581, opened 2024-08-05
[ubuntu/noble]: patch out breaking changes

holmanb:
new snapshot and:

- patch out single process changes
- patch out nocloud changes


TheRealFalcon (commented):
Change LGTM, but changelog entries should be separate commits.

TheRealFalcon (approved):
LGTM!

PR #5582, opened 2024-08-05
[ubuntu/devel] Add netcat dependency

holmanb:
This is already part of ubuntu-minimal, but for good dependency hygiene we should track this.


TheRealFalcon (commented):
Changelog entry should be a separate commit.

Also, 24.3~1... has already been released. We should start a new entry with a bumped version and UNRELEASED label. Since we won't know the hash to use until we release, I think `0ubuntu2` is fine for now.

TheRealFalcon (approved):
LGTM!

Issue #5575, opened 2024-08-01
Cloud-init depends on undefined yaml behavior in LXD implementation


holmanb:
# Bug report
The user-defined meta-data key gets appended as a string to the lxd-provided meta-data. This means that duplicate keys can be added, which creates a configuration that isn't well defined. Both 1.1 and 1.2 of the YAML spec state that keys are unique, which this violates.

The configuration received by cloud-init:
```python
{'_metadata_api_version': '1.0',
 'config': {'user.meta-data': 'instance-id: test_2'},
 'devices': {'eth0': {'hwaddr': '00:16:3e:e3:ed:2c',
                      'name': 'eth0',
                      'network': 'lxdbr0',
                      'type': 'nic'},
             'root': {'path': '/', 'pool': 'default', 'type': 'disk'}},
 'meta-data': '#cloud-config\n'
              'instance-id: 0b6c31e2-403c-44eb-b610-ad7eafea777e\n'
              'local-hostname: oracular\n'
              'instance-id: test_2'}
```

Cloud-init's implementation uses PyYAML which happens to use the last defined key - and therefore produces the desired outcome (allow user to override the default meta-data), but it depends on [undefined behavior](https://github.com/yaml/pyyaml/issues/165). If cloud-init were ever to move to a different YAML library this behavior could break or need to be manually worked around.

In order to preserve the current behavior while creating a path to using standard-compliant yaml while preserving backwards compatibility, we could do the following:

1) cloud-init could be updated to make values in `metadata['config']['user.meta-data']` override values in `metadata['meta-data']`. This wouldn't change cloud-init's current behavior, which ignores the values in `metadata['config']`. We could optionally check for a bump to the value in `_metadata_api_version` before doing this, but this wouldn't be strictly required since this is functionally identical currently.

2) Once stable distributions have this update, we could update the api to no longer append user meta-data to the default metadata (and bump the meta-data api, if desired). While we're making this change, we might want to drop the `#cloud-config` comment too. This isn't necessary because meta-data isn't part of cloud-config.

holmanb:
Filed against LXD as well: https://github.com/canonical/lxd/issues/13853

PR #5584, opened 2024-08-05
fix: Update default LXD meta-data with user meta-data

holmanb:
## Proposed Commit Message

```
fix: Update default LXD meta-data with user meta-data

This was previously unnecessary because:

a. LXD automatically appends the user.meta-data key to default meta-data.
b. In the presence of duplicate keys, PyYAML uses the last key.

This change is part of a series of changes that will enable cloud-init to
avoid depending on undefined behavior. In the future LXD may stop appending
user-defined meta-data to its default meta-data. This change makes
cloud-init forward compatible to LXD for when that change is implemented.

GH-5575
```

## Additional context

Fixes https://github.com/canonical/cloud-init/issues/5575.
Should make cloud-init forward compatible with LXD changes which might be made to address https://github.com/canonical/lxd/issues/13853

## Merge type

### DIFF
--- a/cloudinit/sources/DataSourceLXD.py
+++ b/cloudinit/sources/DataSourceLXD.py
@@ -210,8 +210,8 @@ class DataSourceLXD(sources.DataSource):
         config = self._crawled_metadata.get("config", {})
         user_metadata = config.get("user.meta-data", {})
         if user_metadata:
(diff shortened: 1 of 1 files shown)


TheRealFalcon (approved):
LGTM!

PR #5586, opened 2024-08-06
chore: Stop running mypy on tests

TheRealFalcon:
This should be ready to review but in draft state as it may be controversial.

## Proposed Commit Message

```
chore: Stop running mypy on tests

Often times we need to test error conditions or invalid input. That
necessitates doing things that mypy considers unsound. While mypy
can also provide some benefit to testing, the cost of silencing errors
isn't worth the benefit.
```

## Merge type


a-dubs:
after having to do weird things for the ibm datasources tests, +1

blackboxsw (commented):
It does sometimes feel a bit like busy-work to be updating typing on unittests tests.

I see a bit more value in still running mypy on the integration tests which can be exposed to using pycloudlib and other library dependencies where we may more likely misuse some dependencies.

What do you think about the middle ground of dropping mypy from `tests/unittests` but still including `tests/integration_tests` subdir?

TheRealFalcon:
@holmanb , you make some good points, but contrast that with this sort of thing:
```python
def parse_json(my_json) -> Optional[Dict]:
    ...

def read_file(filename, decode) -> Union[str, bytes]:
    ...

def test_json():
    x = parse_json('{"a": 1}')
    assert x['a'] == 1

def test_file():
    x = read_file('file.txt', True)
    assert x == 'content'
```

We either need to sprinkle `# type: ignore` everywhere, or we'll wind up doing stuff like:
```python
def test_json():
    x = parse_json('{"a": 1}')
    if x is None:
        pytest.fail("oh noes")
    assert x['a'] == 1
```

I suspect we'll see more of the latter as we'll want to "properly" silence errors rather than just ignoring them. To me this is worse than silencing errors because we're needlessly bloating tests testing for conditions that are implicitly part of the assertion that already exists. Given how much of the cloud-init code base returns Optionals and Unions, I think this will be too common of a problem.

holmanb:
> @holmanb , you make some good points, but contrast that with this sort of thing:
>
> ...
>
> I suspect we'll see more of the latter as we'll want to "properly" silence errors rather than just ignoring them. To me this is worse than silencing errors because we're needlessly bloating tests testing for conditions that are implicitly part of the assertion that already exists. Given how much of the cloud-init code base returns Optionals and Unions, I think this will be too common of a problem.

For optionals it usually suffices to toss in an `or <blah>` in the assignment or start the assertion with a check for truthiness.

```python
def parse_json(my_json) -> Optional[Dict]:
    ...

def test_json1():
    x = parse_json('{"a": 1}')
    assert x and x['a'] == 1

def test_json2():
    x = parse_json('{"a": 1}') or {}
    assert x['a'] == 1
```

Neither of these feels particularly burdensome to me.

The unions are more annoying because of [mypy's strict union checking](https://github.com/python/mypy/issues/1693). Some projects deal with this by avoiding Union types altogether.

The workaround that typeshed uses is to use [`object` or `Any` instead of `Union`](https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#using-any-and-object). While this is less beneficial for type-based editor-completion, it avoids the `isinstance()` requirement and allows mypy to still do dynamic type checking. I'm not a fan of this approach because it sacrifices both the quality of type checking and completion - though it has merit especially for a library that has widely consumed types.

I try to avoid `Union` return types when writing new code because `Union`s make reasoning about usage harder - in addition to the mypy-driven boilderplate that results from it. We've made some changes to [commonly used utilities in the cloud-init codebase](https://github.com/canonical/cloud-init/commit/0e90131d57e3a91776172047f6e852f1c9dafe72) for the same reason. Personally I like the little reminder that mypy provides about `Union` types because it feels like encouragement to write (or refactor) code to be easy to understand. I recognize, however, that this is probably not a commonly held perspective.

holmanb:
@aciba90 You've done a fair bit of typing work in cloud-init. I'd be curious to know your thoughts on the topic.

a-dubs:
@holmanb when he's right

⣀⣀⣀⣀⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠄⠀⠀⢀⠀⠀⠀⠀⠀⠐⠀⢢⠆⠐⠁⠀⠀⠀⠀⠇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣶⠀⠀⡀⢀⣤⠀⠀⠐⢂⠀⠀⢈⣀⠀⠀⠀⢀⣠⠖⠋⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⣿⣿⣿⣿⣿⡇⢳⠠⠀⢄⡐⢂⣤⡄⠀⣂⠜⣿⣧⠚⠒⠒⣀⠀⠤⠐⡚⠀⠍⠀⠂⠉⢠⡶⠋⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⣀⣀⣀⣀⣠⣤⣤⣀⣀⣀⣀⣀⣀
⣿⣿⣿⣿⣿⣇⠸⡇⠷⠼⠄⠐⠉⢁⣉⡠⠤⠼⣗⠂⠉⠁⠀⠀⢀⣀⣠⣤⠤⠤⢤⡶⠿⠶⣤⣤⣤⣄⣀⣀⣀⣀⣀⣀⣠⣤⣶⡾⠿⠟⠛⠉⠉⠉⠉⢉⠉⠉⠉⢭⣉⡛⠻⢿⣿
⣿⣿⣿⣿⣿⣿⣀⡧⠤⠔⢒⣊⠉⢤⠂⠀⠀⠉⢢⠀⠀⠀⠀⣿⣿⡟⠉⠀⠀⢠⡞⠀⣴⣶⣿⣶⣭⡻⣿⣿⣿⣿⣿⣿⣿⠟⠁⠀⠀⠀⠀⠀⠀⠀⢀⣀⣁⣀⠀⠀⠀⠙⠈⠀⠙
⣿⣿⣿⣿⣿⣿⡟⠒⠒⠉⣁⣀⣔⡇⠀⠀⠀⠀⠘⡣⡀⠀⠀⠙⣿⠀⠀⠀⢀⡟⠀⣸⢿⣿⣿⣿⣿⣿⣿⣿⡟⠛⠛⢿⣿⠀⠀⠀⠀⣀⣐⢛⣲⣴⣿⣿⣿⣿⣷⠀⠀⠀⠀⠀⠀
⣿⣿⣿⣿⣿⣿⣇⠀⢠⠀⢖⣦⣼⠁⠀⠀⠀⠀⠀⡿⠒⠂⠠⢤⣿⡄⠀⠠⣾⡇⠀⣿⡘⢿⣧⣯⣿⢿⣿⠏⠆⠀⠀⠈⣿⣇⠀⠀⠀⣿⣿⣿⣿⠏⠘⣿⣿⣿⣿⡆⠀⠀⠀⠀⢠
⣿⣿⣿⣿⣿⣿⣿⡟⣻⣧⣸⣿⣿⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠛⣧⠀⣴⣿⡇⠀⠻⠿⢿⣿⡿⠁⣺⡟⠀⠀⠀⠀⠀⢘⣿⣆⢀⡄⣿⠘⠻⣿⣷⣤⣙⣿⣿⣿⡿⠄⣀⢂⣠⣾
⣿⣿⣿⣿⣿⣿⣿⣧⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠀⡙⣷⣿⡟⠁⠀⠀⠀⠀⠋⣠⡾⠋⠀⠀⠀⠀⠀⠀⠀⠈⢿⣮⡝⣿⣿⡇⠈⠛⠛⠛⠛⠛⠋⠐⠀⣸⣶⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⣿⣻⣿⡿⡟⠀⠀⠀⠀⠀⠀⣿⠀⠀⠀⠀⠀⡼⠀⠙⠛⠲⠶⠦⠶⢶⠟⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢝⡿⣿⣿⣿⣧⡄⠀⠀⠀⠀⠀⠀⣀⣠⣾⣿⣿
⣿⣿⣿⣿⣿⣿⣿⣿⡿⠿⠷⣦⣿⠀⠀⠀⠀⠀⠀⠘⠀⠀⠀⠀⣼⠃⠀⠀⠀⠀⠀⠀⢰⠏⠀⢀⣠⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠙⠿⣿⠛⠻⣿⣛⠿⠟⠛⠛⠛⠋⠉⠀⠁
⣿⠟⠋⠉⠙⠻⡟⠁⠀⠀⠀⠀⠙⡆⠀⠀⠀⠀⠀⣸⣤⣤⣤⠤⣇⠀⠀⠀⠀⠀⠀⣰⠃⢀⣴⣿⣿⣧⣤⣤⣶⣦⣀⠀⠳⣄⠀⠀⠀⠀⠀⢹⡆⠀⠈⢻⣿⡄⠀⠀⠀⠀⠀⠀⠀
⡇⠀⠀⠀⠀⠀⠸⠀⠀⠀⠀⠀⠀⠂⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⠈⣀⠀⠀⠀⢀⡼⠁⠀⣾⣿⣿⣿⠛⠿⢿⣿⣿⣿⣷⣤⣼⣄⠀⠀⠀⠀⠈⢿⣆⠀⠀⠹⣧⠀⠀⠀⠀⠀⠀⠀
⢁⠀⠀⠀⠀⠀⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢸⠀⠀⠀⠀⡰⠊⠁⠀⢠⠏⠀⠀⠰⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀⠈⢿⣷⡀⠀⠘⠀⠀⠀⠀⠀⠀⠀
⢸⠀⠀⠀⠀⠀⠀⢇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⠀⠀⠀⢀⠏⠀⠀⠀⠎⠀⠀⠀⠀⢿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡏⠀⠀⠀⠀⠀⠀⠈⢿⣿⡦⠀⠀⠀⠀⠀⠀⠀⠀
⢸⡀⢀⠀⠐⠒⠒⠚⠁⠒⠀⠉⠈⠛⠢⢀⠀⠀⠀⢸⠀⠀⠀⣿⠀⠀⠀⠰⠀⠀⠀⠀⠀⠀⠻⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠛⠁⠀⠀⠀⠀⠀⠀⠀⢠⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀
⠊⠱⣾⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠓⠦⢬⡆⠀⠀⡟⠀⠀⠀⠀⠀⠀⠀⠀⠀⣆⠀⠹⠛⠿⣿⣿⣿⠿⠛⣇⠀⠀⠀⠀⠀⠀⠀⢀⠞⠀⣼⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠙⢷⣄⠀⠀⠀⠀⠀⠁⠐⠀⠀⠀⠀⠀⠀⠀⠀⠙⡀⠀⢧⠀⠀⠀⠀⠀⠀⠀⠀⠀⠸⡄⠐⢠⠤⠿⠛⠳⢧⣌⡹⡀⠀⠀⠀⠀⠀⠀⠈⢲⣤⣿⣿⡿⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠉⠓⠶⠦⢤⣬⣤⣄⠀⠀⠀⠀⠀⠀⠀⠀⠀⡅⠀⢸⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣤⣤⣾⠀⠀⠀⠀⠀⠀⠀⣻⣀⣠⠀⠀⠀⠀⣠⣶⣿⣿⣿⡃⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠚⢿⡳⠀⠀⠀⠀⠀⠀⠀⠀⡏⠀⠘⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠻⣿⣿⣆⠀⠀⣨⠀⠀⢠⣿⣿⠟⠛⠓⢶⣾⣿⣿⣿⣿⡿⠁⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠀⠀⠀⠀⠀⠀⠀⢰⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠙⠻⠿⠿⢿⡿⠿⠿⠛⠁⠀⠀⠀⠘⠿⠟⠛⠉⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡼⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣼⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣠⡺⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀

TheRealFalcon:
Good points @holmanb . I'll close this.

PR #5587, opened 2024-08-06
Fix dailies for Ubuntu/focal

TheRealFalcon:
new upstream snapshot and fix quilt patches for `test_main.py`  and dropping env condition from service files.


debian/patches/drop-unsupported-systemd-condition-environment.patch
```diff
@@ -62,16 +52,16 @@ This patch header follows DEP-3: http://dep.debian.net/deps/dep3/
...
- 
+
  [Service]
  Type=oneshot
---- a/systemd/cloud-init-hotplugd.socket
-+++ b/systemd/cloud-init-hotplugd.socket
-@@ -8,7 +8,6 @@ Description=cloud-init hotplug hook sock
- After=cloud-config.target
+--- a/systemd/cloud-init-main.service.tmpl
++++ b/systemd/cloud-init-main.service.tmpl
```
holmanb, line 59: I don't think that we want to be modifying this service file on focal. I plan on submitting a follow-up PR which will patch it out completely.

TheRealFalcon: Updated

debian/patches/drop-unsupported-systemd-condition-environment.patch
```diff
@@ -25,27 +25,17 @@ This patch header follows DEP-3: http://dep.debian.net/deps/dep3/
...
+@@ -28,7 +28,6 @@ Conflicts=shutdown.target
  RequiresMountsFor=/var/lib/cloud
  ConditionPathExists=!/etc/cloud/cloud-init.disabled
  ConditionKernelCommandLine=!cloud-init=disabled
 -ConditionEnvironment=!KERNEL_CMDLINE=cloud-init=disabled
- 
- [Service]
- Type=oneshot
---- a/systemd/cloud-init.service.tmpl
-+++ b/systemd/cloud-init.service.tmpl
```
holmanb, line 42: @TheRealFalcon This actually drops the patch from the re-named cloud-init.service.tmpl (now named cloud-init-network.service.tmpl). I can add it back in in my followup PR if you want - otherwise we should rename it and I'll rename it back. I'm fine with doing that either way - let me know.

TheRealFalcon: Yeah, we're kind if in a weird "post single process PR" but "pre patch it out" state. I think I'm fine leaving it as-is as you'll need to make changes in your patch either way.

TheRealFalcon: Are we good to merge this PR then?

holmanb: Go for it!

holmanb (approved):

PR #5590, opened 2024-08-07
Add paride to CLA signers

paride:
## Proposed Commit Message

```
chore: add myself (paride) to the CLA signers list

I was already present in tools/.lp-to-git-user, but with my old Launchpad
username. Instead of updating that, I'm just dropping the LP -> GitHub
mapping and adding myself directly to tools/.github-cla-signers.
```

## Additional Context

This is a "bare" change, with no real changes to cloud-init code. The reason for this is that I plan to open a PR against the packaging branch (ubuntu/devel), but it doesn't feel right to update the CLA information there.

## Merge type

### DIFF
--- a/tools/.github-cla-signers
+++ b/tools/.github-cla-signers
@@ -144,6 +144,7 @@ onitake
 orndorffgrant
 Oursin
 outscale-mdr
+paride
 philsphicas
 phsm
 phunyguy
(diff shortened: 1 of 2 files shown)


blackboxsw:
Thanks @paride, but there is no real need to shift to .github-cla-signers as our workflow checks both files so you won't be prompted. I think at this point it may make sense for us to just retire the .lp-to-git-user file and pull github usernames into .github-cla-signers. We'll still have this historic commit referencing former LP users that transitioned their CLA signing to github. But I don't expect we will need to use that nowadays. Let's close this PR and we can discuss dropping this .lp-to-git-user file, workflow related functionality and the tools/migrate-lp-user-to-github as a separate pr.

